PR #2415, opened 2019-10-12
Allow specifying start location for epp:parse_file/2 and open/1 

gomoripeti:
The primary goal is to get column numbers by setting start_location to {1,1}.

Specifying start location is already supported internally by epp via the
undocumented open/5 function (used by escript).

If it is not desirable to allow any start location, an alternative would
be a `{column, boolean()}` option.

(TODO: update docs)

### DIFF
--- a/lib/stdlib/src/epp.erl
+++ b/lib/stdlib/src/epp.erl
@@ -110,14 +110,16 @@ open(Name, Path, Pdm) ->
     internal_open([{name, Name}, {includes, Path}, {macros, Pdm}], #epp{}).
 
 open(Name, File, StartLocation, Path, Pdm) ->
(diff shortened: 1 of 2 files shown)


zuiderkwast:
For the record, `merl:quote/2` officially handles `{1,1}` as start location.

```
1> merl:quote({1,1}, "hello,\n world").
[{atom,{1,1},hello},{atom,{2,2},world}]
```

jhogberg:
Thanks for the PR, and sorry for keeping it in limbo for so long.

We've decided to do this ourselves in the near future as this is a bit trickier than just adding column numbers. Support in the rest of the compiler is rather fragmented, and these annotations are sometimes indirectly documented or used as line numbers, such as `return_warnings` in the compiler or parse transformations.

PR #2417, opened 2019-10-12
rpc do not hide throw class failures

mikpe:
Factor out duplicated code for evaluating user-specified function
call to new helper function.  Replace old-style catch with try-catch,
and map thrown exceptions to badrpcs.

Adjust rpc_SUITE to match the new behaviour.

Fixes ERL-1066.

### DIFF
--- a/lib/kernel/doc/src/rpc.xml
+++ b/lib/kernel/doc/src/rpc.xml
@@ -134,6 +134,12 @@
           <anno>Args</anno>)</c> on node <c><anno>Node</anno></c> and returns
           the corresponding value <c><anno>Res</anno></c>, or
           <c>{badrpc, <anno>Reason</anno>}</c> if the call fails.</p>
+        <p>A call that evaluates to <c>{'EXIT', _} = <anno>Reason</anno></c> is
+          considered to have failed, causing <c>{badrpc, <anno>Reason</anno>}</c>
+          to be returned.</p>
+        <p>Prior to OTP-23, if a call failed by throwing <c><anno>Reason</anno></c>,
+          <c><anno>Reason</anno></c> would be returned. In OTP-23 and later,
+          <c>{badrpc, <anno>Reason</anno>}</c> is returned.</p>
       </desc>
     </func>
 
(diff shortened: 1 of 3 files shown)


bjorng:
It looks fine to me, but we can't accept this pull request for `maint`. It could break code if someone happened to depend on the old behavior. I have changed the base to `master`.

Also, I think that changed behavior should be documented. That is, there should be a note in the documentation mentioning how throws were handled before OTP 23, and how they are handled in OTP 23 and later.

Please combine the changes into a single commit.

mikpe:
Rebased on master, added note to documentation, and squashed.

bjorng:
We noticed three failed test cases in our daily builds:

```
global_SUITE:simple_resolve/1
global_SUITE:simple_resolve2/1
mnesia_durability_test:master_on_non_local_tables/1
```
To make run the mnesia test case without running the entire mnesia test suite, you must do like this:

```
cd $ERL_TOP/lib/mnesia/test
make
PATH=$ERL_TOP/bin:$PATH erl -sname a
```
In the Erlang shell, run the test like this:

```
mt:t(durability, master_on_non_local_tables).
```

bjorng:
By the way, we will need to discuss this pull request on the OTP Technical Board.

mikpe:
Fixed.  All three failures were due to rpc:call turning _normal_ `{'EXIT',_}` returns into badrpcs, which is obvious when one looks at the old code, but not implied by the documentation for rpc:call.

bjorng:
Now the test cases succeed in our daily builds.

I think it is time to document that `'EXIT'` tuples are converted into badrpcs. Can you do that? Also, squash everything into one commit.

As a preparation for the OTP Technical Board meeting, it would be useful if you could rewrite the commit message to make it clear what you have changed (with examples) and why and give some more details about your use case. I think that the original intention of `throw` was that there should always be a corresponding `catch` and that it should never leak out through a public API. That intention has probably never been officially documented, though, so we are worried that some of our customers can have used `throw` in unexpected ways and now depend on the current behavior of `rpc:call()`.

mikpe:
Extended the documentation, squashed, and rewrote the commit message, as instructed.

bjorng:
OTP Technical Board decided that it is too risky to change the behavior of the `rpc:call()` functions. There could be code that depends on the old behavior without realizing it.

If you'll need a fix as part of OTP, the way to go is to add a new function. OTB is much more likely to approve a well-thought new function.

PR #2424, opened 2019-10-19
eunit: Allow test cases to retrieve output captured by EUnit

zuiderkwast:
To be able to test programs or functions producing output on
standard output and to verify this output in an EUnit test
case, a new macro ?capturedOutput is defined in eunit.hrl.


IngelaAndin:
I tried to run your branch in our tests and it results in the following dialyzer failiure that needs to be fixed. 

eunit_tests.erl:49: Function get_output_test/0 has no local return

[2019-10-22T13:44:43.046Z] eunit_tests.erl:53: The created fun has no local return

zuiderkwast:
Fixed. Sorry for not checking properly before!

PR #2428, opened 2019-10-23
erts: Fix so that quotes are respected in erlexec args files

garazdawi:
Before this fix any double space ('a  b') would be stripped to a single space even if quoted.

### DIFF
--- a/erts/etc/common/erlexec.c
+++ b/erts/etc/common/erlexec.c
@@ -1716,6 +1716,13 @@ static char **build_args_from_string(char *string, int allow_comments)
 	    case '\0':
 		state = Start;
 		break;
+            case '#':
+                if (allow_comments) {
+                    ++p;
+                    state = BuildComment;
+                    break;
+                }
+                /* fall-through */
 	    default:
 		state = Build;
 		break;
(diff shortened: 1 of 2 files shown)


okeuday:
This fixes [ERL-1051](https://bugs.erlang.org/browse/ERL-1051).  Works with " quotes as expected.

okeuday:
@garazdawi After testing with this change merged into master @ https://github.com/erlang/otp/commit/0dc1a5522657d14a95954af902611748216c7a82 (locally) I am unable to start a release with more complex command-line usage (I confirmed that things work fine without the current change here).  The error I get is:

    {"could not start kernel pid",application_controller,"error in config file \"./.config\" (none): configuration file not found"}

So, I am not sure what the problem with the change here is, but it doesn't appear to be ready to merge.  My command-line (for executing a build of https://github.com/CloudI/CloudI/) is similar to:

    /.../install_test1/lib/cloudi-1.7.5/erts-10.5.3/bin/erlexec -boot /.../install_test1/lib/cloudi-1.7.5/releases/1/cloudi -config /.../install_test1/lib/cloudi-1.7.5/etc/vm.config -args_file /.../install_test1/lib/cloudi-1.7.5/etc/vm.args -mode embedded -shutdown_time 65000 -- console start

garazdawi:
Should be fixed now, please test it again.

okeuday:
@garazdawi The change works fine now.  Thanks for the fix!

PR #2430, opened 2019-10-23
edoc: support multiple clauses in -spec

attah:
Putting this up again...
I'm getting nowhere fast in cleaning this up (and starting to question how much that is needed).
As we all know, the best way to know what the proper solution would be is being wrong on the internet.

Purpose: render specs with multiple clauses in edoc.

### DIFF
--- a/lib/edoc/src/edoc_data.erl
+++ b/lib/edoc/src/edoc_data.erl
@@ -230,9 +230,11 @@ callback({N, A}, _Env, _Opts) ->
 %% <!ELEMENT throws (type, localdef*)>
 %% <!ELEMENT equiv (expr, see?)>
 %% <!ELEMENT expr (#PCDATA)>
-
-function({N, A}, As, Export, Ts, Env, Opts) ->
-    {Args, Ret, Spec} = signature(Ts, As, Env),
+function({N, A}, []=As, Export, Ts, Env, Opts)->
+    function({N, A}, [As], Export, Ts, Env, Opts);
+function({N, A}, [HAs | _]=As, Export, Ts, Env, Opts) when not is_list(HAs) ->
+    function({N, A}, [As], Export, Ts, Env, Opts);
+function({N, A}, As0, Export, Ts, Env, Opts) ->
     {function, [{name, atom_to_list(N)},
 		{arity, integer_to_list(A)},
       		{exported, case Export of
@@ -240,13 +242,8 @@ function({N, A}, As, Export, Ts, Env, Opts) ->
 			       false -> "no"
 			   end},
 		{label, edoc_refs:to_label(edoc_refs:function(N, A))}],
-     [{args, [{arg, [{argName, [atom_to_list(A)]}] ++ description(D)}
-	      || {A, D} <- Args]}]
-     ++ Spec
-     ++ case Ret of
-	    [] -> [];
-	    _ -> [{returns, description(Ret)}]
-	end
+     lists:append([get_args(lists:nth(Clause, As0), Ts, Clause, Env)
+                   || Clause <- lists:seq(1, length(As0))])
      ++ get_throws(Ts, Env)
      ++ get_equiv(Ts, Env)
      ++ get_doc(Ts)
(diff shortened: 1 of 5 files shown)


uabboli:
Hi,

> I'm getting nowhere fast in cleaning this up (and starting to question
> how much that is needed).

I'm not sure how to proceed. Do you want to withdraw the PR?

I looked into the code and in general it looks like you're getting somewhere. :-)

A few things though:

If you remove spaces at the end of lines, do it in a separate commit. (Or don't do it at all.)

In edoc_specs:specs() the line
   P = P0#parms{line = L},
is duplicated.

The indentation in edoc_layout:function() is wrong.

docgen_edoc_xml_cb:t_utype() is not correct. Using
lists:append() instead of lists:merge() is an improvement, but more
work is needed to make it look OK.

Best regards,

Hans Bolinder, Erlang/OTP team, Ericsson

attah:
I'm not sure how i want to proceed either, so i am basically fishing for comments (much like yours), or suggestions/contributions.

I have addressed your comments. Thanks.
Well, all but docgen_edoc_xml_cb:t_utype() not being correct, do you have an example of what it fails at?

Hmm, seems i managed to break the docbuild... it succeeded on my machine just moments before.
Does it merge first or something?

Side note:
It is an uphill battle trying to use modern editors on legacy code, not automatically removing trailing newlines is behind a plugin and a tab size that is not the same as an indentation step is just not configurable. I really wish these sort of things made it in to the languages themselves and did not exist in just one or tow editors.

attah:
Seems i failed to get the change to append to stick, and that it is indeed merging with master that gives the error seen in Travis. Ugh.

uabboli:
> Well, all but docgen_edoc_xml_cb:t_utype() not being correct, do you
> have an example of what it fails at?

I chose some function, edoc_extract:header/4 say, and
wrote a -spec:
```
-spec header(Forms,
             File :: file:filename(), edoc_lib:edoc_env(),
             Options :: proplists:property()) ->
                    {ok, Tags} | {error, Reason} when
      Forms :: erl_syntax:syntaxTree() | [erl_syntax:syntaxTree()],
      Tags :: [term()],
      Reason :: term()
                ; (A, B, C, D) -> ok when
                    A :: atom(),
                    B :: integer(),
                    C :: pid(),
                    D :: term().
```

The resulting string is a concatenation of the two "clauses" with no
line break between them.

Is the order of the "clauses" reversed, BTW?

attah:
> Is the order of the "clauses" reversed, BTW?

Yes, apparently... it was another case of me using merge instead of join.
I have now fixed all of those.

> The resulting string is a concatenation of the two "clauses" with no line break between them.

I am unable to reproduce this. The edoc looks just fine, with headers and parameters in their own paragraph tags. If you by "string" mean something else, i'm not getting it. This problem does however sound like a lists:merge "sorting away" sags or newlines (which should now be fixed), but that is pure speculation on my part.

uabboli:
> > The resulting string is a concatenation of the two "clauses" with no
> > line break between them.

> I am unable to reproduce this. The edoc looks just fine, with headers
> and parameters in their own paragraph tags. If you by "string" mean
> something else, i'm not getting it. This problem does however sound
> like a lists:merge "sorting away" sags or newlines (which should now
> be fixed), but that is pure speculation on my part.

If I change edoc_extract.erl as above, and run 'make release_tests edoc'
on $ERL_TOP with your latest updates, I see:
```
    header(Forms, File::file:filename(), Env::edoc_lib:edoc_env(), Options::proplists:property()) -> {ok, Tags} | {error, Reason}(A, B, C, D) -> ok
```

"(A, B, " follows "Reason}" without newline or any other separator.

Maybe you're talking about the output from edoc:file()? It looks OK,
I think.

attah:
> Maybe you're talking about the output from edoc:file()?

Yes. What are you talking about? Where do i observe said string?

attah:
@uabboli I had a senior colleague look at what you said, but neither of us understand where you are looking. Could you please elaborate?

attah:
I got some help on where to look. I'll look in to it.

uabboli:
For the moment the priority is low on this issue. I'll have a look as soon as time permits.

attah:
The latest iteration takes care of the previously mentioned problem of no newlines between header clauses.
The failed test is just a timeout on the s390x.

uabboli:
I think it looks good. Thanks! I've added the branch to our daily testing.

uabboli:
Thanks again. The ticket number is OTP-16407.

PR #2431, opened 2019-10-24
stdlib: Add 'async_call' and 'yield' to generic behaviors

dgud:
Simplify and encourage users to do more async work, the usage
pattern is already available in 'rpc' module and similar usages
are available in other languages and standards.

Async calls can be implemented via cast or regular messages,
but then the user need to implement it both in client and server.
In this implementation the server does not need to know that the
client are making async calls.

yield/[1|2] returns {reply, Reply} instead of {ok, Reply} since Reply
may be positive or negative answer i.e.:
{reply, {ok, Value}} and {reply, {error, Reason}} looks better than
{ok, {error, Reason}} or {ok, {ok, Value}}.

And we need to encapsulate the return value to
differ between client timeouts and server response which may be
the atom timeout, e.g. timeout vs {reply, timeout}

We don't want do exit(timeout) since then you can't do non_blocking yields
without catching the call to yield(Promise, 0).

yield_msg/2 return {reply, Reply} or false (change negative result?)

yield_msg(Msg, Promise) is introduced to be able handle concurrent,
async_calls in receive loop or in handle_info in another gen_server.

Promise = gen:async_call(..),
...

dispatch(State) ->
receive
  Msg -> case gen:yield_msg(Msg, Promise) of
            {reply, Reply} -> {Reply, State};
	    false ->
	    	  State0 = handle_msg(Msg, State0),
		  dispatch(State0)
         end
end,

Both yield and yield_msg leaves the monitor until a proper response
is received or the moniter fires. To be able to poll for response.


lib/stdlib/doc/src/gen_event.xml
```diff
@@ -240,6 +240,45 @@ gen_event:stop             ----->  Module:terminate/2
...
+        <v>&nbsp;GlobalName = ViaName = term()</v>
+	<v>Handler = Module | {Module,Id}</v>
+        <v>&nbsp;Module = atom()</v>
+        <v>&nbsp;Id = term()</v>
+        <v>Request = term()</v>
+        <v>Promise = term()</v>
+      </type>
+      <desc>
+        <p>
+	  Makes a synchronous call to event handler <c>Handler</c>
```
essen: synchronous?

essen:
Sounds good. We do lose the mailbox optimization when the reference was created just before the receive though, don't we? If so it's probably worth mentioning somewhere.

`yield_msg` is a weird name. Maybe something like `async_reply/2`?

ferd:
I think `yield_msg` is probably the most confusing one if you go straight from the docs. As you mention in the PR description here, the objective is to be able to match a message _already received out of the mailbox_ and know if it is associated to a specific promise that was already generated and held in state, and this is mostly useful when you have your own receive loop with mixed messages (which is what using it from `handle_info/3` accomplishes as well), but that context is missing from the actual documentation. Adding it would probably make it clearer.

I otherwise like that it's replicating the `gen_server:call` semantics in terms of exiting when the other process is down and handling monitors in a deferred manner.

lib/stdlib/doc/src/gen_event.xml
dgud: Of course

dgud:
async_call and yield is the same as in rpc module so I kept that, and is used in other languages.

I wanted yield_msg named something like yield so that it is obvious that they are variants of
the same thing but I'm open for suggestions, naming stuff.
The docs could be made more explicit about what the function is intended for, users might not read the commit comments :-)

lib/stdlib/src/gen_server.erl
```diff
@@ -223,6 +234,38 @@ call(Name, Request, Timeout) ->
...
+-spec async_call(Name::server_ref(), Request::term()) -> promise().
+async_call(Name, Request) ->
+    try gen:async_call(Name, '$gen_call', Request)
+    catch _:Reason ->
+            erlang:error({Reason, {?MODULE, ?FUNCTION_NAME, [Name, Request]}})
+    end.
+
+-spec yield(Promise::promise()) -> {reply, Reply::term()}.
+yield(Promise) ->
+    try gen:yield(Promise, infinity)
```
josevalim: Should this reflect the other timeouts in the GenServer module? I.e. 5000?

dgud: Good question.
I don't like the 5s timeout but now it's there.
Opinions?

josevalim:
For completeness, I would like to point out there is some potential for confusion with the name `async_call`. `call` in `gen_server` is defined as "Makes a synchronous call to the ServerRef of the gen_server process and waiting ..." and `cast` is defined as "Sends an asynchronous request to the ServerRef of the gen_server process". Adding `async_call` blurries those lines.

The other issue with `async_call` is that "async" often means fire-and-forget, so I am worried people will use `async_call` with said purpose and it will leave stray messages in the inbox.

Part of the issue is that "call is sync" and "cast is async" has been drilled down in many places. For example, both "Erlang and OTP in Action" and "Designing for Scalability" mention something like "handle_call/3 - callback for synchronous requests" and that becomes slightly confusing with the presence of `async_call`.

To be clear, I am arguing against `async_call` because books would be slightly inaccurate, I just want to make sure the points above have been considered.

I know naming is hard, so I would like to suggest at least one option: `yieldable_call` or similar. It reduces the confusion that would arise with "async" and it also indicates that the call must be yielded. I am aware the current naming mirrors `rpc`, but given this API is being added to multiple modules, if you believe there is a better name, it makes sense to adopt it and make `rpc` mirror it.

> I wanted yield_msg named something like yield so that it is obvious that they are variants of
the same thing but I'm open for suggestions, naming stuff.

To me the biggest confusion in regards to `yield_msg` is that at first glance I expected it to cause some effect on the inbox, similar to yield. However, it is just a check if something has already happened. So maybe it should be in the past, such as `yielded_message`, `was_yielded`, or similar?

Also, I am not convinced `yield_msg` should exit. I often use something like `yield_msg` inside a `gen_server` and most of the times I want to act on the DOWN message explicitly rather than let it crash. Having an API that doesn't raise also allows us to implement things such as "at least one of these N calls must succeed". Therefore I would also propose the return type to be `{reply, Reply} | {exit, Reason} | false`. Callers can always match exit and explicitly `exit` if they want to.

lib/stdlib/src/gen_server.erl
josevalim: Yeah, I hear you. :/

I am worried someone may start with `call`, then move to `async_call`+`yield` as some optimization but by doing so they have changed the meaning of the code in an unexpected way.

IngelaAndin:
maybe promise_call and yield_promise would be better names?

lib/stdlib/src/gen_server.erl
garazdawi: I think I agree with José. All other timeouts are 5s, then this should be as well.

garazdawi: though, really if what would be the same thing would be if async_call would setup a timer for 5 seconds when it is called and then yield/yield_message would timeout when they receive the timer. Don't think we want to have that though.

dgud:
Pushed some doc fixes, have not taken all comments into consideration yet.
The comments and reviewing it is appreciated.

lib/stdlib/src/gen_server.erl
IngelaAndin: I think that the "async_call" leaves all timeout logic up the caller, and there should be no timeout option.

dgud: Timer would not be set in `async_call` but in `yield/2` which contains the `receive`

essen: The timer argument lets the user do something like getting the current time `StartTime`, do the async call, do something else and then at the end call `{ok, Result} = yield(Promise, max(0, Timeout + CurrentTime - StartTime))`. Only it would `{badarg,timeout}` instead of `exit(timeout)` but that's probably not a big deal?

essen:
What happens if I do a lot of `async_call` at once? Right now it creates an `MRef` for every call, is that likely to cause a performance problem? If it would, can it be mitigated? We would probably need more than the monitor to differentiate between replies.

> Part of the issue is that "call is sync" and "cast is async" has been drilled down in many places. For example, both "Erlang and OTP in Action" and "Designing for Scalability" mention something like "handle_call/3 - callback for synchronous requests" and that becomes slightly confusing with the presence of `async_call`.

It's a little misleading to say that `call` is synchronous, as the reply may be deferred by the other side. I get it though, it *feels* synchronous from the caller's side.

I think a better description would be that `call` is a blocking call and `async_call` is a non-blocking call (not saying the function should be renamed, it's fine). `cast` is something different and with the addition of `async_call` there would be no need to use `cast` to simulate a non-blocking call anymore so `cast` would be clearly distinct from calls. A `cast` has no reply and is therefore not a `call`, it's more like an explicit `!` message that's part of the interface to speak to the process.

The old book will remain confusing, perhaps even more so than they are now, but the distinction should be a lot clearer to the user.

josevalim:
> What happens if I do a lot of async_call at once? Right now it creates an MRef for every call, is that likely to cause a performance problem? 

I don't think it would be a performance problem because every `call` today already creates a ref?

> I think a better description would be that call is a blocking call and async_call is a non-blocking call (not saying the function should be renamed, it's fine).

Even though you *didn't* propose a name change, imo `non_blocking_call` is clearer than `async_call` and less likely to cause confusion. :) I know everyone will have different opinions and preferences in regards to the name though.

essen:
> > What happens if I do a lot of async_call at once? Right now it creates an MRef for every call, is that likely to cause a performance problem?
> 
> I don't think it would be a performance problem because every `call` today already creates a ref?

Right but it's not the same, I can't create many references at once with `call` because the function is blocking. I can only do that with `async_call`.

``` erlang
%% This creates 1000 monitors but only one exists at a time.
[call(Pid, hello) || _ <- lists:seq(1,1000)]

%% This creates 1000 monitors and up to 1000 will exist at the same time.
[async_call(Pid, hello) || _ <- lists:seq(1,1000)]
```

I therefore wonder if this situation could lead to degradation (beyond the fact we're sending so many messages all at once) or to unusual situations. For example if I do these 1000 monitors and the process dies, then I will receive up to 1000 `'DOWN'` messages back. There's probably other considerations worth figuring out.

dgud:
We decided to change the function names and divert from rpc's naming:

> async_call => send_request
> yield => wait_response
> yield_msg => check_response

Will update the PR (eventually)

dgud:
Changed the names, and returns errors instead doing exit.
Removed default timer value for now in gen_server and gen_event.

PR #2433, opened 2019-10-28
Fix typo

benbro:


benbro:
The same typo repeats several times but I only fixed it in one place.

bmk:
Hi,

Three people have now looked at the (socket) documentation
without finding any more 'to that should be too'.
So, could you please point the one you found.

Regards,
	/BMK

benbro:
Look for:
if the buffer size is **to** small, the message will be truncated

PR #2434, opened 2019-10-28
Fix unsafe optimization of receives

bjorng:
Fix a problem where a receive marker would be set even if it
was not guaranteed that execution would reach a receive statement.

This optimization bug caused the problem for the `socket` module
reported here:

http://erlang.org/pipermail/erlang-questions/2019-October/098591.html

https://bugs.erlang.org/browse/ERL-1076


garazdawi:
Should the `beam_validator` be modified to catch these types of faults?

PR #2435, opened 2019-10-28
Add function/arity to shadow clauses warnings when possible

josevalim:
When functions can be generated dynamically, we may
have multiple functions defined in the same line, which
makes the shadow clause warning a bit hard to debug.

This patch addresses this by using the a function
annotation, if available, and including that in the
printed message.

### DIFF
--- a/lib/compiler/src/sys_core_fold.erl
+++ b/lib/compiler/src/sys_core_fold.erl
@@ -380,11 +380,11 @@ expr(#c_case{}=Case0, Ctxt, Sub) ->
     %%   according to the rules above).
     %%
     case opt_bool_case(Case0, Sub) of
-	#c_case{arg=Arg0,clauses=Cs0}=Case1 ->
+	#c_case{anno=Anno,arg=Arg0,clauses=Cs0}=Case1 ->
 	    Arg1 = body(Arg0, value, Sub),
 	    LitExpr = cerl:is_literal(Arg1),
 	    {Arg2,Cs1} = case_opt(Arg1, Cs0, Sub),
-	    Cs2 = clauses(Arg2, Cs1, Ctxt, Sub, LitExpr),
+	    Cs2 = clauses(Arg2, Cs1, Ctxt, Sub, LitExpr, Anno),
 	    Case = Case1#c_case{arg=Arg2,clauses=Cs2},
 	    warn_no_clause_match(Case1, Case),
 	    Expr = eval_case(Case, Sub),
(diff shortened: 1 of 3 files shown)


josevalim:
I am not sure if this is the best approach (relying on annotation), but by the time we reach `sys_core_fold`, it is no longer clear if the source of the case is a function clause or not, so we have to pass this information somehow.

lib/compiler/src/sys_core_fold.erl
```diff
@@ -1456,19 +1456,22 @@ clauses(E, [C0|Cs], Ctxt, Sub, LitExpr) ->
...
+shadow_warning([C|Cs], none, Anno) ->
     add_warning(C, nomatch_shadow),
-    shadow_warning(Cs, none);
-shadow_warning([C|Cs], Line) ->
-    add_warning(C, {nomatch_shadow, Line}),
-    shadow_warning(Cs, Line);
-shadow_warning([], _) -> ok.
+    shadow_warning(Cs, none, Anno);
+shadow_warning([C|Cs], Line, Anno) ->
+    case lists:keyfind('fun', 1, Anno) of
```
bjorng: Omit the `lists:` prefix (this a convention in the compiler source code).

lib/compiler/src/sys_core_fold.erl
```diff
@@ -1456,19 +1456,22 @@ clauses(E, [C0|Cs], Ctxt, Sub, LitExpr) ->
...
+	{'fun', {Name, Arity}} -> add_warning(C, {nomatch_shadow, Line, {Name, Arity}});
```
bjorng: Long line. Please break it after the `->`. I would break the next line too in the same way.

lib/compiler/src/sys_core_fold.erl
```diff
@@ -2901,6 +2904,10 @@ format_error({embedded_unit,Unit,Size}) ->
 format_error(bad_unicode) ->
     "binary construction will fail with a 'badarg' exception "
 	"(invalid Unicode code point in a utf8/utf16/utf32 segment)";
+format_error({nomatch_shadow,Line,{Name, Arity}}) ->
+    M = io_lib:format("this clause for ~ts/~B cannot match because a previous clause at line ~p "
```
bjorng: Long line. Please break it.

josevalim:
Thanks for the positive review. All feedback address.

bjorng:

PR #2436, opened 2019-10-29
Really fix ftp loop

richcarl:
The previous fix in 9a1c36c10a was not sufficient, and the code was still looping, eventually consuming all available memory. This new patch has been verified in production running many thousands of ftp transfers with no problems.


richcarl:
This patch is in 21 and master but never made it to maint-22, so it's not yet been included in any OTP 22 release.

HansN:
It is available also in maint (as all maint-* patches) and will therefor be released  in OTP-22.2, planned to be released 2019-12-11.

PR #2437, opened 2019-10-29
httpc: do not pass error tuple to uri_string:parse/1

mikpe:
Fixes crashing part of ERL-1083.


IngelaAndin:
Looks good and does not break anything. Would you please add a regression test case and then we will merge this.

mikpe:
Added test case, verified it works with the fix and fails without the fix.

IngelaAndin:
Thank you, running last verification round in our nightly builds now.

PR #2438, opened 2019-10-29
erl_child_setup: reduce number of calls to close()

jeckersb:
On systems without closefrom(), such as Linux, iterating over all
possible file descriptors and calling close() for each is inefficient.
This is markedly so when the maximum number of file descriptors has
been tuned to a large number.

Instead, walk the open descriptors under /dev/fd and close only those
which are open.

### DIFF
--- a/erts/emulator/sys/unix/erl_child_setup.c
+++ b/erts/emulator/sys/unix/erl_child_setup.c
@@ -418,6 +418,7 @@ main(int argc, char *argv[])
     int uds_fd = 3, max_fd = 3;
 #ifndef HAVE_CLOSEFROM
     int i;
+    DIR *dir;
 #endif
     struct sigaction sa;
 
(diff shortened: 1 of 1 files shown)


garazdawi:
Hello! Thanks for your contribution!

In what scenario did you find that this was a problem?

jeckersb:
> Hello! Thanks for your contribution!
> 
> In what scenario did you find that this was a problem?

We have rabbitmq running inside of a container with the fd limit set to 65536 (lots of client connections), and health checks inside are regularly running rabbitmqctl which launches lots of short-lived erlang VM instances.

This issue was spotted while doing some system-wide profiling and noticing that an unusually large amount of time was spent in the kernel handling the close() syscall along with the code to context switch back and forth.

garazdawi:
> We have rabbitmq running inside of a container with the fd limit set to 65536 (lots of client connections), and health checks inside are regularly running rabbitmqctl which launches lots of short-lived erlang VM instances.

That makes sense. Thanks for the explanation.

PR #2439, opened 2019-10-29
Fix sigq message 

garazdawi:
Sister bugfix to #2434, see commit messages for more info.

This change turned into something quite large, so maybe only 
765fa32 will be released as a patch on maint-21 and maint-22 and the full fix will be part of maint and master.


garazdawi:
The small fix will be released in OTP 21 and OTP 22 patches, the larger patch set has been merged to maint to release in the next maintenance release.

PR #2440, opened 2019-10-29
[ERL-1078] Validate auth tag when using chacha20_poly1305 cipher

mtrudel:
Adds support for validating auth tag when using `chacha20_poly1305` cipher with `crypto_one_time_aead`

Details at https://bugs.erlang.org/browse/ERL-1078


HansN:
Thanks for the PR and fault report! However, we already had a fix "in the pipe" also including a test case, so therefor I close yours. 
But you are most welcome with other PRs.

PR #2441, opened 2019-10-29
Do not merge maps when the value is the same

josevalim:
If merging two maps that are precisely the same
in memory, we can avoid the whole merge operation
and simply return the given map.

This can reduce memory allocation and CPU work for
algorithms that may be updating a map and merging
into a previous value. The check is also very cheap,
which means the downsides are minimal.

The commit also adds coverage for a similar behaviour
already implemented for maps:put/3.


sverker:
What about `maps:merge(NotAMap, NotAMap)`?

josevalim:
@sverker oh! good catch! :D I will push a fix later.

josevalim:
@sverker fixed!

josevalim:
@sverker I have also noticed we can apply a similar optimization of one the maps is empty. Should I update this PR or should I send a separate one?

josevalim:
@sverker the PR has been updated to also check for empty maps.

PR #2442, opened 2019-10-30
Allow ignoring log message from function

hauleth:
This introduces ~~2~~ 1 change:

- `msg_fun/0` can now return `ignore` to ignore logging message at all
- ~~`Metadata` argument can accept 0-ary function that is supposed to
  return computed map with metadata~~

### DIFF
--- a/lib/kernel/src/logger.erl
+++ b/lib/kernel/src/logger.erl
@@ -88,7 +88,8 @@
                               single_line := boolean()}.
 -type msg_fun() :: fun((term()) -> {io:format(),[term()]} |
                                    report() |
-                                   unicode:chardata()).
+                                   unicode:chardata() |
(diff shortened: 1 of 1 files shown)


josevalim:
Having the metadata be a separate function is not that useful because often you will compute the log message and the metadata at the same time (if they are expensive) and if they are separate functions it means you have to do the work twice, which is not as beneficial.

hauleth:
@josevalim removed that change and keept only the change about being able to return `ignore` from `msg_fun`, so it is possible to ignore message from within it.

josevalim:
This changes looks simple enough now. :+1: Should it also include a test?

garazdawi:
I'm closing this PR in favor of #2721

PR #2443, opened 2019-11-01
Allow linewidth and indent to be given to erl_pp

josevalim:
This gives a bit of flexibility and control to the
formatted code. For example, Erlang official docs
uses 3 spaces when indenting types, with this patch
we can reproduce this effect elsewhere.


josevalim:
Although the changes look large, they are quite straight-forward. I replaced the users of `?MAXLINE` by `PP#pp.linewidth` and `?IND` by `PP#pp.indent`. Nothing else was changed.

Also note I picked the name `linewidth` to mirror `erl_prettypr`.

uabboli:
Thanks, looks good.

PR #2444, opened 2019-11-01
According to spec format can be io:format/0

hauleth:
Earlier it accepted only charlist which could be confusing for users that would expect it to accept the same types as `io:fwrite/2`.


garazdawi:
Shouldn't it also allow binaries if the spec is to be followed? Also there needs to be tests for this.

hauleth:
It checks for string as well, `IS_STRING` already does that.

garazdawi:
> It checks for string as well, IS_STRING already does that.

Ah, yes. Missed that part.

hauleth:
@garazdawi added tests.

PR #2445, opened 2019-11-01
fix tcp fragment timeout in DIAMETER

RoadRunnr:
Two consecutive messages with the just the right combination of
fragmentation and delay would cause the fragment_timeout to trigger
on the wrong (second) fragment.

### DIFF
--- a/lib/diameter/src/transport/diameter_tcp.erl
+++ b/lib/diameter/src/transport/diameter_tcp.erl
@@ -657,6 +657,10 @@ transition({diameter, {close, Pid}}, #transport{parent = Pid,
 transition({timeout, TRef, flush}, #transport{tref = TRef} = S) ->
     flush(S#transport{tref = false});
 
+transition({timeout, _, flush}, S) ->
+    %% race between of stop_timer and already queued message
+    S;
+
 %% Request for the local port number.
 transition({resolve_port, Pid}, #transport{socket = Sock,
                                            module = M})
@@ -726,7 +730,7 @@ tls(accept, Sock, Opts) ->
 %% Receive packets until a full message is received,
 
 recv({Msg, Rest}, S) ->  %% have a complete message ...
-    recv(acc(Rest), message(recv, Msg, S));
+    recv(acc(Rest), message(recv, Msg, stop_fragment_timer(S)));
 
 recv(Frag, #transport{recv = B,
                       socket = Sock,
@@ -833,6 +837,18 @@ start_fragment_timer(#transport{frag = B, tref = TRef} = S)
 
 start_fragment_timer(#transport{timeout = Tmo} = S) ->
     S#transport{tref = erlang:start_timer(Tmo, self(), flush)}.
+
+%% stop_fragment_timer/1
+%%
+%% Stop a timer only if there's one running
+
+stop_fragment_timer(#transport{tref = TRef} = S)
+  when is_reference(TRef) ->
+    erlang:cancel_timer(TRef, [{async, true}, {info, false}]),
+    S#transport{tref = false, flush = false};
+
+stop_fragment_timer(S) ->
+    S.
 
 %% accept/2
 
--- a/lib/diameter/test/diameter_transport_SUITE.erl
+++ b/lib/diameter/test/diameter_transport_SUITE.erl
@@ -38,6 +38,7 @@
          sctp_accept/1,
          sctp_connect/1,
          reconnect/1, reconnect/0,
+         tcp_fragment/1,
          stop/1]).
 
 -export([accept/1,
@@ -102,7 +103,8 @@ tc() ->
      tcp_connect,
      sctp_accept,
      sctp_connect,
-     reconnect].
+     reconnect,
+     tcp_fragment].
 
 init_per_suite(Config) ->
     [{sctp, ?util:have_sctp()} | Config].
@@ -171,6 +173,18 @@ sctp_connect(Config) ->
 connect(Prot) ->
     T = {Prot, make_ref()},
     [] = ?util:run([{?MODULE, [init, X, T]} || X <- [gen_accept, connect]]).
+
+%% ===========================================================================
+%% tcp_fragment/1
+%%
+%% Test TCP fragment timeout.
+
+tcp_fragment(_) ->
+    fragment(tcp).
+
+fragment(Prot) ->
+    T = {Prot, make_ref()},
+    [] = ?util:run([{?MODULE, [init, X, T]} || X <- [accept_recv2, gen_fragment]]).
 
 %% ===========================================================================
 %% reconnect/1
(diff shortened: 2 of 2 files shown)


otp-anders:
Hi Andreas.

Sorry I haven't responded to this earlier, but I seem to have missed your erlang-questions post when I was on vacation, and the pull request because the notification didn't end up where I thought it would. :)

I will take a look, but it would be good if you could describe the "right combination of fragmentation and delay" for starters. (Or point me at where you've already described it.)

Anders

RoadRunnr:
> I will take a look, but it would be good if you could describe the "right combination of fragmentation and delay" for starters. (Or point me at where you've already described it.)

I'm, not sure if I can put it into a comprehensible form, but let me try

Lets assume a fragment timeout of 1000 ms.

It requires two diameter messages that are fragmented at the TCP layer and a slow enough connection (or packet loss).

- t + 0 ms: the first fragment of the first message will trigger the fragment timer,
- t + 1000 ms: the fragment timer triggers, it sets flush = true and restarts the timer
- t + 1800 ms: the second fragment of the first message and the first fragment of the second message arrive, the first message is now complete and will be delivered. The fragment from the second will stay in the buffer. *The fragment timer will not be canceled or extended, flush is true*,
- t + 2000 ms: the fragment timer triggers, flush is true, it finds the fragment of the second message and delivers that fragment as partial message, *the transport connection is closed*
- t + 2200 ms: the second fragment of the second message arrives, but *the socket is already closed*


The problem is that when the first message is completed, the fragment timer is not cleared.

Does that explanation make sense?

otp-anders:
I'm not sure I follow your third point; in particular that flush = true. When recv/2 returns it sets flush = false in the second clause, and the call to start_fragment_timer/1 doesn't change that. This then tells flush/1 at timer expiry to start a new timer (and set flush = true) instead of flushing the buffer. The timer is purposely never canceled, just restarted after an expiry. What am I missing?

otp-anders:
If I run the new testcase on the baseline it does indeed fail, but when I look at the timing in the testcase, I expect it to fail. I'm looking at the gen_fragment clause of the init/2 clause. We have fragment_timer = 100. Let's say t = 0 (time) before the first call to gen_send. You send the initial bytes of the first message at this time, which causes the server to start a fragment timer that will expire at t = 100. Since new bytes aren't sent until t = 150, the timer expires and a new timer is started, with flush = true. This timer will expire at t = 200. At t = 150 you send the rest of the first message and some initial bytes from the second. This causes the server to receive the first message and set flush = false. It's timer still expires at t = 200 though, and when it does a new timer is started and flush = true. This timer expires at t = 300. However, you wait another 150 ms in the client before sending the rest of the second message, so that's at the same t = 300 that the fragment timer is due to expire in the server, so it's not unexpected that these bytes arrive at the server too late and the initial bytes of the second message are flushed: the fragment timer has expired twice without the second message being completely received.

There is one fragment timer that is restarted after each expiry as long as there are message bytes that still haven't been received. With fragment_timer = 100, two expiries means that the client might get away with 200 ms between fragments (if the server receives bytes immediately after having reset the timer), but it might need to send them within 100 ms (if the server receives bytes just before it resets the timer). The fragment_timer config says that messages will be flushed between one and two times the configured number of ms after bytes are received, depending on timing.

Correct me if I've completely missed something. :)

otp-anders:
You can play with the timers in your testcase to see the behaviour. For example, set the first timer to 195, so that the rest of the first message is received shortly before the second timer expiry. Then the testcase passes if you set the second timer to 100, so that the rest of the second message is received before the third expiry of the timer, but fails if you set it to 110. That's because receiving the start of the second message 5 ms before the second timer expiry means that it's only 105 ms until two subsequent expiries.

(The third timer in the testcase does nothing, since the next call is just to receive the reply to the second message.)

otp-anders:
Similarly, set your first timer to 105, so that the server receives the rest of the first message just after the first expiry of the fragment timer. Then the testcase passes if the second timer is 190, but fails if it's 200; again, because the rest of the second message is received on opposite sides of the third fragment timer expiry.

RoadRunnr:
> If I run the new testcase on the baseline it does indeed fail, but when I look at the timing in the testcase, I expect it to fail. I'm looking at the gen_fragment clause of the init/2. We have fragment_timer = 100. Let's say t = 0 (time) before the first call to gen_send. You send the initial bytes of the first message at this time, which causes the server to start a fragment timer that will expire at t = 100. Since new bytes aren't sent until t = 150, the timer expires and a new timer is started, with flush = true. This timer will expire at t = 200. At t = 150 you send the rest of the first message and some initial bytes from the second. This causes the server to receive the first message and set flush = false. It's timer still expires at t = 200 though, and when it does a new timer is started and flush = true. This timer expires at t = 300.

And that is the problem, it should IMHO not expire at t = 300. The first fragment to the second message was received at t = 150, therefore the second timer expiry should happen and t = 350.

For a 100 ms second timeout a 50 ms difference seems small, but when you run with the connection with the default 1000 ms fragment timeout over a link with a round trip delay of 200 ms, a 500 ms difference does matter. A lost packet will take at least another 200 ms to arrive and under certain conditions that will just fall into that window. If the timer where full reset the moment the first fragment arrives, the connection would survive.

> However, you wait another 150 ms in the client before sending the rest of the second message, so that's at the same t = 300 that the fragment timer is due to expire in the server, so it's not unexpected that these bytes arrive at the server too late and the initial bytes of the second message are flushed: the fragment timer has expired twice without the second message being completely received.
> 
> There is one fragment timer that is restarted after each expiry as long as there are message bytes that still haven't been received. With fragment_timer = 100, two expiries means that the client might get away with 200 ms between fragments (if the server receives bytes immediately after having reset the timer), but it might need to send them within 100 ms (if the server receives bytes just before it resets the timer). The fragment_timer config says that messages will be flushed between one and two times the configured number of ms after bytes are received, depending on timing.

That is not what the documentation states. There is reads:

> Option fragment_timer specifies the timeout, in milliseconds, of a timer used to flush messages from the incoming byte stream even if the number of bytes indicated in the Message Length field of its Diameter Header have not yet been accumulated: such a message is received over the transport interface after two successive timeouts without the reception of additional bytes. Defaults to 1000.

I would expect that is exactly two timeouts for a give message, not between one and two.

> Correct me if I've completely missed something. :)

otp-anders:
> > It's timer still expires at t = 200 though, and when it does a new timer is started and flush = true. This timer expires at t = 300.
> 
> And that is the problem, it should IMHO not expire at t = 300. The first fragment to the second message was received at t = 150, therefore the second timer expiry should happen and t = 350.

Fair enough, but that's not how it works. :) It's a single timer that's restarted at fixed intervals of fragment_timer, and flushing happens when there are two successive timeouts without additional bytes of an incomplete message being received. When the initial bytes of this incomplete messages were received doesn't affect the timer.

The timer used to be restarted after each fragment reception (before it was documented), and the reason it was changed is that this constant cancelling and restarting of timers resulted in quite a performance penalty, especially in the server case in which there were many peer connections. Cancelling/starting timers has become cheaper since then I think, but that's the background. The current scheme means the timer is only restarted at fixed intervals.

> For a 100 ms second timeout a 50 ms difference seems small, but when you run with the connection with the default 1000 ms fragment timeout over a link with a round trip delay of 200 ms, a 500 ms difference does matter. A lost packet will take at least another 200 ms to arrive and under certain conditions that will just fall into that window. If the timer where full reset the moment the first fragment arrives, the connection would survive.

But if you can't reasonably expect fragments to be spaced at intervals of less then 1000 ms then then fragment_timer needs to be increased. With the current semantics, fragment_timer = 1000 ms means that the connection can tolerate a delay of between 1000 and 2000 ms between fragments, depending on timing. Restarting the timer after the reception of an initial fragment means that a second fragment has exactly 2000 ms to arrive, but it doesn't change the semantics for subsequent fragments. For this you need to restart the timer after each fragment, the way it used to be, so why should the second fragment be treated differently?

> That is not what the documentation states. There is reads:
> 
> > Option fragment_timer specifies the timeout, in milliseconds, of a timer used to flush messages from the incoming byte stream even if the number of bytes indicated in the Message Length field of its Diameter Header have not yet been accumulated: such a message is received over the transport interface after two successive timeouts without the reception of additional bytes. Defaults to 1000.
> 
> I would expect that is exactly two timeouts for a give message, not between one and two.

You're reading more into the documentation than it says. It could be more explicit, but this doesn't say that it's initial message bytes that starts the timer, and it isn't message reception that's required within two timeouts, just additional message bytes. (I've edited one place in an earlier message where I confused the two.)

otp-anders:
If it's a bound on the time to receive a complete message you're after, then that's a different timer. It runs into the same problem with continually cancelling/restarting though, unless it's also implemented as fixed-interval timer, resulting in a range of acceptability rather than a fixed time per message. I don't really see a problem with a range though, and it will be cheaper than cancel/restart, even if the difference might not be as much as it used to be.

otp-anders:
Any comment on this? Is it a message timer you're after or, if not, what is it you want to solve that increasing the fragment timer doesn't address?

One issue with the current timer is that it's not particularly fine-grained with larger values of fragment_timer given that the range or acceptability is 1 to 2 times the configured value. One way to address that would be to allow the frequency of the recurring timer to be increased and trigger on a higher number of traffic-free timeouts than the current 2.

otp-anders:
Just to add to my last comment, instead of just an integer number of milliseconds, could allow a pair with a positive integer to divide the timeout by to get the frequency of the recurring timer. Flushing would then happen when there are one more than this number of timeouts without required fragments being received. The current {fragment_timer, Tmo} would be equivalent to {fragment_timer, {Tmo, 1}}.

otp-anders:
I've implemented a variant of what I mentioned previously, but let me know if that's what you're after. I'll close this PR for now.

PR #2448, opened 2019-11-02
Add openssh vendor extension to sftp rename.

easink:
To use openssh rename extension it use the same logic as openssh scp client and verify if extension exists and choose that first.


HansN:
Looks good so far. What about a test case? Is this option always enabled on for example openssh servers?

easink:
The option is there but its an undocumented "feature" to disable their extension. The "original" way of rename is to link() the file and then unlink() the source. The extension uses rename(), and is in my opinion better, one operation iso two.

HansN:
Thanks for the PR!

PR #2453, opened 2019-11-08
Do not allow whitespace in http header at all

garazdawi:
In RFC 7230 headers must not include any whitespaces because of security reasons. This change returns a http_error when a header ends with a whitespace. i.e. The below is no longer allowed:

    Host : localhost:8000

See https://bugs.erlang.org/browse/ERL-1053 for details.


garazdawi:
merged to master

PR #2454, opened 2019-11-12
Fix erl dist with no atom cache

garazdawi:
Fix the Erlang distribution to handle the scenario when a node connects that can handle message fragmentation but can not handle the atom cache. This bug only affects users that have implemented a custom distribution carrier. It has been present since OTP-21.

This fix will only be merged to master as solving it in maint is not trivial due to other refactorings.

Thanks @essen for reporting

### DIFF
--- a/erts/doc/src/alt_dist.xml
+++ b/erts/doc/src/alt_dist.xml
@@ -785,12 +785,14 @@
 	  <taglist>
 	    <tag><c>DFLAG_DIST_HDR_ATOM_CACHE</c></tag>
 	    <item>Do not use atom cache over this connection.</item>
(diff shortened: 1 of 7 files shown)


essen:
I'm sometimes getting crashes but that might be caused by https://github.com/erlang/otp/pull/2432 as the core dump backtrace looks very similar to the one in the ticket. Your branch starts just before this PR was merged I believe.

garazdawi:
rebased on latest master. Please let me know if you still get segfaults.

essen:
I still get crashes but they're different now, see https://bugs.erlang.org/browse/ERL-1098

PR #2455, opened 2019-11-13
Document the allowed float sizes for the bit syntax

richcarl:
The current implementation of the bit syntax only allows floating point
fields of size 64 or 32. Make this clear in the Erlang Reference Manual.


bjorng:
I cherry-picked your commit to the `maint` branch.

PR #2456, opened 2019-11-13
Build with wsl

dgud:
Allow "cross" building from WSL to windows native with Microsoft toolchain.

Also while doing this make sure that we can upgrade toolchain to newest tools,
such as MSVC-2019, nsis-3.0.4, ssl-1.1.1 and wxWidgets-3.1.1.

Note the actual build results are not test yet, though 'werl.exe' starts, 'crypto' can load nif library
and wx:demo() runs.


dch:
looks awesome! what's the long-term default plan here - to enable building native windows erl* within WSL, as a way to replace MINGW/CYGWIN? Interested in this as we rely on this toolchain for building Apache CouchDB.

dgud:
Yes, the plan is to replace Msys and Msys 2 which we currently use to build native windows, 
and hopefully get a platform that is more stable, faster and easier to maintain. 
Lot of wishes :-)

wohali:
@dgud Experimenting with this, I see another case you've not planned for in `lib/crypto/configure.in`, specific to when `vcpkg` is used to build `openssl`. Would you accept a PR with an additional check for it?

dgud:
I can't promise that until we have seen it, but we do want to improve the configure scripts, and make it
easier to build on windows. So please send in a PR.

PR #2457, opened 2019-11-14
Add support for default metadata to the logger

hauleth:
This adds new `kernel` configuration option `logger_default_metadata` that will contain default metadata for processes.  It can be overriden via `logger:set_process_metadata/1`, so if someone is using that form of setting metadata it will remove the defaults.  Changing the defaults after the local process metadata were set will not be reflected in the metadata

Close ERL-1904

### DIFF
--- a/lib/kernel/doc/src/kernel_app.xml
+++ b/lib/kernel/doc/src/kernel_app.xml
@@ -200,6 +200,12 @@
             <c>logger:set_primary_config(level, Level)</c></seemfa>.</p>
 	<p>Defaults to <c>notice</c>.</p>
       </item>
+      <tag><marker id="logger_metadata"/><c>logger_metadata = Metadata</c></tag>
+      <item>
+	<p>Specifies primary metadata for log events.</p>
+	<p><c>Metadata = map()</c></p>
+	<p>Defaults to <c>#{}</c>.</p>
+      </item>
       <tag><marker id="logger_sasl_compatible"/>
 	<c>logger_sasl_compatible = true | false</c></tag>
       <item>
--- a/lib/kernel/doc/src/logger.xml
+++ b/lib/kernel/doc/src/logger.xml
@@ -199,17 +199,28 @@ logger:error("error happened because: ~p", [Reason]). % Without macro
 	  <item><c>file => ?FILE</c></item>
 	  <item><c>line => ?LINE</c></item>
 	</list>
-	<p>You can add custom metadata, either by specifying a map as
-	  the last parameter to any of the log macros or the API
-	  functions, or by setting process metadata
-	  with <seemfa marker="#set_process_metadata/1">
+	<p>You can add custom metadata, either by:</p>
+        <list>
+          <item>specifying a map as the last parameter to any of the
+            log macros or the logger API functions.</item>
+          <item>setting process metadata with <seemfa marker="#set_process_metadata/1">
 	    <c>set_process_metadata/1</c></seemfa>
-	  or <seemfa marker="#update_process_metadata/1">
-	    <c>update_process_metadata/1</c></seemfa>.</p>
+	    or <seemfa marker="#update_process_metadata/1">
+	  <c>update_process_metadata/1</c></seemfa>.</item>
+          <item>setting primary metadata with <seemfa marker="#set_primary_config/1">
+            <c>set_primary_config/1</c></seemfa> or through the kernel
+            configuration parameter <seeapp marker="kernel_app#logger_metadata">
+            logger_metadata</seeapp></item>
+        </list>
+        <note>
+          <p>When adding custom metadata, make sure not to use any of the keys
+            mentioned above as that may cause a lot of confusion about the log
+            events.</p>
+        </note>
 	<p>Logger merges all the metadata maps before forwarding the
 	  log event to the handlers. If the same keys occur, values
-	  from the log call overwrite process metadata, which in turn
-	  overwrite values set by Logger.</p>
+	  from the log call overwrite process metadata, which overwrites
+          the primary metadata, which in turn overwrite values set by Logger.</p>
 	<p>The following custom metadata keys have special meaning:</p>
 	<taglist>
 	  <tag><c>domain</c></tag>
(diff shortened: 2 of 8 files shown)


hauleth:
elixir-lang/elixir#9521

cc @josevalim @fertapric

garazdawi:
I think that this should be added here: https://github.com/erlang/otp/blob/master/lib/kernel/src/logger_backend.erl#L31 as part of the primary logger config. There should be a `maps:merge/2` there so that the primary metadata is in the lowest priority.

hauleth:
@garazdawi I have added it to the `logger:proc_meta/0` which is then handled in `log_allowed/4`, so all metadata is there before landing in `logger_backend:log_allowed/2`.

garazdawi:
Your way does one extra `ets:lookup/2` per logging call which I would like to avoid.

hauleth:
Ahhh, ok. I now see what you meant. I will provide fix in a moment.

hauleth:
Fixed. Now there are 2 places to define metadata (presented here in increasing importance):

- Primary
- Handlers

So it is possible to have more important "defaults" per handler as well.

garazdawi:
hmm, why put it in the handler config as well? Wouldn't you just put it directly in your template? There is also the template condition syntax for doing this type of thing.

Also, the primary metadata should be part of the Log map when the primary filters are called.

And docs and tests.

hauleth:
@garazdawi I put it into the handlers as well just to make API uniform, but yeah, it doesn't make that much sense, except for a case of reusing the filters. I will update docs and tests as well.

Good point about metadata for filters.

hauleth:
Ok, I am removing the `metadata` support in the handlers. However I see different problem there, when users will use `logger:get_process_metadata/0` they will not see default metadata. I think that this can be problematic in case of external utilities that will want to share metadata with Logger (for example OpenTelemetry). I do not know how much of the problem will that be.

cc @tsloughter

hauleth:
@garazdawi seems that s390x build [failure is caused not by this change](https://travis-ci.org/erlang/otp/jobs/627974872#L2053).

tsloughter:
If you mean `get_process_metadata` doesn't have the defaults but it is still there when the actual log is formatted then I don't see an issue.

josevalim:
After putting more thought into this PR and the original issue, I would say this probably suits better as a handler metadata. For instance, the original issue was about adding things like Commit Hash, which I definitely want to send to certain handlers such as exception tracking, but I probably don't care about it on my general logs. Given it is handler specific, it probably doesn't even need to be part of `logger` either. Any handler can support some sort of default metadata if they wish to.

hauleth:
@josevalim the problem is with what @garazdawi said - the metadata should be available in all filters, even primary ones. And formatters/handlers can handle filtering the metadata if needed.

Also about version/hash in the log messages. It may seem that the version is unneeded in the general log, but as soon as at least one of your servers will not update and you have centralised log aggregator, you may face [Knightmare](https://dougseven.com/2014/04/17/knightmare-a-devops-cautionary-tale/). I would say, that default metadata should be included, well, by default, everywhere, and if some handler want to add more to it (for example Sentry) then they can do so without any problems, with current approach. And from what I have seen, the default metadata was one of the most common requested features from the Elixir's Logger.

lib/kernel/doc/src/kernel_app.xml
```diff
@@ -194,6 +194,12 @@
             <c>logger:set_primary_config(level, Level)</c></seealso>.</p>
 	<p>Defaults to <c>notice</c>.</p>
       </item>
+      <tag><marker id="logger_default_metadata"/><c>logger_default_metadata = Metadata</c></tag>
+      <item>
+	<p>Specifies default value of process metadata for log events.</p>
+	<p><c>Metadata = map()</c></p>
+	<p>Defaults to <c>#{}</c>.</p>
+      </item>
```
garazdawi, line 202: <c>logger_chapter.xml</c> also should be updated with documentation regarding this option.

lib/kernel/doc/src/logger.xml
```diff
@@ -814,10 +814,12 @@ start(_, []) ->
       <name name="set_primary_config" arity="2" clause_i="1" since="OTP 21.0"/>
       <name name="set_primary_config" arity="2" clause_i="2" since="OTP 21.0"/>
       <name name="set_primary_config" arity="2" clause_i="3" since="OTP 21.0"/>
+      <name name="set_primary_config" arity="2" clause_i="4" since="OTP 21.0"/>
```
garazdawi, line 817:
```suggestion
      <name name="set_primary_config" arity="2" clause_i="4" since="OTP @OTP-17181@"/>
```

garazdawi (commented):
We have decided to include this in OTP-24. Sorry for the huge delay in making a decision.

Would you mind fixing the small changes needed to include this? If not I will do it later this week.

hauleth:
Sure, I will update it later today.

garazdawi:
I started looking at finishing the docs and found that I did not like the way that the implementation behaved. I can't remember what you originally submitted, but I think what I changed it to now is closer to that.

So what I did was:

* rebase the branch to latest master
* change name of the kernel parameter to `logger_metadata` and in the docs it is referred to as "primary metadata"
* change the merge of the metadata so that we get `Location MD -> Primary MD -> Process MD -> Call MD` where "Location MD" has the lower priority and "Call MD" the highest.
* fixed tests and updated docs

hauleth:
@garazdawi seems ok.

hauleth:
However I am worried what will happen if someone will set `#{module := foo}` in their Primary MD, will that override all entries with `foo` as their module?

garazdawi:
> However I am worried what will happen if someone will set `#{module := foo}` in their Primary MD, will that override all entries with `foo` as their module?

Yes, it will override. Not really sure what to do about that though, or if we should do anything about it. I suppose the alternatives are to either reject setting such a primary metadata, or moving the merge order of metadata around so that `Location` metadata has a higher priority.

Then again process metadata has always merged like this, and so far I haven't heard anyone complain about that, though that does not mean that it has not happened.

josevalim:
> Then again process metadata has always merged like this, and so far I haven't heard anyone complain about that, though that does not mean that it has not happened.

Yeah. I think having a note in the docs that points to "reserved" keys and say that you most likely don't want to ever set those, in any context, is sufficient.

garazdawi:
Added a note in the metadata docs about this.

garazdawi:
Squashed and merges, thanks for the PR!

PR #2100, opened 2019-01-21
Apply type optimizations across local function calls

jhogberg:
The current type pass works on a per-function level and assumes that every argument and return value could be anything, which means they have to be type-checked before they're used. This is a bit inefficient as most functions always return something of the same type, and their argument types seldom differ.

This PR makes the type pass track the return and argument types of local functions, letting us eliminate many redundant type checks.

```erlang
example1() ->
    %% We no longer check that this is a 2-tuple before extracting A & B
    {A, B} = returns_2_tuple(), 
    A + B.

returns_2_tuple() ->
    {a:a(), b:b()}.

-record(a, {n=0}).

example2() ->
    loop(lists:seq(1, 20), #a{}).

%% Record type is no longer checked on each iteration
loop([I | Is], #a{n=N}=St) ->
    loop(Is, St#a{n=N+I});
loop([], St) ->
    St.
```

PR #2460, opened 2019-11-18
compiler: Improve type inference

jhogberg:
The cross-function type optimization introduced in PR #2100 removed a lot of redundant type checks, but the implementation was very naive and often lost type information in recursive functions which greatly limited its effectiveness. For example, it couldn't figure out that the following always produced a proper list:

```erlang
foo([A | As]) -> [A | foo(As)];
foo([]) -> [].
```

This PR improves the compiler's understanding of recursive functions and leverages that to keep track of types inside lists, allowing us to remove more type tests in loops.

I'm still waiting for the benchmarks to finish running, but until then [here's an assembly diff](https://github.com/erlang/otp/files/3858858/improved_type_inference_diff.txt) showing the impact this PR has on OTP and parts of Elixir.

### DIFF
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -29,17 +29,13 @@
 
 -define(UNICODE_MAX, (16#10FFFF)).
 
--record(d,
-        {ds :: #{beam_ssa:b_var():=beam_ssa:b_set()},
-         ls :: #{beam_ssa:label():=type_db()},
-         once :: cerl_sets:set(beam_ssa:b_var()),
-         params :: [beam_ssa:b_var()],
-         func_id :: func_id(),
-         func_db :: func_info_db(),
-         sub = #{} :: #{beam_ssa:b_var():=beam_ssa:value()},
-         succ_types = [] :: success_type_set()}).
+%% Constant metadata used by both subpasses.
+-record(metadata,
+        { func_id :: func_id(),
+          params :: [beam_ssa:b_var()],
+          used_once :: cerl_sets:set(beam_ssa:b_var()) }).
 
--type type_db() :: #{beam_ssa:var_name():=type()}.
+-type type_db() :: #{ beam_ssa:var_name() := type() }.
 
 -spec opt_start(Linear, Args, Anno, FuncDb) -> {Linear, FuncDb} when
       Linear :: [{non_neg_integer(), beam_ssa:b_blk()}],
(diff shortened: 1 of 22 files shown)


lib/compiler/src/beam_types.erl
```diff
@@ -347,8 +347,13 @@ subtract(#t_integer{elements={Min, Max}}, #t_integer{elements={N,N}}) ->
     end;
 subtract(number, #t_float{elements=any}) -> #t_integer{};
 subtract(number, #t_integer{elements=any}) -> #t_float{};
-subtract(list, cons) -> nil;
-subtract(list, nil) -> cons;
+subtract(#t_list{type=TypeA}=T, #t_cons{head=TypeB}) ->
+    case meet(TypeA, TypeB) of
+        TypeA -> nil;
```
bjorng: This seems to be unsafe for a non-proper list. If the list is not known to be proper, you can't be sure that the tail of the final cons is nil.

lib/compiler/src/beam_types.hrl
```diff
@@ -51,14 +51,16 @@
...
 -record(t_bitstring, {size_unit=1 :: pos_integer()}).
 -record(t_bs_context, {tail_unit=1 :: pos_integer(),
                        slots=0 :: non_neg_integer(),
                        valid=0 :: non_neg_integer()}).
 -record(t_bs_matchable, {tail_unit=1}).
+-record(t_cons, {head=any :: type(), proper=false :: boolean()}).
+-record(t_float, {elements=any :: 'any' | {float(),float()}}).
+-record(t_fun, {arity=any :: arity() | 'any'}).
+-record(t_integer, {elements=any :: 'any' | {integer(),integer()}}).
+-record(t_list, {type=any :: type(), proper=false :: boolean()}).
```
bjorng: You have forgotten to update the comment at the top of the file. It stil says that the atoms `cons` and `list` represent the cons and list types, respectively.

lib/compiler/src/beam_types.erl
jhogberg: The subtraction is on the list itself, so returning `nil` is correct even when the list isn't proper.

Thanks for pointing this out though, we should only do it when the cons cell is at least as wide as the one in the list (lists being `#t_cons{} | nil`), which includes whether they're proper.

lib/compiler/src/beam_types.hrl
jhogberg: Fixed locally, thanks!

lib/compiler/src/beam_types.erl
```diff
@@ -382,7 +382,12 @@ subtract(#t_union{tuple_set=#t_tuple{}=Tuple}=A, #t_tuple{}=B) ->
 subtract(#t_union{other=Other}=A, B) ->
     shrink_union(A#t_union{other=subtract(Other, B)});
 
-subtract(T, _) -> T.
+subtract(A, B) ->
+    %% There's nothing left if A is strictly more specific than B.
```
bjorng: I think that the word *strictly* should be omitted. There will be nothing left if the types are equally specific.

lib/compiler/src/beam_ssa_type.erl
```diff
@@ -17,75 +17,383 @@
...
+%% types as in the paper because we need to know all possible inputs including
+%% those that will not return.
+%%
+%% [1] http://www.it.uu.se/research/group/hipe/papers/succ_types.pdf
+%%
+
+-record(worklist,
+        { counter = 0 :: integer(),
+          elements = gb_trees:empty() :: gb_trees:tree(integer(), term()),
+          indexes = #{} :: #{ term() => integer() } }).
```
bjorng: For clarity, I suggest making #worklist{} an abstract type. (It is *almost* an abstract type already.) That is, move the `worklist` record declaration to just before the implementation of the functions that operate on it, add a new `wl_new/0` function, and define a `worklist()` type. It is allowed to reference a type that is being defined later in the same module, so you can use the `worklist()` type for the `wl` field in `#sig_st{}`.

lib/compiler/src/beam_types.erl
jhogberg: Good point, fixed locally.

lib/compiler/src/beam_ssa_type.erl
jhogberg: Sure thing, fixed locally.

PR #2461, opened 2019-11-20
Allow changing the -mode when the system restarts

josevalim:
In order for Elixir developers to configure their
systems using Elixir syntax, the Elixir application
must be started. However, in order to start the
Elixir application, kernel and stdlib need to be
started, which means we can't configure those
applications.

Elixir addresses this by booting twice. First, the
system boots with a minimal set of apps started, then
it executes the configuration files, writes them to
disk and restarts the system, which will pick up the
new configuration.

The problem with such approach is that, when running
in embedded mode, starting and shutting down the system
takes a long period of time, due to the loading and
purging of all modules.

This PR addresses the problem by allowing the mode to
be changed to embedded on init:restart/1. This allows
Elixir to first boot in interactive mode, load the
configuration, and then start the system in embedded
mode.

This reduces the boot time in the sample application
that reproduces the problem from 5s to 1s.

### DIFF
--- a/erts/doc/src/init.xml
+++ b/erts/doc/src/init.xml
@@ -165,12 +165,24 @@
       <name name="restart" arity="0" since=""/>
       <fsummary>Restart the running Erlang node.</fsummary>
       <desc>
+        <p>The same as
+    <seealso marker="#restart/1"><c>restart([])</c></seealso>.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="restart" arity="1" since="OTP 23.0"/>
+      <fsummary>Restart the running Erlang node.</fsummary>
+      <desc>
         <p>The system is restarted <em>inside</em> the running Erlang
           node, which means that the emulator is not restarted. All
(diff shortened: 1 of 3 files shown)


josevalim:
Note: I have not updated the preloaded files and this is required when merging. AFAIK, this is done by the OTP team manually but let me know if I should do it myself.

josevalim:
Hi @rickard-green. I apologize if you are not the correct person to ping but is there any chance this can be looked at for inclusion in Erlang/OTP 23.0? This will provide a very meaningful speed up for Elixir releases. Thank you!

rickard-green:
@josevalim I'll have a look at it

rickard-green:
Please change the `init:restart/1` function to accept an option list instead of the mode argument, so that it is possible to extend it with other features in the future.

Also it would be good to test both supported operations. Restarting interactive -> embedded as well as embedded -> interactive.

josevalim:
@rickard-green thank you. Feedback has been addressed.

PR #2462, opened 2019-11-22
Preserve http headers original case while erlang:decode_packet/3

tank-bohr:
Function `erlang:decode_packet/3` titleizes header names (convert the case to `X-Awesome-Header` shape) for some reason. By [RFC](https://tools.ietf.org/html/rfc2616#section-4.2) headers are case-insensitive. But in some cases in can be important to keep the original case of the header name. For example for testing purposes. bookish_spork library tries to give a user the most precise means for http request introspection as possible. Preserving headers case will allow to check how http-clients and some api libraries manage headers

See: https://github.com/tank-bohr/bookish_spork/issues/44

### DIFF
--- a/erts/emulator/beam/packet_parser.c
+++ b/erts/emulator/beam/packet_parser.c
@@ -780,7 +780,6 @@ int packet_parse_http(const char* buf, int len, int* statep,
         }
     }
     else {
-        int up = 1;      /* make next char uppercase */
         http_atom_t* name;
         char name_buf[HTTP_MAX_NAME_LEN];
         const char* name_ptr = name_buf;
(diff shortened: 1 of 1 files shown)


garazdawi:
> Function erlang:decode_packet/3 titleizes header names for some reason.

The reason is in order to make it simpler for the user (they only have to deal with matching on `X-Awesome-Header`) and be compliant with the RFC.

This breaks backward compatibility in a huge way, so we cannot accept it. If you need this functionality I recommend looking on hex for a package that does it. Or if there are none, create one and publish it for others to use.

tank-bohr:
I bet nobody relies on it. It's a good change. It eliminates useless case conversions. Availability of built-in http-parser inspired me to write this library 😿

garazdawi:
> I bet nobody relies on it.

https://github.com/klacke/yaws/blob/b703951f7a13d5ef44054b884690f0d146f27118/src/yaws_api.erl#L468

I'm not so sure.

PR #2465, opened 2019-11-24
Add uri_string:resolve/2,3

essen:
Please consider including them in OTP directly. For some backstory:

I was looking for this function because as part of implementing the parsing for the Link HTTP header it's necessary to resolve potentially relative URIs according to https://tools.ietf.org/html/rfc3986#section-5

After asking it turns out @darkling already needed this function in https://github.com/darkling/lagra/blob/30bba5322630382f046decb9bb83fd4ea75e058c/src/lagra_parser_turtle_parser.erl#L585

I asked him about license but I needed it before I got a response so I reimplemented it in Cowlib here: https://github.com/ninenines/cowlib/blob/master/src/cow_link.erl#L261

This is this implementation that I submit here as a PR, which is more or less the algorithm from https://tools.ietf.org/html/rfc3986#section-5.2.2

The documentation is currently missing and the whitespace/indentation is probably wrong, I'll add/fix them once I get some feedback.

### DIFF
--- a/lib/stdlib/src/uri_string.erl
+++ b/lib/stdlib/src/uri_string.erl
@@ -228,7 +228,7 @@
 %%-------------------------------------------------------------------------
 -export([compose_query/1, compose_query/2,
          dissect_query/1, normalize/1, normalize/2, parse/1,
-         recompose/1, transcode/2]).
+         recompose/1, resolve/2, resolve/3, transcode/2]).
 -export_type([error/0, uri_map/0, uri_string/0]).
 
 
@@ -372,6 +372,45 @@ recompose(Map) ->
             catch
                 throw:{error, Atom, RestData} -> {error, Atom, RestData}
             end
+    end.
+
+
+%%-------------------------------------------------------------------------
+%% Resolve URIs
+%%-------------------------------------------------------------------------
+-spec resolve(RefURI, BaseURI) -> TargetURI when
+      RefURI :: uri_string() | uri_map(),
+      BaseURI :: uri_string() | uri_map(),
+      TargetURI :: uri_string()
+                 | error().
+resolve(URIMap, BaseURIMap) ->
+    resolve(URIMap, BaseURIMap, []).
+
+
+-spec resolve(RefURI, BaseURI, Options) -> TargetURI when
+      RefURI :: uri_string() | uri_map(),
+      BaseURI :: uri_string() | uri_map(),
+      Options :: [return_map],
+      TargetURI :: uri_string() | uri_map()
+                 | error().
+resolve(URIMap, BaseURIMap, Options) when is_map(URIMap) ->
+    case resolve_map(URIMap, BaseURIMap) of
+        TargetURIMap when is_map(TargetURIMap) ->
+            case Options of
+                [return_map] ->
+                    TargetURIMap;
+                [] ->
+                    recompose(TargetURIMap)
+            end;
+        Error ->
+            Error
+    end;
+resolve(URIString, BaseURIMap, Options) ->
+    case parse(URIString) of
+        URIMap when is_map(URIMap) ->
+            resolve(URIMap, BaseURIMap, Options);
+        Error ->
+            Error
     end.
 
 
@@ -1706,6 +1745,69 @@ maybe_to_list(Comp) -> Comp.
 
 encode_port(Port) ->
     integer_to_binary(Port).
+
+
+%%-------------------------------------------------------------------------
+%% Helper functions for resolve
+%%-------------------------------------------------------------------------
+
+resolve_map(URIMap=#{scheme := _}, _) ->
+    normalize_path_segment(URIMap);
+resolve_map(URIMap, #{scheme := _}=BaseURIMap) ->
+    resolve_map(URIMap, BaseURIMap, resolve_path_type(URIMap));
+resolve_map(_URIMap, BaseURIMap) when is_map(BaseURIMap) ->
+    {error,invalid_scheme,""};
+resolve_map(URIMap, BaseURIString) ->
+    case parse(BaseURIString) of
+        BaseURIMap = #{scheme := _} ->
+            resolve_map(URIMap, BaseURIMap, resolve_path_type(URIMap));
+        BaseURIMap when is_map(BaseURIMap) ->
+            {error,invalid_scheme,""};
+        Error ->
+            Error
+    end.
+
+resolve_path_type(URIMap) ->
+    case iolist_to_binary(maps:get(path, URIMap, <<>>)) of
+        <<>> -> empty_path;
+        <<$/,_/bits>> -> absolute_path;
+        _ -> relative_path
+    end.
+
+resolve_map(URI=#{host := _}, #{scheme := Scheme}, _) ->
+    normalize_path_segment(URI#{scheme => Scheme});
+resolve_map(URI, BaseURI, empty_path) ->
+    Keys = case maps:is_key(query, URI) of
+        true -> [scheme, userinfo, host, port, path];
+        false -> [scheme, userinfo, host, port, path, query]
+    end,
+    maps:merge(URI, maps:with(Keys, BaseURI));
+resolve_map(URI, BaseURI, absolute_path) ->
+    normalize_path_segment(maps:merge(
+        URI,
+        maps:with([scheme, userinfo, host, port], BaseURI)));
+resolve_map(URI=#{path := Path}, BaseURI, relative_path) ->
+    normalize_path_segment(maps:merge(
+        URI#{path => merge_paths(Path, BaseURI)},
+        maps:with([scheme, userinfo, host, port], BaseURI))).
+
+merge_paths(Path, BaseURI=#{path := BasePath0}) ->
+    case {BaseURI, iolist_size(BasePath0)} of
+        {#{host := _}, 0} ->
+            merge_paths_absolute(Path);
+        _ ->
+            case string:split(BasePath0, <<$/>>, trailing) of
+                [BasePath, _] when is_binary(Path) -> unicode:characters_to_binary([BasePath, $/, Path]);
+                [BasePath, _] when is_list(Path) -> unicode:characters_to_list([BasePath, $/, Path]);
+                [_] -> Path
+            end
+    end.
+
+merge_paths_absolute(Path) when is_binary(Path) ->
+    <<$/, Path/binary>>;
+merge_paths_absolute(Path) when is_list(Path) ->
+    unicode:characters_to_list([$/, Path]).
+
 
 %%-------------------------------------------------------------------------
 %% Helper functions for transcode
--- a/lib/stdlib/test/uri_string_SUITE.erl
+++ b/lib/stdlib/test/uri_string_SUITE.erl
@@ -44,6 +44,8 @@
          recompose_query/1, recompose_parse_query/1,
          recompose_path/1, recompose_parse_path/1,
          recompose_autogen/1, parse_recompose_autogen/1,
+         resolve_normal_examples/1, resolve_abnormal_examples/1,
+         resolve_base_uri/1, resolve_return_map/1,
          transcode_basic/1, transcode_options/1, transcode_mixed/1, transcode_negative/1,
          compose_query/1, compose_query_latin1/1, compose_query_negative/1,
          dissect_query/1, dissect_query_negative/1,
@@ -123,6 +125,10 @@ all() ->
      recompose_parse_path,
      recompose_autogen,
      parse_recompose_autogen,
+     resolve_normal_examples,
+     resolve_abnormal_examples,
+     resolve_base_uri,
+     resolve_return_map,
      transcode_basic,
      transcode_options,
      transcode_mixed,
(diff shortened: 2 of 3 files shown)


lib/stdlib/src/uri_string.erl
```diff
@@ -375,6 +375,80 @@ recompose(Map) ->
...
+-spec resolve(RefURI, BaseURI, Options) -> TargetURI when
+      RefURI :: uri_string() | uri_map(),
+      BaseURI :: uri_string() | uri_map(),
+      Options :: [return_map],
+      TargetURI :: uri_string() | uri_map()
+                 | error().
+resolve(URIMap, BaseURIMap, Options) ->
+    try resolve1(URIMap, BaseURIMap, Options)
+    catch
+        throw:{error, Atom, RestData} -> {error, Atom, RestData}
```
peterdmv: Can we get this exception here?

lib/stdlib/src/uri_string.erl
```diff
@@ -375,6 +375,80 @@ recompose(Map) ->
...
+    end.
+
+resolve1(URI, BaseURI, Options) ->
```
peterdmv: I would move helper functions to "Internal functions" to keep them separated from the API.

lib/stdlib/src/uri_string.erl
essen: Don't think so but I wasn't sure until this morning. I'll remove it.

lib/stdlib/src/uri_string.erl
essen: Yes I guessed I would have to. I will do that when I update the documentation.

lib/stdlib/src/uri_string.erl
```diff
@@ -375,6 +375,80 @@ recompose(Map) ->
...
+
+resolve1(URI=#{scheme := _}, _) ->
+	URI;
+resolve1(URI=#{host := _}, BaseURI) ->
+	#{scheme := Scheme} = ensure_map_uri(BaseURI),
+	URI#{scheme => Scheme};
+resolve1(URI=#{path := <<>>}, BaseURI0) ->
+	BaseURI = ensure_map_uri(BaseURI0),
+	Keys = case maps:is_key(query, URI) of
+		true -> [scheme, host, port, path];
```
peterdmv: +               true -> [scheme, host, port, userinfo, path];
Add also userinfo as "authority" has 3 subcomponents: host, port and userinfo.
Same applies to the other clauses.

lib/stdlib/src/uri_string.erl
```diff
@@ -375,6 +375,80 @@ recompose(Map) ->
...
+resolve(URIMap, BaseURIMap, Options) ->
+    try resolve1(URIMap, BaseURIMap, Options)
+    catch
+        throw:{error, Atom, RestData} -> {error, Atom, RestData}
+    end.
+
+resolve1(URI, BaseURI, Options) ->
+	TargetURI = case resolve1(ensure_map_uri(URI), BaseURI) of
+		TargetURI0 = #{path := Path0} ->
+			%% We remove dot segments. Normalizing the entire URI
```
peterdmv: The extra slash is added by the scheme based normalization see: 

```
6.2.3.  Scheme-Based Normalization

For example,
   because the "http" scheme makes use of an authority component, has a
   default port of "80", and defines an empty path to be equivalent to
   "/", the following four URIs are equivalent:

      http://example.com
      http://example.com/
      http://example.com:/
      http://example.com:80/

   In general, a URI that uses the generic syntax for authority with an
   empty path should be normalized to a path of "/".
```
I think it is safe to normalize the whole URI here!
See my comment below regarding the failing TC.

lib/stdlib/test/uri_string_SUITE.erl
```diff
@@ -825,6 +829,59 @@ parse_recompose_autogen(_Config) ->
...
     lists:map(fun run_test_parse_recompose/1, Tests).
 
+resolve_normal_examples(_Config) ->
+    BaseURI = <<"http://a/b/c/d;p?q">>,
+    <<"g:h">> = uri_string:resolve(<<"g:h">>, BaseURI),
+    <<"http://a/b/c/g">> = uri_string:resolve(<<"g">>, BaseURI),
+    <<"http://a/b/c/g">> = uri_string:resolve(<<"./g">>, BaseURI),
+    <<"http://a/b/c/g/">> = uri_string:resolve(<<"g/">>, BaseURI),
+    <<"http://a/g">> = uri_string:resolve(<<"/g">>, BaseURI),
+    <<"http://g">> = uri_string:resolve(<<"//g">>, BaseURI),
```
peterdmv: This specific example is not normalized by scheme in RFC3986 and I think that even sheme-based normalization should have been applied here to avoid ambiguity as all other exampes are normalized. I think the correct result is <<"http://g/">>.

essen: There doesn't seem to be an errata for this one and the algorithm does not mention normalization (which may be applied afterwards), I just used `normalize` because I didn't have a `remove_dot_segments` in Cowlib. But there's probably one in `uri_string` so I probably just need to swap that `normalize` call with `remove_dot_segments`.

lib/stdlib/src/uri_string.erl
```diff
@@ -375,6 +375,80 @@ recompose(Map) ->
...
+    resolve(URIMap, BaseURIMap, []).
+
+
+-spec resolve(RefURI, BaseURI, Options) -> TargetURI when
+      RefURI :: uri_string() | uri_map(),
+      BaseURI :: uri_string() | uri_map(),
+      Options :: [return_map],
+      TargetURI :: uri_string() | uri_map()
+                 | error().
+resolve(URIMap, BaseURIMap, Options) ->
```
peterdmv:
```erlang
resolve(URIMap, BaseURIMap, Options) ->
    normalize(resolve1(ensure_map_uri(URIMap), BaseURIMap), Options).
```
We can also implement this in a one liner and call ```resolve1/2``` directly. This would mean that the target URI is always normalized. Is there any practical use for producing non-normalized URIs?
E.g.
```
135>  uri_string:resolve(<<"../../../g">>, <<"http://a/b/c/d;p?q">>).
<<"http://a/b/c/../../../g">>

```

essen: It should not produce those URIs because the dot segments must be removed according to the algorithm. That's what I was using `normalize` for but it really should just be `remove_dot_segments` as I mentioned. I'll work on it tomorrow or Thursday.

essen:
Pushed code changes. I still need to move the "helper" functions, fix the whitespace and add documentation.

lib/stdlib/src/uri_string.erl
essen: I removed the try/catch.

lib/stdlib/src/uri_string.erl
essen: Done.

lib/stdlib/src/uri_string.erl
essen: I am no longer using the (full) normalize function, it's now using the `normalize_path_segment` function which just calls `remove_dot_segments` on the path. The comment is no longer relevant as a result.

lib/stdlib/test/uri_string_SUITE.erl
essen: Now using the correct function.

lib/stdlib/src/uri_string.erl
essen: Now using the correct function.

essen:
... I somehow forgot to remove the unneeded normalize step.

essen:
One more thing that needs fixing: currently it converts string to binary, I suspect we want to leave as string like all the other functions.

essen:
I think the PR is pretty good now, still need to move helper functions and fix the whitespace though.

I think I've adressed all other stuff.

essen:
```
Error in uri_string:resolve/1: unknown spec
```

Am I supposed to do something to make this pass?

lib/stdlib/src/uri_string.erl
```diff
@@ -375,6 +375,98 @@ recompose(Map) ->
...
+        maps:with([scheme, userinfo, host, port], BaseURI)));
+resolve_map(URI=#{path := Path}, BaseURI, relative_path) ->
+	normalize_path_segment(maps:merge(
+		URI#{path => merge_paths(Path, BaseURI)},
+		maps:with([scheme, userinfo, host, port], BaseURI))).
+
+merge_paths(Path, #{path := BasePath0}) ->
+    case iolist_size(BasePath0) of
+        0 ->
+            merge_paths_absolute(Path);
```
peterdmv:
```
If the base URI has a defined authority component and an empty
path, then return a string consisting of "/" concatenated with the
reference's path; otherwise,
```
```
1> uri_string:resolve("a/b/c","").
"/a/b/c"
```
I've found this special case with an empty BaseURI where I think the correct result should be "a/b/c".

essen: There's a mistake but your interpretation is wrong. It says if the base URI *has* an authority component, so this does not include `""`. Note that based on the algorithm the `merge_path` function does not seem to ever be allowed to return a relative path.

The mistake is that when refactoring I dropped the check on the authority component entirely. I will add it back.

essen:
I changed the target branch to maint, I messed up when opening the PR I was already developing against maint.

I will do the two changes and move functions/fix whitespace tomorrow. I want to add some tests with a different base URI as well to cover more cases, question is how do I run tests with code coverage enabled?

lib/stdlib/src/uri_string.erl
peterdmv: You are right, it cannot return _just_ a relative path, but we need to add more checks even for the base URI. According to chapter 5.2.1, the scheme component is required to be present in a base URI.
```
5.2.1.  Pre-parse the Base URI

   The base URI (Base) is established according to the procedure of
   Section 5.1 and parsed into the five main components described in
   Section 3.  Note that only the scheme component is required to be
   present in a base URI; the other components may be empty or
   undefined.  A component is undefined if its associated delimiter does
   not appear in the URI reference; the path component is never
   undefined, though it may be empty.
```
This requirement should be fulfilled with those in chapter 5.2.3:

```
5.2.3.  Merge Paths

   The pseudocode above refers to a "merge" routine for merging a
   relative-path reference with the path of the base URI.  This is
   accomplished as follows:

   o  If the base URI has a defined authority component and an empty
      path, then return a string consisting of "/" concatenated with the
      reference's path; otherwise,

   o  return a string consisting of the reference's path component
      appended to all but the last segment of the base URI's path (i.e.,
      excluding any characters after the right-most "/" in the base URI
      path, or excluding the entire base URI path if it does not contain
      any "/" characters).
```

So my previous example should return an error tuple instead as the base URI is not valid. Then we have the case when the base URI consists of only the scheme component e.g. ```http:```. In that case we cannot apply the first paragraph of 5.2.3 as the base URI has an empty path but the authority component is missing. We can apply the second paragraph and exclude the entire base URI path as it does not contain any "/" characters. So we get a target URI with a scheme and a relative path:

```
1> uri_string:resolve("a/b/c","http:").
"http:a/b/c"
```
I think this could be added to the abnormal examples.

essen: We need tests with many different types of base URIs. It's weird that they didn't add any.

By the way what does `"http:a/b/c"` even mean? For some URI schemes I suppose they would be valid, but for `http` it's invalid according to https://tools.ietf.org/html/rfc7230#section-2.7.1

```
   A sender MUST NOT generate an "http" URI with an empty host
   identifier.  A recipient that processes such a URI reference MUST
   reject it as invalid.
```

Do we want to special case `http`, `https` and perhaps other schemes and return an error for them if the base URI has no `host` component?

peterdmv: I'm more inclined to have special handling for at least `http `and `https`. RFC3986 includes for instance the scheme based normalization for `http`, `https`, `ftp`, `ssh`, `sftp` and `tftp`. We also took a pragmatic approach for the handling of urlencoded query strings based on HTML 5.0 and HTML 5.2 (WHAT WG URL). I think having a special check for `http `and `https ` would have more practical benefits compared to keeping the module's knowledge of the `http` scheme on a bare minimum.

essen: Hmm but then there's the `http:g` example where `http:g` must be returned, so as far as HTTP is concerned perhaps a separate function would be better to check that you have a URI that's valid according to the scheme. I don't know. I'm fine checking it in user code for now and perhaps we can discuss this again after some real world experience.

I will add tests with different types of base URIs for now.

essen:
Added base URI tests, squashed, fixed whitespace, moved the internal functions down the file, applied your documentation fix.

peterdmv:
You can use the following commands to get code coverage:
```
cd $TESTROOT/stdlib_test
erl
> ct:run_test([{suite, uri_string_SUITE}, {cover,"stdlib.cover"},{cover_stop,false}]).
```

peterdmv:
Thanks for the contribution!

essen:
Woot!!!

PR #2079, opened 2019-01-07
Add packet_spec packet mode to packet parser

Vagabond:
packet_spec is a way to parse simple arbitrary binary packet protocols
where the length field is not the first field. Packet specs consist of a
list of [ 'u8' | 'u16' | u16le' | 'u32' | 'u32le' | 'varint' ]. The last
field listed is the length of the payload. Fields before the final field
contribute to the header length. Fixed length footers can be described
with the `{packet_spec_extra_bytes, non_neg_integer()}` option.

**Why do we need this new feature?**

Erlang's built in packet parser currently supports big-endian length-prefixed packets, ASN.1 BER packets, CORBA packets, SUN RPC packets, HTTP packets, etc. However there are many packet formats that *almost* work with Erlang's existing packet parser. For example many packets have a fixed length leader before the length field, or have a little-endian length field. This feature is an attempt to provide a simple way to delegate packet parsing of simple binary packets that are not simply big-endian length prefixed to the packet parser.

**Risks or uncertain artifacts?**

I could use guidance on the best way to implement some of this, especially how the `packet_spec` struct is initialized and passed around and how the prim_inet changes should look. I believe the feature itself is not a risk because it does not allow unbounded or infinite packets and does not consume unbounded memory.

**How did you solve it?**

As can be seen from the patch, a new packet mode and 2 new packet options have been added. I tried to avoid any dynamic allocation and do everything on the stack, although I am not sure this is correct as it is an optional feature. Additionally I added special encoding for the packet_spec in prim_inet and I'm not sure that is the correct approach.

Please let me know if this feature is useful. I wrote a blog post detailing the motivations behind it in more detail here: http://vagabond.github.io/programming/2018/12/30/notes-on-extending-the-erlang-packet-parser

PR #2466, opened 2019-11-25
Change Reserved to unmodified HTTP header-field

matyasmarkovics:
Propagate unmodified header fields in the previously Reserved and
undefined 4th element of the http_header tuple.

**Why do we need this new feature?**

While[ Section 4.2 of RFC 2616 for HTTP/1.1](https://tools.ietf.org/html/rfc2616#section-4.2) states
that Field names are case-insensitive, some non-compliant
web services may rely on peculiar casing of Field names.
Such web-services cannot be accessed through a Proxy-server,
that is implemented in Erlang/OTP, certainly not over HTTPS.

Despite the RFC, even the [IANA lists a number of permanently
registered headers](https://www.iana.org/assignments/message-headers/message-headers.xml#perm-headers) in all-caps or mixed-cased format,
e.g.: ALPN, HTTP2-Settings, WWW-Authenticate, etc.
This leads to confusion and to service implementations,
that only accept headers formatted the same way.
A production issue was experienced using the common, drafted,
but not yet standard header: [DNT](http://tools.ietf.org/html/draft-mayer-do-not-track-00).

There are a number of HTTP servers implemented in Erlang
that could be used in a Proxy implementation. However,
there is none that would respect header-casing.
* [cowboy](https://github.com/ninenines/cowboy/blob/master/src/cowboy_http.erl#L605), httpd - their header parser lower-cases the fields
* [yaws](https://github.com/klacke/yaws/blob/master/src/yaws.erl#L2701) - emulators' internal HTTP header parsing via ssl/inet
* [mochiweb](https://github.com/scrapinghub/mochiweb/blob/master/src/mochiweb_http.erl#L103), [elli](https://github.com/elli-lib/elli/blob/develop/src/elli_http.erl#L465) - both use erlang:decode_packet/3

As a side note: An HTTPS-Proxy requires CONNECT method support.
Therefor [cowboy is not a real option](https://github.com/ninenines/cowboy/blob/master/src/cowboy_http.erl#L420) despite its popularity.

With this proposal yaws, mochiweb and elli could all be patched
to respect casing of header fields and propagate the original
header format to the application level modules.
The DNT header issue was experienced using a patched mochiweb.

**Risks or uncertain artifacts?**

These changes are minimalistic and backward compatible,
given application developers respected the documentation.
All above mentioned http servers do so, they ignore the 4th
element of http_header tuple. It was verified in the master
branches of the project repositories.

**How did you solve it?**

The proposal is that packet_parser.c should also propagate
unmodified fields to both inet_drv.c and erl_bif_port.c,
where these would be put in the 4th place of the http_header
tuple. Its recognised, that this element was Reserved
for future and/or internal use. Hopefully you'd agree, that
sending the original header fields is a productive use-case
to free up the reservation.

Adding a new httph_cs option, cs for case-sensitive, was also
considered. I got the [impression](https://github.com/erlang/otp/pull/2079) thought, that these these
packet decoders exit for compatibility reasons only and
the OTP-Team is not keen on adding new options.
That said I'm happy to add a `httph_cs` and `http_bin_cs`,
if that is what is need for the merge to happen.

*PS*: A related [Rejected PR](https://github.com/erlang/otp/pull/2462) from a colleague.


garazdawi:
Yes, this is a much better approach. It just needs some updates to the docs and should be good.

matyasmarkovics:
Thanks for the feedback @garazdawi. Both `ssl` and `inet` app. docs refer back to the documentation of `erlang:decode_packet/3`, so I've amended that one only. Please let me know if you think updates needed elsewhere. I have looked but couldn't find other relevant places to update.

matyasmarkovics:
While I was searching through `cowboy`'s commit history to create this PR's description, I've came across a comment about how `decode_packet` title-cases headers with double hyphens incorrectly: e.g.: `Content--Length` will be `Content--length`, so I've added a test-case for this. It's not necessarily an issue as I'm not aware of legitimate headers with such a format. However, maybe it's good to have a test, so that the casing behaviour is known.

garazdawi:
> Please let me know if you think updates needed elsewhere.

`erlang:decode_packet/3` was were I meant, so this is fine.

> However, maybe it's good to have a test, so that the casing behaviour is known.

Agreed. Good initiative, thanks.

Putting it in testing over the weekend.

erts/emulator/drivers/common/inet_drv.c
```diff
@@ -2610,7 +2612,7 @@ http_header_inetdrv(void* arg, const http_atom_t* name, const char* name_ptr,
       i = LOAD_INT(spec, i,  0);
       i = http_load_string(desc, spec, i, name_ptr, name_len);
     }
-    i = LOAD_ATOM(spec, i, am_undefined);
+    i = http_load_string(desc, spec, i, oname_ptr, oname_len);
     i = http_load_string(desc, spec, i, value_ptr, value_len);
     i = LOAD_TUPLE(spec, i, 5);
```
garazdawi: The `ASSERT(i <= 26)` a couple of lines down needs to be updated.

garazdawi: Also the `spec` buffer needs to be adjusted to fit the new data.

matyasmarkovics: @garazdawi - In my understanding the new `26` should be `27`. Previously the `undefined` atom needed 2 placeholders, now the `list` or `binary` data need 3 places in the `spec` buffer. Hope this logic is correct.

garazdawi: yes, that seems to be correct.

matyasmarkovics:
Hey @garazdawi , thanks for merging this in `master`.
* Am I correct in assuming that, these changes would be included in the next major release, ie.: OTP-23?
* Do you see any chances to have the changes back-ported to OTP-22? If so, anything I can do to initiate that process? :)

garazdawi:
> Do you see any chances to have the changes back-ported to OTP-22?

There will only be a couple of months difference between the 22.3 release and 23.0, so I think the risk outweighs the rewards. It is already too late for 22.2.

matyasmarkovics:
> > Do you see any chances to have the changes back-ported to OTP-22?
> There will only be a couple of months difference between the 22.3 release and 23.0.
Good to hear that. Thanks!

PR #2467, opened 2019-11-25
uri_string:normalize/2 may return error()

essen:
~~Also remove an unnecessary try/catch in uri_string:normalize/1.~~

### DIFF
--- a/lib/stdlib/src/uri_string.erl
+++ b/lib/stdlib/src/uri_string.erl
@@ -306,7 +306,8 @@ normalize(URIMap) ->
 -spec normalize(URI, Options) -> NormalizedURI when
       URI :: uri_string() | uri_map(),
       Options :: [return_map],
-      NormalizedURI :: uri_string() | uri_map().
+      NormalizedURI :: uri_string() | uri_map()
+                     | error().
 normalize(URIMap, []) when is_map(URIMap) ->
     recompose(normalize_map(URIMap));
 normalize(URIMap, [return_map]) when is_map(URIMap) ->


peterdmv:
Can you please rebase also this PR to maint?

essen:
I did it on top of maint I just messed up when opening the PR.

lib/stdlib/src/uri_string.erl
```diff
@@ -297,16 +297,14 @@
       NormalizedURI :: uri_string()
                      | error().
 normalize(URIMap) ->
-    try normalize(URIMap, [])
-    catch
-        throw:{error, Atom, RestData} -> {error, Atom, RestData}
-    end.
+    normalize(URIMap, []).
```
peterdmv: uri_string_SUITE:normalize_negative was failing in our CI due to this change.
```
{error,invalid_utf8,<<0,0,0,246>>} = uri_string:normalize("//%00%00%00%F6")
```
To make it more complicated I found problems even with the original code. The following examples trigger uncaught exceptions:
```
uri_string:normalize("//%00%00%00%F6", [return_map])
uri_string:normalize(#{host => "%00%00%00%F6",path => []}, [return_map]).
```
So if we keep the scope of this PR, I would propose to keep the try catch.

essen: Seems that `uri_string:normalize("/%XX")` would also trigger a throw.

I'll put the try/catch back.

peterdmv:
Thanks for the contribution!

PR #2473, opened 2019-12-04
Fix stupid limitation for support IPv6

MstrVLT:
fix stupid limitation for support IPv6
Windows XP fully support IPv6

https://www.sites.google.com/site/jrey42/Home/ipv6/winxp

`Install base IPv6 on XP support:
From command line (netsh int ipv6 install - note: is preferred to not use ipv6 commands without netsh)`


jhogberg:
Thanks for the PR, but we don't support Windows XP anymore.

PR #2475, opened 2019-12-09
dbg improvments

garazdawi:
This PR does a couple of things to `dbg`:

* It adds a new tracer sink for `logger`. So it is now possible to do `dbg:tracer(logger)` to start tracing to logger instead. These log events have the domain `[otp,runtime_tools,dbg]` which means that they will not show up in the default log unless the default filters are modified.
* Printing of terms using the device sink (i.e. the one for stdout) has been modified to take the width of the terminal into consideration. Also "smart" new-lines are added to avoid printouts that were terrible to read.
* The default `dbg` process sink has been re-written to use the `logger_olp` module instead of relying on suspending processes. This will make the `dbg` sink kill itself if it is overloaded.

Things left to do:

- [ ] documentation about `logger` integration and options to `logger_olp`
- [ ] testcases for `logger` and `logger_olp` integrations
- [ ] testcases for the printing of terms

I'm publishing this PR before completely done in order to get feedback before finishing the work.

### DIFF
--- a/erts/emulator/beam/erl_nif.c
+++ b/erts/emulator/beam/erl_nif.c
@@ -926,6 +926,13 @@ int enif_send(ErlNifEnv* env, const ErlNifPid* to_pid,
             msgq = msgq->next;
         }
 
+        if (t_p && scheduler > 0 && copy_sz > ERTS_MSG_COPY_WORDS_PER_REDUCTION) {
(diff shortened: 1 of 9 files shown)


garazdawi:
Closing this as it seems I will not have much time to complete it. Maybe someday.

PR #2478, opened 2019-12-10
Optimize handling of empty_list in ping_list()

jpjodoin:
Currently, calling this method with an empty list can take several seconds for the function to return. For system that can be run in both single and multiple instance mode, this add unnecessary delays for the single instance scenario.

### DIFF
--- a/lib/kernel/src/net_adm.erl
+++ b/lib/kernel/src/net_adm.erl
@@ -129,6 +129,9 @@ dns_hostname(Hostname) ->
 
 -spec ping_list([atom()]) -> [atom()].
 
+ping_list([]) ->
+    [];
+
 ping_list(Nodelist) ->
     ok = net_kernel:monitor_nodes(true),
     Sofar = ping_first(Nodelist, nodes()),


jpjodoin:
Not sure why the s390x build fails, but I doubt it's related to this pull request.

garazdawi:
I assume that you are aware that this is an undocumented function and thus should not be used outside kernel? What do you use it for?

jpjodoin:
Oh, actually I didn't realize that. I am using this function as an alternative to lists:map({net_adm, ping}, [], Nodelist).  

The comment over that function pretty much describe my use case:
% A common situation in "life" is to have a configuration file with a list
% of nodes, and then at startup, all nodes in the list are ping'ed

Any particular reason why this function shouldn't be public? Seems like a common use case. I realize this function is way more complex then  lists:map({net_adm, ping}, [], Nodelist).

garazdawi:
Any reason why `net_adm:world_list/1` would not work? Does it create a too large a cluster?

jpjodoin:
I never used world_list, I suppose that would work. You can close this issue.

PR #2479, opened 2019-12-11
ssh: fix typespec in ssh_sftpd

attah:
file_handler is a callback module, tupled optionally with a state.
Fix that modules tend to be atoms.
Introduce ssh_sftpd_file_api:state() :: term() and use that,
but let the default state remain an empty list for compatibility.

### DIFF
--- a/lib/ssh/src/ssh_sftpd.erl
+++ b/lib/ssh/src/ssh_sftpd.erl
@@ -60,14 +60,15 @@
 %%====================================================================
 -spec subsystem_spec(Options) -> Spec when
       Options :: [ {cwd, string()} |
-                   {file_handler, CallbackModule::string()} |
+                   {file_handler, CbMod | {CbMod, FileState}} |
                    {max_files, integer()} |
                    {root, string()} |
                    {sftpd_vsn, integer()}
                  ],
       Spec :: {Name, {CbMod,Options}},
       Name :: string(),
-      CbMod :: atom() .
+      CbMod :: atom(),
+      FileState :: term().
 
 
 subsystem_spec(Options) ->


HansN:
By introducing the type state() in ssh/src/ssh_sftpd_file_api.erl and using it in ssh/src/ssh_sftpd.erl, the generated parts of the documentation in ssh/doc/src/ssh_sftpd.xml gets a documentation link error.
That is because the subsystem_spec in  ssh_sftpd.xml  will have a generated link to ssh_sftpd_file_api.xml for that new type.
Since ssh_sftpd_file_api.xml does not exist, it must be created to introduce the anchor for that link.

attah:
Is that a request for additions, or a note for internal consideration?

We use edoc to not run in to these issues, so i'm not surprised i missed it.
I'm also entirely fine with not introducing that type and do a more bare-minimum correction.
I'd prefer you guys decide, i just want the broken spec gone :)

lib/ssh/src/ssh_sftpd.erl
```diff
@@ -60,14 +60,15 @@
...
+                   {file_handler, CbMod | {CbMod, FileState}} |
                    {max_files, integer()} |
                    {root, string()} |
                    {sftpd_vsn, integer()}
                  ],
       Spec :: {Name, {CbMod,Options}},
       Name :: string(),
-      CbMod :: atom() .
+      CbMod :: atom(),
+      FileState :: ssh_sftpd_file_api:state().
```
HansN: It seems best to just do the minimum error fix, that is instead of ssh_sftpd_file_api:state() just use term().  We are also a bit reluctant to introduce too many types.

HansN:
Thanks for the PR!

PR #2480, opened 2019-12-12
ct: Don't hide error reasons from user

attah:
Depending on what was given as an argument to erlang:error(),
parts would not be printed. Specifically two-tuples of atom and
something, would cause only the atom to be printed.
Also remove now redundant clause of exempting {'ERROR', _} from this.


IngelaAndin:
As this is an obvious improvement and the code is covered by our tests and does not break anything I have accepted it.

PR #2484, opened 2019-12-17
sctp_assoc_change record have incorrect type in documentation

sebastiw:
Everywhere else the error field in the SCTP record is an integer.
The default value for this field is '0'.


RaimoNiskanen:

PR #2487, opened 2019-12-20
ERL-991 Strip control codes from eunit_surefire output

ampleyfly:
Without this, test cases which output control codes, or with names or
descriptions containing control codes, would cause the generated XML files to be
invalid.

Backport from master.


IngelaAndin:
Please rebase your branch on maint and we will include it in the last planned patch release OTP-22.3 
The maint-22 branch is only used for unplanned patches, that are mainly done for serious bugs, vulnerabilities or features considered important by our product manager.

ampleyfly:
So, I rebased my branch on maint and changed the target branch to maint.
I don't know what's up with the CLA check, I have signed the CLA and the check passed the last time.

IngelaAndin:
Thanks, I  have put it in testing. I think that the CLA check might have been failing because of the branch being based on something else then maint, it can make it look like other commits than yours where part of the PR.

IngelaAndin:
Thank you for your contribution. It has passed the test phase and is now merged to maint.

PR #2491, opened 2019-12-27
Allow binaries as parts of logger_formatter template

hauleth:
This small change will simplify integration of Erlang's logger with
Elixir applications where users are more accustomed to usage of binaries
instead of char lists as a "string" type.

### DIFF
--- a/lib/kernel/src/logger_formatter.erl
+++ b/lib/kernel/src/logger_formatter.erl
@@ -532,8 +532,8 @@ check_template([Str|T]) when is_list(Str) ->
         false -> error
     end;
 check_template([Bin|T]) when is_binary(Bin) ->
-    case unicode:character_to_list(Bin) of
-        Str -> check_template([Str|T]);
+    case unicode:characters_to_list(Bin) of
+        Str when is_list(Str) -> check_template([Str|T]);
         _Error -> error
     end;
 check_template([]) ->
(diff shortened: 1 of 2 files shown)


lib/kernel/src/logger_formatter.erl
```diff
@@ -531,6 +531,9 @@ check_template([Str|T]) when is_list(Str) ->
         true -> check_template(T);
         false -> error
     end;
+check_template([Bin|T]) when is_binary(Bin) ->
+    Str = binary:bin_to_list(Bin),
```
ferd: this wouldn't be unicode-safe compared to `unicode:characters_to_list/2-3`.

hauleth: Fixed.

garazdawi:
The `string()` type should probably just be changed to `chardata()` instead of adding `binary()`. However, that means that we should add tests for deep `chardata()` in all the places. i.e.

```
Template = [{metakey,[[$e,<<"x">>,$i,[<<"s">>],":"],metakey],[[[[[[[[["does ",<<"not">>," exist"]]]]]]]]]},[[[<<"more">>]],"deep"]].
```

hauleth:
@garazdawi added such test as well.

lib/kernel/src/logger_formatter.erl
```diff
@@ -35,7 +35,7 @@
                     template        => template(),
                     time_designator => byte(),
                     time_offset     => integer() | [byte()]}.
--type template() :: [metakey() | {metakey(),template(),template()} | string()].
```
garazdawi: The type should be:

    -type template() :: [metakey() | {metakey(),template(),template()} | unicode:chardata()].

garazdawi:
Better late than never! :) Thanks!

PR #2492, opened 2019-12-28
export uri_string:decode/1

galdor:
In the current state, the uri_string module parses URIs to maps of components
but does not decode those which can be percent-encoded such as the
fragment. It is not a problem with the query since `uri_string:dissect_query/1`
will decode it, but other components need decoding.

Exporting `uri_string:decode/1` solves this problem and lets users decode
components when they need it.

In the future, it could be nice to add `uri_string:parse/2`, the second
argument being a list of options which could include `decode_components`.

### DIFF
--- a/lib/stdlib/doc/src/uri_string.xml
+++ b/lib/stdlib/doc/src/uri_string.xml
@@ -84,6 +84,9 @@
       <item>Dissecting form-urlencoded query strings into a list of key-value pairs<br></br>
       <seealso marker="#dissect_query/1"><c>dissect_query/1</c></seealso>
       </item>
+      <item>Decoding percent-encoded strings or binaries.<br></br>
+      <seealso marker="#decode/1"><c>decode/1</c></seealso>
+      </item>
     </list>
     <p>There are four different encodings present during the handling of URIs:</p>
     <list type="bulleted">
@@ -206,6 +209,21 @@
 2> {<<"city">>,<<"東京"/utf8>>}], [{encoding, latin1}]).]]>
 <![CDATA[<<"foo+bar=1&city=%26%2326481%3B%26%2320140%3B">>]]>
 	</pre>
+      </desc>
+    </func>
+
+    <func>
+      <name name="decode" arity="1" />
+      <fsummary>Percent-encoding string or binary decoding.</fsummary>
+      <desc>
+        <p>Decodes a percent-encoded string or binary as defined in section
+        2.1 of <url href="https://www.ietf.org/rfc/rfc3986.txt">RFC
+        3986</url>.</p>
+        <p><em>Example:</em></p>
+        <pre>
+          1> <input><![CDATA[uri_string:decode(<<"hello%20world">>).]]></input>
+          <![CDATA[<<"hello world">>]]>
+	    </pre>
       </desc>
     </func>
 
--- a/lib/stdlib/src/uri_string.erl
+++ b/lib/stdlib/src/uri_string.erl
@@ -226,7 +226,7 @@
 %%-------------------------------------------------------------------------
 %% External API
 %%-------------------------------------------------------------------------
--export([compose_query/1, compose_query/2,
+-export([compose_query/1, compose_query/2, decode/1,
          dissect_query/1, normalize/1, normalize/2, parse/1,
          recompose/1, resolve/2, resolve/3, transcode/2]).
 -export_type([error/0, uri_map/0, uri_string/0]).


galdor:
This is my first PR on OTP, let me know if something is wrong so I can fix it :)

peterdmv:
Isn't it normalization that you are after? `uri_string:decode/1` is an internal function and used for percent-encoding normalization. You can normalize your input URI and return a map using `uri_string:normalize/2`:

```erlang
1> uri_string:normalize("http://localhos%74-%c3%b6rebro:80/a/b/c/./../../g#%c3%b6rebro", [return_map]).
#{fragment => "örebro",host => "localhost-örebro",
  path => "/a/g",scheme => "http"}
```

essen:
The problem is that some elements, like the path, still have %s in the path segments, and that you need to decode them in order to map them onto other things, like file names for example.

By the way I plan to send a separate PR adding an interface for encoding as well. Currently I have code in Cowlib that does it but it might be better to just use `uri_string`. Maybe next week.

peterdmv:
@essen Can you give an example? As far as I see that can only happen if the path segments have percent-encoded triplets that represent reserved characters.

essen:
> @essen Can you give an example? As far as I see that can only happen if the path segments have percent-encoded triplets that represent reserved characters.

Doesn't sound like I need to? You answered the question already. :-) It's typically about the path or the query string when the segment or value in question may contain reserved characters. For example `/platforms/Erlang%2FOTP`.

galdor:
As Essen explained, `normalize` does not correctly decode all components. For example, it will not decode `%26` sequences: `uri_string:normalize("http://example.com#a%26b", [return_map]).` yields `#{fragment => "a%26b",host => "example.com",path => "/", scheme => "http"}`; the fragment is not decoded, and cannot be since `uri_string:decode` is not exported.

In general, from what I see in RFC 3986, all percent-encoded components in a URI can be automatically decoded but the query string (decoding of the path is debatable, but I never saw any case where it would be a problem). Since `parse` does not do it, we need `decode`.

The tricky part with `uri_string` is that it uses the same representation (a map) for both encoded and decoded components. It cannot be fixed without breaking backward compatibility, but at least exporting `decode` makes it useable.

essen:
For a code example, in Cowboy when doing the routing URL decoding is used after splitting up path segments, see https://github.com/ninenines/cowboy/blob/master/src/cowboy_router.erl#L315

peterdmv:
> As Essen explained, `normalize` does not correctly decode all components. For example, it will not decode `%26` sequences: `uri_string:normalize("http://example.com#a%26b", [return_map]).` yields `#{fragment => "a%26b",host => "example.com",path => "/", scheme => "http"}`; the fragment is not decoded, and cannot be since `uri_string:decode` is not exported.

In this specific example the fragment contains a reserved character that is percent-encoded. Characters in the reserved set are protected from normalization (_encoding_/_decoding_) as it would change the meaning of the component. This concept is defined in chapter 2.2:

```
Percent-encoding a reserved character, or decoding a percent-encoded octet
that corresponds to a reserved character, will change how the URI is
interpreted by most applications.  Thus, characters in the reserved
set are protected from normalization and are therefore safe to be
used by scheme-specific and producer-specific algorithms for
delimiting data subcomponents within a URI.
```

> In general, from what I see in RFC 3986, all percent-encoded components in a URI can be automatically decoded but the query string (decoding of the path is debatable, but I never saw any case where it would be a problem). Since `parse` does not do it, we need `decode`.

Even the exported `uri_string:decode` will not _decode_ percent-encoded triplets that represent reserved characters and follows the rules described above. The query string is special as the uri_string module supports form-urlencoded query strings based on [HTML 5.2](https://www.w3.org/TR/html52/)

```erlang
8> uri_string:decode("a%26b").                   
"a%26b"
```
 
> The tricky part with `uri_string` is that it uses the same representation (a map) for both encoded and decoded components. It cannot be fixed without breaking backward compatibility, but at least exporting `decode` makes it useable.

Note that the library is not aware of any application level encoding rules in the URI. It uses two basic types: `uri_string()` represents a valid URI according to RFC3986; `uri_map()` on the other hand is a map of URI components where each component is a `unicode:chardata()`. Moreover the library knows  only non-normalized and normalized URIs. It doesn't implement a generic context free encode/decode for percent-encoding.

What is your use case? Do you need encode/decode functions that can convert arbitrary percent-encoded triplets into the characters and vice versa?

essen:
@peterdmv There's definitely a need for a function that can decode percent-encoded components, that you can integrate in the user application's own parsing code. Whether this should be in `uri_string` or OTP at all is the question. Currently you have to use a third party library for just this one tiny thing, it's not ideal.

About percent-encoding on the other hand, since the code is already there it's basically just a matter of exporting the functions (or adding a single function that lets you choose which component you are percent-encoding, which is what I'm leaning toward). I have two such functions in Cowlib but they're not as specific as `uri_string` and encode more than necessary depending on the component. Being more precise would be great. Also `uri_string` supports different types, not just binary, so it works out better for users.

galdor:
@peterdmv I believe the confusion comes from the fact that `normalize` decodes percent encoded characters, which seems incorrect. According to RFC 3986 6.2.2.2., "These URIs should be normalized by decoding any percent-encoded octet that corresponds to an unreserved character". But `normalize` decodes other sequences, such as `%20` (the space character is not in the set of unreserved character), or `%c3%a9` (multibyte character sequences are also not unreserved).

I now see that `decode` ignores percent-encoded reserved characters, which makes it unusable in practice.

> Moreover the library knows only non-normalized and normalized URIs. It doesn't implement a generic context free encode/decode for percent-encoding.
I'm sorry if I'm missing something, but it seems to me that it makes `uri_string` useless: if I need to parse an URI and use the path or the fragment, I need to be able to decode them to use them. If the fragment is `"a&b"`, it will be encoded as `"a%26b"` in the URI, and must be decoded.

A good example would be the `net/url` [Go library](https://golang.org/pkg/net/url/), which automatically decode host, path and fragment, and provide functions to decode the userinfo and query components.

Now that I have a better understanding of `uri_string`, it feels to me that exporting `decoding` is actually useless, and that I need to write a separate URI library if I need to actually use URI components.

peterdmv:
> @peterdmv There's definitely a need for a function that can decode percent-encoded components, that you can integrate in the user application's own parsing code. Whether this should be in `uri_string` or OTP at all is the question. Currently you have to use a third party library for just this one tiny thing, it's not ideal.

@essen I agree that we need such a function after this discussion.

> @peterdmv I believe the confusion comes from the fact that `normalize` decodes percent encoded characters, which seems incorrect. According to RFC 3986 6.2.2.2., "These URIs should be normalized by decoding any percent-encoded octet that corresponds to an unreserved character". But `normalize` decodes other sequences, such as `%20` (the space character is not in the set of unreserved character), or `%c3%a9` (multibyte character sequences are also not unreserved).

@galdor The default (and also standard) `normalize` operation, where both input and output data is a `uri_string()`, **does not** decode percent-encoded octets that correspond to unreserved characters.

```
1> uri_string:normalize("http://exa%20mple.com#a%c3%b6b").
"http://exa%20mple.com/#a%C3%B6b
```

I think the confusion arises from the fact that the `normalize/2` function has a shortcut with the option `[return_map]`, that skips the last recompose step and returns a `uri_map()` to the user. This `uri_map()` will have unicode characterts and decodes all percent-encoded octets that correspond to unreserved characters. E.g. %20 and %C3%B6. **It is designed to not decode percent-encoded octets that represent reserved characters and let the user/application handle them.** By following these rules we gained two properties in our property tests that we would like to keep:

```erlang
prop_recompose() ->
    ?FORALL(Map, map_no_unicode(),
           Map =:= uri_string:parse(uri_string:recompose(Map))).

prop_normalize() ->
    ?FORALL(Map, map(),
            uri_string:normalize(Map, [return_map]) =:=
                uri_string:normalize(uri_string:parse(uri_string:recompose(Map)),
                                     [return_map])).
```
> I'm sorry if I'm missing something, but it seems to me that it makes `uri_string` useless: if I need to parse an URI and use the path or the fragment, I need to be able to decode them to use them. If the fragment is `"a&b"`, it will be encoded as `"a%26b"` in the URI, and must be decoded.

After some prototyping I realized that it would have been possible to have an API with a different behavior that decodes all percent-encoded octets in the `uri_map()` and does an additional encode step in `recompose`. However, such a change today would not be backward compatible.

As I see it, we need a new option in `normalize/2` that sets a different percent-encoding normalization that decodes all percent-encoded triplets. At the same time a new `recompose/2` function shall also be added with the same not yet defined option in order to be able to convert the `uri_map()` into a standard `uri_string()`. I think it is also possible to define a property to do property tests on this solution.

> Now that I have a better understanding of `uri_string`, it feels to me that exporting `decoding` is actually useless, and that I need to write a separate URI library if I need to actually use URI components.

I agree on that exporting the `decode` function does not solve your original problem.

essen:
> @galdor The default (and also standard) `normalize` operation, where both input and output data is a `uri_string()`, **does not** decode percent-encoded octets that correspond to unreserved characters.

You mean it does not decode percent-encoded octets that correspond to **reserved** characters, correct? Because from a quick testing I see unreserved characters decoded via normalize, regardless of `return_map` usage. And since it gets decoded there's no way it'll know to encode it again from the map.

``` erlang
8> uri_string:normalize("http://exa%20mple.com#a%c3%b6b%4d", [return_map]).           #{fragment => "aöbM",host => "exa mple.com",path => "/",
  scheme => "http"}
9> uri_string:normalize(uri_string:parse("http://exa%20mple.com#a%c3%b6b%4d"), [return_map]).
#{fragment => "aöbM",host => "exa mple.com",path => "/",
  scheme => "http"}
```

> After some prototyping I realized that it would have been possible to have an API with a different behavior that decodes all percent-encoded octets in the `uri_map()` and does an additional encode step in `recompose`. However, such a change today would not be backward compatible.

I'm not terribly convinced about this? What would you do for the path, for example? A list of segments? Hosts? List as well? It might be a better idea to provide separate functions to do this extra decoding instead, this way you can use them on the parts you need to decode. I think it would be pretty rare to need this for everything.

In future Cowboy versions I am considering switching to URI Templates for the router, which means I would no longer need this for routing, just the normalized URI. But users might need them or the urldecode.

peterdmv:
> > @galdor The default (and also standard) `normalize` operation, where both input and output data is a `uri_string()`, **does not** decode percent-encoded octets that correspond to unreserved characters.
> 
> You mean it does not decode percent-encoded octets that correspond to **reserved** characters, correct? Because from a quick testing I see unreserved characters decoded via normalize, regardless of `return_map` usage. And since it gets decoded there's no way it'll know to encode it again from the map.
> 
Yes, you are right, I meant **reserved** characters.

> > After some prototyping I realized that it would have been possible to have an API with a different behavior that decodes all percent-encoded octets in the `uri_map()` and does an additional encode step in `recompose`. However, such a change today would not be backward compatible.
> 
> I'm not terribly convinced about this? What would you do for the path, for example? A list of segments? Hosts? List as well? It might be a better idea to provide separate functions to do this extra decoding instead, this way you can use them on the parts you need to decode. I think it would be pretty rare to need this for everything.
> 
I agree. I think that all those percent-encoded octets that represent characters in the reserved set can only be handled by the application. If an application already has the parsed components, it can apply its application specific decoding functions on those components.

The question is, if it's worth to add an API that allows users to apply application specific decode functions on the parsed URI components. It would clutter the current API, so maybe it's more appropriate to have such functions in a separate library, and keep uri_string simple, implementing only the standard.

essen:
I think a good start would be to explicitly say which characters remain percent encoded in the documentation. This would make implementing such functions obvious (for example split on `/` and decode `%2F`). Similar to the last paragraph of http://erlang.org/doc/man/uri_string.html#compose_query-2

In that case there is no need for an uncompromising urldecode function, since the characters that actually need decoding would be very few (from what I understand, path would ONLY have `%2F` remaining, and other components would have less than a handful as well).

peterdmv:
I created an internal ticket to improve the documentation.

PR #2496, opened 2020-01-06
Do not check Type/Proto combination

hauleth:
Some OSes (for example Linux) allow "non-standard" pairs and we should just pass them as is to the BIFs. That will make OS test whether given combination is allowed or not.

### DIFF
--- a/erts/preloaded/src/socket.erl
+++ b/erts/preloaded/src/socket.erl
@@ -1192,8 +1192,8 @@ open(Domain, Type, Protocol, Extra) when is_map(Extra) ->
     try
         begin
             EDomain   = enc_domain(Domain),
-            EType     = enc_type(Domain, Type),
-            EProtocol = enc_protocol(Type, Protocol),
+            EType     = enc_type(Type),
+            EProtocol = enc_protocol(Protocol),
             case nif_open(EDomain, EType, EProtocol, Extra) of
                 {ok, SockRef} ->
                     Socket = #socket{ref = SockRef},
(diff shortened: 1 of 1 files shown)


erts/preloaded/src/socket.erl
```diff
@@ -3982,8 +3980,8 @@ invalid_type(Type) ->
       Type  :: term(),
       Proto :: term().
 
-invalid_protocol(Type, Proto) ->
-    error({invalid_protocol, {Type, Proto}}).
+invalid_protocol(Proto) ->
+    error({invalid_protocol, Proto}).
```
RaimoNiskanen: The spec must be changed too

erts/preloaded/src/socket.erl
```diff
@@ -2787,22 +2787,20 @@ enc_type(_, raw)       -> ?SOCKET_TYPE_RAW;
...
+enc_protocol(default) -> ?SOCKET_PROTOCOL_DEFAULT;
+enc_protocol(ip)      -> ?SOCKET_PROTOCOL_IP;
+enc_protocol(tcp)     -> ?SOCKET_PROTOCOL_TCP;
+enc_protocol(udp)     -> ?SOCKET_PROTOCOL_UDP;
+enc_protocol(sctp)    -> ?SOCKET_PROTOCOL_SCTP;
+enc_protocol(icmp)    -> ?SOCKET_PROTOCOL_ICMP;
+enc_protocol(igmp)    -> ?SOCKET_PROTOCOL_IGMP;
+enc_protocol({raw, P} = RAW) when is_integer(P) -> RAW;
+enc_protocol(Proto) ->
+    invalid_protocol(Proto).
```
RaimoNiskanen: There is a call to `enc_protocol(Type, Protocol)` in `open/4` that needs to be changed too.  This does not compile.

erts/preloaded/src/socket.erl
```diff
@@ -2787,22 +2787,20 @@ enc_type(_, raw)       -> ?SOCKET_TYPE_RAW;
 enc_type(_, seqpacket) -> ?SOCKET_TYPE_SEQPACKET;
 enc_type(_, Type)      -> invalid_type(Type).
```
RaimoNiskanen: `enc_type/2` could also be rewritten into arity 1 for symmetry, and the call to it in `open/4`

RaimoNiskanen:
We would like this change on `maint`, so can you please rebase your branch?
Please wait a few hours for us to merge a biggie affecting `socket.erl` first.  I will post here when that is done...

RaimoNiskanen:
The merge is done.  You can now do the rebase...

erts/preloaded/src/socket.erl
hauleth: Which specs?

hauleth:
@RaimoNiskanen done.

erts/preloaded/src/socket.erl
RaimoNiskanen: The spec for this function: `invalid_protocol(Proto)`. The spec 4 lines up is for `invalid_protocol(Type, Proto)` which is a now nonexistent function.

RaimoNiskanen:
This branch (f420b19), with corrected spec for `socket:invalid_protocol/1`, is now running in our daily build system

RaimoNiskanen:

hauleth:
@RaimoNiskanen I need it as I want to provide library for integration with [systemd](https://github.com/hauleth/erlang-systemd) socket activation, which requires me to open some existing sockets. That was part of work on such library.

RaimoNiskanen:
Ok. Cool!

PR #2498, opened 2020-01-07
stdlib: add string:next_lexeme/2

lego12239:
Get a string and separators as input and return {lexeme, rest} tuple.
Suitable for a sequential parsing of a string.

This is useful for parsing the strings like:

HEADER SEPARATOR VALUE
KEY SEPARATOR VALUE

Where VALUE may contains SEPARATOR characters and there isn't good to touch it.

### DIFF
--- a/lib/stdlib/doc/src/string.xml
+++ b/lib/stdlib/doc/src/string.xml
@@ -287,6 +287,24 @@ true</pre>
     </func>
 
     <func>
+      <name name="next_lexeme" arity="2" since="OTP 20.0"/>
+      <fsummary>Get next lexeme from string.</fsummary>
+      <desc>
+	<p>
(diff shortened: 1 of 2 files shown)


dgud:
Seems like useful function to me (without discussing this anyone yet), 
but please extend test suite with tests.
There are som obvious problems with the code, the spec is wrong and should return {String, String}, and also the corner cases where there is no separator returns a list.

dgud:
ping

lego12239:
Sorry. Have no time for this now :-(. But i made a better (imho) implementation of next_lexeme:

```
next_lexeme(Str, Sep) ->
        {_, Str_trimmed} = string:take(Str, Sep),
        string:take(Str_trimmed, Sep, true).
```

Simpler and has not bigger time than the first version.

IngelaAndin:
Welcome to reopen if you find the time to work on it.

PR #2502, opened 2020-01-10
Add support for utf8 atoms in enif_get_atom/enif_get_atom_length

define-null:
Hi this little PR adds support for utf8 atoms in enif_get_atom/enif_get_atom_length functions.

### DIFF
--- a/erts/emulator/beam/erl_nif.c
+++ b/erts/emulator/beam/erl_nif.c
@@ -1588,27 +1588,40 @@ int enif_has_pending_exception(ErlNifEnv* env, ERL_NIF_TERM* reason)
 }
 
 int enif_get_atom(ErlNifEnv* env, Eterm atom, char* buf, unsigned len,
-		  ErlNifCharEncoding encoding)
+                  ErlNifCharEncoding encoding)
 {
     Atom* ap;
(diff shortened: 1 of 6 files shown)


sverker:
Nice initiative!
But to be released I think we want all functions receiving enum `ErlNifCharEncoding` to also accept the new `ERL_NIF_UTF8`.
```
enif_make_existing_atom
enif_make_existing_atom_len
enif_get_string
enif_make_string
enif_make_string_len
```

define-null:
Can we consider them to be undocumented for now, or it's more or less either support for all or nothing ?)

sverker:
I would prefer a full documented implementation released in a major release.

A reason for this comes from the fact that the existing functions do not reject invalid encoding values (apart from ASSERTS only executed in debug build). If someone builds against a new erl_nif.h with ERL_NIF_UTF8, but the NIF module is later loaded into an older VM without this support, strange behavior may arise due to wrong character encodings. To protect against this we can increase the interface version number which will prevent a new lib from being loaded into an old VM. But it doesn't feel right to impose such a limitation for an undocumented feature.

rickard-green:
Please make a new PR when/if fully implemented

PR #2506, opened 2020-01-16
Optimize filelib:wildcard/2 with double star patterns

josevalim:
The previous implementation would execute file:list_dir/2
twice for every file found in a double star (**) pattern.
This patch changes it so it only calls it once per file,
literally cutting the traversal time in half.

### DIFF
--- a/lib/stdlib/src/filelib.erl
+++ b/lib/stdlib/src/filelib.erl
@@ -333,6 +333,7 @@ match_part([_|_], []) ->
     false.
 
 will_always_match([accept]) -> true;
+will_always_match([double_star]) -> true;
 will_always_match(_) -> false.
 
 prepare_base(Base0) ->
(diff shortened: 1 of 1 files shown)


bjorng:
I have pushed a fixup commit with the following changes:

* Following the convention in the rest of the file to end the incoming version of a variable with a zero (i.e. I change `Result` to `Result0`).

* Reindented  your code with Erlang mode in Emacs.

If you agree with my changes, please squash the commits and force push.

josevalim:
@bjorng squashed and force pushed.

bjorng:
@josevalim Thanks!

michaelst:
It looks like this fix was included in 22.2.5, however, when I build the docker images with 22.2.5 there doesn't seem to be any change in the performance of this function. Is there possibly something I am missing in building the elixir container from Erlang:22.2.5?

josevalim:
Hi @michaelst! Looking at the commits in the 22 branch, I don't see this particular change. Were you able to find it?

michaelst:
Oh I read that wrong, it is showing up in the commits to master since that tag. Sorry about that.

PR #2509, opened 2020-01-17
Add s390x support in travis build

sangitanalkar:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,27 @@ matrix:
             - xsltproc
             - libxml2-utils
 
+    - env: Linux-s390x-SmokeTest
+      arch: s390x
+      script:
+        - ./scripts/build-otp
+        - ./otp_build tests
+        - ./scripts/run-smoke-tests
+      addons:
+        apt:
+          update: true
+          packages:
+            - autoconf
+            - libncurses-dev
+            - build-essential
+            - libssl-dev
+            - libwxgtk3.0-dev
+            - libglu1-mesa-dev
+            - default-jdk
+            - g++
+            - xsltproc
+            - libxml2-utils
+
     - env: Linux64Docbuild
       script:
         - ./scripts/build-otp docs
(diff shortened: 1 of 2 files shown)


garazdawi:
Thanks! Could you please include the reason for the change in the test case in the commit message?

sangitanalkar:
Hi @garazdawi , I have added the reason for the change in the test case in the commit message.
Please let me know if anything else is needed from my side.

garazdawi:
@sangitanalkar https://travis-ci.org/erlang/otp/jobs/642813859 the build stopped working again...

sangitanalkar:
Hi @garazdawi I tried running the erlang/otp travis build  and it passed ([travis_log_link](https://travis-ci.com/linux-on-ibm-z/otp/jobs/281442606))

sangitanalkar:
Hi @garazdawi ,  I see s390x support for travis is disabled in commit at [link](https://github.com/erlang/otp/commit/3c8e73d26a5abf7d78c255669566151c99eb53b2#diff-354f30a63fb0907d4ad57269548329e3). May I please know if there is any specific reason in doing so?

garazdawi:
It still keeps failing sporadically. Maybe 1 in 5 builds fail, so I removed it again.

sangitanalkar:
Thanks for letting us know. 
Travis-CI has the option of `allow_failures` ([link](https://docs.travis-ci.com/user/customizing-the-build/#jobs-that-are-allowed-to-fail)), which allows specific jobs to fail without causing entire build to fail. We can add s390x job in `allow_failures` since it fails intermittently. Please let us know if that can be done.

garazdawi:
I'm not sure I see the point of having a check that does not fail anything.

sangitanalkar:
Yes, thought of adding s390x job to `allow_failures` until travis build becomes stable since the issue seems to be related to travis environment and there is no build failure as such. Since travis is moving from travis-ci.org to travis-ci.com, we can check after the migration.

garazdawi:
Since we cannot seem to get a stable build for it I would need to see a long streak of s390x working on travis for Erlang/OTP in order to enable it again.

PR #2511, opened 2020-01-18
crypto: fix ng mem leak

christianjgreen:
This PR fixes potential memory leaks involving calls to the ng nifs.
Following along with the original PR in this file: https://github.com/erlang/otp/commit/98feb2577dc64a622f51dac6562135aa4c60274c

### DIFF
--- a/lib/crypto/c_src/api_ng.c
+++ b/lib/crypto/c_src/api_ng.c
@@ -261,7 +261,7 @@ static int get_update_args(ErlNifEnv* env,
     if (!enif_inspect_binary(env, indata_arg, &in_data_bin) )
         {
             *return_term = EXCP_BADARG(env, "Bad 2:nd arg");
-            goto err;
+            goto err0;
         }
 
     ASSERT(in_data_bin.size <= INT_MAX);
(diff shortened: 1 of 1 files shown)


lib/crypto/c_src/api_ng.c
```diff
@@ -820,14 +822,15 @@ ERL_NIF_TERM ng_crypto_one_time(ErlNifEnv* env, int argc, const ERL_NIF_TERM arg
...
+                    goto err_realloc;
                 }
             memcpy(out_data_bin.data + first_part_size, final_data_bin.data, (size_t)final_data_bin.size);
         }
 
     /* Exit here */
     ret = enif_make_binary(env,&out_data_bin);
-
+ err_realloc:
+    enif_release_binary(&out_data_bin);
```
HansN: This will also release the ErlNifBinary that was made read-only in line 831.
If the reallocation of out_data_bin failed in line 822, the ErlNifBinary is still read-only and will be removed by GC.
So this enif_release_binary is not necessary in either case.

christianjgreen: Looks like I missed that part! Thanks for the review :)

HansN:
Good catch!
However, I have already changed ng_crypto_one_time but not pushed it yet, so please remove both changes in that function (line 825 and 832-833)
If you also rebase the resulting PR onto maint, you have made a needed fix in OTP-22 also!

christianjgreen:
> If you also rebase the resulting PR onto maint, you have made a needed fix in OTP-22 also!

Thanks for the review! I've reverted the changes on lines 825 and 832-833. As for OTP-22, would you like the change in a secondary PR?

HansN:
It is enough if you make one PR on maint (= coming OTP-22.3) because everything we merge into maint is also merged into master (= coming OTP-23), but not the reverse.

HansN:
Thanks for the PR!

PR #2512, opened 2020-01-20
kernel: allow opening unbinded connections

hauleth:
This is useful for socket-activation features where the supervisor daemon (like `systemd`) or other daemon (like `xinitd`) pass aready opened and bounded file descriptor to the system. This allows to keep sockets opened between application restarts.

- [x] UDP
- [ ] TCP
- [ ] SCTP

### DIFF
--- a/lib/kernel/src/inet6_tcp.erl
+++ b/lib/kernel/src/inet6_tcp.erl
@@ -137,30 +137,42 @@ do_connect(Addr = {A,B,C,D,E,F,G,H}, Port, Opts, Time)
 	{ok, _} -> exit(badarg)
     end.
 
-%% 
+%%
 %% Listen
 %%
 listen(Port, Opts) ->
     case inet:listen_options([{port,Port} | Opts], ?MODULE) of
-	{error, Reason} -> exit(Reason);
-	{ok,
-	 #listen_opts{
-	    fd = Fd,
-	    ifaddr = BAddr = {A,B,C,D,E,F,G,H},
-	    port = BPort,
-	    opts = SockOpts} = R}
-	when ?ip6(A,B,C,D,E,F,G,H), ?port(BPort) ->
-	    case inet:open(
-		   Fd, BAddr, BPort, SockOpts,
-		   ?PROTO, ?FAMILY, ?TYPE, ?MODULE) of
-		{ok, S} ->
-		    case prim_inet:listen(S, R#listen_opts.backlog) of
-			ok -> {ok, S};
-			Error -> prim_inet:close(S), Error
-		    end;
-		Error -> Error
-	    end;
-	{ok, _} -> exit(badarg)
+        {error, Reason} -> exit(Reason);
+        {ok,
+         #listen_opts{
+            fd = Fd,
+            ifaddr = BAddr = {A,B,C,D,E,F,G,H},
+            port = BPort,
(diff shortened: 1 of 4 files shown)


bmk:
The change (assuming I found it) was not a problem, but it was
well hidden in a restructure that I did not really see a need for.
It just made it more difficult to see what was actually changed.
Please redo, with only the added feature (R#udp_opts{port = 0}).

hauleth:
It is not about allowing `port = 0` but `ifaddr = undefined` which will make `inet_udp` to skip call to `bind`. So it still need some changes that would needlessly duplicate most of the code in such functions.

bmk:
Hi,

Yes I know, I was just trying to point to where thought the change was.

My point was that the restructure made it difficult to see what was
the substantial addition and not just the restructure.
Here is (basically) the old code *with* your addition:

```
open(Port, Opts) ->
    case inet:udp_options([{port,Port} | Opts], ?MODULE) of
	{error, Reason} ->
	    exit(Reason);
	{ok, #udp_opts{fd     = Fd,
		       ifaddr = BAddr = {A,B,C,D,E,F,G,H},
		       port   = BPort,
		       opts   = SockOpts}}
	  when ?ip6(A,B,C,D,E,F,G,H), ?port(BPort) ->
	    inet:open(Fd, BAddr, BPort, SockOpts,
		      ?PROTO, ?FAMILY, ?TYPE, ?MODULE);
	{ok, #udp_opts{fd     = Fd,
		       ifaddr = BAddr = undefined,
		       opts   = SockOpts}} ->
	    inet:open(Fd, BAddr, 0, SockOpts,
		      ?PROTO, ?FAMILY, ?TYPE, ?MODULE);
	{ok, _} ->
	    exit(badarg)
    end.

```
It has the added benefit that its easy for us to identify the addition.

hauleth:
@bmk, ok I can use that change and move refactoring to separate PR. The question left is whether this will be needed when the `socket` integration lands, because if there would be a way to create `gen_tcp` from the `socket` then it would be less of the problem.

bmk:
_Eventually_ `gen_tcp` (and `gen_udp`) based on `socket` will be feature complete.
But when that will happen I don't know (we hope `gen_tcp` on `socket` will be included in OTP 23,
but still experimental).
We have not even started on `gen_udp`...
Also, its safe to assume that many/most existing implementations using `gen_udp`/`gen_tcp`
will take some time to port (even to `socket`-based `gen_udp`/`gen_tcp`). So, I am guessing
that the inet-variants has to (co-) exist for some time...

bmk:
Sorry for the mess.
I pushed to your repo instead of mine (and then removed the commit)...
I made an update to your branch, because there where
two undefined variables (BAddr). I pushed the branch to my public git repo:

(https://github.com/bmk/otp)

The branch has the same name as yours. You can copy from that branch.

bmk:
Even with my fixes, a couple of DTLS test cases(~10) in the SSL application fails when 
this branch is added to the testing.
You will have to test this further on your end and resubmit.

Regards,
   /BMK

IngelaAndin:
ping

hauleth:
I haven’t worked on it in some time, so I would need to update it a little. I will try to do so when I will have some time to spare.

IngelaAndin:
You are welcome to reopen when you have time fix it so that it works.

PR #2514, opened 2020-01-21
Fix compile on systems without getifaddrs, add PASE getifaddrs

NattyNarwhal:
* There was a typo in the error declaration, which lacked the e in enotsup.

 * PASE has a getifaddrs, but it has a different name because AIX lacks the API. Handle this with some light ifdef.

Example:

```
1> inet:getifaddrs().
{ok,[{"127.0.0.1",
      [{flags,[up,loopback,running,multicast]},
       {addr,{127,0,0,1}},
       {netmask,{255,0,0,0}}]},
     {"<redacted>",
      [{flags,[up,broadcast,running,multicast]},
       {addr,{<redacted>}},
       {netmask,{255,255,255,0}},
       {broadaddr,{<redacted>}}]},
     {"<redacted>",
      [{flags,[up,broadcast,running,multicast]},
       {addr,{<redacted>}},
       {netmask,{255,255,252,0}},
       {broadaddr,{<redacted>}}]}]}
2> os:type().
{unix,os400}
```

### DIFF
--- a/erts/emulator/nifs/common/prim_net_nif.c
+++ b/erts/emulator/nifs/common/prim_net_nif.c
@@ -65,6 +65,14 @@
 
 #ifdef HAVE_IFADDRS_H
 #include <ifaddrs.h>
+#elif defined(__PASE__)
+/* PASE has this, but under a different name because AIX doesn't have it. */
+#include <as400_protos.h>
+/*
+ * We don't redefine the function names because they're used in other
+ * contexts, but the struct is safe to rename.
+ */
+#define ifaddrs ifaddrs_pase
 #endif
 
 #ifdef HAVE_NETPACKET_PACKET_H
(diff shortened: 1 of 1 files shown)


bmk:
We know what AIX is (although we currently have no such machine),
but we have no idea what PASE is. Could you please explain.

NattyNarwhal:
PASE is the AIX compatibility layer (syscall translation, etc.) for IBM i (perhaps better known as OS/400), analogous to WSL providing a Linux compatibility layer on Windows. It's mostly compatible, but has limitations to be worked around and additional features AIX does not.

bmk:
I seem to remember AS400...
Anyway, where does the __PASE__ come from? You have not done any
changes to configure. Is it defined in the OS/400 version of ifaddrs.h?

NattyNarwhal:
It comes from the default GCC specs from IBM’s packages for PASE.

bmk:
Ok, so it is defined in one of the "system" include files?

NattyNarwhal:
#response_container_BBPPID{font-family: initial; font-size:initial; color: initial;} It should be one of those "ambient" options that GCC sets even when not including anything. (Before, this wasn't defined, making the default includes identical to AIX. Now it is. so it should be used when needing to differentiate.)                                                                                                                                                                           From: notifications@github.comSent: January 30, 2020 07:21To: otp@noreply.github.comReply to: reply@reply.github.comCc: calvin@cmpct.info; author@noreply.github.comSubject: Re: [erlang/otp] Fix compile on systems without getifaddrs, add PASE getifaddrs (#2514)  Ok, so it is defined in one of the "system" include files?

bmk:
Funny thing. I was trying to figure out where in the code the we generated the "redacted"
addresses, when I realized that, that was of course your edit of the printout :)

PR #2515, opened 2020-01-22
ssl: Prioritise rsa_suites above rc4 and des

jwheare:
https://bugs.erlang.org/browse/ERL-1145

> @IngelaAndin: OpenSSL considers  RC4 and DES legacy and not default supported. They still support RSA. The not anonymous  PSK and SRP are prioritized over RSA.   So  you are welcome to make a PR that prioritizes  RSA over DES and RC4  in "all-list" and we will accept it.

### DIFF
--- a/lib/ssl/src/ssl_cipher.erl
+++ b/lib/ssl/src/ssl_cipher.erl
@@ -290,8 +290,8 @@ all_suites({3, _} = Version) ->
 	++ psk_suites(Version)
 	++ srp_suites(Version)
         ++ rsa_suites(Version)
-        ++ rc4_suites(Version)
-        ++ des_suites(Version);
+        ++ des_suites(Version)
+        ++ rc4_suites(Version);
 
 all_suites(Version) ->
     dtls_v1:all_suites(Version).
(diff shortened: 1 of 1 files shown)


IngelaAndin:
I was just thinking that while you are at it would make sense to swap places of DES and RC4 too,
so that it becomes RSA, DES, RC4.  Please.

jwheare:
Done.

IngelaAndin:
Thanks, I will put it in testing now and we can merge tomorrow when confirmed that it does not break anything. Not that I think it will but it is a good principal to always test!

IngelaAndin:
By the way are you sure you only want this PR for master and not for maint?

jwheare:
Oh I thought I had based it on maint. For our own use case we are no longer using this full list of ciphers, so it's not an urgent patch for us. And it might be an unexpected change of behaviour for some people so maybe master is better anyway? It's your call.

IngelaAndin:
Hopefully users do not use the all list for configuring their ciphers. I will clarify this in the documentation related to ERL-1144 on maint. And the rsa-key exchange suites creates a vulnerability if present at all.  So plague or cholera. I will  just merge it to master.

PR #2517, opened 2020-01-22
`init:get_argument(home).` uses `$HOME`

nathanl:
Demo:

```
HOME="silly_value" /path/to/erl
> init:get_argument(home).
{ok,[["silly_value"]]}
```


nathanl:
We needed to be able to control the value this returns on a server, so we had to figure this out.

PR #2518, opened 2020-01-24
ssl: Doc enhancment

IngelaAndin:
Improve documentation of cipher suites handling due to misunderstandings by users.

### DIFF
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -348,20 +348,35 @@
     <datatype>
       <name name="cipher_suites"/>
       <desc>
-	<p>Supported cipher suites. The function
-	<c>cipher_suites/2</c> can be used to find all ciphers that
-	are supported by default. <c>cipher_suites(all, 'tlsv1.2')</c> can be
-	called to find all available cipher suites. Pre-Shared Key
-	(<url href="http://www.ietf.org/rfc/rfc4279.txt">RFC
-	4279</url> and <url
-	href="http://www.ietf.org/rfc/rfc5487.txt">RFC 5487</url>),
-	Secure Remote Password (<url
-	href="http://www.ietf.org/rfc/rfc5054.txt">RFC 5054</url>),
-	RC4, 3DES, DES cipher suites, and anonymous cipher suites only work if
-	explicitly enabled by this option; they are supported/enabled
-	by the peer also.  Anonymous cipher suites are supported for
-	testing purposes only and are not be used when security
-	matters.</p>
+	<p>A list of cipher suites that should be supported</p>
+
+	<p>The function <seealso
+	marker="#cipher_suite-2"> ssl:cipher_suites/2 </seealso>
+	can be used to find all cipher suites that
+	are supported by default and all ciphers suites that may be configured.</p>
+
+	<p>If you compose your own <c>cipher_suites()</c> make
+	sure they are filtered for cryptolib support
+	<seealso
+	    marker="#filter_cipher_suites-2"> ssl:filter_cipher_suites/2 </seealso>
+	Additionaly the functions <seealso
+	marker="#append_cipher_suites-2"> ssl:append_cipher_suites/2 </seealso>
+	, <seealso
+	marker="#prepend_cipher_suites-2"> ssl:prepend_cipher_suites/2</seealso>,
+	<seealso marker="#suite_to_str/1">ssl:suite_to_str/1</seealso>,
+	<seealso marker="#str_to_suite/1">ssl:str_to_suite/1</seealso>,
+	and <seealso marker="#suite_to_openssl_str/1">ssl:suite_to_openssl_str/1</seealso>
+	also exist to help creating customized cipher suite lists.</p>
+
+	<note><p>Note that TLS-1.3 and TLS-1.2 cipher suites are not overlapping
+	sets of cipher suites so to support both these versions cipher
+	suites from both versions need to be included. If supporting
+	TLS-1.3 versions prior to TLS-1.2 can not be supported.	</p></note>
+
+	<p>Non default cipher suites including anonymous cipher suites (PRE TLS-1.3) are supported for
+	interop/testing purposes and may be used by adding them to your cipher suite list.
+	Note that they must also be supported/enabled by the peer to actually be used.
+	</p>
       </desc>
     </datatype>
     
@@ -1258,6 +1273,17 @@ fun(srp, Username :: string(), UserState :: term()) ->
       <desc><p>Returns all default or all supported (except anonymous),
       or all anonymous cipher suites for a
       TLS version</p>
+
+      <note><p>The cipher suites returned by this function are the
+      cipher suites that the OTP ssl application can support provided
+      that they are supported by the cryptolib linked with the OTP
+      crypto application. Use <seealso
+      marker="#filter_cipher_suites-2"> ssl:filter_cipher_suites(Suites,
+      []).</seealso> to filter the list for the current
+      cryptolib. Note that cipher suites may be filtered out because
+      they are too old or too new depending on the
+      cryptolib</p></note>
+
     </desc>
     </func>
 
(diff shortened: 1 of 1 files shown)


lib/ssl/doc/src/ssl.xml
```diff
@@ -348,20 +348,36 @@
...
+	are supported by default and all ciphers suites that may be configured.
+
+	<p>If you compose your own cipher suite lits make
+	sure it is filtered for cryptolib support
+	<seealso
+	    marker="#filter_cipher_suite-2"> ssl:filter_cipher_suites/2 </seealso>
+	Additionaly the functions <seealso
+	marker="#append_cipher_suites-2"> ssl:append_cipher_suites/2 </seealso>
+	, <seealso
+	marker="#prepend_cipher_suites-2"> ssl:prepned_cipher_suites/2</seealso>,
```
jwheare: prepned -> prepend

lib/ssl/doc/src/ssl.xml
```diff
@@ -348,20 +348,36 @@
...
+	<seealso marker="#suite_to_str/1">ssl:suite_to_str/1</seealso>,
+	<seealso marker="#str_to_suite/1">ssl:str_to_suit/1</seealso>,
```
jwheare: str_to_suit -> str_to_suite

lib/ssl/doc/src/ssl.xml
```diff
@@ -348,20 +348,36 @@
...
+	and <seealso marker="#suite_to_openssl_str/1">ssl:suite_to_openssl_str/1</seealso>
+	also exist to help creating customized cipher suite lists.
+	</p>
+
+	<note><p>Note that TLS-1.3 and TLS-1.2 cipher suites are not overlapping
+	sets of cipher suites so to support both these versions cipher
+	suites form both versions needs to be included. If supporting
```
jwheare: "form both versions needs" -> from both versions need"

lib/ssl/doc/src/ssl.xml
```diff
@@ -1424,10 +1451,14 @@ fun(srp, Username :: string(), UserState :: term()) ->
...
-      returns false for any part of the cipher suite. This function
-      also calls default filter functions to make sure the cipher
-      suites are supported by crypto. If no filter function is supplied for some
-      part the default behaviour is fun(Algorithm) -> true.</p>
+      returns false for any part of the cipher suite. If no filter function is supplied for some
+      part the default behaviour is fun(Algorithm) -> true.
+
+      Additionaly this function also calls default filter functions to make sure the cipher
+      suites are supported by the cryptolib used by the OTP crypto application.
+      That is calling ssl:filter_cipher_suites(Suites, []) will be eqvivalent to only
```
jwheare: eqvivalent -> equivalent

IngelaAndin:
Thanks for catching the typos, should probably have proof read it again. Do you think it all is clearer now?

lib/ssl/doc/src/ssl.xml
```diff
@@ -348,20 +348,35 @@
...
+	, <seealso
+	marker="#prepend_cipher_suites-2"> ssl:prepend_cipher_suites/2</seealso>,
+	<seealso marker="#suite_to_str/1">ssl:suite_to_str/1</seealso>,
+	<seealso marker="#str_to_suite/1">ssl:str_to_suite/1</seealso>,
+	and <seealso marker="#suite_to_openssl_str/1">ssl:suite_to_openssl_str/1</seealso>
+	also exist to help creating customized cipher suite lists.</p>
+
+	<note><p>Note that TLS-1.3 and TLS-1.2 cipher suites are not overlapping
+	sets of cipher suites so to support both these versions cipher
+	suites form both versions need to be included. If supporting
```
jwheare: form -> from :)

jwheare:
Yup, much better now.

lib/ssl/doc/src/ssl.xml
```diff
@@ -1424,10 +1450,14 @@ fun(srp, Username :: string(), UserState :: term()) ->
...
-      also calls default filter functions to make sure the cipher
-      suites are supported by crypto. If no filter function is supplied for some
-      part the default behaviour is fun(Algorithm) -> true.</p>
+      returns false for any part of the cipher suite. If no filter function is supplied for some
+      part the default behaviour is fun(Algorithm) -> true.
+
+      Additionaly this function also calls default filter functions to make sure the cipher
+      suites are supported by the cryptolib used by the OTP crypto application.
+      That is calling ssl:filter_cipher_suites(Suites, []) will be equivalent to only
+      applying the filters for cryptolib support.
```
ferd: I would avoid showing one thing (`fun(Algorithm) -> true`) if the paragraph right after contradicts it since it also mentions.

Maybe just go: "for each part of the algorithm that can be filtered, the default behavior is to make sure the cipher suites are supported by the cryptolib used uby the OTP Crypto application"

Actually now that I write that, is cryptolib support _always_ checked regardless? If so it could be made clearer that the user's filters are only more restrictive. Right now it's unclear whether supplying filters removes that "default behavior" because of the usage of "default" in there.

lib/ssl/doc/src/ssl.xml
IngelaAndin: fixed

lib/ssl/doc/src/ssl.xml
```diff
@@ -348,20 +348,35 @@
...
-	testing purposes only and are not be used when security
-	matters.</p>
+	<p>A list of cipher suites that should be supported</p>
+
+	<p>The function <seealso
+	marker="#cipher_suite-2"> ssl:cipher_suites/2 </seealso>
+	can be used to find all cipher suites that
+	are supported by default and all ciphers suites that may be configured.</p>
+
+	<p>If you compose your own cipher suite lits make
```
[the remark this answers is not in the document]
IngelaAndin: Argh (I must be tried today ;)).  I just rephrased this
	<p>If you compose your own <c>cipher_suites()</c> make
	sure they are filtered for cryptolib support ....
	<seealso

lib/ssl/doc/src/ssl.xml
```diff
@@ -348,20 +348,35 @@
...
+	marker="#prepend_cipher_suites-2"> ssl:prepend_cipher_suites/2</seealso>,
+	<seealso marker="#suite_to_str/1">ssl:suite_to_str/1</seealso>,
+	<seealso marker="#str_to_suite/1">ssl:str_to_suite/1</seealso>,
+	and <seealso marker="#suite_to_openssl_str/1">ssl:suite_to_openssl_str/1</seealso>
+	also exist to help creating customized cipher suite lists.</p>
+
+	<note><p>Note that TLS-1.3 and TLS-1.2 cipher suites are not overlapping
+	sets of cipher suites so to support both these versions cipher
+	suites form both versions need to be included. If supporting
+	TLS-1.3 versions prior to TLS-1.2 can not be supported.	</p></note>
```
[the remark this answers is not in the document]
IngelaAndin: Yes TLS-1.3 specifications says that TLS-1.3 servers must support TLS-1.2 clients but obsoletes all prior TLS versions.

lib/ssl/doc/src/ssl.xml
```diff
@@ -1424,10 +1450,14 @@ fun(srp, Username :: string(), UserState :: term()) ->
       <name since="OTP 20.3" name="filter_cipher_suites" arity="2" />
       <fsummary></fsummary>
       <desc><p>Removes cipher suites if any of the filter functions
-      returns false for any part of the cipher suite. This function
-      also calls default filter functions to make sure the cipher
-      suites are supported by crypto. If no filter function is supplied for some
-      part the default behaviour is fun(Algorithm) -> true.</p>
+      returns false for any part of the cipher suite. If no filter function is supplied for some
+      part the default behaviour is fun(Algorithm) -> true.
```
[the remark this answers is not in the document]
IngelaAndin: That is what I mean.  You would write a fun that accepts exchange algorithms, an other that 
accepts bulk cipher algorithms and so on.

lib/ssl/doc/src/ssl.xml
IngelaAndin: Rewrote it a bit!

ferd: regard it -> regards it
Otherwise that's much clearer!

IngelaAndin: fixed

IngelaAndin:
On the subject of typos, we might as well get it as correct as possible. Have created a new branch!

PR #2520, opened 2020-01-28
mnesia_frag: Fix add and create of external_copies

hazsul:
The #cstruct.external_copies param is expected to
be a list of tuples:
{{Alias, Mod}, Nodes}
instead of just Nodes as in disk_copies/disk_only_copies

when mnesia_frag creates new frags with n_external_copies
it puts the Nodes directly in the #cstruct.external_copies
which results in a crash during creation and addition
of table fragments

This fix properly populates the #cstruct.external_copies
when n_external_copies is specified.

Trigger the bug:
- compile https://github.com/hazsul/mnesia-ext-frag-test/blob/master/test_mnesia_be.erl with OTP version before this fix
- Add the module test_mnesia_be to your code path
- Call `test_mnesia_be:register() ` to register Alias test_mnesia_be

- Then use it to create a normal table:
`mnesia:create_table(test_table, [{type, set}, {test_mnesia_be, [node()]}, {attributes,[key,val]}]).`
The table is created properly

- Then create a table with 4 fragments with external_copies:
`mnesia:create_table(test_table_2, [{type, set}, {test_mnesia_be, [node()]}, {attributes,[key,val]},{frag_properties, [{n_fragments, 4}, {node_pool, [node()]}, {n_external_copies, 1}]}]).`

this results in a crash below:
```
{aborted,{function_clause,[{mnesia_schema,'-verify_nodes/1-fun-2-',
                                          [mnesia_frag_crash@kingfisher],
                                          [{file,"mnesia_schema.erl"},{line,1243}]},
                           {lists,foreach,2,[{file,"lists.erl"},{line,1338}]},
                           {mnesia_schema,verify_nodes,1,
                                          [{file,"mnesia_schema.erl"},{line,1242}]},
                           {mnesia_schema,assert_correct_cstruct,1,
                                          [{file,"mnesia_schema.erl"},{line,1103}]},
                           {mnesia_schema,verify_cstruct,1,
                                          [{file,"mnesia_schema.erl"},{line,1019}]},
                           {mnesia_schema,unsafe_make_create_table,1,
                                          [{file,"mnesia_schema.erl"},{line,1552}]},
                           {mnesia_schema,do_create_table_1,1,
                                          [{file,"mnesia_schema.erl"},{line,1543}]},
                           {lists,foreach,2,[{file,"lists.erl"},{line,1338}]}]}}
```
Complete logs from local testing:
[frag_crash.log](https://github.com/erlang/otp/files/4121417/frag_crash.log)

Test the fix:
- compile https://github.com/hazsul/mnesia-ext-frag-test/blob/master/test_mnesia_be.erl with OTP version after this fix
- Repeat the tests above
- The fragmented table is created properly and the properties can be viewed using:
`rp(mnesia:table_info(test_table_2, all)).`

- Logs for fix working:
[frag_fix.log](https://github.com/erlang/otp/files/4121453/frag_fix.log)

### DIFF
--- a/lib/mnesia/src/mnesia_frag.erl
+++ b/lib/mnesia/src/mnesia_frag.erl
@@ -570,10 +570,19 @@ do_expand_cstruct(Cs, FH, N, Pool, NR, ND, NDO, NExt, Mode) ->
 verify_n_fragments(N, Cs, Mode) when is_integer(N), N >= 1 ->
     case Mode of
 	create ->
-	    Cs#cstruct{ram_copies = [],
-		       disc_copies = [],
-		       disc_only_copies = [],
-		       external_copies = []};
+            case Cs#cstruct.external_copies of
+                [{Mod, _Ns}] ->
+                    NExtCopies = [{Mod, []}],
+                    Cs#cstruct{ram_copies = [],
+                               disc_copies = [],
+                               disc_only_copies = [],
+                               external_copies = NExtCopies};
+                _ ->
+                     Cs#cstruct{ram_copies = [],
+                               disc_copies = [],
+                               disc_only_copies = [],
+                               external_copies = []}
+            end;
 	activate  ->
 	    Reason = {combine_error, Cs#cstruct.name, {n_fragments, N}},
 	    mnesia_schema:verify(1, N, Reason),
@@ -668,8 +677,8 @@ set_frag_nodes(NR, ND, NDO, NExt, Cs, [Head | Tail], Acc) when NDO > 0 ->
     {Cs2, Head2} = set_frag_node(Cs, Pos, Head),
     set_frag_nodes(NR, ND, NDO - 1, NExt, Cs2, Tail, [Head2 | Acc]);
 set_frag_nodes(NR, ND, NDO, NExt, Cs, [Head | Tail], Acc) when NExt > 0 ->
-    Pos = #cstruct.external_copies,
-    {Cs2, Head2} = set_frag_node(Cs, Pos, Head),
+    %% Pos = #cstruct.external_copies,
+    {Cs2, Head2} = set_ext_frag_node(Cs, Head),
     set_frag_nodes(NR, ND, NDO, NExt - 1, Cs2, Tail, [Head2 | Acc]);
 set_frag_nodes(0, 0, 0, 0, Cs, RestDist, ModDist) ->
     {Cs, ModDist, RestDist};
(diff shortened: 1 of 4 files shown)


dgud:
Can you please add a test case for this.

hazsul:
A test case (nice_ext) has now been added to mnesia_frag_test.
I have also tried this test on master and it detects the problem there and is fixed with this pull request.

dgud:
Hmm, you are expecting the external copies to be a list of one element, but you could have have multiple different external copy types, if I understand the code correctly.

hazsul:
Thank you @dgud for looking into this and bringing up this question. I agree that if you look at `mnesia:create_table/2` it allows you to create a table with multiple different external_copies spread over different nodes. 

However, when I started digging into the code on mnesia_frag I realized that we wouldnt be able to create multiple types of external copies for a fragmented table with the current setup that mnesia_frag have. The reason for that stems from the way nodes for external_copies are specified which conflicts with the node_pool parameter in the frag_properties. 

I will try to explain how I got to this conclusion.:
- During creation of an unfragmented table, you can use option `{test_mnesia_be,[ListOfNodes]}` and it will create those copies on those node. you can add multiple of these to the create_table I.e. `{test_mnesia_be,[ListOfNodes]}, {test_mnesia_be2, [AnotherListOfNodes]}` and that will work as well.

- In the mnesia_frag, we receive the `{{Alias, Mod}, [ListOfNodes]}` along with frag_properties including node_pool. For all other "copies" (I.e. disk_only_copies/disk_copies) the node_pool which overrides the [ListOfNodes], as is the case with the normal 'disk_copies\ fragments and the fragments should be distributed to node_pool, not the [ListOfNodes]. This opens up the question that how do you ensure correct types of external backends are created on correct nodes. This wouldnt be possible without reworking the complete setup of how the external_copies of a fragmented table are created and that it takes into account not only the node_pool but also the provided [ListOfNodes] with each {Alias,Mod} pair that is received.

So, in the end, this pull fixes the most basic case:
- You can have only one type of external_copies where the fragmentation is done on the same [ListOfNodes] where the external_copies are allowed to be created I.e. that the user adds the same nodes in node_pool as they would in the [ListOfNodes] for the external_copies.
- there is no check to prevent the user from adding different nodes in node_pool and [ListOfNode], however the node_pool nodes are used to "distribute" the external_copies fragments.

I hope with this is a good first step towards adding more support for external backends with fragmented tables in mnesia.

dgud:
Can we change the API to make it work?
Because this feels like we will get a bug report in 6months and we are stuck with a broken api as well.

hazsul:
@dgud I understand your point, I will try to come back with a suggestion for the API for this and if OTP team approves I can try to implement one. Hopefully will get back to you in next couple of days.

hazsul:
So below is a proposal for what can be implemented to be able to support multiple types of external_copies on the different nodes with fragments:

- The n_external_copies parameter is changed from integer to a list of two-tuple `{ExtBE, Int}` where:
   ExtBE = Mnesia external backend
   Int = Number of copies on that ExtBe
- Then in mnesia_frag, keep the allowed list with each ExtBE along with the node_pool and when its time to distribute on external_backends, then for each node chosen the external_backend selected is the type that is allowed on that node. so in principle the API would look something like this:

`mnesia:create_table(test_table_2, [{type, set}, {test_mnesia_be, [node1, node2]}, {test_mnesia_be2, [node3, node4]} {attributes,[key,val]},{frag_properties, [{n_fragments, 4}, {node_pool, [node1, node2, node3, node4, node5, node6]}, {n_external_copies, [{test_mnesia_be,2}, {test_mnesia_be2, 2}]]}]).`

So, this means we want 2 of each test_mnesia_be and test_mnesia_be2 distributed over nodes node1, node2, node3, node4, node5 and node6 while test_mnesia_be is allowed on node1 and node2 and test_mnesia_be2 is allowed on node3 and node4 which results that it will not distribute the external_copies on node5 and node6 as they are not in list for any external_backend.
If you think that is a good way to fix this, I can look into implementing a solution based on this.

dgud:
Sounds like it should work, though it was a long time I looked at mnesia_frag and external_backends 
implementation.
It would be great if you can implement that and I'm sure doing that will reveal if works or not.

dgud:
Closing this, due to inactivity, please open a new PR with the changes requested.

PR #2521, opened 2020-01-29
EEP 52: Allow key and size expressions in map and binary matching

bjorng:
For binary matching, this PR allows the size of a segment be given as a guard expression. For example:

```
example1(<<Size:8,Payload:((Size-1)*8)/binary,Rest/binary>>) ->
   {Payload,Rest}.
```

In the current map matching syntax, the keys in a map pattern must be
a single value or a literal.  That leads to unnatural code if the keys
in a map are complex terms.  For example:

    example2(M, X) ->
        Key = {tag,X},
        #{Key := Value} = M,
        Value.

This PR allows the key in a map pattern to be a guard expression. That will allow the previous example to be written like this:

    example2(M, X) ->
        #{{tag,X} := Value} = M,
        Value.

There is more information in [EEP-52](https://github.com/erlang/eep/blob/master/eeps/eep-0052.md).


vans163:
Hell yes, been waiting for this a long long time (the binary matching part of the PR).

PR #2522, opened 2020-01-30
Stop mangling FQDNs in inet:gethostname()

ulidtko:
Dear Ericsson team! Hallå från Ukraina! :)

I'm well aware that touching code which hasn't seen any change in 10+ years is a bold move. But please consider. Someone on #erlang IRC told me you're working on switching to the *'socket' NIF interface* (honestly, no idea what's that), so things are going to break anyway — thus more documented inet defects & more patches wouldn't harm.

### Bug description ###

We've stumbled into the OTP's implementation for `inet:gethostname()` while solving a 5-second delay during login to RabbitMQ, from any client. RabbitMQ server needs to compute its hostname during AMQP connection opening — and while doing that it [tries to resolve][1] the result of `inet:gethostname()`.

Two factors come together and materialize the bug for us.
1. Our system is using `foobar.cluster-eu3` FQDN-like hostname (uname -n, /proc/sys/kernel/hostname).
2. Hostname **is not** aliased to 127.0.0.1 in /etc/hosts (which is [rather popular][2], but wrong thing to do). Neither is the "short" (stripped) hostname, `foobar`, valid or resolvable.

Now, the bug is in the line I'm proposing to remove in this patch. The line causes `inet:gethostname()` to return `foobar` where it should return `foobar.cluster-eu3`.

As a consequence, RabbitMQ spends 5 seconds trying to resolve `foobar` via the system (Erlang "native") resolver — times out, then immediately succeeds thanks to `search cluster-eu3` in /etc/resolv.conf. Still, 5 second delay has passed, and it's a big deal here.

Furthermore, on the same system we're getting `foobar.cluster-eu3` from `net_adm:localhost()` function — while `inet:gethostname()` returns `foobar`. Observe the inconsistency.

[1]: https://github.com/rabbitmq/rabbitmq-common/blob/v3.8.0/src/rabbit_net.erl#L269-L275
[2]: https://github.com/hashicorp/vagrant/issues/7263#issuecomment-567983125

### Workarounds ###

Everyone's immediate knee-jerk reaction is to suggest putting `foobar` (the "short" hostname) into /etc/hosts. Well, we cannot go this way, since it [breaks other tools][2] we use.

A viable workaround is to plumb together an `erl_hosts` file via ERL_INETRC. Not fun at all, neither reliable — but doable if you control *all of* hostname-changing code on the system.

Another viable workaround is to *vendor* the OTP — patch it, and use a "shop-local" build.

### Change analysis ###

Obviously, this change has the potential to break *bug-compatibility*: it's somewhat likely that there's software which worked by chance (relying on the bug), and it will stop working.

However, this change **cannot affect** any systems which use "short" (non-FQDN) hostnames.

On the other hand — for systems which do use multiple-dotted-component FQDN-like hostnames,
this mangling I'm removing causes a lot of grief, unnecessary /etc/hosts manipulation,
broken `libnss_myhostname` /etc/nsswitch.conf plugin, spurious resolution timeouts, etc etc etc.

The "nohost.nodomain" fallback winks extra irony: we'll mangle your hostname, remove components,
and ALWAYS return you a single short name without dots. Oh, except not always; be ready for "nohost.nodomain".

This splitting & stripping just shouldn't be here, as I see it.

With sincere regards for your Open-Source work
Max, Senior DevOps, Ukraine

### DIFF
--- a/lib/kernel/src/inet.erl
+++ b/lib/kernel/src/inet.erl
@@ -484,8 +484,7 @@ gethostname() ->
 	{ok,U} ->
 	    {ok,Res} = gethostname(U),
 	    inet_udp:close(U),
-	    {Res2,_} = lists:splitwith(fun($.)->false;(_)->true end,Res),
-	    {ok, Res2};
+	    {ok, Res};
 	_ ->
 	    {ok, "nohost.nodomain"}
     end.


vans163:
I doubt this will make it in as it can potentially break a lot of previous deployments.  But a inet:gethostname2() makes sense here with the correct behaviour.  Or you could just write your own at this point and ship it with your app.

"когда я пошел бухать, я уже был пьян" - Зеленский President of Ukraine

ulidtko:
> I doubt this will make it in as it can potentially break a lot of previous deployments.

Well, that assessment doesn't do justice to the **Change analysis** section I done.

A lot of deployments *will not* be broken (namely, those using "short" non-FQDN hostnames) — that can be seen by trivial reasoning.

"Long" hostname deployments can indeed become broken. Although I have trouble contriving scenarios leading to that — I still admit that it's possible in rare circumstances. The prevalent effect would be that various workarounds (fiddling with /etc/hosts and the like) stop being necessary.

> But a inet:gethostname2() makes sense here with the correct behaviour.

Nope, that's useless. Continuing with the RabbitMQ example — to get rid of the bug, not only should I upgrade OTP, but also patch RabbitMQ server to use the "fixed" gethostname2? That's just dumb, sorry.

> Or you could just write your own at this point and ship it with your app.

Nope; same argument as above. Our apps don't deal with Erlang, they deal with an AMQP broker, for which we use RabbitMQ. "Writing our own" would involve patching/vendoring RabbitMQ.

Rather than doing that, a much better plan for us is to insert yet another kludge into our platform (`/etc/rabbitmq/erl_hosts` file, written from every place where hostname gets changed).

That's the impact of the bug: the best way to deal with it (apart from fixing it upstream) is to build yet more workarounds.

RaimoNiskanen:
Hello @ulidtko!

We intend to break as little as possible when introducing the `socket` API, by mostly adding stuff, not changing.

Yes this is old code, so Lord knows how much code worldwide that depends on this behaviour...

The `inet:gethostname/0` function is clearly written to return the short host name, i.e before the first dot.

The `net_adm:localhost/0` function is clearly written to return the "long" host name (FQDN) since it takes the short host name and appends the domain name.  Your patch would probably break this since in your use case it would add the domain to `foobar.cluster-eu3`, which I think would duplicate the subdomain `cluster-eu3`.

On some systems (e.g FreeBSD, OpenBSD) you set the FQDN with `sethostname()`, and get the FQDN from `gethostname()`.  On other systems (e.g Solaris, Linux) you set the short host name with `sethostname()` and get it with `gethostname()`.  I do not recall how this is handled on Windows... The code in `inet:gethostname/0` compensates for this difference by always stripping the domain from the host name so it will always return the short host name.  The code in `net_adm:localhost/0` compensates for this difference by adding the domain to the short host name to not be depending on what was set with `sethostname()` but instead add the domain we managed to figure out (not always successfully) so it will always return the FQDN (best effort).

So blindly changing the behaviour of `inet:gethostname/0` sounds beyond risky.

I agree that `nohost.nodomain` is an ugly wart.  Really unfortunate, or straight up idiotic!

I am not certain you can call the current behaviour a bug, even though it does not suit your needs, and maybe could have been different.  Maybe you need `inet:gethostname([long])` a'la `hostname --long` instead?


But what can we do for you, then?

* Why do you have to set a host name with a dot in it on a Linux system?  It does not sound as the intentions in `hostname(1)`, and `gethostname(2)`.
* Why does your system try to resolve the result of `inet:gethostname/0`
* Why cannot the short host name be valid to resolve?

ulidtko:
@RaimoNiskanen hello, thanks a lot for your response!

> Why do you have to set a host name with a dot in it on a Linux system? It does not sound as the intentions in hostname(1), and gethostname(2).

We ship a platform, a complete OS image, and the end-user is supposed to configure network settings (including hostname) as they see fit. Well yes, oftentimes we ourselves are this end-user (CI/CD); so we configure both short and long hostnames — just to test that everything works in both styles. Our goal is to avoid imposing restrictions on which network architectures our platform can be deployed on.

> Why does your system try to resolve the result of inet:gethostname/0

We use RabbitMQ server as one of platform components, and it just does this. [These lines][1], see?

I don't get why they do this, too — but this is what they do. And it's not exactly easy for us to change that.

In general, resolving own hostname does vaguely make sense; I can be wrong, but that's about the only good way programs can compute "IP address where I'm running on", so that it just works™ with IPv4/v6 and on different base OS's.

[1]: https://github.com/rabbitmq/rabbitmq-common/blob/v3.8.0/src/rabbit_net.erl#L269-L275

> Why cannot the short host name be valid to resolve?

Well it can, sort of, this is currently exactly what we do (because we have to). The problem is, it's done via unreliable trickery, and we can't do any better.

Consider [`libnss_myhostname`][2]. That's a plugin one drops into `/etc/nsswitch.conf` (it's already there by default on many Linux systems), and its sole purpose is to guarantee that system can quickly resolve it's own hostname. This works beautifully well, reliable 0-configuration — except not in case of Erlang + FQDN. The plugin does resolve `A.B`, but is being asked to resolve just `A`.

And once you get into workaround-land, it's only downhill from there. Suppose hypothetically, a rare hash collision occurs, so two nodes appear: `Nf41e129e.cluster-eu3` and `Nf41e129e.cluster-eu4`. Hopefully, they don't have to communicate! Because that'll definitely be screwed up by our "shortened hostname resolution" trickery, no matter how we do it.

[2]: https://www.freedesktop.org/software/systemd/man/nss-myhostname.html

----------------------------------

> Your patch would probably break this since in your use case it would add the domain to foobar.cluster-eu3, which I think would duplicate the subdomain cluster-eu3.

Oh yes, you're absolutely right. Don't merge the patch! :)

This can be corrected though, by aligning the outputs of `inet:gethostname/0` and `net_adm:localhost/0`... but I agree in general, it's best to avoid changing any of this.

Now, with your illuminating explanation on domains, I think my issue is best solved by patching RabbitMQ. It shouldn't be trying to resolve the shortened hostname; either resolve full hostname, or nothing at all.

Question: would it be correct to suggest using `net_adm:localhost/0` instead of `inet:gethostname/0` to RabbitMQ developers? I'm still a bit confused on the difference between `net_adm` and `inet` modules.

> Maybe you need inet:gethostname([long]) a'la hostname --long instead?

This sounds right; yes. That way app developers could see there's a choice, and pick the right option.

Is it difficult to introduce? I mean, API guarantees, compatibility, docs, changelog... I could certainly do another PR adding the function, but not so sure about the rest.

Thanks again for the explanation, it was much needed and helpful.

Please keep this issue open for awhile — I'll need to refer to it when talking to RabbitMQ guys.

RaimoNiskanen:
It is hard or impossible for a host to figure out its own FQDN.

The classical approach is to set a short name without dots, assuming that this name is unique on the host's domain.  Then it depends on the network interface what the FQDN is - it can only be defined from the outside.  If you can reach a machine from anywhere using a name that is a possible FQDN.

The classical approach continues: (Solaris; the Network is the Computer), had NIS on the network and all machines used short names between themselves, on the NIS domain.  So e.g "gethostname -d" returns the short host name (dot) the NIS domain.

Enter DNS.  Now we want the NIS domain to be the same as the DNS domain since NIS never was intended to work much over Internet, possibly between a few NIS domains. A machine knows what its NIS domain is or else it can not talk to the NIS server.

Exit NIS.  We now only have DNS.  Then we can ask our name resolver, it should be able to tell us what our FQDN is...  And this is what RabbitMQ's code does, as does lots of code in Erlang/OTP and other's applications.

The code in `net_adm:localhost/0` does roughly the same, but it picks up stored values instead.  The domain was looked up during Erlang VM boot, and may have failed, so a later lookup of the short host name may have better luck.  That is the rational behind the RabbitMQ code.  But, if you have an /etc/resolv.conf file that gets rewritten later, with the "domain' tag, that may be discovered later since /etc/resolv.conf is monitored, so then the stored domain may get updated, and then `net_adm:localhost/0` may succeed and be better than looking up the host's short name.

This is a mess, and I have no good solution.  Anyone with a half good solution may speak up, please!

The suggestion for `inet:gethostname([long])` was only half serious, since I do not know how to implement it reliably.  It will in some cases have to ask DNS resolvers and then possibly be forced to wait.  There may be clever solutions such as doing all search path queries in parallel, but then it will be funny if you get contradictions after answering the caller.



If you set a short host name, will not `libnss_myhostname` help you to immediately look up your host's own name?

So this is the use case for when you use long host names, right?  Then it may be argued that if a node runs with long host names, then RabbitMQ should use the node name (`net_adm:localhost()`) instead, to follow cluster practices.  If that does not work we could maybe call it a network configuration error, or improve how the domain name is discovered during VM boot, so it works for RabbitMQ, whatever problem their code solves...

michaelklishin:
FWIW it is possible to [disable reverse DNS lookups](https://www.rabbitmq.com/networking.html#dns-reverse-dns-lookups) (also in [rabbitmq.conf.example](https://github.com/rabbitmq/rabbitmq-server/blob/v3.8.x/docs/rabbitmq.conf.example#L56)) in RabbitMQ, which is the feature that ultimately leads to the function linked to in this issue. So maybe that counts for a workaround in your specific case.

RaimoNiskanen:
I'll close this PR.

If you still need to refer to it it should still be reachable via the link https://github.com/erlang/otp/pull/2522

PR #2523, opened 2020-01-31
Allow using custom IO devices in logger_std_h

hauleth:
This is needed for simpler integration with Elixir logger which allows any device to be set as an output device. It also provides greater flexibility when needed to integrate with log catchers.

### DIFF
--- a/lib/kernel/doc/src/logger_std_h.xml
+++ b/lib/kernel/doc/src/logger_std_h.xml
@@ -55,7 +55,7 @@
        is stored in a sub map with the key <c>config</c>, and can contain the
        following parameters:</p>
     <taglist>
-      <tag><marker id="type"/><c>type = standard_io | standard_error | file | {device, atom() | file:io_device()}</c></tag>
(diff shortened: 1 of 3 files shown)


garazdawi:
I think that the type you want to include is `io:device()`. The `file:io_device()` type includes what is returned by `file:open(Path,[raw])` which I don't think will work in this case.

hauleth:
Yeah, that was probably the type I wanted. I will update the data.

lib/kernel/doc/src/logger_std_h.xml
```diff
@@ -55,7 +55,7 @@
        is stored in a sub map with the key <c>config</c>, and can contain the
        following parameters:</p>
     <taglist>
-      <tag><marker id="type"/><c>type = standard_io | standard_error | file</c></tag>
+      <tag><marker id="type"/><c>type = standard_io | standard_error | file | {device, atom() | io:device()}</c></tag>
```
garazdawi: `io:device()` already includes `atom()`, so no need to include it in the list.

hauleth: Now it should be fixed.

PR #2526, opened 2020-02-07
Optimize =:= when comparing maps

bjorng:
Teach `=:=` to immediately return `false` when comparing two maps
of different sizes. That can be much faster for huge maps.

Reported in the elixir forum:

https://elixirforum.com/t/erlang-2-does-not-seem-to-check-map-sizes-before-performing-the-thoroughful-comparison/28910

### DIFF
--- a/erts/emulator/beam/utils.c
+++ b/erts/emulator/beam/utils.c
@@ -2973,10 +2973,14 @@ tailrecur_ne:
 		    bb = hashmap_val(b) + 1;
 		    switch (hdr & _HEADER_MAP_SUBTAG_MASK) {
 		    case HAMT_SUBTAG_HEAD_ARRAY:
+                        if (aa[0] != bb[0])
+                            goto not_equal;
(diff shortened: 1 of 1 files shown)


max-au:
I wonder if it also fixes bug with ETS table match for maps:

Tab = ets:new(tab, [ordered_set, named_table]),
true = ets:insert(tab, {#{}, 1}),
catch ets:select_replace(tab, , [{{#{}, 1}, [], [{const, {#{}, 42}}]}]).

bjorng:
@max-au Have you reported that bug?

I can't see that the optimization in this pull request should fix any bug. It only speeds up the comparison of maps of size 33 or greater.

sverker:
@max-au That does not look like a bug.
The match expression to select_replace must be guaranteed to retain the key element, regardless the content of the table.
The first #{} will match any map as key and the second #{} means an empty map as key.

PR #2531, opened 2020-02-11
gen_statem: change callback module

RaimoNiskanen:
#### Changing the callback module
This is a new feature, proposed for `gen_statem`, that does not exist in e.g `gen_server`.

By returning the action `{change_callback_module, NewModule}` the `gen_statem` engine changes the callback module.  Simple as that!

We have not figured out any hairy implications other than that the *callback mode* has to be read for `NewModule` before calling any callback function in it, which was not very hairy.  Also, changing the callback module has to be regarded as orthogonal to changing the state.  I have furthermore made it not possible to change callback module from a *state enter call* since that would be just too weird.

#### Why, indeed?
The inception is our work on TLS 1.3.  This protocol has evolved so much that the state machines for TLS 1.3 has so little in common with the one for TLS 1.2, so it makes much sense to have a common state machine for the version negotiation, and when that is done switch to a version specific state machine.

This feature makes the code clearer to read since the version specific state machines can be cleansed from irrelevant code.

There might be other applications, but I foresee mostly state machines for protocols with similar properties.

Sure, it is yet another knob, which increases the possibility for confusion, but it is well contained and suited for specific use cases.

#### Alternatives?
I prototyped a state machine wrapper doing the same thing that became 500 lines of mostly boilerplate code and around 80 lines in the dispatch function.

The worst bit with this solution is that it turned out to not be possible to write a generic variant suitable for `stdlib`. It has to be adapted to the *callback mode* of the target callback module, so there needs to be 4 variants: `state_functions`, `handle_event_function,` `[state_functions,state_enter]` and `[handle_event_function,state_enter]`.

And for the SSL application it meant that two wrapper modules would have to be written since one of its state machines use `state_enter` and the other does not.

#### The implementation

Most of the diff is documentation changes, and a test case.  The first commit is preparation that moves reading the *callback mode* to a little bit later in the engine loop, so it happens in the right place for the feature addition.

The actual change is in `gen_statem.erl`, in the second commit (095cd40) and is quite straightforward and small.

### DIFF
--- a/lib/stdlib/doc/src/gen_statem.xml
+++ b/lib/stdlib/doc/src/gen_statem.xml
@@ -4,7 +4,7 @@
 <erlref>
   <header>
     <copyright>
-      <year>2016</year><year>2019</year>
+      <year>2016</year><year>2020</year>
       <holder>Ericsson AB. All Rights Reserved.</holder>
     </copyright>
     <legalnotice>
@@ -99,6 +99,14 @@
 	  </seealso>
 	  were added.
 	</item>
+	<item>
+	  In OTP 22.3 the possibility to change the callback module
+	  with action
+	  <seealso marker="#type-action">
+	    <c>change_callback_module</c>
+	  </seealso>
+	  was added.
+	</item>
       </list>
     </note>
     <p>
@@ -119,6 +127,7 @@
 	Reply from other state than the request, <c>sys</c> traceable
       </item>
       <item>Multiple <c>sys</c> traceable replies</item>
+      <item>Changing the callback module</item>
     </list>
 
 
@@ -129,15 +138,17 @@
     </p>
     <list type="bulleted">
       <item>
-        <p>One for finite-state machines
+        <p>
+	  One for finite-state machines
           (<seealso marker="gen_fsm"><c>gen_fsm</c></seealso> like),
           which requires the state to be an atom and uses that state as
-          the name of the current callback function
+          the name of the current callback function.
         </p>
       </item>
       <item>
-        <p>One without restriction on the state data type
-          that uses one callback function for all states
+        <p>
+	  One that allows the state to be any term and
+          that uses one callback function for all states.
         </p>
       </item>
     </list>
@@ -150,7 +161,7 @@
       </seealso> <c>gen_fsm</c> to <c>gen_statem</c>.
     </p>
     <p>
-      A generic state machine process (<c>gen_statem</c>) implemented
+      A generic state machine server process (<c>gen_statem</c>) implemented
       using this module has a standard set of interface functions
       and includes functionality for tracing and error reporting.
       It also fits into an OTP supervision tree. For more information, see
@@ -626,9 +637,12 @@ handle_event(_, _, State, Data) ->
 	  If the
 	  <seealso marker="#type-callback_mode"><em>callback mode</em></seealso>
 	  is <c>state_functions</c>,
-	  the state must be of this type.
+	  the state must be an atom.
 	  After a <em>state change</em> (<c>NextState =/= State</c>),
 	  all postponed events are retried.
+	  Note that the state <c>terminate</c> is not possible
+	  to use since it would collide with the optional callback function
+	  <seealso marker="#Module:terminate/3"><c>Module:terminate/3</c></seealso>.
 	</p>
       </desc>
     </datatype>
@@ -707,10 +721,9 @@ handle_event(_, _, State, Data) ->
       <name name="callback_mode"/>
       <desc>
 	<p>
-	  The <em>callback mode</em> is selected when starting the
-	  <c>gen_statem</c> and after code change
-	  using the return value from
-	  <seealso marker="#Module:callback_mode/0"><c>Module:callback_mode/0</c></seealso>.
+	  The <em>callback mode</em> is selected
+	  with the return value from
+	  <seealso marker="#Module:callback_mode/0"><c>Module:callback_mode/0</c></seealso>:
 	</p>
 	<taglist>
 	  <tag><c>state_functions</c></tag>
@@ -732,6 +745,16 @@ handle_event(_, _, State, Data) ->
 	    </p>
 	  </item>
 	</taglist>
+	<p>
+	  The function
+	  <seealso marker="#Module:callback_mode/0"><c>Module:callback_mode/0</c></seealso>
+	  is called when starting the <c>gen_statem</c>,
+	  after code change
+	  and after changing the callback module with action
+	  <seealso marker="#type-action"><c>change_callback_module</c></seealso>.
+	  The result is cached for subsequent calls to
+	  <seealso marker="#state callback">state callbacks</seealso>.
+	</p>
       </desc>
     </datatype>
     <datatype>
@@ -1166,6 +1189,46 @@ handle_event(_, _, State, Data) ->
 	      <seealso marker="#type-event_type"><c>internal</c></seealso>
 	      is to be used when you want to reliably distinguish
 	      an event inserted this way from any external event.
+	    </p>
+	  </item>
+	  <tag><c>change_callback_module</c></tag>
+	  <item>
+	    <p>
+	      Changes the callback module to
+	      <c><anno>NewModule</anno></c>
+	      which will be used when calling all subsequent
+	      <seealso marker="#state callback">state callbacks</seealso>.
+	    </p>
+	    <p>
+	      The <c>gen_statem</c> engine will find out the
+	      <seealso marker="#type-callback_mode">
+		<em>callback mode</em>
+	      </seealso>
+	      of <c><anno>NewModule</anno></c> by calling
+	      <seealso marker="#Module:callback_mode/0">
+		<c>NewModule:callback_mode/0</c>
+	      </seealso>
+	      before the next
+	      <seealso marker="#state callback">state callback</seealso>.
+	    </p>
+	    <p>
+	      Changing the callback module does not affect the
+	      <em>state transition</em> in any way,
+	      it only changes which module that handles the events.
+	      Be aware that all relevant callback functions in
+	      <c><anno>NewModule</anno></c>
+	      such as the
+	      <seealso marker="#state callback">state callback</seealso>,
+	      <seealso marker="#Module:code_change/4"><c><anno>NewModule</anno>:code_change/4</c></seealso>,
+	      <seealso marker="#Module:format_status/2">
+		<c><anno>NewModule</anno>:format_status/2</c>
+	      </seealso>
+	      and
+	      <seealso marker="#Module:terminate/3">
+		<c><anno>NewModule</anno>:terminate/3</c>
+	      </seealso>
+	      must be able to handle the state and data
+	      from the old module.
 	    </p>
 	  </item>
 	</taglist>
@@ -1917,7 +1980,9 @@ handle_event(_, _, State, Data) ->
   <funcs>
     <func>
       <name since="OTP 19.1">Module:callback_mode() -> CallbackMode</name>
-      <fsummary>Update the internal state during upgrade/downgrade.</fsummary>
+      <fsummary>
+	Define the callback mode for the callback module.
+      </fsummary>
       <type>
 	<v>
 	  CallbackMode =
@@ -1933,10 +1998,11 @@ handle_event(_, _, State, Data) ->
 	  <seealso marker="#type-callback_mode"><em>callback mode</em></seealso>
 	  of the callback module.  The value is cached by <c>gen_statem</c>
 	  for efficiency reasons, so this function is only called
-	  once after server start and after code change,
+	  once after server start, after code change,
+	  and after changing the callback module,
 	  but before the first
 	  <seealso marker="#state callback"><em>state callback</em></seealso>
-	  in the current code version is called.
+	  in the current callback module's code version is called.
 	  More occasions may be added in future versions
 	  of <c>gen_statem</c>.
 	</p>
@@ -1945,9 +2011,14 @@ handle_event(_, _, State, Data) ->
 	  <seealso marker="#Module:init/1"><c>Module:init/1</c></seealso>
 	  returns or when
 	  <seealso marker="#enter_loop/4"><c>enter_loop/4-6</c></seealso>
-	  is called.  Code change happens when
+	  is called.
+	  Code change happens when
 	  <seealso marker="#Module:code_change/4"><c>Module:code_change/4</c></seealso>
 	  returns.
+	  A change of the callback module happens when a
+	  <seealso marker="#state callback"><em>state callback</em></seealso>
+	  returns the action
+	  <seealso marker="#type-action"><c>change_callback_module</c></seealso>.
 	</p>
 	<p>
 	  The <c>CallbackMode</c> is either just
@@ -1970,7 +2041,9 @@ handle_event(_, _, State, Data) ->
       <name since="OTP 19.0">Module:code_change(OldVsn, OldState, OldData, Extra) ->
         Result
       </name>
-      <fsummary>Update the internal state during upgrade/downgrade.</fsummary>
+      <fsummary>
+	Update the internal state during upgrade/downgrade.
+      </fsummary>
       <type>
         <v>OldVsn = Vsn | {down,Vsn}</v>
         <v>&nbsp;&nbsp;Vsn = term()</v>
@@ -2058,7 +2131,7 @@ handle_event(_, _, State, Data) ->
     <func>
       <name since="OTP 19.0">Module:init(Args) -> Result(StateType)</name>
       <fsummary>
-	Initializing process and internal state.
+	Initialize process and internal state.
       </fsummary>
       <type>
         <v>Args = term()</v>
@@ -2101,8 +2174,9 @@ init(Args) -> erlang:error(not_implemented, [Args]).</pre>
       <name since="OTP 19.0">Module:format_status(Opt, [PDict,State,Data]) ->
         Status
       </name>
-      <fsummary>Optional function for providing a term describing the
-        current <c>gen_statem</c> status.</fsummary>
+      <fsummary>
+	Describe the current <c>gen_statem</c> status (optional).
+      </fsummary>
       <type>
         <v>Opt = normal | terminate</v>
         <v>PDict = [{Key, Value}]</v>
--- a/lib/stdlib/src/gen_statem.erl
+++ b/lib/stdlib/src/gen_statem.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 2016-2019. All Rights Reserved.
+%% Copyright Ericsson AB 2016-2020. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -694,7 +694,7 @@ enter(
            hibernate_after = HibernateAfterTimeout},
     S = #state{state_data = {State,Data}},
     Debug_1 = ?sys_debug(Debug, Name, {enter,State}),
-    loop_callback_mode(
+    loop_state_callback(
       P, Debug_1, S, Q, {State,Data},
       %% Tunneling Actions through CallbackEvent here...
       %% Special path to go to action handling, after first
@@ -1046,79 +1046,7 @@ loop_event_handler(
     %% restored when looping back to loop/3 or loop_event/5.
     %%
     Q = [Event|Events],
-    loop_callback_mode(P, Debug, S, Q, State_Data, Event).
-
-%% Figure out the callback mode
-%%
-loop_callback_mode(
-  #params{callback_mode = undefined} = P, Debug, S,
-  Q, State_Data, CallbackEvent) ->
-    %%
-    Module = P#params.module,
-    try Module:callback_mode() of
-	CallbackMode ->
-	    loop_callback_mode_result(
-              P, Debug, S,
-              Q, State_Data, CallbackEvent,
-              CallbackMode, listify(CallbackMode), undefined, false)
-    catch
-	CallbackMode ->
-	    loop_callback_mode_result(
-              P, Debug, S,
-              Q, State_Data, CallbackEvent,
-              CallbackMode, listify(CallbackMode), undefined, false);
-	Class:Reason:Stacktrace ->
-	    terminate(
-	      Class, Reason, Stacktrace, P, Debug, S, Q)
-    end;
-loop_callback_mode(P, Debug, S, Q, State_Data, CallbackEvent) ->
-    loop_state_callback(P, Debug, S, Q, State_Data, CallbackEvent).
-
-%% Check the result of Module:callback_mode()
-%%
-loop_callback_mode_result(
-  P, Debug, S, Q, State_Data, CallbackEvent,
-  CallbackMode, [H|T], NewCallbackMode, NewStateEnter) ->
-    %%
-    case callback_mode(H) of
-        true ->
-            loop_callback_mode_result(
-              P, Debug, S, Q, State_Data, CallbackEvent,
-              CallbackMode, T, H, NewStateEnter);
-        false ->
-            case state_enter(H) of
-                true ->
-                    loop_callback_mode_result(
-                      P, Debug, S, Q, State_Data, CallbackEvent,
-                      CallbackMode, T, NewCallbackMode, true);
-                false ->
-                    terminate(
-                      error,
-                      {bad_return_from_callback_mode,CallbackMode},
-                      ?STACKTRACE(),
-                      P, Debug, S, Q)
-            end
-    end;
-loop_callback_mode_result(
-  P, Debug, S, Q, State_Data, CallbackEvent,
-  CallbackMode, [], NewCallbackMode, NewStateEnter) ->
-    %%
-    case NewCallbackMode of
-        undefined ->
-            terminate(
-              error,
-              {bad_return_from_callback_mode,CallbackMode},
-              ?STACKTRACE(),
-              P, Debug, S, Q);
-        _ ->
-            P_1 =
-                P#params{
-                  callback_mode = NewCallbackMode,
-                  state_enter = NewStateEnter},
-            loop_state_callback(
-              P_1, Debug, S, Q, State_Data, CallbackEvent)
-    end.
-
+    loop_state_callback(P, Debug, S, Q, State_Data, Event).
 
 %% Make a state enter call to the state function, we loop back here
 %% from further down if state enter calls are enabled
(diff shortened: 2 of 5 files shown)


ferd:
How would you say this compares to 'enter_loop' as a callback, which works exactly as a more general way to change both the state machine and the state? Calling it as way to get out of hibernation would also get rid of the stack without showing any sign of it.

RaimoNiskanen:
Hello @ferd, that was fast!

I guess a bit what you mean here...  Calling `gen_statem:enter_loop` from the callback module would not be tail recursive since the `gen_statem` engine loop would still expect a return from the callback function.  Also it is not possible to pass all `gen_statem` engine state to the new `gen_statem` engine loop.  So maybe that is not what you meant...

Going into hibernation from the callback module (by an API function in `gen_statem`) would get rid of that return address from the stack, but it is not possible to pass the `gen_statem` internal data such as event queue and whatnot from the callback module into a new engine loop.

essen:
It's something I could use in Cowboy, where it's possible to go from HTTP/1.1 to HTTP/2 or Websocket. In the case of HTTP/2 I believe it could help me solve the problem of "read the request and request body as HTTP/1.1, BUT send the response headers and response body as HTTP/2" via an intermediate callback module.

It may be useful in Gun as well where it could go from a "connecting" state machine to http1, http2, Websocket and so on but I'll need to review in more details if this PR would be enough for this as there are a lot of different interactions possible especially considering proxies. I'll give it a thought tomorrow.

ferd:
@RaimoNiskanen yeah that's correct. Tracking the event queues and so on I guess is a necessity only for gen_statem, since deferring to having other behaviours (gen_server, and so on) track their one timers internally makes the enter_loop + hibernation trampolining workable for them. I guess it is a bit of a warning regarding the expansion of these other behaviours' offering; the added conveniences, especially if around side-effects, risk making it difficult to use more general mechanisms to do their specific switches, and would require framework support at the individual level.

Given that context it seems there's not really a way around it for gen_statem though. It interestingly enough kind of maps well to the original "become" semantics of actor systems, but for state machines specifically.

RaimoNiskanen:
For the discussion: the internal state in `gen_server` is: ParentPid, SysDebugState, ServerName, CallbackModule, HibernateAfterTimeout.  `gen_server:enter_loop/5` requires the caller to supply CallbackModule, Options, ServerName (and the new State).  It will pick up ParentPid from the process dictionary, and HibernateAfterTimeout from Options.  The SysDebugState, I guess will be re-initialized from Options.

So I have never heard of the enter_loop + hibernation trampoline before, but it seems that using it requires you to keep your ServerName and start Options in your callback State up to the trampoline jump, and you will get the SysDebugState cleared.

And, I assume it is `proc_lib:hibernate/3` that should be used, right?

RaimoNiskanen:
Name poll: what is the best name for the callback action?
* `{change_callback_module, NewCallbackModule}`
* `{set_callback_module, CallbackModule}`
* `{callback_module, CallbackModule}`

The current suggestion; `change_callback_module` indicates it is a change, although `NewCallbackModule` can be equal to the current (no reason to check for that stupidity), so maybe `set_callback_module` is more neutral, or maybe the even shorter `callback_module` will do fine.

RaimoNiskanen:
Regarding the `gen_server:enter_loop` + `proc_lib:hibernate` trampoline: is there a desire for being able to change more of the engine internal state, for `gen_server` and `gen_statem`?  `{hibernate_after,Time}` is one clear candidate.  Changing the server name is another less clear candidate, is that even useful?  Changing the parent another albeit strange one.

The last one, changing the parent, should maybe be done from the outside, via new functions in `sys:`, and it has implications on the supervision tree.  I do not know if it is a desirable feature to re-plant a server to a different supervisor, or if you just should restart the whole tree...

`sys:log` and `sys:trace` can now only be affected from the "outside" from those API functions and from server start arguments.  There could be actions to change these from the callback module, but that is also a new and probably really strange thing to do. (?)

IngelaAndin:
@ferd  As Raimo pointed out enter_loop will not work. The rational for the enter_loop function is to be able to have a non synchronous init phase of a process under a simple_one_for_one supervisor (I know as I implemented it for that reason once upon a time). It could be argued that this should be better integrated into existing behavior infrastructure but that is a separate issue.

lib/stdlib/doc/src/gen_statem.xml
```diff
@@ -1168,6 +1191,46 @@ handle_event(_, _, State, Data) ->
...
+	      <seealso marker="#Module:callback_mode/0">
+		<c>NewModule:callback_mode/0</c>
+	      </seealso>
+	      before the next
+	      <seealso marker="#state callback">state callback</seealso>.
+	    </p>
+	    <p>
+	      Changing the callback module does not affect the
+	      <em>state transition</em> in any way,
+	      it only changes which module that handles the events.
```
essen: From what I understand, that means that you can't change the state or data at the same time. That's probably too limited for my case because I will need to convert the data at the same time (HTTP/1.1 and HTTP/2 have very different state data). I can probably do an intermediate state in order to convert the data before changing the callback module so it's not a blocker. (It can't be done after module change because the new module doesn't have the record definition.)

I am wondering however about the state name itself. In my case the states do not match because when transitioning to HTTP/2 the preface must be sent and therefore if I transition from a `connected` state I can't just go to the corresponding `connected` state for HTTP/2. I also need to do some protocol-specific initialization (not just converting the state), setup timeouts, call callbacks and so on. So it would be ideal if I could do something like this:

``` erlang
statem_callback(..., StateData0=#st{a=A, b=B, c=C...}) ->
    cancel_timeouts_etc(StateData0),
    StateData = new_statem:convert_state(A, B, C...),
    [{next_state, preface, StateData}, {callback_module, new_statem}].
```

And it brings me into the `preface` state of `new_statem` with `StateData` instead of `StateData0`. The bonus I get from this is that I can share the `preface` state between the different entry points to `new_statem` (for HTTP/2, there are 3 different entry points, but as mentioned yesterday one of them would be better done as an intermediate callback module since it needs to juggle with both HTTP/1.1 and HTTP/2 at the same time).

Again nothing impossible, I would just have to have a transition state `switch_protocol` that exists in both protocols and just go `connected` -> (change state data a the same time) -> `switch_protocol` -> (change callback module) -> `switch_protocol` -> `preface`, but if I can do all in one go I'd take it.

essen:
> * `{change_callback_module, NewCallbackModule}`
> * `{set_callback_module, CallbackModule}`

Either of those two really, the verb indicates an action. First one is probably more accurate, because you don't just "set" it, you may also call `callback_mode`.

lib/stdlib/doc/src/gen_statem.xml
RaimoNiskanen: @essen: Since changing the callback module does not affect the state transition in any way, it is absolutely possible to combine a state transition that changes State and updates Data with changing callback module.  Changing State and updating Data is orthogonal to changing callback module.

```erlang
statem_callback(..., StateData0=#st{a=A, b=B, c=C...}) ->
    cancel_timeouts_etc(StateData0),
    StateData = new_statem:convert_state(A, B, C...),
    {next_state, preface, StateData,
     [{change_callback_module, new_statem}]}.
```
The next callback that will be called is `new_statem:preface(Type, Content, StateData)`.

You can even insert an event in the same state transition to force a call of the new callback module, so it is possible to instead of having an API function `new_statem:convert_state/N` insert `{next_event, internal, convert_state}` to the `preface` state, so that there is a function clause that converts the state instead of an API function.  (If that would be clearer)

Does that solve all the problems in the rest of your post?

Do you have any suggestion for improvement of the documentation wording, to make this important point clearer?

RaimoNiskanen: > The next callback that will be called is `new_statem:preface(Type, Content, StateData)`.

More accurately the next callback that will be called is `new_statem:callback_mode()`, and then either `new_statem:preface(Type, Content, StateData)`, or `new_statem:handle_event(Type, Content, preface, StateData)` depending on `new_statem`'s callback mode, or even, if state enter calls are used by `new_statem`; `new_statem:preface(enter, statem_callback, StateData)` or `new_statem:handle_event(Type, enter, statem_callback, StateData)` inbetween...

essen: Ah I thought perhaps I was wrong about that. No problem at all then.

About inserting an event to update the state that won't work for me because the new statem doesn't have the record definition of the old statem so some conversion needs to be done beforehand (or manually looking into the record, or exposing the record definition, neither of which I'm willing to do). But in theory that would be possible indeed.

The documentation wording is fine, it's just a complex subject and I get easily confused with the different types of events and actions.

attah:
I too have experimented with things along these lines, but as a shim module implementing the behaviour, and forwarding the calls. I'd certainly prefer it if similar functionality (like proposed here) became part of the official behaviour. 

My solution is not cleaner, but i did experiment with a few more things within the same concept, like "inheriting" handler functions, which i know from previous experience is quite useful if done right. Sure, a handler function could call a library function that is shared among several, but there is something that appeals to me with the write-once concept.

Another thing i experimented with was "pushing" new handler modules, so that handler functions could end in a "pop". This is useful for at some initial stage pushing a couple of procedures, which themselves does not care about order.

One use-case for this would be to for example send a request in the enter function, explicitly handle positive and negative responses, but "inherit" handling of unrelated requests and common error handling. Here i found it quite annoying to not be able to have logic in the enter function that voids the procedure, and switches module. Maybe i'm just doing it wrong :) But i find that approach (pop, enter, conditionally move on) cleaner than a dozen calls to maybe_send_req_and_enter_cbmod(), or worse yet, duplicated logic.

Happy to see initiatives like this, hopefully my musings can be of some inspiration.

RaimoNiskanen:
Hello @attah!  Interesting ideas!

In my view, inheriting handler functions strays a bit too far from "the Erlang way" (if there is one such way) of explicitness.  I prefer having a catch-all at the bottom of a state that explicitly calls some other handler.  To inherit handler functions feels very object oriented in that you can not tell from the code in front of you where an event is handled...

Switching callback module in state enter calls i think goes against what I intended them for, i.e to place state entry code right next to the state handling code.  It is somewhat a matter of preference.  Then also I think I recall from implementing this feature that in the current implementation of `gen_statem` it is problematic to change callback module, having to invalidate and re-read the callback mode from the new callback module, in the middle of doing state enter calls.  I am not even sure it possible to get coherent semantics...

attah:
Hi @RaimoNiskanen 

Right, yes, that is probably the way to go. It is just something that rubs me the wrong way with (explicitly) having the last "state function" call a handle_event. However, in this case the concept makes a bit more sense. Good point.

I'm not trying to push object orientation, but it was a problem we had and a solution that worked great. The language back then just happened to be object oriented. With what i prototyped in Erlang recently i'm rather trying to go declarative instead of explicit if that makes any sense.

You and OTP seem to be in agreement that switching callback module is impure, and possibly hard. The former argument is perhaps best left for another day (feel free to expand on it), but i must say i don't see what would be so (conceptually) hard about it. A concrete implementation can of course be more or less stuck in its ways, but since the enter function already is allowed to change state data, i can't see how state would be more contended for by other actions. 

By logic it should be the very last thing done in the gen_statem code itself as part of changing state. Detecting that it changed again and repeating the steps doesn't sound so bad. I'll have to look in to the code and see if the argument makes more sense then. :)

Anyway, about pushing handler modules; do you have any opinions there? I'm not really suggesting you include it now, unless you really want to. But rather just keep it in the back of your mind for if/when decisions that could affect it are made as part of this improvement. Perhaps you think of some way to evolve that concept too. :)

RaimoNiskanen:
@attah: Pushing and popping handler modules; If I understand it correctly would be just to keep a stack (list) of handler modules instead of now just an atom.  It does not sound conceptually very different from `change_callback_module`.  Seems like an interesting idea!  Should not be hard to implement. (should the server crash or ignore pop of empty module stack?)

About switching callback module being impure...  It has never been possible for a behaviour to switch callback module before this, so it is a bit a scary thought.  I hope there will be nothing colliding with the release handler or the application concept in OTP.

Switching callback module from state enter calls is what I think is very ugly, and probably hard to implement; this since the callback mode has to be updated in the middle of what I recall is the wrong part of the loop.

RaimoNiskanen:
...and should `change_callback_module` change the stack top or clear the stack?

attah:
@RaimoNiskanen Precisely. Apart from the "blind return", it is indeed just about the same thing. I'm trying to hide the state transitions in less logic, while still enabling reuse.

In true Erlang spirit i have considered popping an empty stack/list as a critical error. I think pushing and changing state need to be separate in that regard, i.e. pushing leaves the current module at the bottom of the stack. `change_callback_module` should just replace the current one, not the entire stack. Needing to empty the stack should not be needed if the design is reasonable. The stacker should take care to only stack popping modules too of course, anything else should probably have started with a module change.

I see i expressed myself poorly, sorry. I blame running to a meeting. Switching callback module is impure _in the enter function_ it should have said. I really like your idea, and don't see any real bad sides to it. I agree switching callback module in the enter function isn't pretty, but nether is convoluted logic around deciding the next state or callback module. This is what i'm trying to get rid of.

Without the pop semantics, it can most probably be fine the way it stands now, where the enter function is limited in what it can do. (Although i still find it artificial) But so far i can't see a good way to reduce state/module change logic and still be able to skip stacked states without it. (Well, one can send signals to oneself, but seriously?)

ferd:
I'm not sure I properly understand what a stack of callbackmodule implies for a state machine. I guess the closest parallel I can think for it would be of nested state machines, and adding a handler going a level deeper in the state machine, while popping just getting back to the top-level state machine? I can imagine at least using something like having a way to specify a sub-state and sub-handler and exiting that rather than literally adopting stack semantics but either case it's one hell of an extension to bolt on.

I already consider gen_statem to be rather complex and has so far been a bit of a challenge to explain in the few sessions I ran with newcomers to the language (instead we only go for a subset of the features). This module will soon catch up to `sofs` in the ease of adopting it for the first time based on the docs.

attah:
@ferd I'll offer you an explanation that a colleague of mine uses for the concept. Think of it as queuing a set of procedures. Each with a few states (possibly just one), that are otherwise irrelevant for the main state handler and would just serve to clutter it up. Concise handler modules are always nice.

The other key part of the concept is the queuing itself, facilitated by the stack and blind popping of states as opposed to convoluted logic around determining next state/module.

So i guess this can either mean going N levels deep in substates (and working your way back up), or just going one level deep, with the substate ending in switching to another substate, which it has no real knowledge about other than that it was promised to be stacked. Or a combination of the two...

I do agree gen_statem, and OTP behaviours in general are not as approachable as i'd like them to be. I too struggle here sometimes. However, i don't primarily blame all the return options, although they could be more organized, but the amount of boilerplate and hurdle to get to the first handled call. Admittedly there is a correlation between the two, but still.

@RaimoNiskanen Feels i may be starting to derail your PR, do tell me to go elsewhere with this discussion at your earliest discomfort.

RaimoNiskanen:
@attah I will try to schedule `{push_callback_module, NewModule}` and `pop_callback_module` for a Fun Day in the future.
* Should `{push_callback_module, ?MODULE}` just do what it is told or try to be clever and be a no-op?
* Should there be a `{pop_callback_module, N}` with `{pop_callback_module, 1}` being equivalent to `pop_calback_module`? ( `N` is the number of levels to pop)

You are right, maybe the discussion about features not in this PR should end here, after those questions have been answered...

Other than that there seems to be no problems with this PR, and I will probably merge soon...

@ferd I love the comparison with `sofs`, and hope the [`gen_statem` behaviour description](https://erlang.org/doc/design_principles/statem.html) is now good enough to avoid being `sofs`.

attah:
> Should {push_callback_module, ?MODULE} just do what it is told or try to be clever and be a no-op?

No trying to be clever :) I think that generally a handler module would not push itself, but one of its "substates" as part of a procedure. That could also involve pushing more than one other callback module.... and what if the top one is in that list, etc, etc... too complicated to reason around imo.

> - Should there be a {pop_callback_module, N} with {pop_callback_module, 1} being equivalent to pop_calback_module? ( N is the number of levels to pop)

Going with my reasoning that the pushed module should not (need to) know about what it is popping to, i'd say that goes for the number of modules remaining in the stack, just the same as the names of them. So, i don't think it is a good idea. Possibly a way to completely replace the stack can be useful for enabling "become"-semantics regardless of how deep in substates you are. However, i think substating often goes well with postponing certain events, so for sake of sanity becoming another set of handlers is best done at an empty (or 1-deep, depending how you see it) stack.

RaimoNiskanen:
Ok.  I think we had the same answer to all of my questions now and before :-)
Thank you for your input!

okeuday:
@RaimoNiskanen This is a nice feature to help manage the complexity of state machine source code.  However, it doesn't seem like the impact on hot code loading was mentioned.  If you want to update a gen_statem module and you don't know what module to update, due to it changing based on the state handling, it seems like it should be a problem.  I would assume you wouldn't think this is a problem because you should know all the callback modules that should be used by a single gen_statem module and the list of those modules can be passed to `code:atomic_load/1`.  

To avoid problems, you may want to provide a way of specifying all the callback modules that will be used so the callback module atom is checked before the change.  For example, if you added an optional `Module:callback_modules() -> list(module())` function (if not provided it would be the same as an empty list, with the current module added if it wasn't already present) there would be a way of knowing all the modules that need to be updated as a group.  Having `Module:callback_modules/0` defined would help avoid human error when updating and would help the developer understand the size of the state machine source code.

RaimoNiskanen:
Code reloading (live application upgrade) has not been mentioned, that is right.

There are two facets of that problem, that I can think of.

#### Code reload of a server
Code change is controlled by `sys:change_code/4,5` that calls `Module:code_change/4` in the current callback module.  So for code change to work, all callback modules must implement `Module:code_change/4` and handle the state and data upgrade.  This is just as for one callback module, but for all callback modules.  Should be no big surprise.

#### Application upgrade
In this case I think there is no new problem with multiple callback modules since you within the application concept in OTP already have to enumerate all application's modules in your application specification, and it does not matter which are callback modules and which are helper (library) modules.  During an application upgrade all servers in the application are halted, all the application's modules are reloaded, and then all its servers are un-halted.

Having multiple ways to specify your application modules i.e through `Module:callback_modules()` would be redundant and in effect only for documentation purposes.  I think such documentation should be elsewhere (comments, application specification file, ...).

If this information should be in a callback module, then I think it should enforce something (as you suggest) so it can not go stale.  For example it could be used only from the first callback module, and after that you are only allowed to change between known modules, but in that form it is a bad idea since the module list could change after an upgrade.  Then we need to reload the list after code upgrade, from the current callback module,... this gets messy.

So I think it that specifying callback modules in such a way would be over-engineering...

ferd:
Unless I'm wrong, the code upgrades only calls for the upgraded module to be upgraded in the worker process if the specific module is mentioned in the supervisor's spec.

So if you work with gen_statem that way potentially having multiple modules, you'd likely have to switch to a 'dynamic' type (like for gen_event) or a static multi-module list. Otherwise, scanning the supervision tree may leave the worker non-upgraded on the old version, and will cause crashes.

The gotcha with the static list is that if _any_ of the modules in the list is upgraded, it will get suspended and `change_code` will be called on it. But if your list has `[ModA, ModB]` you may end up calling `ModB`'s code_change when ModA is running. This would likely not happen with the dynamic mode which could specify only which module is currently running. 

I believe that the latter is handled transparently in the gen.erl implementation when release_handler_1 calls things, but this also requires a breaking change in how everyone specifies and supervises their gen_statem to specify the `dynamic` type.

ferd:
Actually I think gen_statem does not implement the `get_modules` message handling which would be required for `dynamic` to work and provide safe upgrades: https://github.com/erlang/otp/blob/83c68a5e9cebdd00d00bbf8e5de20ce928e0172f/lib/stdlib/src/gen_event.erl#L366-L368 

So uh, any multi-state FSM is going to be hit or miss for an upgrade on a node using the static list of callback modules as far as I can tell.

okeuday:
If all related gen_statem processes are suspended together with `sys:suspend/1,2` called multiple times, once for each process, then `sys:change_code/4,5` would work as described.  So, handling the update manually would work.

RaimoNiskanen:
@ferd You are right.  It is the child specification that specifies which modules a server uses:

http://erlang.org/doc/design_principles/sup_princ.html#child-specification

And in that chapter it says:
> `modules` are to be a list with one element `[Module]`, where `Module` is the name of the callback module, if the child process is a supervisor, `gen_server`, `gen_statem`. If the child process is a `gen_event`, the value shall be `dynamic`.

So `dynamic` is only supported for `gen_event`, not for `gen_statem` (or `gen_server` (or once `gen_fsm`)).

I think there is no gotcha for a static child specification modules list.  It is the server that is ordered to suspend, with a `sys` call to the server, handled within the server engine.  Then all modules are changed, and the server is un-suspended.

In https://erlang.org/doc/man/supervisor.html it says:
> `modules` is used by the release handler during code replacement to determine which processes are using a certain module. As a rule of thumb, if the child process is a supervisor, `gen_server` or, `gen_statem`, this is to be a list with one element `[Module]`, where `Module` is the callback module. If the child process is an event manager (`gen_event`) with a dynamic set of callback modules, value `dynamic` must be used. For more information about release handling, see Release Handling in OTP Design Principles.

That section indicates that a list of more than one module is possible.

I think the documentation in both sections could be improved to state that for `gen_statem` the `modules` list should list all callback modules.

In fact, the SSL application has since long (the `gen_fsm` days) had two modules listed here because it has got a proxy module and a back-end module that implements the state machine...  I think  this list should really contain all modules that would be affected by the upgrade such as modules that handle the server's state (data), or helper modules that the server calls with a changed API.

RaimoNiskanen:
@ferd You are more right!  I am unfortunately a bit rusty regarding OTP Design Principals and exactly how they work.  Finally I think I get what you are saying. :-)

The modules list, as you said, in the supervisor child specification, lists which modules, when upgraded, that should cause a server suspend, followed by code upgrade (load), ordering the server to call `code_change()` in the current callback module, lastly server un-suspend.

So, as you say: if the modules list is `[mod_a, mod_b]`, and `mod_b` is being upgraded, the result might be that `mod_a:code_change/4` is called.

The list of modules that are upgraded for an application comes from the .appup file that has a list of modules that are changed in the upgrade.  This list is manually created, as is the supervisor child specification.

One solution could be that the modules list should list all callback modules, since any of them may do the `code_change/4`, and when updating any of them; all of them must be able to handle the state+data upgrade in their `code_change/4`.  Note that all of them often need to be able to handle the new state+data anyway, so updating all of them should be the norm.  It might be appropriate to have `code_change/4` implemented in one module and the others calling it.

Also part of that solution is that all callback modules, since all of them handles the new state (maybe), should be listed in the .appup file so all gets upgraded together.

Another solution would be to have just one module, for example the first callback module, in the supervisor child spec modules list.  This is sufficient as long as all callback modules are in the .appup file modules list to get upgraded together.  Then they can all for example call `code_change/4` in the first callback module, as long as all callback modules have a `code_change/4` that handles the state+data upgrade, and that all callback modules can run with the updated state+data.

Boiling it down even further...

Theoretically, I think, the modules in the supervisor child spec modules list need not be callback modules at all, as long as such modules always are in the .appup file modules list for all future soft upgrades.  All callback modules that handles the new state must be updated in a soft upgrade, and therefore also be in the .appup file modules list.  All callback modules that might get the call to `code_change/4` must be able to handle the state+data upgrade.

I think that was the minimal requirements...

Too many words.  \*Sigh!\*

ferd:
Yeah. It's getting fairly tricky to manage these. It kind of feels like going for 'dynamic' would be safer as well, but as I was thinking about that one, it turns out that it would also risk giving a racy behaviour. Unless `gen_event`, where the multiple modules are potentially all running at once, `gen_statem` has the ability to possibly change its callback module in the interval of time between the `get_modules` message is sent from the release handler and when it sends the `suspend` message through the `sys` module. As such, even `dynamic` is *not* safe for `gen_statem`

That does leave you in the uncomfortable position to have to something like:
-  call _all_ the callback modules that need upgrading at once in one suspend (is this a reasonable idea? I haven't thought that one through)
- have any of the callback modules able to understand the upgrade for all of the other callback modules (this is not going to work well when private records in `mod_b` are being used in the state but `mod_a` is asked to handle them, and will be very risky)
- have the `gen_statem` module notice and no-op the code upgrade to an identity function when the `Mod` for the upgrade doesn't match the callback module  but would still unnecessarily suspend the process in question

The latter should be doable with: https://github.com/erlang/otp/blob/83c68a5e9cebdd00d00bbf8e5de20ce928e0172f/lib/stdlib/src/gen_statem.erl#L808-L813 and could work well for the first option as well.

I can't imagine having all callback modules able to handle all code changes working well due to potentially multiple of them having similar `#state{}` or `#data{}` records that are private, or shaped in such ways (valid for maps as well) that knowing which worker you're upgrading is not obvious.

But yeah, this change has to come with a change of how OTP modules are being specified for code upgrades and get the accompanying code in place, otherwise anyone who tries to hot load a release with new versions of state machines in it (or ssl/tls being used) will find themselves unable to safely do hot code loading when this is released next.

RaimoNiskanen:
FYI.  For completeness have I merged a branch into 'maint' that adds `{push_callback_module, NewModule}` and `pop_callback_module' to the possible actions just like `{change_callback_module, NewModule}` since I thought the can of worms (if it is one) was already open...

ferd:
so what's the solution for hot code upgrades?

RaimoNiskanen:
Since the solution would be the same for push & pop as for change, I would not add a new problem and simply implemented them...

The solution for a soft upgrade with state change, I think, is to have one, the "original", callback module in the child specification, "traditionally".

Then the .appup file should specify an advanced upgrade for the same module, and all other callback modules also has to be loaded (reloaded).  This would make the release handler do the code change dance for our server, upgrading all callback modules.

Now the `M:code_change/4` function will be called in the current callback module, so any callback module that may be the current during upgrade (probably every callback module) must be able to handle the state change.

One way to accomplish that would be to implement the `M:code_change/4` function just in the "original" callback module and let the other callback modules call it, which in its turn may call out to some other function in the other callback modules for change of their specific state parts.  One can be certain that the current callback module at least knows how to update its state and sometimes that might be sufficient.

Reasonable?

ferd:
It's gonna blow up with shared tuple definitions based on context, but I guess "clashing tuple definitions" are kind of frequent already...

shawarmaz:
can `change_callback_module` be used to run `:gen_statem` in series to create hierarchical state machines?

RaimoNiskanen:
That depends on what you mean by "in series" and "hierarchical"...

When you have changed callback module it is the new module that handles all events.  The new module may also change to yet another callback module.

If you push the current callback module you change to a new callback module, that handles all events.  That new callback module may when it so desires pop the callback module to get back to the next on the stack (the one that pushed), so the popping callback module does not need to know which callback module that pushed.  After a pop it is the now new (the old) callback module that handles all events.

This sounds, in some way, "hierarchical", but maybe not so much "in series".
I do not know what you had in mind...

PR #2534, opened 2020-02-13
pg: ensure argument is expected local pid(s)

rnowak:
Hi,

Looking through the new pg module that was recently merged in (https://github.com/erlang/otp/pull/2524), I came across something that might be unintended.

https://github.com/erlang/otp/blob/8eefeda8d81996bfe8249017fa0be2595eb64cff/lib/kernel/src/pg.erl#L130

There appear to be two issues here. One is that in the list case the (in)comprehension deals in dark arts, and two, that in the pid case, it is never ensured in the caller to be a pid.

#### List Case

The intent in join/3 when PidOrPids is a list seems to be to ensure that each entry is a pid and that it is local to the node, otherwise throw an error.

Unfortunately, the filters in comprehensions treat functions that can be guards as guards would be treated in clause heads ([read more](http://erlang.org/pipermail/erlang-questions/2010-October/053649.html)), and unless wrapped in for instance a block expression or a fun, exceptions are devoured and the entire filter is treated as false -- in this case, sidestepping the intended protection mechanism.

As such:
```erlang
1> Node = node().
pgfixes@something
2> PidOrPids = [123, 456, 789].
[123,456,789]
3> is_list(PidOrPids) andalso [error({nolocal, Pid}) || Pid <- PidOrPids, node(Pid) =/= Node orelse not is_pid(Pid)].
[]
```
the is_pid/1 call never gets a chance as node/1 throws a badarg which magically turns the entire filter into a 'false'.

A wrapping in a block expression seems to dispel the magic:
```erlang
4> is_list(PidOrPids) andalso [error({nolocal, Pid}) || Pid <- PidOrPids, begin node(Pid) =/= Node orelse not is_pid(Pid) end].
     in function  node/1
        called as node(123)
```

#### Pid Case

However, when PidOrPids is a pid instead of a list, the checks aren't afforded at all. As such, pg:join(group, 123) for instance passes through into the process and it eventually crashes with a no function clause in join_monitors/3 as it has an is_pid guard for the only non-list clause.

The same thing occurs for leave/3 where it crashes with a no function clause in leave_monitors/3.

While raising these issues was my main concern, I took some liberties in rewriting this handling.

I kept the semantics intact where {nolocal, Pid} is thrown as an error. Perhaps this should do something else in case a list element isn't a pid at all.

Minor spec and doc adjustments in the related parts to reflect current state.

While at a glance there are a few other bits that could be improved in this module, this is what I could do for now.

Thanks.

### DIFF
--- a/lib/kernel/src/pg.erl
+++ b/lib/kernel/src/pg.erl
@@ -117,31 +117,29 @@ start_link(Scope) when is_atom(Scope) ->
 
 %%--------------------------------------------------------------------
 %% @doc
-%% Joins a single process
+%% Joins a single or a list of processes.
 %% Group is created automatically.
-%% Process must be local to this node.
+%% Processes must be local to this node.
 -spec join(Group :: group(), PidOrPids :: pid() | [pid()]) -> ok.
 join(Group, PidOrPids) ->
     join(?DEFAULT_SCOPE, Group, PidOrPids).
 
 -spec join(Scope :: atom(), Group :: group(), PidOrPids :: pid() | [pid()]) -> ok.
-join(Scope, Group, PidOrPids) ->
-    Node = node(),
-    is_list(PidOrPids) andalso [error({nolocal, Pid}) || Pid <- PidOrPids, node(Pid) =/= Node orelse not is_pid(Pid)],
+join(Scope, Group, PidOrPids) when is_pid(PidOrPids); is_list(PidOrPids) ->
+    ok = ensure_local(PidOrPids),
     gen_server:call(Scope, {join_local, Group, PidOrPids}, infinity).
 
 %%--------------------------------------------------------------------
 %% @doc
-%% Single process leaving the group.
-%% Process must be local to this node.
+%% Single or list of processes leaving the group.
+%% Processes must be local to this node.
 -spec leave(Group :: group(), PidOrPids :: pid() | [pid()]) -> ok.
 leave(Group, PidOrPids) ->
     leave(?DEFAULT_SCOPE, Group, PidOrPids).
 
--spec leave(Scope :: atom(), Group :: group(), Pid :: pid() | [pid()]) -> ok | not_joined.
-leave(Scope, Group, PidOrPids) ->
-    Node = node(),
-    is_list(PidOrPids) andalso [error({nolocal, Pid}) || Pid <- PidOrPids, node(Pid) =/= Node orelse not is_pid(Pid)],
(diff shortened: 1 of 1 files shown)


rnowak:
Force-pushed an update to correct an oversight where empty nested lists in the list would fall through.

lib/kernel/src/pg.erl
```diff
@@ -117,33 +117,45 @@ start_link(Scope) when is_atom(Scope) ->
...
+%% Single or list of processes leaving the group.
+%% Processes must be local to this node.
 -spec leave(Group :: group(), PidOrPids :: pid() | [pid()]) -> ok.
 leave(Group, PidOrPids) ->
     leave(?DEFAULT_SCOPE, Group, PidOrPids).
 
--spec leave(Scope :: atom(), Group :: group(), Pid :: pid() | [pid()]) -> ok | not_joined.
-leave(Scope, Group, PidOrPids) ->
-    Node = node(),
-    is_list(PidOrPids) andalso [error({nolocal, Pid}) || Pid <- PidOrPids, node(Pid) =/= Node orelse not is_pid(Pid)],
```
michalmuskala: Could this be fixed by swapping the order of `not is_pid(Pid)` and `node(Pid) =/= Node`? This should ensure it properly returns in case the argument is not a pid instead of silently failing.

rnowak: That does not fix the issue as it is an `orelse` and even if `not is_pid(Pid)` is `true`, any `Pid` input to `node/1` there will cause the entire filter to become `false`. It could probably be massaged by different grouping and `not` placement, but it is a bit unwieldy in my opinion with the special casing where the exceptions are eaten and all. The case of where it is a pid could be made proper by just rearranging, but the good case already works, so it'd achieve little. Even then, the case of a single pid wouldn't be checked as line 144 is only concerned if it is a list.

max-au:
Thank you. It is indeed my oversight.

May I make a few suggestions?
1. I added a test, see this branch: https://github.com/max-au/otp/tree/pg_nolocal_test
2. Could you move "ensure_local" function to line 340 (after gen_server callbacks), just to keep code style (local non-exported functions come after API-exported functions), right after "Internal implementation" comment.

rnowak:
Hi @max-au,

Thank you for your reply, and I apologize for the delay. As for 2. I have moved the function down -- I wasn't sure where you'd want non-exported functions used by the public API. Thanks. As for 1. Do you want me to pull in your commit into this PR or how would you like that handled?

If there's anything else you'd like me to change, please let me know.

PR #2535, opened 2020-02-14
Clarify try-of documentation

sverker:
- `try_clause` instead of `case_clause`
- `try_clause` is not caught

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1414,9 +1414,10 @@ end</code>
       the patterns <c>Pattern</c> are sequentially matched against
       the result in the same way as for a
       <seealso marker="#case">case</seealso> expression, except that if
-      the matching fails, a <c>try_clause</c> run-time error occurs.</p>
-    <p>An exception occurring during the evaluation of <c>Body</c> is
-      not caught.</p>
+      the matching fails, a <c>try_clause</c> run-time error occurs instead of a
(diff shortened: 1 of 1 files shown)


ferd:
The thing I've found most useful when explaining the difference in the past was to give a good reason to use it:

`try ... of -> Body catch ... end` lets you put recursive calls in `Body` that will not get protected by the `catch`, which means that it will not interfere with tail recursion by staying in the call stack and expecting a potential exception.

system/doc/reference_manual/expressions.xml
```diff
@@ -1414,9 +1414,11 @@ end</code>
       the patterns <c>Pattern</c> are sequentially matched against
       the result in the same way as for a
       <seealso marker="#case">case</seealso> expression, except that if
-      the matching fails, a <c>try_clause</c> run-time error occurs.</p>
-    <p>An exception occurring during the evaluation of <c>Body</c> is
-      not caught.</p>
+      the matching fails, a <c>try_clause</c> run-time error occurs instead of a
+      <c>case_clause</c>.</p>
+    <p>Only exceptions occurring during the evaluation of <c>Exprs</c> can be
+      caught by the <c>catch</c> section. Exceptions occuring in a <c>Body</c>
```
mikpe: nit: occurring (two r:s)

sverker: :+1:

PR #2540, opened 2020-02-18
Adjust spawn_request() features

rickard-green:
* Remove (the not yet released) `timeout` option of `spawn_request()`
* Introduce a new BIF <s>`abandon_spawn_request/1`</s>[`spawn_request_abandon/1`](http://erlang.org/~rickard/OTP-13450_OTP-15251/erts-10.6.4/doc/html/erlang.html#spawn_request_abandon-1) as a replacement for the removed `timeout` option. This BIF is more flexible for handling timeouts than the `timeout` option, since you can determine at any time when you want to abandon the request instead having to determine that at the time of the call to `spawn_request()`. The BIF also requires a lot less functionality in the VM than the `timeout` option.
* Introduce a `reply` option to the [`spawn_request/5`](http://erlang.org/~rickard/OTP-13450_OTP-15251/erts-10.6.4/doc/html/erlang.html#spawn_request-5) BIF making it possible to silence replies that one is not interested in.


lpgauth:
Wouldn't `erlang:cancel_spawn_request/1` be more inline with the normal naming?

e.g. `erlang:cancel_timer/1`, `httpc:cancel_request/1`

rickard-green:
> Wouldn't `erlang:cancel_spawn_request/1` be more inline with the normal naming?
> 
> e.g. `erlang:cancel_timer/1`, `httpc:cancel_request/1`

At least to me 'cancel' implies that the request will be withdrawn, stopped or similar which is not the case here. The spawn-request is sent and we do not try to withdraw the request and cannot do that. If no failures occur and it just takes a long time for the signal to reach its destination, a new process *will* be spawned. I think 'abandon' explains what the BIF does quite well, while 'cancel' does not.

IngelaAndin:
Maybe in the httpc case abandon would also have been a better name as it is impossible to
guarantee that request does note reach the server, it is more a way of saying you are no longer interested in the response. Although depending on timing a canceled request might not be sent to the server.

essen:
> Maybe in the httpc case abandon would also have been a better name as it is impossible to
> guarantee that request does note reach the server, it is more a way of saying you are no longer interested in the response. Although depending on timing a canceled request might not be sent to the server.

No it's the right name. In HTTP/2 you can tell the server to cancel requests as well, with `cancel` as the reason, even though there's no guarantee that the request does not get processed. So `cancel` is part of HTTP with the same properties you attribute to "abandon".

PR #2543, opened 2020-02-18
stdlib/sys: change behavior for sensitive processes

Pouriya-Jahanbakhsh:
Prevent leaking process dictionary and prevent calling `ReplaceStateFun` (which can change sensitivity) for sensitive processes.  
If I understand the codes correctly; `gen_server`, `gen_event` and `gen_statem` codes should change in lines [925](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_server.erl#L925), [1202](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_server.erl#L1202), [1204](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_server.erl#L1204), [777](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_event.erl#L777) and [2349](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_statem.erl#L2349) too.

### DIFF
--- a/lib/stdlib/src/sys.erl
+++ b/lib/stdlib/src/sys.erl
@@ -510,27 +510,41 @@ do_get_state(Mod, Misc) ->
     end.
 
 do_replace_state(StateFun, Mod, Misc) ->
-    case erlang:function_exported(Mod, system_replace_state, 2) of
-	true ->
-	    try
-		{ok, State, NMisc} = Mod:system_replace_state(StateFun, Misc),
-		{State, NMisc}
-	    catch
-		Cl:Exc ->
-		    {{error, {callback_failed,{Mod,system_replace_state},{Cl,Exc}}}, Misc}
-	    end;
-	false ->
-	    try
-		NMisc = StateFun(Misc),
-		{NMisc, NMisc}
-	    catch
-		Cl:Exc ->
-		    {{error, {callback_failed,StateFun,{Cl,Exc}}}, Misc}
-	    end
-    end.
+Sensitivity = erlang:process_flag(sensitive, false),
+    Result = 
+        case {Sensitivity, erlang:function_exported(Mod, system_replace_state, 2)} of
+            {true, _} ->
+                {{error, sensitive}, Misc};
+            {_, true} ->
+                try
+                    {ok, State, NMisc} = Mod:system_replace_state(StateFun, Misc),
+                    {State, NMisc}
+                catch
+                    Cl:Exc ->
+                        {{error, {callback_failed,{Mod,system_replace_state},{Cl,Exc}}}, Misc}
+                end;
+            _ -> % {false, false}
+                try
+                    NMisc = StateFun(Misc),
+                    {NMisc, NMisc}
+                catch
+                    Cl:Exc ->
+                        {{error, {callback_failed,StateFun,{Cl,Exc}}}, Misc}
+                end
+        end,
+    _ = erlang:process_flag(sensitive, Sensitivity),
+    Result.
 
 get_status(SysState, Parent, Mod, Debug, Misc) ->
-    PDict = get(),
+    PDict = 
+        case erlang:process_flag(sensitive, false) of
+            false ->
+                get();
+            _ -> % true
+                _ = erlang:process_flag(sensitive, true),
+                % Because erlang:process_info/1 Shows empty list:
+                []
+        end,
     FmtMisc =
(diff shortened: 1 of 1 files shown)


mikpe:
If this is applied, we'll need to split the "sensitive" process_flag into two, one which means "restrict access", and a new one that implements the "In crash dumps, the stack, messages, and the process dictionary are omitted" part of the current semantics.  We use sensitive on some processes exactly to prevent leaking data in crash dumps, but we also rely on sys:replace_state on occasion (we have long-running systems).

ferd:
Currently the sensitive processes also disable tracing and backtraces so the meaning is definitely extending further than crash dumps already.

I'd argue that the current description does line up better with the "everything goes":
> When a process has been marked as sensitive by calling process_flag(sensitive, true), features in the runtime system that can be used for examining the data or inner working of the process are silently disabled.

Except the one bit about "the runtime system"; since `sys` and OTP are technically above, I generally interpreted things as meaning that the hole left by `sys` was intentional.

lib/stdlib/src/sys.erl
```diff
@@ -510,27 +510,41 @@ do_get_state(Mod, Misc) ->
...
-	false ->
-	    try
-		NMisc = StateFun(Misc),
-		{NMisc, NMisc}
-	    catch
-		Cl:Exc ->
-		    {{error, {callback_failed,StateFun,{Cl,Exc}}}, Misc}
-	    end
-    end.
+Sensitivity = erlang:process_flag(sensitive, false),
```
essen: Disabling the sensitive flag here exposes the process information until you set the sensitive flag back to what it was. Meaning one can look into the sensitive process data by concurrently calling `replace_state` and `process_info`.

Pouriya-Jahanbakhsh: Yeah. I think we need to fetch this flag without changing it.

dszoboszlay:
I agree with @mikpe that we need a way to limit the amount of data logged in crash reports without making hot code upgrades and debugging more difficult. Also keep in mind the other (mis)use of the sensitive flag is on processes that are expected to have very large but uninteresting message queues and/or state in the process dictionary but holding no secrets at all. In this case the sensitive flag is just a convenient way to make the error logger less verbose in case of a crash.

Ideally I think `proc_lib` should provide an API for customising on a per process level whether to include messages and process dictionary in crash reports (the rest of the data cannot really grow in size).

As for our code upgrade use-case that this PR would interfere with, we can try to switch to `sys:change_code` and implementing the state transform code in the `code_change/3` callback - this would work for the `gen_*` processes at least.

dszoboszlay:
I have one philosophical concern about this PR though. So far the restrictions on sensitive processes seemed to be designed to prevent accidental data leaks: the threat is that generic error logging and monitoring tools that collect info on _all processes_ accidentally reveal secrets from sensitive processes. This kind of protection makes total sense to me.

But this change feels like it's protecting against active attacks, someone targeting one specific process to steal data from it. At least I can't imagine a scenario where a tool would like to `sys:replace_state` on all the processes to do something humble, but accidentally messing up sensitive processes. And this new threat model leads to dangerous territory, because I don't think Erlang could really protect a sensitive process against it. To exploit `sys:replace_state` I basically have to have arbitrary code execution in the Beam, at which point it's game over. OK, I can't use `sys:replace_state` after this patch, but I can e.g. load a compromised version of the `sys` module that doesn't have this check. Or any other module the sensitive process may call into. Or just read up the whole memory of the VM from the `/proc` file system. To protect against an attacker with arbitrary code execution, the sensitive process would at least have to run in a different OS process. This level of separation is not what the sensitive flag is meant to provide, at least in my humble opinion.

But don't get me wrong, there's nothing wrong with this PR, except that it may enforce a false sense of security.

Pouriya-Jahanbakhsh:
Thanks for sharing your opinions, and I'm glad you were so forthright about it!  
If the purpose of `sensitive` flag is to get empty lists for process dictionary and message queue just in case of calling `process_info` then this PR has nothing to do. But if the idea behind `sensitive` flag is to get empty lists for all OTP APIs then we need to correct this.
@dszoboszlay is right and we can't really protect sensitive data. But because of this conflict in OTP API calls I think the name `sensitive` should change to something like `verbose`.

rickard-green:
Closing this PR. Reasons for closing it has been expressed well by others above.

PR #2545, opened 2020-02-19
Introduce EEP-48 and help functions in shell

garazdawi:
This PR implements generation of [EEP-48](http://erlang.org/eeps/eep-0048.html) style documentation and also functions in the shell to display the documention.

This is a work in progress, but it should be functional enough for most usecases. At the moment it is not possible to create EEP-48 style documentation for anything but Erlang/OTP modules.

## Generation

The documentation is built when all other Erlang/OTP documentation is built. The format for the documentation is an erlang term format using three tuples describing an HTML document. Example:

```
[{p,[],
      [<<"Returns an integer or float that is the arithmetical absolute value of ">>,
       {a,[{type,anno}],[<<"Float">>]},
       <<" or ">>,
       {a,[{type,anno}],[<<"Int">>]},
       <<", for example:">>]},
   {pre,[],[<<"> abs(-3.33).\n3.33\n> abs(-3).\n3">>]},
   {p,[],[<<"Allowed in guard tests.">>]}]
```

Each function entry may also contains various metadata such as `edit_url`, `signature` and `since`.

The intent is for this format to be documented in the Kernel's User's Guides, but that document has not yet been written.

## Rendering

Six new functions have been introduced in the `shell_default` which means that they
will be available in all shells that do not implement their of `shell_default`. The
functions are `h(Module)`, `h(Module, Function)`, `h(Module, Function, Arity)`,
`ht(Module)`, `ht(Module, Function)` and `ht(Module, Function, Arity)`.

The `h/1` function displays the documentation for a module. Example:

```
> h(erlang).

        erlang

  By convention, most Built-In Functions (BIFs) are included in this
  module. Some of the BIFs are viewed more or less as part of the
  Erlang programming language and are auto-imported. Thus, it is
  not necessary to specify the module name. For example, the calls 
  atom_to_list(erlang) and erlang:atom_to_list(erlang) are
  identical.
```

The `h/2` and `h/3` functions show the function documentation. Example:

```
> h(erlang,abs,1).

-spec abs(Float) -> float() when Float :: float();
         (Int) -> non_neg_integer() when Int :: integer().


  Returns an integer or float that is the arithmetical absolute
  value of Float or Int, for example:

    > abs(-3.33).
    3.33
    > abs(-3).
    3

  Allowed in guard tests.
```

The `ht/1` function shows a list of all available types in a module. Example:

```
> ht(array).
   array

-opaque array(Type) ::
            #array{default :: Type, elements :: elements(Type)}.
-type array() :: array(term()).
-type array_indx() :: non_neg_integer().
-type array_opts() :: array_opt() | [array_opt()].
-type array_opt() ::
          {fixed, boolean()} |
          fixed |
          {default, Type :: term()} |
          {size, N :: non_neg_integer()} |
          (N :: non_neg_integer()).
-type indx_pairs(Type) :: [indx_pair(Type)].
-type indx_pair(Type) :: {Index :: array_indx(), Type}.
```

and finally `ht/2` and `ht/3` shows the documentation for a specific type. Example:

```
> ht(array,array,1).

-opaque array(Type) ::
            #array{default :: Type, elements :: elements(Type)}.


  A functional, extendible array. The representation is not
  documented and is subject to change without notice. Notice that
  arrays cannot be directly compared for equality.
```

## Extras

This PR also includes fixes to the documentation to make it possible to parse and functionality needed in order to build and render the documentation.

Added API's are:

* code:all_available()
  * Returns all modules that are either loaded or loadable in a system.
* code:get_doc(Module)
  * Returns the EEP-48 documentation for a specific module.
* shell_docs
  * A module that can be used to render EEP-48 style docs of the format "application/erlang+html".


essen:
Looks good! Should be easy enough to apply to my projects.

Querying with the arity is probably not going to work all that well though, and that's true also for OTP. For example `gen_server:enter_loop/4` http://erlang.org/doc/man/gen_server.html#enter_loop-3 has the fourth argument being either `ServerName` or `Timeout` and they make better sense when you consider `gen_server:enter_loop/3` and `/5` at the same time: you provide none, either one, or both. In those cases I'll probably provide the same documentation for all arities rather than try to split things up.

I notice that there are no title tags at the moment in the format, they can certainly be done via normal text but I wonder if any thoughts was given about them, or if they're absent only because OTP manuals don't use them.

garazdawi:
> In those cases I'll probably provide the same documentation for all arities rather than try to split things up.

Yes, this is what I decided to do as well.

> I notice that there are no title tags at the moment in the format, they can certainly be done via normal text but I wonder if any thoughts was given about them, or if they're absent only because OTP manuals don't use them.

At the moment there are no header tags, but this morning I found some uses of them so I think I will add them. I assume `<h1>`, `<h2>` and `<h3>` should be enough?

essen:
> > I notice that there are no title tags at the moment in the format, they can certainly be done via normal text but I wonder if any thoughts was given about them, or if they're absent only because OTP manuals don't use them.
> 
> At the moment there are no header tags, but this morning I found some uses of them so I think I will add them. I assume `<h1>`, `<h2>` and `<h3>` should be enough?

Yes, I don't think I ever use more than 2 header levels in my projects, excluding the signature (so 3 header levels total at most depending on how it gets rendered). But even then 2 header levels is rare, it's typically the document title (which would be the signature here) and section titles on the same level.

One additional question. Should callbacks be documented as well, and if so how? It's probably difficult to do it outside of the module itself considering `Module:StateName/3` http://erlang.org/doc/man/gen_statem.html#Module:StateName-3

garazdawi:
> One additional question. Should callbacks be documented as well, and if so how?

Yes, they should. It is on my todo list. The EEP suggests doing `{{callback,'StateName',3},....}` and that seems like a good way to go. Then I'll just expose `hcb` or something like that in the shell to inspect them.

lib/kernel/src/code.erl
```diff
@@ -785,6 +834,97 @@ where_is_file(Tail, File, Path, Files) ->
...
+                {ok, Bin} ->
+                    {ok, binary_to_term(Bin)};
+                {error,enoent} ->
+                    get_doc_chunk(Dir, Mod);
+                {error,Reason} ->
+                    {error,Reason}
+            end
+    end.
+
+get_doc_chunk_from_ast(Filename) ->
```
josevalim: I wonder if building a "fake chunk" from the AST on the fly is the way to go. Maybe `code:get_doc/1` should remain truthful to what is actually in the chunk/disk and returned an error in cases like this. What are the cases the "fake chunk" is necessary? :)

lib/stdlib/src/shell_docs.erl
```diff
@@ -0,0 +1,672 @@
...
+normalize_trim([{Tag,Attr,Content}|T],Trim) ->
+    [{Tag,Attr,normalize_trim(Content,Trim)} | normalize_trim(T,Trim)];
+normalize_trim([<<>>|T],Trim) ->
+    normalize_trim(T,Trim);
+normalize_trim([B1,B2|T],Trim) when is_binary(B1),is_binary(B2) ->
+    normalize_trim([<<B1/binary,B2/binary>> | T],Trim);
+normalize_trim([H|T],Trim) ->
+    [normalize_trim(H,Trim) | normalize_trim(T,Trim)];
+normalize_trim([],_Trim) ->
+    [].
```
josevalim: Should we do this when writing to the chunk? 🤔 Otherwise every tool processing the chunk will have to do it.

josevalim: Please ignore. I see this is already used by docgen. :D

josevalim:
> At the moment there are no header tags, but this morning I found some uses of them so I think I will add them. I assume <h1>, <h2> and <h3> should be enough?

FWIW, Elixir assumes that <h1> is always the signature (i.e. it is implicit, you never write it yourself) and people may use <h2> and <h3> in the docs. For instance, we would have a <h2>Examples</h2> heading when listing examples.

lib/stdlib/src/shell_docs.erl
garazdawi: My plan is to call it both when generating and when rendering. That way tools that generate can skip it if they want and renderers that want to render Erlang/OTP docs dont have to normalize.

lib/kernel/src/code.erl
garazdawi: I added it mostly because for modules without any chunk but still with a spec I thought it would be nice of `c:h` to actually return the spec.

This is mainly for erlang libraries without any API docs at all.

garazdawi:
> FWIW, Elixir assumes that <h1> is always 0the signature (i.e. it is implicit, you never write it yourself) and people may use <h2> and <h3> in the docs. For instance, we would have a <h2>Examples</h2> heading when listing examples.

Yes, I've pushed changes that does the same for our docs. See `h(ssh)` as an example.

lib/kernel/src/code.erl
josevalim: I see. We do the same but the generation of the "fake chunk" is inside our "shell_docs". I guess some tools would find the "fake chunk" useful but others may not, so I guess both approaches work. Thanks for the info!

lib/stdlib/src/shell_docs.erl
josevalim: Fantastic!

lib/kernel/src/code.erl
```diff
@@ -785,6 +834,97 @@ where_is_file(Tail, File, Path, Files) ->
...
+    case beam_lib:chunks(Filename, [abstract_code]) of
+        {error,beam_lib,{missing_chunk,_,_}} ->
+            {error,missing};
+        {ok, {_Mod, [{abstract_code,
+                      {raw_abstract_v1, AST}}]}} ->
+            Docs = get_function_docs_from_ast(AST),
+            {ok, #docs_v1{ anno = 0, beam_language = erlang, format =  <<"text/erlang_doc">>,
```
josevalim: Should this be `?NATIVE_FORMAT` or `<<"text/erlang_doc">>`?

lib/erl_docgen/priv/bin/chunk.escript
```diff
@@ -0,0 +1,738 @@
...
+                          MetaWSpec = add_spec(AST,FMeta),
+                          [docs_v1_entry(function, Anno, Name, Arity, Signature, MetaWSpec, Fdoc)]
+                  end
+          end, Functions),
+
+    docs_v1(ModuleDocs, Anno, TypeMeta, FuncEntrys ++ TypeEntries).
+
+docs_v1(DocContents, Anno, Metadata, Docs) ->
+    #docs_v1{ anno = Anno,
+              module_doc = #{<<"en">> => term_to_binary(shell_docs:normalize(DocContents))},
```
josevalim: We had discussions about this and we were considering updating EEP 48 to allow any term to be given here, so we don't do a double `term_to_binary`. What are your final thoughts? I can send a PR to EEP 48 later today if we want to lift this restriction. We already need to update it to mention `edit_url`.

lib/kernel/src/code.erl
garazdawi: fixed

lib/erl_docgen/priv/bin/chunk.escript
garazdawi: Yes, I think that would be a nice change.

josevalim: PR sent: https://github.com/erlang/eep/pull/9

wojtekmach:
@garazdawi this is fantastic!

I started using the chunk in ExDoc. There's more work to be done on our side with regards to type specs in particular, but here are our early results: [1]

image

Given ExDoc code highlighting and auto linking currently works on markdown the way it works is: `erlang+html->markdown->html` but our end goal is to have `erlang+html->html`. [2]

As I was interpreting the `erlang+html` format I noticed non-standard tags like:

* `c`
* `anno`

is the intention to keep them or transform them to html equivalents before storing in the chunk?

[1] http://wojtekmach.pl/otp_docs/stdlib/array.html
[2] https://github.com/elixir-lang/ex_doc/commit/d259e6ffef1703a2b9744ee8b7ab98fed394cb43#diff-fc9775865b82e656523d7cbbe16526d2R102

garazdawi:
> As I was interpreting the erlang+html format I noticed non-standard tags like:
> * `c`
> * `anno`
> is the intention to keep them or transform them to html equivalents before storing in the chunk?

I think we will remove `anno` and translate `c` to `code` and then we would translate our `code` to `pre`.

josevalim:
@garazdawi Proposal: translate `c` to `code` and `code` to `<pre><code>` since `<pre>` by itself does not mean code is inside. If you have multiple languages, then you can also emit `<pre><code class="some-language">` or `<pre><code data-language="some-language">`. If no language is `code`, then downstream can assume the default is the one declare in the chunk.

lib/erl_docgen/priv/bin/chunk.escript
```diff
@@ -0,0 +1,747 @@
...
+              {datatype,NameAttr,ContentsNoName};
+          {name,[],Content} ->
+              [{Name,Arity}] = func_to_tuple(Content),
+              Signature = strip_tags(Content),
+              {datatype,[{name,Name},{n_vars,integer_to_list(Arity)},
+                         {signature,Signature}],ContentsNoName}
+      end || N = {name,_,_} <- Dom].
+
+transform_seealso(_S = {seealso,_Attr,_Content}) ->
+    _Content.
```
wojtekmach: here are couple examples of `seealso` in action in xml:

* `<p>See also <seealso marker="#new-2"><c>new/2</c></seealso>`
* `<seealso marker="kernel:file"><c>file</c></seealso> module.</p>`

currently we transform this into `{c, [], ["new/2"]}` and `{c, [], ["file"]}` respectively. This is a bit broader discussion and probably not a big priority at the moment, but currently we lose semantic information which is especially relevant for the `file` atom: was it suppose to mean just an _atom_ or a _module_? To preserve references, perhaps this should be transformed into something like:

* `{a, [{href, "#new-2"}], [{c, [], "new/2"}]}`
* `{a, [{href, "kernel:file"}], [{c, [], "file"}]}`

and then the downstream tools could interpret these to create proper cross-references. To differentiate from external URLs (we sometimes link to e.g. RFCs) the `a` element could have additional attributes.

josevalim: EEP 48 supports cross references in metadata, so we could move those there. This may even allow us to remove paragraphs such as above completely and instead automatically generate them.

wojtekmach: So the more complete snippet from `filelib.xml` is:

```
  <description ghlink="maint/lib/stdlib/doc/src/filelib.xml#L38">
    <p>This module contains utilities on a higher level than the
      <seealso marker="kernel:file"><c>file</c></seealso> module.</p>
```

is the idea that this would be converted to:

```html
<p>This module contains utilities on a higher level than the
  <code>file</code>module.</p>

See also: <a href="kernel:file"><code>file</code></a>, ...
```

that is, we wouldn't have "inline" cross-references but they'd always happen and the end of a given block of docs?

josevalim: I think you would want both. Sometimes you want to provide some information when linking, but sometimes a direct “see also”, like in Man pages, is desired.

garazdawi:
> Proposal: translate `c` to `code` and `code` to `<pre><code>` since `<pre>` by itself does not mean code is inside. If you have multiple languages, then you can also emit `<pre><code class="some-language">` or `<pre><code data-language="some-language">`. If no language is `code`, then downstream can assume the default is the one declare in the chunk.

Sure, but then I think that just `<pre>` should also be `<pre><code>` as that is what is rendered in our current html docs.

lib/erl_docgen/priv/bin/chunk.escript
garazdawi: I've not really looked at any of the linking yet as it is not needed for the shell docs. I don't have time to look at it now so it will have to wait until after OTP 23-rc1.

garazdawi: Having re-read what you wrote.

Yes, `seealso` should be translated to a `<a>` tag. The reason why it is not done right now is because it was not needed for the shell documentation. Also, we do not want to do a direct mapping of what we have in `marker` to what is in the `<a>` as some of the information is unnecessarily hard to figure out later on.

In order to get something working at all with links I'll include `<a marker="kernel:file">` in our chunks for now. But don't do too much work on parsing these yet as they are bound to change in OTP 23-rc2.

wojtekmach: Sounds good, thank you!

lib/erl_docgen/priv/bin/chunk.escript
garazdawi: thanks!

KennethL:
We have discussed to translate the see also to a but with some nonstandard
attribute like linktype which should indicate if it is a reference to a
module, function or application or even type, since this is important for
the html and we already have that info hidden in the format of the href
value which we can parse during transformation to chunk. Will happen after
23rc1

/Kenneth

essen:
> We have discussed to translate the see also to a but with some nonstandard attribute like linktype which should indicate if it is a reference to a module, function or application or even type, since this is important for the html and we already have that info hidden in the format of the href value which we can parse during transformation to chunk. Will happen after 23rc1 /Kenneth

You don't need to use a non-standard attribute, this is what `rel` is for. If the relation type cannot be found in those that are already registered ( https://www.iana.org/assignments/link-relations/link-relations.xhtml ), you can use a URI as the value for `rel`. For example:

```
<a href="..." rel="https://erlang.org/function">kernel:file</a>
```

The URI doesn't have to actually exist though if it does it should point to a description of the relation type.

It's also possible to register new types at IANA but might not be worth the effort.

wojtekmach:
For anyone interested I create a list of all unique tag/attribute-names combinations currently present in erlang+html chunks on top master:

```
[
  a: [:href],
  a: [:id],
  a: [:marker],
  br: [],
  code: [],
  code: [:type],
  dd: [],
  dl: [],
  dt: [],
  em: [],
  h2: [],
  h3: [],
  i: [],
  li: [],
  li: [:class],
  li: [:name],
  li: [:name, :class],
  ol: [],
  p: [],
  p: [:class],
  p: [:ghlink],
  pre: [],
  ul: [],
  ul: [:class]
]
```

(https://gist.github.com/wojtekmach/60e5bf7a79cac7a940ce0daecb070066)

Thanks again @garazdawi

PR #2546, opened 2020-02-19
Don't keep stacktraces forever

bjorng:
The `erlang:get_stacktrace/0` BIF retrieves the stacktrace from
the previous error in the process. The problem is that the very
existence of `erlang:get_stacktrace/0` means that the stacktrace
and potentially function arguments must be kept indefinitely.

Therefore, in OTP 21, the `erlang:get_stacktrace/0` BIF was deprecated
and the syntax of try/catch extended to allow matching out the
stacktrace directly.

This commit changes `erlang:get_stacktrace/0` for OTP 23 to always
return an empty list (`[]`) and eliminates the need to keep the
stacktrace forever.

`erlang:get_stacktrace/0` is scheduled for removal in OTP 24.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -2120,8 +2120,9 @@ true</pre>
       <fsummary>Get the call stack back-trace of the last exception.</fsummary>
       <type name="stack_item"/>
       <desc>
-	<warning><p><c>erlang:get_stacktrace/0</c> is deprecated and will stop working
-	in a future release.</p></warning>
+	<warning><p><c>erlang:get_stacktrace/0</c> is deprecated and
+	will be removed in OTP 24. Starting from OTP 23,
+	<c>erlang:get_stacktrace/0</c> returns <c>[]</c>.</p></warning>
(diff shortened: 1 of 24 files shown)


erts/doc/src/erlang.xml
```diff
@@ -2120,8 +2120,10 @@ true</pre>
       <fsummary>Get the call stack back-trace of the last exception.</fsummary>
       <type name="stack_item"/>
       <desc>
-	<warning><p><c>erlang:get_stacktrace/0</c> is deprecated and will stop working
-	in a future release.</p></warning>
+	<warning><p><c>erlang:get_stacktrace/0</c> is deprecated and
+	will be removed in OTP 24. Starting from OTP 23,
+	<c>erlang:get_stacktrace/0</c> returns <c>[]</c>.</p></warning>
+      <p><c>erlang:get_stacktrace/0</c> returns <c>[]</c>.</p>
```
jhogberg: Nit: are the double `<c>erlang:get_stacktrace/0</c> ...` intentional?

PR #2547, opened 2020-02-22
Implement RFC 6066 extension max_fragment_length

dotsimon:
This PR implements the max_fragment_length extension specified in RFC 6066, allowing a client to negotiate a smaller plaintext.

Support is provided for DTLS/TLS 1, 1.1 & 1.2 as well as TLS 1.3

### DIFF
--- a/lib/ssl/src/dtls_connection.erl
+++ b/lib/ssl/src/dtls_connection.erl
@@ -448,9 +448,12 @@ init({call, From}, {start, Timeout},
     Hello = dtls_handshake:client_hello(Host, Port, ConnectionStates0, SslOpts,
 					Session#session.session_id, Renegotiation, Cert),
 
+    MaxFragEnum = maps:get(max_frag_enum, Hello#client_hello.extensions, undefined),
+    ConnectionStates1 = ssl_record:set_max_fragment_length(MaxFragEnum, ConnectionStates0),
     Version = Hello#client_hello.client_version,
     HelloVersion = dtls_record:hello_version(Version, Versions),
-    State1 = prepare_flight(State0#state{connection_env = CEnv#connection_env{negotiated_version = Version}}),
+    State1 = prepare_flight(State0#state{connection_env = CEnv#connection_env{negotiated_version = Version},
+                                         connection_states = ConnectionStates1}),
     {State2, Actions} = send_handshake(Hello, State1#state{connection_env = CEnv#connection_env{negotiated_version = HelloVersion}}),  
     State = State2#state{connection_env = CEnv#connection_env{negotiated_version = Version}, %% RequestedVersion
 			  session = Session,
@@ -580,8 +583,9 @@ hello(internal, #server_hello{} = Hello,
          handshake_env = #handshake_env{renegotiation = {Renegotiation, _}},
          connection_env = #connection_env{negotiated_version = ReqVersion},
          connection_states = ConnectionStates0,
+         session = #session{session_id = OldId},
          ssl_options = SslOptions} = State) ->
-    case dtls_handshake:hello(Hello, SslOptions, ConnectionStates0, Renegotiation) of
+    case dtls_handshake:hello(Hello, SslOptions, ConnectionStates0, Renegotiation, OldId) of
 	#alert{} = Alert ->
 	    handle_own_alert(Alert, ReqVersion, ?FUNCTION_NAME, State);
 	{Version, NewId, ConnectionStates, ProtoExt, Protocol} ->
@@ -1107,9 +1111,11 @@ send_handshake_flight(#state{static_env = #static_env{socket = Socket,
 			     connection_states = ConnectionStates0,
                              ssl_options = #{log_level := LogLevel}} = State0,
                       Epoch) ->
-    %% TODO remove hardcoded Max size
+    PMTUEstimate = 1400, %% TODO make configurable
+    #{current_write := #{max_fragment_length := MaxFragmentLength}} = ConnectionStates0,
+    MaxSize = min(MaxFragmentLength, PMTUEstimate),
     {Encoded, ConnectionStates} =
-	encode_handshake_flight(lists:reverse(Flight), Version, 1400, Epoch, ConnectionStates0),
+	encode_handshake_flight(lists:reverse(Flight), Version, MaxSize, Epoch, ConnectionStates0),
     send(Transport, Socket, Encoded),
     ssl_logger:debug(LogLevel, outbound, 'record', Encoded),
    {State0#state{connection_states = ConnectionStates}, []};
@@ -1123,8 +1129,11 @@ send_handshake_flight(#state{static_env = #static_env{socket = Socket,
 			     connection_states = ConnectionStates0,
                              ssl_options = #{log_level := LogLevel}} = State0,
                       Epoch) ->
+    PMTUEstimate = 1400, %% TODO make configurable
+    #{current_write := #{max_fragment_length := MaxFragmentLength}} = ConnectionStates0,
+    MaxSize = min(MaxFragmentLength, PMTUEstimate),
     {HsBefore, ConnectionStates1} =
-	encode_handshake_flight(lists:reverse(Flight0), Version, 1400, Epoch, ConnectionStates0),
+	encode_handshake_flight(lists:reverse(Flight0), Version, MaxSize, Epoch, ConnectionStates0),
     {EncChangeCipher, ConnectionStates} = encode_change_cipher(ChangeCipher, Version, Epoch, ConnectionStates1),
 
     send(Transport, Socket, [HsBefore, EncChangeCipher]),
@@ -1141,12 +1150,15 @@ send_handshake_flight(#state{static_env = #static_env{socket = Socket,
 			     connection_states = ConnectionStates0,
                              ssl_options = #{log_level := LogLevel}} = State0,
                       Epoch) ->
+    PMTUEstimate = 1400, %% TODO make configurable
+    #{current_write := #{max_fragment_length := MaxFragmentLength}} = ConnectionStates0,
+    MaxSize = min(MaxFragmentLength, PMTUEstimate),
     {HsBefore, ConnectionStates1} =
-	encode_handshake_flight(lists:reverse(Flight0), Version, 1400, Epoch-1, ConnectionStates0),
+	encode_handshake_flight(lists:reverse(Flight0), Version, MaxSize, Epoch-1, ConnectionStates0),
     {EncChangeCipher, ConnectionStates2} = 
 	encode_change_cipher(ChangeCipher, Version, Epoch-1, ConnectionStates1),
     {HsAfter, ConnectionStates} =
-	encode_handshake_flight(lists:reverse(Flight1), Version, 1400, Epoch, ConnectionStates2),
+	encode_handshake_flight(lists:reverse(Flight1), Version, MaxSize, Epoch, ConnectionStates2),
     send(Transport, Socket, [HsBefore, EncChangeCipher, HsAfter]),
     ssl_logger:debug(LogLevel, outbound, 'record', [HsBefore]),
     ssl_logger:debug(LogLevel, outbound, 'record', [EncChangeCipher]),
@@ -1162,10 +1174,13 @@ send_handshake_flight(#state{static_env = #static_env{socket = Socket,
 			     connection_states = ConnectionStates0,
                              ssl_options = #{log_level := LogLevel}} = State0,
                       Epoch) ->
+    PMTUEstimate = 1400, %% TODO make configurable
+    #{current_write := #{max_fragment_length := MaxFragmentLength}} = ConnectionStates0,
+    MaxSize = min(MaxFragmentLength, PMTUEstimate),
     {EncChangeCipher, ConnectionStates1} = 
 	encode_change_cipher(ChangeCipher, Version, Epoch-1, ConnectionStates0),
     {HsAfter, ConnectionStates} =
-	encode_handshake_flight(lists:reverse(Flight1), Version, 1400, Epoch, ConnectionStates1),
+	encode_handshake_flight(lists:reverse(Flight1), Version, MaxSize, Epoch, ConnectionStates1),
     send(Transport, Socket, [EncChangeCipher, HsAfter]),
     ssl_logger:debug(LogLevel, outbound, 'record', [EncChangeCipher]),
     ssl_logger:debug(LogLevel, outbound, 'record', [HsAfter]),
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -392,6 +392,7 @@
                                 {psk_identity, client_psk_identity()} |
                                 {srp_identity, client_srp_identity()} |
                                 {server_name_indication, sni()} |
+                                {max_fragment_length, max_fragment_length()} |
                                 {customize_hostname_check, customize_hostname_check()} |
                                 {signature_algs, client_signature_algs()} |
                                 {fallback, fallback()} |
@@ -414,6 +415,7 @@
 -type client_srp_identity()             :: srp_identity().
 -type customize_hostname_check() :: list().
 -type sni()                      :: HostName :: hostname() | disable. 
+-type max_fragment_length()      :: undefined | 512 | 1024 | 2048 | 4096.
 -type client_signature_algs()    :: signature_algs().
 -type fallback()                 :: boolean().
 -type ssl_imp()                  :: new | old.
@@ -465,6 +467,7 @@
                                   alpn =>  app_level_protocol(),
                                   srp  => binary(),
                                   next_protocol => app_level_protocol(),
+                                  max_frag_enum  => 1..4,
                                   ec_point_formats  => [0..2],
                                   elliptic_curves => [public_key:oid()],
                                   sni => hostname()}. % exported
@@ -2257,6 +2260,13 @@ validate_option(server_name_indication, undefined) ->
     undefined;
 validate_option(server_name_indication, disable) ->
     disable;
+
+%% RFC 6066, Section 4
+validate_option(max_fragment_length, I) when I == ?MAX_FRAGMENT_LENGTH_BYTES_1; I == ?MAX_FRAGMENT_LENGTH_BYTES_2;
+                                             I == ?MAX_FRAGMENT_LENGTH_BYTES_3; I == ?MAX_FRAGMENT_LENGTH_BYTES_4 ->
+    I;
+validate_option(max_fragment_length, undefined) ->
+    undefined;
 
 validate_option(sni_hosts, []) ->
     [];
--- a/lib/ssl/src/ssl_handshake.erl
+++ b/lib/ssl/src/ssl_handshake.erl
@@ -73,11 +73,11 @@
 
 %% Extensions handling
 -export([client_hello_extensions/7,
-	 handle_client_hello_extensions/9, %% Returns server hello extensions
-	 handle_server_hello_extensions/9, select_curve/2, select_curve/3,
+	 handle_client_hello_extensions/10, %% Returns server hello extensions
+	 handle_server_hello_extensions/10, select_curve/2, select_curve/3,
          select_hashsign/4, select_hashsign/5,
 	 select_hashsign_algs/3, empty_extensions/2, add_server_share/3,
-	 add_alpn/2, add_selected_version/1, decode_alpn/1
+	 add_alpn/2, add_selected_version/1, decode_alpn/1, max_frag_enum/1
 	]).
 
 -export([get_cert_params/1,
@@ -698,6 +698,10 @@ encode_extensions([#sni{hostname = Hostname} | Rest], Acc) ->
                               ?BYTE(?SNI_NAMETYPE_HOST_NAME),
                               ?UINT16(HostLen), HostnameBin/binary,
                               Acc/binary>>);
+encode_extensions([#max_frag_enum{enum = MaxFragEnum} | Rest], Acc) ->
+    ExtLength = 1,
+    encode_extensions(Rest, <<?UINT16(?MAX_FRAGMENT_LENGTH_EXT), ?UINT16(ExtLength), ?BYTE(MaxFragEnum),
+                              Acc/binary>>);
 encode_extensions([#client_hello_versions{versions = Versions0} | Rest], Acc) ->
     Versions = encode_versions(Versions0),
     VerLen = byte_size(Versions),
@@ -1091,7 +1095,8 @@ client_hello_extensions(Version, CipherSuites, SslOpts, ConnectionStates, Renego
 add_tls12_extensions(_Version,
                      #{alpn_advertised_protocols := AlpnAdvertisedProtocols,
                       next_protocol_selector := NextProtocolSelector,
-                      server_name_indication := ServerNameIndication} = SslOpts,
+                      server_name_indication := ServerNameIndication,
+                      max_fragment_length := MaxFragmentLength} = SslOpts,
                      ConnectionStates,
                      Renegotiation) ->
     SRP = srp_user(SslOpts),
@@ -1102,7 +1107,8 @@ add_tls12_extensions(_Version,
       next_protocol_negotiation =>
           encode_client_protocol_negotiation(NextProtocolSelector,
                                              Renegotiation),
-      sni => sni(ServerNameIndication)
+      sni => sni(ServerNameIndication),
+      max_frag_enum => max_frag_enum(MaxFragmentLength)
      }.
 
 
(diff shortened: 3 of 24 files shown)


dotsimon:
Found a bug with {handshake, hello}, fixed & added test case

IngelaAndin:
Your PR is certainly interesting and I think that we can get it into a mergeable shape. Alas there is not time to squeeze this in before OTP-22.3 so it will be included in an OTP-23.X release (where X might be 0 depending on how much time it will take). We have been quite preoccupied lately so please be patient and we will come back with some code review comments "soonish"

lib/ssl/src/ssl_handshake.hrl
```diff
@@ -389,8 +389,14 @@
...
 
+%% enum{ 2^9(1), 2^10(2), 2^11(3), 2^12(4), (255) } MaxFragmentLength;
+-define(MFL_EXT, 1).
+
+-record(mfl, {
+          mfl_enum = undefined  %% contains the enum value 1..4
+        }).
+
 %% Other possible values from RFC 6066, not supported
--define(MAX_FRAGMENT_LENGTH, 1).
```
peterdmv: We think it's better to keep this longer name for this macro.

lib/ssl/src/ssl_handshake.hrl
```diff
@@ -389,8 +389,14 @@
           hostname = undefined
         }).
 
+%% enum{ 2^9(1), 2^10(2), 2^11(3), 2^12(4), (255) } MaxFragmentLength;
+-define(MFL_EXT, 1).
+
+-record(mfl, {
```
peterdmv: mfl -> max_frag_len

lib/ssl/src/ssl_handshake.hrl
```diff
@@ -389,8 +389,14 @@
...
+          mfl_enum = undefined  %% contains the enum value 1..4
```
peterdmv: mfl_enum -> max_fragment_length

lib/ssl/src/ssl_handshake.erl
```diff
@@ -689,6 +689,10 @@ encode_extensions([#sni{hostname = Hostname} | Rest], Acc) ->
                               ?BYTE(?SNI_NAMETYPE_HOST_NAME),
                               ?UINT16(HostLen), HostnameBin/binary,
                               Acc/binary>>);
+encode_extensions([#mfl{mfl_enum = MFL} | Rest], Acc) ->
```
peterdmv: Please see our comments on the record type.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1095,7 +1100,8 @@ add_tls12_extensions(_Version,
       next_protocol_negotiation =>
           encode_client_protocol_negotiation(NextProtocolSelector,
                                              Renegotiation),
-      sni => sni(ServerNameIndication)
+      sni => sni(ServerNameIndication),
+      mfl => mfl(MaxFragmentLength)
```
peterdmv: We think this belong to the common extensions as it can be present in all TLS versions.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2558,6 +2590,12 @@ decode_extensions(<<?UINT16(?SNI_EXT), ?UINT16(Len),
     decode_extensions(Rest, Version, MessageType,
                       Acc#{sni => dec_sni(NameList)});
 
+decode_extensions(<<?UINT16(?MFL_EXT), ?UINT16(1), ?BYTE(MFL), Rest/binary>>,
```
peterdmv: > If a server receives a maximum fragment length negotiation request    for a value other than the allowed values, it MUST abort the  handshake with an "illegal_parameter" alert.

If we receive a faulty value the whole extension will be dropped but we should instead send an illegal_parameter alert.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2919,6 +2957,13 @@ handle_alpn_extension([ServerProtocol|Tail], ClientProtocols) ->
             false -> handle_alpn_extension(Tail, ClientProtocols)
     end.
 
+handle_mfl_extension(#mfl{mfl_enum = MFLEnum}=MFL) when MFLEnum >= 1, MFLEnum =< 4 ->
+    MFL;
+handle_mfl_extension(#mfl{}) ->
+    throw(?ALERT_REC(?FATAL, ?ILLEGAL_PARAMETER));
```
peterdmv: This will never throw because the extension is dropped before.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -3137,6 +3182,17 @@ sni(disable) ->
 sni(Hostname) ->
     #sni{hostname = Hostname}.
 
+mfl(512) ->
+    #mfl{mfl_enum = 1};
```
peterdmv: Please use macros for the enumerated values. E.g
```erlang
-define(MAX_FRAG_LEN_512, 1).
%% ...
#max_frag_len{max_fragment_length = ?MAX_FRAG_LEN_512};
```

lib/ssl/src/tls_connection.erl
```diff
@@ -601,8 +601,15 @@ init({call, From}, {start, Timeout},
     %% Update pre_shared_key extension with binders (TLS 1.3)
     Hello1 = tls_handshake_1_3:maybe_add_binders(Hello, TicketData, HelloVersion),
 
+    ConnectionStates1 = case maps:get(mfl, Hello1#client_hello.extensions, undefined) of
```
peterdmv: I think code could be moved into a separate function. I appears two times.

IngelaAndin:
Apart from fixing review comments would you also please rebase your PR on master as our next
planned release is OTP-23.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -552,13 +560,14 @@ build_content(Context, THash) ->
...
                        extensions = Extensions} = _Hello,
-         #state{connection_states = _ConnectionStates0,
+         #state{connection_states = ConnectionStates0,
                 ssl_options = #{ciphers := ServerCiphers,
                                 signature_algs := ServerSignAlgs,
                                 supported_groups := ServerGroups0,
                                 alpn_preferred_protocols := ALPNPreferredProtocols,
                                 honor_cipher_order := HonorCipherOrder},
-                session = #session{own_certificate = Cert}} = State0) ->
+                session = #session{own_certificate = Cert}} = State00) ->
```
peterdmv: It would be easier to read the code if the state variables would follow the same naming convention: starting with State0 and incrementing all other occurrences.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -922,6 +941,16 @@ do_wait_ee(#encrypted_extensions{extensions = Extensions}, State0) ->
     {Ref, Maybe} = maybe(),
 
     try
+        %% RFC 6066: handle received/expected maximum fragment length
```
peterdmv: It would be cleaner if this block of code would be factored into a separate function following the pattern e.g.
```erlang
Maybe(handle_max_fragment_length(State0, Extensions)),
```

lib/ssl/src/tls_record.erl
```diff
@@ -112,12 +112,18 @@ encode_handshake(Frag, {3, 4}, ConnectionStates) ->
 encode_handshake(Frag, Version, 
 		 #{current_write :=
 		       #{beast_mitigation := BeastMitigation,
+                         max_fragment_length := MaxFragmentLength,
 			  security_parameters :=
 			     #security_parameters{bulk_cipher_algorithm = BCA}}} = 
 		     ConnectionStates) ->
+    MaxLength = if is_integer(MaxFragmentLength) ->
```
peterdmv: This code block appears twice, it could be factored into a separate function.

peterdmv:
I've run into a new standard while doing the code review: [Record Size Limit Extension for TLS](https://tools.ietf.org/html/rfc8446). This new RFC deprecates the max_fragment_length extension for multiple technical reasons and the IANA registry lists max_fragment_length as a not recommended extension.
What is your use case? Wouldn't it be better to implement the latest standard that corrects many features of the previous one?

dotsimon:
Hi,
The point to implement RFC 6066 is that it is already in use in M2M applications. RFC 8449 was borne out of discussion about browsers and unfortunately missing the pure client-based constraints of some embedded system.
If you really feel strongly about RFC 8449, I can remove all the client-side max_fragment_length handling. Then you will still have a more feature rich server without a "deprecated" client.
Unless I get a lot more spare time (or a use-case) I won't be implementing RFC 8449
.simon

peterdmv:
We've just had an internal discussion about these RFCs and we think you can go ahead with the implementation of RFC 6066 (both client and server side).

IngelaAndin:
Just to clarify, we reason that we probably will add support for the new RFC sometime in the future. Interoperability with the old extension is still a valid use case, and as long as the default behavior of
TLS is not changed (I do believe that is not the case in your PR) it can be a tradeoff left to the user.
Probably we ought to mention in the documentation something about the security concerns to think of before enabling this option. It is good for testing to have both sides being able to do it.

IngelaAndin:
@dotsimon  ping

dotsimon:
I haven't abandoned this PR, just a few more important things going on in the world.
WRT the change requests I will
1) revisit the DTLS part for correctness
2) check on if illegal parameter is sent in all cases it should
As far as I can see all the other changes are administrative or cosmetic so if they're important to you then I will roll them in. However, I have a few comments :-)

- The use of atoms for setting the client max_fragment_length option seems weird to me. I can't imagine anyone using the Erlang/OTP client in this way except for testing in which case you should already know what you're trying to do. Also the allowed values are documented. An alternate approach taken in openssl is to round down to the nearest allowed value. Think about it and let me know which way you want to go.

- In the code mfl vs max_fragment_length was to distinguish between the places where the value was the enum (mfl) or the number of octets (max_fragment_length). Not the cleverest solution but at least it helped keep it consistent. I'm not opposed to making mfl something more readable but I think the suggestion of max_frag_len is too much like max_fragment_length possibly leading to confusion. At the other extreme, max_fragment_length_enum seems a bit overblown.

.simon

IngelaAndin:
That is quite alright, we are not in a big hurry just wondering what happened when it was so quiet for a while. 

RFCs are full of details, I find you often have to read them many times to get it right. To not speak of newer RFCs that might update the behavior of old ones. So if we misunderstood something your welcome to try convince us of your understanding.

Regarding atoms we will discuss and get back to you. On the part of mfl,  I then suggest max_frag_enum.

lib/ssl/src/ssl.erl
```diff
@@ -2132,6 +2135,12 @@ validate_option(server_name_indication, undefined) ->
 validate_option(server_name_indication, disable) ->
     disable;
 
+%% RFC 6066, Section 4
+validate_option(max_fragment_length, I) when I == 512; I == 1024; I == 2048; I == 4096 ->
```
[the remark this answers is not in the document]
dotsimon: Currently the case of faulty input (ie. the value is not one of the allowed integers, or undefined) throws
`{error, {options, {max_fragment_length, FaultyValue}}}` around line 2238

lib/ssl/src/ssl_handshake.erl
dotsimon: SNI is also applicable to other TLS versions but resides in add_tls12_extensions which is why I put MFL here.
Please confirm you want MFL in the common extensions.

lib/ssl/src/dtls_connection.erl
```diff
@@ -1107,9 +1117,9 @@ send_handshake_flight(#state{static_env = #static_env{socket = Socket,
 			     connection_states = ConnectionStates0,
                              ssl_options = #{log_level := LogLevel}} = State0,
                       Epoch) ->
-    %% TODO remove hardcoded Max size
+    #{current_write := #{max_fragment_length := MaxFragmentLength}} = ConnectionStates0,
     {Encoded, ConnectionStates} =
-	encode_handshake_flight(lists:reverse(Flight), Version, 1400, Epoch, ConnectionStates0),
+	encode_handshake_flight(lists:reverse(Flight), Version, MaxFragmentLength, Epoch, ConnectionStates0),
```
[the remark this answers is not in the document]
dotsimon: As I see it, encode_handshake_flight/5 calls dtls_handshake:fragment_handshake/2 with the plaintext (Flight) and MaxFragmentLength. So the plaintext is being broken up into chunks not larger than MaxFragmentLength.
dtls_record:encode_handshake/4 handles the ciphering so the final packet is generated there which will be larger than the hardcoded 1400.
I agree the hardcoding the MTU is a problem but the code does what it is supposed to with the MaxFragmentLength operating on the plaintext.
If you prefer I can make this max(1400, MaxFragmentLength) to keep the new option from making the DTLS plaintext larger than what would be generated with the existing code.

dotsimon:
Rebased to master and also left some comments regarding Péter's review. Could you respond to those and also if you made a decision about the atom vs integer for the ssl client option?

dotsimon:
I rebased again to resolve a whitespace conflict introduced by d8076b77d12

lib/ssl/src/dtls_connection.erl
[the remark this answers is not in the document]
peterdmv: I have some thoughts here that the hardcoded size was never really applied. Maybe, we should hardcode it to `?MAX_PLAIN_TEXT_LENGTH` to get the original behavior? @IngelaAndin

lib/ssl/src/ssl.erl
```diff
@@ -402,6 +403,7 @@
 -type client_srp_identity()             :: srp_identity().
 -type customize_hostname_check() :: list().
 -type sni()                      :: HostName :: hostname() | disable. 
+-type max_fragment_length()      :: undefined | 512 | 1024 | 2048 | 4096.
```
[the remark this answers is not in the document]
peterdmv: I'll discuss this with @IngelaAndin.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1321,13 +1336,28 @@ handle_server_hello_extensions(RecordCB, Random, CipherSuite, Compression,
...
+            #{current_write := #{max_fragment_length := ConnMaxFragLen}} = ConnectionStates,
+            ClientMaxFragEnum = max_frag_enum(ConnMaxFragLen),
+
+            if ServerMaxFragEnum == ClientMaxFragEnum ->
+                    ok;
+               true ->
+                    throw(?ALERT_REC(?FATAL, ?ILLEGAL_PARAMETER))
+            end;
+       true ->
+            ok
```
peterdmv, line 1358: Maybe we should add a comment here that the `max_fragment_length `extension is ignored if this is a session resumption.

>  The negotiated length applies for the duration of the session
   including session resumptions.

lib/ssl/src/ssl_record.erl
```diff
@@ -202,6 +203,33 @@ set_renegotiation_flag(Flag, #{current_read := CurrentRead0,
...
+set_max_fragment_length(#max_frag_enum{enum = MaxFragEnum},
+                        #{current_read := CurrentRead0,
+                          current_write := CurrentWrite0,
+                          pending_read := PendingRead0,
+                          pending_write := PendingWrite0}
+                        = ConnectionStates) ->
+    MaxFragmentLength = if MaxFragEnum == 1 -> ?MAX_FRAGMENT_LENGTH_BYTES_1;
+                           MaxFragEnum == 2 -> ?MAX_FRAGMENT_LENGTH_BYTES_2;
+                           MaxFragEnum == 3 -> ?MAX_FRAGMENT_LENGTH_BYTES_3;
+                           MaxFragEnum == 4 -> ?MAX_FRAGMENT_LENGTH_BYTES_4
```
peterdmv, line 220: I'm just thinking if we should add a guard expression `true` here to properly handle a fault even at this level.

lib/ssl/src/dtls_connection.erl
```diff
@@ -448,9 +448,12 @@ init({call, From}, {start, Timeout},
     Hello = dtls_handshake:client_hello(Host, Port, ConnectionStates0, SslOpts,
 					Session#session.session_id, Renegotiation, Cert),
 
+    MaxFragEnum = maps:get(max_frag_enum, Hello#client_hello.extensions, undefined),
+    ConnectionStates1 = ssl_record:set_max_fragment_length(MaxFragEnum, ConnectionStates0),
```
peterdmv, line 452: Can it be a problem if we allow the fragmentation of a ClientHello in DTLS? @IngelaAndin

lib/ssl/src/tls_connection.erl
```diff
@@ -604,8 +604,11 @@ init({call, From}, {start, Timeout},
     %% Update pre_shared_key extension with binders (TLS 1.3)
     Hello1 = tls_handshake_1_3:maybe_add_binders(Hello, TicketData, HelloVersion),
 
+    MaxFragEnum = maps:get(max_frag_enum, Hello1#client_hello.extensions, undefined),
+    ConnectionStates1 = ssl_record:set_max_fragment_length(MaxFragEnum, ConnectionStates0),
+
```
peterdmv, line 609: Can it be a problem if ClientHello is fragmented? @IngelaAndin

dotsimon: No. Hello fragmentation is required by TLS, as noted in RFC 6066:

> Note that TLS already requires clients and servers to support fragmentation of handshake messages.

IngelaAndin: @dotsimon  is correct all handshake messages can be fragmented regardless of the max_fragment_length extension.

lib/ssl/src/dtls_connection.erl
IngelaAndin: It should already be allowed, even if it may not happen!

lib/ssl/src/dtls_connection.erl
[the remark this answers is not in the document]
IngelaAndin: So I have been thinking about this, re reading the DTLS RFC. As I remember the hard coded 1400 was chosen to be a little smaller than than expected "common case"  MTU (of course the tm correct thing to do is to have a discovery mechanism that we have not had time to implement so far, at the leas we should make it configurable), so that the encrypted DTLS record holding a handshake fragment normally should fit into one datagram, as this is the requirement of the RFC.  So in my mind it would it not be the min(1400, MaxFragmentLen) that we want?!

lib/ssl/src/tls_connection.erl
```diff
@@ -1164,6 +1168,8 @@ next_tls_record(Data, StateName,
             _ ->
                 State0#state.connection_env#connection_env.negotiated_version
         end,
+    #{current_write := #{max_fragment_length := MaxFragLen}} = State0#state.connection_states,
+    SslOpts = SslOpts0#{max_fragment_length => MaxFragLen},
```
peterdmv, line 1172: I wonder if we can avoid setting `max_fragment_length` in each call to `next_tls_record/3`? @IngelaAndin

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -902,6 +921,9 @@ do_wait_ee(#encrypted_extensions{extensions = Extensions}, State0) ->
...
+        Maybe(maybe_max_fragment_length(Extensions, State0)),
+
```
peterdmv, line 926: This a comment to follow up if the presence of `max_fragment_length ` in `EncryptedExtensions` affects TLS 1.3 session resumption interoperability with openssl. @IngelaAndin

lib/ssl/src/tls_record.erl
peterdmv: I looked it up, this code block appears 5 times. It would be better if they were factored into a function.

```shell
tls_record_1_3.erl
48:    MaxLength = if is_integer(MaxFragmentLength) ->
49-                        MaxFragmentLength;
50-                   true ->
51-                        %% TODO: Consider padding here
52-                        ?MAX_PLAIN_TEXT_LENGTH
--
80:    MaxLength = if is_integer(MaxFragmentLength) ->
81-                        MaxFragmentLength;
82-                   true ->
83-                        ?MAX_PLAIN_TEXT_LENGTH
84-                end,

tls_record.erl
119:    MaxLength = if is_integer(MaxFragmentLength) ->
120-                        MaxFragmentLength;
121-                   true ->
122-                        ?MAX_PLAIN_TEXT_LENGTH
123-                end,
--
168:    MaxLength = if is_integer(MaxFragmentLength) ->
169-                        MaxFragmentLength;
170-                   true ->
171-                        ?MAX_PLAIN_TEXT_LENGTH
172-                end,

dtls_record.erl
220:    MaxLength = if is_integer(MaxFragmentLength) ->
221-                        MaxFragmentLength;
222-                   true ->
223-                        ?MAX_PLAIN_TEXT_LENGTH
224-                end,
```

lib/ssl/test/ssl_mfl_SUITE.erl
```diff
@@ -0,0 +1,530 @@
...
+    PostF = fun(Server, Client) ->
+                    receive {Server, {ext, ServerExt}} ->
+                            ct:log("Server handshake Ext ~p~n", [ServerExt]),
+                            MflEnum = maps:get(max_frag_enum, ServerExt, undefined)
+                    end,
+                    receive {Client, {ext, ClientExt}} ->
+                            ct:log("Client handshake Ext ~p~n", [ClientExt]),
+                            case maps:get(server_hello_selected_version, ClientExt, undefined) of
+                                {3,4} ->
+                                    %% For TLS 1.3 the ssl {handshake, hello} API is inconsistent:
```
peterdmv, line 281: This is a todo item for us. @IngelaAndin

lib/ssl/test/ssl_mfl_SUITE.erl
```diff
@@ -0,0 +1,530 @@
...
+            4096
+    end;
+get_openssl_max_fragment_length_ext(_Acc, Port, Acc2) ->
+    get_openssl_max_fragment_length_line(Port, Acc2).
+
+
+get_line(Data) ->
+    get_line(Data, []).
+
+get_line([$\r|T], A) ->
```
peterdmv, line 497: We will see in the nightly tests if we should handle old macOS line brakes, too.

lib/ssl/src/dtls_connection.erl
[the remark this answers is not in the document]
peterdmv: Then I think we can resolve this item, too, as it correctly sets it to 1400.

peterdmv:
2/2 review done. We will further discuss the open items and we will come back to you!

lib/ssl/src/tls_connection.erl
IngelaAndin: It seems this is used to propagate the value of the negotiated max_fragment_length tls_record:validate_tls_record_length/7 without adding an extra argument to all relevant functions. I think it makes the code confusing!

lib/ssl/test/ssl_mfl_SUITE.erl
IngelaAndin: Yes, the API was created before TLS-1.3 and hence TLS-1.3 differences where not considered. We are aware that considerations for TLS-1.3 are needed, but have not prioritized that yet.

peterdmv:
We have found a failing testcase in one of our nightly test runs (solaris 2.11, x86):
`ssl_session_SUITE:session_table_stable_size_on_tcp_close` failed due to a `function_clause `error in `tls_record:encode_handshake`. It is easily reproducable made a [gist](https://gist.github.com/peterdmv/4c2861c8d4987a7df9dd3e0bf3af5fb8) for it.

IngelaAndin:
When handling fragmentation of DTLS handshake messages reliability is more important than accurate fragment length. And trying to avoid IP fragmentation is important. Now our DTLS implementation needs some more work to take all such aspects in to account. However we believe that the code in
dtls_connection for send_handshake_flight should be:

```
PMTUEstimate = 1400, %% TODO make configurable
{HsBefore, ConnectionStates1} =
	encode_handshake_flight(lists:reverse(Flight0), Version, min(MaxFragmentSize, PMTUEstimate), Epoch, ConnectionStates0),
```

IngelaAndin:
So the review comments where quite many, and I am unsure if we communicated the atom vs integer issue. So anyway we decided that an integer will be fine in this particular case. Normally enumerates are represented as atoms. But as this enumerate is quite special and its replacement will be an integer. We will go for that.

dotsimon:
Ok I will fix the next_tls_record thing sometime this week.

Regarding the failed test case, did you have the correct version of ssl_session_SUITE.erl installed under the test_server tree?
Your gist fails because it does not include max_fragment_length => undefined, but this is present in the last line in that file on this branch. I find it strange that the test cases create this map since it is internal. Why not use (d)tls_record:init_connection_states to get the initial map and then modify the parts actually used by any particular test case?

Finally, I had also suggested min(MaxFragmentSize, 1400) as a comment to the review from @peterdmv so I will make the change as you proposed. Just a note for you though, if/when you implement any MTU estimation it should **not** be used when SCTP is the transport layer (RFC 6083 s3.2)

peterdmv:
Regarding the failing testcase, it's not failing anymore, it was some strange glitch in the test framework. Good comment on ssl_session_SUITE!

IngelaAndin:
Just to be clear OTP-23 code stop is coming up.  So if you managed to finish this week it can probably make it for OTP-23, otherwise it will  OTP-23.1.  I does not really matter to us, just so you know.

dotsimon:
Thanks for the info about the code freeze, hopefully this now resolves all the outstanding items.

peterdmv:
Thank you for your contribution! Great work!

dotsimon:
But not for OTP 23? When do you think this will get into an actual release?

IngelaAndin:
It is part of the OTP-23 release, it did make it!

PR #2548, opened 2020-02-23
erts: Polymorphic specs for hd/1, tl/1, min/2 and max/2

zuiderkwast:
Relates the result to the domain by using the same type variable on both sides of the arrow.


kostis:
Please see the comment on $2549.

In Erlang, there is _nothing_ that forces the elements of a list to have the same type.
The call hd([42,gazonk,<<42>>]) happily returns (i.e. succeeds with) 42 -- and from that you cannot deduce that the rest of the list contains integers.

uabboli:
See https://github.com/erlang/otp/pull/2549.

PR #2549, opened 2020-02-23
stdlib: Improved specs for lists:{fold,mapfold}{l,r}/3

zuiderkwast:
By using the same type variables for Acc0, Acc1, AccIn and AccOut, tools can check that these are of the same type.

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -1252,10 +1252,11 @@ flatmap(F, []) when is_function(F, 1) -> [].
 
 -spec foldl(Fun, Acc0, List) -> Acc1 when
       Fun :: fun((Elem :: T, AccIn) -> AccOut),
-      Acc0 :: term(),
-      Acc1 :: term(),
-      AccIn :: term(),
-      AccOut :: term(),
+      Acc0 :: Acc,
+      Acc1 :: Acc,
+      AccIn :: Acc,
+      AccOut :: Acc,
+      Acc :: term(),
       List :: [T],
       T :: term().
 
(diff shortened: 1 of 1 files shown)


zuiderkwast:
I don't really see the point of the constraints on the form `T :: term()`, but I have followed the pattern of existing code for consistency. IMO, a more simple (and equivalent, hopefully) spec would be as follows:

```Erlang
-spec foldl(fun((T, Acc) -> Acc), Acc, [T]) -> Acc.
```

kostis:
This change is _wrong_!

zuiderkwast:
> This change is *wrong*

I assume because it implies that Acc1 and Acc0 has to be the same value, right?

Is there a way to relate the Acc types without this implication, e.g. `AccIn :: Acc0, AccIn :: AccOut, AccOut :: Acc1`? Or something like that....

kostis:
Even that is wrong.  Consider
<pre>
1> lists:foldl(fun (T, Acc) when is_atom(T), is_map(Acc) -> 42 end, gazonk, []).
gazonk
</pre>

gazonk is not a map, and the list does not contain atoms...

zuiderkwast:
...  with `Acc1 :: AccOut | Acc0` then, and perhaps some more changes....?

Obviously I haven't thought this through enough, but I'd like to believe there is a way to relate the Acc types, at least some of them, a little bit. Not possible at all?

elbrujohalcon:
@kostis … in your example, wouldn't it be fair to say that `T :: term()` and `Acc :: gazonk | 42`?
Or relaxing the spec a bit, like @zuiderkwast proposes…

```erlang
-spec foldl(Fun, InitValue, List) -> Result when
      Fun :: fun((Elem :: T, AccIn) -> AccOut),
      AccIn :: Acc,
      Result :: Acc,
      Acc :: InitValue | AccOut,
      AccOut :: term(),
      InitValue :: term(),
      Acc :: term(),
      List :: [T],
      T :: term().
```

(I changed the names a bit because I don't like the `Xxx0`, `Xxx1` names for variables)

uabboli:
I'll reject the PR since Kostis pointed out that it is wrong. Other
tools like Gradualizer seem to interpret type variables differently, but for
the foreseeable future we will stick to Dialyzer's interpretation.

PR #2551, opened 2020-02-25
Allow sys_core_prepare to be skipped

josevalim:
Some tools may store v3_core and feed into the compiler,
in those cases, there is no need to prepare it.

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -808,7 +808,7 @@ core_passes(CoreStatus) ->
     case CoreStatus of
         non_verified_core ->
             [?pass(core_lint_module),
-             {pass,sys_core_prepare},
+             {unless,no_core_prepare,{pass,sys_core_prepare}},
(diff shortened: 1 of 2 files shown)


jhogberg:
Thanks for the PR! Can you add tests and documentation?

josevalim:
@jhogberg surely! Do you have any suggestions on how to test that a pass is not invoked? :)

jhogberg:
I'd export and call `compile:passes/2` or compile a file with the `time` option and see if it's listed, neither is particularly pretty but they should suffice.

bjorng:
> Do you have any suggestions on how to test that a pass is not invoked? :)

You could use OTP 22 to produce Core Erlang for a small module that uses `receive` (and hard-code the result into the test case). If you skip the prepare pass when attempting to compile such code, the compiler will crash in a later pass.

josevalim:
Test added, I went with @bjorng's suggestion. Thanks!

jhogberg:

PR #2298, opened 2019-06-21
Add SSH agent support

pmeinhardt:
Hey 👋 This is a take on adding `ssh-agent` support to OTP.

_Note: This is an early draft intended to collect feedback and iterate on the solution. ✌️_

_Note 2: There are a few helper files (`scripts/*` and `Dockerfile`) which I intend to remove but kept around for now, because they make development a bit more convenient._

**Overview of changes**

1. Add an `ssh_agent` module handling communication with `ssh-agent` processes
    - determines the agent socket from the process environment
    - handles encoding and decoding of agent messages into records
2. Add `ssh_file_with_agent` module (loosely) implementing the [`ssh_client_key_api`](http://erlang.org/doc/man/ssh_client_key_api.html) `behaviour`
    - handles calls to the `user_key/2` callback by fetching a list of keys from an agent
    - delegates calls to the host key callbacks to the default [`ssh_file`](http://erlang.org/doc/man/ssh_file.html) module (agents do not handle host keys)
3. Update the `ssh_auth` module to handle keys returned from an agent
    - keys returned from an agent never include the private key part
    - delegate "signing" to the agent for agent keys to prove the client identity

**Test-drive**

If you would like to test-drive these changes, here is an escript I used to try them out:
https://gist.github.com/pmeinhardt/a39494d43f6acf606fc7249a3fe2cb50

This also illustrates how these changes would be used in code.

**Discussion**

The current approach taken tries to fit into the existing code structure and APIs without breaking backwards-compatibility. However, there are some limitations because of this which may be worth addressing in the future (not part of this PR).

In particular, while working on this I noticed that the `ssh_client_key_cb` behavior:

1. makes the assumption that there is only ever one key for a given algorithm (`user_key/2` is expected to only ever return one key).

    This means that even though you can add multiple RSA keys to an agent for instance (or you may have multiple key files for this algorithm), the OTP `ssh` application will only try one.

2. does not include "signing" to verify the client identity.

    This makes it rather odd to integrate the agent behavior, as the agent-specific keys and handling of data signing "bleeds" into the `ssh_auth` module.

In the long run ("future work"), it might therefore be desirable to potentially replace/extend the `ssh_client_key_api` to address these issues.

**References**

The relevant RFCs and specific sections are also referenced in the code, but I will also list them here for easy access:

- https://tools.ietf.org/html/draft-miller-ssh-agent-02
- https://tools.ietf.org/html/rfc4251
- https://tools.ietf.org/html/rfc4252
[...]

PR #2554, opened 2020-02-27
Add SSH agent support

andreasknoepfle:
Hey 👋 
Last year our college @pmeinhardt opened this PR https://github.com/erlang/otp/pull/2298 on adding SSH agent support to OTP. There were still some missing tests and documentation and was it hanging in there for quite a bit, so me and @maltoe we picked it up this week since we had some spare time.

Changes since:
- we merged `ssh_file_with_agent` and `ssh_agent`, since we thought it may be nicer as one module
- added some documentation
- added tests, including a `ssh_agent_mock_server` implementation to simulate a ssh agent in the tests and tests the `ssh_agent` module in integration
- implemented @HansN's suggestion https://github.com/erlang/otp/pull/2298#pullrequestreview-252912629 to have a more explicit return value for `user_key`  

Here is a preview of the generated documentation:
Screenshot 2020-02-27 at 15 48 17


lib/ssh/src/ssh_agent.erl
```diff
@@ -0,0 +1,134 @@
...
+-type socket_path_option() :: {socket_path,  string()}.
+-type timeout_option() :: {timeout, integer()}.
+
+%% ssh_client_key_api implementation
+
+add_host_key(Host, Key, Opts) ->
+    ssh_file:add_host_key(Host, Key, Opts).
+
+is_host_key(Key, PeerName, Algorithm, Opts) ->
+    ssh_file:is_host_key(Key, PeerName, Algorithm, Opts).
```
HansN: Just a thought: if someone makes a special ssh_file by some reason, they probably want that to be called from ssh_agent instead.
So an option {call_ssh_file, SshFile} could be useful -- with a better name maybe:

```suggestion
add_host_key(Host, Key, Opts) ->
    KeyCbOpts = proplists:get_value(key_cb_private, Opts, []),
    SshFileCb = proplists:get_value(call_ssh_file, KeyCbOpts, ssh_file),
    SshFileCb:add_host_key(Host, Key, Opts).

is_host_key(Key, PeerName, Algorithm, Opts) ->
    KeyCbOpts = proplists:get_value(key_cb_private, Opts, []),
    SshFileCb = proplists:get_value(call_ssh_file, KeyCbOpts, ssh_file),
    SshFileCb:is_host_key(Key, PeerName, Algorithm, Opts).
```

lib/ssh/src/ssh_auth.erl
```diff
@@ -140,47 +141,67 @@ keyboard_interactive_msg([#ssh{user = User,
...
+
+            Sig = case Key of
+                {ssh2_pubkey, PubKeyBlob} ->
+                  {ssh_agent, KeyCbOpts} = ?GET_OPT(key_cb, Opts),
+                  % OpenSSH does not seem to care when these flags are set for
+                  % signature algorithms other than RSA, so we always send them.
+                  SignFlags = ?SSH_AGENT_RSA_SHA2_256 bor ?SSH_AGENT_RSA_SHA2_512,
+                  SignRequest = #ssh_agent_sign_request{key_blob = PubKeyBlob, data = SigData, flags = SignFlags},
+                  SignResponse = ssh_agent:send(SignRequest, KeyCbOpts),
+                  #ssh_agent_sign_response{signature = #ssh_agent_signature{blob = Blob}} = SignResponse,
```
HansN: I don't like the idea that ssh_auth has knowledge about agents and also is bound to a certain name of the callback file.
What about a new optional callback function in the ssh_client_key_api so the call would be roughly:
```suggestion
                  Blob = ssh_transport:call_KeyCb(sign, [SigData, ....], Opts);
```

lib/ssh/test/ssh_agent_SUITE.erl
```diff
@@ -0,0 +1,130 @@
...
+
+suite() ->
+    [{timetrap, {seconds, 30}}].
+
+all() ->
+    [request_identities, sign_request, connect_with_ssh_agent].
+
+init_per_suite(Config) ->
+    ok = ssh:start(),
+    ssh_agent_mock_server:check_mktemp(Config).
```
HansN: We run the tests on 25+ machines with varying architecture, OSes and cryptolib variants.
Some does not even have a cryptolib, so our ssh (and ssl/tls) test suites need to test for crypto availability. If not available, the test suite is skipped. So I suggest:
```suggestion
    ?CHECK_CRYPTO(
       begin
           ok = ssh:start(),
           ssh_agent_mock_server:check_mktemp(Config)
       end
      ).
```
The ?CHECK_CRYPTO will do the begin ... end if crypto is available and skip the suite if it is not available.

lib/ssh/test/ssh_agent_mock_server.erl
```diff
@@ -0,0 +1,151 @@
...
+
+-record(state, {socket, priv_key, pub_key, socket_path}).
+
+-define(SIG_ALG, 'ssh-rsa').
+
+start_link(PrivKeyDir) ->
+    {ok, PrivKey} =ssh_file:user_key(?SIG_ALG, [{user_dir, PrivKeyDir}]),
+
+    %% We cannot use priv_dir because unix socket paths are limited to 108 bytes.
+    SocketPath = string:chomp(os:cmd("mktemp -u")),
```
HansN: Does this work on Windows?

lib/ssh/src/ssh_auth.erl
maltoe: 💯 👍  We actually debated a solution pretty exactly like what you suggest. Sounds perfect.

Would you move the current signing logic (i.e. the second branch in this `case`) to a `ssh_file:sign` or should it stay where it is?

lib/ssh/test/ssh_agent_mock_server.erl
maltoe: No, I don't  think windows has a `mktemp` utility. But probably also not a likely OS to have an (openssh) ssh agent, so we'll skip it.

Otherwise the `mktemp` approach is ok? We checked that temporary files are cleaned up after test runs.

lib/ssh/src/ssh_auth.erl
HansN: Good question - I have to think about it.  Keep it in ssh_auth.erl, I can move it later.

lib/ssh/test/ssh_agent_mock_server.erl
HansN: I see.  Forget my comment.

andreasknoepfle: We already check if `mktemp` is in the path and skip if it is not in there. I guess this should be fine for `windows` or do you think we still need an additional os:type check @HansN?

andreasknoepfle:
Thanks for the fast review @HansN. Updated the PR to your suggestions :)

HansN:
Hi @andreasknoepfle @pmeinhardt @maltoe
The nightly tests seem good, so I'll merge this.  But, there are a lot of commits, so either I or you squash them to a reasonable amount before the merge.  Do you want to do it?

HansN:
Thanks for the PR!

PR #2555, opened 2020-02-28
Improve compilation speed

jhogberg:
Certain modules were much slower to compile in `OTP 23-rc1` than in `OTP 22`. This PR is an attempt to remedy this.

https://bugs.erlang.org/projects/ERL/issues/ERL-1181


jhogberg:
Merged as e321fd9ddfcb86e2e74620ca96a84394a96f765e

PR #2556, opened 2020-02-29
Added -delayed-detach to 'erl' (unix only).

mbj4668:
This flag works like -detached, except it delayes the detaching until
the boot script has been successfully executed.  This makes it
possible for user code to validate config etc before detaching.

The function init:boot_msg/1 can be used to print messages to stderr
during boot time.

[This function was developed at tail-f 2006, and was contributed as
part of a big chunk of code around that time.  It has been used in
production since then.  This PR is a cleaner version with just the
delayed detach functionality.]


bjorng:
We don't accept pull requests that implement a feature only on some platforms (when it would have been technically possible to implement on all platforms). Feel free to commit a new PR when you (or someone else) has implemented it for Windows.

mbj4668:
The new flag works exactly like -detached, i.e., only present on unix.
Would it be accepted if it was done as an optional argument to -detached instead?

bjorng:
> The new flag works exactly like -detached, i.e., only present on unix.

`-detached` **is** supported on Windows.

mbj4668:
Ok.  Never mind then.  At least we tried!


/martin

Den tis 10 mars 2020 kl 07:24 skrev Björn Gustavsson <
notifications@github.com>:

PR #2557, opened 2020-03-01
pg: handle race condition between scope 'DOWN' and processes leaving

max-au:
Handles following sequence:
 - nodes A and B share scope S, and were just disconnected
 - some process on node B leaves scope S at the same time
 - node B processes leave request and (while 'nodedown' from A is in the
   mailbox) attempts to send 'leave' broadcast to A
 - node A already received and processes 'nodedown' from B, and removed
   all processes from scope S
 - node B establishes connection to A, and 'leave' message gets sent
 - scope S on node A crashes, because it does not expect messages from
   B (it is assumed to be down, and not yet participating in the overlay
   network)

It also guards against other cases when nodes outside of overlay network
are trying to make changes.

### DIFF
--- a/lib/kernel/src/pg.erl
+++ b/lib/kernel/src/pg.erl
@@ -264,30 +264,48 @@ handle_cast(_, _State) ->
 
 %% remote pid or several pids joining the group
 handle_info({join, Peer, Group, PidOrPids}, #state{scope = Scope, nodes = Nodes} = State) ->
-    join_remote(Scope, Group, PidOrPids),
-    % store remote group => pids map for fast sync operation
-    {MRef, RemoteGroups} = maps:get(Peer, Nodes),
-    NewRemoteGroups = join_remote_map(Group, PidOrPids, RemoteGroups),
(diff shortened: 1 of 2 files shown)


lib/kernel/src/pg.erl
```diff
@@ -264,30 +264,49 @@ handle_cast(_, _State) ->
...
-    join_remote(Scope, Group, PidOrPids),
-    % store remote group => pids map for fast sync operation
-    {MRef, RemoteGroups} = maps:get(Peer, Nodes),
-    NewRemoteGroups = join_remote_map(Group, PidOrPids, RemoteGroups),
-    {noreply, State#state{nodes = Nodes#{Peer => {MRef, NewRemoteGroups}}}};
+    case maps:get(Peer, Nodes, []) of
+        {MRef, RemoteGroups} ->
+            join_remote(Scope, Group, PidOrPids),
+            %% store remote group => pids map for fast sync operation
+            {MRef, RemoteGroups} = maps:get(Peer, Nodes),
```
rickard-green: This `maps:get(Peer, Nodes)` can be removed

PR #2561, opened 2020-03-09
Fix ERL-1189

montag451:


IngelaAndin:
Thank you for the PR. We would like you to also provide a test case and sign the CLA.

montag451:
Hi @IngelaAndin,

As requested I have added a test case. The test tries to use a specialized decode function and fails if the specialized function does not exist because of a failure to generate it. It's a bit ugly but I didn't find a better way to test it as the compilation errors generated during the creation of specialized decode functions are catched by `asn1ct:generate/4`.

Regarding the CLA I have already signed it if I'm not mistaken.

IngelaAndin:
CLA looks good now. Thank you for making a test. I have included your branch for testing and requested a review.

IngelaAndin:
Tests are running fine just waiting for the code reviews.

PR #2565, opened 2020-03-15
add file:del_dir/2

mikpe:
This adds file:del_dir/2, a variant of file:del_dir/1 taking a list of options to extend its behaviour.  Available options are:
- recursive: recursively delete the contents of a directory before deleting the directory itself
- force: ignore errors when accessing or deleting files or directories
- keeptop: the top-most directory is not deleted

file:del_dir(Dir, [recursive, force]) is equivalent to "rm -rf Dir" in a Unix shell.
file:del_dir(Dir, [recursive, force, keeptop]) is similar to "rm -rf Dir/\*" in a Unix shell, but will also remove "."-files not expanded by "\*".
file:del_dir(Dir, []) is equivalent to file:del_dir(Dir).

Symbolic links are not followed.

In recursive and non-force mode, if there is an error accessing or deleting a file, the error return specifies both the error and which specific file it applies to.

Fixes ERL-1164.

### DIFF
--- a/lib/kernel/doc/src/file.xml
+++ b/lib/kernel/doc/src/file.xml
@@ -402,6 +402,28 @@ f.txt:  {person, "kalle", 25}.
       </desc>
     </func>
     <func>
+      <name name="del_dir_r" arity="1" since="OTP 23.0"/>
+      <fsummary>Delete a file or directory.</fsummary>
+      <desc>
+        <p>Deletes file or directory <c><anno>File</anno></c>.
+          If <c><anno>File</anno></c> is a directory, its contents
+          is first recursively deleted.  Returns:</p>
+        <taglist>
+          <tag><c>ok</c></tag>
+          <item>
+            <p>The operation completed without errors.</p>
+          </item>
+          <tag><c>{error, posix()}</c></tag>
+          <item>
+            <p>An error occurred when accessing or deleting <c><anno>File</anno></c>.
+             If some file or directory under <c><anno>File</anno></c> could not be
+             deleted, <c><anno>File</anno></c> cannot be deleted as it is non-empty,
+             and <c>{error, eexist}</c> is returned.</p>
+          </item>
+        </taglist>
+      </desc>
+    </func>
+    <func>
       <name name="delete" arity="1" since=""/>
       <fsummary>Delete a file.</fsummary>
       <desc>
--- a/lib/kernel/src/file.erl
+++ b/lib/kernel/src/file.erl
@@ -28,7 +28,7 @@
 -export([format_error/1]).
 %% File system and metadata.
 -export([get_cwd/0, get_cwd/1, set_cwd/1, delete/1, rename/2,
-	 make_dir/1, del_dir/1, list_dir/1, list_dir_all/1,
+	 make_dir/1, del_dir/1, del_dir_r/1, list_dir/1, list_dir_all/1,
 	 read_file_info/1, read_file_info/2,
 	 write_file_info/2, write_file_info/3,
 	 altname/1,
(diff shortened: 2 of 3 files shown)


lib/kernel/doc/src/file.xml
```diff
@@ -401,6 +401,52 @@ f.txt:  {person, "kalle", 25}.
...
+          </item>
+          <tag><c>{error, posix()}</c></tag>
+          <item>
+            <p>An error occurred when accessing or deleting the
+             top-level directory, in non-recursive mode.</p>
+          </item>
+          <tag><c>{error, {posix(), File}}</c></tag>
+          <item>
+            <p>An error occurred when accessing or deleting a file or
+             directory, in recursive mode.</p>
```
josevalim: Should it clarify that, if an error happens in the middle of recursive deletion, the directory will be in a state where part of files were deleted and the other part was not?

lib/kernel/doc/src/file.xml
```diff
@@ -401,6 +401,52 @@ f.txt:  {person, "kalle", 25}.
...
+             before the directory itself is.</p>
+          </item>
+          <tag><c>force</c></tag>
+          <item>
+            <p>Ignores errors when accessing or deleting a file or
+             directory.</p>
+          </item>
+          <tag><c>keeptop</c></tag>
+          <item>
+            <p>The top-level directory is not deleted.</p>
```
josevalim: Should we clarify that if this option is given and `recursive` isn't, then the operation is a no-op? :)

lib/kernel/doc/src/file.xml
mikpe: I would think that's obvious, but if OTP wants that clarification I'll do it.

lib/kernel/doc/src/file.xml
mikpe: Why deny people of that easter-egg? :-)  But yes, I could clarify that if needed.

mikpe:
Let me just add that clearly del_dir/2 is intended to only be used together with recursive, so a possible simplification could be to rename it to del_dir_r/2, which implies recursive but optionally allows to specify force and/or keeptop.  I really don't want to have to specify what happens in non-recursive mode.

lib/kernel/doc/src/file.xml
jhogberg: It's obvious that the file that couldn't be deleted will be left alone, but I'm not sure it's as obvious that it exits early on the first error.

If you run `rm -r` on a folder that has a bunch of files that the user has permission to delete and some files they lack permission for, then only the latter will be kept. When using `file:del_dir/2` it might exit early and keep some of the files that could've been deleted if not for the error.

I think both approaches are reasonable, but if we exit early I'd like the documentation to say that it may leave things half-finished.

jhogberg:
Thanks for the PR!

> Let me just add that clearly del_dir/2 is intended to only be used together with recursive, so a possible simplification could be to rename it to del_dir_r/2 ... snip ...

I'd prefer this approach.

lib/kernel/doc/src/file.xml
mikpe: I agree that bailing out early would be against users' expectations.  In that case the reasonable thing would be to imply 'force' in the recursive calls, and just return whatever del_dir/1 returns on the top-level directory; if it's non-empty we get eexist.  We will however not get any indication about which file(s) it failed to remove.  I'm not sure I'd want to collect a list of all failed files and their specific errors.  (The user could list the directory themselves if they really want to know.)

jhogberg: > and just return whatever del_dir/1 returns on the top-level directory; if it's non-empty we get eexist. We will however not get any indication about which file(s) it failed to remove

How would this interact with `keeptop`? Would there be files left in the directory without an indication of error?

mikpe: The short answer is that I've decided to remove the `keeptop` option.

I started removing the `recursive` option and implying `force` in recursive deletes, and it became clear that `force` too was entirely redundant, as it only affects whether a failure to delete the top-level component returns `ok` on an `error`-tuple.  And the `keeptop` option would require more code than I though it worth.  So in the end I reduced it to `del_dir_r/1` with no options and equivalent to `rm -rf`.

jhogberg:
The tests ran fine over the weekend, so I'll merge it once RC2  is released. Can you squash the PR down to one commit?

mikpe:
Squashed

jhogberg:

PR #2570, opened 2020-03-20
Use effective UID and GID in checks

hauleth:
In most situations effective and real IDs will match, but in theory
someone can create NIF that will change the effective IDs to other one.
In such cases the returned value could be different from the expected.


jhogberg:
Thanks for the PR! It looks good but I missed the merge window for RC2, so I'll merge it as soon as that's released.

jhogberg:
Merged, thanks for the PR!

PR #2571, opened 2020-03-21
Prevent left nav overlapping content during scroll

maxnordlund:
When viewed on a smaller screen the content can be scrolled not only up
and down, but also left and right. More specifically it can be scrolled
back over the navigation menu to the left.

This prevents that, and as an bonus it fixes the scrolling to the right
where there is no content.

---

Compare:

image

With:

image

---

I wasn't sure if this is to be merged to master or maint. I mean the bug is in master, but the file hasn't been touched since 2015.

I can easily rebase on maint and update the PR if needed.

### DIFF
--- a/lib/erl_docgen/priv/css/otp_doc.css
+++ b/lib/erl_docgen/priv/css/otp_doc.css
@@ -52,6 +52,7 @@ a:visited      { color: #1b6ec2; text-decoration: none }
 #content {
   margin-left: 340px; /* set left value to WidthOfFrameDiv */
   max-width: 42em;
+  overflow-x: hidden;
 }
 
 .frontpage 


sverker:
It turns out that this change hides all the "since" versions out in the right margin as they are actually beyond 100% width.
I think I will revert this commit in next OTP-23 patch to begin with and then we'll see how to get this to work with since versions intact. Contributions are welcome.

maxnordlund:
Do you have an example I could look at? Screenshots would also be useful.

maxnordlund:
I can see it now, the culprit is that `.td-since` is set to `width: 200%`, which puts it off screen.

With some more tweaks I can get it to look like this, and still not scroll out of bounds:

image

sverker:
You can look at the module rand for example.

OTP 22.3: http://erlang.org/documentation/doc-10.7/lib/stdlib-3.12/doc/html/rand.html
OTP 23.0: https://erlang.org/doc/man/rand.html

Scroll down to the functions. Every function should have a light grey "OTP 18.0" (or higher) out in the right margin.

In Firefox I can right click and choose "Inspect Element (Q)" to dynamically edit the html. Find the <div id="content"> and disable the attribute "overflow-x: hidden" and that will bring the light grey "since" versions back.

maxnordlund:
What do you think about the changes above in https://github.com/erlang/otp/pull/2571#issuecomment-631573704 ?

dgud:
Does that patch fix the row breaks in [https://erlang.org/doc/man/mnesia.html#delete-3](https://erlang.org/doc/man/mnesia.html#delete-3) as well?

That kind of line-breaks was introduced with this patch according to rumors :-)

maxnordlund:
Will check

maxnordlund:
I see no difference with or without overflow hidden. Maybe only on large screens? Could you take a screen shot please?

dgud:
Mnesia didn't have generated types in previous releases so not seen there, I noticed the bad brakes when I added a PR that added the types.

maxnordlund:
Hm, if you disable the overflow hidden using the web inspector, what happens?

dgud:
Hmm I don't know how I do that, my web knowledge is not where it should be.
Using chrome

sverker:
I'm not thrilled about the line break between "OTP" and the version number.

maxnordlund:
Right click somewhere on the page and open the web inspector, navigate the DOM/HTML tree to the container div and on the left you should be able to uncheck overflow hidden

maxnordlund:
Yeah, I tried to get ride of it. But in order to make it nice I would need to change the HTML. Tables are not fun to work with, since they have their own layout algorithm among other limitations.

maxnordlund:
@dgud I applied my changes to the mnesia page and there's no difference between the current version, the previous or the proposed changes.

(And I said the wrong div, it's `content`, not `container`)

dgud:
Setting the width to 230% solved it, couldn't find the 'overflow' yet.
It's dinner time, and long weekend will try to remember to check it later.

maxnordlund:
Ok, I'm going hiking and will be back Saturday. Otherwise we can resume this perhaps on Monday?

sverker:
The tables and the 200% width was a cheap way (with limited html knowledge) to add the "since" versions without disturbing the existing layout.

The html is generated from xml by lib/erl_docgen/priv/xsl/db_html.xsl

Resume on Monday sounds good.

PR #2573, opened 2020-03-23
docgen: Rework seealso to be aware of the type of the target

garazdawi:
This PR removes `<seealso>` from the erl_docgen dtd and replaces it with type aware tags instead. The tags are: `seemfa|seeerl|seetype|seeapp|seecom|seecref|seefile|seeguide`. This means that if a link is pointing to an MFA, it should use `seemfa`. If the link is pointing to a User's Guide, it should use `seeguide`. This is done in order to make it easier for third-party tools to work with the documentation. 

A validation script has been added that is called when you do `make xmllint` which checks that all link types are correct and that they exist.

The EEP-48 embedded documentation has been updated to use the new `see*` links.

Also, a new tag has been added to `<funcs>` called `<fsdescription>` which can be used to describe the content of a `<funcs>` block without it ending up as a part of the EEP-48 module documentation.


system/doc/general_info/deprecations_23.inc
```diff
@@ -2,11 +2,7 @@
       <title>Crypto Old API</title>
       <p>The <seeguide marker="crypto:new_api#the-old-api">Old API</seeguide> is now deprecated
       and has also been
-<<<<<<< HEAD
```
bjorng: Shouldn't this merge conflict be resolved in an earlier commit?

KennethL:
I don't exactly understand the use of the new <fsdescription> tag. I assume it is supposed to contain the sections that occur between and after <funcs>. In that case I just regard this as a temporary workaround since I think that everything about a modules documentation should have a mapping to the chunk format. If this is not the case it will not work to generate complete documentation for a module from the chunk in the future, via for example Exdoc or similar tools. I think the way to go is to forbid the use of more than one <funcs> and to have the module level descriptions with sections in one place only. Another approach would be to allow several funcs with fsdescriptions and let that structure be available in the chunks format as well.

garazdawi:
> I think the way to go is to forbid the use of more than one and to have the module level descriptions with sections in one place only. Another approach would be to allow several funcs with fsdescriptions and let that structure be available in the chunks format as well.

I tried to do this, but gave up after spending two days doing the rewrite and not being close to finishing. So this is the compromise that I came up with that could be done without spending a week or two rewriting the docs.

system/doc/general_info/deprecations_23.inc
garazdawi: done

PR #2575, opened 2020-03-26
erts: Fix external fun application of erlang:apply/3

jhogberg:
http://erlang.org/pipermail/erlang-questions/2020-March/099343.html

@kostis


kostis:
Thanks for the quick fix -- most likely it's complete.
If you merge it on master soon, I can test it on more examples on PropEr's test suite that fail.

kostis:
All tests of PropEr now pass.  Thanks again for the quick reaction!

PR #2584, opened 2020-04-03
Simplify erl_parse.yrl using operator precedence

michalmuskala:
This is one of the changes in the parser done during development of the flexible parser for the formatter.  It will probably benefit the mainline parser as well. I'm exploring possibilities of upstreaming more changes.

### DIFF
--- a/lib/stdlib/src/erl_parse.yrl
+++ b/lib/stdlib/src/erl_parse.yrl
@@ -26,12 +26,8 @@ form
 attribute attr_val
 function function_clauses function_clause
 clause_args clause_guard clause_body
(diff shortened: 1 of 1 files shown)


bjorng:
Added to our daily builds.

uabboli:
A little late, but...

It has hitherto been necessary to put parentheses around 'catch' under
certain circumstances. One example is mentioned in the docs,
http://erlang.org/doc/reference_manual/expressions.html#catch-and-throw:

  A = (catch 1+2).

With this pull request it seems the parentheses are no longer needed.

Can we remove the paragraph from the docs, or is there some other
example where the parentheses are needed? I've tried to find one, so
far without success.

PR #2585, opened 2020-04-03
Introduce strict_include_lib compiler option

michalmuskala:
Current behaviour of `include_lib` of first trying the same resolution
as `inlcude` is unintuitive and error prone. Most users expect the two
options to behave in different ways - `include` to include headers from
current application/project and `include_lib` to include public headers
from other applications.

Current behaviuor can easily result in erronous behaviour where a wrong
header is included in complex repositories with convoluted structure.

To combat this issue, a new epp and compiler option `strict_include_lib`
is introduced that separates behaviour of `include` and `include_lib`
into the one described above.

---

If this is generally accepted as a good idea, I will follow up with another commit that adds documentation for the new option.

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -1035,18 +1035,7 @@ parse_module(_Code, St0) ->
     end.
 
 do_parse_module(DefEncoding, #compile{ifile=File,options=Opts,dir=Dir}=St) ->
-    SourceName0 = proplists:get_value(source, Opts, File),
-    SourceName = case member(deterministic, Opts) of
-                     true -> filename:basename(SourceName0);
-                     false -> SourceName0
-                 end,
-    R = epp:parse_file(File,
-                       [{includes,[".",Dir|inc_paths(Opts)]},
-                        {source_name, SourceName},
-                        {macros,pre_defs(Opts)},
-                        {default_encoding,DefEncoding},
-                        extra]),
-    case R of
+    case epp:parse_file(File, epp_opts(File, Dir, DefEncoding, Opts)) of
 	{ok,Forms,Extra} ->
 	    Encoding = proplists:get_value(encoding, Extra),
 	    case find_invalid_unicode(Forms, File) of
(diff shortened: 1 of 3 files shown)


garazdawi:
Hello,

This has more to do with the build system than the code the include lives in. So would it make sense to add a compiler option called `strict_include_lib` that makes all `include_libs` behave the way you want? This could possibly be made the default if the benefit is considered big enough.

michalmuskala:
I'm not sure I understand your suggestion. The PR already adds a `strict_include_lib` compiler option.

garazdawi:
Sorry, my bad. I missunderstood what you were trying to do.

Yes, this approach seems like a good one.  Docs would be most welcome.

KennethL:
We have had other discussions around `include_lib` and the workings of `rebar3` quite recently, I want to recap what we discussed and concluded there before adding this. It it is a good idea, but the real, real problem with include_lib is the use of the code path of the system running the compiler. This use of the code path is not deterministic and not under good control, a change of current directory will for example have impact.

michalmuskala:
Yes, I fully agree with that. There are couple places in the compiler where it is "stateful" - `include_lib` resolved through code path is one, loading behaviour modules to check `-behaviour` definitions would be another.  The final one is parse transforms, but that obviously has to load modules and execute code one way or another.

max-au:
Would it make sense to deprecate -include_lib() searching code paths, and keep only (more) deterministic -include()?

essen:
> Would it make sense to deprecate -include_lib() searching code paths, and keep only (more) deterministic -include()?

How would you include `eunit.hrl` using only `-include()`?

max-au:
Good question. With no other changes, it would be
-include("eunit-2.4.1/include/eunit.hrl")
Which is not convenient given OTP application version has to be explicitly specified. This actually solves an issue with non-reproducible builds (depending on ordering in code path), but inconvenient.

jhogberg:
We plan to solve the root of this problem in OTP 25, so I'm closing this PR. See https://github.com/erlef/eef-build-and-packaging-wg/discussions/16

PR #2586, opened 2020-04-05
removed LDFLAGS

mrcdll:


rickard-green:
This change alone will mess up non-macosx builds.

This and further changes to make it work on other platforms as well has already been made and merged into master. This will also be released in the soon to come OTP 22.3.1 patch.

Also, when submitting a pull request in the future, you need to motivate why the change is needed.

mrcdll:
@rickard-green sorry I inadvertently open the pull request. I'll take the occasion to ask a question. Even if the fix will be shipped with 22.3.1 users will not able to install 21.3.8.3 on Catalina. Is that correct?

rickard-green:
Correct, OTP 21.3.8.3 won't build on Catalina. I don't think there will be an OTP 21 patch released for this, but I'm not sure. If we were to release an OTP 21 patch, it would be based on the latest OTP 21 version (currently OTP 21.3.8.14) though.

I suggest that you cherry pick the commits made for fixing macosx builds once OTP 23.3.1 is out. There might come more fixes. I'm not sure if the just reported https://bugs.erlang.org/browse/ERL-1217 needs more fixes or not.

PR #2593, opened 2020-04-09
Fix misspelling in ssl test cases

dotsimon:
Fix misspelling in calls to ssl_test_lib:check_sane_openssl_renegotiate/2


dotsimon:
re-pushed the same commit to fix the CLA oops

IngelaAndin:
Good catch! If you rebase to latest master I have found the alpn typo but missed the npn one.

PR #2594, opened 2020-04-14
Fix slow lists:mapfoldl/3

bjorng:
This PR fixes two issues in the compiler that made `lists:mapfoldl/3` slow. See the commit messages for further details.

https://bugs.erlang.org/browse/ERL-1216

### DIFF
--- a/lib/compiler/src/beam_ssa_opt.erl
+++ b/lib/compiler/src/beam_ssa_opt.erl
@@ -40,7 +40,8 @@
 -include("beam_ssa_opt.hrl").
 
 -import(lists, [all/2,append/1,duplicate/2,flatten/1,foldl/3,
-                keyfind/3,last/1,member/2,reverse/1,reverse/2,
+                keyfind/3,last/1,mapfoldl/3,member/2,
+                partition/2,reverse/1,reverse/2,
                 splitwith/2,sort/1,takewhile/2,unzip/1]).
 
 -define(MAX_REPETITIONS, 16).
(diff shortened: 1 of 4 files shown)


fenollp:
Very interesting!
So to recap fast-growing O(n) space can be saved thanks to a reordering of calls:
```diff
 mapfoldl(F, Acc0, [Hd|Tail]) ->
     Res1 = F(Hd, Acc0),
+    R = element(1, Res1),
     Res2 = mapfoldl(F, element(2, Res1), Tail),
-    {[element(1, Res1)|element(1, Res2)],element(2, Res2)};
+    {[R|element(1, Res2)],element(2, Res2)};
 mapfoldl(F, Acc, []) ->
     {[],Acc}.
```
How can we make the compiler attempt such reorderings more often?

Here the call `element(1, Res1)` can be moved to happen before `element(2, Res1)` as the former cannot fail if the latter didn't (`Res1` is thus proven to be a tuple of size >= 2).

Could we instruct the compiler on the semantics of some of the basic functions such as element/2, map_key/2, ... (maybe only guards even) so these reorderings can be attempted?

Maybe there's even a heuristic similar to the number of stack slots mentioned here that can be used to decide when to reorder or better: which ordering to pick.
I'll have a try (albeit probably overfitted to the case at hand):
*Pick the reodering that minimizes the amount of "basic function" calls happening after non-"basic function" calls.*

I guess these "basic functions" can be any function the compiler knows is pure. Guards are a good subset of these here for memory saving as their output is often smaller than their input, trading off a small amount of computation (except for `length/1`).

Should I open an issue on https://bugs.erlang.org/?

bjorng:
@fenollp I only have time for a quick answer.

> So to recap fast-growing O(n) space can be saved thanks to a reordering of calls:

No, that is not what my PR does. The compiler usually sinks (executing later) tuple extraction instructions. When compiling `mapfoldl/3`, that results in `slow_mapfoldl/3`. In this particular case, this is a pessimization because `Res1` will be kept alive too long. So my PR uses a heuristic to disable this particular optimization of sinking tuple extraction instructions when it would be potentially harmful.

> Should I open an issue on https://bugs.erlang.org/?

No. I don't think that your suggested transformation (hoisting tuple extraction instructions) would be generally beneficial. Especially in OTP 22 and 23.

> Could we instruct the compiler on the semantics of some of the basic functions such as element/2, map_key/2, ... (maybe only guards even) so these reorderings can be attempted?

The compiler already knows the semantics of many BIFs and uses that knowledge to do a multitude of optimizations.

lib/compiler/src/beam_ssa_opt.erl
```diff
@@ -2245,6 +2295,176 @@ used_blocks([{L,Blk}|Bs], Def, Acc0) ->
...
+    %% Separate sinks that will force the reference to the tuple to be
+    %% saved on the stack from sinks that don't force.
+    {DefLocGC0,DefLocNoGC} =
+        partition(fun({{_,_}=FromTo,_}) ->
+                          map_get(FromTo, WillGC)
+                  end, Paths),
+
+    %% Here is an heuristic to avoid harmful sinking in
+    %% lists:mapfold/3 and similar functions.
+    DefLocGC = case DefLocGC0 of
```
jhogberg, line 2351: This looks a bit heavy, could you break it out into a separate function?

lib/compiler/src/beam_trim.erl
```diff
@@ -131,33 +147,47 @@ take_last_dead_1([{dead,Reg}|Is]) ->
...
-    recipe_cost_1(Ks, Cost+1);
-recipe_cost_1([_|Ks], Cost) ->
-    recipe_cost_1(Ks, Cost);
-recipe_cost_1([], Cost) -> Cost.
-
 %% try_remap([Recipe], [Instruction], FrameSize) ->
 %%           {[Instruction],[TrimInstruction]}.
 %%  Try to renumber Y registers in the instruction stream. The
-%%  first rececipe that works will be used.
+%%  first reccipe that works will be used.
```
jhogberg, line 190:
```suggestion
%%  first recipe that works will be used.
```

lib/compiler/src/beam_ssa_opt.erl
bjorng: Will do.

lib/compiler/src/beam_trim.erl
bjorng: Will fix.

PR #2598, opened 2020-04-16
feat: Adds ssh_sftp:list_dir_details/2,3

isaacsanders:
https://bugs.erlang.org/browse/ERL-235?jql=project%20%3D%20ERL%20AND%20resolution%20%3D%20Unresolved%20AND%20component%20%3D%20ssh

### DIFF
--- a/lib/ssh/doc/src/ssh_sftp.xml
+++ b/lib/ssh/doc/src/ssh_sftp.xml
@@ -227,6 +227,16 @@
     </func>
 
     <func>
+      <name name="list_dir_details" arity="2" since=""/>
+      <name name="list_dir_details" arity="3" since=""/>
+      <fsummary>Lists the directory.</fsummary>
+      <desc>
+        <p>Lists the given directory on the server, returning the
+          filenames and info as a list of pairs of strings and records.</p>
+      </desc>
+    </func>
+
+    <func>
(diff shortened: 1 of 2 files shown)


HansN:
Welcome First-Time Contributor!

THnak you, but documentation + a test case increases the likelihood to get the PR accepted :) See [Adding a new feature](https://github.com/erlang/otp/blob/master/CONTRIBUTING.md#adding-a-new-feature) in the wiki.

I can't handle this PR before Wednesday, since we have a code stop for OTP-23.0-rc3, then there is a less than two weeks window to get contributions into the OTP-23.0

isaacsanders:
Any advice on compiling with LibreSSL?

HansN:
The function returns `Details :: #ssh_xfer_attr{}`. This is an internal record in the ssh application since it is defined in an internal .hrl file in src.
If a record is returned from an API-function, it must be documented and exported in an include/X.hrl.

So either you need to do all of that, or preferably return some other datatype as a property list or a map.

The ssh_sftp:read_file_info returns a [file:#file_info{}](https://erlang.org/doc/man/file.html#type-file_info). Could that record be used for some of the data returned by list_dir_details?

HansN:
Closed due to inactivity

PR #2600, opened 2020-04-20
socket-nif: Fix string argument passing for setsockopt

RoadRunnr:
When passing a string into setsockopt we have to use the pointer to
the string, not a pointer to a pointer.

Affects the socket bindtodevice and tcp congestion option.


RoadRunnr:
Note: the same fix should IMHO also be applied to all 22.x releases.

PR #2602, opened 2020-04-22
Add a cost for copying binaries with the binary syntax

bjorng:
When constructing binaries using the binary syntax, there would
not be any cost in terms of reductions for copying large binaries,
and a process could copy many large binaries before being scheduled
out.

Add a cost of 1 reduction per 1 KiB of binary data copied in
binary construction.

### DIFF
--- a/erts/emulator/test/bs_construct_SUITE.erl
+++ b/erts/emulator/test/bs_construct_SUITE.erl
@@ -27,7 +27,8 @@
 	 mem_leak/1, coerce_to_float/1, bjorn/1, append_empty_is_same/1,
 	 huge_float_field/1, system_limit/1, badarg/1,
 	 copy_writable_binary/1, kostis/1, dynamic/1, bs_add/1,
-	 otp_7422/1, zero_width/1, bad_append/1, bs_append_overflow/1]).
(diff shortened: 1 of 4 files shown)


erts/emulator/test/bs_construct_SUITE.erl
```diff
@@ -889,6 +890,33 @@ bs_append_overflow_unsigned() ->
...
+    reds_at_least(2500, fun() -> <<TwoMeg/binary,TwoMeg:(2000*1024)/binary>> end),
+    ok.
+
+reds_at_least(N, Fun) ->
+    receive after 1 -> ok end,
+    {reductions,Red0} = process_info(self(), reductions),
+    _ = Fun(),
+    {reductions,Red1} = process_info(self(), reductions),
+    Diff = Red1 - Red0,
+    io:format("Expected at least ~p; got ~p\n", [N,Diff]),
```
lrascao, line 912: a question: is it ok allowing console output in OTP test runs?

bjorng: The output from `io:format/2` will go to the log file for the test case, not to the console.

PR #2604, opened 2020-04-23
Added a fun/4 with State access for keyboard-interactive

patrikbwi:
Hi,

When using 'keyboard-interactive' SSH authentication we need access to the user state to be able to generate prompts. This to be able to integrate with e.g. https://duo.com/

This patch adds a (backwards compatible) 'keyboard-interactive' fun/4 that has access to the user state.


HansN:
As usual:  a testcase and documentation helps acceptance :)

HansN:
Thanks for the PR!

PR #2488, opened 2019-12-21
[ERL-588] add mnesia funcs to table of contents

jeffweiss:
Because the mnesia docs did not have xml specs built for them, the functions were not appearing in the table of contents. This PR addresses that issue and also updates the xml documentation to use the `<name name="fun_name" arity="3" since=""/>` format instead of the older `<name since="">fun_name(A, B, C) -> :ok</name>` format.

I also removed xml documentation for `mnesia:dirty_slot/2` and `mnesia:set_debug_level/1` because they did not have specs in `mnesia.erl`.  I hope that's ok.

This is also the first time I've done this large of a documentation change, so if I've gotten wrong in some way, please let me know.

PR #2605, opened 2020-04-23
Dgud/mnesia/pr 2488

dgud:
Rebased (and fix issues) in #2488


dgud:
Thanks @jeffweiss for your work on this

PR #2609, opened 2020-04-24
Type erl_anno:anno() should not be opaque

josevalim:
The [docs](http://erlang.org/doc/apps/erts/absform.html)
refer to the annotation simply as LINE, however, if we were
to add the lines manually and give it to erl_eval:expr/2,
Dialyzer will complain as erl_anno:anno() is listed as opaque.

This makes erl_anno:anno() as non-opaque, with the
location() as public type but the annotation() as private.

### DIFF
--- a/lib/stdlib/src/erl_anno.erl
+++ b/lib/stdlib/src/erl_anno.erl
@@ -53,9 +53,9 @@
                     | {'text', string()}.
 
 -ifdef(DEBUG).
--opaque anno() :: [annotation(), ...].
+-type anno() :: [annotation(), ...].
 -else.
--opaque anno() :: location() | [annotation(), ...].
+-type anno() :: location() | [annotation(), ...].
 -endif.
 -type anno_term() :: term().
 


kostis:
What exactly do you mean by
> if we were to add the lines manually

Write just `42`?

Why can't you write `erl_anno:new(42)` instead as e.g. [PropEr](https://github.com/proper-testing/proper/) does all over the place? (For example, search for `?anno` in [proper_typeserver](https://github.com/proper-testing/proper/blob/master/src/proper_typeserver.erl).

Dialyzer is rarely an excuse to break programming abstractions.

josevalim:
I am aware. If LINE is no longer accepted, then the docs have to be updated and it needs to be listed as a breaking change. At the moment, the docs and specs are inconsistent.

kostis:
Yes. IMO it's the docs that need to be updated -- they should have been when the `erl_anno` module was introduced (18.x ?)

uabboli:
I agree with Kostis. I'll write a ticket on updating the docs.

If you cannot find the cause of Dialyzer warnings and suspect erl_anno i involved, you can compile erl_anno with DEBUG. erl_anno will then only work with annotations that are lists.

josevalim:
Thanks everyone! I will close to this then and make sure we are using `erl_anno` across the board.

PR #2610, opened 2020-04-26
[Fix] ERL-1227: mnesia_recover

zhangsiyang0629:
https://bugs.erlang.org/browse/ERL-1227?jql=text%20~%20%22mnesia%22

When a Mnesia node startup, it could not execute the transaction recovery process. 
The transaction recovery process entry is in mnesia_recover.erl file, but never called 
when a Mnesia node startup. This commit fix the bug.

### DIFF
--- a/lib/mnesia/src/mnesia_recover.erl
+++ b/lib/mnesia/src/mnesia_recover.erl
@@ -449,6 +449,9 @@ wait_for_decision(D, InitBy, N) ->
     if 
 	Outcome =:= committed -> {Tid, committed};
 	Outcome =:= aborted   -> {Tid, aborted};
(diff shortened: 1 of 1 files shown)


zhangsiyang0629:
The more information about this bug is descriped in ERL-1227

dgud:
It would make my life easier if you could squash the commits.

zhangsiyang0629:
> squash

I'm sorry. Because of the bad network, I couldn't clone the branch to my local computer, I don't know how squash the commits on the website without git Terminal. I'm sorry.

dgud:
No problem I'll fix it.

dgud:
Thanks, squashed and merged.

PR #2613, opened 2020-04-29
Remove docs for specs for functions in other modules

gomoripeti:
They are rejected by the compiler/linter since OTP 20 (commit 60e47cba).

See also https://bugs.erlang.org/browse/ERL-845.

The parser still seems to support spec with module (https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_parse.yrl#L90,
https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_parse.yrl#L681) and it is even documented in the Abstract Format (http://erlang.org/doc/apps/erts/absform.html#module-declarations-and-forms)
Should it be kept or removed?

### DIFF
--- a/system/doc/reference_manual/typespec.xml
+++ b/system/doc/reference_manual/typespec.xml
@@ -475,22 +475,22 @@
       <c>-spec</c> attribute. The general format is as follows:
     </p>
     <pre>
-  -spec Module:Function(ArgType1, ..., ArgTypeN) -> ReturnType.</pre>
+  -spec Function(ArgType1, ..., ArgTypeN) -> ReturnType.</pre>
     <p>
-      The arity of the function must match the number of arguments,
+      An implementation of the function with the same name must exist in the same module,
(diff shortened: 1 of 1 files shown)


michalmuskala:
AFAIK the linter still accepts:

```
-spec ?MODULE:fun() -> ok.
```

But yeah, I don't think having that is useful.

gomoripeti:
oh boy, you are right

uabboli:
Looks good in general.

The Module:Function syntax is still valid but the Module has to be the
current one. The syntax is used to signal that the module name is to
be included in the docs, see for instance erlang(3). Functions like
erlang:yield/0 cannot be called without the "erlang:" prefix.

Would you care to update the PR accordingly?

gomoripeti:
thanks for the extra insight, I'll update the text accordingly

uabboli:
Thanks for the pull request.

Merged commit c87d564896d73896627f8fde19c880c6da8365c6 into erlang:maint.

PR #2619, opened 2020-04-30
Fix compilation when enabling tracing in dist_util.hrl

JeromeDeBretagne:
Recent commits in dist_util.erl have broken compilation when tracing is enabled in dist_util.hrl
This PR fixes the few impacted ?trace and ?debug lines


JeromeDeBretagne:
Thanks Sverker!

PR #2620, opened 2020-05-02
New erl_uds_dist.erl example entirely in Erlang

JeromeDeBretagne:
Update `uds_dist.erl`, the alternative distribution protocol example using Unix Domain Sockets which was originally using a port driver written in C, by switching to distribution controllers implemented by Erlang processes instead, so with the code now written entirely in Erlang.

Add support for recent distribution features as well, such as dynamic node names and 32-bit big Creation numbers.

EDIT: The updated pull request now creates a new `uds2_dist.erl` module instead of replacing the original one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -337,6 +337,7 @@ JAVADOC-GENERATED
 
 /lib/kernel/src/inet_dns_record_adts.hrl
 /lib/kernel/test/gen_tcp_dist.erl
+/lib/kernel/test/erl_uds_dist.erl
 
 # kernel
 
--- a/erts/doc/src/alt_dist.xml
+++ b/erts/doc/src/alt_dist.xml
@@ -63,9 +63,9 @@
 	Erlang code will of course have to take care of retransmissions,
 	etc in this example). That is, depending on what you want
 	to do you perhaps do not need to implement a driver at all
-	and can then skip the driver related sections below.
-	The <c>gen_tcp_dist</c> example described in the
-	<seeguide marker="#distribution_module">Distribution
+	and can then skip the driver related sections below. The
+	<c>gen_tcp_dist</c> and <c>erl_uds_dist</c> examples described
+	in the <seeguide marker="#distribution_module">Distribution
 	Module</seeguide> section utilize distribution controller
 	processes and can be worth having a look at if you want to
 	use distribution controller processes.
@@ -187,13 +187,20 @@
       then voluntarily terminate when an exit signal is received.
     </p>
     <p>
-      An example implementation of a distribution module can be found
-      in
+      An example implementation of a distribution module can be found in
       <url href="gen_tcp_dist.erl">$ERL_TOP/lib/kernel/examples/gen_tcp_dist/src/gen_tcp_dist.erl</url>.
       It implements the distribution over TCP/IP using the <c>gen_tcp</c>
       API with distribution controllers implemented by processes. This
       instead of using port distribution controllers as the ordinary TCP/IP
       distribution uses.
+    </p>
+    <p>
+      Another example implementation of a distribution module can be found in
+      <url href="erl_uds_dist.erl">$ERL_TOP/lib/kernel/examples/erl_uds_dist/src/erl_uds_dist.erl</url>.
+      It implements the distribution over Unix domain sockets using the
+      <c>gen_tcp</c> API with distribution controllers implemented by processes.
+      Compared to the original <c>uds_dist</c> example using a port driver
+      written in C, <c>erl_uds_dist</c> is written entirely in Erlang.
     </p>
 
     <section>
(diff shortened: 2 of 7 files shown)


essen:
Nice.

rickard-green:
Thanks for the PR.

We don't want to lose the example implementation using a driver, though. This since we then only got examples of implementations using distribution processes. We can of course have two examples of how to implement Erlang distribution over uds, one using distribution ports and one using distribution processes.

We don't want to encourage usage of, not supported, internal interfaces such as `prim_file`, so usage of it in an example does not fit very well.

JeromeDeBretagne:
Thanks a lot Rickard for your feedback.

1/ I can update this pull request to create a new example, rename it to uds2_dist.erl (or any other naming suggestion that would be better/clearer) instead of replacing the existing one. I had thought that having access to the previous implementation in git history could be acceptable, especially since its code didn't receive any update in 10+ years, but I can see the value in having them side-by-side. 

2/ About the usage of prime_file, I didn't want to promote it either, that's why I've even added a comment explaining why it's used here exceptionnally (= distribution can be brought up very early during the early boot stage). I haven't found a better approach so far. I'm not familiar with the file_server early loading and the current rationale that prevents functions from the file module to at least work on the local filesystem early at boot (I didn't find this documented either btw). Do you have a suggestion? Would it be acceptable if I made the comment even more explicity with a warning mention?

Let me know how you would like me to proceed for both 1/ and 2/ and I'll work on an updated version. No rush on my side as I imagine your are really busy with the coming OTP 23.0 release.

Cheers, Jérôme

josevalim:
Regarding the usage of `prim_file`, I *think* you can pass the `raw` flag to `file:open/2`, as that does not use the `file_server`:

```erlang
file:open(SocketPathname ++ ".uds", [raw, read, write, binary])
```

JeromeDeBretagne:
Thanks José, I will try your suggestion. If it works, that should allow me to propose an updated version in the coming days.

JeromeDeBretagne:
Well, I was a bit too optimistic as my new implementation also needs `delete` to work but `file:delete/1` can't take a raw flag so I can't remove the 2 calls to `prim_file:delete/1` :-( It seems that I'm facing a current limitation of the `file.erl` module.

JeromeDeBretagne:
Thanks to José's comment, I've been able to remove all calls to functions from `prim_file` except one for which I haven't found a solution yet. I've created the [ERL-1244](https://bugs.erlang.org/browse/ERL-1244) issue to discuss that separately.

Rickard, I've followed your suggestion by creating a new `uds2_dist.erl` example, leaving the existing `uds_dist.erl` untouched, so that users can have both side-by-side.

I've also updated the comment to make it clear that the last remaining use of `prim_file:delete/1` is kept only as a short-term stopgap, to be replaced as soon as there is a fix for [ERL-1244](https://bugs.erlang.org/browse/ERL-1244), I hope this is an acceptable approach.

Let me know if you have any other feedback.

JeromeDeBretagne:
Now that [ERL-1244](https://bugs.erlang.org/browse/ERL-1244) has been implemented and merged, here is an updated version with all calls to `prim_file` removed as suggested previously. 

Let me know Rickard if you have any other feedback.

Thanks, Jérôme

rickard-green:
Great!

I'm not too fond of the "2" in the name, though. Perhaps `erl_uds_dist`, or `euds_dist`? Or if you can come up with something else that isn't numbered.

It would be nice with a test-case as well. It does not have to be possible to run all test-suites as in the `gen_tcp_dist` case, just a smoke test in `lib/kernel/test/erl_distribution_SUITE.erl` starting two nodes using your dist module checking that they can connect to each other. Modify `lib/kernel/test/Makefile` in a similar way as done for `gen_tcp_dist`, so your dist module become available in the test-suite. The `test_server` node wont be able to talk to these nodes, so you'll have to start them using `open_port` and let the nodes report the result in a file or connect back to the `test_server` node using `gen_tcp`.

JeromeDeBretagne:
Rickard, I've renamed the new distribution module to `erl_uds_dist` as that's more explicit than `uds2_dist` indeed.

About adding a test-case, I'm happy to look into it in a separate PR later on as it will be a steep learning curve for me (I'm not familiar with the notions of `test_server` or `open_port` you've mentioned at all) and I don't have the same bandwidth available at the moment. Thanks for your understanding.

rickard-green:
I can add a smoke test.

The `erl_uds_dist` implementation isn't mentioned in the documentation at all, so I guess it will go unnoticed by most users. Add a couple of words to `erts/doc/src/alt_dist.xml` where it talks about `uds_dist` explaining that there also exist a pure erlang example utilizing distribution controller processes in the `kernel/examples` directory.

rickard-green:
I've pushed a smoke-test to your branch in your repo. It should be quite easy to write more tests using the helper functions used by `erl_uds_dist_smoke_test()` if you want to add some more tests.

JeromeDeBretagne:
I've just added some documentation in `erts/doc/src/alt_dist.xml`, following the same format as used to describe `gen_tcp_dist`.

Thanks a lot for the smoke-test contribution, really appreciated, I will have to go through it calmly to properly understand how the test suite works and see how I could add some more tests then.

rickard-green:
Thanks! Merged into master now.

PR #2621, opened 2020-05-06
digraph: Avoid false edges for vertices with names containing a '_'

frej:
This patch fixes a bug in the digraph module where in_edges/2 and
out_edges/2 return false edges in the case they are called for a
vertex which has a '_' in its name term. The following example shows
the problem:

    G = new([acyclic]),
    A = add_vertex(G, 'A'),
    B = add_vertex(G, '_'),
    AB = add_edge(G, A, B),
    0 = out_degree(G, B), %% ok
    [] = out_edges(G, B), %% Fails as [AB] is returned

An edge, E, emanating from and incident on a vertex is stored in the
ntab-ets table as a record of the form {{out, V}, E} and {{in, V}, E}
respectively. As out_edges/2 and in_edges/2 use ets:select/2 to look
up the relevant edges, a vertex with a '_' somewhere a in the term
used as its name will trigger ets to do a wildcard match. In the
example above, this will return all out-edges in the whole graph. This
problem is fixed by switching from ets:select to ets:lookup.

This patch includes an extended test case which checks the correctness
of the fix described above as well as checking that the other
edge-related functions in digraph do not suffer from the same problem.

### DIFF
--- a/lib/stdlib/src/digraph.erl
+++ b/lib/stdlib/src/digraph.erl
@@ -230,7 +230,7 @@ in_neighbours(G, V) ->
       Edges :: [edge()].
 
 in_edges(G, V) ->
-    ets:select(G#digraph.ntab, [{{{in, V}, '$1'}, [], ['$1']}]).
+    [E || {{in, _}, E} <- ets:lookup(G#digraph.ntab, {in, V})].
 
 -spec out_degree(G, V) -> non_neg_integer() when
       G :: graph(),
@@ -255,7 +255,7 @@ out_neighbours(G, V) ->
       Edges :: [edge()].
 
 out_edges(G, V) ->
-    ets:select(G#digraph.ntab, [{{{out, V}, '$1'}, [], ['$1']}]).
+    [E || {{out, _}, E} <- ets:lookup(G#digraph.ntab, {out, V})].
 
 -spec add_edge(G, V1, V2) -> edge() | {'error', add_edge_err_rsn()} when
       G :: graph(),
--- a/lib/stdlib/test/digraph_SUITE.erl
+++ b/lib/stdlib/test/digraph_SUITE.erl
@@ -31,7 +31,7 @@
 	 init_per_group/2,end_per_group/2]).
 
 -export([opts/1, degree/1, path/1, cycle/1, vertices/1,
-	 edges/1, data/1, otp_3522/1, otp_3630/1, otp_8066/1]).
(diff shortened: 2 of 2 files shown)


frej:
Fixed a typo in the commit message.

lib/stdlib/src/digraph.erl
```diff
@@ -230,7 +230,7 @@ in_neighbours(G, V) ->
       Edges :: [edge()].
 
 in_edges(G, V) ->
-    ets:select(G#digraph.ntab, [{{{in, V}, '$1'}, [], ['$1']}]).
+    [E || {{in, _}, E} <- ets:lookup(G#digraph.ntab, {in, V})].
```
josevalim, line 233: I may be missing something, but I believe you can still keep it as a `ets:select/2`, you'd need to wrap '$1' in `{const, '$1'}`.

bjorng: @josevalim I can't see why wrapping `'$1'` would help. Trying to wrap `{in, V}` doesn't work either because `const` tuples are not allowed in patterns. The following works:

    ets:select(G#digraph.ntab, [{'$1',
                                 [{'=:=', {element, 1, '$1'}, {const, {in, V}}}],
                                 [{element, 2, '$1'}]}]).

But that is not an improvement over using `ets:lookup/2` and a list comprehension because the guard expression will have to be evaluated for every record in the table.

frej: I might be missing something too, but wrapping the edge term (`$1`) won't help us, the `_` in `V` will still be interpreted as a wildcard. What we would like to do is to make ets interpret the `_` in the match-head as a literal. Reading  [Match Specifications in Erlang](https://erlang.org/doc/apps/erts/match_spec.html), it looks like we cannot use `const` in match heads, only in the guard, so:

```Erlang
    ets:select(G#digraph.ntab,
               [{{{in, V}, '$1'},
                 [{'=:=', {const, V}, {element, 2, {element, 1, '$_'}}}],
                 ['$1']}]).
``` 

would give us what we want. Considering that a digraph might have a large number of edges, this would probably have much worse overhead than the extra garbage we produce by the list comprehension digging out the edges. The `select` version will have to traverse the complete table and match against each edge. The  `lookup`-based implementation will only process the edges incident to `V`, and with `ntab` being a bag, I assume this won't require a complete scan.

PR #2623, opened 2020-05-08
Make erlang:get_stacktrace/0 a regular function

michalmuskala:
Since the function now always returns an empty list, it does not
need to be implemented as a BIF, a regular function will work
just fine.
Additionally changing the spec should give some additional signal
forcing people to migrate.

---

This probably needs to update the preloaded modules as well, but I'll leave that to the maintainers


bjorng:
We plan to remove `erlang:get_stacktrace/0` in OTP 24.

michalmuskala:
Should I close this, then?

bjorng:
Yes.

PR #2624, opened 2020-05-09
inets: httpc:request parse url without normalization

eduardbcom:
https://bugs.erlang.org/browse/ERL-1215

### DIFF
--- a/lib/inets/src/http_client/httpc.erl
+++ b/lib/inets/src/http_client/httpc.erl
@@ -501,10 +501,7 @@ service_info(Pid) ->
 %%% Internal functions
 %%%========================================================================
 normalize_and_parse_url(Url) ->
-    case uri_string:normalize(Url) of
-        {error, _, _} = Error -> Error;
-        UriString -> uri_string:parse(UriString)
-    end.
+	uri_string:parse(Url).
 
 handle_request(Method, Url, 
                URI,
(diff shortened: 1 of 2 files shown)


lib/inets/test/httpc_SUITE.erl
```diff
@@ -381,10 +382,19 @@ get_space() ->
     [{"Test http get request with '%20' in the path of the URL."}].
 get_space(Config) when is_list(Config) ->
     Request  = {url(group_name(Config), "/space%20.html", Config), []},
-    {ok, {{_,200,_}, [_ | _], Body = [_ | _]}} = httpc:request(get, Request, [], []),
+    {ok, {{_,201,_}, [_ | _], Body = [_ | _]}} = httpc:request(get, Request, [], []),
```
eduardbcom, line 385: that was changed to 201, default response has same values
this can lead to hidden errors

lib/inets/src/http_client/httpc.erl
```diff
@@ -501,10 +501,7 @@ service_info(Pid) ->
 %%% Internal functions
 %%%========================================================================
 normalize_and_parse_url(Url) ->
```
eduardbcom, line 503: in case everything else is okay will rename it to parse_url

peterdmv:
Thank you for your effort trying to solve our support case, but we have decided to offer an other solution in #2629.

PR #2626, opened 2020-05-11
Allow any element subpatterns in LHS of ++-patterns

richcarl:
There is no actual need to restrict the elements of the LHS list
of a ++-pattern to be characters like $x or integers like 65. The
rewriting by the compiler is the same even if patterns like
`[foo,bar,baz] ++ Rest` or `[{x,42},{y,Y}]++Rest` are allowed.
Dropping this restriction removes any real need to use the traditional
list syntax [...|...].

### DIFF
--- a/lib/stdlib/src/erl_eval.erl
+++ b/lib/stdlib/src/erl_eval.erl
@@ -1185,10 +1185,8 @@ match1({bin,_,_}, _, _Bs, _BBs) ->
     throw(nomatch);
 match1({op,_,'++',{nil,_},R}, Term, Bs, BBs) ->
     match1(R, Term, Bs, BBs);
-match1({op,_,'++',{cons,Ai,{integer,A2,I},T},R}, Term, Bs, BBs) ->
-    match1({cons,Ai,{integer,A2,I},{op,Ai,'++',T,R}}, Term, Bs, BBs);
-match1({op,_,'++',{cons,Ai,{char,A2,C},T},R}, Term, Bs, BBs) ->
-    match1({cons,Ai,{char,A2,C},{op,Ai,'++',T,R}}, Term, Bs, BBs);
+match1({op,_,'++',{cons,Ai,H,T},R}, Term, Bs, BBs) ->
+    match1({cons,Ai,H,{op,Ai,'++',T,R}}, Term, Bs, BBs);
 match1({op,_,'++',{string,Ai,L},R}, Term, Bs, BBs) ->
     match1(string_to_conses(L, Ai, R), Term, Bs, BBs);
 match1({op,Anno,Op,A}, Term, Bs, BBs) ->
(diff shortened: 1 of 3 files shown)


alavrik:
With this restriction dropped, wouldn't we end up with two ways for doing the same thing?

The main problem I see with this is users beginning to use both styles in the same codebase, which isn't going to help readability. And then users would be continuously arguing which one is better and which one should be used.

In order to avoid this problem, a couple of things need to happen. 1) Erlang would need to settle on `++` being the new standard, 2) there should be a fully automated way to bring everybody on the new standard (e.g. gofix like tool), 3) there should be a deprecation period for the old pattern

Even with this, this would be a radical move for the current Erlang being a 25 year old language. There are so many artifacts (books, papers, ...) that won't ever be upgraded.

richcarl:
I'm not expecting people to be in any hurry to rewrite existing code this way, but new code could certainly use it. Just like the old `catch Expr` has to be kept around, but is not recommended for new code (it's not even officially deprecated yet). Yes, there's a whole lot of documentation out there with the normal cons syntax, but it could be slowly modernized.

But I think that this could be a great enabler - the Prolog style list notation is one of the most confusing things for new users of Erlang. It has the conceptual disadvantage that it is assymmetric - the LHS is an element, the RHS is a list. Learning to see the crucial difference between `[1,2,3,[4,5]]`, and `[1,2,3|[4,5]]` takes time - even experienced people occasionally mistype it, and the error can go undetected for a long time. Writing `[1,2,3]++[4,5]` in the latter case is much clearer, only one character longer, and the compiler will generate the same code anyway. Allowing you to use the corresponding notation for any kind of list patterns as well would make it possible to write all your code without ever bothering about the more obscure Prolog notation. Even just adding or matching a single element, I find `[X]++Rest` a better notation than `[X|Rest]` (for one, it doesn't require putting a closing bracket after the RHS).

And if in the future it becomes standard to use an auto-formatting editor, it would be easy to plug in erl_tidy to do this sort of rewrite automatically, if you want to update existing code to this style.

zuiderkwast:
Very nice! There's no reason for the limitation to exist IMO. I think [this documentation](http://erlang.org/doc/reference_manual/expressions.html#string-prefix-in-patterns) should use the word "list" rather than "string" and perhaps mention that the LHS needs to be a fixed-length list expression.

> ### String Prefix in Patterns
>
> When matching strings, the following is a valid pattern:
>
>     f("prefix" ++ Str) -> ...
>
> This is syntactic sugar for the equivalent, but harder to read:
>
>     f([$p,$r,$e,$f,$i,$x | Str]) -> ...

richcarl:
Do the OTP team have any feedback on this? Does it need its own EEP?

richcarl:
Closing this because of lack of interest from upstream. Would require an EEP to reopen.

PR #2627, opened 2020-05-12
Drop old code for handling pre-R15 abstract format

richcarl:
This code seems safe to remove now.


gomoripeti:
What's the approach for occurrence of such old format in the docs? (does not hurt to keep for historical insight, but keeping all formats that ever existed would clutter the docs) https://github.com/erlang/otp/blob/master/erts/doc/src/absform.xml#L374

richcarl:
@gomoripeti Thanks, removed from doc as well.

bjorng:
The test case `xref_SUITE:fun_mfa_r14/1` in the tools application fails.

bjorng:
@richcarl I have pushed a fixup commit that removes the failing test case. Please squash!

PR #2628, opened 2020-05-12
OCSP stapling client

ao-song:
Implementation of OCSP stapling client


lib/public_key/src/public_key.erl
```diff
@@ -1228,6 +1232,41 @@ pkix_test_data(#{} = Chain) ->
...
+%%--------------------------------------------------------------------
+get_ocsp_responder_id(Cert) ->
+    pubkey_ocsp:get_ocsp_responder_id(Cert).
+
+%%--------------------------------------------------------------------
+-spec get_nonce(boolean()) -> binary() | undefined.
+%%
+%% Description: Get an OCSP nonce
+%%--------------------------------------------------------------------
+get_nonce(OcspNonce) ->
```
IngelaAndin, line 1267: I do not think we need to have such an API function in public_key. I think the generating of the nonce should be handled in the ssl code and inputted into  API functions above.

lib/public_key/src/public_key.erl
```diff
@@ -1228,6 +1232,41 @@ pkix_test_data(#{} = Chain) ->
...
+    [Extn || Extn <- [pubkey_ocsp:get_nonce_extn(Nonce),
+                      pubkey_ocsp:get_acceptable_response_types_extn()],
+             erlang:is_record(Extn, 'Extension')].
+
+%%--------------------------------------------------------------------
+-spec get_ocsp_responder_id(#'Certificate'{}) -> binary().
+%%
+%% Description: Get the OCSP responder ID der
+%%--------------------------------------------------------------------
+get_ocsp_responder_id(Cert) ->
```
IngelaAndin, line 1259: I think we can remove get from the function name. It is good if all API functions start with ocsp and get does not add much value, and none of the other  API functions have "get" in their names.  See for example public_key:pkix_issuer_id/2

lib/public_key/src/public_key.erl
```diff
@@ -1228,6 +1232,41 @@ pkix_test_data(#{} = Chain) ->
 pkix_test_root_cert(Name, Opts) ->
     pubkey_cert:root_cert(Name, Opts).
 
+%%--------------------------------------------------------------------
+-spec ocsp_status(binary(), list(), undefined | binary()) ->
```
IngelaAndin, line 1236: We need to better know the return value of this function as it is an API function. It looks to me like it returns {ok, [#'SingleResponse'{}]  | {error, Reason}   where we need to specify reason too.

lib/ssl/src/ssl_connection.erl
```diff
@@ -1179,6 +1179,34 @@ certify(internal, #client_key_exchange{exchange_keys = Keys},
...
+                       ocsp_nonce := OcspNonce} = OcspState} = HsEnv,
+               connection_env = #connection_env{
+                   negotiated_version = Version}} = State,
+        Connection) ->
+    Result = public_key:ocsp_status(
+        OcspRespDer, ResponderCerts, OcspNonce),
+    NewOcspState = OcspState#{
+        ocsp_stapling_result => Result},
+    NewState = State#state{handshake_env =
+        HsEnv#handshake_env{ocsp_stapling_state = NewOcspState}},
```
IngelaAndin, line 1199: We can not have this check below here!  We need to just save the result in the state here, and all checks should be done during the certificate path_validation in the verify fun. In this first version, there will only ever be one  response,
as you have not implemented multiple responses.

lib/public_key/src/public_key.erl
peterdmv: I put this comment here so that we won't forget to DER encode the nonce according to RFC6066.

> In the case of the "id-pkix-ocsp-nonce" OCSP extension, [RFC2560] is
unclear about its encoding; for clarification, the nonce MUST be a
DER-encoded OCTET STRING, which is encapsulated as another OCTET
STRING (note that implementations based on an existing OCSP client
will need to be checked for conformance to this requirement).

lib/ssl/src/dtls_handshake.erl
```diff
@@ -79,7 +79,7 @@ client_hello(_Host, _Port, Cookie, ConnectionStates,
     Extensions = ssl_handshake:client_hello_extensions(TLSVersion, CipherSuites,
                                                        SslOpts, ConnectionStates, 
                                                        Renegotiation, undefined,
-                                                       undefined),
+                                                       undefined, undefined),
```
peterdmv, line 82: Do we want to implement OCSP even for DTLS or we document that it's only supported with TLS? @IngelaAndin

lib/ssl/doc/src/ssl.xml
```diff
@@ -1006,6 +1006,29 @@ fun(srp, Username :: string(), UserState :: term()) ->
...
+  included in the client hello to indicate the desire to receive certificate status information.
+  If false (the default), OCSP stapling will be disabled.</p>
+	</desc>
+      </datatype>
+
+      <datatype>
+	<name name="ocsp_responder_certs"/>
+	<desc><p>This option is a list of DER encoded certificates of OCSP responders which the client
+  trusts. This option is an empty list by default which means that the responders are implicitly
+  known to the server.</p>
```
peterdmv, line 1021: We should probably document here how the `ocsp_reponder_certs` option interacts with the list of `certs` received in `BasicOCSPResponse`. We should also consider if this option should take the value of `cacerts`/`cacertfile` if it is not set.

lib/public_key/src/public_key.erl
IngelaAndin: For the DER encoding, we want to use a public_key function. I think we should be able to use
public_key:der_encode/2. I am not sure if we need to create a special  ASN1 type spec to solve it or if we can use an existing one.

lib/ssl/src/dtls_handshake.erl
IngelaAndin: There should not be anything DTLS specific for this functionality, only that it might need some
inclusion in DTLS code as initial hello handling has some differences for DTLS. So there is really no reason not to support it for DTLS as well.

lib/ssl/src/ssl_connection.erl
```diff
@@ -1179,6 +1179,34 @@ certify(internal, #client_key_exchange{exchange_keys = Keys},
...
 	    handle_own_alert(Alert, Version, ?FUNCTION_NAME, State)
     end;
+%% The response will be handled only when a certificate status request has
+%% been sent by the client and confirmed by the server.
+certify(internal, #certificate_status{response = OcspRespDer},
+        #state{handshake_env = #handshake_env{
+                   ocsp_stapling_state = #{
+                       is_client_hello_requested := true,
+                       is_server_hello_requested := true,
+                       ocsp_responder_certs := ResponderCerts,
```
peterdmv, line 1189: ocsp_reponder_certs should probably go to static_env. Should we put these certs into cert_db?

```erlang
-record(static_env, {
                     role                  :: client | server,
                     transport_cb          :: atom(),   % callback module
                     protocol_cb           :: tls_connection | dtls_connection,
                     data_tag              :: atom(),   % ex tcp.
                     close_tag             :: atom(),   % ex tcp_closed
                     error_tag             :: atom(),   % ex tcp_error
                     passive_tag           :: atom(),   % ex tcp_passive
                     host                  :: string() | inet:ip_address(),
                     port                  :: integer(),
                     socket                :: port() | tuple(), %% TODO: dtls socket
                     cert_db               :: reference() | 'undefined',
                     session_cache         :: db_handle(),
                     session_cache_cb      :: atom(),
                     crl_db                :: term(),
                     file_ref_db          :: db_handle(),
                     cert_db_ref          :: certdb_ref() | 'undefined',
                     trackers              :: [{atom(), pid()}] | 'undefined' %% Trac
or listen socket
                    }).
```

lib/ssl/src/tls_connection.erl
```diff
@@ -618,13 +622,22 @@ init({call, From}, {start, Timeout},
     %% negotiated_version is also used by the TLS 1.3 state machine and is set after
     %% ServerHello is processed.
     RequestedVersion = tls_record:hello_version(Versions),
+    IsCertStatusRequested = maps:is_key(
```
peterdmv, line 625: This variable is redundant as we already have this information in `ssl_options`. `ssl_options` is used to build the ClientHello message and we know that the ClientHello will have an ocsp_stapling extension when it is enabled in `ssl_options`.

lib/ssl/src/ssl_connection.erl
```diff
@@ -1179,6 +1179,34 @@ certify(internal, #client_key_exchange{exchange_keys = Keys},
 	#alert{} = Alert ->
 	    handle_own_alert(Alert, Version, ?FUNCTION_NAME, State)
     end;
+%% The response will be handled only when a certificate status request has
+%% been sent by the client and confirmed by the server.
+certify(internal, #certificate_status{response = OcspRespDer},
+        #state{handshake_env = #handshake_env{
+                   ocsp_stapling_state = #{
+                       is_client_hello_requested := true,
```
peterdmv, line 1187: See my comment about this variable in tls_connection.erl. I think you can use the ssl option here that enables stapling. A compliant ssl server will only send a `certificate_status` if the client has sent a CertificateStatusRequest extension.
I also think we don't need the is_server_hello_requested variable in this case, and this way the ocsp_stapling_state can be expressed by using the above mentioned ssl option that enables ocsp stapling and the cert_db/ocsp_cert_db in static_env.

lib/ssl/src/ssl_connection.erl
peterdmv: I agree with @IngelaAndin and I'd like to add here that I think we can have multiple reponses here even according to rfc6960. Current implementation won't handle responses with multiple SingleReponse elements.
```asn.1
ResponseData ::= SEQUENCE {
      version              [0] EXPLICIT Version DEFAULT v1,
      responderID              ResponderID,
      producedAt               GeneralizedTime,
      responses                SEQUENCE OF SingleResponse,
      responseExtensions   [1] EXPLICIT Extensions OPTIONAL }
```

lib/ssl/src/ssl_connection.hrl
```diff
@@ -81,7 +81,8 @@
                         public_key_info      :: ssl_handshake:public_key_info() | 'undefined',
                         premaster_secret     :: binary() | secret_printout() | 'undefined',
                         server_psk_identity         :: binary() | 'undefined',  % server psk identity hint
-                        ticket_seed
+                        ticket_seed,
+                        ocsp_stapling_state = #{}
```
peterdmv, line 85: See my comments above, we probably don't need the ocsp_stapling_state.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -349,12 +349,12 @@ certify(#certificate{asn1_certificates = ASN1Certs}, CertDbHandle, CertDbRef,
           crl_check := CrlCheck,
           log_level := Level,
           depth := Depth} = Opts, CRLDbHandle, Role, Host) ->
-    
+
     ServerName = server_name(ServerNameIndication, Host, Role),
-    [PeerCert | ChainCerts ] = ASN1Certs,
```
peterdmv, line 354: This is a general comment about fixing whitespaces. We usually put them into a separate commit, it makes it easier to review the code.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -796,6 +838,14 @@ decode_handshake(Version, ?SERVER_HELLO, <<?BYTE(Major), ?BYTE(Minor), Random:32
        extensions = HelloExtensions};
 decode_handshake(_Version, ?CERTIFICATE, <<?UINT24(ACLen), ASN1Certs:ACLen/binary>>) ->
     #certificate{asn1_certificates = certs_to_list(ASN1Certs)};
+%% RFC 6066, servers return a certificate response along with their certificate
+%% by sending a "CertificateStatus" message immediately after the "Certificate"
+%% message and before any "ServerKeyExchange" or "CertificateRequest" messages.
+decode_handshake(_Version, ?CERTIFICATE_STATUS, <<?BYTE(StatusType),
```
peterdmv, line 844: As `status_type` can only take one value, is it better to match on this value and let the last clause generate a `HANDSHAKE_FAILURE` if there is no match. This way we can remove the `get_ocsp_reponse/2` function.
```
enum { ocsp(1), (255) } CertificateStatusType;
```
```erlang
decode_handshake(_Version, ?CERTIFICATE_STATUS, <<?BYTE(?CERTIFICATE_STATUS_TYPE_OCSP), ...
```

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1151,6 +1202,30 @@ maybe_add_tls13_extensions({3,4},
...
+            OcspRequestExtns = public_key:ocsp_extensions(OcspNonce),
+            Req = #ocsp_status_request{responder_id_list = OcspResponderList,
+                                       request_extensions = OcspRequestExtns},
+            CertStatusReqExtn = #certificate_status_request{
+                status_type = ?CERTIFICATE_STATUS_TYPE_OCSP,
+                request = Req
+            },
+            HelloExtensions#{ocsp_stapling => CertStatusReqExtn}
+    end.
+
```
peterdmv, line 1223: I would call this function maybe_add_certificate_status_request to follow the pattern. `maybe_` expresses that there is a condition for adding the extension. As there is only one OCSP extension defined by RFC6066 it is easier to understand if we name that specific extension.

Also it could be factored into two separate clauses:
```erlang
maybe_add_certificate_status_request(_Version, #{ocsp_stapling := false}, _OcspNonce, HelloExtensions) ->
    HelloExtensions;
maybe_add_certificate_status_request(
    _Version, #{ocsp_stapling := true,
                ocsp_responder_certs := OcspResponderCerts}, _OcspNonce, HelloExtensions) ->
...
```

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1151,6 +1202,30 @@ maybe_add_tls13_extensions({3,4},
...
+get_ocsp_responder_list([], Acc) ->
```
peterdmv, line 1224: This is a pattern that we usually use to hide the initialization of the list accumulator:
```erlang
get_ocsp_responder_list(L) ->
    get_ocsp_responder_list(L, []).
%%
get_ocsp_responder_list([], Acc) -> ...
```

lib/public_key/src/public_key.erl
ao-song: > I think we can remove get from the function name. It is good if all API functions start with ocsp and get does not add much value, and none of the other API functions have "get" in their names. See for example public_key:pkix_issuer_id/2

done

lib/public_key/src/public_key.erl
ao-song: > I do not think we need to have such an API function in public_key. I think the generating of the nonce should be handled in the ssl code and inputted into API functions above.

maybe change the name to ocsp_nonce in public_key? since we may need encode the nonce first in public_key according to Peter's comment.

lib/ssl/src/ssl_connection.erl
ao-song: > See my comment about this variable in tls_connection.erl. I think you can use the ssl option here that enables stapling. A compliant ssl server will only send a `certificate_status` if the client has sent a CertificateStatusRequest extension.
> I also think we don't need the is_server_hello_requested variable in this case, and this way the ocsp_stapling_state can be expressed by using the above mentioned ssl option that enables ocsp stapling and the cert_db/ocsp_cert_db in static_env.

how we ensure the following in rfc6066 in this case?

If a server returns a "CertificateStatus" message, then the server MUST have included an extension of type 
"status_request" with empty "extension_data" in the extended server hello.

lib/ssl/src/ssl_connection.erl
ao-song: > I agree with @IngelaAndin and I'd like to add here that I think we can have multiple reponses here even according to rfc6960. Current implementation won't handle responses with multiple SingleReponse elements.
> 
> ```asn
> ResponseData ::= SEQUENCE {
>       version              [0] EXPLICIT Version DEFAULT v1,
>       responderID              ResponderID,
>       producedAt               GeneralizedTime,
>       responses                SEQUENCE OF SingleResponse,
>       responseExtensions   [1] EXPLICIT Extensions OPTIONAL }
> ```

I have noticed this, but it is hard to find the response related to the server certificate from a list of responses as the order of the list is related to the order of request which is send from the server and there is no field as I have checked can help to find the right cert.

lib/ssl/src/ssl_connection.hrl
ao-song: > See my comments above, we probably don't need the ocsp_stapling_state.

Currently I store the ocsp status result in the state, if we will remove this, where should we store the result then?

lib/public_key/src/public_key.erl
ao-song: > We need to better know the return value of this function as it is an API function. It looks to me like it returns {ok, [#'SingleResponse'{}] | {error, Reason} where we need to specify reason too.

done

lib/ssl/doc/src/ssl.xml
ao-song: > We should probably document here how the `ocsp_reponder_certs` option interacts with the list of `certs` received in `BasicOCSPResponse`. We should also consider if this option should take the value of `cacerts`/`cacertfile` if it is not set.

partially done, since I have checked the reply seems only cert is provided in the certs field.

lib/ssl/src/tls_connection.erl
ao-song: > This variable is redundant as we already have this information in `ssl_options`. `ssl_options` is used to build the ClientHello message and we know that the ClientHello will have an ocsp_stapling extension when it is enabled in `ssl_options`.

done

lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -0,0 +1,290 @@
...
+
+    #'ResponseData'{
+        responderID = ResponderID
+    } = ResponseData,
+
+    case verify_ocsp_signature(
+        public_key:der_encode('ResponseData', ResponseData),
+        SignatureAlgo#'AlgorithmIdentifier'.algorithm,
+        Signature, Certs ++ ResponderCerts,
+        ResponderID) of
```
peterdmv, line 63: The certs attribute of the BasicOCSPResponse is meant to help us to verify the signature of the response:
```
   BasicOCSPResponse       ::= SEQUENCE {
      tbsResponseData      ResponseData,
      signatureAlgorithm   AlgorithmIdentifier,
      signature            BIT STRING,
      certs            [0] EXPLICIT SEQUENCE OF Certificate OPTIONAL }

   The value for signature SHALL be computed on the hash of the DER
   encoding of ResponseData.  The responder MAY include certificates in
   the certs field of BasicOCSPResponse that help the OCSP client verify
   the responder's signature.  If no certificates are included, then
   certs SHOULD be absent.
```
The ResponderCerts on the other hand is a list of certs that the client trusts with the following definition:
```
   In the OCSPStatusRequest, the "ResponderIDs" provides a list of OCSP
   responders that the client trusts.  A zero-length "responder_id_list"
   sequence has the special meaning that the responders are implicitly
   known to the server, e.g., by prior arrangement.
```
So we have multiple scenarios here, depending on if certs is empty or not, ocsp_responder_certs configured or not, and also if we trust the certs that we got in the BasicOCSPReponse. I'm thinking that maybe we need an additional options just for that.

**Scenario 1: certs is empty, ocsp_responder_certs not defined**
Responders implicitly known to the client -> use cacerts?

**Scenario 2: certs is not empty, ocsp_responder_certs not defined**
Responders implicitly known to the client -> use cacerts then use certs if we trust it? new option?

**Scenario 2: ocsp_responder_certs is defined**
Responders implicitly known to the client -> use only ocsp_responder_certs if cert not found check cacerts, if cert still not found check certs if we trust it?

lib/ssl/src/ssl_handshake.erl
ao-song: > As `status_type` can only take one value, is it better to match on this value and let the last clause generate a `HANDSHAKE_FAILURE` if there is no match. This way we can remove the `get_ocsp_reponse/2` function.
> 
> ```
> enum { ocsp(1), (255) } CertificateStatusType;
> ```
> 
> ```erlang
> decode_handshake(_Version, ?CERTIFICATE_STATUS, <<?BYTE(?CERTIFICATE_STATUS_TYPE_OCSP), ...
> ```

done

lib/ssl/src/ssl_handshake.erl
ao-song: done

ao-song: > I would call this function maybe_add_certificate_status_request to follow the pattern. `maybe_` expresses that there is a condition for adding the extension. As there is only one OCSP extension defined by RFC6066 it is easier to understand if we name that specific extension.
> 
> Also it could be factored into two separate clauses:
> 
> ```erlang
> maybe_add_certificate_status_request(_Version, #{ocsp_stapling := false}, _OcspNonce, HelloExtensions) ->
>     HelloExtensions;
> maybe_add_certificate_status_request(
>     _Version, #{ocsp_stapling := true,
>                 ocsp_responder_certs := OcspResponderCerts}, _OcspNonce, HelloExtensions) ->
> ...
> ```

done

lib/ssl/src/ssl_handshake.erl
ao-song: > usually

done

lib/ssl/src/ssl_connection.erl
peterdmv: Additional comment about the elements of the response. A server can be configured with multiple certificates and the BasicOCSPReponse will return a SingleReponse for each server cert and we should choose the right one based on which one was chosen for the handshake (server's Certificate message).

lib/public_key/src/public_key.erl
ao-song: > I put this comment here so that we won't forget to DER encode the nonce according to RFC6066.
> 
> > In the case of the "id-pkix-ocsp-nonce" OCSP extension, [RFC2560] is
> > unclear about its encoding; for clarification, the nonce MUST be a
> > DER-encoded OCTET STRING, which is encapsulated as another OCTET
> > STRING (note that implementations based on an existing OCSP client
> > will need to be checked for conformance to this requirement).

done

lib/ssl/src/ssl_handshake.erl
```diff
@@ -745,6 +754,38 @@ encode_extensions([#pre_shared_key_server_hello{selected_identity = Identity} |
...
+encode_responderID_list(List) ->
+    do_encode_responderID_list(List, <<>>).
+
+%% ResponderID is DER-encoded ASN.1 type defined in RFC6960.
+do_encode_responderID_list([Responder | Rest], Acc)
+  when is_binary(Responder) ->
+    Len = byte_size(Responder),
+    do_encode_responderID_list(
+        Rest, <<Acc/binary, ?UINT16(Len), Responder/binary>>);
+do_encode_responderID_list([], Acc) ->
```
peterdmv, line 777: This could be the first clause instead to follow to pattern.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2648,6 +2724,31 @@ decode_extensions(<<?UINT16(?PRE_SHARED_KEY_EXT), ?UINT16(Len),
...
 
+%% RFC6066, if a server returns a "CertificateStatus" message, then
+%% the server MUST have included an extension of type "status_request"
+%% with empty "extension_data" in the extended server hello.
+decode_extensions(<<?UINT16(?STATUS_REQUEST), ?UINT16(Len),
+                    _ExtensionData:Len/binary, Rest/binary>>, Version,
+                    MessageType = server_hello, Acc)
+  when Len =:= 0 ->
+    decode_extensions(Rest, Version, MessageType,
+                      Acc#{ocsp_stapling => undefined});
```
peterdmv, line 2735: It would be better if the map key is called the same as the decoded extension, in this case `status_request`. This change should be propagated to other parts of the code.
`status_request` itself is overloaded as it can contain either a `CertificateStatusRequest` or `CertificateStatus` (TLS 1.3). There is also a missing encode_extension clause for encoding a `status_request`. It will be used on the server side in the future. With the help of that encode clause we will be able to extend the property tests suite.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2648,6 +2724,31 @@ decode_extensions(<<?UINT16(?PRE_SHARED_KEY_EXT), ?UINT16(Len),
...
+%% RFC6066.
+decode_extensions(<<?UINT16(?STATUS_REQUEST), ?UINT16(Len),
+                    CertStatus:Len/binary, Rest/binary>>, Version,
+                    MessageType, Acc) ->
+    case CertStatus of
+        <<?BYTE(?CERTIFICATE_STATUS_TYPE_OCSP),
+          ?UINT24(OCSPLen),
+          ASN1OCSPResponse:OCSPLen/binary>> ->
+            decode_extensions(Rest, Version, MessageType,
+                      Acc#{ocsp_stapling => ASN1OCSPResponse});
```
peterdmv, line 2747: Same comment: ocsp_stapling -> status_request

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2648,6 +2724,31 @@ decode_extensions(<<?UINT16(?PRE_SHARED_KEY_EXT), ?UINT16(Len),
...
+        _Other ->
+            decode_extensions(Rest, Version, MessageType, Acc)
```
peterdmv, line 2749: At this point we have recognized a `status_request` extension but it contains invalid data. I think we should probably abort the handshake instead of dropping the extension @IngelaAndin ?
```erlang
throw(?ALERT_REC(?FATAL, ?HANDSHAKE_FAILURE, {malformed_status_request, Other})
```
But we should test if this works or not. This would be a better way to handle protocol errors.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -3319,4 +3421,4 @@ empty_extensions(_, server_hello) ->
       ec_point_formats => undefined}.
 
 handle_log(Level, {LogLevel, ReportMap, Meta}) ->
-    ssl_logger:log(Level, LogLevel, ReportMap, Meta).
+    ssl_logger:log(Level, LogLevel, ReportMap, Meta).
```
peterdmv, line 3424: Missing new line may confuse our scripts @IngelaAndin ?

lib/ssl/src/ssl_handshake.hrl
```diff
@@ -441,4 +440,29 @@
 
 -record(signature_algorithms_cert, {signature_scheme_list}).
 
--endif. % -ifdef(ssl_handshake).
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%% OCSP stapling
+%% RFC 6066 section 8 Certificate Status Request
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
```
peterdmv, line 447: I think this section belongs to the RFC 6066 section further up in this file. We forgot to update to comment there, max_fragment_length is already supported.

lib/ssl/src/ssl_connection.erl
IngelaAndin: So we have a new understanding of this. And we should actual  postpone the path_validation to we have this info in the stapling case. And let save this until we prototped the new path_validation handling.

lib/ssl/src/tls_connection.erl
```diff
@@ -580,24 +580,28 @@ init({call, From}, {start, Timeout},
...
-                            versions := [HelloVersion|_] = Versions,
-                            session_tickets := SessionTickets} = SslOpts,
-	    session = NewSession,
-	    connection_states = ConnectionStates0
-	   } = State0) ->
+                        %% Use highest version in initial ClientHello.
+                        %% Versions is a descending list of supported versions.
+                        versions := [HelloVersion|_] = Versions,
+                        session_tickets := SessionTickets,
+                        ocsp_responder_certs := OcspResponderCerts,
```
peterdmv, line 587: In this function we should first match out the option `ocsp_stapling` to see if stapling is enabled. If it is enabled and only if it is enabled we calculate the OcspNonce. `public_key:ocsp_nonce` could be wrapped in another function that takes care of this. Current code will always calculate it even if `ocsp_stapling` is set to false.

lib/ssl/src/tls_connection.erl
```diff
@@ -622,9 +626,15 @@ init({call, From}, {start, Timeout},
                          connection_env = CEnv#connection_env{
                                             negotiated_version = RequestedVersion},
                          session = Session,
-                         handshake_env = HsEnv#handshake_env{tls_handshake_history = Handshake},
+                         handshake_env = HsEnv#handshake_env{
+                             tls_handshake_history = Handshake,
+                             ocsp_stapling_state =
+                                 #{is_server_hello_requested => undefined,
```
peterdmv, line 632: `is_server_hello_requested` and `is_client_hello_requested` could be replaced by one variable. We have only one shared state for OCSP stapling. First, we evaluate the client options, and then we set a variable here in the state that represent the negotiated state of the stapling function. Before the negotiation it will be set to the same value as the `ocsp_stapling` option. Later we will change its value based on the the received ServerHello extensions (the negotiated value).

lib/ssl/src/tls_connection.erl
```diff
@@ -662,8 +672,18 @@ hello(internal, #server_hello{extensions = Extensions} = Hello,
...
+    %% "status_request" with empty "extension_data" in the extended
+    %% server hello.
+    OcspState = HsEnv#handshake_env.ocsp_stapling_state,
+    IsCertStatusRequested = maps:is_key(ocsp_stapling, Extensions),
+    {next_state, user_hello,
+     State#state{start_or_recv_from = undefined,
+                 handshake_env = HsEnv#handshake_env{
+                     hello = Hello,
+                     ocsp_stapling_state = OcspState#{
+                         is_server_hello_requested := IsCertStatusRequested}}},
```
peterdmv, line 686: We can set that above mentioned single variable here to the negotiated value.

lib/ssl/src/tls_connection.erl
```diff
@@ -662,8 +672,18 @@ hello(internal, #server_hello{extensions = Extensions} = Hello,
...
              handshake_env = HsEnv,
              start_or_recv_from = From} = State) ->
-    {next_state, user_hello, State#state{start_or_recv_from = undefined,
-                                         handshake_env = HsEnv#handshake_env{hello = Hello}},
+    %% RFC6066.8, If a server returns a "CertificateStatus" message,
+    %% then the server MUST have included an extension of type
+    %% "status_request" with empty "extension_data" in the extended
+    %% server hello.
+    OcspState = HsEnv#handshake_env.ocsp_stapling_state,
+    IsCertStatusRequested = maps:is_key(ocsp_stapling, Extensions),
```
peterdmv, line 680: We could verify here that the `status_request` extension is empty as mandated by the RFC and fail the handshake otherwise.

lib/ssl/src/tls_connection.erl
```diff
@@ -713,27 +733,41 @@ hello(internal, #client_hello{client_version = ClientVersion} = Hello,
...
       #state{connection_states = ConnectionStates0,
              connection_env = #connection_env{negotiated_version = ReqVersion} = CEnv,
 	     static_env = #static_env{role = client},
-             handshake_env = #handshake_env{renegotiation = {Renegotiation, _}},
+             handshake_env = #handshake_env{
+                 renegotiation = {Renegotiation, _},
+                 ocsp_stapling_state = OcspState} = HsEnv,
 	     ssl_options = SslOptions} = State) ->
+    %% check if server has sent a certifucate status message in hello
+    IsCertStatusRequested = maps:is_key(ocsp_stapling, Extensions),
```
peterdmv, line 745: Same comment here about the error handling.

lib/ssl/src/tls_connection.erl
```diff
@@ -713,27 +733,41 @@ hello(internal, #client_hello{client_version = ClientVersion} = Hello,
...
     case tls_handshake:hello(Hello, SslOptions, ConnectionStates0, Renegotiation) of
 	#alert{} = Alert -> %%TODO
 	    ssl_connection:handle_own_alert(Alert, ReqVersion, hello,
                                             State#state{connection_env =
-                                                            CEnv#connection_env{negotiated_version = ReqVersion}});
+                                                            CEnv#connection_env{negotiated_version = ReqVersion},
+                                                        handshake_env = HsEnv#handshake_env{
+                                                            ocsp_stapling_state = OcspState#{
+                                                                is_server_hello_requested := IsCertStatusRequested}}
```
peterdmv, line 753: Same comment about the single variable that represent the negotiated ocsp state.

lib/ssl/src/tls_connection.erl
```diff
@@ -713,27 +733,41 @@ hello(internal, #client_hello{client_version = ClientVersion} = Hello,
...
+                                                        });
         %% Legacy TLS 1.2 and older
 	{Version, NewId, ConnectionStates, ProtoExt, Protocol} ->
 	    ssl_connection:handle_session(Hello, 
-					  Version, NewId, ConnectionStates, ProtoExt, Protocol, State);
+					  Version, NewId, ConnectionStates, ProtoExt, Protocol,
+                      State#state{handshake_env = HsEnv#handshake_env{
+                          ocsp_stapling_state = OcspState#{
+                              is_server_hello_requested := IsCertStatusRequested}}});
```
peterdmv, line 761: Same comment about the single variable that represent the negotiated ocsp state.

lib/ssl/src/tls_connection.erl
```diff
@@ -713,27 +733,41 @@ hello(internal, #client_hello{client_version = ClientVersion} = Hello,
...
         %% TLS 1.3
         {next_state, wait_sh, SelectedVersion} ->
             %% Continue in TLS 1.3 'wait_sh' state
             {next_state, wait_sh,
              State#state{
-               connection_env = CEnv#connection_env{negotiated_version = SelectedVersion}},
+               connection_env = CEnv#connection_env{negotiated_version = SelectedVersion},
+               handshake_env = HsEnv#handshake_env{
+                   ocsp_stapling_state = OcspState#{
+                       is_server_hello_requested := IsCertStatusRequested}}},
```
peterdmv, line 770: Same comment about the single variable that represent the negotiated ocsp state.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -680,9 +681,10 @@ do_start(#server_hello{cipher_suite = SelectedCipherSuite,
         %% of the triggering HelloRetryRequest.
         ClientKeyShare = ssl_cipher:generate_client_shares([SelectedGroup]),
         TicketData = get_ticket_data(self(), SessionTickets, UseTicket),
+        OcspNonce = public_key:ocsp_nonce(OcspNonceOpt),
```
peterdmv, line 684: We should only calculate nonce if stapling is enabled.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -712,7 +714,9 @@ do_start(#server_hello{cipher_suite = SelectedCipherSuite,
         State = State3#state{
                   connection_states = ConnectionStates,
                   session = Session0#session{session_id = Hello#client_hello.session_id},
-                  handshake_env = HsEnv#handshake_env{tls_handshake_history = HHistory},
+                  handshake_env = HsEnv#handshake_env{
+                      tls_handshake_history = HHistory,
+                      ocsp_stapling_state = #{ocsp_nonce => OcspNonce}},
```
peterdmv, line 719: We should store here if OCSP has been negotiated.

lib/ssl/src/ssl_handshake.erl
ao-song: > This is a general comment about fixing whitespaces. We usually put them into a separate commit, it makes it easier to review the code.

OK

lib/ssl/src/ssl_handshake.erl
ao-song: > This could be the first clause instead to follow to pattern.

Done, but shouldn't we put the match clause with highest possibility to the first?

lib/ssl/src/ssl_handshake.erl
ao-song: > Same comment: ocsp_stapling -> status_request

done

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -1302,7 +1306,18 @@ process_certificate(#certificate_1_3{certificate_list = Certs0},
                                             SslOptions, CRLDbHandle, Role, Host) of
                 {ok, {PeerCert, PublicKeyInfo}} ->
                     State = store_peer_cert(State0, PeerCert, PublicKeyInfo),
-                    {ok, {State, wait_cv}};
+                    NewState = check_ocsp(State, Certs0),
```
peterdmv, line 1309: We should refactor this part. process_certificate should use the negotiated OCSP state set when we received the ServerHello. All this should be handled in the validate_certificate_chain function.

lib/ssl/src/ssl_handshake.hrl
ao-song: > I think this section belongs to the RFC 6066 section further up in this file. We forgot to update to comment there, max_fragment_length is already supported.

done

lib/ssl/test/ssl_ocsp_SUITE.erl
```diff
@@ -0,0 +1,367 @@
...
+     {group, revoked_1_3}].
+
+groups() -> 
+    [{tls1_2, [], tls1_2_tests()},
+     {tls1_3, [], tls1_3_tests()},
+     {revoked_1_2, [], tls1_2_revoked_tests()},
+     {revoked_1_3, [], tls1_3_revoked_tests()}].
+
+tls1_2_tests() ->
+    [ocsp_stapling_without_nonce_and_responder_certs_tls1_2,
```
peterdmv, line 45: If this is the base case without nonce and without responder certs that it could just be called ocsp_stapling_tls1_2.

lib/ssl/test/ssl_ocsp_SUITE.erl
```diff
@@ -0,0 +1,367 @@
...
+     ocsp_stapling_with_nonce_tls1_2,
+     ocsp_stapling_with_responder_cert_tls1_2].
+
+tls1_2_revoked_tests() ->
+    [ocsp_stapling_revoked_tls1_2].
+
+tls1_3_tests() ->
+    [ocsp_stapling_without_nonce_and_responder_certs_tls1_3,
```
peterdmv, line 53: Same comment here, base case could have a short name.

lib/ssl/src/tls_connection.erl
ao-song: > In this function we should first match out the option `ocsp_stapling` to see if stapling is enabled. If it is enabled and only if it is enabled we calculate the OcspNonce. `public_key:ocsp_nonce` could be wrapped in another function that takes care of this. Current code will always calculate it even if `ocsp_stapling` is set to false.

done

lib/ssl/src/ssl_handshake.erl
IngelaAndin: I think such optimizations shall be done by the compiler or JIT (that is worked on). Unless it
is measured to be a bottleneck, which is seldom the case.

lib/ssl/src/tls_connection.erl
ao-song: > `is_server_hello_requested` and `is_client_hello_requested` could be replaced by one variable. We have only one shared state for OCSP stapling. First, we evaluate the client options, and then we set a variable here in the state that represent the negotiated state of the stapling function. Before the negotiation it will be set to the same value as the `ocsp_stapling` option. Later we will change its value based on the the received ServerHello extensions (the negotiated value).

done

lib/ssl/src/tls_connection.erl
ao-song: > We can set that above mentioned single variable here to the negotiated value.

done

lib/ssl/src/tls_handshake_1_3.erl
ao-song: > We should only calculate nonce if stapling is enabled and negotiated.

done

lib/ssl/test/ssl_ocsp_SUITE.erl
```diff
@@ -0,0 +1,367 @@
...
+%% Common Test interface functions -----------------------------------
+%%--------------------------------------------------------------------
+all() -> 
+    [{group, tls1_2},
+     {group, tls1_3},
+     {group, revoked_1_2},
+     {group, revoked_1_3}].
+
+groups() -> 
+    [{tls1_2, [], tls1_2_tests()},
```
peterdmv, line 39: We could try to use the "standard" group names so that we get the right ssl version configured when calling init_per_group from ssl_test_lib.

lib/ssl/test/ssl_ocsp_SUITE.erl
```diff
@@ -0,0 +1,367 @@
...
+
+
+end_per_suite(Config) ->
+    ResponderPid = ?config(responder_pid, Config),
+    stop_ocsp_responder(ResponderPid),
+    ok = ssl:stop(),
+    application:stop(crypto).
+
+%%--------------------------------------------------------------------
+init_per_group(tls1_2, Config) ->
```
peterdmv, line 95: We could improve the ssl_test_lib to move some initialization there.

lib/ssl/test/ssl_ocsp_SUITE.erl
```diff
@@ -0,0 +1,367 @@
...
+init_per_suite(Config) ->
+    catch crypto:stop(),
+    try crypto:start() of
+	ok ->
+        ssl_test_lib:clean_start(),
+        DataDir = proplists:get_value(data_dir, Config),
+        PrivDir = proplists:get_value(priv_dir, Config),
+
+        %% Prepare certs
+        {ok, _} = make_certs:all(DataDir, PrivDir),
```
peterdmv, line 69: We should only generate one cert type.

lib/public_key/asn1/OCSP-2013-88.asn1
```diff
@@ -0,0 +1,148 @@
...
+id-pkix-ocsp-nonce           OBJECT IDENTIFIER ::= { id-pkix-ocsp 2 }
+id-pkix-ocsp-crl             OBJECT IDENTIFIER ::= { id-pkix-ocsp 3 }
+id-pkix-ocsp-response        OBJECT IDENTIFIER ::= { id-pkix-ocsp 4 }
+id-pkix-ocsp-nocheck         OBJECT IDENTIFIER ::= { id-pkix-ocsp 5 }
+id-pkix-ocsp-archive-cutoff  OBJECT IDENTIFIER ::= { id-pkix-ocsp 6 }
+id-pkix-ocsp-service-locator OBJECT IDENTIFIER ::= { id-pkix-ocsp 7 }
+id-pkix-ocsp-pref-sig-algs   OBJECT IDENTIFIER ::= { id-pkix-ocsp 8 }
+id-pkix-ocsp-extended-revoke OBJECT IDENTIFIER ::= { id-pkix-ocsp 9 }
+
+END
```
peterdmv, line 148: I put this comment only on this file, but it's valid for all modules. It seems to be safer to keep that last new line at the end of each module.

lib/ssl/src/ssl_connection.hrl
peterdmv: I think we will not need to store it, the result can be used directly when we verify the certificate chain.

lib/ssl/src/tls_connection.erl
```diff
@@ -669,21 +670,24 @@ hello(internal, #client_hello{extensions = Extensions} = Hello,
...
+                 ocsp_stapling := OcspStapling},
              handshake_env = HsEnv,
              start_or_recv_from = From} = State) ->
     %% RFC6066.8, If a server returns a "CertificateStatus" message,
     %% then the server MUST have included an extension of type
     %% "status_request" with empty "extension_data" in the extended
     %% server hello.
     OcspState = HsEnv#handshake_env.ocsp_stapling_state,
-    IsCertStatusRequested = maps:is_key(ocsp_stapling, Extensions),
+    IsCertStatusRequested = (undefined == maps:get(status_request, Extensions, false)),
```
peterdmv, line 683: At this point we have received the `server_hello` and we had sent out an OCSPRequest in ClientHello based on the value of the `ocsp_stapling` option. Next, we should check the status_request extension:
- if `ocsp_stapling`  is set to `false`, that is the client never sent out an OCSPRequest, then the presence of the status_request extension is a protocol error and we should abort the handshake. 
- if `ocsp_stapling` is set to `true`, then we should check if the `status_request` is empty (undefined). If `status_request` is empty we have successfully negotiated OCSP and we can set the `IsCertStatusRequested` (should be called `OcspNegotiated`) to true. If `status_request` not empty that it is a protocol error and we should abort the handshake.
I think this logic should go into a separate function with 2 input parameters: `OcspStapling` and `Extensions`:
```erlang
OcspNegotiated = handle_status_request_extension(OcspStapling, Extenions),
...
handle_status_request_extension(false = OcspStapling, Extensions) ->
    case maps:get(status_request, Extensions, false) of
        false -> %% extension is not present
            OcspStapling;
        _Else -> %% protocol error
           %% error handling here
    end;
handle_status_request_extension(true = OcspStapling, Extensions) ->
    case maps:get(status_request, Extensions, false) of
        undefined -> %% OCSP negotiated
            OcspStapling;
        false -> %% server won't negotiate OCSP
            false;
        _Else -> %% protocol error, status_request not empty
           %% error handling here
    end.
```
@IngelaAndin

lib/ssl/src/tls_connection.erl
```diff
@@ -669,21 +670,24 @@ hello(internal, #client_hello{extensions = Extensions} = Hello,
...
     {next_state, user_hello,
      State#state{start_or_recv_from = undefined,
                  handshake_env = HsEnv#handshake_env{
                      hello = Hello,
                      ocsp_stapling_state = OcspState#{
-                         is_server_hello_requested := IsCertStatusRequested}}},
+                         ocsp_negotiated :=
```
peterdmv, line 689: Following the above logic this can be simplified to
```erlang
ocsp_negotiated := OcspNegotiated,
```

lib/ssl/src/tls_connection.erl
```diff
@@ -741,24 +745,27 @@ hello(internal, #server_hello{extensions = Extensions} = Hello,
                  renegotiation = {Renegotiation, _},
                  ocsp_stapling_state = OcspState} = HsEnv,
 	     ssl_options = SslOptions} = State) ->
-    %% check if server has sent a certifucate status message in hello
-    IsCertStatusRequested = maps:is_key(ocsp_stapling, Extensions),
+    %% check if server has sent an empty certifucate status message in hello
+    IsCertStatusRequested = (undefined == maps:get(status_request, Extensions, false)),
```
peterdmv, line 749: Same comments here, we can reuse the new function to handle all cases.

lib/ssl/src/tls_connection.erl
```diff
@@ -741,24 +745,27 @@ hello(internal, #server_hello{extensions = Extensions} = Hello,
...
+    #{ocsp_stapling := OcspStapling} = SslOptions,
     case tls_handshake:hello(Hello, SslOptions, ConnectionStates0, Renegotiation) of
 	#alert{} = Alert -> %%TODO
 	    ssl_connection:handle_own_alert(Alert, ReqVersion, hello,
                                             State#state{connection_env =
                                                             CEnv#connection_env{negotiated_version = ReqVersion},
                                                         handshake_env = HsEnv#handshake_env{
                                                             ocsp_stapling_state = OcspState#{
-                                                                is_server_hello_requested := IsCertStatusRequested}}
+                                                                ocsp_negotiated :=
```
peterdmv, line 758: I think in this part we will not need to update the ocsp state as we are about to abort the handshake.

lib/ssl/src/tls_connection.erl
```diff
@@ -741,24 +745,27 @@ hello(internal, #server_hello{extensions = Extensions} = Hello,
...
+                                                                    OcspStapling andalso IsCertStatusRequested}}
                                                         });
         %% Legacy TLS 1.2 and older
 	{Version, NewId, ConnectionStates, ProtoExt, Protocol} ->
 	    ssl_connection:handle_session(Hello, 
 					  Version, NewId, ConnectionStates, ProtoExt, Protocol,
                       State#state{handshake_env = HsEnv#handshake_env{
                           ocsp_stapling_state = OcspState#{
-                              is_server_hello_requested := IsCertStatusRequested}}});
+                              ocsp_negotiated :=
```
peterdmv, line 767:
```erlang
ocsp_negotiated := OcspNegotiated
```

lib/ssl/src/tls_connection.erl
```diff
@@ -767,7 +774,8 @@ hello(internal, #server_hello{extensions = Extensions} = Hello,
                connection_env = CEnv#connection_env{negotiated_version = SelectedVersion},
                handshake_env = HsEnv#handshake_env{
                    ocsp_stapling_state = OcspState#{
-                       is_server_hello_requested := IsCertStatusRequested}}},
+                       ocsp_negotiated :=
```
peterdmv, line 777:
```erlang
ocsp_negotiated := OcspNegotiated
```

lib/ssl/src/ssl_connection.erl
ao-song: > So we have a new understanding of this. And we should actual postpone the path_validation to we have this info in the stapling case. And let save this until we prototped the new path_validation handling.

pending

lib/ssl/src/tls_connection.erl
```diff
@@ -580,24 +580,29 @@ init({call, From}, {start, Timeout},
...
+                        ocsp_nonce := OcspNonceOpt} = SslOpts,
+	                    session = NewSession,
+	                    connection_states = ConnectionStates0
+	    } = State0) ->
     KeyShare = maybe_generate_client_shares(SslOpts),
     Session = ssl_session:client_select_session({Host, Port, SslOpts}, Cache, CacheCb, NewSession),
     %% Update UseTicket in case of automatic session resumption
     {UseTicket, State1} = tls_handshake_1_3:maybe_automatic_session_resumption(State0),
     TicketData = tls_handshake_1_3:get_ticket_data(self(), SessionTickets, UseTicket),
+    OcspNonce = public_key:ocsp_nonce(OcspNonceOpt, OcspStaplingOpt),
```
peterdmv, line 598: I had a comment about to wrap this function, I can't find it anymore, but it was about not to leak the abstraction of the public_key API. `public_key:ocsp_nonce`should not be aware of any variables from applications using the public_key API. So `public_key:ocsp_nonce/0` should just return a DER encoded nonce, no parameters needed.

lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -0,0 +1,290 @@
...
+get_public_key(#'OTPCertificate'{tbsCertificate = TbsCert}) ->
+    PKInfo = TbsCert#'OTPTBSCertificate'.subjectPublicKeyInfo,
+    PKInfo#'OTPSubjectPublicKeyInfo'.subjectPublicKey.
+
+%%--------------------------------------------------------------------
+-spec get_nonce(boolean()) -> binary() | undefined.
+%%
+%% Description: Get an OCSP nonce
+%%--------------------------------------------------------------------
+get_nonce(false) ->
```
peterdmv, line 262: This should be `public_ocsp:get_nonce/0`.

lib/public_key/src/public_key.erl
```diff
@@ -1228,6 +1232,43 @@ pkix_test_data(#{} = Chain) ->
...
+%%--------------------------------------------------------------------
+ocsp_responder_id(Cert) ->
+    pubkey_ocsp:get_ocsp_responder_id(Cert).
+
+%%--------------------------------------------------------------------
+-spec ocsp_nonce(boolean(), boolean()) -> binary() | undefined.
+%%
+%% Description: Get an OCSP nonce
+%%--------------------------------------------------------------------
+ocsp_nonce(OcspNonce, true) ->
```
peterdmv, line 1267: This should be `public_key:ocsp_nonce/0`. See my other comment in ssl where you call this function!

lib/public_key/src/pubkey_cert.erl
```diff
@@ -33,7 +33,8 @@
 	 is_fixed_dh_cert/1, verify_data/1, verify_fun/4, 
 	 select_extension/2, match_name/3,
 	 extensions_list/1, cert_auth_key_id/1, time_str_2_gregorian_sec/1,
-         gen_test_certs/1, root_cert/2]).
```
peterdmv, line 36: I would remove this change as it only changes whitespaces in a non-related module.

lib/ssl/test/ssl_ocsp_SUITE.erl
```diff
@@ -0,0 +1,367 @@
...
+%%--------------------------------------------------------------------
+%% Test Cases --------------------------------------------------------
+%%--------------------------------------------------------------------
+
+ocsp_stapling_without_nonce_and_responder_certs_tls1_2() ->
+    [{doc, "Verify OCSP stapling works without nonce "
+           "and responder certs for tls1.2."}].
+ocsp_stapling_without_nonce_and_responder_certs_tls1_2(Config)
+  when is_list(Config) ->
+    Port = ?config(server_port, Config),
```
peterdmv, line 127: A general comment, you should use `proplists:get_value/2` instead of the "?config" macro to remove the dependency on the old test_server.

ao-song: done

lib/public_key/src/pubkey_cert.erl
ao-song: ok

lib/ssl/src/tls_connection.erl
ao-song: done

lib/public_key/src/pubkey_ocsp.erl
ao-song: done

lib/public_key/src/public_key.erl
ao-song: done

lib/ssl/src/tls_connection.erl
ao-song: done

lib/ssl/src/tls_connection.erl
ao-song: done

lib/ssl/src/tls_connection.erl
ao-song: done

lib/public_key/asn1/OCSP-2013-88.asn1
ao-song: done

lib/ssl/src/ssl_handshake.erl
ao-song: done

lib/ssl/test/ssl_ocsp_SUITE.erl
peterdmv: I set this to resolved as this is the only way today to generate revoked certificates.

lib/kernel/doc/src/application.xml
```diff
@@ -244,7 +244,7 @@ Nodes = [cp1@cave, {cp2@cave, cp3@cave}]</code>
       <desc>
         <p>Sets the configuration <c><anno>Config</anno></c> for multiple
           applications. It is equivalent to calling <c>set_env/4</c> on
-          each application individially, except it is more efficient.
+          each application individually, except it is more efficient.
```
peterdmv, line 247: This is the typo fix that is already merged to upstream/maint.

lib/public_key/src/public_key.erl
```diff
@@ -1262,6 +1280,56 @@ pkix_test_data(#{} = Chain) ->
...
     pubkey_cert:root_cert(Name, Opts).
 
+%%--------------------------------------------------------------------
+-spec ocsp_status(binary | #'Certificate'{} | #'OTPCertificate'{},
+                  list(), list()) ->
+    {good, term()} | {unknown, term()} | {revoked, term()} | no_matched_response.
+%%
+%% Description: Get Certificate status
+%%--------------------------------------------------------------------
+ocsp_status(Cert, CertPath, Responses) ->
```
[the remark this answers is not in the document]
ao-song: > I just mark here that my tracing shows that Cert is included in CertPath (it was the second element in the list when running ocsp_stapling_basic, TLS 1.2).

these two parameters was inherited from previous crl_check/5, isn't it be the first on in the cert list, why second?

lib/public_key/test/pubkey_ocsp_SUITE.erl
```diff
@@ -0,0 +1,369 @@
...
+
+%%--------------------------------------------------------------------
+%% Test Cases --------------------------------------------------------
+%%--------------------------------------------------------------------
+ocsp_test() ->
+    [{doc, "Test functions in pubkey_ocsp"}].
+ocsp_test(Config) when is_list(Config) ->
+    %% Feeding data
+    SingleResponseGood =
+        [{'SingleResponse',
```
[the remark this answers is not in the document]
ao-song: > A general comment to this module. We could use the record syntax instead of the tagged tuples. It is more future proof in case any of those records are changed.

done

lib/public_key/src/public_key.erl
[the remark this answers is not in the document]
IngelaAndin: Please wait to handle this. My refactor to use the state machine cleaner will probably make this problem obsolete.

ao-song:
An experimental OCSP client support has been merged.

PR #2630, opened 2020-05-13
Drop remnants of elib_malloc (removed in R15)

richcarl:
While working on https://github.com/erlang/otp/pull/2627, I found that this elib_malloc stuff was still in there even though the docs said it was to be removed in R15. See https://github.com/erlang/otp/blob/master/erts/doc/src/notes.xml#L16543

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -8269,7 +8269,7 @@ Metadata = #{ pid => pid(),
     </func>
 
     <func>
-      <name name="system_info" arity="1" clause_i="77" since=""/>
+      <name name="system_info" arity="1" clause_i="76" since=""/>
       <fsummary>System info overview.</fsummary>
       <desc>
         <p>Returns information about the current system.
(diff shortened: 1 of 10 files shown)


michalmuskala:
Is this related to `ERL_MALLOC_LIB` and `+MYm` emulator option that only support `libc`? If so, processing of that could probably be also removed.

https://github.com/erlang/otp/blob/9cab4f6d7b373babdfeedc9913a63162ba280d44/erts/etc/common/erlexec.c#L491-L505

garazdawi:
You also need to update the clause_i numbers for all the system_info clauses in erlang.xml

richcarl:
> You also need to update the clause_i numbers for all the system_info clauses in erlang.xml

Oh joy!

richcarl:
> Is this related to `ERL_MALLOC_LIB` and `+MYm` emulator option that only support `libc`? If so, processing of that could probably be also removed.

Yes, it's related, but the question is whether OTP want to keep the `+MYm` option around in case they want to try using it again. @garazdawi ?

garazdawi:
> Yes, it's related, but the question is whether OTP want to keep the +MYm option around in case they want to try using it again. @garazdawi ?

Please remove that as well.

garazdawi:
thanks!

PR #2635, opened 2020-05-17
User defined tls alert for selfsigned cert

sircinek:
This PR aims at giving user possibility to generate custom TLS alert for path validation errors.
Previously, no matter what user defined verify_fun returned, the alert was always Bad Certificate.
The changes allow user to control TLS alert in similar way that is done for certificates with valid paths.

### DIFF
--- a/lib/public_key/doc/src/public_key.xml
+++ b/lib/public_key/doc/src/public_key.xml
@@ -448,6 +448,10 @@ fun(OtpCert :: #'OTPCertificate'{},
 	verifying application-specific extensions. If called with an
 	extension unknown to the user application, the return value
 	<c>{unknown, UserState}</c> is to be used.</p>
+  <note><p>
+  Note that user defined custom verify_fun may alter original 
+  path validation error (e.g <c>selfsigned_peer</c>). Use with caution.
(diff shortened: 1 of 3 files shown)


IngelaAndin:
Thank you for your contribution. We think that the principal is sound, that is the return of the users verify_fun should be honored.  We have included the PR in our test runs.  We also want to ask you to
change the documentation <note> to  a  <warning>.

sircinek:
> Thank you for your contribution. We think that the principal is sound, that is the return of the users verify_fun should be honored. We have included the PR in our test runs. We also want to ask you to
> change the documentation to a .

How would you want me to change it? I was looking for right words to express the changed behavior and that important information might be lost along the way now.

IngelaAndin:
Argh, I should have code quoted. I edited my comment above!

sircinek:
> Argh, I should have code quoted. I edited my comment above!

Sure ! Pushed  requested changes.

IngelaAndin:
I rebased the change to maint as master is now for the upcoming OTP-24 release.  We will merge it once it has passed our test and we reviewed the new documentation (once it is built).

PR #2638, opened 2020-05-23
Haiku support

NattyNarwhal:
May need some additional integration work so it can be managed by Haiku's package manager, but this gets it building and running. I've done the same with Linux as well.

01aff7d is a workaround for ERL-1255.

### DIFF
--- a/erts/lib_src/pthread/ethread.c
+++ b/erts/lib_src/pthread/ethread.c
@@ -47,6 +47,10 @@
 
 #include <limits.h>
 
+#if defined (__HAIKU__)
+#include <os/kernel/OS.h>
+#endif
+
 #define ETHR_INLINE_FUNC_NAME_(X) X ## __
 #define ETHREAD_IMPL__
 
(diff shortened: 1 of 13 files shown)


erts/lib_src/pthread/ethread.c
```diff
@@ -337,7 +341,11 @@ ethr_thr_create(ethr_tid *tid, void * (*func)(void *), void *arg,
     twd.stacksize = 0;
 
     if (opts && opts->name) {
+#ifdef __HAIKU__
+        snprintf(twd.name_buff, B_OS_NAME_LENGTH, "%s", opts->name);
```
mikpe, line 345: This differs from the declared size of `.name_buff[]`.  IMO this (and all other `snprintf`:s you change) should use `sizeof` of the destination array instead of hard-coding its size.

rickard-green: Yes, I agree with @mikpe on the `sizeof` approach

rickard-green:
I've rebased and merged this into the `maint` branch. That is, it will be part of next maintenance patch package (OTP 23.1) instead of next release (OTP 24). I also fixed the issue mentioned above about `snprintf()` and destination buffers.

PR #2639, opened 2020-05-24
Fixup erl_lint:format_error/1 deprecation clause

starbelly:
In the case of deprecation warnings erl_lint:format_error/1 would include the word `use` while all `-deprecated/1` calls included the word `use` resulting in `use use` in the warning message.

I'd be happy to write a test but it didn't look like there were any tests concerned with this level of detail and assumed it's probably unwanted.

Before : 

```
src/mod.erl:line_no: Warning: crypto:block_encrypt/4 is deprecated and will be removed in OTP 24; use use crypto:crypto_one_time/5, crypto:crypto_one_time_aead/6,7 or crypto:crypto_(dyn_iv)?_init + crypto:crypto_(dyn_iv)?_update + crypto:crypto_final instead
```

After:

```
src/mod.erl:line_no:  Warning: crypto:block_encrypt/4 is deprecated and will be removed in OTP 24; use crypto:crypto_one_time/5, crypto:crypto_one_time_aead/6,7 or crypto:crypto_(dyn_iv)?_init + crypto:crypto_(dyn_iv)?_update + crypto:crypto_final instead
```

The other way to tackle this is to keep `use` in the `format_error/1` clause and remove `use` from all the string elements in the `-deprecated/1` clause.

### DIFF
--- a/lib/stdlib/src/erl_lint.erl
+++ b/lib/stdlib/src/erl_lint.erl
@@ -249,7 +249,7 @@ format_error({redefine_bif_import,{F,A}}) ->
     io_lib:format("import directive overrides auto-imported BIF ~w/~w~n"
 		  " - use \"-compile({no_auto_import,[~w/~w]}).\" to resolve name clash", [F,A,F,A]);
 format_error({deprecated, MFA, String, Rel}) ->
-    io_lib:format("~s is deprecated and will be removed in ~s; use ~s",
+    io_lib:format("~s is deprecated and will be removed in ~s; ~s",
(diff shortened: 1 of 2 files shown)


jhogberg:
Thanks for the PR! Can you rebase to `maint` and fix up the `-removed()` attributes for `erl_scan:attributes_info/_` and `erl_scan:token_info/_`?

starbelly:
@jhogberg Can do 👍

jhogberg:

PR #2641, opened 2020-05-27
Add socket:getprotoby*/1 functions

hauleth:
This allows to get the supported protocol raw numbers in platform-independent way. While user can use IANA protocol list I think that it is safer way, especially if some OS would support some experimental protocol.


erts/emulator/nifs/common/prim_socket_nif.c
```diff
@@ -16198,8 +16202,83 @@ ERL_NIF_TERM esock_cancel_mode_select(ErlNifEnv*       env,
...
+{
+    ERL_NIF_TERM name = MKS(env, entry->p_name);
+    ERL_NIF_TERM aliases = MKEL(env);
+
+    for (char *alias = entry->p_aliases; alias; ++alias) {
+        ERL_NIF_TERM item = MKS(env, alias);
+        aliases = enif_make_list_cell(env, item, aliases);
+    }
+
+    ERL_NIF_TERM protocol = MKI(env, entry->p_proto);
```
hauleth, line 16268: Question is whether it shouldn't be a tuple in form of `{raw, Num}` to make it simpler for use together with `socket:open/3`.

lib/kernel/src/socket.erl
```diff
@@ -2489,6 +2498,14 @@ flush_abort_msg(SockRef, Ref) ->
 %%
 %% ===========================================================================
 
+-spec getprotobyname(string()) -> protoent().
+getprotobyname(Name) ->
+    prim_socket:getprotobyname(Name).
```
hauleth, line 2503: Should we support binaries and atoms there as well as strings?

RaimoNiskanen:
Hello!

Sorry for the delay... I am working on the socket code and messing a bit with the API.  I have a branch in our daily builds with rewrites, that I should try to merge soonish...

I do not know when the {raw, Num} form is necessary, if ever, it could be obvious from the type (integer).

Also, maybe this functionality could be hidden from the user and built in to the open() call, when the protocol is a string or a binary.  I think both string and binary should be supported with strings being UTF-8 encoded and binaries in raw encoding.

RaimoNiskanen:
I think it would be nicer to not commit all unaffected preloaded .beam files.  Use "./otp_build update_preloaded --no-commit" and then manually commit only the .beam files you know must have been changed.  That lowers the odds for a binary merge conflict.

RaimoNiskanen:
Would hiding getprotobyname() and getprotobynumber() from the API be a good solution?  This could be implemented with two NIFs that are called from socket:open() so we do not need to clutter the NIF for open().

Or does getprotobyname() and getprotobynumber have other use cases than when opening a socket  and to translate when you get a number from socket:getopt(S, {socket, protocol}) ?

hauleth:
@RaimoNiskanen there could be build map of `#{ProtoName => ProtoEnt}` during the `socket` initialisation that would contain all protocols. In that case the `socket:open` would not need to handle `tcp`, etc. as a special cases and would always use `getprotobyname` (or cached value) when needed. It could work.

RaimoNiskanen:
@hauleth I have not gotten a good grip of this subject yet, so I am maybe thinking out loud here...

Does the /etc/protocols database, accessible through getprotobyname(), getprotobynumber(), getprotoent(), et.al define protocols only for the AF_INET and AF_INET6 protocol domains, and if so why is the protocols "ip" and "ipv6" on the list (are they for use through protocol type SOCK_RAW?)

Does the user have to know which combination of protocol type (SOCK_STREAM, SOCK_DGRAM, SOCK_RAW, and SOCK_SEQPACKET) that works with which protocol, or how are you supposed to know which arguments to give to socket(domain, type, protocol)?

Yes, building a database using getprotoent() during initialization of `socket` is a neat idea.  I am not sure how to present that data to the Erlang code, though.  An extra NIF call to look up the protocol number before calling `socket:open()` is a bit awkward.  Maybe a persistent term?

Do we have to cater for systems that lack getprotobyname()?  How common are they?  This is a POSIX function, right?  If so we can not rely on acquiring a lookup table during initialization, (or have to do a hard coded one from the important protocols).

Also, is your suggestion missing thread locking?  The getprotobyname() function is on Linux defined as not tread safe, and on BSD there are tread safe variants e.g getprotobyname_r().  We should have to handle both.

RaimoNiskanen:
@hauleth Also, is this database ever to change during a systems lifetime; can we have a one time snapshot that we never need to update, or does it have to be updated?

RaimoNiskanen:
@hauleth The question remains: should we provide a `socket:getprotobyname()` API function or should we only provide the possibility to use protocol names in `socket:open/3,4`?

hauleth:
@RaimoNiskanen if database changes then kernel probably needed to change as well. I do not know if we want to maintain that level of compatibility.

About API, I am not sure. It truly depends whether we want to be "as close to BSD sockets as possible" or have convenience API already cooked in. If the second, then I think that hiding `socket:getprotobyname/1` behind the `socket:open/{3,4}` would be good approach. If we want to stay "as close to original as possible" then I think it would be better to have the separate calls.

> Does the `/etc/protocols` database, accessible through `getprotobyname()`, `getprotobynumber()`, `getprotoent()`, et.al define protocols only for the `AF_INET` and `AF_INET6` protocol domains, and if so why is the protocols "ip" and "ipv6" on the list (are they for use through protocol type `SOCK_RAW`?)

No, it defines all protocols known by the kernel. For example on my macOS I have AX.25 listed there.

> Does the user have to know which combination of protocol type (`SOCK_STREAM`, `SOCK_DGRAM`, `SOCK_RAW`, and `SOCK_SEQPACKET`) that works with which protocol

Yes, this is up to user, especially as some protocols can be used with multiple protocol types (ex. ICMP can be used with `SOCK_RAW` and on some OSes with `SOCK_STREAM` for non-root ICMP sockets).

> Yes, building a database using `getprotoent()` during initialization of socket is a neat idea. I am not sure how to present that data to the Erlang code, though. An extra NIF call to look up the protocol number before calling `socket:open()` is a bit awkward. Maybe a persistent term?

I was thinking about map in persistent term as well.

> Do we have to cater for systems that lack getprotobyname()? How common are they? This is a POSIX function, right? If so we can not rely on acquiring a lookup table during initialization, (or have to do a hard coded one from the important protocols).

Do we support any OS that do not have. such function? I have checked and both [`getprotobyname`](https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-getprotobyname) and [`getprotobynumber`](https://docs.microsoft.com/pl-pl/windows/win32/api/winsock/nf-winsock-getprotobynumber) exists on Windows.

I have also checked and there are man pages for:

- [HP-UX 11.22](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=HP-UX+11.22&arch=default&format=html)
- [NetBSD 1.0](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=NetBSD+1.0&arch=default&format=html)
- [OpenBSD 2.0](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=OpenBSD+2.0&arch=default&format=html)
- [SunOS 4.1.3](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=SunOS+4.1.3&arch=default&format=html)
- [ULTRIX 4.2](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=ULTRIX+4.2&arch=default&format=html)
- [FreeBSD 1.0](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=FreeBSD+1.0-RELEASE&arch=default&format=html)
- [386BSD 0.0](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=386BSD+0.0&arch=default&format=html)
- [Darwin 7.0.1](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=Darwin+7.0.1&arch=default&format=html)
- [Debian 2.2.7](https://manpage.me/index.cgi?apropos=0&q=getprotobyname&sektion=0&manpath=Debian+2.2.7&arch=default&format=html)

Platforms that I couldn't find whether they support such:

- Plan 9
- Minix
- Unix Seventh Edition
- "Classic" BSD
- MACH
- HP-UX older than 11.22

I also couldn't find the list of supported OSes by Erlang, so I cannot check whether above list is enough or not.

RaimoNiskanen:
> About API, I am not sure. It truly depends whether we want to be "as close to BSD sockets as possible" or have convenience API already cooked in. If the second, then I think that hiding socket:getprotobyname/1 behind the socket:open/{3,4} would be good approach. If we want to stay "as close to original as possible" then I think it would be better to have the separate calls.

We want to stay close to the original, but in this case, since we do not want to use include files and can not have platform dependent include files; we can not mimic the C APIs property of using a symbol from system header files in compile time.  So a more Erlang:ish way would, I think, to do it in runtime if possible.  We have also added the possibility to wait for a message to perform non-blocking send, which also is an Erlang:ism.  Therefore I think we would like this feature in the `socket:open/{3,4}` call.

New detail to decide on is if we should do a lookup only for strings and binaries, or for all atoms, and then if to do it through a loaded /etc/protocols table or only for unknown protocols with getprotobyname() per call...

> I was thinking about map in persistent term as well.

There should be two maps: binary name to number and reverse, to be able to translate in `socket:getopt(S, {socket, protocol})`.

### Protocols
So the protocol number is a namespace of its own, and the 'domain' and 'type' arguments are more like qualifiers to the protocol number.  Except that you may supply only 'domain' and 'type' and still get a default protocol.  Not very clean in my ears...

Nevertheless.  The protocol number is then the only one that makes sense to look up.  'family' and 'type' are known and fixed small sets. Right?

### OS:es
The OS lists you present looks reassuring.  I think we do not have a list of supported OS:es, only a list of what we run in our daily builds, and a list of what we have heard works.  Anyway, there is always someone doing esoteric stuff, and it would be harsh to shut them out.  So we better have a fallback for systems without getprotobyname() and friends, at least not systemize it impossible.

RaimoNiskanen:
I have just merged a rather big rewrite (into master) of chiefly the socket options but also some about socket open, so PRs touching the socket functionality should be based on that master or later.

Here's my plan:

##### Next steps
Not necessarily in this order.
* Add a `protocols` query probably to `nif_supports()`, that will loop over getprotoent() while holding a lock to protect it from itself (not strictly necessary since this NIF will only be called from `prim_socket:on_load/1`, but for good measure and as a warning sign) and return a list of `{Names :: [binary()], Number: integer()}` entries.  The calling Erlang code will then create the lookup map(s) and store them as persistent term(s).  The returned list from the NIF will contain prepended or appended predefined entries for known protocols to enable systems without getprotoent(); should such exist.
* Rewrite `nif_open()` to take atoms for `domain` and `type` and remove the encoding/decoding in `prim_socket.erl`.  I saw that there are already functions in `prim_socket_nif.c` that does such encoding/decoding down there, for socket options, so why not for `open()`, since the value sets are so small.
* Rewrite `nif_open()` to only take a native protocol number as `protocol`, and rewrite `socket:open()` to use the lookup table.

A future possibility emerges, and that is to return native option numbers with `nif_supports(options)` and do option translation in the same way...

Maybe I should do this work on a new PR, and close this one?

Unfortunately (not in every way :-) I have scheduled vacation for myself next week, so there will not be much progress in a while.

RaimoNiskanen:
I am working on this idea, and will eventually publish a different PR, therefore closing this one.

RaimoNiskanen:
@hauleth: I have created #2670 that contains ongoing work on this feature

PR #2648, opened 2020-06-04
Fix small typo in using_ssl.xml

aronisstav:
The return value does not match the pattern on the preceding line.


IngelaAndin:
Thank you for fixing this.  Would you mind moving your fix to maint  so that we do not need to wait for OTP-24 to correct the documentation. I tried to move it, but it probably needs a rebase --onto

aronisstav:
Oops, missed this message! Unfortunately I did the patch via the online editor; could you try to cherry-pick it to `maint`?

IngelaAndin:
I did a rebase onto of your branch and will merge it via PR-2660

PR #2650, opened 2020-06-08
Initial check in.

shijiang1130:
To windows Active Domain,  get user last password attribute is common task, this simple code guide you how to implement via erlang way.

### DIFF
--- a/lib/eldap/example/ad_expiring_account.erl
+++ b/lib/eldap/example/ad_expiring_account.erl
@@ -1,0 +1,50 @@
+%%
+%% %CopyrightBegin%
+%%
+%% Copyright Ericsson AB 1996-2018. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


HansN:
I'm not familiar with Windows Active Directory (is it the same as Active Domain?) at all, but a quick search gives a couple of sources (for example [this](https://theitbros.com/ldap-query-examples-active-directory/) one), using a search pattern like

   `(objectCategory=person)(objectClass=user)(pwdLastSet=0) ... `

I'm just curios, is there some reason behind you not using that one?

A second question is, have you run your example towards a real AD server with good result?

shijiang1130:
I'm just curios, is there some reason behind you not using that one?
It because of i don't found useful example in the erlang itself,  look for internet take more than hours got wanted. just to see can save some time to new jump the language lover.

A second question is, have you run your example towards a real AD server with good result?
That is true, it test and can got the result. but for now i don't know how to ultilize by eldap filter sytax : (objectCategory=person)(objectClass=user)(pwdLastSet=0)

HansN:
Thanks for the PR!

PR #1968, opened 2018-09-29
support for FTP over SSL

jimdigriz:
https://bugs.erlang.org/browse/ERL-740

PR #2652, opened 2020-06-09
contributing: each commit should only change one application

jimdigriz:
Discovered during putting together PR #1968, the maintainers like each commit to only make changes to a single application.  This PR just notes this in CONTRIBUTING.md.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,9 @@ compiled and that it works.
 * Check for unnecessary whitespace before committing with `git diff --check`.
 However, do not fix preexisting whitespace errors in otherwise untouched source lines.
 
+* Each commit should only change one application.
+** If you need to change (for example) both the `ftp` and `ssl` applications, a single commit should not change both.
+
 Check your coding style:
 
 * Make sure your changes follow the coding and indentation style of the code surrounding your changes.


IngelaAndin:
Thank you. Could you please also base it on maint so that we can get this improvement in place as soon as possible and not have it wait for OTP-24 release. I tried the GUI change base but i did not work, maybe it has to be a `rebase --onto`

jimdigriz:
Sure, done.

IngelaAndin:
So after some internal comments we want to suggest an alternative change. That would make the following section read:

`* To make it possible to use the powerful `git bisect` command, make sure that each commit can be
compiled and that it works. After fulfilling this requirement it will facilitate code reviews if it is possible to make separate commits for different OTP applications and logical steps in the PR.`

This diff introduces the change

`diff --git a/CONTRIBUTING.md b/CONTRIBUTING.md
index b754fbd662..6327e97367 100644
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -110,7 +110,8 @@ Make sure that your branch contains clean commits:
 conflicts or include the latest changes.
 
 * To make it possible to use the powerful `git bisect` command, make sure that each commit can be
-compiled and that it works.
+compiled and that it works. After fulfilling this requirement it will facilitate code reviews if it is possible to make
+sepearate commits for diffrent OTP applications and logical steps in the PR.
 
 * Check for unnecessary whitespace before committing with `git diff --check`.
 However, do not fix preexisting whitespace errors in otherwise untouched source lines.

`

jimdigriz:
This probably would be easier if you and your team adjust the working and commit as using me as a remote keyboard seems a poor use of both our time and causes an unnecessary delay?

IngelaAndin:
We wanted to make sure the intent of your change is still valid, just that we discovered that there might be some exceptions so we want to make it clearer for all cases. But you have a valid point , so I will make a PR from OTP instead and you can comment if you feel something is not clear enough.

jimdigriz:
Not a problem.

As this is a non-technical 'word smithing'/documentation patch, the wording I suggested was *never* going to make the cut :)

As long as it helps others avoid the same problem I tripped over, I really am not opinionated on this.

PR #2654, opened 2020-06-10
inet_tls_dist: make TLS handshake concurrent

max-au:
TLS handshake may take considerable amount of time, so it should
be done in a different process, not accepting process itself.
Also, there is no need to kill the accepting process if TLS
handshake fails.
Handshake process keeps max priority that original implementation had.

### DIFF
--- a/lib/ssl/src/inet_tls_dist.erl
+++ b/lib/ssl/src/inet_tls_dist.erl
@@ -223,27 +223,62 @@ gen_accept(Driver, Listen) ->
     monitor_pid(
       spawn_opt(
         fun () ->
-                accept_loop(Driver, Listen, Kernel)
+            process_flag(trap_exit, true),
+            LOpts = application:get_env(kernel, inet_dist_listen_options, []),
+            MaxPending =
+                case lists:keyfind(backlog, 1, LOpts) of
+                    {backlog, Backlog} -> Backlog;
+                    false -> 128
+                end,
+            DLK = {Driver, Listen, Kernel},
+            accept_loop(DLK, spawn_accept(DLK), MaxPending, #{})
         end,
         [link, {priority, max}])).
 
-accept_loop(Driver, Listen, Kernel) ->
-    case Driver:accept(Listen) of
-        {ok, Socket} ->
-	    case check_ip(Driver, Socket) of
-                true ->
-                    accept_loop(Driver, Listen, Kernel, Socket);
-                {false,IP} ->
-		    ?LOG_ERROR(
-                      "** Connection attempt from "
-                      "disallowed IP ~w ** ~n", [IP]),
-		    ?shutdown2(no_node, trace({disallowed, IP}))
-	    end;
-	Error ->
-	    exit(trace(Error))
+%% Concurrent accept loop will spawn a new HandshakePid when
+%%  there is no HandshakePid already running, and Pending map is
+%%  smaller than MaxPending
+accept_loop(DLK, undefined, MaxPending, Pending) when map_size(Pending) < MaxPending ->
+    accept_loop(DLK, spawn_accept(DLK), MaxPending, Pending);
+accept_loop(DLK, HandshakePid, MaxPending, Pending) ->
+    receive
+        {continue, HandshakePid} when is_pid(HandshakePid) ->
+            accept_loop(DLK, undefined, MaxPending, Pending#{HandshakePid => true});
+        {'EXIT', Pid, Reason} when is_map_key(Pid, Pending) ->
+            Reason =/= normal andalso
+                ?LOG_ERROR("TLS distribution handshake failed: ~p~n", [Reason]),
+            accept_loop(DLK, HandshakePid, MaxPending, maps:remove(Pid, Pending));
+        {'EXIT', HandshakePid, Reason} when is_pid(HandshakePid) ->
+            %% HandshakePid crashed before turning into Pending, which means
+            %%  error happened in accept. Need to restart the listener.
+            exit(Reason);
+        Unexpected ->
+            ?LOG_WARNING("TLS distribution: unexpected message: ~p~n" ,[Unexpected]),
+            accept_loop(DLK, HandshakePid, MaxPending, Pending)
     end.
 
-accept_loop(Driver, Listen, Kernel, Socket) ->
+spawn_accept({Driver, Listen, Kernel}) ->
+    AcceptLoop = self(),
+    spawn_link(
+        fun () ->
+            case Driver:accept(Listen) of
+                {ok, Socket} ->
+                    AcceptLoop ! {continue, self()},
+                    case check_ip(Driver, Socket) of
+                        true ->
+                            accept_one(Driver, Kernel, Socket);
+                        {false,IP} ->
+                            ?LOG_ERROR(
+                                "** Connection attempt from "
+                                "disallowed IP ~w ** ~n", [IP]),
+                            trace({disallowed, IP})
+                    end;
+                Error ->
+                    exit(Error)
+            end
+        end).
+
+accept_one(Driver, Kernel, Socket) ->
     Opts = setup_verify_client(Socket, get_ssl_options(server)),
     wait_for_code_server(),
     case
@@ -259,13 +294,18 @@ accept_loop(Driver, Listen, Kernel, Socket) ->
                    Driver:family(), tls}),
             receive
                 {Kernel, controller, Pid} ->
-                    ok = ssl:controlling_process(SslSocket, Pid),
-                    trace(
-                      Pid ! {self(), controller});
+                    case ssl:controlling_process(SslSocket, Pid) of
+                        ok ->
+                            trace(Pid ! {self(), controller});
+                        Error ->
+                            trace(Pid ! {self(), exit}),
+                            ?LOG_ERROR(
+                                "Cannot control TLS distribution connection: ~p~n",
+                                [Error])
+                    end;
                 {Kernel, unsupported_protocol} ->
-                    exit(trace(unsupported_protocol))
-            end,
-            accept_loop(Driver, Listen, Kernel);
+                    trace(unsupported_protocol)
+            end;
         {error, {options, _}} = Error ->
             %% Bad options: that's probably our fault.
             %% Let's log that.
@@ -273,10 +313,10 @@ accept_loop(Driver, Listen, Kernel, Socket) ->
               "Cannot accept TLS distribution connection: ~s~n",
               [ssl:format_error(Error)]),
             gen_tcp:close(Socket),
-            exit(trace(Error));
+            trace(Error);
         Other ->
             gen_tcp:close(Socket),
-            exit(trace(Other))
+            trace(Other)
     end.
 
 
@@ -412,9 +452,9 @@ gen_accept_connection(
 
 do_accept(
   _Driver, AcceptPid, DistCtrl, MyNode, Allowed, SetupTime, Kernel) ->
-    {ok, SslSocket} = tls_sender:dist_tls_socket(DistCtrl),
     receive
 	{AcceptPid, controller} ->
+            {ok, SslSocket} = tls_sender:dist_tls_socket(DistCtrl),
 	    Timer = dist_util:start_timer(SetupTime),
             NewAllowed = allowed_nodes(SslSocket, Allowed),
             HSData0 = hs_data_common(SslSocket),
@@ -427,7 +467,11 @@ do_accept(
                   this_flags = 0,
                   allowed = NewAllowed},
             link(DistCtrl),
-            dist_util:handshake_other_started(trace(HSData))
+            dist_util:handshake_other_started(trace(HSData));
+        {AcceptPid, exit} ->
+            %% this can happen when connection was initiated, but dropped
+            %%  between TLS handshake completion and dist handshake start
+            ?shutdown2(MyNode, connection_setup_failed)
     end.
 
 allowed_nodes(_SslSocket, []) ->
--- a/lib/ssl/test/ssl_dist_SUITE.erl
+++ b/lib/ssl/test/ssl_dist_SUITE.erl
@@ -39,6 +39,8 @@
          basic/1,
          payload/0,
          payload/1,
+         dist_port_overload/0,
+         dist_port_overload/1,
          plain_options/0,
          plain_options/1,
          plain_verify_options/0,
@@ -91,6 +93,7 @@ start_ssl_node_name(Name, Args) ->
 all() ->
     [basic,
      payload,
+     dist_port_overload,
      plain_options,
      plain_verify_options,
      nodelay_option,
@@ -163,6 +166,61 @@ payload() ->
     [{doc,"Test that send a lot of data between the ssl distributed nodes"}].
 payload(Config) when is_list(Config) ->
     gen_dist_test(payload_test, Config).
+
+%%--------------------------------------------------------------------
+dist_port_overload() ->
+    [{doc, "Test that TLS distribution connections can be accepted concurrently"}].
+dist_port_overload(Config) when is_list(Config) ->
+    %% Start a node, and get the port number it's listening on.
+    #node_handle{nodename = NodeName} = NH1 = start_ssl_node(Config),
+    [Name, Host] = string:lexemes(atom_to_list(NodeName), "@"),
+    {ok, NodesPorts} = apply_on_ssl_node(NH1, fun net_adm:names/0),
+    {Name, Port} = lists:keyfind(Name, 1, NodesPorts),
+    %% Run 4 connections concurrently. When TLS handshake is not concurrent,
+    %%  and with default net_setuptime of 7 seconds, only one connection per 7
+    %%  seconds is closed from server side. With concurrent accept, all 7 will
+    %%  be dropped in 7 seconds
+    RequiredConcurrency = 4,
+    Started = [connect(self(), Host, Port) || _ <- lists:seq(1, RequiredConcurrency)],
+    %% give 10 seconds (more than 7, less than 2x7 seconds)
+    Responded = barrier(RequiredConcurrency, [], erlang:system_time(millisecond) + 10000),
+    %% clean up
+    stop_ssl_node(NH1),
+    [R ! exit || R <- Responded],
+    [exit(P, kill) || P <- Started -- Responded],
+    %% Ensure some amount of concurrency was reached.
+    (length(Responded) >= RequiredConcurrency) orelse
+        ct:fail({actual, length(Responded), expected, RequiredConcurrency}),
+    success(Config).
+
+barrier(0, Responded, _Until) ->
+    Responded;
+barrier(RequiredConcurrency, Responded, Until) ->
+    Timeout = Until - erlang:system_time(millisecond),
+    receive
+        {waiting, Pid} ->
+            barrier(RequiredConcurrency - 1, [Pid | Responded], Until);
+        {error, Error} ->
+            ct:fail(Error)
+    after
+        Timeout -> Responded
+    end.
+
+connect(Control, Host, Port) ->
+    spawn(
+        fun () ->
+            case gen_tcp:connect(Host, Port, [{active, true}]) of
+                {ok, Sock} ->
+                    receive
+                        {tcp_closed, Sock} ->
+                            Control ! {waiting, self()};
+                        exit ->
+                            gen_tcp:close(Sock)
+                    end;
+                Error ->
+                    Control ! {error, Error}
+            end
+        end).
 
 %%--------------------------------------------------------------------
 plain_options() ->


IngelaAndin:
Thank you for your contribution. We will start by testing it. Review might take a little longer as it is vacation time, please have patience.

RaimoNiskanen:
Creative!

I would like to bounce this against the right VM guys, but they/he/she are on vacation another 3 weeks, just in case.

Other than that I think that `check_ip()` should be part of what is done before spawning.

Also, I am worried about the lack of flow control.  Many attaching nodes might get past `accept()` and start parallel TLS handshakes on high priority.  Before the change there could be only one at the time, which was the problem this pull request is solving, but now the node might get drowned in CPU intensive work.

And, I am not certain about the link structure between the processes. Now the Accept process (known to `net_kernel`, looping in `accept_loop()` gets linked to the spawned processes calling `accept()`, and the Accept process dies if `check_ip()` fails which could kill a previous ongoing handshake.  So maybe the Accept process should not die but just log and loop, or maybe these two processes should not be linked, or rather unlink after `{continue,self()}` has been sent.  But then we have a free running process doing the handshake.  Maybe just fine.  Something for the VM guys...

max-au:
> check_ip() should be part of what is done before spawning

I'm not entirely sure how check_ip() is supposed to work, this seems to be undocumented. My reading of source code suggests that it rejects connections attempts made from a different network (IP addresses are masked and compared). So it looks like a firewall of a sort, should this functionality be expanded? I'd actually love some kind of an accept-level firewall, probably even a whitelist of IP addresses that are allowed to connect.

> Many attaching nodes might get past accept() and start parallel TLS handshakes on high priority

This was the part I was also concerned about. My general concern is that accepting process has 'max' priority, and if we keep it this way, it may create more and more of 'max' priority connections. However it replicates existing approach with TCP connections. I might suggest lowering priority of the accepting process back to normal (and therefore handshake processes are also normal). However this will be a change from existing schema, when all dist processes run with max priority (including acceptor). I may not have enough context to understand the implications.

I ran this patch with 17k TLS cluster, and while it works, it surely creates a possibility of self-inflicted DoS with standard setup (7 seconds for any network operation during dist connection setup). When there are 17k nodes attempting to connect, it creates rather long accept queue. Handshake processes run with max priority, however they are stuck waiting for tls_sender processes (which seems to be running on normal).

> Accept process dies if check_ip() fails which could kill a previous ongoing handshake

This appears to be a bug. The intent was to loop Accept process infinitely, unless it receives shutdown signal, or accepting socket closes. I did not notice ?shutdown2() in check_ip (probably because we aren't using check_ip, and there is no test for the combination with check_ip set to true and connection originating outside of local network). I'll make changes, thank you!

max-au:
I managed to re-work this patch. It became larger, as it also fixes a number of issues uncovered testing on large (>15k nodes) clusters.

Some notes:
1. Now, accept_loop maintains a list of concurrently running handshake processes, limiting concurrency to {backlog, B} - taken from kernel configuration parameter inet_dist_listen_options. Default is 128 (same as for backlog).
2. Spawned handshake processes are no longer crashing the listener. Before, any abnormally exiting handshake process caused listener to exit. For example, "connection attempt from a disallowed IP" was also closing listening socket (and losing everything accumulated in socket backlog).
3. There was a race condition. After handshake but before ssl:controlling_process call, socket could be closed by peer, that was causing listener to exit (losing socket backlog). This commit prevents listener restart.

Side note.
I also considered an opportunity to move TLS handshake into do_accept/7, next to "dist_util:start_timer(SetupTime)". It seems to be a more logical solution.
However it would not let me use tls_connection process as DistCtrl (because tls_connection process is spawned during handshake), and also would not allow to limit concurrency (unless I patch net_kernel as well).

@RaimoNiskanen, @IngelaAndin - I'd love to listen to your feedback, whether this change can be improved further. I tested it on some large clusters, so was able to catch all mentioned cases, but I might've missed some other considerations.

IngelaAndin:
We hope to have some feedback soon for this PR too. Just this week things are getting back to more normal after summer vacation. Thank you for you patience.

IngelaAndin:
@max-au Initially we think this is looking good so I have put it into testing now. As it is a complicated part of the code we still need some time for a deeper review.

RaimoNiskanen:
These are gory internals, so I am not entirely sure about what is "right", for instance about max priority...

We had a little internal chat about this, and managed to recall that distribution connections that dropped during high load, and connection setups timing out, are problems that max priority aims to mitigate.

The question is now how to handle encrypted communication?  It is special since it is much heavier on the CPU, with encryption setup very much heavier than bulk data encryption.

We may have to re-evaluate how to set process priority for encrypted communication.  To have so many processes burn so much CPU may starve the processes doing real work, to the extent that they can not keep up with the communication.  One possibility may be to have asymmetrical priorities i.e high for outgoing encryption and normal for incoming, if that should work good in most cases.

#### Regarding the PR

It seems the commit comment "Handshake process runs with normal priority to avoid node overload" is incorrect.  For starters, accept_loop/5 spawns HandshakePid with max priority, and furthermore in the SSL code (currently) the handshake is done with max priority thanks to the undocumented flag {erl_dist,true}.

I wonder if the backlog value is a good parameter for the number of parallel TLS handshake processes?  If you run more processes than the number of schedulers through ssl:handshake in parallel they will compete for CPU and run slower, which may cause timeouts.  So may the number of schedulers be a better parameter for the number of parallel accepts?

accept_loop exits if it receives {'EXIT', Pid, Reason}, from any process.  Maybe check that Pid is in Pending.  And would not accept_loop now exit for any crash in check_ip or accept_one and bring down all processes still are in ssl_handshake; is that intentional?  Sounds sensitive to me, but I may have missed something.  If the intention is to exit for errors from Driver:accept, then maybe send an explicit e.g {accept_error, self()} back to accept_loop as for {continue, self()} and exit due to that, and ignore the exit reason of the link (maybe log/warn for non-normal).

IngelaAndin:
At the moment we also see this error during test. Only occasionally.  I will remove the testing label for now. Until
we figure out how we want this.  

*** CT Error Notification 2020-08-29 12:06:35.606 ***
erlang:port_close failed
Reason: {badarg,[{erlang,...},{...}|...]}

=== Ended at 2020-08-29 12:06:35
=== Location: [{erlang,port_close},
 {ssl_dist_test_lib,start_ssl_node,105},
 {ssl_dist_SUITE,gen_dist_test,420},
 {test_server,ts_tc,1754},
 {test_server,run_test_case_eval1,1263},
 {test_server,run_test_case_eval,1195}]
=== === Reason: bad argument
  in function  port_close/1
     called as port_close(#Port<0.6473>)
  in call from ssl_dist_test_lib:start_ssl_node/2 (ssl_dist_test_lib.erl, line 105)
  in call from ssl_dist_SUITE:gen_dist_test/2 (ssl_dist_SUITE.erl, line 420)
  in call from test_server:ts_tc/3 (test_server.erl, line 1754)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1263)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 1195)

max-au:
@RaimoNiskanen thank you review & historical context. I understand there is no simple answer around communications setup, and clearly see how to implement DoS using distribution over TLS listening socket.
So I added a test case reproducing the behaviour, and proving there is some level of protection/concurrency added within this PR.

> and furthermore in the SSL code (currently) the handshake is done with max priority thanks to the undocumented flag {erl_dist,true}.

Yes, I also saw this code, and decided that it should be addresses separately.

> I wonder if the backlog value is a good parameter for the number of parallel TLS handshake processes? If you run more processes than the number of schedulers through ssl:handshake in parallel they will compete for CPU and run slower, which may cause timeouts. So may the number of schedulers be a better parameter for the number of parallel accepts?

Number of concurrent TLS handshake processes depends on expected TCP connection latency. In most cases TLS is required for long-distance communications, crossing single datacenter boundary and traveling over public networks. Limiting concurrency to a number of schedulers dramatically increases amount of time needed to connect to a large number of nodes. Napkin calculation: with 50 ms latency, 3 round-trips to establish a connection and 16 schedulers accepting side is limited to 1000/150*16 ~ 100 connections per second. Given 7 seconds default timeout for a distribution connection, it makes at most 700 nodes in a cluster.

In my experience backlog is often used as a proxy for amount of connections expected - the more expected, the larger backlog is. Same applies to accept concurrency. Hence my attempt to piggy-back on the backlog setting. However I can surely make it a different configuration parameter of ssl application.

> accept_loop exits if it receives {'EXIT', Pid, Reason}, from any process. Maybe check that Pid is in Pending. And would not accept_loop now exit for any crash in check_ip or accept_one and bring down all processes still are in ssl_handshake; is that intentional?

It is another hard choice. I removed "exit" for all expected errors (e.g. failure to execute ssl:controlling_process, or unsupported protocol). However I am not sure whether it would be correct to suppress all unexpected crashes of handshake processes. The reason is, we may completely lose visibility here, and listening socket would be never reopen. There may be a better way to handle it, and I'd love to find it with help of OTP team.


@IngelaAndin how do I run the test you mention? In most cases I'm using either `ts:run(ssl, ssl_dist_SUITE, [batch]).` or `ts:run(ssl, [batch]).`, but I'm not seeing output you provided on either Linux, or Mac OS. Is there a specific OS/version I need to run?

IngelaAndin:
@max-au  the particular fault I see happens on an OpenBSD machine. I have not had time to analyze it further I could perhaps be a test case error. But it has not happened before I added your branch. I am currently looking to confirm that it goes away when your branch is not included.  Of course we want to be as certain as possible we are not introducing an new problems. But until we are sure the code is like we want it I would not spend to much time on chasing this.

IngelaAndin:
@max-au @RaimoNiskanen  On the subject of  unexpected crashes of handshake processes I do not think there will be many as they are designed to always go down gracefully (using try ...catch) sending a TLS-ALERT as that is what the protocol expects. This of course is not always optimal during development and debugging. But I have tried to compensate with good logger messages for debugging and during development I can of course have temporary code doing what I want.

RaimoNiskanen:
@max-au 
> Yes, I also saw this code, and decided that it should be addresses separately.

Agreed!

> Napkin calculation: with 50 ms latency, 3 round-trips to establish a connection and 16 schedulers accepting side is limited to 1000/150*16 ~ 100 connections per second. Given 7 seconds default timeout for a distribution connection, it makes at most 700 nodes in a cluster.

I'd say that is an optimistic calculation that ignores CPU load, whilst looking only at the number of schedulers is a pessimistic number that ignores that network latency significantly decreases CPU load.

A good default should lie somewhere in between, but I think your parameter (the backlog value) is the better in practice.
Maybe the geometric mean of both values could be used, or a scaled geometric mean.  But that gets kind of convoluted.

It would be nice to avoid an explicit configuration item for this, so why not start with the backlog value, as you suggest, and refine it later if needed.

> The reason is, we may completely lose visibility here, and listening socket would be never reopen. There may be a better way to handle it, and I'd love to find it with help of OTP team.

In the current proposal, an error from Driver:accept/1 will cause the accept_loop/4 to exit, which will bring down all concurrent connection attempts in progress, and net_kernel will restart a new acceptor process on the same listen socket.  This is probably just fine since it should be a rare error, and since (I think) if net_kernel wants to shut down the distribution it kills the acceptor - accept_loop/4, which _should_ take down all processes handling connection setup.

I'll write a few nitpicks on the code as well...

lib/ssl/src/inet_tls_dist.erl
```diff
@@ -215,27 +215,53 @@ gen_accept(Driver, Listen) ->
...
-        {ok, Socket} ->
-	    case check_ip(Driver, Socket) of
+%% Concurrent accept loop will spawn a new HandshakePid when
+%%  there is no HandshakePid already running, and Pending map is
+%%  smaller than MaxPending
+accept_loop(DLK, undefined, MaxPending, Pending) when map_size(Pending) < MaxPending ->
+    accept_loop(DLK, spawn_accept(DLK), MaxPending, Pending);
+accept_loop(DLK, HandshakePid, MaxPending, Pending) ->
+    receive
+        {continue, HandshakePid} ->
```
RaimoNiskanen, line 237: A guard here for when is_pid(HandshakePid) would be nice to make the possibility that the process should receive {continue, undefined} impossible

lib/ssl/src/inet_tls_dist.erl
```diff
@@ -215,27 +215,53 @@ gen_accept(Driver, Listen) ->
...
+            accept_loop(DLK, undefined, MaxPending, Pending#{HandshakePid => true});
+        {'EXIT', Pid, _} when is_map_key(Pid, Pending) ->
+            accept_loop(DLK, HandshakePid, MaxPending, maps:remove(Pid, Pending));
+        {'EXIT', HandshakePid, Reason} ->
```
RaimoNiskanen, line 241: when is_pid(HandshakePid)

RaimoNiskanen:
...that net_kernel is silent when restarting an acceptor is a different problem.  I think it should log something...

lib/ssl/src/inet_tls_dist.erl
```diff
@@ -215,27 +215,53 @@ gen_accept(Driver, Listen) ->
...
+%%  there is no HandshakePid already running, and Pending map is
+%%  smaller than MaxPending
+accept_loop(DLK, undefined, MaxPending, Pending) when map_size(Pending) < MaxPending ->
+    accept_loop(DLK, spawn_accept(DLK), MaxPending, Pending);
+accept_loop(DLK, HandshakePid, MaxPending, Pending) ->
+    receive
+        {continue, HandshakePid} ->
+            accept_loop(DLK, undefined, MaxPending, Pending#{HandshakePid => true});
+        {'EXIT', Pid, _} when is_map_key(Pid, Pending) ->
+            accept_loop(DLK, HandshakePid, MaxPending, maps:remove(Pid, Pending));
```
RaimoNiskanen, line 240: Maybe check for Reason ( _) =/= normal and log something in that case

lib/ssl/src/inet_tls_dist.erl
```diff
@@ -215,27 +215,53 @@ gen_accept(Driver, Listen) ->
...
+            %% HandshakePid crashed before turning into Pending, which means
+            %%  error happened in accept. Need to restart the listener.
+            exit(Reason)
+    end.
+
+spawn_accept({Driver, Listen, Kernel}) ->
+    AcceptLoop = self(),
+    spawn_link(
+        fun () ->
+            {ok, Socket} = Driver:accept(Listen),
```
RaimoNiskanen, line 251: inet_tcp_dist exits with Error = Driver:accept(Listen) if not {ok, Socket}.  Maybe better to follow that pattern than to exit with a badarg here

lib/ssl/src/inet_tls_dist.erl
```diff
@@ -215,27 +215,53 @@ gen_accept(Driver, Listen) ->
...
+    receive
+        {continue, HandshakePid} ->
+            accept_loop(DLK, undefined, MaxPending, Pending#{HandshakePid => true});
+        {'EXIT', Pid, _} when is_map_key(Pid, Pending) ->
+            accept_loop(DLK, HandshakePid, MaxPending, maps:remove(Pid, Pending));
+        {'EXIT', HandshakePid, Reason} ->
+            %% HandshakePid crashed before turning into Pending, which means
+            %%  error happened in accept. Need to restart the listener.
+            exit(Reason)
+    end.
```
RaimoNiskanen, line 245: A catchall to waste unknown messages is missing.  Either log and ignore or just ignore.

lib/ssl/src/inet_tls_dist.erl
max-au: Good catch. I did put the log statement before, as originally TLS handshake process was both logging, and exiting with some non-normal reason. 
Now it logs, and exits normally, so logging here means that handshake process crashed unexpectedly.

max-au:
Thank you, good finds.

> and net_kernel will restart a new acceptor process on the same listen socket. 

Interesting! I assumed that net_kernel closes and re-opens listening socket when acceptor process fails. So I experimented with gen_tcp:close() on a listening port. This caused max_restart_intensity of net_kernel and subsequently entire node failed. It is outside of this patch context, but may be worth changing.

IngelaAndin:
@max-au @RaimoNiskanen  On the subject of  unexpected crashes of handshake processes I do not think there will be many as they are designed to always go down gracefully (using try ...catch) sending a TLS-ALERT as that is what the protocol expects. This of course is not always optimal during development and debugging. But I have tried to compensate with good logger messages for debugging and during development I can of course have temporary code doing what I want.

RaimoNiskanen:
@max-au - Yes.  I wonder why there is no reopening of the listen socket.  It sounds like a good remedy for unforseen problems. Could be because reopening the listen socket on the same port may cause "address in use" depending on the distribution module implementation (reuseaddr or not, fixed port or not, ...) so net_kernel, maybe do not want to assume it can be done...

RaimoNiskanen:
@rickard-green did not have any other theories, so we'll leave that problem for the future.
@max-au I have put the branch in testing again, so we hopefully can merge this soonish.

RaimoNiskanen:
@max-au: We have daily test errors for the in-this-pr introduced test case `ssl_dist_SUITE:dist_port_overload`.  I will take one at the time...

On FreeBSD machines `dos/3` fails on line 216 with case clause `{error,econnreset}`, and on our Windows 10 machine it is `{error,econnrefused}` instead.  So it seems the test case's assumption that a filled backlog will cause connection timeouts does not hold on all platforms.  These ones instead seem to refuse new incoming connections.

On Darwin and Solaris we also have a case clause on the same line, but there it is `{error,emfile}`, so I will have to figure out if we have some low system limit or what is happening there.  I will be back on this one...

On One Linux machine we get a test case timetrap timeout on the same line, in `prim_inet:connect0`, after 24 s, which is odd since the test case uses `gen_tcp:connect` with a timeout of 2 s.  This is an odd one that will take some investigation...

RaimoNiskanen:
@max-au: Regarding the `{error,emfile}` error on Darwin and Solaris(2.11) those systems have the limit of open file descriptors for a process (ulimit -n) as 256 per default.  Linux has 1024.  This is only half an explanation since the backlog is set to 128.  The other half may be (from stackoverflow):
> Also, the backlog argument can have different meanings in different socket implementations.
> * In most it means the size of the half-open connection queue, in some it means the size of the completed connection queue.
> * In many implementations, the backlog argument will multiplied to yield a different queue length.
> * If a value is specified that is too large, all implementations will silently truncate the value to maximum queue length anyways.

RaimoNiskanen:
@max-au: In conclusion; we (you and others in this PR) need to figure out what to do on platforms with refused connection for full backlog.  Either find a different way to test, or skip the test on those platforms.

We also need to figure out what to do on platforms with lower descriptor limit than backlog size, maybe lower the backlog size so we do not get `emfile`, or, again, skip the test on those platforms.

It would be a bit sad to only test on Linux, though.

The strange timeout error the OTP team will have to figure out, since it seems to be a bug in inet_drv, the VM, or the OS, maybe in combination.  It does not seem to be this PR:s fault, anyway.

RaimoNiskanen:
@max-au, again:  We have further investigated the problem with the strange timeout error, and that also seems to be a problem with how the test case assumes that the backlog should work.  On this machine (an Ubuntu 16, possibly with some tweaked network settings), the dos() loop never finishes, seemingly connects are delayed and old connections are dropped as fast as new ones are created.

So the dos() loop never reaches a full backlog.

max-au:
Thank you! This is indeed a lot of variety in how systems treat backlog. 
I will take some time re-thinking how to test this reliably on systems that have different backlog treatment. Probably it can be done with simulated delay at TLS connection stage (in hostname check callback, test can introduce a reliable delay). This will not test real world DoS but at least will test that TLS connections are being concurrently established, and only to some degree.

max-au:
I do not have a way to test on Solaris or custom Ubuntu, but at least Darwin seems to pass consistently (and fail the test before the patch). Hope this version works on all supported OSed.

IngelaAndin:
We will test it again and hopefully it  will pass.

IngelaAndin:
We are seeing some new failures now, however I do not think it is your fault!  It is rather a legacy problem and I think it will be fixed by adding the line
	    `ssl_test_lib:clean_start(),`   as first  action after starting crypto  in init_per_suite in ssl_dist_SUITE.  Could  you add that line, as your test fails due to  this  it would be easiest if your branch fixes it .  The previous problems looks as if they are gone.

IngelaAndin:
@max-au  I am afraid that the test cases seems to break almost all  the time now with the reason {actual,0,expected,64}.  I have not had time to analyze  it yet, it seems strange. If  you run them locally with my suggested  change do they pass for you?

max-au:
Yes, and I believe I updated (force-pushed) the branch a few days ago.

I wonder if I am running tests correctly. I usually do "./otp_build autoconf && ./configure && make -j16 && make -j16 tests", then "cd $ERL_TOP/cd release/tests/test_server/", run "$ERL_TOP/bin/cerl" and enter:
ts:install().
ts:run(ssl, ssl_dist_SUITE, [batch]).


(or, ts:run(ssl, [batch]) - which takes a huge amount of time, but only fails oscp stabling tests with "{ssl_test_lib,check_active_receive,2487}"

IngelaAndin:
That is a correct way to run the tests and how our daily builds run them.  I would have run erl and not cerl (but it seems to be eqvivalent). The ssl test is however not dependent on any c-code so I usually run them with pure common_test as then the test are run in your current shell and you can easily use tracing. And yes running all TLS/SSL test takes time.

 ```
cd $ERL_TOP/cd release/tests/ssl_test  
 $ERL_TOP/bin/erl
ct:run_test([{suite, ssl_dist_SUITE}]). 
```
The OCSP code is purely experimental, and not supported yet. Probably it is an interop problem with your version of OpenSSL. And this is not related to Erlang distribution over TLS.  

The test has already failed this morning on 13 machines running ubuntu, freebsd, darwin, solaris and netbsd. 

```
ssl_dist_SUITE,dist_port_overload,223}
{{actual,0,expected,...},[{ssl_dist_SUITE,...},{...}|...]}

```

**Comment by max-au** [comment]
This is indeed strange. I am running tests on MacOS X (Catalina), CentOS7 and CentOS8, all tests pass. 
I was not able to build OTP for FreeBSD 12.2, it fails with "Internal compiler error: Segmentation fault signal terminated program cc1plus" while compiling beam_asm.o file.

Will try Ubuntu, to see whether I am missing something.

**Comment by max-au** [comment]
I tested on OpenBSD 6.8 GENERIC.MP#98 amd64 and also Ubuntu 20.04.1 LTS, test passes just fine. Could it be some special setting that I am missing, e.g. custom socket backlog size?

**Comment by IngelaAndin** [comment]
@RaimoNiskanen  any  ideas?

**Comment by RaimoNiskanen** [comment]
I started to figure out heads or tails of that test case.  I suspect it is some SSL problem with some machines in our lab, or induced by our lab environment.  Maybe the test node does not manage to start.  Since no connection is succesful it is some general connection problem, not the backlog size.  Me and @IngelaAndin should do a common debug session...

**Comment by IngelaAndin** [comment]
@max-au  We decide the test suite needs some refreshing to make sure we have a solid base. I will ask you to rebase in a couple of days and use the new test suite to implement your test.

**Comment by max-au** [comment]
I wonder if it would make sense to split the PR in two.
First part implements actual concurrent TLS distribution (which has been extensively tested in production environment) and test suite. This would simplify later merge/rebase procedure.

**Comment by IngelaAndin** [comment]
@max-au  You mean first the concurrent version (like the original PR) and then the overload protection? The first version did not include any new test cases but passed the old ones?! Hopefully we will find some more time for this soon. We been pretty preoccupied and on Christmas holidays.

**Comment by max-au** [comment]
I mean I could split this PR (in the most recent form) in two: one touching "inet_tls_dist.erl" (production code), and another one implementing a test case separately. But if there is progress expected, I can surely wait longer.

**Comment by IngelaAndin** [comment]
I think we might be making some progress on test cuddling now. I need at least one runt more to be sure. Thank you for your patience.

**Comment by IngelaAndin** [comment]
So I fixed some problems in the dist suite, but your branch still fails. Tonight I will be testing my larger rewrite of the test that had some odd problems earlier and that I think might actually have been related to the other problems now fixed. I also have a fix for the ERL-1458 which would be good to have fixed next time trying out your branch. So slowly getting there.

**Comment by IngelaAndin** [comment]
@max-au I have now merged some updates to the test suites to make sure we do not have any legacy problems with certificates used or illogical option combinations. So please rebase and lets try again. The CRL test removed would not execute any different code path then normal CRL test as they are all handled by common TLS code and not handled at all in specific dist code. We might still add such test later when modernizing the test suite further for completeness testing of the supervisor tree, but that would require more work to ensure sane input for all platforms and will prioritized later.

**Comment by max-au** [comment]
Thank you! I rebased the patch to the current master, and now all SSL test cases pass, except for `ocsp_stapling_basic` (TLS 1.2, DTLS 1.2) and `ocsp_stapling_with_nonce` for DTLS1.2.

**Comment by IngelaAndin** [comment]
I would not worry about the OCSP tests as this functionality is not ready yet.  Alas your test case is still  failing big time in out tests. I think that the timeout calculation in the barrier function could be problematic on some operating systems. I would suggest that you rewrite the test to make a timestamp wait for 64 answers and take a new timestamp and check that the difference is in an acceptable size.

**Comment by max-au** [comment]
@IngelaAndin  would it be possible to paste the error code, and OS where it happens?  I was not able to reproduce it on my set of VMs, but I will follow the suggestion and rewrite the test.

**Comment by IngelaAndin** [comment]
The testcase failed in the same way as before, no error other than that no handshakes where registered as compleated 

```
ssl_dist_SUITE,dist_port_overload,223}
{{actual,0,expected,...},[{ssl_dist_SUITE,...},{...}|...]}

```

max-au:
This failure is really unexpected. I can only think of 0 actual connection established when a socket cannot be opened.

I rewrote the test to completely avoid using ssl library, and rely now on pure TCP instead. The way test works now relies on net_setuptime behaviour. The setup time starts ticking form the moment dist connection is accepted. I establish 4 TCP connections concurrently, so if accept is concurrent, all 4 start ticking immediately (which means, in 7 seconds all 4 are closed). 

However if acceptor is not concurrent, it will create a streak of 4x7 = 28 seconds timeout. In the test I wait for 10 seconds, and if I don't get all 4 connections closed, I declare failure.
I have force-pushed latest squashed commit.

IngelaAndin:
@max-au Thank you for the rewrite, it is looking promising now :) Have to wait for some more test runs to complete before we can merge.

IngelaAndin:
Thank you for the PR and your patience now merged to master in time for release candidate on OTP-24

PR #2655, opened 2020-06-10
erts: set quota to 1 if cpu quota is less than 1

tony612:
Fixes https://bugs.erlang.org/browse/ERL-1280

With this, if CPU quota is less than 1, we will try to use `1` as the quota. Without this, cores number will be used.


jhogberg:
Thanks for the PR! Can you rebase it to maint?

tony612:
@jhogberg Done. I rebased and changed the base branch to `maint`.

jhogberg:

PR #2657, opened 2020-06-10
Clarify how to make contributions

IngelaAndin:
We want it to be as easy as possible to review a pull request without
breaking the possibility to use bisect.

The need for improvement was discovered due to PR-1968 and a first
enhancement proposal made in PR-2652, but we discovered it needed to
be reworded to fulfill our whole requirement.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,8 +109,11 @@ Make sure that your branch contains clean commits:
 * Don't merge `maint` or `master` into your branch. Use `git rebase` if you need to resolve merge
 conflicts or include the latest changes.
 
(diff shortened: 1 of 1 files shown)


jhogberg:
I find the wording a bit unclear, how about something like:

>* Each commit should represent a logical change, such as a feature added or bug fixed.
>
>* Changes to multiple applications should be made in separate commits when possible.
>    
>    Generally speaking, a commit should only change multiple applications when they're mutually dependent on each other.

While I'm at it, maybe we could sneak this in too:

> * When making a code change, don't split the corresponding changes to documentation or tests into separate commits.

PR #2663, opened 2020-06-17
Fix for delete_object at read on a set table

emafr:
On a set table, if the delete_object doesnt delete (due to no match) at commit
return whatever is (or is not) stored in source table as a result of a read in
the same transaction as the delete_object.
And, if a delete already has been done on Oid in the same transaction do not
record a delete_object operation on the same Oid.
Added delete_object tests on set table and made the existing tests checking more.

Before applying the fix in mnesia.erl one of the added tests (`%% Delete other object than already existing"`) in mnesia_trans_access_test:delete_object_set/1 fail:

`
{mnesia_trans_access_test,delete_object_set},
   [{'REASON',"mnesia_trans_access_test.erl",303,
     [{aborted,
        {{badmatch,[]},
        ...
`

The other tests are added to cover for the new behaviour.


dgud:
Looks good, can you fix the CLA issue above.

PR #2664, opened 2020-06-18
Make compiler report column as well as line number

richcarl:
This adds column number support for the compiler, for passes before the transformation to Core Erlang.

There is one problem with the warnings_SUITE latin1_fallback test, which appears to be nondeterministic. The reported column in the invalid_unicode error message (apparently from the io server) is sometimes 24, sometimes 25, and sometimes 1.


nox:
My own patches included documentation. :P

uabboli:
Since we've made many updates and also rebased the branch, we've decided to create a new PR. See also https://github.com/erlang/otp/pull/3006.

PR #2665, opened 2020-06-20
break.c: Rephrase "Eh?"

henrik:
This is intended mostly as a starting point for a discussion – checking if the powers that be are open to tweaking this text at all.

I've found the "Eh?" prompts a little grating – the tone comes across to me as the system being annoyed.

Also, this prompt could help the user out by clarifying what it's after. I seem to remember getting stuck in this state once or twice as a newcomer to the ecosystem.

I considered repeating the options ("Sorry? Please enter one of these options: …"), but figured the options will usually still be on screen, so it would add noise to the output and slight complexity to the code, keeping them in sync.

What do you think?

### DIFF
--- a/erts/emulator/beam/break.c
+++ b/erts/emulator/beam/break.c
@@ -563,6 +563,8 @@ dump_attributes(fmtfn_t to, void *to_arg, byte* ptr, int size)
 void
 do_break(void)
 {
+    const char *helpstring = "BREAK: (a)bort (A)bort with dump (c)ontinue (p)roc info (i)nfo\n"
+        "       (l)oaded (v)ersion (k)ill (D)b-tables (d)istribution\n";
     int i;
 #ifdef __WIN32__
     char *mode; /* enough for storing "window" */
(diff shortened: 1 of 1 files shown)


mikpe:
Factor out the initial printout to a help() function, add 'h' to show the help text, then rephrase the Eh? bit to say something like "Invalid option.  Type h to show available options.".

But yes, "Eh?" is suboptimal.

henrik:
@garazdawi Just checking – does self-assigned mean you're planning on implementing this, or are you waiting for me to update the PR per @mikpe's feedback?

I don't usually work in C and wouldn't mind at all if someone else did this, but I'm also happy to give it a shot if you'd prefer :)

garazdawi:
I meant to take a look at it before at the time, but other things got in the way. Now I'm on vacation so will get back to you when I get back to the (home) office.

garazdawi:
I rebased to maint and polished the code a bit.

Thanks for you contribution!

PR #2666, opened 2020-06-23
release_handler more resilient of exiting processes during upgrade

sverker:
Proposed fix for https://bugs.erlang.org/browse/ERL-1247


zsoci:
Seems to be OK.

PR #2674, opened 2020-06-26
EDoc: Allow user defined macros to be functions

maxno-kivra:
This makes them equal to the builtin ones, and allows for neat things
like linking to the offical docs page:

```erlang
man_doc_macro(MacroArg, _Line, _Env) ->
    {ok, Pattern} = re:compile("([^:]+):([^/]+)/(\\d+)"),
    {match, [Module, Function, Arity]} = re:run(
        MacroArg, Pattern, [{capture, all_but_first, binary}]
    ),
    lists:flatten(io_lib:format(
        "<a href=\"https://erlang.org/man/~ts.html#~ts-~ts\"><code>~ts</code></a>",
        [Module, Function, Arity, MacroArg]
    )).
```

---

Which allows for:

```erlang
%%% {@man maps:merge/2}
```

To output:

```html
<a href="https://erlang.org/man/maps.html#merge-2"><code>maps:merge/2</code></a>
```

### DIFF
--- a/lib/edoc/src/edoc_macros.erl
+++ b/lib/edoc/src/edoc_macros.erl
@@ -58,6 +58,8 @@ std_macros(Env) ->
 
 check_defs([{K, D} | Ds]) when is_atom(K), is_list(D) ->
     check_defs(Ds);
+check_defs([{K, D} | Ds]) when is_atom(K), is_function(D, 3) ->
+    check_defs(Ds);
 check_defs([X | _Ds]) ->
     report("bad macro definition: ~P.", [X, 10]),
     exit(error);


maxno-kivra:
@uabboli is there something I can do to help move this forward?

uabboli:
I'm not a frequent user of EDoc macros, but I think that your suggestion can come in handy.
What is missing is an example in edoc_SUITE and an update of the description of edoc:get_doc/2.

zuiderkwast:
It would be nice if the regular references like `@link` and `@see` (as in `@see maps:merge/2`) would link to the official documentation automatically, if the module belong to an application which is in OTP.

For 3rd party applications (as in `@see cowboy:start_tls/3`), maybe it could be configurable/extensible...

uabboli:
Thank you for the pull request. I've updated the docs and merged to master.

maxno-kivra:
Thanks for getting this in.

PR #2675, opened 2020-06-29
Add optional_applications to .app resource files

josevalim:
Both Mix and Rebar allow some applications to be absent
at runtime - sometimes also known as optional dependencies.

However, given optional applications are not stored in .app
resource files, releases do not consider optional applications
in its boot order, leaving it up to chance if an optional app
will be started before its parent.

Users can try to explicitly list optional applications on their
release definition files, but given the order is not enforced,
this manual specification may be reordered when new apps
are added, leaving developers with broken releases.

This PR introduces the "optional_applications" field to .app
resource files. This field must be a subset of the "applications"
field. The reason why it is a subset rather than its own list
of applications is exactly to keep application start order
on "application:start/2" and during releases.

If application "b" is an optional application for application "a",
and application "b" is missing, "application:start(a)" will still
succeed.

If application "b" is an optional application for application "a",
and application "b" is available, "application:ensure_all_started(a)"
will automatically start application "b" before "a".

### DIFF
--- a/lib/kernel/doc/src/app.xml
+++ b/lib/kernel/doc/src/app.xml
@@ -59,6 +59,7 @@
    {maxT,         MaxT},
    {registered,   Names},
    {included_applications, Apps},
+   {optional_applications, Apps},
    {applications, Apps},
    {env,          Env},
    {mod,          Start},
@@ -148,10 +149,21 @@ ApplicationVersion = string()</code>
       <tag><c>applications</c></tag>
       <item>
         <p>All applications that must be started before this
-          application is allowed to be started. <c>systools</c> uses
-          this list to generate correct start scripts. Defaults to
-          the empty list, but notice that all applications have
-          dependencies to (at least) Kernel and STDLIB.</p>
+          application. If an application is also listed in
+          <c>optional_applications</c>, then the application
+          is not required to exist (but if it exists, it is
+          also guaranteed to be started before this one).</p>
+        <p><c>systools</c> uses this list to generate correct start
+          scripts. Defaults to the empty list, but notice that all
+          applications have dependencies to (at least) Kernel and STDLIB.</p>
+      </item>
+      <tag><c>optional_applications</c></tag>
+      <item>
+        <p>A list of <c>applications</c> that are optional.
+          Note if you want an optional dependency to be
+          automatically started before the current application
+          whenever it is available, it must be listed on both
+          <c>applications</c> and <c>optional_applications</c>.</p>
       </item>
       <tag><c>env</c></tag>
       <item>
@@ -239,4 +251,3 @@ ApplicationVersion = string()</code>
       <seeerl marker="sasl:systools"><c>systools(3)</c></seeerl></p>
   </section>
 </fileref>
-
--- a/lib/kernel/src/application.erl
+++ b/lib/kernel/src/application.erl
@@ -130,7 +130,7 @@ ensure_all_started(Application) ->
       Started :: [atom()],
       Reason :: term().
 ensure_all_started(Application, Type) ->
-    case ensure_all_started(Application, Type, []) of
+    case ensure_all_started([Application], [], Type, []) of
 	{ok, Started} ->
 	    {ok, lists:reverse(Started)};
 	{error, Reason, Started} ->
@@ -138,23 +138,45 @@ ensure_all_started(Application, Type) ->
 	    {error, Reason}
     end.
 
-ensure_all_started(Application, Type, Started) ->
-    case start(Application, Type) of
-	ok ->
-	    {ok, [Application | Started]};
-	{error, {already_started, Application}} ->
-	    {ok, Started};
-	{error, {not_started, Dependency}} ->
-	    case ensure_all_started(Dependency, Type, Started) of
+ensure_all_started([App | Apps], OptionalApps, Type, Started) ->
+    case ensure_loaded(App) of
+	{ok, Name} ->
+	    case ensure_started(Name, App, Type, Started) of
 		{ok, NewStarted} ->
-		    ensure_all_started(Application, Type, NewStarted);
+		    ensure_all_started(Apps, OptionalApps, Type, NewStarted);
 		Error ->
 		    Error
 	    end;
+	{error, {"no such file or directory", _} = Reason} ->
+	    case lists:member(App, OptionalApps) of
+		true ->
+		    ensure_all_started(Apps, OptionalApps, Type, Started);
+		false ->
+		    {error, {App, Reason}, Started}
+	    end;
 	{error, Reason} ->
-	    {error, {Application, Reason}, Started}
-    end.
+	    {error, {App, Reason}, Started}
+    end;
+ensure_all_started([], _OptionalApps, _Type, Started) ->
+    {ok, Started}.
 
+ensure_started(Name, App, Type, Started) ->
+    {ok, ChildApps} = get_key(Name, applications),
+    {ok, OptionalApps} = get_key(Name, optional_applications),
+
+    case ensure_all_started(ChildApps, OptionalApps, Type, Started) of
+	{ok, NewStarted} ->
+	    case application_controller:start_application(Name, Type) of
+		ok ->
+		    {ok, [App | NewStarted]};
+		{error, {already_started, App}} ->
+		    {ok, NewStarted};
+		{error, Reason} ->
+		    {error, {App, Reason}, NewStarted}
+	    end;
+	Error ->
+	    Error
+    end.
 
 -spec start(Application) -> 'ok' | {'error', Reason} when
       Application :: atom(),
@@ -169,12 +191,19 @@ start(Application) ->
       Reason :: term().
 
 start(Application, RestartType) ->
+    case ensure_loaded(Application) of
+	{ok, Name} ->
+	    application_controller:start_application(Name, RestartType);
+	Error ->
+	    Error
+    end.
+
+ensure_loaded(Application) ->
     case load(Application) of
 	ok ->
-	    Name = get_appl_name(Application),
-	    application_controller:start_application(Name, RestartType);
+	    {ok, get_appl_name(Application)};
 	{error, {already_loaded, Name}} ->
-	    application_controller:start_application(Name, RestartType);
+	    {ok, Name};
 	Error ->
 	    Error
     end.
--- a/lib/kernel/test/application_SUITE.erl
+++ b/lib/kernel/test/application_SUITE.erl
@@ -31,7 +31,7 @@
 	 otp_3002/1, otp_3184/1, otp_4066/1, otp_4227/1, otp_5363/1,
 	 otp_5606/1,
 	 start_phases/1, get_key/1, get_env/1,
-	 set_env/1, set_env_persistent/1, set_env_errors/1,
+	 set_env/1, set_env_persistent/1, set_env_errors/1, optional_applications/1,
 	 permit_false_start_local/1, permit_false_start_dist/1, script_start/1, 
 	 nodedown_start/1, init2973/0, loop2973/0, loop5606/1, otp_16504/1]).
 
@@ -60,7 +60,7 @@ all() ->
      permit_false_start_dist, get_key, get_env, ensure_all_started,
      set_env, set_env_persistent, set_env_errors,
      {group, distr_changed}, config_change, shutdown_func, shutdown_timeout,
-     shutdown_deadlock, config_relative_paths,
+     shutdown_deadlock, config_relative_paths, optional_applications,
      persistent_env, handle_many_config_files, format_log_1, format_log_2,
      configfd_bash, configfd_port_program].
 
@@ -970,7 +970,7 @@ ensure_all_started(_Conf) ->
     w_app9(Fd9),
     file:close(Fd9),
     {ok, Fd10} = file:open("app10.app", [write]),
-    w_app10_dep9(Fd10),
+    w_app10(Fd10, [app9], []),
     file:close(Fd10),
     {ok, FdErr} = file:open("app_chain_error.app", [write]),
     w_app(FdErr, app_chain_error()),
@@ -1037,6 +1037,60 @@ ensure_all_started(_Conf) ->
     ok = application:unload(app10),
     ok = application:unload(app_chain_error2),
     ok = application:unload(app_chain_error),
+    ok.
+
+optional_applications(_Conf) ->
+    {ok, Fd10} = file:open("app10.app", [write]),
+    w_app10(Fd10, [app9], []),
+    file:close(Fd10),
+
+    {error,{not_started,app9}} = application:start(app10),
+    {ok, []} = application:get_key(app10, optional_applications),
+    ok = application:unload(app10),
+
+    %% List app9 as an optional application and app10 starts
+    {ok, Fd10Opt} = file:open("app10.app", [write]),
+    w_app10(Fd10Opt, [app9], [app9]),
+    file:close(Fd10Opt),
+
+    ok = application:start(app10),
+    {ok, [app9]} = application:get_key(app10, optional_applications),
+    false = lists:keymember(app9,1,application:which_applications()),
+
+    ok = application:stop(app10),
+    ok = application:unload(app10),
+
+    %% If app9 is defined, we can still start app10 without app9
+    {ok, Fd9} = file:open("app9.app", [write]),
+    w_app9(Fd9),
+    file:close(Fd9),
+
+    ok = application:start(app10),
+    false = lists:keymember(app9,1,application:which_applications()),
+
+    ok = application:stop(app10),
+    ok = application:unload(app10),
+
+    %% But if we use ensure all started, then app9 is started too
+    {ok, [app9, app10]} = application:ensure_all_started(app10, temporary),
+    true = lists:keymember(app9,1,application:which_applications()),
+
+    ok = application:stop(app9),
+    ok = application:unload(app9),
+    ok = application:stop(app10),
+    ok = application:unload(app10),
+
+    %% Finally, let's have an optional dependency with a start error
+    {ok, Fd10Error} = file:open("app10.app", [write]),
+    w_app10(Fd10Error, [app_start_error], [app_start_error]),
+    file:close(Fd10Error),
+
+    {ok, FdAppError} = file:open("app_start_error.app", [write]),
+    w_app_start_error(FdAppError),
+    file:close(FdAppError),
+
+    {error,{app_start_error,_}} = application:ensure_all_started(app10, temporary),
+    ok = application:unload(app10),
     ok.
 
 %%%-----------------------------------------------------------------
@@ -1430,7 +1484,7 @@ otp_4227(Conf) when is_list(Conf) ->
         rpc:multicall(Cps, application, load, [app9()]),
     ?UNTIL(is_loaded(app9, Cps)),
     {[ok,ok],[]} = 
-        rpc:multicall(Cps, application, load, [app10_dep9()]),
+        rpc:multicall(Cps, application, load, [app10([app9], [])]),
     {error, {not_started, app9}} = 
 	rpc:call(Cp1, application, start, [app10]),
 
@@ -1642,6 +1696,7 @@ get_key(Conf) when is_list(Conf) ->
 		{maxT, infinity}, 
 		{registered, []}, 
 		{included_applications, [appinc1, appinc2]}, 
+		{optional_applications, []},
 		{applications, [kernel]}, 
 		{env, Env}, 
 		{mod, {application_starter, [ch_sup, {appinc, 41, 43}] }}, 
@@ -1686,6 +1741,7 @@ get_key(Conf) when is_list(Conf) ->
 		{maxT, infinity}, 
 		{registered, []}, 
 		{included_applications, [appinc1, appinc2]}, 
+		{optional_applications, []},
 		{applications, [kernel]}, 
 		{env, Env}, 
 		{mod, {application_starter, [ch_sup, {appinc, 41, 43}] }}, 
@@ -2623,13 +2679,14 @@ app9() ->
       {applications, [kernel]},
       {mod, {ch_sup, {app9, 19, 19}}}]}.
 
-app10_dep9() ->
+app10(Apps, OptionalApps) ->
     {application, app10,
      [{description, "ERTS  CXC 138 10"},
       {vsn, "2.0"},
       {modules, []},
       {registered, []},
-      {applications, [kernel, app9]},
+      {applications, [kernel] ++ Apps},
+      {optional_applications, OptionalApps},
       {mod, {ch_sup, {app10, 20, 20}}}]}.
 
 appinc() ->
@@ -2953,8 +3010,8 @@ w_app8(Fd) ->
 w_app9(Fd) ->
     io:format(Fd, "~p.\n", [app9()]).
 
-w_app10_dep9(Fd) ->
-    io:format(Fd, "~p.\n", [app10_dep9()]).
+w_app10(Fd, Deps, Optional) ->
+    io:format(Fd, "~p.\n", [app10(Deps, Optional)]).
 
 w_app_start_error(Fd) ->
     io:format(Fd, "~p.\n", [app_start_error()]).
--- a/lib/kernel/doc/src/application.xml
+++ b/lib/kernel/doc/src/application.xml
@@ -73,18 +73,16 @@
       <desc>
         <p>Equivalent to calling
         <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
-            repeatedly on all dependencies that are not yet started for an
-            application that is not yet started.</p>
-        <p>Returns <c>{ok, AppNames}</c>, where <c>AppNames</c> is a list of the application names
-            that was actually started by this call.
-            The list might be empty, or not contain all dependencies if the application
-            or some of its dependencies are already started.</p>
+            repeatedly on all dependencies that are not yet started for an application.
+            Optional dependencies will also be loaded and started if they are available.</p>
+        <p>Returns <c>{ok, AppNames}</c> for a successful start or for an already started
+            application (which is, however, omitted from the <c>AppNames</c> list).</p>
         <p>The function reports <c>{error, {AppName,Reason}}</c> for errors, where
         <c>Reason</c> is any possible reason returned by
         <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
             when starting a specific dependency.</p>
-        <p>If an error occurs, the applications started by the function are stopped
-            to bring the set of running applications back to its initial state.</p>
+	<p>If an error occurs, the applications started by the function are stopped
+	    to bring the set of running applications back to its initial state.</p>
       </desc>
     </func>
     <func>
@@ -352,16 +350,22 @@ Nodes = [cp1@cave, {cp2@cave, cp3@cave}]</code>
         <p>The application controller checks the value of
           the application specification key <c>applications</c>, to
           ensure that all applications needed to be started before
-          this application are running. Otherwise,
+          this application are running. If an application is missing
+          and the application is not marked as optional,
           <c>{error,{not_started,App}}</c> is returned, where <c>App</c>
-          is the name of the missing application.</p>
-        <p>The application controller then creates an <em>application master</em>
-          for the application. The application master becomes the
-          group leader of all the processes in the application. I/O is
-          forwarded to the previous group leader, though, this is just
-          a way to identify processes that belong to the application.
-          Used for example to find itself from any process, or,
-          reciprocally, to kill them all when it terminates.</p>
+          is the name of the missing application. Note this function
+          makes no attempt to start any of the applications listed in
+          <c>applications</c>, not even optional ones. See
+          <seemfa marker="#ensure_all_started/1"><c>ensure_all_started/1,2</c></seemfa>
+          for recursively starting the current application and its
+          dependencies.</p>
+        <p>Once validated, the application controller then creates an
+          <em>application master</em> for the application. The application
+          master becomes the group leader of all the processes in the
+          application. I/O is forwarded to the previous group leader,
+          though, this is just a way to identify processes that belong
+          to the application. Used for example to find itself from any
+          process, or, reciprocally, to kill them all when it terminates.</p>
         <p>
           The application master starts the application by calling
           the application callback function <c>Module:start/2</c> as
--- a/lib/kernel/src/application_controller.erl
+++ b/lib/kernel/src/application_controller.erl
@@ -158,7 +158,7 @@
 %% Env         = [{Key, Value}]
 %%-----------------------------------------------------------------
 
--record(appl, {name, appl_data, descr, id, vsn, restart_type, inc_apps, apps}).
+-record(appl, {name, appl_data, descr, id, vsn, restart_type, inc_apps, opt_apps, apps}).
 
 %%-----------------------------------------------------------------
 %% Func: start/1
@@ -373,6 +373,8 @@ get_key(AppName, Key) ->
 		    {ok, (Appl#appl.appl_data)#appl_data.regs};
 		included_applications ->
 		    {ok, Appl#appl.inc_apps};
+                optional_applications ->
+                    {ok, Appl#appl.opt_apps};
 		applications ->
 		    {ok, Appl#appl.apps};
 		env ->
@@ -404,6 +406,7 @@ get_all_key(AppName) ->
 		  {maxT, (Appl#appl.appl_data)#appl_data.maxT},
 		  {registered, (Appl#appl.appl_data)#appl_data.regs},
 		  {included_applications, Appl#appl.inc_apps},
+                  {optional_applications, Appl#appl.opt_apps},
 		  {applications, Appl#appl.apps},
 		  {env, get_all_env(AppName)},
 		  {mod, (Appl#appl.appl_data)#appl_data.mod},
@@ -1277,15 +1280,15 @@ do_load_application(Application, S) ->
 
 %% Recursively load the application and its included apps.
 %load(S, {ApplData, ApplEnv, IncApps, Descr, Vsn, Apps}) ->
-load(S, {ApplData, ApplEnv, IncApps, Descr, Id, Vsn, Apps}) ->
+load(S, {ApplData, ApplEnv, IncApps, OptApps, Descr, Id, Vsn, Apps}) ->
     Name = ApplData#appl_data.name,
     ConfEnv = get_env_i(Name, S),
     NewEnv = merge_app_env(ApplEnv, ConfEnv),
     CmdLineEnv = get_cmd_env(Name),
     NewEnv2 = merge_app_env(NewEnv, CmdLineEnv),
     add_env(Name, NewEnv2),
-    Appl = #appl{name = Name, descr = Descr, id = Id, vsn = Vsn, 
-		 appl_data = ApplData, inc_apps = IncApps, apps = Apps},
+    Appl = #appl{name = Name, descr = Descr, id = Id, vsn = Vsn, apps = Apps,
+		 appl_data = ApplData, inc_apps = IncApps, opt_apps = OptApps},
     ets:insert(ac_tab, {{loaded, Name}, Appl}),
     NewS =
 	foldl(fun(App, S1) ->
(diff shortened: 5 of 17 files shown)


tsloughter:
Big +1 from me. 

But to be clear, at least with relx/systools, the order is not subject to be "reordered at any time". A relx release definition passed to systools uses a stable sort even for the top level applications listed for the release. So you can ensure a dependency is started before another by using this. The issue with it is just that it is that it is annoying and pushed on to the end user to get right, meaning every time it is important it must be documented for that particular project, like OpenTelemetry does https://github.com/open-telemetry/opentelemetry-erlang#using

lib/kernel/doc/src/app.xml
```diff
@@ -148,10 +149,17 @@ ApplicationVersion = string()</code>
...
       <item>
         <p>All applications that must be started before this
-          application is allowed to be started. <c>systools</c> uses
-          this list to generate correct start scripts. Defaults to
-          the empty list, but notice that all applications have
-          dependencies to (at least) Kernel and STDLIB.</p>
+          application. If an application is also listed in
+          <c>optional_applications</c>, then there will be an
+          attempt to start the optional application, but failing
+          to run it won't stop this application from starting.</p>
```
tsloughter, line 155: I disagree here. If the optional dep is in the code path I would expect it to fail startup if that application fails to start.

lib/kernel/src/application.erl
```diff
@@ -139,22 +139,40 @@ ensure_all_started(Application, Type) ->
...
 
+ensure_children([], _OptionalApps, _Type, Started) ->
+    {ok, Started};
+ensure_children([App | Apps], OptionalApps, Type, Started) ->
+    case ensure_all_started(App, Type, Started) of
+	{ok, NewStarted} ->
+	    ensure_children(Apps, OptionalApps, Type, NewStarted);
+	{error, _AppReason, NewStarted} = Error ->
+	    case lists:member(App, OptionalApps) of
+		true -> ensure_children(Apps, OptionalApps, Type, NewStarted);
```
tsloughter, line 172: As I said in my docs comment I think this should only be ignored if `App` is not in the code path.

This might be tricky because the same error for not being in the code path is used for not having a `.app` file (so in most cases not compiled because that is when that file is generated): `{error,{"no such file or directory", "App.app"}}`

But even with that caveat I think it is better and more expected to the user to have this whole thing fail if an optional dep fails to start for any reason other than `no such file or directory`.

An example might be a database driver. If I include a certain postgres library as an optional dep and it fails to boot I'd rather know during application start that this has failed than to later get errors when calls to the database are attempted.

tsloughter:
So scratch that, I'm a "big +1" only with the change I mention in my review comments -- without it, assuming it at least correctly fails in the case of a release when the included optional dep fails to boot, I'm a small +1 :)

I believe in the case of a release the boot script does a `start` on each application so it won't be the case that it would silently skip an optional dep that fails to boot. Which I think is correct. So the issue is also the different result to the user whether they are booting up locally for testing with like `rebar3 shell` or running a release -- in one case it silently ignores a failed boot and in the other it errors out the whole boot process.

josevalim:
Thanks @tsloughter for the review. "at any time" was poorly phrased on my end, I should have said "at any time new dependencies are introduced", so I have edited it accordingly. Your other feedback is also great, I will do those changes and push them soon.

josevalim:
All feedback from @tsloughter has been addressed. Docs have been amended and tests have been improved.

josevalim:
Pushed support for optional applications on systools too.

garazdawi:
We'll take a look at this once vacation times are over, i.e. sometime in august.

tsloughter:
Bumping this :)

lib/kernel/src/application.erl
```diff
@@ -130,31 +130,53 @@ ensure_all_started(Application) ->
...
+	    {error, {App, Reason}, Started}
+    end;
+ensure_all_started([], _OptionalApps, _Type, Started) ->
+    {ok, Started}.
 
+ensure_started(Name, App, Type, Started) ->
+    {ok, ChildApps} = get_key(Name, applications),
+    {ok, OptionalApps} = get_key(Name, optional_applications),
+
+    case ensure_all_started(ChildApps, OptionalApps, Type, Started) of
```
sverker, line 167: `ensure_all_started/4` does not protects itself against cyclic dependences which would seem to cause an infinite recursion.
Seem to me that a dependency cycle could "hide" in certain combination of available optional applications.

josevalim: My understanding is that this behaviour is unchanged in this patch because OptionalApplications are only taken into account if the `.app` file cannot be found on disk. In case of cyclic apps, assuming both of them are available, then you ran into a loop as we will start both. If one of them is not available, then it won't be started, but by then you don't have a cycle either (as one part is missing).

sverker: I agree that behaviour is unchanged and existing system will not be affected.
My concern was that with the introduction of optional applications it would be possible to construct a system with several optional applications with a "hidden" cycle that is not detected until a certain combination of optional applications are available together. Maybe I'm overstating this risk as I can't really come up with a simple enough example :-)

sverker:
I think `reltool` needs updating too. For example `reltool_target:sort_apps/4` contains the same algorithm as `systools_make:sort_appls/4`.

sverker:
What do you think about allowing applications in `optional_applications` that are not part of `applications`. That would be optional  library applications that do not need to be started. By allowing that, the `optional_applications` list can also serve the `runtime_dependencies` list.

tsloughter:
@sverker do you have examples of how `runtime_dependencies` are used? You say "do not need to be started", is this referring to an application that may or may not be used? Like they are only used if a user triggers a specific action and you don't want the application to start unless such an action is taken?

That is the only thing I could think of as being a reason to include but not start an application.

ferd:
One way we use `applications` in a lot of places for rebar3 is to use it as the path of "what will end up in the release". xref analysis, for example, has shifted to only using these. More importantly, releases also build using these for transitive dependencies. 

The gotcha is that using `optional_applications` as a disjoint set from `applications` is that you risk losing backwards compatibility with libraries relying on it as you move forwards with support for things. Having the sets overlap is trickier for declaration, but allows easy backwards compatibility with people on older toolchains using newer apps.

tsloughter:
> I could have taken the other approach of removing all missing optional applications from the `applications` entry.

They should be removed. Unless I'm misunderstanding. Removal from the `.app` file is definitely what relx will do with this feature.

josevalim:
I briefly talked to @tsloughter on IRC and, to make sure we are all on the same page, keeping the missing optional applications is not an issue because we are keeping both `applications` and `optional_applications` in the release app files and the `application_controller` knows that these missing apps are optional and the system should still boot just fine.

rickard-green:
@tsloughter 

> @sverker do you have examples of how runtime_dependencies are used?

`runtime_dependencies` are used by the [`otp_patch_apply`](https://erlang.org/doc/installation_guide/OTP-PATCH-APPLY.html) tool to determine that runtime dependencies for the applications to patch are satisfied in the installed system being patched. The user can also verify that all runtime dependencies are satisfied using [`system_information:sanity_check()`](https://erlang.org/doc/man/system_information.html#sanity_check-0)

`runtime_dependencies` list all direct dependencies to other applications that may be needed. These dependencies also include minimum version requirements. That is, optional as well as required dependencies. A way to inform about which dependencies are optional (apart from writing it in the documentation) has been requested multiple times.

> You say "do not need to be started", is this referring to an application that may or may not be used? Like they are only used if a user triggers a specific action and you don't want the application to start unless such an action is taken?

Yes this, and library applications which have no requirement to be started.

We also have the scenario with mutual dependency. In that case it would be problematic for both applications to list each other under `applications` in their `.app` files. The documentation say "All applications that must be started before this application is allowed to be started" which also solves this otherwise problematic scenario. That is, we potentially have cases where an application should not be listed in `applications` even though it must be started.

`ssl` and `inets` is an example of applications with mutual dependency. Both include each other in their `runtime_dependencies`. None of them, however, include each other in `applications` since this is optional functionality.

@josevalim 

> Regarding the adding support for optional applications not in the applications list, is it ok if we have this discussion later on? I am hoping this patch is functionally complete and if we add more scenarios, we can do so in a separate discussion.

I'm not too fond of merging things without ironing out the details. I also think it would be really unfortunate to unnecessarily have to introduce yet another parameter if we can use this new one to solve both problems. I'm not sure using `optional_applications` for both is the best way, but I think we should have the discussion.

josevalim:
> I'm not too fond of merging things without ironing out the details. I also think it would be really unfortunate to unnecessarily have to introduce yet another parameter if we can use this new one to solve both problems.

Ah, of course, makes sense. :)

In the current patch, if you list an application in `optional_applications` but not in `applications`, the runtime won't do anything at all with it during start/systools/reltool. But given nothing happens (at the moment), there is no purpose in doing it.

From my (rudimentary) understanding of `runtime_dependencies`, the biggest issue with making it a general mechanism is the version aspect of them. For example, what does "some_app-2.0" allows? 1.0? 2.1? 2.0.1? And is this versioning schema generally applicable outside of Erlang/OTP?

The other issue is what happen if we were to allow this:

    {applications, [kernel, stdlib, some_app]},
    {optional_applications, ["some_app-2.0"]},

Where `some_app` is listed in both `applications` and in `optional_applications` with a version. Does it mean if I only have "some_app-1.0" in the system, it won't be started before? How will the application controller (plus systools/reltool) know these versioning semantics?

I am not sure if this helps but that's my current understanding of what would need to be addressed for the options to work as one. :)

rickard-green:
@josevalim There is no need for versions in the `optional_applications` entry. The required versions are already determined by the `runtime_dependencies` entry. The only extra information needed to solve the `runtime_dependency` issue is whether an application is optional or not.

Regarding the `applications` entry. I guess that if you don't have a mutual dependency between applications, it is ok to list all applications that you depend on in the `applications` entry. Is that how you use it in `relx`, `rebar3`, and `mix`? Or do you only include those applications needed to be started before the application itself? @tsloughter @ferd @josevalim 

@josevalim Just for information, even though versions should not need to be involved in `optional_applications`. Application versions that satisfy a dependency in `runtime_dependencies` are all versions that compare as larger than or equal to the dependency. Besides comparing as larger than, equal to, and smaller than, a version can also be unordered with the dependency. See "order of versions" in the [versions](https://erlang.org/doc/system_principles/versions.html#version-scheme) documentation for more info. No, the version schema is not generally applicable outside of OTP.

ferd:
At least for Rebar3 and relx we don't allow any circular dependency, whether in deps declaration for build-time deps or in applications (which releases would complain about). If you need that, our recommendation is usually to factor the circular parts into a third application that the two others can depend on safely.

We do nothing with 'runtime_applications'  because it never felt like it made sense as a format with the version strings in there (how is something optional even more accurate than something mandatory? even then, many of our version values are unknown until build time because people inject git values into version strings and all apps in a repo get the same version).

The doc for that value says (still today) that "we hope they can be correct for OTP-18" and we never really touched it. It does mention minimum requirements, but as @josevalim mentioned, does little regarding max versions either. It always felt like something that required a ton more effort than would be useful, and considering the community in general, we also always knew we could never reliably expect this value to be there and never built anything that came to expect it. I think this is the first time I hear of the meaning of runtime_dependencies as "optional dependencies when there's an otherwise circular dependency", and I don't think those are clear semantics at all. Closest thing I get to it from the doc is:

> Some dependencies can only be required in specific runtime scenarios. When such optional dependencies exist, these are specified and documented in the corresponding "App" documentation of the specific application.

rickard-green:
@ferd 
> I think this is the first time I hear of the meaning of runtime_dependencies as "optional dependencies when there's an otherwise circular dependency", and I don't think those are clear semantics at all.

I don't recognize that description. It is all direct dependencies that the application have, regardless of whether these are mandatory dependencies or not. That is, all dependencies needed for full functionality of the application.

The mutual dependency issue only regards the `applications` entry, and was there, in the same form, long before the `runtime_dependency` entry was introduced. One of the reasons that `runtime_dependency` was introduced was that `applications` did not reflect the dependencies that the application had, just applications needed to be started before the application itself.

However, if I understand you correct, you list all dependencies in the `applications` entry?

josevalim:
> The only extra information needed to solve the runtime_dependency issue is whether an application is optional or not.

I see. If that's the case, the `optional_applications` field should suffice. We can document that an optional application may not be in `applications` but the runtime will not do anything at all with it.

tsloughter:
I don't think this can be combined with the `runtime_dependencies` feature. I with `runtime_dependencies` wasn't in the `.app` file and instead defined somewhere else, like a `RUNTIME_DEPENDENCIES` file containing each app for the OTP install and its "runtime dependencies", `[{app, ["other-app-vsn", ...]}]`. I find it to be a confusing name and attribute to find in the `.app` file since it is only used for `otp_patch_apply`.

tsloughter:
> However, if I understand you correct, you list all dependencies in the `applications` entry?

Yes. We consider any dependency of an application to have to "start" before the app that uses it and do not allow circular dependencies (so have no concept of a mutual dependency). 

Since we don't have mutual dependencies there is no reason not to include all applications in the `applications` list since nothing is harmed by assuming it must also boot before the application that includes it in the list. Which also ensures it is included in the release.

rickard-green:
@tsloughter 
> I don't think this can be combined with the runtime_dependencies feature.

I fail to see an argument to why, more than that you wish it wasn't there.

> I find it to be a confusing name and attribute to find in the .app file since it is only used for otp_patch_apply.

I have a really hard time seeing that the name is confusing, since it *is* the runtime dependencies of the application.

rickard-green:
>>    However, if I understand you correct, you list all dependencies in the `applications` entry?
>
> Yes. We consider any dependency of an application to have to "start" before the app that uses it and do not allow circular dependencies (so have no concept of a mutual dependency).

Thanks

tsloughter:
> I have a really hard time seeing that the name is confusing, since it _is_ the runtime dependencies of the application.

There are two issues, one may simply be having differing definitions of what a "runtime dependency" is. The `stdlib` `runtime_dependencies` list includes `compiler`, the `kernel`'s includes `sasl`. I get why they are, each of those are potentially used by modules in the application that includes them as `runtime_dependencies`, but that sounds like optional dependencies to me while `runtime_dependency` makes me think "required to run".

The other issue is related to the definition but builds on it the fact that the community tools ignore the field. I think this can be confusing to users who might encounter it in the `.app` file.

josevalim:
Based on my code browsing, it feels like `runtime_dependencies` should be called `otp_dependencies` as it is a mechanism in which otp does resolution for its own apps.

Since `otp` has everything bundled into the same app, it needs its own resolution mechanism and this mechanism can be done at runtime. This is a contrast to tools like rebar/mix, that needs to do resolution in a build stage, because we need to get the packages from the network in the first place.

So a name like `otp_dependencies` may reflect better when it is used and the versioning schema it follows. Again, my understanding of this is quite limited, but I thought I would share my $.02 while we are on the topic. :)

tsloughter:
> I fail to see an argument to why, more than that you wish it wasn't there.

It just seems to be used for something different because of the use of adding version constraints for both optional and required applications. It is a useful attribute for patching otps but may be confusing to users when we require they only give atom name of optional dependencies while the field is also used in a different context with string names containing a version.

I know I discussed this at the time 17 came out with someone but I can't recall what came out of it. Maybe that discussion only ended with "don't use it"-- use for defining optional dependencies when for building releases with relx -- because it was potentially going to be removed since it was "experimental".

rickard-green:
@tsloughter 
> > I have a really hard time seeing that the name is confusing, since it _is_ the runtime dependencies of the application.
> 
> There are two issues, one may simply be having differing definitions of what a "runtime dependency" is. The `stdlib` `runtime_dependencies` list includes `compiler`, the `kernel`'s includes `sasl`. I get why they are, each of those are potentially used by modules in the application that includes them as `runtime_dependencies`, but that sounds like optional dependencies to me while `runtime_dependency` makes me think "required to run".
>

It makes me think of _all_ runtime dependencies. I don't think it is possible to get everyone to associate the expression exactly the same way.

> The other issue is related to the definition but builds on it the fact that the community tools ignore the field. I think this can be confusing to users who might encounter it in the `.app` file.

Yes, it would have been better with a solution that could have been used by more projects. It should be possible to extend it, and probably even make incompatible changes to it, in order to better fit the needs of other projects. Do you have any suggestions?

> > > I don't think this can be combined with the runtime_dependencies feature.
> >
> > I fail to see an argument to why, more than that you wish it wasn't there.
> 
> It just seems to be used for something different because of the use of adding version constraints for both optional and required applications.

This does not make sense to me. Why wouldn't you want version constraints on all your dependencies? If I'm going to use a part of the application that needs an optional dependency, I would very much like that to work as well.

> It is a useful attribute for patching otps but may be confusing to users when we require they only give atom name of optional dependencies while the field is also used in a different context with string names containing a version.
> 

I have a hard time seeing that it is confusing, but perhaps more convenient with the application name as an atom for some.

All of the above are reasons to why you don't like the `runtime_dependencies` entry, but not arguments for why we should not be able to utilize the `optional_applications` entry for runtime dependencies as well.

The only thing that I can see that would prevent `optional_applications` from being able to be utilized by `runtime_dependencies`, is if `optional_applications` would be required to be a subset of `applications`. That would, however, also imply that applications using the `applications` entry as documented would not be able to express all optional applications. Often this could be solved by adding all applications to the `applications` entry, but not in case you have a mutual dependency between applications, such as in the `ssl` and `inets` case. By this I think the whole idea just got a lot less attractive.

> I know I discussed this at the time 17 came out with someone but I can't recall what came out of it. Maybe that discussion only ended with "don't use it"-- use for defining optional dependencies when for building releases with relx -- because it was potentially going to be removed since it was "experimental".

The plan was not to remove it, but improve it. For example, so that optional dependencies could be expressed. This however did not get prioritized at the time.

ferd:
> This does not make sense to me. Why wouldn't you want version constraints on all your dependencies? If I'm going to use a part of the application that needs an optional dependency, I would very much like that to work as well.

Because for most of our users the versions are already covered by their build tool's config files (rebar.config, mix.exs) and the way the system fetches things using either a single OTP version's standard library or these files means there isn't going to be a hard requirement for any constraints.

So what you get instead is versions you have to track in many places, which can clash with each other (the versions resolved by the build tool are incompatible with those specified in the .app.src file), and which you can't easily override without patching the library since no such mechanism exists. A floor value on version is not useful the moment you start having multiple major version and they don't guarantee anything very useful. Hell, for the most part, versions don't tend to guarantee that much useful stuff because the parts of the APIs you call may or may not have changed according to what the version states.

At this point our users (especially newcomers) are already kind of annoyed having to understand why they need to specify deps in rebar.config and .app files. We ended up explaining this as the rebar.config files being compile-time deps and the .app files being the runtime deps, and now we're talking of adding another tuple for runtime_deps but that is _the actual real runtime deps_ and holy crap this is absolutely terrible ergonomics. 

If they're optional they should just be labelled optional, at least the intent will be clear without having to read documentation from the last few years just to get going.

> All of the above are reasons to why you don't like the runtime_dependencies entry, but not arguments for why we should not be able to utilize the optional_applications entry for runtime dependencies as well.

In my case, it's mostly that `applications` is already about runtime dependencies. They are needed at runtime, you put them there. There's one weird edge case and it's "app you don't want started but still need loaded" and today this workaround is done at the release level by declaring the dep there but taking it out of the app file. This is technically "a runtime dep that you only want loaded and not started" which may or may not be a safe way to use a given library.

Would a loaded-but-not-started-runtime-dep (or optional dep for that matter) be expected to work? We got extremely clear semantics around app-handling for releases: loaded, started. There is another set of semantics for dependencies: optional, meaning you may or may not need Wx for say observer to work, but if it's there it's expected to be started.

If we're debating clarifying semantics then we should probably do it in a way that is far less ambiguous regarding the actual runtime properties of the system. I would expect an optional dependency to be about whether the code must use it or not, not around whether it needs to be loaded or started. If it really needs to be loaded-but-not-started, then "runtime_dep" is still plenty ambiguous: it has to be ambiguous because that's the term or description we've been using for `applications` tuple the last few years and we don't agree on what it means in this thread.

Clearly, it has been ambiguous for people maintaining build tools, designing languages on the BEAM, and the OTP team since we can't agree, therefore I posit that it can't be simple for people who don't have the decades of experience we have in this thread.

> The only thing that I can see that would prevent optional_applications from being able to be utilized by runtime_dependencies, is if optional_applications would be required to be a subset of applications. 

At this point I don't know how it would interact with runtime_dependencies because I've been avoiding it since I couldn't understand what it was supposed to do and was not required by anything ever so I could just pretend it didn't exist.

> That would, however, also imply that applications using the applications entry as documented would not be able to express all optional applications. Often this could be solved by adding all applications to the applications entry, but not in case you have a mutual dependency between applications, such as in the ssl and inets case. By this I think the whole idea just got a lot less attractive.

I think that this is correct. If we wanted to express all optional applications in the application tuple I would likely advocate for something like `{applications, [a, {b, [start]}, {c, [load]}, {d, [optional, load]}, {e, [optional, start], {f, [optional]}}]}` which would logically expand to `{applications, [{a, [start]}, {b,[start]}, {c,[load]}, {d, [optional,load]}, {e, [optional,start]}, {f, [optional,start]}]}`. I don't care for the versions at this point because I believe the collection of acceptable versions should be defined way before runtime when you already packaged and built your release and made it ready to deploy.

Having ssl have an optional/load dependency on inets and inets have an optional/load dependency on ssl would let both be loaded first and make who-starts-before-who a soft dependency that can be broken safely. If inets had an optional/start dependency on ssl, then you know you could safely make ssl start before inets if a third application requires both to be started.

The problem is that any old OTP release trying to load an app using this specification format would instantly blow up for not being able to parse the `applications` value. This implies that to do the change in a backwards-compatible manner, the annotation must be put in a distinct field that current OTP versions can safely ignore without making the application itself unusable. 

Whether you want to use optional_applications or runtime_applications for that doesn't make that much of a difference but the format IMO: can't be a version; must specify load vs. start semantics; must specify optionality as a distinct factor from load vs. start.

Anything else will be ambiguous and unclear with regards to how releases work.

tsloughter:
> This does not make sense to me. Why wouldn't you want version constraints on all your dependencies? If I'm going to use a part of the application that needs an optional dependency, I would very much like that to work as well.

I wouldn't want them in the `.app` file. rebar3 and relx handle versions, and they do so in a way that would conflict if versions in the `.app` file were supposed to mean that you expect the release to error if there is a conflict. rebar3 chooses the first version found when traversing the tree of dependencies (the maven method of dependency resolution) and this does not give the same result as checking that the version is greater than any version listed in the `.app` file for that dependency.

tsloughter:
> must specify load vs. start semantics

Note that if this (the `Type` of the app currently defined in the `.rel` file) were to be added to the `.app` we'd have to define the precedence for what happens when there are conflicts where some app defines the dep as `load`, another has that same dep as `permanent` and another as `none`.

I think its fairly simple and straightforward to say `none < load < temporary < transient < permanent` -- just important to define.

rickard-green:
> > > > > I don't think this can be combined with the runtime_dependencies feature.
> > > >
> > > > I fail to see an argument to why, more than that you wish it wasn't there.
> > >
> > > It just seems to be used for something different because of the use of adding version constraints for both optional and required applications.
> >
> > This does not make sense to me. Why wouldn't you want version constraints on all your dependencies? If I'm going to use a part of the application that needs an optional dependency, I would very much like that to work as well.
>
> [Response from @ferd](https://github.com/erlang/otp/pull/2675#issuecomment-709279484)
> [Response from @tsloughter](https://github.com/erlang/otp/pull/2675#issuecomment-709389769)

Clarification: It does not make sense to me to apply version constraints, or any other type of constraints, on one sub-set of the dependencies _but not on the other_.



> > All of the above are reasons to why you don't like the runtime_dependencies entry, but not arguments for why we should not be able to utilize the optional_applications entry for runtime dependencies as well.
>
> In my case, it's mostly that applications is already about runtime dependencies. They are needed at runtime, you put them there.

That is however not how the `applications` entry is documented to be used. I'm not saying that you aren't allowed to use it the way you use it. I'm saying that you cannot ignore the fact that there are systems that don't use it that way.

> There's one weird edge case and it's "app you don't want started but still need loaded" and today this workaround is done at the release level by declaring the dep there but taking it out of the app file. This is technically "a runtime dep that you only want loaded and not started" which may or may not be a safe way to use a given library.

It would perhaps have been good not to allow mutual dependencies, but it was allowed when this was designed in the 90s. I don't know how much it has been used, but it has been around for a very long time, so I don't think we can forbid it now. I can imagine that splitting up applications to get rid of such dependencies can be quite a lot of work. This of course depends very much on the applications.


> > The only thing that I can see that would prevent optional_applications from being able to be utilized by runtime_dependencies, is if optional_applications would be required to be a subset of applications.
>
> At this point I don't know how it would interact with runtime_dependencies because I've been avoiding it since I couldn't understand what it was supposed to do and was not required by anything ever so I could just pretend it didn't exist.

The tools using `runtime_dependencies` can determine which dependencies are optional by also inspecting `optional_applications` and give information accordingly based on that information. The usage of `optional_applications` in combination with `runtime_dependencies` is in no way in conflict with the way you want to use `optional_applications`.

I think the discussion has diverted from what this PR is about. The important question, or at least the contended question, is whether `optional_applications` should be a subset of `applications` or not. If `optional_applications` must be a subset of `applications`, I don't see `optional_applications` as especially useful unless you also use the `applications` entry as you do. Are there are any reasons to why `optional_applications` should/must be a subset of `applications`?

lib/kernel/doc/src/app.xml
```diff
@@ -148,10 +149,17 @@ ApplicationVersion = string()</code>
...
+          <c>optional_applications</c>, then the application
+          is not required to exist (but if it exists, it is
+          also guaranteed to be started before this one).</p>
+        <p><c>systools</c> uses this list to generate correct start
+          scripts. Defaults to the empty list, but notice that all
+          applications have dependencies to (at least) Kernel and STDLIB.</p>
+      </item>
+      <tag><c>optional_applications</c></tag>
+      <item>
+        <p>A subset of <c>applications</c> that are optional.</p>
```
josevalim, line 162:
```suggestion
        <p>A list of <c>applications</c> that are optional.
          Note if you want an optional dependency to be
          automatically started before the current application
          whenever it is available, it must be listed on both
          <c>applications</c> and <c>optional_applications</c>.</p>
```

josevalim (commented):
>  Are there are any reasons to why optional_applications should/must be a subset of applications?

My understanding of the situation is the following:

1. Because Mix/Rebar relies exclusively on applications, there is little benefit for `optional_applications` to not be a subset of `applications`

2. Because OTP also relies on runtime_dependencies, there is little benefit if `optional_applications` is only a subset of `applications`

In any case, `optional_applications` does not have to be a subset of `applications`.

To support optional applications in this PR we only need to amend the docs. See my suggestion below for the necessary changes. If you agree with the change, I will merge it, squash it, and update the commit message.

tsloughter:
> Are there are any reasons to why `optional_applications` should/must be a subset of `applications`?

I believe the sole reason it was decided to be a subset was to allow the user to specify startup order. Since all applications in `.app` files are sorted with a  stable sort when creating the release that order in `applications` is important.

Restricting it to a subset, I think, just would make it easier to understand for the user, but it isn't something required by the solution beyond that, the remaining applications can just be appended to the list.

tsloughter:
> That is however not how the `applications` entry is documented to be used. I'm not saying that you aren't allowed to use it the way you use it. I'm saying that you cannot ignore the fact that there are systems that don't use it that way.

To be clear, this is still basically the case with `applications` use in rebar3 and mix.

The reason we push the idea of putting all runtime deps in there is it simplifies the concept for users and since we don't allow cycles there isn't a time that you would hit an issue with defining an app in the `applications` list. Even if you don't want it started you can put it in the list but then have `{theapp, load}` in the release configuration.

Leaving an app out of `applications` but it still being used and wanting it in the release is not blocked, so we basically use `applications` entry the same way as it is documented, but we encourage users to have all apps an application uses to "start" before it, meaning put in `applications`, and partly do this encouraging and simplification of the concept by referring to it as the "runtime dependencies".

And with rebar3 optional dependencies are currently done by leaving the app out of `applications` and then adding it to the release. Well that or using a `.app.src.script` file, but we don't even document that I don't think since we prefer declarative configuration.

IngelaAndin:
I think that the application is not really optional if you have to start it. At least for the cases I know of the reason to have them as optional is that a user that does not use the optional application feature should not need to have the optional application at all, and hence it can not start it. The user that wants the optional application will have to make sure that it is started.

josevalim:
> The user that wants the optional application will have to make sure that it is started.

This is fine by me. All of our optional applications are started by hand by the user today. The main goal of this patch is to define the application ordering so releases load any optional application before the current one.

Therefore, an alternative approach would be to introduce an `release_applications` field, which is a superset of `applications`, which defines the ordering of applications in releases, and it may include optional ones. I can rework on this direction if everyone agrees that's the preferred approach.

IngelaAndin:
@josevalim  Why is it important to start an optional  application  before the  current one ?  I think  I must  have missed something!

josevalim:
Imagine you have application `foo` that may use libraries `bar` or `baz`. One of them have to be available for `foo`, so in `foo`'s start/2 callback I want to start one of them and configure the rest of my system based on them.

This works fine in development but it often fails on releases because the release algorithm does not guarantee `bar` or `baz` will be available before `foo`. I need them to be loaded before `foo` in the release so the code in `start/2` for `foo` still works.

I don't need Erlang to start them for me, but I need the release ordering to include them before `foo`. The work around is for me to explicitly configure my release ordering and include `bar` or `baz` before `foo`.

However, if I ship `foo` as a library for other developers, then users of my library need to be aware and solve this dependency issue as well. Most often they forget about it and what happens is that they end-up with a broken release in production, which leads to frustration.

ferd:
I think Wx is a good example of an application that is optional _and_ must be started to work.let's use something like observer where there is both a local and a remote section to it.

On my server without any GUI toolkit, I would expect the GUI to be optional. I don't include Wx, never start it, never even load it, but the code within observer handling all the node introspection is in place and working.
On my client with the GUI toolkit, I need Wx and want it started as a dependency. I then call over distributed Erlang to get the data required to display.

As such it sounds reasonable to expect that an optional application may require to be started. It also sounds confusing that in this one specific case the application is included but not started and the starting needs to be manual. The latter isn't much better than what we have today with the release-based final configuration.

lib/kernel/doc/src/app.xml
rickard-green: I'm satisfied with this change

tsloughter:
Why would a new field be added? Just allow applications in `optional_applications` that aren't in the `applications` list. Then they get appended and its essentially not caring about them being booted. But they will still `start`.

If the desire is to be able to set `load` in an application for a dependency that is another story and requires handling the merging of that when conflicting configs exists between applications, which is why it is nice to just leave that up to the release definition.

tsloughter:
`opentelemetry` is also a good example.

If only the `opentelemetry_api` applications is included then nothing is recorded, it is all noops. But if `opentelemetry` is included then it needs to be booted before everything else so that it is ready to handle starting and ending spans before any application that uses it.

At this time we simply document that you need to include it in the release list of apps before your applications that use opentelemetry https://github.com/open-telemetry/opentelemetry-erlang#including-in-release

If any app that used it could put it in their `applications` list but also in the `optional_applications` list they could ensure it is started without configuring it in the release's list of applications (though optionally setting it to `temporary` would still be documented for users who want to keep it from ever taking down their node, but that would be separate from boot order, so easier to explain and document.)

IngelaAndin:
OK, I understand that it could be important to start an optional application if it exists. And as I understand it  
that  is what will happen, but it will not fail if the optional application does not exist  which is good for the persons that does not want that application at all.

We also need to consider the case of circular optional dependencies, which I can agree on should be avoided, but they do exist and might not be so easy to resolve (as inets (optional HTTPS) and ssl (optional CRL handling with HTTP),  a required start will not work in this case.  There are probably other such cases as this is something that was not previously disallowed, even if it is in rebar case.

lib/kernel/test/application_SUITE.erl
```diff
@@ -1037,6 +1037,59 @@ ensure_all_started(_Conf) ->
     ok = application:unload(app_chain_error),
     ok.
 
+optional_applications(_Conf) ->
+    {ok, Fd10} = file:open("app10.app", [write]),
```
garazdawi, line 1041: I don't know why, but for some reason I think this testcase is interacting with the code_SUITE:big_boot_embedded and making it fail. The below is the log form big_boot_embedded.

```
=== Started at 2020-10-27 02:35:10



app10: File not found: "app10.app"



*** CT Error Notification 2020-10-27 02:35:11.032 ***
code_SUITE:create_big_boot failed on line 1424
Reason: {badmatch,error}

=== WARNING: Attempt to stop a nonexisting slavenode (big_boot_embedded@artemis)
===          Trying to kill it anyway!!!
```

garazdawi:
There are a bunch of testcase failures in sasl related to this change, mostly in systools_SUITE and release_handler_SUITE.

josevalim:
Thanks @garazdawi, I will investigate.

garazdawi:
I also suspect system:upgrade_SUITE may fail as a result of this change, though it does not fail on all our machines, so I'm not 100% sure.

```
*** CT Error Notification 2020-10-27 04:11:29.734 ***
upgrade_SUITE:upgrade_system failed on line 235
Reason: {badmatch,error}

=== Ended at 2020-10-27 04:11:29
=== Location: [{upgrade_SUITE,upgrade_system,235},
              {upgrade_SUITE,upgrade_test1,144},
              {test_server,ts_tc,1754},
              {test_server,run_test_case_eval1,1263},
              {test_server,run_test_case_eval,1195}]
=== === Reason: no match of right hand side value error
  in function  upgrade_SUITE:upgrade_system/6 (upgrade_SUITE.erl, line 235)
  in call from upgrade_SUITE:upgrade_test1/3 (upgrade_SUITE.erl, line 144)
  in call from test_server:ts_tc/3 (test_server.erl, line 1754)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1263)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 1195)
```

garazdawi:
Removed from testing while waiting for testcases to be fixed.

lib/kernel/test/application_SUITE.erl
josevalim: I have addressed this one locally. I forgot to unload app10 after a failed application start on the added test. Thanks!

josevalim:
@garazdawi just a quick update. I am unable to reproduce the sasl failures locally, I am investigating why that is the case.

garazdawi:
I don't know if it makes any difference in this case, but it is always good to test sasl using a released Erlang/OTP and not the source tree. i.e. `make release` and use `release/x86....*/bin/` as your erl when testing.

josevalim:
This is where [the bug was](https://github.com/erlang/otp/pull/2675/files#diff-132af868f46343bd340a5f62c55093cf2cd1fe124b0020c69be594de720201e9R669)! The reason it passed locally is because apparently I had fixed it 10 days ago but I forgot to push. :D

All tests pass for me now except for 5 tests in systools because they can't find "lib/kernel-7.1". I assume this is exactly because I was running the tests from `$ERL_ROOT/bin/erl`. I have tried the approach described by @garazdawi by doing this:

```
$ make release
$ cd releases/x...
$ ./Install `pwd`
$ cd ../test/test_server
$ ../../x.../bin/erl
erl> ts:install().
erl> ts:run(sasl, [batch]).
```

But unfortunately the suite never started. I assume I messed up the installation somehow.

garazdawi:
You may need to add it to the PATH for the test to catch the correct `erl`.

I'll re-add this to the tests tomorrow in order to get one night without this patch to more easily see which testcases fail because of it.

josevalim:
> You may need to add it to the PATH for the test to catch the correct erl.

It has ran now! We are down to one failure in `release_handler_SUITE:client1` with reason `{badmatch,{badrpc,{'EXIT',{disallowed_function_call,{trace,<...}`. It seems to be unrelated to this PR.

josevalim:
Hi! :wave: I have rebased the PR to latest master. Please let me know if there is anything I can do to move this forward. Thank you!

sverker:
@josevalim  I have rebased your single commit on top of a fresh master and force pushed it to jv-opt-apps in your repo. (You had done a merge of master into topic branch which we usually try to avoid).

I will put it into test again and if no more issues merge it.

PR #2680, opened 2020-07-01
Prepare for 64-bit pids

sverker:
Today Erlang process identifiers are restricted to 32 bits (including 4 bits type info) even on 64-bit architectures.

This PR is the first step in an upgrade path toward 64-bit pids. The local node (OTP-24) will still only generate 32-bit pids but we accept 64-bit pids from other nodes.


vans163:
This is nice but there are problems with spawning many pids, like allocation contention.  Often noticeable when taking an app on a  8core system to 64/128 cores, and not getting a horizontal speedup.

okeuday:
@sverker The external term format doesn't need to change to support 64bit pids because `NEW_PID_EXT` already exists, right?

sverker:
@vans163 Our main purpose for 64-bit pids is to practically eliminate the problem of pid reuse by wraparound.

sverker:
@okeuday Yes, NEW_PID_EXT already contains space for 64 bits. The PR just accepts all bits to be used by removing checks for too large values (for external pids) and makes sure all 64 bits are stored.

crownedgrouse:
Could a 64bit pid format include the flavor of a Pid? Io server or process etc...? This wait we could guard on pid input, while is_pid/1 still available.

PR #2681, opened 2020-07-01
sys_core_alias: Bail out when expressions get too deep

jhogberg:
https://bugs.erlang.org/browse/ERL-1297

### DIFF
--- a/lib/compiler/src/sys_core_alias.erl
+++ b/lib/compiler/src/sys_core_alias.erl
@@ -284,24 +284,31 @@ nodes_to_alias(Kind, Inner, Anno, Node, Keys0) ->
 %% Builds the key used to check if a value can be
 %% replaced by an alias. It considers literals,
 %% aliases, variables, tuples and cons recursively.
+%%
+%% We bail out after an arbitrary amount of nodes
+%% as very complex expressions may take forever to
+%% process otherwise.
+-define(DEPTH_LIMIT, 100).
 nodes_to_key(Kind, Nodes) ->
-    nodes_to_key(Nodes, [], Kind).
+    ntk_1(Nodes, [], Kind, ?DEPTH_LIMIT).
 
-nodes_to_key([#c_alias{var=Var}|T], Acc, Kind) ->
-    nodes_to_key([Var|T], Acc, Kind);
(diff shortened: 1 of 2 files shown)


lib/compiler/src/sys_core_alias.erl
```diff
@@ -284,24 +284,31 @@ nodes_to_alias(Kind, Inner, Anno, Node, Keys0) ->
 %% Builds the key used to check if a value can be
 %% replaced by an alias. It considers literals,
 %% aliases, variables, tuples and cons recursively.
+%%
+%% We bail out after an arbitrary amount of nodes
+%% as very complex expressions may take forever to
+%% process otherwise.
+-define(DEPTH_LIMIT, 1000).
```
josevalim, line 291: Awesome, thank you!

I am wondering if the depth should be smaller. This pass requires a pattern to exist that matches the (sub)tree. If you have a large depth, it is unlikely that you have a pattern matching that tree. At the same tree, if you have a very deep tree and only a tiny subtree matches it, the cost of not allocating that subtree is very minor compared to allocating the rest of the tree. So I would say a depth of 100 should be enough for all practical purposes?

jhogberg: > I am wondering if the depth should be smaller. [...] So I would say a depth of 100 should be enough for all practical purposes?

Good point, 1000 is far too cautious. I can't see any difference between 20 (!) and 1000 when compiling OTP, so 100 ought to be enough.

josevalim: I see! I wouldn't honestly mind making it 20 as well. If I reduce the reported tree to depth of 100, the pass still takes 50ms (which is still relatively high), while 20 is only 3ms. Although I guess none of those are really common enough to matter? :)

josevalim:
@jhogberg actually, I just thought of another optimization. In the code I have shared with you, we actually don't have any substitution, because we don't match in any tuple or cons at any moment. So we are building those keys for no purpose at all. It seems we can skip this traversal in the first place. I will send a PR soon. :)

josevalim:
PR here: #2683 - I have also improved the docs so I struggle less the next time around. :P

PR #2683, opened 2020-07-01
Only compute vars and keys if we have patterns

josevalim:
The goal of the sys_core_alias pass is to optimize
code like this:

    foo({ok, Val}) ->
      {ok, Val}.

into:

    foo({ok, Val} = Var) ->
      Var.

To do so, we need to find all patterns and find all
tuples and lists that may mirror their patterns.

Previously, we would compute the keys for each data
structure even if no pattern was stored. This optimizes
it so we only compute the keys if we have a pattern
so far.

### DIFF
--- a/lib/compiler/src/sys_core_alias.erl
+++ b/lib/compiler/src/sys_core_alias.erl
@@ -43,12 +43,28 @@
 -include("core_parse.hrl").
 
 -define(NOTSET, 0).
+-define(HAS_SUBS(Sub), Sub#sub.t =/= none).
 
 -record(sub, {p=#{} :: #{term() => ?NOTSET | atom()},              %% Found pattern substitutions
               v=cerl_sets:new() :: cerl_sets:set(cerl:var_name()), %% Variables used by patterns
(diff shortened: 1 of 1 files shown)


lib/compiler/src/sys_core_alias.erl
```diff
@@ -43,13 +43,29 @@
...
+%%    we update the existing sub instead of replacing it
+%% 3. sub is the previous substitution - this is used when
+%%    we replace the current sub
+%% 4. clause is used when there are clauses with patterns
+%%
+-type temp() ::
+        none
+        | {temp, temp()}
+        | {sub, sub()}
+        | {clause, [term()], [term()], cerl_sets:new(), temp()}.
```
jhogberg, line 67: Thanks for the PR!

`dialyzer` is unhappy with `cerl_sets:new()`. Travis caught this but the notifications are broken for some reason, sorry about that. :(

```suggestion
        | {clause, [term()], [term()], cerl_sets:set(cerl:var_name()), temp()}.
```

josevalim:
@jhogberg fixed!

jhogberg:

PR #2684, opened 2020-07-02
Support for openssh ed25519, rsa and dsa private keys

dotsimon:
Add support for openssh private keys of type rsa & dsa.
Cleaned up the code & comments for ed25519 key handling
Removed unused function `pubkey_ssh:new_openssh_encode/1`


HansN:
I must admit that I don't understand why you do this PR.

In OTP-23.0 the ticket OTP-15434 implemented the ssh-specific key knowledge from public_key into ssh_file and ssh_message in the SSH app. It was also refactored and extended.
The now for OTP obsolete ssh-specific code in public_key was left as-is because it is documented, but the intention - which you can't know - is to deprecate it in a future release, maybe OTP-24.
The experimental code for handling the new openssh format (openssh-key-v1) in the internal non-published module pubkey_ssh was just left to be removed when I have time.

So I wonder if you just were unaware of what I wrote above, or have a deeper thought or use case that is not obvious for me.

dotsimon:
This problem originated on a production OTP 22 system.

Consider a module that implements the `ssh_client_key_api` and handles openssh key blobs but is not limited to files and/or the naming & directory restrictions of ssh_file. Currently `public_key:pem_decode/1`, `public_key:pem_decode/1,2` & `public_key:ssh_decode/2` provide ways to extract the keys. That this is implemented in an internal non-published module is irrelevant if it is accessed via a public documented module.

I looked at the new refactored and extended code in the ssh app you mentioned. Amusingly it is quite similar to this code.

Since you seem upset with this PR perhaps instead you will consider exporting `ssh_file:decode_ssh_file/4` which would be sufficient for my purpose.

Ultimately if the goal is to deprecate all the ssh/openssh key handling from the public_key API then surely you will want to expose more of the new code via similar functions in the the ssh API.

HansN:
Hi,
I have merged and pushed an exported version of  `ssh_file:decode_ssh_file/4` to maint and master.  It will be available (without documentation or release note) in OTP-23.1 planned for mid September.

The exported one has limitations, so PRs for further needs are welcome!

PR #2685, opened 2020-07-04
do not bind pseudo key for delete char

velimir:
consider to use a key representation directly as the `<backspace>` can be
occupied by other modes (e.g. as a leader key for a modal keybinding
mode)


kjellwinblad:
I think this PR looks alright. Can you just clarify why (kbd "DEL") and [(backspace)] does not always translate to the same thing? I have found that (kbd "DEL") should represent the backspace key but I have not been able to find what [(backspace)] should represent.

IngelaAndin:
ping

velimir:
The root of this, as to my understanding, is that `backspace` is a key on a keyboard, that is bound to function. `(kbd "DEL")` is a control symbol `^?` (keystorke, that in treated in terminals as a symbol that is supposed to delete the previous character). `backward-delete-char-untabify` should be called when there's an intention to delete the character, rather than just bound to a key that is under a default configuration is supposed to delete the character.

Previously, the map was bound to `backspace` key, that is not necessarily is supposed to delete a previous character, but could be bound to a different action (e.g. leader key for a [modal keybinding](https://github.com/xahlee/xah-fly-keys)).

PR #2695, opened 2020-07-24
Update copyright year

ChristianTovar:
Updated the copyright year to `2020`


jhogberg:
Thanks for the effort, but we want the copyright year to reflect when the files last changed (or in the case of generated files, their source file).

I'm closing this PR as we update them ourselves when making new releases.

PR #2696, opened 2020-07-25
Fix stateless ticket window check problem IR-1316

mmzeeman:
The function checking the window assumed a tuple was passed, but it was passed as an integer.

Sorry I did not include a test case. I really don't know how to set that up for the ssl app. There are so many moving parts it is scary. I checked this PR with Chrome and Firefox, which can connect via TLS 1.3 with stateless tickets with replay detection. Both can correctly connect and use stateless tickets.


mmzeeman:
Sorry, wrong branch

PR #2697, opened 2020-07-25
Fix stateless ticket window check. ERL-1316

mmzeeman:
The window is passed as an integer, not as tuple.

Sorry I did not include a test with this PR. There are so many moving parts in ssl. I checked this PR by using Chrome and Firefox and connect it to an erlang web server using TLS 1.3 with stateless tickets with bloomfilter replay detection.

### DIFF
--- a/lib/ssl/doc/src/using_ssl.xml
+++ b/lib/ssl/doc/src/using_ssl.xml
@@ -424,12 +424,12 @@ ssl:connect("localhost", 9999,
     is set as <em>current</em>.
     </p>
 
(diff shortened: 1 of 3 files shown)


peterdmv:
The change looks good! Can you please add this testcase to the commit?
https://gist.github.com/peterdmv/08b3afc9a11c59689df6f3cf26da5824

mmzeeman:
Added the test. Very nice to see how this is done.

peterdmv:

PR #2699, opened 2020-07-28
Fix css problem with leftnav

GuillaumeMilan:
Avoid documentation overlapping strange behavior

*Old behavior*
image

*New behavior*
image

### DIFF
--- a/lib/erl_docgen/priv/css/otp_doc.css
+++ b/lib/erl_docgen/priv/css/otp_doc.css
@@ -37,6 +37,7 @@ a:visited      { color: #1b6ec2; text-decoration: none }
 }
 
 #leftnav {
+  background: #fefefe;
   position: fixed;
   float: left;
   top: 0;


GuillaumeMilan:
Hello can I close this?
What does mean stalled? And why some css stay stalled for 1 year?

IngelaAndin:
Stalled means that due to other more high prioritized things nobody has had time to investigate if we want to include this PR or not. This does not mean it is not good or we do not want it. It could perhaps help if you add some more information on where you experienced this problem (what browser, OS etc).  And maybe also an explanation of your solution, what is obvious to you might not be obvious to others.  I do have a vague memory of one of my coworkers looking into this a while back but I think it ended in that he could not reproduce it, and then got bogged down with other issues. I do apologies if this has somewhat fallen into oblivion.

IngelaAndin:
Also we do not think it is evident from the picture that the bottom is an improvement to the top picture. So steps in how to reproduce would be welcome.

lhoguin:
On the top picture the text from the content would overlap the menu on the left is you continued scrolling. I have not seen this problem in quite some time though so I too am curious as to the OS/browser combination that has this issue.

GuillaumeMilan:
Hello, 
I was on Chrome on Mac OSX.

What I am currently seeing on the online documentation is that the problem has been solved in an other way which is better (by removing the horizontal scroll.)

I will close the pull request as it's not required anymore. (and is taking my fork's master branch).

PR #2704, opened 2020-08-04
win32: Use QPC timer on Windows

dgud:
If resolution is good enough use QueryPerformanceTimer on windows,
which with thread compensation gives an ~0.1ms resolution instead of 15ms.

The cost is ~30% higher for QPC than for GetTickCound64
and the thread compensation cost adds another 0.5% amortized cost.

### DIFF
--- a/erts/emulator/sys/win32/sys_time.c
+++ b/erts/emulator/sys/win32/sys_time.c
@@ -161,7 +161,7 @@ os_monotonic_time_qpc(void)
         temp = (ErtsMonotonicTime) pc.QuadPart;
     } while(!(temp & SKIP));
 
-    return temp & (0xFFFFFFFFFFFFFFFF-SKIP);
+    return temp & (ERTS_I64_LITERAL(0xFFFFFFFFFFFFFFFF)-SKIP);
 }
 
 static void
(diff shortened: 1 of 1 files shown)


erts/emulator/sys/win32/sys_time.c
```diff
@@ -141,15 +141,27 @@ SystemTime2MilliSec(SYSTEMTIME *stp)
...
+     * are monotonic over several threads, reduce resolution to ensure
+     * that we return monotonic clocks even on multiple threads.
+     */
+    do {
+        if (!(*internal_state.r.o.pQueryPerformanceCounter)(&pc))
+            erts_exit(ERTS_ABORT_EXIT, "QueryPerformanceCounter() failed\n");
+        temp = (ErtsMonotonicTime) pc.QuadPart;
+    } while(!(temp & SKIP));
+
+    return temp & (0xFFFFFFFFFFFFFFFF-SKIP);
```
rickard-green, line 164: Use the `ERTS_I64_LITERAL()` macro on large integer literals, so you get the correct literal suffix for the specific compiler. That is `ERTS_I64_LITERAL(0xFFFFFFFFFFFFFFFF)` in this case.

PR #2705, opened 2020-08-04
Inline always required layer in raw_file_io

josevalim:
The raw option is always available to raw_file_io:open/2
but the previous implementation made it look like it was
optional as the other layers. This commit inlines
raw_file_io_raw into raw_file_io so it is clear it always
ends up by calling prim_file:open/2.


josevalim:
@jhogberg, I was a bit confused about when match_raw would ever return false but it seems that it is never the case, so I wrote this PR. However, if you don't think this isn't an improvement, just go ahead and close it. :D

jhogberg:
Merged, thanks for the PR!

PR #2707, opened 2020-08-06
Do not consider dotfiles to be an extension

josevalim:
Previously, filename:extension(".gitignore") would
return ".gitignore" as an extension while ".gitignore"
is actually part of the filename. Similarly,
filename:rootname("foo/.gitignore") would return "foo/".

One argument for .gitignore to be treated as filename
is that we should expect this property to hold:

    length(filename:split(Path)) =:=
        length(filename:split(filename:rootname(Path)))

But that was violated for dotfiles prior to this commit.

Another point of discussion is filenames such as "..ext".
This commit takes the approach that the rootname is "."
and the extension is ".ext".

### DIFF
--- a/lib/stdlib/src/filename.erl
+++ b/lib/stdlib/src/filename.erl
@@ -398,22 +398,31 @@ extension(Name) when is_binary(Name) ->
 	[] ->
 	    <<>>;
 	List ->
-	    {Pos,_} = lists:last(List),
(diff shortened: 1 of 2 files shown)


bjorng:
The test case `filename_SUITE:extension_bin/1` fails.

josevalim:
Apologies, it should be fixed now.

PR #2700, opened 2020-07-29
Update config scripts to compile for Apple Silicon (arm64)

markpeek:
Detect Apple Silicon (arm64) during configuration for compilation on macOS 11.0 (beta at this time). Fixes ERL-1305.

Main changes:
- Fix the integer macOS version correctly for the build version check
- Map correctly to 64bit builds (-m64)

Verified via `./otp_build tests`

Note: the current config.guess uses "arm-apple-darwin*" for the build/host system type while a newer version uses "aarch64-apple-darwin*". However, the diffs between 2015-03-04 to 2020-07-12 were outside the scope of being able to test for this PR.

PR #2708, opened 2020-08-07
Fix: Variadic function cast causing segfaults on Apple ARM64

tehprofessor:
## Overview

Howdy!

This is a fix for segfaults happening on Apple ARM64. The segfaults are caused by differences in how Apple ARM64 treats variadic functions. x86 will only use the stack when necessary, after filling the available registers; while, 🍎 ARM64 will always places variadic arguments onto the stack regardless of many registers are available. The segfault manifests as what look like errors in bif functions (e.g. `is_integer_1`) and do not leave an erl crash dump (only available logs were from the system's crash report).

This was originally developed against to `master` (otp-24) but has been back-ported here to `maint`. The updated build configuration provided by #2700 is required so it's currently included in this branch.

I'm running the tests now and will get back with the results when they finish (in the morning).

FWIW, so far, Cowboy's test suite\[1] and `cowlib` passed all tests without error on master (`otp-24`). Using _this_ branch (based off `maint`) with Elixir's `master-otp-23` (from asdf I'll get the elixir `sha` later) Phoenix's test suite passed all tests without error.

## Issue

The problem itself (seems) to boil down to variadic casts happening when calling bifs in `erts/emulator/beam/erl_db_util.c - db_prog_match`:

```c
bif = (Eterm (*)(Process*, ...)) *pc++;
```

`bif` points to functions from `erts/emulator/beam/erl_bif_op.c` which have a fixed number of arguments e.g.

```c
BIF_RETTYPE is_integer_1(BIF_ALIST_1)
{
    if (is_integer(BIF_ARG_1)) {
	BIF_RET(am_true);
    }
    BIF_RET(am_false);
}
```

Casting the fixed arguments to be variadic then causes the registers to be incorrectly marshaled and the VM segfaults.

## The Fix

All of the functions in `erts/emulator/beam/erl_bif_op.c` are defined using a set of macros defined in `erts/emulator/beam/bif.h` with the relevant parts being:

```c
#define BIF_RETTYPE Eterm

#define BIF_P A__p

#define BIF_ALIST Process* A__p, Eterm* BIF__ARGS, BeamInstr *A__I

#define BIF_ALIST_0 BIF_ALIST
/* omitted for brevity */
#define BIF_ALIST_4 BIF_ALIST

#define BIF_ARG_1  (BIF__ARGS[0])
/* omitted for brevity */
#define BIF_ARG_4  (BIF__ARGS[3])
```

Using these macros (on 🍎 ARM64), we were then able to declare function pointers in `db_prog_match` matching the expected function signature:

```c
BIF_RETTYPE (*bif_call_1)(BIF_ALIST_1);
```

and then call it:

```c
/* omitted for brevity */
	case matchCall1:
             bif_call_1 = (BIF_RETTYPE (*)(BIF_ALIST_1)) *pc++;
             t = (*bif_call_1)(build_proc, esp-1, bif_args);
```

## Closing Remarks

I know, after battling these segfaults for a week. this is a very hot path of code execution , and I'll gladly admit I don't have the expertise required right now to know if there is a better approach for this fix or if this poses its own set of problems. This has been way out of my comfort zone (but damn it felt good! I'll be back for more); so, I'd like to emphasize I'm _very_ receptive to feedback. The change in the PR as of opening it, is the simplest and most obvious (?) solution I could muster... and so that's exactly what's here.

I have a gigantic amount of symbolicated crash logs, I'd normally attach to help others without the platform see the issues, but... I'm not sure how much I can and can't share since I'm bound to the 🍎 developer kit NDA. I'm going to ask in their forums later if I can share some of the crash logs here or via email (if Apple needs that) for posterity.

Finally I'm happy, and waiting 😄 , to rebase after #2700 can be backported to `maint` as that's a dependency for this patch to work... Thank you for time!

<sub>\[1] I'm not really sure how cowboy's test suite ran as it _looks_ (based on the output) like it needs `go` and I _definitely_ do not have a working instance of `go` on the machine.

Co-Authored-By: mjc <mjc@kernel.org>

### DIFF
--- a/erts/emulator/beam/erl_db_util.c
+++ b/erts/emulator/beam/erl_db_util.c
@@ -2031,7 +2031,7 @@ Eterm db_prog_match(Process *c_p,
     Process *tmpp;
     Process *current_scheduled;
     ErtsSchedulerData *esdp;
-    Eterm (*bif)(Process*, ...);
+    BIF_RETTYPE (*bif)(BIF_ALIST);
     Eterm bif_args[3];
     int fail_label;
 #ifdef DMC_DEBUG
@@ -2338,8 +2338,8 @@ restart:
             *esp++ = t;
             break;
 	case matchCall0:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
-	    t = (*bif)(build_proc, bif_args);
+	    bif = (BIF_RETTYPE (*)(BIF_ALIST)) *pc++;
+	    t = (*bif)(build_proc, bif_args, NULL);
 	    if (is_non_value(t)) {
 		if (do_catch)
 		    t = FAIL_TERM;
@@ -2349,8 +2349,8 @@ restart:
 	    *esp++ = t;
 	    break;
 	case matchCall1:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
-	    t = (*bif)(build_proc, esp-1);
+	    bif = (BIF_RETTYPE (*)(BIF_ALIST)) *pc++;
+	    t = (*bif)(build_proc, esp-1, NULL);
 	    if (is_non_value(t)) {
 		if (do_catch)
 		    t = FAIL_TERM;
@@ -2360,10 +2360,10 @@ restart:
 	    esp[-1] = t;
 	    break;
 	case matchCall2:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
+	    bif = (BIF_RETTYPE (*)(BIF_ALIST)) *pc++;
 	    bif_args[0] = esp[-1];
 	    bif_args[1] = esp[-2];
-	    t = (*bif)(build_proc, bif_args);
+	    t = (*bif)(build_proc, bif_args, NULL);
 	    if (is_non_value(t)) {
 		if (do_catch)
 		    t = FAIL_TERM;
@@ -2374,11 +2374,11 @@ restart:
 	    esp[-1] = t;
 	    break;
 	case matchCall3:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
+	    bif = (BIF_RETTYPE (*)(BIF_ALIST)) *pc++;
 	    bif_args[0] = esp[-1];
 	    bif_args[1] = esp[-2];
 	    bif_args[2] = esp[-3];
-	    t = (*bif)(build_proc, bif_args);
+	    t = (*bif)(build_proc, bif_args, NULL);
 	    if (is_non_value(t)) {
 		if (do_catch)
 		    t = FAIL_TERM;


erts/emulator/beam/erl_db_util.c
```diff
@@ -2338,8 +2348,13 @@ Eterm db_prog_match(Process *c_p,
...
 	case matchCall0:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
-	    t = (*bif)(build_proc, bif_args);
+#if defined(__APPLE__) && defined(__aarch64__)
+        bif_call_0 = (BIF_RETTYPE (*)(BIF_ALIST_0)) *pc++;
+        t = (*bif_call_0)(build_proc, esp-1, bif_args);
+#else
+        bif = (Eterm (*)(Process*, ...)) *pc++;
+        t = (*bif)(build_proc, bif_args);
+#endif
```
mikpe, line 2357: This and the other similar changes should be unconditional.  It is _NEVER_ correct to call a C function via a function pointer type that differs from its original declaration (you can use such casts for transporting the function pointer, but any call must see the correct type).  Abusing this tends to work in older ABIs because pre-C89 K&R C didn't have full prototypes, but newer ABIs are free to handle variadic functions differently from non-variadic ones, as you've dicovered.

Many thanks for working on this.

tehprofessor: Howdy!

Thank you very much for the suggestion and information! Also a big thanks for y'all getting the PR this was dependent on back ported and merged so damn quickly! 🤘 As I was making the original change, I couldn't figure out why it was done that way previously; so, I'm glad to know it was mostly just abusing older ABIs and not a deeper rabbithole. 

The changes have been made and are now unconditional. I've given it ago on my x86 and ARM64 systems and it works great. Hopefully the change as it is now, is what you had in mind. Please let me know if there's anything else I can do! 

Lastly, I'm still waiting for a good test run on ARM64, not because of BEAM, but because the system will randomly restart (there haven't been any segfaults in the system crash logs though 🎉 ).

erts/emulator/beam/erl_db_util.c
```diff
@@ -2031,7 +2031,10 @@ Eterm db_prog_match(Process *c_p,
     Process *tmpp;
     Process *current_scheduled;
     ErtsSchedulerData *esdp;
-    Eterm (*bif)(Process*, ...);
+    BIF_RETTYPE (*bif_call_0)(BIF_ALIST_0);
+    BIF_RETTYPE (*bif_call_1)(BIF_ALIST_1);
+    BIF_RETTYPE (*bif_call_2)(BIF_ALIST_2);
+    BIF_RETTYPE (*bif_call_3)(BIF_ALIST_3);
```
mikpe, line 2037: See erts/emulator/beam/bif.h.  It should be enough to change the type of "bif" to ErtsBifFunc (they all look the same as the parameters are passed in an array nowadays).

erts/emulator/beam/erl_db_util.c
```diff
@@ -2338,8 +2341,8 @@ Eterm db_prog_match(Process *c_p,
             *esp++ = t;
             break;
 	case matchCall0:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
-	    t = (*bif)(build_proc, bif_args);
+	    bif_call_0 = (BIF_RETTYPE (*)(BIF_ALIST_0)) *pc++;
+	    t = (*bif_call_0)(build_proc, esp-1, bif_args);
```
mikpe, line 2345: bif_args should be the second parameter, not the third.  The third parameter is normally the beam instruction pointer, which most bifs won't need, I _think_ it might work to just pass NULL there.

erts/emulator/beam/erl_db_util.c
```diff
@@ -2374,11 +2375,8 @@ Eterm db_prog_match(Process *c_p,
 	    esp[-1] = t;
 	    break;
 	case matchCall3:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
-	    bif_args[0] = esp[-1];
-	    bif_args[1] = esp[-2];
-	    bif_args[2] = esp[-3];
-	    t = (*bif)(build_proc, bif_args);
+	    bif_call_3 = (BIF_RETTYPE (*)(BIF_ALIST_3)) *pc++;
+	    t = (*bif_call_3)(build_proc, esp-1, bif_args);
```
mikpe, line 2379: No, you need to preserve the argument copy into bif_args, and pass bif_args as the second parameter.

erts/emulator/beam/erl_db_util.c
bjorng: Yes, passing in NULL will work. None of the BIFs that can actually be called will use it.

erts/emulator/beam/erl_db_util.c
bjorng: Yes, because the elements in esp[] are in the wrong order. The same goes for `matchCall2`. The third argument should be NULL.

erts/emulator/beam/erl_db_util.c
tehprofessor: Thank you! I'll get on these next round of changes.

erts/emulator/beam/erl_db_util.c
```diff
@@ -2349,8 +2349,8 @@ Eterm db_prog_match(Process *c_p,
 	    *esp++ = t;
 	    break;
 	case matchCall1:
-	    bif = (Eterm (*)(Process*, ...)) *pc++;
-	    t = (*bif)(build_proc, esp-1);
+	    bif = (BIF_RETTYPE (*)(BIF_ALIST)) *pc++;
+	    t = (*bif)(build_proc, esp-1, NULL);
```
tehprofessor, line 2353: I think I've got everything updated. The one thing I'm not entirely sure of is if this _should_ be copied into `bif_args`, if this is an optimization for this _specific_ case, or something else?  I'm new to `c`; so, if I'm understanding this correctly, `esp-1` works _here_ without copying into `bif_args` because `esp-1` and `bif_args[0]` essentially point to the same value and both are pointers so semantically it's all good? Then by not copying into `bif_args[0]` we save an allocation/step?

And, please feel free to tell me to just _google_ more 😄 !

Thank you both for your time reviewing this, and please let me know if there's anything else I can change, improve, or have missed 😅  here.

bjorng: The copying of the arguments to `bif_args` when there are 2 or 3 arguments is only necessary because the arguments would be in the wrong order if we would pass `esp-2` or `esp-3`, respectively. When there is a single argument, there is no need to copy it because it can't be in the wrong order.

bjorng:
Many thanks for working on this.

I have added this pull request to our daily builds. If everything goes well, we will merge it within a few days.

erts/emulator/beam/erl_db_util.c
tehprofessor: Thank you @bjorng

bjorng:
Thanks for your pull request.

tehprofessor:
Just wanted to say "thank you" @bjorng and @mikpe for the code reviews and feedback! I really appreciate you both taking the time to do so.

PR #2709, opened 2020-08-09
Remove private ram_file:compress in favor of zlib:gzip

josevalim:
A bit more background: I am planning on rewriting ram_file to
use NIFs as a learning exercise and part of the initial work is
on cleaning the existing API. During this process, I have noticed
that ram_file:compress is only used by beam_lib and the
functionality is already available as zlib:gzip/gunzip.

Therefore I decided to remove the private compress/uncompress
functionality in favor of the public zlib API.


lib/stdlib/src/beam_lib.erl
```diff
@@ -906,19 +906,15 @@ beam_filename(File) ->
...
-    {ok, Fd} = ram_file:open(Binary0, [write, binary]),
-    {ok, _} = ram_file:compress(Fd),
-    {ok, Binary} = ram_file:get_file(Fd),
-    ok = ram_file:close(Fd),
-    Binary.
+uncompress(Binary) ->
+    try
+	zlib:gunzip(Binary)
+    catch
+	_:_ -> Binary
```
josevalim, line 913: Maybe this should just fail if an uncompressed file is given but for this pull request I decided to preserve the previous behaviour which would simply return the binary.

jhogberg:
Thanks for the PR!

The build fails as `erts/emulator/Makefile.in` still lists `gzip.o`, can you remove those lines?

josevalim:
@jhogberg oops, it should be good now!

jhogberg:

jhogberg:
Sorry, but we had to revert this as it introduced a build error on `master`, which went undetected as another branch in our test runs had removed `erts_gzinflate_buffer` from `beam_load.c`.

Copying `erts_gzinflate_buffer` into `beam_load.c` works fine but there's an issue with `beam_lib`. We get the following error after updating primary bootstrap and rebuilding:

```
 GEN    asn1ct_rtt.erl
{"init terminating in do_boot",{{badmatch,{error,beam_lib,{not_a_beam_file,'asn1rtt_check.beam'}}},[{prepare_templates,abstract,1,[{file,"prepare_templates.erl"},{line,98}]},{prepare_templates,'-gen_asn1ct_rtt/1-lc$^1/1-1-',1,[{file,"prepare_templates.erl"},{line,36}]},{prepare_templates,gen_asn1ct_rtt,1,[{file,"prepare_templates.erl"},{line,36}]},{init,start_em,1,[]},{init,do_boot,3,[]}]}}
init terminating in do_boot ({{badmatch,{error,beam_lib,{not_a_beam_file,asn1rtt_check.beam}}},[{prepare_templates,abstract,1,[{_},{_}]},{prepare_templates,-gen_asn1ct_rtt/1-lc$^1/1-1-,1,[{_},{_}]},{p
```

josevalim:
@jhogberg apologies for the confusion, I will investigate.

I have some questions:

1. Is it better to copy `erts_gzinflate_buffer` or make it use the underlying `zlib` NIF API?

2. If `erts_gzinflate_buffer` has already been removed from another branch, should I wait until said branch is merged instead?

jhogberg:
> If erts_gzinflate_buffer has already been removed from another branch, should I wait until said branch is merged instead?

The branch removes the use in `beam_load.c` but leaves it in `ram_file_drv.c`. The plan is to merge the branch in September (during/after the BEAM conf) so if you're not in a hurry you can leave it until then.

> Is it better to copy erts_gzinflate_buffer or make it use the underlying zlib NIF API?

If you don't want to wait it's easiest just to copy `erts_gzinflate_buffer`, as the aforementioned branch already does the latter and we'd get a merge conflict otherwise.

josevalim:
If you give me an option to procrastinate, I will take it. :D So I will wait until september and resubmit this! Thanks a lot!

PR #2711, opened 2020-08-10
Remove unused and private ram_file functions

josevalim:
This commit removes the following unused private APIs:

  * ram_file:open/2 with deprecated file modes
    (it is now equivalent to raw_file_io:open/2)

  * ram_file:set_file/2 and ram_file:get_file_close/1
    which are private and not used anywhere in Erlang/OTP

  * ram_file:uuencode/1 and ram_file:uudecode/2 which
    which are private and not used anywhere in Erlang/OTP

Note that ram_file:get_size/1 also fits the criteria above
but it was kept because there is some usage in the wild.

### DIFF
--- a/erts/emulator/drivers/common/ram_file_drv.c
+++ b/erts/emulator/drivers/common/ram_file_drv.c
@@ -40,12 +40,12 @@
 
 /* other operations */
 #define RAM_FILE_GET           30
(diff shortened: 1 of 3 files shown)


jhogberg:
Merged, thanks for the PR!

PR #2713, opened 2020-08-11
Use github actions instead of travis

garazdawi:
Travis seems to have stopped working for some reason that I cannot figure out, so that gave me the opportunity to experiment with github workflows instead.

There are still a couple of things that need to be migrated from travis, but most of the functionality (and more) is there.


prebuild.skip
```diff
@@ -1,4 +1,4 @@
 Makefile
 bin
 bootstrap/bin
-ibin
+ibin
```
bjorng, line 4: Missing newline.

PR #2714, opened 2020-08-17
Name cipher suites consistently across ssl docs

britto:
They are referenced as "cipher suites" most of the time, but as "cipher suits" in a few places. This change turns these remaining occurences into "cipher suites". Also fixes typo: chipher -> cipher.

### DIFF
--- a/lib/ssl/doc/src/notes.xml
+++ b/lib/ssl/doc/src/notes.xml
@@ -3578,7 +3578,7 @@
         <item>
           <p>
 	    Add support for PSK (Pre Shared Key) and SRP (Secure
-	    Remote Password) chipher suits, thanks to Andreas
+	    Remote Password) cipher suites, thanks to Andreas
 	    Schultz.</p>
           <p>
 	    Own Id: OTP-10450 Aux Id: kunagi-269 [180] </p>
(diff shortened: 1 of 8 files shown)


IngelaAndin:
Thank you for fixing these typos, could you do the change on the maint branch instead of master. 
I think it needs to be a rebase --onto.  As then the documentation can be improved already for OTP-23.1 and not only in OTP-24. This can be considered a bug fix. 

You could fix your own typo too, in the comment,  "cypher suites"  -> "cipher suites"

lib/ssl/doc/src/notes.xml
```diff
@@ -3578,7 +3578,7 @@
         <item>
           <p>
 	    Add support for PSK (Pre Shared Key) and SRP (Secure
-	    Remote Password) chipher suits, thanks to Andreas
+	    Remote Password) chipher suites, thanks to Andreas
```
garazdawi, line 3581:
```suggestion
	    Remote Password) cipher suites, thanks to Andreas
```

britto: Good catch! 👀

britto:
Thanks, @IngelaAndin! Will do.

britto:
Done. I've changed base to `maint` and also fixed other chipher -> cipher typos.

PR #2715, opened 2020-08-19
supervisor: Eliminate a subtle internal state corruption

bjorng:
When a temporary child of a `simple_one_for_one` supervisor died, the  internal state of the supervisor would be corrupted in a way that would cause the supervisor to retain the start arguments for subsequent children started by the supervisor, causing unnecessary growth of the supervisor's heap.
    
This bug was introduced in 59d418205486.
    
Thanks to Zeyu Zhang (@zzydxm) for noticing this bug and to Maxim Fedorov for making it known to me.


max-au:
Thank you! 
Actually, this issue was found & reported by @zzydxm (Zeyu Zhang), I merely passed it over.

PR #2717, opened 2020-08-20
Change application set/unset env timeouts to infinity

josevalim:
There has been a recent report where starting multiple
processes (8) to perform concurrent httpc requests in
a Raspberry Pi Zero would trigger set_env timeouts.

That's because httpc would call crypto which would call
application:set_env with a cache of the existing curvers
and that would make the aplication_controller unable to
reply in time.

One option to address this issue would be to pass the
timeout option when calling set_env from crypto but
reallistically speaking we would most likely just move
the error to next set_env call that may happen.

Given all functions in application already default to
a timeout of infinity, I chose to fix the problem at
the source and set the default timeout to infinity.

This also aligns with more recent trends in Erlang/OTP
that use an infinity timeout as default.

### DIFF
--- a/lib/kernel/doc/src/application.xml
+++ b/lib/kernel/doc/src/application.xml
@@ -247,10 +247,10 @@ Nodes = [cp1@cave, {cp2@cave, cp3@cave}]</code>
           each application individually, except it is more efficient.
           The given <c><anno>Config</anno></c> is validated before the
           configuration is set.</p>
-        <p><c>set_env/2</c> uses the standard <c>gen_server</c> time-out
-          value (5000 ms). Option <c>timeout</c> can be specified
-          if another time-out value is useful, for example, in situations
-          where the application controller is heavily loaded.</p>
+        <p><c>set_env/2</c> uses the default <c>timeout</c> of <c>infinity</c>.
+          Option <c>timeout</c> can be specified if another time-out value
+          is useful, for example, in situations where the application controller
+          is heavily loaded.</p>
         <p>Option <c>persistent</c> can be set to <c>true</c>
           to guarantee that parameters set with <c>set_env/2</c>
           are not overridden by those defined in the application resource
@@ -312,10 +312,10 @@ Nodes = [cp1@cave, {cp2@cave, cp3@cave}]</code>
       <desc>
         <p>Sets the value of configuration parameter <c><anno>Par</anno></c> for
           <c><anno>Application</anno></c>.</p>
-        <p><c>set_env/4</c> uses the standard <c>gen_server</c> time-out
-          value (5000 ms). Option <c>timeout</c> can be specified
-          if another time-out value is useful, for example, in situations
-          where the application controller is heavily loaded.</p>
+        <p><c>set_env/4</c> uses the default <c>timeout</c> of <c>infinity</c>.
+          Option <c>timeout</c> can be specified if another time-out value
+          is useful, for example, in situations where the application controller
+          is heavily loaded.</p>
         <p>If <c>set_env/4</c> is called before the application is loaded,
           the application environment values specified in file <c>Application.app</c>
           override the ones previously set. This is also true for application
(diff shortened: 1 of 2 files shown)


zzantozz:
[Infinity is a bad default timeout](https://techblog.bozho.net/infinity-bad-default-timeout/). Infinite timeouts are the root of hard-to-diagnose issues that leave you with nothing but silence in the logs. Nobody ever really wants to wait infinitely for a computer to do something. It needs to fail at some point and say why and where it failed.

dmorneau:
@zzantozz FWIW the OTP team has expressed disagreement with that line of thought in the past: http://erlang.org/pipermail/erlang-questions/2010-August/052636.html

josevalim:
While I agree infinity can have its downsides, I would be nice to discuss if the use of infinity in this particular case can be problematic, otherwise it is easy to get side tracked into a bigger discussion. Given we don’t use any external resources or call any other process, this change should be pretty safe.

zzantozz:
@dmorneau I'm fine disagreeing with the OTP team. I stand by my opinion, based on experience and shared with others in our field, that infinity is almost never the right timeout to use, much less to use as a default. Many hours of my life have been wasted over this exact thing in multiple languages. It was like a breath of fresh air to come to Erlang and find that most things actually had a non-infinite timeout by default because I could be fairly certain that an unusual thing like a hung TCP connection wouldn't just hang up a process forever without giving any hint as to why. Making timeouts infinite makes things harder for the users of the language.

zzantozz:
@josevalim I would say it's problematic because it's using infinity as a timeout. It isn't reasonable to wait an infinite amount of time to set or unset a config parameter. If the goal is to set it to something outrageously long, how about an hour? Would you agree there's definitely a problem if it takes that long to change a config parameter? If not, how about a day? The point is that infinity never closes the loop. Providing a timeout gives a definite point at which something will crash and tell you, "This thing right here failed." With infinity, you never get that.

essen:
> While I agree infinity can have its downsides, I would be nice to discuss if the use of infinity in this particular case can be problematic, otherwise it is easy to get side tracked into a bigger discussion. Given we don’t use any external resources or call any other process, this change should be pretty safe.

It's never safe. The process code could have a bug and get stuck. The VM could have a bug and make the process get stuck. The VM could have a bug and prevent the process from receiving the message. You cannot predict bugs. A timeout has the advantage of being visible when an issue like this occurs, unlike when `infinity` is used.

RabbitMQ has been moving away from `infinity` timeouts precisely for these reasons.

For this particular call I would say a 1 or 5 minutes timeout would be large enough. If the `application_controller` doesn't respond in 5 minutes you've got trouble either way.

josevalim:
I am not particularly convinced by the "what if the VM has bugs" angle because, if that's the premise, then we cannot even assume that timeouts will work as expected. The operations performed here are trivial and if they are taking too long or failing inexplicably, you have far deeper problems.

In any case, practically speaking 5 minutes is the same as infinity, so I am fine with it too. Someone just let me know what is the preferred choice and I will change it accordingly.

okeuday:
While I hate infinity timeouts for the same reasons mentioned previously and made sure [CloudI](https://github.com/CloudI/CloudI) service requests help you never have infinity for a timeout value, I understand the tendency for people to feel the need of infinity timeout values in Erlang when trying to simplify the source code (trying not to think about the possibility of the timeout and only focus on the sequential logic, i.e., often trying to avoid the timeout exception that would require an impure catch).  The use of infinity with the `application_controller` is safe, i.e., avoiding deadlock because the `application_controller` should always spawn a separate process for anything that attempts a `set_env` or `unset_env`.

However, if you have a timeout with the `application_controller` doing a `set_env` that takes longer than 5000 milliseconds, that should be a problem to solve elsewhere that the developer would want to know about.  For example, it could be container execution with no cpu quota set, having a scheduler get stuck due to container execution.  5 seconds is a long time and ignoring the problem is much worse than solving the problem, so I prefer keeping the default timeout as 5000.

josevalim:
@okeuday i am not running into a theoretical problem though. This timeout is being triggered on a Raspberry Pi Zero when doing 8 httpc requests in parallel. I could reduce the concurrency here but  that would drastically affect performance because the Pi Zero can handle the I/O concurrency just fine, the issue is just the initial setup of crypto.

What do you suggest then as a solution that won’t degrade the performance of the system?

okeuday:
@josevalim My view is that usage of `application:set_env` in `crypto` is the problem.  `application:set_env` should not be used (it is good the docs have a warning) unless it is setting application configuration that may only be determined at runtime (env state is best for static state from config files and the .app files).  I haven't looked into the details, but added concurrency shouldn't be a problem (8 processes instead of 1 process shouldn't make `crypto` calling `application:set_env` slow).  If the `application:set_env` needs to be used for runtime state, I think the `application` behavior start function is the best place to use it.  If the `application_controller` is the bottleneck due to unavoidable concurrency it would be better to create a separate `ets` table for the data.

essen:
> I am not particularly convinced by the "what if the VM has bugs" angle

It's not a "what if". It happened recently. It'll happen again.

> we cannot even assume that timeouts will work as expected

Yes, it's possible that a timeout never triggers as well. It's never safe. But there's more chances to have the timeout trigger if it's not set to `infinity`. :-)

PR #2718, opened 2020-08-21
stdlib: fix lists:sublist/3 if Start>length(L)+1 

shantiii:
lists:sublist/3 fails when Start (the 2nd argument) was greater than
len(List)+1, due to nthtail/2 throwing an exception. This change
replaces the use of nthtail/2 with a recursive function definition.

This fixes [ERL-1334](https://bugs.erlang.org/browse/ERL-1334).

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -341,8 +341,12 @@ max([],    Max)              -> Max.
       Len :: non_neg_integer(),
       T :: term().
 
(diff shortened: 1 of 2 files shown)


rickard-green:
Please change the first line of the commit message to something that  explains the change instead of referring to the issue at bugs.erlang.org. Otherwise it looks good. I'll merge it once you've fixed the commit message.

shantiii:
@rickard-green thanks for the review; I've pushed up an updated commit message and updated the PR title as well.

PR #2720, opened 2020-08-21
Cache crypto:supports(curves) on application start

josevalim:
Otherwise the ssl app may start the crypto app indirectly
as part of its dependencies without a curves cache, which
can lead to crypto:supports/0 being called concurrently,
considerably slowing down the machine performance.

### DIFF
--- a/lib/crypto/src/Makefile
+++ b/lib/crypto/src/Makefile
@@ -37,6 +37,7 @@ RELSYSDIR = $(RELEASE_PATH)/lib/crypto-$(VSN)
 
 MODULES= \
 	crypto \
+	crypto_app \
 	crypto_ec_curves
 
 HRL_FILES=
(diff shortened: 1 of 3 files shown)


IngelaAndin:
@josevalim  I think maybe the question about the start function should apply to all OTP application that has such functions. application:ensure_all_started/1 was not around when these utility functions where first created, but I have always felt they are a little misleading as they are just sort of connivance functions for developers and not the way applications will be started in a release.

HansN:
I close this PR, but we have noted your problem and arguments, and we have started an internal discussion.  We agree that the correct way is to use application:start(X) and not X:start(), which is only a help for developers.

In the case of the crypto application it is a bit special. Supervisors and a crypto_app.erl is a bit strange since crypto is a library module without a server in the Erlang/OTP meaning.
I agree that crypto has a state, but that is down in cryptolib and in the NIF, and is hard to handle from a supervisor at the Erlang level.

For the problem at hand, I have made a prototype and written an internal ticket on moving all of the curve's cache down into the NIF.  It will be initialized when the NIF is loaded (for example by code:load_file(crypto)) as the rest of the NIF is.  This will make synchronization handled by the NIF load mechanism if more than one process simultaneously make a  initial call to crypto.

Thanks for raising those problems!

PR #2721, opened 2020-08-22
Allow metadata and ignore to be computed in logger funs

josevalim:
This commit allows the logger metadata to be
computed lazily via the logger `msg_fun`.
The `msg_fun` itself can also return `ignore`.

These features are useful when computing the
metadata or if the message should be logged or
not is expensive and you want to do so only
after the log level match.

### DIFF
--- a/lib/kernel/doc/src/logger.xml
+++ b/lib/kernel/doc/src/logger.xml
@@ -450,12 +450,44 @@ logger:error("error happened because: ~p", [Reason]). % Without macro
       <type variable="FunArgs" name_i="4"/>
       <type variable="Metadata"/>
       <desc>
-        <p>Log the given message.</p>
+        <p>Create a log event at the given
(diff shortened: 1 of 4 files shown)


josevalim:
Hi @garazdawi, do you have any concerns about moving this forward?  I am drafting the changes we will do on Elixir's Logger in the next version and this will help inform us which direction to go. Thank you!

hauleth:
@josevalim could you point me to that document? Maybe I could help with that.

garazdawi:
I had forgotten about this.

I don't see any problems with this, will talk with the others at OTP to make sure that they don't have any objections.

garazdawi:
Putting into test for inclusion in OTP-24.

garazdawi:
I rebased your commit on latest master and also added a commit expanding the documentation a bit. I'll let this run one more night in tests and then merge.

garazdawi:
Merged! Thanks!

PR #2729, opened 2020-08-27
Adjust the shell path in installation scripts for Android

JeromeDeBretagne:
This patch replaces /bin/sh into /system/bin/sh in the Release target for Android to adjust the various Erlang/OTP release scripts. I've updated the Android documentation accordingly.

I'm not totally sure to see the purpose of the local_setup target in erts/Makefile when cross compiling for Android but I've adjusted it as well.

Cheers, Jérôme

### DIFF
--- a/HOWTO/INSTALL-ANDROID.md
+++ b/HOWTO/INSTALL-ANDROID.md
@@ -45,7 +45,8 @@ Use the following instead when compiling a 32-bit version.
 
 ### Compile Erlang/OTP ###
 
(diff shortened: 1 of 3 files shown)


garazdawi:
I don't suppose that there is an environment variable that we can use to find that `sh` is not in its standard location?

JeromeDeBretagne:
I haven't found one. This fix is the best way I've come up so far to adjust the path for Android, previously I was doing it manually...

PR #2730, opened 2020-08-27
Updated documentation for application:ensure_all_started

netrome:
Before this change, the documentation incorrectly describes the
case of running application:ensure_all_started for an application
which is started, but has unstarted dependencies.

### DIFF
--- a/lib/kernel/doc/src/application.xml
+++ b/lib/kernel/doc/src/application.xml
@@ -67,22 +67,25 @@
   </datatypes>
   <funcs>
     <func>
-        <name name="ensure_all_started" arity="1" since="OTP R16B02"/>
-        <name name="ensure_all_started" arity="2" since="OTP R16B02"/>
-        <fsummary>Load and start an application and its dependencies, recursively.</fsummary>
-        <desc>
-          <p>Equivalent to calling
-	  <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
-          repeatedly on all dependencies that are not yet started for an application.</p>
-          <p>Returns <c>{ok, AppNames}</c> for a successful start or for an already started
-          application (which is, however, omitted from the <c>AppNames</c> list).</p>
-	  <p>The function reports <c>{error, {AppName,Reason}}</c> for errors, where
-	  <c>Reason</c> is any possible reason returned by
-	  <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
-	  when starting a specific dependency.</p>
-	  <p>If an error occurs, the applications started by the function are stopped
-	  to bring the set of running applications back to its initial state.</p>
-        </desc>
+      <name name="ensure_all_started" arity="1" since="OTP R16B02"/>
+      <name name="ensure_all_started" arity="2" since="OTP R16B02"/>
+      <fsummary>Load and start an application and its dependencies, recursively.</fsummary>
+      <desc>
+        <p>Equivalent to calling
+        <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
+            repeatedly on all dependencies that are not yet started for an
+            application that is not yet started.</p>
+        <p>Returns <c>{ok, AppNames}</c>, where <c>AppNames</c> is a list of the application names
+            that was actually started by this call.
+            The list might be empty, or not contain all dependencies if the application
+            or some of its dependencies are already started.</p>
+        <p>The function reports <c>{error, {AppName,Reason}}</c> for errors, where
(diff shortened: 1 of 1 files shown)


lib/kernel/doc/src/application.xml
```diff
@@ -73,9 +73,10 @@
         <desc>
           <p>Equivalent to calling
 	  <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
-          repeatedly on all dependencies that are not yet started for an application.</p>
+          repeatedly on all dependencies that are not yet started for an
+          application that is not yet started.</p>
           <p>Returns <c>{ok, AppNames}</c> for a successful start or for an already started
-          application (which is, however, omitted from the <c>AppNames</c> list).</p>
+          application (in which case AppNames is always an empty list).</p>
```
ferd, line 79: I think one of the things here is that the original docs (which I think I wrote?) refers to two things:

1. starting an application that is already started returns an empty list
2. a dependency that was already started will be omitted from the list

I would say something like "Returns `{ok Appnames}` for a successful start. Any application that was already started is omitted from the list of applications"

netrome: What do you think about "Returns {ok Appnames} for a successful start. Any application that was already started or is a dependency of an application that is already started remains unchanged by this operation" - to make it clear that this does not start unstarted dependencies of a started application?

ferd: I don't think this is that significant because the application would refuse to start without its own dependencies already being started. By definition, if app A depends on B and A is started, B has to have been started before for it to start.

The one edge case is that if A and B were both started and that then B has been manually shut down in a way that didn't crash A, then starting A (which still runs) will indeed not start B again.

netrome: Another case is that A is updated while running. The new version of A depends on C, not B. But C is not started. This is a significant case in my opinion, since this presents itself as the perfect function to call when you change dependencies to ensure that all dependencies are started.

IngelaAndin: How about "Returns a list of the application names that was actually started by this call, that is the list might be empty, or not contain all dependencies if the application and some or all of its dependencies where already started previously."

netrome: Sure, I would modify it to  "Returns `ok, AppNames`, where `AppNames` a list of the application names that was actually started by this call. The list might be empty, or not contain all dependencies if the application or some of its dependencies are already started." to have it more concise and accurate. What do you think?

lib/kernel/doc/src/application.xml
```diff
@@ -67,22 +67,25 @@
...
+      <fsummary>Load and start an application and its dependencies, recursively.</fsummary>
+      <desc>
+        <p>Equivalent to calling
+        <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
+            repeatedly on all dependencies that are not yet started for an
+            application that is not yet started.</p>
+        <p>Returns <c>{ok, AppNames}</c>, where <c>AppNames</c> a list of the application names
+            that was actually started by this call.
+            The list might be empty, or not contain all dependencies if the application
+            or some of its dependencies are already started.</p>
```
garazdawi, line 81:
```suggestion
            if some of its dependencies are already started.</p>
```

lib/kernel/doc/src/application.xml
```diff
@@ -67,22 +67,25 @@
...
-        </desc>
+      <name name="ensure_all_started" arity="1" since="OTP R16B02"/>
+      <name name="ensure_all_started" arity="2" since="OTP R16B02"/>
+      <fsummary>Load and start an application and its dependencies, recursively.</fsummary>
+      <desc>
+        <p>Equivalent to calling
+        <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
+            repeatedly on all dependencies that are not yet started for an
+            application that is not yet started.</p>
+        <p>Returns <c>{ok, AppNames}</c>, where <c>AppNames</c> a list of the application names
```
garazdawi, line 78:
```suggestion
        <p>Returns <c>{ok, AppNames}</c>, where <c>AppNames</c> is a list of the application names
```

garazdawi (commented):
Look good to me. Does anyone else have any more thoughts?

garazdawi:
I squashed and rebased your change to `maint` so that it is released in the next release.

Thanks for your contribution.

netrome:
Thanks. Happy to contribute :slightly_smiling_face:

PR #2732, opened 2020-08-28
otp_ring0.beam: remove from preloaded beam files

max-au:
PR-1825 renamed otp_ring0 into erl_init, however preloaded
otp_ring0 stayed. This commit removes unused file.


garazdawi:
thanks!

PR #2733, opened 2020-08-30
Default to Unicode for filename encoding on Android

JeromeDeBretagne:
UTF-8 is used for filename encoding on Android, for reference from the official documentation:

https://developer.android.com/reference/java/io/File#interoperability-with-java.nio.file-package
> On Android strings are converted to UTF-8 byte sequences when sending filenames to the operating system, and byte sequences returned by the operating system (from the various list methods) are converted to strings by decoding them as UTF-8 byte sequences.

Cheers,
Jérôme


garazdawi:
thanks!

PR #2735, opened 2020-09-01
Process aliases preventing late replies reaching clients

rickard-green:
Currently there exists no lightweight mechanism for preventing late replies from a server to a client after a timeout or connection loss has occurred. The only way to prevent late replies today is to make the request via a proxy process.

The proposed process alias feature is a lightweight mechanism that solves the above problem. A process alias is similar to a registered name that is used temporarily while a request is outstanding. If the request times out or the connection to the server is lost, the alias is deactivated which prevents a late reply from reaching the client.

This pull request serves as reference implementation for [EEP 53](https://github.com/erlang/eep/blob/master/eeps/eep-0053.md). See the EEP for more information.

Documentation of BIFs and BIF options introduced by this pull request:

* [`alias/0` and `alias/1`](https://erlang.org/~rickard/OTP-16718/erts-11.1.1/doc/html/erlang.html#alias-0)
* [`unalias/1`](https://erlang.org/~rickard/OTP-16718/erts-11.1.1/doc/html/erlang.html#unalias-1)
* [`monitor/3`](https://erlang.org/~rickard/OTP-16718/erts-11.1.1/doc/html/erlang.html#monitor-3)
* [`{monitor, MonitorOpts}` option of `spawn_opt()`](https://erlang.org/~rickard/OTP-16718/erts-11.1.1/doc/html/erlang.html#spawn_opt-4)
* [`{monitor, MonitorOpts}` option of `spawn_request()`](https://erlang.org/~rickard/OTP-16718/erts-11.1.1/doc/html/erlang.html#spawn_request-5)

The gen behaviours such as `gen_server` also have been modified to utilize the new alias feature. Due to this it is now also possible to implement `receive_response()` functionality similar to [`erpc:receive_response()`](https://erlang.org/~rickard/OTP-16718/lib/kernel-7.1/doc/html/erpc.html#receive_response-2) for these behaviours:

* [`gen_server:receive_response/2`](https://erlang.org/~rickard/OTP-16718/lib/stdlib-3.13.2/doc/html/gen_server.html#receive_response-2)
*  [`gen_statem:receive_response/2`](https://erlang.org/~rickard/OTP-16718/lib/stdlib-3.13.2/doc/html/gen_statem.html#receive_response-2)
*  [`gen_event:receive_response/2`](https://erlang.org/~rickard/OTP-16718/lib/stdlib-3.13.2/doc/html/gen_event.html#receive_response-2)

Updates:
* 29/10 2020 - Added the `tag` option of `monitor/3`
* 3/11 2020 - Added test cases, minor fixes of implementation, and minor doc fixes


lib/stdlib/src/gen.erl
```diff
@@ -198,17 +199,41 @@ call(Process, Label, Request, Timeout)
...
+        {Mref, Reply} ->
+            erlang:demonitor(Mref, [flush]),
+            {ok, Reply};
+        {'DOWN', Mref, _, _, Reason} ->
+            exit(Reason)
+    end;
+do_call(Process, Label, Request, Timeout) when is_atom(Process) =:= false ->
+    Mref = erlang:monitor(process, Process, [{alias,demonitor}]),
+
+    Tag = [alias | Mref],
```
josevalim, line 225: Is the compiler capable of tracking selective receives when the reference is stored in a variable that is then used in the `receive`? Or does the compiler need to be improved here? Also, do we need to teach the compiler about `monitor/3` when talking about selective receives?

rickard-green: I forgot to make the compiler aware of `monitor/3`.

rickard-green: Just pushed a fix of this

RoadRunnr:
Somewhat related, the two things related to this functionality that are missing are:
* having a way to access the monitor reference from inside the function spawned through spawn_monitor, the current mechanism of setting up a monitor and passing the reference into a message to that process only works with a already running processes. It would be use full to spawn_monitor a function and send results back from it using the monitor reference/alias,
* the ability to match in gen_{event,server,statem}:handle_info methods on the reference returned from send_request

rickard-green:
> Somewhat related, the two things related to this functionality that are missing are:
> 
> * having a way to access the monitor reference from inside the function spawned through spawn_monitor, the current mechanism of setting up a monitor and passing the reference into a message to that process only works with a already running processes. It would be use full to spawn_monitor a function and send results back from it using the monitor reference/alias,

We cannot change the behaviour of `spawn_monitor()` to always create an alias. If you want an alias with the monitor, you should use `{alias, UnaliasOpt}` as a `MonitorOpts` in `spawn_opt()` or `spawn_request()`.  After that it is the same scenario as if you were about to communicate with an already existing process.

* [`{monitor, MonitorOpts}` option of `spawn_opt()`](https://erlang.org/~rickard/OTP-16718/erts-11.0.3/doc/html/erlang.html#spawn_opt-4)
* [`{monitor, MonitorOpts}` option of `spawn_request()`](https://erlang.org/~rickard/OTP-16718/erts-11.0.3/doc/html/erlang.html#spawn_request-5)

I don't see any need for functionality to retrieve the alias used by the parent. As a comparison, you cannot get the process identifier of the parent from within the spawned function either.

> * the ability to match in gen_{event,server,statem}:handle_info methods on the reference returned from send_request

Perhaps I'm misunderstanding what you are saying, but I think that you are saying that you want to match for the response message corresponding to the `send_request()`. The API was intentionally designed so that you should *not* do that. This since we would then need to document internal details of the `gen` protocol which in turn would prevent changes to the protocol when such are needed. For example changes like the one introduced by this PR. The user should use `check_response()` in order to determine if a message is a response to a previous `send_request()`.

rickard-green:
@RoadRunnr 

> > 
> > * having a way to access the monitor reference from inside the function spawned through spawn_monitor, the current mechanism of setting up a monitor and passing the reference into a message to that process only works with a already running processes. It would be use full to spawn_monitor a function and send results back from it using the monitor reference/alias,
> 
> We cannot change the behaviour of `spawn_monitor()` to always create an alias. If you want an alias with the monitor, you should use `{alias, UnaliasOpt}` as a `MonitorOpts` in `spawn_opt()` or `spawn_request()`. After that it is the same scenario as if you were about to communicate with an already existing process.
> 
>     * `{monitor, MonitorOpts}` option of `spawn_opt()`
> 
>     * `{monitor, MonitorOpts}` option of `spawn_request()`
> 
> 
> I don't see any need for functionality to retrieve the alias used by the parent. As a comparison, you cannot get the process identifier of the parent from within the spawned function either.
> 

My comparison with the pid of the parent process is bad, since you can pass that pid via the arguments. In the `spawn_request()` case you typically also do not want to pass the alias using ordinary message passing since you then have to wait for the spawn reply before you can send it which in the distributed case will cause a substantial delay.

What you want to do is create the alias before the `spawn_request()` and pass the alias as an argument to the child process. The only issue with this is that you now have different references in response and DOWN message which will prevent the message queue optimization that can skip parts of the message queue. I'll soon make a modification to the PR and the EEP that address this issue.

rickard-green:
@RoadRunnr I've made an update of the EEP and the PR with the addition of a `tag` option to the `monitor/3` BIF in order to address this issue. See the [Why is the tag option of monitor/3 introduced?](https://github.com/erlang/eep/blob/master/eeps/eep-0053.md#why-is-the-tag-option-of-monitor3-introduced) section of the updated EEP.

RoadRunnr:
> @RoadRunnr I've made an update of the EEP and the PR with the addition of a `tag` option to the `monitor/3` BIF in order to address this issue. See the [Why is the tag option of monitor/3 introduced?](https://github.com/erlang/eep/blob/master/eeps/eep-0053.md#why-is-the-tag-option-of-monitor3-introduced) section of the updated EEP.

Very nice, that is exactly what I was looking for.

With the new functionality, would it be possible to make is possible to match on the RequestId returned from `send_request` ?

What I'm looking for is roughly like this, example of gen_statem async calling a gen_server

```
handle_event({call, From}, {forward, What}, _, #data{server = Server} = Data)
    ReqId = gen_server:send_request(Server, What),
    {keep_state, Data#data{req_id = ReqId};

handle_event(info,{[alias | ReqId], Reply}, _, #data{req_id = ReqId}) ->
            erlang:demonitor(Mref, [flush]),	            erlang:demonitor(Mref, [flush]),
            {ok, Reply};	            {ok, Reply};
        {'DOWN', Mref, _, _, noconnection} ->	        {'DOWN', Mref, _, _, noconnection} ->


 {update_session, Session, Events}, _State, _Data) ->
    ?LOG(debug, "SessionEvents: ~p~n       Events: ~p", [Session, Events]),
    Actions = [{next_event, internal, {session, Ev, Session}} || Ev <- Events],
    {keep_state_and_data, Actions};

rickard-green:
> 
> With the new functionality, would it be possible to match on the RequestId returned from `send_request` ?
> 

No, this is out of the scope of this change.

> What I'm looking for is roughly like this, example of gen_statem asynchronously calling a gen_server
> 
[code removed]
> 
> Exposing internal details like `[info|ReqId]` might not be desirable, but maybe that could be wrapped with a macro and put in a shipped included file?

I'm not in favor of a macro like that, since users needs to recompile their code in order to be up to date with changes in the internals of gen behaviors. I can check what people here at OTP think of it though.

rickard-green:
@RoadRunnr wrote:
> Somewhat related, the two things related to this functionality that are missing are:
>
[snip ...]
> *  the ability to match in gen_{event,server,statem}:handle_info methods on the reference returned from send_request

The PR #5792 improves handling of multiple outstanding request. It wont let you match on the request identifier (which I am strongly against), but the `check_response()` functions can now take a collection of request identifiers instead of just one request identifier which I think should solve your problem.

PR #2736, opened 2020-09-02
inets, stdlib: move rfc1123_date and related functions from httpd_uti…

ao-song:
…l to calendar

Change-Id: Id45e4004ad1e39cdd93ba181607dcaf3e22b2528


garazdawi:
Shouldn't the usage of the httpd_util functions first be deprecated before removed?

ao-song:
> Shouldn't the usage of the httpd_util functions first be deprecated before removed?

I think you are right, the functions should be deprecated first before totally moved to another module.

IngelaAndin:
You can move them and deprecate them at the same time. That is you must keep wrapper functions in httpd_util for now, but they can be implemented using the new function. And inets does not have to use the httpd_util  module internally.

josevalim:
Should we name the functions `calendar:weekday_abbreviation/1` and `calendar:month_abbreviation/1`? Or `calendar:weekday_abbr/1` and `calendar:month_abbr()`? I believe it will make it clearer that it is not returning the full names for those reading the code.

lib/dialyzer/test/r9c_SUITE_data/src/inets/httpd_response.erl
```diff
@@ -317,7 +317,7 @@ transform({content_type,undefined})->
     ["Content-Type:text/plain\r\n"];
 
 transform({date,undefined})->
```
IngelaAndin, line 319: @uabboli  Please comment on how you want to handle this test code.

uabboli: Modify the code any way you like to get rid of the warning(s).

ao-song:
Close this PR as we think need more investigation in a better date format both for http and calendar, the code will not be merged for now, thanks for your valuable comments!

PR #2740, opened 2020-09-03
Add os:env/0

sverker:
https://bugs.erlang.org/browse/ERL-1332

`os:env/0` returns all environment variables in a tuple list `[{VarName, Value}]`

 The same as the undocumented `os:list_env_vars/0`, but with a better name.

A nicer format than the existing os:getenv/0 which returns a list of strings `"VarName=Value"`.

### DIFF
--- a/erts/emulator/beam/bif.tab
+++ b/erts/emulator/beam/bif.tab
@@ -393,7 +393,6 @@ bif ets:match_spec_run_r/3
 bif os:get_env_var/1
 bif os:set_env_var/2
 bif os:unset_env_var/1
-bif os:list_env_vars/0
 bif os:getpid/0
 bif os:timestamp/0
 bif os:system_time/0
@@ -773,3 +772,4 @@ bif erts_internal:get_creation/0
 bif erts_internal:prepare_loading/2
 bif erts_internal:beamfile_chunk/2
 bif erts_internal:beamfile_module_md5/1
+bif os:env/0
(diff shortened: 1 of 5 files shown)


Stratus3D (approved):
Looks great!

fenollp:
A map would provide constant access. Why not return a map?

vans163:
> A map would provide constant access. Why not return a map?

Yea why not a map since there can never be 2 of the same key here, as far as I know.

sverker:
@fenollp @vans163
The typical use of os:env/0 is to iterate over all variables and that's easier to do with a list.
To lookup a particular variable you typically use the existing os:getenv/1.

galdor:
Having `env/1` with an option argument, and the `return_map` option would offer flexibility regarding that aspect. The name `return_map` is already used for example by `uri_string`, so it's not something new.

josevalim:
`os:getenv/1` is faster for lookups and should be generally preferred, as it won't allocate all keys and values. If `os:env/0` returns a map, it may mislead developers to write something like:

    maps:get("USER", os:env(), undefined)

which will be very expensive compared to `os:getenv/1`. So it makes sense for `os:env/0` to prioritize traversal, which a list is better suited for. if you really want a map, there is always `maps:from_list/1`.

I am not a big fan of using extra arguments to change the return type, I would personally prefer to use another name. However, we can always add an extra function (or even an extra argument) later. So I am :+1: for merging this as is.

sverker:
Undocumented `os:list_env_vars/0`  removed.

sverker:
OTP technical board decided to accept `os:env/0` as is, returning a list.

PR #2741, opened 2020-09-04
Fix deprecation warning message

ergl:
Found this while upgrading old code, correct replacement should be `rand:uniform/1`, the current warning points to a non-existant function.


HansN:
Could you please rebase it onto maint since it is a bug fix?

ergl:
Hi @HansN, all done, thanks for the help. When I opened the PR, I think the `maint` branch didn't have the wrong deprecation, but I can see now that it's there.

HansN:
Thanks for the PR!

PR #2743, opened 2020-09-07
Added wxTaskBarIcon:new/1 to create popupmenu

dominicletz:
With the new cocoa based wxWidgets backend for macos call PopupMenu( menu ) to create a taskicon menu is officially not supported anymore for mac. Thus the only way to create a popup menu is to provide an override for the "wxMenu* CreatePopupMenu()" virtual function.

This patch adds the ability to provide a callback function in the constructor which then servers as CreatePopupMenu() implementation.

I've also supplied a small taskbar sample app, to exemplify the usage.

Selection_307


dgud:
Thanks for the PR, I'm in the process of rewriting the code generator for wx (base it of interface files in 3.1) instead of "real" headers, so I will let this wait a while.

dominicletz:
That's great to hear! Would be awesome to have the constants be defined (or at least accessible) through a function, so they become accessible to other beam languages (Elixir in my case)

dominicletz:
Also another thing that I realized is that virtual impl callbacks like this one can create deadlocks when creating new wx objects. Not sure if that is fixable in your effort.

dgud:
> That's great to hear! Would be awesome to have the constants be defined (or at least accessible) through a function, so they become accessible to other beam languages (Elixir in my case)

I'll keep that in mind then, didn't know that you couldn't.

> Also another thing that I realized is that virtual impl callbacks like this one can create deadlocks when creating new wx objects. Not sure if that is fixable in your effort.

Can you email me, dgud at erlang org,  a short example or an explanation. 
The way the callbacks are handled today is to avoid those deadlocks :-(
You can't do calls to other processes from a callback, since the graphics thread is locked waiting for a response
from the callback.

dominicletz:
@dgud sorry didn't have time yet to reproduce the menu issue, hopefully will do that later. Regarding your rewrite of the generator, when do you expect that to go into otp? IMHO the current situation is rather buggy since macos is now the only OS you can't use the taskbaricon on. If your rewrite is longer term I would opt to include this as a bugfix until then. Even though the bug is in wx, for people trying to port their apps from windows or linux to macos it's a showstopper right now :-(
I don't even think there is any workaround possible without this patch, if you're aware of one let me know.

Would also be happy to make any changes to the calling interface for injecting the fun if that helps making the current solution more forward compatible.

dgud:
Ok, will take a better look at this for OTP-23.2 then.

The rewrite targets OTP-24 as it will be backwards incompatible in a couple (many) hopefully not used functions.

dgud:
Can you rebase this on maint, since we want to release this in the next maint release.

Also do we need a new example can it not be added to the existing menu example?

dominicletz:
Sure, can merge the examples and rebase

dominicletz:
here you go @dgud

PR #2744, opened 2020-09-08
beam_validator: Track the unit of saved bs match positions

jhogberg:
https://bugs.erlang.org/browse/ERL-1340

### DIFF
--- a/lib/compiler/src/beam_validator.erl
+++ b/lib/compiler/src/beam_validator.erl
@@ -203,7 +203,15 @@ validate_0([{function, Name, Arity, Entry, Code} | Fs], Module, Level, Ft) ->
          %%
          %% 'initialized' means we've saved a message position, and 'committed'
          %% means the next loop_rec instruction will use it.
-         recv_marker=none :: none | undecided | initialized | committed
+         recv_marker=none :: none | undecided | initialized | committed,
+         %% Holds the current saved position for each `#t_bs_context{}`, in the
+         %% sense that the position is equal to that of their context. They are
+         %% invalidated whenever their context advances.
+         %%
+         %% These are used to update the unit of saved positions after
+         %% operations that test the incoming unit, such as bs_test_unit and
+         %% bs_get_binary2 with {atom,all}.
+         ms_positions=#{} :: #{ Ctx :: #value_ref{} => Pos :: #value_ref{} }
         }).
 
 -type label()        :: integer().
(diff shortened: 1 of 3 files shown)


lib/compiler/src/beam_validator.erl
```diff
@@ -1403,21 +1405,37 @@ validate_bs_start_match({f,Fail}, Live, Slots, Src, Dst, Vst) ->
...
     CtxType = #t_bs_context{tail_unit=gcd(Stride, TailUnit)},
 
+    validate_bs_get_1(Op, Fail, Ctx, Live, CtxType, ValueType, Dst, Vst).
+
+validate_bs_get_all(Op, Fail, Ctx, Live, Stride, ValueType, Dst, Vst) ->
+    assert_type(#t_bs_context{}, Ctx, Vst),
+
+    %% Set the tail unit to the stride of the current operation to match the
+    %% SSA pass. This is safe as the tail unit is irrelevant at the very end.
+    CtxType = #t_bs_context{tail_unit=Stride},
```
jhogberg, line 1424: This is more of a band-aid than an actual fix. We need to track the units for saved positions, but that's a bit too complicated for an emergency patch. I'll try to get this fixed in OTP 23.2.

lib/compiler/test/beam_validator_SUITE.erl
```diff
@@ -808,6 +810,17 @@ mrt_1(Bool) ->
...
 
+%% ERL-1340: bs_get_binary2 didn't match the SSA pass when updating match
+%% context units on tail matches ({atom,all}).
+bs_get_binary_context_type(Config) when is_list(Config) ->
+    [<<0,1,2,3,4>>, <<5,6,7,8,9>>] = bgbct_1(id(<<0,1,2,3,4,5,6,7,8,9>>)),
+    [] = bgbct_1(id(<<>>)),
+
+    ok.
+
+bgbct_1(<<Bin/binary>> = Bin) ->
```
michalmuskala, line 821: Shouldn't something like this be rejected by the linter in the first place, though?

jhogberg: I suppose we could warn about it from OTP 24 onwards. As iffy as this seems it's still legal, and I'd rather not change it in a patch.

jechol: I'm original reporter of this issue for Elixir (https://github.com/elixir-lang/elixir/issues/10309) and I want explain why I happened to write code like that.

I need this iffy expression to include argument name generated by ex_doc.
If I only include `<<bin::binary>>`, then argument in document is named as `arg`.
And I prefer `<<bin::binary>>` pattern matching to `is_binary(bin)` guard.

bjorng: @jechol Would it work for the documentation if you would write the function head like this?

    def encode_bin(<<_::binary>> = bin) do

The compiler will generate efficient code for that.

jechol: @bjorng That's what I want. Thanks.

michalmuskala: I was specifically referring to variables on the left & right of patterns being "compared". I don't think it can ever succeed beyond this special case in binary where the pattern match is a no-op. Everywhere else a part of a data structure can't be equal to the data structure itself, e.g. `{X, _} = X` shouldn't even compile as it will always fail.

bjorng: @michalmuskala I think that `{X, _} = X` should produce a warning. That is what we do for other unmatchable patterns such as `[X,Y] = {X,Y}`. Our reasoning is that macros, parse transformations and naive code generators could produce nonsensical but working code.

lib/compiler/src/beam_validator.erl
```diff
@@ -2593,6 +2707,23 @@ mv_args([], _VsA, _VsB, Acc) ->
 merge_fragility(FragileA, FragileB) ->
     cerl_sets:union(FragileA, FragileB).
 
+merge_ms_positions(MsPosA, MsPosB, Vs) ->
+    Keys = if
+               map_size(MsPosA) =< map_size(MsPosB) -> maps:keys(MsPosA);
+               map_size(MsPosA) > map_size(MsPosB) -> maps:keys(MsPosB)
```
bjorng, line 2713: This is a matter of taste but I think that the code below is clearer:

```suggestion
               true -> maps:keys(MsPosB)
```

lib/compiler/test/beam_validator_SUITE.erl
```diff
@@ -808,6 +810,90 @@ mrt_1(Bool) ->
...
+                     {t_bs_context,12,0,0},
+                     {t_bs_context,8,0,0}}}}] = Errors,
+
+    [<<0,1,2,3,4>>, <<5,6,7,8,9>>] = bspu_1(id(<<0,1,2,3,4,5,6,7,8,9>>)),
+    [] = bspu_1(id(<<>>)),
+
+    ok.
+
+bspu_1(<<Bin/binary>> = Bin) ->
+    [Chunk || <<Chunk:5/binary>> <= Bin].
```
bjorng, line 895: I suggest moving this part of the test case to `bs_match_SUITE` so that the code will be tested also without optimization and compiled for OTP 21.

bjorng (approved):

lib/compiler/src/beam_validator.erl
jhogberg: This is a copy of `merge_regs/4` and I'd like to keep the same style. We can change both later on.

lib/compiler/test/beam_validator_SUITE.erl
jhogberg: Fixed.

lib/compiler/test/beam_validator_SUITE.erl
michalmuskala: Introducing a warning in such cases would be a good improvement, agreed.

PR #2746, opened 2020-09-11
Replace CAR, CDR and CONS macros to inline functions 

kvakvs:
Replaces 3 macros: CAR, CDR and CONS with better more typesafe and read/write const-safe inlines.
Gathering OTP team opinion whether this is a good idea, and whether I should do more like this.
If the idea is liked, this opens way for more similar changes:

* Cleaner `Eterm` API: list ops, binary ops, small int ops, atom ops...
* Abstract away bit ops on terms, hiding bit ops from the logic and making the logic cleaner and safer.
* Prevent bugs when an integer is passed/received unintentionally, which is not an Eterm

## Pros

* Renames old style `CAR`, `CDR` to `cell_head` and `cell_tail` which reads easier.
* Ensures that `Eterm*` and `Eterm` goes in/returns `Eterm` and `Eterm*`. Some compile time checks on types, and sizes.
* Has some `const` safety (more in the future changes). The user has to choose whether the `list_cell` will be writable or not.
* Debugging separate functions is also more pleasant (more debug info, stepping in, seeing locals), than macros.
* Should not affect the performance, but could (see Cons below).

## Cons (not CONS)

* While `Eterm` is identical to an integer, the type safety is not yet enforced. But it is possible also to replace `Eterm` with a single field struct. This can give it strong type and prevents confusing it with integers.
* This potentially could affect the performance, should be investigated. On small code examples in Godbolt's Compiler Explorer the assembly coming from the inline code looks identical to macros.

### DIFF
--- a/erts/emulator/beam/beam_bif_load.c
+++ b/erts/emulator/beam/beam_bif_load.c
@@ -309,7 +309,7 @@ finish_loading_1(BIF_ALIST_1)
 
     for (i = 0; i < n; i++) {
 	Eterm* cons = list_val(BIF_ARG_1);
-	Eterm term = CAR(cons);
+	Eterm term = cell_head(cons);
 
 	if (!is_internal_magic_ref(term)) {
 	    goto badarg;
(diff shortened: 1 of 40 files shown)


bjorng:
We think this is a too big diff without any real benefit.

PR #2748, opened 2020-09-12
Escript errors to stderr

mikpe:
Fixes [https://bugs.erlang.org/browse/ERL-1338](url)

It's awkward to get the stderr from a spawned executable, so for now escript_SUITE simply merges stderr of the spawned executable with its stdout, which means there's no real change as far as the suite is concerned.  (There's less bogus quoting in one case though.)

All solutions I've seen to retrieving a spawned executable's stderr have involved I/O redirection to a temporary file.  I didn't think Windows would support that, but some searches indicate it does, so I'll investigate that possibility next.

### DIFF
--- a/lib/stdlib/src/escript.erl
+++ b/lib/stdlib/src/escript.erl
@@ -276,7 +276,7 @@ start(EscriptOptions) ->
             [File|Args] ->
                 parse_and_run(File, Args, EscriptOptions);
             [] ->
-                io:format("escript: Missing filename\n", []),
+                io:format(standard_error, "escript: Missing filename\n", []),
                 my_halt(127)
         end
     catch
@@ -635,7 +635,7 @@ parse_source(S, File, Fd, StartLine, HeaderSz, CheckOnly) ->
             ok = file:close(Fd),
 	    check_source(S3, CheckOnly);
 	{error, Reason} ->
-	    io:format("escript: ~tp\n", [Reason]),
+	    io:format(standard_error, "escript: ~tp\n", [Reason]),
 	    fatal("Preprocessor error")
     end.
 
@@ -693,7 +693,7 @@ epp_parse_file(Epp, S, Forms) ->
     epp_parse_file2(Epp, S, Forms, Parsed).
 
 epp_parse_file2(Epp, S, Forms, Parsed) ->
-    %% io:format("~p\n", [Parsed]),
+    %% io:format(standard_error, "~p\n", [Parsed]),
     case Parsed of
         {ok, Form} ->
             case Form of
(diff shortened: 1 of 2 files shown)


garazdawi:
On some machines (mostly Linux) the `errors` testcase now fails with this:

```
Run: "compile_error"

Expected: <<"/ldisk/daily_build/24_master-opu_o.2020-09-14_21/test/stdlib_test/escript_SUITE_data/compile_error:5: syntax error before: '*'\n/ldisk/daily_build/24_master-opu_o.2020-09-14_21/test/stdlib_test/escript_SUITE_data/compile_error:8: syntax error before: blarf\nescript: There were compilation errors.\nExitCode:127">>

Actual:   <<"ExitCode:127">>
```
In this example it would seem like the stderr output does not arrive at all, which is odd as I think all IO should be synchronous and the port should be flushed...

```
Run: "escript -s c:\\ldisk\\daily_build\\24_master-opu_64_551.2020-09-15_20\\test\\stdlib_test\\escript_SUITE_data\\lint_error"

Expected: <<"c:\\ldisk\\daily_build\\24_master-opu_64_551.2020-09-15_20\\test\\stdlib_test\\escript_SUITE_data\\lint_error:6: function main/1 already defined\nc:\\ldisk\\daily_build\\24_master-opu_64_551.2020-09-15_20\\test\\stdlib_test\\escript_SUITE_data\\lint_error:8: variable 'ExitCode' is unbound\nescript: There were compilation errors.\nExitCode:127">>

Actual:   <<"c:\\ldisk\\daily_build\\24_master-opu_64_551.2020-09-15_20\\test\\stdlib_test\\escript_SUITE_data\\lint_error:6: function main/1 already defined\nescript: There were compilation errors.\nc:\\ldisk\\daily_build\\24_master-opu_64_551.2020-09-15_20\\test\\stdlib_test\\escript_SUITE_data\\lint_error:8: variable 'ExitCode' is unbound\nExitCode:127">>
```

while here it arrives but is mixed with the stdout output in an unexpected way... which is not that surprising with the stderr_to_stdout flag set.

mikpe:
I cannot reproduce the failure in the `errors` testcase, it passes for me on x86_64-linux (Fedora 31).  The second issue you mention appears to be on Windows, which I'm not set up to build on (would have to be Cygwin in that case), but I imagine file descriptor merging probably does not work exactly like it does on Unix/Linux.
I'll try to redirect stderr to a temporary file instead.

garazdawi:
> The second issue you mention appears to be on Windows, which I'm not set up to build on (would have to be Cygwin in that case), but I imagine file descriptor merging probably does not work exactly like it does on Unix/Linux.
I'll try to redirect stderr to a temporary file instead.

Another idea might be to use regexp to make sure the correct texts are present, but maybe not in the same order?

garazdawi:
On windows many tests now fail with `Failed to create temp file: ebusy`, which I assume means that for some reason it did not manage to create the temp file. Maybe put it in the test `priv_dir` instead of in `/tmp/`? That way you don't have to contend with any other processes creating files as well so should be able to use a much simpler naming scheme.

mikpe:
Now uses priv_dir and command name instead of /tmp + pid + random nr to create temporary stderr files, and also logs both file name and actual error instead of just "ebusy" which should help if Windows keeps causing problems.

mikpe:
The logs from that failed build are just woefully incomplete.  I can see that some unrelated test suite had a skipped test case, but not why it was skipped (there must be a crash there somewhere, but it's not visible in the logs).

garazdawi:
Everything seems to work fine now, I just pushed a small fix for the windows tests to work. I'll merge this as soon as I verify that windows works as it should.

garazdawi:
Finally managed to fix all the issues with testing on Windows. Thank you for your contribution!

PR #2750, opened 2020-09-15
Xref: Fix local calls from on_load being reported as unused

asabil:
Xref incorrectly reported functions called exclusively from an on_load
function as being unused.


uabboli:
Thanks for the PR! I'll try to include it in the upcoming 23.1.

PR #2752, opened 2020-09-17
Suppress `xref` output for "unused export" on behaviour module's `behaviour_info/1`

paulo-ferraz-oliveira:
This is a special (albeit optional) case - `behaviour_info`, generated internally by Erlang, so should not be part of the analysis' output.
Tools depending on the analysis output (like e.g. `rebar3 xref`) will force the consumer to add exceptions to them, when otherwise not required.

### DIFF
--- a/lib/tools/src/xref_reader.erl
+++ b/lib/tools/src/xref_reader.erl
@@ -95,6 +95,8 @@ form({function, _, module_info, 0, _Clauses}, S) ->
     S;
 form({function, _, module_info, 1, _Clauses}, S) ->
     S;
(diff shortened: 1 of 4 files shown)


uabboli:
Thanks for the PR! It looks good. It's too late for 23.1, but if you rebase it upon maint after the release of 23.1 next week, it'll make into 23.2.

paulo-ferraz-oliveira:
Rebasing onto `maint`...

paulo-ferraz-oliveira:
Hm... too many conflicts. I'm going to hand-pick it.

PR #2753, opened 2020-09-17
Send exit status from ssh_cli

isvilen:
Make sure that ssh_cli module always send exit-status message,
so SSH clients can use it to check for successful command execution.

[https://bugs.erlang.org/browse/ERL-1349](https://bugs.erlang.org/browse/ERL-1349)


HansN:
Thanks for the PR!

PR #2755, opened 2020-09-17
Optimize find_yregs pass from beam_ssa_pre_codegen

josevalim:
The current pass gets very expensive the longer the
code block is. This commit address it. For example,
imagine this code:

    -module(bar).
    -export([bar/0]).
    bar() ->
      foo:bar(),
      % repeat foo:bar() 999 times
      ok.

Prior to this patch, it would take:

    find_yregs                 :      0.212 s  73 %

After this patch, it takes:

    find_yregs                 :      0.013 s  14 %

Another example is the code in this gist:
https://gist.github.com/josevalim/acac154996e1e61058f367a5b3a161aa

After the patch, find_yregs goes from 3.5s to 0.4s.

This patch is not ready yet for merge, please see comments inline for
places where I would like style-related feedback.

### DIFF
--- a/lib/compiler/src/beam_ssa_pre_codegen.erl
+++ b/lib/compiler/src/beam_ssa_pre_codegen.erl
@@ -1749,7 +1749,7 @@ find_rm_act([]) ->
 %%%
 
 -record(dk, {d :: ordsets:ordset(var_name()),
-             k :: ordsets:ordset(var_name())
+             k :: cerl_sets:set(var_name())
             }).
 
 %% find_yregs(St0) -> St.
@@ -1772,10 +1772,10 @@ find_yregs(#st{frames=[_|_]=Frames,args=Args,ssa=Blocks0}=St) ->
     St#st{ssa=Blocks}.
 
 find_yregs_1([{F,Defs}|Fs], Blocks0) ->
-    DK = #dk{d=Defs,k=[]},
+    DK = #dk{d=Defs,k=cerl_sets:new()},
     D0 = #{F=>DK},
     Ls = beam_ssa:rpo([F], Blocks0),
-    Yregs0 = [],
+    Yregs0 = cerl_sets:new(),
     Yregs = find_yregs_2(Ls, Blocks0, D0, Yregs0),
     Blk0 = map_get(F, Blocks0),
     Blk = beam_ssa:add_anno(yregs, Yregs, Blk0),
@@ -1830,7 +1830,7 @@ find_update_succ([S|Ss], #dk{d=Defs0,k=Killed0}=DK0, D0) ->
     case D0 of
         #{S:=#dk{d=Defs1,k=Killed1}} ->
             Defs = ordsets:intersection(Defs0, Defs1),
-            Killed = ordsets:union(Killed0, Killed1),
+            Killed = cerl_sets:union(Killed0, Killed1),
             DK = #dk{d=Defs,k=Killed},
             D = D0#{S:=DK},
             find_update_succ(Ss, DK0, D);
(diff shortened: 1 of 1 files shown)


lib/compiler/src/beam_ssa_pre_codegen.erl
```diff
@@ -1841,24 +1841,55 @@ find_update_succ([S|Ss], #dk{d=Defs0,k=Killed0}=DK0, D0) ->
...
 find_yregs_is([], Ys, Yregs) -> {Yregs,Ys}.
 
 find_yregs_terminator(Terminator, #dk{k=Killed}, Yregs0) ->
-    Used = beam_ssa:used(Terminator),
-    Yregs = ordsets:intersection(Used, Killed),
-    ordsets:union(Yregs0, Yregs).
+    Yregs = intersect_used(Terminator, Killed),
+    cerl_sets:union(Yregs0, Yregs).
+
+intersect_used(#b_blk{is=Is,last=Last}, Set) ->
```
josevalim, line 1861: The main idea of this patch is replacing `beam_ssa:used+intersection` by this new `intersect_used` with computes the "intersect used" values by checking against a `cerl_sets` at the same time we traverse.

Question: do we want to move `intersect_used` to beam_ssa? The only benefit is that we won't have to duplicate `beam_ssa:used_args/1` below?

lib/compiler/src/beam_ssa_pre_codegen.erl
```diff
@@ -2303,7 +2336,7 @@ reserve_yregs(#st{frames=Frames}=St0) ->
 
 reserve_yregs_1(L, #st{ssa=Blocks0,cnt=Count0,res=Res0}=St) ->
     Blk = map_get(L, Blocks0),
-    Yregs = beam_ssa:get_anno(yregs, Blk),
+    Yregs = ordsets:from_list(cerl_sets:to_list(beam_ssa:get_anno(yregs, Blk))),
```
josevalim, line 2339: Is there a more efficient way to go from cerl_sets to ordsets? 🤔

josevalim:
Finally, I would like to mention that this patch would be simpler if we implemented [EEP 50](https://www.erlang.org/erlang-enhancement-proposals/eep-0050.html). Today, I have to change data structures and the algorithm at the same time, which complicates things. With EEP 50, I would only have to worry about the algorithm.

The other aspect to note is that this optimization could be applied elsewhere, for example in `beam_ssa:def_unused/3` but we can't do so easily because `def_unused/3` works with a different set than the cerl_sets used in this PR (ordsets vs cerl_sets). With EEP 50, it would be easier to reuse those algorithms too.

lib/compiler/src/beam_ssa_pre_codegen.erl
bjorng: I don't think so.

lib/compiler/src/beam_ssa_pre_codegen.erl
bjorng: `beam_ssa` is meant for functions that are used from more than one module, or are expected to be used from other modules in the future. If you don't think that it can be used from other modules, I would suggest that you simplify it and keep in place.

josevalim:
Code simplified! This should be good to go now!

lib/compiler/src/beam_ssa_pre_codegen.erl
```diff
@@ -1841,24 +1841,51 @@ find_update_succ([S|Ss], #dk{d=Defs0,k=Killed0}=DK0, D0) ->
...
+    intersect_used_return(V, Set);
+intersect_used(#b_set{op=phi,args=Args}, Set) ->
+    cerl_sets:from_list([V || {#b_var{}=V,_} <- Args, cerl_sets:is_element(V, Set)]);
+intersect_used(#b_set{args=Args}, Set) ->
+    cerl_sets:from_list(intersect_used_keep(used_args(Args), Set));
+intersect_used(#b_switch{arg=#b_var{}=V}, Set) ->
+    intersect_used_return(V, Set);
+intersect_used(_, _) -> cerl_sets:new().
+
+intersect_used_return(V, Set) ->
```
bjorng, line 1873: I find the name of this function confusing. What about `intersect_used_keep_singleton`?

bjorng (changes requested):
Looks good to me except for the naming of one function.

lib/compiler/src/beam_ssa_pre_codegen.erl
josevalim: Done!

bjorng (approved):

bjorng:
Thanks for the pull request!

PR #2756, opened 2020-09-18
erlfmt: experiment by formatting 9 "random" files in otp

awalterschulze:
We are looking for feedback on `erlfmt` from the erlang community.
To make this easy, we thought to create some draft pull requests on some projects.
This allows you to easily comment and provide feedback, instead of having to learn how to use the tool before being able to provide feedback.

**Note: This pull request is not intended to be merged**

If you would like me to format some other files for you to also comment on, please let me know.
Thank you for your time and I hope this is appropriate.

### DIFF
--- a/lib/crypto/src/crypto.app.src
+++ b/lib/crypto/src/crypto.app.src
@@ -17,14 +17,15 @@
 %%
 %% %CopyrightEnd%
 %%
-{application, crypto,
-   [{description, "CRYPTO"},
+{application, crypto, [
+    {description, "CRYPTO"},
     {vsn, "%VSN%"},
-    {modules, [crypto,
-	       crypto_ec_curves]},
+    {modules, [
+        crypto,
+        crypto_ec_curves
+    ]},
     {registered, []},
     {applications, [kernel, stdlib]},
     {env, [{fips_mode, false}, {rand_cache_size, 896}]},
-    {runtime_dependencies, ["erts-9.0","stdlib-3.4","kernel-5.3"]}]}.
-
-
+    {runtime_dependencies, ["erts-9.0", "stdlib-3.4", "kernel-5.3"]}
+]}.
--- a/lib/crypto/src/crypto.erl
+++ b/lib/crypto/src/crypto.erl
@@ -22,8 +22,17 @@
 
 -module(crypto).
 
--export([start/0, stop/0, info_lib/0, info_fips/0, supports/0, enable_fips_mode/1,
-         version/0, bytes_to_integer/1]).
+-export([
+    start/0,
+    stop/0,
+    info_lib/0,
+    info_fips/0,
+    supports/0,
+    enable_fips_mode/1,
+    version/0,
+    bytes_to_integer/1
+]).
+
 -export([equal_const_time/2]).
 -export([cipher_info/1, hash_info/1]).
 -export([hash/2, hash_init/1, hash_update/2, hash_final/1]).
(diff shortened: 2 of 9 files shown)


okeuday:
It would be nice if the formatting aided your view of the Erlang data as your eye scans the file, starting from the upper-left and moving down, gradually to the right.  To do that, it should be best to make the indentation level clear with the left-most character, so `lib/crypto/src/crypto.app.src` would look like (possibly with the other lists starting on the next line too):

    {application, crypto,
     [{description, "CRYPTO"},
      {vsn, "%VSN%"},
      {modules,
       [crypto,
        crypto_ec_curves]},
      {registered, []},
      {applications, [kernel, stdlib]},
      {env, [{fips_mode, false}, {rand_cache_size, 896}]},
      {runtime_dependencies, ["erts-9.0", "stdlib-3.4", "kernel-5.3"]}]}.

lib/crypto/src/crypto.app.src
```diff
@@ -17,14 +17,15 @@
...
 %% %CopyrightEnd%
 %%
-{application, crypto,
-   [{description, "CRYPTO"},
+{application, crypto, [
+    {description, "CRYPTO"},
     {vsn, "%VSN%"},
-    {modules, [crypto,
-	       crypto_ec_curves]},
+    {modules, [
```
paulo-ferraz-oliveira, line 23: Does it do this only if "not in same line" (as it the case with :28, below)?

lib/crypto/src/crypto.erl
```diff
@@ -47,533 +56,572 @@
...
+        "use crypto:crypto_one_time/4 or crypto:crypto_init/3 + "
+        "crypto:crypto_update/2 + crypto:crypto_final/1 instead"},
+    {block_decrypt, 4,
+        "use crypto:crypto_one_time/5, crypto:crypto_one_time_aead/6,7 "
+        "or crypto:crypto_(dyn_iv)?_init + "
+        "crypto:crypto_(dyn_iv)?_update + crypto:crypto_final instead"}
+]).
+
+-export([
+    stream_init/2, stream_init/3,
```
paulo-ferraz-oliveira, line 224: Why are some kept in the same line? Is it possible to use an option to further "expand" them?

lib/crypto/src/crypto.erl
```diff
@@ -47,533 +56,572 @@
...
+    engine_remove/1,
+    engine_get_id/1,
+    engine_get_name/1,
+    ensure_engine_loaded/2,
+    ensure_engine_loaded/3,
+    ensure_engine_unloaded/1,
+    ensure_engine_unloaded/2
+]).
+
+%% A minimum exported: only what public_key needs.
```
paulo-ferraz-oliveira, line 282: Oh, hahaha, this one's nice!

lib/crypto/src/crypto.erl
```diff
@@ -47,533 +56,572 @@
...
--type dh_public() :: key_integer() .
--type dh_private() :: key_integer() .
--type dh_params() :: [key_integer()] . % [P, G] | [P, G, PrivateKeyBitLength]
+% Always binary() when used as return value
+-type key_integer() :: integer() | binary().
 
--type ecdh_public()  :: key_integer() .
--type ecdh_private() :: key_integer() .
--type ecdh_params()  :: ec_named_curve() | edwards_curve_dh() | ec_explicit_curve() .
+%%% Keys
```
paulo-ferraz-oliveira, line 339: Does `%%%` have special meaning?

lib/crypto/src/crypto.erl
```diff
@@ -47,533 +56,572 @@
...
-                        | wtls9
-                          .
-
--type edwards_curve_dh() :: x25519 | x448 .
-
--type edwards_curve_ed() :: ed25519 | ed448 .
+-type ec_explicit_curve() ::
+    {Field :: ec_field(), Curve :: ec_curve(), BasePoint :: binary(), Order :: binary(),
+        % FIXME: Really?
+        CoFactor ::
```
paulo-ferraz-oliveira, line 383: Is this on purpose? Or does it seem "weird" only because there's a comment? Also, the comment moved from "after `none`", to "before element"...

```erlang
{Field :: ec_field(),
 Curve :: ec_curve(),
 BasePoint :: binary(),
 Order :: binary(),
 CoFactor :: none % FIXME: Really?
           | binary()}.
```
^ how does this look? (I imply that `|` could be optioned to "start in next line".

lib/crypto/src/crypto.erl
```diff
@@ -47,533 +56,572 @@
...
-                     | aes_256_ccm
-
-                     | aes_128_gcm
-                     | aes_192_gcm
-                     | aes_256_gcm
-
-                     | chacha20_poly1305 .
-
+-type cipher() ::
+    cipher_no_iv() |
```
paulo-ferraz-oliveira, line 493: Is "`|` before atom" a choice?

paulo-ferraz-oliveira: I mean, I don't actually think that
```erlang
-type this() :: 'OR' |
                that.
```
(or even)
```erlang
-type this() ::
    'OR' |
    that.
```
is much less readable than
```erlang
-type this() :: 'OR'
              | that.
```
but I guess there are different preferences.

lib/crypto/src/crypto.erl
```diff
@@ -642,87 +696,96 @@ nif_stub_error(Line) ->
...
+
 version() -> ?CRYPTO_VSN.
 
--spec start() -> ok | {error, Reason::term()}.
+-spec start() -> ok | {error, Reason :: term()}.
 start() ->
     case application:start(crypto) of
         ok ->
-            _ = supports(curves), % Build curves cache if needed
+            % Build curves cache if needed
```
paulo-ferraz-oliveira, line 706: Is "have comments after code" a choice?

lib/crypto/src/crypto.erl
```diff
@@ -642,87 +696,96 @@ nif_stub_error(Line) ->
...
-                                      | {macs,    Macs}
-                                      | {curves,  Curves}
-                                      | {rsa_opts, RSAopts},
-                             Hashs :: [sha1() | sha2() | sha3() | blake2() | ripemd160 | compatibility_only_hash()],
-                             Ciphers :: [cipher()],
-                             PKs :: [rsa | dss | ecdsa | dh | ecdh | eddh | ec_gf2m],
-                             Macs :: [hmac | cmac | poly1305],
-                             Curves :: [ec_named_curve() | edwards_curve_dh() | edwards_curve_ed()],
-                             RSAopts :: [rsa_sign_verify_opt() | rsa_opt()] .
+-spec supports() -> [Support] when
```
paulo-ferraz-oliveira, line 717: I do prefer stuff like
```erlang
-spec supports(Input) -> Result
      when Input :: binary(),
           Result :: ok.
```
but maybe it's not "norm".

lib/crypto/src/crypto.erl
```diff
@@ -744,28 +807,29 @@ enable_fips_mode(Enable) ->
...
                                 %% Make the next state change fast:
-                                OldCs = application:get_env(?MODULE,?CURVES), 
+                                OldCs = application:get_env(?MODULE, ?CURVES),
                                 application:set_env(?MODULE, var_name(OldState), OldCs);
                             _ ->
                                 ok
                         end,
                         %% Now rebuild the list by hard work:
-                        application:unset_env(?MODULE, ?CURVES), % This will force a re-build
+                        application:unset_env(?MODULE, ?CURVES),
```
paulo-ferraz-oliveira, line 816: Strange choice, it seems.

lib/crypto/src/crypto.erl
HansN: Yes, three % makes Emacs put the comment at the beginning of the line.

lib/crypto/src/crypto.erl
```diff
@@ -605,30 +653,36 @@
...
-                error(badarg);
-            error:{E, {_File,_Line}, _Reason} when E==notsup ; E==badarg ->
-                error(E)
-        end).
+    try
+        begin
+            CALL
+        end
+    catch
+        error:{error, {_File, _Line}, _Reason} ->
```
HansN, line 668: One of my favorite opinions: Colde should be readable.  You read code much more often than you write it.  Therefore the reader should be helped:
 `{A, {B, C}, D} ` looks like arity four, while
`{A, {B,C},  D}`  shows more clearly that it is an arity three tuple with a tuple as second argument

lib/crypto/src/crypto.erl
```diff
@@ -47,533 +56,572 @@
...
+    aes_ige256 |
+    blowfish_cbc |
+    blowfish_cfb64 |
+    blowfish_ofb64 |
+    chacha20 |
+    des_ede3_cbc |
+    des_ede3_cfb |
+    des_cbc |
+    des_cfb |
+    rc2_cbc.
```
HansN, line 527: Personally I think the original formatting is much more readable.  The grouping with empty lines emphasis the naming schema that first the numbers are varied, then the letter suffix and finally the letter prefix.  This makes the grouping look logical: All aes_*_cbc, then aes_*_ctr etc.  This has a meaning for the semantics.

And I personally also favor the bars vertically aligned in front like in grammars.

lib/crypto/src/crypto.erl
```diff
@@ -642,87 +696,96 @@ nif_stub_error(Line) ->
...
-                             Curves :: [ec_named_curve() | edwards_curve_dh() | edwards_curve_ed()],
-                             RSAopts :: [rsa_sign_verify_opt() | rsa_opt()] .
+-spec supports() -> [Support] when
+    Support ::
+        {hashs, Hashs} |
+        {ciphers, Ciphers} |
+        {public_keys, PKs} |
+        {macs, Macs} |
+        {curves, Curves} |
+        {rsa_opts, RSAopts},
```
HansN, line 724: The original shows that there alternatives just by the aligned vertical bars in front, while you have to read all to understand that in the new formatting.

lib/crypto/src/crypto.erl
awalterschulze: Added this as a data point to an ongoing discussion here https://github.com/WhatsApp/erlfmt/issues/134

lib/crypto/src/crypto.erl
awalterschulze: I have created an issue where we can start to discuss this design decision https://github.com/WhatsApp/erlfmt/issues/167

lib/crypto/src/crypto.erl
awalterschulze: Since this is a repeated issue, I created an issue for a centralized discussion https://github.com/WhatsApp/erlfmt/issues/167

lib/crypto/src/crypto.erl
awalterschulze: Regarding the pipes I opened an issue, where it is easier to gather all opinions https://github.com/WhatsApp/erlfmt/issues/167

awalterschulze: Regarding the newlines, it is still possible to preserve this grouping by adding a comment in between.  How do you feel about that?

lib/crypto/src/crypto.erl
awalterschulze: Thanks again for pointing this out.  Just for reference here is the issue where we can collect the feedback about vertical bars or pipes https://github.com/WhatsApp/erlfmt/issues/167

lib/crypto/src/crypto.erl
awalterschulze: That is very interesting.  Do you think you can find more examples, this could be worth exploring.

lib/crypto/src/crypto.erl
awalterschulze: What is happening here is that the comments 
`% This will force a re-build` and `%% Re-build curves cache. Not strictly needed here.` are seen as a single comment block.
This means that `% This will force a re-build` is no longer seen as a trailing comment, so it is moved to live with  `%% Re-build curves cache. Not strictly needed here.` instead of moved to the line above, as all trailing comments usually are.

I have opened an issue for this https://github.com/WhatsApp/erlfmt/issues/168

lib/crypto/src/crypto.erl
```diff
@@ -779,21 +843,18 @@ enable_fips_mode_nif(_) -> ?nif_stub.
 equal_const_time(X1, X2) ->
     equal_const_time(X1, X2, true).
 
-
-equal_const_time(<<B1,R1/binary>>, <<B2,R2/binary>>, Truth) ->
+equal_const_time(<<B1, R1/binary>>, <<B2, R2/binary>>, Truth) ->
```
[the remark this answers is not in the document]
awalterschulze: Another interesting case :) What would you propose as a consistent way to format binary pattern matching? Sorry for asking, but I struggle to think about how to solve this and you seem to have new ideas.

awalterschulze (commented):
Thank you @okeuday , @HansN  and @paulo-ferraz-oliveira for taking the time to check out the formatter and provide feedback.  This is much appreciated <3

Maybe it is possible to also expand with some things you like, so we can try to keep those decisions in place?

awalterschulze:
> It would be nice if the formatting aided your view of the Erlang data as your eye scans the file, starting from the upper-left and moving down, gradually to the right. To do that, it should be best to make the indentation level clear with the left-most character, so `lib/crypto/src/crypto.app.src` would look like (possibly with the other lists starting on the next line too):
> 
> ```
> {application, crypto,
>  [{description, "CRYPTO"},
>   {vsn, "%VSN%"},
>   {modules,
>    [crypto,
>     crypto_ec_curves]},
>   {registered, []},
>   {applications, [kernel, stdlib]},
>   {env, [{fips_mode, false}, {rand_cache_size, 896}]},
>   {runtime_dependencies, ["erts-9.0", "stdlib-3.4", "kernel-5.3"]}]}.
> ```

So it seems that this is support for tuples, but how we format lists differs from your style, is that correct or is there more that I might be misinterpreting?

Do you feel this way about lists in general or are there counter examples?

lib/crypto/src/crypto.erl
awalterschulze: We have found that this is a bug, good catch :D

okeuday:
> > It would be nice if the formatting aided your view of the Erlang data as your eye scans the file, starting from the upper-left and moving down, gradually to the right. To do that, it should be best to make the indentation level clear with the left-most character, so `lib/crypto/src/crypto.app.src` would look like (possibly with the other lists starting on the next line too):
> > ```
> > {application, crypto,
> >  [{description, "CRYPTO"},
> >   {vsn, "%VSN%"},
> >   {modules,
> >    [crypto,
> >     crypto_ec_curves]},
> >   {registered, []},
> >   {applications, [kernel, stdlib]},
> >   {env, [{fips_mode, false}, {rand_cache_size, 896}]},
> >   {runtime_dependencies, ["erts-9.0", "stdlib-3.4", "kernel-5.3"]}]}.
> > ```
> 
> @okeuday So it seems that this is support for tuples, but how we format lists differs from your style, is that correct or is there more that I might be misinterpreting?

The formatting above is closer to what `io:format/2` provides with `~p` pretty-printing.  So, the format may be more familiar to Erlang users.  However, the main concern is that the graphic design idea of formatting based on how a human eye scans a page is what I wanted to mention.  The erlfmt formatting, with the data example is obscuring the indentation level by using a character that is right-most, due to the human eye scanning down, reading from the left.  The first thing we try to look at is structure and making the structure in the page more difficult to see won't help the reader.  That leads me to believe that `io:format/2` pretty-printing is closer to an easy-to-read formatting method.

> Do you feel this way about lists in general or are there counter examples?

The problem likely exists with things other than lists, but it seems simplest to look at the data example with lists.  I found it disturbing that comments are deleted or moved, so there are likely other things I would want to avoid, but the indentation structure should be the most important concern.

lib/crypto/src/crypto.app.src
paulo-ferraz-oliveira: It makes sense. I guess, in the future, in any case, there'll probably be more options to adapt to certain "preferences" (e.g. if they're widespread).

lib/crypto/src/crypto.erl
paulo-ferraz-oliveira: > functions that have the same name, but different arity
Great. I thus understand the rationale. I like it. Surely grouping per name (e.g. sorted, as you say) would probably benefit other use cases, but that might be a future option too.

lib/crypto/src/crypto.erl
paulo-ferraz-oliveira: > we have to pick one

I imagine that, like `elvis` (for example), there'll be configuration options (I'm sorry, I didn't read the whole manual). I do prefer operators before my code, in any case, albeit this is probably like discussing tabs v. spaces.

lib/crypto/src/crypto.erl
paulo-ferraz-oliveira: Hm... if I understand correctly this is an opinionated formatter 😄 (saw this now on the GitHub page). I get it, but I didn't think it was 😢 I prefer "options". Since that's the case, I'd have to run it over my personal code base to make more informed decisions. Some it'll change stuff and I'll like it, some it'll change stuff and I won't. Is there a particular reason for it to not be configurable, in any case?

lib/crypto/src/crypto.erl
paulo-ferraz-oliveira: I don't think I'll find many examples. This was the result I reached after a few iterations of working with 4-space tabs, tabs v. spaces, same line variables, etc., over the years.

paulo-ferraz-oliveira: This is similar, for example: https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_eval.erl#L100, but not an exact match.

lib/stdlib/src/supervisor.erl
```diff
@@ -22,15 +22,28 @@
...
+    start_link/2, start_link/3,
+    start_child/2,
+    restart_child/2,
+    delete_child/2,
+    terminate_child/2,
+    which_children/1,
+    count_children/1,
+    check_childspecs/1,
+    get_childspec/2
+]).
```
IngelaAndin, line 35: I can see why you grouped start_link variants, but I am thinking you should not make this special case as they are technical different functions and it would be easier to just have one rule. I use to flatten my exports earlier but I find that having one per row is gives a better overview and avoids merge conflicts.

lib/stdlib/src/supervisor.erl
```diff
@@ -41,170 +54,230 @@
...
+        {restarting, pid() | undefined} |
+        [pid()],
+    id :: child_id(),
+    mfargs :: mfargs(),
+    restart_type :: restart(),
+    shutdown :: shutdown(),
+    child_type :: worker(),
+    modules = [] :: modules()
+}).
+
```
IngelaAndin, line 181: I believe this is easier to automate but the left side is easier to read.

lib/stdlib/src/supervisor.erl
```diff
@@ -41,170 +54,230 @@
...
+    strategy => one_for_one,
+    intensity => 1,
+    period => 5
+}).
+
+-define(default_child_spec, #{
+    restart => permanent,
+    type => worker
+}).
+
```
IngelaAndin, line 162: I prefer the new one.

lib/stdlib/src/supervisor.erl
```diff
@@ -216,102 +289,103 @@ delete_child(Supervisor, Id) ->
...
+                    init_dynamic(State, StartSpec);
+                {ok, State} ->
+                    init_children(State, StartSpec);
+                Error ->
+                    {stop, {supervisor_data, Error}}
+            end;
+        ignore ->
+            ignore;
+        Error ->
+            {stop, {bad_return, {Mod, init, Error}}}
```
IngelaAndin, line 388: There are lot of such indenting changes. In general I think we should adhere to how the Emacs mode indents it in these cases as that is the most spread way at least in the OTP code base. How well supervisor adheres I do not know some of this code is really old.

lib/crypto/src/crypto.erl
HansN: A possibly empty comment is ok.  I see that is hard to automatically classify a blank line as semantic information :)

lib/crypto/src/crypto.erl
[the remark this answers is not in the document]
HansN: I'm just happy to tell what I like and dislike :)
In arguments or elements I personally prefer not to have spaces in the binary as in the commented line, but as standalone patterns or topmost right expressions it could be clearer:
`<<Code, Rest/binary>> = Data,`
 and
`NewData = <<Code, SomethingElse, Rest/binary>> ,`

lib/crypto/src/crypto.erl
HansN: Yes, to formulate a rule telling this is nice or not nice is not easy...

lib/crypto/src/crypto.erl
HansN: The way @paulo-ferraz-oliveira prefer also gives nicer documentation if you use the feature to generate the xml from type specs.

lib/crypto/src/crypto.app.src
awalterschulze: `erlfmt` is an opinionated formatter.  We have very limited configuration and we intend to keep it this way.
We are trying to reach a format that most erlang developers can conform to.
Not saying this is going to be easy, but lots of languages have achieved this, but retro fitting a formatter onto an existing community is going to be even tougher.
Either way it would be great if we could talk through our different opinions and assume we can't configure them later.

lib/crypto/src/crypto.erl
awalterschulze: Great to hear you like it :)

Yes we have decided to leave sorting exports to another tool that can be used in conjunction with `erlfmt`.

lib/crypto/src/crypto.erl
awalterschulze: No `erlfmt` is not going to provide many configuration options.
We have one which is the `width` at which we break expression up over lines, but that is it and we intend to keep this set very very minimal.  So lets continue the discussion :)

lib/crypto/src/crypto.erl
awalterschulze: We should probably create a separate issue, but the opinionated nature of `erlfmt` is being discussed here https://github.com/WhatsApp/erlfmt/issues/167

Please try it out on your own code and give us feedback, so we can try to make as best a decision as possible.

lib/crypto/src/crypto.erl
awalterschulze: Thank you for understanding :)

lib/crypto/src/crypto.erl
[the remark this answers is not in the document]
awalterschulze: Ha ha, yes even knowing what you like/dislike is also helpful :)

Okay so in this case it would also be tough to make an automated rule, or am I misinterpreting?

lib/crypto/src/crypto.app.src
paulo-ferraz-oliveira: Fair enough. It's your call in the end, in any case 😄

lib/stdlib/src/supervisor.erl
awalterschulze: Aha avoiding merge conflicts is good :)
I guess this would be true for anything we try to leave on the same line though.

Still it could be worth discussing flattening vs having each export on a separate line.
In this case it does seem as though we created at least a more flattened list.

lib/stdlib/src/supervisor.erl
awalterschulze: Great to hear :)

lib/stdlib/src/supervisor.erl
awalterschulze: I have added the comment as a data point to another issue https://github.com/WhatsApp/erlfmt/issues/134

What part do you think is making the left side easier to read other wise?

IngelaAndin: The alignment of "::" makes it easier to read!

lib/stdlib/src/supervisor.erl
IngelaAndin: Well it is typically annoying for export lists as you most of the time want to keep both unrelated changes but it is quite easy to resolve the conflict incorrectly when the export list is a big not so structured blob of information. And as the younger people say all modern IDEs will let you collapse that part of the code anyway ;)

lib/stdlib/src/supervisor.erl
awalterschulze: In general we want to avoid alignment that is not at 4 spaces, but we also have an open issue to discuss emacs indentation https://github.com/WhatsApp/erlfmt/issues/166

awalterschulze:
> > > It would be nice if the formatting aided your view of the Erlang data as your eye scans the file, starting from the upper-left and moving down, gradually to the right. To do that, it should be best to make the indentation level clear with the left-most character, so `lib/crypto/src/crypto.app.src` would look like (possibly with the other lists starting on the next line too):
> > > ```
> > > {application, crypto,
> > >  [{description, "CRYPTO"},
> > >   {vsn, "%VSN%"},
> > >   {modules,
> > >    [crypto,
> > >     crypto_ec_curves]},
> > >   {registered, []},
> > >   {applications, [kernel, stdlib]},
> > >   {env, [{fips_mode, false}, {rand_cache_size, 896}]},
> > >   {runtime_dependencies, ["erts-9.0", "stdlib-3.4", "kernel-5.3"]}]}.
> > > ```
> > 
> > 
> > @okeuday So it seems that this is support for tuples, but how we format lists differs from your style, is that correct or is there more that I might be misinterpreting?
> 
> The formatting above is closer to what `io:format/2` provides with `~p` pretty-printing. So, the format may be more familiar to Erlang users. However, the main concern is that the graphic design idea of formatting based on how a human eye scans a page is what I wanted to mention. The erlfmt formatting, with the data example is obscuring the indentation level by using a character that is right-most, due to the human eye scanning down, reading from the left. The first thing we try to look at is structure and making the structure in the page more difficult to see won't help the reader. That leads me to believe that `io:format/2` pretty-printing is closer to an easy-to-read formatting method.
> 
> > Do you feel this way about lists in general or are there counter examples?
> 
> The problem likely exists with things other than lists, but it seems simplest to look at the data example with lists. I found it disturbing that comments are deleted or moved, so there are likely other things I would want to avoid, but the indentation structure should be the most important concern.

@okeuday if you find a comment that is deleted, please report it, that is a HUGE bug that we would love to address asap.

Comments are moved since we do not currently support trailing comments and we move them to the line above.
Although this is an ongoing discussion https://github.com/WhatsApp/erlfmt/pull/124

How the human eye scan is very interesting.
I have created an issue to discuss this further.
https://github.com/WhatsApp/erlfmt/issues/182

lib/crypto/src/crypto.erl
awalterschulze: FYI: We are doing some data analysis and early findings seems to suggest that we will change `erlfmt` and move pipes to the start of the line, instead of the end
https://github.com/WhatsApp/erlfmt/pull/186

lib/crypto/src/crypto.erl
awalterschulze: FYI: Data analysis is suggesting we move pipes to before the "atom"
https://github.com/WhatsApp/erlfmt/pull/186

lib/stdlib/src/supervisor.erl
awalterschulze: The IDE might collapse it, but the merge conflict won't go away.
I hope that this way of formatting and grouping functions with the same name is a happy compromise.
But I do feel I am still missing you are trying to say?

IngelaAndin: My point is that the merge conflicts will be rare if you have an export list with one function per
line but quite common if you have a few lines with many functions per line.  And I think the trade off between a long list
that I usually do not care to see and having more "compressed" list but get many merge conflicts, I prefer not having the merge conflicts. I was wondering if is worth the extra effort to group functions with the same name in the same line, as they are different functions even though they normally are related. But that is your trade off!

lib/stdlib/src/supervisor.erl
awalterschulze: Oh wow yes, really good point.  This is a good example.

The reason we decided not to make the formatter do any alignment is because we do not want changing a single line to result in a big change.

For example if we take the example here:
```erlang
-record(child, {% pid is undefined when child is not running
             pid = undefined :: child()
                              | {restarting, pid() | undefined}
                              | [pid()],
             id              :: child_id(),
             mfargs          :: mfargs(),
             restart_type    :: restart(),
             shutdown        :: shutdown(),
             child_type      :: worker(),
             modules = []    :: modules()}).
```

If we provide `restart_type` with a default or add a new field, then we have to shift all types to keep them all aligned.

```erlang
-record(child, {% pid is undefined when child is not running
             pid = undefined :: child()
                              | {restarting, pid() | undefined}
                              | [pid()],
             id              :: child_id(),
             mfargs          :: mfargs(),
             restart_type    :: restart(),
             shutdown        :: shutdown(),
             child_type      :: worker(),
             new_longer_field :: longer_type(),
             modules = []    :: modules()}).
```

This would result in a modification of each line, which is something we would like to avoid.

lib/stdlib/src/supervisor.erl
awalterschulze: Yes it is a trade off, I can hope that when functions have the same name, they are more likely to be edited at the same time, but that is pure speculation.

awalterschulze:
It seems that all questions have been answered and follow up issues have been created.
Are there any other concerns?

lib/crypto/src/crypto.erl
awalterschulze: This is very interesting, but I am struggling to understand how the placement of `when` impacts the documentation.  Could you provide an example, maybe as a screen shot here or whatever is easiest?  Sorry for the inconvenience.

lib/dialyzer/src/dialyzer.erl
```diff
@@ -49,181 +51,189 @@
...
-  F = fun() ->
-	  {Ret, _Warnings} = dialyzer_cl:start(Opts),
-	  Ret
-      end,
-  doit(F).
+    F = fun() ->
+        {Ret, _Warnings} = dialyzer_cl:start(Opts),
+        Ret
+    end,
+    doit(F).
```
zuiderkwast, line 160: This is commonly (at least within Ericsson projects) indented with 'end' aligned under 'fun', like this:

```Erlang
cl(Opts) ->
    F = fun() ->
            {Ret, _Warnings} = dialyzer_cl:start(Opts),
            Ret
        end,
    doit(F).
```

The same with case, if, receive and try.

awalterschulze: Thank you for commenting and bringing this up.
`erlfmt` does not do any alignment to avoid a change of the name `F` to result in a bigger diff over multiple lines.
I hope that makes sense.

lib/crypto/src/crypto.erl
awalterschulze: So I tried to generate edocs using two different placements of when

```
-spec supports(Input) -> Result
   when Input :: binary(),
        Result :: ok.
supports(_) -> ok.

-spec supports1(Input) -> Result when
    Input :: binary(),
    Result :: ok.
supports1(_) -> ok.
```

But the edoc looks the same

Screen Shot 2021-01-27 at 2 40 43 PM

Were you using some other tool to generate docs?

awalterschulze: We are nearing finalizing our format, so it would be really helpful to hear what I am missing sooner rather than later.  Sorry for the short timeline, I know I have expected much more leniency for myself. @HansN @paulo-ferraz-oliveira

uabboli:
A bit off-topic as it's not about formatting, but it should be noted
that PropEr, https://github.com/proper-testing/proper, has a module
called "proper_erlang_abstract_code" which generates abstract code for
the 'master' branch of Erlang/OTP, that is, what will be Erlang/OTP
24.0.

It seems the module has not made it into the latest docs yet,
http://proper.softlab.ntua.gr/doc/, but it should be useful anyway.

We've used it at OTP to check, among other things, the formatters
'erl_pp' and 'erl_prettypr'.

lib/crypto/src/crypto.erl
awalterschulze: @HansN @paulo-ferraz-oliveira am I using the wrong tool to generate these docs, what tool were you referring to?

awalterschulze: @richcarl do you maybe have an idea at the top of your head what is being referred to, to generate xml from type specs?

PR #2758, opened 2020-09-20
fix deadlock during async httpc:request when request body is a function

jimdigriz:
This is resolves the async calling of `httpc:request/4` deadlock bug described in https://github.com/erlang/otp/issues/4329 (was https://bugs.erlang.org/browse/ERL-1321)

Unit tests pass locally for me as much as they usually do for `maint`; `https_sim` seems to flip flop on `stream_no_length` and `stream_large_not_200_or_206` and the IPv6 ones skip for some reason claiming I do not have host support for IPv6.

### DIFF
--- a/lib/inets/doc/src/httpc.xml
+++ b/lib/inets/doc/src/httpc.xml
@@ -133,6 +133,11 @@
         The <c>iolist()</c> in <c>body_processing_result()</c> returns the next part
         of your body and the stream is ended with <c>eof</c>.
       </p>
+      <p>
+        <c>{chunkify, ...}</c> is <url href="https://tools.ietf.org/html/rfc2616#section-14.41">only
+        supported for HTTP/1.1</url>, attempts to use this function with earlier versions
+        will throw a <c>bad_body</c> exception.
+      </p>
     </note>
     <p>For more information about HTTP, see
     <url href="http://www.ietf.org/rfc/rfc2616.txt">RFC 2616</url>.</p>
--- a/lib/inets/doc/src/httpc.xml
+++ b/lib/inets/doc/src/httpc.xml
@@ -113,18 +113,30 @@
     <taglist>
       <tag><c>body()</c></tag>
       <item><p>= <c>string() | binary()</c></p>
-      <p>| <c>{fun(accumulator())</c></p>
-      <p><c> -> body_processing_result(), accumulator()}</c></p>
-      <p>| <c>{chunkify, fun(accumulator())</c></p>
-      <p><c> -> body_processing_result(), accumulator()}</c></p>
+            <p>| <c>body_callback()</c></p>
+            <p>| <c>{chunkify, body_callback()}</c></p>
       </item>
     </taglist>
-    <p><c>body_processing_result() = eof | {ok, iolist(), accumulator()}</c></p>
+    <p><c>body_callback() = {fun(accumulator()) -> body_processing_result() | eof, accumulator()}</c></p>
+    <p><c>body_processing_result() = {ok, iolist(), accumulator()}</c></p>
     <p><c>accumulator() = term()</c></p>
-    <p><c>filename() = string()</c></p>
- <p>For more information about HTTP, see
- <url href="http://www.ietf.org/rfc/rfc2616.txt">RFC 2616</url>.</p> 
-</section>
+    <note>
+      <p>
+        <c>body_callback()</c> is useful to handle dynamic and/or streaming
+        content which can alternatively be supplied using <c>{chunkify, ...}</c>
+        to provide automatic handling of
+        <url href="https://tools.ietf.org/html/rfc2616#section-3.6.1">chunked transfer-encoding</url>
+        where the application will handle the encoding and additionally add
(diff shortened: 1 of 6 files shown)


IngelaAndin:
The patch seems to resolve the issue in your example and I think it provides some insight to the problem. It would be great if you could turn your example into a regression test using inets httpd instead of an external web server. 

The ipv6 is due to the  {require, ipv6_hosts}  in that test. Which means you need to have a common test configuration file that specifies {ipv6_hosts,[myhostname]}  to get the test run. Or  you can comment out the require from the test case. This is due to that it is hard to check in runtime, in a way that for sure works on all our test systems, if iv6 is setup correctly and working.   

stream_large_not_200_or_206 is a wobbling test case on some machines, we have not seen any failures on stream_no_length though.

jimdigriz:
> The patch seems to resolve the issue in your example and I think it provides some insight to the problem.

I am not convinced my patch actually is suitable as it now changes some edge cases that would likely succeed (as they start a new fresh connection) into immediate failures. I do not know what an acceptable solution would look like but if someone pointed me to how the OTP would like it resolved I can put some effort into it.

> It would be great if you could turn your example into a regression test using inets httpd instead of an external web server.

I will try to find the time soon to convert it as requested.

IngelaAndin:
Having a test case that we can control and run automatically is a good start to find a good solution to the problem. The complexity of handling HTTP connections that automate keep alive or pipelining retrying request appropriately, being able to be synchronous or asynchronous, supporting chunk-encoding or length headers handling bodies or no bodies depending on becomes quite complex.

jimdigriz:
As requested, the unit tests have been introduced by https://github.com/erlang/otp/pull/2758/commits/686ae1245adbe79494a4cc57950c2906a65f8d8d

Maybe of interest, pipelining seems to be okay, only persistent connections cause this deadlock to occur.

My fix in https://github.com/erlang/otp/pull/2758/commits/a15e35ef9b46cdaa9504abffba7452e818a11bde does workaround this but breaks (at least for me) either the 'stream_no_length' or 'stream_large_not_200_or_206' unit tests...depending on the alignment of the planets.

IngelaAndin:
Thank you for the test case. I can confirm that your observation of failing those two test case sporadically occurs in our builds too. It is not surprising that pipe lining works as post request are not safe to pipeline and you will get a new connection each time and not get into the deadlock situation. I suppose you have confirmed that the test fails without your fix?  We will keep digging, maybe new timing  uncovered other problems. I can not make this top priority at the moment, but we will get to the bottom of this.

jimdigriz:
Yes, the (persistent connections) test fails every time.

IngelaAndin:
@jimdigriz  So I tried adding this to our test build again as our test for inets have been running cleaner even if there still exists some test cuddling or potentially bugs. It seems like this branch is only causing the stream_no_length test case to fail. Which claims to be a HTTP-1.0 test case as in HTTP-1.0 there could be a length header or a server close if the length header is missing. I am however not convinced the this is testing what it claims. I do not have time to look into this more right now, if you have time feel free to have a look. I will try to return to this before OTP-24 as it would be nice to fix it even it is not top prio for Ericsson.

jimdigriz:
Rebased (leading to 255328b6dd70e7fb487a0ad0c12b2164b0792b02) this work on top of `maint` (also tried against 'OTP-24.0-rc1' and 'master' too) and found all the `inets` tests pass for me:
```
2> ts:run(inets, httpc_SUITE, [batch]).
[snipped]
Testing tests.inets_test.httpc_SUITE: TEST COMPLETE, 160 ok, 0 failed, 1 skipped of 161 test cases

Updating /usr/src/otp/release/tests/test_server/index.html ... done
Updating /usr/src/otp/release/tests/test_server/all_runs.html ... done
ok
```

Including the `stream_no_length` test:
image

Thoughts?

IngelaAndin:
We have a code freeze now for OTP-23.3 but I will test it again after OTP-23.3

jimdigriz:
...does seem non-deterministic, just ran again from scratch and it failed:

image

Then tried again from scratch at it worked:

image

I'll go digging to see if I can figure out why it is flip-flopping...

jimdigriz:
Running the `httpc_SUITE` test in a loop I can still trigger the `stream_large_not_200_or_206` as well as `stream_no_length`, I'll keep digging...

image
image

lisztspace:
After spending time on investigating both the issue(s) and the code, we have concluded that we will close this ticket for the time being.  We acknowledge that there is a problem here, but the proposed solution does not generalise to scenarios with multiple requests.  There is also an amount of negative feature interaction going on in the code from the start.  The documentation isn't very clear on how different features actually should behave.  Instead of doing some additional patches, we really need to do a major overhaul of this code, possibly a full rewrite with the clear goals of providing a minimal http client with well documented semantics.  If there is need for a more full fledged client there are other alternatives that can be used today.  When this will be done is not yet decided.  We will thus close this ticket for the time being.

Also, it is possible to do a workaround by spawning processes to deliver a body in chunks before it is actually sent.

Thank you for your contribution and making us aware of these problems.

PR #2759, opened 2020-09-21
Ingela/ssl/session reuse

IngelaAndin:
This PR adds a small session reuse feature (pre TLS-1.3 that uses session tickets instead). 
Also adds User Guide documentation for TLS sessions. 

@jimdigriz The feature is useful in PR-1968 and should remove the need for the ssl based changes made there.


jimdigriz:
This is great, exactly what I needed to save the couple of hours figuring it out digging through the OTP source code. The walk though examples are a nice touch too!

Ship it! :)

PR #2762, opened 2020-09-22
Improvements to the documentation for the Erlang module

kjellwinblad:
This pull request contains an attempt to improve the documentation for functions in the Erlang module. The pull request adds more examples, fixes errors, and adds more links. Most of the changes are based on comments from @sverker and @rickard-green.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -530,9 +530,28 @@ Z = erlang:adler32_combine(X,Y,iolist_size(Data2)).</code>
       <name name="binary_to_existing_atom" arity="2" since=""/>
       <fsummary>Convert from text representation to an atom.</fsummary>
       <desc>
-        <p>As
-          <seemfa marker="#binary_to_atom/2"><c>binary_to_atom/2</c></seemfa>,
-          but the atom must exist.</p>
+        <p>As <seemfa
+        marker="#binary_to_atom/2"><c>binary_to_atom/2</c></seemfa>,
+        but the atom must exist.</p>
+
+        <p>The Erlang system has a <seeguide
+        marker="system/efficiency_guide:advanced">configurable
+        limit</seeguide> for the total number of atoms that can exist,
+        and atoms are not garbage collected. Therefore, it is not safe
+        to create many atoms from binaries that come from an untrusted
+        source (for example, a file fetched from the Internet), for
+        example, using <seemfa
+        marker="#binary_to_atom/2"><c>binary_to_atom/2</c></seemfa>. This
+        function is thus the appropriate option when the input binary
+        comes from an untrusted source.</p>
+
+        <p>An atom exists in an Erlang system when included in a
+        loaded Erlang module or when created programmatically (for
+        example, by <c>binary_to_atom/2</c>). See the next note for an
+        example of when an atom exists in the source code for an
+        Erlang module but not in the compiled version of the same
+        module.</p>
+
         <p>Failure: <c>badarg</c> if the atom does not exist.</p>
 	<note>
 	  <p>Note that the compiler may optimize away atoms. For
@@ -544,6 +563,12 @@ Z = erlang:adler32_combine(X,Y,iolist_size(Data2)).</code>
 	  to <c>binary_to_existing_atom(&lt;&lt;"some_atom"&gt;&gt;, utf8)</c>
 	  will fail.</p>
 	</note>
+        <note>
+          <p>The number of characters that are permitted in an atom
+          name is limited. The default limits can be found in the
+          <seeguide marker="system/efficiency_guide:advanced">
+          efficiency guide (section Advanced)</seeguide>.</p>
+        </note>
       </desc>
     </func>
 
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -4067,7 +4067,7 @@ RealSystem = system + MissedSystem</code>
         <p>Returns the node where <c><anno>Arg</anno></c> originates.
           <c><anno>Arg</anno></c> can
           be a process identifier, a reference, or a port.
-          If the local node is not
+          If <c><anno>Arg</anno></c> originates from the local node and the local node is not
           alive, <c>nonode@nohost</c> is returned.</p>
         <p>Allowed in guard tests.</p>
       </desc>
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -498,7 +498,7 @@ Z = erlang:adler32_combine(X,Y,iolist_size(Data2)).</code>
         <note>
           <p>There is configurable limit on how many atoms that can
           exist and atoms are not garbage collected. Therefore, it is
-          recommended to consider if
+          recommended to consider whether
           <c>binary_to_existing_atom/2</c> is a better option
           than <c>binary_to_atom/2</c>. The default limits can
           be found in <seeguide
@@ -777,7 +777,7 @@ hello
           emulator, the reduction counter is normally incremented by
           one for each function and BIF call. A context switch is
           forced when the counter reaches the maximum number of
-          reductions for a process (4000 reductions in Erlang/OTP 24).</p>
+          reductions for a process (4000 reductions in Erlang/OTP 19.2 and later).</p>
         <warning>
           <p>This BIF can be removed in a future version of the Beam
             machine without prior warning. It is unlikely to be
@@ -1707,7 +1707,7 @@ b</pre>
           list of arguments for the current function; in Beam it is used
           to provide the arguments for the current function in
           the term <c>Where</c>. As evaluating this function causes
-          an exception to be thrown, it has no return value.
+          an exception to be raised, it has no return value.
           The intent of the exception class <c>error</c> is to
           signal that an unexpected error has happened (for example, a
           function is called with a parameter that has an incorrect
@@ -1741,7 +1741,7 @@ example_fun(A1, A2) ->
         <p>Raises exception of class <c>exit</c> with exit reason
         <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
         is any term. As evaluating this function causes an exception
-        to be thrown, it has no return value. This function should be
+        to be raised, it has no return value. This function should be
         used when the intent is to stop the current process. This
         function differ from <c>erlang:error/1</c> and
         <c>erlang:error/2</c> by causing an exception of a different
@@ -1813,9 +1813,9 @@ example_fun(A1, A2) ->
           should be used when the intent is to stop the current
           process while <c>erlang:exit/2</c> should be used when the
           intent is to send an exit signal to another process. Note
-          also that <c>erlang:exit/1</c> throws an exception that can
+          also that <c>erlang:exit/1</c> raises an exception that can
           be caught while <c>erlang:exit/2</c> does not cause any
-          exception to be thrown.</p>
+          exception to be raised.</p>
         </note>
         <warning>
           <p>The only scenario that has not been covered by the
@@ -2276,8 +2276,7 @@ true</pre>
       <desc>
         <p>Returns the process dictionary as a list of
         <c>{<anno>Key</anno>, <anno>Val</anno>}</c> tuples.
-        This function does not guarantee anything about the order between
-        the items in the returned list.
+        The items in the returned list can be in any order.
         Example:</p>
         <pre>
 > <input>put(key1, merry),</input>
@@ -2323,8 +2322,7 @@ true</pre>
       </fsummary>
       <desc>
         <p>Returns a list of all keys present in the process dictionary.
-        This function does not guarantee anything about the order between
-        the keys in the returned list.
+        The items in the returned list can be in any order.
         Example:</p>
         <pre>
 > <input>put(dog, {animal,1}),</input>
@@ -2341,8 +2339,7 @@ true</pre>
       <desc>
         <p>Returns a list of keys that are associated with the value
         <c><anno>Val</anno></c> in the process dictionary.
-        This function does not guarantee anything about the order between
-        the keys in the returned list.
+        The items in the returned list can be in any order.
         Example:</p>
         <pre>
 > <input>put(mary, {1, 2}),</input>
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -4456,6 +4456,13 @@ RealSystem = system + MissedSystem</code>
       <fsummary>Portable hash function.</fsummary>
       <type_desc variable="Range">Range = 1..2^32, Hash = 1..Range</type_desc>
       <desc>
+        <warning>
+          <p>This function is deprecated as <seemfa
+          marker="erlang#phash2/2"><c>erlang:phash2/2</c></seemfa>
+          should be used for new code. Note that
+          <c>erlang:phash(X,N)</c> is not necessary equal to
+          <c>erlang:phash2(X,N)</c></p>
+        </warning>
         <p>Portable hash function that gives the same hash for
           the same Erlang term regardless of machine architecture and
           ERTS version (the BIF was introduced in ERTS 4.9.1.1).
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -12155,16 +12155,22 @@ true</pre>
       <name name="yield" arity="0" since=""/>
       <fsummary>Let other processes get a chance to execute.</fsummary>
       <desc>
-        <p>Voluntarily lets other processes (if any) get a chance to
-          execute. Using this function is similar to
-          <c>receive after 1 -> true end</c>, except that <c>yield()</c>
-          is faster.</p>
+        <p>Tries to give other processes with the same or
+        higher priority (if any) a chance to execute before
+        returning. There is no guarantee that any other process runs
+        between the invocation and return of <c>erlang:yield/0</c>.</p>
+
+        <p>See the documentation for <seeguide
+        marker="system/reference_manual:expressions#receive"><c>receive-after</c>
+        expressions</seeguide> for how to make the current process
+        sleep for a specific number of milliseconds.</p>
         <warning>
-          <p>There is seldom or never any need to use this BIF
-            as other processes have a chance to run in another scheduler
-            thread anyway.
+          <p>There is seldom or never any need to use this BIF.
             Using this BIF without a thorough grasp of how the scheduler
-            works can cause performance degradation.</p>
+            works can cause performance degradation. The current implementation
+            of this function puts the current process last in the current
+            scheduler's queue for processes of the same priority as the
+            current process.</p>
         </warning>
       </desc>
     </func>
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -752,7 +752,7 @@ hello
           emulator, the reduction counter is normally incremented by
           one for each function and BIF call. A context switch is
           forced when the counter reaches the maximum number of
-          reductions for a process (2000 reductions in Erlang/OTP R12B).</p>
+          reductions for a process (4000 reductions in Erlang/OTP 24).</p>
         <warning>
           <p>This BIF can be removed in a future version of the Beam
             machine without prior warning. It is unlikely to be
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -1704,7 +1704,7 @@ b</pre>
           <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
           is a list of the functions most recently called (the current
           function first). <c><anno>Args</anno></c> is expected to be the
-          list of arguments for the current function; in Beam it is used
+          list of arguments for the current function; it is used
           to provide the arguments for the current function in
           the term <c>Where</c>. As evaluating this function causes
           an exception to be raised, it has no return value.
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -489,10 +489,27 @@ Z = erlang:adler32_combine(X,Y,iolist_size(Data2)).</code>
             encoded atoms</seeguide>
             in section "External Term Format" in the User's Guide.</p>
         </note>
+        <note>
+          <p>The number of characters that are permitted in an atom
+          name is limited. The default limits can be found in the
+          <seeguide marker="system/efficiency_guide:advanced">
+          efficiency guide (section Advanced)</seeguide>.</p>
+        </note>
+        <note>
+          <p>There is configurable limit on how many atoms that can
+          exist and atoms are not garbage collected. Therefore, it is
+          recommended to consider if
+          <c>binary_to_existing_atom/2</c> is a better option
+          than <c>binary_to_atom/2</c>. The default limits can
+          be found in <seeguide
+          marker="system/efficiency_guide:advanced">efficiency guide
+          (section Advanced)</seeguide>.</p>
+        </note>
         <p>Examples:</p>
         <pre>
 > <input>binary_to_atom(&lt;&lt;"Erlang"&gt;&gt;, latin1).</input>
-'Erlang'
+'Erlang'</pre>
+        <pre>
 > <input>binary_to_atom(&lt;&lt;1024/utf8&gt;&gt;, utf8).</input>
 'Ѐ'</pre>
       </desc>
@@ -2731,7 +2748,23 @@ false</code>
           above 255. For more information on Unicode support in atoms, see
           <seeguide marker="erl_ext_dist#utf8_atoms">note on UTF-8
           encoded atoms</seeguide>
-          in section "External Term Format" in the User's Guide.</p>
+        in section "External Term Format" in the User's Guide.</p>
+        <note>
+          <p>The number of characters that are permitted in an atom
+          name is limited. The default limits can be found in the
+          <seeguide marker="system/efficiency_guide:advanced">
+          efficiency guide (section Advanced)</seeguide>.</p>
+        </note>
+        <note>
+          <p>There is configurable limit on how many atoms that can
+          exist and atoms are not garbage collected. Therefore, it is
+          recommended to consider if
+          <c>binary_to_existing_atom/2</c> is a better option
+          than <c>list_to_atom/1</c>. The default limits can
+          be found in <seeguide
+          marker="system/efficiency_guide:advanced">efficiency guide
+          (section Advanced)</seeguide>.</p>
+        </note>
         <p>Example:</p>
         <pre>
 > <input>list_to_atom("Erlang").</input>
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -6061,18 +6061,18 @@ RealSystem = system + MissedSystem</code>
 
     <func>
       <name name="raise" arity="3" since=""/>
-      <fsummary>Stop execution with an exception of specified class, reason,
+      <fsummary>Raises exception of specified class, reason,
         and call stack backtrace.</fsummary>
       <type name="raise_stacktrace"/>
       <desc>
-        <p>Stops the execution of the calling process with an
-          exception of the specified class, reason, and call stack backtrace
+        <p>Raises exception of the specified class, reason, and call stack backtrace
           (<em>stacktrace</em>).</p>
         <p><c><anno>Class</anno></c> is <c>error</c>, <c>exit</c>, or
-          <c>throw</c>. So, if it were not for the stacktrace,
+        <c>throw</c>. So, if it were not for the stacktrace,
           <c>erlang:raise(<anno>Class</anno>, <anno>Reason</anno>,
           <anno>Stacktrace</anno>)</c> is equivalent to
-          <c>erlang:<anno>Class</anno>(<anno>Reason</anno>)</c>.</p>
+          <c>erlang:<anno>Class</anno>(<anno>Reason</anno>)</c> (given that <c><anno>Class</anno></c>
+        is a valid class).</p>
         <p><c><anno>Reason</anno></c> can be any term.</p>
         <p><c><anno>Stacktrace</anno></c> is a list as provided in a try-catch
 	  clause.</p>
@@ -6102,6 +6102,10 @@ end</pre>
           <c>error(erlang:raise(<anno>Class</anno>, <anno>Reason</anno>,
           <anno>Stacktrace</anno>))</c>
           and hope to distinguish exceptions later.</p>
+        <p>See the reference manual about <seeguide
+        marker="system/reference_manual:errors">errors and error
+        handling</seeguide> for more information about exception
+        classes and how to catch exceptions.</p>
       </desc>
     </func>
 
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -1372,7 +1372,12 @@ end</code>
           protocols.</p>
         <p>Returns <c>true</c> if disconnection succeeds,
           otherwise <c>false</c>. If the local node is not alive,
-          <c>ignored</c> is returned.</p>
+        <c>ignored</c> is returned.</p>
+        <note>
+          <p>This function returns before <seemfa
+          marker="#monitor_node/2"> <c>nodedown</c> messages</seemfa>
+          have been delivered.</p>
+        </note>
       </desc>
     </func>
 
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -1764,33 +1764,82 @@ example_fun(A1, A2) ->
       <desc>
         <p>Sends an exit signal with exit reason <c><anno>Reason</anno></c> to
           the process or port identified by <c><anno>Pid</anno></c>.</p>
-        <p>The following behavior applies if <c><anno>Reason</anno></c>
-          is any term, except <c>normal</c> or <c>kill</c>:</p>
+        <p>The following behavior applies if
+        <c><anno>Reason</anno></c> is any term, except <c>normal</c>
+        or <c>kill</c>, and <c>P</c> is the process or port identified
+        by <c><anno>Pid</anno></c>:</p>
         <list type="bulleted">
-          <item><p>If <c><anno>Pid</anno></c> is not trapping exits,
-            <c><anno>Pid</anno></c>
-            itself exits with exit reason <c><anno>Reason</anno></c>.</p>
+          <item><p>If <c>P</c> is not <seemfa marker="#process_flag/2">trapping exits</seemfa>,
+            <c>P</c> exits with exit reason <c><anno>Reason</anno></c>.</p>
           </item>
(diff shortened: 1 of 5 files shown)


system/doc/general_info/DEPRECATIONS
```diff
@@ -74,6 +74,8 @@ snmpm:async_get_bulk/6 since=23 remove=25
 snmpm:async_get_bulk/7 since=23 remove=25
 snmpm:async_get_bulk/8 since=23 remove=25
 
+erlang:phash/2 since=23
```
jhogberg, line 77: Minor nit: this should be under "Added in OTP 24" with `since=24` if you're deprecating it in master.

erts/doc/src/erlang.xml
```diff
@@ -11661,12 +11661,18 @@ improper_end</pre>
...
           for example:</p>
         <pre>
-> <input>trunc(5.5).</input>
+> <input>trunc(5.7).</input>
+5</pre>
+        <pre>
+> <input>trunc(-5.7).</input>
+-5</pre>
+        <pre>
+> <input>trunc(5).</input>
```
garazdawi, line 11993: Maybe add an example of `trunc(36028797018963970.0)`? To illustrate that printed floats and integers may not work as expected.

system/doc/general_info/DEPRECATIONS
kjellwinblad: Thanks for finding this! Fixed: 4ba65a0

erts/doc/src/erlang.xml
kjellwinblad: I have added it here: 516a316 . Maybe one should add an explanation as well but I'm not sure how to explain this in a good way. Feel free to comment if you have any suggestion?

garazdawi: Maybe we need a separate chapter somewhere talking about the printing of floats? There are a bunch of places where this issue pops up, so having a central place to talk about it makes sense.

I tend to link to https://0.30000000000000004.com/ when explaining why things work as they do.

michalmuskala: That's a good site. I made the Elixir docs link to that and also https://floating-point-gui.de/

sverker: The lack of representation of large integers as floats (and how they are printed) is not specific to trunc. So I think that should be explained some other place with maybe links from trunc and round.
For example:
trunc(36028797018963970.0) returns ..3968 because it got ..3968.0 as argument, not because of truncation.

erts/doc/src/erlang.xml
```diff
@@ -10251,23 +10575,36 @@ hello
...
-        <p>Failure: <c>nocatch</c> if not evaluated within a catch.</p>
+        {hello,there}</pre>
+         <p>If evaluated within a
+          <c>try</c>-block of a <seeguide marker="system/reference_manual:expressions#try">try expression</seeguide>, the value <c><anno>Any</anno></c> can be caught within the catch block.
+          Example:</p>
+        <code>
+try
+    throw({my_exception, "Something happened"})
+catch
+    {my_exception, Desc} ->
```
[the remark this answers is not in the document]
garazdawi: I think it is better to be explicit.

erts/doc/src/erlang.xml
garazdawi: http://erlang.org/doc/reference_manual/data_types.html#number I think we should add a paragraph here that we link to in as many places as possible (trunc, round, float_to_list, io:format come to mind). That section should explain in brief why float printing works as it does and also link to some external sources for more detailed explanations.

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
...
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
           is any term. The exit reason is
           <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
           is a list of the functions most recently called (the current
           function first). <c><anno>Args</anno></c> is expected to be the
           list of arguments for the current function; in Beam it is used
           to provide the arguments for the current function in
           the term <c>Where</c>. As evaluating this function causes
-          the process to terminate, it has no return value.</p>
+          an exception to be thrown, it has no return value.
```
bjorng, line 1710: I think it is better to avoid "thrown" as it could be misunderstood.

```suggestion
          an exception to be raised, it has no return value.
```

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
...
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
-          is any term. As
-          evaluating this function causes the process to terminate, it
-          has no return value. Example:</p>
+      <fsummary>Raise exception of class <c>exit</c> with a specified reason.</fsummary>
+      <desc>
+        <p>Raises exception of class <c>exit</c> with exit reason
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+        is any term. As evaluating this function causes an exception
+        to be thrown, it has no return value. This function should be
```
bjorng, line 1744:
```suggestion
        to be raised, it has no return value. This function should be
```

erts/doc/src/erlang.xml
```diff
@@ -1654,33 +1764,82 @@ b</pre>
...
+        processes.
+        </p>
+        <note>
+          <p>The functions <seemfa marker="#exit/1"><c>erlang:exit/1</c></seemfa> and
+          <c>erlang:exit/2</c> are named similarly but provide very
+          different functionalities. The <c>erlang:exit/1</c> function
+          should be used when the intent is to stop the current
+          process while <c>erlang:exit/2</c> should be used when the
+          intent is to send an exit signal to another process. Note
+          also that <c>erlang:exit/1</c> throws an exception that can
```
bjorng, line 1816:
```suggestion
          also that <c>erlang:exit/1</c> raises an exception that can
```

erts/doc/src/erlang.xml
```diff
@@ -1654,33 +1764,82 @@ b</pre>
...
+          be caught while <c>erlang:exit/2</c> does not cause any
+          exception to be thrown.</p>
```
bjorng, line 1818:
```suggestion
          exception to be raised.</p>
```

erts/doc/src/erlang.xml
```diff
@@ -1960,8 +2123,43 @@ true</pre>
...
       <desc>
-        <p>Returns a string corresponding to the text
-          representation of <c><anno>Fun</anno></c>.</p>
+        <warning>
+          <p>This function is <em>not</em> guaranteed to work as described here
+          when <c><anno>Fun</anno></c> was created while executing
+          uncompiled code (uncompiled escripts, the Erlang shell, and
+          other code executed by the <seeerl
+          marker="stdlib:erl_eval"><c>erl_eval</c> module</seeerl>).</p>
+        </warning>
```
bjorng, line 2132: I find this warning confusing and unclear. I think it would be much better to remove it and add an example of a fun created in the shell and short explanation. I think that example and explanation should be a the end of the description.

(Our documentation guidelines says that warnings should used be when there could be serious consequences if the warnings are ignored; notes should be used for less serious things.)

erts/doc/src/erlang.xml
```diff
@@ -2077,7 +2275,10 @@ true</pre>
       <fsummary>Return the process dictionary.</fsummary>
       <desc>
         <p>Returns the process dictionary as a list of
-          <c>{<anno>Key</anno>, <anno>Val</anno>}</c> tuples, for example:</p>
+        <c>{<anno>Key</anno>, <anno>Val</anno>}</c> tuples.
+        This function does not guarantee anything about the order between
+        the items in the returned list.
```
bjorng, line 2280:
```suggestion
        The items in the returned list can be in any order.
```

erts/doc/src/erlang.xml
```diff
@@ -2117,8 +2322,10 @@ true</pre>
       <fsummary>Return a list of all keys from the process dictionary.
       </fsummary>
       <desc>
-        <p>Returns a list of all keys present in the process dictionary,
-          for example:</p>
+        <p>Returns a list of all keys present in the process dictionary.
+        This function does not guarantee anything about the order between
+        the keys in the returned list.
```
bjorng, line 2327:
```suggestion
        The items in the returned list can be in any order.
```

erts/doc/src/erlang.xml
```diff
@@ -2133,7 +2340,10 @@ true</pre>
       <fsummary>Return a list of keys from the process dictionary.</fsummary>
       <desc>
         <p>Returns a list of keys that are associated with the value
-          <c><anno>Val</anno></c> in the process dictionary, for example:</p>
+        <c><anno>Val</anno></c> in the process dictionary.
+        This function does not guarantee anything about the order between
+        the keys in the returned list.
```
bjorng, line 2345:
```suggestion
        The items in the returned list can be in any order.
```

bjorng (commented):
Overall looks good to me. I found some minor issues that I think can be improved.

erts/doc/src/erlang.xml
kjellwinblad: @garazdawi I think your suggestion is good. I will try to implement it.

erts/doc/src/erlang.xml
```diff
@@ -1308,7 +1374,12 @@ end</code>
           protocols.</p>
         <p>Returns <c>true</c> if disconnection succeeds,
           otherwise <c>false</c>. If the local node is not alive,
-          <c>ignored</c> is returned.</p>
+        <c>ignored</c> is returned.</p>
+        <note>
+          <p>This function returns before <seemfa
+          marker="#monitor_node/2"> <c>nodedown</c> messages</seemfa>
+          have been delivered.</p>
```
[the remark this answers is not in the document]
kjellwinblad: Fixed: 2358fe5

erts/doc/src/erlang.xml
```diff
@@ -1654,33 +1764,82 @@ b</pre>
...
+        sends a signal with the <c>normal</c> reason to itself is
+        described in the warning):</p>
+        <list type="bulleted">
+          <item>If <c>P</c> is <seemfa
+          marker="#process_flag/2">trapping exits</seemfa>, the exit signal is
+          transformed into a message <c>{'EXIT', From, normal}</c>,
+          where <c>From</c> is the process identifier of the process
+          that sent the exit signal, and delivered to <c>P</c>'s
+          message queue.</item>
+          <item>The message has no effect if <c>P</c> is not trapping
```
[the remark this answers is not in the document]
kjellwinblad: Fixed: 2358fe5

erts/doc/src/erlang.xml
```diff
@@ -10251,23 +10575,36 @@ hello
       <name name="throw" arity="1" since=""/>
       <fsummary>Throw an exception.</fsummary>
       <desc>
-        <p>A non-local return from a function. If evaluated within a
-          <c>catch</c>, <c>catch</c> returns value <c><anno>Any</anno></c>.
+        <p>Used to raise exceptions and to do non-local returns from functions.</p>
```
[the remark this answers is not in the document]
kjellwinblad: Fixed: 2358fe5

erts/doc/src/erlang.xml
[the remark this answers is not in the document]
kjellwinblad: Fixed: 2358fe5

erts/doc/src/erlang.xml
kjellwinblad: Would it be better if the warning said something like this:

"The fun_to_list/1 function does not work in a useful way when applied to fun values created while executing fun expressions in uncompiled code (uncompiled escripts, the Erlang shell, and other code executed by the <seeerl marker="stdlib:erl_eval"><c>erl_eval</c> module</seeerl>). All funs created from fun expressions in uncompiled code with the same arity are mapped to the same list by fun_to_list/1."

I agree with you that the original text in the warning was confusing and unclear but I think a warning is appropriate here as I think users can be very confused by fun_to_list/1 if they use it in an escript or shell without being aware of what the warning says. I don't think an example is needed to demonstrate what the new warning text says.

I also agree that it is better to move the warning to the end and change the text so that it first speaks about funs created by fun expressions in compiled code.

erts/preloaded/src/erlang.erl
```diff
@@ -944,8 +945,9 @@ fun_info_mfa(_Fun) ->
     erlang:nif_error(undefined).
 
 %% fun_to_list/1
--spec erlang:fun_to_list(Fun) -> string() when
-      Fun :: function().
+-spec erlang:fun_to_list(Fun) -> String when
+      Fun :: function(),
+      String :: string().
```
uabboli, line 950: I see no reason to introduce String :: string() (a function constraint) for the return value.

When we introduce these extra function constraints, in most cases for arguments, it's to make it easier for the pretty-printer to create readable specs. In this case it's not necessary.

One could argue that it's not necessary to introduce a function constraint for the argument Fun either, but that's another topic.

erts/preloaded/src/erlang.erl
```diff
@@ -1642,7 +1644,7 @@ put(_Key, _Val) ->
     erlang:nif_error(undefined).
 
 %% raise/3
--spec erlang:raise(Class, Reason, Stacktrace) -> no_return() when
+-spec erlang:raise(Class, Reason, Stacktrace) -> badarg when
```
uabboli, line 1647: Note that 'badarg' is an atom.

erts/preloaded/src/erlang.erl
```diff
@@ -2612,7 +2614,7 @@ term_to_iovec(_Term, _Options) ->
 
 %% Shadowed by erl_bif_types: erlang:tl/1
 -spec tl(List) -> term() when
-      List :: [term(), ...].
+      List :: nonempty_maybe_improper_list().
```
uabboli, line 2617: Note that how you write it in the code does not affect how it looks in the docs.

kjellwinblad: [term(), ...] and nonempty_maybe_improper_list() are different types, right?

uabboli: Yes, didn't see that. But what you says makes the spec strange, considering we have success typings. The tail of a list() cannot be improper.

uabboli: I take back what I wrote. Considering the return value is term(), you're right. Hope I got it now.

erts/doc/src/erlang.xml
```diff
@@ -1592,15 +1667,20 @@ b</pre>
...
-          is any term. The exit reason is
-          <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
-          is a list of the functions most recently called (the current
-          function first). As evaluating this function causes
-          the process to terminate, it has no return value. Example:</p>
+      <fsummary>Raise exception of class error.</fsummary>
+      <desc>
+        <p>Raises exception of class <c>error</c> with the reason
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+        is any term. The exit reason is <c>{<anno>Reason</anno>,
```
rickard-green, line 1674: It is not an "exit reason". It should be either "error reason" or just "reason".

erts/doc/src/erlang.xml
```diff
@@ -1592,15 +1667,20 @@ b</pre>
...
-        <p>Stops the execution of the calling process with the reason
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
-          is any term. The exit reason is
-          <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
-          is a list of the functions most recently called (the current
-          function first). As evaluating this function causes
-          the process to terminate, it has no return value. Example:</p>
+      <fsummary>Raise exception of class error.</fsummary>
+      <desc>
+        <p>Raises exception of class <c>error</c> with the reason
```
rickard-green, line 1672: an exception

erts/doc/src/erlang.xml
```diff
@@ -1592,15 +1667,20 @@ b</pre>
...
-      <fsummary>Stop execution with a specified reason.</fsummary>
-      <desc>
-        <p>Stops the execution of the calling process with the reason
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
-          is any term. The exit reason is
-          <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
-          is a list of the functions most recently called (the current
-          function first). As evaluating this function causes
-          the process to terminate, it has no return value. Example:</p>
+      <fsummary>Raise exception of class error.</fsummary>
```
rickard-green, line 1670: an exception

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
       <name name="error" arity="2" since=""/>
       <fsummary>Stop execution with a specified reason.</fsummary>
       <desc>
-        <p>Stops the execution of the calling process with the reason
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+        <p>Raises exception of class <c>error</c> with the reason
```
rickard-green, line 1701: an exception

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
...
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
           is any term. The exit reason is
```
rickard-green, line 1703: not an exit reason

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
...
     <func>
       <name name="exit" arity="1" since=""/>
-      <fsummary>Stop execution with a specified reason.</fsummary>
-      <desc>
-        <p>Stops the execution of the calling process with exit reason
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
-          is any term. As
-          evaluating this function causes the process to terminate, it
-          has no return value. Example:</p>
+      <fsummary>Raise exception of class <c>exit</c> with a specified reason.</fsummary>
```
rickard-green, line 1739: an

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
...
+      <desc>
+        <p>Raises exception of class <c>exit</c> with exit reason
```
rickard-green, line 1741: an

erts/preloaded/src/erlang.erl
kjellwinblad: I think the constraint was there even before my change so I have not added any extra constraint. I introduced a name for the return value to make it easier to refer to it in the documentation. This is not necessary to do but I think it makes the documentation better. Btw, I'm currently rewriting the description of `erlang:fun_to_list/1` based on suggestions from @bjorng so it will change soon.

uabboli: You can also use an annotation, like "-> String :: string()".

erts/preloaded/src/erlang.erl
kjellwinblad: Yes, but in a type spec `badarg` reffers to the type of the atom badarg, or have I done something wrong? Btw, is it better to write the return type as `no_return() | 'badarg'` in this case to illustrate that the function may throw an exception?

erts/doc/src/erlang.xml
```diff
@@ -1618,28 +1698,58 @@ b</pre>
...
+{ok,test}
+<input>7> test:example_fun(arg1,"this is the second argument").</input>
+** exception error: my_error
+     in function  test:example_fun/2
+         called as test:example_fun(arg1,"this is the second argument")
+ </pre>
       </desc>
     </func>
 
     <func>
```
rickard-green, line 1737: It can also be worth noting that if a process calls `exit(kill)` and does not catch the exception, it will terminate with exit reason `kill` and also emit exit signals with exit reason `kill` (not `killed`) to all linked processes. Such exit signals with exit reason `kill` *can* be trapped by the linked processes.

essen: Wow I did not know that. Then why is `kill` changed into `killed` when an `exit(P, kill)` propagates to linked processes?

erts/preloaded/src/erlang.erl
kjellwinblad: Ok, thanks, I did not know that.

erts/doc/src/erlang.xml
rickard-green: This is a design decision made a *really* long time ago and probably originates from the beginning. The only reason I can come up with is that `killed` hints about a process being "brutally" killed using `exit(P, kill)`.

The draft erlang specification http://erlang.org/download/erl_spec47.ps.gz (see section 10.6.3) describes exit signals as consisting of `Sender`, `Cause`, and a flag `ProcessCompleted`. The flag is true if the sender terminated otherwise it is false. If it is false, it implies that the signal was sent using `exit/2`. Only exit signals with `Cause=kill` and `ProcessCompleted=false` terminates the receiver regardless of its trap exit state and translate `Cause` (reason) `kill` to `killed`.

essen: Interesting, thanks a lot!

erts/doc/src/erlang.xml
```diff
@@ -1654,33 +1764,82 @@ b</pre>
...
+          exits.</item>
+        </list>
+        <p>If <c><anno>Reason</anno></c> is the atom <c>kill</c>, that
+        is, if <c>exit(<anno>Pid</anno>, kill)</c> is called, an
+        untrappable exit signal is sent to the process that
+        is identified by <c><anno>Pid</anno></c>, which
+        unconditionally exits with exit reason <c>killed</c>.  The
+        exit reason is changed from <c>kill</c> to <c>killed</c> so
+        that the <c>kill</c> signal will not cascade to all linked
+        processes.
```
rickard-green, line 1807: The "The exit reason is changed from `kill` to `killed` so that the `kill` signal will not cascade to all linked processes." sentence isn't really true.

There are exit signals with exit reason `kill` that does not unconditionally terminate the receiver. See comment under `exit/1`. If `kill` reason was kept `ProcessCompleted` would in this case be `true` on all exit signals propagated due to links. That is, the propagated exit signals would not unconditionally terminate receivers.

erts/doc/src/erlang.xml
```diff
@@ -5712,18 +6058,18 @@ RealSystem = system + MissedSystem</code>
 
     <func>
       <name name="raise" arity="3" since=""/>
-      <fsummary>Stop execution with an exception of specified class, reason,
+      <fsummary>Raises exception of specified class, reason,
```
rickard-green, line 6061: an exception

erts/doc/src/erlang.xml
```diff
@@ -5712,18 +6058,18 @@ RealSystem = system + MissedSystem</code>
...
         and call stack backtrace.</fsummary>
       <type name="raise_stacktrace"/>
       <desc>
-        <p>Stops the execution of the calling process with an
-          exception of the specified class, reason, and call stack backtrace
+        <p>Raises exception of the specified class, reason, and call stack backtrace
```
rickard-green, line 6065: an

erts/doc/src/erlang.xml
```diff
@@ -10251,23 +10609,39 @@ hello
       <name name="throw" arity="1" since=""/>
       <fsummary>Throw an exception.</fsummary>
       <desc>
-        <p>A non-local return from a function. If evaluated within a
-          <c>catch</c>, <c>catch</c> returns value <c><anno>Any</anno></c>.
+        <p>Raises exception of class <c>throw</c>. Intended to be used to do non-local returns from functions.</p>
```
rickard-green, line 10612: an

erts/preloaded/src/erlang.erl
uabboli: My bad. Too fast on the trigger. No, 'badarg' is OK, of course.

bjorng: It is recommended to single-quote atoms in specs.

erts/doc/src/erlang.xml
kjellwinblad: Fixed in 5088294

erts/preloaded/src/erlang.erl
kjellwinblad: Fixed in 0d1d237

erts/preloaded/src/erlang.erl
kjellwinblad: Fixed in 0d1d237

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: Fixed: 3b5bfbb

erts/doc/src/erlang.xml
kjellwinblad: I have updated the text 3b5bfbb. @rickard-green can you check my update?

erts/doc/src/erlang.xml
kjellwinblad: I have updated the text 3b5bfbb. @rickard-green can you check my update?

erts/doc/src/erlang.xml
kjellwinblad: @garazdawi I have tried to implement your suggestion 002d1ef. Can you check it?

garazdawi (changes requested):

system/doc/reference_manual/data_types.xml
```diff
@@ -79,6 +79,25 @@
 11> <input>1_234.333_333</input>
 1234.333333
 </pre>
+<marker id="float_represenation_problem" /><note>
```
uabboli, line 82: Maybe you want to correct the spelling of `representation'.

erts/doc/src/erlang.xml
```diff
@@ -6347,6 +6353,14 @@ true</pre>
         <pre>
 <input>round(-5.5).</input>
 -6</pre>
+        <pre>
+<input>round(36028797018963969.0).</input>
```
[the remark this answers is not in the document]
kjellwinblad: I changed the float literal that you suggested for another example were I wanted to show that a float is not always printed the same way as the float literal that the float was created from and then I thought that it would be pedagogically advantageous to use the same float literal in all examples.

system/doc/reference_manual/data_types.xml
```diff
@@ -79,6 +79,25 @@
...
+  <p>Float literals cannot always be
```
uabboli, line 83: Besides the confusion about floating point numbers not being
represented exactly, there is also confusion about floating point
numbers not being printed as expected. Typically it's about numbers
such as float(1 bsl 53), which is indeed represented exactly, but has
the same representation as, for example, float(1 bsl 53 + 1).

I think all basic information can be found in
https://bugs.erlang.org/browse/ERL-1308; the problem is to explain it
correctly and succinctly so that everyone can understand.

erts/doc/src/erlang.xml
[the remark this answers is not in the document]
garazdawi: aha!

system/doc/reference_manual/data_types.xml
uabboli: `1 bsl 53' is not a good example. One needs to go up to `1 bsl 55'.

kjellwinblad: @uabboli I have added a comment to https://bugs.erlang.org/browse/ERL-1308 . Can you have a look? Can you also have a look at @garazdawi's suggestion for improving this above and let us know if you think it explains the "not being printed as expected" part in a satisfactory way?

system/doc/reference_manual/data_types.xml
kjellwinblad: Thanks for seeing this. Fixed: 855323f

system/doc/reference_manual/data_types.xml
```diff
@@ -79,6 +79,25 @@
...
+    0.30000000000000004
+    > <input>36028797018963969.0.</input>
+    36028797018963970.0
+  </pre>
+  <p>See external resources for a detailed description of why this is
+  the case (for example <url
+  href="https://floating-point-gui.de/">https://floating-point-gui.de/</url>
+  and <url
+  href="https://0.30000000000000004.com/">https://0.30000000000000004.com/</url>)</p>
+</note>
```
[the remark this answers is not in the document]
kjellwinblad: Looks good to me. I have tested the XML and added your suggestion in 855323f

system/doc/reference_manual/data_types.xml
uabboli: Interesting that Python's output differs. I've always thought Scheme does everything right when it comes to printing floating point numbers, so I downloaded mit-scheme: they print Lukas' example like Erlang does. It seems Python is using a modified algorithm.

BTW, there is another example where a zero can be misunderstood: the syntax of Erlang's floating point numbers requires a digit before and after the dot, while Scheme's syntax does not, which is to their advantage.

Regarding Lukas' suggestion. I find it a bit too succinct. Given the example '36028797018963969.0.', how am I supposed to think? Isn't it possible, or even likely, that I write a bug report?

KennethL: There is a spelling error in the marker (line 82)
<marker id="float_represenation_problem" /><note>
  <p>Float literals cannot always be

Should be <marker id="float_representation_problem" />....

garazdawi: My thinking was that we rely on external pages to do the in-depth description of why floating points behave the way they do and just have a basic description of the problem in our documentation. We could of course have a larger better description, like for instance [elixir](https://hexdocs.pm/elixir/Float.html#module-known-issues).

uabboli: Elixir's explanation does not mention the rounding behavior of the pretty-printer, nor does the external paper. You refer to "the problem", but I claim there are two "problems", the other one being the pretty-printer algorithm. Both problems seem to generate bug reports.

kjellwinblad: I have looked into this and I think it makes sense to change the printing algorithm so the example becomes:

```
> 36028797018963969.0.
3.602879701896397e16
```

instead of:

```
> 36028797018963969.0.
36028797018963970.0
```

This should make the example less confusing. I will add a commit with this change and an attempt to do a better explanation next week.

kjellwinblad: I have changed the printing algorithm so that 36028797018963969.0 is printed as 3.602879701896397e16 and updated the documentation accordingly. See 09bc530

garazdawi (approved):

uabboli (approved):

erts/doc/src/erlang.xml
```diff
@@ -1616,35 +1696,83 @@ b</pre>
...
-        <p>Stops the execution of the calling process with the reason
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
-          is any term. The exit reason is
+        <p>Raises an exception of class <c>error</c> with the reason
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+          is any term. The error reason is
           <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
           is a list of the functions most recently called (the current
           function first). <c><anno>Args</anno></c> is expected to be the
           list of arguments for the current function; in Beam it is used
```
bjorng, line 1707: This sentence must have been rewritten a long time ago when JAM still existed.

```suggestion
          list of arguments for the current function; it is used
```

sverker:
Removed top two commits that changed the pretty printing of large floats. Creating a dedicated PR for that.

erts/doc/src/erlang.xml
```diff
@@ -533,6 +581,10 @@ Z = erlang:adler32_combine(X,Y,iolist_size(Data2)).</code>
         <pre>
 > <input>binary_to_float(&lt;&lt;"2.2017764e+0">>).</input>
 2.2017764</pre>
+        <p>The float string format is the same as the format for
+        <seeguide marker="system/reference_manual:data_types">Erlang
+        float literals</seeguide> except for that underscores are not
+        permitted.</p>
```
zuiderkwast, line 587: Remove "for" => "... except that underscores are not permitted.".

(Disclaimer: I'm not a linguist.)

erts/doc/src/erlang.xml
```diff
@@ -2802,6 +3126,10 @@ false</code>
         <pre>
 > <input>list_to_float("2.2017764e+0").</input>
 2.2017764</pre>
+        <p>The float string format is the same as the format for
+        <seeguide marker="system/reference_manual:data_types">Erlang
+        float literals</seeguide> except for that underscores are not
```
zuiderkwast, line 3131: "except for that" => "except that"

erts/doc/src/erlang.xml
```diff
@@ -1616,35 +1696,83 @@ b</pre>
...
       <desc>
-        <p>Stops the execution of the calling process with the reason
-          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
-          is any term. The exit reason is
+        <p>Raises an exception of class <c>error</c> with the reason
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+          is any term. The error reason is
           <c>{<anno>Reason</anno>, Where}</c>, where <c>Where</c>
           is a list of the functions most recently called (the current
           function first). <c><anno>Args</anno></c> is expected to be the
```
zuiderkwast, line 1706: The word "stacktrace" could be used here, since it is used in the documentation of raise/3 and probably other places.

zuiderkwast (commented):
Very nice! Sorry for posting comments after merge.

PR #2764, opened 2020-09-22
Use SSE/AVX to shrink code generated by the JIT

bjorng:
When two words that are consecutive (or reverse consecutive), SSE and AVX can be used to move two words at once, shrinking the code size.

### DIFF
--- a/erts/emulator/beam/jit/instr_common.cpp
+++ b/erts/emulator/beam/jit/instr_common.cpp
@@ -816,20 +816,46 @@ void BeamModuleAssembler::emit_i_trim(const ArgVal &Words) {
     }
 }
 
-void BeamModuleAssembler::emit_move(const ArgVal &Src, const ArgVal &Dst) {
+void BeamModuleAssembler::emit_i_move(const ArgVal &Src, const ArgVal &Dst) {
     mov_arg(Dst, Src);
 }
 
-/* Move two words at consecutive addresses to consecutive destinations. */
-void BeamModuleAssembler::emit_move_two_words(const ArgVal &Src,
-                                              const ArgVal &Dst) {
-    x86::Mem dst_ptr = getArgRef(Dst);
-    x86::Mem src_ptr = getArgRef(Src);
+/* Move two words at consecutive addresses to consecutive or reverse
+ * consecutive destinations. */
+void BeamModuleAssembler::emit_move_two_words(const ArgVal &Src1,
+                                              const ArgVal &Dst1,
+                                              const ArgVal &Src2,
+                                              const ArgVal &Dst2) {
+    x86::Mem src_ptr = getArgRef(Src1, 16);
 
-    dst_ptr.setSize(16);
-    src_ptr.setSize(16);
-    a.movups(x86::xmm0, src_ptr);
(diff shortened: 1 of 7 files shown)


erts/emulator/beam/jit/instr_common.cpp
```diff
@@ -617,18 +617,50 @@ void BeamModuleAssembler::emit_i_make_fun(const ArgVal &Fun,
...
-    mov_arg(ARG1, Src);
+    x86::Gp boxed_ptr = emit_ptr_val(src, src);
 
-    x86::Gp boxed_ptr = emit_ptr_val(ARG1, ARG1);
+    switch (ArgVal::register_relation(Hd, Tl)) {
+    case ArgVal::Relation::consecutive: {
+        comment("(moving head and tail together)");
+        x86::Mem dst_ptr = getArgRef(Hd);
+        x86::Mem src_ptr = getCARRef(boxed_ptr);
+        dst_ptr.setSize(16);
```
jhogberg, line 630: We seem to change the size of memory operands pretty often now, maybe we should add a size overload to `getArgRef` and friends, similar to `getSchedulerRegRef`?

erts/emulator/beam/jit/instr_common.cpp
```diff
@@ -617,18 +617,50 @@ void BeamModuleAssembler::emit_i_make_fun(const ArgVal &Fun,
...
+        }
 
-    a.mov(ARG2, getCARRef(boxed_ptr));
-    a.mov(ARG3, getCDRRef(boxed_ptr));
+        comment("(moving and swapping head and tail together)");
+        x86::Mem dst_ptr = getArgRef(Tl);
+        x86::Mem src_ptr = getCARRef(boxed_ptr);
+        dst_ptr.setSize(16);
+        src_ptr.setSize(16);
+        a.vpermilps(x86::xmm0, src_ptr, 0x4E); /* Load and swap */
```
jhogberg, line 646: I think `vpermilpd` with a control byte of 1 is a bit easier to follow in this case. It's otherwise identical with regards to size, speed, latency, etc so it's just a matter of taste.

```
IF (imm8[0] = 0) THEN DEST[63:0]←SRC1[63:0]
IF (imm8[0] = 1) THEN DEST[63:0]←SRC1[127:64]
IF (imm8[1] = 0) THEN DEST[127:64]←SRC1[63:0]
IF (imm8[1] = 1) THEN DEST[127:64]←SRC1[127:64]
DEST[MAXVL-1:128]←0
```

jhogberg (approved):

bjorng:
@jhogberg I have implemented your suggestions. I have also fixed bugs in the fallback code for CPUs without AVX.

vans163:
I am not sure how relevant frequency down clocking still is in regards to AVX, https://blog.cloudflare.com/on-the-dangers-of-intels-frequency-scaling/. But bringing it up.

bjorng:
@vans163 That's interesting. I didn't know about that.

However, I don't think it applies to how we use AVX. We only use the 128-bit AVX instructions, and only AVX instruction sequences that C compilers choose to emit.

PR #2765, opened 2020-09-22
Add new make_fun3 instruction

sverker:
Compared to old make_fun2:
- No GC done.
- Environment and destination are arguments of instruction and not assumed x-registers.

Results in more effective register allocations.

### DIFF
--- a/lib/compiler/src/genop.tab
+++ b/lib/compiler/src/genop.tab
@@ -578,7 +578,7 @@ BEAM_FORMAT_NUMBER=0
 
 ## @spec put_tuple2  Destination Elements
 ## @doc  Build a tuple with the elements in the list Elements and put it
-##       put into register Destination.
+##       into register Destination.
 164: put_tuple2/2
 
 ## @spec bs_get_tail Ctx Dst Live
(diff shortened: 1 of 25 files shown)


bjorng:
Looks good to me.

I have only one minor nit to pick: For `Co-authored-by` to work as intended, it must be placed last in the commit message.

lib/compiler/src/genop.tab
```diff
@@ -608,3 +608,5 @@ BEAM_FORMAT_NUMBER=0
 ##       it will never fail at runtime, or 'resume' when we know the input is
 ##       a match context.
 170: bs_start_match4/4
+
+171: make_fun3/3
```
bjorng, line 612: Please document make_fun/3.

bjorng:
@sverker Two test cases fail in the compiler test suite on host ecthelion (to recreate locally, give the `cover` option when runnning the test suites).

You will need to include `no_make_fun3` must be included in `test_lib:opt_opts/1`.

lib/compiler/src/genop.tab
sverker: done

sverker:
*** MERGED *** MERGED *** MERGED ***
This PR is merged to master at 1073fa2

PR #2770, opened 2020-09-24
beam_ssa_opt: Optimize sinking of get_tuple_element instructions

jhogberg:
https://bugs.erlang.org/browse/ERL-1359


bjorng:
Looks good to me.

PR #2771, opened 2020-09-24
Documentation of when exit signals are emitted by gen_server

zuiderkwast:
The conditions for when an exit signal is emitted by a terminating
gen_server are added in various sections of the gen_server manual:

* In the Description section
* Under start_link/3,4
* Under stop/1,3
* Under Module:init/1
* Under Module:terminate/2

### DIFF
--- a/lib/stdlib/doc/src/gen_server.xml
+++ b/lib/stdlib/doc/src/gen_server.xml
@@ -99,6 +99,14 @@ gen_server:abcast     -----> Module:handle_cast/2
       in place of the time-out or hibernation value, which will immediately
       invoke the <c>handle_continue/2</c> callback.</p>
 
+    <p>If the <c>gen_server</c> process terminates, e.g.
+      as a result of a function in the callback module returning
+      <c>{stop,Reason,NewState}</c>, an exit signal with this <c>Reason</c>
+      is sent to linked processes and ports. See
+      <seeguide marker="system/reference_manual:processes#errors">
+      Processes</seeguide> in the Reference Manual for details
+      regarding error handling using exit signals.</p>
+
   </description>
 
   <funcs>
@@ -546,7 +554,10 @@ gen_server:abcast     -----> Module:handle_cast/2
           the function returns <c>{error,Reason}</c>. If
           <c>Module:init/1</c> returns <c>{stop,Reason}</c> or
           <c>ignore</c>, the process is terminated and the function
-          returns <c>{error,Reason}</c> or <c>ignore</c>, respectively.</p>
+          returns <c>{error,Reason}</c> or <c>ignore</c>, respectively.
+	  An exit signal with the same <c>Reason</c> (or <c>normal</c> if
+	  <c>Module:init/1</c> returns <c>ignore</c>) is sent to linked
+	  processes and ports.</p>
       </desc>
     </func>
 
@@ -609,6 +620,8 @@ gen_server:abcast     -----> Module:handle_cast/2
           <c>shutdown</c>, or <c>{shutdown,Term}</c> causes an
           error report to be issued using
           <seeerl marker="kernel:logger"><c>logger(3)</c></seeerl>.
+          An exit signal with the same reason is sent to linked processes
+	  and ports.
           The default <c>Reason</c> is <c>normal</c>.</p>
         <p><c>Timeout</c> is an integer greater than zero that
           specifies how many milliseconds to wait for the server to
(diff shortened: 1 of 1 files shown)


lib/stdlib/doc/src/gen_server.xml
```diff
@@ -99,6 +99,11 @@ gen_server:abcast     -----> Module:handle_cast/2
       in place of the time-out or hibernation value, which will immediately
       invoke the <c>handle_continue/2</c> callback.</p>
 
+    <p>If the <c>gen_server</c> process terminates with any reason other
+      than <c>normal</c>, e.g. as a result of a function in the callback
+      module returning <c>{stop,Reason,NewState}</c> with a reason other
+      than <c>normal</c>, an exit signal is emitted.</p>
```
ferd, line 105: I'm pretty sure you get a `normal` exit signal regardless and that you can see it with `trap_exit`, otherwise supervisors would have a hard time

zuiderkwast: Ah, right! I'll change this and other mentions of "normal".

zuiderkwast: Ah, you're right. I wanted to mention that the exit signal is the same as Reason in various places. I'll revert all mentions of "normal" in this PR.

rickard-green:
I think you should clarify that exit signals are sent to linked processes and ports in these scenarios. If nothing is linked to the server, no exit signals at all will be emitted.

zuiderkwast:
> I think you should clarify that exit signals are sent to linked processes and ports in these scenarios. If nothing is linked to the server, no exit signals at all will be emitted.

Thanks for the advice. I updated as suggested. I also added a link to the Reference Manual, page Processes, section Error Handling.

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -973,7 +984,10 @@ gen_server:abcast     -----> Module:handle_cast/2
           <c>proc_lib:hibernate/3</c></seemfa>).</p>
         <p>If the initialization fails, the function is to return
           <c>{stop,Reason}</c>, where <c>Reason</c> is any term, or
-          <c>ignore</c>.</p>
+          <c>ignore</c>. An exit signal with this reason is sent to linked
+          processes and ports, notably to the process starting the gen_server
+          when <seemfa marker="#start_link/3"><c>start_link/3,4</c></seemfa>
+          is used.</p>
```
rickard-green, line 990: If `ignore` is returned the `gen_server` process exits with exit reason `normal`.

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -546,7 +554,8 @@ gen_server:abcast     -----> Module:handle_cast/2
           the function returns <c>{error,Reason}</c>. If
           <c>Module:init/1</c> returns <c>{stop,Reason}</c> or
           <c>ignore</c>, the process is terminated and the function
-          returns <c>{error,Reason}</c> or <c>ignore</c>, respectively.</p>
+          returns <c>{error,Reason}</c> or <c>ignore</c>, respectively.
+	  An exit signal with the same reason is emitted.</p>
```
rickard-green, line 558: If `ignore` is returned the `gen_server` process exits with exit reason `normal`. I think it also should mention here that exit signals are sent to all linked processes and ports.

lib/stdlib/doc/src/gen_server.xml
zuiderkwast: Changing to "`An exit signal with this <c>Reason</c> (or with reason <c>normal</c> if <c>ignore</c> is returned) is sent to`..."

lib/stdlib/doc/src/gen_server.xml
zuiderkwast: Changing to "`An exit signal with the same <c>Reason</c> (or <c>normal</c> if
	  <c>Module:init/1</c> returns <c>ignore</c>) is sent to linked
	  processes and ports.`"

RaimoNiskanen:
I just want to remind that there is also the `gen_statem` behaviour documentation that might need some attention in this area.

rickard-green:
@zuiderkwast Can you add this to `gen_statem` as well?

zuiderkwast:
> @zuiderkwast Can you add this to gen_statem as well?

Maybe I can find time next week or so. Separate PR or the same one? I was hoping to keep this PR small andalso I'm not as familiar with the gen_statem manpage as with gen_server... Seems we can simply add the same content to the same sections as for gen_server (in Description, start_link, Module:init, Module:terminate), right?

(I made this PR because I was confused myself about the connection between `{stop, Reason, _}` and the reason in the exit signal and when I found out, I wanted to share it. Doing the extra copy-pasting for another module isn't that much fun however, so I wouldn't mind if someone else wants to do it...)

rickard-green:
@zuiderkwast no problem, it wasn't a requirement

@RaimoNiskanen @IngelaAndin gen_statem for the todo list

PR #2772, opened 2020-09-24
Fixes ERL-1335

fcristovao:
* When using `release_shell` option, the return is a pid, instead of the usual test return.
* For those cases, just skip that entry and continue result accumulation as usual

As requested in https://bugs.erlang.org/browse/ERL-1335

### DIFF
--- a/lib/common_test/src/ct_run.erl
+++ b/lib/common_test/src/ct_run.erl
@@ -1148,7 +1148,8 @@ run_all_specs([], _, _, TotResult) ->
 				    {Ok1,Fail1,{UserSkip1,AutoSkip1}}) ->
 					{Ok1+Ok,Fail1+Fail,
 					 {UserSkip1+UserSkip,
-					  AutoSkip1+AutoSkip}}
+					  AutoSkip1+AutoSkip}};
+				(Pid, Acc) when is_pid(Pid) -> Acc
 				end, {0,0,{0,0}}, TotResult1)
 	    end
     end;


fcristovao:
I don't see how my change would be succeeding on 32-bit, and failing on 64-bit?
I'm not sure what can I do here. Any help from the maintainers is appreciated :)

IngelaAndin:
I do not think this is your fault, so just ignore it for now and hopefully some of my colleagues will fix what ever is going on.

IngelaAndin:
We think this PR is good and if you could come up with a test case it would be even better. I got the advice to look at  
kernel/test/interactive_shell_SUITE.erl for inspiration on how to test.

IngelaAndin:
@fcristovao do you think you will be able to add a test?  That is always best practice. As it is an improvement to a test tool and not  critical to applications and also not Ericssons top prio, we might consider it for inclusion with an internal ticket for making an automated test case when we can find the time.  Alas that could be much later.

IngelaAndin:
We decided this will be included once we verified it does not break anything.

fcristovao:
@IngelaAndin I'm really sorry, I only got back now.
I hadn't checked this in a while. 
I see that it has been merged in the meantime, but if I can manage the time, I'll add a test for it.

Kind Regards,

IngelaAndin:
Thank you, it would be really nice if you could. Having regression test is always better. It is an exception that we merge a bug fix without, and it was a trade off between the possible harm that could be done if it breaks later and the possible benefit of having it fixed before it has a chance of making our priority list.

PR #2775, opened 2020-09-25
httpd_util: fix deprecation reference

attah:
Refer to erlang:integer_to_list/2, not integer_to_hexlist/2


ao-song:
Hi, Thanks for your PR, but since this is about deprecation reference, I have to update some other files as well rather than merge it directly here, you can find it in the latest master branch now. Thanks for pointing out this mistake.

PR #2778, opened 2020-09-29
Implement binary:linear_compare/2

starbelly:
- added linear_compare/2 to binary via erl_bif_binary
- added tests for main functionality
- added smoke test to check if a compiler optimization has caused the
  function to return early.
- updated binary docs

### DIFF
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2513,7 +2513,7 @@ static BIF_RETTYPE binary_linear_compare(Process *p, Eterm bin1, Eterm bin2)
         x |= b1[i] ^ b2[i];
     }
 
-    if (x == 0U)
+    if ((1 & ((x - 1) >> 8)) - 1 == 0)
         goto match;
 
     goto nomatch;
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2484,11 +2484,11 @@ static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
         if (i <= (size1 - 1) && i <= (size2 - 1))
             acc |= bytes1[i] ^ bytes2[i];
 
-        if ((i > (size1 - 1) && i <= (size2 - 1)))
-            acc |= bytes1[size1 - i] ^ bytes2[i];
-
         if ((i <= (size1 - 1) && i > (size2 - 1)))
             acc |= bytes1[i] ^ bytes2[size2 - 1];
+        
+        if ((i > (size1 - 1) && i <= (size2 - 1)))
+            acc |= bytes1[size1 - 1] ^ bytes2[i];
     }
 
     if (acc == 0)
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2481,7 +2481,7 @@ static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
     max_size = size1 > size2 ? size1 : size2;
 
     for (i=0; i < max_size; i++) {
-        if (i <= (size1 - 1) && i <= (size2 - 1)) { 
+        if (i < size1 && i < size2) {
             acc |= bytes1[i] ^ bytes2[i];
         } else {
             acc |= 'a' ^ 'b';
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -70,7 +70,7 @@ binary_match(Process *p, Eterm arg1, Eterm arg2, Eterm arg3, Uint flags);
 static BIF_RETTYPE
 binary_split(Process *p, Eterm arg1, Eterm arg2, Eterm arg3);
 
-static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2);
+static BIF_RETTYPE binary_linear_compare(Process *p, Eterm bin1, Eterm bin2);
 
 void erts_init_bif_binary(void)
 {
@@ -2459,38 +2459,38 @@ static int cleanup_copy_bin_state(Binary *bp)
     return 1;
 }
 
-static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
+static BIF_RETTYPE binary_linear_compare(Process *p, Eterm bin1, Eterm bin2)
 {
     byte *bytes1;
     byte *bytes2;
     Uint bitoffs1, bitsize1;
     Uint bitoffs2, bitsize2;
-    Uint size1, size2, max_size;
+    Uint size;
     size_t i;
-    volatile unsigned char acc = 0; 
-    const volatile char A = 'a';
-    const volatile char B = 'b';
+    volatile unsigned char x = 0U;
 
     if (is_not_binary(bin1) || is_not_binary(bin2))
+        goto bad_arg;
+
+    size = binary_size(bin1);
+
+    if (size != binary_size(bin2))
         goto bad_arg;
     
     ERTS_GET_BINARY_BYTES(bin1, bytes1, bitoffs1, bitsize1); 
     ERTS_GET_BINARY_BYTES(bin2, bytes2, bitoffs2, bitsize2);
+    
+    const volatile unsigned char *volatile b1 =
+        (const volatile unsigned char *volatile) bytes1;
 
-    size1 = binary_size(bin1);
-    size2 = binary_size(bin2);
+    const volatile unsigned char *volatile b2 =
+        (const volatile unsigned char *volatile) bytes2;
 
-    max_size = size1 > size2 ? size1 : size2;
-   
-    for (i=0; i < max_size; i++) {
-        if (i < size1 && i < size2) {
-            acc |= bytes1[i] ^ bytes2[i];
-        } else {
-            acc |= A ^ B;
-        }
+    for (i=0; i < size; i++) {
+        x |= b1[i] ^ b2[i];
     }
 
-    if (acc == 0)
+    if ((1 & ((x - 1) >> 8)) - 1 == 0)
         goto match;
 
     goto nomatch;
@@ -2503,9 +2503,9 @@ static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
     BIF_ERROR(p,BADARG);
 }
 
-BIF_RETTYPE binary_secure_compare_2(BIF_ALIST_2)
+BIF_RETTYPE binary_linear_compare_2(BIF_ALIST_2)
 {
-    return binary_secure_compare(BIF_P, BIF_ARG_1, BIF_ARG_2);
+    return binary_linear_compare(BIF_P, BIF_ARG_1, BIF_ARG_2);
 }
 
 /*
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2513,7 +2513,7 @@ static BIF_RETTYPE binary_linear_compare(Process *p, Eterm bin1, Eterm bin2)
         x |= b1[i] ^ b2[i];
     }
 
-    if ((1 & ((x - 1) >> 8)) - 1 == 0)
+    if (x == 0U)
         goto match;
 
     goto nomatch;
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2459,6 +2459,21 @@ static int cleanup_copy_bin_state(Binary *bp)
     return 1;
 }
 
+#ifdef HAVE_WEAK_SYMBOLS
+__attribute__((weak)) void
+_dummy_symbol_to_prevent_linear_compare_lto(const unsigned char *b1,
+                                           const unsigned char *b2,
+                                           const size_t         len);
+__attribute__((weak)) void
+_dummy_symbol_to_prevent_linear_compare_lto(const unsigned char *b1,
+                                           const unsigned char *b2,
+                                           const size_t         len)
+{
+    (void) b1;
+    (void) b2;
+    (void) len;
+}
+#endif
 static BIF_RETTYPE binary_linear_compare(Process *p, Eterm bin1, Eterm bin2)
 {
     byte *bytes1;
@@ -2479,12 +2494,20 @@ static BIF_RETTYPE binary_linear_compare(Process *p, Eterm bin1, Eterm bin2)
     
     ERTS_GET_BINARY_BYTES(bin1, bytes1, bitoffs1, bitsize1); 
     ERTS_GET_BINARY_BYTES(bin2, bytes2, bitoffs2, bitsize2);
-    
+
+#ifdef HAVE_WEAK_SYMBOLS
+    const unsigned char *b1 = (const unsigned char *) bytes1;
+    const unsigned char *b2 = (const unsigned char *) bytes2;
+#else
     const volatile unsigned char *volatile b1 =
         (const volatile unsigned char *volatile) bytes1;
-
     const volatile unsigned char *volatile b2 =
         (const volatile unsigned char *volatile) bytes2;
+#endif
+
+#if HAVE_WEAK_SYMBOLS
+    _dummy_symbol_to_prevent_linear_compare_lto(b1, b2, size);
+#endif
 
     for (i=0U; i < size; i++) {
         x |= b1[i] ^ b2[i];
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2481,11 +2481,14 @@ static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
     max_size = size1 > size2 ? size1 : size2;
 
     for (i=0; i < max_size; i++) {
-        if (i <= (size1 - 1) && i < (size2 - 1))
+        if (i <= (size1 - 1) && i <= (size2 - 1))
             acc |= bytes1[i] ^ bytes2[i];
 
-        if ((i > (size1 - 1) || i > (size2 - 1)) && acc == 0)
-            acc = 1;
+        if ((i > (size1 - 1) && i <= (size2 - 1)))
+            acc |= bytes1[size1 - i] ^ bytes2[i];
+
+        if ((i <= (size1 - 1) && i > (size2 - 1)))
+            acc |= bytes1[i] ^ bytes2[size2 - 1];
     }
 
     if (acc == 0)
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -2467,8 +2467,10 @@ static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
     Uint bitoffs2, bitsize2;
     Uint size1, size2, max_size;
     size_t i;
-    volatile unsigned char acc = 0;
-  
+    volatile unsigned char acc = 0; 
+    const volatile char A = 'a';
+    const volatile char B = 'b';
+
     if (is_not_binary(bin1) || is_not_binary(bin2))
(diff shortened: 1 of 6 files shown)


lib/stdlib/doc/src/binary.xml
```diff
@@ -575,6 +575,17 @@ store(Binary, GBSet) ->
       </desc>
     </func>
 
+    <func>
+      <name name="secure_compare" arity="2" since="OTP R24"/>
+      <fsummary>Compares two binaries in constant time.</fsummary>
+      <desc>
+          <p>Given two binaries this function will iterate and compare each byte of both binaries in constant time
```
starbelly, line 582: Docs could be improved here. Specifically some use case examples perhaps and when not to use this function. Possibly also link to https://en.wikipedia.org/wiki/Timing_attack

lib/stdlib/test/binary_module_SUITE.erl
```diff
@@ -1373,3 +1376,38 @@ check_no_invalid_read_bug(I) ->
     binary:encode_unsigned(N+N),
     binary:encode_unsigned(N+N, little),
     check_no_invalid_read_bug(I+1).
+
+secure_compare(Config) when is_list(Config) ->
+    true  = binary:secure_compare(<<"">>, <<"">>),
```
starbelly, line 1381: Could add some tests that check against random inputs.

starbelly: Could also add some tests to ensure that time to return is approximately the for two arguments regardless of their order, may be brittle.

erts/emulator/beam/erl_bif_binary.c
```diff
@@ -2457,6 +2459,53 @@ static int cleanup_copy_bin_state(Binary *bp)
     return 1;
 }
 
+static BIF_RETTYPE binary_secure_compare(Process *p, Eterm bin1, Eterm bin2)
```
starbelly, line 2462: This implementation differs from the original implementation in #2749 in that we do not short circuit if lengths do not match. Instead we always iterate based on the size of the largest binary which feels like the right thing to do. I think if you want to short circuit on differing lengths that is particular to a caller's case and thus should be handled by the caller.

erts/emulator/beam/erl_bif_binary.c
```diff
@@ -2457,6 +2459,53 @@ static int cleanup_copy_bin_state(Binary *bp)
...
+    size1 = binary_size(bin1);
+    size2 = binary_size(bin2);
+
+    max_size = size1 > size2 ? size1 : size2;
+
+    for (i=0; i < max_size; i++) {
+        if (i <= (size1 - 1) && i <= (size2 - 1)) { 
+            acc |= bytes1[i] ^ bytes2[i];
+        } else {
+            acc |= 'a' ^ 'b';
```
starbelly, line 2487: Added so we always perform an equivalent operation (striving for "near constant" time).

vans163:
This is fine but I dont see why you cant just do a `compare(), Rng = rand.uniform(10), receive after Rng -> ok end `

starbelly:
> This is fine but I dont see why you cant just do a `compare(), Rng = rand.uniform(10), receive after Rng -> ok end `

@vans163 Not sure I follow... could you explain this comment and suggestion more?

jhogberg:
> This is fine but I dont see why you cant just do a `compare(), Rng = rand.uniform(10), receive after Rng -> ok end `

Because additional delays still leak information. For the sake of argument let's say the operation takes either 0 or 10 seconds depending on input; with a random delay of 1-10 seconds you'd see it take 1-10s or 11-21s.

A less contrived scenario will of course be far more annoying for an attacker, but the leak will remain regardless of how you tweak the numbers, and it's just a question of time before they figure it out.

erts/emulator/beam/erl_bif_binary.c
yanmaani: This seems dodgy. The compiler will probably optimize a xor of constants out.
How about this:
```
acc |= bytes1[i rem size1] ^ bytes2[i rem size2];
acc |= size1 ^ size2;
```
Or alternately, this:
```
const volatile char A = 'a';
const volatile char B = 'b';
...
acc |= A ^ B;
```

okeuday: A variation on the above may help keep this simpler:

    {   
        volatile char c1 = '\0';
        volatile char c2 = '\0';
        if (i < size1)
            c1 = bytes1[i];
        if (i < size2)
            c2 = bytes2[i];
        acc |= c1 ^ c2;
    }

jhogberg: I hate to be a downer but `volatile` reads and writes still go through processor cache so all proposals thus far will leak the size of the secret. Comparing hashes would work fine, but only if the secret is hashed before it can be interacted with and we can't do that behind this interface.

I'd much rather leave all this to a library that does it for a living. Rolling your own crypto is fun, but I wouldn't trust something of my own design and neither should anyone else.

starbelly: @jhogberg The only other alternative is to do it in pure erlang. I will be back off from making changes until I hear either this PR is rejected in full, the C impl is good enough, or we want it in erlang.

okeuday: @jhogberg Your argument about the processor cache isn't relevant.  It is just best-effort to have a constant time compare.  All source code executes using the processor cache.  Here, volatile is used to avoid getting a register.

IngelaAndin: If it is just best effort why go to all this trouble trying to make it look secure.

okeuday: @IngelaAndin Having a constant time comparison is good, but the compiler can always insert malicious source code.  The hardware all generally lacks formal verification and if it was formally verified it is difficult to confirm hardware matches the proof.  Typically we use monolithic kernels for operating systems that lack formal verification.  Erlang/OTP lacks any formal verification.  Why do we bother with development?  To gradually improve on the situation, making the big ball of mud better in a gradual way.

yanmaani:
OpenSSL has `CRYPTO_memcmp`. It might make sense to just wrap that in `crypto`.

erts/emulator/beam/erl_bif_binary.c
starbelly: I mean that's my take.... it's best effort, there's no way we or anyone can actually get constant time. Not even with ASM AFAIK. I think this function is generally needed and desired.

starbelly:
> OpenSSL has `CRYPTO_memcmp`. It might make sense to just wrap that in `crypto`.

This was the original PR https://github.com/erlang/otp/pull/2749, it started with `CRYPTO_memcmp` but it's not documented and it was also thought that this function which really isn't a cryptographic function shouldn't be tied to the crypto module and thus unavailable if you didn't compile with openssl (or a compat library).

erts/emulator/beam/erl_bif_binary.c
starbelly: > A variation on the above may help keep this simpler:
> 
> ```
> {   
>     volatile char c1 = '\0';
>     volatile char c2 = '\0';
>     if (i < size1)
>         c1 = bytes1[i];
>     if (i < size2)
>         c2 = bytes2[i];
>     acc |= c1 ^ c2;
> }
> ```

I could be missing something, but switching to this, which is similar to an approach I initially tried returns true when it should not.

```
binary:secure_compare(<<"ttestxx">>, <<"ttestx">>).
true
```

okeuday: > I could be missing something, but switching to this, which is similar to an approach I initially tried returns true when it should not.
> 
> ```
> binary:secure_compare(<<"ttestxx">>, <<"ttestx">>).
> true
> ```
@starbelly I tried it as separate C source code and Erlang source code to show that acc should be 120 for your test when doing this approach:

    #include <stdio.h>
    #include <string.h>
    int main()
    {
        char const * const bytes1 = "ttestxx";
        char const * const bytes2 = "ttestx";
        size_t const size1 = strlen(bytes1);
        size_t const size2 = strlen(bytes2);
        size_t const max_size = (size1 > size2) ? size1 : size2;
        char acc = 0;
        for (size_t i = 0; i < max_size; i++)
        {
            volatile char c1 = '\0';
            volatile char c2 = '\0';
            if (i < size1)
                c1 = bytes1[i];
            if (i < size2)
                c2 = bytes2[i];
            acc |= c1 ^ c2;
        }
        printf("%d\n", acc);
        return 0;
    }



    -module(test).
    -export([compare_constant/2]).
    compare_constant(Test, [_ | _] = Correct) ->
        compare_constant(Test, Correct, 0) =:= 0.
 
    compare_constant([], [], Bits) ->
        io:format("~p~n", [Bits]),
        Bits;
    compare_constant([] = Test, [C2 | Correct], Bits) ->
        compare_constant(Test, Correct, Bits bor (0 bxor C2));
    compare_constant([C1 | Test], [] = Correct, Bits) ->
        compare_constant(Test, Correct, Bits bor (C1 bxor 0));
    compare_constant([C1 | Test], [C2 | Correct], Bits) ->
        compare_constant(Test, Correct, Bits bor (C1 bxor C2)).

    1> l(test).
    {module,test}
    2> test:compare_constant("ttestxx","ttestx").
    120
    false

jhogberg: > @jhogberg Your argument about the processor cache isn't relevant. It is just best-effort to have a constant time compare.  All source code executes using the processor cache. Here, volatile is used to avoid getting a register.

Why would I mention it if it isn't relevant, and why do you assume I don't know why they reached for `volatile`? There's no need to patronize me, if you think I'm unclear you should ask me to clarify instead. I'll readily accept that I was, things that seem obvious to me might not be to someone else.

```
acc |= bytes1[i % size1] ^ bytes2[i % size2];
```

Since we start over from the beginning of the shorter one once we're past its size, we'll have a timing difference once it wraps around since it's bound to be hot in cache (having visited it recently). Figuring out the size to the nearest multiple of the cache line size is easy, and with enough prodding you'll be able to figure out the exact size. It'll take time but it's not hard (assuming you control one of the inputs).

The other variants are vulnerable to similar attacks, easier ones at that.

Does knowing the size of the secret matter? I hope not, but it depends entirely on what the user is doing and I don't like advertising something as "secure" when I can't say that with confidence. Getting this right is _difficult_ and I'd much rather leave this to a library that does it for a living.

starbelly: @jhogberg I hear your points and did the first time and I respect them and your expertise in these areas. I guess at this point, I'm just not clear on what to do? Would renaming the function and eliminating the word secure from the doc suffice? Or are you completely opposed?

peerst:
Why not have a look at OpenSSLs function implementation to see what they are doing?

Would give us some insight and confidence to really have the least possible leaking function.

peerst:
And I think the comment of @IngelaAndin is very valid.  Making something look more secure than it can be is a dangerous thing.  If the function claims to replicate crypto functionality we need to make sure it’s at least as good as OpenSSLs implantation.  

And make it sure that to be really minimal leaking one should only use it on secure hashes.  Which does open the question again in what scenarios it’s really useful without having crypto?

josevalim:
I believe PR #2749 discussed the OpenSSL route and why it may be complicated to use it or copy it (private APIs, use of asm, etc).

It seems this function should not be called secure_compare, because we cannot really guarantee it is secure (can anyone?). So one suggestion would be to call it either linear_compare (requires both size to be equal and time is linear to size) or longest_compare (takes the size of the longest one). @jhogberg do you think those guarantees would be easier to provide? Or do you still think it would be hard?

okeuday:
OpenSSL has had development problems in the past and [LibreSSL](https://www.libressl.org/) was pursuing improving its source code.  Their source code for CRYPTO_memcmp is:

https://github.com/libressl-portable/openbsd/blob/f5193c1aee6e2ada87e97282a0344f03baa6f990/src/lib/libcrypto/cryptlib.c#L365-L377

    int
    CRYPTO_memcmp(const void *in_a, const void *in_b, size_t len)
    {
        size_t i;
        const unsigned char *a = in_a;
        const unsigned char *b = in_b;
        unsigned char x = 0;
 
        for (i = 0; i < len; i++)
            x |= a[i] ^ b[i];
 
        return x;
    }

It looks like it still remains undocumented in the OpenBSD manpages.  They avoid differing lengths by using a single length.  The alternative is to stick with the length of one array as the length of the correct data, not the data being tested.

starbelly:
> I believe PR #2749 discussed the OpenSSL route and why it may be complicated to use it or copy it (private APIs, use of asm, etc).
> 
> It seems this function should not be called secure_compare, because we cannot really guarantee it is secure (can anyone?). So one suggestion would be to call it either linear_compare (requires both size to be equal and time is linear to size) or longest_compare (takes the size of the longest one). @jhogberg do you think those guarantees would be easier to provide? Or do you still think it would be hard?

I just wanted to voice my support for not calling this `secure_` , I think `linear_` is a much better name! 💯 . I'm also ok with returning a badarg if the lengths differ.  FWIW this is the implementation I favor : https://github.com/jedisct1/libsodium/blob/ae4add868124a32d4e54da10f9cd99240aecc0aa/src/libsodium/sodium/utils.c#L179

peerst:
I'd go for badarg when length differ, that matches e.g. the libsodium  implementation where there is  only one length anyway.

This solves the problem of leaking the size because if someone uses this function for something where leaking the size makes a difference they are on a quite wrong path already and this prevents the abuse.  At the same time we get out of the problem of impossibility of successfully keeping the size secret (well its not quite impossible but would need a much higher effort).

okeuday:
@peerst If the single length is attached to the tested argument (not the argument holding the correct data), the function would be more useful:

https://github.com/CloudI/CloudI/blob/0f7c1d0e19f9c9b573f2ea818ebeb74757e14901/src/lib/cloudi_core/src/cloudi_string.erl#L283-L293

    compare_constant(Test, [_ | _] = Correct) ->
        compare_constant(Test, Correct, 0) =:= 0.
 
    compare_constant([], [], Bits) ->
        Bits;
    compare_constant([], [_ | _], _) ->
        1;
    compare_constant([C | Test], [] = Correct, Bits) ->
        compare_constant(Test, Correct, Bits bor (C bxor -1));
    compare_constant([C1 | Test], [C2 | Correct], Bits) ->
        compare_constant(Test, Correct, Bits bor (C1 bxor C2)).

starbelly:
> I'd go for badarg when length differ, that matches e.g. the libsodium implementation where there is only one length anyway.
> 
> This solves the problem of leaking the size because if someone uses this function for something where leaking the size makes a difference they are on a quite wrong path already and this prevents the abuse. At the same time we get out of the problem of impossibility of successfully keeping the size secret (well its not quite impossible but would need a much higher effort).

I agree.

jhogberg:
> This solves the problem of leaking the size because if someone uses this function for something where leaking the size makes a difference they are on a quite wrong path already and this prevents the abuse. At the same time we get out of the problem of impossibility of successfully keeping the size secret (well its not quite impossible but would need a much higher effort). 

+1. Higher effort means more room for mistakes.

> @peerst If the single length is attached to the tested argument (not the argument holding the correct data), the function would be more useful:

Your code leaks the size.

I think it's rather telling that the libraries we've found _don't allow the arguments to have different sizes_. I fail to see why we shouldn't follow their lead.

okeuday:
@jhogberg My example leaks the "Test" argument length in the same way the `CRYPTO_memcmp` example leaks the `len` variable.  However, I understand your desire to follow the lead of other software for Erlang/OTP.

jhogberg:
> @jhogberg My example leaks the "Test" argument length in the same way the `CRYPTO_memcmp` example leaks the `len` variable. However, I understand your desire to follow the lead of other software for Erlang/OTP.

You leak the length of the `Correct` argument as well.

okeuday:
@jhogberg The example at https://github.com/erlang/otp/pull/2778#issuecomment-706764742 only leaks the size of "Test" because that length is used for the comparison.

jhogberg:
> @jhogberg The example at [#2778 (comment)](https://github.com/erlang/otp/pull/2778#issuecomment-706764742) only leaks the size of "Test" because that length is used for the comparison.

This is getting tiresome. I would not have said it unless I could prove it.

```erlang
compare_constant([C | Test], [] = Correct, Bits) ->
    compare_constant(Test, Correct, Bits bor (C bxor -1));
compare_constant([C1 | Test], [C2 | Correct], Bits) ->
    compare_constant(Test, Correct, Bits bor (C1 bxor C2)).
```

1. Extracting values from a list cell takes longer than not doing that.
2. `bxor` with a compile-time constant is faster than with a variable.

As the time taken per cell decreases once `Test` is longer than `Correct`, the attacker can infer the size by prodding with different lengths for `Test`.

jhogberg:
@starbelly I apologize for derailing the thread. I'm happy to have this functionality, and I'm confident you'll arrive at a good solution.

starbelly:
> @starbelly I apologize for derailing the thread. I'm happy to have this functionality, and I'm confident you'll arrive at a good solution.

No need to apologize. This is been a highly education set of threads and it's not only good for me, it's good for anyone who comes across this in the future. Cheers for the call outs; you're doing what is right! :)

okeuday:
@jhogberg My view was that the difference between a compile-time constant and the variable should be negligible, at least for pursuing the best solution in Erlang source code, given the currently available Erlang/OTP functions.  The main purpose for providing the example was to communicate the idea, which could be implemented in C to avoid the potential impact from Erlang optimizations (and JIT).

yanmaani:
Couldn't you implement it for identical lengths at first, and add in a version for nonidentical lengths later on if needed?

```
-module(cc).
-export([linear_compare/2]).

linear_compare(<<>>, <<>>, Bits) ->
    Bits;
linear_compare(<<C1:8, Test/binary>>, <<C2:8, Correct/binary>>, Bits) ->
    linear_compare(Test, Correct, Bits bor (C1 bxor C2)).
linear_compare(Test, Correct) ->
    linear_compare(Test, Correct, 0) =:= 0.
```

erts/emulator/beam/erl_bif_binary.c
```diff
@@ -2457,6 +2459,55 @@ static int cleanup_copy_bin_state(Binary *bp)
...
+
+    size = binary_size(bin1);
+
+    if (size != binary_size(bin2))
+        goto bad_arg;
+    
+    ERTS_GET_BINARY_BYTES(bin1, bytes1, bitoffs1, bitsize1); 
+    ERTS_GET_BINARY_BYTES(bin2, bytes2, bitoffs2, bitsize2);
+    
+    const volatile unsigned char *volatile b1 =
```
starbelly, line 2483: @jhogberg @IngelaAndin This is a best effort to prevent compiler optimizations. It makes sense to me, but it could also be seen as a little too hand wavey. Thoughts? 

Note I did not do the weak dummy function hack as done in libsodium as that seemed to me as a form of mitigation for a very specific case that does not apply to us (i.e., the case where another lib in C has hard coded values it checks against).

jhogberg:
I wanted to leave this alone, but I'll reply since @starbelly appreciates it:

> @jhogberg My view was that the difference between a compile-time constant and the variable should be negligible

The difference between bailing out early and sweeping through the entire thing is also "negligible." That doesn't mean an attacker can't figure it out with enough patience.

> The main purpose for providing the example was to communicate the idea

That idea is fundamentally flawed; _you do different things depending on how your inputs compare_ (lengthwise in this case), and different things take different time to execute. Anything with a branch is obviously bad, but timing differences arise even when it's not immediately apparent as I showed [in an earlier comment](https://github.com/erlang/otp/pull/2778#discussion_r502716767).

It doesn't even matter if you write it in assembly, you'll end up leaking one way or the other and it's much simpler to follow everyone else and disallow arguments with different sizes.

I think we've wasted enough time on this, I'm not going to discuss it further.

erts/emulator/beam/erl_bif_binary.c
jhogberg: The commit message hints at there being a good reason for this by pointing to two dead links. :(

https://github.com/jedisct1/libsodium/commit/113091b2a0fbc999f6ef585c10bd622e94f8b6eb

HansN:
This PR is closed as of a decision in the OITP Technical Board.
See the JIRA [ERL-1385](https://bugs.erlang.org/browse/ERL-1385) for the notes and for a means to make comments or questions.

PR #2779, opened 2020-10-01
compiler: Suppress stack traces for throw/1 when they're not used

jhogberg:
The runtime scans the stack twice when exceptions are raised; once to build a stack trace, and again to find the appropriate handler (see `handle_error` in `beam_common.c` for an explanation why). While the first scan finishes as soon as it has gathered enough frames (8 by default), it de-duplicates adjacent entries which makes `throw/1` more expensive than just returning up the stack in some scenarios.

This PR improves the performance of `throw/1` by skipping the stack trace if we can prove that it will never be needed. It doesn't do much on its own, but opens up for further optimizations that will eliminate stack scanning altogether.

### DIFF
--- a/lib/compiler/src/beam_ssa_throw.erl
+++ b/lib/compiler/src/beam_ssa_throw.erl
@@ -1,0 +1,495 @@
+%%
+%% %CopyrightBegin%
+%%
+%% Copyright Ericsson AB 2020-2020. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% %CopyrightEnd%
+%%
+
+%%%
+%%% This pass skips stack trace generation on erlang:throw/1 when they're
+%%% guaranteed not to be inspected.
+%%%
+%%% It does so by finding all exception handlers in the module, determining
+%%% which functions they cover, and then simulating the result of every throw/1
+%%% in the module; if a particular throw/1 never reaches an instruction that
+%%% creates a stack trace, then it's safe to rewrite it.
+%%%
+
+-module(beam_ssa_throw).
+
+-export([module/2]).
+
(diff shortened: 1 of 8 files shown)


lib/compiler/src/beam_ssa_type.erl
```diff
@@ -1671,6 +1671,18 @@ type(call, [#b_literal{val=Fun} | Args], _Anno, _Ts, _Ds) ->
...
+    case map_get(V, Ds) of
+        #b_set{op=landingpad} when Idx =:= 0 ->
+            %% Class
+            #t_atom{elements=[error,exit,throw]};
+        #b_set{op=landingpad} when Idx =:= 1 ->
+            %% Reason
+            any;
+        #b_set{op=landingpad} when Idx =:= 2 ->
+            %% Stack trace
+            #t_list{}
```
bjorng, line 1692: The raw stacktrace is an opaque data type that the compiler should not know anything about. The type should be `any`. (There is no difference in `diffable` if I change the type to `any`.)

jhogberg: Good point, I'll fix it here and in the validator.

lib/compiler/src/beam_ssa_throw.erl
```diff
@@ -0,0 +1,495 @@
...
+si_handler_end(Is, Id, EndLbl, Last, Stacktrace, Lst0, Gst) ->
+    #lst{suitability=Suitability0} = Lst0,
+
+    Marker = case Suitability0 of
+                 #{ Stacktrace := {tentative, _} } ->
+                     %% Multiple uses make partitioning more difficult, so
+                     %% we'll disqualify this handler to save us the hassle.
+                     unsuitable;
+                 #{} ->
+                     {tentative, EndLbl}
```
bjorng, line 252: I think it would be useful to add a comment similar to the one in `si_handler_start` to note that handler's suitability depends on what is actually thrown.

lib/compiler/src/beam_ssa_throw.erl
```diff
@@ -0,0 +1,495 @@
...
+                       Path = beam_ssa:between(StartLbl, EndLbl, Preds, Blocks),
+                       Partition = si_copy_blocks(Path, Blocks, #{}),
+
+                       Handler = {tentative, StartLbl, Vars, Partition},
+                       Handlers0#{ HandlerId => Handler };
+                   #{} when Stacktrace =/= none ->
+                       %% Stack trace is used in an unsupported manner.
+                       Handlers0#{ HandlerId => unsuitable };
+                   #{} when Stacktrace =:= none ->
+                       Handlers0#{ HandlerId => suitable }
```
bjorng, line 230: Just for clarity, a comment that notes that the stacktrace is never used would be useful.

bjorng (approved):

PR #2781, opened 2020-10-06
Expand the spec for erl_epmd:listen_port_please/2

pjk25:
The spec for erl_epmd:listen_port_please/2 is overly strict, resulting in
dialyzer errors for the valid snippet:

[Name, Host] = string:split(atom_to_list(node()), "@"),
R = erl_epmd:port_please(Name, Host),

This commit relaxes the spec for listen_port_please to match that of port_please

### DIFF
--- a/lib/kernel/src/erl_epmd.erl
+++ b/lib/kernel/src/erl_epmd.erl
@@ -130,8 +130,8 @@ getepmdbyname(HostName, _Timeout) ->
     {ok, HostName}.
 
 -spec listen_port_please(Name, Host) -> {ok, Port} when
-      Name :: atom(),
-      Host :: string() | inet:ip_address(),
+      Name :: atom() | string(),
+      Host :: atom() | string() | inet:ip_address(),
       Port :: non_neg_integer().
 listen_port_please(_Name, _Host) ->
     try


garazdawi:
There is a conflict here in the documentation about what the Erlang/OTP implementation of `erl_epmd` accepts, and what third-party implementations need to accept. I don't think we should change this as it would mean that third-party implementation may break.

pjk25:
@garazdawi I was until now unaware of the possibility of alternate discovery modules, and I agree with your concern. However, would it not be reasonable to then tighten the spec for `erl_epmd:port_please/2` and `erl_epmd:port_please/3`? The call is in fact unsafe with a string anyway, but the tighter spec would result in a more readable dialyzer warning. If one is calling the function with a string and knows the default discovery mechanism is in place, they will need to ignore a dialyzer either way, or instead unnecessarily convert the string to an atom before the call anyway.

garazdawi:
On second thought, I think that we can change the spec for listen_port as you described.

We did it for port_please (#2117) recently and the listen_port callback is relatively new so we should be able to expand it without breaking too many usages.

pjk25:
Great, that sounds good. Also, I realized I misnamed the function in my commit message (although the code is fine). The altered spec is for `erl_epmd:listen_port_please/2`.

garazdawi:
Would you mind fixing the commit message to be correct so that we can search in the commit log for this function in the future?

pjk25:
Sure, the message has been updated.

garazdawi:
Thanks for your contribution!

PR #2785, opened 2020-10-06
Simplify lint vts

richcarl:
Make the pattern variable tracking in erl_lint somewhat intelligible by removing unused parameters, changing misleading names, and correcting and improving comments.

Note that the Bvt0 argument to pattern_bin() was always [], so there was no point in passing it around the main pattern() loop.


bjorng:
Thanks for the pull request!

PR #2787, opened 2020-10-06
Fix os_mon application (disksup). replace hardcoded path for 'df' command in linux environment

halturin:
This PR is resolving issue https://bugs.erlang.org/browse/ERL-1370 and related to https://github.com/NixOS/nixpkgs/issues/99387

Retrieving some details of available disks is implemented via regular `df` command with a hardcoded path to it for all the OSs the Erlang does support.
https://github.com/erlang/otp/blob/master/lib/os_mon/src/disksup.erl#L242

In NixOS the location of `df` command differs from the defined location in [FHL](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). 

Solution: replace the hardcoded path to `df`command with a more flexible way via `env`.

### DIFF
--- a/lib/os_mon/src/disksup.erl
+++ b/lib/os_mon/src/disksup.erl
@@ -240,7 +240,7 @@ check_disk_space({unix, irix}, Port, Threshold) ->
     Result = my_cmd("/usr/sbin/df -lk",Port),
     check_disks_irix(skip_to_eol(Result), Threshold);
 check_disk_space({unix, linux}, Port, Threshold) ->
-    Result = my_cmd("/bin/df -lk -x squashfs", Port),
+    Result = my_cmd("/usr/bin/env df -lk -x squashfs", Port),
     check_disks_solaris(skip_to_eol(Result), Threshold);
 check_disk_space({unix, posix}, Port, Threshold) ->
     Result = my_cmd("df -k -P", Port),
--- a/lib/os_mon/src/disksup.erl
+++ b/lib/os_mon/src/disksup.erl
@@ -228,6 +228,19 @@ newline([13|_], B) -> {ok, lists:reverse(B)};
 newline([H|T], B) -> newline(T, [H|B]);
 newline([], B) -> {more, B}.
 
+%%-- Looking for Cmd location ------------------------------------------
+find_cmd(Cmd) ->
+    os:find_executable(Cmd).
+
+find_cmd(Cmd, Path) ->
+    %% try to find it at the specific location
+    case os:find_executable(Cmd, Path) of
+        false ->
+            find_cmd(Cmd);
+        Found ->
+            Found
+    end.
+
 %%--Check disk space----------------------------------------------------
 
 check_disk_space({win32,_}, not_used, Threshold) ->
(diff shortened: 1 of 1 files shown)


mikpe (approved):

KennethL:
I think we have absolute paths for a good reason and this PR using /usr/bin/env is not different from just executing "df -lk..." directly as I can understand it. Both "/usr/bin/env df..." and "df..." will traverse the PATH and execute the first `df` it finds.
The problem is if the PATH contains references to for example NFS mounted directories and something goes wrong with the NFS access, then this Erlang process and a scheduler might be blocked for minutes. And this happens even if `df` always is available on local disk. It just need an NFS mounted path before the `df`s path to make this vulnerable.
At least this is what I remember about why we have absolute paths here and in some other places where it seems unnecessary.

halturin:
@KennethL , thanks for the comment. As I've already mentioned in comments for https://bugs.erlang.org/browse/ERL-1370 the behavior of `env` differs from just executing `df ...` - it ignores an alias for `df` if it was defined before. 

I see your concern about NFS and erlang scheduler. But using hardcoded path for these commands is not preventing such kind of problem. I'm not sure if should resolve any OS environment issues this way. OS must provide a correct and working environmend for the erlang instance. I believe, there is a bunch of ways to make erlang vm crazy )

KennethL:
I think that if the absolute path to `df` is residing on the local file
system and that other directories are mounted via NFS then you are safe and
can not be blocked in executing `df` because of NFS problems. We have had
such cases with real customers. Correct me if I am wrong. I agree that the
OS environment must be working correctly, but in this case it
is nothing wrong with having NFS mounted directories in the PATH it is just
wrong to rely on the PATH if you want a robust Erlang system. If there was
a specific environment variable telling the path to `df` that would be
better. I don't understand why you say that a hard coded path to `df`
(known to be local in most cases) is not preventing the problem. I can
agree that it does not remove the problem to 100% but it reduce it to be
very unlikely.

I am afraid we will break some of our customers' systems if we change this
for Linux in general. I would suggest an approach where first the hard
coded path is tested and if it does not work os:find_executable is used to
find where `df` is. With this we get the same behavior as today for all
systems where `/usr/bin/df` is present and will find `df` in another place
as well.
If we don't what to do this for every call of `df` we can do it on startup
of os_mon or disk_sup and remember the path and use it for the next calls.
We should also add comments in the code explaining why we have absolute
paths.

halturin:
@KennethL thanks for the update. I agree it's a good idea to use os:find_executable for that. just updated my PR. 

As a first attempt we are trying to find `df` at a specified location (e.g. `os:find_executable("df", "/bin")` ) and if it wasn't found there we are making another attempt via `os:find_executable("df")` which provides the same result as `env` command. So, we don't need `env` anymore here. 

we can also use this approach for the rest.

halturin:
do I need to update anything else in my PR?

garazdawi:
I was just waiting for the tests to make sure that everything passed. Thanks for your contribution!

deliciouslytyped:
@KennethL Just a general comment, if people think it's important to keep corner cases like that working they should really add tests or comment it and not need someone to raise their hand in a PR for anyone to notice... :(

deliciouslytyped:
Also, wouldn't the way to solve that for everyone be to add extra logic to find_Executable, and make sure people use that function when trying to...find executables?

garazdawi:
Yes, I will add such a comment.

yorickvP:
If erlang requires `df` as a runtime dependency, the path should probably be configurable during build time.

PR #2789, opened 2020-10-08
Refactor ad-hoc entropy acquisition calls

yanmaani:
This patch amends four modules in `kernel` to use `rand:uniform/1` instead of ad-hoc combinations of `erlang:monotonic_time/0` and `erlang:unique_integer/0`. This makes the codebase slightly cleaner. It also means that any future improvements in `rand` will affect these parts of the codebase.


bjorng (changes requested):

bjorng:
Closing due to inactivity.

PR #2791, opened 2020-10-09
Fold over queues

Maria-12648430:
The first commit in this PR introduces a `fold/3` function for the `queue` module. Up to now, this had to be done by converting a queue to a list and folding over that, which is both cumbersome and more performance-intensive (`O(length(Rear))` for the conversion, and `O(length(List))` for the fold over that list) than a fold over the internal lists making up the `Rear` and `Front`, inside the `queue` module (`O(len(Queue))`, roughly equivalent to only `O(length(List))`). The queue is traversed in queue order (front to rear), which is what would be expected.

<hr />

The second commit in this PR addresses a "glitch" in the `filter/2` function, namely that while returning `true` or `false` from the filter function is semantically equivalent to returning `[Item]` or `[]` respectively, the latter generates more garbage. By handling the two special cases "single-element list" and "empty list" the same as `true` and `false` respectively, this is no longer the case, and the paragraph in the function documentation has been removed accordingly.

Also, to align `queue:filter/2` more closely with `lists:filtermap/2` (`queue:filter/2` is really a filtermap with insert capability), the filter function may return `{true, NewItem}` to signal that the current item should be replaced with `NewItem` in the result queue.

### DIFF
--- a/lib/stdlib/doc/src/queue.xml
+++ b/lib/stdlib/doc/src/queue.xml
@@ -54,6 +54,7 @@
 
     <p>All operations have an amortized O(1) running time, except
       <seemfa marker="#filter/2"><c>filter/2</c></seemfa>,
+      <seemfa marker="#fold/3"><c>fold/3</c></seemfa>,
       <seemfa marker="#join/2"><c>join/2</c></seemfa>,
       <seemfa marker="#len/1"><c>len/1</c></seemfa>,
       <seemfa marker="#member/2"><c>member/2</c></seemfa>,
@@ -131,6 +132,27 @@
           as returning <c>[]</c> is semantically equivalent to
           returning <c>false</c>. But returning a list builds
           more garbage than returning an atom.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="fold" arity="3" since=""/>
+      <fsummary>Fold a function over a queue.</fsummary>
+      <desc>
+        <p>Calls <c><anno>Fun</anno>(<anno>Item</anno>, <anno>AccIn</anno>)</c>
+          on successive items <c>Item</c> of <c>Queue</c>, starting
+          with <c><anno>AccIn</anno> == <anno>Acc0</anno></c>. The queue is
+          traversed in queue order, that is, from front to rear.
+          <c><anno>Fun</anno>/2</c> must return a new accumulator, which is
+          passed to the next call. The function returns the final value of
+          the accumulator. <c><anno>Acc0</anno></c> is returned if the queue is
+          empty.</p>
+        <p><em>Example:</em></p>
+        <pre>
+> <input>queue:fold(fun(X, Sum) -> X + Sum end, 0, queue:from_list([1,2,3,4,5])).</input>
+15
+> <input>queue:fold(fun(X, Prod) -> X * Prod end, 1, queue:from_list([1,2,3,4,5])).</input>
+120</pre>
       </desc>
     </func>
 
--- a/lib/stdlib/doc/src/queue.xml
+++ b/lib/stdlib/doc/src/queue.xml
@@ -54,6 +54,7 @@
 
     <p>All operations have an amortized O(1) running time, except
       <seemfa marker="#filter/2"><c>filter/2</c></seemfa>,
+      <seemfa marker="#filtermap/2"><c>filtermap/2</c></seemfa>,
       <seemfa marker="#fold/3"><c>fold/3</c></seemfa>,
       <seemfa marker="#join/2"><c>join/2</c></seemfa>,
       <seemfa marker="#len/1"><c>len/1</c></seemfa>,
@@ -132,6 +133,21 @@
           as returning <c>[]</c> is semantically equivalent to
           returning <c>false</c>. But returning a list builds
           more garbage than returning an atom.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="filtermap" arity="2" since=""/>
+      <fsummary>Filter and map a queue.</fsummary>
+      <desc>
+        <p>Returns a queue <c><anno>Q2</anno></c> that is the result of calling
+          <c><anno>Fun</anno>(<anno>Item</anno>)</c> on all items in
+          <c><anno>Q1</anno></c>, in order from front to rear.</p>
+        <p>If <c><anno>Fun</anno>(<anno>Item</anno>)</c> returns <c>true</c>,
+          <c>Item</c> is copied to the result queue. If it returns <c>false</c>,
+          <c><anno>Item</anno></c> is not copied. If it returns
+          <c>{true, NewItem}</c>, the queue element at this position is replaced
+          with <c>NewItem</c> in the result queue.</p>
       </desc>
     </func>
 
--- a/lib/stdlib/src/queue.erl
+++ b/lib/stdlib/src/queue.erl
@@ -27,7 +27,7 @@
 -export([get/1,get_r/1,peek/1,peek_r/1,drop/1,drop_r/1]).
 
 %% Higher level API
--export([reverse/1,join/2,split/2,filter/2,fold/3]).
+-export([reverse/1,join/2,split/2,filter/2,filtermap/2,fold/3]).
 
 %% Okasaki API from klacke
 -export([cons/2,head/1,tail/1,
@@ -400,6 +400,42 @@ filter_r(Fun, [X|R0]) ->
 	    R;
 	L when is_list(L) ->
 	    lists:reverse(L, R)
+    end.
+
+%% Filter and map a queue, traverses in queue order.
+%%
+%% O(len(Q1))
+-spec filtermap(Fun, Q1) -> Q2 when
+      Fun :: fun((Item) -> boolean() | {'true', Value}),
+      Q1 :: queue(Item),
+      Q2 :: queue(Item | Value),
+      Item :: term(),
+      Value :: term().
+filtermap(Fun, {R0, F0}) when is_function(Fun, 1), is_list(R0), is_list(F0) ->
+    F = lists:filtermap(Fun, F0),
+    R = filtermap_r(Fun, R0),
+    if R =:= [] ->
+	    f2r(F);
+       F =:= [] ->
+	    r2f(R);
+       true ->
+	    {R,F}
+    end;
+filtermap(Fun, Q) ->
+    erlang:error(badarg, [Fun,Q]).
+
+%% Call Fun in reverse order, i.e tail to head
+filtermap_r(_, []) ->
+    [];
+filtermap_r(Fun, [X|R0]) ->
+    R = filtermap_r(Fun, R0),
+    case Fun(X) of
+	true ->
+	    [X|R];
+	{true, Y} ->
+	    [Y|R];
+	false ->
+	    R
     end.
 
 %% Fold a function over a queue, in queue order.
--- a/lib/stdlib/src/queue.erl
+++ b/lib/stdlib/src/queue.erl
@@ -373,7 +373,11 @@ filter_f(Fun, [X|F]) ->
     case Fun(X) of
 	true ->
 	    [X|filter_f(Fun, F)];
+	[Y] ->
+	    [Y|filter_f(Fun, F)];
 	false ->
+	    filter_f(Fun, F);
+	[] ->
 	    filter_f(Fun, F);
 	L when is_list(L) ->
 	    L++filter_f(Fun, F)
@@ -388,7 +392,11 @@ filter_r(Fun, [X|R0]) ->
     case Fun(X) of
 	true ->
 	    [X|R];
+	[Y] ->
+	    [Y|R];
 	false ->
+	    R;
+	[] ->
 	    R;
 	L when is_list(L) ->
 	    lists:reverse(L, R)
--- a/lib/stdlib/src/queue.erl
+++ b/lib/stdlib/src/queue.erl
@@ -27,7 +27,7 @@
 -export([get/1,get_r/1,peek/1,peek_r/1,drop/1,drop_r/1]).
 
 %% Higher level API
--export([reverse/1,join/2,split/2,filter/2]).
+-export([reverse/1,join/2,split/2,filter/2,fold/3]).
 
 %% Okasaki API from klacke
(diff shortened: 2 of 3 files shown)


jhogberg:
Thanks for the PR!

We'll need to run it past the technical board (next week) as it changes an interface, but I think it looks great.

Maria-12648430:
> We'll need to run it past the technical board (next week) as it changes an interface, but I think it looks great.

You mean because of the `{true, NewItem}` as a new possible return for the filter function?

jhogberg:
> You mean because of the `{true, NewItem}` as a new possible return for the filter function?

Yes, though we'll discuss `fold/3` as well. We want as many eyeballs as possible on the documentation.

josevalim:
Nice, I would definitely have used `queue:fold/3` in the past! 

I am not convinced about the return type for `queue:filter` though. I understand the rationale in aligning it with filtermap but given they are already different (filter allows a list to be returned), a new return type complicates the API and doesn't add anything over returning `[NewItem]` directly.

RaimoNiskanen:
In that case, @josevalim & @Maria-12648430, maybe write a new `queue:filtermap` that aligns with `lists:filtermap` and silently let `queue:filter` keep its pecularity of being able to also do filtermap.  Oddly enough `queue:filter` is actually more general than `lists:filtermap` since the former allows replacing an element with any number of elements.

@Maria-12648430 - Note that your change to optimize the code handling of `[Item]` does not change the amount of created garbage. The garbage is created when creating the return value `[Item]` which is one cons cell referring to `Item`, and that cons cell (2 VM words) becomes garbage when building the queue internal result list. Both `[Item | filter_f(Fun, F)]`, and `L ++ filter_f(Fun, F)` does exactly the same: every cons cell in `L` (1 pcs) becomes garbage and a new one is created to hold the list that is appended to.  Returning `true` does not produce garbage because an atom is an immediate value.

Your change optimizes away external calls to `lists:append` and `lists:reverse` for the important case of replacing an item (filtermap), which I think is a good optimization - keep it.  But you can actually not remove the documentation about more created garbage.

Side note: returning `{true, NewItem}` produces 3 VM words of garbage since a tuple has a header word and then one per item, which is 50% more than when returning `[NewItem]`.

A question for decision: would we need both `queue:foldl` and `queue:foldr` instead of just `queue:fold`?  I think maybe not, since a `queue:reverse` is O(1).

josevalim:
> A question for decision: would we need both queue:foldl and queue:foldr instead of just queue:fold? I think maybe not, since a queue:reverse is O(1).

I pondered the same and my conclusion was that if there is ever a need for a “reverse fold”, it could be called “fold_r”, which aligns well with the other functions in the module. WDYT?

okeuday:
You may want to consider the need for a separate function (possibly in a separate pull request) that would avoid processing the whole queue.  I have found it necessary, due to testing of the time taken in `queue:filter/2`, to create a separate function to remove a single unique value from a queue (https://github.com/okeuday/pqueue/blob/ebfe8e76e87f05caae84827334662726629f5f54/src/pqueue4.erl#L11602-L11619).  With a situation like that, `queue:fold` is unable to help.  It is definitely good to have a `queue:fold` function.  When removing a unique value from a queue, the order doesn't matter, so the queue lists have minimal impact (i.e., no reverses or concatenations).

RaimoNiskanen:
@josevalim: `fold` and `fold_r` (the latter if needed).  I buy that!

RaimoNiskanen:
@okeuday: A napkin calculation of a possible `queue:remove/2` function:  It would, on average, only have to traverse half of the queue, and only have to rebuild on quarter of the cons cells (the ones preceeding the found element in either the front or the rear list), so it would be between 2..4 times faster than `queue:filter/2`, but still O(N).  Sounds like a nice feature anyway.

Would one want both a `queue:remove(PredicateFun, Queue)` and a `queue:delete(Item, Queue)` for a predicate fun and a `=:=` match respectively?  Matching equality is significantly faster than calling a fun().

okeuday:
@RaimoNiskanen I needed the predicate fun to access a tuple element but the `=:=` equality match function should be helpful for other data.

Maria-12648430:
Oh, quite a response :)

<hr />

@RaimoNiskanen  & @josevalim:
> In that case, @josevalim & @Maria-12648430, maybe write a new `queue:filtermap` that aligns with `lists:filtermap` and silently let `queue:filter` keep its pecularity of being able to also do filtermap. Oddly enough `queue:filter` is actually more general than `lists:filtermap` since the former allows replacing an element with any number of elements.

Ok, I second that. So, the filtermap function may return `true` to keep, `false` to drop, and `{true, NewItem}` to replace. No insert capability, right?

Maria-12648430:
@RaimoNiskanen 
> @Maria-12648430 - Note that your change to optimize the code handling of `[Item]` does not change the amount of created garbage. The garbage is created when creating the return value `[Item]` which is one cons cell referring to `Item`, and that cons cell (2 VM words) becomes garbage when building the queue internal result list. Both `[Item | filter_f(Fun, F)]`, and `L ++ filter_f(Fun, F)` does exactly the same: every cons cell in `L` (1 pcs) becomes garbage and a new one is created to hold the list that is appended to. Returning `true` does not produce garbage because an atom is an immediate value.

Ah, got that wrong then.

> Your change optimizes away external calls to `lists:append` and `lists:reverse` for the important case of replacing an item (filtermap), which I think is a good optimization - keep it. But you can actually not remove the documentation about more created garbage.

Ok, I'll add it back.

> Side note: returning `{true, NewItem}` produces 3 VM words of garbage since a tuple has a header word and then one per item, which is 50% more than when returning `[NewItem]`.

That would be obsoleted with a separate `filtermap` function.

Maria-12648430:
@RaimoNiskanen & @josevalim 
> A question for decision: would we need both `queue:foldl` and `queue:foldr` instead of just `queue:fold`? I think maybe not, since a `queue:reverse` is O(1).

TBH, I had both `foldl` and `foldr` in the beginning, but then removed them for `fold` because of the reason you described.

> @josevalim: `fold` and `fold_r` (the latter if needed). I buy that!

Decided then? :)

Maria-12648430:
@okeuday 
> You may want to consider the need for a separate function (possibly in a separate pull request) that would avoid processing the whole queue. I have found it necessary, due to testing of the time taken in `queue:filter/2`, to create a separate function to remove a single unique value from a queue

I need something like that every now and then, too. Maybe later, in a seperate PR :)

> With a situation like that, `queue:fold` is unable to help.

Yes

> When removing a unique value from a queue, the order doesn't matter, so the queue lists have minimal impact (i.e., no reverses or concatenations).

Hm, I would argue that it is not guaranteed that an item is unique, at least not from the perspective of the queue.

I would suggest a function (or two) that removes the first element matching the condition (either predicate function or exact match), then stops traversing further.

The front list can be traversed without reversal, so best case would be if the element is in that list (ie, near the front). The rear lists has to be reversed to find the first element (instead of the last), so that would be a worse case. Worst case is if no element is found, which means full traversal plus a reversal in the middle.

A more general version of this would be something like `remove(N, PredicateFun, Queue)` or `delete(N, Item, Queue)`, meaning "remove (at most) `N` matching elements".

Maria-12648430:
@okeuday & @RaimoNiskanen

Another function that could be handy would be `contains(PredicateFun, Queue)`, as the existing `member/` function only finds exact matches. Slower than `member/2` which employs `lists:member/2`, yes, but removes the need for converting to list and such. More general forms that allow more flexibility even would be `any` and `all`, like their counterparts in `lists`. What do you think?

(Probably material for another PR, anyway ;))

RaimoNiskanen:
@Maria-12648430 
> > Side note: returning `{true, NewItem}` produces 3 VM words of garbage since a tuple has a header word and then one per item, which is 50% more than when returning `[NewItem]`.
> 
> That would be obsoleted with a separate `filtermap` function.

Obsolete or not, it would still be so that `queue:filter/2` with a `Fun` returning `[Item]` would produce slightly less garbage than`queue:filtermap/2` with a `Fun` returning `{true, NewItem}`.  Certainly irrelevant.
Nevertheless, `queue:filtermap/2` should be compatible with `lists:filtermap/2`, just as you suggest.

RaimoNiskanen:
@Maria-12648430 
> >    @josevalim: fold and fold_r (the latter if needed). I buy that!
> Decided then? :)

I agree.  And I think fold_r can wait for its demand, since `queue:reverse(queue:fold(Fun, queue:reverse(Q)))` is not that inefficient.  Unless you just can't wait to implement it...

Maria-12648430:
> Obsolete or not, it would still be so that `queue:filter/2` with a `Fun` returning `[Item]` would produce slightly less garbage than`queue:filtermap/2` with a `Fun` returning `{true, NewItem}`. Certainly irrelevant.

I meant that my addition of `{true, NewItem}` as a possible return for the filter function would be obsoleted by `filtermap`, ie should be removed from `filter` again ;)

Maria-12648430:
> Unless you just can't wait to implement it...

I'll see ;)

essen:
More functions in `queue` is more than welcome but `fold_r` sounds unnecessary at this point. Perhaps wait to get feedback on the normal `fold` first.

RaimoNiskanen:
@Maria-12648430: `contains/1`, and `any/2` and `all/2` would be welcome additions.  Then `lists:contains/2` would be missing.

That would be a future PR...

I think there is a greater need for high level functions in `queue` because unlike lists there is no nice way in the language to traverse the elements with own code.

RaimoNiskanen:
Here are two variants more shaped for using a `PredicateFun`:

``` erlang
delete_last(Item, [X | Rest]) ->
    case delete_last(Item, Rest) of
        false ->
            case Item of
                X -> Rest;
                _ -> false
            end;
        R -> [X | R]
    end;
delete_last(_, []) -> false.

delete_first(Item, [X | Rest]) ->
    case Item of
        X -> Rest;
        _ ->
            case delete_first(Item, Rest) of
                false -> false;
                F -> [X | F]
            end
    end;
delete_first(_, []) -> false.
```

Maria-12648430:
Ok, I'll recap:

* this PR:
  * remove `{true, NewItem}` from the return values of the filter function (and accordingly, from the documentation)
  * add back the paragraph regarding garbage in the documentation
  * add a new `filtermap/2` function to mirror `lists:filtermap`
* `remove`, `delete`, (`remove_first`, `delete_first`?), `any` and `all` are material for a future PR or PRs
* `fold_r` in a future PR if needed

Agree so far?

Maria-12648430:
>   * remove `{true, NewItem}` from the return values of the filter function (and accordingly, from the documentation)
>   * add back the paragraph regarding garbage in the documentation
>   * add a new `filtermap/2` function to mirror `lists:filtermap`

Ok, all done :)

A note on `queue:filtermap/2`: I'm running `lists:filtermap/2` on the front list, and iterate a custom function mimicking `queue:filter_r/2` over the rear list, in order to traverse in queue order. The latter could also be done via `lists:foldr/3`, but it would require another function enveloping the filtermap function. Also I wasn't sure about "garbage" etc (my knowledge doesn't run _that_ deep =^^=).

lib/stdlib/doc/src/queue.xml
```diff
@@ -134,6 +136,42 @@
...
+    </func>
+
+    <func>
+      <name name="fold" arity="3" since=""/>
+      <fsummary>Fold a function over a queue.</fsummary>
+      <desc>
+        <p>Calls <c><anno>Fun</anno>(<anno>Item</anno>, <anno>AccIn</anno>)</c>
+          on successive items <c>Item</c> of <c>Queue</c>, starting
+          with <c><anno>AccIn</anno> == <anno>Acc0</anno></c>. The queue is
+          traversed in queue order, i.e. from front to rear.
```
RaimoNiskanen, line 161: We do not use such abbreviations in documentation.  Write "i.e." as "that is".

RaimoNiskanen:
@Maria-12648430: We brought this up on our Technical Board, and have a few style comments.

* The commit message on the third commit is too long so it causes a line break in GitHub's overview.
Our [guidelines](https://github.com/erlang/otp/wiki/Writing-good-commit-messages) says max 50 characters, then a blank line and after that more explanation if needed.
* I have commented inline about not using abbreviations like e.g. i.e. in our documentation.
* A personal opinion; we had some for, and some against, so do as you like:  I would prefer that since you had to write an own function for the reverse list for `filtermap/2`, write one for the front list too, so it is obvious from the code that both lists are handled the same way and to avoid an external call.
Those against says that since the code exists in `lists:filtermap/2`, it is better to use it.

Maria-12648430:
I'll do all that on monday, afk atm :)

Maria-12648430:
@RaimoNiskanen everything done :) Please review again.

> * A personal opinion; we had some for, and some against, so do as you like:  I would prefer that since you had to write an own function for the reverse list for `filtermap/2`, write one for the front list too, so it is obvious from the code that both lists are handled the same way and to avoid an external call.
>   Those against say that since the code exists in `lists:filtermap/2`, it is better to use it.

I decided to leave it as it is (front list via `lists:filtermap/2`), for the same reason as "those against" bring up ;)

Maria-12648430:
A more general question/request for opinions (especially in view of future additions to the `queue` module):

Should it be _explicitly mandatory_ for the queue module to perform _any_ traversals in queue order, at least when predicate functions are involved? It may (`fold/3`) or may not (`filter/2`, `filtermap/2`, or the future `any/2` and `all/2`) matter for the return value of those functions, but it _does_ always matter when it comes to side effects of the predicate function.

juhlig:
> Should it be _explicitly mandatory_ for the queue module to perform _any_ traversals in queue order, at least when predicate functions are involved? It may (`fold/3`) or may not (`filter/2`, `filtermap/2`, or the future `any/2` and `all/2`) matter for the return value of those functions, but it _does_ always matter when it comes to side effects of the predicate function.

My two cents: it should always be done in queue order when user functions are involved, for the possible side effects. `filter/2` does it (so does your implementation of `filtermap/2`), and I think it is just what one would expect.

RaimoNiskanen:

Maria-12648430:
> Merged. Thank you for your pull request!

You're welcome :)
I'm preparing another PR for `any/2`, `all/2`, `delete/2`, `delete_r/2` right now.

PR #2792, opened 2020-10-09
Read file rename race

mikpe:
file:read_file/1 and prim_file:read_file/1 retrieve the file size before opening the file, and then only read as much data as that file size stated.  Between the file size retrieval (stat() on Unix) and the open, the file may get renamed to a different, larger one.  This causes read_file to return truncated data.  Fixed by retrieving the file size from the open file handle instead.

Fixes https://bugs.erlang.org/browse/ERL-1376.

### DIFF
--- a/erts/emulator/nifs/common/prim_file_nif.c
+++ b/erts/emulator/nifs/common/prim_file_nif.c
@@ -1266,12 +1266,16 @@ static ERL_NIF_TERM read_file_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM a
 
     if((posix_errno = efile_marshal_path(env, argv[0], &path))) {
         return posix_error_to_tuple(env, posix_errno);
-    } else if((posix_errno = efile_read_info(&path, 1, &info))) {
-        return posix_error_to_tuple(env, posix_errno);
     } else if((posix_errno = efile_open(&path, EFILE_MODE_READ, efile_resource_type, &d))) {
         return posix_error_to_tuple(env, posix_errno);
     }
 
+    /* read_file() wants to know the file size, so retrieve it now from the
+       open file handle.  In theory, efile_read_handle_info() may fail with
+       ENOTSUP, fall back to the "unknown size" logic if that happens.  */
+    if (efile_read_handle_info(d, &info) != 0) {
+        info.size = 0;
+    }
     posix_errno = read_file(d, info.size, &result);
 
     erts_atomic32_set_acqb(&d->state, EFILE_STATE_CLOSED);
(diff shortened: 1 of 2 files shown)


erts/emulator/nifs/common/prim_file_nif.c
```diff
@@ -1266,12 +1266,16 @@ static ERL_NIF_TERM read_file_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM a
...
     } else if((posix_errno = efile_open(&path, EFILE_MODE_READ, efile_resource_type, &d))) {
         return posix_error_to_tuple(env, posix_errno);
     }
 
+    /* read_file() wants to know the file size, so retrieve it now from the
+       open file handle.  In theory, efile_read_handle_info() may fail with
+       ENOTSUP, fall back to the "unknown size" logic if that happens.  */
+    if (efile_read_handle_info(d, &info) != 0)
+        info.size = 0;
+
```
jhogberg, line 1278: Style fix:

```suggestion
    if (efile_read_handle_info(d, &info) != 0) {
        info.size = 0;
    }
```

jhogberg (commented):
LGTM, though I'd prefer the fix and test were in the same commit.

mikpe:
Applied your style fix and squashed.  I had them as separate commits to simplify running the TC w/o the fix present.

mikpe:
Backporting this to OTP-22 unfortunately fails because OTP-22 doesn't have efile_read_handle_info(), it was added in the rather large 924cd70f8b7cf1fa2256055af39723b24fd6238e.  What are your thoughts on that?  Backport all of that commit (including new API functions and test cases), just the internal refactoring and additions for efile_read_handle_info(), or create something special for OTP-22 (after all, in Unix terms all we need is an fstat() and the struct stat st_size field, though I don't know off-hand how to do that for Windows).

jhogberg:
It's a lot of code to release in a patch for something that is arguably not even a bug, so my first thought is "don't."

(You need to leave the file server for it to happen, and [the documentation](http://erlang.org/doc/man/file.html) warns that you may see nonsense when you do.)

jhogberg:
The testcase fails on Windows and older versions of OS X (Darwin 9.8.0), can you skip it on those platforms, or write it in a way that doesn't use rename?

mikpe:
> The testcase fails on Windows and older versions of OS X (Darwin 9.8.0), can you skip it on those platforms, or write it in a way that doesn't use rename?

I'll try skipping it on those platforms.  (It's _very_ odd that Unix-based Darwin would fail.)

jhogberg:
> (It's _very_ odd that Unix-based Darwin would fail.)

It runs fine on anything released the past decade though, so I'm guessing it's a bug specific to that version or older. We've had quite a few those relating to files.

mikpe:
Updated TC to skip on Windows and old Darwin.

jhogberg:

PR #2000, opened 2018-10-24
Make ct a behaviour for test suites

richcarl:
Common Test uses SUITE modules as callbacks, but there is no behaviour definition that can be used to check the callbacks. This patch makes it possible to put "-behaviour(ct)." in a SUITE module.

PR #2793, opened 2020-10-09
stdlib: Improve API and documentation of uri_string

peterdmv:
This PR intends to improve the API of the ```uri_string``` module and adds a new chapter (Uniform Resource Identifiers) to stdlib's Users' Guide.

- ```uri_string:normalize/1``` is changed to only decode those percent-encoded triplets that correspond to a character in the unreserved set.

```
      unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~" 
```
- A new function ```uri_string:percent_decode/1``` is added for decoding percent-encoded triplets in URI components.
- New utility function ```uri_string:allowed_characters()``` added for listing the allowed characters in each major URI component, and also in the most important characters sets.
```erlang
1> uri_string:allowed_characters().
[{scheme,"+-.0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"},
 {userinfo,"!$%&'()*+,-.0123456789:;=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"},
 {host,"!$&'()*+,-.0123456789:;=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"},
 {ipv4,".0123456789"},
 {ipv6,".0123456789:ABCDEFabcdef"},
 {regname,"!$%&'()*+,-.0123456789;=ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"},
 {path,"!$%&'()*+,-./0123456789:;=@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"},
 {query,"!$%&'()*+,-./0123456789:;=?@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"},
 {fragment,"!$%&'()*+,-./0123456789:;=?@ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"},
 {reserved,"!#$&'()*+,/:;=?@[]"},
 {unreserved,"-.0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz~"}]
```

PR #2794, opened 2020-10-09
Make ct a behaviour for test suites (original by richcarl)

paulo-ferraz-oliveira:
This PR is related to https://github.com/erlang/otp/pull/2000 (I used the same title and all), and continues from https://github.com/erlang/otp/pull/2000#issuecomment-706234891.

**What's missing?**
* something else (?)


paulo-ferraz-oliveira:
OK, so here's the story:
1. found https://github.com/erlang/otp/pull/2000 while searching for a Common Test suite behaviour-defining module,
2. [asked](https://github.com/erlang/otp/pull/2000#issuecomment-705804158) if I could help,
3. was [given an OK](https://github.com/erlang/otp/pull/2000#issuecomment-706035991) to help,
4. asked for [guidance](https://github.com/erlang/otp/pull/2000#issuecomment-706208037) on how to proceed on top of previous work,
5. got the required [guidance](https://github.com/erlang/otp/pull/2000#issuecomment-706234891),
6. started working on the current PR - if all goes well this will supersede https://github.com/erlang/otp/pull/2000 and it can be closed.

paulo-ferraz-oliveira:
[`common_test Module:group/1`](https://erlang.org/doc/man/common_test.html#Module:group-1) refers to `CTHPriority` but this is not specified. I'm tentatively specifying it as `term()`.

Also, the doc. itself doesn't refer to any "priority", and I've never used this element either.

**Edit**: the same goes for [`common_test Module:suite/0`](https://erlang.org/doc/man/common_test.html#Module:suite-0).

paulo-ferraz-oliveira:
I proceeded to: 
1. cherry pick from [the original](https://github.com/richcarl/otp/tree/ct-behaviour),
2. moving the existing code to lib/common_test/src/ct_suite.erl,
3. improving the `-spec(_).`s by adding missing elements and going deeper in specification,
4. formatting that code as per other elements in Erlang/OTP,
5. tweaking exported elements.

paulo-ferraz-oliveira:
What's missing:
1. knowing if this module is named correctly and placed where it should be placed,
2. knowing if there's any other type we should be exporting (and if naming is correct),
3. knowing if this change involves writing tests (or otherwise applying the behaviour to already-existing test elements - I can put it in all suites, for example),
4. knowing if this is going to `maint` or `master`,
5. understanding how I should proceed to add documentation (the missing `man` page).

lib/common_test/src/ct_suite.erl
```diff
@@ -0,0 +1,132 @@
...
+            skipped |
+            failed.
+-type ct_group_props() :: [
+                parallel |
+                sequence |
+                shuffle |
+                {shuffle, Seed :: {integer(), integer(), integer()}} |
+                {ct_group_repeat_type(), ct_test_repeat()}
+            ] |
+            default.
```
paulo-ferraz-oliveira, line 32: This option (`default`) isn't documented as valid for [`common_test Module:groups/0`](https://erlang.org/doc/man/common_test.html#Module:groups-0). If it comes to it I'll have to segregate it - along with a bunch of other types.

lib/common_test/src/ct_suite.erl
```diff
@@ -0,0 +1,132 @@
...
+-type ct_info_required() :: Key :: atom() |
+            {Key :: atom(), SubKeys :: ct_info_required_subkeys()} |
+            {Key :: atom(), Subkey :: atom()} |
+            {Key :: atom(), Subkey :: atom(), SubKeys :: ct_info_required_subkeys()}.
+-type ct_info_required_subkeys() :: SubKey :: atom() |
+            [SubKey :: atom()].
+-type ct_hooks() :: [
+                CTHModule :: atom() |
+                {CTHModule :: atom(), CTHInitArgs :: term()} |
+                {CTHModule :: atom(), CTHInitArgs :: term(), CTHPriority :: term()}
```
paulo-ferraz-oliveira, line 78: [`common_test Module:group/1`](https://erlang.org/doc/man/common_test.html#Module:group-1) refers to `CTHPriority` but this is not specified. I'm tentatively specifying it as `term()`.

Also, the doc. itself doesn't refer to any "priority", and I've never used this element either.

The same goes for [`common_test Module:suite/0`](https://erlang.org/doc/man/common_test.html#Module:suite-0).

IngelaAndin: suite/0 can be used to specify 'ct_hooks', 'timetrap', 'userdata' and 'require' configuration for a suite according to the documentation (common_test).
group/1 can override the values for suite/0 for a special group.  Example from base64_SUITE.erl in stdlib 
```

suite() ->
    [{ct_hooks,[ts_install_cth]},
     {timetrap,{minutes,4}}].

all() ->
    [base64_encode, base64_decode, base64_otp_5635,
     base64_otp_6279, big, illegal, mime_decode, mime_decode_to_string,
     {group, roundtrip}].

groups() ->
    [{roundtrip, [parallel],
      [roundtrip_1, roundtrip_2, roundtrip_3, roundtrip_4]}].

group(roundtrip) ->
    %% valgrind needs a lot of time
    [{timetrap,{minutes,10}}].

```
CTHPriority is an integer, better  specified in the ct_hooks documentation (both manual page and UG for common_test) It defines in which order hooks should be run.

lib/common_test/src/ct_suite.erl
garazdawi: It is not valid for `Module:groups`, so the `group_props` for `group_ref` is different from the `group_props` of `group_def`. Which means that they have to be segregated. I would suggest creating a type called `group_props_ref` that is `group_props() | default`, or something along those lines.

paulo-ferraz-oliveira: I'll "fix" it but it'll involve more than just (simply) a new type.

paulo-ferraz-oliveira: How does 40d021e look?

lib/common_test/src/ct_suite.erl
paulo-ferraz-oliveira: fe68aa0

paulo-ferraz-oliveira:
Could you guide me as per the questions in https://github.com/erlang/otp/pull/2794#issuecomment-706438205? Thanks.

IngelaAndin:
You can look at gen_statem and ssh_server_channel as examples of callback behaviours documentation using specs.  Some of the documentation in the common_test module (that is the current documentation of a test suite callbacks) is of course relevant but it can be improved and better structured with help of the usage of specs to generate parts of the documentation.

paulo-ferraz-oliveira:
> You can look at gen_statem and ssh_server_channel as examples of callback behaviours documentation using specs. Some of the documentation in the common_test module (that is the current documentation of a test suite callbacks) is of course relevant but it can be improved and better structured with help of the usage of specs to generate parts of the documentation.

Thank you, @IngelaAndin. I'll add a `ct_suite.xml` to `doc/src` then and start working on it. Should there be any link from the common_test man to that one?

Can I assume that:
1. module `ct_suite` is named correctly and placed where it should be?
2. there's no other type we should be exporting?
3. type naming is correct?
4. this is going to `maint`?

**About testing**
Does this change involve writing tests - or can I, otherwise, apply the newly added behaviour to already-existing test elements - all suites, for example?

paulo-ferraz-oliveira:
@IngelaAndin, do you think it would be OK to document the **Data Types** in [common_test.html](https://erlang.org/doc/man/common_test.html), like [dict.html](https://erlang.org/doc/man/dict.html ) does, for example?

The main difference is the former aren't exported by `common_test` (so we'd probably have to do that too - ?).

IngelaAndin:
Can I assume that:
1.   module ct_suite is named correctly and placed where it should be?

Yes

2.     there's no other type we should be exporting?

Our policy is to be restrictive with exporting types.  Only input and return types should be necessary to export. 

3.     type naming is correct?

I think that might still be up for discussion, but it will be simpler to discuss it when we have a built documentation proposal and not to difficult to change. 

4.     this is going to maint?

Yes we will accept it for maint.

It does not require any special tests. It might be challenging to apply it to all OTP test suites as there might still be legacy things
around in a lot of test suites dated before common_test.  You could perhaps apply it to some highly developed application such as ssl as a proof of concept.

IngelaAndin:
When it comes to the data types, they should be documented in the same fashion as in dict, this you should get almost for free when using specs to generate the docs. We usually put the spec in the relevant Erlang file an not in a a header file. So you can export the relevant types from ct_suite.erl    The common_test.xml  should not have any documentation for the callbacks this
module ought to only have information about possible application environment variables, it is originally a man(6) information and is often called  <application>_app.xml 

The xml will need something like this (a not full lexample from ssl.xml) :

```
<datatypes>
    <datatype_title>Types used in TLS/DTLS</datatype_title>

    <datatype>
      <name name="socket"/>
    </datatype>
    
    <datatype>
      <name name="sslsocket"/>
      <desc>
	<p>An opaque reference to the TLS/DTLS connection, may be used for equality matching.</p>
	</desc>
    </datatype>

    <datatype>
      <name name="tls_option"/>
    </datatype>
        
    <datatype>
      <name name="tls_client_option"/>
    </datatype>
    
    <datatype>
      <name name="tls_server_option"/>
    </datatype>
   
```

paulo-ferraz-oliveira:
> when using specs to generate the docs

How do you use "specs to generate the docs"? I'm used to `edoc` but haven't found any documentation on how to do it for Erlang/OTP base. I looked at:
* https://github.com/erlang/otp/blob/master/CONTRIBUTING.md
* https://github.com/erlang/otp/wiki/Documentation
but haven't run the appropriate commands yet.

paulo-ferraz-oliveira:
> The common_test.xml should not have any documentation for the callbacks this
module ought to only have information about possible application environment variables, it is originally a man(6) information and is often called _app.xml

I'm a bit confused here, since there's no `common_test.xml` and types are exported from `ct_suite.erl`.

I think for the doc. to be clear:
1. `common_test_app.xml` shall be _fixed_ (as per the issues I found before - missing types, for example),
2. a Data Types section must be added somewhere (I don't know if the correct place is `common_test_app.xml`),
3. there should be some indication that the behavior is implemented by module `ct_suite`.

lib/common_test/doc/src/common_test_app.xml
```diff
@@ -272,21 +272,22 @@
...
 
     <func>
-      <name>Module:group(GroupName) -> [Info] </name>
+      <name>Module:group(GroupName) -> [Info]</name>
       <fsummary>Test case group information function (providing default data
         for a test case group, that is, its test cases and
         subgroups).</fsummary>
       <type>
-	<v>Info = {timetrap,Time} | {require,Required} | {require,Name,Required} | {userdata,UserData} | {silent_connections,Conns} | {stylesheet,CSSFile} | {ct_hooks, CTHs}</v>
+	<v>GroupName = atom()</v>
```
paulo-ferraz-oliveira, line 280: Added, since "missing before".

lib/common_test/doc/src/common_test_app.xml
```diff
@@ -299,6 +300,7 @@
 	<v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;{CTHModule, CTHInitArgs, CTHPriority}]</v>
 	<v>CTHModule = atom()</v>
 	<v>CTHInitArgs = term()</v>
+	<v>CTHPriority = integer()</v>
```
paulo-ferraz-oliveira, line 303: Added, since "missing before".

IngelaAndin:
We do not use edoc.  We want you to create a manual page ct_suite.xml that includes the data types and documentation for the ct_suite behaviour.  The xml from code above refers to data types exported from ssl.erl and creates the datatype part of the manual page.

You can refer to types with seetype: (examle from gen_statem)

```

<func>
      <name since="OTP 19.0">Module:format_status(Opt, [PDict,State,Data]) ->
        Status
      </name>
      <fsummary>
	Describe the current <c>gen_statem</c> status (optional).
      </fsummary>
      <type>
        <v>Opt = normal | terminate</v>
        <v>PDict = [{Key, Value}]</v>
	<v>
	  State =
	  <seetype marker="#state">state()</seetype>
	</v>
	<v>
	  Data =
	  <seetype marker="#data">data()</seetype>
	</v>
```
Alas I do not think it is publicly documented at the moment.  

If we should remove common_test.xml  or perhaps replace it  with common_test_app.xml is an other issue. I think we can address that once we are getting ct_suite.xml  in place.

IngelaAndin:
The documentation existing of the callback API today in common_test.xml is not at all in a place that we usually would expect it to be. I think it has some legacy reasons for being so.  The origin of common_test was a very non common test framework that ended up becoming a common test framework replacing the old OTP test framework, instead of being used for an other not so common use case.   I think the relevant part should be moved to ct_suite.xml

paulo-ferraz-oliveira:
> not at all in a place that we usually would expect it to be

Hadn't understood this.

> I think the relevant part should be moved to ct_suite.xml

Sure, that's why I included that in the initial description. In any case I made the changes in the existing man page so it's easier to validate once I copy-paste them (since I will mostly copy-paste).

paulo-ferraz-oliveira:
I tried to address all documentation _issues_ in the most recent commits:
* e5c4ace: add ct_suite
* c4cf0b2: use seetype references

IngelaAndin:
You know that you need to build the documentation from the top once, and then you should be able to use the `local_docs` make target for the application in app/doc/src directory. There should also be an xmllint target.

IngelaAndin:
I just got the tip that erl_docgen  Users  Guide describes a lot of the doc features although not everything.  Keep up the good work :) and  do not hesitate to ask.

paulo-ferraz-oliveira:
> You know that you need to build the documentation from the top once, and then you should be able to use the `local_docs` make target for the application in app/doc/src directory. There should also be an xmllint target.

I've tried to build the doc, from the top, using `make release_docs`, but am still trying to understand some errors (I've rebased in the meantime, on top of `maint`, and at least the previous errors went away 😄).

paulo-ferraz-oliveira:
> I just got the tip that erl_docgen Users Guide describes a lot of the doc features although not everything. Keep up the good work :) and do not hesitate to ask.

Good to know, thanks. For future (and present) reference: http://erlang.org/doc/apps/erl_docgen/users_guide.html.

paulo-ferraz-oliveira:
In the process of `make release_docs` I'm getting:

```
=== Leaving application tftp
make[2]: *** No rule to make target `release_docs'.  Stop.
make[1]: *** [release_docs] Error 2
make: *** [release_docs] Error 2
```

(going to investigate further)

IngelaAndin:
I do:
 cd <your-otp-build-directory>
export ERL_TOP=`pwd`
./otp_build setup -a
make docs 

just keeping everything in the development environment

paulo-ferraz-oliveira:
> just keeping everything in the development environment

I see 😄. Probably we could update [the docs](https://github.com/erlang/otp/wiki/Documentation) to mention this. I'll try it. Thanks.

paulo-ferraz-oliveira:
I'm getting

```erlang
=== Leaving application tftp
make[2]: *** No rule to make target `opt'.  Stop.
make[1]: *** [opt] Error 2
make: *** [libs] Error 2
```

Trying to understand what's causing it...

paulo-ferraz-oliveira:
From [Create XML files from code](http://erlang.org/doc/apps/erl_docgen/doc-build.html#utilities-to-prepare-xml-files): I'm getting `$(ERL_TOP)`, from the example, executed. Is this OK in a given shell (interpreted as an env. var)?

paulo-ferraz-oliveira:
I was able to generate it, finally. I see some types that aren't linked, like `ct_group_props()`.

IngelaAndin:
I think you missed to include the behaviour module ct_suite.erl in the Makefile. Even if it does not have any explicit code some code will be generated for it. And that is the cause of the failing doc build.

lib/common_test/src/Makefile
```diff
@@ -40,6 +40,9 @@ RELSYSDIR = $(RELEASE_PATH)/lib/common_test-$(VSN)
 # Target Specs
 # ----------------------------------------------------
 
+BEHAVIOUR_MODULES= \
+	ct_suite
```
paulo-ferraz-oliveira, line 44: @IngelaAndin, shouldn't this be enough? I'm trying to understand which is the `Makefile` missing this file.

paulo-ferraz-oliveira:
`ct_group_props()` isn't exported (in `ct_suite.erl`) but shows up as part of the documentation (link is obviously broken). I'll look around, in other doc. files, about how we should make this work, but any help from you would be most welcome.

lib/common_test/src/Makefile
IngelaAndin: Sorry yesterday was crazy. I was just trying to decipher the error messages in the checks below. I cloned your repo and could also build the docs. I will try including it in the test and see if the build errors below goes away.

IngelaAndin:
All sub-types does not need to be exported to be used in the docs. Although input and output types to functions needs to be exported. But I think that is mainly for being able to reference them from code and no for the sake of the documentation.  I think  you  just need to add it to the "datatypelist" 
```
 <datatype>
        <name name="ct_group_props"/>
      </datatype>
```

IngelaAndin:
I get the  same error!

I think you need   

```
diff --git a/lib/common_test/src/Makefile b/lib/common_test/src/Makefile
index 7913512b1b..c6bf080063 100644
--- a/lib/common_test/src/Makefile
+++ b/lib/common_test/src/Makefile
@@ -91,7 +91,8 @@ MODULES= \
 
 
 TARGET_MODULES= $(MODULES:%=$(EBIN)/%)
-BEAM_FILES= $(MODULES:%=$(EBIN)/%.$(EMULATOR))
+BEAM_FILES= $(MODULES:%=$(EBIN)/%.$(EMULATOR)) \
+           $(BEHAVIOUR_MODULES:%=$(EBIN)/%.$(EMULATOR)) \
 
 ERL_FILES= \
        $(MODULES:=.erl) \

```

to avoid:

```
[2020-10-20T07:03:32.792Z] /usr/bin/install -c -m 644 ../ebin/ct_suite.beam  ../ebin/ct.beam ../ebin/ct_logs.beam ../ebin/ct_framework.beam ../ebin/ct_ftp.beam ../ebin/ct_ssh.beam ../ebin/ct_snmp.beam ../ebin/ct_gen_conn.beam ../ebin/ct_rpc.beam ../ebin/ct_run.beam ../ebin/ct_master.beam ../ebin/ct_telnet.beam ../ebin/ct_util.beam ../ebin/ct_cover.beam ../ebin/ct_testspec.beam ../ebin/ct_event.beam ../ebin/ct_master_event.beam ../ebin/ct_master_logs.beam ../ebin/ct_master_status.beam ../ebin/ct_repeat.beam ../ebin/ct_telnet_client.beam ../ebin/ct_make.beam ../ebin/unix_telnet.beam ../ebin/ct_config.beam ../ebin/ct_config_plain.beam ../ebin/ct_config_xml.beam ../ebin/ct_slave.beam ../ebin/ct_hooks.beam ../ebin/ct_hooks_lock.beam ../ebin/cth_log_redirect.beam ../ebin/cth_surefire.beam ../ebin/ct_netconfc.beam ../ebin/ct_conn_log_h.beam ../ebin/cth_conn_log.beam ../ebin/ct_groups.beam ../ebin/ct_property_test.beam ../ebin/ct_release_test.beam ../ebin/ct_default_gl.beam ../ebin/erl2html2.beam ../ebin/test_server_ctrl.beam ../ebin/test_server_gl.beam ../ebin/test_server_io.beam ../ebin/test_server_node.beam ../ebin/test_server.beam ../ebin/test_server_sup.beam ../ebin/common_test.app ../ebin/common_test.appup "/daily_build/otp/lib/common_test-1.19/ebin"

[2020-10-20T07:03:32.792Z] /usr/bin/install: cannot stat '../ebin/ct_suite.beam': No such file or directory

[2020-10-20T07:03:32.792Z] Makefile:160: recipe for target 'release_spec' failed

```

paulo-ferraz-oliveira:
I'm looking into this again (thanks, @IngelaAndin, will try your fix in a minute). I hope there's no conflict (since macOS in case-insensitive for filenames) between `ct_SUITE` (already exists) and `ct_suite` (the new file, being added).

lib/common_test/src/Makefile
paulo-ferraz-oliveira: 👍

paulo-ferraz-oliveira:
> I get the same error!
> 
> I think you need
> 
> ...

Thanks for the help. I hope 6fbd047 does it. It's not quite what you wrote, but I assume you assumed I hadn't created an issue before (which I might have - and which, hopefully, this commit fixes).

paulo-ferraz-oliveira:
@IngelaAndin: regarding "some types aren't linked, like `ct_group_props()` (is probably mis-use of `<datatypes>`)", could you help me? The HTML file is generated with broken links, but this is surely due to my lack of knowledge regarding the use of datatypes. The datatypes I'm using also refer to some internal ones (that don't get "generated"). Is there a way to have them "generated" inline?

e.g.
```
ct_test_def() :: ct_testname() | ct_group_ref() | ct_testcase_ref()
```
shouldn't have references to type `ct_group_ref()`, for example, since that one's internal (as in "not exported"). Is there a way to have it expanded in the doc.? Should the "generator" know about "not exported" and behave differently (this would be the subject of a different issue, of course).

paulo-ferraz-oliveira:
I solved the compilation issues, along with the test issues.
I think the documentation generation issues are solved.
I'd like validation on the solution chosen for the hyperlinks (datatypes refer to the functions that, themselves, describe the types).

paulo-ferraz-oliveira:
After this, I'd like to know if you feel there's something missing.

IngelaAndin:
The documentation  still does not  build, some links need fixing. If  you do  ` git grep "common_test#"`  in common_test/doc/src  and  `"query  replace" `  the result  with  ` ct_suite#   ` I think it should  then work and  you should then  have done the structural improvement of  making a behaviour for  the test suites and further improvements to the actual  text  should  be scheduled into our  backlog.

IngelaAndin:
We are aware that you can not currently  see which types  are exported and hence may be refereed to in other specs and which are only sub types that are for documentation purposes. There is no obvious solution at the moment,  time will tell but this is as you say an other issue than the one of this PR.

paulo-ferraz-oliveira:
> We are aware that you can not currently see which types are exported and hence may be refereed to in other specs and which are only sub types that are for documentation purposes. There is no obvious solution at the moment, time will tell but this is as you say an other issue than the one of this PR.

I kind of hacked a solution for this:
1. in the `datatypes` section I wrote e.g. `ct_info(): the return type for :suite()`, and then
2. in the `:suite()` part of the doc. I refer to the type `ct_info()` as containing other sub-types (and they're all in that context).

paulo-ferraz-oliveira:
> The documentation still does not build, some links needs fixing. If you do ` git grep "common_test#"` in common_test/doc/src and `"query replace" ` the result with `ct_suite# ` I think it should then work and you should then have done the structural improvement of making a behaviour for the test suites and further improvements to the actual text should be scheduled into our backlog.

I'll do this :) I just keep piling up issues haha. (shouldn't the build process, locally, complain about that, though?)

IngelaAndin:
I think it is the link checker that finds the problem, so it is a later step.

paulo-ferraz-oliveira:
> I think it is the link checker that finds the problem, so it is a later step.

If only there was a `make check-doc` target 😄 

Is anything missing, @IngelaAndin. I want to help further if further help is required.

lib/common_test/doc/src/ct_suite.xml
```diff
@@ -0,0 +1,619 @@
...
+      <title>Callback Functions</title>
+      <p>
+        The following functions are to be exported from a
+        <c>ct_suite</c> callback module in order to define
+        the callback interface for a test suite.
+      </p>
+    </fsdescription>
+
+    <func>
+      <name since="">Module:all() -> [ct_test_def()] | {skip, Reason}</name>
```
peterdmv, line 84: All type information can be auto-generated from the specs by following the pattern below. This is a general comment for all function definitions.
 
```xml
    <func>
      <name name="all" arity="0" since=""/>
      <fsummary> ... 1-liner summary...</fsummary>
      <desc>
	<p> ... details ...</p>
      </desc>
    </func>
```

lib/common_test/doc/src/ct_suite.xml
```diff
@@ -0,0 +1,619 @@
...
+
+	<p>Returns the list of all test cases and test case groups in the
+          test suite module to be executed. This list also specifies the
+          order the cases and groups are executed by <c>Common Test</c>.
+          A test case is represented by an atom,
+	  the name of the test case function, or a <c>testcase</c> tuple
+	  indicating that the test case shall be repeated. A test case group is
+	  represented by a <c>group</c> tuple, where <c>GroupName</c>,
+	  an atom, is the name of the group (defined in
+	  <seemfa marker="#Module:groups/0"><c>Module:groups/0</c></seemfa>).
```
peterdmv, line 111: It would be better to use  ```<seetype>``` tags instead of  ```<seemfa>``` when referring to types throughout this document. An example from #2793 :
```xml
<seetype marker="uri_string#uri_string"><c>uri_string()</c></seetype>
```

paulo-ferraz-oliveira: I can do that, but most of those tags were copied from common_test_app.xml.

paulo-ferraz-oliveira: I can't see a misuse of `seemfa`: will look for misuse of `seetype`. The place where you add the comment seems to be as expected, since `groups/0` is a function, not a type.

paulo-ferraz-oliveira: I can't see a misuse of `seetype` either.
Where did you find an issue, specifically?

lib/common_test/doc/src/ct_suite.xml
paulo-ferraz-oliveira: I don't understand your comment (what is "to follow a pattern"?). Is there something you wish me to change? Looking at the generated documentation it seems good to me.

paulo-ferraz-oliveira: You mean removing `<type>`?
I can do that, but will that work with what @IngelaAndin and I discussed before [here](https://github.com/erlang/otp/pull/2794#issuecomment-713422455)?

paulo-ferraz-oliveira: I'm now getting `Error in ct_suite:all/0: unknown spec` for
```
      <name name="all" arity="0" since=""/>
```

lib/common_test/doc/src/ct_suite.xml
peterdmv: You are right, please disregard this comment.

lib/common_test/doc/src/ct_suite.xml
peterdmv: The "pattern" I meant was to try to auto-generate the type sections using the specs in the code. I did a fast prototype to see if it works and I realized that the doc generation is missing some functionality when documenting callback functions. There are ways to work around the limitations but it is not that elegant. Try the diff below to see the difference. Ideally the types section should be generated from the specs with working links. I'll discuss this with team and I'll come back soon.

```diff
diff --git a/lib/common_test/doc/src/ct_suite.xml b/lib/common_test/doc/src/ct_suite.xml
index 0289a569f0..eb8dea7735 100644
--- a/lib/common_test/doc/src/ct_suite.xml
+++ b/lib/common_test/doc/src/ct_suite.xml
@@ -61,8 +61,8 @@
       <desc><p>The test group definition, as returned by <seemfa marker="#Module:groups/0"><c>Module:groups/0</c></seemfa>.</p></desc>
     </datatype>
     <datatype>
-      <name>ct_test_def()</name>
-      <desc><p>The test suite definition, as returned by <seemfa marker="#Module:all/0"><c>Module:all/0</c></seemfa>.</p></desc>
+      <name name="ct_test_def"/>
+      <desc><p>The test suite definition, as returned by <seemfa marker="#all/0"><c>Module:all/0</c></seemfa>.</p></desc>
     </datatype>
     <datatype>
       <name>ct_info()</name>
@@ -81,22 +81,9 @@
     </fsdescription>
 
     <func>
-      <name since="">Module:all() -> [ct_test_def()] | {skip, Reason}</name>
+      <name name="all" arity="0" since=""/>
       <fsummary>Returns the list of all test case groups and test cases
 	in the module.</fsummary>
-      <type>
-	<v><seetype marker="#ct_test_def">ct_test_def()</seetype> = TestCase | {group, GroupName} | {group, GroupName, Properties} | {group, GroupName, Properties, SubGroups}</v>
-	<v>TestCase = <seetype marker="#ct_testname">ct_testname()</seetype></v>
-	<v>GroupName = <seetype marker="#ct_groupname">ct_groupname()</seetype></v>
-	<v>Properties = [parallel | sequence | Shuffle | {RepeatType, N}] | default</v>
-	<v>SubGroups = [{GroupName, Properties} | {GroupName, Properties, SubGroups}]</v>
-	<v>Shuffle = shuffle | {shuffle, Seed}</v>
-	<v>Seed = {integer(), integer(), integer()}</v>
-	<v>RepeatType = repeat | repeat_until_all_ok | repeat_until_all_fail | repeat_until_any_ok | repeat_until_any_fail</v>
-	<v>N = integer() | forever</v>
-	<v>Reason = term()</v>
-      </type>
-
       <desc>
 	<p>MANDATORY</p>
 
diff --git a/lib/common_test/src/ct_suite.erl b/lib/common_test/src/ct_suite.erl
index a2d23e15ef..6f3c287593 100644
--- a/lib/common_test/src/ct_suite.erl
+++ b/lib/common_test/src/ct_suite.erl
@@ -12,6 +12,8 @@
               ct_info/0
              ]).
 
+-export([all/0]).
+
 -type ct_testname() :: atom().
 -type ct_groupname() :: atom().
 -type ct_config() :: [{Key :: atom(), Value :: term()}].
@@ -77,6 +79,8 @@
             ].
 -type ct_test_def() :: ct_testname() | ct_group_ref() | ct_testcase_ref().
 
+
+
 -callback all() ->
     [TestDef :: ct_test_def()] |
     {skip, Reason :: term()}.
@@ -128,3 +132,10 @@
                      init_per_testcase/2,
                      end_per_testcase/2
                     ]).
+
+-spec all() -> [TestDef] | {skip, Reason} when
+      TestDef :: ct_test_def(),
+      Reason :: term().
+
+all() ->
+    [].

```

paulo-ferraz-oliveira: I'll wait for your feedback before moving forward. Thanks.

peterdmv: We have decided to handle this in a future improvement.

paulo-ferraz-oliveira:
@peterdmv, is there something missing that I can do, still?

paulo-ferraz-oliveira (commented):
Had an open review (by mistake). Tried to post something to delete it afterward; couldn't. Sorry about the noise.

paulo-ferraz-oliveira:
Thanks for this and your guidance.

PR #2795, opened 2020-10-11
Do not try to convert already absolute path in uri_string:recompose/1

robertoaloi:
According to the documentation, the `path` passed to `uri_string:recompose/1` via the `uri_map/0` is of type `unicode:chardata()`. From what I understand, a `path` expressed as `[<<"/">>, <<"tmp">>]` should be valid input, but that would be considered a relative path and therefore an extra slash would be prepended.

### DIFF
--- a/lib/stdlib/src/uri_string.erl
+++ b/lib/stdlib/src/uri_string.erl
@@ -1780,6 +1780,8 @@ make_path_absolute(<<"/",_/binary>> = Path) ->
     Path;
 make_path_absolute([$/|_] = Path) ->
     Path;
+make_path_absolute([<<"/">>|_] = Path) ->
+    Path;
 make_path_absolute(Path) when is_binary(Path) ->
     concat(<<$/>>, Path);
 make_path_absolute(Path) when is_list(Path) ->


essen:
Don't know if code or doc is at fault here, but if it's the code then this fix is only partial, it won't work for `[<<"/t">>, <<"mp">>]`, or even for `["/", <<"tmp">>]` and any other `unicode:charlist` variant. Might be simpler to convert to list/binary before calling `make_path_absolute`.

peterdmv (changes requested):
I agree with @essen, it is simpler to convert a possibly deep list to a list or binary. That change would handle all the special cases:
```erlang
1> unicode:characters_to_binary(["/", <<"tmp">>]).
<<"/tmp">>
2> unicode:characters_to_list([<<"/t">>, <<"mp">>]).  
"/tmp"
```
As this a bug, can you please base this PR on maint and provide testcases?

robertoaloi:
@essen Indeed. If we agree the code should support `unicode:charlist/0` (and that the documentation is correct), then your suggestion makes sense. I wonder if there's already a function in OTP which converts relative to absolute.
@peterdmv Yes, that are also needed.

@peterdmv I won't be able to work on this for the next couple of days, so feel free to jump on it. Otherwise I will try to provide a proper solution as soon as I can.

peterdmv:
I'll provide a fix for this issue in #2801. Therefore, I close this PR.

robertoaloi:
Thank you for following up!

PR #2798, opened 2020-10-13
escript: Allow the user to request a particular emulator flavor

frej:
Escript allows the user to use the (undocumented) `ESCRIPT_EMULATOR`
environment variable to specify which emulator binary to use, but
there is no corresponding way to select the emulator flavor. This
patch adds support for selecting the flavor by setting the
`ESCRIPT_FLAVOR` environment variable. Setting `ESCRIPT_FLAVOR` to
`<flavor>` will add `-emu_flavor <flavor>` to the emulator command
line when running escripts.

### DIFF
--- a/erts/doc/src/escript_cmd.xml
+++ b/erts/doc/src/escript_cmd.xml
@@ -446,5 +446,11 @@ ok
       <item>Compiles the escript using flag <c>+native</c>.
       </item>
     </taglist>
+    <note>
+      <p>The configuration of the Erlang emulator invoked by
+        <c>escript</c> can be controlled using the <seecom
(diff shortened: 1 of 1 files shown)


garazdawi:
I'm not sure that we need to add this. The same thing is achieved by doing:

```
ERL_FLAGS="-emu_flavor jit" ./myscript.es
```

frej:
> I'm not sure that we need to add this. The same thing is achieved by doing:

I agree, its obvious when you tell me. Would a patch to the `escript` documentation adding a `<note>...</note>` pointing out that this is how you should send flags to the interpreter be palatable?

garazdawi:
Yes, that would be acceptable.

PR #2800, opened 2020-10-14
Since the string from strerror should never be modified, use const.

spth:
This is just a very minor issue.

erlang contains some calls to strerror where the result is assigned to a char *. However, according to the strerror specification, the
string returned by strerror should never be changed, so I think it would be better to use const char * instead.

Philipp


garazdawi:

PR #2802, opened 2020-10-14
Remove private ram_file:compress in favor of zlib:gzip

josevalim:
This is just a rebase of #2709. /cc @jhogberg


jhogberg:
This PR hangs while building `asn1`, can you test it locally?

https://github.com/erlang/otp/pull/2802/checks?check_run_id=1253505324

josevalim:
@jhogberg I ran ` ./otp_build all -a && ./otp_build tests` locally and ran `ram_file_SUITE` before submitting the PR. I just did it again and everything passed. Nothing got stuck. :(

garazdawi:
Maybe try running the dockerfile locally?

You can find it here: https://github.com/erlang/otp/blob/master/.github/dockerfiles/Dockerfile.64-bit

The source file should be packed using https://github.com/erlang/otp/blob/master/.github/scripts/init-pre-release.sh

garazdawi:
> Changed Dockerfile.64-bit to use FROM ubuntu-base (since I don't have access to the one in the github org)

hmm, well that was too bad. I thought that it would work for you without having logged in, [but apparently not](https://github.community/t/docker-pull-from-public-github-package-registry-fail-with-no-basic-auth-credentials-error/16358/1).

> Now to figure out what went wrong...

Happy debugging!

josevalim:
@garazdawi thanks!

So the build was getting stuck because beam_lib would always call uncompress, even if the file was already in the proper .beam format. The previous implementation would normalize iolist to binaries in the `catch` clause, the current implementation didn't, so we were passing an iolist to the internals of beam_lib, which would fail. However, since the asn1rtt generation happens early on with a limited version of the runtime, no error was spewed and it just got stuck. :D

I have fixed it by doing two changes:

1. Making sure we normalize iolists to binaries in the catch clause.

2. Avoid calling gunzip in the first place if the file is already in the proper format (we did this check for when binaries were given to beam_lib but now I have generalized it).

I have force pushed to see how the build will behave. I am also re-running things from scratch locally. :)

josevalim:
The build is stuck because we need to update the primary bootstrap. I was told in the past I should not do that in a PR (as external users should not commit compiled code), so I have done so in a separate commit. If you prefer me to squash it, please let me know.

jhogberg:
Thanks, I'll add it to our tests with an updated bootstrap. Can you reset the PR to e95f648dd49d8699225ba46cea81ab6b831525bc so I can merge it once they pass?

josevalim:
@jhogberg done!

jhogberg:

michaelklishin:
FYI, this breaks Rebar3 bootstrap https://github.com/erlang/rebar3/issues/2402.

josevalim:
Right, there is a fix upstream already. The issue is that hex_core has its own copy of erl_tar for some bug fixes and that included calling these (and possibly others) private APIs. Ideally they want to push those fixes upstream.

My suggestion would be for the Rebar and Hex teams to release updated versions asap so in the next months people upgrade their local installs reducing the odds of running into issues by the time Erlang/OTP 24 comes out. We can discuss this in the next Build and Packaging WG meeting. /cc @t_sloughter

tsloughter:
Yup, hex already had a release I think, so we just need to update the vendoring and make a release. Hopefully before the next WG meeting.

PR #2808, opened 2020-10-18
Using WX_APP_TITLE for wxEntry()

dominicletz:
For additional windows in GTK this replaces the default "Erlang" displayed title when the environment variable WX_APP_TITLE is specified. In the current version this happens already on macOS but is not effective on GTK. This change adds the same effect for GTK by changing the initial parameter of wxEntry() based on the environment variable.

@dgud fyi wxWidgets maintenance fix

### DIFF
--- a/lib/wx/c_src/wxe_main.cpp
+++ b/lib/wx/c_src/wxe_main.cpp
@@ -118,7 +118,20 @@ void *wxe_main_loop(void *vpdl)
 {
   int result;
   int  argc = 1;
-  const wxChar temp[10] = L"Erlang";
+  wxChar temp[128] = L"Erlang";
+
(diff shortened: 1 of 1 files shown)


dominicletz:
A note from implementing this, I've adapted the same process used in wxe_ps_init.c for macos but I think that there is be a bug in that macos code for very large environment variables and have created another pull request specific to that #2810

dgud:
Maint patches need to be based on maint :-)

dominicletz:
😳 of course, just rebased, but it seems i can't change the target branch from my side.

dgud:
You can (or at least I can) change the base at github via 'edit' button on the right of the PR header.

PR #2814, opened 2020-10-19
Handle uncompressed binaries given as compressed to erl_tar

josevalim:
While one could argue that given an uncompressed binary with the compressed flag should fail, earlier OTP versions implicitly supported this behaviour, so this commit adds it back.

This fixes a regression from #2802. /cc @jhogberg 

Thanks @wojtekmach for the report.


jhogberg:
Merged, thanks for the PR!

PR #2818, opened 2020-10-21
Do not centralize type/cname lookups through inet_db

josevalim:
Before this patch, inet_res:getbyname by name lookups
were centralized through the inet_db process, which
made it a bottleneck on concurrent machines that need
to establish many connections to several hosts quickly.

Given these lookups are concurrent, the update of
resolve.conf has been changed to use atomic operations
that replace the list of servers and list of names.

The inet_db API has also been cleaned up to no longer
expose operations that are not used anywhere within OTP,
especially to avoid del+ins/add operations which
wouldn't be safe concurrently.

### DIFF
--- a/lib/kernel/src/inet_db.erl
+++ b/lib/kernel/src/inet_db.erl
@@ -33,12 +33,10 @@
 
 %% External exports
 -export([start/0, start_link/0, stop/0, reset/0, clear_cache/0]).
--export([add_rr/1,add_rr/5,del_rr/4]).
--export([add_ns/1,add_ns/2, ins_ns/1, ins_ns/2,
-	 del_ns/2, del_ns/1, del_ns/0]).
--export([add_alt_ns/1,add_alt_ns/2, ins_alt_ns/1, ins_alt_ns/2, 
-	 del_alt_ns/2, del_alt_ns/1, del_alt_ns/0]).
--export([add_search/1,ins_search/1,del_search/1, del_search/0]).
+-export([add_rr/1, add_rr/5, del_rr/4]).
+-export([add_ns/1, add_ns/2]).
+-export([ins_alt_ns/2, del_alt_ns/2]).
+-export([add_search/1, ins_search/1]).
 -export([set_lookup/1, set_recurse/1]).
 -export([set_socks_server/1, set_socks_port/1, add_socks_methods/1,
 	 del_socks_methods/1, del_socks_methods/0,
@@ -154,41 +152,12 @@ add_ns(IP,Port) ->
     call({listop, nameservers, add, {IP,Port}}).
 
 %% insert at head of name server list
-ins_ns(IP) ->
-    ins_ns(IP, ?NAMESERVER_PORT).
-ins_ns(IP,Port) ->
-    call({listop, nameservers, ins, {IP,Port}}).
-
-%% delete this name server entry (delete all ns having this ip)
-del_ns(IP) -> 
-    del_ns(IP, ?NAMESERVER_PORT).
-del_ns(IP, Port) ->
-    call({listop, nameservers, del, {IP,Port}}).
-
-del_ns() -> 
-    call({listdel, nameservers}).
-
-%% ALTERNATIVE NAME SERVER
-%% add to the end of name server list
-add_alt_ns(IP) -> 
-    add_alt_ns(IP, ?NAMESERVER_PORT).
-add_alt_ns(IP,Port) ->
-    call({listop, alt_nameservers, add, {IP,Port}}).
-
-%% insert at head of name server list
-ins_alt_ns(IP) -> 
-    ins_alt_ns(IP, ?NAMESERVER_PORT).
 ins_alt_ns(IP,Port) ->
     call({listop, alt_nameservers, ins, {IP,Port}}).
 
 %% delete this name server entry
-del_alt_ns(IP) ->
-    del_alt_ns(IP, ?NAMESERVER_PORT).
 del_alt_ns(IP, Port) ->
     call({listop, alt_nameservers, del, {IP,Port}}).
-
-del_alt_ns() -> 
-    call({listdel, alt_nameservers}).
 
 %% add this domain to the search list
 add_search(Domain) when is_list(Domain) -> 
@@ -196,12 +165,6 @@ add_search(Domain) when is_list(Domain) ->
 
 ins_search(Domain) when is_list(Domain) ->
     call({listop, search, ins, Domain}).
-
-del_search(Domain) ->
-    call({listop, search, del, Domain}).
-
-del_search() ->
-    call({listdel, search}).
 
 %% set host name used by inet
 %% Should only be used by inet_config at startup!
@@ -638,7 +601,12 @@ lookup_cname(Domain) ->
 %% Have to do all lookups (changes to the db) in the
 %% process in order to make it possible to refresh the cache.
 lookup_rr(Domain, Class, Type) ->
-    call({lookup_rr, Domain, Class, Type}).
+    match_rr(#dns_rr{domain = tolower(Domain), class = Class,type = Type,
+		     cnt = '_', tm = '_', ttl = '_',
+		     bm = '_', func = '_', data = '_'}).
+
+match_rr(RR) ->
+    filter_rr(ets:match_object(inet_cache, RR), times()).
 
 %%
 %% hostent_by_domain (newly resolved version)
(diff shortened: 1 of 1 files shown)


josevalim:
@sneako has run this in production and he reported it performs better in his app which makes very intensive use of `inet_res`. However, he did find a bug too. [`ets:next` on cache refresh timeout can fail](https://github.com/erlang/otp/blob/e5a7c5f59f731ce6e52c124487c40d75349c2f4e/lib/kernel/src/inet_db.erl#L1686) if `filter_rr` happens to remove an entry at the same time the cache is being cleaned up.

I can think of two options to fix this issue:

1. Have filter_rr collect expired entries and send them to the server for deletion
2. Use safe_fixtable around ets:next

I am more inclined to go with option 1 but I would love to hear your preferences. Thank you.

RaimoNiskanen:
I'd say that lookup_rr / match_rr / filter_rr not using safe_fixtable feels scary because a lookup updates table records (not only delete, but also insert).  Why not move up the filter_rr code to right under match_rr, remove the comment above lookup_rr about doing lookups in the server process, and wrap the lookup in safe_fixtable true/false calls (your option 2)

If the performance is disappointing we might look at your option 1.  A call to match_object does a safe_fixtable internally, so I do not think that extending that safe_fixtable to a subsequent match_delete and insert will be that bad.

Other than that I have a few concerns about the existing code.  Why does it use match_delete instead of delete_object?  Why does it use a type bag table with Domain as key instead of {Domain, Class, Type} as key in a table of type set?  I think I will try a rewrite - there seems to be room for optimization...

RaimoNiskanen:
Ok, that rewrite went down like a lead balloon ;-)  The `#dns_rr` record is used in many places, so keeping a `bag` is the simple way to do it.

But I do not see why the calls to `ets:match_delete(inet_cache, RR)` in `inet_db:filter_rr/2` could not be changed into `ets:delete_object(inet_cache, RR)`.

Maybe we should add a clause to avoid unnecessary updates on heavily loaded systems:
``` Erlang
filter_rr([RR | RRs], Time) when RR#dns_rr.cnt =:= Time ->
    [RR | filter_rr(RRs, Time)];
filter_rr([RR | RRs], Time) ->
    ets:delete_object(inet_cache, RR),
    ets:insert(inet_cache, RR#dns_rr { cnt = Time }),
    [RR | filter_rr(RRs, Time)];
```

josevalim:
@RaimoNiskanen good call on `match_delete` vs `delete_object`. It seems that `delete_object` can be used in other places as well, such as `do_refresh_cache` and `delete_older`.

Btw, should we do the `safe_fixtable` on `match_rr` (the client) or on `refresh_cache`+`add_rr` (the server)? Does it make a semantic difference which one we pick?

RoadRunnr:
> Other than that I have a few concerns about the existing code. Why does it use match_delete instead of delete_object? Why does it use a type bag table with Domain as key instead of {Domain, Class, Type} as key in a table of type set? I think I will try a rewrite - there seems to be room for optimization...

Since you mentioned "optimization", there is also room for feature completion. The various DNS APIs (that are supported by the functions in inet_db) are lacking in features. Modelling them after old POSIX API doesn't help (although this PR might not be the place to discuss that).

* the higher level DNS APIs in inet are restricted to `IN {A,AAA}` lookups, TXT, NAPTR, SRV and others would be quite useful. Especially using the caching features in the inet APIs would be very helpful for those records.
* there is no API that is makes it simple to do A and AAA lookups in parallel
* using inet_res directly is sometimes a good option, but this usually leads to the desire to cache the found RRs (and negative cache failures). Being able to use the existing functionality for inet_db is often desirable for this.

RaimoNiskanen:
@RoadRunnr: I agree on all points (except `AAA` -> `AAAA`)

RaimoNiskanen:
@josevalim: I have done some more tinking, reading and conferring with the ETS guy(s).

In `filter_rr/2` an `ets:safe_fixtable/2` would not do any good.  The traversal is done with `ets:match_object/2` which is already as safe as it gets.  Then some deletes and inserts are done which `ets:safe_fixtable/2` does not affect.  Maybe the calls to `ets:delete_object/2` and `ets:insert/2` with a new `cnt` field should be swapped so there temporarily is two objects in the cache and not none.

For `do_refresh_cache/1,4`, an `ets:safe_fixtable/2` is needed since otherwise `ets:next/2` might fail or miss objects.  The same goes for `delete_older/5`.

For `cache_rr/3`, an `ets:safe_fixtable/2` will not matter.  Also, swapping the order between `ets:match_delete/2` and `ets:insert/2` will not work so there will be a small time delta with no cache entry during an update.  The only way to remedy that would be to insert first, then match_objects, iterate the list and delete all but the inserted object.  Much ado about (almost) nothing.

It seems to me that `do_refresh_cache/1,4` and `delete_older/5` iterates strangely (using some `ets:next/1` / `ets:lookup/2` weirdness, I hope it is weirdness, and not something fancy I have not understood yet).  Since the cache is rather small (100 entries by default) it could work with an `ets:tab2list/1`, iterate the list and delete the oldies while finding the new OldestT, but since someone might increase the cache size arbitrarily it would be safer to instead iterate with `ets:match_object(Tab, '_', ?CACHE_LIMIT/2)` / `ets:match_object(Cont)`.

josevalim:
> Maybe the calls to ets:delete_object/2 and ets:insert/2 with a new cnt field should be swapped so there temporarily is two objects in the cache and not none.

Ah, there is a concern about concurrent lookups too. Although I think it is fine if one overrides the other since the concurrent accesses are likely to be milliseconds from each other and that precision is not relevant here. It is at least something we should document.

At the same time, I wonder why we need to remove entries when we are looking up. We already expire cache entries every one hour and, if the cache is full, `add_rr` will also purge entries. Maybe a saner approach would be to skip the deletes on lookup and make sure that the current delete+insert becomes an update? Although I am not sure how to express it as an update given the table is a bag.

RaimoNiskanen:
@josevalim: I can also not really understand why cache entries are expired during lookup.  The comment for `ttl =:= 0 -> %% at least once` might be a clue, since that would be impossible to accomplish without deletion during lookup. On the other hand that special clause feels like an incomplete kludge because if you insert an entry with `ttl = 1` and then do not look it up until after one millisecond, that entry will never be used, while one with `ttl = 0` will be used exactly once despite its age.  Ugly.

So I think it sounds good to stop expiring entries during lookup, and simply ignore old ones.  As you said, entries are expired before insertion and once every hour.  That should be enough.

I was sketching on an update in a bag, the sketch was approximately:
``` Erlang
ets:insert(Tab, RR),
[ets:delete_object(Tab, Match)
    ||  Match <- ets:match(Tab, RR#dns_rr{...wildcards...}),
        Match =/= RR]`
```
But I do not think it is as thread safe as one might want.  Two processes inserting in parallel might delete each others insertion ending up in no inserted entry.  The current match_delete + insert approach might end up in one entry per parallel insert and has a small gap where there is no entry, where some other process might also decide to insert one, which only causes an unnecessary full lookup and insertion but probably not another duplicate entry.

It is a cache, so best effort should be good enough.  I see no really bad corner cases for missing or duplicate entries.

josevalim:
@RaimoNiskanen I am thinking about going back to your earlier suggestion and try to make it a set. We can either store it as a tuple `{{Domain,Class,Type}, #dns_rr}` or add a `cache_key` entry to `#dns_rr`. It duplicates a bit of the data but we at least have easy atomic updates...

RaimoNiskanen:
I ditched the idea when I realized I had to duplicate data.  But that is no good reason.
I'd go with {{Domain,Class,Type} = Key,#dns_rr{}} since changing #dns_rr{} might cause problems somewhere.

josevalim:
@RaimoNiskanen I have sent PR #2891 that moves it to a set. I will close this one for now and resubmit it once #2891 is merged.

lessless:
Hi @josevalim,

we were also bitten by this issue recently. 

Would you like to resubmit this PR again since #2891 is already merged?

josevalim:
Yup, it is on my todo!

josevalim:
@lessless I have submit a new PR here: #3041. If you can try it out and report back, it would be very helpful.

lessless:
@josevalim thanks for the heads up! we ended up copy-pasting & cleaning up  inet_res implementation. It was fast enough for all our purposes and there is no chance for me to check this updated version ATM because I moved to the next area 🙇

PR #2820, opened 2020-10-22
zlib: Hard deprecate legacy functions

wojtekmach:
I found this comment when browsing the source code.


jhogberg:
Thanks for the PR!

Can you change the declarations to [fit our new format](https://github.com/erlang/otp/pull/2824)? You can copy over the description from the documentation, and mark them for removal in OTP 27.

wojtekmach:
@jhogberg updated, sorry for the delay!

Since I updated `zlib.erl`, I needed to run `./otp_build update_preloaded` but I assume I wasn't supposed to commit the updated .beam files.

I also updated `zip` since it wasn't using the deprecated functions.

lib/stdlib/src/zip.erl
```diff
@@ -1534,45 +1534,46 @@ get_file_name_extra(FileNameLen, ExtraLen, B, GPFlag) ->
...
+copy_data_loop(CompSize, In0, Out0, Input, Output, CRC0) ->
     N = erlang:min(?READ_BLOCK_SIZE, CompSize),
     case Input({read, N}, In0) of
-	{eof, In1} -> {Out0, In1};
-	{Uncompressed, In1} ->
-	    CRC1 = zlib:crc32(Z, CRC0, Uncompressed),
-	    Out1 = Output({write, Uncompressed}, Out0),
-	    copy_data_loop(CompSize-N, In1, Out1, Input, Output, CRC1, Z)
+        {eof, In1} ->
+            {Out0, In1};
```
wojtekmach, line 1557: I wasn't sure if this should return a 3 element tuple to match the other function head. Removing this clause didn't make the zip_SUITE tests fail, I wasn't sure how to trigger this code path either. Any assistance with this would be appreciated.

jhogberg: When extracting it will only happen when a file ends before the size given in its header, but that should be caught earlier on when checking the central directory. I think the most reasonable thing to do is to ignore it and crash.

When compressing (`put_z_data_loop`) it will happen if one of the files shrink before we're done reading them, but it seems rather benign there.

wojtekmach: Thank you, I’ll be happy to send another PR that raises when extracting.

wojtekmach:
The CI failure:

```
2020-11-17T11:42:51.8136541Z Testing tests.system_test.otp_SUITE.cases: Starting test, 5 test cases
2020-11-17T11:42:58.5827343Z 
2020-11-17T11:42:58.5858882Z - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
2020-11-17T11:42:58.5869806Z test_server:ts_tc failed on line 1754
2020-11-17T11:42:58.5870336Z Reason: {test_case_failed,{11,obsolete_but_not_deprecated}}
2020-11-17T11:42:58.5871042Z - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
2020-11-17T11:42:58.5871226Z 
2020-11-17T11:42:58.6048132Z Testing tests.system_test.otp_SUITE.cases: *** FAILED test case 3 of 5 ***
2020-11-17T11:42:59.1122357Z Testing tests.system_test.otp_SUITE.cases: TEST COMPLETE, 4 ok, 1 failed of 5 test cases
```

seems possibly related to me not committing the updated `zlib.beam`. Anything I should do next?

jhogberg:
> seems possibly related to me not committing the updated zlib.beam. Anything I should do next?

Yeah, there's not much that can be done about the CI failure on GitHub, but it ran fine internally so I've merged it as d3256b64e7dbe8b628f7779c9fe170bd302866ac. Thanks again for the PR!

PR #2822, opened 2020-10-23
Clarify how add_sup_handler establishes links

aronisstav:
Make the following comment visible in the docs: 
https://github.com/erlang/otp/blame/maint/lib/stdlib/src/gen_event.erl#L525

### DIFF
--- a/lib/stdlib/doc/src/gen_event.xml
+++ b/lib/stdlib/doc/src/gen_event.xml
@@ -209,6 +209,10 @@ gen_event:stop             ----->  Module:terminate/2
           <seemfa marker="#add_handler/3"><c>add_handler/3</c></seemfa>,
           but also supervises the connection between the event handler
           and the calling process.</p>
+	<p>This call will also set up a link between the event handler and
+	  the calling process, which will <em>not</em> be removed if the
+	  handler is removed, in case another supervised handler also links
+	  to the same process.</p>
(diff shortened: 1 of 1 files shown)


IngelaAndin:
This code is very old, written before monitors existed. We really would not like to document this flawed handling of links. The tm correct way of implementing it before monitors would be to have a reference count on the links. The best solution would probably be to rewrite it to use monitors, however we do not know if this would break old code. Hopefully not, but it needs to be investigated. That would be something for master.

aronisstav:
Nice! I went for the least possible amount of work to raise the issue really, and I'd be happy to help with a different fix (reference counting links, or switching to monitors), if you need such help.

I discovered this by writing a test that subscribes to a gen_event process with a supervised connection, unsubscribes, and then tells gen_event to shutdown, and noticed that my test process was being taken down as well. :-)

IngelaAndin:
I am thinking that any old code that discovered this behavior might do an unlik later, and if so that should be ok as unlik will not return an error if there is no link!  It would however be safer to change such a thing in master and mark it as potentially incompatible. You are more than welcome to help :) Reference counting now would perhaps no longer guarantee the correct behavior as old code might have already broke it by unlinking without being the last reference ...

IngelaAndin:
So we will close this one, looking forward to OTP-24 PR that uses monitors.

PR #2823, opened 2020-10-23
Implement NSS-style keylog in ssl:connection_information/2

dotsimon:
This commit provides the ability to retreive keylog information that
can be used by tools (e.g. Wireshark) to decrypt TLS connections.

The information returned by the keylog connection_information item is
a list of keylog items that can be written to a file or otherwise
communicated to a tool understanding keylog.

For example:
```
{ok, [{keylog, KeylogItems}]} = ssl:connection_information(S, [keylog]).
file:write_file("/tmp/key.log", [[KeylogItem,$\n] || KeylogItem <-KeylogItems]).
```

For TLSv1.3 connections more information must be stored in the connection state. In this case the initial `ssl:listen` or `ssl:connect` call must include the option `{keylog, true}`

For TLSv1.2 connections, the required information could be obtained from the existing `master_secret` and `client_random` items and the keylog item constructed manually. However this PR provides a common method regardless of the protocol.

References:
[NSS](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Key_Log_Format)
[Wireshark](https://wiki.wireshark.org/TLS#Using_the_.28Pre.29-Master-Secret)

### DIFF
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -788,6 +788,16 @@ fun(srp, Username :: binary(), UserState :: term()) ->
       </desc>
     </datatype>
 
+    <datatype>
+      <name name="keep_secrets"/>
+      <desc><p>Configures a TLS 1.3 connection for keylogging</p>
+      <p>In order to retrieve keylog information on a TLS 1.3 connection, it must be configured
+      in advance to keep the client_random and various handshake secrets.</p>
+      <p>The keep_secrets functionality is disabled (<c>false</c>) by default.</p>
+      <p>Added in OTP 23.2</p>
+      </desc>
+    </datatype>
+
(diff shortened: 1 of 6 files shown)


lib/ssl/doc/src/ssl.xml
```diff
@@ -788,6 +788,15 @@ fun(srp, Username :: binary(), UserState :: term()) ->
       </desc>
     </datatype>
 
+    <datatype>
+      <name name="keep_secrets" since="OTP 23.1"/>
```
peterdmv, line 792: ```since``` tags only work with functions. It would be better to document it in the description: "Since OTP 23.2."

peterdmv (changes requested):
Looks good, I have just a few comments to the documentation.

dotsimon:
> Looks good, I have just a few comments to the documentation.

Should be all fixed now. Thanks for the review.

peterdmv (approved):

peterdmv:

PR #2825, opened 2020-10-26
filter makedep_side_effect

tonyrog:
Avoid dialyzer (and other) crash when using -MMD -MF dir/file.d options to generate beam file. Otherwise makedep_side_effect is passed to compiler that tries to create dir/file.d out-of context.


bjorng:
Thanks for your pull request.

PR #2827, opened 2020-10-27
[inet_dns] Support URI and CAA resource records

ruudk:
This adds support for querying [URI](https://tools.ietf.org/html/rfc7553) and [CAA records](https://tools.ietf.org/html/rfc6844#section-3).

Personally I only needed CAA records but I added URI to make the list increment nicer :) 

I assume I need to add some testcases but I wish to get some advice with that. But before I do I first would like to know if this is something worth adding.


jimdigriz:
CAA is not a TXT encoded field, it has a [tag value](https://tools.ietf.org/html/rfc6844#section-5.1) in there?

My CAA encoder I had to use the following before passing into `inet_dns:encode(inet_dns:make_msg([..., {anlist,ANList}]))`:
```
ANList = lists:map(fun(AN) ->
        inet_dns:make_rr(case lists:keyfind(type, 1, AN) of
                {type,caa} ->
                        {data,{F,T0,V0}} = lists:keyfind(data, 1, AN),
                        T = list_to_binary(T0),
                        TL = byte_size(T),
                        V = list_to_binary(V0),
                        Data = <<F:8/integer, TL:8/integer, T/binary, V/binary>>,
                        lists:keystore(type, 1, lists:keystore(data, 1, AN, {data,Data}), {type,?T_CAA});
                _Else ->
                        AN
        end)
end, ANList0),
```

jimdigriz:
Ignore https://github.com/erlang/otp/pull/2827#issuecomment-721837629, probably not relevant as your work is for queries, mine is for encoding responses.

ruudk:
@jimdigriz Actually, you're right. This is what I had to do to decode it properly in Elixir:
```elixir
defp transform_caa(value, 257) do
    <<head::binary-size(2), tail::binary>> = value

    [bit, length] = head |> :binary.bin_to_list()

    <<tag::binary-size(length), value::binary>> = tail

    ~s(#{bit} #{tag} "#{value}")
  end
```

I think it will be hard for me to continue with this PR as I don't have any experience with Erlang. But basically I need to convert my Elixir code to Erlang, and use that to convert the CAA record to text.

RaimoNiskanen:
I added 3 commits.  Try and see if they work for you.  I augmented the test case for decoding.

What I am not certain about is the character encoding.  The decoding for NAPTR records uses utf8 character decoding, but I think that is not correct.  This suggested code decodes as a byte list a.k.a latin1 since RFC 6844 says that the Value field should be regarded as binary data.

I did not (yet) decode the Tags field into an atom, but instead as a character list.  If it should be decoded as an atom we should use `binary_to_existing_atom(Bin, latin1)`, and have to do a case over the known values `issue | issuewild | iodef` to make the atoms existing.  Would that be a better approach, and if so should future unknown values be decoded as byte lists or render a `?DECODE_ERROR`?

RaimoNiskanen:
I added a correction commit.  Awaiting comments...

jimdigriz:
I would leave the tag as a list, meaning there are no edge cases for the programmer to deal with.

The programmer is going to have their own view on 'valid' and having to deal with atoms and a possible fallback (which may change over future OTP versions) is just going to lead to gnashing of teeth.

I'll see about trying out the encoder part this at the weekend for you with my project.

RaimoNiskanen:
@jimdigriz Valid point about tag as list.

Looking forward to results about the encoder :-)

jimdigriz:
I checked out your [ruudk:uri-and-caa-rr branch](https://github.com/ruudk/otp/tree/uri-and-caa-rr), compiled, removed my own CAA support from my codebase and it all just worked straight off the bat.

SHIP IT!

Note that my codebase only uses the DNS *encoder* so the only other testing (`dig CAA @127.0.0.1 -p 5553 wibble.azure.coremem.com`) I was able to do was `io:format`ing the output from `{ok,D} = inet_dns:decode(R), inet_dns:msg(D).`

```
[{header,{dns_header,61814,false,query,false,false,true,false,false,0}},
 {qdlist,[{dns_query,"wibble.azure.coremem.com",caa,in}]},
 {anlist,[]},
 {nslist,[]},
 {arlist,[{dns_rr_opt,".",opt,4096,0,0,0,
                      <<0,10,0,8,224,37,204,122,191,208,50,124>>}]}]
```

Sorry, do not have the environment for testing the decoder of a DNS response with CAA data in there.

Thanks.


```
alex@aineko:~/src/dns-wingman$ make
rebar3 auto
===> Verifying dependencies...
===> Analyzing applications...
===> Compiling acme
===> Compiling dnswingman
Setting up watches.  Beware: since -r was given, this may take a while!
Setting up watches.  Beware: since -r was given, this may take a while!
Watches established.
Watches established.
Erlang/OTP 24 [DEVELOPMENT] [erts-11.1.1] [source-f8bfc9ebe0] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V11.1.1  (abort with ^G)
1> 2020-11-16T12:50:46.581100+00:00 warning: dns listening on 5553/{udp,tcp}
2020-11-16T12:50:46.602109+00:00 warning: httpd listening on 4443/tcp (rfc8738 only works on 443/tcp)
===> Booted ranch
===> Booted acme
===> Booted dnswingman
===> Booted sasl
2020-11-16T12:50:47.196418+00:00 notice: pull azure.coremem.com success
2020-11-16T12:50:47.196721+00:00 notice: notify azure.coremem.com

1> 2020-11-16T12:50:52.197009+00:00 notice: Query from ::ffff:127.0.0.1 for azure.coremem.com (aa:false, rd:false, type:axfr)
```

In my other console:
```
alex@aineko:~/src/otp$ dig AXFR @127.0.0.1 -p 5553 azure.coremem.com

; <<>> DiG 9.11.5-P4-5.1+deb10u2-Debian <<>> AXFR @127.0.0.1 -p 5553 azure.coremem.com
; (1 server found)
;; global options: +cmd
azure.coremem.com.      3600    IN      SOA     azureprivatedns.net. azureprivatedns-host.microsoft.com. 1605531047 10 300 2419200 10
cheese.azure.coremem.com. 3600  IN      MX      20 example.invalid.
wibble.azure.coremem.com. 3600  IN      CAA     12 cheese "moocow"
steak.azure.coremem.com. 3600   IN      AAAA    2a02:390:735e::
test.azure.coremem.com. 3600    IN      A       9.8.6.6
test.azure.coremem.com. 3600    IN      A       3.4.5.1
test.azure.coremem.com. 3600    IN      A       1.1.1.1
azure.coremem.com.      172800  IN      NS      ns1-08.azure-dns.com.
azure.coremem.com.      172800  IN      NS      ns2-08.azure-dns.net.
azure.coremem.com.      172800  IN      NS      ns3-08.azure-dns.org.
azure.coremem.com.      172800  IN      NS      ns4-08.azure-dns.info.
azure.coremem.com.      3600    IN      SOA     azureprivatedns.net. azureprivatedns-host.microsoft.com. 1605531047 10 300 2419200 10
;; Query time: 18 msec
;; SERVER: 127.0.0.1#5553(127.0.0.1)
;; WHEN: Mon Nov 16 12:50:52 GMT 2020
;; XFR size: 12 records (messages 1, bytes 440)
```

RaimoNiskanen:
@jimdigriz:
> I checked out your ruudk:uri-and-caa-rr branch, compiled, removed my own CAA support from my codebase and it all just worked straight off the bat.
> 
> SHIP IT!

Good to know!

> Sorry, do not have the environment for testing the decoder of a DNS response with CAA data in there.

No worries.  The test case I added tests against BIND as a server.  Passes on my Ubuntu 18.04 laptop. (it takes some messing with the system to make it run, thanks to Aparmor)

I still need to fix the problem that in our lab we have a slightly older BIND version (at least 9.9.5) that does not recognize CAA records, so I have to change the test case to work for such BIND versions too, at least to not produce an error...  All other tests work from BIND 8.

jimdigriz:
> I still need to fix the problem that in our lab we have a slightly older BIND version (at least 9.9.5) that does not recognize CAA records, so I have to change the test case to work for such BIND versions too, at least to not produce an error... All other tests work from BIND 8.

You have my deepest sympathies, and respect, I would have just had a bonfire, salted the earth and moved to use `[djbdns](https://cr.yp.to/djbdns.html)` as a testing target ;)

RaimoNiskanen:
@jimdigriz: Intriguing idea.  It seems tinydns or djbdns is hard to find as a package for Ubuntu, so maybe move to nsd that is preferred in OpenBSD, can run chrooted, and can be package installed on Ubuntu and thereby on enough machines.  I will have a look at it.  Although I just found the "RFC 3597 syntax" that should be applicable to any BIND version...

jimdigriz:
> ...so maybe move to nsd that is preferred in OpenBSD

I can confirm that NSD is less awful than BIND.

I am not convinced even ISC likes BIND much and probably is destined to go the same way of their DHCP server, but this is getting off topic and pushing me back towards my grumpy sysadmin ways.

Let me know if there is anything I can do to help.

RaimoNiskanen:
I have added a commit in our nightly builds that changes the test counterpart server to _nsd_ instead.  It should run at least one night without errors before I can commit this PR.

RaimoNiskanen:
Finally I got the test cases to work on all our lab machines.  nsd versions range from 4.0.? to 4.3.?
It has become easier to test now.  Just install nsd and run inet_res_SUITE (depends unfortunately on inet_SUITE, so it looks better to run that first).  No hassling with Apparmor.

The test case change is in one commit on top of this pull request.  I did not push it here because it would have brought in merge conflicts with the master branch.

ruudk:
Thanks for taking care of this 🙌

PR #2828, opened 2020-10-27
Clarify the role of underscores

bjorng:
Clarify the difference between the anonymous variable and variables
starting with underscore.

https://bugs.erlang.org/browse/ERL-1113

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -91,9 +91,19 @@ _Height</pre>
       <p><em>Example:</em></p>
     <pre>
 [H|_] = [1,2,3]</pre>
+    <p>The anonymous variable can also be used to suppress warnings for
+    expressions whose values are ignored.</p>
+    <p><em>Example:</em></p>
+    <pre>
+_ = element(1, Tuple)</pre>
+    <p>By assigning the result of the call to <c>element/2</c> to <c>_</c>,
+    the compiler will not emit a warning for the value being ignored.
+    In most cases, such expressions can be rewritten in a clearer way:</p>
+    <pre>
+true = tuple_size(Tuple) >= 1</pre>
     <p>Variables starting with underscore (_), for example,
-      <c>_Height</c>, are normal variables, not anonymous. They are
-      however ignored by the compiler in the sense that they do not
+      <c>_Height</c>, are normal variables, not anonymous. However,
+      they are ignored by the compiler in the sense that they do not
       generate any warnings for unused variables.</p>
       <p><em>Example:</em></p>
       <p>The following code:</p>
@@ -105,8 +115,8 @@ member(_, []) ->
 member(Elem, []) ->
     [].</pre>
     <p>This causes a warning for an unused variable,
-      <c>Elem</c>, if the code is compiled with the flag
-      <c>warn_unused_vars</c> set. Instead, the code can be rewritten
+      <c>Elem</c>, unless the code is compiled with the option
+      <c>nowarn_unused_vars</c>. Instead, the code can be rewritten
       to:</p>
     <pre>
 member(_Elem, []) ->
@@ -118,6 +128,10 @@ member(_Elem, []) ->
     <p>But this fails:</p>
     <pre>
 {_N,_N} = {1,2}</pre>
+    <p>Also notice that warnings for ignoring the result of an
+    expression will <em>not</em> be suppressed by assigning them to a
+    variable starting with underscore. The only way to suppress such
+    warnings is to assign the expression to the anonymous variable.</p>
     <p>The scope for a variable is its function clause.
       Variables bound in a branch of an <c>if</c>, <c>case</c>, 
       or <c>receive</c> expression must be bound in all branches 


dantswain:
@bjorng Thanks for updating the documentation.  I'm afraid this doesn't really clear it up for me, honestly.  Here is an example snippet of Elixir code (I know, not Erlang, but I don't think that matters here):

```
defmodule UnusedWarning do
  @map %{foo: :bar}

  def no_warn(key) do
    m = @map
    _ = Map.fetch!(m, key)
    Map.get(m, key)
  end

  def also_no_warn(key) do
    m = @map
    _foo = Map.fetch!(m, :foo)
    Map.get(m, key)
  end

  def warn(key) do
    m = @map
    _foo = Map.fetch!(m, key)
    Map.get(m, key)
  end

  def also_warn(key) do
    _foo = Map.fetch!(@map, key)
    Map.get(@map, key)
  end
end
```

The first two functions do not generate a warning, the last two do.  As someone who has quite a bit of Erlang and Elixir experience and is trying to onboard several teammates new to the language, I struggle to find a good explanation for this.  I don't think that this update to the documentation really clears it up.

(Feel free to ask me to move discussion back to JIRA if that's preferred).

bjorng:
There are two distinct warnings. `_` can suppress both types of warnings, a variable starting with underscore only one of them. I'll try to illustrate that with an example in Erlang:

```
-module(double_warning).
-export([double_warning/1]).

double_warning(Map) ->
    Foo = map_get(key, Map),
    ok.
```

The compiler will emit **two** warnings:

```
$ erlc double_warning.erl
double_warning.erl:5: Warning: the result of the expression is ignored (suppress the warning by assigning the expression to the _ variable)
double_warning.erl:5: Warning: variable 'Foo' is unused
$ 
```

The second warning (`Warning: variable 'Foo' is unused`) can be suppressed by using a variable starting with an underscore or by `_`, but the the first warning can only be suppressed with `_`.

Trying to do the same thing in Elixir:

```
defmodule DoubleWarnings do
  def double_warnings(map, key) do
    foo = Map.fetch!(map, key)
  end
end
```

I only get one warning:

```
$  elixirc DoubleWarnings.elx
warning: variable "foo" is unused (if the variable is not meant to be used, prefix it with an underscore)
  DoubleWarnings.elx:3: DoubleWarnings.double_warnings/2
$
```
I'm not sure why there is only one warning. Probably the Elixir compiler does some filtering of Erlang compiler warnings.

Does this explanation make it any clearer?

bjorng:
We might consider changing the behavior of variables that begin with underscore, but that will require an EEP.

The reason that `_` and variables beginning with underscore differently is because we want code like the following to work:

```
some_function(Arg1, _SomeDebugArg) ->
    %% io:format("%p\n", [_SomeDebugArg]),
       .
       .
       .
```

That is, if the comment is removed, the `io:format/2` call should work. Therefore, we would still want variables that begin with underscore to be real variables, but perhaps they could suppress more warnings.

dantswain:
OK Interesting.  The way I have always understood this is that

```
some_function(Arg) ->
  _IgnoredResult = some_other_action(Arg),
  ok.
```

would compile without warning and was generally "better form" than `_ = some_other_action(Arg)` since you could use the part of the variable name after the underscore to provide context for what was being ignored.  That has never emitted a warning for me until OTP 23.

I'm not trying to claim that the way I understood it is necessarily correct (though it made a lot of sense), but in 5+ years I've never seen any documentation or guides or books that made me think I was wrong.  And, to the point at hand, maybe I'm having trouble correcting my wrong way of looking at it but I'm not sure that the way this documentation is worded clarifies my source of confusion.  I can have some of my learners look at it to see if it makes sense to them.

bjorng:
Your example will still compile without a warning. The compiler assumes that `some_other_action/1` is a function with a side effect (e.g. sending a message or updating an ETS table) and will not emit a warning.

However, if the function call is one of the BIFs or operators known by the compiler to be without side effect, the compiler will warn that the result of the call is being ignored. (BIFs without side effect are BIFs for which [erl_bifs:is_pure/3](https://github.com/erlang/otp/blob/a0be3885d65265fe74da6ec601e2bea752e9506d/lib/compiler/src/erl_bifs.erl#L54) returns `true`.)

What has changed in OTP 23 is that `maps:get(Key, Map)` will be rewritten to `map_get(Key, Map)`. `map_get/2` is one of the BIFs that the compiler knows about and will complain if its value is ignored.

What I don't understand why you would call `Map.fetch!(m, key)` and ignore the value. Is that because of some macro expansion in Elixir or some style convention? If I wanted an exception if a certain key was missing in a map I would write it more directly like so:

```
true = Map.has_key?(m, key)
```

bjorng:
After a discussion with some members of the OTP team we have decided that we will probably change the compiler so that all types of warnings can be suppressed by assigning to a variable that starts with underscore. (Variables starting with underscore will still be variables, though.) The formal decision will have to be taken by the OTP Technical Board. If OTB approves the suggestion, it will be implemented in OTP 24.

dantswain:
> Your example will still compile without a warning. The compiler assumes that some_other_action/1 is a function with a side effect (e.g. sending a message or updating an ETS table) and will not emit a warning.

> However, if the function call is one of the BIFs or operators known by the compiler to be without side effect, the compiler will warn that the result of the call is being ignored. (BIFs without side effect are BIFs for which erl_bifs:is_pure/3 returns true.

Wow, I've been using Erlang for I think 7 years and I did not know about this.  I think this is key to what I was not understanding.  Is this in the docs anywhere?

> What I don't understand why you would call Map.fetch!(m, key) and ignore the value. Is that because of some macro expansion in Elixir or some style convention? 

It's not really what I would prefer to do generally, either.  This was based on code that a learner wrote; I was unable to explain to them why this would cause the "ignored value" warning but also explained that I would prefer to handle the underlying business logic a little more directly.

I have occasionally used this pattern - `_val = Map.fetch!(map, key)` as a type of lazy assertion that `key` must be in `map` unless something extremely unusual is going on.  `true = Map.has_key?(map, key)` would give the same info but previously I think I might have preferred `Map.fetch!/2` because if `key` is not in `map` it throws a `KeyError` whereas your version would throw a `MatchError`.  The `KeyError` gets a little more directly to the point imho.

This situation is fairly rare for me - generally I'd either be using `Map.get/3` with a default or I'd check if the key exists and then explicitly handle the case that it does not.  Like I said, it's lazy, but lazy happens sometimes ;)

> After a discussion with some members of the OTP team we have decided that we will probably change the compiler so that all types of warnings can be suppressed by assigning to a variable that starts with underscore. (Variables starting with underscore will still be variables, though.) The formal decision will have to be taken by the OTP Technical Board. If OTB approves the suggestion, it will be implemented in OTP 24.

Wow, it's cool to see this happening.  I hope it's the right thing and not just my confusion - but I trust the team is doing the right thing.

In the meantime, if you don't think that the side effect explanation is elsewhere in the documentation, would you be willing to take a pull request if I wanted to try to help improve the docs?  Documentation improvements are included in patches, right? So it might help someone else in the time between now and OTP 24?

Thanks so much for taking the time to explain all of this, I really appreciate it!

bjorng:
> Wow, I've been using Erlang for I think 7 years and I did not know about this. I think this is key to what I was not understanding. Is this in the docs anywhere?

No, it is not really documented. There is a brief mention of warnings that are emitted during optimization and code generation in the documentation:

https://github.com/erlang/otp/blob/f3d8a1b1051c8c1af7efb0144c03eab8c562c578/lib/compiler/doc/src/compile.xml#L697-L701

In case you are interested, here is the code that emits that type of warnings:
https://github.com/erlang/otp/blob/f3d8a1b1051c8c1af7efb0144c03eab8c562c578/lib/compiler/src/sys_core_fold.erl#L2778-L2851

> previously I think I might have preferred Map.fetch!/2 because if key is not in map it throws a KeyError whereas your version would throw a MatchError

This is a good point.

> In the meantime, if you don't think that the side effect explanation is elsewhere in the documentation, would you be willing to take a pull request if I wanted to try to help improve the docs?

Yes, a pull request based on `maint` will be included in the next patch release. We will accept it if we think it improves the documentation. I don't think that we want a lot of details about compiler warnings in the Reference Manual, but it could be a good idea to expand the paragraph in the documentation of the `compile` module that I quoted above.

bjorng:
I have now created #2831 that allows underscore variables to suppress all warnings.

bjorng:
I am closing this pull request since the behavior of underscore variables will be changed in OTP 24 (assuming that OTB approves it).

PR #2829, opened 2020-10-27
Make printing of floats more user friendly

kjellwinblad:
The pretty printer of floats may round the float value before printing
as long as the following property holds (X is a float value):

X=:=erlang:list_to_float(lists:flatten(io_lib:format("~w",[X]))).

This commit makes sure that the rounded digit always appear last, or
just before e if the float is printed using scientific notation. This
is accomplished by always printing using scientific notation if the
number is so large that it might be rounded before the fractional
part.

Before this commit, users could see the following in the Erlang shell:

> 36028797018963970.0.
36028797018963970.0
> 36028797018963970.0 == 36028797018963970.
false
> 36028797018963970.0 == 36028797018963968.
true

36028797018963970.0 is represented as the float value
36028797018963968.0 but the pretty printer rounds it to
36028797018963970.0 as all numbers in the range [36028797018963966.0,
36028797018963972.0] are represented by the float value
36028797018963968.0. However, this can be confusing to users
(https://bugs.erlang.org/browse/ERL-1308) as it is difficult to guess
that this is what is happening.

After this commit, the above example becomes as follows:

> 36028797018963968.0.
3.602879701896397e16
> 3.602879701896397e16 == 36028797018963970.
false
> 3.602879701896397e16 == 36028797018963968.
true

Which is less confusing because it is natural to think that the last
digit of a pretty printed float value can be rounded.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -2038,10 +2038,10 @@ true</pre>
 "3.00000000000000044409e-01"
 </pre>
 <p>In the last example, <c>float_to_list(0.1+0.2)</c> evaluates to
-<c>"3.00000000000000044409e-01"</c>. The reason for this is that
-<seeguide marker="system/reference_manual:data_types#float_representation_problem"><c>0.1</c> and
-<c>0.2</c> cannot be represented exactly as float
-values</seeguide>.</p>
+<c>"3.00000000000000044409e-01"</c>. The reason for this is explained
+in <seeguide
+marker="system/reference_manual:data_types#float_representation_problem">Representation
+of Floating Point Numbers</seeguide>.</p>
       </desc>
     </func>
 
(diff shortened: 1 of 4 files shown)


erts/doc/src/erlang.xml
```diff
@@ -12127,10 +12131,15 @@ improper_end</pre>
...
-<p>In the last example, <c>trunc(36028797018963969.0)</c> evaluates to
-<c>36028797018963968</c>. The reason for this is that <seeguide
-marker="system/reference_manual:data_types#float_representation_problem"><c>36028797018963969.0</c>
-cannot be represented exactly as a float value</seeguide>.</p>
+      <p>In the last example, <c>trunc(36028797018963969.0)</c>
+      evaluates to <c>36028797018963968</c>. The reason for this is
+      that the float literal <c>36028797018963969.0</c> gets converted
+      to the float value <c>36028797018963968.0</c> as
+      <c>36028797018963968.0</c> is the closest value that can be
+      represented exactly as a float value. See <seeguide
```
zuiderkwast, line 12139: I think this sentence can be improved. Repetitions of 36028797018963968.0 and the word "value" and the formulation "gets converted" seems odd.

How about something like this?: "The reason for this is that the number 36028797018963969.0 cannot be represented exactly as a float value. Instead, the float literal is represented as 36028797018963968.0, which the closest number that can be represented exactly as a float value."

lib/stdlib/src/io_lib_format.erl
```diff
@@ -649,7 +660,19 @@ insert_decimal(Place, S) ->
                     end;
                 true ->
                     if
-                        Place - L + 2 =< ExpCost ->
+                        %% All integers in the range [-2^53, 2^53] can
+                        %% be stored without lose of precision in an
```
zuiderkwast, line 664: *loss* of precision (multiple occurrences)

zuiderkwast (commented):
Interesting, nice solution!

lib/stdlib/src/io_lib_format.erl
kjellwinblad: Thanks for seeing this!

erts/doc/src/erlang.xml
kjellwinblad: Thanks, I like your suggestion. I will add a commit that changes the text according to your suggestion.

kjellwinblad: Fixed in 9443da9

lib/stdlib/src/io_lib_format.erl
kjellwinblad: Fixed in 9443da9

PR #2830, opened 2020-10-28
jit: Fix overflow check in 64-bit bs_get_integer

jhogberg:
https://bugs.erlang.org/browse/ERL-1391


erts/emulator/test/bs_match_int_SUITE.erl
```diff
@@ -347,7 +347,13 @@ bignum(Config) when is_list(Config) ->
 
     BignumBin = id(<<0:512/unit:8,258254417031933722623:9/unit:8>>),
     <<258254417031933722623:(512+9)/unit:8>> = BignumBin,
-    erlang:garbage_collect(),			%Search for holes in debug-build.
+    erlang:garbage_collect(),			%Search for holes in debug-build
+
+    %% ERL-1391
+    V = -1 bsl 63,
```
garazdawi, line 353: Maybe add testcases for `-1 bsl 59..64` just to be on the safe side?

wojtekmach:
Thank you!

PR #2833, opened 2020-10-29
monitor_node/2: Raise 'notalive' instead of 'badarg'

bjorng:
The `monitor_node/2` BIF raises a `badarg` exception when the
node is not alive. This commit changes `monitor_node/2` to
instead raise a `notalive` exception in that situation.

Specification
-------------

Currently, `monitor_node/2` raises a `badarg` exception when
the node is not distributed (not alive):

    1> monitor_node(node@host, true).
         in function  monitor_node/2
            called as monitor_node(node@host,true)

Change that to raise a `notalive` exception:

    1> monitor_node(node@host, true).
    ** exception error: the node cannot be part of a distributed system
         in function  monitor_node/2
            called as monitor_node(node@host,true)
    2> catch monitor_node(node@host, true).
    {'EXIT',{notalive,[{erlang,monitor_node,[node@host,true],[]},
                       {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,684}]},
                       {erl_eval,expr,5,[{file,"erl_eval.erl"},{line,437}]},

Motivation
----------

`badarg` is illogical if there is nothing wrong with the arguments. To
avoid confusion, it is better to only raise a `badarg` exception if
the arguments are bad.

Backwards Compatibility
-----------------------

Applications or libraries that catch the exception from
`monitor_nodes/2` and expect the exception reason to always be
`badarg` will need to be updated. However, in practice, there is
probably little or no code that does that. If an application uses
`monitor_node/2`, it probably expects that the distribution is
enabled.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -4031,7 +4031,7 @@ RealSystem = system + MissedSystem</code>
           If this fails, a <c>nodedown</c> message is delivered.</p>
         <p>Nodes connected through hidden connections can be monitored
           as any other nodes.</p>
-        <p>Failure: <c>badarg</c> if the local node is not alive.</p>
+        <p>Failure: <c>notalive</c> if the local node is not alive.</p>
       </desc>
     </func>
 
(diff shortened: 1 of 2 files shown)


bjorng:
This pull request has been approved by the OTP Technical Board.

PR #2834, opened 2020-10-29
Skip literals on v3_core and sys_core_alias

josevalim:
Skip literals on v3_core and sys_core_alias

When compiling code with large amount of literals,
there was a performance degradation between Erlang/OTP
22.3 and 23.1.

For example, the gisted code below takes 10 extra seconds:

https://gist.github.com/josevalim/694c1799143fcf25e43aa27e3e11e4c1

In particular, the v3_core pass went from 1.8 seconds to 3.2
seconds. When profiling the pass, we could see a lot of time
was was spent on the new lowering pass in v3_core. In particular,
cerl_trees:mapfold/3 would show up in profilers:

    cerl_trees:mapfold/4   3220377    19.14   2447684  [0.76]

Which is quite curious, as the linked code does not trigger
any of the lowering optimizations.

Under further analysis, this happened because cerl_trees would
traverse literals, which is a waste given there is no lowering
to do inside literals.

This commit speeds up both v3_core since cerl_trees is invoked
33% less frequently and the pass is overall 18% faster.
sys_core_alias also got 33% faster. The same technique may be
appliable to other passes.

### DIFF
--- a/lib/compiler/src/cerl_trees.erl
+++ b/lib/compiler/src/cerl_trees.erl
@@ -352,131 +352,141 @@ mapfold(F, S0, T) ->
 
 
 %% @spec mapfold(Pre, Post, Initial::term(), Tree::cerl()) -> {cerl(), term()}
-%%       Pre  = (cerl(), term()) -> {cerl(), term()}
+%%       Pre  = (cerl(), term()) -> {cerl(), term()} | skip
 %%       Post = (cerl(), term()) -> {cerl(), term()}
 %%
 %% @doc Does a combined map/fold operation on the nodes of the
 %% tree. It begins by calling <code>Pre</code> on the tree, using the
-%% <code>Initial</code> value. It then deconstructs the top node of
+%% <code>Initial</code> value. <code>Pre</code> must either return a
+%% tree with an updated accumulator or the atom <code>skip</code>.
+%%
+%% If a tree is returned, this function deconstructs the top node of
 %% the returned tree and recurses on the children, using the returned
 %% value as the new initial and carrying the returned values from one
 %% call to the next. Finally it reassembles the top node from the
 %% children, calls <code>Post</code> on it and returns the result.
+%%
+%% If <code>skip</code> is returned, it returns the tree and accumulator
+%% as is.
 
--spec mapfold(fun((cerl:cerl(), term()) -> {cerl:cerl(), term()}),
+-spec mapfold(fun((cerl:cerl(), term()) -> {cerl:cerl(), term()} | skip),
               fun((cerl:cerl(), term()) -> {cerl:cerl(), term()}),
 	      term(), cerl:cerl()) -> {cerl:cerl(), term()}.
 
 mapfold(Pre, Post, S00, T0) ->
-    {T, S0} = Pre(T0, S00),
-    case type(T) of
- 	literal ->
-	    case concrete(T) of
-		[_ | _] ->
+    case Pre(T0, S00) of
+	{T, S0} ->
+	    case type(T) of
+		literal ->
+		    case concrete(T) of
+			[_ | _] ->
+			    {T1, S1} = mapfold(Pre, Post, S0, cons_hd(T)),
+			    {T2, S2} = mapfold(Pre, Post, S1, cons_tl(T)),
+			    Post(update_c_cons(T, T1, T2), S2);
+			V when tuple_size(V) > 0 ->
+			    {Ts, S1} = mapfold_list(Pre, Post, S0, tuple_es(T)),
+			    Post(update_c_tuple(T, Ts), S1);
+			_ ->
+			    Post(T, S0)
+		    end;
+		var ->
+		    Post(T, S0);
+		values ->
+		    {Ts, S1} = mapfold_list(Pre, Post, S0, values_es(T)),
+		    Post(update_c_values(T, Ts), S1);
+		cons ->
 		    {T1, S1} = mapfold(Pre, Post, S0, cons_hd(T)),
 		    {T2, S2} = mapfold(Pre, Post, S1, cons_tl(T)),
-		    Post(update_c_cons(T, T1, T2), S2);
-		V when tuple_size(V) > 0 ->
+		    Post(update_c_cons_skel(T, T1, T2), S2);
+		tuple ->
 		    {Ts, S1} = mapfold_list(Pre, Post, S0, tuple_es(T)),
-		    Post(update_c_tuple(T, Ts), S1);
-		_ ->
-		    Post(T, S0)
+		    Post(update_c_tuple_skel(T, Ts), S1);
+		map ->
+		    {M , S1} = mapfold(Pre, Post, S0, map_arg(T)),
+		    {Ts, S2} = mapfold_list(Pre, Post, S1, map_es(T)),
+		    Post(update_c_map(T, M, Ts), S2);
+		map_pair ->
+		    {Op,  S1} = mapfold(Pre, Post, S0, map_pair_op(T)),
+		    {Key, S2} = mapfold(Pre, Post, S1, map_pair_key(T)),
+		    {Val, S3} = mapfold(Pre, Post, S2, map_pair_val(T)),
+		    Post(update_c_map_pair(T,Op,Key,Val), S3);
+		'let' ->
+		    {Vs, S1} = mapfold_list(Pre, Post, S0, let_vars(T)),
+		    {A, S2} = mapfold(Pre, Post, S1, let_arg(T)),
+		    {B, S3} = mapfold(Pre, Post, S2, let_body(T)),
+		    Post(update_c_let(T, Vs, A, B), S3);
+		seq ->
+		    {A, S1} = mapfold(Pre, Post, S0, seq_arg(T)),
+		    {B, S2} = mapfold(Pre, Post, S1, seq_body(T)),
+		    Post(update_c_seq(T, A, B), S2);
+		apply ->
+		    {E, S1} = mapfold(Pre, Post, S0, apply_op(T)),
+		    {As, S2} = mapfold_list(Pre, Post, S1, apply_args(T)),
+		    Post(update_c_apply(T, E, As), S2);
+		call ->
+		    {M, S1} = mapfold(Pre, Post, S0, call_module(T)),
+		    {N, S2} = mapfold(Pre, Post, S1, call_name(T)),
+		    {As, S3} = mapfold_list(Pre, Post, S2, call_args(T)),
+		    Post(update_c_call(T, M, N, As), S3);
+		primop ->
+		    {N, S1} = mapfold(Pre, Post, S0, primop_name(T)),
+		    {As, S2} = mapfold_list(Pre, Post, S1, primop_args(T)),
+		    Post(update_c_primop(T, N, As), S2);
+		'case' ->
+		    {A, S1} = mapfold(Pre, Post, S0, case_arg(T)),
+		    {Cs, S2} = mapfold_list(Pre, Post, S1, case_clauses(T)),
+		    Post(update_c_case(T, A, Cs), S2);
+		clause ->
+		    {Ps, S1} = mapfold_list(Pre, Post, S0, clause_pats(T)),
+		    {G, S2} = mapfold(Pre, Post, S1, clause_guard(T)),
+		    {B, S3} = mapfold(Pre, Post, S2, clause_body(T)),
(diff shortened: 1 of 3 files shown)


josevalim:
@jhogberg @bjorng, I haven't run the tests yet, so apologies if things fail, but I would like to get your thoughts before I commit further to this.

I am mostly concern this slowdown is also present in other parts of the compiler, especially because functions like `cerl_trees:mapfold/4` promote traversing all nodes, while manual recursion would make you think harder if you need to go deeper into a subtree or not.

This also doesn't bring the compilation times in v3_core back to Erlang/OTP 22. It may be the best way to address this is by merging the passes in v3_core into a single pass but I have no idea how feasible that is.

josevalim:
Actually, I realize this optimization can be a bit (quite?) more aggressive. We cannot have cases or receives inside a tuple nor inside a binary. I assume we won't find them inside `call`s either. Is there a function inside `cerl` that gives me this property? Or do I need to "handwrite" it?

michalmuskala:
What do you mean by no cases or receives inside tuples? I think this is perfectly valid code:
```erlang
{
    case X of
        true -> 1;
        false -> 0
    end,
    receive
        Value -> Value
    end
}
```

josevalim:
@michalmuskala in this particular pass I *believe* they have already been hoisted outside of the tuple.

michalmuskala:
Ah that's fair. We're in the world of core erlang - sorry I forgot about that.

josevalim:
Here is another diff that I have been exploring while working on the same file linked above:

```diff
diff --git a/lib/compiler/src/beam_ssa.erl b/lib/compiler/src/beam_ssa.erl
index 2a02eb3cd0..694aad21d9 100644
--- a/lib/compiler/src/beam_ssa.erl
+++ b/lib/compiler/src/beam_ssa.erl
@@ -43,6 +43,8 @@
          trim_unreachable/1,
          used/1,uses/1,uses/2]).
 
+-export([uses_rpo/2, dominators_rpo/2]).
+
 -export_type([b_module/0,b_function/0,b_blk/0,b_set/0,
               b_ret/0,b_br/0,b_switch/0,terminator/0,
               b_var/0,b_literal/0,b_remote/0,b_local/0,
@@ -386,7 +388,13 @@ dominators(Blocks) ->
       Preds :: predecessor_map(),
       Result :: {dominator_map(), numbering_map()}.
 dominators(Blocks, Preds) ->
-    Top0 = rpo(Blocks),
+    dominators_rpo(rpo(Blocks), Blocks, Preds).
+
+dominators_rpo(Top, Blocks) ->
+    Preds = predecessors(Blocks),
+    dominators_rpo(Top, Blocks, Preds).
+
+dominators_rpo(Top0, _Blocks, Preds) ->
     Df = maps:from_list(number(Top0, 0)),
     [{0,[]}|Top] = [{L,map_get(L, Preds)} || L <- Top0],
 
@@ -662,6 +670,9 @@ uses(Blocks) ->
 uses(From, Blocks) ->
     fold_rpo(fun fold_uses_block/3, From, #{}, Blocks).
 
+uses_rpo(Top, Blocks) ->
+    fold_rpo_1(Top, fun fold_uses_block/3, Blocks, #{}).
+
 fold_uses_block(Lbl, #b_blk{is=Is,last=Last}, UseMap0) ->
     F = fun(I, UseMap) ->
                 foldl(fun(Var, Acc) ->
diff --git a/lib/compiler/src/beam_ssa_bool.erl b/lib/compiler/src/beam_ssa_bool.erl
index 208f27a891..c7b484da5b 100644
--- a/lib/compiler/src/beam_ssa_bool.erl
+++ b/lib/compiler/src/beam_ssa_bool.erl
@@ -148,10 +148,11 @@ opt_function(#b_function{bs=Blocks0,cnt=Count0}=F) ->
     DefVars = interesting_defs(Blocks1),
     if
         map_size(DefVars) > 1 ->
-            Dom = beam_ssa:dominators(Blocks1),
-            Uses = beam_ssa:uses(Blocks1),
+            Top = beam_ssa:rpo(Blocks1),
+            Dom = beam_ssa:dominators_rpo(Top, Blocks1),
+            Uses = beam_ssa:uses_rpo(Top, Blocks1),
             St0 = #st{defs=DefVars,count=Count1,dom=Dom,uses=Uses},
-            {Blocks2,St} = bool_opt(Blocks1, St0),
+            {Blocks2,St} = bool_opt(Top, Blocks1, St0),
             Count = St#st.count,
 
             %% When merging blocks, phi nodes must have the same
@@ -202,7 +203,7 @@ pre_opt(Blocks, Count) ->
 
     %% Collect information to help the pre_opt pass to optimize
     %% `switch` instructions.
-    Sub0 = #{uses => {uses,Blocks}},
+    Sub0 = #{uses => {uses,Top,Blocks}},
     Sub1 = get_phi_info(Top, Blocks, Sub0),
     Sub = maps:remove(uses, Sub1),
 
@@ -280,8 +281,8 @@ get_phi_info_single_use(Var, Sub) ->
                  #{Var:=[_]} -> true;
                  #{Var:=[_|_]} -> false
              end,Sub};
-        {uses,Blocks} ->
-            Uses = beam_ssa:uses(Blocks),
+        {uses,Top,Blocks} ->
+            Uses = beam_ssa:uses_rpo(Top, Blocks),
             get_phi_info_single_use(Var, Sub#{uses => Uses})
     end.
 
@@ -565,9 +566,6 @@ interesting_defs_is([], _L, Acc) -> Acc.
 %%% interior '=:=' instruction we will visit the blocks in postorder.
 %%%
 
-bool_opt(Blocks, St) ->
-    bool_opt(beam_ssa:rpo(Blocks), Blocks, St).
-
 bool_opt([L|Ls], Blocks0, St0) ->
     {Blocks,St1} = bool_opt(Ls, Blocks0, St0),
     case Blocks of
```

It seems that many SSA operations rely on `beam_ssa:rpo/1` and we end-up computing those over and over again. Often, we end-up computing the same rpo, as seen in the diff above. The changes above make `beam_ssa_bool` roughly 10% faster and ultimately we go from this:

    beam_ssa:rpo_1/4                                        802391    10.72   375843  [      0.47]

to this:

    beam_ssa:rpo_1/4                                        481526     6.35   203949  [      0.42]

I saw opportunity to apply this same optimization in other places, such as a couple passes in `beam_ssa_precodegen`. It partially feels that maybe those functions should always except the "rpo" as input? This will make much more clearer if we are recomputing the same data over and over again and when to share it.

bjorng:
> Actually, I realize this optimization can be a bit (quite?) more aggressive. We cannot have cases or receives inside a tuple nor inside a binary. I assume we won't find them inside calls either. Is there a function inside cerl that gives me this property? 

I am not aware of one function that includes all of them.

Within `v3_core` the function `is_safe/1` would give you some of them. There is also `cerl:is_data/1` that returns `true` for 
lists, tuples, and literals.

Note that Core Erlang code could be generated by other code generators than `v3_core`. So in `sys_core_fold` and `sys_core_alias` you can't really assume anything (I think it's OK if you make an assumption and that only result in a missed opportunity for optimization).

bjorng:
> It may be the best way to address this is by merging the passes in v3_core into a single pass but I have no idea how feasible that is.

I would say not feasible. The first three passes were created by Robert Virding. I remember that he spent some time trying to use fewer passes. The fourth pass was created by me. I don't remember whether it was completely impossible to combine the third and fourth passes, but it would certainly be a non-trivial task.

bjorng:
This looks good to me.

>  I haven't run the tests yet, so apologies if things fail, but I would like to get your thoughts before I commit further to this.

Update the primary bootstrap and build all of OTP from scratch. Then run the compiler test suite. (Don't push the commit with the primary bootstrap to github; we will take care of adding a primary bootstrap in our daily builds.) Tell me when you have done that successfully;  after that will we add this PR to our daily builds.

josevalim:
> Note that Core Erlang code could be generated by other code generators than v3_core. So in sys_core_fold and sys_core_alias you can't really assume anything (I think it's OK if you make an assumption and that only result in a missed opportunity for optimization).

The lowering code must always run though, right? If we don't, later passes will break? If that's the case, then this must stay as is, at least the v3_core one. We could be more aggressive in the `sys_core_alias` one... but I am not really sure if it is worth it.

> I would say not feasible. 

I will continue exploring other venues then. The other places that are doing full traversals and could be fixed are:

1. sys_core_prepare although not a concern at first because it was not generated by v3_core and likely not having the tree explosion issue associated with binstrings

2. sys_core_bsm - today the pass is extremely simple and could even be done in v3_kernel: we could re-order when we handle #c_case in v3. However, this pass has shrunk considerably over the last years, so I suspect you already have a plan for it

Thank you for the time and the replies, as always! :heart:

josevalim:
@bjorng apologies, I have one extra question. Would you accept a patch that changes `beam_ssa` functions to accept the `Rpo` instead of `From`, as I mentioned in an earlier comment? Let me know if you:

a. Would accept a PR
b. Welcome the change but you'd prefer if someone in the compiler team did it (because I think you will spot more optimizations than I would)
c. Not interested

If I did it, I would also do it in parts (i.e. I would start with three functions then slowly convert a couple more). :)

bjorng:
> The lowering code must always run though, right? If we don't, later passes will break? If that's the case, then this must stay as is, at least the v3_core one.

Yes, the lowering must always run. You **can** be more aggressive in `v3_core` because you KNOW that the previous sub passes in `v3_core` have broken out cases from other constructs such as building of tuples.

> sys_core_prepare although not a concern at first because it was not generated by v3_core and likely not having the tree explosion issue associated with binstrings

I would say not bother until we have a real-world use case (i.e. slow compilation of real code for one of the BEAM-based languages that generate Core Erlang code).

> sys_core_bsm - today the pass is extremely simple and could even be done in v3_kernel: we could re-order when we handle #c_case in v3. However, this pass has shrunk considerably over the last years, so I suspect you already have a plan for it

No, we don't have a plan for `sys_core_bsm`. We have a plan to eliminate `v3_kernel` (without eliminating any of the optimizations done in it). We would prefer to avoid adding more code to `v3_kernel` in the meantime. If there would be a substantial win for  a compilation time I would suggest that you investigate whether you could move it into `sys_core_fold`.

> Let me know if you:

> a. Would accept a PR

Yes.

> If I did it, I would also do it in parts (i.e. I would start with three functions then slowly convert a couple more).

That is better for us, too.

josevalim:
Ack on all fronts. Hopefully one last question, is this list correct for v3_core lowering pass?

```erl
%% Those nodes do not have case or receive within them,
%% so we can speed up lowering by avoiding traversing them.
has_case_or_receive(#c_binary{}) -> false;
has_case_or_receive(#c_bitstr{}) -> false;
has_case_or_receive(#c_call{}) -> false;
has_case_or_receive(#c_cons{}) -> false;
has_case_or_receive(#c_literal{}) -> false;
has_case_or_receive(#c_map{}) -> false;
has_case_or_receive(#c_map_pair{}) -> false;
has_case_or_receive(#c_primop{}) -> false;
has_case_or_receive(#c_tuple{}) -> false;
has_case_or_receive(_) -> true.
```

bjorng:
> is this list correct for v3_core lowering pass?

Yes.

But I don't think you'll need `#c_map_pair{}`.

(Run the compiler test suite with the `cover` option and check that you code is covered.)

josevalim:
Good idea on cover! We didn't need c_map_pair nor c_bitstr.

I have updated the code to apply it to more cases, improved the commit message, updated the bootstrap, recompiled, and ran the test suite. All tests pass.

bjorng:
> I have updated the code...

It seems that you have not actually pushed the updated code to github.

josevalim:
@bjorng now properly force pushed, thanks.

lib/compiler/src/cerl_trees.erl
```diff
@@ -352,7 +352,7 @@ mapfold(F, S0, T) ->
 
 
 %% @spec mapfold(Pre, Post, Initial::term(), Tree::cerl()) -> {cerl(), term()}
-%%       Pre  = (cerl(), term()) -> {cerl(), term()}
+%%       Pre  = (cerl(), term()) -> {cerl(), term()} | skip
 %%       Post = (cerl(), term()) -> {cerl(), term()}
 %%
 %% @doc Does a combined map/fold operation on the nodes of the
```
bjorng, line 358: Please update the documentation to mention what `skip` means.

josevalim: Done!

bjorng:
Added to our daily builds.

PR #2835, opened 2020-10-29
compile: drop dead code

richcarl:
@bjorng If this code has effectively been running as an assertion for the last 4 years, it's probably safe to say that it's not actually doing anything anymore.

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -1958,7 +1958,6 @@ listing(LFun, Ext, Code, St) ->
     Lfile = outfile(St#compile.base, Ext, St#compile.options),
     case file:open(Lfile, [write,delayed_write]) of
 	{ok,Lf} ->
-            Code = restore_expanded_types(Ext, Code),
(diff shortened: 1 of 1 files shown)


bjorng:
Thanks for noticing this unnecessary code.

I ran the compiler test suite with the `cover` option and noticed that the code is actually unreachable. The reason is that the following clause will never be executed:

https://github.com/erlang/otp/blob/0f0100618e3d050ea9f00517a4d7ba00bfb740b9/lib/compiler/src/compile.erl#L1992-L1994

The reason is that the intermediate representation of the code passed in the second argument is now a list instead of a tuple. My guess is that the clause became unreachable when the representation was changed when the `sys_pre_expand` pass was eliminated.

Can you update the commit message to explain why you are removing the code? We want the information in the commit message and not in the pull request description so that when someone stumbles over the change in 2030 the reason for the change is immediately available.

bjorng:
Thanks for noticing the unreachable code. I have committed the code removal with more details in the commit message.

PR #2836, opened 2020-10-29
compile: make select_passes clause order match the comment text

richcarl:
If you tried to use a {listing, Fun} pass, it would be consumed by the {listing, Ext} clause before it reached the generic {_, Fun} clause further below (and similar for any other case where the name would match even though the second element is a fun).

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -694,6 +694,11 @@ select_passes([{pass,Mod}|Ps], Opts) ->
 		end
 	end,
     [{Mod,F}|select_passes(Ps, Opts)];
+select_passes([{_,Fun}=P|Ps], Opts) when is_function(Fun) ->
+    [P|select_passes(Ps, Opts)];
+select_passes([{_,Test,Fun}=P|Ps], Opts) when is_function(Test),
+					      is_function(Fun) ->
+    [P|select_passes(Ps, Opts)];
 select_passes([{src_listing,Ext}|_], _Opts) ->
     [{listing,fun (Code, St) -> src_listing(Ext, Code, St) end}];
 select_passes([{listing,Ext}|_], _Opts) ->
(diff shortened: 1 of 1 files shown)


michalmuskala:
I don't think this change is correct - in particular a listing pass halts execution and does not process any following passes. I believe with this change those custom listing passes won't behave like this.

richcarl:
> I don't think this change is correct - in particular a listing pass halts execution and does not process any following passes. I believe with this change those custom listing passes won't behave like this.

You can add 'done' after your custom pass to get that effect.

bjorng:
I have a some questions:

How do you actually run custom passes? I wasn't aware that there is a way to do that.

Do you have to call your pass a `{listing, Fun}`? Could you just call it `{custom_listing,Fun}`?

Assuming that you do need this change, we would want an explanation in the **commit message** why this change is needed (similar to the description in the pull request description). Also, please update the comment for `select_passes/2` to reflect your changes to the code.

richcarl:
> I have a some questions:
> 
> How do you actually run custom passes? I wasn't aware that there is a way to do that.

There isn't through the current API, but if you edit the code to make experiments you would like this sort of thing to just work without surprises (even for maintainers at OTP).

> Do you have to call your pass a `{listing, Fun}`? Could you just call it `{custom_listing,Fun}`?

Sure, but then you'd need to make select_passes understand yet another case, which seemed like unnecessary clutter.

> Assuming that you do need this change, we would want an explanation in the **commit message** why this change is needed (similar to the description in the pull request description). Also, please update the comment for `select_passes/2` to reflect your changes to the code.

bjorng:
> Sure, but then you'd need to make select_passes understand yet another case, which seemed like unnecessary clutter.

OK. I just wanted be sure. We'll accept this PR when you have made the changes I suggested.

richcarl:
My first attempt was a minimalistic change, but now I have rewritten it to move up the generic {Name, Fun} clauses to match the comment text. This seems easier to motivate and is a better change in the long run to minimize surprises.

bjorng:

bjorng:
Thanks for this pull request!

PR #2837, opened 2020-10-29
Also collapse strings in patterns

josevalim:
Prior to this patch, v3_core would expand each
character in a string to an integer. This made
patterns with large strings to become large and
expensive ASTs. For example, this file:

https://gist.github.com/josevalim/694c1799143fcf25e43aa27e3e11e4c1

would grow its representation in more than 10x!

This commit makes it so we squeeze strings into
large integers also when building patterns,
mimicking an optimization that we already did for
strings outside of patterns.

Before this patch:

    expand_records   :  0.077 s   19988.7 kB
    core             :  2.552 s  373294.0 kB
    sys_core_fold    :  0.868 s  370212.9 kB
    sys_core_alias   :  0.237 s  370212.9 kB
    core_transforms  :  0.000 s  370212.9 kB
    sys_core_bsm     :  0.677 s  370212.9 kB
    v3_kernel        :  2.662 s  169439.0 kB

After this patch:

    core             :  0.653 s   72136.4 kB
    sys_core_fold    :  0.482 s   69055.3 kB
    sys_core_alias   :  0.146 s   69055.3 kB
    core_transforms  :  0.000 s   69055.3 kB
    sys_core_bsm     :  0.098 s   69055.3 kB
    v3_kernel        :  2.250 s  169439.0 kB

This commit also changes the maximum collapse size
from 2048 to 1024 in order to mirror the expansion
value in v3_kernel. Making sure that collapses made
in v3_core are expanded in v3_kernel.

### DIFF
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -91,6 +91,9 @@
 
 -include("core_parse.hrl").
 
+%% Matches expansion max segment in v3_kernel.
+-define(COLLAPSE_MAX_SIZE_SEGMENT, 1024).
+
 %% Internal core expressions and help functions.
 %% N.B. annotations fields in place as normal Core expressions.
 
@@ -1152,7 +1155,7 @@ bitstrs([], St) ->
     {[],[],St}.
 
 bitstr({bin_element,Line,{string,_,S},{integer,_,8},_}, St) ->
-    bitstrs(bin_expand_string(S, Line, 0, 0), St);
+    bitstrs(bin_expand_string(S, Line, 0, 0, []), St);
 bitstr({bin_element,Line,{string,_,[]},Sz0,Ts}, St0) ->
     %% Empty string. We must make sure that the type is correct.
     {[#c_bitstr{size=Sz}],Eps0,St1} =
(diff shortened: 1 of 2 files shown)


josevalim:
It may be that keeping those as strings would actually be the most efficient thing to do and handle them only on v3_kernel. But I believe it would require changes across `v3_core` (perhaps even syntax/parser ones).

okeuday:
@josevalim Doesn't this change make matching on a prefix of a large string slower?  If the string has to be 1024 character or longer to have a slowdown, that is quite large, though I wanted to check if that is possible.

josevalim:
@okeuday, this patch should not change performance/behaviour at all. The reason it doesn't affect matching is because v3_kernel expands integers with size <= 1024 back into integers of size 8 in order to build a matching tree. One way to interpret this patch is as if it was compressing AST nodes into large integers because they are cheaper to represent inside binaries.

bjorng:
Looks good to me.

Have you run tests (built OTP with an updated primary bootstrap + compiler test suite)?

Have you run `scripts/diffable` to verify that the generated code didn't change? (Invoke `scripts/diffable` without arguments to print usage instructions.)

josevalim:
I will do all that!

> Have you run tests (built OTP with an updated primary bootstrap + compiler test suite)?

To make sure I don't screw this up, I should:

1. Apply my changes
2. `cd lib/compiler && make`
3. `./otp_build update_primary`
4. `./otp_build all -a && ./otp_build tests`
5. Run compiler suite

bjorng:
Yes.

josevalim:
@bjorng all tests have passed after the primary has been updated and the source recompiled.

bjorng:

lib/compiler/src/v3_core.erl
```diff
@@ -2117,7 +2120,9 @@ pat_bin(Ps0, St) ->
     pat_segments(Ps, St).
 
 pat_bin_expand_strings(Es0) ->
-    foldr(fun ({bin_element,Line,{string,_,S},Sz,Ts}, Es1) ->
+    foldr(fun ({bin_element,Line,{string,_,S},default,_Ts}, Es1) ->
```
josevalim, line 2123:
```suggestion
    foldr(fun ({bin_element,Line,{string,_,S},default,default}, Es1) ->
```

Hi @bjorng, I forgot to run scripts/diffable. I did it just now. I think this change is necessary otherwise we will wrongly compile any encoded string.

Although I am thinking maybe we should just convert other types into integers right here... I already do a similar optimization in `v3_kernel` (but only for UTF-8).

josevalim:
Ok, I have force pushed. It includes two fixes:

1. fixes xmerl related to utf-16 binaries
2. fixes scripts/diffable difference related to empty strings

We should be good to go now!

bjorng:
I have included the updated PR in our daily builds.

bjorng:
Thanks for this pull request!

PR #2838, opened 2020-10-30
Improve compile:forms/2 doc with newly exported type

paulo-ferraz-oliveira:
This is a very localised enhancement (i.e. I didn't improve the whole `compile.xml`). If there's anything not "too out of context" that you'd like me to fix in the same pull request, let me know.

### DIFF
--- a/lib/compiler/doc/src/compile.xml
+++ b/lib/compiler/doc/src/compile.xml
@@ -43,6 +43,10 @@
       <name>option() = term()</name>
       <desc><p>See <seemfa marker="#file/2">file/2</seemfa> for detailed description</p></desc>
     </datatype>
+    <datatype>
+      <name>forms() = term()</name>
+      <desc><p>List of Erlang abstract or Core Erlang format representations, as used by <seemfa marker="#forms/2">forms/2</seemfa></p></desc>
+    </datatype>
   </datatypes>
 
   <funcs>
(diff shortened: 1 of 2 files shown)


lib/compiler/doc/src/compile.xml
```diff
@@ -774,11 +778,15 @@ module.beam: module.erl \
       <name since="">forms(Forms, Options) -> CompRet</name>
       <fsummary>Compiles a list of forms.</fsummary>
       <type>
-        <v>Forms = [Form]</v>
+        <v>Forms = <seetype marker="#forms">forms()</seetype></v>
+        <v>forms() = [<seetype marker="erl_parse#abstract_form">erl_parse:abstract_form</seetype>] | <seetype marker="cerl#type-c_module">cerl:c_module()</seetype></v>
```
garazdawi, line 782:
```suggestion
        <v>forms() = [<seetype marker="erl_parse#abstract_form">erl_parse:abstract_form</seetype>] | <seeerl marker="cerl#type-c_module">cerl:c_module()</seeerl></v>
```

Since cerl docs has been generated using edoc the types in it do not use types from specs, so you need to use the `<seeerl>` tag for links.

paulo-ferraz-oliveira: 😄 Good to know, thanks.

paulo-ferraz-oliveira: I've force-pushed my last commit. Will check back again to see if tests are passing.

paulo-ferraz-oliveira: Everything's green now. Thanks much.

bjorng:
Can you also update the explanation inside the parentheses to also mention Core Erlang?

https://github.com/erlang/otp/blob/9225b8d5ba477cde52cd1814e4ede109d2b4cc6a/lib/compiler/doc/src/compile.xml#L784-L785

Also, please squash at least the last two commits.

paulo-ferraz-oliveira:
> Can you also update the explanation inside the parentheses to also mention Core Erlang?

Sure, can you tell me the exact text you'd prefer there? Would this be enough as a change?

```
(in either Erlang abstract format representation or Core Erlang)
```

paulo-ferraz-oliveira:
@bjorng: I've squashed the commits as requested and pushed my understanding of the changes you want on the doc: 59f9771

bjorng:
Thanks for your pull request.

PR #2839, opened 2020-10-30
Expliticly pass in rpo as arguments

josevalim:
Many functions in beam_ssa would receive labels
and compute the rpo from labels. This lead to
multiple calls to rpo which can be inefficient
in some scenarios.

This commit changes these functions to receive
the rpo as argument and shares them in the most
trivial cases found so far.

beam_ssa:fold_po/2 was removed as it is not used
anywhere. The remaining functions will be converted
in another PR.

### DIFF
--- a/lib/compiler/src/beam_ssa.erl
+++ b/lib/compiler/src/beam_ssa.erl
@@ -24,14 +24,14 @@
          between/4,
          clobbers_xregs/1,def/2,def_unused/3,
          definitions/1,
-         dominators/1,dominators/2,common_dominators/3,
-         flatmapfold_instrs_rpo/4,
-         fold_po/3,fold_po/4,fold_rpo/3,fold_rpo/4,
-         fold_instrs_rpo/4,
+         dominators/2,dominators_from_predecessors/2,common_dominators/3,
+         flatmapfold_instrs/4,
+         fold_blocks/4,
+         fold_instrs/4,
          is_loop_header/1,
          linearize/1,
-         mapfold_blocks_rpo/4,
-         mapfold_instrs_rpo/4,
+         mapfold_blocks/4,
+         mapfold_instrs/4,
          merge_blocks/1,
          normalize/1,
          no_side_effect/1,
@@ -41,7 +41,7 @@
          split_blocks/3,
          successors/1,successors/2,
          trim_unreachable/1,
-         used/1,uses/1,uses/2]).
+         used/1,uses/2]).
 
 -export_type([b_module/0,b_function/0,b_blk/0,b_set/0,
               b_ret/0,b_br/0,b_switch/0,terminator/0,
@@ -361,7 +361,7 @@ def_unused(Ls, Unused, Blocks) ->
     Preds = cerl_sets:from_list(Top),
     def_unused_1(Blks, Preds, [], Unused).
 
-%% dominators(BlockMap) -> {Dominators,Numbering}.
+%% dominators(Labels, BlockMap) -> {Dominators,Numbering}.
 %%  Calculate the dominator tree, returning a map where each entry
 %%  in the map is a list that gives the path from that block to
 %%  the top of the dominator tree. (Note that the suffixes of the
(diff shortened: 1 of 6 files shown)


lib/compiler/src/beam_ssa.erl
```diff
@@ -404,26 +404,24 @@ common_dominators(Ls, Dom, Numbering) ->
     Doms = [map_get(L, Dom) || L <- Ls],
     dom_intersection(Doms, Numbering).
 
--spec fold_instrs_rpo(Fun, From, Acc0, Blocks) -> any() when
+-spec fold_instrs_rpo(Fun, Labels, Acc0, Blocks) -> any() when
```
bjorng, line 407: This name no longer makes sense. The blocks will be visited in the order of the labels in `Labels`. I suggest you rename it to `fold_instrs`. Also, while at it, please add comment in the style of the other comments in this module.

lib/compiler/src/beam_ssa.erl
```diff
@@ -404,26 +404,24 @@ common_dominators(Ls, Dom, Numbering) ->
...
       Fun :: fun((b_blk()|terminator(), any()) -> any()),
-      From :: [label()],
+      Labels :: [label()],
       Acc0 :: any(),
       Blocks :: block_map().
 
-fold_instrs_rpo(Fun, From, Acc0, Blocks) ->
-    Top = rpo(From, Blocks),
+fold_instrs_rpo(Fun, Top, Acc0, Blocks) ->
```
bjorng, line 413: Since the order is arbitrary, `Labels` would be a better name than `Top`.

lib/compiler/src/beam_ssa.erl
```diff
@@ -404,26 +404,24 @@ common_dominators(Ls, Dom, Numbering) ->
...
     fold_instrs_rpo_1(Top, Fun, Blocks, Acc0).
 
 %% Like mapfold_instrs_rpo but at the block level to support lookahead and
 %% scope-dependent transformations.
--spec mapfold_blocks_rpo(Fun, From, Acc, Blocks) -> Result when
+-spec mapfold_blocks_rpo(Fun, Labels, Acc, Blocks) -> Result when
```
bjorng, line 418: I suggest you rename this function to `mapfold_blocks`. Also rewrite the comment before to a comment in the proper style.

lib/compiler/src/beam_ssa.erl
```diff
@@ -404,26 +404,24 @@ common_dominators(Ls, Dom, Numbering) ->
...
       Fun :: fun((label(), b_blk(), any()) -> {b_blk(), any()}),
-      From :: [label()],
+      Labels :: [label()],
       Acc :: any(),
       Blocks :: block_map(),
       Result :: {block_map(), any()}.
-mapfold_blocks_rpo(Fun, From, Acc, Blocks) ->
-    Successors = rpo(From, Blocks),
+mapfold_blocks_rpo(Fun, Successors, Acc, Blocks) ->
```
bjorng, line 424: Rename `Successors` to `Labels`.

lib/compiler/src/beam_ssa.erl
```diff
@@ -434,84 +432,41 @@ mapfold_blocks_rpo_1(Fun, Lbl, {Blocks0, Acc0}) ->
...
       Acc :: any(),
       Blocks0 :: block_map(),
       Blocks :: block_map().
 
-mapfold_instrs_rpo(Fun, From, Acc0, Blocks) ->
-    Top = rpo(From, Blocks),
+mapfold_instrs_rpo(Fun, Top, Acc0, Blocks) ->
     mapfold_instrs_rpo_1(Top, Fun, Blocks, Acc0).
 
--spec flatmapfold_instrs_rpo(Fun, From, Acc0, Blocks0) -> {Blocks,Acc} when
```
bjorng, line 449: Same thing: delete `_rpo` prefix, proper comment, called the argument `Labels`

lib/compiler/src/beam_ssa.erl
```diff
@@ -434,84 +432,41 @@ mapfold_blocks_rpo_1(Fun, Lbl, {Blocks0, Acc0}) ->
...
-    fold_rpo(Fun, [0], Acc0, Blocks).
-
 %% fold_rpo(Fun, [Label], Acc0, Blocks) -> Acc.  Fold over all blocks
-%%  reachable from a given set of labels in a reverse postorder
-%%  traversal of the block graph; that is, first visit a block, then
-%%  visit its successors.
+%%  from a given set of labels in a reverse postorder traversal of the
+%%  block graph; that is, first visit a block, then visit its successors.
 
 -spec fold_rpo(Fun, Labels, Acc0, Blocks) -> any() when
```
bjorng, line 463: I suggest you call it `fold_blocks`.

lib/compiler/src/beam_ssa.erl
```diff
@@ -646,21 +601,18 @@ used(_) -> [].
...
                        (_Terminator, Acc) ->
                             Acc
-                    end, [0], #{}, Blocks).
-
--spec uses(Blocks :: block_map()) -> usage_map().
-uses(Blocks) ->
-    uses([0], Blocks).
+                    end, Top, #{}, Blocks).
 
--spec uses(From, Blocks) -> usage_map() when
```
bjorng, line 659: Please add a comment.

lib/compiler/src/beam_ssa.erl
```diff
@@ -646,21 +601,18 @@ used(_) -> [].
...
-      From :: [label()],
+-spec uses(Labels, Blocks) -> usage_map() when
+      Labels :: [label()],
       Blocks :: block_map().
-uses(From, Blocks) ->
-    fold_rpo(fun fold_uses_block/3, From, #{}, Blocks).
+uses(Top, Blocks) ->
```
bjorng, line 614: Rename `Top` to `Labels`.

bjorng (changes requested):
When I started writing the SSA passes I used Top as an abbreviation of Topological. This terminology is not correct when talking about graphs, so in newer code I tend to use RPO instead. Please rename `Top0' and `Top` to RPO. (Also, the convention in the compiler is to number only variables that will later be updated, such as `Block0` which is frequently updated.)

It could be nice to add some information to the commit message about how much these changes sped up the compiler on some huge modules.

While at, you might want to a few typos in the comment message:

Expliticly => Explicitly
inneficient => inefficient

Otherwise looks good to me.

josevalim:
Pushed. I should have addressed all feedback. Tests still pass.

bjorng (approved):

bjorng:
Added to our daily builds.

PR #2840, opened 2020-10-30
jit: Support systems that enforce W^X (OpenBSD, SELinux)

jhogberg:
This PR makes the JIT work on systems that don't allow writable pages to become executable, like OpenBSD and SELinux (without `allow_execmem`).


tsloughter (approved):

erts/emulator/beam/jit/beam_asm.cpp
```diff
@@ -147,7 +149,16 @@ void beamasm_init() {
 
     cpuinfo = CpuInfo::host();
 
-    bga = new BeamGlobalAssembler();
+    JitAllocator::CreateParams params;
+
+    params.options = JitAllocator::kOptionUseDualMapping |
+                     JitAllocator::kOptionUseMultiplePools |
+                     JitAllocator::kOptionFillUnusedMemory;
+    params.blockSize = 4 << 20;
```
garazdawi, line 157: I understand why you use DualMapping, but maybe add a comment about the other options and the blockSize?

PR #2842, opened 2020-10-31
erl_term.h: correct size of ExternalThing

max-au:
Fixes regression introduced by ccfc6b3. offsetof returns value in bytes, while EXTERNAL_THING_HEAD_SIZE is expected to be in words.

### DIFF
--- a/erts/emulator/beam/erl_term.h
+++ b/erts/emulator/beam/erl_term.h
@@ -1037,7 +1037,7 @@ typedef struct external_thing_ {
     } data;
 } ExternalThing;
 
-#define EXTERNAL_THING_HEAD_SIZE (offsetof(ExternalThing,data))
+#define EXTERNAL_THING_HEAD_SIZE (offsetof(ExternalThing,data) / sizeof(Uint))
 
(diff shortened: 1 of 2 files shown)


sverker:
@max-au Good catch! That would have been 168 bytes wasted per external pid/port/ref in OTP-24.

We don't have any asserts catching that kind of mistake apparently.

Thank you.

sverker:
@max-au Just curios, did you troubleshoot excessive heap usage or did you just happen to eyeball that code?

max-au:
I noticed a significant bump in 'pg' ETS table memory usage (3x), and also pg heap size, while amount of elements in the table/state remained constant. We have huge amount of processes published in pg, so total RAM hike was clearly visible.
It took me some time to find `erts_debug:flat_size` and discover that OTP 23 has external pid size of 4 words. Initially I though ot 64 bit pids but after reading the code realised it must've been something else.

sverker:
@max-au Pushed a new commit with some more white box testing that will detect bug like this.

max-au:
Looks even better, thank you!

PR #2847, opened 2020-11-03
Explicit pass RPO to beam_ssa functions (cont)

josevalim:
This patch finishes the work started on 0ce3fb91
by explicitly passing RPO to the remaining functions
in beam_ssa. This commit avoids the double RPO
computation in 4 other occasions.

### DIFF
--- a/lib/compiler/src/beam_ssa_bsm.erl
+++ b/lib/compiler/src/beam_ssa_bsm.erl
@@ -459,12 +459,14 @@ combine_matches(#b_function{bs=Blocks0,cnt=Counter0}=F, ModInfo) ->
                           {Block0#b_blk{is=Is}, State}
                   end,
                   RPO,
-                  #cm{ definitions = beam_ssa:definitions(Blocks0),
+                  #cm{ definitions = beam_ssa:definitions(RPO, Blocks0),
                        dominators = Dominators,
                        blocks = Blocks0 },
                   Blocks0),
 
-            Blocks2 = beam_ssa:rename_vars(State#cm.renames, [0], Blocks1),
+            %% The fun in mapfold_blocks does not update terminators,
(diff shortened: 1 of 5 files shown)


lib/compiler/src/beam_ssa_bsm.erl
```diff
@@ -450,21 +450,22 @@ combine_matches({Fs0, ModInfo}) ->
...
-                  RPO,
-                  #cm{ definitions = beam_ssa:definitions(Blocks0),
+                  RPO0,
+                  #cm{ definitions = beam_ssa:definitions(RPO0, Blocks0),
                        dominators = Dominators,
                        blocks = Blocks0 },
                   Blocks0),
 
-            Blocks2 = beam_ssa:rename_vars(State#cm.renames, [0], Blocks1),
+            RPO1 = beam_ssa:rpo(Blocks1),
```
bjorng, line 467: You can reuse the previous RPO here, because RPO0 and RPO1 are guaranteed to be identical (the fun passed to `beam_ssa:mapfold_blocks/4` does not update terminators).

bjorng (changes requested):

lib/compiler/src/beam_ssa_bsm.erl
josevalim: Fantastic, I will update this and leave a comment explaining why. :)

josevalim: Done!

bjorng (approved):

bjorng:
Added to our daily builds.

bjorng:
Thanks for this pull request.

josevalim:
Btw, with all 4 PRs combined, the Unicode module in Elixir compiles 18% faster on machine, which is pretty neat! :D

PR #2848, opened 2020-11-03
Reset inet_db timer even if file is up to date

josevalim:
Prior to this patch, if 5 seconds passed and
the resolv_conf or hosts_file were up to date,
we would continue to check for the file update
on every inet_res or inet_hosts operation.

This PR makes sure that the timer is updated
even if the file is up to date, so we check
again only in the next 5 seconds.

This addresses a performance regression on
inet_res which would lookup every 5 seconds
on Erlang/OTP 22 but performs it on every
request after 5 seconds on Erlang/OTP 23.

### DIFF
--- a/lib/kernel/src/inet_db.erl
+++ b/lib/kernel/src/inet_db.erl
@@ -1214,6 +1214,8 @@ handle_set_file(ParseFun, File, Bin, From, State) ->
 
 handle_update_file(
   Finfo, File, TagTm, TagInfo, ParseFun, From, #state{db = Db} = State) ->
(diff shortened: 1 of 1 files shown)


RaimoNiskanen:
I think the code worked as intended before this PR.  The assumption was that a read_file_info would be so lightweight it could be done for every DNS query.  Apparently not, so the delay in refreshing the file content that this PR adds is probably worth the prize for fixing the performance issues with the previously intended behaviour...

RaimoNiskanen:

PR #2849, opened 2020-11-03
Implement EEP-54 (extended error information)

bjorng:
This commit is the reference implementation for EEP-54, which proposes a mechanism
for providing more informational messages than simply "bad argument" when a call
to a BIF fails. Here is an example:

    1> element(a, b).
         in function  element/2
            called as element(a,b)
            *** argument 1: not an integer
            *** argument 2: not a tuple

All the details are found in EEP-54, but here are the main points:

* An extension of the format of the call-stack back trace (**stacktrace**) format
  to indicate that there exists extended error information for that call, and
  a convention for how extended error information can be provided.

* A new `erlang:error/3` BIF to allow libraries and applications to
  raise an exception with extended error information.

* New functions `erl_error:format_exception/3` and
  `erl_error:format_exception/4` to allow libraries and application to
  format stacktraces in the same style as the shell.

This implementation of EEP-54 implements extended error information for all
C-implemented BIFs in the `erlang` and `ets` modules.


josevalim:
Link to EEP 54: http://erlang.org/eeps/eep-0054.html :)

erts/preloaded/src/erlang.erl
```diff
@@ -821,6 +833,16 @@ error(_Reason) ->
 error(_Reason, _Args) ->
     erlang:nif_error(undefined).
 
+%% error/3
+%% Shadowed by erl_bif_types: erlang:error/3
+-spec error(Reason, Args, Options) -> no_return() when
+      Reason :: term(),
+      Args :: [term()],
+      Options :: [Option],
+      Option :: {'error_info', #{'reason' := term()}}.
```
josevalim, line 842: My Dialyzer is not great but should this be:

```suggestion
      Option :: {'error_info', #{'reason' := term(), any() := any()}}.
```

?

Also, is there a reason (ha) to require a `reason` key? The `format_error/4` written so far all do `maps:get(reason, Map, none)`, which implies reason is not being strictly required.

erts/doc/src/erlang.xml
```diff
@@ -1734,6 +1734,27 @@ example_fun(A1, A2) ->
...
+    <func>
+      <name name="error" arity="3" since="24.0"/>
+      <fsummary>Raise an exception of class <c>error</c></fsummary>
+      <desc>
+        <p>Raises an exception of class <c>error</c> with the reason
+        <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+        is any term. <c><anno>Args</anno></c> is expected to be the
+        list of arguments for the current function; it is used
+        to provide the arguments for the current function in
+        the stack back-trace.  If the <c>{error_info, Term}</c> option
```
josevalim, line 1746: Should it be `map()` instead of `Term`?

```suggestion
        the stack back-trace.  If the <c>{error_info, map()}</c> option
```

erts/preloaded/src/erlang.erl
bjorng: > Also, is there a reason (ha) to require a `reason` key?

Not really. I will talk with the team and perhaps change it.

Changing the spec in the way you suggested would allow a library to pass arbitrary keys and values in the map that its own `format_error/4` function would understand. That could be nice, but I am not sure that is a feature that is needed. 

Our reason for using a map was that we could add additional keys to the map that `erl_error` would know about (updating the spec at the same time).

josevalim: I see: it only allows the `reason` and nothing more. It makes sense to me now, thanks.

zuiderkwast:
Very nice!

A comment on the EEP:
> The map ErrorInfoMap contains further information about the error. This should contain at least the key reason. The corresponding value could contain additional information about the error. When there is no additional information avaiable, the value is a dummy value such as none or [].

Wouldn't it be cleaner to leave out the `reason` key instead of mapping it to a dummy value?

Also, is there a risk of confusing that two things now are called "reason"?

bjorng:
> Wouldn't it be cleaner to leave out the reason key instead of mapping it to a dummy value?

Yes, we have just discussed that in the VM team and decided to leave out the `reason` key if there is no useful value.

> Also, is there a risk of confusing that two things now are called "reason"?

Yes, there is. Perhaps it would be better to call the key `cause` instead?

josevalim:
What about `detail` or `info` for the key? They are more vague and don't add meaning to the value.

zuiderkwast:
I think `cause` is better than `reason`. It still shows the intent of the typical usage, that a formatting function would normally need the cause to be able to explain why an error happened, if I understand correctly. If it's optional anyway, a module can choose to use other keys anyway.

ferd:
I'd just use `reason` because every bit of erlang code and tutorial I can think of uses `{error, Reason}`.
In the OTP codebase for example:

```shell
→ ack -r '{error, ?Reason}' | wc -l   # literally the 'Reason' variable
9536
→ ack -r '{error, ?[A-Z][^},]+}' | wc -l # any variable (excluding underscored ones)
14235
```
In cases where the error is labelled  and used, 67% of the variables in current OTP code is using `Reason`

josevalim:
@ferd the confusion though is exactly because we already use it in other places. The first argument of `erlang:error/3` is already `Reason`, so you end-up with `erlang:error(Reason, Args, #{reason := ...})`, right?

zuiderkwast:
> I'd just use `reason` because every bit of erlang code and tutorial I can think of uses `{error, Reason}`.

Yeah, but here it's not `{error, Reason}` but rather `erlang:error(Reason, Args, [{error_info, #{reason => ExtraReason}}])`.

josevalim:
@zuiderkwast it was a mistake from me but I think they require a proplist in case they want to add more entries in the future.

ferd:
In that case I would go with a broader `context` but that's also overloaded in a lot of places

zuiderkwast:
The whole map `ErrorInfoMap` is a piece of info or a context, so these names are a bit meaningless. IMHO, if any keys should be recommended they better be meaningful. Otherwise, it would be better to not recommend any particular key and let each module pick its own keys. The map is handled by the formatting function in the same module anyway.

bjorng:
> The EEP says Options should be [{error_info,ErrorInfoMap}] but I think only the map would be cleaner. Any reason for the longer variant?

Yes, to allow the `Options` to be extended with other options in the future if needed.

> The whole map ErrorInfoMap is a piece of info or a context, so these names are a bit meaningless. IMHO, if any keys should be recommended they better be meaningful.

That was how we was reasoning, too. We want the keys in the map to be clearly defined in case we would want `erl_error` to use some of of them in the future (e.g. `module` and `function` to name a `format_error/4` function). If we would allow libraries outside of OTP to use any keys in the map it would be harder to make those kind of extensions without compatibility issues.

lib/kernel/src/erl_erlang_errors.erl
```diff
@@ -0,0 +1,811 @@
...
+
+-module(erl_erlang_errors).
+-export([format_error/4]).
+
+-spec format_error(Function, Args, ExceptionReason, ErrorInfoMap) -> ErrorMap when
+      Function :: atom(),
+      Args :: [term()],
+      ExceptionReason :: term(),
+      ErrorInfoMap :: #{},
+      ErrorMap :: #{pos_integer() => unicode:chardata()}.
```
essen, line 29: `pos_integer()` might not be enough in general. Take the example of `ssl:connect/3`. It has a couple arguments followed by `Opts`. There may be tons of options. Being limited strictly to argument position and text would potentially leave you with a wall of text full of errors.

Could the error map also allow something like `#{{pos_integer(), atom()} => <<"unknown protocol nottls">>}` for these cases? Then it can be formatted as `argument 3 (protocol): unknown protocol nottls`.

The intent is about separating independent errors, although a library could also use this to describe what the argument is beyond its position.

bjorng: I don't see anything in `ssl:connect/3` documentation that it would raise exceptions. This EEP is about exceptions, not errors returned from a function.

essen: I only used `ssl:connect/3` as an example of a function with lots of options, not necessarily as one that immediately applies to this. Functions like `term_to_binary` for example will only tell you `"invalid option in list"`. That's fine because the options list is small. But if a user library has many options, similar to what `ssl:connect/3` has, and triggers an exception, then knowing the exact location would be helpful.

The user's `format_error` function could very well put the exact location directly in the message, but the message might grow large if there are many different errors in the options. Having something like `{pos_integer(), atom()}` or even `{pos_integer(), any()}` allowed as key would instead result in multiple errors for the same argument which is more readable.

We could also argue that functions such as `ssl:connect/3`, like many other functions I've written, should not return an error tuple when the arguments are wrong, but that they only do so because the `error:badarg` exception was not helpful before this PR. This PR will most definitely impact the way we report errors to the user. I'd love to change functions like `gun:open` to trigger an exception if only I can properly inform the user of where the problem is, and that function has multiple levels of options. So output would ideally look something like:

```
argument 3 (http2_opts.bad): unknown option
argument 3 (keepalive): must be a timeout()
```

And it would also need to apply both to the shell and to logger reports.

Anyway this is not something that can't be improved later on so the stakes are low to get this right immediately.

bjorng: > We could also argue that functions such as `ssl:connect/3`, like many other functions I've written, should not return an error tuple when the arguments are wrong, but that they only do so because the error:badarg exception was not helpful before this PR.

A function that raises an exception does not have to raise a `badarg` exception. It could just as easily raise a `{badoption,DetailedReason}` exception. We are stuck with `badarg` for BIFs for backward compatibility reasons.

> Anyway this is not something that can't be improved later on so the stakes are low to get this right immediately.

Yes, that's why we have the maps so that this mechanism can be improved as needed in the future.

bjorng:
I have force-pushed an update to the EEP that renames `reason` to `cause` and makes it optional.

I have also created a PR that updates the EEP: erlang/eep#15

lib/kernel/src/erl_erlang_errors.erl
garazdawi: A place where this may be useful for us would be for `badopt` in `erlang:spawn_request`.

lib/stdlib/doc/src/erl_error.xml
```diff
@@ -0,0 +1,96 @@
...
+  <datatypes>
+    <datatype>
+      <name name="format_options"/>
+      <desc><p>A map with formatting options.</p></desc>
+    </datatype>
+    <datatype>
+      <name name="stack_trim_fun"/>
+      <desc><p>A fun used to trim the end of the stacktrace.  It is
+      called with module, function, and arity from an entry from the
+      stacktrace.  The fun is to return <c>true</c> if the entry
```
paulo-ferraz-oliveira, line 54: is to return > returns?

bjorng: We have an internal style document, and according to that we should avoid certain words such as "should" and "shall" because they can be ambiguous. Instead we are to write "is to..." or "are to...".

lib/stdlib/doc/src/erl_error.xml
```diff
@@ -0,0 +1,96 @@
...
+    </datatype>
+  </datatypes>
+
+  <funcs>
+    <func>
+      <name name="format_exception" arity="3" since="OTP 24.0"/>
+      <name name="format_exception" arity="4" since="OTP 24.0"/>
+      <fsummary>Format an exception.</fsummary>
+      <desc>
+        <p>Format the error reason and stack back-trace catched using
```
paulo-ferraz-oliveira, line 81: catched > caught? (catched is acceptable, but less modern)

lib/stdlib/src/ets.erl
```diff
@@ -1824,3 +1826,345 @@ re_match([H|T], Re) ->
...
+expand_error(bad_matchspec) ->
+    <<"not a valid match specification">>;
+expand_error(bad_options) ->
+    <<"invalid options">>;
+expand_error(bad_table_name) ->
+    <<"invalid table name (must be an atom)">>;
+expand_error(bad_update_op) ->
+    <<"not a valid update operation">>;
+expand_error(counter_not_integer) ->
+    <<"the value in the given position in the object is not an integer">>;
```
paulo-ferraz-oliveira, line 2146: in the object > , in the object, ?
(add commas)

system/doc/reference_manual/errors.xml
```diff
@@ -109,26 +109,29 @@
...
           tuples. The field <c>Arity</c> in the first tuple can be the
           argument list of that function call instead of an arity integer,
           depending on the exception.</p>
-
-	<p><c>Location</c> is a (possibly empty) list of two-tuples
-	that can indicate the location in the source code of the
-	function.  The first element is an atom describing the type of
-	information in the second element. The following items can
-	occur:</p>
+	  <p><c>ExtraInfo</c> is a (possibly empty) list of two-tuples
```
paulo-ferraz-oliveira, line 124: Does this two- refer to two elements?
If so would "two-element tuples" convey a better message?

bjorng:
I have pushed an update to implement the changes requested by the OTP Technical Board. See https://github.com/erlang/eep/pull/16.

I have also moved out the `format_error/2` functions to `erl_erts_errors`, `erl_kernel_errors`, and `erl_stdlib_errors` to handle all errors for the ERTS, Kernel, and STDLIB applications, respectively.

garazdawi:
Yay! I'm really looking forward to this!

filmor:
Is an extension of [enif_raise_exception](http://erlang.org/doc/man/erl_nif.html#enif_raise_exception) planned to let arbitrary NIFs take advantage of this feature?

PR #2850, opened 2020-11-03
Add queue:all/2, queue:any/2, queue:delete/2, and queue:delete_r/2

Maria-12648430:
This PR is a follow-up to #2791 which adds 4 new functions to the `queue` module. The new functions perform the same tasks as their counterparts in the `lists` module, but avoid the need to convert queues to lists first, which is both inconvenient and inefficient.

* `queue:all/2`: like `lists:all/2`, checks if all items in a queue satisfy a predicate; processed in queue order to retain the order of side effects of the predicate function
* `any/2`: like `lists:any/2`, checks if at least one item in a queue satisfies a predicate; procesed in queue order to retain the order of side effects of the predicate function
* `delete/2`: like `lists:delete/2`, removes the _first_ occurence of an item from a queue, according to queue order
* `delete_r/2`: there is no counterpart for this in the `lists` module, but I believe this function is still handy in a queue context; it removes the _last_ occurence of an item from a queue, according to queue order; the naming `..._r` has been chosen to follow the naming convention of other "from the rear" functions of the module, namely `in_r` and `out_r`

In #2791, @okeuday requested another function that removes the first/last item from a queue matching a predicate. Suggested naming was `remove/2`, IIRC. While this may be handy and could be added to this PR, I'm not sure what the consensus was, as such a function is not present in the `lists` module. What do you think?

### DIFF
--- a/lib/stdlib/src/queue.erl
+++ b/lib/stdlib/src/queue.erl
@@ -27,7 +27,7 @@
 -export([get/1,get_r/1,peek/1,peek_r/1,drop/1,drop_r/1]).
 
 %% Higher level API
--export([reverse/1,join/2,split/2,filter/2,filtermap/2,fold/3]).
+-export([reverse/1,join/2,split/2,filter/2,filtermap/2,fold/3,any/2,all/2]).
 
 %% Okasaki API from klacke
 -export([cons/2,head/1,tail/1,
@@ -452,6 +452,28 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
     lists:foldr(Fun, Acc1, R);
 fold(Fun, Acc0, Q) ->
     erlang:error(badarg, [Fun, Acc0, Q]).
+
+%% Check if any item satisfies the predicate, traverse in queue order.
+%%
+%% O(len(Q)) worst case
+-spec any(Pred, Q :: queue(Item)) -> boolean() when
+      Pred :: fun((Item) -> boolean()).
+any(Pred, {R, F}) when is_function(Pred, 1), is_list(R), is_list(F) ->
+    lists:any(Pred, F) orelse
+    lists:any(Pred, R);
+any(Pred, Q) ->
+    erlang:error(badarg, [Pred, Q]).
+
+%% Check if all items satisfy the predicate, traverse in queue order.
+%%
+%% O(len(Q)) worst case
+-spec all(Pred, Q :: queue(Item)) -> boolean() when
+      Pred :: fun((Item) -> boolean()).
+all(Pred, {R, F}) when is_function(Pred, 1), is_list(R), is_list(F) ->
+    lists:all(Pred, F) andalso
+    lists:all(Pred, R);
+all(Pred, Q) ->
+    erlang:error(badarg, [Pred, Q]).
 
 %%--------------------------------------------------------------------------
 %% Okasaki API inspired by an Erlang user contribution "deque.erl" 
--- a/lib/stdlib/src/queue.erl
+++ b/lib/stdlib/src/queue.erl
@@ -27,7 +27,8 @@
 -export([get/1,get_r/1,peek/1,peek_r/1,drop/1,drop_r/1]).
 
 %% Higher level API
--export([reverse/1,join/2,split/2,filter/2,filtermap/2,fold/3,any/2,all/2]).
+-export([reverse/1,join/2,split/2,filter/2,filtermap/2,fold/3,any/2,all/2,
+	 delete/2,delete_r/2]).
 
 %% Okasaki API from klacke
 -export([cons/2,head/1,tail/1,
@@ -474,6 +475,71 @@ all(Pred, {R, F}) when is_function(Pred, 1), is_list(R), is_list(F) ->
     lists:all(Pred, R);
 all(Pred, Q) ->
     erlang:error(badarg, [Pred, Q]).
+
+%% Delete the first occurence of an item in the queue,
+%% according to queue order.
+%%
+%% O(len(Q1)) worst case
+-spec delete(Item, Q1) -> Q2 when
+      Item :: T,
+      Q1 :: queue(T),
+      Q2 :: queue(T),
+      T :: term().
+delete(Item, {R0, F0} = Q) when is_list(R0), is_list(F0) ->
+    case delete_front(Item, F0) of
+        false ->
+            case delete_rear(Item, R0) of
+                false ->
+                    Q;
+                [] ->
+                    f2r(F0);
+                R1 ->
+                    {R1, F0}
+            end;
+        [] ->
+            r2f(R0);
+        F1 ->
+            {R0, F1}
+    end;
+delete(Item, Q) ->
+    erlang:error(badarg, [Item, Q]).
+
+%% Delete the last occurence of an item in the queue,
+%% according to queue order.
+%%
+%% O(len(Q1)) worst case
+-spec delete_r(Item, Q1) -> Q2 when
+      Item :: T,
+      Q1 :: queue(T),
+      Q2 :: queue(T),
+      T :: term().
+delete_r(Item, {R0, F0}) when is_list(R0), is_list(F0) ->
+    {F1, R1}=delete(Item, {F0, R0}),
+    {R1, F1};
+delete_r(Item, Q) ->
+    erlang:error(badarg, [Item, Q]).
+
+delete_front(Item, [Item|Rest]) ->
+    Rest;
+delete_front(Item, [X|Rest]) ->
+    case delete_front(Item, Rest) of
+        false -> false;
+        F -> [X|F]
+    end;
+delete_front(_, []) ->
+    false.
+
+delete_rear(Item, [X|Rest]) ->
+    case delete_rear(Item, Rest) of
+        false when X=:=Item ->
+            Rest;
+        false ->
+            false;
+        R ->
+            [X|R]
+    end;
+delete_rear(_, []) ->
+    false.
 
 %%--------------------------------------------------------------------------
 %% Okasaki API inspired by an Erlang user contribution "deque.erl" 
--- a/lib/stdlib/src/queue.erl
+++ b/lib/stdlib/src/queue.erl
@@ -28,7 +28,7 @@
 
 %% Higher level API
 -export([reverse/1,join/2,split/2,filter/2,filtermap/2,fold/3,any/2,all/2,
-	 delete/2,delete_r/2]).
+	 delete/2,delete_r/2,delete_with/2,delete_with_r/2]).
 
 %% Okasaki API from klacke
 -export([cons/2,head/1,tail/1,
@@ -539,6 +539,79 @@ delete_rear(Item, [X|Rest]) ->
             [X|R]
     end;
 delete_rear(_, []) ->
+    false.
+
+%% Delete the first occurence of an item in the queue
+%% matching a predicate, according to queue order.
+%%
+%% O(len(Q1)) worst case
+-spec delete_with(Pred, Q1) -> Q2 when
+      Pred :: fun((Item) -> boolean()),
+      Q1 :: queue(Item),
+      Q2 :: queue(Item),
+      Item :: term().
+delete_with(Pred, {R0, F0} = Q) when is_function(Pred, 1), is_list(R0), is_list(F0) ->
+    case delete_with_front(Pred, F0) of
+	false ->
+	    case delete_with_rear(Pred, R0) of
+		false ->
+		    Q;
+		[] ->
+		    f2r(F0);
+		R1 ->
+		    {R1, F0}
+	    end;
+	[] ->
+	    r2f(R0);
+	F1 ->
+	    {R0, F1}
+    end;
+delete_with(Pred, Q) ->
+    erlang:error(badarg, [Pred, Q]).
+
+%% Delete the last occurence of an item in the queue
+%% matching a predicate, according to queue order.
+%%
+%% O(len(Q1)) worst case
+-spec delete_with_r(Pred, Q1) -> Q2 when
+      Pred :: fun((Item) -> boolean()),
+      Q1 :: queue(Item),
+      Q2 :: queue(Item),
+      Item :: term().
+delete_with_r(Pred, {R0, F0}) when is_function(Pred, 1), is_list(R0), is_list(F0) ->
+    {F1, R1} = delete_with(Pred, {F0, R0}),
+    {R1, F1};
+delete_with_r(Pred, Q) ->
+    erlang:error(badarg, [Pred, Q]).
+
+delete_with_front(Pred, [X|Rest]) ->
+    case Pred(X) of
+	true ->
+	    Rest;
+	false ->
+	    case delete_with_front(Pred, Rest) of
+		false ->
+		    false;
+		F ->
+		    [X|F]
+	    end
+    end;
+delete_with_front(_, []) ->
+    false.
+
+delete_with_rear(Pred, [X|Rest]) ->
+    case delete_with_rear(Pred, Rest) of
+	false ->
+	    case Pred(X) of
+		true ->
+		    Rest;
+		false ->
+		    false
+	    end;
+	R ->
+	    [X|R]
+    end;
+delete_with_rear(_, []) ->
     false.
 
 %%--------------------------------------------------------------------------
--- a/lib/stdlib/doc/src/queue.xml
+++ b/lib/stdlib/doc/src/queue.xml
@@ -53,6 +53,8 @@
       assuming knowledge of the format is running on thin ice.</p>
 
     <p>All operations have an amortized O(1) running time, except
+      <seemfa marker="#all/2"><c>all/2</c></seemfa>,
+      <seemfa marker="#any/2"><c>any/2</c></seemfa>,
       <seemfa marker="#filter/2"><c>filter/2</c></seemfa>,
       <seemfa marker="#filtermap/2"><c>filtermap/2</c></seemfa>,
       <seemfa marker="#fold/3"><c>fold/3</c></seemfa>,
@@ -115,6 +117,28 @@
     <fsdescription>
       <title>Original API</title>
     </fsdescription>
+    <func>
+      <name name="all" arity="2" since=""/>
+      <fsummary>Return <c>true</c> if all items in a queue satisfy
+        <c>Pred</c>.</fsummary>
+      <desc>
+        <p>Returns <c>true</c> if <c><anno>Pred</anno>(<anno>Item</anno>)</c>
+          returns <c>true</c> for all items <c><anno>Item</anno></c> in
+          <c><anno>Q</anno></c>, otherwise <c>false</c>.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="any" arity="2" since=""/>
+      <fsummary>Return <c>true</c> if any of the items in a queue satisfy
+        <c>Pred</c>.</fsummary>
+      <desc>
+        <p>Returns <c>true</c> if <c><anno>Pred</anno>(<anno>Item</anno>)</c>
+          returns <c>true</c> for at least one item <c><anno>Item</anno></c>
+          in <c><anno>Q</anno></c>, otherwise <c>false</c>.</p>
+      </desc>
+    </func>
+
     <func>
       <name name="filter" arity="2" since=""/>
       <fsummary>Filter a queue.</fsummary>
--- a/lib/stdlib/doc/src/queue.xml
+++ b/lib/stdlib/doc/src/queue.xml
@@ -57,6 +57,8 @@
       <seemfa marker="#any/2"><c>any/2</c></seemfa>,
       <seemfa marker="#delete/2"><c>delete/2</c></seemfa>,
       <seemfa marker="#delete_r/2"><c>delete_r/2</c></seemfa>,
+      <seemfa marker="#delete_with/2"><c>delete_with/2</c></seemfa>,
+      <seemfa marker="#delete_with_r/2"><c>delete_with_r/2</c></seemfa>,
       <seemfa marker="#filter/2"><c>filter/2</c></seemfa>,
       <seemfa marker="#filtermap/2"><c>filtermap/2</c></seemfa>,
       <seemfa marker="#fold/3"><c>fold/3</c></seemfa>,
@@ -147,7 +149,7 @@
       <desc>
         <p>Returns a copy of <c><anno>Q1</anno></c> where the first item
           matching <c><anno>Item</anno></c> is deleted, if there is such an
-          element.</p>
+          item.</p>
       </desc>
     </func>
 
@@ -157,7 +159,29 @@
       <desc>
         <p>Returns a copy of <c><anno>Q1</anno></c> where the last item
           matching <c><anno>Item</anno></c> is deleted, if there is such an
-          element.</p>
+          item.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="delete_with" arity="2" since=""/>
+      <fsummary>Delete an item matching a predicate from the front of a
+        queue.</fsummary>
+      <desc>
+        <p>Returns a copy of <c><anno>Q1</anno></c> where the first item
+          for which <c><anno>Pred</anno></c> returns <c>true</c> is deleted,
+          if there is such an item.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="delete_with_r" arity="2" since=""/>
+      <fsummary>Delete an item matching a predicate from the rear of a
+        queue.</fsummary>
+      <desc>
+        <p>Returns a copy of <c><anno>Q1</anno></c> where the last item
+          for which <c><anno>Pred</anno></c> returns <c>true</c> is deleted,
+          if there is such an item.</p>
       </desc>
     </func>
 
--- a/lib/stdlib/doc/src/queue.xml
+++ b/lib/stdlib/doc/src/queue.xml
@@ -188,7 +188,7 @@
       <desc>
         <p>Returns a queue <c><anno>Q2</anno></c> that is the result of calling
           <c><anno>Fun</anno>(<anno>Item</anno>)</c> on all items in
-          <c><anno>Q1</anno></c>, in order from front to rear.</p>
+          <c><anno>Q1</anno></c>.</p>
         <p>If <c><anno>Fun</anno>(<anno>Item</anno>)</c> returns <c>true</c>,
           <c>Item</c> is copied to the result queue. If it returns <c>false</c>,
           <c><anno>Item</anno></c> is not copied. If it returns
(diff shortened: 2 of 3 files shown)


lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
+%% Delete the first occurence of an item in the queue,
+%% according to queue order.
+%%
+%% O(len(Q1)) worst case
+-spec delete(Item, Q1) -> Q2 when
+      Item :: T,
+      Q1 :: queue(T),
+      Q2 :: queue(T),
+      T :: term().
+delete(Item, {R0, F0}) when is_list(R0), is_list(F0) ->
```
RaimoNiskanen, line 488: `{R0, F0} = Q`

lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
+    case delete_front(Item, F0) of
+        false ->
+            case delete_rear(Item, R0) of
+                false ->
+                    {R0, F0};
```
RaimoNiskanen, line 493: `Q`;
Bound in the function head

lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
+      Q1 :: queue(T),
+      Q2 :: queue(T),
+      T :: term().
+delete_r(Item, {R0, F0}) when is_list(R0), is_list(F0) ->
+    {F1, R1}=delete(Item, {F0, R0}),
+    {R1, F1};
+delete_r(Item, Q) ->
+    erlang:error(badarg, [Item, Q]).
+
+delete_front(Item, [Item|Rest]) ->
```
RaimoNiskanen, line 522: Come to think of it; there should be a guard `when is_list(Rest)` here or else an improper list with `false` in the tail would confuse the code.

lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
+    Rest;
+delete_front(Item, [X|Rest]) ->
```
RaimoNiskanen, line 524: `when is_list(Rest)`

lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
+    case delete_front(Item, Rest) of
+        false -> false;
+        F -> [X|F]
+    end;
+delete_front(_, []) ->
+    false.
+
+delete_rear(Item, [X|Rest]) ->
```
RaimoNiskanen, line 532: `when is_list(Rest)`

lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
+    erlang:error(badarg, [Pred, Q]).
+
+%% Check if all items satisfy the predicate, traverse in queue order.
+%%
+%% O(len(Q)) worst case
+-spec all(Pred, Q :: queue(Item)) -> boolean() when
+      Pred :: fun((Item) -> boolean()).
+all(Pred, {R, F}) when is_function(Pred, 1), is_list(R), is_list(F) ->
+    lists:all(Pred, F) andalso
+    lists:all(Pred, lists:reverse(R));
```
RaimoNiskanen, line 475: `all_r(Pred, R);`

``` Erlang
all_r(Pred, [X|R]) when is_list(R) ->
    all_r(Pred, R) andalso Pred(X);
all_r(_Pred, []) ->
    true.
```
Not tested, nor compiled.  Does not build any garbage like `lists:reverse(R)`.

RaimoNiskanen:
Regarding `remove/2`: I do not remember if there was a consensus.  It solves a problem that the list module does not have.  But the question is if its small performance gain is worth implementing.  It will on average be twice as fast but with the same Ordo as `delete/2`, and since it removes from any end probably only usable when you know all elements are unique.

lib/stdlib/src/queue.erl
```diff
@@ -453,6 +454,93 @@ fold(Fun, Acc0, {R, F}) when is_function(Fun, 2), is_list(R), is_list(F) ->
...
     erlang:error(badarg, [Fun, Acc0, Q]).
 
+%% Check if any item satisfies the predicate, traverse in queue order.
+%%
+%% O(len(Q)) worst case
+-spec any(Pred, Q :: queue(Item)) -> boolean() when
+      Pred :: fun((Item) -> boolean()).
+any(Pred, {R, F}) when is_function(Pred, 1), is_list(R), is_list(F) ->
+    lists:any(Pred, F) orelse
+    lists:any(Pred, lists:reverse(R));
```
RaimoNiskanen, line 464: `any_r(Pred, R);`

``` Erlang
any_r(Pred, [X|R]) when is_list(R) ->
    any_r(Pred, R) orelse (not not Pred(X));
any_r(_Pred, []) ->
    false.
````
Not tested nor compiled. Does not build any garbage like `lists:reverse(R)`.

**Comment by okeuday** [comment]
@RaimoNiskanen `remove/2` could go in-order to avoid being limited to only the use-case with unique items.  The main reason a `remove/2` function should be necessary is to be able to go into items that are likely typically tuples.  I don't think usage of `delete/2` would be as common as `remove/2` because it seems typical to be queuing data for processing that is unable to be removed based on equality (typically due to an id or key being associated with additional data getting queued in a tuple).

File: lib/stdlib/src/queue.erl (hunk 3 above)
Comment (Maria-12648430) [reply lib/stdlib/src/queue.erl:522]: Hm, granted, and I will add guards in the suggested places... But actually, an improper list would only be possible if what is given to the exported functions `delete/2` and `delete_r/2` is not a real queue (as returned by `queue` API functions) but a queue-like tuple.

File: lib/stdlib/src/queue.erl (hunk 6 above)
Comment (Maria-12648430) [reply lib/stdlib/src/queue.erl:475]: Why the double negation `(not not Pred(X))`?

File: lib/stdlib/src/queue.erl (hunk 7 above)
Comment (Maria-12648430) [reply lib/stdlib/src/queue.erl:464]: Same question as for `all_r/2`: why the double negation?

File: lib/stdlib/src/queue.erl (hunk 6 above)
Comment (Maria-12648430) [reply lib/stdlib/src/queue.erl:475]: Please reconsider, I didn't use `lists:reverse/1` here out of sheer lazyness ;)

(But first, I have to admit that I don't know how `lists:reverse/1` is really implemented. In any case, it is _fast_.)

As I see it, the difference between the two approaches is this:
* My approach only traverses the (reversed) `R` list down to the point where `Pred` returns `false`, then stops. In the worst case, when `Pred` returns `true` for all (or all but the last) elements in `R`, that means a full traversal of `R`, all other cases stop earlier. So, this approach is `O(length(R))` in the worst case (plus a bit for the reversal, yes ;)).
* The suggested approach traverses `R` in full _always_, then unrolls the recursion again. With this, even the best case (`Pred` returns `false` for the last element in `R`) would equal the worst case in my approach, plus the unrolling of the _entire_ recursion. So, this approach is `O(2*length(R))` in _any_ case.

The same holds for `any` (and actually, for `delete_rear` too, which could be implemented by ways of `lists:reverse` + `lists:delete` + `lists:reverse`, but since it involves two reversals, I refrained from it there).

**Comment by RaimoNiskanen** [comment]
@okeuday: I thought the point of `remove(Pred, Q)` vs. `delete(Pred, Q)` was that `remove/2` would be faster since it traverses both lists the fast direction.  If `remove/2` would traverse the reverse list in the queue order would not then be the same as `delete/2`?  Or is the difference that it returns `{Removed :: boolean(), NewQ}`?

**Comment by Maria-12648430** [comment]
> @okeuday: I thought the point of `remove(Pred, Q)` vs. `delete(Pred, Q)` was that `remove/2` would be faster since it traverses both lists the fast direction. If `remove/2` would traverse the reverse list in the queue order would not then be the same as `delete/2`? Or is the difference that it returns `{Removed :: boolean(), NewQ}`?

As I understood it, `remove` is like `delete` in that it deletes the first (or last) item, but with a predicate instead of exact match... And as such, it should work in queue order like all the other queue functions that involve user functions, for the order of side effects.

File: lib/stdlib/src/queue.erl (hunk 3 above)
Comment (RaimoNiskanen) [reply lib/stdlib/src/queue.erl:522]: Quite right!  Ignore my suggestion.  Unnecessary safeguard.

Comment (Maria-12648430) [reply lib/stdlib/src/queue.erl:522]: Ok, I'll mark those as resolved then ;)

File: lib/stdlib/src/queue.erl (hunk 6 above)
Comment (RaimoNiskanen) [reply lib/stdlib/src/queue.erl:475]: > Why the double negation `(not not Pred(X))`?

Because for the last iteration, `Pred`, that is user supplied, might return something non-boolean, which would become the return value of the function.  `(not not)` fails for a non-booelan operand, while `andalso` | `orelse` ignores their right operand.  But it is redundant to have it on every iteration since the result is recursed up to the first operand of `andalso` | `orelse`, so the better code would be:
``` Erlang
    not not all_r(Pred, R);

all_r(Pred, [X|R]) ->
    all_r(Pred, R) andalso Pred(X);
all_r(_Pred, []) ->
    true.
```

jhogberg: > Please reconsider, I didn't use lists:reverse/1 here out of sheer lazyness ;)

The suggested approach doesn't create any garbage which is a significant win for processes with large heaps.

Why not leave the order undefined and just traverse it as-is? Honoring side-effects in _predicate functions_ feels rather strange and neither "all" nor "any" imply order per se.

RaimoNiskanen: > Please reconsider, I didn't use `lists:reverse/1` here out of sheer lazyness ;)

I am sorry that I suspected that :-)

`lists:reverse/1` is implemented in C and it is as you say _fast_, but it is still O(`length(R)`).  It builds a complete reversed list on the heap, so it traverses all elements of `R`.  Then your iteration goes tail recursively down to a bad `Pred(X)` value and stops at that.  So it builds the whole reverse `R` on the heap, which is one list cell per element.  This has to be garbage collected later.

My suggestion iterates body recursively down to the end of the list, which builds stack, one return address per iteration.  Then it applies `Pred(X)` to the list elements up to a bad value and skips all after.  This just builds stack during the iteration, and it should be one word (half a list cell) per iteration, so it should be the half amount of memory.  The stack is free immediately after the iteration is complete without garbage collection.

Now, garbage collection happens anyway, so the reversing approach probably just makes it happen earlier than the body recursion approach.  Then the garbage collection time is mostly depending on the amount of live data.  So if this function is called from a process with much live data it might get an expensive garbage collection earlier.  The body recursion approach on the other hand has to do one `andalso` | `orelse` test per iteration even after `Pred(X)` has returned a bad value, while the reversing approach can finish right there.  `lists:reverse/1` is probably much faster than an erlang loop testing a boolean for the same list length.

So which is fastest depends on stuff.  Both are O(`length(R)`) (since garbage collection is not O(ProducedGarbage)).  Note that both O(2*N) and O(0.5*N) are O(N).  We are just playing with constant factors here.  The reversing approach is probably faster (since `lists:reverse/1` has so low constant factor on O(N)), at least when called from a process with small heap (say benchmark, and the question is how large heap that is needed to be noticed).  The body recursion approach probably uses less maximum memory.

Either approach is good, and I guess you have benchmarked.  I just like body recursion, and enjoy minimizing garbage. ;-)

RaimoNiskanen:
@Maria-12648430 and @okeuday; I have been reading sloppy and thought `delete/2` used a predicate.  Then I agree with @okeuday that `remove(Pred, Q)` probably is even more useful than `delete(Item, Q)`, and that both would be useful.

Then `lists:remove(Pred, L)` is missing.  Bummer!  Annoying that `maps:remove(Key, Map)` takes a key to remove, and `lists:delete(Elem, List)` takes an element to match.  Impedance mismatch...

okeuday:
@RaimoNiskanen A `remove_unique(Pred, Q)` function could be added that doesn't go in-order, to provide that smaller speed benefit with more minimal modification of the queue lists.  However, `remove(Pred, Q)` is the main improvement to be more efficient than code currently that would be using `queue:filter(Pred, Q)` to remove a single item.  So, avoiding `queue:filter(Pred, Q)` use is the main goal with `remove(Pred, Q)`.

RaimoNiskanen:
@jhogberg asked a very good question kind of hidden in the buzz above so I repeat it here:
> Why not leave the order undefined and just traverse it as-is? Honoring side-effects in _predicate functions_ feels rather strange and neither "all" nor "any" imply order per se.

`lists:any/2` and `lists:all/2` does not guarantee anything about evaluation order.  It is just assumed that it will be traversed from the head and bail out as early as possible, but e.g guaranteeing to traverse all elements would also be useful semantics, if you want to rely on the side effects.  But for these functions I think you are happy with as fast as possible.  You can not know beforehand if all elements will be traversed anyway, and making use of knowing that all before the true predicate in queue order are traversed feels far-fetched...

lib/stdlib/src/queue.erl
Maria-12648430: > @jhogberg asked a very good question kind of hidden in the buzz above so I repeat it here:
> 
> > Why not leave the order undefined and just traverse it as-is? Honoring side-effects in _predicate functions_ feels rather strange and neither "all" nor "any" imply order per se.
> 
> `lists:any/2` and `lists:all/2` does not guarantee anything about evaluation order. It is just assumed that it will be traversed from the head and bail out as early as possible, but e.g guaranteeing to traverse all elements would also be useful semantics, if you want to rely on the side effects. But for these functions I think you are happy with as fast as possible. You can not know beforehand if all elements will be traversed anyway, and making use of knowing that all before the true predicate in queue order are traversed feels far-fetched...

I'm ok with that, but let me point out that `queue:filter/2` is explicitly designed to operate in queue order. As the filter function is a predicate (with the additional capability of inserting/replacing), by @jhogberg's argument, this is equally not required and could be faster if the rear list was also just traversed as-is.

The traversal in queue order is not mentioned in the user documentation (only in the code), but it sets a precedent that strengthens the intuitive assumtion that traversal will be done in queue order in all functions.

I guess what I'm trying to say is, if the queue module is to go against the intuitive assumption, it should be mentioned in the documentation. That is, only mention that `queue:any` and `queue:all` do _not_ obey queue order, not that `filter` _does_.

Maria-12648430:
> @jhogberg asked a very good question kind of hidden in the buzz above so I repeat it here:
> 
> > Why not leave the order undefined and just traverse it as-is? Honoring side-effects in _predicate functions_ feels rather strange and neither "all" nor "any" imply order per se.
> 
> `lists:any/2` and `lists:all/2` does not guarantee anything about evaluation order. It is just assumed that it will be traversed from the head and bail out as early as possible, but e.g guaranteeing to traverse all elements would also be useful semantics, if you want to rely on the side effects. But for these functions I think you are happy with as fast as possible. You can not know beforehand if all elements will be traversed anyway, and making use of knowing that all before the true predicate in queue order are traversed feels far-fetched...

I'm ok with that, but let me point out that `queue:filter/2` is explicitly designed to operate in queue order. As the filter function is a predicate (with the additional capability of inserting/replacing), by @jhogberg's argument, this is equally not required and could be faster if the rear list was also just traversed as-is.

The traversal in queue order is not mentioned in the user documentation (only in the code), but it sets a precedent that strengthens the intuitive assumtion that traversal will be done in queue order in all functions.

I guess what I'm trying to say is, if the queue module is to go against the intuitive assumption, it should be mentioned in the documentation. That is, only mention that `queue:any` and `queue:all` do _not_ obey queue order, not that `filter` _does_.

lib/stdlib/src/queue.erl
Maria-12648430: @RaimoNiskanen @jhogberg I will go with the approach that you suggested. While my approach _is_ faster than the suggested, it only matters in really extreme cases (some millions of items in a queue, likely a bad idea to begin with). Also, I cannot repeat my initial benchmarks that showed a significantly large difference, so I must have done something wrong back then (I didn't keep the setup, my bad :\)

jhogberg:
> I'm ok with that, but let me point out that queue:filter/2 is explicitly designed to operate in queue order. As the filter function is a predicate (with the additional capability of inserting/replacing), by @jhogberg's argument, this is equally not required and could be faster if the rear list was also just traversed as-is.

I think it would've been documented if it were designed with that in mind. From reading the code it looks slightly more efficient to traverse in queue order so I'd wager that's the reason, but I wasn't around back then.

> The traversal in queue order is not mentioned in the user documentation (only in the code), but it sets a precedent that strengthens the intuitive assumtion that traversal will be done in queue order in all functions.
>
> I guess what I'm trying to say is, if the queue module is to go against the intuitive assumption, it should be mentioned in the documentation. That is, only mention that queue:any and queue:all do not obey queue order, not that filter does.

The documentation is the public interface, not the code, and I often find myself regretting the times I've tied myself to a certain implementation (often by accident). Interfaces tend to ossify once users start making assumptions and I don't like encouraging that by implying that all other functions in the module follow queue order.

This is just my personal opinion, mind you, and I think having these functions is great. I'm just picking a nit :)

Maria-12648430:
> I think it would've been documented if it were designed with that in mind. From reading the code it looks slightly more efficient to traverse in queue order so I'd wager that's the reason, but I wasn't around back then.

Judging from the date in the documentation, 1997-01-15, this must be one of the oldest modules around =^^= `queue:filter/2` probably appeared in it later, but still 😅

> The documentation is the public interface, not the code, and I often find myself regretting the times I've tied myself to a certain implementation (often by accident). Interfaces tend to ossify once users start making assumptions

I have heard horror stories in that line from a friend, yes ;) But I have to admit that I have also used `lists:any/2` or `lists:all/2` as "`lists:foreach/2` with early stopping" here and there, shame on me XD

> and I don't like encouraging that by implying that all other functions in the module follow queue order.

Ah, right, it didn't occur to me that saying that `any` and `all` do _not_ follow queue order could be understood as that the other functions _do_.

> This is just my personal opinion, mind you, and I think having these functions is great. I'm just picking a nit :)

No, you're right :) And as @RaimoNiskanen seemed to agree, I will go change `queue:any/2` and `queue:all/2` as you suggested here.

RaimoNiskanen:
We'll take up the traversal order on a future OTP Technical Board.  But it so far most members in the OTP group seems to agree that the traversal order should be irrelevant for `any/2` and `all/2`, and documented as undefined.

For `filter/2` most also agree that the traversal order ought to be irrelevant, but since the documentation says:
> Returns a queue Q2 that is the result of calling Fun(Item) on all items in Q1, in order from front to rear.

it apparently defines the traversal order.  The intention of that sentence was actually to state that Q2 has got the same element order as Q1.  So here the question is if we can get away with claiming a documentation bug/clarification and change the implementation.  We have a hard time figuring out reasonable code that would be affected by a change in traversal order.

What do you or all other interested think about changing the traversal order for `filter/2`?

The only functions that traverse a queue with a `fun()` is (right?) `filter/2`, your new `filtermap/2` and `fold/3`, and the ones in this PR.  I think the documentation for all of these (also `delete/2` and `delete_r/2`) should state the traversal order as undefined (except possibly for `filter/2` depending on what we decide).

Maria-12648430:
> What do you or all other interested think about changing the traversal order for `filter/2`?

Just a feeling, but I don't think it can be done without breaking a lot of code retroactively.

> The only functions that traverse a queue with a `fun()` is (right?) `filter/2`, your new `filtermap/2` and `fold/3`, and the ones in this PR. I think the documentation for all of these (also `delete/2` and `delete_r/2`) should state the traversal order as undefined (except possibly for `filter/2` depending on what we decide).

Oh, for `fold` traversal in queue order _is_ important:
```
queue:fold(fun (X, Acc) -> [X|Acc] end, Q)
```
With undefined traversal order, the order of list elements would also be undefined and no way around it.

RaimoNiskanen:
My bad about `fold/3`.  It has to be in queue order, of course.  Too close to the weekend ;-)

(Note that it _would_ be possible to have an unordered fold that traverses all elements in some order.  I doubt it would be very useful since the accumulated value had to be queue order independent, but for example; creating a histogram?)

Concerning `delete/2` and `delete_r/2`: It is the _traversal_ order that could be undefined.  They should still remove the first/last matching element in queue order, which would be sufficient guarantee.  It is just the principle that if we do not _have_ to define a traversal order it is safer to state it as undefined.  I can not think of any reason to traverse in any other order, but with a different implementation, who knows?

  

> > What do you or all other interested think about changing the traversal order for filter/2?

> Just a feeling, but I don't think it can be done without breaking a lot of code retroactively.

I tried hard to figure how such code might be put together, but failed.

Maria-12648430:
> My bad about `fold/3`. It has to be in queue order, of course. Too close to the weekend ;-)

Indeed :)

> (Note that it _would_ be possible to have an unordered fold that traverses all elements in some order. I doubt it would be very useful since the accumulated value had to be queue order independent, but for example; creating a histogram?)

I know, but would that justify to have another function, like, let's call it `accumulate`?

> Concerning `delete/2` and `delete_r/2`: It is the _traversal_ order that could be undefined. They should still remove the first/last matching element in queue order, which would be sufficient guarantee. It is just the principle that if we do not _have_ to define a traversal order it is safer to state it as undefined. I can not think of any reason to traverse in any other order, but with a different implementation, who knows?

Ah, so just _state_ it as undefined in the documentation, in order to not block any paths that don't need to be blocked, right? :)

> > > What do you or all other interested think about changing the traversal order for filter/2?
> 
> > Just a feeling, but I don't think it can be done without breaking a lot of code retroactively.
> 
> I tried hard to figure how such code might be put together, but failed.

Dunno, just call it female intuition ;) But seriously, after I wrote that, I tried to find a reasonable example, but also failed ^^; Not even an unreasonable example

Maria-12648430:
Oh, just noticed: `delete/2`/`delete_r/2` don't need any mention of traversal order, as it does not use a predicate function =^^= That would be the future `remove/2`/`remove_r/2` :)

lib/stdlib/src/queue.erl
Maria-12648430: Done.

lib/stdlib/src/queue.erl
Maria-12648430: Done.

lib/stdlib/src/queue.erl
Maria-12648430: Obsolete, changed to `lists:all(Pred, F) andalso lists:all(Pred, R)` according to the discussion below, sacrificing traversal in queue order.

Maria-12648430:
I updated the PR, and hope I didn't break anything just before the weekend:
* `queue:any/2` and `queue:all/2` now run the respective `lists` function over the front and rear lists as-is, thereby sacrificing queue order traversal
* the documentation for `queue:any/2` and `queue:all/2` has been augmented with the sentence "The order of queue traversal is unefined."
* `delete/2` has been slightly changed to return `Q`, bound in the function head, if no matching item was found, instead of re-constructing the tuple from individual rear and front components

The documentation for `filtermap/2` still has "in order from front to rear" in it. I didn't touch that since `filtermap/2` was in another PR. So, organizational question, how should I best go in removing it (or rather, replacing it with "... undefined...")? Separate commit in this PR? Separate PR?

Regarding `remove/2`/`remove_r/2`: The agreement seems to be that those are useful and should be implemented, yes? If so, in this PR, too?

Finally, have a nice weekend :)

Maria-12648430:
> @Maria-12648430 and @okeuday; I have been reading sloppy and thought `delete/2` used a predicate. Then I agree with @okeuday that `remove(Pred, Q)` probably is even more useful than `delete(Item, Q)`, and that both would be useful.
> 
> Then `lists:remove(Pred, L)` is missing. Bummer! Annoying that `maps:remove(Key, Map)` takes a key to remove, and `lists:delete(Elem, List)` takes an element to match. Impedance mismatch...

I have been thinking about this over the weekend. To avoid the clash of `remove` in `lists`/`queue` with `maps:remove`, the function could be named `deletewith` (and `deletewith_r`) instead, which could also be added to `lists` later. `delete_with` would also be possible, but there is `lists:zipwith` which goes without the underscore, plus `queue:dlete_with_r` looks strange ^^;

jhogberg:
> What do you or all other interested think about changing the traversal order for filter/2?

I don't think changing the traversal order is warranted as it's doubtful that it will turn out better in practice (taking garbage into account). 

I believe changing the documentation is the right thing to do though, it's too easy to misinterpret right now. It may even be better to remove all mention of order since it's obvious that removing things from a queue shouldn't the affect order of the remaining elements (the function would be pointless if it did).

> the documentation for queue:any/2 and queue:all/2 has been augmented with the sentence "The order of queue traversal is unefined."

You don't need to point this out, everything unmentioned is undefined to begin with.

> The documentation for filtermap/2 still has "in order from front to rear" in it. I didn't touch that since filtermap/2 was in another PR. So, organizational question, how should I best go in removing it (or rather, replacing it with "... undefined...")? Separate commit in this PR? Separate PR?

None of these features have been released into the wild yet, so I think a separate commit in this PR is fine.

Maria-12648430:
I updated the PR again, removing any mention of traversal order from the documentation of `any/2`, `all/2`, `filtermap/2` and `filter/2`.

juhlig:
> I have been thinking about this over the weekend. To avoid the clash of `remove` in `lists`/`queue` with `maps:remove`, the function could be named `deletewith` (and `deletewith_r`) instead, which could also be added to `lists` later. `delete_with` would also be possible, but there is `lists:zipwith` which goes without the underscore, plus `queue:delete_with_r` looks strange ^^;

I'm in favor of `deletewith` (`deletewith_r`), and I think it would fit well in this PR ;)

Maria-12648430:
Last commit adds `deletewith/2` and `deletewith_r/2`, ie. "delete by ways of a predicate".

IngelaAndin:
Naming is hard and us not being consistent from the start makes it harder. The PS team discussed the underscore or not underscore names. Quite a few people think the underscore names are easier to read. The idle thing is to have consistency, but as we can not have that we think we should strive for consistency within a module.  The lists module consistently does not use underscores but the queue module does use underscores. We will probably never reach a consensus on which looks best.

Maria-12648430:
> Naming is hard

Yes =^^=

> and us not being consistent from the start makes it harder. The PS team discussed the underscore or not underscore names. Quite a few people think the underscore names are easier to read. The idle thing is to have consistency, but as we can not have that we think we should strive for consistency within a module. The lists module consistently does not use underscores but the queue module does use underscores. We will probably never reach a consensus on which looks best.

Ok, so to be clear, I should use `delete_with` and `delete_with_r` here, yes?
The base term `delete_with` is ok, rather than the formerly suggested `remove`, which would clash (somewhat semantically) with `maps:remove`?

IngelaAndin:
Yes that is right. And we also think it is good to avoid using the same name if there are semantically clashes in order to avoid as much confusion as possible.

Maria-12648430:
Done :)

RaimoNiskanen:

PR #2852, opened 2020-11-04
Remove installed crypto/priv/obj directory and all its object files

sverker:
This is a residue from the time of binary only releases,
now it just consumes disk space.


HansN:
Looks good to me.

PR #2853, opened 2020-11-05
erts: Fix exception handling during tracing

jhogberg:
Since 25fe3fb23c594d735cb6ebae120910e44f0cdae4, trace parameters always live next to their CP so we don't need to skip any "stack frame variables." This was generally harmless since the first argument was usually a CP anyway, but we'd skip too much when it wasn't. For example when return time tracing was used, and we didn't have any prev_info:

```
E -= 2;
E[1] = prev_info ? make_cp(erts_codeinfo_to_code(prev_info)) : NIL;
E[0] = (Eterm) beam_return_time_trace;
c_p->stop = E;
```

The loop to skip stack frame variables would thus skip the trace CP (which is OK) and then skip the `prev_info` (which is `[]`), and then we'd add one to skip what we thought was `prev_info`, ruining the scan.

### DIFF
--- a/erts/emulator/test/trace_call_time_SUITE.erl
+++ b/erts/emulator/test/trace_call_time_SUITE.erl
@@ -35,7 +35,7 @@
 
 -export([seq/3, seq_r/3]).
 -export([loaded/1, a_function/1, a_called_function/1, dec/1, nif_dec/1, dead_tracer/1,
-        return_stop/1]).
+        return_stop/1,catch_crash/1]).
 
 -define(US_ERROR, 10000).
 -define(R_ERROR, 0.8).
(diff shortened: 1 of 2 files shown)


erts/emulator/test/trace_call_time_SUITE.erl
```diff
@@ -634,6 +635,25 @@ spinner(N) ->
...
+
+    _ = erlang:trace_pattern({'_','_','_'}, false, [call_time]),
+    _ = erlang:trace(self(), false, [call]),
+
+    id(Res),
+
+    ok.
+
+catch_crash_1() ->
+    erlang:error(crash).
```
bjorng, line 656:
```suggestion
    error(crash).
```

PR #2854, opened 2020-11-05
otp: Remove HiPE

jhogberg:
As [mentioned on our mailing list](http://erlang.org/pipermail/erlang-questions/2020-June/099645.html) earlier this year, we're going to remove HiPE unless someone can commit to maintaining it long-term. It has been half-broken since OTP 22 and no longer works in OTP 24.

We cannot commit any resources to fixing it before release, let alone maintain it over time, so now's the time to step up if someone wants to keep it.


essen:
RIP.

bjorng:
`compile_SUITE:forms_2/1` fails.

bjorng:
(Testfallet testar fortfarande optionen `from_beam` som inte längre stöds.)

bjorng:
Because you have removed `from_beam` you should also remove the mapping of the `.beam` extension in `erl_compile:compiler/1` and `compile:from_beam/3`.

jhogberg:
> Because you have removed `from_beam` you should also remove the mapping of the `.beam` extension in `erl_compile:compiler/1` and the function `compile:from_beam/3`.

Done, also done the same for `c.erl` etc.

PR #2861, opened 2020-11-11
jit: Make mov_imm(Gp, NIL) shorter

jhogberg:
This PR shaves off a few bytes here and there by making the empty list constant (`NIL`) positive:

Before:
```
    ; When NIL is negative, we need to sign-extend it when moving to a 64-bit register,
    ; which uses 7 bytes:
    ;
    ; 48 c7 c0 fb ff ff ff
    mov rax, -5
```
After:
```
    ; When NIL is positive, we can use the fact that a move to the lower 32 bits of any
    ; 64-bit register implicitly zeroes its upper 32 bits. This uses 5 bytes instead:
    ;
    ; b8 3b 00 00 00
    mov eax, 0x3b
```

### DIFF
--- a/erts/emulator/beam/erl_term.h
+++ b/erts/emulator/beam/erl_term.h
@@ -283,7 +283,7 @@ _ET_DECLARE_CHECKED(Sint,signed_val,Eterm)
 #endif
 
 /* NIL access methods */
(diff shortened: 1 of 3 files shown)


garazdawi:
Do we know why NIL was negative to start with? If we do, maybe add that in the commit message?

jhogberg:
Not for sure, our best guess is that it may have had something to do with the old tag scheme but that's ancient history by now.

PR #2862, opened 2020-11-11
Clarify ETS traversal doc

sverker:


lib/stdlib/doc/src/ets.xml
```diff
@@ -195,10 +195,20 @@
...
+      traversal. The result of each concurrently updated object may be seen (or
+      not) depending on if it has happened when the traversal visits that part
+      of the table. The only way to guarantee a full consistent table snapshot
+      (if you really need that) is to disallow concurrent updates during the
+      entire traversal.
+    </p>
+    <p>
+      Moreover, traversals not done in a <em>safe</em> way, on tables where
+      keys are inserted or deleted during the traversal, may yield the
+      following undesired effects:
```
ferd, line 210: This is a thing that is possibly worth mentioning: since `ets:tab2file` scans the table to dump it (through a select), the same risk of inconsistencies also exists when creating the dump, even if ets:fixtable is used in it. 
The `safe_fixtable` option used prevents deletions from removing records but still doesn't guarantee any sort of consistency since the speed of the traversal can still impact the logical result of data dumped.

We had a bug at a previous job that was based on the insertion of entries, say `[A,E,F,K,M,P]` becoming `[A,B,E,F,K,M,N,P,T]`. We could end up with a snapshot that shows `[A,E,F,K,N,P,T]` even though `B` was inserted before either `N` or `T` (and could have been used to calculate `N` or `T`) because it ended up sorted in a bucket of the table that was already scanned when it happened. Repairing from this one was not obvious and took quite a while to figure out :)

sverker: The point of this PR was to clarify the distinction between table traversal over a consistent snapshot (which does not exists in ETS) and a "safe" traversal which avoids missed objects, double hits and badarg.

So, "done in safe way" **does** refer to `safe_fixtable` or the internal fixtable done by functions like `ets:tab2file`.

Any suggestions how to make the docs more clear are always welcome.

ferd: safe_fixtable just prevents deletions from interrupting the traversal, but does not actually prevent race conditions and other concurrency errors.

sverker: safe_fixtable does two things:

- It keeps deleted objects in the table (marked as deleted) so the key can still be used as an entry points by a traversal. Without it ets:next(DeletedKey) would badarg.
- It prevents the hash table from growing/shrinking and rehash objects. Without it a traversal may miss any objects or hit any objects twice if they are moved within the table.

Object created or deleted during a **safe** traversal will be "randomly" hit 0 or 1 time. All other objects in the table will be hit 1 time.

The section https://erlang.org/doc/man/ets.html#traversal tries to explain this.

PR #2797, opened 2020-10-13
Add maps:merge_with/3, maps:intersect/2, and maps:intersect_with/2

kjellwinblad:
This pull request extends the maps module with three new functions:

* `maps:merge_with/3` -- similar to `maps:merge/2` but takes a fun as a third argument that is used to resolve merge conflicts (this function is similar to Elexir's `map.merge(map1, map2, fun)`)
* `maps:intersect/2` -- Computes the intersection of two maps
* `maps:intersect_with/3` -- similar to `maps:intersect/2` but takes a fun as a third argument that is used to resolve conflicts

Motivation
-------------
[A function similar to `maps:intersect_with/3` is used in the Erlang compiler](https://github.com/erlang/otp/blob/cbd3c8e077506d1d6450541e9ccf3e1adac92b4c/lib/compiler/src/beam_ssa_type.erl#L2042) and an [Erlang user has requested `maps:merge_with/3`](https://bugs.erlang.org/browse/ERL-1367).

PR #2864, opened 2020-11-12
Initial implementation of EEP 50

josevalim:
This pull request adds support for maps in the sets module. According to my [benchmarks](https://github.com/josevalim/sets_bench), using maps is faster in the majority of cases, sometimes by multiple orders of magnitude, and in the few cases it is slower, it is only by 1.1x-1.4x, which can be improved with the steps outlined below. Note that, although we are using maps, the type is still opaque.

---

Fully adopting EEP 50 will require multiple PRs. In particular, following PRs should:

1. Remove cerl_sets

2. ~Incorporate the intersection implementation in #2797~

3. Write intersection, subtract, is_subset, is_disjoint, and from_list as NIFs so `sets` becomes the fastest set implementation in OTP on all possible scenarios

### DIFF
--- a/lib/stdlib/src/sets.erl
+++ b/lib/stdlib/src/sets.erl
@@ -18,10 +18,13 @@
 %% %CopyrightEnd%
 %%
 
-%% We use the dynamic hashing techniques by Per-Åke Larsson as
-%% described in "The Design and Implementation of Dynamic Hashing for
-%% Sets and Tables in Icon" by Griswold and Townsend.  Much of the
-%% terminology comes from that paper as well.
+%% The new version 2 has moved to use maps under the roof whenever a
+%% map is given.
+
+%% The previous version (version 1) uses the dynamic hashing techniques
+%% by Per-Åke Larsson as described in "The Design and Implementation
+%% of Dynamic Hashing for Sets and Tables in Icon" by Griswold and
+%% Townsend.  Much of the terminology comes from that paper as well.
 
 %% The segments are all of the same fixed size and we just keep
 %% increasing the size of the top tuple as the table grows.  At the
@@ -44,8 +47,14 @@
 -export([is_disjoint/2]).
 -export([subtract/2,is_subset/2]).
 -export([fold/3,filter/2]).
+-export([new/1, from_list/2]).
 
 -export_type([set/0, set/1]).
+
+%% This is the value used when sets are represented as maps.
+%% We use an empty list instead of an atom as it is cheaper
+%% to serialize.
+-define(VALUE, []).
 
 %% Note: mk_seg/1 must be changed too if seg_size is changed.
 -define(seg_size, 16).
@@ -54,6 +63,7 @@
 -define(contract_load, 3).
 -define(exp_size, ?seg_size * ?expand_load).
 -define(con_size, ?seg_size * ?contract_load).
+-compile({no_auto_import,[size/1]}).
 
 %%------------------------------------------------------------------------------
 
@@ -74,7 +84,7 @@
 
 -type set() :: set(_).
 
--opaque set(Element) :: #set{segs :: segs(Element)}.
+-opaque set(Element) :: #set{segs :: segs(Element)} | #{Element => ?VALUE}.
 
 %%------------------------------------------------------------------------------
 
@@ -84,10 +94,41 @@ new() ->
     Empty = mk_seg(?seg_size),
     #set{empty = Empty, segs = {Empty}}.
 
+-spec new([{version, 1..2}]) -> set().
+new([{version, 2}]) ->
+    #{};
+new(Opts) ->
+    case proplists:get_value(version, Opts, 1) of
+        1 -> new();
+        2 -> new([{version, 2}])
+    end.
+
+%% from_list([Elem]) -> Set.
+%%  Build a set from the elements in List.
+-spec from_list(List) -> Set when
+      List :: [Element],
+      Set :: set(Element).
+from_list(Ls) ->
+    lists:foldl(fun (E, S) -> add_element(E, S) end, new(), Ls).
+
+-spec from_list(List, [{version, 1..2}]) -> Set when
+      List :: [Element],
+      Set :: set(Element).
+from_list(Ls, [{version, 2}]) ->
+    maps:from_list([{K,?VALUE}||K<-Ls]);
+from_list(Ls, Opts) ->
+    case proplists:get_value(version, Opts, 1) of
+        1 -> from_list(Ls);
+        2 -> from_list(Ls, [{version, 2}])
+    end.
+
+%%------------------------------------------------------------------------------
+
 %% is_set(Set) -> boolean().
 %%  Return 'true' if Set is a set of elements, else 'false'.
 -spec is_set(Set) -> boolean() when
       Set :: term().
+is_set(#{}) -> true;
 is_set(#set{}) -> true;
 is_set(_) -> false.
 
@@ -95,35 +136,36 @@ is_set(_) -> false.
 %%  Return the number of elements in Set.
 -spec size(Set) -> non_neg_integer() when
       Set :: set().
-size(S) -> S#set.size. 
+size(#{}=S) -> map_size(S);
+size(#set{size=Size}) -> Size.
 
 %% is_empty(Set) -> boolean().
 %%  Return 'true' if Set is an empty set, otherwise 'false'.
 -spec is_empty(Set) -> boolean() when
       Set :: set().
-is_empty(S) -> S#set.size=:=0.
+is_empty(#{}=S) -> map_size(S)=:=0;
+is_empty(#set{size=Size}) -> Size=:=0.
 
 %% to_list(Set) -> [Elem].
 %%  Return the elements in Set as a list.
 -spec to_list(Set) -> List when
       Set :: set(Element),
       List :: [Element].
-to_list(S) ->
+to_list(#{}=S) ->
+    maps:keys(S);
+to_list(#set{} = S) ->
     fold(fun (Elem, List) -> [Elem|List] end, [], S).
-
-%% from_list([Elem]) -> Set.
-%%  Build a set from the elements in List.
--spec from_list(List) -> Set when
-      List :: [Element],
-      Set :: set(Element).
-from_list(L) ->
-    lists:foldl(fun (E, S) -> add_element(E, S) end, new(), L).
 
 %% is_element(Element, Set) -> boolean().
 %%  Return 'true' if Element is an element of Set, else 'false'.
 -spec is_element(Element, Set) -> boolean() when
       Set :: set(Element).
-is_element(E, S) ->
+is_element(E, #{}=S) ->
+    case S of
+        #{E := _} -> true;
+        _ -> false
+    end;
+is_element(E, #set{}=S) ->
     Slot = get_slot(S, E),
     Bkt = get_bucket(S, Slot),
     lists:member(E, Bkt).
@@ -133,7 +175,9 @@ is_element(E, S) ->
 -spec add_element(Element, Set1) -> Set2 when
       Set1 :: set(Element),
       Set2 :: set(Element).
-add_element(E, S0) ->
+add_element(E, #{}=S) ->
+    S#{E=>?VALUE};
+add_element(E, #set{}=S0) ->
     Slot = get_slot(S0, E),
     Bkt = get_bucket(S0, Slot),
     case lists:member(E, Bkt) of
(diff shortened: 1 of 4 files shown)


lib/stdlib/src/sets.erl
```diff
@@ -206,10 +235,20 @@ union1(S1, []) -> S1.
       Set1 :: set(Element),
       Set2 :: set(Element),
       Set3 :: set(Element).
-intersection(S1, S2) when S1#set.size < S2#set.size ->
-    filter(fun (E) -> is_element(E, S2) end, S1);
-intersection(S1, S2) ->
-    filter(fun (E) -> is_element(E, S1) end, S2).
+intersection(#{}=S1, #{}=S2) ->
+    if
```
jhogberg, line 239: I noticed that maps were slower here, can you try the implementation from PR #2797?

josevalim: My plan is to not change the `cerl_sets` implementation for now and once that PR is merged, we can just delete to maps:intersection. What do you think?

jhogberg: Sure, I just figured I'd point it out since you mentioned benchmarks and this was the only operation that was slower. #2797 should take care of that :)

josevalim: Oh, I see. Perfect, thanks. :)

lib/stdlib/src/sets.erl
```diff
@@ -47,6 +50,11 @@
 
 -export_type([set/0, set/1]).
 
+%% This is the value used when sets are represented as maps.
+%% We use an empty list instead of atom as it is very cheap
```
paulo-ferraz-oliveira, line 54: of atom > of an atom?

lib/stdlib/src/sets.erl
```diff
@@ -74,7 +82,7 @@
 
 -type set() :: set(_).
```
paulo-ferraz-oliveira, line 83: I know it's probably outside the scope of this pull request, but would it make sense to make this type opaque, too?

michalmuskala: Would't this mean that dialyzer couldn't unify them? So if you had a function returning `set()` you wouldn't be able to pass the result to another function expecting `set(_)` as an argument?

josevalim:
I went ahead and tackled the pending issues. In particular, I chose for `sets:new/0` and `sets:from_list/1` to return maps but I can revert that if deemed too forward.

lib/stdlib/src/sets.erl
josevalim: Moved the suggestion to the PR description.

lib/stdlib/src/sets.erl
paulo-ferraz-oliveira: I'm unaware if that would be the case, but would like to understand the same as you. I can try it locally though, to see if it makes sense. If what you ask is replied in the sense of "Yes, dialyzer couldn't unify them", then my request is no good and can be scratched.

paulo-ferraz-oliveira: You are right 😄. It breaks opacity, for `dialyzer`. In any case (not the scope of this PR, surely) I wonder if that should be the case, though.

lib/stdlib/src/sets.erl
```diff
@@ -74,56 +84,74 @@
...
+deprecated_from_list(L) ->
+    lists:foldl(fun (E, S) -> add_element(E, S) end, deprecated_new(), L).
+
+%%------------------------------------------------------------------------------
+
 %% is_set(Set) -> boolean().
 %%  Return 'true' if Set is a set of elements, else 'false'.
 -spec is_set(Set) -> boolean() when
       Set :: term().
+is_set(#{}) -> true;
```
asabil, line 117: Would it make sense wrapping the map into a record, say `map_set`, so as to avoid any map being interpreted as a set?

lib/stdlib/src/sets.erl
```diff
@@ -74,56 +84,74 @@
...
 
--opaque set(Element) :: #set{segs :: segs(Element)}.
+-opaque set(Element) :: #set{segs :: segs(Element)} | #{Element => ?VALUE}.
 
 %%------------------------------------------------------------------------------
 
 %% new() -> Set
 -spec new() -> set().
-new() ->
+new() -> #{}.
```
jhogberg, line 93: After discussing this internally we'd like the old implementation to be the default in order not to break distribution: an application on OTP 24 may send a set to another one on OTP 23, which will crash if you send a map.

We'll change the default in OTP 26 when all supported versions support maps.

josevalim: Thanks! Any suggestions on how to call the `new/0` and `from_list/1` variants that return `maps`? Or we don't want to introduce such functions as they will be removed on OTP 26 anyway?

jhogberg: Not at the moment, we'll take it up at the next technical board meeting.

josevalim: Sounds good, I will wait until a decision on this front then. In the worst case scenario, we keep them as `map_new/0` and `map_from_list/1` so we can use it within OTP and in tests, but we don't make them public.

jhogberg: We've decided to add `new/1` and `from_list/2` that take an option list, where one can specify `{version,1}` for the old and `{version,2}` for the new version.

We'll change the default in OTP 26.

josevalim:
@jhogberg I have updated the PR to support a version flag. Docs have been updated too.

I have also benchmarked maps:intersect/2 and it is slower then the current cerl_sets implementation. You can see [the benchmark here](https://github.com/josevalim/sets_bench/blob/master/bench/intersection.exs) and [the benchmark results here](https://github.com/josevalim/sets_bench/blob/master/results/intersection.txt). For this reason, I decided to keep the cerl_sets implementation for intersection for now. If you want to give it a try yourself, you can clone the repo, run `mix deps.get` and then `mix run bench/intersection.exs`.

jhogberg:
> @jhogberg I have updated the PR to support a version flag. Docs have been updated too.

Great :)

> I have also benchmarked maps:intersect/2 and it is slower then the current cerl_sets implementation. You can see [the benchmark here](https://github.com/josevalim/sets_bench/blob/master/bench/intersection.exs) and [the benchmark results here](https://github.com/josevalim/sets_bench/blob/master/results/intersection.txt). For this reason, I decided to keep the cerl_sets implementation for intersection for now. If you want to give it a try yourself, you can clone the repo, run `mix deps.get` and then `mix run bench/intersection.exs`.

The current implementation uses `maps:intersect_with/3` which turned out to be more expensive than I thought. A quick-and-dirty specialization of `maps:intersect/2` makes it much faster than the alternatives:


```erlang
-spec intersect(Map1,Map2) -> Map3 when
    Map1 :: #{ Key => term() },
    Map2 :: #{ term() => Value2 },
    Map3 :: #{ Key => Value2 }.

intersect(Map1, Map2) when is_map(Map1), is_map(Map2) ->
    case map_size(Map1) < map_size(Map2) of
        true ->
            Iterator = maps:iterator(Map1),
            intersect_1(maps:next(Iterator), Map1, Map2);
        false ->
            Iterator = maps:iterator(Map2),
            intersect_2(maps:next(Iterator), Map2, Map1)
    end;
intersect(Map1, Map2) ->
    erlang:error(error_type_two_maps(Map1, Map2),
                 [Map1, Map2]).

intersect_1({Key, LHS_Value, Iterator}, LHS0, RHS) ->
    LHS = case RHS of
              #{ Key := LHS_Value } ->
                  LHS0;
              #{ Key := RHS_Value } ->
                  LHS0#{ Key := RHS_Value };
              _ ->
                  maps:remove(Key, LHS0)
          end,
    intersect_1(maps:next(Iterator), LHS, RHS);
intersect_1(none, Res, _) ->
    Res.

intersect_2({Key, _RHS_Value, Iterator}, RHS0, LHS) ->
    RHS = case LHS of
              #{ Key := _LHS_Value } ->
                  %% The value in RHS has precedence, so it doesn't need to
                  %% be updated even if they differ.
                  RHS0;
              _ ->
                  maps:remove(Key, RHS0)
          end,
    intersect_2(maps:next(Iterator), RHS, LHS);
intersect_2(none, Res, _) ->
    Res.
```

```
$ mix run bench/intersection.exs
Error trying to dermine erlang version enoent
Operating System: Linux
CPU Information: Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz
Number of Available Cores: 8
Available memory: 31.17 GB
Elixir 1.12.0-dev
Erlang ok

warmup: 2 s
time: 5 s
memory time: 0 ns
parallel: 1
inputs: large eq bin, large eq int, medium eq bin, medium eq int, small eq bin, small eq int
Estimated total run time: 2.10 min

Benchmarking cerl_sets with input large eq bin...
Benchmarking cerl_sets with input large eq int...
Benchmarking cerl_sets with input medium eq bin...
Benchmarking cerl_sets with input medium eq int...
Benchmarking cerl_sets with input small eq bin...
Benchmarking cerl_sets with input small eq int...
Benchmarking maps with input large eq bin...
Benchmarking maps with input large eq int...
Benchmarking maps with input medium eq bin...
Benchmarking maps with input medium eq int...
Benchmarking maps with input small eq bin...
Benchmarking maps with input small eq int...
Benchmarking sets with input large eq bin...
Benchmarking sets with input large eq int...
Benchmarking sets with input medium eq bin...
Benchmarking sets with input medium eq int...
Benchmarking sets with input small eq bin...
Benchmarking sets with input small eq int...

Name                ips        average  deviation         median         99th %
maps             166.72        6.00 ms     ±8.78%        5.95 ms        6.28 ms
sets              65.36       15.30 ms     ±5.28%       15.13 ms       16.82 ms
cerl_sets         45.05       22.20 ms     ±7.66%       22.59 ms       25.78 ms

Comparison: 
maps             166.72
sets              65.36 - 2.55x slower +9.30 ms
cerl_sets         45.05 - 3.70x slower +16.20 ms

Name                ips        average  deviation         median         99th %
maps             251.17        3.98 ms     ±9.02%        3.97 ms        4.08 ms
sets             106.50        9.39 ms     ±4.05%        9.34 ms       10.02 ms
cerl_sets         49.25       20.31 ms     ±3.88%       20.21 ms       21.04 ms

Comparison: 
maps             251.17
sets             106.50 - 2.36x slower +5.41 ms
cerl_sets         49.25 - 5.10x slower +16.32 ms

Name                ips        average  deviation         median         99th %
maps            19.80 K       50.51 μs    ±43.19%       49.99 μs       53.41 μs
sets             7.16 K      139.58 μs    ±25.66%      138.41 μs      187.03 μs
cerl_sets        6.85 K      145.93 μs    ±24.94%      143.19 μs      207.91 μs

Comparison: 
maps            19.80 K
sets             7.16 K - 2.76x slower +89.07 μs
cerl_sets        6.85 K - 2.89x slower +95.42 μs

Name                ips        average  deviation         median         99th %
maps            34.16 K       29.28 μs    ±29.85%       28.79 μs       34.87 μs
sets            11.64 K       85.88 μs    ±33.19%       84.88 μs      130.80 μs
cerl_sets        8.52 K      117.40 μs    ±28.48%      115.68 μs      167.77 μs

Comparison: 
maps            34.16 K
sets            11.64 K - 2.93x slower +56.60 μs
cerl_sets        8.52 K - 4.01x slower +88.12 μs

Name                ips        average  deviation         median         99th %
maps          1212.89 K        0.82 μs  ±2932.23%        0.75 μs        1.06 μs
cerl_sets      816.49 K        1.22 μs  ±1638.32%        1.10 μs        1.47 μs
sets           658.45 K        1.52 μs  ±1169.52%        1.40 μs        1.82 μs

Comparison: 
maps          1212.89 K
cerl_sets      816.49 K - 1.49x slower +0.40 μs
sets           658.45 K - 1.84x slower +0.69 μs

Name                ips        average  deviation         median         99th %
maps             4.80 M      208.52 ns  ±9696.18%         147 ns         453 ns
cerl_sets        1.90 M      525.39 ns  ±4505.73%         405 ns         741 ns
sets             0.79 M     1270.84 ns  ±1164.76%        1138 ns        1556 ns

Comparison: 
maps             4.80 M
cerl_sets        1.90 M - 2.52x slower +316.87 ns
sets             0.79 M - 6.09x slower +1062.31 ns
```

Adding a fast-path for `maps:intersect(Same, Same)` makes it 10-30x faster on my machine, but may be slower in some cases so I've left that out for now.

Either way I'm happy with this PR as-is and can merge it now if you want me to, I can always switch to `maps:intersect/2` in the branch that improves its performance.

josevalim:
@jhogberg awesome, i ran your branch too and got 3-4x solid improvement. I changed this PR to use maps:intersect, updated the commit message and pushed.

Next I will update the EEP and then send a separate PR to remove cerl_sets.

jhogberg:
Merged, thanks for the PR!

PR #2865, opened 2020-11-13
Fix the make configuration for macOS 11.0.1

idyll:
This change is needed to ensure that we're able to configure make on macOS 11.0.1.

A smilar fix was applied for 11.0 in #2700 

Main changes:

 - Fix the integer macOS version correctly for the build version check

Verified via ./otp_build tests

### DIFF
--- a/make/configure.in
+++ b/make/configure.in
@@ -398,7 +398,7 @@ if test $CROSS_COMPILING = no; then
 	       [1-9][0-9].[0-9])
 	          int_macosx_version=`echo $macosx_version | sed 's|\([^\.]*\)\.\([^\.]*\)|\10\200|'`;;
 	       [1-9][0-9].[0-9].[0-9])
-	          int_macosx_version=`echo $macosx_version | sed 's|\([^\.]*\)\.\([^\.]*\)\.\([^\.]*\)|\1\2\3|'`;;
+	          int_macosx_version=`echo $macosx_version | sed 's|\([^\.]*\)\.\([^\.]*\)\.\([^\.]*\)|\10\20\3|'`;;
 	       [1-9][0-9].[1-9][0-9])
 	          int_macosx_version=`echo $macosx_version | sed 's|\([^\.]*\)\.\([^\.]*\)|\1\200|'`;;
 	       [1-9][0-9].[1-9][0-9].[0-9])


idyll:
Let me know if I need to open this against `maint` to make sure that it is applied to the next 23.1.x release

bjorng:
> Let me know if I need to open this against maint to make sure that it is applied to the next 23.1.x release

Yes, please do. (You can edit this pull request to base it on maint. If your commit is currently is based on master, you will also need to rebase it and force-push.)

idyll:
> Yes, please do. (You can edit this pull request to base it on maint. If your commit is currently is based on master, you will also need to rebase it and force-push.)

Done.  :)

aselder:
yay!!! Can we get this merged to master and maint ASAP?

theoks:
Thank you for the contribution, I hope this appears on maint soon. Will this fix be included in a 22.* version as well?

verrchu:
I have successfully built maint branch with this patch on macOS 11.0.1
please merge it

ryandesign:
Both this fix and the previous one for macOS 11.0 in b93b9f95438fc3c4ccfef1b7da33379abe45cbb1 are wrong and break things for Mac OS X 10.x. See https://bugs.erlang.org/browse/ERL-1407.

bjorng:
How does it break 10.x? The branch builds fine on my Mac with macOS 10.15.7 as well on a Mac with 10.5.8 in our lab. On which version would it be a problem?

I have not yet tested the branch myself on macOS 11.0.1, but @verrchu state above that it works.

I am open to suggestions for a better fix of the problem. Perhaps it would be better to remove the version check. I don't remember why we added it in the first place.

ryandesign:
It would be a problem on OS X 10.9 and earlier. See https://bugs.erlang.org/browse/ERL-1407.

bjorng:
> It would be a problem on OS X 10.9 and earlier.

Not on all versions before 10.9. As I said, it built fine on 10.5.8.

I'll try to take closer look on this problem before the release of 23.2 to avoid breaking certain version of macOS. For now, this pull request does seem to fix the immediate problem of building on macOS 11.0.1.

ryandesign:
It may build fine, but the test is no longer testing what you think it's testing.

bjorng:
I see no reason to keep the test, so I'm removing it in #2871.

PR #2869, opened 2020-11-16
port_please/2 returns {port, _, _}, not {ok, _, _}

jesperes:
This is a followup PR to https://github.com/erlang/otp/pull/2616 which fixed the type spec for port_please/3, but missed the incorrect `{ok, _, _}` term in port_please/2 which should be `{port, _, _}`.


garazdawi:
Thanks!

PR #2870, opened 2020-11-16
Fix configure on macOS Big Sur 11.0.1

ylh:
I'll just reproduce the commit message below:

```
Generate plausible synthetic __ENVIRONMENT_MAC_OS_X_VERSION_MIN_REQUIRED__

This generates comparison values for the predefined macro in line with the
precedent set by Apple. Single-digit 10.x versions followed 10x0, and starting
with 10.10, moved to MMmmPP, Major/minor/Point. Left-padding on the minor is an
educated guess. This should now produce the correct value, from OS versions
Erlang never even ran on, to the current Big Sur Patch (11.0.1) that broke
configure, to future versions.
```

### DIFF
--- a/make/configure.in
+++ b/make/configure.in
@@ -113,8 +113,8 @@ AS_HELP_STRING([--enable-bootstrap-only],
 [ if test "X$enableval" = "Xyes"; then
      BOOTSTRAP_ONLY=yes
   else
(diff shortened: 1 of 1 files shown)


bjorng:
Already fixed in #2865, which has been merged.

PR #2871, opened 2020-11-17
macOS: Remove version check in configure script

bjorng:
There is a version check in one of the configure scripts that
tests the version of macOS to ensure that the executables that
are built can be run on the same computer that built them.

The reason for this test has been lost in the mists of time.
Presumably there was a good reason for the test when it was
first added.

Since the test recently has become more complicated because of the
introduction of macOS 11.0 and 11.0.1, and is incorrect for some
versions of macOS, remove it. If there turns out that the version test
would be useful, it can be added back later (but in a less buggy way,
preferably using `awk` instead of `sed` to make the version parsing
simpler and more accurate).

### DIFF
--- a/make/configure.in
+++ b/make/configure.in
@@ -387,50 +387,6 @@ if test X${enable_native_libs} = Xyes -a X${enable_hipe} != Xno; then
 fi
 AC_SUBST(NATIVE_LIBS_ENABLED)
 
-if test $CROSS_COMPILING = no; then
-   case $host_os in
-   	darwin*)
-	   macosx_version=`sw_vers -productVersion`
-	   test $? -eq 0 || {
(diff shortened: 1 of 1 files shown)


andyduong1920:
👍

andyduong1920:
Thank you for your works @bjorng 🎉 . Might I know when will the next release for this one @bjorng ?

bjorng:
@andyduong1920 It will be included OTP 23.2 in December.

andyduong1920:
> @andyduong1920 It will be included OTP 23.2 in December.

Is there any chance to have this fix is released sooner @bjorng, maybe 23.1.4 as MasOS (Big Sur) users can't install 23.1.3 and before that including me 😃

gromgit:
Will this be backported to earlier versions, like 22.x? Homebrew helpers like me would like to know. Thanks much!

PR #2872, opened 2020-11-18
Add -Werror to github actions C/C++ compilation

garazdawi:
This PR does three things:

* Add -Werror to all CFLAGS
* Start using clang to compile 64-bit in order to catch clang warnings
* Fix all warnings in applications that came up along the way

### DIFF
--- a/erts/emulator/beam/sys.h
+++ b/erts/emulator/beam/sys.h
@@ -1160,60 +1160,66 @@ ERTS_GLB_INLINE size_t sys_strlen(const char *s)
 
 /* Standard set of integer macros  .. */
 
-#define get_int64(s) (((Uint64)(((unsigned char*) (s))[0]) << 56) | \
-                      (((Uint64)((unsigned char*) (s))[1]) << 48) | \
-                      (((Uint64)((unsigned char*) (s))[2]) << 40) | \
-                      (((Uint64)((unsigned char*) (s))[3]) << 32) | \
-                      (((Uint64)((unsigned char*) (s))[4]) << 24) | \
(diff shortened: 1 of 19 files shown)


erts/emulator/beam/sys.h
```diff
@@ -1167,14 +1167,14 @@ ERTS_GLB_INLINE size_t sys_strlen(const char *s)
...
 
-#define put_int64(i, s) do {((char*)(s))[0] = (char)((Sint64)(i) >> 56) & 0xff;\
-                            ((char*)(s))[1] = (char)((Sint64)(i) >> 48) & 0xff;\
-                            ((char*)(s))[2] = (char)((Sint64)(i) >> 40) & 0xff;\
-                            ((char*)(s))[3] = (char)((Sint64)(i) >> 32) & 0xff;\
-                            ((char*)(s))[4] = (char)((Sint64)(i) >> 24) & 0xff;\
-                            ((char*)(s))[5] = (char)((Sint64)(i) >> 16) & 0xff;\
-                            ((char*)(s))[6] = (char)((Sint64)(i) >> 8)  & 0xff;\
-                            ((char*)(s))[7] = (char)((Sint64)(i))       & 0xff;\
+#define put_int64(i, s) do {((char*)(s))[0] = (char)((Sint64)(i) >> 56) & (char)0xff;\
```
jhogberg, line 1170: Nitpick; casting `0xff` to char is implementation-defined.

```suggestion
#define put_int64(i, s) do {((byte*)(s))[0] = (byte)(((Sint64)(i) >> 56) & 0xff);\
```

PR #2873, opened 2020-11-19
Add erl option for reading config from file descriptor

kjellwinblad:
Make it possibe for users to specify the option "-configfd FD" when
executing the erl command. When this option is given the system will
try to read and parse configuration parameters from the file
descriptor. At the moment the only configuration format that is
supported is the format that is supported when using the -config
option.

TODO:

* Test that it works correctly on windows


erts/emulator/nifs/common/prim_file_nif.c
```diff
@@ -521,6 +524,61 @@ static ERL_NIF_TERM open_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[]
...
+     * blocking operation which must not be done in the GC callback, and we
+     * can't defer it as the resource is gone as soon as it returns.
+     *
+     * We instead keep the resource alive until efile_close is called, after
+     * which it's safe to leave things to the GC. If the controlling process
+     * were to die before the user had a chance to close their file, the above
+     * monitor will tell the erts_prim_file process to close it for them. */
+    result = enif_make_resource(env, d);
+
+    return enif_make_tuple2(env, am_ok, result);
```
jhogberg, line 579: Consider sharing this part with `open_nif`, I think it's a bit too tricky to duplicate.

erts/emulator/nifs/win32/win_prim_file.c
```diff
@@ -491,6 +491,42 @@ posix_errno_t efile_open(const efile_path_t *path, enum efile_modes_t modes,
     }
 }
 
+posix_errno_t efile_from_fd(int fd,
+                            enum efile_modes_t modes,
+                            ErlNifResourceType *nif_type,
+                            efile_data_t **d) {
+    HANDLE handle;
+
+    handle = (HANDLE)_get_osfhandle(fd);
```
jhogberg, line 500: `_get_osfhandle` invokes the invalid parameter handler when `fd` is invalid, which crashes in debug builds. I'll try to find a workaround that doesn't involve disabling the handler.

erts/emulator/nifs/common/prim_file_nif.c
```diff
@@ -521,6 +524,61 @@ static ERL_NIF_TERM open_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[]
...
+    ASSERT(argc == 2);
+    if(!enif_is_list(env, argv[1])) {
+        return enif_make_badarg(env);
+    }
+
+    if(!enif_is_number(env, argv[0])) {
+        return enif_make_badarg(env);
+    }
+
+    modes = efile_translate_modelist(env, argv[1]);
```
jhogberg, line 545: The `fd` is already open, why require a mode list?

lib/kernel/src/application_controller.erl
```diff
@@ -1799,48 +1805,99 @@ do_config_diff([{Env, Value} | AppEnvNow], AppEnvBefore, {Changed, New}) ->
...
+                    NewEnv
+            end;
+        {error, {Line, _Mod, Str}} ->
+            throw({error, {FName, Line, Str}})
+    end;
+conf_param_to_conf({configfd, FileDescStrP}) ->
+    FileDescStr = unicode:characters_to_nfc_list(FileDescStrP),
+    IsDigit = fun(C) -> lists:member(C, "0123456789") end,
+    FileDesc =
+        try list_to_integer(lists:takewhile(IsDigit, FileDescStr))
```
jhogberg, line 1840: Since you're doing `dropwhile/2` later on, why not `partition/2` instead?

lib/kernel/src/application_controller.erl
```diff
@@ -1799,48 +1805,99 @@ do_config_diff([{Env, Value} | AppEnvNow], AppEnvBefore, {Changed, New}) ->
...
+        catch
+            error:badarg ->
+                throw({error, {file_descriptor,
+                               FileDescStr,
+                               invalid_file_desc,
+                               "The given file descriptor has incorrect format. "
+                               "The format should be \"FileDescId[.FileType]\". "
+                               "Examples: 3 or 3.config"}})
```
jhogberg, line 1848: The documentation for this option doesn't mention trailing `.config`

Out of curiosity, what's the reason for allowing it?

erts/emulator/nifs/win32/win_prim_file.c
jhogberg: There doesn't seem to be a straightforward way to do this, so I think it's best to disable the handler temporarily just for this call.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/set-invalid-parameter-handler-set-thread-local-invalid-parameter-handler?view=msvc-160

erts/emulator/nifs/win32/win_prim_file.c
```diff
@@ -491,6 +491,42 @@ posix_errno_t efile_open(const efile_path_t *path, enum efile_modes_t modes,
...
+
+    if(handle != INVALID_HANDLE_VALUE) {
```
jhogberg, line 502: You should check for `-2` as well.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=msvc-160

erts/emulator/nifs/win32/win_prim_file.c
```diff
@@ -491,6 +491,42 @@ posix_errno_t efile_open(const efile_path_t *path, enum efile_modes_t modes,
...
+            return ENOTDIR;
+        }
+
+        /* We could check if handle has read and write permissions that
+         * correspond to the modes parameter here. However, how to do this
+         * does not seem to be very well documented and is not necessary
+         * for the current usage. */
+
+        w = (efile_win_t*)enif_alloc_resource(nif_type, sizeof(efile_win_t));
+        w->handle = handle;
```
jhogberg, line 517: You need to mark the resource as having been created from a file descriptor, and store the original fd so it can be `_close`d. The documentation says we can't call `CloseHandle` on these.

lib/kernel/src/application_controller.erl
```diff
@@ -1910,6 +2001,59 @@ scan_file(Str) ->
...
+                            MaxSizeBytes) ->
+    ReadLimit = 1024,
+    Read =
+        fun Read(Ref, _Fd, _Data, DataSize) when DataSize > MaxSizeBytes ->
+                Reason = io_lib:format("Max size ~w bytes exceeded",
+                                       [MaxSizeBytes]),
+                {Ref, error, Reason};
+            Read(Ref, Fd, Data, DataSize) ->
+                case file:read(Fd, ReadLimit) of
+                    eof ->
```
josevalim, line 2015: What happens when we are at the end of the file descriptor but the OS or the parent program did not finish writing to it? Does it return `eof`? Does it block? Depending on the behavior, we can get nondeterministic behavior where the configuration will only succeed if whoever is writing it has written it quick enough.

In any case, the fd to file descriptor functionality is really neat! I can't wait to play with it as an alternative to the current port-based one, which buffers all input and can be problematic in some cases.

lib/kernel/src/application_controller.erl
```diff
@@ -1910,6 +2001,59 @@ scan_file(Str) ->
...
+                        {Ref, ok, Data};
+                    {ok, Bin} ->
+                        Read(Ref, Fd, [Data, Bin], DataSize + byte_size(Bin));
+                    {error, Reason} ->
+                        {Ref, error, Reason}
+                end
+        end,
+    Reciver = self(),
```
josevalim, line 2023: Small nitpick: "Reciver" -> "Receiver".

lib/kernel/src/application_controller.erl
garazdawi: The `.config` suffix is part of future extensions to the format of the config files. It does not have to be part of this PR, but we decided that it may as well so that all functionality is added in one place.

lrascao:
I'm curious on the use case behind this feature, maybe this PR could contain a small explanation as to why this is handy

josevalim:
@lrascao my understanding is that, if you want to dynamically configure the Erlang VM in production (for example, imagine you run an escript or bash or whatever to emit a configuration file), you need to write those files to disk but that's an issue in read-only file systems. This allows someone to pipe the configuration to the VM using -configfd 0 or similar.

lrascao:
@josevalim makes sense, never considered/came across the read-only scenario really, thanks

erts/emulator/nifs/common/prim_file_nif.c
kjellwinblad: I thought it could be useful to check that the FD has the expected access flags but I have now realized that one does not get much benefit from that (at least not in how it is used now) as one will notice that anyway when one starts to use the file descriptor. So I think I will remove the mode list parameter.

lib/kernel/src/application_controller.erl
kjellwinblad: That's a good suggestion. I was not familiar with `partition/2` but I have now looked it up and I think it fits very well here.

erts/emulator/nifs/win32/win_prim_file.c
kjellwinblad: Thanks, I should have noticed that text in the note.

lib/kernel/src/application_controller.erl
jhogberg: Err, sorry, `splitwith/2`

erts/emulator/nifs/win32/win_prim_file.c
kjellwinblad: Thanks again, I will fix that.

lib/kernel/src/application_controller.erl
kjellwinblad: Thanks for the question. Yes, `file:read/2` will block and does not return `eof` unless the the OS or the parent program has finished writing to it. However, we can still get the nondeterministic behavior that you describe as there is currently a timeout for reading a configuration from a file descriptor see [here](https://github.com/erlang/otp/pull/2873/files/099232f6de906f05664031b05ac1327a9bc1be2b#diff-2e7a405caf2dc0b5e567c6f01d62ad4b37ef9e517976ffc02c716900f2c87a9fR2046). This is meant as a feature to prevent `erl` from getting struck when reading a configuration from a file descriptor that never ends or is extremely slow. Currently, this timeout is set to two seconds.  If it times out, an error message describing what went wrong will be printed. Perhaps two seconds is too short? Do you think a timeout does more harm than good?

erts/doc/src/erl_cmd.xml
```diff
@@ -216,6 +216,39 @@
...
+$ <input>echo '[{kernel, [{logger_level, error}]}].' > test2.config</input>
+$ <input>erl \
+-noshell \
+-configfd 3 \
+-configfd 4 \
+-eval \
+'io:format("~p~n",[application:get_env(kernel, logger_level)]),erlang:halt()' \
+3&lt; test1.config 4&lt; test2.config</input>
+{ok,error}
+</pre>
```
ferd, line 249: Is there any expected order for configs to be read? Is it `-config` first then `-configfd` entries in declaration order? The base config option can also refer to other config files.

Also what's the merge policy? You mention in the example setting `kernel.logger_level` twice, but what if one of the FDs is a setting of the logger and another one is say, shell history? Are they properly merged within each app definition, or are the setting per-app?

erts/etc/common/erlexec.c
```diff
@@ -651,6 +646,16 @@ int main(int argc, char **argv)
                         } while ((i+1) < argc && argv[i+1][0] != '-' && argv[i+1][0] != '+');
 		    }
 #endif
+                    else if (strcmp(argv[i], "-configfd") == 0) {
+                        if (i+1 >= argc)
+ 			    usage("-configfd");
+                        if ( strcmp(argv[i+1], "0") != 0 ) {
+			    add_arg(argv[i]);
+                        } else {
+                            add_args("-noshell", "-noinput", NULL);
```
ferd, line 655: Any checks required for someone setting it to 1?

lib/kernel/src/application_controller.erl
josevalim: I think a timeout is welcome but 2 seconds seems to be too low? Maybe someone is curling an internal endpoint or a service for this config and I can see it taking a bit more than that.I would go with 10 seconds or similar, enough to prevent hiccups but providing relatively soon feedback if something went wrong. But this is just gut feeling, so your guess is as good as mine. :)

erts/emulator/nifs/common/prim_file_nif.c
kjellwinblad: You are right. It is much cleaner to break this out to a helper function.

lib/kernel/src/application_controller.erl
```diff
@@ -1799,48 +1805,99 @@ do_config_diff([{Env, Value} | AppEnvNow], AppEnvBefore, {Changed, New}) ->
...
+                          BFName ++ ".config"),
+    case load_file(FName) of
+        {ok, NewEnv} ->
+            %% OTP-4867 sys.config may now contain names of other
+            %% .config files as well as configuration parameters.
+            %% Therefore read and merge contents.
+            if
+                BFName =:= "sys" ->
+                    DName = filename:dirname(FName),
+                    {ok, SysEnv, Errors} =
```
tsloughter, line 1820: So including additional config files in a config is still restricted to only a file named `sys.config` and thus looks like its not supported in the config passed with `configfd`?

garazdawi: The problem is that we do not know where to look for the extra files when using `-configfd`. For `sys.config` it looks in the same directory as `sys.config`. I suppose if we could do it if we only allow absolute file names, or if we make the paths relative to the emulators cwd.

erts/doc/src/erl_cmd.xml
garazdawi: > Is there any expected order for configs to be read? Is it -config first then -configfd entries in declaration order?

The order is as declared on the command line. So `-config a -configfd 1 -config b` would read a.config, then fd 1, then b.config. However, `erl -kernel logger_level all` will always override anything given in a config file.

> Also what's the merge policy? You mention in the example setting kernel.logger_level twice, but what if one of the FDs is a setting of the logger and another one is say, shell history? Are they properly merged within each app definition, or are the settings per-app?

The merge policy is per configuration parameter, which is the same as today if you give many `-config`.

lib/kernel/src/application_controller.erl
tsloughter: Ah, didn't think about that. If it was done I would go for relative to the root of the current version of the release. So a config file in `releases/VSN/` can be referenced with `./`.

lib/kernel/src/application_controller.erl
kjellwinblad: You are right. `partion/2` does not do the right thing but `splitwith/2` does.

erts/etc/common/erlexec.c
kjellwinblad: Thanks for the question! Standard input (file descriptor 0) is a special case because the Erlang runtime system automatically tries to buffer its contents, if the user has not specified `-noinput`. This buffering conflicts with reading configuration data from file descriptor 0. There is no such conflict for 1 ( standard output) and 2 (standard error) so no checks are necessary for these file descriptors. However, the Erlang system will not be able to print to standard output if 1 is used as a configuration file descriptor (a configuration file descriptor is closed after the configuration has been read) so it is probably not a good idea to use 1 and 2 as arguments to `-configfd` in most cases.

erts/emulator/nifs/win32/win_prim_file.c
kjellwinblad: Thanks, I will disable the handler temporary.

kjellwinblad: Fixed: 62a29aa

erts/emulator/nifs/win32/win_prim_file.c
kjellwinblad: Fixed: 62a29aa

erts/emulator/nifs/win32/win_prim_file.c
kjellwinblad: Fixed: 62a29aa, 9472dbe

lib/kernel/src/application_controller.erl
kjellwinblad: Thanks for seeing this!

lib/kernel/src/application_controller.erl
garazdawi: We will add so that it will look for other files in `releases/VSN` if running in a release. Is there anywhere else that this could be used within relx/rebar3 so that more directories need to be added?

erts/etc/common/erlexec.c
ferd: Yeah sounds like something we'd want to put in docs, since I assume that using these values means we can't warn (either to display the fd0 warning, or to show warnings about fd1 or fd2 themselves)

lib/kernel/src/application_controller.erl
ferd: Not in what we natively support, but I could expect someone to use /tmp mounted to a ramfs or tmpfs partition to get writeable configs going in some embedded environments where they want to generate the script without doing it with fds. 

I would possibly assume something like:

1. start with the cwd of the release
2. look in the directory of the first config file passed

To be fair though, this is quickly becoming hella obtuse and I don't think it would be mandatory to support all of that as of today since I'm not sure we even do right now and people can still use absolute filenames for that.

garazdawi: > To be fair though, this is quickly becoming hella obtuse and I don't think it would be mandatory to support all of that as of today since I'm not sure we even do right now and people can still use absolute filenames for that.

Hmm, yes that is true. Maybe it is enough to only support absolute paths?

For releases I suppose they could use `$PWD/release/$VSN/` in the sys.config.src for any files that they need to include.

ferd: I'd make a non-absolute path relative to the release's ROOTDIR (or whatever it is we do today), and absolute has no problem regardless. That way you support both instantly.

tsloughter: Will environment variables work when used in a config file?

I would suggest doing the releases dir and absolute, I think that is enough and not going to be confusing.

erts/etc/common/erlexec.c
kjellwinblad: I agree. I will write something in the documentation to warn users about using fd0, fd1 and fd2. I don't understand what you mean with "either to display the fd0 warning" . What is the fd0 warning?

ferd: Uh my bad. I misremembered a warning about fd0, but it was only in docs.

lib/kernel/src/application_controller.erl
garazdawi: > Will environment variables work when used in a config file?

They will work as before, i.e. not at all. I was talking about the way relx helps with `extended_bin`.

> I would suggest doing the releases dir and absolute, I think that is enough and not going to be confusing.

Agreed.

erts/doc/src/erl_cmd.xml
kjellwinblad: I have added a description of the merge policy to the documentation.

See:

https://github.com/erlang/otp/pull/2873/commits/ee2920a5daac95f80e2b522db62acb7a3fc36d39#diff-b4ce526e7b5c99616ffad3906c61ed3f879a8290d5728b1926e991ca790e7f8eR43

erts/etc/common/erlexec.c
kjellwinblad: I have added a note in the documentation for `-configfd` about fd1 and fd2:

See: https://github.com/erlang/otp/commit/ee2920a5daac95f80e2b522db62acb7a3fc36d39#diff-9b388e03a9038a603ba9d7c1d1cb04d38bbdd341f4ef7b4a8452306d8fe8fb79R251

lib/kernel/src/application_controller.erl
kjellwinblad: I have decided to log warnings if a file descriptor is slow instead of timing out (after discussing this with other members of the Ericsson Erlang/OTP team). This approach does not cause any problems for users with very slow file descriptors but can still give the users feedback if they pass broken file descriptors.

See: https://github.com/erlang/otp/pull/2873/commits/fffa1a550e658950d1b9f667873de1f257cc973a

lib/kernel/src/application_controller.erl
garazdawi: So, when looking at implementing looking for config files in release dir we realized that we do not know what the release dir is when starting erlang. The release dir can be set using a sasl configuration parameter, which is not available during boot.

So I think that for now, we will have to make do with absolute paths and if really needed we need to add some command-line option to set the cwd for configfd config.

tsloughter: Does it know the location of the boot script being run?

garazdawi: Yes it does, that may work as an indicator...

tsloughter: Yea, in the case of relx generated releases that will be the same as the release dir `releases/VSN/`. Obviously it isn't guaranteed to be the same but if its documented as "cwd is the directory of the boot script being booted" instead of "the release directory" I'd think it would be fine.

ferd: One thing to note is that release_handler assumes a very specific path, and if you change it (unless the SASL thing you mentioned allows changes), relups start breaking. See https://bugs.erlang.org/browse/ERL-715

For the time being, assuming any other directory for configs sounds dangerous because it just breaks OTP functionality to change it, and as far as I know this isn't documented anywhere but has been in forever.

garazdawi: Yes, release upgrades will not work with `configfd` style config for the new version. I was thinking that maybe it would be a good idea to allow release handler to take the contents of a config file as an argument when doing an upgrade that does not involve en emulator restart. When doing an emulator restart, the heart command will have to deal with supplying the correct `-configfd` setup.

kjellwinblad: I have added support for including files. If a file path is relative, the path will first be searched relative to the directory where the boot script is located and then relative to the current working directory of the emulator.

lib/kernel/src/application_controller.erl
```diff
@@ -2015,6 +2015,7 @@ read_fd_until_end_and_close(FileDescriptorId,
                             TimeBetweenWarningsMilliseconds,
                             MaxSizeBytes) ->
     ReadLimit = 1024,
+    OSType = element(1, erlang:system_info(os_type)),
```
jhogberg, line 2018:
```suggestion
    {OSFamily, _} = os:type(),
```

PR #2875, opened 2020-11-19
ssl: Improve active emulation

IngelaAndin:
If using active once sockets and the socket is closed and all data is already
delivered it is desirable that ssl:setopts(S, [{active, once}] shall succeed
and an active close message be sent instead of ssl:setopts returning {error, closed}
This how gen_tcp/inet behaves.

Example of client side:

> {ok, S} = ssl:connect("localhost", 9999, [{active, false}]).
{ok,{sslsocket,{gen_tcp,#Port<0.10>,tls_connection,
                        undefined},
               [<0.120.0>,<0.119.0>]}}
> ssl:setopts(S, [{active, once}]).
ok
> flush().
Shell got {ssl,{sslsocket,{gen_tcp,#Port<0.10>,tls_connection,undefined},
                          [<0.120.0>,<0.119.0>]},
               "foo"}
ok
> ssl:setopts(S, [{active, once}]).
ok
> flush().
Shell got {ssl_closed,
              {sslsocket,
                  {gen_tcp,#Port<0.10>,tls_connection,undefined},
                  [<0.120.0>,<0.119.0>]}}
ok
> ssl:setopts(S, [{active, once}]).
{error,closed}
> flush().
ok

### DIFF
--- a/lib/ssl/src/ssl_connection.erl
+++ b/lib/ssl/src/ssl_connection.erl
@@ -2898,13 +2898,13 @@ ssl_options_list([{Key, Value}|T], Acc) ->
 handle_active_option(false, connection = StateName, To, Reply, State) ->
     hibernate_after(StateName, State, [{reply, To, Reply}]);
 
-handle_active_option(_, connection = StateName, To, _Reply, #state{static_env = #static_env{role = Role},
+handle_active_option(_, connection = StateName, To, Reply, #state{static_env = #static_env{role = Role},
                                                                    connection_env = #connection_env{terminated = true},
                                                                    user_data_buffer = {_,0,_}} = State) ->
     Alert = ?ALERT_REC(?FATAL, ?CLOSE_NOTIFY, all_data_deliverd),
     handle_normal_shutdown(Alert#alert{role = Role}, StateName, 
                            State#state{start_or_recv_from = To}),
-    {stop,{shutdown, peer_close}, State};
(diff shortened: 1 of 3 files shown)


IngelaAndin:
Updated implementation to avoid busy loop waiting.

bmk:
Is it necessarily a good idea to be "compatible" with gen_tcp?
What if you want to start using 'socket' directly?

IngelaAndin:
We would still want to have the active option even if we start using socket directly. However the user can should not count on that ssl:setopts will always return ok, neither does inet:setopts.  But as we anyway need the mechanism for the connection process to linger after the internal socket is closed by the peer, as there might be more data to deliver, we might as well linger so that we can send a close message to the processes doing ssl:setopts(S, [{active, once}]).  I think it is the least surprising thing to do.

bmk:
I was only talking about the gen_tcp behaviour "emulation" in this case.
Having an active feature is fine, since its conventent.
I think my point was that ok does not mean ok in that case. The actual result may instead come at "some" later time.

IngelaAndin:
I would find it more surprising if setopts returns {error, closed}  and then you get a closed message too.

bmk:
Ok. I thought we where discussing *either* alt A or alt B not both.

peterdmv (approved):

bmk (approved):

PR #2877, opened 2020-11-19
fix crypto hmac performance loss

mikpe:
Fixes the crypto hmac performance loss reported in [ERL-1400](https://bugs.erlang.org/browse/ERL-1400).

When upgrading from OTP-21 to OTP-22/23, the crypto hmac calls change from using openssl's low-level `HMAC()` interface to its higher-level `EVP_PKEY` interface. With openssl-1.0.x (as found in some enterprise environments), this causes each hmac call to require several reference count changes in openssl as pkey objects are accessed and released. openssl-1.0.x allows the client application to register an `add_lock` callback to atomically add to an integer in memory and return its new value. OTP does not register that callback, which causes openssl-1.0.x to fall back to taking and releasing a global pkey mutex in each reference count change. That mutex quickly causes exponential slowdown even with moderate concurrency. The solution is to implement and register the callback. With openssl-1.1.x this is not an issue since it implements this synchronization primitive itself.

With the above fix the exponential slowdown was cured, but we still saw a 2X slowdown compared to OTP-21. Profiling with gprof revealed that the new API has a lot of internal overheads from dynamic object allocation and deletion. A simpler interface not using EVP_PKEY is still present in crypto as `hmac_low_level()`, so we switched to that.

Even when using `hmac_low_level()` we saw a 2X slowdown compared to OTP-21, with both openssl-1.0.x and openssl-1.1.x. A comparison between the `hmac_nif()` in OTP-21 and `hmac_low_level()` in OTP-22/23 revealed the reason for that: the newer code calls `HMAC()` twice where the older code only called it once, which is enough since the size of the result is bounded by a small constant. Reinstating that logic in `hmac_low_level()` eliminated the final 2X slowdown.

### DIFF
--- a/lib/crypto/c_src/crypto.c
+++ b/lib/crypto/c_src/crypto.c
@@ -250,6 +250,7 @@ static int initialize(ErlNifEnv* env, ERL_NIF_TERM load_info)
 #if OPENSSL_VERSION_NUMBER < PACKED_OPENSSL_VERSION_PLAIN(1,1,0)
 #ifdef OPENSSL_THREADS
     if (nlocks > 0) {
+	CRYPTO_set_add_lock_callback(ccb->add_lock_function);
 	CRYPTO_set_locking_callback(ccb->locking_function);
 	CRYPTO_set_id_callback(ccb->id_function);
 	CRYPTO_set_dynlock_create_callback(ccb->dyn_create_function);
--- a/lib/crypto/c_src/crypto_callback.c
+++ b/lib/crypto/c_src/crypto_callback.c
@@ -133,6 +133,28 @@ static INLINE void locking(int mode, ErlNifRWLock* lock)
 }
 
 #if OPENSSL_VERSION_NUMBER < PACKED_OPENSSL_VERSION_PLAIN(1,1,0)
+
+/* TODO: there should be an enif_atomic32_add_return() */
+
+typedef int (*add_lock_function_t)(int *var, int incr, int type, const char *file, int line);
+
+#if defined(__GNUC__) && defined(__ATOMIC_ACQ_REL)
+static int add_lock_function(int *var, int incr, int type, const char *file, int line)
+{
+    return __atomic_add_fetch(var, incr, __ATOMIC_ACQ_REL);
+}
+
+static add_lock_function_t get_add_lock_function(void)
+{
+    return __atomic_always_lock_free(sizeof(int), NULL) ? add_lock_function : NULL;
+}
+#else
+static add_lock_function_t get_add_lock_function(void)
+{
+    return NULL;
+}
+#endif
+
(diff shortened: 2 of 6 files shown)


HansN:
First I would like to thank you @mikpe for the findings, the tests and the PRs.  This one worked out of the box on all our nightly tests architectures, OSes and cryptolib versions which is uncommon.

My only concern is that you removed the EVP calls completely.  I added them to enable usage of (hardware) acceleration that is possible in later OpenSSL releases.  It was also asked for on the Erlang questions list some years ago. In the coming OpenSSL 3.0, the low-level (non-EVP) API is deprecated.

There was a similar problem with EVP and DH for older versions of OpenSSL on some architectures this summer, and we solved it by a configuration flag --disable-evp_dh  in crypto/configure.in that sets the symbol DISABLE_EVP_DH=1.  Then the low-level api is always used.  In case of OpenSSL < 1.0.2 it is always set.

So I would prefer an analogue solution with a --disable-evp-hmac.  In fact I have such a one already, so I could merge your branch with mine on top (keeping your one-call only) and let EVP always be disabled if < 1.0.2 or something. Any comments?

mikpe:
A `--disable-evp-hmac` configure option would work for us. I understand that this only buys us time (a few years perhaps).

HansN:
Good!

Regarding OTP we still support usage of OpenSSL-0.9.8c which was replaced about 20 years ago!  So this non-EVP support will most probably stay decades in OTP crypto app.  That only 1.1.1 is maintained by OpenSSL is another thing.  We always recommend using versions which receives security patches, but you know the importance of that better than me.

I'll merge this PR later this week with the --disable-evp-hmac flag on top of it.

HansN:
Merged with an --disable-evp-hmac config option. EVP HMAC is always disabled for OpenSSL < 1.0.2
Thanks for the PR!

benbro:
Are all changes only relevant when EVP HMAC is disabled? For example isn't the 2X slowdown relevant for EVP HMAC?

HansN:
There are two sources to the speed-up.  One is a fix of a lock problem, the other is to use HMAC low-level functions including a fix of that code with a 2X speedup.  The flag only affects whether to use low-level or not. --disable-evp-hmac  enables the low-level functions, and it is always automatically added for OpenSSL versions < 1.0.2.  If, and how much the lock fix improves depends on the OpenSSL version, the processor architecture, number of cores and mostly by the load.
The reason to keep the EVP variant is that EVP makes it possible to use HW-acceleration.  With the low-level functions that is not possible.
Today is only version 1.1.1 maintained with security fixes by OpenSSL, and 3.0 is on its way out.  From 3.0 the low-level routines are deprecated so if we removed the EVP now we would have been forced to add them again in a couple of years.

PR #2880, opened 2020-11-23
Raise nicer exceptions for bad generator and filters in comprehensions

bjorng:
This pull request introduce a nicer `{bad_generator,Generator}` exception when a generator for list or binary comprehension has the wrong type, and a `{bad_filter,Filter}` exception when a filter does not evaluate to a boolean. This PR also fixes some comprehension bugs and refactor and simplifies compiler code.


lib/compiler/src/beam_ssa_bc_size.erl
```diff
@@ -0,0 +1,596 @@
...
+join_bs_1([], _Bigger, Smaller) -> Smaller.
+
+%%%
+%%% Turn the result of the traversal of the SSA code into an expression tree.
+%%%
+
+make_expr_tree({{call,Alloc0},GenAnno}) ->
+    {Alloc1,Annos} = make_expr_tree_list(Alloc0, none, none, [GenAnno]),
+    Alloc2 = opt_expr(Alloc1),
+    Alloc = case divisible_by_eight(Alloc2) of
```
jhogberg, line 372: I think you should add a comment or break this out into its own function. At a first glance it's not obvious we're rounding to the next byte.

lib/compiler/src/beam_ssa_type.erl
```diff
@@ -1349,6 +1361,10 @@ eval_type_test_bif(I, Op, Types) ->
             I
     end.
 
+t_is_number(#t_integer{}) -> true;
```
jhogberg, line 1364: I'd prefer this were added to `beam_types` where we already have similar functions. `is_numerical_type` would fit right in.

jhogberg (approved):
Looks good to me aside from a few minor nits.

lib/compiler/src/beam_ssa_bc_size.erl
```diff
@@ -0,0 +1,596 @@
...
+    end;
+opt_expr_1({bif,'div'}=Op, [Numerator,#b_literal{val=Denominator}]=Args) ->
+    try
+        opt_expr_div(Numerator, Denominator)
+    catch
+        throw:not_possible ->
+            case Denominator band (Denominator - 1) of
+                0 ->
+                    %% The denominator is a power of two.
+                    {{bif,'bsr'},[Numerator,#b_literal{val=log2(Denominator)}]};
```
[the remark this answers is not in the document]
bjorng: I'll use `round(math:log2(Denominator))`.

bjorng:
I have updated this PR to also raise `{bad_filter,Filter}` exceptions for bad filters.

PR #2883, opened 2020-11-25
Drop the erl_tidy and igor modules from syntax_tools

richcarl:
These can now be found at https://github.com/richcarl/erl_tidy and
https://github.com/richcarl/igor, respectively.


leoliu:
Could you explain a bit on the reason behind this change? Thanks.

richcarl:
> Could you explain a bit on the reason behind this change? Thanks.

It's been requested by the OTP team and others, since these modules are pretty different from the main functionality of the syntax_tools library and are not needed for most uses of syntax_tools. Also, as separate applications, they can be more easily improved on by themselves without touching the OTP distribution.

bjorng:
Unfortunately, when we started to check whether any of our customers actually used `igor` or `erl_tidy`, it turned out that one project uses `erl_tidy:file/2` for pretty printing. That means that we might have to postpone the removal of `erl_tidy` to a later release (or forever).

To avoid having this pull request entering a limbo state, it is probably best to change it to only remove `igor`. We can merge that change to `master` very soon.

Here is my pull request that deprecates `igor` and part of `erl_tidy`: https://github.com/erlang/otp/pull/2902.

richcarl:
Sounds good. I had also prepared a branch after this one to move the "prettypr" module from syntax_tools to the "tools" app - see https://github.com/erlang/otp/compare/master...richcarl:move-prettypr - but there seems to be some problems with dependencies during bootstrapping that I'm not sure how to resolve. Maybe erl_tidy could be moved along with prettypr to tools for OTP24, to avoid having both syntax_tools and tools as bootstrat deps.

bjorng:
Turns out that the customer will stop using `erl_tidy`. I have created #3016 to deprecate all functions in `erl_tidy` in OTP 23.3.

bjorng:
I have force-pushed an update to this branch where I have done the following:

* Rebased on the lastest master
* Kept the `erl_tidy` and `igor` modules, but without all code deleted and a `removed` attribute added
* Added a commit on top that updates `otp_internal`

I have added the resulting branch to our daily builds.

garazdawi:
> Kept the erl_tidy and igor modules, but without all code deleted and a removed attribute added

Will this make erlc issue a warning that the module is removed?

bjorng:
> Will this make erlc issue a warning that the module is removed?

Yes.

richcarl:
> > Will this make erlc issue a warning that the module is removed?
> 
> Yes.

Could this cause any problems for those using erl_tidy or igor from the new separate repos?

bjorng:
> Could this cause any problems for those using erl_tidy or igor from the new separate repos?

Yes, it would. I suppose we will have to remove the modules from OTP and the warnings along with them.

bjorng:
I have now force-pushed another update.

The `igor` and `erl_tidy` modules are now completely gone and `erlc` will emit no warnings for using those modules. There will be a note about the removal on the "Remove Functionality" page in the documentation (which is a new page in OTP 24).

PR #2886, opened 2020-11-26
Add data_only option to binary_to_term

michalmuskala:
This option allows enforcing that only "regular data" terms
are decoded. Forbidden terms include funs, but also pids, ports,
and references. This makes use of ETF somewhat safer for generic data
interchange, where node-specific terms (like ports, pids, and references)
or funs that carry instructions alongside data, are not needed.

See: https://www.alphabot.com/security/blog/2020/elixir/Remote-code-execution-vulnerability-in-Elixir-based-Paginator-project.html

---

This is only a PoC showing the general idea. It lacks documentation,
and thorough testing, but I wanted to open this early to get feedback
on the feature from the OTP team. In particular, it might make more
sense to have more granular options, e.g. `{without, [fun, ref, port, pid]}`
alongside the generic option.
Additionally, I think it would make sense to deprecate the `safe` option
in favour of something more descriptive - e.g. `only_existing_atoms`,
given the option doesn't really make the decoder safe - there are still
possible safety issues to guard against.

\cc @josevalim

### DIFF
--- a/erts/emulator/beam/atom.names
+++ b/erts/emulator/beam/atom.names
@@ -200,6 +200,7 @@ atom current_function
 atom current_location
 atom current_stacktrace
 atom data
+atom data_only
 atom debug_flags
 atom decentralized_counters
 atom decimals
--- a/erts/emulator/beam/erl_nif.c
+++ b/erts/emulator/beam/erl_nif.c
@@ -1408,13 +1408,21 @@ size_t enif_binary_to_term(ErlNifEnv *dst_env,
     Sint size;
     ErtsHeapFactory factory;
     const byte *bp = (byte*) data;
+    Uint32 Xopts = (Uint32)opts;
     Uint32 flags = 0;
 
-    switch ((Uint32)opts) {
-    case 0: break;
-    case ERL_NIF_BIN2TERM_SAFE: flags = ERTS_DIST_EXT_BTT_SAFE; break;
-    default: return 0;
+    if (Xopts & ERL_NIF_BIN2TERM_SAFE) {
+        flags |= ERTS_DIST_EXT_BTT_SAFE;
+        Xopts &= ~ERL_NIF_BIN2TERM_SAFE;
     }
+    if (Xopts & ERL_NIF_BIN2TERM_DATA_ONLY) {
+        flags |= ERTS_DIST_EXT_BTT_DATA_ONLY;
+        Xopts &= ~ERL_NIF_BIN2TERM_DATA_ONLY;
+    }
+    if (Xopts != 0) {
+        return 0;
+    }
+
     if ((size = erts_decode_ext_size(bp, data_sz)) < 0)
         return 0;
 
--- a/erts/emulator/beam/erl_nif.h
+++ b/erts/emulator/beam/erl_nif.h
@@ -258,7 +258,8 @@ typedef enum {
 } ErlNifUniqueInteger;
 
 typedef enum {
-    ERL_NIF_BIN2TERM_SAFE = 0x20000000
+    ERL_NIF_BIN2TERM_SAFE = 0x20000000,
+    ERL_NIF_BIN2TERM_DATA_ONLY = 0x40000000
 } ErlNifBinaryToTerm;
 
 typedef enum {
--- a/erts/emulator/beam/external.c
+++ b/erts/emulator/beam/external.c
@@ -1251,7 +1251,7 @@ Eterm erts_decode_ext(ErtsHeapFactory* factory, const byte **ext, Uint32 flags)
 	return THE_NON_VALUE;
     }
     if (flags) {
-        ASSERT(flags == ERTS_DIST_EXT_BTT_SAFE);
+        ASSERT((flags & ~ERTS_DIST_EXT_BTT_SAFE) & ~ERTS_DIST_EXT_BTT_DATA_ONLY);
         ede.flags = flags; /* a dummy struct just for the flags */
         ede.data = NULL;
         edep = &ede;
@@ -2076,6 +2076,9 @@ BIF_RETTYPE binary_to_term_2(BIF_ALIST_2)
         opt = CAR(list_val(opts));
         if (opt == am_safe) {
             ctx.flags |= ERTS_DIST_EXT_BTT_SAFE;
+        }
+        else if (opt == am_data_only) {
+            ctx.flags |= ERTS_DIST_EXT_BTT_DATA_ONLY;
         }
         else if (opt == am_used) {
             ctx.used_bytes = 1;
@@ -4268,6 +4271,9 @@ dec_term_atom_common:
 	    }
         case PID_EXT:
         case NEW_PID_EXT:
+            if (edep && (edep->flags & ERTS_DIST_EXT_BTT_DATA_ONLY)) {
+                goto error;
+            }
 	    factory->hp = hp;
 	    ep = dec_pid(edep, factory, ep, objp, ep[-1]);
 	    hp = factory->hp;
@@ -4277,6 +4283,9 @@ dec_term_atom_common:
 	    break;
         case PORT_EXT:
         case NEW_PORT_EXT:
+            if (edep && (edep->flags & ERTS_DIST_EXT_BTT_DATA_ONLY)) {
+                goto error;
+            }
 	    {
 		Eterm sysname;
 		ErlNode *node;
@@ -4377,6 +4386,9 @@ dec_term_atom_common:
 		ep += 4;
 
 	    ref_ext_common:
+                if (edep && (edep->flags & ERTS_DIST_EXT_BTT_DATA_ONLY)) {
+                    goto error;
+                }
 
 		if (ref_words > ERTS_MAX_REF_NUMBERS)
 		    goto error;
@@ -4589,6 +4601,9 @@ dec_term_atom_common:
 		break;
 	    }
 	case EXPORT_EXT:
+            if (edep && (edep->flags & ERTS_DIST_EXT_BTT_DATA_ONLY)) {
+                    goto error;
+            }
 	    {
 		Eterm mod;
 		Eterm name;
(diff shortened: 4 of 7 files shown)


okeuday:
Would be nice to have `{without, [fun, ref, port, pid]}` as described instead of the current approach.  I would like to still allow pids but block funs and other things like refs/ports.  Could add `new_atoms` to the without list instead of the current `safe` option or the proposed `only_existing_atoms` option.

mikpe:
I agree with @okeuday 's comment above. First, `data_only` as a name doesn't match what it does (all the things it rejects _are_ data). Second, I disagree with placing the filtering here; the bug if any lies in Elixir or application code blindly executing stuff received from the outside world, preventing transport of such stuff is just a workaround for that bug. Third, the correct way of preventing receipt of such stuff is to add negative filters to `binary_to_term/2`, for instance I can see the value in preventing creation of new atoms. As for the proposed rejection of references, I honestly don't see the point of that (a reference doesn't _do_ anything, and is isomorphic to a unique integer).

josevalim:
> Second, I disagree with placing the filtering here; the bug if any lies in Elixir or application code blindly executing stuff received from the outside world, preventing transport of such stuff is just a workaround for that bug.

Although we are working on fixing the root cause, in practice you want to add as many safety nets as you can. If you are certain you should only receive certain terms, then being able to forbidden certain entries, as proposed here, is very valuable.

For example, in the case linked above, I would say they should also have signed the payload. If they signed it and listed only the terms they expected, they would not have been hit by the Elixir bug. If there is ever a bug in binary_to_term, the other two protections would have helped them. If they sign/encrypt it and their key leaks, then they have other two layers, and so on.

The filtering can also be done *after* calling binary_to_term, but I think having a higher-level API that makes it easy can be a net positive. :) I agree with the comments about data_only not being ideal. There is also an option for explicitly listing the terms you expect (instead of forbidding the ones you don't want).

okeuday:
@mikpe A reference in the binary term format still has a node encoded as a node.  The means the node is consuming an atom when the reference is decoded, even if the node is irrelevant to the machine decoding the reference, unless the reference decoding doesn't decode the node as an atom (I haven't checked the source code).  Trying to limit the number of odd node-related atoms that appear is good to avoid consuming excessive amounts of atoms during a long period of time (the [nodetool](https://github.com/CloudI/CloudI/blob/6015aa82172ce44daea15210abf6f38dad6e08b1/src/rel/files/nodetool.in#L18) release helper script is a good example of where this is beneficial).

mikpe:
I can confirm that `binary_to_term` of an encoded reference does enter its node in the atom table, but a "no new atoms" (or "only existing atoms") filter should cover that.

essen:
There's value in having filtering in both encode and decode steps. For example an encoder might want to crash before sending unexpected types such as funs or references. Similarly, a decoder might want to crash instead of decoding data containing unexpected types.

Such filtering can be done either in binary_to_term/term_to_binary, or in user code, but in the later case you end up going over the data twice and probably generate more garbage when doing so after decoding because the entire binary has already been converted to a term.

I very much like `{without, [fun, ref, port, pid]}` but it should be extended to cover all types. One may not want atoms for example, since those don't necessarily have an equivalent in other languages.

About the reference atom being covered that's true, but one may also not want to expose the node name by mistakenly sending a reference.

okeuday:
@essen If all types are supported, it would be best to have `with` instead of `without` even though it is rare that a type is added to Erlang.  Using `with` to allow types would avoid any need to modify the `without` list in the source code if types are added to Erlang and it would help to provide visibility to the Erlang binary term data (allowing the source code viewer to compare the `with` list to the data being matched).

sverker:
First, to clarify any remaining confusion about the existing `safe` option I would like to quote the documentation of binary_to_term/2:

_"This prevents creation of new atoms directly, creation of new atoms indirectly (as they are embedded in certain structures, such as process identifiers, refs, and funs), and creation of new external function references. None of those resources are garbage collected, so unchecked creation of them can exhaust available memory."_

So, `safe` will reject new atoms in any form and export funs (`fun Mod:Func/Arity`) of not (yet) loaded functions.

Second, no ETF binary contains instructions to be executed, not evens funs which only contain a reference to the function. That is, the code must already exist on the receiving server for a blindly executed fun to have any effect.

Third, each new option to reject a certain type of term will add a branch in the code for binary_to_term to be executed by every invocation. No free lunches. And for many (most?) cases, don't you let the decoded term go through some sort of parsing anyway which will fail if it does not conform to an agreed upon format.

I personally consider this more of a convenient feature and am a bit sceptical if it really is worth the price.

michalmuskala:
> Second, no ETF binary contains instructions to be executed, not evens funs which only contain a reference to the function. That is, the code must already exist on the receiving server for a blindly executed fun to have any effect.

This is not really true if one assumes `erl_eval` exists on the node. Using it, a fun executing arbitrary code can be relatively easily crafted and sent to the node.

essen:
Convenience is very much what is interesting here. Would the extra bitwise operation and `if` test degrade performance that much? Sounds negligible compared to actually decoding terms.

mikpe:
> Convenience is very much what is interesting here. Would the extra bitwise operation and `if` test degrade performance that much? Sounds negligible compared to actually decoding terms.

You're asking the vast majority of binary_to_term calls which to not want this overhead to pay it anyway for the benefit of a few that want this. That's not a good trade-off, especially since those who want it can easily get it by combining the `safe` option with their own post-processing of the raw term.
No-one is arguing against input validation, the issue is where to place it and who to burden.

okeuday:
@sverker If the source code changed in the `dec_term` function https://github.com/erlang/otp/blob/810d6797c6561a72ac30c959938f031da8df8f24/erts/emulator/beam/external.c#L4591-L4626 to check boolean flags that represent the `without` list, there shouldn't be a major slowdown, it would just be part of the switch branches that already need to occur for decoding terms.

essen:
> > Convenience is very much what is interesting here. Would the extra bitwise operation and `if` test degrade performance that much? Sounds negligible compared to actually decoding terms.
> 
> You're asking the vast majority of binary_to_term calls which to not want this overhead to pay it anyway for the benefit of a few that want this. That's not a good trade-off, especially since those who want it can easily get it by combining the `safe` option with their own post-processing of the raw term.
> No-one is arguing against input validation, the issue is where to place it and who to burden.

I had originally started writing a comment suggesting a solution that would not have this overhead for those that don't want it, but then noticed the check is really small. So I ask my question again, would this extra check actually degrade performance noticeably? It's not an expensive check like `safe` where it has to call other functions to double check things, it's just a bitwise operation like what `data_only` is already doing in this PR.

sverker:
> ..., would this extra check actually degrade performance noticeably? It's not an expensive check like `safe` where it has to call other functions to double check things, it's just a bitwise operation like what `data_only` is already doing in this PR.

It would be the same kind of bitwise operation as `safe` does for atoms (no double checks there). But with full type checking it will be done in every switch case in every loop round in code that can be quite hot on behalf of callers that on most cases don't care.

I'm not saying it will noticeably degrade performance (probably not), but I think it's good with a healthy resistance against putting new features especially in hot code paths.

max-au:
I believe ETF, despite the name, is only expected to be used to communicate between Erlang processes. Suggested improvements of `binary_to_term` may create false sense of security and provoke developers to use ETF in a way it was not designed to be used. For example, to communicate with clients (similar to protobuf/Thrift/JSON).

There are no security benefits either. Malicious agent can craft a system message to any `gen_server` guaranteed to be running, e.g. `kernel_sup` (frankly we use this for white-box testing, and implementing things like concurrent application shutdown, abusing `sys:replace_state`).

I think it would be better to follow security guidelines, and avoid using ETF for communications crossing Erlang cluster border (which is a security perimeter).

okeuday:
@max-au Discouraging the use of the Erlang Term Format doesn't help avoid the problem, which isn't necessarily a security problem.  If Erlang terms need to be stored in a binary format for a long period of time, it would be good to block these types (`fun`, `ref`, `port`, `pid`) because they are temporary and depend on the current state of the system.  That would allow more dependable state to be used that doesn't run-into problems, even if those problems may seem odd.

For example, a `fun` could contain the use of private module functions that become invalid with a new module version (it would be ok if the `fun` only used exported functions).  That is just a basic system-consistency problem that you can run into, if you are attempting to store Erlang terms as binary data for an undefined period of time before it is used again as Erlang terms.  Sometimes you need to care about the system changing while the data is static, with some number of updates and restarts involved before the data is decoded in the future.

essen:
@max-au Using the ETF for non-Erlang stuff is something that has been done for a long time. BERT was a compatible subset. Hex uses it in HTTP payloads. Discord probably uses it [somewhere](https://github.com/discord/erlpack) as well. And [many others](https://github.com/search?q=erlang+term+format).

So at the very least the format and the compatibility with `term_to_binary` and `binary_to_term` is interesting. The question is whether these functions should provide finer grained options. Perhaps the right answer is to have a separate Erlang implementation of ETF that does.

Based on the arguments so far I'm leaning toward a separate Erlang implementation.

okeuday:
@essen @sverker If an Erlang-only ETF implementation was added to Erlang/OTP and maintained alongside the C implementation, it may provide a good comparison for JIT optimizations, ideally showing that runtime optimizations can exceed static analysis optimizations.  That would avoid any concerns about slowing down the C source code.  That could allow VM startup to switch from the C implementation to the Erlang implementation.

sverker:
Closed after decision from OTP Technical Board.
"We want a better definition of what problem to solve, this is only a solution for a diffuse problem. We don't see any security problems in Erlang because of the current behavior. Erlang is not executing anything that get decoded. If filtering it could be other things to filter as well , like maxlength of strings and binaries."

josevalim:
Thanks @sverker! I agree with the assessment.

I also think we adding some disclaimers to the docs is a net positive. In particular, the `safe` option means decoding binaries is `safe` for the Erlang runtime but it doesn't mean it is safe for the user code. For example, if the data contains `#{role := user}`, the untrusted source can rewrite it to `#{role := admin}` and gain unwarranted access.

I 100% agree it is not the Erlang VM fault and it would be a silly mistake to make - but I think warning about said cases does not hurt either. I will submit a PR.

Renaming the option `safe` to something more explicit like `existing_atoms_only` can also be a positive change, but that's a separate discussion. :)

PR #2887, opened 2020-11-27
Added gen_server:multi_call optimisation for single local node case.

ssmyczynski:
This PR adds an optimisation in `gen_server:multi_call` function in case when this function is used in a single node environments or more general when the only node that is the target of multi call is the local node.

Note that `gen_server:multi_call` is used in many places (in `global` module, in `pg2` module and indirectly also in `mnesia`), and its disadvantage over `gen_server:call` (in single local node case) is that it does not leverage the recv_mark optimisation.
It means that when caller is under pressure (having big outstanding message queue lenght) it is much slower than normal `call`.

A simple demonstration of effect of lack of this optimisation is here:
```erlang
1> timer:tc(fun() -> global:set_lock({mnesia_adjust_log_writes, self()}, [node()], 1) end).      
{82,true}
2> lists:foreach(fun(I) -> self() ! {{dummy, <<"message">>}, I} end, lists:seq(1,100000)).
ok
3> timer:tc(fun() -> global:set_lock({mnesia_adjust_log_writes, self()}, [node()], 1) end).
{5138,true}
```

While with using direct `gen_server:call`:
```erlang
1> timer:tc(fun() -> gen_server:call(global_name_server, {set_lock, {mnesia_adjust_log_writes, self()}}) end). 
{64,true}
2> lists:foreach(fun(I) -> self() ! {{dummy, <<"message">>}, I} end, lists:seq(1,100000)).                     
ok
3> timer:tc(fun() -> gen_server:call(global_name_server, {set_lock, {mnesia_adjust_log_writes, self()}}) end).
{50,true}
```

The above mentioned example is inspirated by the code in `mnesia_dumper` module:
```erlang
adjust_log_writes(DoCast) ->
    Token = {mnesia_adjust_log_writes, self()},
    case global:set_lock(Token, [node()], 1) of
      ...
```
which occasionally fires inside of `mnesia_tm` process - and from agility of this process depends the overall speed of `mnesia` :)

Well - in fact the background of this PR are the problems I observed on production environment with mnesia latencies, but I think that optimising `gen_server:multi_call` not only solves that problem but also could benefit in other usage cases.

### DIFF
--- a/lib/stdlib/src/gen_server.erl
+++ b/lib/stdlib/src/gen_server.erl
@@ -491,6 +491,15 @@ do_send(Dest, Msg) ->
     end,
     ok.
 
+do_multi_call([Node], Name, Req, infinity) when Node =:= node() ->
+    % Special case when multi_call is used with local node only.
+    % In that case we can leverage the benefit of recv_mark optimisation
+    % existing in simple gen:call.
+    try gen:call(Name, '$gen_call', Req, infinity) of
+        {ok, Res} -> {[{Node, Res}],[]}
+    catch exit:_ ->
+        {[], [Node]}
+    end;
 do_multi_call(Nodes, Name, Req, infinity) ->
     Tag = make_ref(),
     Monitors = send_nodes(Nodes, Name, Tag, Req),


rickard-green:
By utilizing the middleman process also in the case with timeout set to infinity would give a similar effect regardless of the amount of connected nodes. There will be an overhead spawning the middle-man and dispatching the request and responses over it, but it wont suffer from having to scan a huge message queue. This since the middle-man won't have a huge message queue, and the response from the middle-man to the caller utilize the recv_mark optimization. I would prefer such a solution compared to only optimizing the single node special case.

ssmyczynski:
> By utilizing the middleman process also in the case with timeout set to infinity would give a similar effect regardless of the amount of connected nodes.

You are right. I committed appropriate change. But I left single node case as an extra optimised one.

rickard-green:
Two test-cases in `gen_server_SUITE` fail with this change (`send_request` and `call_remote1`).

ssmyczynski:
Indeed, it is because global:set_lock uses global:multi_call with infinity and the lock is immediately released when middleman process dies ... it is a hidden drawback of using middleman process that called gen_server cannot rely on From field in its handle_call.

rickard-green:
Thanks for the debugging! I think we'll go for your first approach of just optimizing the local infinity case. Please squash your first two commits into one commit and drop the third commit.

In the end we'll rewrite this using the new [process alias](https://github.com/erlang/eep/blob/master/eeps/eep-0053.md) feature (in order to avoid late replies) together with an improved recv_mark optimization which will be able to handle multiple replies using the same reference. This will however not be possible until earliest OTP 24.

ssmyczynski:
Both features - process alias & improved recv_mark - sounds fantastic. Can't wait for them.

PR #2889, opened 2020-11-27
Remove potential lag from mnesia_log:log/slog functions

ssmyczynski:
This PR addresses the very rare but unfortunately existing problem with occassional latencies (up to 250ms) in mnesia operations described in more detail here:
https://bugs.erlang.org/browse/ERL-1423


dgud:
Nice find

PR #2890, opened 2020-11-28
Support IEEE 754-2008 16-bits floats in bitstrings

josevalim:
This format is often used with computer graphics and
AI, also in cases where data is large enough that the
decreased bandwidth and storage compensates the loss
of precision.

The _Float16 data type is defined in the C11 extension
ISO/IEC TS 18661-3:2015. Recent compilers and platforms
provide support for it and recent CPUs provide specific
instruction sets for them too. In some, _Float16 is storage
only, which is enough for our use cases. If _Float16 is not
available, functions for conversion from float to half-floats
are used, ported from https://github.com/Maratyszcza/FP16.

I am sending this pull request to collect feedback. I will
work on a complete test suite if there is interest in moving
this forward.


erts/emulator/beam/erl_bits.c
```diff
@@ -40,6 +40,209 @@
...
+#else
+typedef uint16_t erlfp16;
+
+#define FP16_FROM_FP64(x) (fp16_ieee_from_fp32_value((float) x))
+#define FP16_TO_FP64(x) ((double) fp16_ieee_to_fp32_value(x))
+
+/*
+ * C-implementation of Float16 - https://github.com/Maratyszcza/FP16
+ *
+ * The MIT License (MIT)
```
jhogberg, line 57: This notice needs to be added in `system/COPYRIGHT` as well.

jhogberg (commented):
Thanks for the PR! I think it's a pretty reasonable feature, we'll take it up on our next technical board meeting.

erts/emulator/beam/erl_bits.c
josevalim: Will do. Btw, how would you handle the separate file? Would it be a erl_bits_fp16.h that you just include at the top?

jhogberg: Yep.

jhogberg:
Sorry about the late reply, I missed the technical board meeting last month due to illness. We're interested in having this :)

josevalim:
Yes!!! I will work on the test suite today! Thanks for the updates! :)

josevalim:
Tests added and feedback addressed!

erts/emulator/beam/erl_bits.c
```diff
@@ -40,6 +40,17 @@
 
 #define BIT_IS_MACHINE_ENDIAN(x) (((x)&BSF_LITTLE) == BIT_ENDIAN_MACHINE)
 
+#if (SIZEOF__FLOAT16 == 2)
+typedef _Float16 erlfp16;
+#define FP16_FROM_FP64(x) ((_Float16) x)
+#define FP16_TO_FP64(x) ((double) x)
+#else
+typedef uint16_t erlfp16;
```
jhogberg, line 48: This seems to break our Windows builds, can you change `uint16_t` to `Uint16`?

josevalim: Oops, fixed! I have also rebased to fix a tiny conflict on erl_lint.

jhogberg: Perfect, thank you. :)

lib/stdlib/src/erl_lint.erl
```diff
@@ -2009,13 +2009,19 @@ bit_size_check(Anno, Size, #bittype{type=Type,unit=Unit}, St) ->
...
 elemtype_check(_Anno, float, 32, St) -> St;
 elemtype_check(_Anno, float, 64, St) -> St;
 elemtype_check(Anno, float, _Size, St) ->
     add_warning(Anno, {bad_bitsize,"float"}, St);
 elemtype_check(_Anno, _Type, _Size, St) ->  St.
 
 
+%% add_bit_size(Line, ElementSize, BinSize, Build, State) -> {Size,State}.
+%%  Add bits to group size.
+>>>>>>> Support IEEE 754-2008 16-bit floats in bitstrings
```
garazdawi, line 2022: Seems like you have committed part of the merge?

josevalim: GAAAAAAAAAAAAH. Thank you! Fixed.

jhogberg:
The `debugger` test case `bs_match_misc_SUITE:t_float/1` has started to fail because 16-bit floats no longer crash, can you change it to try matching 15 bits instead?

josevalim:
@jhogberg done. I changed it to 8 which is what I did in the other similar suites.

erts/emulator/beam/erl_bits_f16.h
```diff
@@ -0,0 +1,193 @@
...
+ * or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+static inline float fp32_from_bits(uint32_t w) {
```
jhogberg, line 27: The Windows build fails here too, can you change `inline` to `ERTS_INLINE`?

josevalim: Done!

erts/emulator/beam/erl_bits_f16.h
```diff
@@ -0,0 +1,193 @@
...
+ * or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+ * INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR
+ * PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE
+ * FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+ * OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+static ERTS_INLINE float fp32_from_bits(uint32_t w) {
```
jhogberg, line 27: Sorry about the back and forth, but we need to change all instances of `uint??_t` in this file too. :(

josevalim: No problem! I should probably get a Windows machine next time I am working on this part of ERTS. :)

I have done the requested changes. I have also added a short notice at the top of this file (below the copyright disclaimer) to list all of the modifications we have done. This will make it easier to track changes or update the code in the future.

erts/emulator/beam/erl_bits_f16.h
```diff
@@ -0,0 +1,202 @@
...
+    if (bias < UINT32_C(0x71000000)) {
+	bias = UINT32_C(0x71000000);
+    }
+
+    base = fp32_from_bits((bias >> 1) + UINT32_C(0x07800000)) + base;
+    bits = fp32_to_bits(base);
+    exp_bits = (bits >> 13) & UINT32_C(0x00007C00);
+    mantissa_bits = bits & UINT32_C(0x00000FFF);
+    nonsign = exp_bits + mantissa_bits;
+    return (sign >> 16) | (shl1_w > UINT32_C(0xFF000000) ? UINT16_C(0x7E00) : nonsign);
```
jhogberg, line 86:
```
[2021-02-03T10:53:10.718Z] erl_bits.o : error LNK2019: unresolved external symbol UINT32_C referenced in function erts_bs_get_float_2
[2021-02-03T10:53:10.718Z] erl_bits.o : error LNK2019: unresolved external symbol UINT16_C referenced in function fp16_ieee_from_fp32_value
```

josevalim: Thanks @jhogberg! Which platform is it coming from? It seems those macros are part of the C99 standard, so I am wondering why they are not available. Although uint16_t is part of the C99 standard too.

josevalim: Follow up: do we support any 16bit platforms? My understanding is that this will matter only for those platforms, otherwise adding a U suffix would be enough.

jhogberg: A relatively old version of MSVC that doesn't fully support C99 yet. We don't support any 16-bit platforms so it should be safe to use a `U` prefix.

josevalim: Updated! I also re-ran the added tests and they continue to pass.

jhogberg: Thank you, it builds fine now :)

erts/emulator/beam/erl_bits.c
```diff
@@ -1132,17 +1178,21 @@ erts_new_bs_put_float(Process *c_p, Eterm arg, Uint num_bits, int flags)
 #ifdef WORDS_BIGENDIAN
```
jhogberg, line 1178: `bs_construct_SUITE:fp16/1` fails on our big endian machine, can you look into that?

josevalim: Can you please post the failure if handy? I will try to figure out how to run this patch on a big endian machine. :)

jhogberg: The first line of the test fails with `{badmatch,false}`, it doesn't say much more :/

I think the lines below are the culprit as they will always be zero when given an fp16, if I'm reading the code right you want to write the first 16 bits instead:

```c
	    t[0] = a >> 24;
	    t[1] = a >> 16;
```

josevalim: Ah, good point, thanks! I believe I have addressed your changes. I kept it as a separate commit for now as it also has some debugging information for the tests. If it works, then I can squash and we can discuss if we should keep the debugging info or no.

jhogberg: Thanks, the fix looks a bit strange though: why set `t[2]` and `t[3]` for a 16-bit value? Shouldn't it be `t[0]` and `t[1]`?

```suggestion
#ifdef WORDS_BIGENDIAN
	    if (num_bits == 16) {
	        t[0] = a >> 8;
	        t[1] = a;
	    } else if (num_bits >= 32) {
		t[0] = a >> 24;
		t[1] = a >> 16;
	        t[2] = a >> 8;
	        t[3] = a;
	        
	        if (num_bits == 64) {
		    t[4] = b >> 24;
		    t[5] = b >> 16;
		    t[6] = b >> 8;
		    t[7] = b;
	        }
	    }
```

josevalim: @jhogberg of course you fix makes much more sense. I have applied it to both bigendian branches.

jhogberg: Thanks, the tests pass and I'll merge this as soon as I can. There are some bureaucratic hurdles with importing the dependency that I hope will clear soon.

jhogberg:
Merged, thanks for the PR!

PR #2891, opened 2020-11-28
Convert inet_cache to a set

josevalim:
This patch is just a refactoring but it should provide
faster operations at the cost of slightly higher storage
(which is acceptable since the table is relatively small).
This will also make it possible to have decentralized
lookups in the future.

This is prep-work for the PR in #2818.

### DIFF
--- a/lib/kernel/src/inet_db.erl
+++ b/lib/kernel/src/inet_db.erl
@@ -752,9 +752,7 @@ res_lookup_type(Domain,Type,RRs) ->
 gethostbyaddr(IP) ->
     case dnip(IP) of
 	{ok, {IP1, HType, HLen, DnIP}} ->
-	    RRs = match_rr(#dns_rr { domain = DnIP, class = in, type = ptr,
-				     cnt = '_', tm = '_', ttl = '_',
-				     bm = '_', func = '_', data = '_' }),
+	    RRs = match_rr({DnIP, in, ptr}),
 	    ent_gethostbyaddr(RRs,  IP1, HType, HLen);
 	Error -> Error
     end.
@@ -899,8 +897,7 @@ init([]) ->
     end,
     Db = ets:new(inet_db, [public, named_table]),
     reset_db(Db),
-    CacheOpts = [public, bag, {keypos,#dns_rr.domain}, named_table],
-    Cache = ets:new(inet_cache, CacheOpts),
+    Cache = ets:new(inet_cache, [public, set, named_table]),
     HostsByname = ets:new(inet_hosts_byname, [named_table]),
     HostsByaddr = ets:new(inet_hosts_byaddr, [named_table]),
     HostsFileByname = ets:new(inet_hosts_file_byname, [named_table]),
@@ -992,11 +989,11 @@ handle_call(Request, From, #state{db=Db}=State) ->
 	{del_rr, RR} when is_record(RR, dns_rr) ->
 	    %% note. del_rr will handle wildcards !!!
 	    Cache = State#state.cache,
-	    ets:match_delete(Cache, RR),
+	    ets:delete(Cache, cache_key(RR)),
 	    {reply, ok, State};
 
 	{lookup_rr, Domain, Class, Type} ->
-	    {reply, do_lookup_rr(Domain, Class, Type), State};
+	    {reply, match_rr({Domain, Class, Type}), State};
 
 	{listop, Opt, Op, E} ->
 	    El = [E],
@@ -1166,7 +1163,7 @@ handle_call(Request, From, #state{db=Db}=State) ->
 	    {reply, ok, State};
 
 	clear_cache ->
-	    ets:match_delete(State#state.cache, '_'),
+	    ets:delete_all_objects(State#state.cache),
 	    {reply, ok, State};
 
 	reset ->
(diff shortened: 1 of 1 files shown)


lib/kernel/src/inet_db.erl
```diff
@@ -1621,46 +1618,39 @@ do_add_rr(RR, Db, State) ->
...
+    Time = times(),
+    case ets:lookup(inet_cache, Key) of
+	[{_,RR}] when RR#dns_rr.ttl =:= 0 -> %% at least once
+	    ets:delete(inet_cache, Key),
+	    [RR];
+	[{_,RR}] when RR#dns_rr.tm + RR#dns_rr.ttl < Time ->
+	    ets:delete(inet_cache, Key),
+	    [];
+	[{_,RR}] ->
+	    ets:update_element(inet_cache, Key, {2, RR#dns_rr{cnt = Time}}),
```
RaimoNiskanen, line 1646: Since this function returns `true` or `false`, maybe indicate that: `_ = ets:update_element(...`

RaimoNiskanen:
Looks fine to me.  I'll include it in our nightly tests.

lib/kernel/src/inet_db.erl
josevalim: Good call! I have added the underscore as well as a comment detailed why we are ignoring it.

lib/kernel/src/inet_db.erl
```diff
@@ -1621,46 +1618,41 @@ do_add_rr(RR, Db, State) ->
...
+
+match_rr({_, _, _} = Key) ->
+    Time = times(),
+    case ets:lookup(inet_cache, Key) of
+	[{_,RR}] when RR#dns_rr.ttl =:= 0 -> %% at least once
+	    ets:delete(inet_cache, Key),
+	    [RR];
+	[{_,RR}] when RR#dns_rr.tm + RR#dns_rr.ttl < Time ->
+	    ets:delete(inet_cache, Key),
+	    [];
```
RaimoNiskanen, line 1644: These lines are not covered in our daily tests by running all test suites in Kernel.

lib/kernel/src/inet_db.erl
```diff
@@ -1746,16 +1738,17 @@ do_refresh_cache(CacheDb) ->
...
-    OldT = lists:foldl(Fun, OldestT, ets:lookup(CacheDb, Key)),
+    OldT =
+        case ets:lookup(CacheDb, Key) of
+	    [{_,RR}] when RR#dns_rr.tm + RR#dns_rr.ttl < Now ->
+		ets:delete(CacheDb, Key),
+		OldestT;
+	     [{_,#dns_rr{cnt = C}}] when C < OldestT ->
+		C;
+	     _ ->
+		OldestT
```
RaimoNiskanen, line 1750: These lines are not covered in our daily tests by running all test suites in Kernel.

lib/kernel/src/inet_db.erl
```diff
@@ -1800,17 +1793,18 @@ delete_older(CacheDb, TM, N) ->
...
-		  MM
+delete_older(Key, CacheDb, TM, N, M) ->
+    Next = ets:next(CacheDb, Key),
+    M1 =
+	case ets:lookup(CacheDb, Key) of
+	    [{_,RR}] when RR#dns_rr.cnt =< TM ->
+		ets:delete(CacheDb, Key),
+		M + 1;
+	    _ ->
+		M
```
RaimoNiskanen, line 1806: These lines are not covered in our daily tests by running all test suites in Kernel.

lib/kernel/src/inet_db.erl
```diff
@@ -1800,17 +1793,18 @@ delete_older(CacheDb, TM, N) ->
...
 	  end,
-    M1 = lists:foldl(Fun, M, ets:lookup(CacheDb, Domain)),
     delete_older(Next, CacheDb, TM, N, M1).
```
RaimoNiskanen, line 1808: These lines are not covered in our daily tests by running all test suites in Kernel.

lib/kernel/src/inet_db.erl
josevalim: Was this line covered in the previous implementation? I am asking just to know if it is something in my implementation or something that carried over from the previous code.

RaimoNiskanen:
The PR looks fine in our daily tests.  I just went through the coverage logs after all Kernel suites, and that was a bit disappointing.

Nevertheless, having full coverage in daily tests of a PR is something I do not think we have as a blocking requirement.  I just wanted to complain a bit, not to you, but to whomever, that increased test coverage would be nice to have in addition to new features ;-)

lib/kernel/src/inet_db.erl
josevalim: That makes sense. Everything is serialized on the server, there is no chance something may have been deleted meanwhile. Awesome catch, I will remove this and below.

josevalim:
@RaimoNiskanen I agree! I can definitely help not make things worse and fix the coverage in the branches I added but I am afraid some of those are carried over from before my patch. I will look into them too but no guarantees. :)

RaimoNiskanen:
@josevalim The test coverage problem is an old one; any help we might get would be much appreciated!

lib/kernel/src/inet_db.erl
RaimoNiskanen: Oh, sorry!  I thought it was a test case problem and merged.
Well, there will be future PR:s to do code cleanup in... ;-)

lib/kernel/src/inet_db.erl
RaimoNiskanen: I have not checked, but would not be surprised if they were not.  Our tests pay too little attention to `inet_res`.

lib/kernel/src/inet_db.erl
josevalim: I will clean it up on the PR that removes the deletes and makes it concurrent. :)

josevalim: Good news, those are needed for the PR that makes it concurrent. :) I am submitting it soon!

PR #2893, opened 2020-11-30
maps: Optimize intersect/2 and intersect_with/3

jhogberg:
The benchmarks in #2864 showed that `maps:intersect/2` was too slow to be used for set intersection. This PR fixes that by specializing it instead of wrapping `maps:intersect_with/3`.

While we're at it, we'll optimize  `maps:intersect_with/3` by having different iteration functions for the two argument orders instead of wrapping the `Combiner` function.

@josevalim

### DIFF
--- a/lib/stdlib/src/maps.erl
+++ b/lib/stdlib/src/maps.erl
@@ -68,57 +68,92 @@ find(_,_) -> erlang:nif_error(undef).
 from_list(_) -> erlang:nif_error(undef).
 
 -spec intersect(Map1,Map2) -> Map3 when
-    Map1 :: #{Key => term()},
-    Map2 :: #{term() => Value2},
-    Map3 :: #{Key => Value2}.
+      Map1 :: #{ Key => term() },
+      Map2 :: #{ term() => Value2 },
+      Map3 :: #{ Key => Value2 }.
 
-intersect(Map1, Map2) when is_map(Map1), is_map(Map2) ->
-    intersect_with(fun intersect_combiner/3, Map1,Map2);
-intersect(Map1, Map2) ->
-    erlang:error(error_type_two_maps(Map1, Map2),
-                 [Map1,Map2]).
-
-intersect_combiner(_K, _V1, V2) ->
-    V2.
-
--spec intersect_with(Combiner, Map1, Map2) -> Map3 when
-    Map1 :: #{Key => Value1},
-    Map2 :: #{term() => Value2},
-    Combiner :: fun((Key, Value1, Value2) -> CombineResult),
-    Map3 :: #{Key => CombineResult}.
-
-intersect_with(Combiner, Map1, Map2) when is_map(Map1),
-                                          is_map(Map2),
-                                          is_function(Combiner, 3) ->
-    case map_size(Map1) < map_size(Map2) of
+intersect(LHS, RHS) when is_map(LHS), is_map(RHS) ->
+    case map_size(LHS) < map_size(RHS) of
         true ->
-            Iterator = maps:iterator(Map1),
-            intersect_with_1(maps:next(Iterator),
-                             Map1,
-                             Map2,
-                             Combiner);
+            Iterator = maps:iterator(LHS),
+            intersect_lhs(maps:next(Iterator), LHS, RHS);
         false ->
-            Iterator = maps:iterator(Map2),
-            intersect_with_1(maps:next(Iterator),
-                             Map2,
-                             Map1,
-                             fun(K, V1, V2) -> Combiner(K, V2, V1) end)
+            Iterator = maps:iterator(RHS),
+            intersect_rhs(maps:next(Iterator), RHS, LHS)
     end;
-intersect_with(Combiner, Map1, Map2) ->
-    error(error_type_merge_intersect(Map1, Map2, Combiner),
-          [Combiner, Map1, Map2]).
+intersect(LHS, RHS) ->
+    erlang:error(error_type_two_maps(LHS, RHS), [LHS, RHS]).
 
-intersect_with_1({K, V1, Iterator}, Map1, Map2, Combiner) ->
-    case Map2 of
-        #{ K := V2 } ->
-            NewMap1 = Map1#{ K := Combiner(K, V1, V2) },
-            intersect_with_1(maps:next(Iterator), NewMap1, Map2, Combiner);
-        _ ->
-            intersect_with_1(maps:next(Iterator), maps:remove(K, Map1), Map2, Combiner)
-    end;
-intersect_with_1(none, Res, _, _) ->
+intersect_lhs({Key, LHS_Value, Iterator}, LHS0, RHS) ->
+    LHS = case RHS of
+              #{ Key := LHS_Value } ->
+                  LHS0;
+              #{ Key := RHS_Value } ->
+                  %% The value in RHS has precedence, so we must overwrite the
+                  %% current value if it differs.
+                  LHS0#{ Key := RHS_Value };
+              _ ->
+                  maps:remove(Key, LHS0)
+          end,
+    intersect_lhs(maps:next(Iterator), LHS, RHS);
+intersect_lhs(none, Res, _) ->
     Res.
 
+intersect_rhs({Key, _RHS_Value, Iterator}, RHS0, LHS) ->
+    RHS = case LHS of
+              #{ Key := _LHS_Value } ->
+                  %% The value in RHS has precedence, so it doesn't need to
+                  %% be updated even if they differ.
+                  RHS0;
+              _ ->
+                  maps:remove(Key, RHS0)
+          end,
+    intersect_rhs(maps:next(Iterator), RHS, LHS);
+intersect_rhs(none, Res, _) ->
+    Res.
+
+-spec intersect_with(Combiner, Map1, Map2) -> Map3 when
+      Map1 :: #{ Key => Value1 },
+      Map2 :: #{ term() => Value2 },
+      Combiner :: fun((Key, Value1, Value2) -> CombineResult),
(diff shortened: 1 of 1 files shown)


josevalim:
@jhogberg I have been playing with the benchmarks and it happens that this version is faster as long as the sets have a lot in common (I was benchmarking the extreme where they have everything in common). Once I added the opposite extreme, where they have nothing in common, then it is (much) slower (see below).

I will play a bit more with those extremes to find which approach is faster and when. I also want to play with a `sets:intersection` implementation that collects the different keys and the equal keys, and then decide to either delete from the existing one or create a new one at the end. It may also be the worst of both words too. :D

In any case, this is just a heads up. This PR is definitely an improvement over the previous code and I don't think these new results change it.

<details>
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
Number of Available Cores: 8
Available memory: 16 GB
Elixir 1.12.0-dev
Erlang 24.0-rc0

warmup: 2 s
time: 5 s
memory time: 0 ns
parallel: 1
inputs: large distinct bin, large distinct int, large eq bin, large eq int, medium distinct bin, medium distinct int, medium eq bin, medium eq int, small distinct bin, small distinct int, small eq bin, small eq int
Estimated total run time: 2.80 min

Benchmarking cerl_sets (filter) with input large distinct bin...
Benchmarking cerl_sets (filter) with input large distinct int...
Benchmarking cerl_sets (filter) with input large eq bin...
Benchmarking cerl_sets (filter) with input large eq int...
Benchmarking cerl_sets (filter) with input medium distinct bin...
Benchmarking cerl_sets (filter) with input medium distinct int...
Benchmarking cerl_sets (filter) with input medium eq bin...
Benchmarking cerl_sets (filter) with input medium eq int...
Benchmarking cerl_sets (filter) with input small distinct bin...
Benchmarking cerl_sets (filter) with input small distinct int...
Benchmarking cerl_sets (filter) with input small eq bin...
Benchmarking cerl_sets (filter) with input small eq int...
Benchmarking cerl_sets (iterator) with input large distinct bin...
Benchmarking cerl_sets (iterator) with input large distinct int...
Benchmarking cerl_sets (iterator) with input large eq bin...
Benchmarking cerl_sets (iterator) with input large eq int...
Benchmarking cerl_sets (iterator) with input medium distinct bin...
Benchmarking cerl_sets (iterator) with input medium distinct int...
Benchmarking cerl_sets (iterator) with input medium eq bin...
Benchmarking cerl_sets (iterator) with input medium eq int...
Benchmarking cerl_sets (iterator) with input small distinct bin...
Benchmarking cerl_sets (iterator) with input small distinct int...
Benchmarking cerl_sets (iterator) with input small eq bin...
Benchmarking cerl_sets (iterator) with input small eq int...

##### With input large distinct bin #####
Name                           ips        average  deviation         median         99th %
cerl_sets (filter)          131.05        7.63 ms    ±11.24%        7.59 ms        8.48 ms
cerl_sets (iterator)         42.48       23.54 ms     ±6.68%       23.38 ms       30.22 ms

Comparison: 
cerl_sets (filter)          131.05
cerl_sets (iterator)         42.48 - 3.09x slower +15.91 ms

##### With input large distinct int #####
Name                           ips        average  deviation         median         99th %
cerl_sets (filter)          137.75        7.26 ms    ±67.12%        6.00 ms       32.30 ms
cerl_sets (iterator)         49.84       20.06 ms     ±3.44%       19.98 ms       22.79 ms

Comparison: 
cerl_sets (filter)          137.75
cerl_sets (iterator)         49.84 - 2.76x slower +12.80 ms

Name                           ips        average  deviation         median         99th %
cerl_sets (iterator)        133.91        7.47 ms    ±11.42%        7.41 ms        8.11 ms
cerl_sets (filter)           55.86       17.90 ms     ±9.82%       17.58 ms       21.43 ms

Comparison: 
cerl_sets (iterator)        133.91
cerl_sets (filter)           55.86 - 2.40x slower +10.43 ms

Name                           ips        average  deviation         median         99th %
cerl_sets (iterator)        221.86        4.51 ms     ±7.81%        4.46 ms        5.33 ms
cerl_sets (filter)           60.18       16.62 ms    ±15.18%       15.90 ms       20.93 ms

Comparison: 
cerl_sets (iterator)        221.86
cerl_sets (filter)           60.18 - 3.69x slower +12.11 ms

##### With input medium distinct bin #####
Name                           ips        average  deviation         median         99th %
cerl_sets (filter)         17.59 K       56.85 μs    ±37.20%          55 μs       96.10 μs
cerl_sets (iterator)        7.00 K      142.95 μs    ±58.27%         135 μs         232 μs

Comparison: 
cerl_sets (filter)         17.59 K
cerl_sets (iterator)        7.00 K - 2.51x slower +86.11 μs

##### With input medium distinct int #####
Name                           ips        average  deviation         median         99th %
cerl_sets (filter)         22.98 K       43.52 μs   ±105.15%          42 μs          78 μs
cerl_sets (iterator)        9.30 K      107.53 μs    ±28.31%         101 μs      188.68 μs

Comparison: 
cerl_sets (filter)         22.98 K
cerl_sets (iterator)        9.30 K - 2.47x slower +64.01 μs

Name                           ips        average  deviation         median         99th %
cerl_sets (iterator)       17.07 K       58.58 μs    ±40.04%          56 μs      107.25 μs
cerl_sets (filter)          7.01 K      142.56 μs    ±59.49%         137 μs         214 μs

Comparison: 
cerl_sets (iterator)       17.07 K
cerl_sets (filter)          7.01 K - 2.43x slower +83.98 μs

Name                           ips        average  deviation         median         99th %
cerl_sets (iterator)       32.19 K       31.07 μs    ±53.01%          30 μs          65 μs
cerl_sets (filter)          9.13 K      109.51 μs    ±66.17%         104 μs         180 μs

Comparison: 
cerl_sets (iterator)       32.19 K
cerl_sets (filter)          9.13 K - 3.52x slower +78.44 μs

##### With input small distinct bin #####
Name                           ips        average  deviation         median         99th %
cerl_sets (filter)        750.52 K        1.33 μs  ±1371.59%           1 μs           2 μs
cerl_sets (iterator)      648.92 K        1.54 μs   ±908.78%           1 μs           2 μs

Comparison: 
cerl_sets (filter)        750.52 K
cerl_sets (iterator)      648.92 K - 1.16x slower +0.21 μs

##### With input small distinct int #####
Name                           ips        average  deviation         median         99th %
cerl_sets (filter)          1.97 M      506.89 ns  ±4441.35%           0 ns        1000 ns
cerl_sets (iterator)        1.55 M      646.76 ns  ±1588.35%        1000 ns        1000 ns

Comparison: 
cerl_sets (filter)          1.97 M
cerl_sets (iterator)        1.55 M - 1.28x slower +139.86 ns

Name                           ips        average  deviation         median         99th %
cerl_sets (iterator)      991.11 K        1.01 μs  ±1656.71%           1 μs           1 μs
cerl_sets (filter)        708.81 K        1.41 μs  ±1351.58%           1 μs           2 μs

Comparison: 
cerl_sets (iterator)      991.11 K
cerl_sets (filter)        708.81 K - 1.40x slower +0.40 μs

Name                           ips        average  deviation         median         99th %
cerl_sets (iterator)        2.91 M      343.61 ns  ±8551.97%           0 ns        1000 ns
cerl_sets (filter)          1.37 M      727.47 ns  ±3635.59%        1000 ns        1000 ns

Comparison: 
cerl_sets (iterator)        2.91 M
cerl_sets (filter)          1.37 M - 2.12x slower +383.86 ns
</details>

josevalim:
Ok, I have more benchmarks. To run these benchmarks, I have added a `maps:from_keys(List, Value)`, which creates a map from keys with a fixed value. This speeds up many functions in the `cerl_sets`/`sets` module and reduces the amount of garbage. I will send a PR for `maps:from_keys/2` separately from this convo.

The benchmarks compare three implementations:

  * `iterator_based` - it is the one in this PR but even more optimized. For sets, we don't need to check values, so we simplified that.
  * `filter_based` - as stands on `cerl_sets` right now
  * `mixed` - it happens `iterator_based` is faster as long as you are keeping more than 75% of the keys. So we start accumulating the keys we are keeping and, once we are past 75%, we start deleting

The results are good. The bookkeeping of keys on the `mixed` approach seems to be minimal, so it is often as fast as `iterator_based`, but it solves quite well the scenarios where `iterator_based` completely fails, which is when the intersection set is small. In the results, you can see mixed is either the fastest or it stays within 10% of the fastest one.

The benchmarks are organized according to this naming schema: "SIZE EQ_OR_DISTINCT TYPE". SIZE is small (10 elements), medium (1000 elements) or large (100k elements). EQ_OR_DISTINCT says if the sets are equal or distinct. `20%eq` means one set is a 20% subset of the other. `distinct+20%eq` means there is are N elements that are distinct + 20% that are equal.

The implementations are [here](https://github.com/josevalim/sets_bench/blob/jv-intersection-fork/src/fast_maps.erl). The results are [here](https://github.com/josevalim/sets_bench/blob/jv-intersection-fork/results/intersection.txt).

jhogberg:
Awesome, sounds like we should roll with `mixed` both for sets and plain maps. Can you make a PR that updates `maps:intersect/2` and `intersect_with/3`?

josevalim:
Sure, I will do that, expect a PR today!

jhogberg:
Awesome, I'll close this PR then. :)

josevalim:
@jhogberg btw, would you still split `intersect` and `intersect_with` or it is fine if `intersect/2` calls `intersect_with/3`? I am more inclined to go with the latter, especially because the implementation will be considerably longer.

jhogberg:
Of course, we can always make it faster later on.

PR #2895, opened 2020-12-01
Add maps:from_keys/2 BIF

josevalim:
It creates a new map from a set of keys and a
single value. This function can be used to
to optimize sets operations such as from_list/1,
filter/2, intersection/2, and subtract/2.


lib/dialyzer/src/erl_bif_types.erl
```diff
@@ -1536,6 +1536,14 @@ type(lists, zipwith3, 4, Xs, Opaques) ->
...
 
 %%-- maps ---------------------------------------------------------------------
+type(maps, from_keys, 2, Xs, Opaques) ->
+  strict(maps, from_keys, 2, Xs,
+	 fun ([List, Value]) ->
+	     case t_is_nil(List, Opaques) of
+	       true -> t_from_term(#{});
+	       false -> t_map([], t_list_elements(List, Opaques), Value)
+	     end
+	 end, Opaques);
```
josevalim, line 1546: Please review this one carefully because it is the first time I change `erl_bif_types`.

jhogberg: You need to update `erl_bif_types:arg_types/3` too:

```erlang
arg_types(maps, from_keys, 2) ->
  [t_list(), t_any()];
```

There's also `beam_call_types:types/3`, which the compiler uses.

```erlang
types(maps, from_keys, [Keys, Value]) ->
    RetType = #t_map{super_key=erlang_hd_type(Keys),
                     super_value=Value},
    sub_unsafe(RetType, [proper_list(), any]);
```

josevalim:
Feedback addressed. Thanks for the hints!

erts/emulator/beam/erl_map.c
```diff
@@ -299,16 +328,27 @@ static Eterm flatmap_from_validated_list(Process *p, Eterm list, Uint size) {
...
+	vs[0] = kv[2];
+    }
     size  = 1;
     item  = CDR(list_val(item));
 
     /* insert sort key/value pairs */
     while(is_list(item)) {
-
-	kv = tuple_val(CAR(list_val(item)));
+	if (fill_value != THE_NON_VALUE) {
```
jhogberg, line 344:
```suggestion
	if (is_value(fill_value)) {
```

josevalim:
Nits addressed!

PR #2897, opened 2020-12-01
Use maps:from_list/1 on maps:intersect_with/3

josevalim:
The current implementation is not efficient when
the maps being compared are mostly distinct.
It happens that collecting all keys and building
the map once at the end is the most efficient
way of implementing intersection.

Benchmarks show this approch is often 4x faster in
the mostly distinct cases, otherwise it behaves
within 5-10% of the current implementation. The only
case it is considerably slower is when working with
small sets of integers (~10 elements) but those
are done in nanoseconds anyway.

Benchmaking code and results:
https://gist.github.com/josevalim/759f3a9c2613119d3e38cd4224840d32

### DIFF
--- a/lib/stdlib/src/maps.erl
+++ b/lib/stdlib/src/maps.erl
@@ -73,13 +73,18 @@ from_list(_) -> erlang:nif_error(undef).
     Map3 :: #{Key => Value2}.
 
 intersect(Map1, Map2) when is_map(Map1), is_map(Map2) ->
-    intersect_with(fun intersect_combiner/3, Map1,Map2);
+    case map_size(Map1) =< map_size(Map2) of
+        true ->
+            intersect_with_small_map_first(fun intersect_combiner_v2/3, Map1, Map2);
+        false ->
+            intersect_with_small_map_first(fun intersect_combiner_v1/3, Map2, Map1)
+    end;
 intersect(Map1, Map2) ->
     erlang:error(error_type_two_maps(Map1, Map2),
                  [Map1,Map2]).
 
-intersect_combiner(_K, _V1, V2) ->
-    V2.
+intersect_combiner_v1(_K, V1, _V2) -> V1.
+intersect_combiner_v2(_K, _V1, V2) -> V2.
 
 -spec intersect_with(Combiner, Map1, Map2) -> Map3 when
     Map1 :: #{Key => Value1},
@@ -90,35 +95,33 @@ intersect_combiner(_K, _V1, V2) ->
 intersect_with(Combiner, Map1, Map2) when is_map(Map1),
                                           is_map(Map2),
                                           is_function(Combiner, 3) ->
-    case map_size(Map1) < map_size(Map2) of
+    %% Use =< because we want to avoid reversing the combiner if we can
+    case map_size(Map1) =< map_size(Map2) of
         true ->
-            Iterator = maps:iterator(Map1),
-            intersect_with_1(maps:next(Iterator),
-                             Map1,
-                             Map2,
-                             Combiner);
+            intersect_with_small_map_first(Combiner, Map1, Map2);
         false ->
-            Iterator = maps:iterator(Map2),
-            intersect_with_1(maps:next(Iterator),
-                             Map2,
-                             Map1,
-                             fun(K, V1, V2) -> Combiner(K, V2, V1) end)
+            RCombiner = fun(K, V1, V2) -> Combiner(K, V2, V1) end,
+            intersect_with_small_map_first(RCombiner, Map2, Map1)
     end;
 intersect_with(Combiner, Map1, Map2) ->
     error(error_type_merge_intersect(Map1, Map2, Combiner),
           [Combiner, Map1, Map2]).
 
-intersect_with_1({K, V1, Iterator}, Map1, Map2, Combiner) ->
+intersect_with_small_map_first(Combiner, SmallMap, BigMap) ->
+    Next = maps:next(maps:iterator(SmallMap)),
+    intersect_with_iterate(Next, [], SmallMap, BigMap, Combiner).
+
+intersect_with_iterate({K, V1, Iterator}, Keep, Map1, Map2, Combiner) ->
+    Next = maps:next(Iterator),
     case Map2 of
         #{ K := V2 } ->
-            NewMap1 = Map1#{ K := Combiner(K, V1, V2) },
-            intersect_with_1(maps:next(Iterator), NewMap1, Map2, Combiner);
+            V = Combiner(K, V1, V2),
+            intersect_with_iterate(Next, [{K,V}|Keep], Map1, Map2, Combiner);
         _ ->
-            intersect_with_1(maps:next(Iterator), maps:remove(K, Map1), Map2, Combiner)
+            intersect_with_iterate(Next, Keep, Map1, Map2, Combiner)
     end;
-intersect_with_1(none, Res, _, _) ->
-    Res.
-
+intersect_with_iterate(none, Keep, _Map1, _Map2, _Combiner) ->
+    maps:from_list(Keep).
 
 %% Shadowed by erl_bif_types: maps:is_key/2
 -spec is_key(Key,Map) -> boolean() when


lib/stdlib/src/maps.erl
```diff
@@ -90,35 +95,53 @@ intersect_combiner(_K, _V1, V2) ->
...
+intersect_with_sorted(Combiner, SmallMap, BigMap) ->
+    Next = maps:next(maps:iterator(SmallMap)),
+    Limit = floor(map_size(SmallMap) * 0.75),
+    intersect_with_heuristic(Next, [], [], Limit, SmallMap, BigMap, Combiner).
+
+%% If we are keeping/changing more than 75% of the keys,
+%% then it is cheaper to update in place. Stop accumulating
+%% and start replacing and deleting.
+intersect_with_heuristic(Next, Keep, Delete, 0, Map1, Map2, Combiner) ->
+    intersect_with_decided(Next, replace(Keep, without(Delete, Map1)), Map2, Combiner);
```
josevalim, line 119: Btw, if we really want to roflscale, we could introduce a `maps:change(PairsToAddOrReplace, KeysToDelete, Map)`, that would do all of this in a single pass. The biggest benefit would be avoiding the multiple allocations from updating and removing keys one by one.

It seems we could reuse the existing implementations for put and remove but this is way above my paygrade. :D

michalmuskala: It's almost as if we could use limited mutability 😅

garazdawi:
```
change([{K1,V1},{K2,V2},{K3,V3}|T],M) ->
  change(T,M#{ K1 := V1, K2 := V2, K3 := V3}).
```

You just need to make a compiler pass that optimized this into a single `put_map_exact` as is done when the keys are literals :)

josevalim: We can't do deletes tho. :'(

josevalim: > You just need to make a compiler pass that optimized this into a single put_map_exact as is done when the keys are literals :)

I explored the current implementation and it seems that for large maps, we end-up doing multiple allocations for large maps. It makes sense though, we want to allocate only the hash buckets that are necessary, and I can see it being very complicated to do all of this in one pass. Both updates and deletions.

The only benefit would be for small maps but then I don't think this matters much. So thanks for the hint @garazdawi!

josevalim: New morning, new insights!

So I also realized that part of the operation I want to perform here is actually a `maps:merge`. It just happens though that one of the sides is a list, not a map. However, could it be that a from_list+merge is faster than a fold+put?

Yup, that's actually consistently the case:

```
##### With input large 100%eq bin #####
Name                      ips        average  deviation         median         99th %
from_list+merge         30.63       32.65 ms    ±10.43%       32.10 ms       62.62 ms
fold+put                14.94       66.96 ms    ±14.26%       65.35 ms      112.48 ms

Comparison: 
from_list+merge         30.63
fold+put                14.94 - 2.05x slower +34.31 ms

##### With input large 20%eq bin #####
Name                      ips        average  deviation         median         99th %
from_list+merge        144.51        6.92 ms    ±49.30%        6.57 ms        9.28 ms
fold+put                92.47       10.81 ms    ±43.82%       10.55 ms       15.86 ms

Comparison: 
from_list+merge        144.51
fold+put                92.47 - 1.56x slower +3.89 ms

##### With input large 40%eq bin #####
Name                      ips        average  deviation         median         99th %
from_list+merge         70.41       14.20 ms    ±33.88%       13.59 ms       17.48 ms
fold+put                37.53       26.65 ms    ±27.91%       25.68 ms       94.18 ms

Comparison: 
from_list+merge         70.41
fold+put                37.53 - 1.88x slower +12.44 ms

##### With input large 60%eq bin #####
Name                      ips        average  deviation         median         99th %
from_list+merge         48.03       20.82 ms    ±27.83%       19.90 ms       62.63 ms
fold+put                23.07       43.35 ms    ±21.53%       41.65 ms      111.16 ms

Comparison: 
from_list+merge         48.03
fold+put                23.07 - 2.08x slower +22.53 ms

##### With input large 80%eq bin #####
Name                      ips        average  deviation         median         99th %
from_list+merge         37.28       26.82 ms    ±25.75%       25.86 ms       90.08 ms
fold+put                11.24       88.97 ms    ±16.09%       89.39 ms      150.98 ms

Comparison: 
from_list+merge         37.28
fold+put                11.24 - 3.32x slower +62.15 ms

Name                      ips        average  deviation         median         99th %
from_list+merge         30.45       32.84 ms    ±10.95%       32.22 ms       63.70 ms
fold+put                13.79       72.53 ms    ±11.70%       72.19 ms      119.20 ms

Comparison: 
from_list+merge         30.45
fold+put                13.79 - 2.21x slower +39.69 ms

Name                      ips        average  deviation         median         99th %
from_list+merge         39.96       25.02 ms    ±24.96%       24.19 ms       79.95 ms
fold+put                18.99       52.65 ms    ±21.83%       50.95 ms      132.91 ms

Comparison: 
from_list+merge         39.96
fold+put                18.99 - 2.10x slower +27.63 ms

Name                      ips        average  deviation         median         99th %
from_list+merge        7.79 K      128.43 μs   ±138.12%      117.99 μs      231.99 μs
fold+put               6.80 K      147.01 μs   ±308.33%      123.99 μs      295.99 μs

Comparison: 
from_list+merge        7.79 K
fold+put               6.80 K - 1.14x slower +18.58 μs

Name                      ips        average  deviation         median         99th %
fold+put              10.82 K       92.43 μs    ±28.47%       80.99 μs      190.99 μs
from_list+merge       10.20 K       98.00 μs   ±160.52%       88.99 μs      186.05 μs

Comparison: 
fold+put              10.82 K
from_list+merge       10.20 K - 1.06x slower +5.57 μs

##### With input small distinct int #####
Name                      ips        average  deviation         median         99th %
from_list+merge        3.43 M      291.16 ns ±12257.25%           0 ns         990 ns
fold+put               1.32 M      758.02 ns  ±5222.34%         990 ns         990 ns

Comparison: 
from_list+merge        3.43 M
fold+put               1.32 M - 2.60x slower +466.86 ns

Name                      ips        average  deviation         median         99th %
from_list+merge      919.60 K        1.09 μs  ±4157.18%        0.99 μs        1.99 μs
fold+put             664.28 K        1.51 μs  ±1621.71%        0.99 μs        4.99 μs

Comparison: 
from_list+merge      919.60 K
fold+put             664.28 K - 1.38x slower +0.42 μs

Name                      ips        average  deviation         median         99th %
from_list+merge        2.09 M      479.54 ns ±18429.34%           0 ns         990 ns
fold+put               1.62 M      616.33 ns  ±3559.07%           0 ns         990 ns

Comparison: 
from_list+merge        2.09 M
fold+put               1.62 M - 1.29x slower +136.79 ns
```

This is always be the case as long as we are merging with large maps. I will amend this PR accordingly.

josevalim:
Ok, I have updated the implementation, it is simpler now. Because of the combiner function (and therefore the fact values can change), using `maps:from_list/1` is the best approach here. We don't need to make it mixed at all. Mixed still works on sets though (since values are guaranteed to be the same).

lib/stdlib/src/maps.erl
```diff
@@ -90,35 +95,33 @@ intersect_combiner(_K, _V1, V2) ->
...
-                             fun(K, V1, V2) -> Combiner(K, V2, V1) end)
+            RCombiner = fun(K, V1, V2) -> Combiner(K, V2, V1) end,
+            intersect_with_sorted(RCombiner, Map2, Map1)
     end;
 intersect_with(Combiner, Map1, Map2) ->
     error(error_type_merge_intersect(Map1, Map2, Combiner),
           [Combiner, Map1, Map2]).
 
-intersect_with_1({K, V1, Iterator}, Map1, Map2, Combiner) ->
+intersect_with_sorted(Combiner, SmallMap, BigMap) ->
```
kjellwinblad, line 110: The sorted word confused me a little bit as it made me think that the function would use sorting in some way before I realized that it just meant that the function expects that the parameters have some properties. Perhaps this can be explained in a comment or "_sorted" can be changed to "_smaller_map_first". I don't think this is a big issue so I'm ok with keeping it as it is if you think that is better.

kjellwinblad (approved):
Thanks for the pull-request and the benchmarking! Looks solid!

lib/stdlib/src/maps.erl
josevalim: Oh, I like that, much clearer. I will change it.

jhogberg:
Merged, thanks for the PR!

PR #2900, opened 2020-12-01
Add maps:filtermap/2

Maria-12648430:
This PR introduces `filtermap/2` for the `maps` module to combine the common tasks of filtering and mapping a map into one operation.

Compared with two separate calls to `maps:filter/2` and `maps:map/2` as required now, this is both...
* Faster, as the map needs to be "traversed" only once, and there is only one final (internal) `maps:from_list/1` call involved.
* More convenient, as only one call with one filter+map function is needed, instead of two calls with two different functions.

According to a discussion in #2850 ([@IngelaAndin's comment](https://github.com/erlang/otp/pull/2850#issuecomment-733519962)), `filter_map/2` might be a better name in the context of the `maps` module, but I'm really not sure. What do you think?

### DIFF
--- a/lib/stdlib/doc/src/maps.xml
+++ b/lib/stdlib/doc/src/maps.xml
@@ -73,6 +73,33 @@
     </func>
 
     <func>
+      <name name="filtermap" arity="2" since=""/>
+      <fsummary>Filter and map a map.</fsummary>
+      <desc>
+        <p>Returns a map <c><anno>Map</anno></c> that is the result of calling
+          <c><anno>Fun</anno>(<anno>Key</anno>, <anno>Value1</anno>)</c> for
+          every <c><anno>Key</anno></c> to value <c><anno>Value1</anno></c>
+          association in <c><anno>MapOrIter</anno></c> in any order.</p>
+        <p>If <c><anno>Fun</anno>(<anno>Key</anno>, <anno>Value1</anno>)</c>
+          returns <c>true</c>, the association is copied to the result map. If
+          it returns <c>false</c>, the association is not copied. If it returns
+          <c>{true, NewValue}</c>, the value for <c><anno>Key</anno></c> is
+          replaced with <c>NewValue</c>at this position is replaced in the
+          result map.</p>
+        <p>The call fails with a <c>{badmap,Map}</c> exception if
+          <c><anno>MapOrIter</anno></c> is not a map or valid iterator,
+          or with <c>badarg</c> if <c><anno>Fun</anno></c> is not a
+          function of arity 3.</p>
+        <p><em>Example:</em></p>
+        <code type="none">
+> Fun = fun(K,V) when is_atom(K) -> {true, V*2}; (_,V) -> (V rem 2) =:= 0 end,
+  Map = #{k1 => 1, "k2" => 2, "k3" => 3},
+  maps:filtermap(Fun,Map).
+#{k1 => 2,"k2" => 2}</code>
+      </desc>
+    </func>
+
+    <func>
       <name name="find" arity="2" since="OTP 17.0"/>
       <fsummary></fsummary>
       <desc>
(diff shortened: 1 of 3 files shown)


jhogberg:
> Compared with two separate calls to maps:filter/2 and maps:map/2 as required now, this is both...

You can easily do it with a single call to `maps:fold/3` though.

> and there is only one final (internal) maps:from_list/1 call involved.

This is an iffy justification: many C BIFs are only faster because they don't yield properly.

Maria-12648430:
> Edit: Woops, I slipped and accidentally clicked "close with comment," closing it and posting my response way too early. :(
> 
> > Compared with two separate calls to maps:filter/2 and maps:map/2 as required now, this is both...
> 
> You can do it with a single call to `maps:fold/3` though,

myeees... =^^=

> and I'm not sure this operation is common enough to warrant another function, but that's just my personal take. We'll discuss it at our next technical board. :)

Judging from the reactions to the initial message (👍 3 and ❤️ 1 ;)), it looks like it is appreciated ;)

> > and there is only one final (internal) maps:from_list/1 call involved.
> 
> Many C BIFs are only faster because they don't yield properly, `maps:from_list/1` included. If you pass it a long enough list it'll hog a scheduler and prevent other processes from running. Building the resulting map one element at a time avoids that problem at the cost of speed, so "faster" doesn't always mean better.

No, I think you got me wrong :)
In the current implementation, both `filter/2` and `map/2` build the result as a list and convert it to a map as the final step:
https://github.com/erlang/otp/blob/25a13ce030f346dfac2aac435eb6f5a16f220894/lib/stdlib/src/maps.erl#L288-L291
and
https://github.com/erlang/otp/blob/25a13ce030f346dfac2aac435eb6f5a16f220894/lib/stdlib/src/maps.erl#L335-L338

I just followed suit with `filtermap/2`, assuming there was a good reason for doing it this way ;)
Anyway, so when using `filter/2` and `map/2` one after another, there are _two_ such calls to `from_list/1`, one for `filter/2`, one for `map/2`. Combining those two functions into one uses only _one_ call to `from_list/1` instead of two.

jhogberg:
> Judging from the reactions to the initial message (+1 3 and heart 1 ;)), it looks like it is appreciated ;)

People tend to have different opinions, hence the technical board. :)

> No, I think you got me wrong :)
> In the current implementation, both filter/2 and map/2 build the result as a list and convert it to a map as the final step:
> ... snip ...
> Anyway, so when using filter/2 and map/2 one after another, there are two such calls to from_list/1, one for filter/2, one for map/2. Combining those two functions into one uses only one call to from_list/1 instead of two.

Righto. I personally don't see much benefit over using `maps:fold/3` though.

Maria-12648430:
> > Judging from the reactions to the initial message (+1 3 and heart 1 ;)), it looks like it is appreciated ;)
> 
> People tend to have different opinions, hence the technical board. :)

Sure ;) You (and the board) decide :)

> Righto. I personally don't see much benefit over using `maps:fold/3` though.

Really? Personally, I find the `filtermap` functions some of the most useful and convenient ones ever invented :)
(Sorry, not trying to argue or push things here. Just my personal opinion ;))

jhogberg:
> Sure ;) You (and the board) decide :)

My opinions are my own and it's not up to me to decide, we'll discuss it on the next technical board. :)

> Really? Personally, I find the filtermap functions some of the most useful and convenient ones ever invented :)

The concept is useful, but `true | false | {true, NewItem}` feels awkward and creates extra garbage whenever you want to use the `map` portion of it. I personally find it easier to read a straight `maps:fold/3`.

Maria-12648430:
Ok then :)

IngelaAndin:
About the name, I guess it will depend on which principal the board deems most important.  To be consistent in the module or to have the same name for two functions with the same semantics. For another recent naming issue, we decided that functions with different semantics should not have the same name.

jhogberg:
We've decided to take it in as-is. I've added it to our nightly builds and will merge it tomorrow or on Monday. :)

jhogberg:
Merged, thanks for the PR!

Maria-12648430:
My pleasure :)

PR #2903, opened 2020-12-02
Fix negative zero to string

josevalim:
Both float_to_binary and float_to_list will correctly
print negative zeros when using the scientific mode but
not when using the decimal representation. This patch
makes it consistent to always show negative floats.

This is important because some operations, such as math:atan2,
will return drastically different results if negative zeros
are given compared to positive zeros. Negative zeros also
are encoded differently in bitstrings.

### DIFF
--- a/erts/emulator/test/num_bif_SUITE.erl
+++ b/erts/emulator/test/num_bif_SUITE.erl
@@ -187,6 +187,17 @@ t_float_to_string(Config) when is_list(Config) ->
     test_fts("1.2300000000e+20",1.23e20, [{scientific, 10}, compact]),
     test_fts("1.23000000000000000000e+20",1.23e20, []),
 
+    %% Negative zero
+    <<NegZero/float>> = <<16#8000000000000000:64>>,
+    "-0.0" = float_to_list(NegZero, [{decimals, 1}, compact]),
+    "-0.0" = float_to_list(NegZero, [{decimals, 1}]),
+    "-0.0e+00" = float_to_list(NegZero, [{scientific, 1}]),
+    "-0.0e+00" = float_to_list(NegZero, [{scientific, 1}, compact]),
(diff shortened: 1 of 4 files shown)


josevalim:
PR updated to also fix io:format.

erts/emulator/test/num_bif_SUITE.erl
```diff
@@ -187,6 +187,17 @@ t_float_to_string(Config) when is_list(Config) ->
...
+    %% Negative zero
+    <<NegZero/float>> = <<16#8000000000000000:64>>,
+    "-0.0" = erlang:float_to_list(NegZero, [{decimals, 1}, compact]),
+    "-0.0" = erlang:float_to_list(NegZero, [{decimals, 1}]),
+    "-0.0e+00" = erlang:float_to_list(NegZero, [{scientific, 1}]),
+    "-0.0e+00" = erlang:float_to_list(NegZero, [{scientific, 1}, compact]),
+    <<"-0.0">> = erlang:float_to_binary(NegZero, [{decimals, 1}, compact]),
+    <<"-0.0">> = erlang:float_to_binary(NegZero, [{decimals, 1}]),
+    <<"-0.0e+00">> = erlang:float_to_binary(NegZero, [{scientific, 1}]),
+    <<"-0.0e+00">> = erlang:float_to_binary(NegZero, [{scientific, 1}, compact]),
```
bjorng, line 199: You don't need the `erlang:` prefix.

bjorng (changes requested):

erts/emulator/test/num_bif_SUITE.erl
josevalim: Fixed.

bjorng (approved):

bjorng:
Added to our daily builds.

PR #2904, opened 2020-12-03
eldap: Add ability to specify size limit on ldap requests

AbdullahShehabi:


lib/eldap/src/eldap.erl
```diff
@@ -1088,6 +1093,9 @@ v_bool(true)  -> true;
 v_bool(false) -> false;
 v_bool(_Bool) -> throw({error,concat(["not Boolean: ",_Bool])}).
 
+v_size_limit(I) when is_integer(I), I>=0 -> I;
+v_size_limit(_I) -> throw({error,concat(["size_limit not positive integer: ",_I])}).
```
paulo-ferraz-oliveira, line 1097: `=:= 0` is non-negative, not positive 😄 (I see there's an error just like it, below your change)

AbdullahShehabi: So update the error message, or the actual check evaluation, seems like we are enforcing the number being greater than 0, so must be a positive integer. Didn't really understand what you mean 😅

paulo-ferraz-oliveira: I was stating that since you're checking if something is bigger or equal to 0 the "correct" message to throw in case it isn't should be "size_limit not non-negative integer" or better still "size_limit negative integer", but it's a minor thing (it would make it incompatible with other messages outside of the scope of the pull request, though).

HansN:
Hi @AbdullahShehabi 
You have to sign the CLA (see above) before we can handle your PR.

AbdullahShehabi:
Hi Hans, I already signed it, so not sure what is going on. 
`You have signed the CLA already but the status is still pending? Let us recheck it.`

HansN:
Hi @AbdullahShehabi ,
It could be that you have committed with an email-address that is not in your github user.  I checked my own profile, and it is possible to have several in it.

AbdullahShehabi:
Thanks for pointing that out, didnt realize i had the wrong email in the commit 😅

HansN:
Looks good so far.  Do you plan to also submit on e or more test cases and some documentation?   See [Contributing to Erlang/OTP](https://github.com/erlang/otp/blob/master/CONTRIBUTING.md)

HansN:
ping

AbdullahShehabi:
Sorry, I am not familiar with writing tests in erlang so i will need to take a look into it, would you be able to point me to the test cases for the other filters so i can append my test to it?

HansN:
Great! I'm a bit busy now, but in a couple of days I could help you.

HansN:
Thanks for the PR! It's rebased and merged into maint

I made two test cases ( search_filter_or_sizelimit_exceeded/1 and search_filter_or_sizelimit_ok/1) in  eldap/test/eldap_basic_SUITE.erl.

PR #2905, opened 2020-12-03
Do not consider zero floats as singleton

josevalim:
There is currently a bug in the compiler where 0.0
is considered a singleton. However, 0.0 can match both
-0.0 and 0.0 which makes this assumption invalid. Here
is a failing test:

    float_zero(Config) when is_list(Config) ->
        <<16#0000000000000000:64>> = to_fbin(1*0.0),
        <<16#8000000000000000:64>> = to_fbin(-1*0.0),
        ok.

    to_fbin(X = 0.0) -> <<X/float>>.

This commit changes beam_types:get_single_value/1 to not
consider float singletons and therefore not perform constant
propagation.

Furthermore, we also need to instruct beam_asm to encode
0.0 and -0.0 as two distinct literals.

### DIFF
--- a/lib/compiler/test/float_SUITE.erl
+++ b/lib/compiler/test/float_SUITE.erl
@@ -22,14 +22,14 @@
 	 init_per_group/2,end_per_group/2,
 	 pending/1,bif_calls/1,math_functions/1,mixed_float_and_int/1,
          subtract_number_type/1,float_followed_by_guard/1,
-         fconv_line_numbers/1]).
+         fconv_line_numbers/1,float_zero/1]).
 
 -include_lib("common_test/include/ct.hrl").
 
 suite() -> [{ct_hooks,[ts_install_cth]}].
 
 all() ->
-    [pending, bif_calls, math_functions,
+    [pending, bif_calls, math_functions, float_zero,
      mixed_float_and_int, subtract_number_type,
      float_followed_by_guard,fconv_line_numbers].
 
@@ -49,6 +49,12 @@ init_per_group(_GroupName, Config) ->
 end_per_group(_GroupName, Config) ->
     Config.
 
+float_zero(Config) when is_list(Config) ->
+    <<16#0000000000000000:64>> = match_on_zero_and_to_binary(1*0.0),
+    <<16#8000000000000000:64>> = match_on_zero_and_to_binary(-1*0.0),
+    ok.
+
+match_on_zero_and_to_binary(0.0 = X) -> <<X/float>>.
 
 %% Thanks to Tobias Lindahl <tobias.lindahl@it.uu.se>
 %% Shows the effect of pending exceptions on the x86.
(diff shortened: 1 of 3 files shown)


lib/compiler/test/float_SUITE.erl
```diff
@@ -49,6 +49,12 @@ init_per_group(_GroupName, Config) ->
 end_per_group(_GroupName, Config) ->
     Config.
 
+float_zero(Config) when is_list(Config) ->
+    <<16#0000000000000000:64>> = match_on_zero_and_to_binary(1*0.0),
+    <<16#8000000000000000:64>> = match_on_zero_and_to_binary(-1*0.0),
+    ok.
+
+match_on_zero_and_to_binary(X = 0.0) -> <<X/float>>.
```
bjorng, line 57:
```suggestion
match_on_zero_and_to_binary(0.0 = X) -> <<X/float>>.
```

bjorng (changes requested):
I only have a nitpick.

josevalim:
Btw, testing this PR is a bit complicated, you need to compile beam_types.erl *before* you compile beam_dict.erl, otherwise the bug in `beam_types` won't allow the `beam_dict` fix to work. Or just compile both twice.

josevalim:
Is the failure on CI a false positive from dialyzer?

```
beam_dict.erl:173: Cons will produce an improper list since its 2nd argument is 
          binary()
```

bjorng:
> Is the failure on CI a false positive from dialyzer?

Dialyzer considers it bad style to construct non-proper lists. You will need to add a suppression. For an example on how to do that, see the `rand` module in STDLIB.

bjorng:
> Btw, testing this PR is a bit complicated, you need to compile beam_types.erl before you compile beam_dict.erl, otherwise the bug in beam_types won't allow the beam_dict fix to work. Or just compile both twice.

That will cause some problems because all our testing is done with scripts. I will have to think about how we can this.

josevalim:
@bjorng I can do this for now:

```erlang
literal(Lit, Dict) when is_float(Lit) ->
    if
        Lit > 0.0; Lit < 0.0 ->
          literal1([term|Lit], Dict);
        true ->
          literal1([binary|my_term_to_binary(Lit)], Dict)
    end.
```

And then "inline" it in the future. Thoughts?

bjorng:
I like your suggested workaround. While at it, please add comments to both `beam_types` and `beam_dict` to explain the code, because it is not immediately obvious.

bjorng:
> Perhaps it is even better to keep in this format permanently so compiler bugs do not compound?

Yes, I don't think this is performance critical. Better write code that is likely to always be correct.

josevalim:
All feedback should be addressed.

bjorng (approved):

bjorng:
Added to our daily builds.

PR #2906, opened 2020-12-03
Handle negative zeros on unary minus

josevalim:
Before this patch, unary minus "-value" was implemented
as "0 - value" but that is not equivalent for floats when
0.0 is given.

This patch addresses this issue by implementing unary
minus as its own operation and improves the coverage.

### DIFF
--- a/erts/emulator/beam/erl_arith.c
+++ b/erts/emulator/beam/erl_arith.c
@@ -79,7 +79,7 @@ BIF_RETTYPE splus_2(BIF_ALIST_2)
 
 BIF_RETTYPE sminus_1(BIF_ALIST_1)
 {
-    BIF_RET(erts_mixed_minus(BIF_P, make_small(0), BIF_ARG_1));
+    BIF_RET(erts_unary_minus(BIF_P, BIF_ARG_1));
 } 
 
 BIF_RETTYPE sminus_2(BIF_ALIST_2)
@@ -460,6 +460,71 @@ erts_mixed_plus(Process* p, Eterm arg1, Eterm arg2)
 	goto badarith;
     }
 }
+
+/*
+ * While "-value" is generally the same as "0 - value",
+ * that's not true for floats due to positive and negative
+ * zeros, so we implement unary minus as its own operation.
+ */
+Eterm
+erts_unary_minus(Process* p, Eterm arg)
+{
+    Eterm hdr, res;
+    FloatDef f;
+    dsize_t sz;
+    int need_heap;
+    Eterm* hp;
+    Sint ires;
+
+    ERTS_FP_CHECK_INIT(p);
+    switch (arg & _TAG_PRIMARY_MASK) {
+    case TAG_PRIMARY_IMMED1:
+        switch ((arg & _TAG_IMMED1_MASK) >> _TAG_PRIMARY_SIZE) {
+        case (_TAG_IMMED1_SMALL >> _TAG_PRIMARY_SIZE):
+            ires = -signed_val(arg);
+            if (IS_SSMALL(ires)) {
+                return make_small(ires);
+            } else {
+                hp = HeapFragOnlyAlloc(p, 2);
+                res = small_to_big(ires, hp);
+                return res;
+            }
+        default:
+        badarith:
+            p->freason = BADARITH;
+            return THE_NON_VALUE;
+        }
+    case TAG_PRIMARY_BOXED:
+        hdr = *boxed_val(arg);
+        switch ((hdr & _TAG_HEADER_MASK) >> _TAG_PRIMARY_SIZE) {
+        case (_TAG_HEADER_POS_BIG >> _TAG_PRIMARY_SIZE):
+        case (_TAG_HEADER_NEG_BIG >> _TAG_PRIMARY_SIZE): {
+            Eterm zero_buf[2] = {make_pos_bignum_header(1), 0};
+            Eterm zero = make_big(zero_buf);
+            sz = big_size(arg);
+            need_heap = BIG_NEED_SIZE(sz);
+            hp = HeapFragOnlyAlloc(p, need_heap);
+            res = big_minus(zero, arg, hp);
+            maybe_shrink(p, hp, res, need_heap);
+            ASSERT(is_not_nil(res));
+            return res;
+        }
+        case (_TAG_HEADER_FLOAT >> _TAG_PRIMARY_SIZE):
+            GET_DOUBLE(arg, f);
+            f.fd = -f.fd;
+            ERTS_FP_ERROR(p, f.fd, goto badarith);
+            hp = HeapFragOnlyAlloc(p, FLOAT_SIZE_OBJECT);
+            res = make_float(hp);
+            PUT_DOUBLE(f, hp);
+            return res;
+        default:
+            goto badarith;
+        }
+    default:
+        goto badarith;
+    }
+}
+
 
 Eterm
 erts_mixed_minus(Process* p, Eterm arg1, Eterm arg2)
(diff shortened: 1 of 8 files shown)


erts/emulator/beam/erl_arith.c
```diff
@@ -461,6 +461,70 @@ erts_mixed_plus(Process* p, Eterm arg1, Eterm arg2)
...
+    int need_heap;
+    Eterm* hp;
+    Sint ires;
+
+    ERTS_FP_CHECK_INIT(p);
+    switch (arg & _TAG_PRIMARY_MASK) {
+    case TAG_PRIMARY_IMMED1:
+        switch ((arg & _TAG_IMMED1_MASK) >> _TAG_PRIMARY_SIZE) {
+        case (_TAG_IMMED1_SMALL >> _TAG_PRIMARY_SIZE):
+            ires = -signed_val(arg);
```
josevalim, line 480: I believe this is correct, as before it would be equivalent to `ires = signed_val(make_small(0)) - signed_val(arg)`.

erts/emulator/beam/erl_arith.c
```diff
@@ -461,6 +461,70 @@ erts_mixed_plus(Process* p, Eterm arg1, Eterm arg2)
...
+            sz = MAX(sz1, sz2)+1;
+            need_heap = BIG_NEED_SIZE(sz);
+            hp = HeapFragOnlyAlloc(p, need_heap);
+            res = big_minus(zero, arg, hp);
+            maybe_shrink(p, hp, res, need_heap);
+            if (is_nil(res)) {
+                p->freason = SYSTEM_LIMIT;
+                return THE_NON_VALUE;
+            }
+            return res;
```
josevalim, line 510: It feels there is a more efficient version of this whole branch but I am well versed enough to come up with it.

erts/emulator/beam/erl_arith.c
```diff
@@ -461,6 +461,70 @@ erts_mixed_plus(Process* p, Eterm arg1, Eterm arg2)
     }
 }
 
+Eterm
+erts_unary_minus(Process* p, Eterm arg)
```
josevalim, line 465: This is a copy and paste of erts_mixed_minus, without the first argument, and with zero inlined.

mikpe, line 465: LGTM. I would just ask for a comment above this function head explaining that the reason it exists is that "-X" is not the same as "0-X" when X is floating-point zero.

erts/emulator/test/float_SUITE.erl
```diff
@@ -56,6 +56,13 @@ otp_7178(Config) when is_list(Config) ->
     {'EXIT', {badarg,_}} = (catch list_to_float("1.0e83291083210")),
     ok.
 
+negative_zero(Config) when is_list(Config) ->
+    <<16#8000000000000000:64>> = <<(my_apply(erlang, '-', [0.0]))/float>>,
+    <<16#8000000000000000:64>> = <<(my_apply(erlang, '*', [-1, 0.0]))/float>>,
+    <<16#8000000000000000:64>> = <<(my_apply(erlang, '*', [-1.0, 0.0]))/float>>,
+    <<16#8000000000000000:64>> = <<(my_apply(erlang, '*', [-1.0, 0]))/float>>,
```
mikpe, line 63: I'll just add that I wrote a C test case and ran it on three different architectures (x86-64, SPARC64, and PowerPC64), and they all agree that negating a floating-point zero sets the sign bit. I haven't read chapter-and-verse from the Erlang spec, but I think people expect compatible behaviour from Erlang.

mikpe (approved):

mikpe:
Since this is IMO a bug fix, it should perhaps be targeting "maint" rather than "master".

erts/emulator/beam/erl_arith.c
josevalim: Done!

erts/emulator/beam/erl_arith.c
bjorng: Big numbers have explicit sign bits (just as floats). The sign bit is in the header word (_TAG_HEADER_POS_BIG and _TAG_HEADER_NEG_BIG). There is indeed only a single bits difference (NEG_BIG_SUBTAG - POS_BIG_SUBTAG). You can implement the negation by simply copying the original bignum, followed by a sign flip of the header word. Be sure to add a call to ERTS_CT_ASSERT() to ensure that the bit flipping is safe, for example (not tested):

```
ERTS_CT_ASSERT((NEG_BIG_SUBTAG - POS_BIG_SUBTAG) >> _TAG_PRIMARY_SIZE == 1)
```

josevalim: Thanks @bjorng!

Unfortunately I tried this and my C knowledge is not good enough to make this happen. In case you or anyone else wants to pick this up, it is totally fine by me. Sorry!

erts/emulator/beam/erl_arith.c
```diff
@@ -461,6 +461,73 @@ erts_mixed_plus(Process* p, Eterm arg1, Eterm arg2)
...
+        case (_TAG_HEADER_NEG_BIG >> _TAG_PRIMARY_SIZE):
+            sz = big_size(arg);
+            need_heap = BIG_NEED_SIZE(sz);
+            hp = HeapFragOnlyAlloc(p, need_heap);
+            res = make_big(hp);
+            ERTS_CT_ASSERT((NEG_BIG_SUBTAG - POS_BIG_SUBTAG) >> _TAG_PRIMARY_SIZE == 1);
+            *hp++ = hdr ^ (NEG_BIG_SUBTAG - POS_BIG_SUBTAG);
+            big_p = big_val(arg) + 1;
+            while (--need_heap) {
+                *hp++ = *big_p++;
```
josevalim, line 511: This was the magical incantation I was missing! 💯

bjorng:
Unary minus in compiled code is still wrong. I have pushed an update to the test case to show it. I will soon push a fix.

josevalim:
@bjorng it is failing even with the changes from #2905?

bjorng:
> it is failing even with the changes from #2905?

I haven't tried with #2905, but I don't see how that would help.

bjorng:
Unfortunately, my code for negating a bignum is wrong. I have reverted it. (The code doesn't handle the case that the bignum should be converted back to a small.) I might be better to keep the original code but simplify the size calculation.

bjorng:
I have implemented a new `i_unary_minus` instruction for the non-jit emulator. Now the test case passes for negations when `-` is called directly from compiled code (and not only when `erlang:'-'/2` is applied). I'll implement it for the JIT on Monday (unless you would like to do it).

josevalim:
@bjorng please do the JIT stuff on Monday as that would certainly be above my paygrade too. :)

I can simplify the size calculation though. Should I just squash all of these commits into one?

bjorng:
Yes, you can squash the commits.

josevalim:
Simplified size and squashed.

bjorng:
I have pushed two commits. One to further simplify negation of bignums, and one to implement unary minus for the JIT. When you have squashed the commits, I will add the branch to our daily builds.

josevalim:
Squashed! Thank you for the time and guidance, very appreciated!

bjorng:
I have run clang-format to correct a minor formatting error that I made, force-pushed, and added the branch to our daily builds.

bjorng:
I have force-pushed a minor fix for the non-JIT runtime system because `beam_literals_SUITE` failed in our daily builds. (A module that did unary minus on a literal compiled with optimizations turned off could not be loaded.)

PR #2909, opened 2020-12-04
EDoc: Fix links in module index when using the file_suffix option

Maria-12648430:
`edoc_doclet` has an option `file_suffix` for generating module files with different file extensions than the default `".html"`, e.g. `".htm"`. The generated module index (`modules-frame.html`) did not consider the different file suffix and always used `".html"`, so the links did not work at all. This is fixed by this PR.

<hr />

Judging from `git blame`, this option was introduced in 2009 and has been broken ever since.

Furthermore, the application of this option is incomplete, as only the module files are generated according to this option, but not the overview, the module index itself, and the index (frameset) files. While it appears to be safe to apply the file suffix option to the overview and module index as well, I'm not sure about the index (frameset), as this seems to be used for linking between applications.

It stands to reason to assume that nobody has ever been seriously using this option (I only happened to notice this bug while working on something different). It _may_ be used in some way or other, but probably only if the generated module files are processed/used further in other ways, not to generate a complete, ready-to-deliver documentation.

Therefore, I put it up to discussion if this option should not be removed from the `edoc` documentation. Removal from the code may not be feasible, or even impossible, so I'm not suggesting that.

### DIFF
--- a/lib/edoc/src/edoc_doclet.erl
+++ b/lib/edoc/src/edoc_doclet.erl
@@ -131,7 +131,7 @@ gen(Sources, App, Modules, Ctxt) ->
     Title = title(App, Options),
     CSS = stylesheet(Options),
     {Modules1, Error} = sources(Sources, Dir, Modules, Env, Options),
-    modules_frame(Dir, Modules1, Title, CSS),
+    modules_frame(Dir, Modules1, Title, CSS, Options),
     overview(Dir, Title, Env, Options),
     index_file(Dir, Title),
     edoc_lib:write_info_file(App, Modules1, Dir),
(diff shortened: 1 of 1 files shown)


garazdawi:
Thanks for the fix, I'll put it into testing.

Regarding the question about whether we should just remove this functionality altogether, I think it is better to be cautious and not remove something that may potentially (though I agrees the likelihood seems very small) break someone's [usage pattern](https://xkcd.com/1172/).

Maria-12648430:
Thanks for the xkcd, how true ;)

garazdawi:
Thanks for your contribution!

Maria-12648430:
No problem :)

PR #2910, opened 2020-12-04
Add map conversion functions to proplists

Maria-12648430:
Converting `proplists` (usually used in older projects/dependencies) from/to `maps` (used in more recent projects) is usually done with `maps:from_list/1`. Generally, you want to use `maps:get/2,3` (or pattern matching) then instead of `proplists:get_value/2,3`.

A conversion via `maps:from_list/1` is fragile in several ways here:
* proplists may contain single atoms as shorthand for `{Atom, true}`. `maps:from_list/1` cannot handle this.
* proplists may even contain _any_ term besides single atoms and key-value 2-tuples. They are accepted and ignored by the proplists API, but `maps:from_list/1` cannot handle those.
* proplists may contain the same keys multiple times with different values, and `proplists:get_value/2,3` will return the _first_ such value for a key. A map created from such a list via `maps:from_list/1` will contain the _last_ value for a key.

The `proplists:to_map/1` function introduced with this PR addresses the cases outlined above:
* (key-value 2-tuples will be converted to `Key => Value` associations, as usual).
* single atoms in the list will be expanded and appear in the map as `Atom => true`.
* any other terms are silently ignored and won't appear in the map at all; in general, it might be appropriate to raise a `badarg` exception here, but since the proplists API accepts and just ignores them, this function necessarily follows suit.
* if a key appears multiple times, the _first_ value will be in the result map, not the _last_, so `maps:get/2,3` on the created map yields the same result as would `proplists:get_value/2,3` on the proplist it was created from.

While `proplists` offers functions to access _all_ values of properties with the same key (`get_all_values/2`, `lookup_all/2`), this is not possible with a map, at least not without taking out all the conveniences you get by using a map in the first place. But if you were using `maps:from_list/1` before for this purpose, you got only single values anyway, just not necessarily the ones you wanted ;)

The other function in this PR, `proplists:from_map/1`, creates a proplist from a map, which is just `maps:to_list/1`. This function is only there for symmetry.

### DIFF
--- a/lib/stdlib/src/proplists.erl
+++ b/lib/stdlib/src/proplists.erl
@@ -35,7 +35,8 @@
 	 lookup_all/2, is_defined/2, get_value/2, get_value/3,
 	 get_all_values/2, append_values/2, get_bool/2, get_keys/1,
 	 delete/2, substitute_aliases/2, substitute_negations/2,
-	 expand/2, normalize/2, split/2, from_map/1, to_map/1]).
+	 expand/2, normalize/2, split/2, from_map/1, to_map/1,
+	 to_map/2]).
 
 %% ---------------------------------------------------------------------
 
@@ -607,14 +608,17 @@ flatten([]) ->
       Expansions :: [{Property :: property(), Expansion :: [term()]}],
       ListOut :: [term()].
 
-normalize(L, [{aliases, As} | Xs]) ->
-    normalize(substitute_aliases(As, L), Xs);
-normalize(L, [{expand, Es} | Xs]) ->
-    normalize(expand(Es, L), Xs);
-normalize(L, [{negations, Ns} | Xs]) ->
-    normalize(substitute_negations(Ns, L), Xs);
-normalize(L, []) ->
-    compact(L).
+normalize(L, Stages) ->
+    compact(apply_stages(L, Stages)).
+
+apply_stages(L, [{aliases, As} | Xs]) ->
+    apply_stages(substitute_aliases(As, L), Xs);
+apply_stages(L, [{expand, Es} | Xs]) ->
+    apply_stages(expand(Es, L), Xs);
+apply_stages(L, [{negations, Ns} | Xs]) ->
+    apply_stages(substitute_negations(Ns, L), Xs);
+apply_stages(L, []) ->
+    L.
 
 %% ---------------------------------------------------------------------
 
@@ -702,6 +706,8 @@ to_map(List) ->
         fun
             ({K, V}, M) ->
                 M#{K => V};
+            %% if tuples with arity /= 2 appear before atoms or
+            %% tuples with arity == 2, get_value/2,3 returns early
             (T, M) when 1 =< tuple_size(T) ->
                 maps:remove(element(1, T), M);
             (K, M) when is_atom(K) ->
@@ -712,6 +718,26 @@ to_map(List) ->
         #{},
         List
     ).
+
+%% @doc Converts the property list <code>List</code> to a map
+%% after applying the normalizations given in <code>Stages</code>.
+%%
+%% @see normalize/2
+%% @see to_map/1
+
+-spec to_map(List, Stages) -> Map when
+      List :: [term()],
+      Stages :: [Operation],
+      Operation :: {'aliases', Aliases}
+                 | {'negations', Negations}
+                 | {'expand', Expansions},
+      Aliases :: [{Key, Key}],
+      Negations :: [{Key, Key}],
+      Expansions :: [{Property :: property(), Expansion :: [term()]}],
+      Map :: #{term() => term()}.
+
+to_map(List, Stages) ->
+    to_map(apply_stages(List, Stages)).
 
 %% @doc Converts the map <code>Map</code> to a property list.
 
--- a/lib/stdlib/src/proplists.erl
+++ b/lib/stdlib/src/proplists.erl
@@ -35,7 +35,7 @@
 	 lookup_all/2, is_defined/2, get_value/2, get_value/3,
 	 get_all_values/2, append_values/2, get_bool/2, get_keys/1,
 	 delete/2, substitute_aliases/2, substitute_negations/2,
-	 expand/2, normalize/2, split/2]).
+	 expand/2, normalize/2, split/2, from_map/1, to_map/1]).
 
 %% ---------------------------------------------------------------------
 
@@ -668,3 +668,58 @@ split([], Store, Rest) ->
 
 maps_prepend(Key, Val, Dict) ->
     Dict#{Key := [Val | map_get(Key, Dict)]}.
+
+%% ---------------------------------------------------------------------
+
+%% @doc Converts the property list <code>List</code> to a map.
+%%
+%% Shorthand atom values in <code>List</code> will be expanded to an
+%% association of the form <code>Atom => true</code>. Tuples of the
+%% form <code>{Key, Value}</code> in <code>List</code> will be
+%% converted to an association of the form <code>Key => Value</code>.
+%% Anything else will be silently ignored.
+%%
+%% If the same key appears in <code>List</code> multiple times, the
+%% value of the one appearing nearest to the head of <code>List</code>
+%% will be in the result map, that is the value that would be returned
+%% by a call to <code>get_value/2</code> with this key.
+%%
+%% <p>Example:<pre>
+%% to_map([a, {b, 1}, {c, 2}, {c, 3}])</pre>
+%% returns<pre>
+%% #{a => true, b => 1, c => 2}</pre>
+%% </p>
+
+-spec to_map(List) -> Map when
+      List :: [Shorthand | {Key, Value} | term()],
+      Map :: #{Shorthand => 'true', Key => Value},
+      Shorthand :: atom(),
+      Key :: term(),
+      Value :: term().
+
+to_map(List) ->
+    lists:foldr(
+        fun
+            ({K, V}, M) ->
+                M#{K => V};
+            (T, M) when 1 =< tuple_size(T) ->
+                maps:remove(element(1, T), M);
+            (K, M) when is_atom(K) ->
+                M#{K => true};
+            (_, M) ->
+                M
+        end,
+        #{},
+        List
+    ).
+
+%% @doc Converts the map <code>Map</code> to a property list.
+
+-spec from_map(Map) -> List when
+    Map :: #{Key => Value},
+    List :: [{Key, Value}],
+    Key :: term(),
+    Value :: term().
+
+from_map(Map) ->
+    maps:to_list(Map).
--- a/lib/stdlib/test/proplists_SUITE.erl
+++ b/lib/stdlib/test/proplists_SUITE.erl
@@ -23,7 +23,8 @@
 -export([all/0, suite/0,groups/0, init_per_suite/1, end_per_suite/1,
          init_per_group/2, end_per_group/2,
 	 init_per_testcase/2, end_per_testcase/2,
-         examples/1, map_conversion/1]).
+         examples/1, map_conversion/1, map_conversion_normalize/1,
+         pm_fold_test/1]).
 
 init_per_testcase(_Case, Config) ->
     Config.
@@ -36,7 +37,7 @@ suite() ->
      {timetrap,{minutes,5}}].
 
 all() ->
-    [examples, map_conversion].
+    [examples, map_conversion, map_conversion_normalize, pm_fold_test].
 
 groups() ->
     [].
@@ -103,39 +104,78 @@ map_conversion(_Config) ->
     %% map yields the same results as proplists:get_value/3 on the
     %% original proplist, ie they either all return the same `Value',
     %% or they all return the `Default' given as respective third argument.
-    Default1 = make_ref(),
-    Default2 = make_ref(),
-    Default3 = make_ref(),
-    InList=[a, b, {a, 1}, {}, {a}, {a, 1, 2}, "foo"],
-    lists:foreach(
-        fun (L1) ->
-            LKs = proplists:get_keys(L1),
-            M = proplists:to_map(L1),
-            L2 = proplists:from_map(M),
-            [] = maps:keys(M) -- LKs,
-            [] = proplists:get_keys(L2) -- LKs, 
-            lists:foreach(
-                fun (K) ->
-                    case
-                        {
-                            maps:get(K, M, Default1),
-                            proplists:get_value(K, L1, Default2),
-                            proplists:get_value(K, L2, Default3)
-                        }
-                    of
-                        {V, V, V} -> true;
-                        {Default1, Default2, Default3} -> true
-                    end
-                end,
-                LKs
-            )
+    Default = make_ref(),
+    InList=[a, b, {a, 1}, {}, {a}, {a, 1, 2}, {c, 1, 2}, "foo"],
+    Fun = fun (L1, Acc) ->
+        LKs = proplists:get_keys(L1),
+        M = proplists:to_map(L1),
+        L2 = proplists:from_map(M),
+        true = lists:sort(maps:keys(M)) =:= lists:sort(proplists:get_keys(L2)),
+        lists:foreach(
+            fun (K) ->
+                case
+                    {
+                        maps:get(K, M, Default),
+                        proplists:get_value(K, L1, Default),
+                        proplists:get_value(K, L2, Default)
+                    }
+                of
+                    {Default, Default, Default} -> ok;
+                    {V, V, V} -> ok
+                end
+            end,
+            LKs
+        ),
+        Acc
+    end,
+    _ = pm_fold(Fun, undefined, InList),
+    ok.
+
+map_conversion_normalize(_Config) ->
+    Stages = [
+        {aliases, [{a, alias_a}]},
+        {negations, [{no_b, b}]},
+        {expand, [{c, [d]}]}
+    ],
+
+    M1 = proplists:to_map([], Stages),
+    true = M1 =:= #{},
+    true = M1 =:= proplists:to_map(proplists:normalize([], Stages)),
+
+    List = [a, no_b, c],
+    M2 = proplists:to_map(List, Stages),
+    true = M2 =:= #{alias_a => true, b => false, d => true},
+    true = M2 =:= proplists:to_map(proplists:normalize(List, Stages)),
+
+    ok.
+
+pm_fold(_, _, []) ->
+    [];
+pm_fold(Fun, Acc0, L) ->
+    pm_fold(Fun, Acc0, L, []).
+
+pm_fold(Fun, Acc, [], Mut) ->
+    Fun(Mut, Acc);
+pm_fold(Fun, Acc, L, Mut) ->
+    lists:foldl(
+        fun
+            (X, AccIn) -> pm_fold(Fun, AccIn, lists:delete(X, L), [X|Mut])
         end,
-        [[A, B, C, D, E, F, G] || A <- InList,
-                                  B <- InList -- [A],
-                                  C <- InList -- [A, B],
-                                  D <- InList -- [A, B, C],
-                                  E <- InList -- [A, B, C, D],
-                                  F <- InList -- [A, B, C, D, E],
-                                  G <- InList -- [A, B, C, D, E, F]]
-    ),
+        Acc,
+        L
+    ).
+
+pm_fold_test(_Config) ->
+    Fun = fun (M, A) -> [M|A] end,
+
+    [] = pm_fold(Fun, [], []),
+
+    [[1]] = lists:sort(pm_fold(Fun, [], [1])),
+
+    Exp1 = lists:sort([[1, 2], [2, 1]]),
+    Exp1 = lists:sort(pm_fold(Fun, [], [1, 2])),
+
+    Exp2 = lists:sort([[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]),
+    Exp2 = lists:sort(pm_fold(Fun, [], [1, 2, 3])),
+
     ok.
--- a/lib/stdlib/test/proplists_SUITE.erl
+++ b/lib/stdlib/test/proplists_SUITE.erl
@@ -23,7 +23,7 @@
 -export([all/0, suite/0,groups/0, init_per_suite/1, end_per_suite/1,
          init_per_group/2, end_per_group/2,
 	 init_per_testcase/2, end_per_testcase/2,
-         examples/1]).
+         examples/1, map_conversion/1]).
 
 init_per_testcase(_Case, Config) ->
     Config.
@@ -36,7 +36,7 @@ suite() ->
      {timetrap,{minutes,5}}].
 
 all() ->
-    [examples].
+    [examples, map_conversion].
 
 groups() ->
     [].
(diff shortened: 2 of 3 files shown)


erszcz:
I like this a lot :+1: The same reasons led me to create https://github.com/erszcz/kvs, but if roughly the same functionality is available upstream, the even better!

RaimoNiskanen:
We had an OTP Technical Board meeting yesterday (CET) and it got ... technical ...

The meeting came to the conclusion that this is a nice feature, and we decided on a definition of exactly how it should behave:

`proplists:get_value(Key, Proplist, Default)`
should be equivalent to
`maps:get(Key, proplists:to_map(Proplist), Default)`

This implicates that the map returned from `proplists:to_map(Proplist)` should have a `Key => Value` association if and only if `proplists:get_value(Key, Proplist, Default)` returns `Value =/= Default` for an unique `Default`.

So, a reference implementation for `proplists:to_map/1`, suggested by me and @jhogberg during the meeting, would be:
``` erlang
to_map(Proplist) ->
    Default = make_ref(),
    lists:foldl(
        fun(Key, Acc) ->
            case get_value(Key, Proplist, Default) of
                Default -> Acc;
                Value -> Acc#{​​​​​​​ Key => Value }​​​​​​​
            end
        end, #{​​​​​​​}​​​​​​​, get_keys(Proplist)).
```

That is not a very efficient implementation, but according to our fresh definition, correct.

Your suggestion is much more efficient but not entirely correct.  The `proplists` module has some devils inside, in this case that it handles lists with `atom()`s and `tuple()`s with arity >= 1, i.e not only arity 2.

Therefore, if there is a `tuple()` with arity 1 or >= 3 before one with arity 2, both with first element `Key`, `proplists:get_value(Key, Proplist, Default)` will return `Default` since it stops the search when it finds the first tuple, deciding that it can not know what to return as `Value`.  Example:
``` erlang
1> proplists:get_value(foo, [{foo}, {foo,4711}], nope).
nope
```

Just before falling asleep I realized that adding this clause to the `foldr` `fun()`, after the `({K, V}, M)` clause in your implementation could be sufficient:
``` erlang
        (T, M) when 1 =< tuple_size(T) ->
            maps:remove(element(1, T), M);
```
...but I would need second and third opinions before trusting it instead of the suggested reference implementation...

Also; just one test case is not enough.  Some of these corner cases should be checked.  Aim for 100% code coverage in the added function, unless very awkward.

Personal side note:  The `proplists` module can do many things, most summarized in `proplists:normalize/2`.  Maybe this function could be a useful addition:
``` erlang
to_map(ListIn, Stages) ->
    to_map(normalize(ListIn, Stages)).
```
...or maybe it is not enough bang for the buck...

Maria-12648430:
OMG, my good man ^^;;; And I thought this would be just so simple. j/k ;)

> `proplists:get_value(Key, Proplist, Default)`
> should be equivalent to
> `maps:get(Key, proplists:to_map(Proplist), Default)`

Agreed, this was actually what I intended ;)

> This implicates that the map returned from `proplists:to_map(Proplist)` should have a `Key => Value` association if and only if `proplists:get_value(Key, Proplist, Default)` returns `Value =/= Default` for an unique `Default`.

Ok... I tried rewording to see if I got that correct, but failed ^^; But I think I grasped the meaning from the refernce implementation: For all keys in the proplist, the map should contain only those keys where `proplists:get_value/3` returns _not_ the default, the distinction is made possible by using a unique reference in the reference implementation.

> So, a reference implementation for `proplists:to_map/1`, suggested by me and @jhogberg during the meeting, would be:
> 
> ```erlang
> to_map(Proplist) ->
>     Default = make_ref(),
>     lists:foldl(
>         fun(Key, Acc) ->
>             case get_value(Key, Proplist, Default) of
>                 Default -> Acc;
>                 Value -> Acc#{​​​​​​​ Key => Value }​​​​​​​
>             end
>         end, #{​​​​​​​}​​​​​​​, get_keys(Proplist)).
> ```
> 
> That is not a very efficient implementation, but according to our fresh definition, correct.

Yes (and yes)... `get_keys/1` builds a set, and each iteration step traverses the list again deeper and deeper, with the worst case being a "well behaved" proplist ;)

> Your suggestion is much more efficient but not entirely correct. The `proplists` module has some devils inside, in this case that it handles lists with `atom()`s and `tuple()`s with arity >= 1, i.e not only arity 2.

Ok, learned something new today =^^= It certainly feels wrong though, is there a reason why it is like this?

> Just before falling asleep I realized that adding this clause to the `foldr` `fun()`, after the `({K, V}, M)` clause in your implementation could be sufficient:
> 
> ```erlang
>         (T, M) when 1 =< tuple_size(T) ->
>             maps:remove(element(1, T), M);
> ```
> 
> ...but I would need second and third opinions before trusting it instead of the suggested reference implementation...

I think it is sound, and looks much better. A good test should ensure that it works as it should.

> Also; just one test case is not enough. Some of these corner cases should be checked. Aim for 100% code coverage in the added function, unless very awkward.

As usual, I simply followed suit with what I already found in place ;) But I agree that we should have more tests, especially since I learned about those weird edge cases here ^^;

> Personal side note: The `proplists` module can do many things, most summarized in `proplists:normalize/2`. Maybe this function could be a useful addition:
> 
> ```erlang
> to_map(ListIn, Stages) ->
>     to_map(normalize(ListIn, Stages)).
> ```
> 
> ...or maybe it is not enough bang for the buck...

OTOH, sure, why not while I'm already on it, I really just need to c&p this ;) OTOH, it isn't that much more work to explicitly write `proplists:to_map(proplists:normalize(Proplist, Stages))`, and @jhogberg was already suspicious about my implementation of `maps:filtermap/2`, so he may not be in favor of another function like this one. Decide among yourselves and let me know ;)

RaimoNiskanen:
#### Definition

> Ok... I tried rewording to see if I got that correct, but failed ^^; But I think I grasped the meaning from the refernce implementation: For all keys in the proplist, the map should contain only those keys where proplists:get_value/3 returns not the default, the distinction is made possible by using a unique reference in the reference implementation.

Precisely.  One wants to say: ...that returns a value, but since the value may be any term, to know that it is *the value* it has to be *not the default* when one is sure that *the default* =/= *the value*.  I tried but could not get it both correct and graspable... :-(

#### Intentional?

> Ok, learned something new today =^^= It certainly feels wrong though, is there a reason why it is like this?

It is at least intentional. From the documentation's first paragraph:
> If there is more than one entry in a list for a certain key, the first occurrence normally overrides any later (irrespective of the arity of the tuples).

And in the code:
``` erlang
get_value(Key, [P | Ps], Default) ->
    if is_atom(P), P =:= Key ->
	    true;
       tuple_size(P) >= 1, element(1, P) =:= Key ->
	    case P of
		{_, Value} ->
		    Value;
		_ ->
		    %% Don't continue the search!
		    Default
	    end;
       true ->
	    get_value(Key, Ps, Default)
    end;
get_value(_Key, [], Default) ->
    Default.
```
The author explicitly warns: *Don't continue the search!*

`get_value/3` is documented to:
> If `lookup(Key, List)` would yield `{Key, Value}`, this function returns the corresponding `Value`, otherwise `Default`.

And `lookup/2` would in this case return a `tuple()` with arity =/= 2, hence `get_value/3` shall return `Default`

So this is indeed intentional...

#### Test coverage

Sorry about the current bad test coverage.  It will not be much work to beat that ;-)

#### `to_map/2`

`proplists:to_map(proplists:normalize(Proplist, Stages))`
vs.
`proplists:to_map(Proplist, Stages)`
e.g:
``` erlang
proplists:to_map(Ps, proplists:normalize(Ps, [{negations, [{no_foo, foo}]}])),
%% vs.
proplists:to_map(Ps, [{negations, [{no_foo, foo}]}]).
```
And the question is also; is there something else that merits a second argument to `to_map()`?

I cannot think of anything close to as useful, so I think it is worth implementing.  See if I can convince @jhogberg...

RaimoNiskanen:
#### `to_map/2`, again
@jhogberg was not hard to convince.  He said something like "It is syntactical sugar that is not in the way.  Extending a function we want with something logical is not a problem compared to adding a function that is redundant."... :-)

Also, it becomes a natural place where there will be a pointer in the documentation to `normalize/2`.
Programmer needs and finds `to_map/1`, sees `to_map/2`, notices the pointer to `normalize/2`, realizes that there is where all fancy features plug in starting with negations...

Maria-12648430:
I updated the PR with the requested changes (the variant you came up with before falling asleep ;)). Passes the test suite I came up with, see below for details on that.
I believe the spec for `to_map` is still correct, yes?

> Precisely. One wants to say: ...that returns a value, but since the value may be any term, to know that it is _the value_ it has to be _not the default_ when one is sure that _the default_ =/= _the value_. I tried but could not get it both correct and graspable... :-(

Yeah, languages for humans are hard ;)

> #### Intentional?
> ...
> So this is indeed intentional...

Yeah. What surprises me is that (n!=2)-tuples are considered special (or property-ish) instead of "other values", I can't imagine where this could be useful =^^= Don't get me wrong, I'm not complaining, just wondering ;)

> #### Test coverage
> Sorry about the current bad test coverage. It will not be much work to beat that ;-)

I updated the test suite with a `map_conversion` test, that covers a few simple cases, plus all 5040 variations of a proplist with `a`-keyed properties and property-ish-ies. Let me know if that way of testing is not ok or something ;)

> #### `to_map/2`
> `proplists:to_map(proplists:normalize(Proplist, Stages))`
> vs.
> `proplists:to_map(Proplist, Stages)`
> e.g:
> 
> ```erlang
> proplists:to_map(Ps, proplists:normalize(Ps, [{negations, [{no_foo, foo}]}])),
> %% vs.
> proplists:to_map(Ps, [{negations, [{no_foo, foo}]}]).
> ```

Ok, you win ;)

> #### `to_map/2`, again
> @jhogberg was not hard to convince. He said something like "It is syntactical sugar that is not in the way. Extending a function we want with something logical is not a problem compared to adding a function that is redundant."... :-)

Ok then :)
I'll put that in a separate commit, probably tomorrow or on monday.

Maria-12648430:
`to_map/2` is up :)

Documentation is a bit minimalistic, not much more than pointers to `normalize/2` and `to_map/1`, but that should be ok, right? No use repeating what is already there.

The test is minimalistic as well, it just ensures that normalizations are applied, ie that `proplists:to_map/2` results in the same as `proplists:to_map/1` applied to the result of `proplists:normalize/2`, ok?

RaimoNiskanen:
> I believe the spec for `to_map` is still correct, yes?

I believe so too.

> What surprises me is that (n!=2)-tuples are considered special (or property-ish) instead of just as (ignored) other values, I can't imagine where this could be useful =^^= Don't get me wrong, I'm not complaining, just wondering ;)

The general idea is apparently that element 1 is the key in a tuple, and if it is a 2-tuple element 2 is the value.  For larger tuples the module should not assume what is the value. `get_keys/1`, `is_defined/2`, `lookup/2`, `lookup_all/2` and `split/2` handles arity >= 1 tuples by only handling the key or by returning the whole tuple. `append_values/2`, `get_all_values/2`, `get_bool/2`, and `get_value/2,3` handle only tuples with value, i.e 2-tuples.

Nice contribution to the test suite!  Excellent!

I have a few nitpicks that I will comment inline, but I have placed the PR in "Testing" anyway...

lib/stdlib/src/proplists.erl
```diff
@@ -668,3 +669,78 @@ split([], Store, Rest) ->
...
+      Key :: term(),
+      Value :: term().
+
+to_map(List) ->
+    lists:foldr(
+        fun
+            ({K, V}, M) ->
+                M#{K => V};
+            (T, M) when 1 =< tuple_size(T) ->
+                maps:remove(element(1, T), M);
```
RaimoNiskanen, line 707: I think a comment here explaining that arity =/= 2 tuples with matching key causes early return in `get_value()` would be appropriate.

lib/stdlib/src/proplists.erl
```diff
@@ -668,3 +669,78 @@ split([], Store, Rest) ->
...
+      Operation :: {'aliases', Aliases}
+                 | {'negations', Negations}
+                 | {'expand', Expansions},
+      Aliases :: [{Key, Key}],
+      Negations :: [{Key, Key}],
+      Expansions :: [{Property :: property(), Expansion :: [term()]}],
+      Map :: #{term() => term()}.
+
+to_map(List, Stages) ->
+    to_map(normalize(List, Stages)).
```
RaimoNiskanen, line 735: Since the last clause in `normalize/2` is `compact(List)`, and that is an annoyingly unnecessary pass before `to_map/1`, I thought about implementing `to_map/2` as a copy of `normalize/2` but with the final clause being `to_map(List)` instead. That would OTOH be an annoying code duplication to save one list rebuild.

I am torn here.  Do as you please with that idea.

lib/stdlib/test/proplists_SUITE.erl
```diff
@@ -77,4 +77,83 @@ examples(_Config) ->
...
+    %% Ensure that maps:get/3 using the created map yields the same
+    %% results as proplists:get_value/3 on the original proplist does,
+    %% and that proplists:get_value/3 on a proplist created from the
+    %% map yields the same results as proplists:get_value/3 on the
+    %% original proplist, ie they either all return the same `Value',
+    %% or they all return the `Default' given as respective third argument.
+    Default1 = make_ref(),
+    Default2 = make_ref(),
+    Default3 = make_ref(),
+    InList=[a, b, {a, 1}, {}, {a}, {a, 1, 2}, "foo"],
```
RaimoNiskanen, line 109: I think a `{c}` or `{c,1,2}` term in the list would be nice, i.e a tuple with a key that has no 2-tuple that can shadow it.

lib/stdlib/test/proplists_SUITE.erl
```diff
@@ -77,4 +77,83 @@ examples(_Config) ->
...
+    lists:foreach(
+        fun (L1) ->
+            LKs = proplists:get_keys(L1),
+            M = proplists:to_map(L1),
+            L2 = proplists:from_map(M),
+            [] = maps:keys(M) -- LKs,
+            [] = proplists:get_keys(L2) -- LKs,
```
RaimoNiskanen, line 116: Here you could instead compare `maps:keys(M)` with `proplists:get_keys(L2)` for equality, after sorting both.

lib/stdlib/test/proplists_SUITE.erl
```diff
@@ -77,4 +77,83 @@ examples(_Config) ->
...
+                fun (K) ->
+                    case
+                        {
+                            maps:get(K, M, Default1),
+                            proplists:get_value(K, L1, Default2),
+                            proplists:get_value(K, L2, Default3)
+                        }
+                    of
+                        {V, V, V} -> true;
+                        {Default1, Default2, Default3} -> true
```
RaimoNiskanen, line 127: To me swapping these two clauses feels more right.  Then it should also be sufficient with just one `Default` value instead of 3.

Maria-12648430:
I'll see to it all tomorrow :)

lib/stdlib/src/proplists.erl
Maria-12648430: Yeah, that would be appropriate I think :)

lib/stdlib/src/proplists.erl
Maria-12648430: Hm, I'm undecided too... How about a helper function that does "normalize-without-compact"? `normalize/2` can call it and pass the result into `compact/1`, and `to_map/2` can call it and pass the result into `to_map/1`. What do you think?

lib/stdlib/test/proplists_SUITE.erl
Maria-12648430: Ok, will do :) I had more in there, but the number of variations is `n!` (with `n` being the number of items), so I had to cut back a little ;) But one more is ok I guess.

lib/stdlib/test/proplists_SUITE.erl
Maria-12648430: Ok

lib/stdlib/test/proplists_SUITE.erl
Maria-12648430: Yes, good idea ;)

lib/stdlib/src/proplists.erl
Maria-12648430: OTOH, `normalize` itself is pretty lean, probably not worth an extra helper function... Ah well, I'll sleep on it ;)

RaimoNiskanen: Maybe worth a helper function since that makes it obvious what is common, and impossible to by mistake do some change to just one of the normalizing functions.  Finding a pretty name is harder...

lib/stdlib/test/proplists_SUITE.erl
RaimoNiskanen: Ouch!  Ok.  It could maybe replace the `{}` term if the size is a problem.

lib/stdlib/src/proplists.erl
Maria-12648430: It won't be exported, so the name doesn't need to be pretty enough to name children after it ;) Would `apply_stages` do?

RaimoNiskanen: `apply_stages` - certainly!

lib/stdlib/src/proplists.erl
Maria-12648430: Done, does that suffice?

lib/stdlib/src/proplists.erl
Maria-12648430: Done :)

lib/stdlib/test/proplists_SUITE.erl
Maria-12648430: Ok, I just added `{c, 1, 2}`, the addition raises then number of combinations from 5040 to 40320, but that's still ok I think.

lib/stdlib/test/proplists_SUITE.erl
Maria-12648430: Done

lib/stdlib/test/proplists_SUITE.erl
Maria-12648430: And done :)

RaimoNiskanen:
Excellent!  I reapplied the Testing label, and will hopefully be able to merge tomorrow.

Maria-12648430:
> Excellent! I reapplied the Testing label, and will hopefully be able to merge tomorrow.

Great :)

Maria-12648430:
I just had an idea for the test suite, about the `n!` permutations (not _variations_ as I said earlier) of the test list (see above).

The way it is now, the list comprehension builds _all_ possible permutations and then iterates over them. This drags a bit on memory, and each added item in the list increases that, heavily. Also, adding items means adding more steps to that list comprehension staircase (though I think it looks kinda pretty ;)).

The staircase thing could be removed by switching to generating the permutations like [this](http://erlang.org/doc/programming_examples/list_comprehensions.html#permutations), but not the drag on memory.

So, I came up with this idea:
```erlang
pm_fold(Fun, Acc0, L) ->
        pm_fold(Fun, Acc0, L, []).

pm_fold(Fun, Acc, [], Mut) -> Fun(Mut, Acc);
pm_fold(Fun, Acc, L, Mut) ->
        lists:foldl(
                fun
                        (X, AccIn) -> pm_fold(Fun, AccIn, L--[X], [X|Mut])
                end,
                Acc,
                L
        ).
```
In a nutshell what this does is "fold as you go": As it generates the permutations, it runs the given function _as soon as it has created one_, then generates the next, and so on, instead of first generating _all_ possibilities and _then_ iterating over them. It has the additional benefit of stopping the generation early when the given function fails for the first time.
In the test suite, `pm_fold` could be given the same function as is given to `lists:foreach` now.

If you wonder why I implemented this with fold capabilities, instead of as a simple foreach (which would suffice for purpose of the map conversion test), this is because I think this function should be internally tested itself, as it is a bit complex. This would be hard to do with a sideeffects-only implementation like a foreach.

Let me know what you think :)

RaimoNiskanen:
I think that is a better idea.  40 k iterations is not much, but the amount of memory to hold all permutations can be awkward.  It should be fairly straightforward to do with explicit iteration functions to avoid the rather expensive `--` operation too.

Maria-12648430:
> It should be fairly straightforward to do with explicit iteration functions to avoid the rather expensive `--` operation too.

Uh... I have been thinking for a while now, but sorry, I don't see how... ^^;;; Because what is needed there is the entire list `L`, except the respective elements `X`, not just a remainder of `L` or something. I can exchange `L--[X]` for `lists:delete(X, L)`, but I don't know if that is better or if it is what `--` does internally anyway.

RaimoNiskanen:
Ok.  Maybe I hallucinated that.  Might give it a go tomorrow to figure out where I was mistaken.

`lists:delete/2` is a small loop in Erlang.  `--` is the same as `erlang:'--'/2` which is the same as `lists:subtract/2` which is quite a large built in function written in C that I suspect is more geared towards not being inefficient for large lists.  For small lists I suspect `lists:delete/2` have smaller overhead.  Anyway, you are just deleting one element, so I also think it is clearer to use such a function.

Maria-12648430:
> Ok. Maybe I hallucinated that. Might give it a go tomorrow to figure out where I was mistaken.
> 
> `lists:delete/2` is a small loop in Erlang. `--` is the same as `erlang:'--'/2` which is the same as `lists:subtract/2` which is quite a large built in function written in C that I suspect is more geared towards not being inefficient for large lists. For small lists I suspect `lists:delete/2` have smaller overhead. Anyway, you are just deleting one element, so I also think it is clearer to use such a function.

Ok, I updated the commit with the changes :)

RaimoNiskanen:
Nope.  It was not a hallucination.  Here is a version:
``` erlang
pm_fold(Fun, Acc, List) ->
    pm_fold(Fun, Acc, List, [], []).
%%
pm_fold(Fun, Acc, [], [], Pm) ->
    Fun(Pm, Acc);
pm_fold(_Fun, Acc, [], _R, _Pm) ->
    Acc;
pm_fold(Fun, Acc, [X|L], R, Pm) ->
    pm_fold(
      Fun, pm_fold(Fun, Acc, L, [X|R], Pm),
      lists:reverse(R, L), [], [X|Pm]).
```
It does away with the `lists:delete/2` call by composing the list with the deleted element from the parts before and after, which probably has a small performance gain.  Use it if you like it.

Maria-12648430:
> It does away with the lists:delete/2 call by composing the list with the deleted element from the parts before and after, which probably has a small performance gain. Use it if you like it.

I had been thinking of this "shovelling from `R` to `L`" approach actually, but discarded the idea again because it would then involve a `++`. You are using the reverse-with-append BIF (`lists:reverse/2`, which I never recognized until just recently) for this purpose, I see. But reversal is not necessary, the order of elements is not important, it only influences the order in which the permutations are generated. Now I'm not sure `lists:reverse/2` does anything else (in terms of performance) than basically reverse and then `++`, but on the surface, `++` alone would be sufficient.

Long story short, I think I will stay with my version for now, if you don't mind ;) I'm currently making a list of things I could do to keep my head busy if the lockdown over here continues to continue, and "improve the `proplists` test suite" is on that list, so if I decide to indulge in this, I might switch to your implementation then ;)

RaimoNiskanen:
`lists:reverse(R, L)` actually showels from `R` to `L`, causing `R` to be reversed with `L` as tail, so it is as efficient as it can be.

`R ++ L` builds a new front list from the elements in `R` in the reverse order, with `L` as tail.  This could be done with `lists:reverse(lists:reverse(R), L)`, or with a body recursive function that build the result in the right order, but is like `lists:reverse/2` a built in C function that creates the result directly on the heap.

So both create as many new cons cells as there are elements in `R`, both are built in C functions.  They actually act very similar in this case.

I guess I just have an old habit of avoiding `++` because it is so short to write, easy to misuse, and might once have been much less efficient than `lists:reverse/2`.

> Long story short, I think I will stay with my version for now, if you don't mind ;)

No problem.

If you want a big problem to chew on; it would very nice to have property based testing e.g with PropEr of the `proplists` module. ;-)

Maria-12648430:
> `lists:reverse(R, L)` ...

Heh, I'm learning a lot and gain many new insights in the course of these conversations, thanks =)

> I guess I just have an old habit of avoiding `++`

Yes, me too, that was what diverted me from the "shovelling" approach 

> If you want a big problem to chew on; it would very nice to have property based testing e.g with PropEr of the `proplists` module. ;-)

Heh... well, I can try, but I doubt that it will turn out very good. I tried property testing before, but writing good properties is really hard if you're not used to it, so I usually just trash it and go with eunit/ct ^^;

RaimoNiskanen:

Maria-12648430:
My pleasure :)

PR #2911, opened 2020-12-07
Change the taskbar api to be forward compat

dgud:
The API will change to an option list in OTP-24,
so change the API now before it is used.


dgud:
@dominicletz  I'm chaning the API on maint.

dominicletz:
perfect! thanks @dgud

PR #2913, opened 2020-12-07
crypto: Fix ENGINE leaks

sverker:
and robustify the NIFs against faulty usage.

* Add `ctx->is_functional` to know if ENGINE_init has been called  and if ENGINE_finish needs to be called.
* Add cleanup to destructor as fallback if engine_free/finish_nif has not been called.
* Set `ctx->engine=NULL` when freed and check it after all calls to enif_get_resource.

Removed memory leaks reported by address-sanitizer:

Test case `engine_SUITE:ctrl_cmd_string`

Direct leak of 192 byte(s) in 1 object(s) allocated from:
    #0 0x7f955fb64b40 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.4+0xdeb40)
    #1 0x7f9556c0accb in CRYPTO_malloc crypto/mem.c:222
    #2 0x7f9556c0acfe in CRYPTO_zalloc crypto/mem.c:230
    #3 0x7f9556bd227e in ENGINE_new crypto/engine/eng_lib.c:34
    #4 0x7f9556bd345c in ENGINE_by_id crypto/engine/eng_list.c:301
    #5 0x7f9556fc1965 in engine_by_id_nif otp/lib/crypto/c_src/engine.c:138
    #6 0x562b780d4e8a in process_main x86_64-unknown-linux-gnu/asan/smp/beam_cold.h:184
    #7 0x562b7810f1eb in sched_thread_func beam/erl_process.c:8486
    #8 0x562b787a1b9b in thr_wrapper pthread/ethread.c:122
    #9 0x7f955f0a26da in start_thread (/lib/x86_64-linux-gnu/libpthread.so.0+0x76da)

### DIFF
--- a/lib/crypto/c_src/engine.c
+++ b/lib/crypto/c_src/engine.c
@@ -23,6 +23,7 @@
 #ifdef HAS_ENGINE_SUPPORT
 struct engine_ctx {
     ENGINE *engine;
+    int is_functional;
     char *id;
 };
 
@@ -44,6 +45,12 @@ static void engine_ctx_dtor(ErlNifEnv* env, struct engine_ctx* ctx) {
         enif_free(ctx->id);
     } else
          PRINTF_ERR0("  empty ctx->id=NULL");
+
+    if (ctx->engine) {
+        if (ctx->is_functional)
+            ENGINE_finish(ctx->engine);
+        ENGINE_free(ctx->engine);
+    }
 }
 
 int get_engine_and_key_id(ErlNifEnv *env, ERL_NIF_TERM key, char ** id, ENGINE **e)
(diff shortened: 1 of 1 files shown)


HansN (approved):
Seem good to me.

lthor (approved):
Looks ok.

sverker:
>  (given a memory leak of 192kb per object)?

The reported leak was 192 **bytes** per object, not kilo bytes.

As I did a small redesign that made the leak go away, I don't know the exact scenario that triggered this leak. It's test case `engine_SUITE:ctrl_cmd_string` invoking `engine_by_id_nif`. Maybe @HansN or @lthor have an idea?

HansN:
The fixed leak occured when one removes (I'm not sure of the terminology) an engine.  If that could be ordered by an adversary depends on the application using the crypto/engine interface.

If that user's application has an operation that an adversary could trigger, and that operation unloads (removes?, not sure of the terminology) the engine, the leakage of 192 bytes is a fact.

Maybe @lthor could comment on the terminology or correct me if I'm wrong.

lthor:
I agree with @HansN that this will only be a problem if an application let some external source trigger unloading the engine.

PR #2914, opened 2020-12-08
EDoc spec cleanup

erszcz:
This extracts spec/type changes from #2803 for easier review. No functional changes to EDoc are made in this PR.

### DIFF
--- a/lib/edoc/src/edoc.erl
+++ b/lib/edoc/src/edoc.erl
@@ -60,23 +60,71 @@
 	 read_comments/1, read_comments/2,
 	 read_source/1, read_source/2]).
 
--compile({no_auto_import,[error/1]}).
+-export_type([module_meta/0,
+	      env/0,
+	      comment/0,
+	      entry/0,
+	      entry_data/0,
+	      tag/0,
+	      xmerl_module/0]).
 
 -include("edoc.hrl").
 
+-type module_meta() :: #module{name :: [] | atom(),
+			       parameters :: none | [atom()],
+			       functions :: ordset(function_name()),
+			       exports :: ordset(function_name()),
+			       attributes :: ordset({atom(), term()}),
+			       records :: [{atom(), [{atom(), term()}]}],
+			       encoding :: epp:source_encoding()}.
+%% Module information.
 
-%% @spec (Name::filename()) -> ok
+-type env() :: #env{}.
+%% Environment information needed by EDoc for generating references.
+
+-type comment() :: #comment{line :: integer(),
+			    text :: string()}.
+%% Simplified comment data.
+
+
+-type entry() :: #entry{name :: function_name() | atom(),
+			args :: [atom()],
+			line :: integer(),
+			export :: boolean(),
+			data :: entry_data()}.
+%% Module Entries (one per function, plus module header and footer).
+
+-type entry_data() :: term().
+
+-type tag() :: #tag{name :: atom(),
+		    line :: integer(),
+		    origin :: comment,
+		    data :: term()}.
+%% Generic tag information.
+
+-type xmerl_module() :: any().
+%% The EDoc documentation data for a module,
+%% expressed as an XML document in {@link //xmerl. XMerL} format. See
+%% the file <a href="edoc.dtd">`edoc.dtd'</a> for details.
+
+-type ordset(T) :: ordsets:ordset(T).
+-type function_name() :: {atom(), integer()}.
+-type filename() :: file:filename().
+-type proplist() :: proplists:proplist().
+-type comment2() :: { Line :: integer(),
+		      Column :: integer(),
+		      Indentation :: integer(),
+		      Text :: [string()] }.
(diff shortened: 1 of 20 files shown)


erszcz:
FYI, I'm working on the broken links.

erszcz:
Apparently, it's not possible to take just the spec/type changes. I tried to make them as nonintrusive as possible.

lib/xmerl/src/xmerl_scan.erl
```diff
@@ -140,11 +140,14 @@
 %% helper functions. To xmerl_lib ??
 -export([accumulate_whitespace/4]).
 
+-export_type([xmlElement/0]).
+
 %-define(debug, 1).
 -include("xmerl.hrl").		% record def, macros
 -include("xmerl_internal.hrl").
 -include_lib("kernel/include/file.hrl").
```
garazdawi, line 149: @lthor do you have any feedback about the addition of this type?

garazdawi (commented):
I think this looks good.

Do we need tests for the app inference logic, or is it tested by building docs for Erlang/OTP?

erszcz:
> Do we need tests for the app inference logic, or is it tested by building docs for Erlang/OTP?

For OTP apps, if the links were wrong CI would fail. That's why https://github.com/erlang/otp/pull/2914/commits/de12adfa521e938694933bdbe94b576e1966e556 originally failed with "Broken links" and "References to missing anchors".

In general, this inference is used when from one app (the app currently processed by EDoc) we refer to a type defined in a module outside the app. If the external module is preloaded or not found under `external_app-1.2.3/ebin/external_mod.beam`, the inference will return `no_app`, therefore the generated link will lead to the currently processed app. This is wrong, but also quite rare and I'm not sure how to address it :| For example, this would occur if in `edoc.erl` we had `-type t() :: zlib:zflush().`, since `zlib` is preloaded.

Outside OTP, it should work fine for as long as we refer to modules in properly structured apps, i.e. following `app-1.2.3/ebin/mod.beam.`

erszcz:
> This is wrong, but also quite rare and I'm not sure how to address it :| 

BTW, this is also the status quo if the inference isn't merged in.

lib/edoc/src/edoc.erl
```diff
@@ -60,23 +60,71 @@
...
+-type edoc_module() :: xmerl_scan:xmlElement().
+%% The EDoc documentation data for a module,
+%% expressed as an XML document in {@link //xmerl. XMerL} format. See
+%% the file <a href="edoc.dtd">`edoc.dtd'</a> for details.
+
+-type ordset(T) :: ordsets:ordset(T).
+-type function_name() :: {atom(), integer()}.
+-type filename() :: file:filename().
+-type proplist() :: proplists:proplist().
+-type comment2() :: { Line :: integer(),
```
garazdawi, line 114: Can we give this a better name than `comment2`?

erszcz: Definitely, just a second...

erszcz: I've pushed a rename. I've also found out that this type is actually an alias of `erl_comment_scan:comment()`, which is exported already. The data instances are read by `erl_comment_scan:read_comments()`. Should I mark it as `-type comment() :: erl_comment_scan:comment()`?

garazdawi: yes, I think that would be better.

erszcz:
> If the external module is preloaded or not found under `external_app-1.2.3/ebin/external_mod.beam`, the inference will return `no_app`

It just occurred to me that some preloaded modules end up having a corresponding `.beam` file under the `erts` app:

```
$ find ~/work/erlang/build.otp -name zlib.beam
/Users/erszcz/work/erlang/build.otp/lib/erlang/lib/erts-11.1.3/ebin/zlib.beam
```

 If that holds for all preloaded modules, then the inference could be improved.

garazdawi:
They are there for tools (such as dialyzer) to be able to get the debug_info when analyzing a released system. I don't think that they are guaranteed to be there, but it may very well be the case that in the scenarios that count they are.

erszcz:
I've added a test following the above assumptions, with the small difference that when tests are run from the source tree, `code_lib:dir(erts)` doesn't have a direct `ebin` subdir. This change should make it possible to properly link to types like `erlang:spawn_opt_type/0` in edoc comments.

erszcz:
Hmm, the last build failed, but it seems to be an environment issue - `Error: read ECONNRESET` - not actually connected with the new changes 🤔

garazdawi:
I restarted the jobs

garazdawi:
I'm happy with these changes now. Would you mind reworking the commit history so that there are only functional working commits? When you are done I will merge this.

garazdawi:
I'll put it in testing once more just to make sure nothing has been broken by the rebase and then merge tomorrow.

erszcz:
Sounds good, thanks for the review!

PR #1987, opened 2018-10-16
Add an option to control the paths used in -include_lib()

jhogberg:
This PR gives the user control over the paths used in `-include_lib()` directives instead of relying on the `ERL_LIBS` of the build environment. It's not airtight since `-behavior()` and parse transformations are still handled with the build environment's paths, but it's a step in the right direction.

More details can be found in [ERL-711](https://bugs.erlang.org/browse/ERL-711).

PR #1274, opened 2016-12-09
Remove behaviour checking from erl_lint

richcarl:
`<tongue-in-cheek>`
Since the Erlang compiler should be pure, not inspecting other beam files during compilation, and since its warnings should be dependable and deterministic, this patch removes the behaviour checking from the Erlang linter. Such checks should instead be performed by a separate tool that has a whole-system context to work with.

No backwards compatibility option is implemented, since this can only be considered a bug fix.
`</tongue-in-cheek>`

PR #2915, opened 2020-12-08
erl_lint: Remove -behaviour() checking

jhogberg:
`-behaviour()` requires that the behavior module is built before the module we're compiling, and that it can be found in the build system's code path (which may have no relation to the host system). This is very annoying to deal with so we've decided to remove it as `dialyzer` already does this check in a far better fashion than the linter.

The only thing `dialyzer` lacks is that it doesn't understand the `behavior_info/1` callback (and raises a warning to that effect), but we consider that less important than fixing this issue. We can fix that separately should it become a problem.

### DIFF
--- a/lib/stdlib/src/erl_lint.erl
+++ b/lib/stdlib/src/erl_lint.erl
@@ -353,23 +353,6 @@ format_error({bad_bitsize,Type}) ->
 format_error(unsized_binary_in_bin_gen_pattern) ->
     "binary fields without size are not allowed in patterns of bit string generators";
 %% --- behaviours ---
-format_error({conflicting_behaviours,{Name,Arity},B,FirstL,FirstB}) ->
-    io_lib:format("conflicting behaviours - callback ~tw/~w required by both '~p' "
-		  "and '~p' ~s", [Name,Arity,B,FirstB,format_where(FirstL)]);
-format_error({undefined_behaviour_func, {Func,Arity}, Behaviour}) ->
-    io_lib:format("undefined callback function ~tw/~w (behaviour '~w')",
-		  [Func,Arity,Behaviour]);
-format_error({undefined_behaviour,Behaviour}) ->
-    io_lib:format("behaviour ~tw undefined", [Behaviour]);
-format_error({undefined_behaviour_callbacks,Behaviour}) ->
-    io_lib:format("behaviour ~w callback functions are undefined",
-		  [Behaviour]);
-format_error({ill_defined_behaviour_callbacks,Behaviour}) ->
-    io_lib:format("behaviour ~w callback functions erroneously defined",
-		  [Behaviour]);
-format_error({ill_defined_optional_callbacks,Behaviour}) ->
-    io_lib:format("behaviour ~w optional callback functions erroneously defined",
-		  [Behaviour]);
 format_error({behaviour_info, {_M,F,A}}) ->
     io_lib:format("cannot define callback attibute for ~tw/~w when "
                   "behaviour_info is defined",[F,A]);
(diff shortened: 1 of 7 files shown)


essen:
What made you decide to remove it specifically? What's "very annoying to deal with"?

jhogberg:
For starters, it prevents `erlc` from building files in parallel as they may depend on each other and there's no way to tell before they've all been built.

It also means that it's unsafe to compile more than file one at a time on a given node as they may need different code paths. This has been a massive headache for `rebar3`, and severely reduces the performance of `erlc`'s compile server option as we need to restart the server node every time `erlc` is invoked from a new directory.

There are of course other dependencies on the code path (notably parse transformations and `-include_lib()`) but we figured we'd start here.

bjorng (approved):

essen:
> For starters, it prevents `erlc` from building files in parallel as they may depend on each other and there's no way to tell before they've all been built.

That doesn't sound right unless you're expecting all users to only call `erlc $ALL_FILES` and not do multiple invocations with dependencies figured out beforehand (for example via `erlc -M`).

> It also means that it's unsafe to compile more than file one at a time on a given node as they may need different code paths. This has been a massive headache for `rebar3`, and severely reduces the performance of `erlc`'s compile server option as we need to restart the server node every time `erlc` is invoked from a new directory.

Are there individual files in a single application that need to be compiled with different paths? I don't think I've seen an application like this. So I'm assuming you mean compile different applications in parallel. I fail to see how this PR would help, unless if you're planning to remove parse transforms/include_lib as well.

> There are of course other dependencies on the code path (notably parse transformations and `-include_lib()`) but we figured we'd start here.

You probably should remove the behaviors from `erlc -M` as well if that's the way you're going for. They're no longer a compile time dependency with this PR.

jhogberg:
> That doesn't sound right unless you're expecting all users to only call erlc $ALL_FILES and not do multiple invocations with dependencies figured out beforehand (for example via erlc -M).

`erlc -M` cannot (and doesn't even attempt to) support `-behaviour()` since we don't know which file to depend on before the behavior module is built. All we know is that it's supposed to be in the code path, but where?

> Are there individual files in a single application that need to be compiled with different paths? I don't think I've seen an application like this. So I'm assuming you mean compile different applications in parallel.

Not that I know of, but it's a problem when compiling different applications serially in the same node. This is a major headache for `rebar3`.

> I fail to see how this PR would help, unless if you're planning to remove parse transforms/include_lib as well.

We're working on a clearer division between compile- and run-time dependencies where tools like `rebar3` can (optionally) provide appropriate paths for `-include_lib()` and handle parse transformations on its own. This PR is a first step to that goal, and while I suppose we could make `-behaviour()` checking optional too, we'd prefer not to have this code in the linter.

richcarl:
Here's some more background discussion for the interested: https://github.com/erlang/otp/pull/1274

essen:
Thanks.

I'm just not sure that removing this is helpful considering the problem remains for other elements, and it sounds like solving those would solve behavior checks as well.

> > That doesn't sound right unless you're expecting all users to only call erlc $ALL_FILES and not do multiple invocations with dependencies figured out beforehand (for example via erlc -M).
> 
> `erlc -M` cannot (and doesn't even attempt to) support `-behaviour()` since we don't know which file to depend on before the behavior module is built. All we know is that it's supposed to be in the code path, but where?

Too bad. Didn't know. I was hoping to be able to use it in the near future. I think in my case it would work out though because it's either a beam file in the code path (so you can find it, not sure what you mean by "where") or given as a source file as argument.

jhogberg:
> I'm just not sure that removing this is helpful considering the problem remains for other elements, and it sounds like solving those would solve behavior checks as well.

This is how they're solved, though.

Granted, we can't tear out parse transformations nor `-include_lib()`, but we can provide options that disable them (e.g. #1987) and let build tools like `rebar3` or the compile server handle it themselves. I'm open to having an option, but I'd rather not keep this baggage in the linter when other tools can do a much better job.

> I think in my case it would work out though because it's either a beam file in the code path (so you can find it, not sure what you mean by "where") or given as a source file as argument.

Is it `./foo.beam`, `app1/ebin/foo.beam`, `app2/ebin/foo.beam`, or `/opt/erlang/lib/stdlib-x.y.z/ebin/foo.beam`? You have no idea where it will be until it's built, and then you have the fun task of figuring out the source file name of said module, which is harder than it sounds in the presence of `+file` and `+deterministic`. :-/

essen:
> > I'm just not sure that removing this is helpful considering the problem remains for other elements, and it sounds like solving those would solve behavior checks as well.
> 
> This is how they're solved, though.
> 
> Granted, we can't tear out parse transformations nor `-include_lib()`, but we can provide options that disable them (e.g. #1987) and let build tools like `rebar3` or the compile server handle it themselves. I'm open to having an option, but I'd rather not keep this baggage in the linter when other tools can do a much better job.

It doesn't have to be in the linter. It was always odd to have this there. I'm not against the removal per se, it can most likely be done after compilation in one way or another without having to invoke Dialyzer. Thanks for the clarifications.

> > I think in my case it would work out though because it's either a beam file in the code path (so you can find it, not sure what you mean by "where") or given as a source file as argument.
> 
> Is it `./foo.beam`, `app1/ebin/foo.beam`, `app2/ebin/foo.beam`, or `/opt/erlang/lib/stdlib-x.y.z/ebin/foo.beam`? You have no idea where it will be until it's built, and then you have the fun task of figuring out the source file name of said module, which is harder than it sounds in the presence of `+file` and `+deterministic`. :-/

Right I get what you mean by "where" now. But it doesn't apply in my case because I only need to worry about dependencies between files inside one application. A behavior from another application has already been built and always will be. So in those cases I would just expect `erlc -M` to assume there's no dependency (same for includes and PTs that belong to other applications). But I suppose I'll keep the dependency resolution code around instead of using `erlc -M`. Anyway that's unrelated to this PR, I don't want to pollute.

kostis:
> The only thing `dialyzer` lacks is that it doesn't understand the `behavior_info/1` callback (and raises a warning to that effect)

FYI: this is more of an opinion from `dialyzer` than an fundamental limitation.

Dialyzer strongly prefers `-callback ` declarations to be present in the behaviour file than a `behaviour_info/1` function, which feels like (is?) a kludge, and can be easily generated from callback declarations anyway.  I know that the OTP team values "backwards compatibility" highly, but perhaps it's a good time to design a deprecation mechanism for `behaviour_info/1` 1-2 years down the road.

FWIW, we've published a [paper on the subject](https://www.researchgate.net/publication/221211388_Typed_callbacks_for_more_robust_behaviours) about a decade ago.  Perhaps its time has come :)

uabboli (approved):

PR #2917, opened 2020-12-08
Fix typo on Erl_Docgen User's Guide

paulo-ferraz-oliveira:
As per https://bugs.erlang.org/browse/ERL-1386.

### DIFF
--- a/lib/erl_docgen/doc/src/doc-build.xml
+++ b/lib/erl_docgen/doc/src/doc-build.xml
@@ -43,7 +43,7 @@
       </p>
       <code>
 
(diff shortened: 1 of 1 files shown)


lthor:
Hi Paulo,

I've added a branch to our internal testing yesterday that fixes this, it was also two more variables further down in another example that was wrong.
I just removed the parentheses in all cases so the change is consistent with the rest of the examples.

It will be released in  23.2 so I won't take your pull request, but thank you for the proposal.

BR Lars

paulo-ferraz-oliveira:
Sure thing. Thanks for the feedback.

PR #2918, opened 2020-12-08
Document compiler options warn_missing_spec and warn_missing_spec_all

paulo-ferraz-oliveira:
As discussed in https://bugs.erlang.org/browse/ERL-1430.


paulo-ferraz-oliveira:
It's still missing tests.

erts/test/erlc_SUITE.erl
```diff
@@ -432,6 +433,21 @@ make_dep_options(Config) ->
     false = exists(BeamFileName),
     ok.
 
+compile_options(Config) ->
+    {SrcDir, _OutDir, Cmd} = get_cmd(Config),
+
+    FilenameMissingOK = filename:join(SrcDir, "erlc_test_warn_missing_spec_ok.erl"),
+    run(Config, Cmd, FilenameMissingOK, [], true),
```
paulo-ferraz-oliveira, line 440: `true` is obviously wrong, here. I'm still trying to get tests running locally (for a single batch).

paulo-ferraz-oliveira: These are now fixed according to test expectation.

bjorng (changes requested):
Thanks for your pull request.

The documentation looks good.

The tests are more complicated than they need to be. We usually put tests for options in the application that implements them. In this case, the spec options are implemented in `erl_lint` (in STDLIB) and should be tested in `erl_lint_SUITE`.

paulo-ferraz-oliveira:
> Thanks for your pull request.
> 
> The documentation looks good.
> 
> The tests are more complicated than they need to be. We usually put tests for options in the application that implements them. In this case, the spec options are implemented in `erl_lint` (in STDLIB) and should be tested in `erl_lint_SUITE`.

Yeah. I was mislead in the bug report, but I'll fix that.

As for the tests, you prefer I make them simpler? I actually think they're as simple as possible, but I'm willing to change them, if you see it fit.

bjorng:
> As for the tests, you prefer I make them simpler?

I mean simpler in the sense that the scaffolding in `erl_lint_SUITE` will allow you to do the same tests using less code and without creating four extra Erlang modules. In `erl_lint_SUITE`, you can just put source code to be tested in one or more binaries and call the `run/2` function.

paulo-ferraz-oliveira:
> I mean simpler in the sense that the scaffolding in erl_lint_SUITE will allow you to do the same tests using less code

Sure thing. Let me simplify these proper. Thanks for your help.

paulo-ferraz-oliveira:
@bjorng: I squashed most of the previous stuff (now force-pushed), then simplified the test suite.

lib/stdlib/test/erl_lint_SUITE.erl
```diff
@@ -4448,6 +4449,46 @@ inline_nifs(Config) ->
            {warnings,[{2,erl_lint,nif_inline}]}}],
     [] = run(Config, Ts).
 
+compile_options(Config) ->
```
bjorng, line 4452: Most of the test cases in this module test compiler options. Please rename it to a more descriptive name.

lib/stdlib/test/erl_lint_SUITE.erl
```diff
@@ -4448,6 +4449,46 @@ inline_nifs(Config) ->
...
+           [warn_missing_spec],
+           []},
+
+    WarnMissSpecNotOkay
+        = {warn_missing_spec_error,
+           <<"-export([external/0]).
+              external() -> internal().
+              -spec internal() -> ok. internal() -> ok.">>,
+           [warn_missing_spec],
+           {warnings, [{2, erl_lint, {missing_spec, {external, 0}}}]}},
```
bjorng, line 4467: You can combine the two tests.

You are not testing that there will **not** be a warning for a local function without a spec.

Here is a suggestion (not tested):

```suggestion
    WarnMissSpec =
        {warn_missing_spec,
           <<"-export([external_with_spec/0,external_without_spec/0]).
              external_without_spec() -> internal_without_spec().
              spec externa_with_specl() -> ok.
              external_with_spec() -> internal_with_spec().
              -spec internal_with_spec() -> ok.
              internal_with_spec() -> ok.
              internal_without_spec() -> ok.>>,
           [warn_missing_spec],
           {warnings, [{2, erl_lint, {missing_spec, {external_without_spec, 0}}}]}},
```

lib/stdlib/test/erl_lint_SUITE.erl
```diff
@@ -4448,6 +4449,46 @@ inline_nifs(Config) ->
...
+           [warn_missing_spec_all],
+           []},
+
+    WarnMissSpecAllNotOkay
+        = {warn_missing_spec_all_error,
+           <<"-export([external/0]).
+              external() -> internal().
+              -spec internal() -> ok. internal() -> ok.">>,
+           [warn_missing_spec_all],
+           {warnings, [{2, erl_lint, {missing_spec, {external, 0}}}]}},
```
bjorng, line 4483: You are not testing that there will be a warning for a local function without a spec. Similar to the first one, these tests can be combined. Note that you can also reuse the identical source code (i.e. assign the binary with the source to a variable in the beginning of the function and use it in the tests for both `warn_missing_spec` and `warn_missing_spec_all`).

bjorng (changes requested):
Thanks for the update. I have some more suggestions for changes. When you have done those changes, please squash to a single commit.

lib/stdlib/test/erl_lint_SUITE.erl
paulo-ferraz-oliveira: Hm... I want to test the following:
1. no error for `warn_missing_spec` when everything's OK,
2. error for `warn_missing_spec` when there's something missing.

lib/stdlib/test/erl_lint_SUITE.erl
paulo-ferraz-oliveira: Hm... I want to test the following:
1. no error for `warn_missing_spec_all` when everything's OK (this includes internal functions),
2. error for `warn_missing_spec_all` when there's something missing (this includes internal functions).

lib/stdlib/test/erl_lint_SUITE.erl
paulo-ferraz-oliveira: In the second test (`WarnMissSpecNotOkay`), I'm making sure that `warn_missing_spec` doesn't care about internal specs.

lib/stdlib/test/erl_lint_SUITE.erl
paulo-ferraz-oliveira: In the second test (`WarnMissSpecNotOkay`), I'm making sure that `warn_missing_spec_all` also cares about external specs.

paulo-ferraz-oliveira: > Note that you can also reuse the identical source code

I can, and will, since this appears to be your general guidance. I just tend to make tests as atomic as possible so one output doesn't influence another and changing one "datum" doesn't affect more than one test.

paulo-ferraz-oliveira:
@bjorng, I think I've handled all your requests. I squashed all previous commits. Thanks for your help.

bjorng (approved):

bjorng:

bjorng:
Thanks for the pull request!

PR #2920, opened 2020-12-08
rand improvements: seed 'default' and generate binaries

RaimoNiskanen:
This PR introduces two improvements to the `rand` module:
* The possibility to seed using the current default algorithm, alias `default`
* Two functions that generate binaries: `bytes/1` and `bytes_s/2`

Stating a `default` algorithm has been deemed not needed since when you seed to reproduce a sequence, you must state the algorithm to determine the sequence.  There is at least one scenario, though, when this does not apply; when you _export_ the start seed (to be able to reproduce the sequence) before generating a sequence, and is happy with whatever algorithm that is the default.

This PR makes `seed/1`, `seed/2`, `seed_s/1` and `seed_s/2` accept the algorithm alias `default` for the default algorithm.

The new functions `bytes(N)` and `bytes_s(N, State)` generate a `binary()` of size `N`.  This is convenient precisely when you need a random binary, with the other properties that comes with the `rand` module such as reproducability and independence of the Crypto application.

### DIFF
--- a/lib/kernel/test/gen_tcp_misc_SUITE.erl
+++ b/lib/kernel/test/gen_tcp_misc_SUITE.erl
@@ -4993,7 +4993,7 @@ port_failed_str(Reason) ->
 %% Verifies that erl_check_io properly handles extra EPOLLIN signals.
 bidirectional_traffic(Config) when is_list(Config) ->
     Workers = erlang:system_info(schedulers_online) * 2,
-    Payload = crypto:strong_rand_bytes(32),
+    Payload = rand:bytes(32),
     {ok, LSock} = gen_tcp:listen(0, [binary, {packet, 0}, {active, false}, {reuseaddr, true}]),
     %% get all sockets to know failing ends
     {ok, Port} = inet:port(LSock),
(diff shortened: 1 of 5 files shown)


RaimoNiskanen:
Performance note: on my laptop the default algorithm for `rand:bytes(N)` is faster than `crypto:strong_rand_bytes(N)` for small `N`:s, up to about 100.  For large `N`:s it seems the crypto algorithm is about 4 times faster.

The last 2 commits changes 2 test suites in Stdlib and Kernel to use the introduced features, as examples.

sverker (approved):
Looks good.

The documentation says _"The default algorithm is ..."_. Could it say _"The **current** default algorithm is ..."_ or some other way make it clearer that default is not set and may change.

essen:
Because the default does not remain the same across versions, would there perhaps be any value in being able to retrieve it (via `rand:default()` or similar?

RaimoNiskanen:
@essen: Your whish becomes possible with this PR, although not in a dedicated function:
``` erlang
    {DefaultAlg, _State} = ExportState = rand:export_seed(rand:seed_s(default)),
```
To reproduce the sequence you need the whole `ExportState`, so using this detour to get the default algorithm might be good enough.

RaimoNiskanen:
@sverker: Fixed.  New commit introduces "default algorithm" as a concept and explains it a wee bit more.

mikpe:
I like this. I have a use-case that hasn't been mentioned here before: being able to generate random numbers _without_ affecting visible state. The issue is that `rand:seed_s/2` requires me to know the names of the available algorithms, which ones are deprecated, and effectively also which one is default. Those properties alas change frequently between major OTP releases. Being able to simply pass `default` there and let `rand` DTRT helps. Note: reproducibility is not a concern, I just want a side-effect free way to generate "good enough" random numbers.

PR #2922, opened 2020-12-09
erts: Support cgroup2 CPU quotas

jhogberg:
@tsloughter


tsloughter:
Awesome :+1:

lrascao:
Does this mean we no longer need to override the number of available schedulers when running in a Kubernetes environment?

jhogberg:
Yes.

tsloughter:
@lrascao note that was already the case, so you don't need the future OTP-24 release to do this. The majority of users of kubernetes or docker are on cgroup v1. Docker just added support in version 20.10.

This will keep it working for a future where users start to move to cgroup v2 in Kubernetes though :).

PR #2925, opened 2020-12-11
Add extended error information for all BIFs in the ERTS application

bjorng:
This pull request adds [extended error information](https://github.com/erlang/eep/blob/master/eeps/eep-0054.md) for all BIFs in the `erlang`, `atomics`, `counters`, and `persistent_term` modules.


erts/emulator/beam/bif.c
```diff
@@ -1934,7 +1937,7 @@ BIF_RETTYPE erts_internal_process_flag_3(BIF_ALIST_3)
                                         (void *) pf3a);
 
    if (is_non_value(res))
-       BIF_RET(am_badarg);
+       BIF_RET(am_badtype);
```
sverker, line 1940: @bjorng I think this should remain `am_badarg` as it's not the wrong type (pid), instead the process is dead. There is a similar case above in `exec_process_flag_3() ` where the process_flag signal is successfully sent but arrives in an exiting process, and there we return `am_badarg`.

This change also conflicts with #2930, but reverting to `am_badarg` will solve that too :-)

bjorng: Yes, I use `badtype` to indicate that the process is dead. (Since those atoms are not seen by the user, to avoid wasting atoms, I chose to use existing atoms that loosely have the meaning I want.)

The `badtype` returned from the `erts_internal` function is translated by the caller in the `erlang` module to a `badarg` exception but with the cause in the error info set to `badtype`:

https://github.com/erlang/otp/blob/2021da227534f09f5ba46b20de05d002171043c2/erts/preloaded/src/erlang.erl#L1690-L1698

Thus:

```
1> DeadPid = self(). a = b.
<0.83.0>
2> a = b.
** exception error: no match of right hand side value b
3> catch process_flag(DeadPid, save_calls, 20).
{'EXIT',{badarg,[{erlang,process_flag,
                         [<0.83.0>,save_calls,20],
                         [{error_info,#{cause => badtype,module => erl_erts_errors}}]},
                 {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,681}]},
                 {erl_eval,expr,5,[{file,"erl_eval.erl"},{line,434}]},
4> process_flag(DeadPid, save_calls, 20).      
     in function  process_flag/3
        called as process_flag(<0.83.0>,save_calls,20)
        *** argument 1: the pid does not refer to an existing process
```

PR #2926, opened 2020-12-11
Export the dst/0 type from erlang.erl

pergu:
The dst/0 type describes what types of values that it is possible to send messages to in Erlang. It would be useful to have one source of truth for this type when referring to it in other modules, since it is somewhat complex it is easy to miss acceptable values when defining it yourself.

### DIFF
--- a/erts/preloaded/src/erlang.erl
+++ b/erts/preloaded/src/erlang.erl
@@ -112,6 +112,9 @@
 -type iovec() :: [binary()].
 -export_type([iovec/0]).
 
+%% Type for the destination of sends.
+-export_type([dst/0]).
+
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 %% Native code BIF stubs and their types
 %% (BIF's actually implemented in this module goes last in the file)


paulo-ferraz-oliveira:
This should be documented under `erts/doc/src/erlang.xml`.

josevalim:
:+1: from me but I feel the current name is a bit cryptic if we are going to export it. Perhaps `process_dest` or `send_destination` would be clearer?

erts/doc/src/erlang.xml
```diff
@@ -248,7 +248,16 @@
...
     </datatype>
-    
+    <datatype>
+      <name name="send_destination"></name>
+      <desc>
+        <p>The destination for a send operation, can be a remote or local process identifier,
+          a (local) port, a locally registered name, or a tuple
+          <c>{<anno>RegName</anno>, <anno>Node</anno>}</c>
+        for a registered name at another node.</p>
+        <p>
```
rickard-green, line 258: You got a stray <p> here

rickard-green: Process alias should also be mentioned here (see alias() BIF)

rickard-green:
Please squash your commits into one commit and update the commit message. If necessary also rebase your branch (we don't want master merged into your branch).

rickard-green:
I've rebased and squashed the commits

PR #2929, opened 2020-12-12
Replace old test BIFs w/o is_ prefix in documentation code examples

dischoen:
As proposed in my email from erlang-questions as of 11.12.2020, here is a patch for the documentation.
It replaces usages of the old BIFs without the is_ prefix.

An initial find-grep for the BIFs showed over 300 occurrences, but finally I only found 4 occurrences in the code
examples.

There is no test case, as the patch does not affect real source code. ´
However, I tried the code examples in an erl shell.

dieter

### DIFF
--- a/lib/tools/doc/src/fprof.xml
+++ b/lib/tools/doc/src/fprof.xml
@@ -598,7 +598,7 @@
 -module(foo).
 -export([create_file_slow/2]).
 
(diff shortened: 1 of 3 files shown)


bjorng:
Thanks for your pull request.

dischoen:
You're welcome!

PR #2932, opened 2020-12-15
erl_gc: fix dynamic tracepoint for heap_grow

max-au:
If stack top is changed before calling heap_grow tracepoint,
the stack is not walk-able, because stack bottom points to
old heap, while stack top is already reset to new heap.

### DIFF
--- a/erts/emulator/beam/erl_gc.c
+++ b/erts/emulator/beam/erl_gc.c
@@ -697,10 +697,10 @@ garbage_collect(Process* p, ErlHeapFragment *live_hf_end,
     ErtsMonotonicTime start_time;
     ErtsSchedulerData *esdp = erts_proc_sched_data(p);
     erts_aint32_t state;
-    ERTS_MSACC_PUSH_STATE();
 #ifdef USE_VM_PROBES
     DTRACE_CHARBUF(pidbuf, DTRACE_TERM_BUF_SIZE);
 #endif
+    ERTS_MSACC_PUSH_STATE();
 
     ERTS_UNDEF(start_time, 0);
     ERTS_CHK_MBUF_SZ(p);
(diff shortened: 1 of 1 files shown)


garazdawi:
Wouldn't it make more sense to let the probe trigger after all the pointers have been updated to the new heap? That way the heap is also valid to traverse if one wants to.

Since we cannot write a testcase for this (I assume?), please add a comment in the code about why that line has to be after/before the dtrace probe.

max-au:
If I understand correctly, old and new heaps are expected to be identical. 

I can change it either way (e.g. move DTRACE invocation after HEAP_END(p) = n_heap + new_sz; or move it before     /* Copy stack to end of new heap */, or add a comment). What would be the preferred solution?

garazdawi:
> If I understand correctly, old and new heaps are expected to be identical.

Yes, but by the time the GC is done the previous heap will have been destroyed by GC forward pointers so it cannot be read, while the new heap will be intact.

> I can change it either way (e.g. move DTRACE invocation after HEAP_END(p) = n_heap + new_sz; or move it before /* Copy stack to end of new heap */, or add a comment). What would be the preferred solution?

I think it would be best to move the DTRACE call until after all of the heap pointers have been updated.

erts/emulator/beam/erl_gc.c
```diff
@@ -1791,6 +1782,15 @@ do_minor(Process *p, ErlHeapFragment *live_hf_end,
     HEAP_TOP(p) = n_htop;
     HEAP_SIZE(p) = new_sz;
     HEAP_END(p) = n_heap + new_sz;
+
+#ifdef USE_VM_PROBES
+    if (HEAP_SIZE(p) != new_sz && DTRACE_ENABLED(process_heap_grow)) {
```
garazdawi, line 1787: `HEAP_SIZE(p) != new_sz` will always be false here.

erts/emulator/beam/erl_gc.c
```diff
@@ -1882,6 +1873,18 @@ major_collection(Process* p, ErlHeapFragment *live_hf_end,
     HEAP_TOP(p) = n_htop;
     HEAP_SIZE(p) = new_sz;
     HEAP_END(p) = n_heap + new_sz;
+
+#ifdef USE_VM_PROBES
+    /* Fire process_heap_grow tracepoint after all heap references have
+     * been updated. This allows to walk the stack. */
+    if (HEAP_SIZE(p) != new_sz && DTRACE_ENABLED(process_heap_grow)) {
```
garazdawi, line 1880: `HEAP_SIZE(p) != new_sz` will always be false here.

garazdawi (changes requested):

PR #2933, opened 2020-12-16
tls_connection: spread RAM/CPU spike induced by flow control

max-au:
When Distribution over TLS is used in large clusters, many
connections are established upfront. All of them have the same
{active_n, 100} setting enabled. When 100 packets are received,
which tends to happen simultaneously for many connections,
a significant spike in RAM and CPU % is observed, due to the way
switching to 'passive' is currently implemented.
Starting with the random number between 1 and 100 helps to spread
the spike and avoid momentarily overload.
This patch also removes unnecessary configuration lookup when
tls_connection is established for Erlang distribution purposes.


IngelaAndin:
Thank you for the PR.

PR #2935, opened 2020-12-17
Added erlang:statistics(message_queues).

max-au:
Counts sum of all messages in all processes message queues. Includes both private message queue and incoming signal/message queue lengths, thus accounts for pending signals too.
Returns process ID that has accumulated the longest queue (or undefined, if all message queues are empty).

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -8219,6 +8219,24 @@ ok
     </func>
 
     <func>
+      <name name="statistics" arity="1" clause_i="20"
+        anchor="statistics_message_queues"  since="OTP 24.0"/>
+      <fsummary>Information about message queues.</fsummary>
+      <desc>
+        <p>Returns a tuple
+          <c>{<anno>MsgqSum</anno>, <anno>MsgqProcNonzero</anno>,
+            <anno>MsgqProcMax</anno>, <anno>MaxPid</anno>}</c>, where
+          <c><anno>MsgqSum</anno></c> is the sum of all message queue lengths
+          in the system, <c><anno>MsgqProcNonzero</anno></c> is
+          the number of processes with non-empty queue,
+          <c><anno>MsgqProcMax</anno></c> is the length of the longest
+          message queue, and <c><anno>MaxPid</anno></c> is the process
+          accumulated the longest queue, or <c>undefined</c> if there is no such
+          process (all queues are empty). </p>
+      </desc>
+    </func>
+
+    <func>
       <name name="suspend_process" arity="1" since=""/>
       <fsummary>Suspend a process.</fsummary>
       <desc>
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -3835,6 +3835,54 @@ BIF_RETTYPE statistics_1(BIF_ALIST_1)
 	    hpp = &hp;
 	}
     }
+    else if (ERTS_IS_ATOM_STR("message_queues", BIF_ARG_1)) {
+	Uint i;
+	Uint msgq_sum = 0;
+	Uint msgq_max = 0;
+	Eterm max_pid = am_undefined;
+	Uint msgq_nproc_nonzero = 0;
+	Eterm res, *hp, **hpp;
+	Uint sz, *szp;
+
+	erts_proc_unlock(BIF_P, ERTS_PROC_LOCK_MAIN);
+	erts_thr_progress_block();
+
+	for (i = 0; i < erts_ptab_max(&erts_proc); i++) {
+	    Process* p = erts_pix2proc(i);
+	    if (p) {
+		Uint len;
+		len = p->sig_qs.len + p->sig_inq.len;
+		if (len) {
+		    msgq_sum += len;
+		    msgq_nproc_nonzero++;
+		    if (len > msgq_max) {
+			msgq_max = len;
+			max_pid = p->common.id;
+		    }
+		}
+	    }
+	}
+
+	erts_thr_progress_unblock();
+	erts_proc_lock(BIF_P, ERTS_PROC_LOCK_MAIN);
+
+	sz = 0;
+	szp = &sz;
+	hpp = NULL;
+	while (1) {
+	    res = erts_bld_tuple(hpp, szp, 4,
+		erts_bld_uint(hpp, szp, msgq_sum),
+		erts_bld_uint(hpp, szp, msgq_nproc_nonzero),
+		erts_bld_uint(hpp, szp, msgq_max),
+		max_pid);
+	    if (hpp) {
+		BIF_RET(res);
+	    }
+	    hp = HAlloc(BIF_P, sz);
+	    szp = NULL;
+	    hpp = &hp;
+	}
+    }
     BIF_ERROR(BIF_P, BADARG);
 }
 
--- a/erts/emulator/test/statistics_SUITE.erl
+++ b/erts/emulator/test/statistics_SUITE.erl
@@ -32,6 +32,7 @@
 	 scheduler_wall_time/1,
          scheduler_wall_time_all/1,
          msb_scheduler_wall_time/1,
+         statistics_message_queues/1,
 	 reductions/1, reductions_big/1, garbage_collection/1, io/1,
 	 badarg/1, run_queues_lengths_active_tasks/1, msacc/1]).
 
@@ -52,6 +53,7 @@ all() ->
      msb_scheduler_wall_time,
      garbage_collection, io, badarg,
      run_queues_lengths_active_tasks,
+     statistics_message_queues,
      msacc].
 
 groups() -> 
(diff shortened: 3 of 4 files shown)


max-au:
We have been using this feature in production for several years, so I'd love to see it in OTP. However I also understand it may not be useful for the entire community, hence suggesting as a draft to start a discussion.

rickard-green:
Scanning the whole process table is a potentially very expensive operation. Doing this without yielding is a problem by itself, but doing so while also blocking all other normal schedulers in the system makes it really problematic. This is not something that we want to introduce in the system.

rickard-green:
By not requiring i consistent snapshot of the system it would be possible to do something similar without risking to cause major issues. This by not blocking schedulers and by yielding when an appropriate amount of processes has been inspected.

We don't want a feature like this to prevent optimization of signal handling in the future, though. Your current implementation gives the sum of all message queue lengths and the amount of all outstanding messages. Inspecting the outer signal queue (`sig_inq`) might become an issue in the future depending on how it is implemented. `sig_qs.len` is the sum of the amount of messages currently present in the message queue (inner signal queue) and the middle signal queue (used for processing of incoming signals) which also are to be considered as outstanding. This makes it a bit hard to define what this function actually measures.

A solution I would prefer would be to send all processes "process_info message queue length request" signals in parallel, and let them calculate and respond to the process requesting this. This wont cause any problems for future signal handling optimizations and the measured value is easy to document since it is the sum of all message queue lengths. It will however require more work to be done by the system in order to retrieve this information.

max-au:
> Your current implementation gives the sum of all message queue lengths and the amount of all outstanding messages. 

Correct. However thinking more about this and collecting production statistics, I realise that signals contribute close to nothing. It probably makes more sense to exclude signals and count only messages.

Context: we call this function periodically to monitor total amount of messages queued. If this number steadily grows (or, stays above some threshold), we raise an alarm. It has proven to be very valuable. Original implementation was done in pure Erlang: 

```
    {Sum, Max, MaxPid} = lists:foldl(fun (Pid, {S, M, P}) -> 
        case process_info(Pid, message_queue_len) of 
            {message_queue_len, N} -> when N > M -> {S + N, N, Pid};
            {message_queue_len, N} -> {S + N, M, P}; 
            undefined -> {S, M, P}
       end
        end).
```

However with 700k+ processes running in the node it was a major CPU hog, running for several seconds (also IIRC there were some unusual interactions with off-heap message queues). 
This patch made it almost instant, albeit in a non-robust way - and I'm looking forward to either make it work well, or move to some sort of "debug" feature with relaxed guarantees.

> A solution I would prefer would be to send all processes "process_info message queue length request" signals in parallel, and let them calculate and respond to the process requesting this.

I haven't tried this approach, but it looks like the "process requesting this" will need to receive a message from every single process in the system. With 1M processes it might be very expensive.

rickard-green:
> > A solution I would prefer would be to send all processes "process_info message queue length request" signals in parallel, and let them calculate and respond to the process requesting this.
> 
> I haven't tried this approach, but it looks like the "process requesting this" will need to receive a message from every single process in the system. With 1M processes it might be very expensive.

The easiest way to implement this would be to send responses as ordinary messages, but it can be done by letting process respond by updating an internal shared data structure as well.

garazdawi:
Getting information about all processes is quite a common operation, is there a possibility to create something that would work as a sort of map-reduce over all processes in the system so that we can do things such as look for the process that uses the most memory, consumed the most reductions etc etc?

Or would writing the map-reduce code en erlang make it too slow?

max-au:
> is there a possibility to create something that would work as a sort of map-reduce over all processes in the system

That'd be very helpful. We already have similar monitoring. For example, reporting for total amount of GC collected). But it has to run at a very low rate due to amount of processing it requires. Same applies to Erlang distribution, iterating over 40,000 distribution connections (to collect statistics like bytes in/out) is very expensive.

I may not have good enough understanding of ERTS frameworks. What would be the best way to collect VM-global statistics?

Another example is https://github.com/max-au/otp/commit/13bafa86ef6659de42f77ec3951a00d1a63773f2 - single process message queue monitoring which we sometimes use. It introduces reasonably small memory overhead per process, but quite helpful to watch process dynamics.

garazdawi:
> I may not have good enough understanding of ERTS frameworks. What would be the best way to collect VM-global statistics?

I'm not sure it is possible to do in a performant generic manner, but I brought it up as I always tend to prefer a solution that can be used by as many use-cases as possible.

I'm not sure it is the best way, but one way it could be done is:
  * disable all process/port migrations
  * run a misc aux job on each scheduler that creates a list of data for each process
    * after N ports/processes it sends the data to the requesting process as 1 message and yields
  * restart process/port migration

The API would look something like this:

```erlang
erlang:process_info(all, process_info_item()) -> reference().
```

and could be used like this:

```erlang
collect_highest_reductions() ->
   Ref = erlang:process_info(all, [reductions]),
   collect_highest_reductions(Ref, {0, self()}).
collect_highest_reductions(Ref, MaxReds) ->
  receive
    {Ref, process_info, ProcInfos} when is_list(ProcInfos) ->
      lists:foldl(
        fun({Pid, [Reds]}, {MaxReds,_} = Acc) ->
           if Reds > MaxReds
             {Pid, Reds};
           true ->
             Acc
           end
        end, MaxReds, ProcInfos);
    {Ref, process_info, stop}
      MaxReds
  end.
```

The above would be a lot heavier than what you have suggested above, but it should also be a lot cheaper than doing all of the process_info's one by one.

rickard-green:
> I'm not sure it is the best way, but one way it could be done is:
> 
>     * disable all process/port migrations
> 
>     * run a misc aux job on each scheduler that creates a list of data for each process
>       
>       * after N ports/processes it sends the data to the requesting process as 1 message and yields
> 
>     * restart process/port migration

Sending requests to all processes and let them respond internally in a shared data structure will most likely be as efficient as the above, but without potential problems that disabling of migration might cause to other work on the system.

The operation may take a while and during that time you may end up executing other jobs on less schedulers than otherwise would be used. In worst case only one scheduler will be used for other work. In the general case, all signals prior to a process-info-request signal have to be handled before you can gather information from a process. Otherwise, it might be inconsistent. message_queue_len is one such value, but there exist others. If there are many such signals that needs to be handled, this whole operation will take even longer time to complete, further increasing the risk that less schedulers than available will handle other work during the time the process info operation is ongoing. Such signals needs to be handled regardless, so they pose no extra work, but they may prolong the time to complete for operations like this.

garazdawi:
> Sending requests to all processes and let them respond internally in a shared data structure will most likely be as efficient as the above.

I don't agree. I think the overhead of putting all processes into the run-queue and scheduling them will be large for a job like this, especially if we are talking hundreds of thousands of processes. But I don't know, would have to prototype it to know for sure. 

Agreed that disabling migration might now be the best way to solve this, maybe it could be done some other way. I suppose we don't actually need to involve all schedulers at all, just have the calling process loop over the process table as done with `erlang:processes/0`, only that it gets data from each process as it loops.

rickard-green:
> > Sending requests to all processes and let them respond internally in a shared data structure will most likely be as efficient as the above.
> 
> I don't agree. I think the overhead of putting all processes into the run-queue and scheduling them will be large for a job like this, especially if we are talking hundreds of thousands of processes.

I don't think it will be as expensive as it might seem.

> But I don't know, would have to prototype it to know for sure.
> 

Yes.

I however think the first thing we should do is implement an asynchronous `process_info()` relying on pure message passing for the replies. That functionality is useful for other scenarios as well, should outperform the synchronous `process_info()` in this scenario by far, and is a good baseline when comparing optimizations for specific scenarios like this.

max-au:
> I however think the first thing we should do is implement an asynchronous process_info() relying on pure message passing for the replies.

Thinking more of this. We have quite an amount of similar needs for visibility. For example, collecting total amount of bytes/messages that went through all Erlang distribution channels. Currently we call `inet:get_stat` for every single port. While it works reasonably well even for 20-25k connections, it is another CPU hog. I see how a mechanism to signal some (or all) processes to update some shared state would be very helpful.
Unless we can just extend existing system counters, similar to `system_info(process_count).` This would need a generic framework with scheduler-local counters.

I wonder if there is a mixed approach between @garazdawi  suggestion (run misc aux job on all schedulers, disabling port/process migration until job is complete) and @rickard-green (sending a signal to all processes and then receiving messages in response). For example, sending a signal to all/selected processes, with a pointer to scheduler-local variable to update (some sort of generic misc aux job).

> implement an asynchronous process_info() relying on pure message passing for the replies

Will this wake up all hibernated processes? If yes, it may be a problem. In our use cases, out of several hundred thousands, only 30-40k processes are awake. Previously we had an issue with code purge routine that was sending a message to all processes in the system and waking up 1M processes (causing OOM). I had to reimplement check_old_code so it runs slower and puts processes back to sleep.

rickard-green:
> > implement an asynchronous process_info() relying on pure message passing for the replies
> 
> Will this wake up all hibernated processes? If yes, it may be a problem. In our use cases, out of several hundred thousands, only 30-40k processes are awake. Previously we had an issue with code purge routine that was sending a message to all processes in the system and waking up 1M processes (causing OOM). I had to reimplement check_old_code so it runs slower and puts processes back to sleep.

Yes they will be woken in the sense that they will handle signals that have arrived, but will then go back to sleep. This is also how it works today when you inspect a hibernated process via `process_info()`. It should however not affect the memory usage of the hibernated process at all. If so, it is a bug that should be fixed.

We previously had a bug which caused hibernated processes to expand to their non-hibernated heap sizes due to a purge. This was however fixed in OTP 19.3.5. If you should experience such problems with purge, it's a bug that should be fixed.

max-au:
I will abandon this attempt in favour of a different implementation. It is unlikely to salvage anything from this code. I will come back with better proposal based on the discussion here.

@rickard-green  process_info (and signals) work fine with hibernated processes. However "old code check" is an actual Erlang message sent to all processes. And they actually have to wake up, reconstruct stack to see whether old module is anywhere in the stack.

rickard-green:
@max-au 

We might be talking about different "old code checks". Are you talking about "system" messages trigging calls to `code_change()` functions in behaviours? I'm talking about the operations performed by a `code:purge()` operation.

`code:purge()` operations are performed by scheduling emulator internal "system tasks" which performs checking of references to modules and copying of literals that might be removed due to the removal of the module. These "system task" execute in the context of the process itself, but the process returns to its hibernated state once the "system task" has completed. A hibernated process might grow due to removal of literal areas which might cause previous literal terms to be copied onto its heap. The process will still return to a hibernated state once the "system task" has completed, but with these terms on the heap. If this should not be the case, it is a bug that should be fixed. I did some quick testing and it seems to work as expected. If this does not work as expected for you, please provide a way to reproduce it. If so, we'll fix the bug.

PR #2937, opened 2020-12-17
erts: fix dtrace integration and MSACC extras 

max-au:
This commit fixes several issues breaking dtrace/systemtap integration.
 - beam_common.h: missing assignment operator breaks compilation
 - dtrace-wrapper.h: generated erlang_dtrace.h contains #include <sys/sdt.h>,
   which has C++ code requiring C++ linkage. However beam_asm.hpp includes
   global.h with C linkage, and compilation fails, because <sdt.h> uses C++
   templates when __cplusplus is defined. Override linkage to always be
   C++ if C++ compiler is used
 - debug FLAVOUR could not be built with DTRACE, because beam_common.c and
   instr_msg.c do not contain erlang_dtrace.h include, which is needed
   for debug builds where inlining is not enabled
- fixed compilation for microstate accounting configured with extra states

### DIFF
--- a/erts/emulator/beam/beam_common.c
+++ b/erts/emulator/beam/beam_common.c
@@ -35,6 +35,10 @@
 #include "beam_catches.h"
 #include "beam_common.h"
 
+#ifdef USE_VM_PROBES
+#include "dtrace-wrapper.h"
+#endif
+
 static Eterm *get_freason_ptr_from_exc(Eterm exc);
 static ErtsCodePtr next_catch(Process* c_p, Eterm *reg);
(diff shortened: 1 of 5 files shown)


max-au:
Also added a fix for microstate accounting compilation failure. It may be worth nothing however that `statistics_SUITE` is failing when OTP is compiled with `--with-microstate-accounting=extra`

garazdawi:
Great, thanks. Putting into nightly tests.

garazdawi:
Worked flawlessly in the nightly tests. Thanks for the fixes!

PR #2938, opened 2020-12-18
Link from timer module doc to the Efficiency Guide

erszcz:
The Efficiency Guide explains the difference between `timer` and `erlang` timer functions and links to both the modules. It seems likely, though, that somebody looking for timer management will first stumble upon `timer`, which doesn't mention any alternatives. This change links from `timer` to the Efficiency Guide.


garazdawi:
Would you mind setting this to be merged into maint instead of master?

erszcz:
Done!

PR #2940, opened 2020-12-18
Use maps:from_keys/2 and heuristics in new sets

josevalim:
maps:from_keys/2 speeds up sets creation by avoiding
a list traversal and reducing memory allocation.

Both intersection/2 and subtract/2 operations were
also augmented with a new heuristic that chooses
between creating a set from scratch or deleting some
of the existing keys, in order to reduce memory
allocation and speed up operations.

As of this patch, I would say the guidelines for
choosing between ordsets, gb_sets and sets (v2) is
the following:

  1. If you only want to traverse, perform unions,
    intersections, or subtractions of values that
    are quick to compare (integers, atoms, etc)
    and they are up to 1k-10k in size, pick ordsets;

  2. If you need take_smallest and take_largest and
    you don't want to do external bookkeeping, use
    gb_sets;

  3. For everything else, choose sets (v2).

Benchmarks for different operations and implementations
can be found at https://github.com/josevalim/sets_bench.

### DIFF
--- a/lib/stdlib/src/sets.erl
+++ b/lib/stdlib/src/sets.erl
@@ -115,7 +115,7 @@ from_list(Ls) ->
       List :: [Element],
       Set :: set(Element).
 from_list(Ls, [{version, 2}]) ->
-    maps:from_list([{K,?VALUE}||K<-Ls]);
+    maps:from_keys(Ls, ?VALUE);
 from_list(Ls, Opts) ->
     case proplists:get_value(version, Opts, 1) of
         1 -> from_list(Ls);
@@ -258,7 +258,14 @@ union1(S1, []) -> S1.
       Set2 :: set(Element),
       Set3 :: set(Element).
 intersection(#{}=S1, #{}=S2) ->
-    maps:intersect(S1, S2);
+    case map_size(S1) < map_size(S2) of
+        true ->
+            Next = maps:next(maps:iterator(S1)),
+            intersection_heuristic(Next, [], [], floor(map_size(S1) * 0.75), S1, S2);
+        false ->
+            Next = maps:next(maps:iterator(S2)),
+            intersection_heuristic(Next, [], [], floor(map_size(S2) * 0.75), S2, S1)
+    end;
 intersection(S1, S2) ->
     case size(S1) < size(S2) of
         true ->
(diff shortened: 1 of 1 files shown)


josevalim:
Here is a status report on EEP 50:

Unfortunately I don't think we can truly replace ordsets by sets v2. Even if we implement intersection and subtract in C, we really can't beat ordsets with immediate terms. Today ordsets beat the other implementations by 2 orders of magnitude.

When it comes to gb_sets, they have take_smallest and take_largest, which can't be implemented in the others without additional (and perhaps expensive) bookkeeping. sets (v2) beats gb_sets in most operations and generally speaking they exhibit similar profiles. They are always within 1-4x of the other in benchmarks.

One operation that could further speed both `subtract` and `intersection` for sets v2 is a `maps:remove_many/2` (i.e. `maps:without/2`). But it doesn't seem to be trivial to implement and, given the points above, I am not sure if worth it.

That said, I would say the only remaining step to conclude EEP 50 is to remove cerl_sets in favor of sets (v2), which I plan to do in a separate PR after this one is merged.

bjorng:
> When it comes to gb_sets, they have take_smallest and take_largest, which can't be implemented in the others without additional (and perhaps expensive) bookkeeping.

For some uses of `take_smallest` (or `take_largest`) a function that would take **any** element of the map would be useful. One example is a workset algorithm:

https://github.com/erlang/otp/blob/c8ef47c44282d1261366c8272038152afd17c65f/lib/compiler/src/beam_ssa_bool.erl#L757-L768

Perhaps it could be useful to add a `maps:take_any/1` function to allow `maps` to be used for workset algorithms.

josevalim:
@bjorng could we use `maps:next(maps:iterator(Map))` and then `maps:take/2` to achieve this behavior?

bjorng:
> @bjorng could we use maps:next(maps:iterator(Map)) and then maps:take/2 to achieve this behavior?

Yes, that would work.

garazdawi:
> > @bjorng could we use maps:next(maps:iterator(Map)) and then maps:take/2 to achieve this behavior?
>
> Yes, that would work.

One of the things with maps:iterator is that it is optimized for taking all the values in the map using the iterator, not just one. So if you keep calling maps:iterator over and over on the same map you will end up building an expensive iterator instance each time.

I don't have any better suggestions though :)

besides possibly changing maps:iterator to only prep for 1 pair and then when maps:next is called to get the second element we build the expensive iterator context.

josevalim:
I know people have asked for `maps:first/1`, that could fit here. The name `first` can be confusing for large maps though (but there is precedent in `ets`).

bjorng:
Thanks for your pull request.

PR #2942, opened 2020-12-19
Fix typos

lagebr:


garazdawi:
Hello! We have chosen to not highlight those sections of the readme because the github highlighter does not do a very good job with them. I think we would like it to remain that way, unless you have another reason for adding them?

We would very much like your original typo fix.

garazdawi:
Thanks for your contribution!

PR #2944, opened 2020-12-20
Detect repeated occurrences of the stacktrace variable in a try clause

richcarl:
Previously, in `try ... catch _:T:T -> ...`, the compiler would treat the two
occurrences of T as separate variables, not adding any check that they have the
same value. Since such a check can be expensive for stacktraces, we don't want
to allow this, just like we don't allow T to be already bound.

This provides a more robust mechanism for handling the stacktrace variable, only
affecting catch-clauses, and sets proper position information on the trace var.


michalmuskala (approved):

garazdawi:
See https://bugs.erlang.org/browse/ERL-1380 for the bug report

PR #1217, opened 2016-10-26
Document the -epmd_module switch

josevalim:

PR #2945, opened 2020-12-20
Document the -no_epmd option

JeromeDeBretagne:
Hello,

The option is already mentioned in https://erlang.org/doc/apps/erts/alt_dist.html but was not documented in https://erlang.org/doc/man/erl.html until now. I'm proposing this pull request as a follow-up to a recent question on the erlang-questions mailing list.

Cheers,
Jérôme

### DIFF
--- a/erts/doc/src/erl_cmd.xml
+++ b/erts/doc/src/erl_cmd.xml
@@ -460,7 +460,7 @@ $ <input>erl \
           become distributed; see <seeerl marker="kernel:net_kernel">
           <c>net_kernel(3)</c></seeerl>. It is also ensured that
           <c><![CDATA[epmd]]></c> runs on the current host before Erlang is
-          started; see <seecom marker="epmd"><c>epmd(1)</c></seecom>.and the
+          started; see <seecom marker="epmd"><c>epmd(1)</c></seecom> and the
           <seecom marker="#start_epmd"><c>-start_epmd</c></seecom> option.</p>
         <p>The node name will be <c><![CDATA[Name@Host]]></c>, where
           <c><![CDATA[Host]]></c> is the fully qualified host name of the
@@ -501,6 +501,22 @@ $ <input>erl \
             network is configured to keep potential attackers out.
           </p>
         </warning>
+      </item>
+      <tag><c><![CDATA[-no_epmd]]></c></tag>
+      <item>
+        <p>Specifies that the distributed node does not need
+          <seecom marker="epmd">epmd</seecom> at all.</p>
+        <p>This option ensures that the Erlang runtime system does not
+          start <seecom marker="epmd">epmd</seecom> and does not start
+          the <seeerl marker="kernel:erl_epmd">erl_epmd</seeerl> process
+          for distribution either.</p>
+        <p>This option only works if Erlang is started as a distributed node
+          with the <seecom marker="#proto_dist">-proto_dist</seecom> option
+          using an alternative protocol for Erlang distribution which
+          does not rely on epmd for node registration and discovery.
+          For more information, see <seeguide marker="erts:alt_dist">How
+          to implement an Alternative Carrier for the Erlang
+          Distribution</seeguide>.</p>
       </item>
       <tag><c><![CDATA[-noinput]]></c></tag>
       <item>
--- a/erts/etc/common/erlexec.c
+++ b/erts/etc/common/erlexec.c
@@ -1,7 +1,7 @@
 /*
  * %CopyrightBegin%
  *
- * Copyright Ericsson AB 1996-2020. All Rights Reserved.
+ * Copyright Ericsson AB 1996-2021. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -185,7 +185,7 @@ static char *plusz_val_switches[] = {
 
 void usage(const char *switchname);
 static void usage_format(char *format, ...);
-void start_epmd(char *epmd);
+void start_epmd_daemon(char *epmd);
 void error(char* format, ...);
 
 /*
@@ -422,6 +422,8 @@ int main(int argc, char **argv)
 				 * of the arguments. */
     int isdistributed = 0;
     int no_epmd = 0;
+    int proto_dist = 0;
+    int start_epmd = 1;         /* Whether epmd should be started */
     int i;
     char* s;
     char *epmd_prog = NULL;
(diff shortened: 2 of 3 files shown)


garazdawi:
This definitely should be documented. I wonder if there are any tests that test that it works... hmm.

josevalim:
There was a previous discussion about this here: #1217

garazdawi:
After doing this a while you end up arguing with your past self...

The original reason was that this option was not public and documented anywhere and it should not be. It has since then slipped into the alternative distribution documentation, which makes it public already which is why I now have changed my opinion.

josevalim:
@garazdawi could the guide be updated to use `-start_epmd`? Then you could keep `-no_epmd` as is and say it was a documentation bug (it is curious we reversed roles, haha).

garazdawi:
The problem is that they do not do the same things and I do not think we can change `-start_epmd false` to imply that `erl_epmd` is not started... so I think this will have to work the way that it does and I'll try to address it when I start the work on converting all of these command-line switches into configuration parameters which is where they really belong.

tsloughter:
@JeromeDeBretagne does `erl -sname foo -start_epmd false -erl_epmd_port 14693 -no_epmd` actually work for you?

tsloughter:
From what I see in the code you can't use `-no_epmd` unless you are setting another distribution module?

So if you want to use the new stuff like relx does you actually don't want to set `-no_epmd` and it will fail to start the node if you do set that.

I think an option to not register with epmd would be useful but in the case like in the mailing list thread and my blog post you actually want it to use the epmd module in the node, it just needs to not register even if epmd is running, so would need a new command line option.

JeromeDeBretagne:
Hi Tristan,

> @JeromeDeBretagne does `erl -sname foo -start_epmd false -erl_epmd_port 14693 -no_epmd` actually work for you?

No, it doesn't work but that's expected. Since you don't specify a -proto_dist option, you end up using the default distribution protocol (inet_tcp) which relies on epmd.

You can try `erl -sname foo -no_epmd -proto_dist erl_uds -pa path_to_erl_uds_dist.beam`, this will switch over to the `erl_uds` distribution protocol which uses Unix Domain Sockets instead and doesn't need epmd at all to work.

I hope this answers your questions.

Cheers, Jérôme

tsloughter:
@JeromeDeBretagne ok, that is what I thought. You should clarify on the mailing list since the thread makes it seem like that option works with the command you were originally running in the first post :)

max-au:
I believe `-no_epmd` switch has somewhat different semantic. It controls whether `kernel` supervision calls `start_link/0` of a supplied `-epmd_module`.

To elaborate. If I specify `-no_epmd`, kernel will not start a child process (`erl_epmd`) as a part of kernel supervision tree. So when I implement my `-epmd_module` it does not need `start_link/0` callback. However it also means that I can't have some state/cache, which I indeed use in some service discovery implementation.

JeromeDeBretagne:
> You should clarify on the mailing list since the thread makes it seem like that option works with the command you were originally running in the first post :)

I didn't intend to make a reference to the original command, I was only trying to answer the following question:

> Is there any option(s) of erl command to avoid registration to epmd?

If you think the thread is slightly confusing, feel free to jump in!

Cheers, Jérôme

shino:
(I'm who asked questions on the ML Thread)

> I didn't intend to make a reference to the original command, I was only trying to answer the following question:

I understood. -no_epmd with -name/-sname does *NOT* work as well for me.

> I think an option to not register with epmd would be useful but in the case like in the mailing list thread and my blog post you actually want it to use the epmd module in the node, it just needs to not register even if epmd is running, so would need a new command line option.

Agree.

erts/doc/src/erl_cmd.xml
```diff
@@ -422,6 +422,22 @@
...
       </item>
+      <tag><c><![CDATA[-no_epmd]]></c></tag>
+      <item>
+        <p>Specifies that the distributed node does not need
+          <seecom marker="epmd">epmd</seecom> at all.</p>
+        <p>This option ensures that the Erlang runtime system does not
+          start <seecom marker="epmd">epmd</seecom> and does not start
+          the <seeerl marker="kernel:erl_epmd">erl_epmd</seeerl> process
+          for distribution either.</p>
+        <p>This only applies if Erlang is started as a distributed node
```
garazdawi, line 433:
```suggestion
        <p>This option only works if Erlang is started as a distributed node
```

garazdawi (commented):
Should we add a check that an alternative `-proto_dist` is given when this is used?

Also, there should be tests for this option now that it is official. Can you add some, if not I will do it, but I don't know when I will have time for that.

JeromeDeBretagne:
Hi Lukas,

My focus with this PR was only on the documentation aspect, simply to complete it (as the option was already officially documented in another page). If you suggest to include larger modifications such as tests, feel free (or other contibutors) to expand this change with additional commits or in a follow-up PR, as I don't have a lot of bandwidth for larger contributions at the moment. I've included your proposed wording change in the updated version, as that was easy to fix.

erts/doc/src/erl_cmd.xml
JeromeDeBretagne: Change included in the updated version, thanks for your review.

erts/etc/common/erlexec.c
```diff
@@ -763,6 +764,19 @@ int main(int argc, char **argv)
 		    }
 		    break;
 
+		  case 'p':
+		    if (strcmp(argv[i], "-proto_dist") == 0) {
+			if (i+1 >= argc)
```
JeromeDeBretagne, line 769: For this check, I've reused the same code as found for other similar checks in the rest of the file, however they are not working as intended currently, cf. issue https://github.com/erlang/otp/issues/4624

JeromeDeBretagne:
Hi Lukas,

I've just added a check that an alternative -proto_dist option is also given, as you had suggsted. FYI, that's when I've detected the issue raised in https://github.com/erlang/otp/issues/4624 .

What else would you suggest, to land this change for OTP 24? I've looked at erlexec_SUITE.erl but it includes very few tests checking for specific options so I've no clue where/how to start to add a useful test. Thanks.

JeromeDeBretagne:
> I would add the test to erts/emulator/test/distribution_SUITE as that is where some other epmd configuration options are started. 

Thanks Lukas for the pointer. Upon a first pass, I've detected that I had broken the start_epmd_false test, so I've updated my second commit to fix this issue to start with. Now, I'll look into adding a new test in the coming days.

JeromeDeBretagne:
Lukas, I've now added a test in distribution_SUITE checking that using -no_epmd without passing the -proto_dist option should fail. Thanks

erts/etc/common/erlexec.c
JeromeDeBretagne: I've switched the previous check to use the new NEXT_ARG_CHECK(); introduced in 022cea1

JeromeDeBretagne:
Hi Lukas,

I've updated this PR to use the new call to NEXT_ARG_CHECK you merged earlier today in 022cea1. That's the only change I've made. As a consequence, I've updated the base branch to target master now instead of maint previously. Thanks

garazdawi:
Great! I will put this in testing and merge soon.

garazdawi:
thanks!

PR #2948, opened 2020-12-22
Fix uses of AC_CONFIG_AUX_DIRS() in configure scripts

rickard-green:
Proposed fix for [ERL-1447](https://bugs.erlang.org/browse/ERL-1447).

Now builds with autoconf version 2.70 although we get lots of warnings for use of obsolete macros. Those will however have to be fixed at a later time.


lrascao:
FYI i've bisected this down to autoconf v2.69d, commit 4c59bf2 "Improve handling of missing aux scripts"

PR #2949, opened 2020-12-23
shell history: allow configurable history module

max-au:
In addition to 'enabled' or 'disabled' setting, kernel
shell_history variable may be an atom that allows to use
any module as callback module for shell history.
This enabled to fetch shell history from different sources,
for example, cloud-based environment, or other nodes of
a cluster.
This commit also fixes test case that was using -kernel
shell_history 'true' instead of 'enabled'.

### DIFF
--- a/lib/kernel/doc/src/kernel_app.xml
+++ b/lib/kernel/doc/src/kernel_app.xml
@@ -446,10 +446,16 @@ MaxT = TickTime + TickTime / 4</code>
 	  using this service.</p>
         <p>Defaults to <c>false</c>.</p>
       </item>
-      <tag><c>shell_history = enabled | disabled </c></tag>
+      <tag><c>shell_history = enabled | disabled | module()</c></tag>
       <item>
         <p>Specifies whether shell history should be logged to disk
-           between usages of <c>erl</c>.</p>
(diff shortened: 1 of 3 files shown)


garazdawi:
> This commit also fixes test case that was using `-kernel shell_history 'true'` instead of `'enabled'`.

How odd, any idea on why the testcase passed at all?

max-au:
>  any idea on why the testcase passed at all?

The testcase verifies that there are no crash reports in the log file. Anything that is not 'enabled' was treated as 'disabled', so 'true' was also treated as 'disabled'. 

When the shell history is disabled, there is no crash, so test passed.

If https://github.com/erlang/otp/commit/4d7a2e58c950983294487563877bdbac40224d9c#diff-0be433b909f9d2267e9fa9a435a68f207fddafcdffb4a01de8e091ab4fac51f8 is reverted, and 'enabled' is passed, then the crash would occur.

wojtekmach:
This is a really nice change, thank you! Sometimes I'd switch back'n'forth between iex & erl and when looking up history in erl I'd see elixir invocations and get slightly confused (as there's just one history), now it's going to be easy to have a separate history per shell type.

PR #2952, opened 2020-12-28
add maps:merge_deep/2

z8674558:
deep-merge functions was proposed in https://github.com/erlang/otp/pull/2797, but not yet implemented.


lib/stdlib/doc/src/maps.xml
```diff
@@ -364,6 +364,26 @@ none</code>
       </desc>
     </func>
 
+    <func>
+      <name name="merge_deep" arity="2" since=""/>
+      <fsummary></fsummary>
+      <desc>
+        <p>Merges two maps, and if two keys exist in both maps
```
paulo-ferraz-oliveira, line 371: By "two keys exist" do you mean "the same key exists"?

lib/stdlib/doc/src/maps.xml
```diff
@@ -364,6 +364,26 @@ none</code>
...
+          and both values are maps, merges them recursively.</p>
```
paulo-ferraz-oliveira, line 372: both values > both their values
?

lib/stdlib/doc/src/maps.xml
```diff
@@ -364,6 +364,26 @@ none</code>
...
+        <p>If either of values at the same key is not map,
```
paulo-ferraz-oliveira, line 373: I would probably replace "If either of values at the same key is not map" with "Otherwise".

lib/stdlib/doc/src/maps.xml
z8674558: thank you, sounds better.

z8674558: fixed

lib/stdlib/doc/src/maps.xml
z8674558: fixed

lib/stdlib/doc/src/maps.xml
z8674558: ty fixed

michalmuskala:
I'd question how useful this is. I think in practice, a function exactly like this won't be particularly useful - frequently you'd want to do more special handling for different values - e.g. concatenating lists, and similar. Given a user-defined implementation (without the strict & specific error handling of the `maps` module) now that we have `merge_with` is very simple, I'm not sure it makes sense to include it.

Please see a discussion on adding exactly the same function to Elixir's `Map` module: https://github.com/elixir-lang/elixir/pull/5339

```erlang
deep_merge(Map1, Map2) ->
  Combine = fun
    (_Key, Value1, Value2) when is_map(Value1), is_map(Value2) -> deep_merge(Value1, Value2);
    (_Key, _Value1, Value2) -> Value2
  end,
  merge_with(Combine, Map1, Map2).
```

z8674558:
@michalmuskala I personally don't think this is a niche feature, because there should not be so many cases that needs special handling, other than what you refer to as list concatenation. Unlike Elixir, we don't have `struct`, thus there is no need to figure out how to merge across types. So if we can select whether two lists with the same key are concatenated or not, it can be useful for everyone.

vans163:
> I think in practice, a function exactly like this won't be particularly useful - frequently you'd want to do more special handling for different values - e.g. concatenating lists, and similar.

Its exactly the same as :maps.merge but unravels nests. In practise that is what you want, often when working with nested merge-able data there are no lists involved, everything is a map (a list of users, becomes a map of users by their uuid). If you are thinking how to merge lists you are giving yourself an unneeded headache for an unusual use-case.

> Please see a discussion on adding exactly the same function to Elixir's `Map` module: [elixir-lang/elixir#5339](https://github.com/elixir-lang/elixir/pull/5339)

No structs to deal with.  Also put_in is still broken on Elixir (does not work with structs) and does not give any stacktrace, so I think the non-struct Map related usage of the lang is lagging. 

Also if I had a dollar for everytime I seen..
```elixir
%{ user  | address:
    %{ address | street:
          %{street | unit:
              %{ unit | no: 5}
              }
          }
     }
}
#VS
:maps.merge_deep(user, %{address: %{street: %{unit: %{no: 5}}}})
```

garazdawi:
I like this feature and have written the equivalent to this function is a number of places already. Though I would like us to think a bit in a broader context before merging this.

The question that comes to my mind is if merge is the only function that should have a deep variant? i.e. do we want maps:deep_put/deep_get? and if so, should this be in a completely different module?

RaimoNiskanen:
@garazdawi: That also touches on the ideas about lenses that Björn-Egil mentioned way back when maps were introduced.

z8674558:
@garazdawi I would say deep_get/deep_put are nice-to-have. Let me explain it with an example; some config files have nested-key syntax, which looks like`foo.bar = x`. When we want to parse these inputs and transform it to map in Erlang, one can write `deep_put` as below and utilize it:

```
deep_put([Key], Val, Map) ->
    merge(Key, Val, Map);
deep_put([Key|More], Val, Map) ->
    merge(Key, deep_put(Paths, Val, #{}), Map).
```

garazdawi:
> That also touches on the ideas about lenses that Björn-Egil mentioned way back when maps were introduced.

Yes, this comes close to that, though it would not be as general. Having something like lenses would be great, though I don't think we ever got to a place where we had a proposal that we were happy with.

> some config files have nested-key syntax

It is also mostly when working with config that I have used the deep maps functions.

lib/stdlib/src/maps.erl
```diff
@@ -26,7 +26,7 @@
          without/2, with/2,
          iterator/1, next/1,
          intersect/2, intersect_with/3,
-         merge_with/3]).
+         merge_with/3, merge_deep/2]).
```
meox, line 29: preserve the alphabetical order

lib/stdlib/src/maps.erl
```diff
@@ -194,6 +194,25 @@ merge_with_1({K, V2, Iterator}, Map1, Map2, Combiner) ->
 merge_with_1(none, Result, _, _) ->
     Result.
 
+-spec merge_deep(Map1,Map2) -> Map3 when
```
meox, line 197:
```suggestion
-spec merge_deep(Map1, Map2) -> Map3 when
```

lib/stdlib/src/maps.erl
```diff
@@ -194,6 +194,25 @@ merge_with_1({K, V2, Iterator}, Map1, Map2, Combiner) ->
...
+
+merge_deep(Map1, Map2) when is_map(Map1), is_map(Map2) ->
+    Combiner = fun (_Key1, Value1, Value2) ->
+        case is_map(Value1) andalso is_map(Value2) of
+            true ->
+                merge_deep(Value1, Value2);
+            false ->
+                Value2
+        end
+    end,
```
meox, line 210:
```suggestion
    Combiner = fun
        (_Key1, Value1, Value2) when is_map(Value1) andalso is_map(Value2) ->
                merge_deep(Value1, Value2);
         (_Key1, _Value1, Value2) ->
                Value2
        end
    end,
```

KennethL: Don't use `andalso` in guards unless really necessary (nested expressions with parentheses). So in this case
```erlang
  Combiner = fun
        (_Key1, Value1, Value2) when is_map(Value1),  is_map(Value2) ->
```

PR #2954, opened 2020-12-29
Also generate SHA256 checksums for releases

tianon:
See https://marc-stevens.nl/research/md5-1block-collision/ and https://security.googleblog.com/2017/02/announcing-first-sha1-collision.html -- both MD5 and SHA1 are considered "broken" and unsafe for serious use, so this updates the release workflow to also generate SHA256 checksums of the built artifacts.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -144,7 +144,7 @@ jobs:
           name: otp_doc_man
 
       ## We add the correct version name into the file names
-      ## and create the MD5 file for all assets
+      ## and create the hash files for all assets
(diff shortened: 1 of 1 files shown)


garazdawi:
Thanks!

tianon:
Any chance of backporting this to the release branches so we don't have to wait for the next major release to reap the benefit? :innocent: :heart: :pray:

(Is there something I can do to help?  I imagine it's probably a pretty easy cherry-pick.)

garazdawi:
> Any chance of backporting this to the release branches so we don't have to wait for the next major release to reap the benefit?

Done

PR #2956, opened 2021-01-02
gen: init_it given atom as parent PID

orzen:
The atom 'self' is passed instead of the parent's PID, using `self()`
when starting some gen-processes with 'monitor' or 'nolink'.

This patch change so that `self()` is called and the parent argument to
proc_lib will result in a PID instead of the atom 'self'.

I don't know the effect of this change. I assume this is a typo since a
corresponding clause, without `Name`, is using `self()` instead of `'self'`.

### DIFF
--- a/lib/stdlib/src/gen.erl
+++ b/lib/stdlib/src/gen.erl
@@ -113,7 +113,7 @@ do_spawn(GenMod, monitor, Mod, Args, Options) ->
 do_spawn(GenMod, _, Mod, Args, Options) ->
     Time = timeout(Options),
     proc_lib:start(?MODULE, init_it,
-		   [GenMod, self(), self, Mod, Args, Options], 
+		   [GenMod, self(), 'self', Mod, Args, Options],
 		   Time,
 		   spawn_opts(Options)).
 
@@ -133,7 +133,7 @@ do_spawn(GenMod, monitor, Name, Mod, Args, Options) ->
 do_spawn(GenMod, _, Name, Mod, Args, Options) ->
     Time = timeout(Options),
     proc_lib:start(?MODULE, init_it,
-		   [GenMod, self(), self, Name, Mod, Args, Options], 
+		   [GenMod, self(), 'self', Name, Mod, Args, Options],
 		   Time,
 		   spawn_opts(Options)).
 


essen (approved):

lib/stdlib/src/gen.erl
```diff
@@ -113,7 +113,7 @@ do_spawn(GenMod, monitor, Mod, Args, Options) ->
 do_spawn(GenMod, _, Mod, Args, Options) ->
     Time = timeout(Options),
     proc_lib:start(?MODULE, init_it,
-		   [GenMod, self(), self, Mod, Args, Options], 
+		   [GenMod, self(), self(), Mod, Args, Options],
```
mikpe, line 116: I believe this change and the one below are incorrect. Using the atom `self` here seems to trigger explicit logic in the gen_* modules' `init_it/6` callbacks, see e.g. the first clause of `gen_server:init_it/6`. The effect is that `Parent` in the spawned server isn't itself, which it apparently wants in these cases.

essen: Nice catch! It seems that when the start is not linked, the parent is set to the started gen process. It makes sense because if we aren't linked to the process that started the gen process then this isn't a parent in the supervision tree sense. It's super confusing that it uses the atom `self` though, maybe something like `no_parent` would have been more clear. A comment could be a nice addition.

orzen: Nice catch indeed! I got confused and did not see how `proc_lib:sync_start/2` would receive the ack to demonitor properly. Now I can see that it's using `Starter` instead of `Parent` to send the ack. Sorry for the inconvenience.

mikpe: I think changing the atom to the more explicit `'self'` form in all relevant places would be a nice cleanup and a reminder to future readers of this code that it's not a typo.

orzen: Sure, I can change it for the gen.erl.

mikpe (approved):

max-au:
If I am reading the patch right, it seems to be no-op.

mikpe:
> If I am reading the patch right, it seems to be no-op.

It's a documentation change to emphasize that the atom `self` used here is deliberate and not a typo for `self()`.

PR #2959, opened 2021-01-05
inet_res: fix dns compression encoder for large responses

jimdigriz:
The [pointer in the label is 14bit wide](https://tools.ietf.org/html/rfc1035#section-4.1.4) and a typo looks to have used the flag value (3) for the left shift rather than just 1; leading to a limit of 49152 being used rather than 16384.

This bug only shows up for large responses involving several hundred records such as those seen in an AXFR transfer.

This PR resolves this more than 11 year old bug.

### DIFF
--- a/lib/kernel/src/inet_dns.erl
+++ b/lib/kernel/src/inet_dns.erl
@@ -697,7 +697,7 @@ encode_labels(Bin, Comp0, Pos, [L|Ls]=Labels)
   when 1 =< byte_size(L), byte_size(L) =< 63 ->
     case gb_trees:lookup(Labels, Comp0) of
 	none ->
-	    Comp = if Pos < (3 bsl 14) ->
+	    Comp = if Pos < (1 bsl 14) ->
 			   %% Just in case - compression
 			   %% pointers cannot reach further
 			   gb_trees:insert(Labels, Pos, Comp0);
--- a/lib/kernel/test/inet_res_SUITE.erl
+++ b/lib/kernel/test/inet_res_SUITE.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 2009-2020. All Rights Reserved.
+%% Copyright Ericsson AB 2009-2021. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -33,7 +33,8 @@
         ]).
 -export([basic/1, resolve/1, edns0/1, txt_record/1, files_monitor/1,
 	 last_ms_answer/1, intermediate_error/1,
-         servfail_retry_timeout_default/1, servfail_retry_timeout_1000/1
+         servfail_retry_timeout_default/1, servfail_retry_timeout_1000/1,
+         label_compression_limit/1
         ]).
 -export([
 	 gethostbyaddr/0, gethostbyaddr/1,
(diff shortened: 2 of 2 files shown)


RaimoNiskanen:
No comments on the patch; it is obviously correct.

I am working on a test case, though.  'bout time that code line gets test coverage.

jimdigriz:
On a related note, I have noticed the DNS decoder currently considers multiple EDNS(0) options as valid when it [should be rejected with a `FORMERR`](https://tools.ietf.org/html/rfc6891#section-6.1.1).

My thoughts is that this should be checked for by OTP in `inet_dns.erl` with something like the following (untested) patch:
```
diff --git a/lib/kernel/src/inet_dns.erl b/lib/kernel/src/inet_dns.erl
index 18103771dc..128a6e2464 100644
--- a/lib/kernel/src/inet_dns.erl
+++ b/lib/kernel/src/inet_dns.erl
@@ -203,6 +203,7 @@ decode_rr_section(Bin, N, Buffer, RRs) ->
 	    RR =
 		case Type of
 		    opt ->
+			case lists:keymember(dns_rr_opt, 1, RRs) of true -> throw(?DECODE_ERROR); false -> ok end,
 			<<ExtRcode,Version,Z:16>> = TTL,
 			#dns_rr_opt{domain=Name,
 				    type=Type,
```

What do you think?

On another note I will soon need to add [TSIG support](https://tools.ietf.org/html/rfc8945) to my application, would the OTP team be interested in something to go into OTP instead or do you consider TSIG something that should remain external?

RaimoNiskanen:
I hope that approach is good enough.  If we would decode a bad message consisting of many `dns_rr_opt` RRs we have a quadratic performance problem.  I do not know how bad that is.
``` erlang
case lists:keymember(dns_rr_opt, 1, RRs) of true -> throw(?DECODE_ERROR); false -> ok end,
```
can be written shorter, but not fancied by all:
``` erlang
lists:keymember(dns_rr_opt, 1, RRs) andalso throw(?DECODE_ERROR),
```

Regarding TSIG support; I predict that we would like to have it implemented in `inet_dns`.

jimdigriz:
The list has a maximum length of 65535 so can never get too large:
```
1> A = lists:map(fun(X) -> {bob, whatever, cheese} end, lists:seq(1, 1 bsl 16)) ++ [{findme,chips}].
[{bob,whatever,cheese},
 {bob,whatever,...},
 {bob,...},
 {...}|...]
2> timer:tc(fun() -> lists:keymember(findme, 1, A) end).
{394,true}
```

Walking lists is faster than people think, folks are always too eager to reach for maps. :)

RaimoNiskanen:
> Walking lists is faster than people think, folks are always too eager to reach for maps. :)

I agree in general.  The fact that the list has a length limitation is often enough.  But, quadratic complexity is often underestimated.

Fortunately I can not see how this could become quadratic, even when constructing an evil message, since the algorithm only traverses when finding an `opt` RR and bails out for the second found, so there will be max 2 traversals summing up to the full list length.

If really a problem, or to be sure that it is not a problem, it can be fixed by holding a special flag in the loop `decode_rr_section/4` or rather then `decode_rr_section(Bin, N, Buffer, OptFlag, RRs)` that just remembers if we have seen a `dns_rr_opt` RR or not.  Ugly special treatment, maybe, but there is already special treatment of `Type =:= opt`, so perhaps not that ugly.

For TSIG it might be possible to (instead of adding another flag variable) expand the special flag to a flag field (map?) to verify that, what was it; (I have only browsed some RFC) all TSIG records are last.  I read that as; no other records come after TSIG records.

jimdigriz:
Maybe not for this time, but when get to my TSIG work did you want this flag checking approach used or keep to `lists:keymember/3`?

I guess the majority of the users of `inet_dns.erl` are DNS clients and as the additional section is typically small it is best to stick with list scanning for reasons of simplicity and ease of understanding. Impacting >99% of the cases for the <1%, especially as it looks like even in the worse bad-actor case the impact is minimal.

RaimoNiskanen:
As I said, I see no problem with list scanning in this case when checking duplicate opt RR:s, since I/we can not find any bad worse case.  The worst case I can find is that instead of looping through the list once, the code would loop through it twice; the processing loop + scanning once (possibly two halves) to check for duplicate opt RR.

I want the code to avoid quadratic complexity, which is easy to get, accidentally, when using list scanning.

If I got it right that the TSIG code has to check that TSIG RR:s are last, would it not be sufficient then to; for every other RR check that the previous one is not a TSIG RR.  That would be an O(1) check in the loop for every non-TSIG RR, which should be no problem.

RaimoNiskanen:

PR #2960, opened 2021-01-05
Use Ryu for io_lib_format:fwrite_g/1

DianaOlympos:
io_lib_format:fwrite_g/1 now use the Ryu algorithm see
(https://dl.acm.org/doi/pdf/10.1145/3192366.3192369). We get an
approximately 3x speedup over the whole doubles, with a more stable
performance profile. The Dragon4 implementation used before had a
performance profile linear with the size of the double to print.

We also see a 10% reduction in memory use with Ryu over the old
algorithm.

It is worth noting that this implementation is slightly slower (10% to
30%) for "small" doubles in some cases. This commit does not try to
optimise this case yet.

The Ryu algorithm depends on a lookup table that is generated upfront.
This commit add a script to generate it (mostly in case a bug is found
in it or we decide to change the size of the lookup table) and the
output of said script in io_lib_format_ryu_table.erl. This is done
because the table do not change from a build to another and should not
be regenerated outside of a bug or a fundamental change in the
algorithm.

The Ryu algorithm has a know worst case scenario for doubles that can be
represented as "small" integer. This commit introduce a optimised
handling for this case.

## Additional comments
The reference implementation of Ryu is at https://github.com/ulfjack/ryu

This is probably optimisable more, in particular there are some redundant divisions in the loops and the core 64bits * 128 bits multiplication could be broken up into multiple 55bits * 59bits one, to avoid using big nums. But this was deemed good enough for a first PR.

In a perfect world, this would be in C, probably as a new format as part of float_to_list/2 and binary_to_list/2. If there is interest, i will offer that in the ML later.
Another reason to want to provide it as a BIF is that this is technically a private API but is used in most JSON encoding library (Poison and Jason at least). It has been really stable over the years, so i think there is a case for it.

I have encountered a lot of problems optimising the case in which it is slower than the current algorithm. Under profiler, Ryu end up being faster due to calling less functions, which makes it particularly hard to understand the behaviour. I am probably going to keep looking for optimisation, but i think it would make more sense to just move it into the BIF. It would also allows to never use big num and use the code coming from the reference implementation directly and to get a better coverage by reusing the Microsoft STL test suite as an inspiration.

Here is a synthetic microbenchmark over 10k random floats.

```
Name           ips        average  deviation         median         99th %
Ryu          73.83       13.54 ms    ±16.18%       13.02 ms       24.83 ms
OTP          24.48       40.85 ms    ±15.28%       38.62 ms       66.18 ms

Comparison: 
Ryu          73.83
OTP          24.48 - 3.02x slower +27.30 ms

Memory usage statistics:

Name    Memory usage
Ryu          7.48 MB
OTP          8.45 MB - 1.13x memory usage +0.97 MB
```

This work is partially funded by the EEF


mko-io:
I've heard of Ryu, and it's very impressive, thanks for making this happen to erlang

max-au:
Indeed impressive!
Would it be possible to add benchmarks in forms of Common Tests? So it could be verified on different platforms (at leas ARM). 
It would also be great to move old (Dragon4) implementation into tests, to compare with new Ryu implementation (ideally quickcheck, should be very simple test to write).

DianaOlympos:
@max-au I mean we could but it means that if we changed the way we display the number, as happened recently, we would have to maintain both version. I get why you want it, but i think CT is really not built for benchmarking. I think it would make more sense to have a way to get a real suite of regular benchmark outside of this repo to track regression.

max-au:
Benchmarks in forms of Common Test can be added to the main repo, but disabled via specs by default. See example in SSL: https://github.com/erlang/otp/blob/master/lib/ssl/test/ssl_bench.spec

For eqc/propery-based check, I was suggesting to have it in io_lib_SUITE (it appears there is no separate test SUITE for io_lib, as most is covered by io_SUITE.erl) just to ensure that current Dragon4 implementation produces the same output as Ryu. However I agree this may overcomplicate tests.

DianaOlympos:
@max-au fwiw i did work this under PropEr against the current version while developping it

max-au:
In that case would it be possible to salvage PropEr test (it is usually compatible or easily converted with eqc/Triq/...) ?

DianaOlympos:
@max-au sure but at that point you need to maintain the second implementation, which is the original problem. I mean the generation is basically take the float generator and check they are equivalent. The problem rise when you need to maintain the second implementation. There is no good solution there.

kjellwinblad:
@DianaOlympos, thanks for the pull-request!

> It is worth noting that this implementation is slightly slower (10% to
> 30%) for "small" doubles in some cases. This commit does not try to
> optimise this case yet.

Can you define what a "small" double is? Does it mean that it has a short string representation or that it represents a small number? Is there some easy way to optimize the algorithm for "small" doubles?

Keeping the old implementation as a reference implementation in a test suite so that we can check the new implementation against the old (e.g., using PropEr) sounds like a good idea. We will indeed need to change both implementations if we want to change the string representation, but historically that has happened extremely rarely, so I don't think that is a big concern. `io_lib_format:fwrite_g/1` is currently tested in `lib/stdlib/test/io_SUITE.erl` so I think this test suite is the most suitable. 

I agree with @max-au that it would be good to have benchmarks in a common test suite. How to write such benchmarks is described in `$ERL_TOP/HOWTO/BENCHMARKS.md`. We (the Ericsson Erlang/OTP team) automatically run such benchmarks daily and plot graphs to find performance problems early.

I will get back to you regarding whether we prefer a C implementation instead soon.

DianaOlympos:
@kjellwinblad mostly, it is one with a relatively small absolute exponent. That is the best i have been able to provide, because when i try to better define it by trying to profile what happens in the code, the profiling itself reverse the results... That is something i still would love to handle, but i am a bit at wits end on that.

For the implementation in tests, fair enough, i will add this.

For the benchmark, sure enough. I will see what i can do. Picking the doubles to test is a relatively hard problem but i may be able to get something out of the Ryu benchmark themselves, they only take a few hours to run.

For the C part, i don't think it is an either/or. We can merge this one, then move it into float_to_list/binary_to_list later.

kjellwinblad:
> For the benchmark, sure enough. I will see what i can do. Picking the doubles to test is a relatively hard problem but i may be able to get something out of the Ryu benchmark themselves, they only take a few hours to run.

What makes it difficult to select doubles for benchmarking? Is it because you want to have something that represents real applications, and it is difficult to know what types of doubles are most common in real applications? Perhaps we can have a set of benchmarks for different ranges/types of doubles that are interesting? It would be easy to see the strengths and weaknesses of different algorithms using this approach.

> For the C part, i don't think it is an either/or. We can merge this one, then move it into float_to_list/binary_to_list later.

This sounds like a reasonable approach to me.

> For the implementation in tests, fair enough, i will add this.

Correctness is extremely important. I think it is worth spending some time coming up with good tests to be confident that the new algorithm is correct and that it gives the same results as the old one.

Please, let me know if you have some questions regarding benchmarking, testing or anything else.

DianaOlympos:
> What makes it difficult to select doubles for benchmarking? Is it because you want to have something that represents real applications, and it is difficult to know what types of doubles are most common in real applications? Perhaps we can have a set of benchmarks for different ranges/types of doubles that are interesting? It would be easy to see the strengths and weaknesses of different algorithms using this approach.

The problem here is that picking doubles at all is hard. There is no good way to know because we are moving in between two representation here that do not lend themselves to nice definition of "ranges". The only real benchmark that has value is over the whole set of 2^63 doubles. I will try to think of something a bit better

> Correctness is extremely important. I think it is worth spending some time coming up with good tests to be confident that the new algorithm is correct and that it gives the same results as the old one.

My problem here is that the previous algorithm was not proven correct either. And even worse, it is technically not correct anymore since we introduced the 2^53 cutoff for integer like. So keeping a whole old implementation that was not tested more than the test already in the suite does not feel to me like injecting more correctness. It make sure there is no change in the behaviour, but that does not mean it is correct. And there is already some pretty complex setup of test to validate the correctness here.

As a side note, the reason i feel relatively confident pointing that correctness problem is that frequently when testing Ryu itself, the Ryu implementers discovered that a lot of widely used float printing functions in stdlib were... incorrect 😄 This is a relatively niche and badly covered topic. That is why i mentionned the MS STL set of test for charconv because they are afaict the most complete set of test for this behaviour as of today. (they also are like 55% of the whole MS STL test suite just for the charchonv stuff, which is scary)

That aside, i will be working on something for the benchmark tomorrow first of all.

DianaOlympos:
@kjellwinblad dumb question, would you prefer the benchmark to be on a separate PR, so that we have a baseline with the previous implementation ?

garazdawi:
> dumb question, would you prefer the benchmark to be on a separate PR, so that we have a baseline with the previous implementation ?

Normally I would say that it does not matter as we run all benchmarks on all branches so that we can compare performance in between releases. However, since the JIT lives only in master, that changes the benchmarks a lot so we cannot really compare maint vs master. So in conclusion, having the benchmarks as a separate PR would be good.

uabboli:
I'll just mention something that seems to have been missed, namely
that io_lib_SUITE:g_t/1 (and /2) can and should be used for testing
generated floats. It implements the spec (without the recent addition)
and does not depend on any implementation.

So there is no need to keep the old implementation; use g_t() instead.

DianaOlympos:
> I'll just mention something that seems to have been missed, namely
> that io_lib_SUITE:g_t/1 (and /2) can and should be used for testing
> generated floats. It implements the spec (without the recent addition)
> and does not depend on any implementation.

I suppose you mean io_SUITE. But yes indeed and the recent addition is added in test on top of it.

> So there is no need to keep the old implementation; use g_t() instead.

I agree. In this case, this PR can be merged as soon as the benchmark are offered, working on it.

DianaOlympos:
See #2983 for benchmarks. I think the small digits benchmark i added should show the discrepancy in speed where the dragon4 algorithm beat ryu. I think the reason is that Dragon4 work left to right to generate digits, while Ryu works right to left. I think it would be possible to optimise Ryu to be on par with Dragon4 here if that would be needed, but would probably a longer endeavour.

DianaOlympos:
@garazdawi could we get the benchmark from #2983  for this one so we can compare ?

@kjellwinblad following what @uabboli is saying, what would you like to get in term of additional tests, if any, to progress in this ?

kjellwinblad:
> @kjellwinblad following what @uabboli is saying, what would you like to get in term of additional tests, if any, to progress in this ?

You mentioned that you used PropEr against the current version while developing the algorithm. My initial thought was that you could add these PropEr based tests to a test suite. Then we can reuse the same tests if we need to change the algorithm in the future and they might catch some bugs if we are lucky.

Perhaps this is still doable, even if we don't keep the previous implementation, if one can use `g_t/1` and `g_t/2` to check the result as  @uabboli mentioned? There is basic property based testing support built into common test (https://erlang.org/doc/man/ct_property_test.html) that can use PropEr as a backend (we have PropEr installed on our test machines that run nightly tests). [Here is an example showing how PropEr is used in one of our test suits](https://github.com/erlang/otp/blob/7ca7a6c59543db8a6d26b95ae434e61a044b0800/erts/emulator/test/hash_property_test_SUITE.erl). If you add PropEr based tests for `fwrite_g`, you can place them in a new test suite called `io_lib_format_fwrite_g_property_test_SUITE.erl` so that they are separated from non-property based tests.

We can probably merge this pull-request even without more tests as we already have quite a lot of tests for `fwrite_g` but I think it can be beneficial to add the PropEr tests that you have already written if they can cover something that the already existing tests do not cover or if they can be helpful when we do changes to the algorithm in the future.

DianaOlympos:
I can look at this, but i would lean into merging without the PropEr test.

The reason i used them is that it was simply simpler than writing unit test for the first pass. But they were not able to find any of the edge case and in general i have found the float generator and shrinker for PropEr to not be really useful for this kind of work. I would love to see if we could create a generator and shrinking strategy that would make sense but i think that this would be paper worthy if we found one.

I would say that this is still open research and merge it as is. If someone want to work with me on that, i would happily participate though :smile:

garazdawi:
> could we get the benchmark from #2983 for this one so we can compare?

float_bench_ryu

Looks like the benchmark shows the Ryu implementation to be about 4 times faster than the Dragon4, which makes the total speedup with JIT + Ryu 5 times faster. Though the numbers seem to be a lot less stable with Ryu. That could just be because the time we are measuring is so small that other factors come into play.

garazdawi:
```
** Inconsistent app file, Modules not included in app file. (superfluous_mods_in_ebin) ->
[io_lib_format_ryu_table]
```

I noticed the stdlib app_SUITE:app_test test case failed as a result of this patch.

bjorng:
This is wrong:

```
1> io:format("~p\n", [10.0]).
0.0e2
ok
```

Fortunately, we caught in a failed compiler test case.

DianaOlympos:
Going to look at these today, thank you so much

bjorng:
Please follow our guidlines [writing good commit message](https://github.com/erlang/otp/wiki/Writing-good-commit-messages).

In particular, the first line should be shortish and the second blank should be blank. Otherwise merge commits and the output of `git log --oneline` will look very ugly.

DianaOlympos:
ok so i did a temporary commit that fix the problem that @bjorng showed. I am asking to not merge for now. My problem is that this bug should have been caught by the existing test suite ( https://github.com/erlang/otp/blob/master/lib/stdlib/test/io_SUITE.erl#L1051) and it did not.

I want to understand why it did not first.

DianaOlympos:
Ok so small update, still investigating. This seems to be a bug in the fwrite test suite for float going back at least to R13

DianaOlympos:
OK this should fix it. First the easy stuff

* fixed the bug reported by @bjorng 
* fixed another bug caught by...
* fix the io_SUITE so that a failing test is not swallowed and does not bypass the following tests silently (see below)
* added the new module to `stdlib.app.src` thanks @garazdawi 


I would love to squash these but first i would like to know if you have a better idea for the test part. Here is what happened.

The IO suite would _pass as green_ if a test in the foreach failed. And then not run the rest of the list of tests. Which make sense, but swallowed a lot of these that _should_ have been caught. This has probably been undetected for a decade or more, because the related code has not changed.

Of note, PropEr on my test code had not caught any of these either...

Good news is, running the fixed tests on otp 23 seems to pass. So there was nothing broken in the old code either ! (no strong proof, i may have runned the wrong code, you may want to run this on the master branch to be sure).

ofc once that is fixed, the bug are far easier to fix. I think we are better now and that should be mergeable now.

kjellwinblad:
@DianaOlympos have you looked into how other Ryu implementations are tested? Perhaps we can get some hints from there about what corner cases that can be good to test. Perhaps white-box testing is the best approach to test this kind of algorithms. By knowing the algorithm and all the corner cases one can come up with test cases that covers the parts that are difficulty to cover by random testing. @uabboli mentioned to me that a white-box testing approach was used to come up with test cases for the current implementation so the current test cases are tailored to the current implementation and are not necessary the best for Ryu.

DianaOlympos:
@kjellwinblad yes definitely. As i mentionned in my github
Yes definitely !
As i mentioned in the initial comment here
> to get a better coverage by reusing the Microsoft STL test suite as an inspiration.

Because Microsoft STL use Ryu. I have preferred to use the existing test suite because it is the best description of the behaviour expected i knew. I do think that keeping that test suite expansion for the C implementation if that ever happens make more sense, but i can try to bring more.

If you want to peruse, the test suite is here https://github.com/microsoft/STL/tree/main/tests/std/tests/P0067R5_charconv and we are mostly interested by this file for the algorithm implemented here https://github.com/microsoft/STL/blob/main/tests/std/tests/P0067R5_charconv/double_to_chars_test_cases.hpp but not only.

Ryu use https://github.com/ulfjack/ryu/blob/master/ryu/tests/d2s_test.cc which afaict are covered by the existing io_SUITE test suite.

The main difference is that we have more edge cases because the existing tests defined a more specific shortest printing than the print loop of Ryu (Ryu only print an exponent version, we print the shortest) which is closer to the charconv.

But we also have added recently, by @jhogberg , a special case for some integer that would be shortest as `.0` to be done as exponent to show a particular lack of precision.

So at this point i would say that the OTP test suite covers better than what i can add. I can for sure try to add some edge cases possibly but i think they are already covered.

DianaOlympos:
I would point out that the reason these edge cases were not caught before was because the test framework used swallowed the errors... Not because they were not in the test suite.

kjellwinblad:
> So at this point i would say that the OTP test suite covers better than what i can add. I can for sure try to add some edge cases possibly but i think they are already covered.

@DianaOlympos thank you for the answer. I think it is better to add one test case too much rather than having to little tests, if there is some corner case that might not be covered.

I will give you more feedback after I have had time to look at Microsoft's test suit.

DianaOlympos:
> @DianaOlympos thank you for the answer. I think it is better to add one test case too much rather than having to little tests, if there is some corner case that might not be covered.

I added some test based on the Ryu white box one from https://github.com/ulfjack/ryu/blob/master/ryu/tests/d2s_test.cc. 

They all pass locally

lib/stdlib/test/io_SUITE.erl
```diff
@@ -1043,6 +1044,90 @@ g_choice() ->
     lists:foreach(fun(V) -> g_t(V) end, L),
     ok.
 
+g_ryu() ->
+    %% specific white box tests that should trigger specific edge cases
+    %% to the ryu algorithm
+
+    %% this list is regression tests from the ryu C ref implementation
```
kjellwinblad, line 1051: I think you can add "https://github.com/ulfjack/ryu/blob/master/ryu/tests/d2s_test.cc" to this comment so that it is easy to find where the test cases re coming from.

kjellwinblad:
> I added some test based on the Ryu white box one from https://github.com/ulfjack/ryu/blob/master/ryu/tests/d2s_test.cc.

Looks good to me except a small comment (see above).

I have had a look at "https://github.com/microsoft/STL/blob/main/tests/std/tests/P0067R5_charconv/double_to_chars_test_cases.hpp". I think it would be an easy task to just take all the doubles from the test cases in this file and add a test case for them in `io_SUITE.erl` (i.e., we check all these doubles with `g_t/1`). The file"double_to_chars_test_cases.hpp" sees to be licensed under "Apache-2.0 WITH LLVM-exception". This seems to be a loser form of Apache-2.0, if I understand the exception correctly, so I don't think there would be any license problem if we write a comment about where the test cases are coming from and what license they are under. @DianaOlympos what do you think about this?

DianaOlympos:
@kjellwinblad sure i will try. More copy pasta with multiple carets editors  😄

DianaOlympos:
@kjellwinblad I added the tests. I bypassed some that were already here or made no sense (specific to hex representation mostly).

Do you want more before i squash ?

kjellwinblad:
@DianaOlympos thank you, I think this looks good. Please go ahead with the squash!

DianaOlympos:
Done, squashed

PR #2961, opened 2021-01-05
Add a type for application:start's result

paulo-ferraz-oliveira:
I don't expect `application:start`'s result to change that much, but I keep having to write the same elements (the ones in this new type) over and over, so it seems more appropriate to do it here, upstream.

I'd also like to note that `{error, Reason}` is not documented in callback `:start_phase` or `:start` except by reference (e.g. in `application:ensure_all_started` or `application:start`, though this is probably expected).

### DIFF
--- a/lib/kernel/doc/src/application.xml
+++ b/lib/kernel/doc/src/application.xml
@@ -58,6 +58,9 @@
       <name name="start_type"/>
     </datatype>
     <datatype>
+      <name name="start_result"/>
+    </datatype>
+    <datatype>
       <name name="restart_type"/>
     </datatype>
     <datatype>
@@ -523,13 +526,14 @@ Nodes = [cp1@cave, {cp2@cave, cp3@cave}]</code>
         <c>application</c> callback module.</p>
     </fsdescription>
     <func>
-      <name since="">Module:start(StartType, StartArgs) -> {ok, Pid} | {ok, Pid, State} | {error, Reason}</name>
+      <name since="">Module:start(StartType, StartArgs) -> StartResult</name>
       <fsummary>Start an application.</fsummary>
       <type>
         <v>StartType = <seetype marker="#start_type"><c>start_type()</c></seetype></v>
         <v>StartArgs = term()</v>
         <v>Pid = pid()</v>
         <v>State = term()</v>
+        <v>StartResult = <seetype marker="#start_result"><c>start_result()</c></seetype></v>
       </type>
       <desc>
         <p>This function is called whenever an application is started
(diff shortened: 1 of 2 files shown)


garazdawi:
Why do you find yourself copying the application return value spec?

Don't you return either a narrower type (i.e. `{ok,pid()}`), or what the main supervisor returns (i.e. `supervisor:startlink_ret()`)?

paulo-ferraz-oliveira:
> Why do you find yourself copying the application return value spec?

Because of templates (e.g. applications that are functional only after they're _instantiated_ - think `rebar3` templates), mostly. I don't exactly know the return type that will be used, but don't want to force the consumer of the template to waste time finding out, either (a further use case is explained below). In any case, a templated generic supervisor can be used with a worker that is not controlled by me, which also adds to this (and a templated generic app can be used with a supervisor that is not controlled by me, ...).

> Don't you return either a narrower type (i.e. `{ok,pid()}`), or what the main supervisor returns (i.e. `supervisor:startlink_ret()`)?

I'm not exactly sure `application`'s spec. is compatible with the `supervisor`'s, since the former doesn't include `ignore` and `startlink_err` is more restrictive than `term` for the error reason, otherwise I could've simply `{ok, pid(), State} | supervisor:startlink_ret()`, I guess.

Do you see a disadvantage to having this type exported? I can see another one to not having it: I have a template that is used to spawn many applications. The template (as-is) defines an application/supervisor/worker structure. The application is `-spec(_).`ed as `-> {ok, Pid} | {ok, Pid, State} | {error, Reason}`, which means all "instances" of the template are spec'ed the same. They all evolve in a different-OTP-versions' ecosystem, but in the same company. Now let's suppose (maybe it's not expected, as I said before) we add `ignore` to the `-spec(_).` whose type I just exported. Later on, I want to make all those 150 apps be compatible with the Erlang version that supports that - let's say OTP 27 (and since they might use different supervisors at different stages of evolution and different purposes): it would be easier to not have to change anything, since the type is controlled by OTP, than to have to create 150 pull requests for maintenance purposes.

Then again one can argue that I could wrap around `application` and have e.g. `gen_application` (probably a bad name) to achieve the same goal, but I though it wouldn't hurt to have the type exported. I'm all for keeping as is, if you see it not fit for OTP and I'll eventually find an alternative solution to my problem.

paulo-ferraz-oliveira:
So... I made my template supervisor return `supervisor:start_link(SupName, Module, Args)`, and my templated application return `{{name}}_sup:start_link()`, which is basically `supervisor:startlink_ret()`. This means my application can now return `ignore` (though I don't explicitly do this in the supervisor). I'll `dialyze` and try to fix to what is "expected", and let this issue drop. Thanks for your review.

PR #2963, opened 2021-01-08
Simplify configuration of cross-compiling to Android x86_64

paulo-ferraz-oliveira:
I was fiddling around with [@JeromeDeBretagne's erlanglauncher's fork](https://github.com/JeromeDeBretagne/erlanglauncher) to get a sense of how to use `jinterface` in an Android env. and wanted to use an x86_64 arch. emulator (that project targets ARM but I couldn't get an ARM emulator to work correctly or with a decent performance). I looked at the [OTP doc.](https://github.com/erlang/otp/blob/master/HOWTO/INSTALL-ANDROID.md) and followed it until I figured I didn't have a readily-available `xcomp.conf` for my use case. This pull request presents one, heavily inspired by `erl-xcomp-arm64-android.conf`.

I'm not sure there's a need to add tests for this, or what other elements I'd need changed in the scope of this pull request, but will gladly follow your guidance.

### DIFF
--- a/xcomp/erl-xcomp-x86_64-android.conf
+++ b/xcomp/erl-xcomp-x86_64-android.conf
@@ -1,0 +1,280 @@
+## -*-shell-script-*-
+##
+## %CopyrightBegin%
+##
+## Copyright Ericsson AB 2021. All Rights Reserved.
+##
+## Licensed under the Apache License, Version 2.0 (the "License");
+## you may not use this file except in compliance with the License.
+## You may obtain a copy of the License at
+##
+##     http://www.apache.org/licenses/LICENSE-2.0
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+##
+## %CopyrightEnd%
+##
+## File: erl-xcomp-x86_64-android.conf
+## Author: Paulo Oliveira
+##
+## -----------------------------------------------------------------------------
+## When cross compiling Erlang/OTP using `otp_build', copy this file and set
+## the variables needed below. Then pass the path to the copy of this file as
+## an argument to `otp_build' in the configure stage:
+##   `otp_build configure --xcomp-conf=<FILE>'
+## -----------------------------------------------------------------------------
+
+## Note that you cannot define arbitrary variables in a cross compilation
+## configuration file. Only the ones listed below will be guaranteed to be
+## visible throughout the whole execution of all `configure' scripts. Other
+## variables needs to be defined as arguments to `configure' or exported in
+## the environment.
+
+## -- Variables for `otp_build' Only -------------------------------------------
+
+## Variables in this section are only used, when configuring Erlang/OTP for
+## cross compilation using `$ERL_TOP/otp_build configure'.
+
+## *NOTE*! These variables currently have *no* effect if you configure using
+## the `configure' script directly.
+
+# * `erl_xcomp_build' - The build system used. This value will be passed as
+#   `--build=$erl_xcomp_build' argument to the `configure' script. It does
+#   not have to be a full `CPU-VENDOR-OS' triplet, but can be. The full
+#   `CPU-VENDOR-OS' triplet will be created by
+#   `$ERL_TOP/erts/autoconf/config.sub $erl_xcomp_build'. If set to `guess',
+#   the build system will be guessed using
+#   `$ERL_TOP/erts/autoconf/config.guess'.
+erl_xcomp_build=guess
+
+# * `erl_xcomp_host' - Cross host/target system to build for. This value will
+#   be passed as `--host=$erl_xcomp_host' argument to the `configure' script.
+#   It does not have to be a full `CPU-VENDOR-OS' triplet, but can be. The
+#   full `CPU-VENDOR-OS' triplet will be created by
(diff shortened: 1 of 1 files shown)


xcomp/erl-xcomp-x86_64-android.conf
```diff
@@ -0,0 +1,280 @@
...
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
+## limitations under the License.
+##
+## %CopyrightEnd%
+##
+## File: erl-xcomp.conf.template
+## Author: Rickard Green
```
garazdawi, line 22: You'll want to change these to be correct

xcomp/erl-xcomp-x86_64-android.conf
```diff
@@ -0,0 +1,280 @@
...
+## not need to set these variables (where `<HOST>' is what has been passed as
+## `--host=<HOST>' argument to `configure').
+
+## All variables in this section can also be used when native compiling.
+
+# * `CC' - C compiler.
+CC=x86_64-linux-$NDK_ABI_PLAT-clang
+
+# * `CFLAGS' - C compiler flags.
+CFLAGS=-g O2
```
garazdawi, line 78:
```suggestion
CFLAGS=-g -O2
```
Did these CFLAGS work for you?

garazdawi (changes requested):
The files in xcomp are examples that we do not test nor update ourselves. I'm happy to add this as another example to help others doing the same thing as you.

xcomp/erl-xcomp-x86_64-android.conf
paulo-ferraz-oliveira: Shall I also change the copyright year to 2021?

garazdawi: Yes, please do that.

xcomp/erl-xcomp-x86_64-android.conf
paulo-ferraz-oliveira: They did. I'll check again, though. It's possible I erased it before committing, by mistake.

paulo-ferraz-oliveira: It does generate an error before configuring the bootstrap build system.

```
./xcomp/erl-xcomp-x86_64-android.conf: line 78: -O2: command not found
```

But then `./otp_build configure` still finishes with success. And I made it print what flags it was using and they were `-fno-common -g -O2 -I...`.

garazdawi: I guess that it falls back to the default for some reason, which is what you are trying to set anyway. I would just comment out the CFLAGS setting all together as it does nothing usefull.

paulo-ferraz-oliveira: They have to be inside `"`, though I'm still trying to figure out why they aren't being taken into account.

paulo-ferraz-oliveira: Strange: in `otp_build`, after `. "$xcomp_conf"`, `CPP_FLAGS` are set correctly, as in the `xcomp.conf` file, but then they're not when I do `./otp_build configure`.

paulo-ferraz-oliveira: These are the `CONFIG_FLAGS` that are getting generated: `--host=x86_64-linux-android CFLAGS=\ -g\ -O2\ -march=native --with-ssl=/usr/local/opt/openssl\@1\.1 --disable-dynamic-ssl-lib --disable-hipe --without-termcap --without-wx --enable-builtin-zlib --with-javac=JAVAC --disable-hipe --without-termcap --without-wx --enable-builtin-zlib CC=x86_64-linux--clang CFLAGS=-g\ -O0 CXX=x86_64-linux--clang++ LD=x86_64-linux-android-ld\.gold` (I used `O0` for easier comparison). Does this some off to you?

xcomp/erl-xcomp-x86_64-android.conf
paulo-ferraz-oliveira: Copyright year, File and Author updated.

xcomp/erl-xcomp-x86_64-android.conf
paulo-ferraz-oliveira: > I guess that it falls back to the default for some reason, which is what you are trying to set anyway. I would just comment out the CFLAGS setting all together as it does nothing usefull.

You mean it's not expected to be used in the `configure` procedure, then? I'll comment it out, as per your suggestion.

paulo-ferraz-oliveira: Hm... I guess you meant it doesn't serve any purpose in _this_ case, but I feel it's not working as expected, either. (?)

garazdawi: > Hm... I guess you meant it doesn't serve any purpose in this case. (?)

Yes, in your case it does not matter as you set it to the same value as the default.

> but I feel it's not working as expected, either

I don't know why CFLAGS is showing up twice in your setup. It does not happen in our tests of another xcomp file (https://github.com/erlang/otp/blob/master/xcomp/erl-xcomp-powerpc-dso-linux-gnu.conf).

paulo-ferraz-oliveira: Cool. Let me fix it (if I haven't already).

garazdawi (approved):
look good to me, if you are happy I will merge this.

paulo-ferraz-oliveira:
> look good to me, if you are happy I will merge this.

I am happy :-)

PR #2964, opened 2021-01-11
Handle macros in patterns

gomoripeti:
The dodger used to replace a macro with a local function call. That works in
expressions but not in patterns like function heads or catch patterns. Instead
of a function call a tuple syntax is used which works in more places.

This change is motivated by the commonly used `?EXCEPTION` macro

```
try
  ...
catch
  ?EXCEPTION(Class, Reason, StackToken) ->
    ...
end.
```

Open source examples:
https://github.com/devinus/poolboy/commit/5d40cc517edc9bb8ee70756544167b63f66662f0
https://github.com/eproxus/meck/commit/558e925b48ce257b12e381080c851dc49c87d7bb

Fixes ERL-1429

### DIFF
--- a/lib/syntax_tools/test/syntax_tools_SUITE_data/syntax_tools_test.erl
+++ b/lib/syntax_tools/test/syntax_tools_SUITE_data/syntax_tools_test.erl
@@ -6,7 +6,7 @@
 
 -module(syntax_tools_test).
 
--export([foo1/0,foo2/2,foo3/0,foo4/3,foo5/1]).
+-export([foo1/0,foo2/2,foo3/0,foo4/3,foo5/1,foo6/2]).
 
 -include_lib("kernel/include/file.hrl").
 -record(state, { a, b, c, d}).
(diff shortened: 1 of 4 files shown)


lib/syntax_tools/test/syntax_tools_SUITE_data/syntax_tools_test.erl
```diff
@@ -48,8 +49,7 @@ foo1() ->
 %% macro test
 foo2(A,B) ->
     % string combining ?
-    [?macro_string, ?macro_string
-     ?macro_string,
+    [?macro_string, ?macro_string ++
```
gomoripeti, line 52: I added a check in the test suite that `epp_dodger` does not return any error marks for better verification that it does its job. This is the only place where it did. Here `epp_dodger` has no chance of realizing that the 2 consecutive macros are two strings so I added an explicit concatenation. Additionally the `epp_dodger_clever` testcase contains scenarios when `epp_dodger` actually has a chance (one of the two positions is a literal string)

bjorng:
Thanks for the pull request. Added to our daily builds.

PR #2966, opened 2021-01-11
Make yielding `maps:from_list/1`

kjellwinblad:
The function `maps:from_list/1` did not trap (yield) before this PR, which could lead to unfair scheduling. This
issue is fixed by making sure that `maps:from_list/1` calls with a long list as parameter value traps.

Trapping versions for all major functions that the implementation of `maps:from_list/1` uses are generated with the [Yielding C Fun tool](https://github.com/erlang/otp/tree/master/erts/lib_src/yielding_c_fun).

I have compared the performance of the trapping version of `maps:from_list/1` with the old version. You can see the results below:

```
List Size	Old(ms)		New(ms)		% Slower than original
100		6.23617e-3	1.31301e-2	110.548088
1000		1.22613e-1	2.08583e-1	70.114630
10000		1.68739e+0	2.81539e+0	66.848444
100000		2.21308e+1	3.39905e+1	53.589057
1000000		2.78427e+2	4.26256e+2	53.094293
10000000	3.28488e+3	4.83911e+3	47.314839

Machine: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz, 32 GB of memory, gcc version 7.5.0, Ubuntu 18.04
```
**Edit:** The performance is much better after some optimizations. See [this comment](https://github.com/erlang/otp/pull/2966#issuecomment-761114951) for updated measurements.

The relative differences will probably be smaller in a real system (at least with the shorter lists). In a real system, less data will be in low-level processor caches, so the overhead added by the yielding will be smaller relative to the total run time.

Previously, `maps:from_list/1` utilized the qsort function from the C standard library. This pull-request adds a new quick-sort implementation to ERTS that can trap. Therefore, some of the performance differences could be related to differences between the qsort function from the C standard library and the new sort function.

The escript used to do the measurements:

```erlang
-module(main).
-mode(compile).
-export([main/1]).

main(["compare", EScriptPath1, EScriptPath2]) ->
    Run = fun(Script, Size) ->
                  Repeats = case 100000000 div Size of 0 -> 1; R -> R end,
                  ResStr = os:cmd(io_lib:format("~s ~s ~p ~p", [Script, escript:script_name(), Size, Repeats])),
                  erlang:list_to_float(ResStr)
          end,
    io:format("List Size\tOld(ms)\t\tNew(ms)\t\t% Slower than original~n"),
    lists:foreach(
      fun(Size) ->
              Time1 = Run(EScriptPath1, Size),
              Time2 = Run(EScriptPath2, Size),
              io:format("~p\t\t~e\t~e\t~f~n",
                        [Size, Time1, Time2, ((Time2 - Time1) / Time1) * 100])
      end,
     [100, 1000, 10000, 100000, 1000000, 10000000]);
main([SizeParameter, RepeatsParam]) ->
    Size = erlang:list_to_integer(SizeParameter),
    List = [{X,X} || X <- [X || {_,X} <- lists:sort([ {rand:uniform(), N} || N <- lists:seq(1,Size)])]],
    Repeats = erlang:list_to_integer(RepeatsParam),
    Time =
        lists:foldl(fun(_, Acc) ->
                            {Time, _} = timer:tc(fun()-> maps:from_list(List) end),
                            Acc + Time
                    end, 0, lists:seq(1, Repeats)),
    io:format("~p", [(Time / Repeats) / 1000.0]),
    ok.

```


bjorng (approved):

kjellwinblad:
I have added a commit that enables trapping for `maps:from_keys/2` as well. Only minor changes had to be done as `maps:from_keys/2` is very similar to `maps:from_list/1`.

I have updated the benchmark script that I posted in the PR description so that it can be used to see how `maps:from_keys/2`'s performance is affected by trapping as well. The results are not surprisingly very similar to the results for `maps:from_list/1`:

```
List Size	Old(ms)		New(ms)		% Slower than original
100		6.39558e-3	1.25624e-2	96.423356
1000		1.27004e-1	2.14930e-1	69.230618
10000		1.75229e+0	2.91144e+0	66.150858
100000		2.23946e+1	3.54021e+1	58.082779
1000000		2.84797e+2	4.23063e+2	48.548828
10000000	3.34947e+3	4.99977e+3	49.270539
```

Here is the updated benchmark script:

```erlang
-module(main).
-mode(compile).
-export([main/1]).

main(["compare", What, EScriptPath1, EScriptPath2]) ->
    Run = fun(Script, Size) ->
                  Repeats = case 100000000 div Size of 0 -> 1; R -> R end,
                  ResStr = os:cmd(io_lib:format("~s ~s ~s ~p ~p", [Script, escript:script_name(), What, Size, Repeats])),
                  erlang:list_to_float(ResStr)
          end,
    io:format("List Size\tOld(ms)\t\tNew(ms)\t\t% Slower than original~n"),
    lists:foreach(
      fun(Size) ->
              Time1 = Run(EScriptPath1, Size),
              Time2 = Run(EScriptPath2, Size),
              io:format("~p\t\t~e\t~e\t~f~n",
                        [Size, Time1, Time2, ((Time2 - Time1) / Time1) * 100])
      end,
     [100, 1000, 10000, 100000, 1000000, 10000000]);
main(["from_list", SizeParameter, RepeatsParam]) ->
    Size = erlang:list_to_integer(SizeParameter),
    List = [{X,X} || X <- [X || {_,X} <- lists:sort([ {rand:uniform(), N} || N <- lists:seq(1,Size)])]],
    Repeats = erlang:list_to_integer(RepeatsParam),
    Time =
        lists:foldl(fun(_, Acc) ->
                            {Time, _} = timer:tc(fun()-> maps:from_list(List) end),
                            Acc + Time
                    end, 0, lists:seq(1, Repeats)),
    io:format("~p", [(Time / Repeats) / 1000.0]),
    ok;
main(["from_keys", SizeParameter, RepeatsParam]) ->
    Size = erlang:list_to_integer(SizeParameter),
    List = [X || {_,X} <- lists:sort([ {rand:uniform(), N} || N <- lists:seq(1,Size)])],
    Repeats = erlang:list_to_integer(RepeatsParam),
    Time =
        lists:foldl(fun(_, Acc) ->
                            {Time, _} = timer:tc(fun()-> maps:from_keys(List, ok) end),
                            Acc + Time
                    end, 0, lists:seq(1, Repeats)),
    io:format("~p", [(Time / Repeats) / 1000.0]),
    ok.
``'

**Comment by kjellwinblad** [comment]
I have commited an optimization for the `erts_qsort` function that is used by both `maps:from_list/1` and `maps:from_keys/2`. Now, qsort from the C standard library is used for small arrays. This substantially improves the performance. Here are the new results for `maps:from_keys/2`:

```
List Size	Old(ms)		New(ms)		% Slower than original
100		6.15298e-3	7.16553e-3	16.456219
1000		1.21962e-1	1.55741e-1	27.696875
10000		1.74244e+0	2.13975e+0	22.801869
100000		2.16292e+1	2.73287e+1	26.350868
1000000		2.74273e+2	3.48314e+2	26.995165
10000000	3.28584e+3	4.24979e+3	29.336327
```

Thanks @sverker for helping me in finding this optimization.

**Comment by sverker** [comment]
After even more optimization of `erts_qsort`:
```
List Size	Old(ms)		New(ms)		% Slower than original
100		7.20136e-3	8.15529e-3	13.246583
1000		1.34849e-1	1.40610e-1	4.272393
10000		1.86830e+0	1.78089e+0	-4.678898
100000		2.46245e+1	2.09362e+1	-14.978234
1000000		3.06814e+2	2.46154e+2	-19.771002
10000000	3.60470e+3	2.84435e+3	-21.093425
```
How about that, looks like we are beating libc `qsort` for large arrays.

File: erts/emulator/beam/utils.c [hunk 1]
```diff
@@ -5208,11 +5208,8 @@ erts_qsort_partion_array(byte *base,
     byte* second_part_start = base + (nr_of_items * item_size);
     byte* curr = base + item_size;
     int found_bigger = 0;
-    size_t pivot_index =
-        (size_t)erts_sched_local_random_hash_64_to_32_shift((Uint64)extra_seed +
-                                                            (Uint64)((UWord)compare) +
-                                                            (Uint64)((UWord)base) -
-                                                            (Uint64)nr_of_items) % nr_of_items;
+    size_t pivot_index = nr_of_items / 2;
```
kjellwinblad, line 5211: @sverker I don't think this change is safe because the input can be designed to make the recursion depth O(N) instead of O(log(N)), if the pivot is not randomly selected.

sverker: Ok, maybe we should find a cheaper way to do it then, because it was quite a performance gain when I removed the random pivot.

erts/emulator/beam/utils.c
```diff
@@ -5206,48 +5198,33 @@ static void *erts_qsort_get_item(void *base, size_t item_size, size_t index)
...
-        if (compare_res == 0) {
-            /* Include in pivot part */
-            pivot_part_end++;
-        } else if (compare_res < 0) {
-            /* Move pivot part one step to the right */
-            erts_qsort_swap(base, item_size, pivot_part_start, pivot_part_end);
-            pivot_part_start++;
-            pivot_part_end++;
+    erts_qsort_swap(item_size, base, base + pivot_index*item_size);
+    while (curr != second_part_start) {
```
kjellwinblad, line 5215: Perhaps there is some better way to optimize the partition function so we can keep the pivot partition and still reduce the number of swaps?

If we skip the pivot partition, we should name the erts_qsort function something else and place it in `erl_map.c` because no pivot partition can lead to O(N^2) run time when all sorted items are equal.

erts/emulator/beam/utils.c
```diff
@@ -5241,27 +5338,35 @@ static void erts_qsort_helper(byte *base,
...
-    }
-    if (nr_of_items == 2) {
-        if (compare(base, base+item_size) > 0) {
-            erts_qsort_swap(item_size, base, base+item_size);
-        }
+    erts_qsort_partion_array_result partion_info;
+    size_t nr_of_items_in_first_partion;
+    size_t nr_of_items_second_partion;
+    const size_t stdlib_qsort_cut_off = 16;
+    if (nr_of_items <= stdlib_qsort_cut_off) {
```
sverker, line 5347: Borde kanske byta namn från `stdlib_qsort_cut_off`.

sverker (approved):

erts/emulator/beam/utils.c
kjellwinblad: Fixed in commit below

erts/emulator/beam/utils.c
kjellwinblad: Fixed

erts/emulator/beam/utils.c
kjellwinblad: Fixed

kjellwinblad:
New numbers for `maps:from_keys/2` after even more improvements (see commit above).

```
escript ~/main.erl  "compare" "from_keys" ../otp_2/bin/escript ../otp_3/bin/escript 
List Size	Old(ms)		New(ms)		% Slower than original
100		6.17723e-3	6.18415e-3	0.112008
1000		1.23210e-1	1.12480e-1	-8.708843
10000		1.68500e+0	1.47917e+0	-12.215200
100000		2.15199e+1	1.75660e+1	-18.373081
1000000		2.70852e+2	2.15536e+2	-20.422987
10000000	3.21278e+3	2.53100e+3	-21.220895
```

michalmuskala:
Is that with random keys? What is the behaviour with already sorted keys? It's probably a fairly common case, In particular for something like `maps:map` that uses `maps:from_list` internally, where the fun updates only only values.

kjellwinblad:
> Those are some impressive results!
> 
> Is that with random keys? What is the behaviour with already sorted keys? It's probably a fairly common case, In particular for something like `maps:map` that uses `maps:from_list` internally, where the fun updates only only values.

Thanks! Yes, the results are for random keys. The big sorting tasks that are done by from_list and from_keys sorts the hashes of the keys so I don't think it will help if the input list is already sorted.

PR #2967, opened 2021-01-11
Hawk/otp start loop 22

hawk:
Fix indefinite loop in emulator startup
    
The system hangs when 1024 file descriptors are open at startup:
    
        for ((i = 3; i < 1024; i++)); do
            eval "exec $i>&1;"
        done
    
        erl


hawk:
Wrong commit

PR #2968, opened 2021-01-11
Fix infinite loop in emulator startup

hawk:
The system hangs when 1024 file descriptors are open at startup:

    for ((i = 3; i < 1024; i++)); do
        eval "exec $i>&1;"
    done

    erl

### DIFF
--- a/erts/emulator/sys/common/erl_poll.c
+++ b/erts/emulator/sys/common/erl_poll.c
@@ -646,6 +646,9 @@ int erts_poll_new_table_len(int old_len, int need_len)
     }
     else {
         new_len = old_len;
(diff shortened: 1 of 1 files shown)


rickard-green:
We already have a fix for this in commit 1b03792d9d3739447b802c5ef8eacf33c7451ae3 waiting for release in the next patch on OTP 23, OTP 21 as well as OTP 22. This fix has also already has been merged into the `maint` and `master` branches.

rickard-green:
@hawk Fix released in [OTP 22.3.4.14](http://erlang.org/download/OTP-22.3.4.14.README) now

PR #2972, opened 2021-01-13
crypto: disable EVP usage also for openssl-1.0.2

mikpe:
The initial fix for the crypto hmac performance loss (ERL-1400 / PR-2877) added code to automatically disable use of the EVP API for OpenSSL < 1.0.2. However, the EVP API is slow also in 1.0.2 as reported in ERL-1400. This adjusts the limit to < 1.1.0.

Tested on CentOS7 and verified it gives a decent speedup (roughly 30-40%) for the ERL-1400 test case.


HansN:
Thanks for the PR!

PR #2973, opened 2021-01-14
Add extended error information for the remaining BIFs in STDLIB

bjorng:


lib/stdlib/src/binary.erl
```diff
@@ -70,16 +85,26 @@ bin_to_list(Subject, Pos, Len) when not is_binary(Subject);
...
     [];
-bin_to_list(_Subject, _Pos, 0) ->
-    erlang:error(badarg);
 bin_to_list(Subject, Pos, Len) when Len < 0 ->
-    bin_to_list(Subject, Pos + Len, -Len);
+    try
+        bin_to_list(Subject, Pos + Len, -Len)
+    catch
+        error:Reason ->
+            error_with_info(Reason, [Subject, Pos, Len])
```
michalmuskala, line 97: I wonder if this will have correct stacktrace. Wouldn't this report the raising function as the `error_with_info`, rather than `bin_to_list` since it's a tail call

bjorng: The call will be inlined. See line 40.

michalmuskala: Ah, that makes sense. Sorry for the noise

PR #2976, opened 2021-01-14
Bring HOWTO/BENCHMARKS.md formatting to modern markdown

DianaOlympos:
The formatting was driving me crazy.

Also uh, should we get rid of the "This will hopefully change in the near future."

This is 9 years old and afaict this is still not available...

### DIFF
--- a/HOWTO/BENCHMARKS.md
+++ b/HOWTO/BENCHMARKS.md
@@ -7,25 +7,25 @@ run benchmarks you have to [release the tests][] just as you normally would.
 
 Note that many of these benchmarks were developed to test a specific feature
 under a specific setting. We strive to keep the benchmarks up-to-date, but alas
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
I [markdownlint](https://github.com/DavidAnson/markdownlint)ed the file and (apart from some trailing white space):
* got a warning that line 70 is not required (it's empty), as isn't line 74,
* got another warning that lines 53, 54, and 55 are using TABS, not SPACEs.

garazdawi:
> Also uh, should we get rid of the "This will hopefully change in the near future."
> 
> This is 9 years old and afaict this is still not available...

:) You can remove the "near" part, I still hope to make it possible to publish the results of our benchmark runs.

garazdawi:
Thanks for your contribution!

PR #2977, opened 2021-01-14
wxNotificationMessage support

dominicletz:
Hey @dgud,

this is my first try on adding a new component to wxWidgets. In the examples/simple/menu.erl I've added in the file menu an option to show the notification. 

Note:
I've done conditional handling of the events since those are only available starting in 3.1.0, have tried that with 3.0.4 and it does seem to deactivate them successfully, haven't yet tried with 3.1.x yet though will do later.

### DIFF
--- a/lib/wx/api_gen/interface/wx/notifmsg.h
+++ b/lib/wx/api_gen/interface/wx/notifmsg.h
@@ -39,13 +39,13 @@
 
 
     @beginEventEmissionTable{wxCommandEvent}
-    @event{wxEVT_NOTIFICATION_MESSAGE_CLICK(id, func)}
+    @event{EVT_NOTIFICATION_MESSAGE_CLICK(id, func)}
            Process a @c wxEVT_NOTIFICATION_MESSAGE_CLICK event, when a notification
            is clicked.
-    @event{wxEVT_NOTIFICATION_MESSAGE_DISMISSED(id, func)}
+    @event{EVT_NOTIFICATION_MESSAGE_DISMISSED(id, func)}
            Process a @c wxEVT_NOTIFICATION_MESSAGE_DISMISSED event, when a notification
            is dismissed by the user or times out.
-    @event{wxEVT_NOTIFICATION_MESSAGE_ACTION(id, func)}
+    @event{EVT_NOTIFICATION_MESSAGE_ACTION(id, func)}
            Process a @c wxEVT_NOTIFICATION_MESSAGE_ACTION event, when the user
            selects on of the actions added by AddAction()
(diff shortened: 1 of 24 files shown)


dominicletz:
And tested against 3.1 as well, the conditionals seem to work well. I can get the click events in 3.1 and the not supported in this version message in the sample.

dominicletz:
When generating the code I'm seeing two warnings I'm not sure what to make of:
```
632:wxNotificationMessage:632: xml unhand: title
  [{text,"Windows        "}]
632:wxNotificationMessage:632: xml unhand: title
  [{text,"macOS          "}]
```
I'm assuming this is just from the documentation explaining details about the different platform versions in a different text format than usually and nothing to worry about.

dgud:
The documentation warnings is just a remainder that the doc generation is not perfect yet.

dominicletz:
I've moved the wxNotificationMessage example to the demo/ app

dgud:
You have a typo in the Makefile so it doesn't build.

Can you edit the commit messsage with motivation of why this is needed,
on windows it popup a almost normal message window.

wx is kind of bloated as it is so I don't want add every single class in wxWidgets just because we can.

dominicletz:
Ah crazy mistake, fixed that and squashed the commits. Added description.

dgud:
I verified the commit by re-generating the data, and I want you to include all the changes that happened when generating the
files otherwise I will be confused later.

I was confused now, when functions was moved around in the wrapper files, I wonder if we should change when to split
it into different files to minimize changes, any ideas?

dominicletz:
Yeah I wanted to keep the diff minimal so I only committed the new entries:
A) Generate only a single huge wrapper file, sorted by function name
B) Keep number of wrappers files but split defs alphanumerically not by count, keeping the same distribution while adding new modules.
C) Increase number of wrapper files and create one per wx module.

I can give B a try and update this pull request correspondingly...

dgud:
We had one file previously, split it to utilize parallel compilation.

dominicletz:
So I've done B) and pushed on top -- The grouping is in `wx_gen_nif.erl` and is first sorting and then statically assigning each class to a specific file based on the first Character after "wx" if there is a "wx" prefix. I've set it up so that ATM the wrapper files are roughly similar size, but the main benefit is that from now on adding or removing a module won't move other modules around anymore.

```erlang
Sorted = lists:sort(fun(#class{name=A}, #class{name=B}) -> A < B end, Defs),
split(Sorted, "ab;cd;efg;hijkl;mnopqr;s;tuv;wxyz"),
```

PR #2979, opened 2021-01-16
ets: use initial salt different from phash2

max-au:
It is convenient to use erlang:phash2 to determine frag (ETS table)
to use. If salt is the same, both phash2 and ets hashing generate
the same frag/bucket number.
Since ETS hash is not portable, there is no need to keep backwards
compatibility with previous releases.

### DIFF
--- a/erts/emulator/beam/utils.c
+++ b/erts/emulator/beam/utils.c
@@ -2061,7 +2061,7 @@ trapping_make_hash2(Eterm term, Eterm* state_mref_write_back, Process* p)
  * One IMPORTANT property must hold (for hamt).
  * EVERY BIT of the term that is significant for equality (see EQ)
  * MUST BE USED AS INPUT FOR THE HASH. Two different terms must always have a
- * chance of hashing different when salted: hash([Salt|A]) vs hash([Salt|B]).
+ * chance of hashing different when salted.
  *
  * This is why we cannot use cached hash values for atoms for example.
  *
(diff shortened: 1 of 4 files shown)


sverker:
make_internal_hash() used by ETS is actually a separate implementation from phash2. It was created as a copy-paste of phash2 and then tweaked, but it still returns same hash for some types. Maybe a better solution would be to tweak make_internal_hash some more to diverge further from phash2.

max-au:
I assumed having a different initial salt would lead to the same outcome, and attempted to avoid larger rewrites. However might be wrong here, e.g. there could be more hash clashes possible. One I can think about is atom hashing. 
Would it make sense to start with documenting how existing hashing works (unless it's already documented somewhere)?

max-au:
It might be challenging to make a larger change on how make_internal_hash works, would it be possible to review this patch as-is, to have a chance of getting it into RC?

sverker:
Sorry for being a bit slow on this.
A simple thing to do is to move the salt from ETS into make_internal_hash, like
`Uint32 hash = salt ^ 33554467UL;
`
By the way, is there a reason you chose a salt only 25 bits long?

max-au:
You are indeed right, I completely missed make_internal_hash definition. Thank you for suggestion!

25-bit salt is a historical choice, based on 32-bit hash, and 6 bits (assuming, wrongly) used by tagging schema. It's a prime closest to the middle of remaining 26 bits. Feel free to adjust it to any other number (prime, ideally).

bjorng:
The test cases `map_SUITE:t_hashmap_balance/1` and `persistent_term_SUITE:collisions/1` fail.

sverker:
I will look at that.

max-au:
Thank you! Indeed it was my fault, I did not realise these tests had hardcoded has values.

PR #2981, opened 2021-01-18
jit: Output size of emitted code in assembly dumps

frej:
With this patch, a debug compiled emulator will output a line with
"; CODE_SIZE: <size in bytes>" at the end of the assembler dump files.


frej:
Pushed a stale version.

PR #2983, opened 2021-01-18
io_lib_format:fwrite_g/1 benchmarks

DianaOlympos:
These benchmarks are added to allow a baseline and a benchmarking of #2960

The small digit test should allows to show the discrepancy between ryu and the dragon4 algorithm for "small number of digits in decimal form" doubles. It may be worth to reduce the cutoff to less digits.

### DIFF
--- a/lib/stdlib/test/stdlib_bench_SUITE.erl
+++ b/lib/stdlib/test/stdlib_bench_SUITE.erl
@@ -29,7 +29,7 @@ suite() -> [{ct_hooks,[{ts_install_cth,[{nodenames,2}]}]}].
 
 
 all() ->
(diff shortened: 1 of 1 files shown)


garazdawi:
io_lib_format_benchmark

The benchmark seems to be working. Nice and stable results. This is without your changes, so the improvement that happens in master is the JIT doing its magic.

If you are happy with this I'll merge it.

DianaOlympos:
let's go, i should probably add a "lot of digits" example" but i think the random is enough for now. (that is because i am pretty sure the current one is O(n) or O(logn) number of digits.

LGTM

PR #2986, opened 2021-01-19
More security recommendations in binary_to_term/2 docs

josevalim:
Follow up on #2886.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -614,8 +614,11 @@ hello
 </pre>
         <warning>
           <p>When decoding binaries from untrusted sources,
-            consider using <c>binary_to_term/2</c> to prevent Denial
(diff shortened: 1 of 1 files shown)


erts/doc/src/erlang.xml
```diff
@@ -652,6 +655,15 @@ hello
...
 hello
 </pre>
+	    <warning>
+	      <p>The <c>safe</c> option ensures the data is safely
+                processed by the Erlang runtime but it does not
+                guarantee the data is safe to your application.
+                You must always validate data from untrusted sources.
+                If the binary is stored or transits through untrusted
+                sources, you should also consider cryptographically
+                signing and encrypting it.</p>
```
voltone, line 665:
```suggestion
                signing it.</p>
```

Signing would be enough, encryption is orthogonal here

PR #2987, opened 2021-01-19
Add TravisCI job to build and run smoke tests on ARM64

martin-g:
More and more development and deployment is being done on ARM64 hardware.
It would be good if the build and the smoke tests are regularly run on TravisCI ARM64 nodes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ addons:
 matrix:
   include:
     # Doc build second as it also takes a long time to run
-    - env: Linux64Docbuild
+    - name: Linux64Docbuild
       script:
         - ./scripts/build-otp docs
       deploy:
@@ -41,7 +41,7 @@ matrix:
           tags: false
           condition: $TRAVIS_PULL_REQUEST = "false"
           repo: erlang/otp
-    - env: Linux-ppc64le-SmokeTest
+    - name: Linux-ppc64le-SmokeTest
       os: linux-ppc64le
       script:
         - ./scripts/build-otp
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,7 +24,7 @@ addons:
 matrix:
   include:
     # Doc build second as it also takes a long time to run
-    - env: Linux64Docbuild
```
martin-g, line 27: I guess the usage of `env` here was a typo.
If I am wrong then please let me know and I will revert this change!

garazdawi:
We have moved away from travis to github actions, so I'm a bit hesitant to include this. We test x86, arm and ppc internally so this would only add extra checks for PRs for early detection.

I've kept the doc build and ppcle as they existed before the merge and I've not yet taken the time to port the doc build so that http://blog.erlang.org/cd/ continues to work.

I suppose the way forward if you want to help add arm support is to cross-compile to it and then use qemu to run the smoke tests.

martin-g:
Here is an attempt to add aarch64 job to the matrix: https://github.com/erlang/otp/compare/master...martin-g:feature/build-and-test-on-arm64?expand=1

But at the moment it fails with: https://github.com/martin-g/otp/runs/1741861402?check_suite_focus=true

I will try to debug and finish it later.
If you have any hints I'll be very thankful!

garazdawi:
I've never done anything like that before so not sure I can be of much help.

However, if you rebase on latest master you can see that I have updated some of the github actions scripts to use the new and improved https://github.com/marketplace/actions/build-and-push-docker-images v2 that uses buildx. In that github action there are descriptions about how to do multiarch builds, unfortunately you can no longer use docker.pkg.github.com to cache the base image, so you would have to build it from scratch everytime. Not a huge issue, but it does add about 3-5 minutes for each type of build.

martin-g:
At the moment I use `arm64v8/debian` instead of docker.pkg.github.com/erlang/otp/debian-base-aarch64 because it seems it needs this iamge to exist to be able to build the base image.

Otherwise the idea in my changes is to enable QEMU for Docker so that one can run a Docker image that is build for a foreign architecture (like arm64v8/*** ones) on x86_64 machines (like GitHub Actions nodes).
The rest is the same - reuse Dockerfile.debian-base by just passing different BASE and Dockerfile.aarch64 is just a copy of Dockerbase.64-bit

I'll take a look at the buildx way!

garazdawi:
I'm closing this PR for now, feel free to re-open or create a new one if you get something working.

PR #2988, opened 2021-01-19
Implement handling of early data in ssl server

peterdmv:
This draft implements the early data feature on the server side. It is not yet feature complete or fully tested.
Current status:

**API**
- [ ]  Option to select 1-RTT or 2-RTT when unwanted early_data received (current default is 1-RTT)
- [x] Delivery of early data to user: delivered as normal application data
- [ ]  Debug Logging of early_data 
- [x] NSS Keylog: store client_early_data_traffic_secret

**State Machine**
- [x] Update state ```start```: handle received early_data extension based on option early_data and set early_data_accepted. Set 0-RTT read traffic keys before transitioning to ```negotiated```.
- [x] Implement state transition from ```negotiated``` to ```wait_eoed```.
- [x] ~~Implement state transition from ```negotiated``` to ```wait_cert```.~~
- [x] Update state ```negotiated```: send early_data_indication if early data is accepted. 
- [x] Implement new state ```wait_eoed```.

**Record Protocol**
- [x] Trial decryption (early_data rejected)
        If the client attempts a 0-RTT handshake but the server
	rejects it, the server will generally not have the 0-RTT record
	protection keys and must instead use trial decryption (either with
	the 1-RTT handshake keys or by looking for a cleartext ClientHello in
	the case of a HelloRetryRequest) to find the first non-0-RTT message.
- [x] Error handling (early_data accepted)
        Specifically, if the server fails to decrypt a 0-RTT record following
        an accepted "early_data" extension, it MUST terminate the connection with a
	"bad_record_mac" alert as per Section 5.2.
- [x] Fallback to 1-RTT (server early_data is not enabled)
       Ignore the extension and return a regular 1-RTT response.  The
        server then skips past early data by attempting to deprotect
        received records using the handshake traffic key, discarding
        records which fail deprotection (up to the configured
        max_early_data_size).  Once a record is deprotected successfully,
        it is treated as the start of the client's second flight and the
        server proceeds as with an ordinary 1-RTT handshake.
- [x] Handle too much early data
- [ ] Fallback to 2-RTT (server early_data is not enabled) - Future improvement?
      Request that the client send another ClientHello by responding
      with a HelloRetryRequest.  A client MUST NOT include the
      "early_data" extension in its followup ClientHello.  The server
      then ignores early data by skipping all records with an external
      content type of "application_data" (indicating that they are
      encrypted), up to the configured max_early_data_size.
- [x] Process early data
      Return its own "early_data" extension in EncryptedExtensions,
      indicating that it intends to process the early data.  It is not
      possible for the server to accept only a subset of the early data
      messages.  Even though the server sends a message accepting early
      data, the actual early data itself may already be in flight by the
      time the server generates this message.
- [ ] Check early_data requirements
      In order to accept early data, the server MUST have accepted a PSK
      cipher suite and selected the first key offered in the client's
      "pre_shared_key" extension.  In addition, it MUST verify that the
      following values are the same as those associated with the
      selected PSK:
      - The TLS version number  ✅
      - The selected cipher suite  ✅
      - The selected ALPN [RFC7301] protocol, if any 

**Other**
- [x] Update session tickets
- [x] FT - erlang client - erlang server
- [x] FT -  openssl s_client - erlang_server
- [x] Statement of Compliance
- [x] Users Guide


lib/ssl/doc/src/using_ssl.xml
```diff
@@ -579,21 +579,37 @@ ok
...
+        application:load(ssl),
+        {ok, _} = application:ensure_all_started(ssl),
+        Port = 11029,
+        LOpts = [{certfile, ?SERVER_CERT},
+             {keyfile, ?SERVER_KEY},
+             {reuseaddr, true},
+             {versions, ['tlsv1.2','tlsv1.3']},
+             {session_tickets, stateless},
+             {early_data, enabled},
+             {keep_secrets, true} %% Enable NSS key log (optional)
```
IngelaAndin, line 598: Maybe we should state it is a debug option !?

peterdmv: Fixed

PR #2989, opened 2021-01-20
inet_tcp, inet6_tcp: do not bind connecting sockets by default

max-au:
There is no need to bind a socket intended to be used for outgoing
connection. This operation exhaust ephemeral port range, preventing
large number of concurrent outgoing connections.
Omitting bind call also saves an extra trip to kernel.

There is no reliable way to test this, as different platforms have
various limitations on number of file descriptors open, ephemeral
port count and socket allocation mechanisms.

Manual testing can be done using this code:

  verify(0) ->
      ok;
  verify(Remain) ->
      receive
          {result, {error, Good}} when Good =:= timeout; Good =:= econnrefused ->
              verify(Remain - 1);
          {result, Any} ->
              Any
      end.

  test_conn() ->
      Self = self(),
      Expect = 66000, %% this number is guaranteed to be larger than ephemeral port count
      [spawn(
          fun() ->
              Res = catch (gen_tcp:connect(
                  {10, 20, (Seq div 254 + 1) rem 254, Seq rem 254 + 1}, 12345, [], 5000)),
              Self ! {result, Res}
          end) || Seq <- lists:seq(1, Expect)],
      verify(Expect).

To run this code, system must be configured for high file descriptor limit
(e.g."ulimit -n 100000").

### DIFF
--- a/lib/kernel/src/inet6_tcp.erl
+++ b/lib/kernel/src/inet6_tcp.erl
@@ -120,12 +120,12 @@ do_connect(Addr = {A,B,C,D,E,F,G,H}, Port, Opts, Time)
 	{ok,
 	 #connect_opts{
 	    fd = Fd,
-	    ifaddr = BAddr = {Ab,Bb,Cb,Db,Eb,Fb,Gb,Hb},
+	    ifaddr = BAddr,
 	    port = BPort,
 	    opts = SockOpts}}
-	when ?ip6(Ab,Bb,Cb,Db,Eb,Fb,Gb,Hb), ?port(BPort) ->
+	when ?port(BPort) ->
 	    case inet:open(
-		   Fd, BAddr, BPort, SockOpts,
+		   Fd, check_ip_format(BAddr), BPort, SockOpts,
 		   ?PROTO, ?FAMILY, ?TYPE, ?MODULE) of
 		{ok, S} ->
 		    case prim_inet:connect(S, Addr, Port, Time) of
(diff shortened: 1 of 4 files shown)


vans163:
Using the new socket api avoids this problem (since you have more control).  Is gen_tcp not going to be deprecated / put ontop of it by default?

But nice.

max-au:
To my knowledge, upcoming release still has original gen_tcp backend as default. Also, I am not sure whether distribution over TLS will use "socket" instead of gen_tcp, or will use "socket" backend for gen_tcp. Therefore I assume this change is still helpful.

max-au:
@RaimoNiskanen is there any chance to review the change?

RaimoNiskanen:
Yes, sorry I have had a hard time finding the time...  I will look at it now, and run it in our daily tests.

RaimoNiskanen:
Sweet!  I did not realize how little there was left to implement this.  Really minimal patch!
We will see if our daily builds show anything, but I doubt that...

RaimoNiskanen:
You missed a spot: the corresponding change in `local_tcp`, found by our daily tests.
I pushed a correcting commit.

max-au:
Thank you! I missed this, distracted by the amount of IPv4 tests failed. In our environment, IPv4 is not fully supported, and pretty much all IPv4 tests tend to fail. I might need to make a PR that would return "skip" instead of failing these tests.

PR #2991, opened 2021-01-20
Allow h4,h5 and h6 headers in application/erlang+html EEP-48 documentation

garazdawi:
This is needed as edoc (implemented in #2803) can emit such headers.


garazdawi:
Added `b` and `strong` to the allowed tags and also made `em` behave as it should (i.e. render as `i`).

paulo-ferraz-oliveira:
Hurrah. I had this in my TODO list. Glad you tackled it.

PR #2992, opened 2021-01-20
No record construction error cascade

mheiber:
Add an option `--tolerant` to stop a cascade of `no local return` errors on bad record constructions.

This is a really rough work in progress. 

TODO:

- [ ] in dialyzer_dataflow.erl, figure out whether we need to union the actual type and specced type of the record construction. When does it make a difference? Should this be controlled by the flag? 
- [  ] better naming than `tolerant`
- [ ] tests
- [ ] stats from some open-source codebase or get permission to share WA stats.


mheiber:
opened accidentally, sorry!!!

PR #2995, opened 2021-01-22
compiler: Warn about matching underscore-prefixed variables

jhogberg:
[An experiment](https://github.com/richcarl/otp/commits/pinning-otp) @richcarl ran for [EEP 55](https://github.com/erlang/eep/blob/master/eeps/eep-0055.md) pointed out a few errors in OTP where underscore-prefixed variables were matched when they clearly weren't meant to be. Underscore-prefixed variables merely suppress "unused variable" warnings but are often misunderstood and thought to behave like `_` which can introduce subtle bugs. For example:

```erlang
case X of
    {ok, Foo} ->
        do_something(Foo);
    _Error ->
        case Y of
            {ok, Bar} ->
                do_something_else(Bar);
            _Error ->
                 %% case_clause exception if _Error is not the same as in the outer case!
                 error
        end
end.
```

This PR fixes these bugs and introduces a warning for similar misuse.


lib/stdlib/src/erl_lint.erl
```diff
@@ -317,6 +317,9 @@ format_error({unsafe_var,V,{What,Where}}) ->
 format_error({exported_var,V,{What,Where}}) ->
     io_lib:format("variable ~w exported from ~w ~s",
                   [V,What,format_where(Where)]);
+format_error({match_underscore_var, V}) ->
+    io_lib:format("variable ~w is matched. Use '_' if you meant to ignore it",
```
richcarl, line 321: I suggest "variable ~w is already bound. If you meant to ignore this value, use a different underscore-prefixed name or just '_'."

jhogberg: Thanks!

bjorng (approved):

mikpe:
I tested this PR on a large, mature, code base, and it did find a (very) small number of errors in the production code, and a fair amount of poorly written tests that accidentally match already-bound underscore-prefixed variables.

mikpe (approved):

richcarl (approved):

lib/stdlib/test/erl_lint_SUITE.erl
```diff
@@ -4477,6 +4478,27 @@ warn_missing_spec(Config) ->
                         {11, erl_lint, {missing_spec, {internal_no_spec, 0}}}]}}
     ]).
 
+underscore_match(Config) when is_list(Config) ->
```
garazdawi, line 4481: Add test for: `foo(_A, _A) -> _A`?

paulo-ferraz-oliveira:
I'm sure I'll find one or two in our lib.s. There are developers that just don't want to change the code that much (understanding fully that they're matching and not suppressing warnings) and actually apply the `case`-like example you posted.

Will there be an option to silence the warnings or will this be a breaking change? (if there's no such option and we're compiling with `warnings_as_errors` we're suddenly forced to change our previously running code)

lib/compiler/doc/src/compile.xml
```diff
@@ -720,6 +720,13 @@ module.beam: module.erl \
 	    Default is to emit no warnings for unused imported functions.</p>
           </item>
 
+          <tag><c>nowarn_underscore_match</c></tag>
+          <item>
+            <p>By default, warnings are emitted when a variable that begins
+              with an underscore is matched after being bound. Use this
+              option to turn off this kind of warnings.</p>
```
paulo-ferraz-oliveira, line 727:
```suggestion
              option to turn off this kind of warning.</p>
```

lib/stdlib/src/erl_lint.erl
```diff
@@ -317,6 +317,10 @@ format_error({unsafe_var,V,{What,Where}}) ->
 format_error({exported_var,V,{What,Where}}) ->
     io_lib:format("variable ~w exported from ~w ~s",
                   [V,What,format_where(Where)]);
+format_error({match_underscore_var, V}) ->
+    io_lib:format("variable ~w is already bound. If you meant to ignore this "
```
paulo-ferraz-oliveira, line 321: "meant" should be "mean", to be in agreement with "use".

Also, I'm not sure about "just", but it doesn't roll of my tongue. I'd use "simply" (or perhaps "a single").

paulo-ferraz-oliveira:
Ah, this is on top of the version that already requires [@uabboli's `parse_trans` pull request :-)](https://github.com/uwiger/parse_trans/pull/42), but on top of `lager`, otherwise I'm getting `error in parse transform 'lager_transform': {badarg,[{erlang,integer_to_list,[{56,9}]`.

uabboli:
> lager, otherwise I'm getting error in parse transform 'lager_transform': {badarg,[{erlang,integer_to_list,[{56,9}].

@paulo-ferraz-oliveira I've updated a few parse transforms on Github, but didn't notice that lager was indeed affected. I'd be great if you fix the issues, otherwise I can have a look as soon as time permits.

paulo-ferraz-oliveira:
> > lager, otherwise I'm getting error in parse transform 'lager_transform': {badarg,[{erlang,integer_to_list,[{56,9}].
> 
> @paulo-ferraz-oliveira I've updated a few parse transforms on Github, but didn't notice that lager was indeed affected. I'd be great if you fix the issues, otherwise I can have a look as soon as time permits.

I only found one issue, pull requested to fix it and it was merged. I asked about my solution being complete, but got no feedback. I'll pull request again if I find other issues there.

PR #3000, opened 2021-01-25
beam_ssa_opt: Don't try to skip 'fcheckerror'

jhogberg:
https://bugs.erlang.org/browse/ERL-1471


bjorng:
Unfortunately, looking closer at the generated code, the optimization to avoid flushing out the float registers turns out to be essential. It is true that the loader will remove the `fclearerror` and `fcheckerror` instructions, but the loading and storing back of float registers will happen for **every** floating point instruction (because each floating point instruction are in separate blocks to begin with). We will have to figure out a different way to fix the bug.

PR #3002, opened 2021-01-25
shell: read records from module in escript/archive

gomoripeti:
Previously the `rr/1` shell function failed with the below error when
called for a module which was loaded from an archive or escript archive,
because `beam_lib:chunks` uses regular `file:open/2` to read from the
file. Now `erl_prim_loader:get_file/1` is used which supports file paths
in archives.

```
{error, beam_lib, {file_error,"<path_to_archive>/test_app.ez/test_app/ebin/test.beam", enotdir}}
```

### DIFF
--- a/lib/stdlib/test/shell_SUITE.erl
+++ b/lib/stdlib/test/shell_SUITE.erl
@@ -64,8 +64,9 @@ end_per_testcase(_Case, Config) ->
     OrigPath = proplists:get_value(orig_path,Config),
     code:set_path(OrigPath),
     application:unset_env(stdlib, restricted_shell),
-    (catch code:purge(user_default)),
-    (catch code:delete(user_default)),
+    purge_and_delete(user_default),
+    %% used by `records' test case
+    purge_and_delete(test),
     ok.
 -endif.
 
@@ -298,8 +299,7 @@ restricted_local(Config) when is_list(Config) ->
 	comm_err(<<"begin shell:stop_restricted() end.">>),
     undefined =
 	application:get_env(stdlib, restricted_shell),
-    (catch code:purge(user_default)),
-    true = (catch code:delete(user_default)),
+    true = purge_and_delete(user_default),
     ok.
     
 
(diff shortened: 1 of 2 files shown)


ranjanified:
Is this [warning](https://erlang.org/doc/man/code.html#loading-of-code-from-archive-files) still relevant, and if so, will it not have any impact here?

garazdawi:
> Is this [warning](https://erlang.org/doc/man/code.html#loading-of-code-from-archive-files) still relevant, and if so, will it not have any impact here?

I don't think there has ever been any formal decision about making archives non-experimental, but I think that the functionality has been used in enough places (especially escript archives) that it will now be impossible to remove. I'll check and see if we can remove that warning.

josevalim:
> I think that the functionality has been used in enough places (especially escript archives) that it will now be impossible to remove

To provide some data points, Elixir unzips .ez archives on install, so we don't need runtime support for .ez, but we do use escripts archives extensively. rebar3 is also distributed as an escript archive, right?

garazdawi:
> rebar3 is also distributed as an escript archive, right?

yes, as is erlang_ls and I'm sure many more. If only we could solve the issue with embedding `.so`/`.dll` files...

essen:
RabbitMQ uses archives for its plugins as well. It would be great to solve the `.so` problem, but in the meantime I believe we just unzip (or don't zip) those. Considering how widely used they are it would be a good idea to make archives non-experimental and to fix the few remaining edge cases with them. I can probably help with that at some point.

oskardrums:
Hey guys, just a bystander here intrigued by the discussion.
If I understand correctly, the `.so`/`.dll` issue is that you can't use `dlopen` and the likes on a memory buffer containing the native code because the syscall takes a path on the disk... Did I get that right?
That's a problem that raises with Python archives as well, [py2exe](https://github.com/py2exe/py2exe/blob/master/source/MyLoadLibrary.c) handles this pretty successfully for Windows, might be worth looking into. Not sure how they fare with the unix case though.
For recent-enough linux systems `memfd_create` can be used to get a cheap memory backed fd, which can be `dlopen`ed via `/proc/self/fd/%d`. Hackish, but should work.

garazdawi:
> If I understand correctly, the `.so`/`.dll` issue is that you can't use `dlopen` and the likes on a memory buffer containing the native code because the syscall takes a path on the disk... Did I get that right?

Yes, this is the problem. I created a ticket for it [here](https://bugs.erlang.org/browse/ERL-1273) so that we do not continue to highjack this PR.

lib/stdlib/test/shell_SUITE.erl
```diff
@@ -428,6 +428,31 @@ records(Config) when is_list(Config) ->
...
+
+    ok = file:write_file(Test, Contents),
+    {ok, test} = compile:file(Test, [{outdir, AppEbinDir}]),
+
+    Ext = init:archive_extension(),
+    Archive = filename:join(PrivDir, "test_app" ++ Ext),
+    {ok, _} = zip:create(Archive, ["test_app"], [{compress, []}, {cwd, PrivDir}]),
+
+    ArchiveEbinDir = filename:join(Archive, "test_app/ebin"),
+    true = code:add_path(ArchiveEbinDir),
```
garazdawi, line 449: The code path is a global state that needs to be reset when the testcase finishes. Otherwise it will interact and possibly fail other testcases. I noticed this as the shell_docs_SUITE testcases started to fail. I suggest that you save the path in init_per_suite and reset it again in end_per_suite.

gomoripeti: ah, thanks, will do

josevalim:
@garazdawi just one more comment about the whole archives/escripts note. If the goal is to mark them as officially supported, then perhaps we should *not* merge this PR and instead make it so `file:read` always looks inside archives? Or is there a high cost associated to this that makes it undesired?

Otherwise I assume `rr/1` is not the only function that will fail for archives/escripts. What if a project needs to load other files inside escripts/archives? Such as the doc/*.chunk files or files from priv in general?

garazdawi:
> @garazdawi just one more comment about the whole archives/escripts note. If the goal is to mark them as officially supported, then perhaps we should not merge this PR

I think that we should still merge this as fixing archives will be a lot of work and will not be done before 24. It will be easy to revert this when/if we implement more support for archives and then we will have a testcase that should work no matter what the implementation is.

> instead make it so file:read always looks inside archives? Or is there a high cost associated to this that makes it undesired?

It would mean that we have to do a `string:find/2` for each `file:open`/`file:read_file`/`file:read_file_info`, which I don't think should be noticeable in relation to the actual file operations...

gomoripeti:
I did not dare to modify more serious modules like `code` or `beam_lib` fearing unexpected change of behaviour elsewhere. That's why I chose the most shallow and harmless place to quick fix my particular issue. But I agree with Jose that archives/escripts support could be tackled on a much more generic way. Maybe not pure `file:read` but with an `include_archives` option to show intent.

garazdawi:
Thanks!

PR #3003, opened 2021-01-25
Reconnect to epmd

dotsimon:
Currently when the connection from a distributed node to epmd is lost, there is no attempt by the node to reconnect.
This leaves the node invisible to other nodes in the distribution that it is not already connected to.
This PR adds simple reconnect functionality.
If connecting to an epmd without big creation numbers, keep reconnecting until the original creation is received in order to avoid a later invocation of the same node name getting the same creation.

### DIFF
--- a/erts/doc/src/erl_dist_protocol.xml
+++ b/erts/doc/src/erl_dist_protocol.xml
@@ -430,9 +430,6 @@ io:format("old/unused name ~ts at port ~p, fd = ~p ~n",
 
       <p>where n = <c>Length</c> - 1.</p>
 
-      <p>The current implementation of Erlang does not care if the connection
-        to the EPMD is broken.</p>
-
       <p>The response for a <c>STOP_REQ</c> is as follows:</p>
 
       <table align="left">
--- a/lib/kernel/doc/src/erl_epmd.xml
+++ b/lib/kernel/doc/src/erl_epmd.xml
@@ -56,8 +56,10 @@
       <desc>
         <p>Registers the node with <c>epmd</c> and tells epmd what port will be
         used for the current node. It returns a creation number. This number is
-        incremented on each register to help with identifying if a node is
-        reconnecting to epmd.</p>
+        incremented on each register to help differentiate a new node instance
+        connecting to epmd with the same name.</p>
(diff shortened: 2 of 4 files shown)


garazdawi:
What kind of scenario is it when Erlang loses the connection to EPMD? Is epmd running on a different host?

dotsimon:
The epmd could crash for example, or be restarted by systemd.
I could not think of any good reason for Eralng **not** to retry.

garazdawi:
Given that EPMD since OTP-23 always returns > 3 creations, I'm not seeing how reconnecting would help. Unless you are running OTP-23+ with an old EPMD, which seems like an odd combination?

dotsimon:
At least in my development environment there can be multiple OTP installations and consequently it could be an older EPMD running.
I admit this isn't a normal scenario but OTOH I could also see the case in a production environment during OTP upgrade if there are multiple nodes on the same machine that are being migrated one at time.
If you consider this too esoteric then I am happy to rewrite the patch for just reconnecting and without regard to the creation ID.

garazdawi:
I will need to think and check a bit to make sure that doing this reconnect does not break some invariant somewhere...

> If you consider this too esoteric then I am happy to rewrite the patch for just reconnecting and without regard to the creation ID.

Yes, I don't think it makes sense to keep that code. Especially since it will be very hard to test that it works in regression tests.

garazdawi:
> I will need to think and check a bit to make sure that doing this reconnect does not break some invariant somewhere...

Talked with some colleagues and they don't see any problem with this, so we are good to go. What is missing is a testcase and documentation. For the documentation I suggest adding a note in http://erlang.org/doc/man/erl_epmd.html about the reconnect.

Writing a testcase may be a little bit harder, but since all tests are guaranteed to run as distributed nodes, maybe it is enough to just add a test that stops the current epmd and then starts it again and make sure that erl_epmd:names() still works as expected? I would put the testcase in lib/kernel/test/erl_distribution_SUITE.erl.

dotsimon:
Added doc & test as requested. Thanks for the review!

garazdawi:
Great! That test case looks much better than what I suggested. Putting it into testing!

garazdawi:
Thanks!

PR #3005, opened 2021-01-26
jinterface: make OtpErlangExternalFun transparent

ilya-klyuchnikov:
Not being able to access the internals of OtpErlangExternalFun is problematic.
The real use case: performing some analysis of Core Erlang (cerl) from inside java/scala - the construct `fun module:fun/arity` is represented as "itself", and when accessing cerl forms from jinterface - it's represented as OtpErlangExternalFun correspondingly.  Unfortunately, - the internals are private, - which makes some forms of Erlang core not analysable through jinterface.

Also, - since all `module`, `function` and `arity` fields are final/immutable - it doesn't make much sense to hide/incapsulate them.


garazdawi:
Thanks!

PR #3006, opened 2021-01-27
Make compiler report column as well as line number

uabboli:
This pull request is a continuation of https://github.com/erlang/otp/pull/2664 by Richard Carlsson.

We've made the following additions to Richard's branch:

A function PT:parse_transform_info/0 is added. It can be used by parse
transforms to signal that they can only handle line numbers in
abstract code.

A compiler option {error_location, line | column} is added. The option
also determines if column numbers are included in abstract code (no
separate option for that). We've chosen `column' as default, and are
aware that this can affect tools and parse transforms in an
incompatible way. To facilitate the transition we've created a few
backwards compatible pull request on Github. If tools stop working,
setting the environment variable ERL_COMPILER_OPTIONS can help
(include "{error_location, line}").

The new compiler option is recognized by Yecc.

Column numbers are also displayed in the Erlang Shell, and the
Dialyzer uses them to pinpoint for instance arguments more accurately.

We have currently no plans to include column numbers in stacktraces.

We currently see no need to add any predefined macros.


lib/compiler/src/compile.erl
```diff
@@ -1148,6 +1173,27 @@ foldl_transform([T|Ts], Code0, St) ->
     end;
 foldl_transform([], Code, St) -> {ok,Code,St}.
 
+%%% It is possible--although unlikely--that parse transforms add
+%%% columns to the abstract code, why this function is called for
+%%% every parse transform.
+maybe_strip_columns(Code, T) ->
+    case erlang:function_exported(T, parse_transform_info, 0) of
```
michalmuskala, line 1180: I'm afraid this check might not be successful since in many cases parse transforms would be lazily loaded, and `function_exported` returns `false` for unloaded modules

PR #3007, opened 2021-01-27
compiler: Stop emitting the fclearerror and fcheckerror instructions

bjorng:
The `fclearerror` and `fcheckerror` instructions have been ignored by the
BEAM loader for many releases. Therefore, update the compiler to stop
emitting them. This also eliminates the bug reported in ERL-1471.

While at it, take the opportunity to clean up and simplify the
floating point optimizations.

(See #3004 for a fix of the same bug for OTP 23.)


lib/compiler/src/beam_validator.erl
```diff
@@ -967,31 +961,10 @@ vi({bs_set_position, Ctx, Pos}, Vst0) ->
...
-    end,
-    set_fls(cleared, Vst);
-vi({fcheckerror,_}, Vst0) ->
-    assert_fls(cleared, Vst0),
-    Vst = set_fls(checked, Vst0),
-    branch(?EXCEPTION_LABEL, Vst);
+vi({fconv,Src,{fr,_}=Dst}, Vst0) ->
+    assert_term(Src, Vst0),
+    Vst = update_type(fun meet/2, number, Src, Vst0),
+    set_freg(Dst, Vst);
```
jhogberg, line 967: Instructions that throw exceptions must branch to `?EXCEPTION_LABEL` to validate the stack and catch state. `update_type/4` also raises type conflict errors when the validator knows the types are incompatible (`none`), causing validation to fail unless used within a branch handler.

```suggestion
vi({fconv,Src,{fr,_}=Dst}, Vst) ->
    assert_term(Src, Vst),
    branch(?EXCEPTION_LABEL, Vst,
           fun(SuccVst0) ->
                   SuccVst = update_type(fun meet/2, number, Src, SuccVst0),
                   set_freg(Dst, SuccVst)
           end);
```

PR #3010, opened 2021-01-28
syntax_tools: Allow merl quote functions to accept an erl_anno:anno()

maxno-kivra:
Merl currently assumes all annotations are locations, which can be
either an integer or a two-tuple of integers. However, erl_anno
supports more fields, like file and text. This is done through an
opaque type, which is actually a proplist.

OTP adds support for column numbers, and it would be nice if
merl would support those somewhere down the line. A first step
would be to allow the quote function(s) to accept the full
erl_anno:anno() type. It currently pulls the location and then
continues as normal.

I also renamed Line to Location in merl_transform to make it clear
that it can be both a integer and tuple of integers. Later perhaps
a full anno, but not now..

---

I'm not sure what to call the test, but I noticed `otp_15291`, is that named after some issue number? Should I change the name to this PR's number?

### DIFF
--- a/lib/syntax_tools/src/merl.erl
+++ b/lib/syntax_tools/src/merl.erl
@@ -310,7 +310,7 @@
 %% while a flat string or binary represents source code containing newlines.
 -type text() :: string() | binary() | [string()] | [binary()].
 
--type location() :: erl_anno:location().
+-type location() :: erl_anno:location() | erl_anno:anno().
 
 
 %% ------------------------------------------------------------------------
@@ -324,7 +324,7 @@ compile(Code) ->
 %% into a binary BEAM object.
 %% @see compile_and_load/2
 %% @see compile/1
-compile(Code, Options) when not is_list(Code)->
+compile(Code, Options) when not is_list(Code) ->
     case type(Code) of
         form_list -> compile(erl_syntax:form_list_elements(Code));
         _ -> compile([Code], Options)
@@ -471,11 +471,11 @@ quote(Text) ->
 %%
 %% @see quote/1
 
-quote({Line, Col}, Text)
-  when is_integer(Line), is_integer(Col) ->
-    quote_1(Line, Col, Text);
-quote(StartPos, Text) when is_integer(StartPos) ->
-    quote_1(StartPos, undefined, Text).
+quote(StartPos, Text) ->
+    Anno = erl_anno:from_term(StartPos),
+    StartLine = erl_anno:line(Anno),
+    StartCol = erl_anno:column(Anno),
+    quote_1(StartLine, StartCol, Text).
 
 quote_1(StartLine, StartCol, Text) ->
     %% be backwards compatible as far as R12, ignoring any starting column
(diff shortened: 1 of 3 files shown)


garazdawi:
> I'm not sure what to call the test, but I noticed otp_15291, is that named after some issue number? Should I change the name to this PR's number?

You can leave it as is. Using the ticket numbers is an old convention that we have moved away from.

ranjanified:
Yes please. anno_test is much more meaningful than otp_1234****. :-) Names
are meant for us, to help each other. Otherwise it will be a journey to go
find out what is otp_123**** or will have to read through the test
implementation to infer anything.

maxno-kivra:
I had a bit of trouble getting dialyzer to run, and a bit of trouble convincing it that the code's fine. But switch to `erl_anno:from_term` fixes that. Let's see if the tests passes. Otherwise I might need some help.

maxno-kivra:
The tests, and dialyzer, passed. Hurray! I think this means it's ready.

uabboli:
I don't quite understand what you're trying to do.

merl:quote() is documented to accept a location() as first argument.
Why are you trying to change that? It doesn't make sense to me.

Note that erl_anno:location() is not an opaque type; you're allowed to
check if it's a tuple, an integer, &c.

Perhaps I'll understand better if you have an example. The test
anno_test() fails on master, but that is what it's supposed to do.

(If you choose continue this PR, you should rebase on master.)

michalmuskala:
> Note that erl_anno:location() is not an opaque type; you're allowed to
check if it's a tuple, an integer, &c.

I'm not sure this is accurate. The dialyzer type is defined with `-opaque` and the [documentation](https://erlang.org/doc/man/erl_anno.html#type-anno) doesn't specify the structure of the type.

maxno-kivra:
I've rebased on master now. The conflict was from #3006 and @richcarl work on using erl_anno opaquely. Here is the commit if you're curious https://github.com/erlang/otp/commit/f30514e3232f63563251a42334912aef8c79d4e8#diff-f2e48eb7504acfca64629944afb87c972e5430f81f047d1c5fff0df1b7651edaR1251-R1254.

Like that commit mentions, what used to be a line number is now a general storage of annotations. Like @michaelklishin pointed out, and the reason I first got dialyzer errors, is that `erl_anno:anno/0` is considered an opaque type, that usually happens to be a number. Sometimes it's a `{Line, Column}` tuple, and sometimes it's a keyword list.

Here's the definition of `new_location/1`. First off it's different with `DEBUG` set or not, but you can also see how locations are stored as a keyword list.

https://github.com/erlang/otp/blob/ad7b57a5dd1ef98f20b90e766e403ca06ebcd10a/lib/stdlib/src/erl_anno.erl#L118-L124

I'm trying to figure out how to compile with `DEBUG` set, to see what happens. @uabboli any pointers on how to do that?

maxno-kivra:
I've been thinking a bit about this one. This PR does not fix a bug, you're absolutely right @uabboli. What it does is pave the road towards supporting column numbers, and end location.

I've updated the commit message, title and description to reflect that. I've also rebased on latest master. What do you think?

maxno-kivra:
Since this hasn't been commented on I quite some time I think it's time to close it. If the need arises in the future it can always be reopened.

PR #3011, opened 2021-01-28
Explain {continue,Continue} in gen_server:Module:init/1 doc

ergl:
The value or the usage of `{continue,Continue}` is not explained on the documentation of `Module:init/1`. Although the meaning of `{continue,Continue}` and `handle_continue/2` are explained at the beginning of the gen_server documentation, I believe the context is lost if one is only reading the documentation of the individual functions, instead of reading the documentation for the entire module.

In the same vein, if one is reading the code for `gen_server`, the term `Timeout` might be confusing, because it can also refer to other values (`hibernate` or `continue`).

### DIFF
--- a/lib/stdlib/src/gen_server.erl
+++ b/lib/stdlib/src/gen_server.erl
@@ -395,9 +395,10 @@ init_it(Starter, Parent, Name0, Mod, Args, Options) ->
 	{ok, {ok, State}} ->
 	    proc_lib:init_ack(Starter, {ok, self()}), 	    
 	    loop(Parent, Name, State, Mod, infinity, HibernateAfterTimeout, Debug);
-	{ok, {ok, State, Timeout}} ->
+	{ok, {ok, State, TimeoutHibernateOrContinue}} ->
 	    proc_lib:init_ack(Starter, {ok, self()}), 	    
-	    loop(Parent, Name, State, Mod, Timeout, HibernateAfterTimeout, Debug);
+	    loop(Parent, Name, State, Mod, TimeoutHibernateOrContinue,
+	         HibernateAfterTimeout, Debug);
 	{ok, {stop, Reason}} ->
 	    %% For consistency, we must make sure that the
 	    %% registered name (if any) is unregistered before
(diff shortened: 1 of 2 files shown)


ranjanified:
Only for the sake of discussion, can we rename "[Extra](https://github.com/erlang/otp/blob/76cea47700240d657e283acd115d0befceb49c71/lib/stdlib/src/gen_server.erl#L398)" to something more meaningful, as it gives a sense of non-mandatory-ness, however the value in that position is very very involved in selecting the function clause of "[loop](https://github.com/erlang/otp/blob/76cea47700240d657e283acd115d0befceb49c71/lib/stdlib/src/gen_server.erl#L439)" which will be executed.  LoopSelector or LoopSelectionData etc.. Just my thoughts.

lib/stdlib/src/gen_server.erl
```diff
@@ -395,9 +395,9 @@ init_it(Starter, Parent, Name0, Mod, Args, Options) ->
 	{ok, {ok, State}} ->
 	    proc_lib:init_ack(Starter, {ok, self()}), 	    
 	    loop(Parent, Name, State, Mod, infinity, HibernateAfterTimeout, Debug);
-	{ok, {ok, State, Timeout}} ->
+	{ok, {ok, State, Extra}} ->
```
zuiderkwast, line 398:
```suggestion
	{ok, {ok, State, TimeoutHibernateOrContinue}} ->
```

lib/stdlib/src/gen_server.erl
```diff
@@ -395,9 +395,9 @@ init_it(Starter, Parent, Name0, Mod, Args, Options) ->
...
 	    proc_lib:init_ack(Starter, {ok, self()}), 	    
-	    loop(Parent, Name, State, Mod, Timeout, HibernateAfterTimeout, Debug);
+	    loop(Parent, Name, State, Mod, Extra, HibernateAfterTimeout, Debug);
```
zuiderkwast, line 400:
```suggestion
	    loop(Parent, Name, State, Mod, TimeoutHibernateOrContinue,
	         HibernateAfterTimeout, Debug);
```

zuiderkwast (approved):

ergl:
Applied suggestions since I wasn't too married to the current naming.

ranjanified:
Thanks. Indeed, sensible and even closer to the context. I only had an opinion. Even I am not married to names, either. :D

PR #3018, opened 2021-02-01
dialyzer: Add file and location to messages about unknowns

uabboli:
See also https://bugs.erlang.org/browse/ERL-1348.


paulo-ferraz-oliveira:
Tested this with `rebar3` from `master`. Got an apparently good result.

```erlang
src/mylib_app.erl
****: Unknown type mytype:unknown/0
****: Unknown type mytype:unknown2/0
```

I don't know why the `****` so I'm digging deeper, but it seems this is `rebar3`-specific.

paulo-ferraz-oliveira:
This is what seems to be breaking it: https://github.com/erlang/otp/pull/3018/files#diff-5650819a7f930097cba1693607578153d0fccf7c0d543269dacd89dc4dd37403R727.

I'm opening a pull request next to `rebar3` so this is tracked properly.

paulo-ferraz-oliveira:
@uabboli, this is an interface-breaking change at least where I point it out, right? Any other place we should be concerned about?

uabboli:
Thanks for the quick review.

I assume you mean "(Filename:Line:Column)" added at the end of the
output?

As a rule, we (the OTP team) do not consider changes in the textual
output as interface changes, so in that respect the change is not
incompatible. It should be mentioned in the release note anyway, of
course.

But I'm still not sure I understand what's going on. I tried to
inflict Dialyzer errors in a rebar3 project (PropEr), and all I
noticed was the extended output. Nothing went wrong. Does rebar3 parse
the output from calling Dialyzer from the command line? Does rebar3
call Dialyzer from Erlang as well?

uabboli:
Realized now that PropEr is most likely a bad example since it has a 'dialyzer' target in its Makefile.

paulo-ferraz-oliveira:
To enable errors, you have to use `{dialyzer, [{warnings, [unknown]}]` since the default is not considering `unknown`.

`rebar3` calls `dialyzer:run/1`.

I don't mean "textual return", I mean the "actual Erlang term return". At a given moment, you are return `{Line, Col}`, if I remember correctly. That's the interface-breaking change I'm referring to... I'll double check.

paulo-ferraz-oliveira:
> I'll double check.

Hm... seems you're not.

paulo-ferraz-oliveira:
Something's funny, here.

I saw this output yesterday:

```erlang
src/mylib_app.erl
****: Unknown type mytype:unknown/0
****: Unknown type mytype:unknown2/0
```

Now, I can't replicate it. 😄

paulo-ferraz-oliveira:
So, yeah...

I compile `rebar3` with your branch's Erlang/OTP.
I run `rebar3 dialyzer` (using the locally compiled `rebar3` and the runtime from your branch).

I get that "strange" `****` output.

I think `rebar3` won't get distributed (compiled with) with OTP 24 for a while, but we can already predict what'll happen.

paulo-ferraz-oliveira:
For reference.

* before:

```erlang
[{warn_unknown,{**[]**,    **0**},
   {unknown_type,{mytype,unknown,0}}},
 {warn_unknown,{**[]**,    **0**},
   {unknown_type,{mytype,unknown2,0}}}]
```

* after:

```erlang
[{warn_unknown,{**File**,  **{Line,Col}**},
   {unknown_type,{mytype,unknown,0}}},
 {warn_unknown,{**File**,  **{Line,Col}**},
   {unknown_type,{mytype,unknown2,0}}}]
```

uabboli:
I tried another project, sourcer (lsp_server), and noticed the same
'****' as you did.

Regarding {Line, Column}: it's due to PR 3006. Thanks for pointing out
that it's an incompatible change of Dialyzer. I think we need to add
an {error_location, line | column} option to Dialyzer.

Regardless, I guess rebar will have to deal with {Line, Column}
eventually.

Thanks again for the feedback.

gomoripeti:
isn't the `****` comes from `io:format` trying to print a term which does not fit in a maximum width? (maybe the width calculation does not include the length of the new column part)

paulo-ferraz-oliveira:
@gomoripeti, you're probably right 😄. In any case, I've opened a [pull request next to `rebar3`](https://github.com/erlang/rebar3/pull/2481) and I asked if we should keep `{Line,Col}` or just `Line` (as was).

@uabboli: "Regardless, I guess rebar will have to deal with {Line, Column} eventually.". Yes, even if we add that option to `rebar3` also, it'll have to know how to parse the result properly.

uabboli:
> I think we need to add an {error_location, line | column} option to Dialyzer.

I take that back. If columns cannot be handled, then the compiler option {error_location, line} can be added.

This is the fallback recommended by OTP's technical board, and hopefully it works for (the users of) rebar.

paulo-ferraz-oliveira:
> > I think we need to add an {error_location, line | column} option to Dialyzer.
> 
> I take that back. If columns cannot be handled, then the compiler option {error_location, line} can be added.
> 
> This is the fallback recommended by OTP's technical board, and hopefully it works for (the users of) rebar.

You mean the output will be `{Line, Col}`, by default, but can be `Line` with that option? (that works for `rebar3`, I believe, it just needs to adapt to both possible output formats).

uabboli:
If all files analyzed by Dialyzer are files compiled with the
{error_location, line} option, then the output will always be Line.

As I wrote this, I realized that I've missed that currently column
numbers are always included when Dialyzer analyzes source files. Don't know
if that affects rebar.

paulo-ferraz-oliveira:
> Don't know if that affects rebar.

Even if it does, it shouldn't be difficult to adapt to the upcoming return type. At the moment, `dialyzer:run/1` is [`-spec(_).`ed as returning `Line :: integer`](https://erlang.org/doc/man/dialyzer.html#run-1). As soon as you've reached what you expect is the OPT 24 definitive output, I'll update my pull request to `rebar3`. I'd asked, and agree with @ferd's comment, that if `Col` (column) is present we should find a format to output it to the consumer.

uabboli:
Since Dialyzer compiles files, I'll have to add an 'error_location' option.

Otherwise, unless something else comes up, I'll merge this PR soon as it does not seem to introduce any new issues.

uabboli:
Merged commit c0ba13bf4a725fb6eaae63e08180dfaeb69a7aab into erlang:master.

PR #3019, opened 2021-02-01
Make compiler accept abstract format as input

richcarl:
This makes life easier for people implementing languages on top of Erlang, allowing them to easily test things and also skipping the normal linter when they know they already have done their own linting. This removes the need for complicated workarounds.

### DIFF
--- a/lib/compiler/doc/src/compile.xml
+++ b/lib/compiler/doc/src/compile.xml
@@ -437,6 +437,14 @@ module.beam: module.erl \
               parsed code before the code is checked for errors.</p>
           </item>
 
+	  <tag><c>from_abstr</c></tag>
+          <item>
+            <p>The input file is expected to contain Erlang terms representing
+            forms in abstract format (default file suffix ".abstr"). Note
+            that the format of such terms can change between releases.</p>
+            <p>See also the <c>no_lint</c> option.</p>
+          </item>
+
 	  <tag><c>from_asm</c></tag>
           <item>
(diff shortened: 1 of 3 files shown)


lib/compiler/doc/src/compile.xml
```diff
@@ -512,13 +520,23 @@ module.beam: module.erl \
...
 
+          <tag><c>no_lint</c></tag>
+          <item>
+            <p>Skips the pass that checks for errors and warnings. Only
+            applicable together with the <c>from_abstr</c> option. This is
+            mainly for implementations of other languages on top of Erlang,
+            which have already done their own checks to guarantee
+            correctness of the code.</p>
+            <p>Caveat: If the code contains errors, the compiler may crash
+            instead of producing any readable error messages.</p>
```
jhogberg, line 537: This part gives me the impression that it will always crash hard in the presence of errors, but at worst it could produce incorrect code without crashing. If we decide to add this option then I think it should be strongly noted that we give no guarantees.

richcarl: How about "Caveat: When this option is used, there are no guarantees that the code output by the compiler is correct and safe to use. The responsibility for correctness lies on the code or person generating the abstract format. If the code contains errors, the compiler may crash or produce unsafe code."

jhogberg: Sounds good to me. :)

bjorng:
Added to our daily builds.

PR #3020, opened 2021-02-01
Quote source as part of error messages

richcarl:
Provides one or more additional lines of context for compiler messages,  showing the relevant source code with the error position indicated by ^, like this:
```
foo.erl:13:10: variable 'X' is unbound
%   13|     case X of
%     |          ^

foo.erl:18:16: illegal guard expression
%   18| foo(A, B) when is_lit(A), is_list(B) ->
%     |                ^

foo.erl:23:22: function bar/1 undefined
%   23|             {ok, {H, bar(T)}}
%     |                      ^

foo.erl:27:5: record foo undefined
%   26| bar(X, Y) ->
%   27|     #foo{a = baz(X, Y),
%     |     ^
%   28|          b = f(),
%     | ...
%   31|          y = false,
%   32|          z = true}.
```
These long messages can be suppressed with the 'brief' compiler option.
The message formatting code is moved out to a separate module.

The followup PR https://github.com/erlang/otp/pull/3026 also adds end position to the parser, allowing the error ranges to be underlined.

### DIFF
--- a/erts/test/erlc_SUITE.erl
+++ b/erts/test/erlc_SUITE.erl
@@ -69,7 +69,8 @@ end_per_group(_GroupName, Config) ->
 %% Tests that compiling Erlang source code works.
 
 compile_erl(Config) when is_list(Config) ->
-    {SrcDir, OutDir, Cmd} = get_cmd(Config),
+    {SrcDir, OutDir, Cmd0} = get_cmd(Config),
+    Cmd = Cmd0 ++ " +brief",
     FileName = filename:join(SrcDir, "erl_test_ok.erl"),
 
     %% By default, warnings are now turned on.
@@ -98,7 +99,6 @@ compile_erl(Config) when is_list(Config) ->
     BadFile = filename:join(SrcDir, "erl_test_bad.erl"),
     run(Config, Cmd, BadFile, "", ["function non_existing/1 undefined\$",
                                    "_ERROR_"]),
-
     ok.
 
 %% Test that compiling yecc source code works.
@@ -221,7 +221,8 @@ deep_cwd_1(PrivDir) ->
 %% Test that a large number of command line switches does not
 %% overflow the argument buffer
 arg_overflow(Config) when is_list(Config) ->
-    {SrcDir, _OutDir, Cmd} = get_cmd(Config),
+    {SrcDir, _OutDir, Cmd0} = get_cmd(Config),
+    Cmd = Cmd0 ++ " +brief",
     FileName = filename:join(SrcDir, "erl_test_ok.erl"),
     %% Each -D option will be expanded to three arguments when
     %% invoking 'erl'.
@@ -266,33 +267,34 @@ erlc() ->
     end.
 
 make_dep_options(Config) ->
-    {SrcDir,OutDir,Cmd} = get_cmd(Config),
+    {SrcDir,OutDir,Cmd0} = get_cmd(Config),
+    Cmd = Cmd0 ++ " +brief",
     FileName = filename:join(SrcDir, "erl_test_ok.erl"),
     BeamFileName = filename:join(OutDir, "erl_test_ok.beam"),
 
     DepRE = ["/erl_test_ok[.]beam: \\\\$",
-             "/system_test/erlc_SUITE_data/src/erl_test_ok[.]erl \\\\$",
-             "/system_test/erlc_SUITE_data/include/erl_test[.]hrl$",
+             "/erlc_SUITE_data/src/erl_test_ok[.]erl \\\\$",
+             "/erlc_SUITE_data/include/erl_test[.]hrl$",
              "_OK_"],
 
     DepRETarget =
     ["^target: \\\\$",
-     "/system_test/erlc_SUITE_data/src/erl_test_ok[.]erl \\\\$",
-     "/system_test/erlc_SUITE_data/include/erl_test[.]hrl$",
+     "/erlc_SUITE_data/src/erl_test_ok[.]erl \\\\$",
+     "/erlc_SUITE_data/include/erl_test[.]hrl$",
      "_OK_"],
 
     DepREMP =
     ["/erl_test_ok[.]beam: \\\\$",
-     "/system_test/erlc_SUITE_data/src/erl_test_ok[.]erl \\\\$",
-     "/system_test/erlc_SUITE_data/include/erl_test[.]hrl$",
+     "/erlc_SUITE_data/src/erl_test_ok[.]erl \\\\$",
+     "/erlc_SUITE_data/include/erl_test[.]hrl$",
      [],
-     "/system_test/erlc_SUITE_data/include/erl_test.hrl:$",
+     "/erlc_SUITE_data/include/erl_test.hrl:$",
      "_OK_"],
 
     DepREMissing =
     ["/erl_test_missing_header[.]beam: \\\\$",
-     "/system_test/erlc_SUITE_data/src/erl_test_missing_header[.]erl \\\\$",
-     "/system_test/erlc_SUITE_data/include/erl_test[.]hrl \\\\$",
+     "/erlc_SUITE_data/src/erl_test_missing_header[.]erl \\\\$",
+     "/erlc_SUITE_data/include/erl_test[.]hrl \\\\$",
      "missing.hrl$",
      "_OK_"],
 
@@ -358,9 +360,9 @@ make_dep_options(Config) ->
     %% since compiler is run on the erlang code a warning will be
     %% issued by the compiler, match that.
     WarningRE =
-         "/system_test/erlc_SUITE_data/src/erl_test_ok.erl:[0-9]+:[0-9]+: "
+         "/erlc_SUITE_data/src/erl_test_ok.erl:[0-9]+:[0-9]+: "
         "Warning: function foo/0 is unused$",
-    ErrorRE = "/system_test/erlc_SUITE_data/src/erl_test_missing_header.erl:"
+    ErrorRE = "/erlc_SUITE_data/src/erl_test_missing_header.erl:"
         "[0-9]+:[0-9]+: can't find include file \"missing.hrl\"$",
 
     DepRE_MMD = insert_before("_OK_", WarningRE, DepRE),
--- a/lib/compiler/doc/src/compile.xml
+++ b/lib/compiler/doc/src/compile.xml
@@ -95,6 +95,13 @@
 
         <p>Available options:</p>
         <taglist>
+	  <tag><c>brief</c></tag>
+          <item>
+            <p>Restricts error and warning messages to a single line of output.
+            As of OTP 24, the compiler will by default also display the part
+            of the source code that the message refers to.</p>
+          </item>
+
           <tag><c>basic_validation</c></tag>
           <item>
             <p>This option is a fast way to test whether a module will
--- a/lib/compiler/src/Makefile
+++ b/lib/compiler/src/Makefile
@@ -98,6 +98,7 @@ MODULES =  \
 	sys_core_fold_lists \
 	sys_core_inline \
 	sys_core_prepare \
+	sys_messages \
 	sys_pre_attributes \
 	v3_core \
 	v3_kernel \
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -1781,8 +1781,8 @@ write_binary(Name, Bin, St) ->
 report_errors(#compile{options=Opts,errors=Errors}) ->
     case member(report_errors, Opts) of
 	true ->
-	    foreach(fun ({{F,_L},Eds}) -> list_errors(F, Eds);
-			({F,Eds}) -> list_errors(F, Eds) end,
+	    foreach(fun ({{F,_L},Eds}) -> sys_messages:list_errors(F, Eds, Opts);
+			({F,Eds}) -> sys_messages:list_errors(F, Eds, Opts) end,
 		    Errors);
 	false -> ok
     end.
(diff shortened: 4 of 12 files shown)


mheiber:
I'm so excited for this change, I think it will be a huge devX improvement 🎉.


Some small bugs with +deterministic:

- With +deterministic, when there is a file with the same basename as the source file in the current directory, lines from the file in the current directory are displayed instead of in the source file. You can see this with an example.erl in `..`, then `yes > example.erl` CTRL-C then `erlc +deterministic ../example.erl`
- With +deterministic, when there is no file with the same name in the current directory, there are no error lines displayed. We hit the error-swallowing case [here](https://github.com/erlang/otp/pull/3020/files#diff-ee923910a9f36565732ca18d7bbd06a69a3e0ac210df5e249f6203e9d1d73608R107).

The root cause seems to be that `SourceName` has two incompatible meanings:
- thing used for display, which can be shortened
- thing used for reading from the fs, which must be accurate

lib/compiler/src/compile_messages.erl
```diff
@@ -0,0 +1,214 @@
...
+        true -> "";
+        false -> quote_source_1(File, StartLoc, EndLoc)
+    end.
+
+quote_source_1(File, Line, Loc2) when is_integer(Line) ->
+    quote_source_1(File, {Line, 1}, Loc2);
+quote_source_1(File, Loc1, Line) when is_integer(Line) ->
+    quote_source_1(File, Loc1, {Line, -1});
+quote_source_1(File, {StartLine, StartCol}, {EndLine, EndCol}) ->
+    case file:read_file(File) of
```
mheiber, line 96: It looks like we are re-reading the source file for each error line. Is there a way to do fewer reads?

richcarl: Absolutely, this was a simplistic implementation. Considering OS file caches, it's probably not too bad as it is even for a hundred or more errors, but you could implement a caching system (which would have to handle include files as well, not just the main source file).

richcarl:
To make things easier for upstream adoption, I am splitting this into two separate parts. This part now only adds the messages and points out the start position, so it does not affect the parser annotations.

lib/compiler/src/compile_messages.erl
```diff
@@ -0,0 +1,214 @@
...
+
+fmt_line(L, Text) ->
+    io_lib:format("~ts~4.ts| ~ts\n", [line_prefix(), line_to_txt(L), Text]).
+
+line_to_txt(0) -> "";
+line_to_txt(L) -> integer_to_list(L).
+
+decorate([{Line, Text} = L | Ls], StartLine, StartCol, EndLine, EndCol) when
+    Line =:= StartLine, EndLine =:= StartLine
+->
```
bjorng, line 147:
```suggestion
    Line =:= StartLine, EndLine =:= StartLine ->
```

lib/compiler/src/compile_messages.erl
```diff
@@ -0,0 +1,214 @@
...
+    S = underline(Text, StartCol, EndCol),
+    decorate(S, L, Ls, StartLine, StartCol, EndLine, EndCol);
+decorate([{Line, Text} = L | Ls], StartLine, StartCol, EndLine, EndCol) when Line =:= StartLine ->
+    %% start with end on separate line
+    S = underline(Text, StartCol, string:length(Text) + 1),
+    decorate(S, L, Ls, StartLine, StartCol, EndLine, EndCol);
+%% decorate([{Line, Text}=L|Ls], StartLine, StartCol, EndLine, EndCol)
+%%   when Line =:= EndLine ->
+%%     S = underline(Text, EndCol,EndCol),  % just mark end
+%%     decorate(S, L, Ls, StartLine, StartCol, EndLine, EndCol);
```
bjorng, line 158: We don't like out-commented code.

bjorng (changes requested):
`compile_SUITE:app_test/1` fails because you have not included `compile_messages` in the app file.

Found two minor style issues.

Otherwise, looks good to me.

lib/compiler/src/compile_messages.erl
bjorng: Seems to be fast enough when compiling the compiler test suite (which has many warnings). In real code, there is probably very few warnings, as many projects use `-Werror` nowadays.

richcarl:
> Otherwise, looks good to me.

Any suggestion for a different module name than "compile_messages", or is it ok for you?

garazdawi:
Should there be tests for this? or are the existing tests sufficient?

lib/compiler/src/compile_messages.erl
richcarl: Some remnants of running erlfmt on my original patch. Fixing.

lib/compiler/src/compile_messages.erl
richcarl: Removing.

ranjanified:
> > Otherwise, looks good to me.
> 
> Any suggestion for a different module name than "compile_messages", or is it ok for you?

If you can make it sound like a "object", that will be good. Modules probably are well-suited with a notion of a "subject". "compile_messages" sound more like a verb. Just my thoughts. Honestly haven't gone through the details of the PR, so excuse me for being shallow.

richcarl:
> "compile_messages" sound more like a verb.

That's basically why I'm asking (and also because someone out there might be using this name for something else already). The main point with the current name is that it indicates its relation to the "compile" module.

bjorng:
> Any suggestion for a different module name than "compile_messages", or is it ok for you?

Traditionally, Robert and Joe use `sys_` as a prefix for compiler modules, and it is still used for some compiler modules.

So, perhaps, `sys_messages`.

richcarl:
Fixed style issues and renamed to sys_messages.

richcarl:
Tried rebasing on master as well to trigger another rebuild, but it's still failing. Docker trouble?

bjorng (approved):

bjorng:
> Tried rebasing on master as well to trigger another rebuild, but it's still failing. Docker trouble?

Yes, seems to be. It builds fine on our Jenkins server, though.

bjorng:
There are failed test cases in erlc_SUITE (system) and escript_SUITE (stdlib) fail.

garazdawi:
```
foo() ->
    A = <<"a𐐀b"/utf16>>, <<"hello">>,
    A.
```

On my machine this prints:

```
t.erl:14:26: Warning: a term is constructed, but never used
%   14|     A = <<"a\x{10400}b"/utf16>>, <<"hello">>,
%     |                          ^
```

garazdawi:
Also, does this need to take `%% coding: latin-1` into consideration?

bjorng:
@garazdawi The problem you found is a problem with `erlc`. I have known about, but haven't got around to fix it. I might try to fix it for OTP 24.

If you compile your function with the `erl` shell, it will work:

```
1> c(t).
t.erl:5:26: Warning: a term is constructed, but never used
%    5|     A = <<"a𐐀b"/utf16>>, <<"hello">>,
%     |                          ^
```

There should be no need take any `coding` into account, because the column position is the number of graphemes. Here is another example:

```
bar() ->
    A = <<"привет"/utf8>>, <<"hello">>,
    B = <<"abcdef"/utf8>>, <<"hello">>,
    {A,B}.
```

The carets are placed in the correct place:

```
t.erl:9:28: Warning: a term is constructed, but never used
%    9|     A = <<"привет"/utf8>>, <<"hello">>,
%     |                            ^

t.erl:10:28: Warning: a term is constructed, but never used
%   10|     B = <<"abcdef"/utf8>>, <<"hello">>,
%     |                            ^
```

garazdawi:
Great!

bjorng:
@richcarl We found that `int_SUITE:init_per_suite/1` in the debugger applications fails with a timetrap timeout on *some* of our test machines. That test case compiles a huge file with many warnings. On my Mac it needs 25 seconds. Some of our test machines needs more than one minute, which is the current timetrap limit.

Could you fix this in one of two ways?

* Raise the timetrap limit, for example to ten minutes. https://github.com/erlang/otp/blob/c28f3011561e3b80cf33f1fb3f30f3504cd67d5d/lib/debugger/test/int_SUITE.erl#L57
* Implement some sort of caching, for example by spawning off a caching process the first time an error or warning is printed.

richcarl:
> @richcarl We found that `int_SUITE:init_per_suite/1` in the debugger applications fails with a timetrap timeout on _some_ of our test machines. That test case compiles a huge file with many warnings. On my Mac it needs 25 seconds. Some of our test machines needs more than one minute, which is the current timetrap limit.

Bumped to 10 min. Caching can wait until later.

richcarl:
> Also, does this need to take `%% coding: latin-1` into consideration?

Found that the underlining was too naive, and could get out of sync with tabs in combination with utf-8 multibytes. Now treats lines as sequences of codepoints, respecting the file encoding. This should always correspond to the column number computed by the scanner. (Forced edit: handle broken encodings in a sensible way instead of crashing.)

bjorng:
> Found that the underlining was too naive, and could get out of sync with tabs in combination with utf-8 multibytes. Now treats lines as sequences of codepoints, respecting the file encoding. This should always correspond to the column number computed by the scanner. (Forced edit: handle broken encodings in a sensible way instead of crashing.)

Do you think you could add a test case to test those things you just fixed? To avoid have to constantly update that test case, it would probably be best to call `sys_messages` directly with a faked list of errors and warnings so that the output will not change if other parts of the compiler are updated.

richcarl:
> Do you think you could add a test case to test those things you just fixed?

Added test case. Running the full compiler to ensure that encodings are checked properly.

bjorng:
There are still two failing test cases in `erlc_SUITE` (`make_dep_options`).

richcarl:
See if you like this solution better before I squash. Also added a fix for the hardcoded paths preventing you from running erlc_SUITE directly from make under erts.

uabboli:
> The followup PR #3026 also adds end position to the parser, allowing
> the error ranges to be underlined.

I think all traces of ranges (#3026) should be removed in this PR.
erl_lint should not be updated at all, and the code in sys_messages,
starting with

 StartLoc = erl_anno:location(Loc).

should not try to be forward compatible with ranges. It breaks the
erl_anno abstraction (the current code will crash if the erl_anno
module is compiled with DEBUG=true).

richcarl:
> I think all traces of ranges (#3026) should be removed in this PR.
> erl_lint should not be updated at all, and the code in sys_messages,
> starting with
> 
> StartLoc = erl_anno:location(Loc).
> 
> should not try to be forward compatible with ranges. It breaks the
> erl_anno abstraction (the current code will crash if the erl_anno
> module is compiled with DEBUG=true).

I agree, so I'll try to remove stuff that shouldn't have been in this part to begin with.

Something is weird with the types, though. in compile.erl, a comp_ret(), which is the return type for compile:file() and compile:forms(), is defined as reporting errors and warnings that contain err_info() :: {erl_anno:line() | 'none', module(), term()}. Note that it says line(), which is defined in erl_anno as non_neg_integer(), and this should be changed to location() to reflect the new support for column numbers, which might fix the DEBUG compilation problem.

Meanwhile, internally in compile.erl the errors and warnings fields of the `#compile{}` record use the type err_warn_info() which is just defined as tuple(). The messages_per_file() function takes the internal err_warn_info() representation and produces the err_info() version.

I don't really want to start on a type annotation whack-a-mole to try to clean up that stuff myself, since it might cause some cascading type issues, at least not until I know what you think how it should look.

bjorng:
> See if you like this solution better before I squash. Also added a fix for the hardcoded paths preventing you from running erlc_SUITE directly from make under erts.

Looks good to me, and the test cases now run clean.

bjorng:
I think you can squash it now. (I would prefer down to a single commit.) If everything goes well in our nightly tests, I intend to merge this PR tomorrow.

uabboli:
> >     I think all traces of ranges (#3026) should be removed in this PR.
> >     erl_lint should not be updated at all, and the code in sys_messages,
> >     starting with

I see that you've reverted erl_lint. Looks OK.

> >     StartLoc = erl_anno:location(Loc).
> >
> >     should not try to be forward compatible with ranges. It breaks the
> >     erl_anno abstraction (the current code will crash if the erl_anno
> >     module is compiled with DEBUG=true).

You've haven't fixed this, as far as I can see.

> Note that it says line(), which is defined in erl_anno as
> non_neg_integer(), and this should be changed to location() to reflect
> the new support for column numbers, which might fix the DEBUG
> compilation problem.

No, the erl_anno-abstraction is broken in a more fundamental way.

The quoted line,

    StartLoc = erl_anno:location(Loc).

breaks the erl_anno abstraction.

    format_messages(F, P, [{Loc, Mod, E} | Es], Opts) ->
        Src = quote_source(F, Loc, Loc, Opts),

should suffice.

lists_errors also breaks the erl_anno-abstraction:

    list_errors(F, [{StartLoc, Mod, E} | Es], Opts) ->
        list_errors(F, [{{StartLoc, StartLoc}, Mod, E} | Es], Opts);

I suppose you can keep the EndLoc in quote_source as long it's
correct, although it's not needed in this PR.

richcarl:
Squashed. Working on a cleanup of type usages, but I can put that in a separate PR after this is merged.

richcarl:
@uabboli Check out https://github.com/erlang/otp/pull/3045 for a patch that tries to unify the type annotations for errors/warnings over the compiler modules.

richcarl:
Accidentally deleted a clause that was actually used. Also finally saw what @uabboli meant, so I removed the calls to erl_anno on a location type.

bjorng:
Thanks, I can confirm that it works with `erl_anno` debugging activated (by uncommenting the defines for `DEBUG` in `erl_anno`, `erl_parse`, and `erl_pp`). Added to our daily builds.

bjorng:
Thanks for your pull request.

PR #3021, opened 2021-02-01
EDoc: Add @vendor tag

Maria-12648430:
The current implementation of `edoc` places an Erlang logo and a link to `http://www.erlang.org/` in the top/bottom right corners of all generated overview and module pages. This has always bothered me, since (to me at least) it suggests that the documented modules are in some way related to Erlang _as an organization_, while most of the time it really is just some software _written_ in Erlang. For an example, see [the API documentation for MySQL-OTP](https://mysql-otp.github.io/mysql-otp/), which is a MySQL driver written in/for Erlang, but otherwise has no relation to Erlang as such.

The logo and link are hardcoded into the `edoc_layout` module, and there is no easy way to change it other than post-processing the generated files (something which I guess is being done with [the API documentation for PropEr](http://proper.softlab.ntua.gr/doc/)) or maybe some CSS/JS trickery.

This PR introduces the `@vendor` tag for the overview and module pages in to allow customization of the link and logo, with the following semantics (when using the standard layout that comes with `edoc`):
* if the `@vendor` tag is absent, the layout renders the Erlang logo and the link to `http://www.erlang.org/` as it does now. While this is actually counter-intuitive, this behavior is required to keep backwards compatibility, especially with software that post-processes the generated files in order to replace the logo and link with their own
* if the `@vendor` tag is present but empty, the layout renders no logo or link, ie "no vendor"
* if the `@vendor` tag is present and not empty, it _must_  contain the vendor name, and _may_ optionally contain a website URI (ie, link to vendor website) in square brackets (`[...]`) and/or a logo URI (ie. the URI of a logo image) in angle brackets (`<...>`), in any order (or rather, the same ordering rules as for the `@author` tags apply). This information is used by the standard layout to render what is displayed in the top/bottom right corners instead of the Erlang logo/link according to the following rules:
  * if only the name is given (no website or logo URIs), only that name appears
  * if a logo URI is given, the logo appears, and the name is placed in the `alt` attribute of the `img` tag
  * if a website URI is given, the name or logo, a link is placed on the name or logo

There is a slight difference between the current (or the equivalent no-@vendor-tag) behavior and the one introduced with the `@vendor` tag: The Erlang logo is copied to the location of the generated documentation (ie, it is _part of_ the documentation), while logo images denoted via the `@vendor` tag are not.

<hr />

About the same could be achieved by introducing a new `{vendor, ...}` option instead of a new `@vendor` tag. This way, a mechanism to copy a logo image to the documentation would probably be easy to implement as well. On the other hand, it takes the vendor information out of the source files and places it in the hands of the caller. For this reason, I decided against it, but I'm open to suggestions of course ;)

### DIFF
--- a/lib/edoc/doc/overview.edoc
+++ b/lib/edoc/doc/overview.edoc
@@ -245,6 +245,10 @@ The following tags can be used in an overview file.
       <em>only</em> be used in an overview file. The content can be
       arbitrary text.</dd>
 
+  <dt><a name="otag-vendor">`@vendor'</a></dt>
+      <dd>See the <a href="#mtag-vendor">`@vendor' module tag</a> for
+      details.</dd>
+
   <dt><a name="otag-version">`@version'</a></dt>
       <dd>See the <a href="#mtag-version">`@version' module
       tag</a> for details.</dd>
@@ -338,6 +342,30 @@ The following tags can be used before a module declaration:
       <dd>Specifies when the module was introduced, with respect to
       the application, release or distribution it is part
       of. The content can be arbitrary text.</dd>
+
+  <dt><a name="mtag-vendor">`@vendor'</a></dt>
+      <dd>Specifies the name of the vendor, along with a website URI and
+      logo. An URI can be given within `[...]', and a logo within `<...>'.
+      The URI and logo are optional. A completely empty `@vendor' tag can
+      be used to explicitly indicate "no vendor". Any surrounding
+      whitespaces are stripped from all strings.
+
+      The name is the first nonempty string that is not within `<...>'
+      or `[...]', and does not contain only whitespace. (In other words,
+      the name can come before, between, or after the URI and logo,
+      but cannot be split up; any sections after the first are ignored.)
+
+      Examples:
+```%% @vendor Example Inc. [http://www.example.net/] <http://www.example.net/logo.png>'''
+
+```%% @vendor Example Inc. [http://www.example.net/]'''
+
+```%% @vendor Example Inc. <http://www.example.net/logo.png>'''
+
+```%% @vendor Example Inc.'''
+
+```%% @vendor'''
+      </dd>
 
   <dt><a name="mtag-version">`@version'</a></dt>
       <dd>Specifies the module version. The content can be arbitrary
--- a/lib/edoc/priv/edoc.dtd
+++ b/lib/edoc/priv/edoc.dtd
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="ISO-8859-1"?>
 <!-- EDoc DTD Version 0.3 -->
 
-<!ELEMENT overview (title, description?, author*, copyright?, version?,
-                    since?, see*, reference*, todo?, modules)>
+<!ELEMENT overview (title, description?, vendor?, author*, copyright?,
+                    version?, since?, see*, reference*, todo?, modules)>
 <!ATTLIST overview
   root CDATA #IMPLIED
   encoding CDATA #IMPLIED>
@@ -12,8 +12,8 @@
 <!ELEMENT modules (module+)>
 
 
-<!ELEMENT module (args?, description?, author*, copyright?, version?,
-		  since?, deprecated?, see*, reference*, todo?,
+<!ELEMENT module (args?, description?, vendor?, author*, copyright?,
+                  version?, since?, deprecated?, see*, reference*, todo?,
 		  behaviour*, callbacks?, typedecls?, functions)>
 <!ATTLIST module
   name CDATA #REQUIRED
@@ -25,6 +25,12 @@
 <!ELEMENT description (briefDescription, fullDescription?)>
 <!ELEMENT briefDescription (#PCDATA)>
 <!ELEMENT fullDescription (#PCDATA)>
+
+<!ELEMENT vendor EMPTY>
+<!ATTLIST vendor
+  name CDATA #REQUIRED
+  website CDATA #IMPLIED
+  logo CDATA #IMPLIED>
 
 <!ELEMENT author EMPTY>
 <!ATTLIST author
--- a/lib/edoc/src/edoc_data.erl
+++ b/lib/edoc/src/edoc_data.erl
@@ -40,7 +40,7 @@
 %% TODO: check that variables in @equiv are found in the signature
 %% TODO: copy types from target (if missing) when using @equiv
 
-%% <!ELEMENT module (args?, description?, author*, copyright?,
+%% <!ELEMENT module (args?, description?, vendor?, author*, copyright?,
 %%                   version?, since?, deprecated?, see*, reference*,
 %%                   todo?, behaviour*, callbacks?, typedecls?,
 %%                   functions)>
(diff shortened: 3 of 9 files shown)


odiferousmint:
I suppose it should show more information on the bottom then, for example as to what "EDoc" is or what it is related to (the programming language Erlang), because without the logo it really is missing as to where it comes from, what it is related to, and so forth.

Maria-12648430:
> I suppose it should show more information on the bottom then, for example as to what "EDoc" is or what it is related to (the programming language Erlang), because without the logo it really is missing as to where it comes from, what it is related to, and so forth.

I disagree. EDoc is, give or take, just for documenting _APIs_. If you get to the _API_ documentation of a project, I would assume that you pretty much know what it is about, what programming language it is in (and for), etc. You want to find out how to _use_ it (and yes, the logo is not needed for _that_ purpose ;))

odiferousmint:
> > I suppose it should show more information on the bottom then, for example as to what "EDoc" is or what it is related to (the programming language Erlang), because without the logo it really is missing as to where it comes from, what it is related to, and so forth.
> 
> I disagree. EDoc is, give or take, just for documenting _APIs_. If you get to the _API_ documentation of a project, I would assume that you pretty much know what it is about, what programming language it is in (and for), etc. You want to find out how to _use_ it (and yes, the logo is not needed for _that_ purpose ;))

*Shrugs*. We could get rid of a lot of things, then, if we keep assuming things. While it is not that incorrect of an assumption, I do not see why we should get rid of this information; other languages keep such information at the very least. Either way, it is not much of a concern to me, after all. :)

odiferousmint:
> > I suppose it should show more information on the bottom then, for example as to what "EDoc" is or what it is related to (the programming language Erlang), because without the logo it really is missing as to where it comes from, what it is related to, and so forth.
> 
> I disagree. EDoc is, give or take, just for documenting _APIs_. If you get to the _API_ documentation of a project, I would assume that you pretty much know what it is about, what programming language it is in (and for), etc. You want to find out how to _use_ it (and yes, the logo is not needed for _that_ purpose ;))

*Shrugs*. We could get rid of a lot of things, then, if we keep assuming things. While it is not that incorrect of an assumption, I do not see why we should get rid of or not include this information; other languages keep such information at the very least. Either way, it is not much of a concern to me, after all. :)

Your rationale for getting rid of it is "This has always bothered me, since (to me at least) it suggests that the documented modules are in some way related to Erlang as an organization,". Personally, I never made the assumption that it belonged to Erlang as an organization, because I have seen official documentation before. We could arguably assume that people who have gotten this far would be able to differentiate between what is and is not related to Erlang as an organization, in which case I could say the same thing: "I would assume that you pretty much know what [...]", which in itself could be an objection to getting rid of the Erlang logo.

In any case, I am fine with allowing them to put their own logo there, for example. I just thought it would not be that much of a deal to include a bit more information on the bottom.

Maria-12648430:
@odiferousmint ok, I guess we started off on a wrong foot here somehow, sorry about that =^^=

For one, I'm certainly not against, say, linking the "EDoc" in the footer to the `edoc` documentation, or putting additional links to `erlang.org` there. But, from what you said...

> I suppose it should show more information on the bottom then, for example as to what "EDoc" is or what it is related to (the programming language Erlang)...

... it sounded to me like you suggested something like "Generated by EDoc, which is the API documentation tool for Erlang (like Javadoc, but for Erlang), which in turn is a functional programming language... etcetc" (I exaggerate a lot, don't take me too seriously ;))

For another, I don't think my assumption is too far-fetched, just maybe a bit biased by my own mindset: When _I_ read such documentation for a project/library/whatever (which I may have found for a new project or in the deps of one I inherited), what I really want to know is _who_ made (and can help me with) _this particular piece of software_, and so I think this information should be placed somewhere prominently. OTOH, I already _know_ that it is _in_ Erlang and I know_ what_ Erlang is  (otherwise, why would _I_ be looking at _this_ in the first place?), and _they_ can't help me with this particular piece of software, so that's not very helpful information for me. Phrased like this, do you agree? :)

For yet another, for anyone who does not care about the vendor information, everything stays as it is, even with these changes: the Erlang logo gets displayed, linked to `erlang.org`. What this PR is really about is that with it, you _can_ customize it ;)

odiferousmint:
> ok, I guess we started off on a wrong foot here somehow, sorry about that =^^=

No worries really. I probably explained myself in an unclear way. It happens quite often actually, so I am pretty much to blame for that.

> it sounded to me like you suggested something like "Generated by EDoc, which is the API documentation tool for Erlang (like Javadoc, but for Erlang), which in turn is a functional programming language... etcetc"

No, no, definitely not of course. :D I apologize for the misunderstanding, we probably spoke over each other. :)

Maria-12648430:
> No hard feelings!

No worries :) Let's see what the OTP people think :)

odiferousmint:
> > No hard feelings!
> 
> No worries :) Let's see what the OTP people think :)

Yeah, in all fairness I would not mind keeping it simple and to me it could work without any of what we said, including the Erlang logo. Just have some nice Erlang-ish template to differentiate like OCaml has[1], for example, and that should be enough.

[1] Example: https://caml.inria.fr/pub/docs/manual-ocaml/libref/Dynlink.html

Maria-12648430:
From what I saw, `edoc` seems to have acquired some dust ^^;

garazdawi:
Hello!

We are not opposed to this addition. I will need some time to dig into the specifics of the code changes, as I (nor anyone else really) is not very familiar with the edoc code base. However, from a first glance I can see that the tests need to be expanded to check that this works.

Maria-12648430:
> We are not opposed to this addition. I will need some time to dig into the specifics of the code changes, as I (nor anyone else really) is not very familiar with the edoc code base. However, from a first glance I can see that the tests need to be expanded to check that this works.

Ok, I agree. I'll see to tests next week then. I just wanted to hear what you think before I invest more work into it.

Have a nice weekend :)

Maria-12648430:
> I'll see to tests next week then.

~Oh what have I gotten myself into again... 🙄~

@garazdawi Looking at the tests that do exist in `edoc`, there is precious little that I can draw from or build on. There are no tests for the output that is generated (granted, that would not be easy). What most of the tests do is pretty much only check if some source and overview files _can_ be processed at all.

So, what I can do is test `edoc_lib:parse_vendor/2`, and make up another overview file with a `@vendor` tag in it to see if `edoc` throws up on it. Would that suffice? =^^=

garazdawi:
> What most of the tests do is pretty much only check if some source and overview files can be processed at all.

I'm aware that edoc does not have the best test coverage :)

Having a test that makes sure that the new tag does not crash seems like a good start. So maybe adding a new module to the myapp test application that uses the new tag and then also generating docs with an alternative overview that contains the tag?

If you want to validate the contents, then maybe look at the output and use a regex to look for the new information that you expect. It does not need to be perfect, just something to prove that it actually works :)

Maria-12648430:
Ok, sounds good :)

Maria-12648430:
@garazdawi tests are now in place :)

garazdawi:
I'll get to this once OTP-24 RC1 is released.

Maria-12648430:
> I'll get to this once OTP-24 RC1 is released.

NEW TOYS! 😍

Maria-12648430:
@garazdawi I just stumbled upon this >2yrs and 2 releases old PR :sweat_smile: Is it still on the list of things you may want, or shall I close it?

PR #3022, opened 2021-02-02
Allow maps as erl_eval bindings

josevalim:
When evaluating snippets with many variables,
sometimes most of the evaluation time is spent
on orddict find, as shown by eprof:

    :orddict.find/2  6611650 89.09 1000982    0.15

This patch allows a map to be given to erl_eval,
which is then kept internally as a map and returned
as map. This improves the performance in some cases
by more than 5 times while keeping backwards
compatibility.

### DIFF
--- a/lib/stdlib/doc/src/erl_eval.xml
+++ b/lib/stdlib/doc/src/erl_eval.xml
@@ -47,7 +47,8 @@
     </datatype>
     <datatype>
       <name name="binding_struct"/>
-      <desc><p>A binding structure.</p></desc>
+      <desc><p>A binding structure. It is either a <c>map</c> or an <c>orddict</c>.
+        <c>erl_eval</c> will always return the same type as the one given.</p></desc>
     </datatype>
     <datatype>
       <name name="expression"/>
--- a/lib/stdlib/src/erl_eval.erl
+++ b/lib/stdlib/src/erl_eval.erl
@@ -45,7 +45,7 @@
 -type(name() :: term()).
 -type(value() :: term()).
 -type(bindings() :: [{name(), value()}]).
--type(binding_struct() :: orddict:orddict()).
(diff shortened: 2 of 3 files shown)


josevalim:
One note: ideally we would change the internal implementation to just use maps, especially because we do have a high-level API to work with bindings, but `binding_struct` is publicly documented to be an orddict and I assume many are building said bindings directly. So I have decided to go in the same route as `sets` and support both formats, keeping the type of bindings that is given. orddicts could be deprecated in the future, if such is desired.

---

Also, for completeness, here is the scenario and some benchmark numbers of this change in practice.

In Elixir we have templates, which is a way to interpolate Elixir code on a mostly static file. Most times those templates are compiled into modules but sometimes those templates are evaluated, because they are meant to run only time to time, for example, to generate dev files.

@aaronrenner noticed that the template evaluation was very slow. When running the tests, we got this:

```
.....................

Finished in 58.2 seconds (0.00s async, 58.2s sync)
21 tests, 0 failures
```

He tried compiling the templates inside .beam files and we got this:

```
.....................

Finished in 2.4 seconds (0.00s async, 2.4s sync)
21 tests, 0 failures
```

I expect the evaluation to be slower but not by this much. After profiling, eprof made it clear the issue was with orddict:

```
:orddict.find/2      6611650 89.09 1000982    0.15
```

After applying this patch and changing Elixir to use maps, now we get:

```
.....................

Finished in 12.2 seconds (0.00s async, 12.2s sync)
21 tests, 0 failures

Randomized with seed 663365
```

As you can see, it got ~5x faster and that's considering the whole test suite times which includes many other options. So in practice you may see more than 5 times, it will depend on how many variables the code is using.

I have also ran the Elixir test suite while using maps on erl_eval and all tests pass. 🎉

PR #3024, opened 2021-02-02
dialyzer: Use types and specs for documentation

uabboli:
Also:

doc/manual.txt is removed.

The native options are no longer documented.


odiferousmint:
How come you deleted `doc/manual.txt`? Was it outdated?

I prefer `format_warning(Msg) -> string()` over `format_warning/1`, the former is more descriptive about what the parameters are and what it returns. Is it not the goal though, or what is the rationale for changing it?

garazdawi:
> I prefer format_warning(Msg) -> string() over format_warning/1, the former is more descriptive about what the parameters are and what it returns. Is it not the goal though, or what is the rationale for changing it? I did not check the code, the reason might be consistency.

The description will still be there, it is just generated from the spec. You can download the built docs from the GH actions checks and see for yourself: https://github.com/erlang/otp/actions/runs/530344973

uabboli:
manual.txt hasn't been updated for many years. I trust @kostis  will
protest if the file is to be kept as is.

I'm not sure I understand what it is that you prefer. Is it "Msg"
rather than "Warning"? I find it confusing that Msg is typed as
{Tag, Id, msg()}, and then later we have

 Warnings :: [{Tag, Id, Msg}]
 Msg = msg()

odiferousmint:
> manual.txt hasn't been updated for many years. I trust @kostis will
> protest if the file is to be kept as is.
> 
> I'm not sure I understand what it is that you prefer. Is it "Msg"
> rather than "Warning"? I find it confusing that Msg is typed as
> {Tag, Id, msg()}, and then later we have
> 
> Warnings :: [{Tag, Id, Msg}]
> Msg = msg()

I was just curious of the state of the manual.txt. As for what I am preferring: displaying the parameter(s) and the return type over displaying only the arity, but I may have confused as to where it would be shown. In case I did, I apologize.

PR #3026, opened 2021-02-02
Add end_location annotations for parse trees

richcarl:
Based on https://github.com/erlang/otp/pull/3020.

This allows the quoted source code in error messages to underline the range of the error, and can also be used by tools that display source code, as in an IDE.

Example:
```
foo.erl:18:16: illegal guard expression
%   18| foo(A, B) when is_lit(A), is_list(B) ->
%     |                ^^^^^^
```
instead of just
```
foo.erl:18:16: illegal guard expression
%   18| foo(A, B) when is_lit(A), is_list(B) ->
%     |                ^
```
This has only been implemented on the parser level. Adding further support for the scanner to annotate all tokens with end position right away would save the time now spent calculating end positions of tokens by taking the length of the text.

### DIFF
--- a/lib/stdlib/src/erl_anno.erl
+++ b/lib/stdlib/src/erl_anno.erl
@@ -24,7 +24,7 @@
 -export([column/1, end_location/1, file/1, generated/1,
          line/1, location/1, record/1, text/1]).
 -export([set_file/2, set_generated/2, set_line/2, set_location/2,
-         set_record/2, set_text/2]).
+         set_end_location/2, set_record/2, set_text/2]).
 
 %% To be used when necessary to avoid Dialyzer warnings.
 -export([to_term/1, from_term/1]).
@@ -144,6 +144,10 @@ is_anno2(location, Line) when ?LN(Line) ->
     true;
 is_anno2(location, {Line, Column}) when ?LN(Line), ?COL(Column) ->
     true;
+is_anno2(end_location, Line) when ?LN(Line) ->
+    true;
+is_anno2(end_location, {Line, Column}) when ?LN(Line), ?COL(Column) ->
+    true;
 is_anno2(generated, true) ->
     true;
 is_anno2(file, Filename) ->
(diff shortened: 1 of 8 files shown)


lib/stdlib/src/erl_anno.erl
```diff
@@ -24,7 +24,7 @@
 -export([column/1, end_location/1, file/1, generated/1,
          line/1, location/1, record/1, text/1]).
 -export([set_file/2, set_generated/2, set_line/2, set_location/2,
-         set_record/2, set_text/2]).
+         set_end_location/2, set_record/2, set_text/2]).
```
gomoripeti, line 27: Dear OTP team,

Allowing `end_location` key as a first class citizen in `erl_anno:anno()` and the addition of `erl_anno:set_end_location/2` would be beneficial in itself.
I started to need this since https://github.com/erlang/otp/commit/f30514e3232f63563251a42334912aef8c79d4e8#diff-9f3b412e8fbd809c6473fcd5bee30f63931c7349010640d76316cfcde504e733
where `erl_syntax:set_pos` started to require a proper `erl_anno:anno()` instead of any term.
Even if the parser or OTP tools don't use it, other tools could store end_location annotation (without text).
This is not a major issue, but only to make dialyzer and `erl_anno:is_anno/1` happy. 

Would the OTP team be open to include only the `erl_anno` changes of this PR in OTP 24 or is it too late?
(minor benefit of including the set_pos and this change in the same release would be that there is a way before OTP 24 and from OTP24 to store start and end location in `erl_syntax:set_pos`, without a gap)

garazdawi: We are not certain that `end_location` is the best way to represent ranges in the `AST`, so we do not want to make a commitment to that design before we know.

gomoripeti: makes sense, thanks for the response
(out of curiosity what alternatives or drawbacks of end_location came up? I can think of text for all abstract format, not just some leafs)

garazdawi: I don't recall what the exact reasons were, but the alternative approach that we have been talking about is including the text that a node represents in `anno`, or something like that. We are not actively working on this, so my memory is a bit rusty.

PR #3037, opened 2021-02-06
Ease static analysis of parse_transform implementations

paulo-ferraz-oliveira:
The goal of this pull request is to provide a mechanism to ease static analysis (e.g. `xref` and `dialyzer`) for implementations of `parse_transform`s, as well as have in/out-source documentation for that.

To this end:
1. I add a new `parse_transform.erl` module,
2. I add a documentation page for this module,
3. I reference this documentation from `compile.xml`.

What's missing?
- [ ] understanding if this is something we want in OTP,
- [ ] understanding if the behaviour-defining module name is acceptable,
- [ ] understanding if the documentation should be in the `compiler` namespace
- [ ] understanding if the new behaviour should be declared in modules like `merl_transform`, `ms_transform`, ...

**Note** on the motivation for this: I was going through some lib.s that implement parse transformations, looking at OTP 24, and figured that the recent change to the compiler's line v. line/col output would be sometimes difficult to achieve, but `dialyzer` could help if there were a well defined behaviour for this, and also not to have to add `rebar3`-specific `ignore_xref` where a parse_transform is defined.

### DIFF
--- a/lib/compiler/doc/src/Makefile
+++ b/lib/compiler/doc/src/Makefile
@@ -31,7 +31,7 @@ APPLICATION=compiler
 # Target Specs
 # ----------------------------------------------------
 XML_APPLICATION_FILES = ref_man.xml
-XML_REF3_FILES = compile.xml
+XML_REF3_FILES = compile.xml parse_transform.xml
 EDOC_REF3_FILES =  cerl.xml cerl_trees.xml cerl_clauses.xml
 
 XML_PART_FILES = internal.xml
@@ -48,5 +48,7 @@ XML_FILES = \
 
 $(XMLDIR)/%.xml: ../../internal_doc/%.md $(ERL_TOP)/make/emd2exml
 	$(ERL_TOP)/make/emd2exml $< $@
+
+NO_CHUNKS = parse_transform.xml
 
 include $(ERL_TOP)/make/doc.mk
--- a/lib/compiler/doc/src/compile.xml
+++ b/lib/compiler/doc/src/compile.xml
@@ -442,7 +442,9 @@ module.beam: module.erl \
           <item>
             <p>Causes the parse transformation function
               <c>Module:parse_transform/2</c> to be applied to the
-              parsed code before the code is checked for errors.</p>
+              parsed code before the code is checked for errors.
+              <c>Module</c> must implement behaviour
+              <seeerl marker="parse_transform">parse_transform</seeerl></p>
           </item>
 
 	  <tag><c>from_abstr</c></tag>
--- a/lib/compiler/doc/src/parse_transform.xml
+++ b/lib/compiler/doc/src/parse_transform.xml
@@ -1,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE erlref SYSTEM "erlref.dtd">
+
+<erlref>
+  <header>
+    <copyright>
+      <year>2021</year>
+      <holder>Ericsson AB, All Rights Reserved</holder>
+    </copyright>
+    <legalnotice>
+      Licensed under the Apache License, Version 2.0 (the "License");
+      you may not use this file except in compliance with the License.
+      You may obtain a copy of the License at
+
+          http://www.apache.org/licenses/LICENSE-2.0
+
+      Unless required by applicable law or agreed to in writing, software
+      distributed under the License is distributed on an "AS IS" BASIS,
+      WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+      See the License for the specific language governing permissions and
+      limitations under the License.
+
+      The Initial Developer of the Original Code is Ericsson AB.
+    </legalnotice>
+    <title>parse_transform</title>
+    <prepared></prepared>
+    <docno></docno>
+    <date></date>
+    <rev></rev>
+  </header>
+  <module since="">parse_transform</module>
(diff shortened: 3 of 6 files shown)


paulo-ferraz-oliveira:
Having issues compiling since `parse_transform` (the new behaviour) is not available when some modules are compile (e.g. `merl_transform`, `ms_transform`, ...).

paulo-ferraz-oliveira:
Hi. Apart from the now-existing conflicts (I may rebase on upstream/`master`) is this something we want? (initial description has open tasks that can serve as reference on how to move forward)

paulo-ferraz-oliveira:
@uabboli, didn't want to bother you while you were preparing OTP 24, but now that it's out, I'd like to continue working on this, if deemed interesting.

uabboli:
I'm not sure about this. I'll come back to you when we have discussed the idea at OTP.

uabboli:
We've had a meeting where we discussed your ideas. I'm a bit on the
negative side since I don't think Dialyzer can help very much when it
comes to column numbers and the like. Its analyses are simply too
weak. (I've used the DEBUG flag in erl_anno and erl_parse to find
violations of the erl_anno abstraction.)

But there were other opinions on the meeting, and I'll leave this
issue open in case anyone would like to add something.

paulo-ferraz-oliveira:
Cool. Thanks, @uabboli.

Even if it doesn't help very much with static analysis (I'm sure that if you write a callback that just returns `ok` there should be a warning), I still believe it's important documentation on how a callback for a parse transformation should be implemented (in terms of spec): I had to search around since there was no "official" documentation, which is what primarily lead me to this.

I might have been misleading in my initial description (it's part of the motivation, not the whole). It's not about error_location, line or column, but the fact that there's no documentation, and I believe callback expectations should be well documented.

If you decide it should be trimmed down/updated I can also work in that sense. I'm curious though, as to why you're "on the negative side": is it because it increases maintenance cost? Is it because you think it serves no purpose? Is it just because you feel that "regarding `dialyzer`" it adds no value?

paulo-ferraz-oliveira:
Actually, if you try to find "Erlang parse_transform" you get almost no official documentation. There's a reference to https://erlang.org/doc/man/erl_id_trans.html, though, and I know of https://erlang.org/doc/man/compile.html#parse-transformations (which leads to the previous).

Another useful use case might be that static analysis understands that a given module implements a parse transformation, for example.

josevalim:
A new documentation manual could indeed be useful and it doesn't require changing the source code or adding new behaviours, right? Perhaps `erl_id_transform` could even be removed once a manual is added.

paulo-ferraz-oliveira:
@josevalim, I also agree with the documentation thing. I'm a novice at it, though, so would need some guidance to properly tackle it. 

Also, the thing with the behaviour, for me, is that it's a declaration of intent, and it kinda binds the Erlang implementation to it (for better and for worse). As as example, we can, surely, work with a non-`gen_server` implementation for a `gen_server` (as long as you have the proper callbacks declared) but it helps identifying a module as such.

KennethL:
As mentioned earlier by @uabboli we had a meeting in the OTP team discussing this and the outcome was that we don't want to introduce a parse_transform behaviour right now because:

We are not convinced that this will make much of a difference when it comes to checking by tools like Dialyzer.
On the plus side this could possible give better and easier to find documentation for parse transforms.
Since we are planning to redesign how behaviours and callbacks are specified (with an EEP first) we want to wait for that before eventually specifying a parse_transform behaviour. 

Another aspect is that we think the whole concept of parse_transforms needs to be improved/refined since it creates problems in the build process (dependencies, code_path, etc).

I agree about that the documentation about parse_transforms is spares and hard to find. There is a `parse_transform` option to the function compile:file/2 and I think the documentation can be extended there and there should be a link to https://erlang.org/doc/man/compile.html#parse-transformations in the same module. I even think the description in erl_id_trans can be moved there.

paulo-ferraz-oliveira:
Hi, @KennethL. Thank you for your reply.

> We are not convinced that this will make much of a difference when it comes to checking by tools like Dialyzer.

Sure.

> On the plus side this could possible give better and easier to find documentation for parse transforms.

Also in my interests, as I replied more recently.

> Since we are planning to redesign how behaviours and callbacks are specified (with an EEP first) we want to wait for that before eventually specifying a parse_transform behaviour.

Great. I'll keep 👀.

> Another aspect is that we think the whole concept of parse_transforms needs to be improved/refined since it creates problems in the build process (dependencies, code_path, etc).

It does, but while we're not there I though this "documentation" part would actually help.

> I agree about that the documentation about parse_transforms is spares and hard to find. There is a parse_transform option to the function compile:file/2 and I think the documentation can be extended there and there should be a link to https://erlang.org/doc/man/compile.html#parse-transformations in the same module. I even think the description in erl_id_trans can be moved there.

Not all might be lost, then. I'll try to look at what you mentioned in further detail and probably open a pull request live to tackle that (no code, only documentation). Is this reasonable?

KennethL:
Yes a PR with only improvements regarding the documentation will be appreciated

paulo-ferraz-oliveira:
Just adding this here for future memory (I tend to forget stuff): a behaviour would also allow for static analysis tools to work-around certain analysis issues (today, `hank` warned me that the second argument of `parse_transform/2` wasn't required, in `inaka/mixer`).

paulo-ferraz-oliveira:
I haven't had time to look at this recently and it has since been removed from my priorities. If I get back to something like this I'll base myself on the PR and comments. Thanks, everyone.

PR #3038, opened 2021-02-08
dialyzer: Allow suppressing localized underspecs

paulo-ferraz-oliveira:
(opening as a draft, since I also asked a question in https://bugs.erlang.org/browse/ERL-1480)

I'm not fully aware of all the implications of suppressing that warning just in this fashion, but my local tests yielded the results I expected (no warning).


paulo-ferraz-oliveira:
Linking the issue, here, for reference. Closes #4033.

PR #3041, opened 2021-02-08
Do not centralize type/cname lookups through inet_db

josevalim:
Before this patch, inet_res:getbyname lookups
were centralized through the inet_db process, which
made it a bottleneck on concurrent machines that need
to establish many connections to several hosts quickly.

Since the lookups are now concurrent, the update of
resolve.conf has been changed to use atomic operations
that replace the list of servers and list of names.

The inet_db API has also been cleaned up to no longer
expose operations that delete all searches/namespaces/etc.
Those functions seem safe to remove as they are not
used anywhere within Erlang/OTP. The atomic replace
operations can be made public if desired in the future.
There are also other unused functions which we can
remove in the future.

This is a regen of #2818 and follow up to #2891.

### DIFF
--- a/lib/kernel/src/inet_db.erl
+++ b/lib/kernel/src/inet_db.erl
@@ -35,10 +35,10 @@
 -export([start/0, start_link/0, stop/0, reset/0, clear_cache/0]).
 -export([add_rr/1,add_rr/5,del_rr/4]).
 -export([add_ns/1,add_ns/2, ins_ns/1, ins_ns/2,
-	 del_ns/2, del_ns/1, del_ns/0]).
--export([add_alt_ns/1,add_alt_ns/2, ins_alt_ns/1, ins_alt_ns/2, 
-	 del_alt_ns/2, del_alt_ns/1, del_alt_ns/0]).
--export([add_search/1,ins_search/1,del_search/1, del_search/0]).
+	 del_ns/2, del_ns/1]).
+-export([add_alt_ns/1,add_alt_ns/2, ins_alt_ns/1, ins_alt_ns/2,
+	 del_alt_ns/2, del_alt_ns/1]).
+-export([add_search/1,ins_search/1,del_search/1]).
 -export([set_lookup/1, set_recurse/1]).
 -export([set_socks_server/1, set_socks_port/1, add_socks_methods/1,
 	 del_socks_methods/1, del_socks_methods/0,
@@ -166,9 +166,6 @@ del_ns(IP) ->
 del_ns(IP, Port) ->
     call({listop, nameservers, del, {IP,Port}}).
 
-del_ns() -> 
-    call({listdel, nameservers}).
-
 %% ALTERNATIVE NAME SERVER
 %% add to the end of name server list
 add_alt_ns(IP) -> 
@@ -188,9 +185,6 @@ del_alt_ns(IP) ->
 del_alt_ns(IP, Port) ->
     call({listop, alt_nameservers, del, {IP,Port}}).
 
-del_alt_ns() -> 
-    call({listdel, alt_nameservers}).
-
 %% add this domain to the search list
 add_search(Domain) when is_list(Domain) -> 
     call({listop, search, add, Domain}).
@@ -200,9 +194,6 @@ ins_search(Domain) when is_list(Domain) ->
 
 del_search(Domain) ->
     call({listop, search, del, Domain}).
-
-del_search() ->
-    call({listdel, search}).
 
 %% set host name used by inet
 %% Should only be used by inet_config at startup!
@@ -706,7 +697,7 @@ lookup_cname(Domain) ->
 %% Have to do all lookups (changes to the db) in the
 %% process in order to make it possible to refresh the cache.
 lookup_rr(Domain, Class, Type) ->
-    call({lookup_rr, Domain, Class, Type}).
+    match_rr({Domain, Class, Type}).
 
 %%
 %% hostent_by_domain (newly resolved version)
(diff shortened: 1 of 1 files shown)


josevalim:
@sneako if by any chance you could give this branch a try and provide us feedback, it would be fantastic! :heart:

sneako:
I tested out this PR, along with #2891 and unfortunately we saw some instability in our cluster as load increased. This instability was accompanied by many `eaddrinuse` errors in our logs.

This morning, I deployed #2891 on its own to test it in isolation. I let it run in production, on a single node, for over an hour under what we consider to be very high load, and it performed well with no issues at all. Therefore it is possible that there is an issue with this PR, and further testing should be done before it is merged. 

However, I am not 100% certain that the instabilities we saw in our application were caused solely by this PR, and we have made several significant improvements to the application in the meantime, so I would like to run another test soon.

In summary, I think we should hold off on merging this PR for now, but I will report back again after my next test which I plan to run very soon.

RaimoNiskanen:
Since practically the only change in this PR is to change cache lookups from server calls to ETS lookups, it seems unlikely that it could cause `eaddrinuse` errors almost anywhere.  So I hope you will find out that it was some other change that was the culprit...

RaimoNiskanen:
@josevalim: Should we maybe think about caching not only gethostbyname and gethostbyaddr lookups, but all record types?  I was under the impression that `inet_res` mainly was used when you have to look up other than IN,A, IN,AAAA and IN,PTR records. (?).

RaimoNiskanen:
A question to these and all users of `inet_res`: is it used on the public Internet or only in "safe" environments?  And why are you using `inet_res` instead of the native resolver (`inet_gethost_native` from `inet`)?

sneako:
Right, I agree that this PR probably did not cause the eaddrinuse errors.

In my case, we are using `inet_res` instead of `inet_gethost_native` because we found `inet_gethost_native` became a bottleneck in our application frequently causing message queue pile ups. We are running in Kubernetes, and found `inet_res` to be much faster. Since we switched to `inet_res` and started running [node local dns pods](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/), we stopped experiencing these message pile ups.

For example, here is the difference we see on our Kubernetes pods:
```
iex(app@host)1> for _ <- 1..50, do: fn -> :inet_gethost_native.gethostbyname('erlang.org') end |> :timer.tc |> elem(0)
[452, 739, 522, 439, 424, 324, 365, 308, 271, 303, 321, 490, 373, 309, 326, 528,
 392, 294, 365, 334, 286, 319, 332, 397, 295, 293, 275, 260, 372, 348, 317, 272,
 351, 277, 479, 442, 387, 419, 804, 372, 302, 386, 360, 590, 415, 300, 287, 538,
 389, 1788]

iex(app@host)2> for _ <- 1..50, do: fn -> :inet_res.gethostbyname('erlang.org') end |> :timer.tc |> elem(0)
[55, 28, 38, 53, 51, 20, 26, 27, 88, 46, 26, 12, 20, 34, 14, 35, 39, 84, 27, 34,
 20, 20, 29, 28, 46, 69, 46, 44, 20, 11, 37, 33, 21, 18, 50, 36, 19, 46, 89, 39,
 43, 38, 39, 32, 32, 59, 32, 43, 34, 58]
```

RaimoNiskanen:
@sneako: Well, I'll be darned! For speed reasons!  In that benchmark you will get almost all answers served from the `inet_db` cache, but I guess `inet_res` indeed improves actual performance.  Maybe also largely thanks to the cache?

So this means that you ignore the platform's resolver configuration and instead ask DNS. (?)
Are you using the `inet` framework https://erlang.org/doc/apps/erts/inet_cfg.html and set the lookup order to e.g `[file, dns]`, or are you calling inet_res directly?

I was thinking about introducing caching in the `inet` framework instead of only for `inet_res` so `inet_gethost_native` could benefit from caching.

sneako:
Hehe yes, we have a need for speed 😄  We are using an inet_cfg file with `{lookup, [dns, native]}.` set in order to use `inet_res`, and do not call it directly.

I definitely like the idea of adding caching for `inet_gethost_native` so that you can simply rely on the system dns, but still avoid the syscalls when a result is cached.

sneako:
I am running another test of this PR right now. So far things are looking good and it seems that our other recent changes have fixed the `eaddrinuse` issue we saw last time. I'm going to let this run for a while a keep increasing the load on this test node, and report back. In the meantime, checkout how fast `inet_res` is now that we skip the gen_server call:

```
> for _ <- 1..50, do: fn -> :inet_res.gethostbyname('erlang.org') end |> :timer.tc |> elem(0)
[17, 8, 27, 6, 10, 7, 14, 4, 3, 4, 3, 3, 3, 3, 4, 6, 14, 12, 8, 6, 7, 4, 4, 18,
 4, 10, 5, 5, 5, 4, 6, 4, 4, 5, 9, 4, 4, 3, 12, 4, 4, 4, 3, 3, 4, 6, 3, 4, 4, 3]
```

sneako:
I have been running this patch in production now for almost 2 hours on one node in our cluster. Using weighted routing, I have been sending extra traffic to this node, keeping the schedulers around 90% saturated, and performance has been very good! It appears that our other recent changes have fixed the `eaddrinuse` errors we saw last time we tried out this patch.

RaimoNiskanen:
Gosh!  Mean time went down from 38 µs to 6 µs.  That is fast!

> I definitely like the idea of adding caching for inet_gethost_native so that you can simply rely on the system dns, but still avoid the syscalls when a result is cached.

I was thinking about adding caching to `inet` for all resolver backends.  I hope it would not be too large a change...

RaimoNiskanen:

PR #3043, opened 2021-02-08
jit: Optimized code generation for select_val

frej:
Two patches optimizing the code generation for select_val:

* Optimize the case when a select_val has exactly two choices which branch to the same label or a fail destination and where the values only differ by one bit.

* The i_select_val_bins instruction has previously been implemented by a call to a global subroutine which does a binary search in a table. This patch changes the code generator to emit the binary search as code.


garazdawi (approved):
lgtm

bjorng:
Thanks for your pull request.

PR #3044, opened 2021-02-08
Add more checking for dynamic M:F(A)

TD5:
This involves two parts:
1. Make Dialyzer check M:F(A) if there is more than one module atom inferred by the success typing
2. Make Dialyzer reject calls where a module exists but the function doesn't (i.e. we know for sure the function is not implemented)

There's still some cruft here to clean up, including some comments that need to be deleted/resolved and some whitespace changes that I should probably undo.


TD5:
Oops, didn't mean to put this here yet

PR #3045, opened 2021-02-08
Error message types

richcarl:
Followup based on https://github.com/erlang/otp/pull/3020.
Makes the compiler use the type declarations in erl_lint as the source of truth for how errors and warnings are represented.
(A minor snag was that the core_lint pass returned a module name (an atom) as the "filename", and the file:filename() type doesn't officially include atoms even if they have always worked.

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -55,10 +55,10 @@
 
 -type option() :: atom() | {atom(), term()} | {'d', atom(), term()}.
 
--type err_info() :: {erl_anno:line() | 'none',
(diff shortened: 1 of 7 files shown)


bjorng:
If this pull request is still relevant, please rebase it on the latest master.

richcarl:
Fixed now.

bjorng:
There are failed test cases in the `core_SUITE` and `core_fold_SUITE` modules for the `compiler` application.

bjorng:
The types look good to me.

richcarl:
Fixed test cases.

bjorng:
Thanks for your pull request.

PR #3049, opened 2021-02-09
ssh: Add longname field  in sftp version 3

jonatascardador:
I encountered some problems using ssh_sftpd erlang because some sftp clients use the long name field in SSH_FXP_NAME response to display file information in sftp version 3. Looking at documentation ([https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02](https://tools.ietf.org/html/draft-ietf-secsh-filexfer-02)), this field is an expanded format for the file name and although not specified by the protocol, the recommended format is similar to what is returned by "ls -l" on Unix systems.
I've made some changes to ssh_xfer and ssh_sftpd modules to add this field in the response for version 3 and it solved the issue for the sftp client I was using, if you could consider taking a look at this PR, I would really appreciate! I tried as much as I could to follow the contribution guide.


HansN:
Thanks for the PR!

PR #3050, opened 2021-02-10
dialyzer: Clarify how to declare records used in match patterns

uabboli:
See also https://bugs.erlang.org/browse/ERL-892.

### DIFF
--- a/system/doc/reference_manual/typespec.xml
+++ b/system/doc/reference_manual/typespec.xml
@@ -467,6 +467,33 @@
       Any unspecified fields are assumed to have the type in the original
       record declaration.
     </p>
+    <note>
+      <p>When records are used to create patterns for ETS and
+        Mnesia match functions, Dialyzer may need some help not to emit
+        bad warnings. For example:</p>
+
+<pre>
+    -type height() :: pos_integer().
+    -record(person, {name :: string(), height :: height()}).
+
+    lookup(Name, Tab) ->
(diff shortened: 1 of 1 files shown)


mheiber:
I'm glad this is being discussed, I think it's important to have a standard way of dealing with records used in match patterns when using Dialyzer.

What about match variables like `'$1'` `'$2'` etc.?

uabboli:
You mean the example should be extended to also include pattern variables?

mheiber:
> You mean the example should be extended to also include pattern variables?

Thanks for following up, fwiw I like your idea of extending the example.

I also question the scalability of the approach for code that has a lot of pattern variables, and wonder if a tweak to Dialyzer itself might be better. But this PR is a big improvement over the status quo so I'd really like to see it get in.

mheiber (approved):
Happy to see documentation for how to solve this common problem.

uabboli:
> wonder if a tweak to Dialyzer itself might be better.

It's not so easy to find a good tweak. Dialyzer needs to somehow know
that the constructed record is be used in match patterns, and nowhere
else.

The example, "ets:match_object(Tab, #person{name = Name, _ = '_'})",
can possibly be handled without much difficulty, but in general I
think types need to be tagged as match patterns internally in
Dialyzer. This resembles how opaque types are implemented, and taking
one step further in that direction would IMHO be a mistake.

But perhaps there is a smart tweak, but until it's found, I think
documentation is the best option.

Issue #4192, opened 2021-02-02
ERL-1476: xref erroneously claims M:behaviour_info/1 is undefined


OTP-Maintainer:
Original reporter: `mikael pettersson`
Affected versions: `OTP-24.0`, `OTP-23.2`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1476

---

```
Upgrading from OTP-23.1.5 to OTP-23.2.3 caused our xref checks to fail, with complaints that calls to M:behaviour_info/1 are undefined, even though they provably are not. A git bisect identified the culprit:

{noformat}
# first bad commit: [d74e9b343a8562aa2ffceb07ad44a1fa8fd45f48] Suppress xref output for unused export on behaviour's behaviour_info/1
{noformat}

That's [https://github.com/erlang/otp/pull/2752|https://github.com/erlang/otp/pull/2752.] aka https://bugs.erlang.org/browse/ERL-1353, which apparently wanted to prevent xref from reporting compiler-generated behaviour_info/1 functions as unused exports. Fine, but the implementation appears too crude, resulting in the problem reported here.

Repeat-by:
{noformat}
> cat a.erl
-module(a).
-export([foo/1]).
-callback foo() -> ok.
foo(Mod) -> Mod:foo().
> cat b.erl
-module(b).
-export([bar/0]).
bar() -> a:behaviour_info(callbacks).
> cat x.erl
-module(x).
-export([ref/0]).
ref() ->
  {ok, XREF} = xref:start([]),
  ok = xref:set_default(XREF, [{warnings, false}]),
  {ok, _} = xref:add_module(XREF, "a.beam"),
  {ok, _} = xref:add_module(XREF, "b.beam"),
  {ok, UndefCalls} = xref:analyze(XREF, undefined_function_calls),
  case UndefCalls of
    [] -> io:format(standard_io, "No undefined function calls\n", []);
    _ -> io:format(standard_io, "Undefined function calls: ~p\n", [UndefCalls])
  end.
{noformat}
{{a}} implements a behaviour, {{b}} a test case, and {{x}} an {{xref}} wrapper. Let's run with OTP-23.2.3:
{noformat}
> erlc +debug_info a.erl b.erl x.erl
> erl -noshell -s x ref -s erlang halt
Undefined function calls: [{{b,bar,0},{a,behaviour_info,1}}]
{noformat}
This is is clearly wrong. The result is the same with current master. A re-run with OTP-23.1.5 shows the expected result:
[...]

PR #4507, opened 2021-02-12
tools: Correct Xref's handling of behaviour_info/1

uabboli:
Solves #4192.

### DIFF
--- a/lib/tools/src/xref_reader.erl
+++ b/lib/tools/src/xref_reader.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 2000-2017. All Rights Reserved.
+%% Copyright Ericsson AB 2000-2020. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lib/tools/src/xref_reader.erl
```diff
@@ -95,16 +96,19 @@ form({function, _, module_info, 0, _Clauses}, S) ->
...
 form({function, Anno, Name, Arity, Clauses}, S) ->
-    MFA0 = {S#xrefr.module, Name, Arity},
-    MFA = adjust_arity(S, MFA0),
-    S1 = S#xrefr{function = MFA},
-    Line = erl_anno:line(Anno),
-    S2 = S1#xrefr{def_at = [{MFA,Line} | S#xrefr.def_at]},
-    S3 = clauses(Clauses, S2),
-    S3#xrefr{function = []};
+    case {Name, Arity, erl_anno:location(Anno)} of
+        {behaviour_info, 1, 0} ->
```
paulo-ferraz-oliveira, line 101: If I get `erl_internal.erl` correctly, `location` is hard-coded as `0`, here: `[{function,0,behaviour_info,1,`.
I'm testing this version (using `:location`), since I don't understand how it could cause a difference in behaviour (but probably there's some other piece of the puzzle I'm missing).

paulo-ferraz-oliveira: Tested it and it's working as intended, in both the shell-based `xref` and the code-based `xref`.

uabboli: The change in xref_reader is no big deal. Matching an annotation with the integer zero breaks the erl_anno abstraction, that's all.

uabboli:
Merged commit ba9f40fb710f3f7b908b045e62ffd07b4186dcc3 to erlang:maint-23 (Erlang OTP 23.2.5).

PR #4508, opened 2021-02-12
Replace uses of maps:map and maps:fold with maps:foreach

michalmuskala:


lib/kernel/src/net_kernel.erl
```diff
@@ -928,22 +928,18 @@ handle_info({From,badcookie,_To,_Mess}, State) ->
...
 handle_info(tick, State) ->
     ?tckr_dbg(tick),
-    ok = maps:fold(fun (Pid, _Node, ok) ->
-                          Pid ! {self(), tick},
-                          ok
-                   end,
-                   ok,
-                   State#state.conn_owners),
+    maps:foreach(fun (Pid, _Node) ->
+                     Pid ! {self(), aux_tick}
```
garazdawi, line 932: This should be `tick`, not `aux_tick`.

garazdawi:
Thanks!

PR #4509, opened 2021-02-12
Fix servfail_retry_timeout to only be used when appropriate

RaimoNiskanen:
This PR changes how the `servfail_retry_timeout` is handled in `inet_res`, so that it is only enforced when a server has answered `servfail`, and only when asking that same server again.

The original implementation was too simplistic and defensive in always waiting for the set time before every retry on the name server list.


bmk (approved):

RaimoNiskanen:
I have updated after OTP internal feedback

PR #4516, opened 2021-02-15
Fix incorrect links on erpc docs

ergl:


rickard-green:
Please rebase this fix onto the `maint` branch so we can release the fix in the next maintenance patch

PR #4518, opened 2021-02-16
Update lib/wx/configure.in to throw only when `--with-wx` is specified

dominicletz:
Based on a comment from @garazdawi 
> The way that we reason about configure tests is that if the user has explicitly enabled something, like for instance with --with-wx, then if configure fails to detect support for wx it should fail. However, if the user did not give any option then configure should only warn that the feature is not useable and continue building.

The configure tests in question worked on our builds yesterday and stopped working today, so something has changed to cause this. @dgud thought that this PR was the issue which is why I posted here.

Command | Error handling
-- | --
./configure --with-wx | Fails if wx fails
./configure | Proceeds if wx fails, but logs info message
./configure --with-wxdir=~/wxWidgets-2.8 | Proceeds if wx fails, but logs info message
./configure --with-wx --with-wxdir=~/wxWidgets-2.8 | Fails if wx fails

fyi @dgud


dominicletz:
Comment from @dgud 
>  I don't like that you have used SKIP instead CONF_INFO,
what should be used is CONF_INFO and WX_CAN_BUILD_DRIVER for default builds,
otherwise erlang applications that need wx headers can not be built.
So this solution needs fixes all over place, which might be needed but not in this PR,
mixed gui applications such as observer, debugger needs to be built even if wx is not available.

Ah so lib/wx is a special case atm. Converting all SKIP CONF_INFO to only log warnings but *always* build lib/wx

dgud:
I don't understand why you moved it to a separate PR, fixing observer, debugger, and the other gui applications 
was what meant with "for another PR" :-)
Sorry for confusing you.

If I'm not completely mistaken, which I can be, the same problem applies to this PR, you can't use SKIP for wx, since that will lead to debugger and observer not compiling if there is no wx available and configure is run without any options.

You can only use SKIP if --with-wx is specified.

dgud:
> Ah so lib/wx is a special case atm. Converting all SKIP CONF_INFO to only log warnings but always build lib/wx
Yes, you need to build the beam files (or at least one) otherwise the other applications don't build.

dominicletz:
Ok, that makes sense. Have look now, keep using ./CONF_INFO but still enforcing build error when `--with-wx` is used. I feel the wxwin-nothrow.m4 could be cleaned up more. E.g. remove all other `AC_MSG_ERROR` as well but ensure that the error messages are persisted into ./CONF_INFO ...

dominicletz:
ok @dgud I think i found an elegant solution. I've added a macro `WX_MSG_ERROR` that will make the configure call fail if and only if `--with-wx` is given in all other cases messages go into ./CONF_INFO.
If you want I can merge this back to the wxWebView PR -- for me it's the same

dgud:
wxwin.m4  was copied from wxWidgets src, so I not done any changes to it.

But great to get some help here. looks good to me.

Yeah, if there is no problem move it back to the other branch,  I can't merge the other branch without some
configure fixes.

dominicletz:
done, closing this pr

PR #4519, opened 2021-02-16
Reduce register shuffling

bjorng:
Consider this code:

    send_deadline_cont(SockRef, Bin, Cont, Deadline, HasWritten) ->
        SelectHandle = make_ref(),
        Sent = prim_socket:send(SockRef, Bin, Cont, SelectHandle),
        {SockRef, Bin, SelectHandle, Deadline, HasWritten, Sent}.

Looking closer at the BEAM code between the two calls we see four
`move` instructions followed by three `swap` instructions:

    {call_ext,0,{extfunc,erlang,make_ref,0}}.
    {move,{y,2},{x,3}}.
    {move,{y,3},{x,2}}.
    {move,{x,0},{y,2}}.
    {move,{y,4},{x,1}}.
    {swap,{x,2},{x,1}}.
    {swap,{x,3},{x,2}}.
    {swap,{x,0},{x,3}}.
    {call_ext,4,{extfunc,prim_socket,send,4}}.

It turns out that the sequence of `move` and `swap` instructions is
longer than it needs to be, because the initial two `move`
instructions move the terms in two of the Y registers to the wrong X
registers. Two of the `swap` instructions are needed merely to move
the terms to the right registers.

This commit updates optimizations in `beam_ssa_pre_codegen` and
`beam_block` to eliminate the extra instructions:

    {call_ext,0,{extfunc,erlang,make_ref,0}}.
    {move,{y,2},{x,2}}.
    {move,{y,3},{x,1}}.
    {move,{y,4},{x,3}}.
    {move,{x,0},{y,2}}.
    {swap,{x,0},{x,3}}.
    {call_ext,4,{extfunc,prim_socket,send,4}}.

By counting the number of inserted and deleted lines in the diff between
a run of the `diffable` before this change and a run after this change,
we can see that this commit results in 1099 fewer instructions in total.


lib/compiler/src/beam_ssa_pre_codegen.erl
```diff
@@ -2064,32 +2064,63 @@ copy_retval_is([], RC, _, Copy, Count, Acc) ->
...
 
 place_retval_copy(I, _Yregs, none, Count, Acc) ->
     {I,Count,Acc};
 place_retval_copy(#b_set{args=[F|Args0]}=I, Yregs, Copy, Count0, Acc0) ->
     #b_set{dst=Avoid} = Copy,
-    {Args,Acc1,Count} = copy_func_args(Args0, Yregs, Avoid, Acc0, [], Count0),
+    {Args,Acc1,Count} = copy_func_args(reverse(Args0), Yregs, Avoid, Acc0, [], Count0),
     Acc = [Copy|Acc1],
-    {I#b_set{args=[F|Args]},Count,Acc}.
+    {I#b_set{args=[F|reverse(Args)]},Count,Acc}.
```
jhogberg, line 2123: The base case of `copy_func_args/6` returns `{reverse(Acc),CopyAcc,Count}` and it's only called from here, so why not remove both reversals?

jhogberg (approved):
LGTM.

lib/compiler/src/beam_ssa_pre_codegen.erl
bjorng: That will not work. (Hint: The `reverse/1` at line 2123 undoes the `reverse/1` at line 2121.) One day I might clean up those functions and their creepy coupling.

jhogberg: Sorry about the confusion, I meant the reversals at lines 2123 and 2137 (base case), not 2121 and 2123.

bjorng: You are right, that will work. I have done that simplification, along with another simplification and renaming of variables to make it clearer what is happening.

PR #4521, opened 2021-02-16
supervisor: add restart type "intrinsic"

juhlig:
This PR introduces a new restart type `intrinsic` for supervisors, inspired by [RabbitMQ's `supervisor2`](https://github.com/rabbitmq/rabbitmq-server/blob/master/deps/rabbit_common/src/supervisor2.erl).

* Like the `transient` restart type, `intrinsic` childs are restarted as usual if they exit abnormally, or if the supervisor's restart strategy is `one_for_all` or `rest_for_one` and a sibling's death causes the restart of `intrinsic` childs.
* However, if an `intrinsic` child exits normally (with exit reason `normal`, `shutdown` or `{shutdown, Term}`), the child's supervisor also shuts down normally.
* Terminating an `intrinsic` child via `supervisor:terminate_child/2` does _not_ cause the supervisor to shut down.

### DIFF
--- a/lib/stdlib/doc/src/supervisor.xml
+++ b/lib/stdlib/doc/src/supervisor.xml
@@ -192,7 +192,15 @@ child_spec() = #{id => child_id(),       % mandatory
           death causes the temporary process to be terminated).
           A <c>transient</c> child process is restarted only if
           it terminates abnormally, that is, with another exit reason
-          than <c>normal</c>, <c>shutdown</c>, or <c>{shutdown,Term}</c>.</p>
+          than <c>normal</c>, <c>shutdown</c>, or <c>{shutdown,Term}</c>.
+          An <c>intrinsic</c> child process is restarted if it terminates
+          abnormally or if the supervisor's restart strategy is 
+          <c>rest_for_one</c> or <c>one_for_all</c> and a sibling's
+          death causes it to terminate. However, if any <c>intrinsic</c>
+          child process terminates normally, that is, with an exit reason
+          of <c>normal</c>, <c>shutdown</c> or <c>{shutdown,Term}</c>,
+          the <c>intrinsic</c> child's supervisor shuts down all remaining
+          childs and then also terminates normally.</p>
         <p>The <c>restart</c> key is optional. If it is not specified,
           it defaults to <c>permanent</c>.</p>
       </item>
--- a/lib/stdlib/src/supervisor.erl
+++ b/lib/stdlib/src/supervisor.erl
@@ -65,7 +65,7 @@
 -type child_id() :: term().
 -type mfargs()   :: {M :: module(), F :: atom(), A :: [term()] | undefined}.
 -type modules()  :: [module()] | 'dynamic'.
--type restart()  :: 'permanent' | 'transient' | 'temporary'.
+-type restart()  :: 'permanent' | 'transient' | 'intrinsic' | 'temporary'.
 -type shutdown() :: 'brutal_kill' | timeout().
 -type worker()   :: 'worker' | 'supervisor'.
 -type sup_name() :: {'local', Name :: atom()}
@@ -141,6 +141,7 @@
 -define(is_simple(State), State#state.strategy =:= simple_one_for_one).
 -define(is_temporary(_Child_), _Child_#child.restart_type=:=temporary).
 -define(is_transient(_Child_), _Child_#child.restart_type=:=transient).
+-define(is_intrinsic(_Child_), _Child_#child.restart_type=:=intrinsic).
 -define(is_permanent(_Child_), _Child_#child.restart_type=:=permanent).
 
 -callback init(Args :: term()) ->
@@ -706,16 +707,28 @@ restart_child(Pid, Reason, State) ->
 do_restart(Reason, Child, State) when ?is_permanent(Child) ->
     ?report_error(child_terminated, Reason, Child, State#state.name),
     restart(Child, State);
+do_restart(normal, Child, State) when ?is_intrinsic(Child) ->
+    NState = del_child(Child, State),
+    {shutdown, NState};
 do_restart(normal, Child, State) ->
     NState = del_child(Child, State),
     {ok, NState};
+do_restart(shutdown, Child, State) when ?is_intrinsic(Child) ->
+    NState = del_child(Child, State),
+    {shutdown, NState};
 do_restart(shutdown, Child, State) ->
     NState = del_child(Child, State),
     {ok, NState};
+do_restart({shutdown, _Term}, Child, State) when ?is_intrinsic(Child) ->
+    NState = del_child(Child, State),
+    {shutdown, NState};
 do_restart({shutdown, _Term}, Child, State) ->
     NState = del_child(Child, State),
     {ok, NState};
 do_restart(Reason, Child, State) when ?is_transient(Child) ->
+    ?report_error(child_terminated, Reason, Child, State#state.name),
+    restart(Child, State);
+do_restart(Reason, Child, State) when ?is_intrinsic(Child) ->
     ?report_error(child_terminated, Reason, Child, State#state.name),
     restart(Child, State);
 do_restart(Reason, Child, State) when ?is_temporary(Child) ->
@@ -1343,6 +1356,7 @@ validFunc(Func)                      -> throw({invalid_mfa, Func}).
 validRestartType(permanent)   -> true;
 validRestartType(temporary)   -> true;
 validRestartType(transient)   -> true;
+validRestartType(intrinsic)   -> true;
 validRestartType(RestartType) -> throw({invalid_restart_type, RestartType}).
 
 validShutdown(Shutdown)
--- a/lib/stdlib/test/supervisor_SUITE.erl
+++ b/lib/stdlib/test/supervisor_SUITE.erl
@@ -48,14 +48,16 @@
 	  child_adm/1, child_adm_simple/1, child_specs/1, extra_return/1,
 	  sup_flags/1]).
 
-%% Tests concept permanent, transient and temporary 
+%% Tests concept permanent, transient, intrinsic and temporary 
 -export([ permanent_normal/1, transient_normal/1,
-	  temporary_normal/1,
+	  intrinsic_normal/1, temporary_normal/1,
 	  permanent_shutdown/1, transient_shutdown/1,
-	  temporary_shutdown/1,
+	  intrinsic_shutdown/1, temporary_shutdown/1,
           faulty_application_shutdown/1,
 	  permanent_abnormal/1, transient_abnormal/1,
-	  temporary_abnormal/1, temporary_bystander/1]).
+	  intrinsic_abnormal/1, temporary_abnormal/1,
+	  temporary_bystander/1, intrinsic_bystander/1,
+	  intrinsic_terminate/1]).
 
 %% Restart strategy tests 
 -export([ multiple_restarts/1,
@@ -103,9 +105,10 @@ all() ->
      do_not_save_start_parameters_for_temporary_children,
      do_not_save_child_specs_for_temporary_children,
      simple_one_for_one_scale_many_temporary_children, temporary_bystander,
-     simple_global_supervisor, hanging_restart_loop,
-     hanging_restart_loop_rest_for_one, hanging_restart_loop_simple,
-     code_change, code_change_map, code_change_simple, code_change_simple_map,
+     intrinsic_bystander, intrinsic_terminate, simple_global_supervisor,
+     hanging_restart_loop, hanging_restart_loop_rest_for_one,
+     hanging_restart_loop_simple, code_change, code_change_map,
+     code_change_simple, code_change_simple_map,
      order_of_children, scale_start_stop_many_children,
      format_log_1, format_log_2].
 
@@ -126,12 +129,12 @@ groups() ->
       [sup_stop_infinity, sup_stop_timeout,
        sup_stop_brutal_kill]},
      {normal_termination, [],
-      [permanent_normal, transient_normal, temporary_normal]},
+      [permanent_normal, transient_normal, intrinsic_normal, temporary_normal]},
      {shutdown_termination, [],
-      [permanent_shutdown, transient_shutdown, temporary_shutdown,
-       faulty_application_shutdown]},
+      [permanent_shutdown, transient_shutdown,intrinsic_shutdown,
+       temporary_shutdown, faulty_application_shutdown]},
      {abnormal_termination, [],
-      [permanent_abnormal, transient_abnormal,
+      [permanent_abnormal, transient_abnormal, intrinsic_abnormal,
        temporary_abnormal]},
      {restart_one_for_one, [],
       [one_for_one, one_for_one_escalation]},
@@ -690,6 +693,7 @@ child_specs(Config) when is_list(Config) ->
     C3 = {child, {m,f,[a]}, temporary, 1000, worker, dynamic},
     C4 = {child, {m,f,[a]}, transient, 1000, worker, [m]},
     C5 = {child, {m,f,[a]}, permanent, infinity, worker, [m]},
+    C6 = {child, {m,f,[a]}, intrinsic, 1000, worker, [m]},
 
     {error, {invalid_mfa,mfa}} = supervisor:start_child(sup_test, B1),
     {error, {invalid_restart_type, prmanent}} =
@@ -717,6 +721,7 @@ child_specs(Config) when is_list(Config) ->
     ok = supervisor:check_childspecs([C3]),
     ok = supervisor:check_childspecs([C4]),
     ok = supervisor:check_childspecs([C5]),
+    ok = supervisor:check_childspecs([C6]),
 
     {error,{duplicate_child_name,child}} = supervisor:check_childspecs([C1,C2]),
 
@@ -803,6 +808,28 @@ transient_normal(Config) when is_list(Config) ->
     [1,0,0,1] = get_child_counts(sup_test).
 
 %%-------------------------------------------------------------------------
+%% An intrinsic child should not be restarted if it exits with reason
+%% normal, but it should cause it's parent to exit.
+intrinsic_normal(Config) when is_list(Config) ->
+    {ok, SupPid} = start_link({ok, {{one_for_one, 2, 3600}, []}}),
+
+    Child1 = {child1, {supervisor_1, start_child, []}, intrinsic, 1000, worker, []},
+    {ok, CPid1} = supervisor:start_child(sup_test, Child1),
+
+    SupRef = monitor(process, SupPid),
+    unlink(SupPid),
+
+    terminate(SupPid, CPid1, child1, normal),
+
+    receive {'DOWN', SupRef, process, SupPid, shutdown} ->
+	ok
+    after 1000 ->
+	catch link(SupPid),
+	demonitor(SupRef, [flush]),
+	ct:fail({supervisor_intrinsic_not_shutdown, SupPid})
+    end.
+
+%%-------------------------------------------------------------------------
 %% A temporary process should never be restarted.
 temporary_normal(Config) when is_list(Config) ->
     {ok, SupPid} = start_link({ok, {{one_for_one, 2, 3600}, []}}),
@@ -869,6 +896,39 @@ transient_shutdown(Config) when is_list(Config) ->
 
     [{child1,undefined,worker,[]}] = supervisor:which_children(sup_test),
     [1,0,0,1] = get_child_counts(sup_test).
+
+%%-------------------------------------------------------------------------
+%% An intrinsic child should not be restarted if it exits with reason
+%% shutdown or {shutdown,Term}, but it should cause it's parent to exit.
+intrinsic_shutdown(Config) when is_list(Config) ->
+    SupRet = {ok, {{one_for_one, 2, 3600}, []}},
+    Child1 = {child1, {supervisor_1, start_child, []}, intrinsic, 1000, worker, []},
+
+    {ok, SupPid1} = start_link(SupRet),
+    {ok, CPid1} = supervisor:start_child(sup_test, Child1),
+    SupRef1 = monitor(process, SupPid1),
+    unlink(SupPid1),
+    terminate(SupPid1, CPid1, child1, shutdown),
+    receive {'DOWN', SupRef1, process, SupPid1, shutdown} ->
+	ok
+    after 1000 ->
+	demonitor(SupRef1, [flush]),
+	catch link(SupPid1),
+	ct:fail({supervisor_intrinsic_not_shutdown, SupPid1})
+    end,
+
+    {ok, SupPid2} = start_link(SupRet),
+    {ok, CPid2} = supervisor:start_child(sup_test, Child1),
+    SupRef2 = monitor(process, SupPid2),
+    unlink(SupPid2),
+    terminate(SupPid2, CPid2, child1, {shutdown, some_info}),
+    receive {'DOWN', SupRef2, process, SupPid2, shutdown} ->
+	ok
+    after 1000 ->
+	demonitor(SupRef2, [flush]),
+	catch link(SupPid2),
+	ct:fail({supervisor_intrinsic_not_shutdown, SupPid2})
+    end.
 
 %%-------------------------------------------------------------------------
 %% A temporary process should never be restarted.
@@ -963,6 +1023,25 @@ transient_abnormal(Config) when is_list(Config) ->
     [1,1,0,1] = get_child_counts(sup_test).
 
 %%-------------------------------------------------------------------------
+%% An intrinsic child should be restarted if it exits with reason abnormal.
+intrinsic_abnormal(Config) when is_list(Config) ->
+    {ok, SupPid} = start_link({ok, {{one_for_one, 2, 3600}, []}}),
+    Child1 = {child1, {supervisor_1, start_child, []}, intrinsic, 1000,
+	      worker, []},
+
+    {ok, CPid1} = supervisor:start_child(sup_test, Child1),
+    terminate(SupPid, CPid1, child1, abnormal),
+
+    [{child1, Pid ,worker,[]}] = supervisor:which_children(sup_test),
+    case is_pid(Pid) of
+	true ->
+	    ok;
+	false ->
+	    ct:fail({intrinsic_child_not_restarted, Child1})
+    end,
+    [1,1,0,1] = get_child_counts(sup_test).
+
+%%-------------------------------------------------------------------------
 %% A temporary process should never be restarted.
 temporary_abnormal(Config) when is_list(Config) ->
     {ok, SupPid} = start_link({ok, {{one_for_one, 2, 3600}, []}}),
@@ -1003,6 +1082,53 @@ temporary_bystander(_Config) ->
     %% Child2 has not been restarted
     [{child1, _, _, _}] = supervisor:which_children(SupPid1),
     [{child1, _, _, _}] = supervisor:which_children(SupPid2).
+
+%%-------------------------------------------------------------------------
+%% An intrinsic process killed as part of a rest_for_one or one_for_all
+%% restart strategy should not cause the parent to exit.
+intrinsic_bystander(_Config) ->
+    Child1 = {child1, {supervisor_1, start_child, []}, permanent, 100,
+	      worker, []},
+    Child2 = {child2, {supervisor_1, start_child, []}, intrinsic, 100,
+	      worker, []},
+    {ok, SupPid1} = supervisor:start_link(?MODULE, {ok, {{one_for_all, 2, 300}, []}}),
+    {ok, SupPid2} = supervisor:start_link(?MODULE, {ok, {{rest_for_one, 2, 300}, []}}),
+    {ok, CPid1} = supervisor:start_child(SupPid1, Child1),
+    {ok, _CPid2} = supervisor:start_child(SupPid1, Child2),
+    {ok, CPid3} = supervisor:start_child(SupPid2, Child1),
+    {ok, _CPid4} = supervisor:start_child(SupPid2, Child2),
+    unlink(SupPid1),
+    unlink(SupPid2),
+    terminate(SupPid1, CPid1, child1, normal),
+    terminate(SupPid2, CPid3, child1, normal),
+    timer:sleep(350),
+    catch link(SupPid1),
+    catch link(SupPid2),
+    %% The supervisors should not have exited as a cause of the intrinsic
+    %% childs restarting
+    true = erlang:is_process_alive(SupPid1),
+    true = erlang:is_process_alive(SupPid2),
+    %% Both childs should have been restarted
+    [{child2, _, _, _}, {child1, _, _, _}] = supervisor:which_children(SupPid1),
+    [{child2, _, _, _}, {child1, _, _, _}] = supervisor:which_children(SupPid2).
+
+%%-------------------------------------------------------------------------
+%% An intrinsic process terminated via supervisor:terminate_child should
+%% not cause the parent to exit.
+intrinsic_terminate(_Config) ->
+    Child1 = {child1, {supervisor_1, start_child, []}, intrinsic, 1000,
+	      worker, []},
+    {ok, SupPid} = start_link({ok, {{one_for_one, 2, 3600}, []}}),
+    {ok, _CPid1} = supervisor:start_child(sup_test, Child1),
+    unlink(SupPid), % otherwise we crash with it
+    ok = supervisor:terminate_child(sup_test, child1),
+    timer:sleep(350),
+    catch link(SupPid),
+    %% The supervisor should not have exited as a cause of the intrinsic
+    %% child's termination
+    true = erlang:is_process_alive(SupPid),
+    %% The child should not have been restarted
+    [{child1, undefined, _, _}] = supervisor:which_children(SupPid).
 
 %%-------------------------------------------------------------------------
 %% Test restarting a process multiple times, being careful not


Maria-12648430:
While we (I collaborated on this 😉) are not sure about RabbitMQ's reason for introducing this restart type, I may elaborate a bit on our reason for picking this up when @essen suggested submitting it to OTP.

In our case, we have a `simple_one_for_one` supervisor under which are groups of cooperating processes (each under their own supervisors, which I will just call "group supervisors" here), with one of those processes acting as a kind of manager/frontend for the others. Said groups of processes are started as needed, which is easy. Stopping them is another story, as (in our case, I should say) it is the manager process which decides when the group should be stopped (including itself, and the group supervisor on top of them, as it would then be empty and idle forever).

There _are_ ways to achieve this, but no "nice" way (that we know of).
We choose the pragmatic approach of making the manager processes `permanent` and setting the restart intensity of their respective group supervisors to `0`. This does what we want, and we can live with the crash reports, but it also prevents restarts of the manager process should it crash.
Another option would be to hand down the top supervisor's and the respective group supervisor's pids to the manager processes, so they can use `terminate_child` to tell the top supervisor to terminate their group supervisor, but doing so from a child in this tree has it's own problems, especially if the child is trapping exits, with potential for deadlocks. In any case, some boilerplate code is required.

So, long story short: This is our attempt at a nice solution, inspired by RabbitMQ's `supervisor2`, for scenarios like the one outlined above ;)

josevalim:
Here is my understanding of the functionality provided, please let me know if I missed something: `intrinsic` behaves exactly like `transient` does, both in regards to restarts and under `rest_for_one`/`one_for_all` strategies. The difference however is that, if an `intrinsic` terminates normally (`normal`, `shutdown`, `{shutdown, _}`), then the supervisor terminates normally too.

If my understanding is correct, would it make sense to introduce a new flag to the child specs instead? This would be non-trivial in the past, but now with child specs using maps, it should be doable. For example, you could say that you want the exit reason in this case to cascade:

    #{cascade_exits := never | normal}

And then if other use cases show up in the future, other options could be added (`abnormal`, `always`, ...).

---

Perhaps it should also be noted that one should avoid using the `intrinsic` option for a child at the application root supervisor, as this means the app is going down without restarts.

Maria-12648430:
> Here is my understanding of the functionality provided, please let me know if I missed something: `intrinsic` behaves exactly like `transient` does, both in regards to restarts and under `rest_for_one`/`one_for_all` strategies. The difference however is that, if an `intrinsic` terminates normally (`normal`, `shutdown`, `{shutdown, _}`), then the supervisor terminates normally too.

That is correct. RabbitMQ's `supervisor2` which this is inspired by states it like this:
```erlang
%% 3) Added an 'intrinsic' restart type. Like the transient type, this
%%    type means the child should only be restarted if the child exits
%%    abnormally. Unlike the transient type, if the child exits
%%    normally, the supervisor itself also exits normally. If the
%%    child is a supervisor and it exits normally (i.e. with reason of
%%    'shutdown') then the child's parent also exits normally.
```
(The last sentence may be a bit misleading, but it is correct and basically says what the sentence before it already stated: If a child (be it a supervisor or worker) exits normally, the supervisor itself (which is the child's parent) exits normally.)

> If my understanding is correct, would it make sense to introduce a new flag to the child specs instead? This would be non-trivial in the past, but now with child specs using maps, it should be doable.

It is doable with map specs, yes, but I'm torn here: It would introduce an incompatibility (or rather, inconsistency) between map specs and tuple specs, which are both allowed and equivalent up to this point.

>For example, you could say that you want the exit reason in this case to cascade:
> 
> ```
> #{cascade_exits := never | normal}
> ```
> 
> And then if other use cases show up in the future, other options could be added (`abnormal`, `always`, ...).

That sounds useful, but yeah, what I just said ;) Let's hear what other people think.

> Perhaps it should also be noted that one should avoid using the `intrinsic` option for a child at the application root supervisor, as this means the app is going down without restarts.

That is correct, too. I would declare this a misconfiguration, or at least a case for "think twice" ;) Another place where you probably don't want to use the intrinsic behavior is in `simple_one_for_one` supervisors, ie supervisors with fully independent, dynamically added/removed childs, as the normal exit of _one_ child would then cause the shutdown of the supervisor and thus all the siblings.

lhoguin:
In RabbitMQ this is mostly used for socket processes. In RabbitMQ's case the acceptor pool (Ranch currently, it was a custom pool before) starts a supervisor and hands off the socket to a process started by that supervisor. It then monitors the supervisor as its child as normal.

A supervisor is used because RabbitMQ needs to start multiple processes tied to the connection process, like the channel supervisor for this connection, the queue collector, etc.

When the socket gets closed its supervisor needs to stop, and that's where things get hairy as mentioned by Maria. The `intrinsic` type allows us to stop this supervisor normally when the connection was closed normally, or to crash verbosely otherwise (because the restart intensity is 0, we don't allow restarts because we deal with sockets). Additional intermediate supervisors can be used for processes we want to allow restarts for.

It's not really about cascading exit reasons, it's about stopping the supervisor automatically when its job is done.

lib/stdlib/doc/src/supervisor.xml
```diff
@@ -192,7 +192,15 @@ child_spec() = #{id => child_id(),       % mandatory
...
-          than <c>normal</c>, <c>shutdown</c>, or <c>{shutdown,Term}</c>.</p>
+          than <c>normal</c>, <c>shutdown</c>, or <c>{shutdown,Term}</c>.
+          An <c>intrinsic</c> child process is restarted if it terminates
+          abnormally or if the supervisor's restart strategy is 
+          <c>rest_for_one</c> or <c>one_for_all</c> and a sibling's
+          death causes it to terminate. However, if an <c>intrinsic</c>
+          child process terminates normally, that is, with an exit reason
+          of <c>normal</c>, <c>shutdown</c> or <c>{shutdown,Term}</c>,
+          the <c>intrinsic</c> child's supervisor also terminates
+          normally.</p>
```
lhoguin, line 203: It should be made clearer that when any child terminates normally then all children will be shutdown and then the supervisor will exit.

lhoguin (changes requested):

lib/stdlib/doc/src/supervisor.xml
Maria-12648430: Thanks for clarifying :) And yes, I agree.

juhlig: Ok, will see to it later :)

juhlig: @lhoguin updated the paragraph, is that better? :)

lhoguin: Yes much better.

lhoguin (approved):

ferd:
So, there's a lot to unpack here.

First of all:

> However, if an intrinsic child exits normally (with exit reason normal, shutdown or {shutdown, Term}), the child's supervisor also shuts down normally.

A supervisor _always_ shuts down normally when it's about children, since the reason `shutdown` is pretty much the only one it returns outside of wrong specs or callbacks returning invalid data. This means that a permanent supervisor with a restart strategy of 0 will always respect the termination _reason_ and be equivalent to this. However, the only thing the `intrinsic` strategy _can_ offer that's different is to **still restart on error, but terminate on success**. I don't think it can offer a lot more. I think this is worth keeping in mind for the rest of my post. 

This explains this warning that first threw me in for a loop:

> Another place where you probably don't want to use the intrinsic behavior is in simple_one_for_one supervisors, ie supervisors with fully independent, dynamically added/removed childs, as the normal exit of one child would then cause the shutdown of the supervisor and thus all the siblings.

This is one hell of a sharp edge, and it surprised me because I first thought that `intrinsic` was supposed to be something like "take down the supervisor once it's empty." However this is not what it does, what it does is make erroneous terminations restartable (if you have more than 0 tolerance for faults) while making successful terminations final. It just happens that normal terminations have a 0 tolerance and would be taking down all other children whether they were still running or not.

This is a bit of a funny one because `intrinsic` seems to sit funny between two configuration values: it's halfway between strategies (`simple_one_for_one`, `one_for_one`, and `rest_for_one` sort of clash with it in weird ways, so it could be a behaviour specified at this level) and `restart` (where it can't work well with any of the existing ones, and needs to be specified at that level).

The thing I find limiting about this PR is that it won't allow a `temporary` supervisor to also clean up after itself upon a success _while not caring for faults_. I also find the idea of a `temporary_intrinsic` to be a really annoying kludge that shouldn't exist.

What I'm sort of detecting here is this idea that what is really desired is a task supervisor: we want a successful task to run to completion _and then clean up after itself_. This last bit is what makes it similar to `transient`, but the current implementation is confusing because _why should one successful child terminate others that aren't done_? Is this an implementation detail (it is simpler to see a good exit as final) or truly a desirable thing?

I think this is what @josevalim is sort of hinting at. "Clean up once done" _is_ desirable in many strategies. But it's not sufficient on its own to be equivalent to "restart on faults but terminate on clean exits", unless it is coupled to the `transient` restart where a tolerance for faults is seen as higher. That way, once all children have successfully exited, the supervisor goes away, but those that are left and fail still restart. 

The advantage there would be that it could also work with a `temporary` restart. Forcing all children to be gone to terminate the parent would also remove the sharp edges from `one_for_one`, `rest_for_one`, and `simple_one_for_one` strategies by making it clear when they should go.

`intrinsic` as implemented just does not compose well with other existing options. If the behaviour that is desired is really that everything should retry until one child succeeds and then it shuts down (which is what the current implementation allows) rather than a distinct "task supervisor cleans after itself once done", I don't know that it can fit well within the current supervisor mechanism because it makes its approach to faults and restart very distinct from any other strategy or restart specification. My approach there would be to do the following:

- have a `one_for_all` supervisor that is supervising other supervisors, with a `permanent` strategy and 0 tolerance to faults
- each of these supervised children are actually `one_for_one` using the "clean up after yourself" mechanism, and they all supervise a single child in a `transient` restart strategy, with a configurable amount of restarts for faults

What would happen then is that each worker is allowed to restart as much as specified, but once one of them succeeds the entire thing shuts down. I'm not sure the converse ("task supervisor cleans itself when everything is done") would be possible to design properly using `intrinsic` as suggested, for example.

Maria-12648430:
@ferd I'll need some time to digest this, it's quite a lot, and at some points I'm having a hard time to figure out what you mean ^^;;;

lhoguin:
There's no real clash with `simple_one_for_one`, it's just that you're unlikely to want both `intrinsic` and `simple_one_for_one`, just like you're unlikely today to want `permanent` and `simple_one_for_one`.

I do not think there's much overlap with restart `strategy`. `intrinsic` means `transient` except that when the intrinsic child completes, the supervisor shuts down as well. The only weird part is that `restart` so far only specified what should be done about the child, and now it also specifies what should be done about the supervisor.

Using a `permanent` strategy along with no fault tolerance is not equivalent because in that case the shut down will produce error or supervisor reports about the failure. We do not want this at all. When we say "normal shutdown" we mean there is no sign of error whatsoever.

So to sum it up, we want:

- for a supervisor to automatically shutdown when a specific child or children have exited normally
- for this supervisor to shut down any other children while shutting down ("normal shutdown")
- for this supervisor to NOT produce error reports when it has to shut down ("normal shutdown")

I have no strong feelings about how it should be specified, except that we must be able to say which children trigger a supervisor shutdown when they terminate normally.

Maria-12648430:
@ferd ok, let's first see where we are on common ground and where we aren't ;) I guess I can't answer all the points, I couldn't couldn't wrap my head around some.

So, it looks like you are not opposed to the general idea behind this PR, a self-exiting "clean up after yourself" supervisor. What you are mainly opposed to is that this is caused by a normal exit of a certain child, and would rather expect a supervisor that exits once it is empty, ie has no more active children. Is that right?

> > However, if an intrinsic child exits normally (with exit reason normal, shutdown or {shutdown, Term}), the child's supervisor also shuts down normally.
>
> A supervisor always shuts down normally when it's about children, since the reason shutdown is pretty much the only one it returns outside of wrong specs or callbacks returning invalid data.

"normally" here is in the sense of "as usual" or "like it does for any other reason".

I agree that `intrinsic` as a restart type seems a bit of a misfit, but you may also think of an `intrinsic` child as a kind of "leader" process. Looking at it this way, the "sharp edge" is actually what you (well, we ;)) _want_, and it also makes the combination of `intrinsic` and `temporary` pointless: A leader process is (arguably) an important process for the group of processes under the same supervisor, but not restarting this process if it fails would mean it is not important at all.

About a exit-if-empty supervisor, this would be useful, too, but it comes with it's own set of problems.
For one, it is not trivial, at least ambiguous, when a supervisor can be understood as being "empty". You could say "when a child exits and there are no more childs left running, duh", but what about childs that were terminated via `terminate_child/2`? They are in a strange state, they could be intended to be restarted later, or they could be intended to stay stopped.
For another, the "sharp edge" remains (just shifts to another place) when you use such a supervisor in combination with `simple_one_for_one`. This supervisor type is typically for independent, dynamically started/stopped, often short-lived processes, so such a supervisor becoming empty at some point is to be expected.

@lhoguin
> There's no real clash with `simple_one_for_one`, it's just that you're unlikely to want both `intrinsic` and `simple_one_for_one`, just like you're unlikely today to want `permanent` and `simple_one_for_one`.

No, not a clash ;) It's something that you _can_ combine, but it usually isn't a good idea. Like combining say, strawberries and garlic ;) You _can_ combine them, but... 🤮 😁

ferd:
> So, it looks like you are not opposed to the general idea behind this PR, a self-exiting "clean up after yourself" supervisor. What you are mainly opposed to is that this is caused by a normal exit of a certain child, and would rather expect a supervisor that exits once it is empty, ie has no more active children. Is that right?
 
Correct, that's the gist of it. It feels like "exits when only one child is successful" causes surprises in the case of `simple_one_for_one` or `one_for_one` or `rest_for_one` in that a non-erroneous termination of a child can cause all other children to terminate and be shut down.

The last point in my post is that "exit on first successful child" can be emulated with "clean up after yourself once empty", but "clean up after yourself once empty" cannot be emulated with "exit on first successful child", which to me makes it sound like one is more useful than the other, and can be a more general setting that applies in more contexts as well.
 
> "normally" here is in the sense of "as usual" or "like it does for any other reason".

Correct. All of the reasons that don't cause a crash report in a supervisor, essentially.

> I agree that `intrinsic` as a restart type seems a bit of a misfit, but you may also think of an `intrinsic` child as a kind of "leader" process. Looking at it this way, the "sharp edge" is actually what you (well, we ;)) _want_, and it also makes the combination of `intrinsic` and `temporary` pointless: A leader process is (arguably) an important process for the group of processes under the same supervisor, but not restarting this process if it fails would mean it is not important at all.
> 

An odd thing for me is what happens if you have a `rest_for_one` strategy where the first child is `permanent`, the second is `intrinsic`. Now let's say the first child dies and causes a restart. Should the restart that will kill the `intrinsic` child take down the supervisor, or this is accepted to not count against the threshold? The `intrinsic` child will die with reason `shutdown` when its parent attempts to kill it, but under the current implementation this will be eaten up by the shutdown loop I believe. Is this properly representing leader semantics to have siblings kill the leader?

The thing about `intrinsic` + `temporary` being pointless is that you still need to define and implement and document that behaviour, and create that self-consistent hierarchy of which one dominates in decision-making. This same thing is what would define whether an intrinsic child being restarted by a peer's error in a `one_for_all` or `rest_for_one` scenario actually kills the supervisor or does not count against it. 

Things will be "surprising" as long as they can't consistently be documented and therefore predictable.

To me a smaller unrelated issue I see is probably that I already use `rest_for_one` to define "leader-like" relationships across supervisors: anything that comes before another is judged more important. It offers more granularity than the rest by allowing to define more than one degree of dependence, and even representing whole chains. The one thing it can't easily do is differentiate between the fault rate of the first child and siblings, but that's generally covered by adding extra layers of supervisors. 
I tend to make a principle out of individual workers not knowing what their supervision structure is like, because I can tweak the handling of faults by shifting workers through the tree as I see fit.

> About a exit-if-empty supervisor, this would be useful, too, but it comes with it's own set of problems.
> For one, it is not trivial, at least ambiguous, when a supervisor can be understood as being "empty". You could say "when a child exits and there are no more childs left running, duh", but what about childs that were terminated via `terminate_child/2`? They are in a strange state, they could be intended to be restarted later, or they could be intended to stay stopped.

Yes. But you can pick one and document it and be done with it. Which one to choose would probably have to do with guessing what would lead to the fewest surprises.

> For another, the "sharp edge" remains (just shifts to another place) when you use such a supervisor in combination with `simple_one_for_one`. This supervisor type is typically for independent, dynamically started/stopped, often short-lived processes, so such a supervisor becoming empty at some point is to be expected.
> 

Yes. A related sharp edge is what do you do _before_ you actually get children. You could sort of work around that by considering the "drop when empty" to become "drop when falling below threshold" where you specify N children, below which the supervisor dies. Specifying a child threshold of 0 keeps the supervisor semantics as they are today. Setting the threshold to 1 means that once empty it vanishes, and you can make it go to as many as you want such that you could decide to keep running as long as 2/3 children are alive, and if these represent peers in a majority setting you go away. Maybe that's a bad idea, but it is somewhat clearer (or more configurable) than all or nothing (additional rough edges: can you re-define the value? does it count when doing partial restarts?). 

Anyway this is becoming a distracting digression from your PR.

> No, not a clash ;) It's something that you _can_ combine, but it usually isn't a good idea. Like combining say, strawberries and garlic ;) You _can_ combine them, but... 🤮 😁

Yeah. I like to try and avoid these when possible. "Totally permitted bad ideas" are nice to remove as much as possible. For reference: 

image

lhoguin:
> An odd thing for me is what happens if you have a `rest_for_one` strategy where the first child is `permanent`, the second is `intrinsic`. Now let's say the first child dies and causes a restart. Should the restart that will kill the `intrinsic` child take down the supervisor, or this is accepted to not count against the threshold? The `intrinsic` child will die with reason `shutdown` when its parent attempts to kill it, but under the current implementation this will be eaten up by the shutdown loop I believe. Is this properly representing leader semantics to have siblings kill the leader?

There's no conflict there, the intrinsic child has not terminated normally on its own, it is only being restarted (it has been terminated, potentially via a "normal" `shutdown` exit signal, by the supervisor). Permanent child dies, supervisor terminates and restarts it and the following children. Same behavior as permanent+transient.

> The thing about `intrinsic` + `temporary` being pointless

You can't have a child that is both intrinsic and temporary, it's one or the other. I think it makes sense to have the child spec defined as it is rather than a separate option because these concerns won't apply.

> This same thing is what would define whether an intrinsic child being restarted by a peer's error in a `one_for_all` or `rest_for_one` scenario actually kills the supervisor or does not count against it.

It does not, only normal self exits, not abnormal exits, not parent exits, of the intrinsic child, can lead to the supervisor shutting down.

If you do use `rest_for_one` with possibility for restarts with children permanent followed by intrinsic, then the intrinsic child must be restartable (so shouldn't hold a socket for example). For example it could be a process generating and pushing data somewhere, that retries on failure but only on failure. Same concerns as for transient.

If you do have an intrinsic child that can't restart, then you cannot allow for restarts anyway so it doesn't really matter if you use `rest_for_one` or any other. In that case any exit should lead to the supervisor exiting as well, only you want a quiet exit if the intrinsic child exited normally.

> To me a smaller unrelated issue I see is probably that I already use `rest_for_one` to define "leader-like" relationships across supervisors: anything that comes before another is judged more important. It offers more granularity than the rest by allowing to define more than one degree of dependence, and even representing whole chains. The one thing it can't easily do is differentiate between the fault rate of the first child and siblings, but that's generally covered by adding extra layers of supervisors.

You can continue using `rest_for_one` as before. The only thing this changes is what happens when the "leader" has finished its task. It will exit, and the supervisor will automatically shutdown the rest of the tree.

> I tend to make a principle out of individual workers not knowing what their supervision structure is like, because I can tweak the handling of faults by shifting workers through the tree as I see fit.

That's why intrinsic is interesting to have. Also it reduces the code size.

"exit when empty" means that the "intrinsic" process would have to take care of terminating its siblings before exiting, which means it has to know about the structure.

One thing to consider however is what are the implications of having more than one intrinsic child. I don't know if there's a use case for this. RabbitMQ does have multiple intrinsic siblings but I'm not sure if that does anything in practice in RabbitMQ's case.

Maria-12648430:
@ferd @lhoguin so are you two going to whack it out in #erlang already, or... j/k ;)

But seriously, I think we're getting to a point where we will be stuck. I believe you both make valid points, and both the intrinsic and "exit-when empty" (we really need a handier name if we are to continue this discussion much longer) approaches have advantages and disadvantages. Both would solve the problem @juhlig and I have (though the intrinsic approach would solve it somewhat better), so we have no _strong_ inclination towards either one...

<hr />

> Should the restart that will kill the intrinsic child take down the supervisor, or this is accepted to not count against the threshold?

I would say, _definitely not_, IMO that would be a surprising and quite annoying behavior, and I'm hard pressed to imagine a use case where this would be desired. I mean, while if you just look at the intrinsic child, it's shutdown is likely to be normal, in the wider picture it is not doing this because it itself decided it was done (@lhoguin said the basically same thing in his first comment above).

> Is this properly representing leader semantics to have siblings kill the leader?

Weeell... if you look at actual history, "getting killed by some relative" seems to be ranking high on the list of job hazards you face by being a leader 😁 (j/k again)

<hr />

> You can't have a child that is both intrinsic and temporary, it's one or the other.

You can't right now (ie, after this PR as it stands), but it _could_ be made possible in theory with something like `intrinsic_temporary` (not saying we should, definitely not saying we should do it like _that_). Quoting from @ferd's first comment on this PR:

> The thing I find limiting about this PR is that it won't allow a temporary supervisor to also clean up after itself upon a success while not caring for faults. I also find the idea of a temporary_intrinsic to be a really annoying kludge that shouldn't exist.

What I'm questioning is if it is something anyone would need, or if it is a question that just won't arise in the real world?

ferd:
It's possible it's neither. You shouldn't see my comments as blockers anyway. I'm flagged as a "member" because rebar3 lives under the Erlang org, but what I say has no actual bearing on whether this gets merged or not. I like to test these ideas as far as possible to try and find weird edge cases, but it doesn't mean that a solution as-is would or wouldn't be acceptable to the OTP team.

I personally have not really encountered a case where I would have needed this feature (but have encountered more cases where "clean-up the empty supervisor" would have been useful), and this tinted the way I approached this review. My experiences don't represent everyone else's though, so I'm fine not imposing anything there and just sticking to using  features I need.

lhoguin:
Same.

I have nothing against "clean up when empty", and it's probably possible to have both. It's less interesting in the RabbitMQ case because there are a lot of processes tied to this leader process (multiple levels of supervisors as), they don't necessarily know about each other (some do, some don't) and just doing an orderly shutdown of the supervisor is simpler than telling every one of them to terminate manually.

juhlig:
Ok, so we will let it be as it currently is, and wait for someone from OTP to chime in.

garazdawi:
Hello!

We are right now in the last preparations for the OTP-24 RC1. So it will be at least until that release is done until we get back to you about this.

juhlig:
> We are right now in the last preparations for the OTP-24 RC1. So it will be at least until that release is done until we get back to you about this.

No sweat :)

Maria-12648430:
> One thing to consider however is what are the implications of having more than one intrinsic child. I don't know if there's a use case for this. RabbitMQ does have multiple intrinsic siblings but I'm not sure if that does anything in practice in RabbitMQ's case.

For one, I would discourage the use of multiple intrinsic childs (if we think of them as leaders, having _multiple_ leaders is usually trouble, if we look at actual history ^^;), though not forbid it.

For another, I have been thinking about this over the weekend, and as I see it, those are the ways in which we could react when faced with multiple intrinsic childs:
* The supervisor shuts down when _any_ intrinsic child under it terminates normally by itself. This is the most straight-forward way, and how this PR as it stands does it, and what I would prefer because of simplicity.
* The supervisor shuts down when the last active intrinsic child stops normally by itself (ie, when all other intrinsic childs are inactive). There are some questions here, though, or glitches. As of this PR, the termination of an intrinsic child via `supervisor:terminate_child/2` does _not_ trigger the supervisor shutdown, resembling the behavior with the `permanent` type where similarly `supervisor:terminate_child/2` does _not_ trigger a restart of that child. If we stick to this here, surprises may occur depending on the order of events... assume 2 intrinsic childs:
  * we terminate one with `supervisor:terminate_child/2`, then the other stops normally by itself --> the supervisor will shut down
  * one stops normally by itself, then we terminate the other with `supervisor:terminate_child/2` --> the supervisor will _not_ shut down
* The supervisor shuts down when the _highest-ranking_ (according to childspec order) intrinsic child terminates normally by itself. This boils down to all intrinsic childs ranking lower than the highest-ranking _active_ intrinsic child to behave just like `transient`. Again, assume 2 intrinsic childs:
  * if the first child is active and the second child stops normally by itself, the supervisor does not shut down, ie the second child's exit is treated as if it was a transient child
  * if the first child is active and is terminated manually via `supervisor:terminate_child/2`, the second child is now the highest-ranking active one, and if it stops normally by itself, the supervisor shuts down; if the first child gets restarted, it is again the highest-ranking active one, and the second child again behaves like `transient`
  * if the first child is active and stops by itself normally, the supervisor shuts down, period

The third option does not remove the problems outlined in the second, but adds some determinism to it. Still, personally I would prefer the first option, because you know, simplicity ;)

lhoguin:
Shut down when any intrinsic child exits normally sounds best.

IngelaAndin:
To start with I like to say we (the PS team) sympathize with the problem that you are trying to solve. We also have that problem and not a really satisfactory solution. We also have been thinking about that we want to further develop supervisors. Supervisors are really useful although I think they suffer a little from development stagnation and people doing their own solutions instead of help improving the existing solution. This might have many reasons that I think we need not discuss here. 

I would say most of the original supervisor design was made with the assumption that you have a static supervisor tree and that restarting is more important than stopping, which is not always the case. So this was before my time so to speak, so this is my opinion from my experience with the code. There are some possibilities for more dynamic supervisor branches in your tree but I feel this mechanism is less developed and lacking. 

Anyway, my colleague @HansN  tried out your solution and he does not think it works to solve his problems in the ssh application. He can tell you more about it himself.

Personally, I think that if you have a dynamic supervisor branch that depends on a TCP connection you would always like to bring down the whole branch when the process of owning the socket terminates. And yes you want to be able to take it down initiated by the termination of that process without having a lot of error reports for "normal shutdown" scenarios. So I was thinking maybe we want a new restart behavior, like a significant_first that would mean if the first process terminates
(for any reason) the whole sub-tree should terminate and we can make this supervisor temporary as to not cause any escalations further up the tree. Then you could perhaps have significant_first_one_for_one if the help processes can be restarted as long as the connection lives and significant_first_one_for_rest etc. I feel that a solution along this line would be easier to understand. I have not prototyped this so there can of course be things I missed to think of. So what do you think? Would this be an interesting approach, that you are willing to explore?

lhoguin:
@IngelaAndin You can't just choose the first child as the significant, it's not going to work in cases where other processes must be started before the significant process (for example a helper supervisor). I think it has to be specified explicitly in the child spec.

IngelaAndin:
@lhoguin Please elaborate on why the help supervisor would need to be started before the process owning the socket? If you have more than one process depending on a specific connection it feels very natural that socket owner is spawned first and all the processes depending on the connection comes later independent of if the rest of the tree are leafs of not.

HansN:
I support the idea of a re-thinking of the supervisor behaviour.  It certainly handles the cases it was thought for, but now many seem to have cases that not fits exactly.  Even with new ideas there are different opinions.  For example with the proposed 'intrinsic' there is a discussion about whether the supervisor should die when the first child dies or when all of them dies.

Why is it so?  It's because the different aspects are (often) independent features like should the supervisor be killed or not if one child dies, should it be killed if the first or all children dies, should there be a message if in situation X or Y or Z etc.

The existing strategies (one_for_all etc) are just names of a set of feature values, and the features are often independent.  So why not let the user specify the features separately in the map?

Compatibility is a case, and it is important. We all want old code to continue to run.  So we should keep the old record specification and the current map keys and introduce new ones for the independent features that users nowadays need.  The current flags and child specs then becomes just a kind of macros and old code continue to work.

I have not thought deep of this, I have no prototype and I have not even checked the supervisor code, but I'm happy for comments!

lhoguin:
> @lhoguin Please elaborate on why the help supervisor would need to be started before the process owning the socket? If you have more than one process depending on a specific connection it feels very natural that socket owner is spawned first and all the processes depending on the connection comes later independent of if the rest of the tree are leafs of not.

RabbitMQ uses `start_child` to start most children because later pids need to know the pids of processes started before them. For example the connection needs the helper sup's pid. The channel needs the limiter's and the writer's pid. As a result the significant process is started last (the connection, or the channel, in these examples).

Passing pids forward is another thing that would be great to solve, by the way. :-) That one can perhaps be done via an option in the map.

juhlig:
@Maria-12648430 OK if I pass this over to you, at least for the time being? I'm pretty busy these days, and you are doing most of the talking already, anyway ;)

Maria-12648430:
@juhlig sure, I have some time on my hands, I'll take it on :)

Maria-12648430:
@IngelaAndin I pretty much agree with everything you said, and I'm willing to explore different approaches. But as we see here, there are already many different ideas and expectations here already, with pros and cons, and if more people chime in, I expect even more ideas and expectations with more pros and cons ^^;

To enumerate, up to here we have suggestions for:

* `intrinsic`, which is supported by @lhoguin, ie shut down the supervisor depending on the (normal) exit of `intrinsic` childs. There are 3 possible variations when to stop the supervisor, namely when...
  *  _any_ `intrinsic` child stops (this is what this PR does now, and what is again supported by @lhoguin)
  * the _last remaining_  `intrinsic` child stops (nobody supporting this yet)
  * the _highest-ranking_ (according to child spec order) active `intrinsic` child stops (nobody supporting this yet)
* "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last remaining active child exits (normally?)
* `significant_first_*`, suggested by @IngelaAndin, ie shut down the supervisor when the first (according to child spec order) child exits, for any reason

As I see it, it should be possible to have all of that really (how to best offer ways of configuration is something I haven't thought about much yet), which I think is what @HansN is hinting at. I'm not yet clear on how combinations of the suggested behaviors would behave under certain conditions, though.

Anyway, right now I have no definite target(s) towards which to explore, so... help? 😅

> Then you could perhaps have significant_first_one_for_one...

I believe @ferd just had a fit here, judging by what he said in an earlier comment for `temporary_intrinsic` ^^; j/k

Maria-12648430:
@HansN about what @IngelaAndin said...

> Anyway, my colleague @HansN tried out your solution and he does not think it works to solve his problems in the ssh application. He can tell you more about it himself.

... could you explain a bit more about why this would not work for your problems, and why/what would? (Maybe we could have a fourth idea to consider 😁 j/k)

Maria-12648430:
@lhoguin

> Passing pids forward is another thing that would be great to solve, by the way. :-)

Definitely :)

IngelaAndin:
Humm ... I see.  Well significant_last could work too. I am conceivable for other solutions too as long as they do not break old code and preferably not adds ambiguities. I would very much like to see that it can be used to solves the SSH supervisor problem too, as I think it is very similar and we should strive to have a general enough solution.

IngelaAndin:
@Maria-12648430 Just as I sent my answer I saw your new comments. I will come back to you. Need some time.  Do not know if will have time to day. But I think that we should be able to find a good solution.

Maria-12648430:
> Just as I sent my answer I saw your new comments. I will come back to you. Need some time. Do not know if will have time to day. But I think that we should be able to find a good solution.

Don't worry, after I post this I guess I'll call it a day/week also, think about it some more over the weekend, and then get back to it on monday :)

> Humm ... I see. Well significant_last could work too.

Ah, you beat @HansN to idea number 4, or 3&frac12;  ;)

`significant_last` would probably solve the problem @lhoguin describes, but it appears to be a bit... fragile (ambiguous? surprise-prone?) in the general sense, I believe. There is always the possibility that more childs get added via `supervisor:start_child/2`, and then another child would be the significant one. The stopping (for any reason) of this now-significant child would cause the supervisor to stop, and no way around it. I don't think this is what is intended, not often.

Also, somewhat along the same lines, I'm not sure this would be a general enough solution. Actually, the significant child might not be the first or last but one somewhere in the middle, and we might have even multiple childs that we may consider significant.

> I am conceivable for other solutions too as long as they do not break old code and preferably not adds ambiguities.

An option addressing this might be to mark specific _childs_ as significant, on the child spec level.
It would be a bit like `intrinsic` (and we have the same three possibilities in the face of multiple such childs, but we could even make that selectable with a new sup flag), only without the requirement that the childs exit _normally_.
(In)significance could/should be specified in a new option in the child spec map instead of as a restart strategy then, because this way it could be combined with the restart strategy to produce different behaviors:
* not `significant` -> same as usual
* `significant` + `temporary` -> the supervisor shuts down when the child exits, for any reason
* `significant` + `transient` -> the supervisor shuts down when the child exits normally, but restarts it (and maybe others, depending on supervisor strategy) when it exits abnormally
* `significant` + `permanent` -> makes no sense I think, boils down to just `permanent`

> I would very much like to see that it can be used to solves the SSH supervisor problem too, as I think it is very similar and we should strive to have a general enough solution.

Let's see what the problem is, :)

josevalim:
I have been following the discussion and to me the clearest proposal so far is the one by @ferd:

> "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last remaining active child exits (normally?)

Perhaps I would just rephrase it as:

> "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last child spec is removed

?

juhlig:
It just occured to me that there might be another solution (just an idea, haven't checked anything, don't know if it's possible or feasible): We could break up the supervisor module and put some functions into callback modules (which would make the distinction of the different strategies clearer, too, instead of having it all baked into the `supervisor` module).

What I'm thinking of for a start is a function that gets called when a child exits, and that tells the (general) supervisor what to do. The possibilities are limited to things like "restart this child" (ie, one_for_one), "restart this and some other childs" (ie, one_for_all and rest_for_one), "shut down the supervisor" (intensity reached/intrinsic/exit-when-empty/significant), so there is no need to hand out the actual handling of restarts etc.

I haven't thought this through to the end, but done right, it would enable anyone to tailor a supervisor to their specific needs, and still be within a canonical environment (vs re-implementing `supervisor` only for some small changes): @lhoguin could have `intrinsic`, @ferd could have "exit-when-empty", @IngelaAndin could have `significant_first` and `significant_last`, and @HansN could (probably) have one that fits his problem in `ssh`.

We could ship the common ones we have already that cover most cases, and have a new option to allow specifying custom callback modules for the very special cases.

If this is possible to do (and it sounds simple enough), it would allow for a very wide degree of freedom, but still be a small enough field to prevent rank growth of supervisor variants.

What do you think?

juhlig:
> I have been following the discussion and to me the clearest proposal so far is the one by @ferd:
> 
> > "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last remaining active child exits (normally?)
> 
> Perhaps I would just rephrase it as:
> 
> > "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last child spec is removed

Hm, but the last (active) child exiting is a different thing than the last child spec being removed... Childs exit for different reasons, usually by themselves. Removing child specs has to be done manually, and from outside (because a child has to be stopped before the spec can be removed, so it cannot remove it's own child spec). And if you have to do it manually from outside, why not just tell the parent supervisor to stop the one in question? ^^;

essen:
> I have been following the discussion and to me the clearest proposal so far is the one by @ferd:
> 
> > "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last remaining active child exits (normally?)
> 
> Perhaps I would just rephrase it as:
> 
> > "exit-when-empty", which is supported by @ferd, ie shut down the supervisor when the last child spec is removed
> 
> ?

Here's a scenario:

* `parent_sup` "exit-when-empty"
* * `helper_sup` "exit-when-empty"
* * * `helper_proc`
* * `conn_proc`

For `parent_sup` to exit, both `helper_sup` and `conn_proc` must exit normally. Let's say the connection is going down, `conn_proc` will exit. In order for `helper_proc` to exit as well, either `helper_proc` must be aware of `conn_proc` and detect the exit, or `conn_proc` must tell `helper_proc` (and any other process in `helper_sup`, there could be more than one) to exit when it itself exits. The complexity cost of doing this is larger the more processes you have.

Then edge cases may arise. Let's say `conn_proc` exited normally, and `helper_proc` detected it and begun shutting down. In the process of shutting down it crashed. When it restarts it has to detect that the connection is gone in order to again attempt to exit. If `helper_proc` was shutting down as a result of a `'DOWN'` message, then I suppose it'll get another `'DOWN'` and try to exit again. If it was `conn_proc` shutting down `helper_proc`, then the only real solution is for this shutdown to be done synchronously by `conn_proc`.

As you can see there's quite a bit more to think about compared to "supervisor does a normal shutdown when important process exited normally on its own", in cases like this. "exit-when-empty" better fits cases where the children exit independently from each other.

Maria-12648430:
> What do you think?

@juhlig how did you... just get out of my head, or at least _ask_ before you help yourself 😅

juhlig:
> how did you... just get out of my head, or at least _ask_ before you help yourself 😅

@Maria-12648430 sorry =^^= ... wait, _what_? really now?

Maria-12648430:
@juhlig yeah really 🙄 I will check it out on monday, ok?

juhlig:
@Maria-12648430 great

Everyone else, thoughts welcome :)

Maria-12648430:
after skimming the supervisor code, the callback idea seems to be doable at least, but really segregating the different strategies into callback modules requires quite a lot of work. especially the simple_on_for_one strategy has a lot of peculiarities that are not easy to accomodate for.

IngelaAndin:
> An option addressing this might be to mark specific _childs_ as significant, on the child spec level.
> It would be a bit like `intrinsic` (and we have the same three possibilities in the face of multiple such childs, but we could even make that selectable with a new sup flag), only without the requirement that the childs exit _normally_.
> (In)significance could/should be specified in a new option in the child spec map instead of as a restart strategy then, because this way it could be combined with the restart strategy to produce different behaviors:
> 
>     * not `significant` -> same as usual
> 
>     * `significant` + `temporary` -> the supervisor shuts down when the child exits, for any reason
> 
>     * `significant` + `transient` -> the supervisor shuts down when the child exits normally, but restarts it (and maybe others, depending on supervisor strategy) when it exits abnormally
> 
>     * `significant` + `permanent` -> makes no sense I think, boils down to just `permanent`
> 
 
I think this sounds reasonable.  I think the reason I was thinking in terms of restart strategy, is as I thought of of it as a new kind of supervisor more than as a child only property. But if we can managed to combine them in a way that makes sense I feel that is good. And it would probably be ok to have some combinations that does not make sense. There are probably such combinations already even if I can not think of one of the top of my head. 

> > I would very much like to see that it can be used to solves the SSH supervisor problem too, as I think it is very similar and we should strive to have a general enough solution.
> 
> Let's see what the problem is, :)

Look at ssh_connection_handler.erl   function stop_subsystem

Maria-12648430:
> I think this sounds reasonable. I think the reason I was thinking in terms of restart strategy, is as I thought of of it as a new kind of supervisor more than as a child only property. But if we can managed to combine them in a way that makes sense I feel that is good. And it would probably be ok to have some combinations that does not make sense. There are probably such combinations already even if I can not think of one of the top of my head.

Ok, then I'll start exploring in that direction :)

> > Let's see what the problem is, :)
> 
> Look at ssh_connection_handler.erl function stop_subsystem

Sorry, I'm still in the dark ^^;;;

ferd:
> after skimming the supervisor code, the callback idea seems to be doable at least, but really segregating the different strategies into callback modules requires quite a lot of work. especially the simple_on_for_one strategy has a lot of peculiarities that are not easy to accomodate for.

callbacks into a supervisor can be rather risky for failures, but also because a reference to a local fun that gets purged after hot code loads sends a kill signal to any process holding a reference. It might be doable, but it introduces some real fun corner cases to handle properly

Maria-12648430:
> callbacks into a supervisor can be rather risky for failures, but also because a reference to a local fun that gets purged after hot code loads sends a kill signal to any process holding a reference. It might be doable, but it introduces some real fun corner cases to handle properly

Ah, pity, it sounded like a nice idea on the surface =^^=

lhoguin:
You can always accept a module or a module/function for the callback.

garazdawi:
> because a reference to a local fun that gets purged after hot code loads sends a kill signal to any process holding a reference. 

It will only be killed if you have the fun in your callstack, i.e. you have called it and upgraded while it was being called. What will happen though is that the fun will crash when called after the upgrade.

IngelaAndin:
> > I think this sounds reasonable. I think the reason I was thinking in terms of restart strategy, is as I thought of of it as a new kind of supervisor more than as a child only property. But if we can managed to combine them in a way that makes sense I feel that is good. And it would probably be ok to have some combinations that does not make sense. There are probably such combinations already even if I can not think of one of the top of my head.
> 
> Ok, then I'll start exploring in that direction :)
> 
> > > Let's see what the problem is, :)
> > 
> > 
> > Look at ssh_connection_handler.erl function stop_subsystem
> 
> Sorry, I'm still in the dark ^^;;;
Humm ... well it was a while since I was more involved in the ssh application. I find it a good way to look at the supervision  tree is to run a simple example for the application an look at the supervisor tree in observer.
But I will try to give you some more tips. So if you look at ssh_system_sup:stop_subsystem you will see that we are trying to find the correct supervisor to shutdown from the top so to speak. And then a process is spawned to do the shutdown and avoid a deadlock as the process initiating the shutdown should be shutdown. Or well that is how I remember it, now there seems to be another process involved  wait_until_dead. (Guess some kind of synchronization purpose)  @HansN ?  
 Anyway on the server side for ssh a subsystem will be a number of process that will depend on one ssh-connection handler process (owner of one TCP socket) so if the socket is closed the subsystem should be closed down. And most of the time this will be a normal termination an we do not want any error-reports to occur.

Maria-12648430:
> You can always accept a module or a module/function for the callback.

@ferd @lhoguin Hm, that was the plan, actually... Now you two are getting me confused =^^=

What I imagined was (simplified and without the other necessary details) callback modules with a function that gets called when a child exits (which child it was, probably also the other child specs around it, and so on), and that tells (or hints at) the supervisor about what to do, _not do it itself_. Like, a callback module for `one_for_one` would then say "restart it" or "forget it, just go on" (depending on restart type); the module for `rest_for_one` would say "restart it, and all the others behind it", etc etc.

Maria-12648430:
> But I will try to give you some more tips. So if you look at ssh_system_sup:stop_subsystem you will see that we are trying to find the correct supervisor to shutdown from the top so to speak. And then a process is spawned to do the shutdown and avoid a deadlock as the process initiating the shutdown should be shutdown. Or well that is how I remember it, now there seems to be another process involved wait_until_dead. (Guess some kind of synchronization purpose) @HansN ?

Thanks for the explanation, I just noticed it too after staring at it for a while ;)

> Anyway on the server side for ssh a subsystem will be a number of process that will depend on one ssh-connection handler process (owner of one TCP socket) so if the socket is closed the subsystem should be closed down. And most of the time this will be a normal termination an we do not want any error-reports to occur.

I'm not sure this PR can do what you want, but it sounds like it could. So I will have a closer look at it again tomorrow :)

Maria-12648430:
@IngelaAndin I started making changes towards https://github.com/erlang/otp/pull/4521#issuecomment-786761719, and it seems to be a better fit than the `intrinsic` restart type. However, it requires putting an additional field in the `child` record, which makes me a bit worried about hot code upgrades... which, I admit, I don't have much experience with, especially since this is `supervisor`/`STDLIB`

IngelaAndin:
@Maria-12648430  When it comes to hot code upgrades by the book, you can not do them if you change the supervisor tree, then you will have to restart the application. I do not think we ever supported real hot upgrade for stdlib, I think both kernel and stdlib always requires a application restart. I think that real hot code upgrade is mostly practical for small critical patches that can then be applied without disturbing the system at all, however I think a more adopted way of hot code upgrades is to just load the new modules that fixes some issue or add some feature and  live with a possible small disturbance of the system.  

We had a Technical Board meeting today about this PR. And a colleague will give you some more feedback from that meeting soon.

okeuday:
@Maria-12648430 The https://github.com/erlang/otp/pull/4521#issuecomment-786761719 approach looks good to me, but I had 2 suggestions.  The `significant` use is meant to mark specific child specs, so the significant idea is meant to relate directly to the child, but the concept may be easier to understand if `terminate` is used instead to refer to how the supervisor will be terminating based on the child spec.

I was expecting that the `restart` data in the child spec could become a list to contain 2 atoms, though it wouldn't be required to be a list.  That should be appropriate because the `significant` atom would be an optional modifier.

juhlig:
> I do not think we ever supported real hot upgrade for stdlib, I think both kernel and stdlib always requires a application restart.

Ah, that's a great relief ^______^

> We had a Technical Board meeting today about this PR. And a colleague will give you some more feedback from that meeting soon.

Nice :) I'll be looking forward to it.

juhlig:
... Deleted my reply by accident =^^= Again...

> I do not think we ever supported real hot upgrade for stdlib, I think both kernel and stdlib always requires a application restart. 

Ah, that's a great relief ^______^

> We had a Technical Board meeting today about this PR. And a colleague will give you some more feedback from that meeting soon.

Nice :) We'll be looking forward to it.

dmorneau:
> There are ways to achieve this, but no "nice" way (that we know of).

Is the "not nice" alternative passing the supervisor pid down to the worker, and having it send the supervisor an exit signal before stopping? I'm curious if this could be done more cleanly if supervisor had an async stop function, like the stop of gen_server, but non-blocking...

Maria-12648430:
> > There are ways to achieve this, but no "nice" way (that we know of).
> 
> Is the "not nice" alternative passing the supervisor pid down to the worker, and having it send the supervisor an exit signal before stopping?

I would call this the "very not nice" way ;) I don't think you're supposed to send supervisors anything manually. I believe the relevant exit signals for a supervisor are documented somewhere, but not as a way to encourage this.

The "just not nice" way is to pass the supervisors pid and the pid of the supervisor on top of it to the child (or use their names, if they are registered), and spawn a process to call `supervisor:terminate_child/2` to tell the supervisor's supervisor to terminate the supervisor to initiate the shutdown.

> I'm curious if this could be done more cleanly if supervisor had an async stop function, like the stop of gen_server, but non-blocking...

A difference here lies in what the child will have to know. For all the above approaches, it will need to know the supervisor's pid, that it may have siblings, and that it is safe to shut them all down (in the "just not nice" approach, it also needs to know the supervisor's supervisor's pid). In the `significant` child approach, a child doesn't have to know anything about supervision and trees and siblings at all, it only really needs to know when it is done with what it was doing, and the supervisor (where you defined the entire setup with siblings and all) will know what to do: restart it, restart others, ignore it, shut down itself, whatever...

Maria-12648430:
@okeuday This is more or less about naming, right? I'm notoriously bad at that XD

Personally, I find "significant" not a bad choice, it expresses it like "this is an important child, without it there is no point in having the others (and the supervisor) around".

Also, I don't think extending the `restart` type is a good idea. For one, having `intrinsic` as a restart type is what seems to have confused many people in the beginning of the discussion here. For another, it is not really about restarting anything, actually the opposite ^^;

But I'm open to suggestions, and look forward to hear what the technical board thinks :)

Maria-12648430:
@dmorneau Actually, I'm not against a stop function for supervisors as you suggest, but it should not be for the reason to shut down supervisors from further down the tree, I guess.

garazdawi:
Hello!

We had an OTB meeting talking about this PR today and I want to share some of our thoughts.

We are very happy to see ideas and work being done in the area around supervisors. It has been an area that has seen very little change for a long time and the way that we write applications has changed since it was originally developed, so an update is long overdue.

There are many reasons why supervisors have not been changed much recently, but part of it is because the people that knew the most about them have moved away to other work and thus we have a knowledge gap about many of the details regarding how supervisors are meant to work. Another reason that we have not changed supervisors a lot lately is that some functionality already feels a bit out of place and when changing things we do not want to add something without being extra sure that it fits into the greater picture of how supervisors should work.

So, with that said, we will need to understand all the intricacies of how this works and that means that this PR will need an EEP describing the problem, changes and also discussing the reasons for not using various other solutions.

We will spend as much time as we can to help guide the conversation, but until OTP-24 is released most of our time will be dedicated to that, so we do not know how much time will be available.

Lukas and the OTP Technical Board

IngelaAndin:
> @dmorneau Actually, I'm not against a stop function for supervisors as you suggest, but it should not be for the reason to shut down supervisors from further down the tree, I guess.

I agree a stop function is missing from the API and I also agree it does not solve the problem we are trying to solve here.

Maria-12648430:
@garazdawi OK, I have compiled an EEP, will check it again and submit tomorrow :)

Maria-12648430:
FYI: [EEP 56](http://erlang.org/eeps/eep-0056.html)

juhlig:
Reference implementation for EEP 56 in #4638

IngelaAndin:
Close in favour of #4638

PR #4526, opened 2021-02-16
Fix init io handling

garazdawi:
This fixes a bug introduced in 95dcdb7754.

While at it, now that I better understand what is going on, I cleaned up the code in init a bit and added a short-circuit in application_master to make io handling a little bit more performant.


lib/kernel/src/application_master.erl
```diff
@@ -162,7 +162,7 @@ start_it(State, Type) ->
 init_loop(Pid, Tag, State, Type) ->
     receive
  	IoReq when element(1, IoReq) =:= io_request ->
-            relay_to_group_leader(IoReq, State),
+            _ = relay_to_group_leader(IoReq, State),
```
ranjanified, line 165: Any particular reason for this '_' binding? With the notion of  "function return value(s)" I can understand, anything else apart from that? Only curious.

garazdawi: It is used to suppress a dialyzer warning regarding unmatched returns.

ranjanified: O!! my my dear little brain... Thanks. :smiley:

garazdawi:
merged to master

PR #4530, opened 2021-02-17
Tweak inet keepalive option documentation

sneako:
The value of the `keepalive` option should be a boolean, but the
documentation states that the default is `disabled` which is not a
boolean.


RaimoNiskanen:
Obviously correct.  Thank you for fixing this issue!

PR #4531, opened 2021-02-18
jit: Generate multi-byte NOPs in the way GAS does

frej:
BeamAssembler::aligned_call() generates a nop (0x90) instruction for
each byte of alignment required. Most modern CPU:s are able to execute
up to four NOP instructions in a cycle [1] during ideal conditions,
but there are no guarantees for the instruction stream generated by
the BEAM loader to be ideal.

Looking at the BEAM code loaded by a typical run of the
`scripts/diffable` benchmark-tool, we find that there are in total
211152 locations requiring alignment. Of those, the number of
alignment bytes needed have the following distribution:

  bytes   number
      0   43502
      1   45703
      2  121947

Interesting to note is that almost 80% of the alignments will, without
this patch, issue more than one instruction. Minimizing the number of
issued "useless" instructions is advantageous as it frees up execution
units for other instructions doing useful work. Therefore this changes
the BeamAssembler::aligned_call() implementation to generate
multi-byte NOPs. The multi-byte NOPs used are the same as the NOPs
produced by the GNU Assembler when doing alignment in a code segment.

[1] https://www.agner.org/optimize/

### DIFF
--- a/erts/emulator/beam/jit/beam_asm.cpp
+++ b/erts/emulator/beam/jit/beam_asm.cpp
@@ -907,3 +907,8 @@ extern "C"
         ba->patchStrings(rw_base, string_table);
     }
 }
+
+const uint8_t *BeamAssembler::nops[3] = {nop1, nop2, nop3};
+const uint8_t BeamAssembler::nop1[1] = {0x90};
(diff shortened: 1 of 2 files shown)


garazdawi (approved):

bjorng:
Thanks for your pull request.

Issue #4502, opened 2021-02-11
Scheduler module: Ignoring online schedulers


tak30:
**Describe the bug**
When running inside docker with the flag --cpus the online schedulers are properly configured but the module scheduler reports all possible schedules utilization and not only the online.  

**To Reproduce**

1. Run otp 23 inside a docker container with cpus flag: 
```
   docker run --cpus 2 -it erlang:23.2.4 /bin/bash 
   erl
```
2. Check online schedulers: ```erlang:system_info(schedulers_online).```
3. Check scheduler utilization: ```scheduler:utilization(scheduler:sample()).```

You will see that the scheduler utlization reports more schedulers than the N that schedulers_online reported.

**Expected behavior**
Calls to the module scheduler should report only utilization of the online schedulers.

**Affected versions**

 OTP 23.*

Note: I didn't put prior versions to otp 23 as affected by this issue as prior to otp 23 there was no support for cpu limits inside containers.

**Additional context**

sverker:
I'm not sure this is worth "fixing". Keeping it as is

- does not risk any backward compatibility problems.
- may give you a couple of extra inactive offline schedulers in the list that can be ignored.
- do result in a different 'total' value. But the 'total' value is dubious anyway as it contains all dirty-cpu schedulers that may very well all be inactive.
- does not affect the 'weighted' value.

tak30:
Hi, thanks a lot for taking a look at the issue.

Which seems strange to me is that all schedulers report utilization, I would expect that those schedulers "not online" should report an utilization of 0%. Also we have seen that most of the time one scheduler is almost 100%. Example of a 12 core host with 2 cores available:

```
2> io:format("~p", [scheduler:utilization(scheduler:sample())]).
[{total,0.06198379665575581,"6.2%"},
 {weighted,0.12396759331151162,"12.4%"},
 {normal,1,1.0,"100.0%"},
 {normal,2,0.0643971273935827,"6.4%"},
 {normal,3,0.061633875799530405,"6.2%"},
 {normal,4,0.050678467996485635,"5.1%"},
 {normal,5,0.06317929065710347,"6.3%"},
 {normal,6,0.04405369144916358,"4.4%"},
 {normal,7,0.030897090040044255,"3.1%"},
 {normal,8,0.05600799519336589,"5.6%"},
 {normal,9,0.02001999093434373,"2.0%"},
 {normal,10,0.026704984358213196,"2.7%"},
[...]

PR #4532, opened 2021-02-18
runtime_tools: Clarify scheduler utilization docs

sverker:
From issue #4502.

### DIFF
--- a/lib/runtime_tools/doc/src/scheduler.xml
+++ b/lib/runtime_tools/doc/src/scheduler.xml
@@ -63,7 +63,10 @@
 	<taglist>
 	  <tag><c>{normal, SchedulerId, Util, Percent}</c></tag>
 	  <item>Scheduler utilization of a normal scheduler with number
-	    <c>SchedulerId</c>.</item>
+	    <c>SchedulerId</c>. Schedulers that are not online will also be
+	    included.
+	    <seeerl marker="erts:erlang#system_info_schedulers_online">Online
+	    schedulers</seeerl> have the lowest <c>SchedulerId</c>.</item>
 	  <tag><c>{cpu, SchedulerId, Util, Percent}</c></tag>
 	  <item>Scheduler utilization of a dirty-cpu scheduler with number
 	    <c>SchedulerId</c>.</item>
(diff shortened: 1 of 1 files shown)


sverker:
@tak30 How about this

@rickard-green What do you think about "a couple of seconds" a recommended minimum measuring interval for scheduler utilization?

tak30:
@sverker Seems fine. That will avoid the misunderstanding I had with the output given by the utilization function and the way samples are needed to take.

lib/runtime_tools/doc/src/scheduler.xml
```diff
@@ -117,6 +120,29 @@
         <p>Calculate scheduler utilizations for the time interval from when
 	<c><anno>Sample</anno></c> was taken and "now". The same as calling
 	<c>scheduler:utilization(Sample, scheduler:sample_all())</c>.</p>
+	<note>
+	  <p>
+	    Scheduler utilization is meassured as an average value over a time
```
josevalim, line 125:
```suggestion
	    Scheduler utilization is measured as an average value over a time
```

rickard-green (approved):

lib/runtime_tools/doc/src/scheduler.xml
```diff
@@ -63,7 +63,10 @@
 	<taglist>
 	  <tag><c>{normal, SchedulerId, Util, Percent}</c></tag>
 	  <item>Scheduler utilization of a normal scheduler with number
-	    <c>SchedulerId</c>.</item>
+	    <c>SchedulerId</c>. Even schedulers that are not online will be
```
paulo-ferraz-oliveira, line 66:
```suggestion
	    <c>SchedulerId</c>. Schedulers that are not online will also be
```

lib/runtime_tools/doc/src/scheduler.xml
```diff
@@ -117,6 +120,29 @@
...
+	    have passed between the two samples. For this reason, you should not
+	    do
+	  </p>
+<pre>
+scheduler:utilization(scheduler:sample()). % DO NOT DO THIS!
+</pre>
+          <p>
+            The above example takes two samples in rapid succession and calculates
+	    the scheduler utilization between them. The resulting values will
+	    probably be more misleading rather than informative.
```
paulo-ferraz-oliveira, line 137:
```suggestion
	    probably be more misleading than informative.
```

paulo-ferraz-oliveira:
`recon` [does it like so](https://github.com/ferd/recon/blob/master/src/recon.erl#L359). We usually call that with `5000` (ms) and it works OK.

Issue #4451, opened 2021-01-31
ERL-1474: emacs mode still requires cl which was deprecated in emacs 27.1


OTP-Maintainer:
Original reporter: `JIRAUSER13400`
Affected version: `Not Specified`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1474

---

```
erlang.el and erlang-test.el still contain {code}
(require 'cl)
{code}
which emits the {{Package cl is deprecated}} warning.
```

ranjanified:
As per [1015](https://bugs.erlang.org/browse/ERL-1015), and as per @kjellwinblad 's [comment](https://bugs.erlang.org/browse/ERL-1015?focusedCommentId=15612&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15612). 

However, a quick rgrep of emacs folder yields on branch master, [erlang.el](https://github.com/erlang/otp/blob/master/lib/tools/emacs/erlang.el#L79) and [erlang-test.el](https://github.com/erlang/otp/blob/master/lib/tools/emacs/erlang-test.el#L65-#L66) entries as such
`(eval-when-compile (require 'cl))`

Any tips please? Can these entries be removed, as they already have been in few other files, as per the comment.

ranjanified:
[auto-highlight-symbol-mode](https://github.com/mhayashi1120/auto-highlight-symbol-mode/blob/master/auto-highlight-symbol.el#L242) has a dependency on (require 'cl). At least this is the one I am loading from source. 

In my emacs config I am doing (require '[edts-start](https://github.com/sebastiw/edts/blob/master/edts-start.el#L51))
1. which loads [edts-mode](https://github.com/sebastiw/edts/blob/master/edts-mode.el). 
2. edts-mode has a [dependency](https://github.com/sebastiw/edts/blob/master/edts-mode.el#L29) on auto-highlight-symbol-mode
3. auto-highlight-symbol-mode has dependency on 'cl

**Hence the warning, even though the 2 entries mentioned in previous comment are taken away.**

ranjanified:
I removed (require 'edts-start) and added (require 'erlang) in my emacs config, and the warning is back. That means, with the current [erlang.el](https://github.com/erlang/otp/blob/master/lib/tools/emacs/erlang.el#L79) and [erlang-test.el](https://github.com/erlang/otp/blob/master/lib/tools/emacs/erlang-test.el#L65-#L66), this warning still shows up.

[...]

PR #4542, opened 2021-02-22
tools: Fix emacs compiler warnings (#4451)

leoliu:
Do not depend on cl. Assume presence of before-save-hook (Emacs 22)
and remove erlang-check-module-name-init. Remove (run-hooks
'erlang-mode-hook) from erlang-mode definition because
define-derived-mode does it already. Replace erlang-caddr with caddr
from Emacs 26.

erlang.el now compiles cleanly in Emacs 24.5, 25, 26, 27.

### DIFF
--- a/lib/tools/emacs/erlang-eunit.el
+++ b/lib/tools/emacs/erlang-eunit.el
@@ -216,7 +216,7 @@ buffer and vice versa"
 
 With prefix arg, compiles for debug and runs tests with the verbose flag set."
   (interactive)
-  (case (erlang-eunit-recent 'mode)
+  (cl-case (erlang-eunit-recent 'mode)
     ('test-mode
      (erlang-eunit-compile-and-test
       'erlang-eunit-run-test (list (erlang-eunit-recent 'module)
--- a/lib/tools/emacs/erlang-test.el
+++ b/lib/tools/emacs/erlang-test.el
@@ -62,8 +62,6 @@
 
 ;;; Code:
 
-(eval-when-compile
-  (require 'cl))
 (require 'ert)
 (require 'erlang)
 
@@ -127,8 +125,8 @@ concatenated to form an erlang file to test on.")
 
 (defun erlang-test-create-erlang-file (erlang-file)
   (with-temp-file erlang-file
-    (loop for (_ . code) in erlang-test-code
-          do (insert code "\n"))))
+    (cl-loop for (_ . code) in erlang-test-code
+             do (insert code "\n"))))
 
 (defun erlang-test-compile-tags (erlang-file tags-file)
   (should (zerop (call-process "etags" nil nil nil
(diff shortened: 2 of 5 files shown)


dgud:
The test case emacs_SUITE:compile_and_load  still fails on emacs-27.1, not really your problem but you know much more about
emacs lisp than me, would appreciate if you have the time to take a look.

leoliu:
I looked at emacs_SUITE.erl and it seems this line might be problematic. Not sure if that's the culprit though.

```
diff --git a/lib/tools/test/emacs_SUITE.erl b/lib/tools/test/emacs_SUITE.erl
index e5587b35..f74a159c 100644
--- a/lib/tools/test/emacs_SUITE.erl
+++ b/lib/tools/test/emacs_SUITE.erl
@@ -107,7 +107,7 @@ compile_and_load(_Config) ->
                 %% Workaround byte-compile-error-on-warn which seem broken in
                 %% Emacs 25.
                 "\"(advice-add #'display-warning :after "
-                    "(lambda (_ f _ _) (error \"%s\" f)))\"";
+                    "(lambda (_ f &optional _ _) (error \\\"%s\\\" f)))\"";
             _ ->
                 "\"(setq byte-compile-error-on-warn t)\""
         end,
```

I also found erldoc.el compiles with a warning under emacs 27. Is this what's causing the failure? Though the warning can be suppressed with the following simple change it might be a regression on emacs' side. I will check with emacs upstream to see what's happening to `eval-and-compile`:

```
diff --git a/lib/tools/emacs/erldoc.el b/lib/tools/emacs/erldoc.el
index a8ac81ec..5d185e1f 100644
--- a/lib/tools/emacs/erldoc.el
+++ b/lib/tools/emacs/erldoc.el
@@ -62,6 +62,7 @@
 
 (eval-when-compile (require 'url-parse))
 (require 'cl-lib)
+(require 'json)
 (require 'erlang)
 
 (eval-and-compile                       ;for emacs < 24.3
@@ -268,7 +269,6 @@ (defun erldoc-parse-all (man-index output &optional json)
     (with-temp-buffer
       (if (not json)
           (pp table (current-buffer))
-        (eval-and-compile (require 'json))
         (let ((json-encoding-pretty-print t))
           (insert (json-encode table))))
       (unless (file-directory-p (file-name-directory output))
```

dgud:
You can run with  `make test ARGS="-suite emacs_SUITE"`  from  `$ERL_TOP/lib/tools`

Yes warnings are errors if emacs > 25 and I understand the testcase correct.

And it fails on erldoc:
```
emacs --batch --quick --directory "/ldisk/daily_build/24_master-opu_o.2021-02-24_21/otp/lib/erlang/lib/tools-3.5/emacs" --eval "(require 'erlang-start)" --eval "(setq byte-compile-error-on-warn t)"  -f batch-byte-compile "/ldisk/daily_build/24_master-opu_o.2021-02-24_21/otp/lib/erlang/lib/tools-3.5/emacs/erldoc.el"

 => 1 Loading erlang-skels...
In toplevel form:
../../../otp/lib/erlang/lib/tools-3.5/emacs/erldoc.el:534:1:Error: the following functions might not be defined at runtime: json-encode, cl-reduce, cl-mapcan
```

The other modules have warnings as well, but doesn't seem to fail the testcase, I have not had time to dig deeper yet.

leoliu:
Thanks for the tip `make test ARGS="-suite emacs_SUITE"`. After some trial and error I was able to fix the failed tests and some warnings and errors that I found in the logs.

Somehow `(eval-when-compile (require 'url-parse))` has to be moved to after `(require 'cl-lib)` in `erldoc.el` to address `the following functions might not be defined at runtime: cl-reduce, cl-mapcan`.

dgud:
Thanks.

PR #4543, opened 2021-02-22
erts: Fix fd leak when an active socket is closed remotely

### LINKED ISSUE #4540: file:send_file leaks file descriptors
**Describe the bug**
If send_file doesn't send all the bytes it leaks the file descriptor it duplicates. see: erts/emulator/drivers/common/inet_drv.c

```
        int dup_file_fd;        /* The file handle to send from; this is
                                 * duplicated when sendfile is issued to
                                 * reduce (but not eliminate) the impact of a
                                 * nasty race, so we have to remember to close
                                 * it. */
```

**To Reproduce**

```
defmodule Leak do

  def leak() do
    File.write!("/tmp/bigfile.txt", String.duplicate("A", 1_000_000))
    port = server()
    client(port)
    client(port)
    pid = :os.getpid()
    {out, _} = System.cmd("lsof", ["-p", to_string(pid)])
    IO.puts(out)
    IO.puts("SLEEPING")
    Process.sleep(300_000)
  end

  def server() do
    {:ok, server} = :gen_tcp.listen(0, [])
    Task.async(fn() -> server_loop(server) end)
    {:ok, {_ip, port}} = :inet.sockname(server)
    port
  end

  def client(port) do
    {:ok, socket} = :gen_tcp.connect('localhost', port, [:binary, {:active, false}])
    {:ok, data} = :gen_tcp.recv(socket, 100)
    IO.puts("RECEIVED_DATA: #{data}")
    :gen_tcp.close(socket)
    IO.puts("LEAKED FD")
  end

  def server_loop(socket) do
    {:ok, client} = :gen_tcp.accept(socket)
    task = Task.async(fn() -> handle_client() end)
    :ok = :gen_tcp.controlling_process(client, task.pid)
    send(task.pid, {:socket, client})
    server_loop(socket)
  end

  def handle_client() do
    client = receive do
      {:socket, client} -> client
    end

    try do
      {:ok, fd} = File.open("/tmp/bigfile.txt", [:binary, :raw, :read])
      try do
        :file.sendfile(fd, client, 0, 1_000_000, [])
      after
        :file.close(fd)
        IO.puts("closed_file")
      end
    after
      :gen_tcp.close(client)
      IO.puts("closed_client")
    end
  end
end
```

**Expected behavior**

file:send_file should ensure it closes the file descriptor it duplicates

**Affected versions**

I've tested it on 23.2.5 but I think it affects a lot of versions.

garazdawi:
Move sendfile dup socker close to stop callback instead of when the read/write error happens.

PR is missing tests for now.

Fixes #4540

### DIFF
--- a/erts/emulator/drivers/common/inet_drv.c
+++ b/erts/emulator/drivers/common/inet_drv.c
@@ -9922,6 +9922,15 @@ static void tcp_inet_stop(ErlDrvData e)
     tcp_close_check(desc);
     tcp_clear_input(desc);
 
+#ifdef HAVE_SENDFILE
+    if(desc->tcp_add_flags & TCP_ADDF_SENDFILE) {
+        desc->tcp_add_flags &= ~TCP_ADDF_SENDFILE;
+        close(desc->sendfile.dup_file_fd);
+        DEBUGF(("tcp_inet_stop(%p): SENDFILE dup closed %d\r\n",
+                desc->inet.port, desc->sendfile.dup_file_fd));
+    }
+#endif
+
     DEBUGF(("tcp_inet_stop(%ld) }\r\n", (long)desc->inet.port));
     inet_stop(INETP(desc));
 }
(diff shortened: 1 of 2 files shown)


garazdawi:
As it turns out, the testcases we had actually trigger the fault, there just wasn't any test to see if any fds leaked. So I've added a check to read the number of fds on Linux and check that they do no increase.

erts/emulator/drivers/common/inet_drv.c
```diff
@@ -10625,11 +10638,9 @@ static void tcp_inet_flush(ErlDrvData e)
 
 #ifdef HAVE_SENDFILE
     /* The old file driver aborted when it was stopped during sendfile, so
-     * we'll clear the flag and discard all output. */
+     * we'll clear the flag and discard all output. It is the job of
+     * tcp_inet_stop to close the extra sendfile fd. */
```
jhogberg, line 10642:
```suggestion
     * we'll discard all output. It is the job of tcp_inet_stop
     * to close the extra sendfile fd. */
```

jhogberg (approved):

PR #4545, opened 2021-02-24
Inline funs that are immediately used

### LINKED ISSUE #4019: ERL-889: Weird codegen when using scoping tricks
Original reporter: `fenollp`
Affected version: `OTP-22.0`
Component: `compiler`
Migrated from: https://bugs.erlang.org/browse/ERL-889

---

```
Scoping of `begin...end` can be improved when using `(fun () -> ... end)()`: with this trick variables bound inside the body of the anonymous function don't leak to the outter scope.
This is a very oft used trick and I have been asking for it to be optimized before.
cc https://bugs.erlang.org/browse/ERL-639
Since this is really just "safer scoping" IMO we shouldn't be paying the cost of anonymous function creation and calling at runtime. Instead the compiler should inline the call and rename some variables at compile time.

This issue notes that we're even sometimes paying that runtime cost for nothing:

{code:erlang}
-module(f).
-export([a/0]).
a() ->
    B = ((fun () -> A = 21, 2*A end)()),
    x:z(B).
{code}


{code:erlang}
{function, a, 0, 2}.
  {label,1}.
    {line,[{location,"f.erl",3}]}.
    {func_info,{atom,f},{atom,a},0}.
  {label,2}.
    {allocate,0,0}.
    {line,[{location,"f.erl",4}]}.
    {call,0,{f,8}}.
    {'%',{type_info,{x,0},{integer,42}}}.
    {move,{integer,42},{x,0}}.
    {line,[{location,"f.erl",5}]}.
    {call_ext_last,1,{extfunc,x,z,1},0}.

{function, '-a/0-fun-0-', 0, 8}.
  {label,7}.
    {line,[{location,"f.erl",4}]}.
    {func_info,{atom,f},{atom,'-a/0-fun-0-'},0}.
  {label,8}.
    {move,{integer,42},{x,0}}.
    return.
{code}

This was with OTP 22-rc1.
```

bjorng:
Funs can be used to hide local variables. Example:

    a() ->
        A = fun() -> A = 21, 2 * A end(),
        foo:bar(A).

To avoid the slight runtime cost for such use of funs, teach the
compiler to inline funs that are used only once immediately after
creation.

Resolves #4019

### DIFF
--- a/lib/compiler/src/sys_core_fold.erl
+++ b/lib/compiler/src/sys_core_fold.erl
@@ -274,7 +274,7 @@ expr(#c_let{}=Let0, Ctxt, Sub) ->
 	    %% The argument for the let is "simple", i.e. has no
 	    %% complex structures such as let or seq that can be entered.
 	    ?ASSERT(verify_scope(Let, Sub)),
-	    opt_simple_let(Let, Ctxt, Sub);
+	    opt_fun_call(opt_simple_let(Let, Ctxt, Sub));
 	Expr ->
 	    %% The let body was successfully moved into the let argument.
 	    %% Now recursively re-process the new expression.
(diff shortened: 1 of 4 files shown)


jhogberg (approved):

josevalim:
Neat! Quick question: does that work if the function is invoked once but across multiple clauses? For example:

```erl
Fun = fun(B) -> A = 21, 2 * B end(),

case Condition of
  true -> Fun(1);
  false -> Fun(-1)
end.
```

bjorng:
> Quick question: does that work if the function is invoked once but across multiple clauses?

No.

PR #4547, opened 2021-02-25
cover: Ensure that LC filters are compiled in guard context

bjorng:
A subtle detail in the semantics of list comprehensions is that
filters will be compiled as guard expressions if possible. That means
that the following list comprehension will never raise an exception as
long as it is passed a proper list:

    lc(L) ->
        [V || V <- L, V =:= a orelse element(1, V) =:= a].

Thus, the following call:

    lc([a, {a,b}, {}, "ignore"])

will return `[a, {a,b}]`. When the calls `element(1, {}` and
`element(1, "ignore")` fail, no exception is raised and the
corresponding list elements are discarded.

The `cover` tools rewrites rewrites `andalso` and `orelse` to `case`
constructions in order to produce better coverage data. For this example,
that rewriting would change the error behavior of the list comprehension:

    lc(L) ->
        [V || V <- L,
              case V =:= a of
                   true -> true;
                   false -> element(1, V) =:= a
              end].

Because of the `case`, the filter is no longer a guard, and the
`element(1, {}` call will now raise a `badarg` exception.

To avoid that problem, stop rewriting `andalso`/`orelse` in filters
that are guard tests.

### DIFF
--- a/lib/tools/src/cover.erl
+++ b/lib/tools/src/cover.erl
@@ -1851,8 +1851,17 @@ expand(Expr) ->
     Expr1.
 
 expand({clause,Anno,Pattern,Guards,Body}, Vs, N) ->
+    %% We must not expand andalso/orelse in guards.
     {ExpandedBody,N2} = expand(Body, Vs, N),
     {{clause,Anno,Pattern,Guards,ExpandedBody},N2};
+expand({lc,Anno,Expr,Qs}, Vs, N) ->
+    {ExpandedExpr,N2} = expand(Expr, Vs, N),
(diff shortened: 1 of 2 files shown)


uabboli:
I think you should just exclude filters (in Qs) such that
erl_lint:is_guard_test/1 returns true. Now non-guard filters and
generators are not expanded, which is a pity.

bjorng:
> I think you should just exclude filters (in Qs) such that erl_lint:is_guard_test/1 returns true.

Done.

uabboli (approved):

Issue #3150, opened 2016-10-13
ERL-282: escript fails when redirecting stdout to less and less terminates


OTP-Maintainer:
Original reporter: `jj1bdx`
Affected version: `OTP-19.1`
Component: `erts`
Migrated from: https://bugs.erlang.org/browse/ERL-282

---

```
When executing the attached script {{lessbug.escript}} as
{{escript lessbug.escript | less}} and terminates less before displaying all the results,
BEAM crashes with a dump.
The system is macOS 10.12.0, Xcode 8.0, with OTP 19.1.3, compiled with the following options:
{{--enable-m64-build --enable-darwin-64bit --disable-hipe --enable-vm-probes --with-dynamic-trace=dtrace --disable-native-libs --enable-kernel-poll --without-odbc --enable-threads --enable-smp-support --with-wx --with-ssl=/usr/local/opt/openssl}}

This crash dump is not always reproducible: the output may not respond at all sometimes.

```

garazdawi:
A bit more info here: #4553

garazdawi:
I want to put down some of the reasons why I think that the issues linked above are duplicates of this.

In OTP-18 I introduced [this](2e84a2ab5e0f0c82c2d79f69cd2f4bff762762a4) commit that sets the stdin and stdout fds to NONBLOCKING on unix. The reason why this change was done is described in the commit message:

> Without this change an stdout write could block an entire scheduler if for some reason the pseudo tty on the other side does not consume the output of the Erlang shell.

As it turns out, this change comes with a host of issues for programs using the stdin/stdout fds of Erlang.

1) Firstly, programs that read the stdout from Erlang may get an FD as its stdin that already is in NONBLOCKING mode. Some versions of programs such as `less` and `cat` are not prepared for this and behave badly. This is what happens when you do `erl | cat`.

2) Secondly, when Erlang exits without setting its stdout/stdin fds to BLOCKING, some versions of bash/sh will not reset it and all of the sudden bash starts working with a NONBLOCKING fd, which it is not prepared for. This is when you get a garbled terminal.

A partial solution to this problem could be to call `sys_tty_reset` (as @jktomer points out in #4553) when SIGINT is encounter. However, this is not a foolproof method as a process can be killed by SIGKILL and then the same problem would remain.

So the solution for this in my mind is to remove the NONBLOCKING behaviour from the `ttsl_driver` and the `fd_driver`. This brings us back to the original problem of what to do when we try to write to stdout and it blocks.

[...]

PR #4553, opened 2021-02-25
fix: Handle SIGUSR1 even under +B i / +B d

### LINKED ISSUE #4552: SIGUSR1 kills beam immediately under +B i/d
**Describe the bug**
The OTP [documentation](https://erlang.org/doc/man/kernel_app.html) states that SIGUSR1 is normally handled by causing a crash dump and halt Erlang. However, when the emulator flags `+B i` or `+B d` are passed, the system call to handle SIGUSR1 is not invoked, so SIGUSR1 will actually cause BEAM to die immediately.

**To Reproduce**
1. start Erlang with `+B i` or `+B d`.
2. send SIGUSR1 to the BEAM process.

**Expected behavior**
Erlang exits producing a crash dump with the slogan "Received SIGUSR1"

**Actual behavior**
Erlang exits immediately; no crash dump is produced (nor does any other cleanup happen).

**Affected versions**
Apparently since at least R13B03.

**Additional context**
See the pull request accompanying this issue.

jktomer:
The OTP documentation states that SIGUSR1 causes a crash dump unconditionally.
However, under +B i or +B d, the call to install the SIGUSR1 handler was not
made before this change, causing SIGUSR1 to kill process immediately instead.

Fixes #4552.

### DIFF
--- a/erts/emulator/sys/unix/sys.c
+++ b/erts/emulator/sys/unix/sys.c
@@ -677,9 +677,6 @@ void erts_replace_intr(void) {
 void init_break_handler(void)
 {
    sys_signal(SIGINT,  request_break);
-#ifndef ETHR_UNUSABLE_SIGUSRX
-   sys_signal(SIGUSR1, generic_signal_handler);
-#endif /* #ifndef ETHR_UNUSABLE_SIGUSRX */
    sys_signal(SIGQUIT, generic_signal_handler);
 }
 
@@ -694,6 +691,9 @@ void
 erts_sys_unix_later_init(void)
 {
     sys_signal(SIGTERM, generic_signal_handler);
+#ifndef ETHR_UNUSABLE_SIGUSRX
+   sys_signal(SIGUSR1, generic_signal_handler);
+#endif /* #ifndef ETHR_UNUSABLE_SIGUSRX */
 
     /* Ignore SIGCHLD to ensure orphaned processes don't turn into zombies on
      * death when we're pid 1. */
--- a/erts/emulator/test/os_signal_SUITE.erl
+++ b/erts/emulator/test/os_signal_SUITE.erl
@@ -52,14 +52,14 @@ suite() ->
 all() ->
     case os:type() of
         {win32, _} -> [];
-        _ -> [set_unset,
-              t_sighup,
+        _ -> [t_sighup,
               t_sigusr1,
               t_sigusr2,
               t_sigterm,
               t_sigalrm,
               t_sigchld,
-              t_sigchld_fork]
+              t_sigchld_fork,
+              set_unset]
     end.
 
 init_per_testcase(Func, Config) when is_atom(Func), is_list(Config) ->
@@ -135,7 +135,7 @@ t_sighup(_Config) ->
 t_sigusr1(_Config) ->
     Pid1 = setup_service(),
     OsPid = os:getpid(),
-    os:set_signal(sigusr1, handle),
+    % shouldn't need to explicitly set_signal to handle USR1, it's the default
     ok = kill("USR1", OsPid),
     ok = kill("USR1", OsPid),
     ok = kill("USR1", OsPid),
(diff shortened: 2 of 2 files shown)


ferd:
This might be the fix behind the shell bugs we see with escripts and rebar3 when exiting the instance.

garazdawi:
> This might be the fix behind the shell bugs we see with escripts and rebar3 when exiting the instance.

Shell bug?

ferd:
> > This might be the fix behind the shell bugs we see with escripts and rebar3 when exiting the instance.
> 
> Shell bug?

I opened it as https://bugs.erlang.org/browse/ERL-1162 and it was closed as a duplicate.

When starting `erl` normally, once can exit the shell without bad consequences on the output when exiting with `CTRL+C`:

```
$ erl
Erlang/OTP 22 [erts-10.5.3] [source] [64-bit] [smp:8:4] [ds:8:4:10] [async-threads:1] [hipe]
Eshell V10.5.3 (abort with ^G)
1>
BREAK: (a)bort (c)ontinue (p)roc info (i)nfo (l)oaded
(v)ersion (k)ill (D)b-tables (d)istribution
^C% λ ~ → ls
<regular output>
```

However, when the `+B` option is passed, the same `CTRL+C` exit causes the resulting listing to be broken, and generally the terminal looks like it has no concept of its own size anymore:

```
$ erl +B
Erlang/OTP 22 [erts-10.5.3] [source] [64-bit] [smp:8:4] [ds:8:4:10] [async-threads:1] [hipe]
Eshell V10.5.3 (abort with ^G)
1>
$ ls
<broken output>
```
I've attached the broken output as an image: 
image


This problem is made a bit worse because escripts mandate the `+B` argument to work, so any interactive shell (i.e. rebar3 shell) gets broken output automatically

garazdawi:
> I opened it as https://bugs.erlang.org/browse/ERL-1162 and it was closed as a duplicate.

Ah, that good oldie. I've linked in all the info in the github issue tracking this.

jktomer:
Not restoring terminal state on Ctrl-C with +B d is related, but not identical, to this problem and the one described in #3150. This commit handles SIGUSR1 only; for +B d, we would want to handle SIGINT with a handler to run (at least) the tty cleanup code in `sys_tty_reset` before exiting. This shouldn't present a compatibility problem with existing Erlang code using `os:set_signal` because any call to `os:set_signal` would replace whatever signal handler is set up at init time.

max-au:
I can confirm that this change does not fix garbled terminal after exiting `rebar3 shell` with Ctrl+C.

jktomer:
Gentle ping: any objection to merging this? I'm having a hard time envisioning a system that *depends* on SIGUSR1 hard-killing BEAM (note that any action by user code to change handling of SIGUSR1, in particular any call to `os:set_signal/2`, or any NIF or port driver that installs a signal handler after early BEAM initialization, would render this change moot, so only programs that did nothing with respect to SIGUSR1 would find themselves affected by this change, and the effect will always be that SIGUSR1 triggers a crashdump and orderly emulator exit rather than brutally killing the process as SIGKILL would).

rickard-green:
Currently a bit short on time. Will have a look at it at the end of the week.

rickard-green:
Short version: I'm for this, but into the `master` branch instead of into the `maint` branch. That is, not until OTP 24.

SIGUSR1 has always (at least since the 90ies) been lumped together with SIGINT and SIGQUIT in the "break handling" and restored to their default actions if +B (or +Bd) was passed as argument at startup. I don't see what SIGUSR1 has to do with "break handlng", though. I guess that someone wanted a way to restore all signal actions to their defaults that could be restored to their defaults without breaking the system.

+Bi was introduced in order to prevent key combinations (Ctrl-C, Ctrl-\\, Ctrl-Z causing signals being sent to the emulator) resulting in the emulator being suspended, halted, etc. The SIGUSR1 probably by mistake got set to default action, when +Bi is passed, just because it had always been lumped together with SIGINT and SIGQUIT in the initialization of "break handling". I don't see any reason why the behavior of SIGUSR1 should change when +Bi is passed.

I don't think we should introduce this change in OTP 23 since it changes a behavior that has been there for a very long time, but I think we should change it in OTP 24 since current behavior makes no sense.

jktomer:
Thanks for the review! Rebased onto master.

PR #4563, opened 2021-03-02
Enhance the optimization of the try/catch elimination optimization

bjorng:
The `beam_ssa_opt:ssa_opt_try/1` optimization would sometimes fail to
eliminate a redundant `try`/`catch` construct because it only looked
for `new_try_tag` instructions at the beginning of an SSA block. Here
is an example:

    foobar({Anno}=T, V) ->
        try
            setelement(1, T, V)
        catch
            _Class:_Error:_Stack ->
                Anno
        end.

In the SSA code, the `get_tuple_element` instruction from the tuple
matching in the function head is at the beginning of the block with the
`new_try_tag` instruction:

    8:
      _10 = get_tuple_element _0, `0`
      @ssa_catch_tag = new_try_tag `'try'`
      br @ssa_catch_tag, ^6, ^5

Generalize the optimization to handle this example. While at it,
also make the optimization run faster by eliminating redundant
set update operations.

### DIFF
--- a/lib/compiler/src/beam_ssa_opt.erl
+++ b/lib/compiler/src/beam_ssa_opt.erl
@@ -1494,35 +1494,45 @@ reduce_try_is([], Acc) ->
 %% unreachable landing pads, removing them from the set whenever we see a
 %% branch to that block. When we encounter a `new_try_tag` instruction that
 %% references a block in the unreachable set, we'll remove the try/catch.
-trim_try([{L, #b_blk{is=[#b_set{op=new_try_tag,dst=Tag}],
-                     last=Last0}=Blk0} | Bs],
-         Unreachable0, Killed0, Acc) ->
-    #b_br{succ=SuccLbl,fail=PadLbl} = Last0,
-    case sets:is_element(PadLbl, Unreachable0) of
-        true ->
-            %% The landing pad can't be reached in any way, remove the entire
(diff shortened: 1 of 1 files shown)


gomoripeti:
I didn't know about the existence of try/catch elimination, very nice. Would it be possible to notify the user about dead code when try/catch elmination is applicable? This sounds useful to clean up code. If not a warning then maybe something similar to bin_opt_info? Or does this mostly occur with macros/generated code and not explicitly written code?

bjorng:
Yes, I would say that redundant catches is far more likely to occur with macros or generated code.

We are unlikely to add more "opportunistic" warnings (warnings emitted for issues noticed while optimizing, such as "this clause ... cannot match because of a previous clause ..."). They are problematic for several reasons, for one that we will have to make an effort to avoid generate annoying warnings for perfectly good code, and for another that users that have got used to such warnings will consider it a bug if some suspicious construct does not generate a warning.

Regarding warnings such as `bin_opt_info` and the upcoming `recv_opt_info`, we will only add such warnings when there is substantial potential gain when the corresponding optimizations can be applied. Redundant try/catch elimination is definitely not one of those optimizations.

gomoripeti:
makes sense, thanks Bjorn

jhogberg (approved):

PR #4564, opened 2021-03-02
inet_res: add option to include the DNS reply in nxdomain errors

RoadRunnr:
A nxdomain error from a name server does contain the SOA record if
the domain exists at all. This record is usefull to determine a TTL
for negative caching of the failed entry.

### DIFF
--- a/lib/kernel/src/inet_res.erl
+++ b/lib/kernel/src/inet_res.erl
@@ -61,7 +61,8 @@
       | {retry, integer()}
       | {timeout, integer()}
       | {udp_payload_size, integer()}
-      | {usevc, boolean()}.
+      | {usevc, boolean()}
+      | {nxdomain_reply, boolean()}.
 
 -type nameserver() :: {inet:ip_address(), Port :: 1..65535}.
 
@@ -262,7 +263,7 @@ do_nslookup(Name, Class, Type, Opts, Timeout) ->
 %% options record
 %%
 -record(options, { % These must be sorted!
-	  alt_nameservers,edns,inet6,nameservers,recurse,
+	  alt_nameservers,edns,inet6,nameservers,nxdomain_reply,recurse,
 	  retry,servfail_retry_timeout,timeout,udp_payload_size,usevc,
 	  verbose}). % this is a local option, not in inet_db
 %%
@@ -304,6 +305,12 @@ make_options([{verbose,Val}|Opts]=Opts0, [verbose|Names]=Names0) ->
        true ->
 	    erlang:error(badarg, [Opts0,Names0])
     end;
+make_options([{nxdomain_reply,Val}|Opts]=Opts0, [nxdomain|Names]=Names0) ->
+    if is_boolean(Val) ->
+	    [Val|make_options(Opts, Names)];
+       true ->
+	    erlang:error(badarg, [Opts0,Names0])
+    end;
 make_options([{Opt,Val}|Opts]=Opts0, [Opt|Names]=Names0) ->
     case inet_db:res_check_option(Opt, Val) of
 	true ->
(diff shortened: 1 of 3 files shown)


RaimoNiskanen:
I would like to have a boolean option instead, such as `{nxdomain_reply, boolean()}`.

Also, you need to rebase to a later master due to merge conflicts.

RaimoNiskanen:
Is negative caching an important feature we should add to the `inet_res`/`inet_db` cache?

RoadRunnr:
> Is negative caching an important feature we should add to the `inet_res`/`inet_db` cache?

Yes and No. Negative caching is one part of what I need in relation to DNS. Other things are parallel resolution attempts on A and AAAA records and caching NAPTR and SRV records.

I'm not sure that all that makes sense in the current scope on inet_res. It basically is just a backend for gethostbyname and friends.

A general purpose DNS client in OTP that can do more that just gethostbyname would IMHO be great.

RoadRunnr:
rebased to master

RaimoNiskanen:
I have a branch with practically parallel lookups for all records, should create a PR of it see what the Community thinks of it.  We have internal customers that are depending on `inet_res` limiting its socket usage, that may have problems with it, though...

RoadRunnr:
> > A general purpose DNS client in OTP that can do more that just gethostbyname would IMHO be great.
> 
> How far from useful are `inet_res:lookup/3,4,5` and `inet_res:resolve/3,4,5`?

Very usable when combined with some custom logic and a custom cache ([here](https://github.com/travelping/ergw/blob/master/src/ergw_inet_res.erl) -- that code also mangles the entries a bit before putting them into the cache to be more aligned for my S-NAPTR use case ).

What my wrapper also does, is converting all list strings to binaries to save some space. I was thinking about adding a option to inet_res to use binaries, but the changes would be quite extensive.

RaimoNiskanen:
Dialyzer error:
```
[2021-03-09T19:01:25.671Z] inet_res.erl:308:1: The pattern 
[2021-03-09T19:01:25.671Z]          <Opts0 = [{nxdomain_reply, Val} | Opts],
[2021-03-09T19:01:25.671Z]           Names0 = [nxdomain | Names]> can never match the type 
[2021-03-09T19:01:25.671Z]          <[tuple()],
[2021-03-09T19:01:25.671Z]           ['alt_nameservers' | 'edns' | 'inet6' | 'nameservers' |
[2021-03-09T19:01:25.671Z]            'nxdomain_reply' | 'recurse' | 'retry' |
[2021-03-09T19:01:25.671Z]            'servfail_retry_timeout' | 'timeout' | 'udp_payload_size' |
[2021-03-09T19:01:25.671Z]            'usevc' | 'verbose']>
```
I will investigate a bit more...

lib/kernel/src/inet_res.erl
```diff
@@ -304,6 +305,12 @@ make_options([{verbose,Val}|Opts]=Opts0, [verbose|Names]=Names0) ->
        true ->
 	    erlang:error(badarg, [Opts0,Names0])
     end;
+make_options([{nxdomain_reply,Val}|Opts]=Opts0, [nxdomain|Names]=Names0) ->
+    if is_boolean(Val) ->
+	    [Val|make_options(Opts, Names)];
+       true ->
+	    erlang:error(badarg, [Opts0,Names0])
+    end;
```
RaimoNiskanen, line 313: I think these two clauses should be merged like this, to avoid misspelling of a `Name` occurence:
``` erlang
make_options([{Name,Val}|Opts]=Opts0, [Name|Names]=Names0)
  when Name =:= verbose;
       Name =:= nxdomain_reply ->
    if is_boolean(Val) ->
        [Val|make_options(Opts, Names)];
```

lib/kernel/src/inet_res.erl
```diff
@@ -313,6 +320,8 @@ make_options([{Opt,Val}|Opts]=Opts0, [Opt|Names]=Names0) ->
     end;
 make_options(Opts, [verbose|Names]) ->
     [false|make_options(Opts, Names)];
+make_options(Opts, [nxdomain_reply|Names]) ->
```
RaimoNiskanen, line 323: Would it be slightly more beautiful to have these clauses in sorted order too, i.e `nxdomain_reply` before `verbose`?

lib/kernel/src/inet_res.erl
RaimoNiskanen: That misspeling is probably what caused the Dialyzer error

RaimoNiskanen:
> Other things are parallel resolution attempts on A and AAAA records ...

I have created #4610 with a suggested algorithm for parallel lookups.

RaimoNiskanen:
No problems in daily tests.  Thank you for the pull request!

Issue #4469, opened 2020-11-24
ERL-1418: Allow size of disk logs to be changed


OTP-Maintainer:
Original reporter: `JIRAUSER17013`
Affected version: `OTP-22.1.3`
Component: `kernel`
Migrated from: https://bugs.erlang.org/browse/ERL-1418

---

```
We wanted more log capacity for our Erlang program.  We changed the log file size parameters in app.config:
{code}
    {handler, disk_log, logger_disk_log_h, #{
         config => #{
           max_no_files => 8,
           max_no_bytes => 4194304,
{code}
But, this didn't go well.  Our app crashes on startup with message "{kernel,{handler_not_added,{size_mismatch" :

{code:erlang}
2020-09-25 08:28:00.322388 crash_report        #{label=>{proc_lib,crash},report=>[[{initial_call,{application_master,init,['Argument__1','Argument__2','Argument__3','Argument__4']}},{pid,<0.46.0>},{registered_name,[]},{error_info,{exit,{bad_return,{{kernel,start,[normal,[]]},{'EXIT',{{badmatch,{error,{bad_config,{handler,{kernel,{handler_not_added,{size_mismatch,{2097152,5},{4194304,8}}}}}}}},[{kernel,start,2,[{file,"kernel.erl"},{line,38}]},
{application_master,start_it_old,4,[{file,"application_master.erl"},{line,277}]}]}}}},[{application_master,init,4,[{file,"application_master.erl"},{line,138}]},
{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,249}]}]}},{ancestors,[<0.45.0>]},{message_queue_len,1},{messages,[{'EXIT',<0.47.0>,normal}]},
{links,[<0.45.0>,<0.44.0>]},{dictionary,[]},{trap_exit,true},{status,running},{heap_size,987},{stack_size,27},{reductions,223}],[]]}
{code}

It seems bad that the contents of log directories can cause the logger to prevent the application from running!

It would be better if the logger module could automatically resize the existing log files, or if that is not possible, automatically remove the existing logs and start new logs with the correct size.
```

uabboli:
Thanks. We've decided to change disk_log as per you suggestion.

PR #4565, opened 2021-03-02
kernel: Let option 'size' of disk_log:open/1 change size

uabboli:
If the value of option 'size' does not match existing wrap log size,
the 'size_mismatch' error is no longer returned, but the size of the
log changed. This applies only if the disk log is not already open.

We have made this incompatible modification to help customers handle
upgrade scenarios.

Fixes GH-4469.


garazdawi (approved):
looks good to me. Once this is merged I will rebase my shell_history test cases on this and see if any other faults show up.

PR #4566, opened 2021-03-02
wxNotificationMessage macos & win polish

dominicletz:
This patch embeds for MacOS a .plist file into the erlang executable. This .plist is there to allow:
1. Showing notifications (under MacOS they will not show at all if no .plist file is found)
2. Using wxWebView for http://localhost/ domains (creating platform parity)

And on windows, this patch adds new methods `UseTaskBarIcon` and `MSWUseToasts` to avoid having icon duplication and match icons in the start menu. For windows to go one step further I was thinking of setting an ApplicationID https://nsis.sourceforge.io/ApplicationID_plug-in but that would require adding an NSIS plugin to the build process and wasn't quite sure if that would make sense for the default distribution.

Let me know if you want me to split any of this out into independent pulls, or whether I should include the NSIS change mentioned above.

@dgud


erts/etc/darwin/Info.plist
```diff
@@ -0,0 +1,29 @@
...
+<dict>
+	<key>CFBundleName</key>
+	<string>Erlang</string>
+	<key>CFBundleDisplayname</key>
+	<string>Erlang</string>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>English</string>
+	<key>CFBundleIdentifier</key>
+	<string>org.erlang.beam</string>
+	<!-- <key>NSUserNotificationAlertStyle</key>
```
dgud, line 13: Can you remove the comments or comment why you still have these left.

dgud:
I don't see the reason of adding it to the windows installer,
the way I see it is that the erlang installer installs a developer package, user programs should have their own installer.
The same goes for Mac (and unix/linux) the user will need create its own application package for distribution, 
but it might be a good thing to have a default Info.plist, so programs like the debugger/observer works as intended.

I know so little about these things, please educate me.

erts/etc/darwin/Info.plist
dominicletz: done

dominicletz:
> but it might be a good thing to have a default Info.plist, so programs like the debugger/observer works as intended.

I like that reasoning as leading. It makes sense to keep the .plist but not add application_id unless/until the debugger is using it.

PR #4570, opened 2021-03-03
￼ Refactored erl_prettypr.erl

matyasmarkovics:
I was bold enough to start a pet-project refactoring the `erl_prettypr.erl` module. This has been a long-time coming and **I still consider it Work-In-Progress**. Never-the-less, I thought I'd share the current state to see if there's interest in this work.

I plan to:
* Rebase the branch and write meaningful commit-messages
* Expand on the documentation
* Extend the tests to verify the formatted layout as opposed to the current verification of compile-abilty

Further-more:
* Make the layout more configurable
* Keep macro-s as is. - I've noticed macro-s are converted, this should be disabled by default, but optional.
* Create alternative callbacks, to the current `hook`, these functions would be specific for each Tree-type.
* ... I'm open to suggestions

I'm aware that there have been attempts to "fix-up" this module and those attempts ended-up in separate stand-alone projects. Despite of that I think there still is a value in improving the code-formatter shipped with Erlang/OTP, as it may be used by other internal libraries.

I've decided to simplify the module before attempting to expand on the Layout-Configurability. So, I consider this PR as stage-1 to be honest towards the end-goal: "Make the layout configurable." - Solving the leading-comma formatting would be a good driving principle for configurabilty.

There are a lot of similarities between Erlang statements, like case-clauses and function-clauses, or between data-structure like lists, tuples and binaries, so I've tried to collapse these cases into one, so that they have the same layout.

The produced code-layout relies only on what's possible with the `prettypr`. Only the 4 different layouts: `sep/1`, `par/1`, `above/2` and `beside/2` are used. These are surprisingly versatile and I've found the `float` layout nondeterministic (don't really understand how it supposed to work).

With this PR a simple, sensible layout is produced for those who just want it to work without any configuration.


paulo-ferraz-oliveira:
@elbrujohalcon, this might interest you.

lib/syntax_tools/src/epp_dodger.erl
```diff
@@ -796,36 +797,69 @@ rewrite(Node) ->
...
+            fun erl_syntax:type_application_name/1,
+            fun erl_syntax:type_application_arguments/1);
+    user_type_application ->
+        rewrite_application(
+            Node,
+            fun erl_syntax:user_type_application_name/1,
+            fun erl_syntax:user_type_application_arguments/1);
+    attribute ->
+        Name = erl_syntax:attribute_name(Node),
+        case catch erl_syntax:concrete(Name) of
```
maxnordlund, line 817: I think this can be simplified a bit to both avoid an unnecessary `catch` and make it clear what kind of node are allowed as the attribute name, i.e. an atom node.

```suggestion
        case erl_syntax:atom_value(Name) of
```

lib/syntax_tools/src/epp_dodger.erl
```diff
@@ -796,36 +797,69 @@ rewrite(Node) ->
...
+            spec ->
+                [SpecTuple] = erl_syntax:attribute_arguments(Node),
+                [FuncName, FuncTypes] = erl_syntax:tuple_elements(SpecTuple),
+                Clauses = erl_syntax:concrete(FuncTypes),
```
maxnordlund, line 821: I like the trick with `concrete`/`abstract` here. Very nice 😄

PR #4572, opened 2021-03-03
Update mnesia.erl

dong50252409:
fix the create_option() spec description error


dgud:
Can you rebase that on top of the maint branch instead, so it can be included in 23.3.

dgud:
Replaced by #4578

PR #4575, opened 2021-03-03
Optimise mnesia table loading time for busy tables

ssmyczynski:
This PR optimizes mnesia table loading time (on adding new table copy or just starting another mnesia node) for busy tables (a lot changes per second).

The problem with current implementation is that table receiver first subscribes for table changes, then starts downloading the content of the table from sender and only after receiving whole content of the table handles all the subscribed changes that happened in the meantime. For busy tables this causes receiver message queue to grow very quickly slowing down the downloading process.

The simple solution is to receive all mnesia_table_events that comes during table copying and store them in private ets table for later processing after table is succesfuly downloaded. 

Currently:
pre

After applying changes in this PR:
post

### DIFF
--- a/lib/mnesia/src/mnesia_loader.erl
+++ b/lib/mnesia/src/mnesia_loader.erl
@@ -320,6 +320,11 @@ start_remote_sender(Node,Tab,Storage) ->
 
 table_init_fun(SenderPid, Storage) ->
     fun(read) ->
+	    % We want to store subscribed mnesia table events received during
+	    % table copying for later processing to not let receiver message queue
+	    % to grow too much (which in consequence would slow down the whole copying process)
+	    SubscrCache = ets:new(subscr_cache, [private, ordered_set]),
+	    put(mnesia_table_receiver_subscr_cache, SubscrCache),
 	    Receiver = self(),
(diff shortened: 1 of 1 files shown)


ssmyczynski:
[demo-script.zip](https://github.com/erlang/otp/files/6081978/demo-script.zip)

dgud:
Why an ets table and not the simple solution of just a adding to a list, performance?

dgud:
Nice find, though..

ssmyczynski:
> Why an ets table and not the simple solution of just a adding to a list, performance?

Actually this was my first thought as also keeping order of events is simple in list - you prepend the list with new events, and just before replying those events just do once a lists:reverse - so the performance was not the reason of using private ets table.

The main reason is that I didn't want to change the return spec of a function returned by table_init_fun (and later make_table_fun), because that it is used later internally in ets module (see ets:init_table_continuation) for ram_copies/disc_copies tables, and in  dets module for disc_only_copies, so I decided to store SubscrCache in process dictionary - and storing a huge list in process dictionary is not a good idea at all :)

And as a bonus we don't have to worry about garbage collection when collecting those subscriptions.

PR #4578, opened 2021-03-04
Fix create_option() spec description error

### LINKED ISSUE #4569: spec description error
https://github.com/erlang/otp/blob/4cdd991df2885398a7c00c85b4340c34197b92d3/lib/mnesia/src/mnesia.erl#L145

dong50252409:
Closes #4569


dgud:
I believe you could have forced pushed you branch on the other PR.
Thanks for finding the bug and the fix.

PR #4579, opened 2021-03-04
Show docs from any textual chunk in the shell

josevalim:
This pull request augments shell_docs to also
handle any chunk with type of "text". Those will
be rendered as is and allow the Erlang shell to
show the documentation of projects that do not
adhere to the application/erlang+html format.
This is useful because many documentation formats
(Markdown, Asciidoc, etc) are readable in their
native format.

### DIFF
--- a/lib/stdlib/doc/src/shell_docs.xml
+++ b/lib/stdlib/doc/src/shell_docs.xml
@@ -54,10 +54,11 @@ Since:
     > maps:new().
     #{}
 </code>
-    <p>This module can only render EEP-48 documentation of the format
-      <c>application/erlang+html</c>. For more information about this format see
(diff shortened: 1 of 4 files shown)


josevalim:
One downside of this PR is that if someone has docs of format text/html, it will also print those as is. The alternative here is to explicitly list which formats we want to allow (text/markdown, text/asciidoc, etc) but I think the current approach is a valid compromise, especially because any structured textual format is still readable. Here is an example of how Elixir docs would look like from the Erlang shell:

Screenshot 2021-03-04 at 09 31 23

garazdawi:
Merged! Thanks! I agree with your tradeoffs in the implementation. We can change it in the future if someone want to emit html in as documentation.

Issue #4573, opened 2021-03-03
OpenSSL configuration leads to build failure in otp_test_engine.c


erwan-le-pape:
**Describe the bug**
In `lib/crypto/c_src/otp_test_engine.c`, function prototypes are declared in a twice-conditional block that depends on `HAVE_EC` (from l.58 to l.378) and `FAKE_RSA_IMPL`. Functions are however are declared on the sole `FAKE_RSA_IMPL` condition (from l.380 to l.456). This breaks compilation on OpenSSL that matches conditions for `HAVE_EC` to be false with typical hardening option `-Werror=format-security`.

**To Reproduce**
Compiling on an OpenSSL 1.1.1i with `OPENSSL_NO_ECDSA` and `-Werror=format-security` in `CFLAGS` works for me, I expect any OpenSSL >= 0.9.8o and with any of `OPENSSL_NO_EC`, `OPENSSL_NO_ECDH` or `OPENSSL_NO_ECSDA` would do the trick.

**Expected behavior**
The function implementations should likely be within the `#if defined(HAVE_EC)` section, in addition to which, I'm unsure why the `HAVE_EC` check looks beyond `OPENSSL_NO_EC` since it doesn't actually depend on `OPENSSL_NO_ECDSA` (it compiles just fine without checking for OPENSSL_NO_ECDSA on OpenSSL 1.1.1i and OTP 23.2.7, although I can't confirm backwards compatibility with obsolete OpenSSL <= 1.1.0).

**Affected versions**
20.3.8.15 through 23.2.7 (verified on 23.2.6 & 23.2.7).

**Additional context**
Compiling fails because of the `implicit declaration of function` but I've left the information the entire file.
```
otp_test_engine.c:387:5: warning: no previous prototype for ‘test_rsa_sign’ [-Wmissing-prototypes]
 int test_rsa_sign(int dtype,
     ^
otp_test_engine.c: In function ‘test_rsa_sign’:
otp_test_engine.c:416:9: error: implicit declaration of function ‘add_test_data’ [-Werror=implicit-function-declaration]
         add_test_data(sigret, (unsigned int)slen); /* The signature is 0,1,2...255,0,1... */
         ^
otp_test_engine.c: At top level:
otp_test_engine.c:426:5: warning: no previous prototype for ‘test_rsa_verify’ [-Wmissing-prototypes]
 int test_rsa_verify(int dtype,
     ^
otp_test_engine.c: In function ‘test_rsa_verify’:
[...]

PR #4580, opened 2021-03-04
Make OpenSSL functions depend on the same conditions as their prototypes

erwan-le-pape:
This PR fixes the otp_test_engine issue where functions don't depend on the same CPP conditions as their prototypes, leading to broken builds in cases where one is true and the other isn't (#4573).


HansN:
Thanks for the PR!

PR #4587, opened 2021-03-06
Expose math:nearbyint/1

josevalim:
IEEE 754 proposes five rounding modes:

  1. To positive infinity (achievable with ceil)
  2. To negative infinity (achievable with floor)
  3. To zero (achievable with floor+ceil)
  4. Ties away from zero (achievable with round)
  5. Ties to even (not achievable)

Erlang behaves according to C in those operations
but it doesn't expose an operation for "ties to even",
which is the recommended operation according to the
standard. We can provide the missing operation by
exposing math:nearbyint/1, added on C99.

### DIFF
--- a/erts/emulator/beam/erl_math.c
+++ b/erts/emulator/beam/erl_math.c
@@ -22,10 +22,12 @@
 #  include "config.h"
 #endif
 
+#include <fenv.h>
 #include "sys.h"
 #include "erl_vm.h"
 #include "global.h"
 #include "erl_process.h"
+#include "erl_math.h"
 #include "error.h"
 #include "bif.h"
 #include "big.h"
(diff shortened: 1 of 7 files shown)


josevalim:
In theory, this behaviour should be consistent across OSes and machines, and I am hoping Erlang's OTP build can confirm if that is effectively the case.

If it is confirmed and there is a chance we can fit this in OTP 24, it would be awesome, but I understand if it is too late for that. Thank you! <3

jhogberg (commented):
Thanks for the PR! I'll see if we can squeeze this in :)

josevalim:
@jhogberg all feedback addressed, thank you!

erts/emulator/beam/erl_math.c
```diff
@@ -293,3 +294,82 @@ BIF_RETTYPE math_fmod_2(BIF_ALIST_2)
...
+	}
+	BIF_ERROR(BIF_P, BADARG);
+    }
+
+    /* get the float */
+    GET_DOUBLE(BIF_ARG_1, f);
+
+    /* Round it and return the resultant integer. Note that the current rounding mode
+     * must be "to nearest" for `nearbyint` to tie to even. */
+    ASSERT(1 == FLT_ROUNDS);
```
jhogberg, line 316: We've decided to take this in, but we'd like the rounding mode to be set to `FE_TONEAREST` and then restored before returning. The initial value is implementation-defined and we can't rely on setting it on boot as NIFs and drivers can change it at will.

https://en.cppreference.com/w/c/numeric/fenv/feround

josevalim: Thank you! Quick question: is changing it thread-safe?

jhogberg: Yes, it's thread-local.

josevalim: In any case, I have pushed the requested update!

erts/emulator/beam/erl_math.c
```diff
@@ -293,3 +295,83 @@ BIF_RETTYPE math_fmod_2(BIF_ALIST_2)
...
+	if (is_integer(BIF_ARG_1)) {
+	    BIF_RET(BIF_ARG_1);
+	}
+	BIF_ERROR(BIF_P, BADARG);
+    }
+
+    /* get the float */
+    GET_DOUBLE(BIF_ARG_1, f);
+
+    curr_direction = fegetround();
```
jhogberg, line 316:
```suggestion
#pragma STDC FENV_ACCESS ON
    curr_direction = fegetround();
```

KennethL:
could it be worth checking if the fegetround() already returns the desired value and then avoid the call to fesetround both before and after the the call to nearbyint?
Or is the fesetround(int) so cheap that it does not matter?
I was thinking that fegetround is only reading from a memory address and fesetround must be writing to memory which should be more expensive?

erts/emulator/beam/erl_math.c
josevalim: @jhogberg there is a build failure once we added the pragma. Reading this [Stack Overflow](https://stackoverflow.com/questions/60716929/if-pragma-stdc-fenv-access-is-absent-does-it-mean-default-rounding-mode) question, it seems the best option is actually to set FENV_ACCESS to off. This will guarantee this always behaves as roundeven, without a need to consult the FENV during execution., WDYT?

josevalim: There is more information [here](https://en.cppreference.com/w/c/numeric/fenv):

> The floating-point environment access and modification is only meaningful when #pragma STDC FENV_ACCESS is set to ON. Otherwise the implementation is free to assume that floating-point control modes are always the default ones and that floating-point status flags are never tested or modified. In practice, few current compilers, such as HP aCC, Oracle Studio, and IBM XL, support the #pragma explicitly, but most compilers allow meaningful access to the floating-point environment anyway. 

My understanding is that for compilers that don't obey the pragma, they can access the environment but setting the environment has no effect on the result? I will try to validate this with GCC.

josevalim: For gcc 10, even without the PRAGMA (or explicitly setting the pragma to off), changing the environment is affecting the result.

My proposal to move forward would be set the PRAGMA to off. For the compilers that support setting the pragma to off, they will emit faster code. Compilers that do not, will ignore it. This should be safe because we want the default behaviour anyway (nearest even).

jhogberg: It means "round even" on platforms that have that as the default (most if not all the ones we test on), but I suppose we can `ERTS_ASSERT(FLT_ROUNDS == 1)` on startup and cross that bridge once we find a platform that doesn't.

jhogberg: Crossposted:

> My proposal to move forward would be set the PRAGMA to off. For the compilers that support setting the pragma to off, they will emit faster code. Compilers that do not, will ignore it. This should be safe because we want the default behaviour anyway (nearest even).

Compilers that don't may warn and fail the build because of `-Werror` (at least `clang` does), so I think it's best to just assume it's off by default.

Sorry about the back and forth on this issue, I just want to try avoiding another locale-like mess if at all possible. It would've been so much easier if we could specify the rounding mode as an argument instead, but alas. :(

josevalim: Ah, I forgot about that detail, thanks.

Doing it in boot still means someone may change it from a NIF. Should we do both? Or perhaps shall we do it only on this NIF, to keep the impact limited, and update the docs to mention that it will return badarg if the native rounding mode is not supported/correct?

josevalim: > Sorry about the back and forth on this issue, I just want to try avoiding another locale-like mess if at all possible. It would've been so much easier if we could specify the rounding mode as an argument instead, but alas. :(

Yes, I completely understand. I am not too comfortable with this approach either. I will close this for now and revisit it later when we are not in a hurry due to OTP 24. Thank you! <3

jhogberg: > Doing it in boot still means someone may change it from a NIF.

Yep, though I suspect it will still work on most platforms as the SystemV and Windows ABIs say that the floating-point control bits are callee-save.

> Should we do both? Or perhaps shall we do it only on this NIF, to keep the impact limited, and update the docs to mention that it will return badarg if the native rounding mode is not supported/correct?

I'd like to find out as soon as someone tries building on such a platform, so at least on startup.

PR #4598, opened 2021-03-09
Support suppressing opportunistic compiler warnings

bjorng:
Opportunistic warnings are the warnings that the compiler's optimization and code generating passes emit when they happen to notice potential issues.

One problem with the opportunistic warnings is that there can be inappropriate warnings for perfectly good code, especially if inlining or macros are involved. Therefore, it has been our TO DO list for a long time to implement a way to turn off those warnings.

#4545, which was recently merged, can cause more spurious warnings to appear (because a `fun` that is now inlined could suppress warnings intentionally or unintentionally). Therefore, it's high time to add options to suppress opportunistic warnings.


lib/compiler/doc/src/compile.xml
```diff
@@ -772,29 +772,44 @@ module.beam: module.erl \
...
-	<p>Another class of warnings is generated by the compiler
-	  during optimization and code generation. They warn about
-	  patterns that will never match (such as <c>a=b</c>), guards
-	  that always evaluate to false, and expressions that
-	  always fail (such as <c>atom+42</c>).</p>
-	  <p>Those warnings cannot be disabled (except by
-	  disabling all warnings).</p>
+	<p>Another kind of warnings is <em>opportunistic
+	warnings</em>. They are generated when the compiler
+	<strong>happens</strong> to notice potential issues during
```
michaelklishin, line 777: Should this stress "potential" and not "happens"? Or is it not a given that in some cases e.g. unused returned values would be missed by the compiler?

bjorng: I wanted to emphasize the other kind of problem with opportunistic warnings, namely that not all potential problems are reported. I think it is better to remove the emphasis on "happens".

lib/compiler/doc/src/compile.xml
```diff
@@ -772,29 +772,44 @@ module.beam: module.erl \
...
         </taglist>
 
-	<p>Another class of warnings is generated by the compiler
-	  during optimization and code generation. They warn about
-	  patterns that will never match (such as <c>a=b</c>), guards
-	  that always evaluate to false, and expressions that
-	  always fail (such as <c>atom+42</c>).</p>
-	  <p>Those warnings cannot be disabled (except by
-	  disabling all warnings).</p>
+	<p>Another kind of warnings is <em>opportunistic
```
paulo-ferraz-oliveira, line 775:
```suggestion
	<p>Other kinds of warnings are <em>opportunistic
```

lib/compiler/doc/src/compile.xml
```diff
@@ -772,29 +772,44 @@ module.beam: module.erl \
...
+	generation.</p>
 
-	  <note>
-	    <p>The compiler does not warn for expressions that it
+	<note>
+	  <p>The compiler does not warn for expressions that it
           does not attempt to optimize. For example, the compiler tries
           to evaluate <c>1/0</c>, detects that it will cause an
 	  exception, and emits a warning. However,
 	  the compiler is silent about the similar expression,
```
paulo-ferraz-oliveira, line 785:
```suggestion
	  the compiler is silent about similar expression
```

lib/compiler/doc/src/compile.xml
```diff
@@ -772,29 +772,44 @@ module.beam: module.erl \
...
 	  <c>X/0</c>, because of the variable in it. Thus, the compiler does
 	  not even try to evaluate and therefore it emits no warnings.</p>
-          </note>
+          <p>Therefore, the absence of warnings does not mean that
```
paulo-ferraz-oliveira, line 788:
```suggestion
          <p>The absence of warnings does not mean that
```

lib/compiler/doc/src/compile.xml
```diff
@@ -772,29 +772,44 @@ module.beam: module.erl \
...
+          <item><p>Disable warnings for expressions that will always
+          fail (such as <c>atom+42</c>).</p></item>
+
+          <tag><c>nowarn_ignored</c></tag>
+          <item><p>Disable warnings for expressions whose
+          values are ignored.</p></item>
+
+          <tag><c>nowarn_nomatch</c></tag>
+          <item><p>Disable warnings for patterns that will never match
+          (such as <c>a=b</c>) and for guards that always evalute to
```
paulo-ferraz-oliveira, line 809:
```suggestion
          (such as <c>a=b</c>) and for guards that always evaluate to
```

lib/compiler/doc/src/compile.xml
bjorng: Looking closer at the entire paragraph (written by me a long time ago), I found it very clumsy and decided I needed to rewrite it:

The compiler does not warn for expressions that it does not attempt to optimize. For example, the compiler will emit a warning for `1/0` but not for `X/0`, because `1/0` is a constant expression that the compiler will attempt to evaluate.

PR #4605, opened 2021-03-09
Windows CI build

dominicletz:
As per discussion in #3027 @dgud @garazdawi this is adding a windows CI build based on Github Windows image + WSL.

This build skips building the docs intentionally but produces a Windows installer with all apps (including wxWidgets 3.1.4 and WebView2).

It's not signing any of the executables, but it's definitely good enough to see whether everything still compiles and for quick downloading and testing.

Cheers!

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -38,6 +38,78 @@ jobs:
           name: otp_git_archive
           path: otp_src.tar.gz
 
+
+  build-windows:
+    defaults:
+      run:
+        shell: wsl-bash {0}
+    name: Build Erlang/OTP on Windows
+    runs-on: windows-latest
+    needs: pack
+    steps:
+      - uses: Vampire/setup-wsl@v1
+        with:
+          distribution: Ubuntu-18.04
+
+      - name: Install WSL dependencies
+        run: apt update && apt install -y g++-mingw-w64 gcc-mingw-w64 make autoconf unzip
+
+      - name: Install openssl
+        shell: cmd
+        run: |
+          choco install openssl
+          move "c:\\Program Files\\OpenSSL-Win64" "c:\\OpenSSL-Win64"
+
+      - name: Download wxWidgets
+        run: |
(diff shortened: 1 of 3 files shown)


garazdawi:
Nice!

garazdawi:
Would it be worth adding the smoke tests as well? Just to test that the installer actually works?

From my perspective this is great, just waiting to see if @dgud has any opinions as well.

dgud:
Testing the installer requires admin rights, I don't know if that is available.
And I don't think it is necessary, we have had an offline discussion about the extra windows dll,
but that is more related to wx then this PR.

.github/workflows/main.yaml
```diff
@@ -38,6 +38,78 @@ jobs:
...
+        with:
+          name: otp_git_archive
+
+      - name: Compile Erlang
+        run: |
+          tar -xzf ./otp_src.tar.gz
+          cd otp
+          export ERL_TOP=`pwd`
+          eval `./otp_build env_win32 x64`
+          ./otp_build autoconf
```
garazdawi, line 99: The otp_src.tar.gz should already be autoconf:ed, so no need to do that.

dominicletz: Thanks! I removed that call

dominicletz:
> Testing the installer requires admin rights, I don't know if that is available.

Admin rights are available, but the installer requires UI interactions I think, e.g. clicking "next" at least. So I don't know how to do that.

But running some kind of smoke tests would be nice. Is there something already available for windows?

garazdawi:
> But running some kind of smoke tests would be nice. Is there something already available for windows?

Could you not run the same smoke test as for Linux? Though being able to run the erlang compiler is a very good smoke test just that, so maybe it is not needed...

dgud:
Use "./Install.exe /S" I think to install without the gui intervention.

garazdawi:
I will include this as it is a great improvement. We can add better functionality of not of this later on if we want to.

Thanks for the PR!

Issue #4607, opened 2021-03-09
No obvious/appropriate replacement for `http_uri:decode/1` or `http_uri:encode/1` in `uri_string`


distortedsignal:
**Is your feature request related to a problem? Please describe.**
There isn't an obvious replacement for either `http_uri:decode/1` or `http_uri:encode/1` exported from `uri_string`. If `http_uri` is going to be deprecated in OTP 25, these functions should be replicated/replaced.

**Describe the solution you'd like**
Looking through the source, it seems like there is a `decode/1` function already [in the `uri_string` module](https://github.com/erlang/otp/blob/205606a0bbae0d4484224a87d074d7eeb9e401df/lib/stdlib/src/uri_string.erl#L1478-L1480). It might be enough to export that function so that users can use it. The currently exported function (`uri_string:percent_decode/1`) seems like it could be a replacement, but I'm curious why it was decided that `Mod1:decode/1` should become `Mod2:percent_decode/1`.

It looks like `encode` might be a little harder. There is an `encode/2` function [in the `uri_string` module](https://github.com/erlang/otp/blob/205606a0bbae0d4484224a87d074d7eeb9e401df/lib/stdlib/src/uri_string.erl#L1575-L1580), but the user would have to pass in a `fun` of unspecified arguments - which seems distinctly un-fun to me. It seems to me (in my layman's lack-of-encoding-understanding) that this `fun` should be fully defined by the IETF RFC for URLs, and should be relatively static in terms of how the encoding should work. It _seems_ like this should be a fine thing to ship in the URI library. But again, I'm a layman in these things.

**Describe alternatives you've considered**
I could continue to use the `http_uri` module, but that seems like something that won't continue to work past OTP 25, as my compiler keeps telling me.

I could publish my own library with these functions, but that seems like it would be counterintuitive - taking a function from something shipped to the standard library to a third-party dependency. That strikes me as regressive.

We could get this from a third party language, but that seems _really bad_.

**Additional context**
I'll try to take a crack at this when I get the chance. I'm guessing the longest part is going to be the `fun` that turns characters into their percent-encoded counterparts.

lhoguin:
There is no encode function because the module already automatically encodes when you give it appropriate components.

Some background is here: https://github.com/erlang/otp/pull/2492 - the `percent_decode` function was added because the `uri_string` parser does not decode reserved characters in components. The decoding has to be done independently by the application afterwards, if desired. The same is not true for encoding.

What do you use the `encode` function for?

[...]

PR #4608, opened 2021-03-10
Add percent_encode function

distortedsignal:
I was working with some URLs earlier this week, and I noticed that the `uri_string` module didn't have a simple `(list()|binary()) -> list()|binary()` function to do percent encoding. I thought this was a nice feature of `http_uri` and I would like to carry it forward after OTP 25 (which is probably coming soon?). I filed #4607 earlier today to try to make the issue clear, then I decided to make this change myself. Along the way, I fixed some typos in the docs (note line 331 in `uri_string.xml`) and I refactored some code to be a little more DRY (lines 1222-1234 in `uri_string.erl`). I added the spec on line 1578 because my text editor wanted me to. There was also a spacing fix in the exports of the test code as well.

Trying to run the tests, I keep getting this error:
```
=WARNING REPORT==== 9-Mar-2021::22:41:37.340494 ===
Process <0.2604.0> on node 'test_server@Dorian' enabled access to the emulator internal state.
NOTE: This is an erts internal test feature and should *only* be used by OTP test-suites.


=ERROR REPORT==== 9-Mar-2021::22:41:37.340654 ===
--- fds in flbk pollset ---------------------------------
--- fds in scheduler pollset ----------------------------
--- fds in pollset --------------------------------------
pollset=0 fd=16 type=sock driver_select ev=IN  inport=#Port<0.0> inname=forker indrv=spawn_forker  flags=USED
pollset=0 fd=17 type=sock driver_select ev=IN  inport=#Port<0.3> inname=tcp_inet indrv=tcp_inet  flags=USED
pollset=0 fd=18 type=sock driver_select ev=IN  inport=#Port<0.5> inname=tcp_inet indrv=tcp_inet  flags=USED
pollset=0 fd=24 type=fifo driver_select ev=IN  inport=#Port<0.10> inname=inet_gethost 4  indrv=spawn  flags=USED

used fds=4
Number of driver_select() structures=4
Number of enif_select() structures=0
internal fds=0
---------------------------------------------------------
```

Seems bad. I'm trying to do this inside WSL 2, so that's an added wrinkle of complication. Anybody seen this before?


distortedsignal:
Things passed! Neat!

luos:
Hello,

Sorry for bumping this closed PR but is there a reason why this was not merged? Is there an alternative?

Thanks

IngelaAndin:
I believe it is a little more complex than many people assume. I recommend reading the "Uniform Resource Identifiers" chapter in the stdlibs Users Guide.

luos:
Hi Ingela, 

Thanks for the quick answer. I checked out the section you mention and I've found this:

```
uri_string:recompose(#{scheme => "https", host => "local#host", path => ""}).
```

I am thinking then it's a missing API, because `path` takes a string instead of a list, therefore I can not specify a path which needs percent encoding containing `/` characters: 

`path => [ "This is a path part", "This is the second part", "örebro", "This#is\the/last/part] `

Thanks

IngelaAndin:
No path is not a string it is a uri_string() = iodata()

```
> Erlang/OTP 23 [erts-11.1.8] [source-3c4d99c4ce] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [hipe]
> 
> Eshell V11.1.8  (abort with ^G)
> 1>  Path = [ "This is a path part", "This is the second part", "örebro", "This#is\the/last/part"].
> ["This is a path part","This is the second part","örebro",
>  "This#is\the/last/part"]
> 2>  uri_string:recompose(#{scheme => "https", host => "local#host", path => Path}).   
> "https://local%23host/This%20is%20a%20path%20partThis%20is%20the%20second%20part%C3%B6rebroThis%23is%09he/last/part"
> 3> 
> 
```

luos:
In my opinion the result of that is incorrect or not what I would expect, it should have been: 

This%20is%20a%20path%20part **/** This%20is%20the%20second%20part **/** %C3%B6rebro **/** This%23is%09he/last/part"

That's why I am saying this functionality is missing, which is only solvable now with `http_uri:encode`

lhoguin:
If you have a list of path segments, then yes, I believe there is currently no function to do this. However getting from path segments to path does not need a general urlencoding function, but rather to just replace `/` with `%2f` in the segments and then put slashes before each segment.

But you'll note that the opposite operation is also not available in the module.

peterdmv:
> ```
> PathSegments = [ "first", "sec/ond"].
> 3>  uri_string:recompose(#{ host => "localhost", path => PathSegments}).
> "//localhost/firstsec/ond"
> ```
> 
> I expect the result should be instead:
> 
> ```
> localhost/first/sec%2fond
> ```
> 
> That's why I am saying this functionality is missing, which is only solvable now with `http_uri:encode` and applying it to the separate path segments each.

The standard defines percent-encoding in the following way: 

```
2.1.  Percent-Encoding

   A percent-encoding mechanism is used to represent a data octet in a
   component when that octet's corresponding character is outside the
   allowed set or is being used as a delimiter of, or within, the
   component. 
```

As the character "/" is allowed in the path component it shall not be percent-encoded by a standard-compliant URI producer. Applying custom percent-encoding at the component level is application-specific and shall be handled by the application, not in the standard library. There is further clarification why such a function has not been implemented in the Users Guide:
```
Percent-encoding is implemented in uri_string:recompose/1 and it happens when converting a uri_map() into a uri_string(). 
There is no equivalent to a raw percent-encoding function as percent-encoding shall be applied on the component
level using different sets of allowed characters. Applying percent-encoding directly on an input URI would not be safe
just as in the case of uri_string:percent_decode/1, the output could be an invalid URI.
```

PR #4610, opened 2021-03-10
inet_res: rewrite to parallel lookup algorithm

RaimoNiskanen:
This is a suggestion that needs to be evaluated.  The concept popped up in the discussion of #4564 and I happened to have this branch already.

I have rewritten the `inet_res` DNS resolver to query name servers in parallel, which seems to be the contemporary way to do it.

To avoid wasting resources there is a delay of half the normal UDP time-out before a new name server is queried, so the queries are not entirely parallel.  The question is if this is too defensive?

To ensure process cleanup there is a Reaper process started for every query, and all queries are performed in parallel processes.  That means at least 2 processes are spawned for a query, and then after half the normal UDP time-out another is spawned up to 1 process per name server + the Reaper.  The question is if this expending too much resources - I did not find a better way to do it.

Because queries are done in parallel this algorithm uses more sockets i.e more IP Ports and more file descriptors.  The current algorithm uses one UDP socket, and possibly one TCP socket if needed.  This algorithm uses one socket per query process which go up to one per name server.  The question is if this might be a problem for systems where IP Ports or file descriptors are a scarce resource.

So, is this a good improvement, or wasting resources?

### DIFF
--- a/lib/kernel/src/inet_res.erl
+++ b/lib/kernel/src/inet_res.erl
@@ -651,398 +651,501 @@ make_query(Dname, Class, Type, Options, Edns) ->
     {Id, Buffer}.
 
 %% --------------------------------------------------------------------------
-%% socket helpers
+%% request algorithm (parallel)
 %%
--record(sock, {inet=undefined, inet6=undefined}).
-
-udp_open(#sock{inet6=I}=S, {A,B,C,D,E,F,G,H}) when ?ip6(A,B,C,D,E,F,G,H) ->
-    case I of
-	undefined ->
-	    case gen_udp:open(0, [{active,false},binary,inet6]) of
-		{ok,J} ->
-		    {ok,S#sock{inet6=J}};
-		Error ->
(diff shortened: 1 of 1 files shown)


lhoguin:
I don't know about environments with scarce resources, but for other environments this is a much welcome change. This is a key component to have proper "happy eyeballs" support in various clients.

RoadRunnr:
What that this does is sending the *same* query to multiple NS in parallel. While that is a welcome addition, what I'm missing more is the ability to send A and AAAA queries in parallel to the same NS. For the native resolvers it is standard to do that when the desire address family is set to PF_UNSPEC. In Erlang every user of inet_res (or inet for that matter) that wants dual stack support would need to implement such a thing themself.

RaimoNiskanen:
@RoadRunnr: I see!
Such a feature could actually use the same socket for both queries, so it is a different kind of parallelism.  Feature parallel vs. server parallel.

PR #4613, opened 2021-03-10
Quote atom in types

dgud:
Some where qouted and some not, qoute them all.


lhoguin:
Qoute -> Quote, is -> in

dgud:
I did it correct in the branch name at least 1/4 :-)

PR #4615, opened 2021-03-10
speed up pg:all_local_pids by using gen_server state

zzydxm:
pg:all_local_pids used to use ets:select to scan the entire ets table.
It will be very slow when there is a lot of remote groups.
Changed it to use gen_server state to only scan local groups.


max-au (approved):
I verified this change, PropEr suite from https://github.com/max-au/spg/blob/master/test/spg_cluster_SUITE.erl succeeds.

PR #4619, opened 2021-03-11
pg: fix empty group removal from internal pg state

max-au:
pg keeps two copies of state: gen_server state and ETS table for
concurrent access. #2866 fixed one potential way to introduce
mismatch between internal state and ETS table, and this patch
completes the fix.

### DIFF
--- a/lib/kernel/src/pg.erl
+++ b/lib/kernel/src/pg.erl
@@ -281,19 +281,7 @@ handle_info({leave, Peer, PidOrPids, Groups}, #state{scope = Scope, nodes = Node
     case maps:get(Peer, Nodes, []) of
         {MRef, RemoteMap} ->
             _ = leave_remote(Scope, PidOrPids, Groups),
-            NewRemoteMap = lists:foldl(
-                fun (Group, Acc) ->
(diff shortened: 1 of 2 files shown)


rickard-green:
We've gotten multiple failures like the one below in the `empty_group_by_remote_leave` test case in our daily testing:
```
=== === Reason: {assertEqual,
                     [{module,pg_SUITE},
                      {line,305},
                      {expression,
                          "pg : get_members ( ? FUNCTION_NAME , ? FUNCTION_NAME )"},
                      {expected,[]},
                      {value,[<47808.93.0>,<47808.96.0>]}]}
  in function  pg_SUITE:empty_group_by_remote_leave/1 (pg_SUITE.erl, line 305)
  in call from test_server:ts_tc/3 (test_server.erl, line 1783)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1292)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 1224)
```

max-au:
It appears to run into rebase conflict. I rebased my commits to master, and added missing sync, now tests are passing consistently.

PR #4629, opened 2021-03-16
erl_call: Add -fetch_stdout and -no_result_term options

kjellwinblad:
This commit adds two options to the erl_call command line tool:

* -fetch_stdout
  Executes the given code (provided by the -a or -e options) in a new
  process with a group leader that forwards all stdout (standard
  output) data so that it is printed to stdout of the erl_call
  process. This means that stdout data that is written during the
  execution of the called code, by the code and by descendant
  processes, will be forwarded (given that the group leader has not
  been changed by a call to erlang:group_leader/2). The printed data
  is UTF-8 encoded.
* -no_result_term
  This option disables printing of the result term.

In order to implement the first of these two options a new function
called ei_xrpc_from has been added to erl_interface. This function is
documented in the erl_interface documentation.

### DIFF
--- a/lib/erl_interface/doc/src/ei_connect.xml
+++ b/lib/erl_interface/doc/src/ei_connect.xml
@@ -1047,6 +1047,7 @@ if (ei_reg_send(&ec, fd, x.buff, x.index) < 0)
     <func>
       <name since=""><ret>int</ret><nametext>ei_rpc(ei_cnode *ec, int fd, char *mod, char *fun, const char *argbuf, int argbuflen, ei_x_buff *x)</nametext></name>
       <name since=""><ret>int</ret><nametext>ei_rpc_to(ei_cnode *ec, int fd, char *mod, char *fun, const char *argbuf, int argbuflen)</nametext></name>
+      <name since="OTP-17048"><ret>int</ret><nametext>ei_xrpc_to(ei_cnode *ec, int fd, char *mod, char *fun, const char *argbuf, int argbuflen, int flags)</nametext></name>
       <name since=""><ret>int</ret><nametext>ei_rpc_from(ei_cnode *ec, int fd, int timeout, erlang_msg *msg, ei_x_buff *x)</nametext></name>
       <fsummary>Remote Procedure Call from C to Erlang.</fsummary>
       <type>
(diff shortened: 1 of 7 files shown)


lib/erl_interface/doc/src/ei_connect.xml
```diff
@@ -1047,6 +1047,7 @@ if (ei_reg_send(&ec, fd, x.buff, x.index) < 0)
     <func>
       <name since=""><ret>int</ret><nametext>ei_rpc(ei_cnode *ec, int fd, char *mod, char *fun, const char *argbuf, int argbuflen, ei_x_buff *x)</nametext></name>
       <name since=""><ret>int</ret><nametext>ei_rpc_to(ei_cnode *ec, int fd, char *mod, char *fun, const char *argbuf, int argbuflen)</nametext></name>
+      <name since=""><ret>int</ret><nametext>ei_xrpc_to(ei_cnode *ec, int fd, char *mod, char *fun, const char *argbuf, int argbuflen, int flags)</nametext></name>
```
garazdawi, line 1050: This since should contain the correct ticket number.

lib/erl_interface/src/prog/erl_call.c
```diff
@@ -989,3 +1023,87 @@ static char* ei_chk_strdup(char *s)
...
+                                 int inbuflen, ei_x_buff* x)
+{
+    int i, index;
+    int got_rex_response = 0;
+    int initial_buff_index = x->index;
+    ei_term t;
+    erlang_msg msg;
+    char rex[MAXATOMLEN];
+ 
+    if (ei_xrpc_to(ec, fd, mod, fun, inbuf, inbuflen, EI_RPC_FETCH_STDOUT) < 0) {
```
garazdawi, line 1051: The return possible values of ei_rpc_to seem to be undocumented. Can you add a note about that in the docs?

lib/kernel/src/rpc.erl
```diff
@@ -1265,3 +1289,114 @@ pinfo(Pid, Item) when node(Pid) =:= node() ->
...
+-spec cnode_call_group_leader_init(CallerPid :: pid()) -> ok | no_return().
+
+cnode_call_group_leader_init(CallerPid) ->
+    State = #cnode_call_group_leader_state{caller_pid = CallerPid},
+    cnode_call_group_leader_loop(State).
+
+-spec cnode_call_group_leader_start(CallerPid :: pid()) -> pid().
+
+cnode_call_group_leader_start(CallerPid) ->
+    spawn_link(fun() -> cnode_call_group_leader_init(CallerPid) end).
```
garazdawi, line 1402: What is the reason for linking the group_leader with the rex process?

lib/erl_interface/doc/src/erl_call_cmd.xml
```diff
@@ -116,6 +116,34 @@
               expressions and returns the result from the last expression.
               Returns <c>{ok,Result}</c> on success.</p>
           </item>
+          <tag><c>-fetch_stdout</c></tag>
+          <item>
+            <p>
+              (<em>Optional.</em>) Executes the code, specified with
```
garazdawi, line 122: Should be a note that this option does not work with Erlang nodes before OTP-24.

garazdawi (commented):
Nice! I just had some minor comments.

lib/erl_interface/doc/src/ei_connect.xml
kjellwinblad: Thanks fixed.

lib/erl_interface/src/prog/erl_call.c
kjellwinblad: Good catch. I have added this to the documentation of ei_rpc_to and ei_xrpc_to

lib/kernel/src/rpc.erl
kjellwinblad: You are right. After looking at the code again, I see no good reason for doing spawn_link instead of spawn so I have changed the code accordingly.

lib/erl_interface/doc/src/erl_call_cmd.xml
kjellwinblad: You are right! Fixed!

lib/erl_interface/doc/src/ei_connect.xml
```diff
@@ -1055,13 +1056,27 @@ if (ei_reg_send(&ec, fd, x.buff, x.index) < 0)
...
+          The <c>ei_xrpc_to()</c> function is equivalent to
+          <c>ei_rpc_to()</c> when its <c>flags</c> parameter is set to
+          <c>0</c>. When the flags parameter of <c>ei_xrpc_to()</c> is
+          set to <c>EI_RPC_FETCH_STDOUT</c>, stdout (standard output)
+          data are forwarded. See the documentation for the flags
+          parameter for more information about the
+          <c>EI_RPC_FETCH_STDOUT</c> flag.
+        </p>
+        <p>
+        <seemfa marker="kernel:rpc#call/4"> <c>rpc:call/4</c></seemfa>
```
garazdawi, line 1077:
```suggestion
        <seemfa marker="kernel:rpc#call/4"><c>rpc:call/4</c></seemfa>
```

Not sure if this space is significant, but may as well remove it.

garazdawi (approved):

lib/erl_interface/doc/src/ei_connect.xml
kjellwinblad: Fixed!

PR #4632, opened 2021-03-16
Inform on how an end_per_testcase crash affects test results

### LINKED ISSUE #4617: ct: failing end_per_testcase not failing test
Using Common Test, a test gets an OK result even if its corresponding `end_per_testcase` fails. Is this by design? It may be, since the report shows the error as "WARNING: end_per_testcase thrown! WARNING: end_per_testcase failed!", but... the result is still OK (with time "Died"), which is not useful if, say, you're running tests in CI.

Should https://erlang.org/doc/man/ct_suite.html#Module:end_per_testcase-2 make this more explicit? This is what it states (among other things), at the moment: "By returning {fail, Reason}, TestCase is marked as faulty (even though it was successful in the sense that it returned a value instead of terminating).".

paulo-ferraz-oliveira:
Closes #4617.

cc: @IngelaAndin

### DIFF
--- a/lib/common_test/doc/src/write_test_chapter.xml
+++ b/lib/common_test/doc/src/write_test_chapter.xml
@@ -11,7 +11,7 @@
       Licensed under the Apache License, Version 2.0 (the "License");
       you may not use this file except in compliance with the License.
       You may obtain a copy of the License at
- 
+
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
Oh, cool. Thanks.

Issue #4624, opened 2021-03-14
Many error checks in erlexec.c are ineffective since OTP 22.1


JeromeDeBretagne:
**Describe the bug**
Looking at the source code from [erlexec.c](https://github.com/erlang/otp/blob/master/erts/etc/common/erlexec.c), some user input mistakes should be detected and should raise an error message, but currently they are not working as intended.

For instance here is the error check for the "-name" option (but there are many other similar ones) :
https://github.com/erlang/otp/blob/master/erts/etc/common/erlexec.c#L733-L735
```
		    if (strcmp(argv[i], "-name") == 0) { /* -name NAME */
			if (i+1 >= argc) <====================== This check is broken
			    usage("-name");
```

**To Reproduce**
$ erl -name
Erlang/OTP [...]

Eshell Vx.y.z  (abort with ^G)
1> 

**Expected behavior**
$ erl -name
Missing argument(s) for '-name'.
Usage: erl ...

**Affected versions**
@garazdawi my guess is that a regression was introduced by this commit https://github.com/erlang/otp/commit/889cc75f40c74ee574b0943524df4f786371efaf from June 2019 as it introduced an extra "--" separator thus incrementing argc, so I would say since OTP 22.1

**Additional context**
I've indeed found that there is an extra "--" argument by adding some debug code.

EDIT: The issue actually happens when an ERL_*FLAGS environment variable is set, cf. https://github.com/erlang/otp/issues/4624#issuecomment-801135677

garazdawi:
This was partly fixed in OTP-23 through 9c113f4deeb7dc1ee02cd58003782e1d70edad56. However, you can still make it happen if you have an ERL_*FLAGS environment variable set. I've opened up #4639 to solve this, just need to think of a way to test this....

JeromeDeBretagne:
> However, you can still make it happen if you have an ERL_*FLAGS environment variable set

Good catch, I have indeed ERL_AFLAGS defined in my setup to enable the Erlang shell history.

> I've opened up #4639 to solve this, just need to think of a way to test this....

I have applied it locally, I can confirm that #4639 is properly fixing the issue, thanks.

Will you apply the fix only on master or do you also plan to backport it to other releases later on? I'm asking to know if I'll have to target master for #2945.

JeromeDeBretagne:
About `erl` argument checks, I've also detected that the following command:

```
erl -name node_name an_unexpected_argument

1> node().
nonode@nohost
```

[...]

PR #4639, opened 2021-03-17
erlexec: Add checks that next argument is valid

garazdawi:
The argument -- is silently added in a number of places in order to seperate arguments from different sources. When checking to see if an argument that erlexec needs to handle is valid we therefore check that it does not start with '-' or '+' which are invalid arguments.

Solves #4624


make/test_target_script.sh
```diff
@@ -231,7 +231,6 @@ $CT_RUN -logdir $MAKE_TEST_CT_LOGS\
         -pa "$ERL_TOP/lib/common_test/test_server"\
         ${ARGS}\
         -erl_args\
-        -env "$PATH"\
```
garazdawi, line 234: @kjellwinblad would you mind checking if this change is correct?

PR #4645, opened 2021-03-18
Makefile.in: Skip lib/jinterface/priv if missing

wojtekmach:
Before this patch these errors were logged:

    $ ./otp_build setup -a --without-jinterface
    (...)
    MAKE tertiary_bootstrap_copy
    cp: lib/jinterface/priv/OtpErlang.jar: No such file or directory
    cp: lib/jinterface/priv/com/ericsson/otp/erlang/*: No such file or directory
    (...)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -832,26 +832,28 @@ tertiary_bootstrap_copy:
 		true; \
 	done
 #	copy jinterface priv directory
-	$(V_at)for x in lib/jinterface/priv/OtpErlang.jar; do \
-		BN=`basename $$x`; \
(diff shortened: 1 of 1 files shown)


garazdawi:
We may be able to remove jinterface from the bootstrap altogether. I think that it was there in order to compile [corba and friends](http://github.com/erlang/corba) when they were part of the main repository.

wojtekmach:
Even better, I'll look into it and change the patch accordingly.

wojtekmach:
Closing in favour of #4646.

rickard-green:
We are currently finalizing RC-2, so we won't test this until after RC-2. It looks ok, so it will most likely be included after RC-2.

PR #4646, opened 2021-03-18
Remove jinterface from bootstrap

wojtekmach:


rickard-green:
Unfortunately this breaks the build of [corba](https://github.com/erlang/corba) since it parasitizes on OTP's build system. You may want to reopen your other PR.

PR #4647, opened 2021-03-18
Fixes to problems reported by codechecker

kjellwinblad:
This pull-request fixes a number of problems reported by [codechecker](https://github.com/Ericsson/codechecker).

### DIFF
--- a/erts/etc/common/erlc.c
+++ b/erts/etc/common/erlc.c
@@ -186,6 +186,9 @@ set_env(char *key, char *value)
     efree(str);
 #endif
 #endif
+    /* codechecker_intentional [Malloc] we may leak str if we don't
+       have copying putenv but that is fine since we only have a
+       constant amount of environment variables */
 }
 
 static void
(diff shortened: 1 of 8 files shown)


erts/etc/common/erlc.c
```diff
@@ -908,7 +913,13 @@ safe_realpath(char* file)
      * Solaris.
      */
     char* real_name = emalloc(PATH_MAX + 1);
-    return realpath(file, real_name);
+    char* result = realpath(file, real_name);
+    if (result == NULL) {
+        free(real_name);
+    }
+    /* codechecker_false_positive [Malloc] we are not leaking
+       real_name because result == real_name if result != NULL */
```
sverker, line 921: To avoid codechecker specific suppression
could this be fixed by changing to
if (result != real_name)

bjorng (approved):

erts/etc/common/erlc.c
kjellwinblad: Thanks, I will give it a try.

kjellwinblad: I have updated the code (we will see if it works after our code checker build has executed).

PR #4651, opened 2021-03-19
Add fullsweep_after support in process_flag/2

lhoguin:
See https://github.com/erlang/otp/pull/4651#issuecomment-806756197

If setting to 0, for example, then the next GC will be fullsweep and the old_heap effectively disabled. An `erlang:garbage_collect()` can force this if necessary.

Would be great to have this in OTP-24.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -5596,7 +5596,18 @@ RealSystem = system + MissedSystem</code>
     </func>
 
     <func>
-      <name name="process_flag" arity="2" clause_i="3"
+      <name name="process_flag" arity="2" clause_i="3" since="OTP @OTP-17285@"/>
+      <fsummary>Set process flag fullsweep_after for the calling process.
+      </fsummary>
+      <desc>
+        <p>Changes the maximum number of generational collections
+          before forcing a fullsweep for the calling process.</p>
+        <p>Returns the old value of the flag.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="process_flag" arity="2" clause_i="4"
 	    anchor="process_flag_min_heap_size" since=""/>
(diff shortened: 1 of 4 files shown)


erts/doc/src/erlang.xml
```diff
@@ -5596,7 +5596,18 @@ RealSystem = system + MissedSystem</code>
     </func>
 
     <func>
-      <name name="process_flag" arity="2" clause_i="3"
+      <name name="process_flag" arity="2" clause_i="3" since="OTP @TODO"/>
```
lhoguin, line 5599: What should I put there?

sverker:
The doc build fails as the fullsweep_after clause is missing in the spec for process_flag/2 in erlang.erl. It should be #3 before min_heap_size.

erts/doc/src/erlang.xml
sverker: You can put `since="OTP @OTP-17285@"`

lhoguin:
I have used it to automatically change the `fullsweep_after` value and it ran fine for a few hours.

I am still working on making it useful. That will take a while.

lhoguin:
It has been invaluable when experimenting with `fullsweep_after` values on a running system. Set the flag (via `sys:replace_state`), wait a few minutes, check memory usage, repeat until satisfied. Then update configuration/code with the value (or don't).

Setting `fullsweep_after` to 0 has also been helpful in identifying when the garbage was generated. With the value set to 0 the memory was peaking on a regular basis, every 100 seconds or so. This corresponded to files being read fully in memory, processed and generating a lot of garbage, which then takes time to cleanup. After a few fixes the memory is lower, stable, with or without `fullsweep_after` configured.

Overall a useful tool. Hope it makes it into OTP-24.

sverker:
Merged to master. Thanks, @lhoguin.

PR #4654, opened 2021-03-21
Add missing documentation for sending to aliases

aronisstav:
The documentation for send is missing explanation of aliases in at least the two fixed places.
The Erlang Reference Manual was also missing mention of ports.

Aliases should arguably also be added to the processes section of the reference manual, but this is not included in this PR.


rickard-green:
Thanks, I'll write something in the reference manual about alias.

PR #4656, opened 2021-03-22
Do not allocate large map when value is the same

josevalim:
This is the same as 6007bf0f8 but applied to large maps.


sverker:
I pushed some refactor commits and put into test for 24.

sverker:
Merged to master. Thanks, @josevalim!

PR #4660, opened 2021-03-23
erts: Fix bug in call time tracing

### LINKED ISSUE #4635: eprof causes segmentation fault upon garbage_collect on OTP 22
I found a segfault on OTP 22 when running `eprof`.

* OTP: 22.3.4.16 (I haven't tested later versions yet)
* OS: Ubuntu 20.04.2 LTS
* Arch: x86_64

Take the below sample program:

```erlang
-module(segfault).
-export([ start/0
        , worker/0
        ]).

start() ->
  Pid = spawn(?MODULE, worker, []),
  eprof:start(),
  eprof:start_profiling([Pid]),
  timer:sleep(1000),
  eprof:stop_profiling(),
  eprof:stop().

worker() ->
  receive
    quit -> ok
  after 100 ->
    erlang:garbage_collect(),
    ?MODULE:?FUNCTION_NAME()
  end.
```

Running the program causes a segmentation fault:

```shell
$ erlc segfault.erl && erl -s segfault
Erlang/OTP 22 [erts-10.7.2.8] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]

Eshell V10.7.2.8  (abort with ^G)
1> Segmentation fault (core dumped)
```

This is what I see in `gdb`:

```
Thread 8 "4_scheduler" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ff403376700 (LWP 360149)]
0x000055a254d687fb in fixup_cp_before_trace (return_to_trace=<synthetic pointer>, c_p=0x7ff405a02c98) at beam/beam_bp.c:670
670	            BeamInstr w = *cp_val(*cpp);
(gdb) print *c_p
$1 = {common = {id = 2817498547491, refc = {atmc = {counter = 1}, sint = 1}, tracer = 140686055703121, trace_flags = 1088, timer = {counter = 0}, u = {alive = {
        started_interval = 0, reg = 0x0, links = 0x0, lt_monitors = 0x0, monitors = 0x0}, release = {later = 0, func = 0x0, data = 0x0, next = 0x0}}}, htop = 0x7ff401815840, 
  stop = 0x7ff401815f70, heap = 0x7ff401815840, hend = 0x7ff401815f88, abandoned_heap = 0x0, heap_sz = 233, min_heap_size = 233, min_vheap_size = 46422, max_heap_size = 3, 
  arity = 0, arg_reg = 0x7ff405a02d48, max_arg_reg = 6, def_arg_reg = {140685976288104, 208907, 18446744073709551611, 0, 0, 4000}, cp = 0x55a255054708 <beam_return_time_trace>, 
  i = 0x7ff4040ff8d8, catches = 0, fcalls = 3997, rcount = 0, schedule_count = 0, reds = 5, flags = 0, group_leader = 2233382994963, fvalue = 18446744073709551611, freason = 0, 
  ftrace = 18446744073709551611, next = 0x0, sig_qs = {first = 0x0, last = 0x7ff405a02dd8, save = 0x7ff405a02dd8, cont = 0x0, cont_last = 0x7ff405a02df0, nmsigs = {next = 0x0, 
      last = 0x0}, saved_last = 0x0, len = 0, flags = 2}, bif_timers = 0x0, dictionary = 0x0, seq_trace_clock = 0, seq_trace_lastcnt = 0, seq_trace_token = 18446744073709551611, 
  u = {real_proc = 0x89e0b, terminate = 0x89e0b, initial = {module = 564747, function = 208907, arity = 0}}, current = 0x7ff4040ff8c0, parent = 2714419332339, static_flags = 0, 
  high_water = 0x7ff401815840, old_hend = 0x0, old_htop = 0x0, old_heap = 0x0, gen_gcs = 0, max_gen_gcs = 65535, off_heap = {first = 0x0, overhead = 0}, mbuf = 0x0, 
  live_hf_end = 0xfffffffffffffff8, msg_frag = 0x0, mbuf_sz = 0, psd = {counter = 140685981587736}, bin_vheap_sz = 46422, bin_old_vheap_sz = 46422, bin_old_vheap = 0, 
  sys_task_qs = 0x0, dirty_sys_tasks = 0x0, state = {counter = 41002}, dirty_state = {counter = 0}, sig_inq = {first = 0x0, last = 0x7ff405a02f10, len = 0, nmsigs = {next = 0x0, 
      last = 0x0}}, trace_msg_q = 0x0, lock = {flags = {counter = 1}, queue = {0x0, 0x0, 0x0, 0x0, 0x0}}, scheduler_data = 0x7ff404463940, run_queue = {
    counter = 140686020325504}}
(gdb) print *c_p->cp
$2 = 94155696147326
(gdb) print c_p->cp
$3 = (BeamInstr *) 0x55a255054708 <beam_return_time_trace>
(gdb) print cpp
$4 = (Eterm *) 0x7ff401815f70
(gdb) print &E[0]
$5 = (Eterm *) 0x7ff401815f70
(gdb) print (BeamInstr*)(E[0])
$6 = (BeamInstr *) 0xfffffffffffffffb
(gdb) print (BeamInstr*)(E[1])
$7 = (BeamInstr *) 0x7ff401a43b80
(gdb) print (BeamInstr*)(E[2])
$8 = (BeamInstr *) 0x55a255054748 <beam_apply+8>
```

So my understanding is that the crash happens [here](https://github.com/erlang/otp/blob/OTP-22.3.4.16/erts/emulator/beam/beam_bp.c#L670).

* The function should have entered the `BeamIsOpCode(w, op_i_return_time_trace)` branch (L663).
* The crash happens in the first iteration within the `for (;;)` loop.
* I don't  fully understand what the code intends to do, but  after checking the `cp` is one of the ops it's looking for it seems to expect that the same op is on the top of the stack. But when I print the stack with `gdb` the top of the stack is `-5` instead. The first op on the stack comes two words later, and it's not the same as the `cp`.
* The `erlang:garbage_collect()` call seems to be important, if I remove it, the segfault goes away.

sverker:
Fix #4635

Looks like this bug was "always" there but was exposed by
1f5fb6de3e20cf433606c8d8c33a9813d6337aa2
when E[0] could also be NIL.

The bug only exists in 22.2 <= OTP < 23.0.

### DIFF
--- a/erts/emulator/test/trace_call_time_SUITE.erl
+++ b/erts/emulator/test/trace_call_time_SUITE.erl
@@ -65,6 +65,7 @@
 -export([all/0, suite/0,
 	 init_per_testcase/2, end_per_testcase/2, not_run/1]).
 -export([basic/1, on_and_off/1, info/1,
+         apply_bif_bug/1, abb_worker/1,
          disable_ongoing/1,
 	 pause_and_restart/1, scheduling/1, called_function/1, combo/1, 
 	 bif/1, nif/1]).
@@ -91,6 +92,7 @@ all() ->
 	false ->
 	    [basic, on_and_off, info, pause_and_restart, scheduling,
              disable_ongoing,
+             apply_bif_bug,
 	     combo, bif, nif, called_function, dead_tracer, return_stop]
     end.
 
@@ -778,3 +780,26 @@ loop() ->
             Pid ! {self(), answer, erlang:apply(M, F, A)},
             loop()
     end.
+
+%% OTP-17290, GH-4635
+apply_bif_bug(_Config) ->
+    Pid = spawn(?MODULE, abb_worker, [self()]),
+    erlang:trace(Pid, true, [call]),
+    erlang:trace_pattern({?MODULE,abb_foo,'_'}, true, [call_time]),
+    erlang:trace_pattern({erlang,display,1}, true, [call_time]),
+    Pid ! {call, erlang, display, ["Hej"]},
+    receive
+        done -> ok
+    end,
+    erlang:trace_pattern({'_','_','_'}, false, [call_time]).
+
+abb_worker(Papa) ->
+    receive
+        {call, M, F, Args} ->
+            abb_foo(M, F, Args),
+            Papa ! done
+    end.
+
+
+abb_foo(M,F,Args) ->
+    apply(M,F,Args).
--- a/erts/emulator/beam/beam_bp.c
+++ b/erts/emulator/beam/beam_bp.c
@@ -661,7 +661,7 @@ static void fixup_cp_before_trace(Process *c_p, int *return_to_trace)
         *return_to_trace = 1;
         cpp = &E[0];
     } else if (BeamIsOpCode(w, op_i_return_time_trace)) {
-        cpp = &E[0];
+        cpp = &E[1];
     } else {
         cpp = NULL;
     }


erts/emulator/test/trace_call_time_SUITE.erl
```diff
@@ -65,6 +65,7 @@
 -export([all/0, suite/0,
 	 init_per_testcase/2, end_per_testcase/2, not_run/1]).
 -export([basic/1, on_and_off/1, info/1,
+         apply_bif_bug/1, abb_worker/1,
```
dszoboszlay, line 68: I think `apply_bif_bug` should be added to `all/0` too.

sverker: Yes! Thank you.

PR #4663, opened 2021-03-23
Allow reconfiguration of logger

hauleth:
This adds function `logger:reconfigure/0` that allows to reset logger configuration to what is configured in `kernel` application configuration. This function is meant to be used by build/run tools like Rebar3 or Mix to provide "production-like" way to configure logging in reliable way.

This function is obviously not safe to be called during application runtime, as it may cause losing some log messages, but it is by far better solution than what we have now, where [Rebar3 tries to reimplement such functionality on their own](https://github.com/erlang/rebar3/blob/8adc139818191542ca9e110f13f78adcd4d83daf/src/rebar_utils.erl#L489) (risking incompatibilities if `logger` configuration will be expanded one day) or just ignore it, and requires users to configure logging on their own (like Mix does, that just relies that users will use Elixir's Logger). Lack of such functionality prevents Elixir from further integration with Erlang's Logger.

Ref elixir-lang/elixir#10791 elixir-lang/elixir#9465

### DIFF
--- a/lib/stdlib/src/maps.erl
+++ b/lib/stdlib/src/maps.erl
@@ -119,18 +119,18 @@ intersect_with(Combiner, Map1, Map2) ->
 
 intersect_with_small_map_first(Combiner, SmallMap, BigMap) ->
     Next = maps:next(maps:iterator(SmallMap)),
-    intersect_with_iterate(Next, [], SmallMap, BigMap, Combiner).
+    intersect_with_iterate(Next, [], BigMap, Combiner).
 
-intersect_with_iterate({K, V1, Iterator}, Keep, Map1, Map2, Combiner) ->
+intersect_with_iterate({K, V1, Iterator}, Keep, BigMap, Combiner) ->
     Next = maps:next(Iterator),
-    case Map2 of
+    case BigMap of
         #{ K := V2 } ->
             V = Combiner(K, V1, V2),
-            intersect_with_iterate(Next, [{K,V}|Keep], Map1, Map2, Combiner);
+            intersect_with_iterate(Next, [{K,V}|Keep], BigMap, Combiner);
         _ ->
-            intersect_with_iterate(Next, Keep, Map1, Map2, Combiner)
+            intersect_with_iterate(Next, Keep, BigMap, Combiner)
     end;
-intersect_with_iterate(none, Keep, _Map1, _Map2, _Combiner) ->
+intersect_with_iterate(none, Keep, _BigMap2, _Combiner) ->
     maps:from_list(Keep).
 
 %% Shadowed by erl_bif_types: maps:is_key/2


lib/kernel/doc/src/logger.xml
```diff
@@ -1189,6 +1189,18 @@ logger:set_proxy_config(maps:merge(Old, Config)).
...
 
+    <func>
+      <name name="timestamp" arity="0" since="OTP 24.0"/>
+      <fsummary>Reconfigure Logger.</fsummary>
+      <desc>
+        <p>Reconfigure Logger using updated <c>kernel</c> configuration
+	  that was set after <c>kernel</c> application was loaded.</p>
+        <p>Beware, that this is meant to be run only by the build tools,
+	  not manually during application lifetime, as this may cause
+	  missing log entires.</p>
```
Nicd, line 1200:
```suggestion
	  missing log entries.</p>
```

garazdawi:
I feel like this problem should be solvable using `erl -kernel logger '[{handler,default,undefined}]'` and `logger:add_handlers(kernel).`.

I tried to do it, and after a small patch, it works. Do you know if this functionality would work for Elixir and rebar3's usecase?

```
diff --git a/lib/kernel/src/logger.erl b/lib/kernel/src/logger.erl
index ea75c8d720..d5a4edd863 100644
--- a/lib/kernel/src/logger.erl
+++ b/lib/kernel/src/logger.erl
@@ -856,6 +856,7 @@ init_kernel_handlers(Env) ->
 %% and then starting the correct handlers. This is done after the
 %% kernel supervisor tree has been started as it needs the logger_sup.
 add_handlers(kernel) ->
+    init_kernel_handlers(get_logger_env(kernel)),
     Env = get_logger_env(kernel),
     case get_proxy_opts(Env) of
         undefined ->
```

```
> erl -kernel logger '[{handler,default,undefined}]'
Erlang/OTP 24 [RELEASE CANDIDATE 1] [erts-11.2] [source-3e0023e933] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V11.2  (abort with ^G)
1> logger:i().
Primary configuration: 
    Level: notice
    Filter Default: log
    Filters: 
        (none)
Handler configuration: 
    Id: simple
        Module: logger_simple_h
        Level:  all
        Formatter:
            Module: logger_formatter
            Config:
                []
        Filter Default: stop
        Filters:
            Id: remote_gl
                Fun: fun logger_filters:remote_gl/2
                Arg: stop
            Id: domain
                Fun: fun logger_filters:domain/2
                Arg: {log,super,[otp,sasl]}
            Id: no_domain
                Fun: fun logger_filters:domain/2
                Arg: {log,undefined,[]}
Proxy configuration: 
    burst_limit_enable: false
    burst_limit_max_count: 500
    burst_limit_window_time: 1000
    drop_mode_qlen: 1000
    flush_qlen: 5000
    overload_kill_enable: false
    overload_kill_mem_size: 3000000
    overload_kill_qlen: 20000
    overload_kill_restart_after: 5000
    sync_mode_qlen: 500
Level set per module: 
    (none)
ok
2> application:unset_env(kernel,logger).
ok
3> logger:add_handlers(kernel).         
ok
4> logger:i().                          
Primary configuration: 
    Level: notice
    Filter Default: log
    Filters: 
        (none)
Handler configuration: 
    Id: default
        Module: logger_std_h
        Level:  all
        Formatter: 
            Module: logger_formatter
            Config:
                legacy_header: true
                single_line: false
        Filter Default: stop
        Filters:
            Id: remote_gl
                Fun: fun logger_filters:remote_gl/2
                Arg: stop
            Id: domain
                Fun: fun logger_filters:domain/2
                Arg: {log,super,[otp,sasl]}
            Id: no_domain
                Fun: fun logger_filters:domain/2
                Arg: {log,undefined,[]}
        Handler Config:
            burst_limit_enable: true
            burst_limit_max_count: 500
            burst_limit_window_time: 1000
            drop_mode_qlen: 200
            filesync_repeat_interval: no_repeat
            flush_qlen: 1000
            overload_kill_enable: false
            overload_kill_mem_size: 3000000
            overload_kill_qlen: 20000
            overload_kill_restart_after: 5000
            sync_mode_qlen: 10
            type: standard_io
Proxy configuration: 
    burst_limit_enable: false
    burst_limit_max_count: 500
    burst_limit_window_time: 1000
    drop_mode_qlen: 1000
    flush_qlen: 5000
    overload_kill_enable: false
    overload_kill_mem_size: 3000000
    overload_kill_qlen: 20000
    overload_kill_restart_after: 5000
    sync_mode_qlen: 500
Level set per module: 
    (none)
ok
```

hauleth:
@garazdawi the problem is that it will allow configuration **only** handlers, it do not support all other options that can be set via `kernel.logger` list:

- primary filters
- module levels
- proxy configuration

Additionally such system need to manually update primary level and primary metadata. If in future there would be more options added, then all these tools needs to be updated manually, which can cause confusion why some options work and other do not, and only in some environments these options fails to be applied.

garazdawi:
So I've thought about this and agree that this function is a good addition to the logger API. After you add test cases for this I'll put into test to make sure that it works.

garazdawi:
Nice that the internal export API could be removed! I hadn't realized that. Putting your changes into testing.

garazdawi:
There are a bunch of tests (> 10) failing in the kernel application due to this change. You can run:
```bash
./otp_build setup -a
(cd lib/kernel && make test ARGS="-spec logger.spec")
```
in order to only run the logger test cases.

garazdawi:
Let me know when you have fixed the tests and I'll put it back into testing.

hauleth:
@garazdawi I needed to republish `internal_init_logger/0` due to different behaviour when `{error_logger,false}` is set (`reconfigure/0` will remove `simple` logger, as expected, while `internal_init_logger/0` will keep it as is). Now all tests are passing.

lib/kernel/doc/src/logger.xml
```diff
@@ -1189,6 +1189,18 @@ logger:set_proxy_config(maps:merge(Old, Config)).
       </desc>
     </func>
 
+    <func>
+      <name name="reconfigure" arity="0" since="OTP 24.0"/>
```
garazdawi, line 1193:
```suggestion
      <name name="reconfigure" arity="0" since="OTP @OTP-17375@"/>
```

garazdawi (commented):
Great! It is too late to include this for OTP-24 now, but we can put it in OTP-24.1.

hauleth:
@garazdawi That mean that Elixir will be able to use that no sooner than 1.16 or later. Well, my bad then, I should have fix that sooner.

lib/kernel/test/logger_SUITE.erl
```diff
@@ -1009,62 +1010,54 @@ app_config(Config) ->
...
 
     %% This shall mean the same as above, but using 'logger' parameter
     %% instead of 'error_logger'
     ok = rpc:call(Node,application,unset_env,[kernel,error_logger]),
     ok = rpc:call(Node,application,set_env,
                   [kernel,logger,[{handler,default,undefined}]]),
-    ok = rpc:call(Node,logger,internal_init_logger,[]),
-    ok = rpc:call(Node,logger,add_handlers,[kernel]),
+    ok = rpc:call(Node,logger,reconfigure,[]),
     LC = rpc:call(Node,logger,get_config,[]),
```
garazdawi, line 1029: This testcase fails on this line with a badmatch.

hauleth: Ok, I need to check, as it was passing on my machine last time I checked.

garazdawi: Any news on this?

I had a quick look and from what I can tell the problem is that logger:reconfigure does not re-add the simple handler before calling internal_init_logger.

The simple handler is the handler that is used during startup before the entire logging system is started and internal_init_logger assumes that that handler is installed.

If you fix that, then the call to `logger:add_handlers(kernel)` should not have to be made anymore.

Also, don't you need to clear the primary filters before doing internal_init_logger?

hauleth: I now have found some time to fix that. So it should be done soonish.

PR #4665, opened 2021-03-24
alloc_SUITE, distribution_SUITE: fix race conditions

max-au:
Distribution suite suffers from the race between logger warning emitted, and "true" printed by the test code. There seems to be no specific order, logger may print before or after test suite outpit.
Alloc may fail when stop_node stops the node really fast, and linked process causes test case to exit with `{'EXIT', noconnection}`


jhogberg:
Merged, thanks for the PR!

PR #4667, opened 2021-03-24
Document BEAM SSA invariants

frej:
This documents some BEAM SSA invariants which would have been helpful to know when I started hacking the Erlang compiler.

The patch to docgen to support internal documentation in the form of both chapters and modules feels somewhat hackish, so helpful suggestions on how to improve it are welcome.


bjorng:
Thanks for your pull request.

PR #4672, opened 2021-03-25
Additionally log message queue len when process killed by max_heap_size

juise:
In most cases, it's very helpful to see not only GC information when the process killed by max_heap_size, but also the message queue len.

```
=ERROR REPORT==== 25-Mar-2021::00:19:17.654030 ===
     Process:            <0.121.0>
     Context:            maximum heap size reached
     Max Heap Size:      10000
     Total Heap Size:    389887
     Kill:               true
     Error Logger:       true
     Message Queue Len:  46
     GC Info:            [{old_heap_block_size,610},
                          {heap_block_size,197026},
                          {mbuf_size,192240},
                          {recent_size,0},
                          {stack_size,10},
                          {old_heap_size,294},
                          {heap_size,363},
                          {bin_vheap_size,0},
                          {bin_vheap_block_size,46422},
                          {bin_old_vheap_size,16},
                          {bin_old_vheap_block_size,46422}]
```

### DIFF
--- a/erts/emulator/beam/erl_gc.c
+++ b/erts/emulator/beam/erl_gc.c
@@ -3477,19 +3477,21 @@ reached_max_heap_size(Process *p, Uint total_heap_size,
             Eterm *o_hp, *hp, args = NIL;
 
             /* Build the format message */
-            erts_dsprintf(dsbufp, "     Process:          ~p ");
+            erts_dsprintf(dsbufp, "     Process:            ~p ");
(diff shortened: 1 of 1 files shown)


garazdawi:
thanks! putting into testing.

garazdawi:
Thanks for the PR! Someone (i.e. me) should really rewrite that error report to use a logger report instead of logging in plain text.

PR #4676, opened 2021-03-26
erts: Support frame pointers in JIT

jhogberg:
One of the nicest side effects of the JIT is that it allows us to use off-the-shelf profilers like `perf`, which have far lower overhead than our own (e.g. `fprof`). Unfortunately they can't always follow what happens due to lack of debug information and frame pointers, making the results somewhat incomplete.

It's possible to get halfway there using `perf`'s last-branch-record (`--call-graph=lbr`) mode that uses processor-specific features to figure out which functions call each other. We've found it very useful but unfortunately it only works on Intel processors, tends to be disabled in virtualized environments (Intel or not), and it can attribute execution time to the wrong functions since it interprets process rescheduling as a function call.

This PR adds frame pointers when `perf` support is turned on (`+JPperf true`), letting us use the frame pointer mode (`--call-graph=fp`) which has none of the above problems.


jhogberg:
> and it can attribute execution time to the wrong functions since it interprets process rescheduling as a function call.

Here's a deliberately provoked example of this, taken from a compiler run with a single scheduler:

image

This looks amazing at a glance -- `epp:scan_toks/2` does communicate with a file IO server in this example -- but unfortunately it only got this right by accident. While the profiler saw how the control flow often went from `file_io_server:server_loop/1` to `epp:scan_toks/2` and vice versa, the processes were rescheduled in between and we could just as well have landed somewhere unrelated and had that attributed to either of these functions. Such "leaks" are difficult to catch because they only rarely show up in the listing, and when they do they're often unwanted noise.

Here's the same example but recorded with `--call-graph=fp` mode:

image

PR #4678, opened 2021-03-26
Remove a test for perl 4

DianaOlympos:
This test in the autoconf is both limited, relatively dangerous
(changequote really...) and simply redundant.

If someone built OTP on a machine where perl 4 is the default
they called it on themselves.

Yes i know it will be considered that this test cost "nothing" but it does have a cost. This code is "write only" and that is not a good thing. It is a small change, but i have more coming if this sound like a good idea.


bjorng:
Thanks for your pull request.

PR #4679, opened 2021-03-26
crypto: Remove OpenSSL search through uninstall

DianaOlympos:
On Windows, we in rare case try a "last chance" search through the uninstall registry.

This basically reward broken installs and is really brittle code. The crypto configuration code is already complex enough as it is.


DianaOlympos:
I updated the commit to delete the extra_dir

Issue #4482, opened 2020-12-17
ERL-1446: Repeated Erlang/OTP compilations from different source directories don't produce identical build results


OTP-Maintainer:
Original reporter: `jeromedebretagne`
Affected version: `OTP-23.1`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1446

---

```
+Steps to reproduce+

$ cd otp_src1
 $ git checkout OTP-23.1.5
 $ export ERL_COMPILER_OPTIONS=deterministic
 $ git clean -xfdq
 $ ./otp_build autoconf
 $ ./otp_build configure
 $ make -j4
 $ make RELEASE_ROOT=/path/to/build/directory1 release
  
 $ cd otp_src2
 $ git checkout OTP-23.1.5
 $ export ERL_COMPILER_OPTIONS=deterministic
 $ git clean -xfdq
 $ ./otp_build autoconf
 $ ./otp_build configure
 $ make -j4
 $ make RELEASE_ROOT=/path/to/build/directory2 release

Comparing the 2 build directory outputs show differences in more than 600 files, in 29 .erl files, in more than 500 .beam files and in many of the libraries and executables.

Cf. the attached output showing all the non-identical files.


  
 +Expected result+
  
 Builds should produce deterministic results.
 Cf. [https://reproducible-builds.org/] for why it matters.
```

JeromeDeBretagne:
I'm re-attaching the original output: [non_deterministic_builds_from_different_dirs.git-status.txt](https://github.com/erlang/otp/files/6079207/non_deterministic_builds_from_different_dirs.git-status.txt)

TD5:
Potentially the fix in #5141 will have some on effect on the non-determinism seen when moving between compilations in different directories, by virtue of respecting the `deterministic` option more reliably for `?FILE` macro usage.

Other sources of non-determinism due to a dependency on the particular compilation directory path being used that I am aware of are:
 - [`asn1ct_gen.erl`](https://github.com/erlang/otp/blob/33e7948d29c6b67f10939fd0b66e1f1079f3ebe1/lib/asn1/src/asn1ct_gen.erl#L1332), which can bake in the `cwd`, `outdir`, etc. via the `options` it records in the `-asn1_info` attribute.
[...]

Issue #4417, opened 2020-12-17
ERL-1445: Repeated Erlang/OTP compilations from the same source directory don't produce identical build results


OTP-Maintainer:
Original reporter: `jeromedebretagne`
Affected version: `OTP-23.1`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1445

---

```
+Steps to reproduce+

$ cd otp_src
 $ git checkout OTP-23.1.5
 $ export ERL_COMPILER_OPTIONS=deterministic
 $ git clean -xfdq
 $ ./otp_build autoconf
 $ ./otp_build configure
 $ make -j4
 $ make RELEASE_ROOT=/path/to/build/directory1 release

$ # from the same otp_src directory
 $ git clean -xfdq
 $ ./otp_build autoconf
 $ ./otp_build configure
 $ make -j4
 $ make RELEASE_ROOT=/path/to/build/directory2 release

Comparing the 2 buld directory outputs show differences in beam.smp, OtpErlang.jar and in a few .hrl files due to the introduction of timestamps.

Cf. the attached diff file.

 

+Expected result+

Builds should produce deterministic results.
 Cf. [https://reproducible-builds.org/] for why it matters.
```

max-au:
This one looks identical to https://github.com/erlang/otp/issues/4482

paulo-ferraz-oliveira:
> This one looks identical to #4482

There's this: https://github.com/erlang/otp/issues/4482#issuecomment-776589648. 👍

JeromeDeBretagne:
I'm re-attaching the original diff file: [non_deterministic_builds.diff](https://github.com/erlang/otp/files/6079185/non_deterministic_builds.diff.txt)

JeromeDeBretagne:
I've made progress with #4683 and #4684 which get rid of most differences, by removing timestamps and version numbers in various generated files.

Trickier now is to understand the source of delta in the 2 remaining binary files: beam.smp and OtpErlang.jar

jhogberg:
> Trickier now is to understand the source of delta in the 2 remaining binary files: beam.smp and OtpErlang.jar

`beam.smp` differs because the C toolchain isn't deterministic, some variants can be made deterministic with the right incantation of options but I'm not sure how reliable we can make it -- we'll most likely have to settle for best effort. I'm guessing something similar affects the JAR file but I haven't looked deeper into it.

[...]

PR #4683, opened 2021-03-28
snmp: Make snmp build reproducibly

JeromeDeBretagne:
Remove the current date/time and snmp version from the header section of .hrl files generated by snmpc_mib_to_hrl.

Contributes to #4417 (ERL-1445) and #4482 (ERL-1446).


bmk:
I am ok with removing the time/date info but I would like to keep the 'version'.

JeromeDeBretagne:
Micael, could you explain the expected value of keeping this version in a comment of the generated header files ? It makes the code change regularly with little (obvious) benefit. If there is a usage I'm missing, I'm happy to add a corresponding comment to document it. In complement, snmpc_mib_to_hrl.erl hasn't changed in the last 10 years so it seems really stable.

bmk:
If the version changes often, then I am thinking of a different 'version'.
I was thinking of 'MIB compiler version', which has not changed for a
long time (so I did not see the problem).

JeromeDeBretagne:
> If the version changes often, then I am thinking of a different 'version'.
> I was thinking of 'MIB compiler version', which has not changed for a
> long time (so I did not see the problem).

If I'm correctly understanding the build system, the 'version' Macro seems defined in this Makefile here:   

https://github.com/erlang/otp/blob/master/lib/snmp/src/compile/Makefile#L32-L34
https://github.com/erlang/otp/blob/master/lib/snmp/src/compile/Makefile#L80-L84

```
include ../../vsn.mk

VSN = $(SNMP_VSN)

ERL_COMPILE_FLAGS += -I../../include \
                     -Dversion=\"$(VSN)$(PRE_VSN)\" \
```

so version is in sync with SNMP_VSN. In my initial tests, it was 5.6.1 and now it is 5.8 indeed.

Since there are many other ways to get the SNMP app version when debugging/reviewing code, I would still propose to remove this instance of 'version' found in a header comment as it adds little benefit I feel.

If you feel strongly it should be kept, let me know and I'll update my PR proposition accordingly. Thank you

bmk:
Still not the version I was thinking about.
Digging a bit deeper I managed to find what I was looking for: mib-format-version
This is not the version printed here though.
Of course there is no way to find out what version of the (mib-) compiler was used when 'generating'
a file. But since we are talking about a rather trivial header file it can be ignored.

bmk (approved):

JeromeDeBretagne:
@jhogberg Let me know if you have any feedback, now that @bmk has reviewed the proposed change. Thank you.

IngelaAndin:
ping @jhogberg

jhogberg:
Sorry about taking so long to get back on this, I missed the notification :(

 It looks good, thanks for the PR!

PR #4685, opened 2021-03-28
reltool: Generate files reproducibly

JeromeDeBretagne:
Remove the current date/time in rel, script and application resource files generated by reltool_target.

Contributes to #4417 (ERL-1445) and #4482 (ERL-1446).


jhogberg:
Merged, thanks for the PR!

PR #4686, opened 2021-03-28
Add aead attribute to map from crypto:cipher_info/1

### LINKED ISSUE #4491: ERL-1479: Function to list AEAD ciphers
Original reporter: `dmorneau`
Affected version: `OTP-23.2.2`
Component: `crypto`
Migrated from: https://bugs.erlang.org/browse/ERL-1479

---

```
It would be useful to have a way to get the list of supported AEAD ciphers.

The concrete use case is a library that uses crypto_one_time internally, but lets the user configure which cipher is used. Right now we have no way to validate that the cipher suite passed by the user is supported by crypto_one_time besides hardcoding the list of AEAD ciphers.

I think this could either be an {{aead_ciphers}} option to crypto:supports, or an additional attribute in the map returned by crypto:cipher_info.

 
```

starbelly:
- changes map returned by cipher_info/1 nif to include an aead attribute
which indictates whether the cipher is AEAD or not

Closes #4491

### DIFF
--- a/lib/crypto/test/crypto_SUITE.erl
+++ b/lib/crypto/test/crypto_SUITE.erl
@@ -1269,7 +1269,7 @@ cipher_info_aead_attr(Config) when is_list(Config) ->
     case lists:foldl(fun(C,Ok) ->
                         case crypto:cipher_info(C) of
                             #{aead := false} ->
-                                true;
+                                true and Ok;
                             _ ->
                                 false
                         end
--- a/lib/crypto/test/crypto_SUITE.erl
+++ b/lib/crypto/test/crypto_SUITE.erl
@@ -1250,7 +1250,7 @@ cipher_info_prop_aead_attr(Config) when is_list(Config) ->
     AeadCiphers = [aes_128_ccm, aes_192_ccm, aes_256_ccm, aes_128_gcm, aes_192_gcm, aes_256_gcm, chacha20_poly1305],
     case lists:foldl(fun(C,Ok) ->
                         case crypto:cipher_info(C) of
-                            #{aead := true} ->
+                            #{prop_aead := true} ->
                                 true and Ok;
                             _ ->
                                 false
--- a/lib/crypto/test/crypto_SUITE.erl
+++ b/lib/crypto/test/crypto_SUITE.erl
@@ -57,6 +57,8 @@
          bad_verify_name/1,
          cipher_info/0,
          cipher_info/1,
+         cipher_info_aead_attr/0,
+         cipher_info_aead_attr/1,
          cipher_padding/1,
          compute/0,
          compute/1,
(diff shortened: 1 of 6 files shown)


lhoguin:
Maybe instead of `#{aead => boolean()}` we could have `#{alg => sym | ae | aead | ...}` to be more future-proof.

starbelly:
> Maybe instead of `#{aead => boolean()}` we could have `#{alg => sym | ae | aead | ...}` to be more future-proof.

I like the idea of future proofing, but of course would like to hear from the OTP team as that changes the scope of this PR.

 I also have a few concerns around things getting a little confusing. For example, `chacha20_poly1305` can be said to be all three of the types you listed in your example. The other issue is simply being able to get this information via the EVP api. 
AEAD was easy as it's a "special" flag if you will, there are no flags at a quick glance for determining the other types, but openssl is quite large so it may be something I overlooked. Of course, hard coding all the information if absolutely needed is a possibility, but don't like that path.

lhoguin:
Maybe `#{alg => aead}` or `#{alg => [aead]}` if others have no value. What I think should be avoided is that one day we end up with `#{aead => boolean(), aead2 => boolean(), aead3 => boolean()...}`.

HansN:
I'm in favor of "single values" like #{attr_1 => true, attr_2 => 17 etc}.
The reason is that it is more verbose and slower to always use lists:member to check the value, and the map data structure is designed to keep lots of values.  Being more verbose often means being harder to read.
There is a different situation in C-like languages where one can back flags in one integer.
One comment I got discussing this was with some people was that the keys should be prefixed by something that classifies them and prevents name clashes.  The PR uses "property" in a comment, so maybe 'prop_'  as prefix?

lib/crypto/test/crypto_SUITE.erl
```diff
@@ -1242,6 +1244,45 @@ cipher_info(Config) when is_list(Config) ->
...
     end.
                                                                          
+cipher_info_aead_attr() ->
+    [{doc, "crypto cipher_info aead attribute testing"}].
+cipher_info_aead_attr(Config) when is_list(Config) ->
+    AeadCiphers = [aes_128_ccm, aes_192_ccm, aes_256_ccm, aes_128_gcm, aes_192_gcm, aes_256_gcm, chacha20_poly1305],
+    case lists:foldl(fun(C,Ok) ->
+                        case crypto:cipher_info(C) of
+                            #{aead := true} ->
+                                true;
```
HansN, line 1254:
```suggestion
                                true and Ok;
```
so one 'false' make the accumulator stay 'false'?  Alternatively use lists:any/2 or lists:all/2.

lib/crypto/test/crypto_SUITE.erl
```diff
@@ -1242,6 +1244,45 @@ cipher_info(Config) when is_list(Config) ->
...
+        true ->
+            ok;
+        false ->
+            ct:fail('AEAD Cipher attribute reported false',[])
+    end,
+    NonAeadCiphers = [aes_ige256, blowfish_cbc, blowfish_cfb64],
+    case lists:foldl(fun(C,Ok) ->
+                        case crypto:cipher_info(C) of
+                            #{aead := false} ->
+                                true;
```
HansN, line 1272:
```suggestion
                                true and Ok;
```

starbelly:
> The PR uses "property" in a comment, so maybe 'prop_' as prefix?

Can you expand on that? Perhaps you mean `prop_aead` ?

HansN:
Yes, please change to prop_aead.

starbelly:
@HansN done 😄

lib/crypto/test/crypto_SUITE.erl
```diff
@@ -1242,6 +1244,45 @@ cipher_info(Config) when is_list(Config) ->
             ct:fail('Cipher unsupported',[])
     end.
                                                                          
+cipher_info_prop_aead_attr() ->
+    [{doc, "crypto cipher_info prop_aead attribute testing"}].
+cipher_info_prop_aead_attr(Config) when is_list(Config) ->
+    AeadCiphers = [aes_128_ccm, aes_192_ccm, aes_256_ccm, aes_128_gcm, aes_192_gcm, aes_256_gcm, chacha20_poly1305],
+    case lists:foldl(fun(C,Ok) ->
+                        case crypto:cipher_info(C) of
+                            #{aead := true} ->
```
HansN, line 1253:
```suggestion
                            #{prop_aead := true} ->
```

lib/crypto/test/crypto_SUITE.erl
```diff
@@ -1242,6 +1244,45 @@ cipher_info(Config) when is_list(Config) ->
...
+    of
+        true ->
+            ok;
+        false ->
+            ct:fail('AEAD Cipher attribute reported false',[])
+    end,
+    NonAeadCiphers = [aes_ige256, blowfish_cbc, blowfish_cfb64],
+    case lists:foldl(fun(C,Ok) ->
+                        case crypto:cipher_info(C) of
+                            #{aead := false} ->
```
HansN, line 1271:
```suggestion
                            #{prop_aead := false} ->
```

starbelly: Thanks for the catches 👍

HansN:
Thanks for the PR!

PR #4688, opened 2021-03-29
SSL: Make the handshake fail when OCSP staple missing

michalpalka:
* Make the TLS handshake fail when OCSP staple is requested
  but missing.
* Add test case for missing OCSP response.
* Add test case for OCSP response with certificate
  status undetermined.


lib/ssl/test/make_certs.erl
```diff
@@ -177,6 +183,18 @@ revoke(Root, CA, User, C) ->
...
+remove_entry(Root, CA, User, C) ->
+    Db = filename:join([Root, CA, "index.txt"]),
+    Cmd = ["grep", " -v"
+       " \"/CN=", User ,"/\" ", Db,
+       " > ", Db, "~new"],
+    cmd(Cmd, []),
+    Cmd2 = ["mv ", Db, "~new ", Db],
+    cmd(Cmd2, []),
+    gencrl(Root, CA, C).
+
```
IngelaAndin, line 197: This is not a portable way of doing this, and it breaks our windows tests. I suggest using file:rename/2  instead.

michalpalka: I will change that. What about running grep? Is this going to be portable?

dgud: Preferable not,
 but grep is available if you use it via wsl, something like
 os:cmd("wsl.exe -e  grep Pattern File"),  with the correct quotes (both window and bash) around Pattern and File,
which might be a bit tricky.

michalpalka: I see. We need a way of removing one line from the OpenSSL CA database file. Do you have a suggestion how to do it in a portable way?

dgud: First thing that came up is:

```
remove_line_with_pattern(File, Pattern) ->
    {ok, Bin} = file:read_file(File),
    AllLines = string:lexemes(Bin, [$\n,"\r\n"]),
    MaybeRemove = fun(Line, Acc) ->
                          case string:find(Line, Pattern) of
                              nomatch -> [Line|Acc];
                              _ -> Acc
                          end
                  end,
    RevLines = lists:foldl(MaybeRemove, [], AllLines),
    Lines = lists:join("\n", lists:reverse(RevLines)),
    ok = file:write_file(File, iolist_to_binary(Lines)).
```

This overwrites the file.  tested once on nice input :-)

michalpalka: Great! Thanks! I will give it a try!

michalpalka: I have used your code in the latest change. I haven't tested it on Windows, but I see no reason why it would not work there.

I have removed the call to iolist_to_binary(), because write_file() accepts an iolist.

IngelaAndin: Thank you I have now re-enabled the test again. They are automatically disabled when you push something new.

IngelaAndin:
Thank you for the PR

PR #4689, opened 2021-03-29
Perf and msacc JIT fixes

garazdawi:
* Flush perf logs after module is added
  * This makes `perf top` recognize the correct symbols right away instead of having to wait until the buffer is flushed.
*  Clear msacc state after exiting light bif call


jhogberg (approved):

garazdawi:
Added an additional change that renames the undocumented option `+asmdump` to `+JDdump true` for consistency.

max-au:
Confirming this fixes MSACC extra states, thank you!

PR #4690, opened 2021-03-29
Base16 algorithm improvements

NelsonVides:
As base16 encoding and decoding was introduced in https://github.com/erlang/otp/pull/3014, I'm now proposing a more performant algorithm. Regardless of the input size, encoding is generally 2x-3x faster, and decoding 1.5x-2x faster. The biggest difference is in memory consumption, where both are constant 128B memory consumption, while the original algorithm grows the heap linearly to the input size. Details below. This algorithm is also very much similar to what Elixir does in https://github.com/elixir-lang/elixir/blob/76d245b6081c53228bf99fc1494add5de7872065/lib/elixir/lib/base.ex

---

Benchmarked using benchee with this code:
```elixir
Benchee.run(
  %{
    "encode optimised" => fn input -> :bench_erl.encode(input) end,
    "encode otp" => fn input -> :bench_erl.encode_hex(input) end
  },
  inputs: %{
    "1. Tiny" => :crypto.strong_rand_bytes(8),
    "4. Medium Big" => :crypto.strong_rand_bytes(512),
    "6. Large" => :crypto.strong_rand_bytes(64000)
  },
  parallel: 12,
  memory_time: 5
)

Benchee.run(
  %{
    "decode optimised" => fn input -> :bench_erl.my_decode(input) end,
    "decode otp" => fn input -> :bench_erl.decode_hex(input) end
  },
  inputs: %{
    "1. Tiny" => :bench_erl.encode(:crypto.strong_rand_bytes(8)),
    "4. Medium Big" => :bench_erl.encode(:crypto.strong_rand_bytes(512)),
    "6. Large" => :bench_erl.encode(:crypto.strong_rand_bytes(64000))
  },
  parallel: 12,
  memory_time: 5
)
```

And with the following results:
```
Operating System: Linux
CPU Information: Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
Number of Available Cores: 12
Available memory: 30.99 GB
Elixir 1.11.0
Erlang 23.3

warmup: 2 s
time: 5 s
memory time: 5 s
parallel: 12
inputs: 1. Tiny, 4. Medium Big, 6. Large
Estimated total run time: 1.20 min

Benchmarking encode optimised with input 1. Tiny...
Benchmarking encode optimised with input 4. Medium Big...
Benchmarking encode optimised with input 6. Large...
Benchmarking encode otp with input 1. Tiny...
Benchmarking encode otp with input 4. Medium Big...
Benchmarking encode otp with input 6. Large...

##### With input 1. Tiny #####
Name                       ips        average  deviation         median         99th %
encode optimised      904.19 K        1.11 μs  ±3872.76%        0.76 μs        2.27 μs
encode otp            463.98 K        2.16 μs  ±2088.18%        1.53 μs        5.30 μs

Comparison:
encode optimised      904.19 K
encode otp            463.98 K - 1.95x slower +1.05 μs

Memory usage statistics:

Name                Memory usage
encode optimised           128 B
encode otp                 408 B - 3.19x memory usage +280 B

**All measurements for memory usage were the same**

##### With input 4. Medium Big #####
Name                       ips        average  deviation         median         99th %
encode optimised       38.20 K       26.17 μs   ±191.69%       24.69 μs       41.38 μs
encode otp             12.01 K       83.24 μs    ±40.94%       82.62 μs       91.06 μs

Comparison:
encode optimised       38.20 K
encode otp             12.01 K - 3.18x slower +57.06 μs

Memory usage statistics:

Name                Memory usage
encode optimised        0.125 KB
encode otp              20.09 KB - 160.69x memory usage +19.96 KB

**All measurements for memory usage were the same**

##### With input 6. Large #####
Name                       ips        average  deviation         median         99th %
encode optimised        316.74        3.16 ms    ±15.09%        3.14 ms        3.76 ms
encode otp               96.11       10.40 ms     ±9.49%       10.30 ms       13.24 ms

Comparison:
encode optimised        316.74
encode otp               96.11 - 3.30x slower +7.25 ms

Memory usage statistics:

Name                Memory usage
encode optimised      0.00012 MB
encode otp               2.44 MB - 20000.69x memory usage +2.44 MB

**All measurements for memory usage were the same**
Operating System: Linux
CPU Information: Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
Number of Available Cores: 12
Available memory: 30.99 GB
Elixir 1.11.0
Erlang 23.3

warmup: 2 s
time: 5 s
memory time: 5 s
parallel: 12
inputs: 1. Tiny, 4. Medium Big, 6. Large
Estimated total run time: 1.20 min

Benchmarking decode optimised with input 1. Tiny...
Benchmarking decode optimised with input 4. Medium Big...
Benchmarking decode optimised with input 6. Large...
Benchmarking decode otp with input 1. Tiny...
Benchmarking decode otp with input 4. Medium Big...
Benchmarking decode otp with input 6. Large...

##### With input 1. Tiny #####
Name                       ips        average  deviation         median         99th %
decode optimised      814.04 K        1.23 μs  ±3774.78%        0.98 μs        2.60 μs
decode otp            460.80 K        2.17 μs  ±2139.78%        1.56 μs        5.41 μs

Comparison:
decode optimised      814.04 K
decode otp            460.80 K - 1.77x slower +0.94 μs

Memory usage statistics:

Name                Memory usage
decode optimised           128 B
decode otp                 408 B - 3.19x memory usage +280 B

**All measurements for memory usage were the same**

##### With input 4. Medium Big #####
Name                       ips        average  deviation         median         99th %
decode optimised       18.81 K       53.16 μs   ±219.64%       51.72 μs       72.63 μs
decode otp             12.12 K       82.52 μs   ±131.22%       81.57 μs       93.18 μs

Comparison:
decode optimised       18.81 K
decode otp             12.12 K - 1.55x slower +29.36 μs

Memory usage statistics:

Name                Memory usage
decode optimised        0.125 KB
decode otp              20.09 KB - 160.69x memory usage +19.96 KB

**All measurements for memory usage were the same**

##### With input 6. Large #####
Name                       ips        average  deviation         median         99th %
decode optimised        148.29        6.74 ms     ±5.34%        6.72 ms        7.13 ms
decode otp               96.59       10.35 ms     ±6.91%       10.28 ms       11.58 ms

Comparison:
decode optimised        148.29
decode otp               96.59 - 1.54x slower +3.61 ms

Memory usage statistics:

Name                Memory usage
decode optimised      0.00012 MB
decode otp               2.44 MB - 20000.69x memory usage +2.44 MB

**All measurements for memory usage were the same**
```


garazdawi:
Hello! The stdlib/binary_module_SUITE:error_info/1 testcase fails when testing this branch. It is specifically `binary:decode_hex(<<"GG">>)` that does not print the correct error message in the shell when invoked.

NelsonVides:
> Hello! The stdlib/binary_module_SUITE:error_info/1 testcase fails when testing this branch. It is specifically `binary:decode_hex(<<"GG">>)` that does not print the correct error message in the shell when invoked.

Ag, missed that change. Should be fixed now.

lib/stdlib/src/binary.erl
```diff
@@ -365,60 +365,123 @@ get_opts_replace(_,_) ->
...
+unhex(25697) -> 218; unhex(25698) -> 219; unhex(25699) -> 220; unhex(25700) -> 221; unhex(25701) -> 222; unhex(25702) -> 223;
+unhex(25904) -> 224; unhex(25905) -> 225; unhex(25906) -> 226; unhex(25907) -> 227; unhex(25908) -> 228; unhex(25909) -> 229; unhex(25910) -> 230; unhex(25911) -> 231; unhex(25912) -> 232; unhex(25913) -> 233;
+unhex(25921) -> 234; unhex(25922) -> 235; unhex(25923) -> 236; unhex(25924) -> 237; unhex(25925) -> 238; unhex(25926) -> 239;
+unhex(25953) -> 234; unhex(25954) -> 235; unhex(25955) -> 236; unhex(25956) -> 237; unhex(25957) -> 238; unhex(25958) -> 239;
+unhex(26160) -> 240; unhex(26161) -> 241; unhex(26162) -> 242; unhex(26163) -> 243; unhex(26164) -> 244; unhex(26165) -> 245; unhex(26166) -> 246; unhex(26167) -> 247; unhex(26168) -> 248; unhex(26169) -> 249;
+unhex(26177) -> 250; unhex(26178) -> 251; unhex(26179) -> 252; unhex(26180) -> 253; unhex(26181) -> 254; unhex(26182) -> 255;
+unhex(26209) -> 250; unhex(26210) -> 251; unhex(26211) -> 252; unhex(26212) -> 253; unhex(26213) -> 254; unhex(26214) -> 255;
+unhex(Char) ->
     badarg_with_cause([<<Char>>], invalid_hex).
```
garazdawi, line 485:
```suggestion
    badarg_with_info([<<Char:16>>]).
```
Since we are decoding two bytes at the same time, the error info needs this. Also, we can remove the invalid_hex as it is not needed.

lib/stdlib/src/erl_stdlib_errors.erl
```diff
@@ -78,7 +78,7 @@ format_binary_error(decode_hex, [Subject], _) ->
         true ->
             [must_be_binary(Subject)]
     end;
-format_binary_error(decode_hex_char, [Subject], _) ->
+format_binary_error(unhex, [Subject], _) ->
     if
         is_integer(Subject),
```
garazdawi, line 83: I think we can remove the `if` here and always return `[range]`. Or maybe even better to return `[<<"not a valid hex">>]`.

garazdawi (changes requested):

NelsonVides:
Applied, and perhaps with a bit opinionated choice for a readable and complete error message 🤔

lib/stdlib/src/binary.erl
```diff
@@ -365,61 +365,124 @@ get_opts_replace(_,_) ->
...
 %% Hex encoding functions
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+-define(HEX(X), (hex(X)):16).
 -spec encode_hex(Bin) -> Bin2 when
       Bin :: binary(),
-      Bin2 :: binary().
-encode_hex(Bin) when is_binary(Bin) ->
-    encode_hex(Bin, <<>>);
+      Bin2 :: <<_:_*16>>.
+encode_hex(Data) when byte_size(Data) band 7 =:= 0 ->
```
hauleth, line 372: Shouldn't it be like that:

```suggestion
encode_hex(Data) when byte_size(Data) band 8 =:= 0 ->
```

lib/stdlib/src/binary.erl
```diff
@@ -365,61 +365,124 @@ get_opts_replace(_,_) ->
...
+encode_hex(Data) when byte_size(Data) rem 5 =:= 0 ->
+    << <<?HEX(A),?HEX(B),?HEX(C),?HEX(D),?HEX(E)>> || <<A,B,C,D,E>> <= Data >>;
+encode_hex(Data) when byte_size(Data) rem 4 =:= 0 ->
+    << <<?HEX(A),?HEX(B),?HEX(C),?HEX(D)>> || <<A,B,C,D>> <= Data >>;
+encode_hex(Data) when byte_size(Data) rem 3 =:= 0 ->
+    << <<?HEX(A),?HEX(B),?HEX(C)>> || <<A,B,C>> <= Data >>;
+encode_hex(Data) when byte_size(Data) rem 2 =:= 0 ->
+    << <<?HEX(A),?HEX(B)>> || <<A,B>> <= Data >>;
+encode_hex(Data) when is_binary(Data) ->
+    << <<?HEX(N)>> || <<N>> <= Data >>;
```
hauleth, line 387: Have you tested it with

```suggestion
encode_hex(Data) when byte_size(Data) > 7 ->
    Size = byte_size(Data),
    Mul8 = 8 * (Size div 8),
    <<Prefix:Mul8/binary, Suffix/binary>>,
    PrefixEncoded =<< <<?HEX(A),?HEX(B),?HEX(C),?HEX(D),?HEX(E),?HEX(F),?HEX(G),?HEX(H)>> || <<A,B,C,D,E,F,G,H>> <= Prefix >>,
    SuffixEncoded = encode_hex(Suffix),
    <<PrefixEncoded/binary, SuffixEncoded/binary>>;
encode_hex(<<A,B,C,D,E,F,G>>) ->
    <<?HEX(A),?HEX(B),?HEX(C),?HEX(D),?HEX(E),?HEX(F),?HEX(G)>>;
encode_hex(Data) when byte_size(Data) ->
    <<?HEX(A),?HEX(B),?HEX(C),?HEX(D),?HEX(E),?HEX(F)>>;
encode_hex(<<A,B,C,D,E>>) ->
    <<?HEX(A),?HEX(B),?HEX(C),?HEX(D),?HEX(E)>>;
encode_hex(<<A,B,C,D,E>>) ->
    <<?HEX(A),?HEX(B),?HEX(C),?HEX(D)>>;
encode_hex(<<A,B,C>>) ->
    <<?HEX(A),?HEX(B),?HEX(C)>>;
encode_hex(<<A,B>>) ->
    <<?HEX(A),?HEX(B)>>;
encode_hex(<<A>>) ->
    <<?HEX(N)>>;
encode_hex(<<>>) ->
    <<>>;
```

NelsonVides: No taking into account that the snippet doesn't compile (but it was obvious how to fix it), this wouldn't bring better results, and it doesn't make the code size (.beam file) smaller either:

```
##### With input 1. Tiny #####
Name                       ips        average  deviation         median         99th %
encode optimised      935.95 K        1.07 μs  ±3968.58%        0.75 μs        2.21 μs
encode suggested      641.82 K        1.56 μs  ±3046.43%        1.11 μs        2.93 μs

Comparison:
encode optimised      935.95 K
encode suggested      641.82 K - 1.46x slower +0.49 μs

Memory usage statistics:

Name                Memory usage
encode optimised           128 B
encode suggested           248 B - 1.94x memory usage +120 B

**All measurements for memory usage were the same**

##### With input 4. Medium Big #####
Name                       ips        average  deviation         median         99th %
encode optimised       38.58 K       25.92 μs   ±182.31%       24.46 μs       40.97 μs
encode suggested       34.58 K       28.92 μs   ±291.90%       27.05 μs       45.45 μs

Comparison:
encode optimised       38.58 K
encode suggested       34.58 K - 1.12x slower +3.00 μs

Memory usage statistics:

Name                Memory usage
encode optimised           128 B
encode suggested           264 B - 2.06x memory usage +136 B

**All measurements for memory usage were the same**

##### With input 6. Large #####
Name                       ips        average  deviation         median         99th %
encode optimised        321.56        3.11 ms     ±6.02%        3.10 ms        3.30 ms
encode suggested        312.56        3.20 ms     ±6.48%        3.20 ms        3.41 ms

Comparison:
encode optimised        321.56
encode suggested        312.56 - 1.03x slower +0.0895 ms

Memory usage statistics:

Name                Memory usage
encode optimised           128 B
encode suggested           264 B - 2.06x memory usage +136 B

**All measurements for memory usage were the same**
```

lib/stdlib/src/binary.erl
NelsonVides: Don't think so. `Size band 7 =:= 0` means that the first three bits of the size are zeroes, that is, the size is not a multiple of any number from 1 to 7, i.e., the size is a multiple of 8. So we can digest in chunks of 8 bytes.
`Size band 8 =:= 0` means that the fourth bit of the size is a zero, that is, the size is not a multiple of 8, which breaks the purpose of this clause.

hauleth: Ok, I was tripped by the fact that this is only place where you use `band` instead of `rem`. Is there any performance penalty there? And if yes, then wouldn't it be better to use `byte_size(Data) band 7 =:= X` in other checks as well?

NelsonVides: That actually left me thinking how the beam handles these bitwise operations. In C there's reason to believe ands and ors to be theoretically faster than rems, unless the compiler is smart enough to see what's happening. Changing the and to a rem seems to worsen performance some 5% on my machine, weird. Wait, I'm going to experiment changing the logic of all clauses to use ands and ors 🤔

hauleth: > In C there's reason to believe ands and ors to be theoretically faster than rems, unless the compiler is smart enough to see what's happening. 

AFAIK any popular optimising compiler is "smart enough". If compilers can do things like changing between shifts/adds and multiply ([in both ways](https://youtu.be/bSkpMdDe4g4?t=1761)).

NelsonVides: Just played a bit with benchee, proper, and a good bunch of clause reorderings and bitwise operations, and they all seem to return pretty much exactly the same performance, variations are too minimal, probably a matter of noise (cpu getting hot or whatever). Compiler should handle that nice, so I'd just leave the most readable options here.

hauleth: I think that `rem` would be most readable, but `band` can be there as well as long as all heads will use the same approach.

garazdawi: If you want to try to chase the last couple of nanoseconds, you may want to try and inline the hex function. Not sure if it will show up in the benchmarks, but it should make things faster and since the literal for the tuple is de-duplicated it won't increase the size by much.

```erlang
-compile({inline,[hex/1]}).
```

NelsonVides: I've just tried the inline and made sure the code-size doesn't grow. It does makes thing slightly faster, somewhere between 5-15% faster, so enough to say it's not benchmarking noises 🙂

NelsonVides:
Just got otp-24 with elixir 1.12 release candidates working, and ran the same benchmarks on the same system, with the JIT. Memory consumption is identical in all cases and inputs. CPU-wise, the ratio optimised/current implementation used to be around 2x to 3.2x, now it is 2.5x to 4x faster, so it gets even better with the JIT.

Some selected details, let's say on inputs ~256bytes big, to get a comparison of each algorithm, and with and without the JIT:

## For OTP23
```
Name                       ips        average  deviation         median         99th %
encode optimised       70.45 K       14.19 μs   ±209.09%       13.17 μs       22.23 μs
encode otp             23.67 K       42.24 μs    ±80.39%       41.49 μs       49.08 μs

Comparison:
encode optimised       70.45 K
encode otp             23.67 K - 2.98x slower +28.05 μs

Name                Memory usage
encode optimised        0.125 KB
encode otp              10.09 KB - 80.69x memory usage +9.96 KB
```

## For OTP24

```
Name                       ips        average  deviation         median         99th %
encode optimised       95.24 K       10.50 μs   ±610.62%        8.99 μs       20.84 μs
encode otp             19.52 K       51.24 μs   ±323.06%       32.70 μs      472.52 μs

Comparison:
encode optimised       95.24 K
encode otp             19.52 K - 4.88x slower +40.74 μs

Name                Memory usage
encode optimised        0.125 KB
encode otp              10.09 KB - 80.69x memory usage +9.96 KB
```

garazdawi (approved):
lgtm, running in tests.

garazdawi:
thanks!

PR #4691, opened 2021-03-29
etp-commands: fix decoding of aux-work SSI flags

jktomer:
The `etp-aux-work-flags` gdb subroutine had gotten out of sync with the actual
flags defined in erl_process.h.


garazdawi:
thanks!

PR #4695, opened 2021-03-31
eunit: ensure the output directory exist for surefire xml files

joaohf:
Hello,

This PR is about eunit, when using surefire reports, creating the output directory if it does not exist. The main use case is for Jenkins or any other CI server.


garazdawi:
thanks!

PR #4701, opened 2021-04-01
Fix race in persistent_term:get/0 and info/0

sverker:
Have seen to cause failed assert in debug VM.


max-au (approved):
Confirming this fixes the race.

sverker:
@max-au Thank you for the bug provoking test code.

max-au:
Thanks goes to @potatosalad, who discovered the actual race, I only happened to know the cause.

sverker:
Merged to master for OTP 24 and in pipe for next OTP-23.3.x.

sverker:
Fix released in OTP-23.3.2.

Issue #4666, opened 2021-03-24
[JIT] Incorrect extra states stats for microstate accounting 


max-au:
**Describe the bug**
Microstate accounting reports incorrect data when ERTS is configured with "--with-microstate-accounting=extra" and JIT is enabled. As one consequence, idle time is reported as BIF state.

**To Reproduce**
This test case reproduces the problem in 100% cases:
https://github.com/erlang/otp/blob/master/erts/emulator/test/statistics_SUITE.erl#L658

**Affected versions**
This bug is only present in OTP 24 with BeamAsm JIT. Non-JIT version passes the test.

**Additional context**
It may be caused with a missing call to ERTS_MSACC_UPDATE_CACHE_X somewhere in instr_bif.cpp, probably in emit_call_light_bif_shared or nearby.

PR #4709, opened 2021-04-02
msacc: fix segfault caused when extra MSACC states are enabled

max-au:
8346580b124a recovered extra MSACC states but caused BEAM to dump
core when msacc was switched off due to accidental rewrite of a
variable.


max-au:
closes #4666 again.

garazdawi:
thanks!

PR #4712, opened 2021-04-06
Fix parse transform column handling

garazdawi:
Fix so that parse_transforms are handed the correct AST without columns if requested by either the user or the parse transform.
    
Added test to verify that the column stripping works.

### DIFF
--- a/erts/doc/src/absform.xml
+++ b/erts/doc/src/absform.xml
@@ -195,11 +195,14 @@
             The word <c>LOCATION</c> represents a location, and denotes the
             number of the last line, and possibly the number of the last
 	    column on that line, in the source file. See <seeerl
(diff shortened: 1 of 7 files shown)


erts/doc/src/absform.xml
```diff
@@ -200,6 +200,9 @@
           </p>
         </item>
       </list>
+      <p>See <seetype marker="stdlib:erl_parse#form_info">
+        <c>the form_info/0</c> type</seetype> in
```
uabboli, line 204: I think  "type" is meant to be outside of the <seetype> tag. What at it: the <c> tag is missing in the <seeerl> just above this line.

uabboli (approved):

PR #4715, opened 2021-04-06
queue, maps, proplists: add since information to docs for new functions

Maria-12648430:
In past PRs #2791, #2850, #2900 and #2910, I contributed some new functions to the `queue`, `maps` and `proplists` modules, but forgot to add the `since` information in the documentation.

### DIFF
--- a/lib/stdlib/doc/src/maps.xml
+++ b/lib/stdlib/doc/src/maps.xml
@@ -73,7 +73,7 @@
     </func>
 
     <func>
-      <name name="filtermap" arity="2" since=""/>
+      <name name="filtermap" arity="2" since="OTP 24.0"/>
       <fsummary>Filter and map a map.</fsummary>
       <desc>
         <p>Returns a map <c><anno>Map</anno></c> that is the result of calling
--- a/lib/stdlib/doc/src/proplists.xml
+++ b/lib/stdlib/doc/src/proplists.xml
@@ -138,7 +138,7 @@ expand([{{foo, true}, [bar, baz]}], [{foo, false}, fie, foo, fum])</code>
     </func>
 
     <func>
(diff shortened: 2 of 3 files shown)


garazdawi:
thanks! Someone should really write a script that makes sure that we always add `since` information for any new functions...

Maria-12648430:
@garazdawi with that in mind, what should a contributor of new functions put as since when writing the documentation? In the past PRs, I wan't sure in which OTP release they would appear so left them blank, and then forgot ^^;;;

garazdawi:
> @garazdawi with that in mind, what should a contributor of new functions put as since when writing the documentation? In the past PRs, I wan't sure in which OTP release they would appear so left them blank, and then forgot ^^;;;

You need to ask the OTP maintainer responsible for the feature to provide a ticket number to put there. Or rather, the OTP maintainer should remember to ask you to put a ticket number there.

Like this: https://github.com/erlang/otp/blob/master/erts/doc/src/erlang.xml#L319

As a placeholder, I normally put `@OTP-12345@` to distinguish from functions that have no `since` version.

We use an automatic versioning tool that determines how a version should be bumped depending on what the tickets that are part of the release contain. For the major OTP versions it is rather predictable, but for the individual applications you can end up with only a patch dump in an OTP major release.

Maria-12648430:
Ok, thanks :)

Maria-12648430:
@garazdawi an afterthought... sorry to bother you =^^=
The copyright years in the modules that I changed, I didn't "bump" those. Should I have? ^^;;;

garazdawi:
We bump all of those with a script automatically, so no need.

Maria-12648430:
Ok, thanks again :)

PR #4716, opened 2021-04-07
EDoc: suggest {preprocess, true} when redundant spec assertion fails

erszcz:
This might happen when `-spec`s are generated conditionally by the preprocessor. EDoc has to be aware of this, so `{preprocess, true}` is required.


garazdawi:
thanks!

PR #4724, opened 2021-04-09
queue: property-based tests

Maria-12648430:
This PR provides a property-based test suite for the `queue` module.

It tests all the current functions of the `queue` module (even `new/0`, for completeness' sake), plus a test for a random sequence of insert and retrieval operations performed on a random queue.
The tests were tried out with PropEr, but I assume they work with eqc and triq just as well.

Most of the individual tests rely heavily on `queue:from_list/1` and `queue:to_list/1`. Those are also tested for symmetry. Anyway, I think this is the only way to go without violating the opacity of the `queue` type.

I have to admit that I'm still new to property-based testing, and I hope I didn't make any big mistakes ^^;

### DIFF
--- a/lib/stdlib/test/property_test/queue_prop.erl
+++ b/lib/stdlib/test/property_test/queue_prop.erl
@@ -1,0 +1,670 @@
+%%
+%% %CopyrightBegin%
+%%
+%% Copyright Ericsson AB 2021. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% %CopyrightEnd%
+%%
+-module(queue_prop).
+
+-compile(export_all).
+
+-proptest(eqc).
+-proptest([triq, proper]).
+
+-ifndef(EQC).
+-ifndef(PROPER).
+-ifndef(TRIQ).
+-define(EQC, true).
+-endif.
+-endif.
+-endif.
+
+-ifdef(EQC).
+-include_lib("eqc/include/eqc.hrl").
+-define(MOD_eqc,eqc).
+
+-else.
+-ifdef(PROPER).
+-include_lib("proper/include/proper.hrl").
+-define(MOD_eqc,proper).
+
+-else.
+-ifdef(TRIQ).
+-define(MOD_eqc,triq).
+-include_lib("triq/include/triq.hrl").
+
+-endif.
+-endif.
+-endif.
+
+%%%%%%%%%%%%%%%%%%
+%%% Properties %%%
+%%%%%%%%%%%%%%%%%%
+
+prop_new() ->
+    [] =:= queue:to_list(queue:new()).
+
+prop_is_queue() ->
+    ?FORALL(
+	{IsQueue, Q},
+	oneof([
+	    {true, queue()},
+	    {false, non_queue()}
+	]),
+	begin
+	    IsQueue =:= queue:is_queue(Q)
+	end
+    ).
+
+prop_list_conversion() ->
+    ?FORALL(
+	List,
+	list(),
+	begin
+	    Queue = queue:from_list(List),
+	    queue:is_queue(Queue) andalso
+	    List =:= queue:to_list(Queue)
+	end
+    ).
+
+prop_all() ->
+    ?FORALL(
+	{L, Q},
+	oneof([list_queue(atom()), list_queue(term())]),
+	begin
+	    lists:all(fun is_atom/1, L) =:= queue:all(fun is_atom/1, Q)
+	end
+    ).
+
+prop_any() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    lists:any(fun is_atom/1, L) =:= queue:any(fun is_atom/1, Q)
+	end
+    ).
+
+prop_cons() ->
+    common_in_r_cons(cons).
+
+prop_daeh() ->
+    common_get_r_last_daeh(daeh).
+
+prop_delete() ->
+    ?FORALL(
+	{X, {L, Q}},
+	{term(), list_queue()},
+	begin
+	    R1 = if
+		L =:= [] ->
+		    true;
+		true ->
+		    {_, Y} = hd(lists:sort([{rand:uniform(), I} || I <- L])),
+		    equal(lists:delete(Y, L), queue:delete(Y, Q))
+	    end,
+	    R2 = equal(lists:delete(X, L), queue:delete(X, Q)),
+
+	    R1 andalso R2
+	end
+    ).
+
+prop_delete_r() ->
+    ?FORALL(
+	{X, {L, Q}},
+	{term(), list_queue()},
+	begin
+	    R1 = if
+		L =:= [] ->
+		    true;
+		true ->
+		    {_, Y} = hd(lists:sort([{rand:uniform(), I} || I <- L])),
+		    equal(lists:reverse(lists:delete(Y, lists:reverse(L))), queue:delete_r(Y, Q))
+	    end,
+	    R2 = equal(lists:reverse(lists:delete(X, lists:reverse(L))), queue:delete_r(X, Q)),
+
+	    R1 andalso R2
+	end
+    ).
+
+prop_delete_with() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    Q1 = queue:delete_with(fun is_atom/1, Q),
+	    L1 = case lists:search(fun is_atom/1, L) of
+		false ->
+		    L;
+		{value, V} ->
+		    lists:delete(V, L)
+	    end,
+	    equal(L1, Q1)
+	end
+    ).
+
+prop_delete_with_r() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    Q1 = queue:delete_with_r(fun is_atom/1, Q),
+	    L1 = lists:reverse(L),
+	    L2 = case lists:search(fun is_atom/1, L1) of
+		false ->
+		    L;
+		{value, V} ->
+		    lists:reverse(lists:delete(V, L1))
+	    end,
+	    equal(L2, Q1)
+	end
+    ).
+
+prop_drop() ->
+    common_drop_tail(drop).
+
+prop_drop_r() ->
+    common_drop_r_init_liat(drop_r).
+
+prop_filter() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    Q1 = queue:filter(
+		fun
+		    (I) when is_atom(I) -> true;
+		    (I) when is_integer(I) -> [I * 2];
+		    (I) when is_float(I) -> [{I, I}];
+		    (I) when is_tuple(I) -> [I, I];
+		    (_) -> false
+		end,
+		Q
+	    ),
+	    L1 = lists:foldr(
+		fun
+		    (I, Acc) when is_atom(I) -> [I|Acc];
+		    (I, Acc) when is_integer(I) -> [I * 2|Acc];
+		    (I, Acc) when is_float(I) -> [{I, I}|Acc];
+		    (I, Acc) when is_tuple(I) -> [I, I|Acc];
+		    (_, Acc) -> Acc
+		end,
+		[],
+		L
+	    ),
+	    equal(L1, Q1)
+	end
+    ).
+
+prop_filtermap() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    F = fun
+		(I) when is_atom(I) ->
+		    true;
+		(I) when is_integer(I) ->
+		    {true, {I, I}};
+		(_) ->
+		    false
+	    end,
+	    Q1 = queue:filtermap(F, Q),
+	    L1 = lists:filtermap(F, L),
+	    equal(L1, Q1)
+	end
+    ).
+
+prop_fold() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    % order-independent fold
+	    F1 = fun
+		(I, Acc) when is_number(I) -> Acc + 2 * I;
+		(_, Acc) -> Acc
+	    end,
+	    RQ1 = queue:fold(F1, 0, Q),
+	    RL1 = lists:foldl(F1, 0, L),
+
+	    % order-dependent fold
+	    F2 = fun
+		(I, Acc) -> [{I, I}|Acc]
+	    end,
+	    RQ2 = queue:fold(F2, [], Q),
+	    RL2 = lists:foldl(F2, [], L),
+
+	    RQ1 =:= RL1 andalso
+	    RQ2 =:= RL2
+	end
+    ).
+
+prop_get() ->
+    common_get_head(get).
+
+prop_get_r() ->
+    common_get_r_last_daeh(get_r).
+
+prop_head() ->
+    common_get_head(head).
+
+prop_in() ->
+    ?FORALL(
+	L,
+	list(),
+	begin
+	    Q = lists:foldl(
+		fun (I, Acc) ->
+		    queue:in(I, Acc)
+		end,
+		queue:new(),
+		L
+	    ),
+	    equal(L, Q)
+	end
+    ).
+
+prop_in_r() ->
+    common_in_r_cons(in_r).
+
+prop_init() ->
+    common_drop_r_init_liat(init).
+
+prop_is_empty() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    (length(L) =:= 0) =:= queue:is_empty(Q)
+	end
+    ).
+
+prop_join() ->
+    ?FORALL(
+	{{L1, Q1}, {L2, Q2}},
+	{list_queue(), list_queue()},
+	begin
+	    equal(L1 ++ L2, queue:join(Q1, Q2))
+	end
+    ).
+
+prop_last() ->
+    common_get_r_last_daeh(last).
+
+prop_len() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    length(L) =:= queue:len(Q)
+	end
+    ).
+
+prop_liat() ->
+    common_drop_r_init_liat(liat).
+
+prop_member() ->
+    ?FORALL(
+        {X, {L, Q}},
+	{term(), list_queue()},
+	begin
+	    % all members of L are members of Q
+	    lists:all(
+		fun (I) ->
+		    queue:member(I, Q)
+		end,
+		L
+	    )
+	    andalso
+	    % all members of Q are members of L
+	    lists:all(
+		fun (I) ->
+		    lists:member(I, L)
+		end,
+		queue:to_list(Q)
+	    )
+	    andalso
+	    % if X is a member of L, it is also a member of Q,
+	    % and if X is not a member of L, it is also not a
+	    % member of Q
+	    lists:member(X, L) =:= queue:member(X, Q)
+	end
+    ).
+
+prop_out() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    case queue:out(Q) of
+		{{value, I}, Q1} ->
+		    I =:= hd(L) andalso
+		    equal(tl(L), Q1);
+		{empty, Q1} ->
+		    L =:= [] andalso
+		    equal(L, Q1)
+	    end
+	end
+    ).
+
+prop_out_r() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    case queue:out_r(Q) of
+		{{value, I}, Q1} ->
+		    L1 = lists:reverse(L),
+		    I =:= hd(L1) andalso
+		    equal(lists:reverse(tl(L1)), Q1);
+		{empty, Q1} ->
+		    L =:= [] andalso
+		    equal(L, Q1)
+	    end
+	end
+    ).
+
+prop_peek() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    case queue:peek(Q) of
+		{value, I} ->
+		    I =:= hd(L);
+		empty ->
+		    L =:= []
+	    end
+	end
+    ).
+
+prop_peek_r() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    case queue:peek_r(Q) of
+		{value, I} ->
+		    I =:= lists:last(L);
+		empty ->
+		    L =:= []
+	    end
+	end
+    ).
+
+prop_reverse() ->
+    ?FORALL(
+	{L, Q},
+	list_queue(),
+	begin
+	    equal(lists:reverse(L), queue:reverse(Q))
+	end
+    ).
+
+prop_snoc() ->
+    ?FORALL(
+	L,
+	list(),
+	begin
+	    Q = lists:foldl(
+		fun (I, Acc) ->
+		    queue:snoc(Acc, I)
+		end,
+		queue:new(),
+		L
+	    ),
+	    equal(L, Q)
+	end
+    ).
+
+prop_split() ->
+    ?FORALL(
+	{N, {L, Q}},
+	{non_neg_integer(), list_queue()},
+	begin
+	    N1 = N rem (length(L) + 1),
+	    {Q1, Q2} = queue:split(N1, Q),
+	    {L1, L2} = lists:split(N1, L),
+	    equal(L1, Q1) andalso
+	    equal(L2, Q2)
+	end
+    ).
+
+prop_tail() ->
+    common_drop_tail(tail).
+
+% Test sequences of insert and retrieval operations
+prop_ops() ->
+    ?FORALL(
+	{Ops, {L, Q}},
+	{
+	    list(
+		oneof([{cons, term()},
+		       daeh,
+		       drop,
+		       drop_r,
+		       get,
+		       get_r,
+		       head,
+		       {in, term()},
+		       {in_r, term()},
+		       init,
+		       liat,
+		       last,
+		       out,
+		       out_r,
+		       peek,
+		       peek_r,
+		       {snoc, term()},
+		       tail])
+	    ),
+	    list_queue()
+	},
+	begin
+	    {Res, RQ, RL} = lists:foldl(
+		fun
+		    (_, Acc = {false, _, _}) ->
+			Acc;
+		    % queue:in/2
+		    ({in, I}, {true, Q1, L1}) ->
+			{true, queue:in(I, Q1), L1 ++ [I]};
+		    % queue:snoc/2
+		    ({snoc, I}, {true, Q1, L1}) ->
+			{true, queue:snoc(Q1, I), L1 ++ [I]};
+		    % queue:in_r/2, queue:cons/2
+		    ({Op, I}, {true, Q1, L1}) when Op =:= in_r; Op =:= cons ->
+			{true, queue:Op(I, Q1), [I|L1]};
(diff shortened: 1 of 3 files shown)


lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+	    equal(L1, Q1) andalso
+	    equal(L2, Q2)
+	end
+    ).
+
+prop_tail() ->
+    common_drop_tail(tail).
+
+% Test sequences of insert and retrieval operations
+prop_ops() ->
```
seriyps, line 451: can it be done with "statefull PBT"? https://propertesting.com/book_stateful_properties.html

Maria-12648430: The `prop_ops` property you mean? Maybe, but I doubt it's worth it TBH...

seriyps: Yep, `prop_ops` just looks like a classical example of a statefull test. Not sure if all the PBT frameworks use the same API for statefull tests though.

IngelaAndin:
I have put it in to testing as I am curious to see if it fills up the coverage gaps. The module has 95% coverage at the moment which is fairly good.

lib/stdlib/test/property_test/queue_prop.erl
IngelaAndin: We have some interests in trying out statefull tests in other places but I think here is probably not the use case.

Maria-12648430:
> I have put it in to testing as I am curious to see if it fills up the coverage gaps. The module has 95% coverage at the moment which is fairly good.

Short answer: It won't 😅

Long answer: The coverage gaps are caused by the omnipresent invalid arguments clauses that every exported function in the queue module has, like this one for example:
https://github.com/erlang/otp/blob/f01ea0ab3d0517fbda145a32346a6cd3310f862a/lib/stdlib/src/queue.erl#L74-L79

They are not tested in the regular ct suite, and they are also not tested in this proptest suite.

garazdawi:
> Long answer: The coverage gaps are caused by the omnipresent invalid arguments clauses that every exported function in the queue module has, like this one for example:
> https://github.com/erlang/otp/blob/f01ea0ab3d0517fbda145a32346a6cd3310f862a/lib/stdlib/src/queue.erl#L74-L79
> 
> They are not tested in the regular ct suite, and they are also not tested in this proptest suite.

Looks like another candidate to write EEP-54 style error information for :) maybe that is what I will do after #4757 is done.

IngelaAndin:
Humm interesting,  I did not note that about the code, I did not look close enough apparently. I was surprised by the coding style, I would have just let it crash with function clause, but different people think differently.  And regardless of how we achieve it newer error handling would make better error messages.  

The property based test seems to be running fine anyway in the builds.

Maria-12648430:
> Humm interesting, I did not note that about the code, I did not look close enough apparently. I was surprised by the coding style, I would have just let it crash with function clause, but different people think differently. And regardless of how we achieve it newer error handling would make better error messages.

Peculiar style I haven't seen anywhere else before or afterwards, yes. But the `queue` module is old, so maybe the style was tried there and just didn't "catch on" 🤷‍♀️

> The property based test seems to be running fine anyway in the builds.

Nice 😃

IngelaAndin:
However, we have looked a little closer at the code and we probably have some enhancements suggestions,  @HansN will provide some feedback when he gets time.  So this is postponed to after OTP-24

Maria-12648430:
> However, we have looked a little closer at the code and we probably have some enhancements suggestions, @HansN will provide some feedback when he gets time.

Always eager to learn and improve :)

> So this is postponed to after OTP-24

Sure, no hurry

IngelaAndin:
I will create an internal ticket to try to get this prioritized for the next sprint so we can have some kind of decision for OTP-25. In general, we like property tests it is just hard sometimes to find the time for all things we like to do. Sorry for letting wait so long.

Maria-12648430:
> I will create an internal ticket to try to get this prioritized for the next sprint so we can have some kind of decision for OTP-25. In general, we like property tests it is just hard sometimes to find the time for all things we like to do. Sorry for letting wait so long.

Thanks @IngelaAndin, and no worries :blush:

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+list_queue(Type) ->
+    ?LET(List, list(Type), {List, queue:from_list(List)}).
+
+queue() ->
+    queue(term()).
+
+queue(Type) ->
+    ?LET(List, list(Type), queue:from_list(List)).
+
+non_queue() ->
```
lisztspace, line 646: This generator shows the problem of the `is_queue/1` function, i.e., only by knowing about the internal representation of queues can we write a generator that we can be sure does not generate a term that can be mistaken for a queue.  This is not very abstract and goes against the documentation saying that one should treat the representation as opaque and not rely on a specification representation.

As a side note, this is also true for other modules implementing data structures, e.g., the `sets` and `array` module.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+
+prop_new() ->
+    [] =:= queue:to_list(queue:new()).
+
+prop_is_queue() ->
+    ?FORALL(
+	{IsQueue, Q},
+	oneof([
+	    {true, queue()},
+	    {false, non_queue()}
```
lisztspace, line 65: Considering the implementation of the `non_queue/0` generator, this will/might start to fail when/if the internal representation of queues change.  While this is very unlikely to happen, it does show the problem of the `is_queue/1` function, i.e., it can return false positives.
On the other hand, this is not to say that this is a bad test. The `true` part of it tests that what is generated by the `queue/0` generator is considered to be a queue, which is good.  The `false` part is more questionable since it relies on knowledge of representation.  It does, however, have sufficient value to keep it.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+prop_delete() ->
+    ?FORALL(
+	{X, {L, Q}},
+	{term(), list_queue()},
+	begin
+	    R1 = if
+		L =:= [] ->
+		    true;
+		true ->
+		    {_, Y} = hd(lists:sort([{rand:uniform(), I} || I <- L])),
```
lisztspace, line 116: This is a rather slow way of picking a random element from a list.  Why not `lists:nth(rand:uniform(length(L)), L)`?

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+	    ),
+	    equal(lists:reverse(L), Q)
+	end
+    ).
+
+%%%%%%%%%%%%%%%%%%
+%%% Generators %%%
+%%%%%%%%%%%%%%%%%%
+
+list_queue() ->
```
lisztspace, line 634: See comment on `prop_join` for a suggestion on how to improve this generator.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+	    andalso
+	    % all members of Q are members of L
+	    lists:all(
+		fun (I) ->
+		    lists:member(I, L)
+		end,
+		queue:to_list(Q)
+	    )
+	    andalso
+	    % if X is a member of L, it is also a member of Q,
```
lisztspace, line 341: Nice with the comments.  It makes it so much easier to read and understand.

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: Yes... period :sweat: As you point out, it can't be done any different way (or if so, I don't see how). The only alternative I can think of would be to only test positive cases. :thinking: Please advise?

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: I don't think that relying on the internal representation is too bad here since where we are testing the module itself, not something that _relies_ on it. If the internal representation changes, many other things are bound to change besides the tests, anyway, right?

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: Yeah... what was I thinking back then, I wonder... :sweat_smile: I'll change it as suggested.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+	    {false, non_queue()}
+	]),
+	begin
+	    IsQueue =:= queue:is_queue(Q)
+	end
+    ).
+
+prop_list_conversion() ->
+    ?FORALL(
+	List,
```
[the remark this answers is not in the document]
Maria-12648430: Ok, I'll change that, too. It makes the left sides of the code look kinda ragged, though.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+		    L;
+		{value, V} ->
+		    lists:reverse(lists:delete(V, L1))
+	    end,
+	    equal(L2, Q1)
+	end
+    ).
+
+prop_drop() ->
+    common_drop_tail(drop).
```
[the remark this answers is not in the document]
Maria-12648430: Ok, will change :wink:

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,670 @@
...
+	list_queue(),
+	begin
+	    (length(L) =:= 0) =:= queue:is_empty(Q)
+	end
+    ).
+
+prop_join() ->
+    ?FORALL(
+	{{L1, Q1}, {L2, Q2}},
+	{list_queue(), list_queue()},
```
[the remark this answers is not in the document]
Maria-12648430: Hm, ok, I'll see to that :blush: But I guess I'll have the internals of the generator create two lists (instead of splitting one), create the queue from them as you suggest, and return it together with the concatenation of the two "source" lists. Would that be ok?

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: See the comment on your comment :grin:

lib/stdlib/test/property_test/queue_prop.erl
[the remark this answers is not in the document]
Maria-12648430: Oh, wait... I got that one wrong I think :sweat_smile: _This_ is because of tabs vs spaces. There are four spaces before the `?FORALL` and a tab (which equals 8 spaces in my editor, but 4 here) before the `List`. Nevertheless, I'll change it.

lib/stdlib/test/property_test/queue_prop.erl
lisztspace: There is really no better way to do it, since the problem is with the `is_queue/1` function, for which the result is reliable only when it answers `false`.  I think it is OK to leave it as it is, but add documentation for the enlightenment of future readers, pointing out why the generator is constructed this way.

lib/stdlib/test/property_test/queue_prop.erl
lisztspace: See previous comment about the `non_queue/0` generator.  Again, adding some documentation to explain the brittleness might be good for future readers.
Also, this also shows the difference between _internal_ (concrete) tests and _external_ (abstract) tests.

lib/stdlib/test/property_test/queue_prop.erl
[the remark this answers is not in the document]
lisztspace: Even though not true for the OTP code base, always using spaces is the way to go.  We can take the extra hit in storage, I think.

lib/stdlib/test/property_test/queue_prop.erl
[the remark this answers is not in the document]
lisztspace: Yes, actual implementation does not really matter.  My main point was to generate _different_ internal representations to make the testing more complete.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -0,0 +1,684 @@
...
+                       peek_r,
+                       {snoc, term()},
+                       tail])
+            ),
+            list_queue()
+        },
+        begin
+            {Res, RQ, RL} = lists:foldl(
+                fun
+                    (_, Acc = {false, _, _}) ->
```
lisztspace, line 481: Good with a test to test a sequence of ops and not just individual operations.

For the sake of readability, I see no reason to have this as an inlined fun.  There are no free names in the fun, so it can be written as a named function instead.  This also means less indentation.
This test is about randomly creating a sequence of operations, applying them to  the `queue` implementation and to a reference implementation (model) and the comparing the results.  For the sake of decreasing complexity one could split these into different parts:
1. The simple reference implementation/model using lists.  This can be done in one function taking two args: operation as generated and the _queue as list_, returning values similar to the implementation to be tested.
2. A mapping from generated operations to `queue` operations, so applying a generated operation is simple
3. A general function for comparing the results of of operations, i.e., that returned values are the "same" and that the model and implementation correspond to each other.
I _think_ that this would be more readable than the current implementation since it does not require looking at every clause and figure out what the test is.  I will admit this is a bit thinking out loud right now and understand if it is vague.  If needed, I can take some time to flesh out some code to clarify what I mean.

lisztspace:
> There, all done I think slightly_smiling_face What do you think?

Very nice.  I did, however, have some more comments. (I was on vacation last week and had left an additional issue for later..)

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: I'm currently drawing up something that does what I think you want. But it seems to turn out in a way that doesn't really reduce the overall complexity IMO. It's not finished yet, probably tomorrow.

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: Something we _could_ do is to only rely on the internal representation in `prop_is_queue`, and let the `non_queue` generator rely on `queue:is_queue` for producing non-queues, like `?SUCHTHAT(..., not queue:is_queue(T))`. What do you think?

lisztspace: Yes, we _could_ do that, but what are you then testing?  Currently the result of the `non_queue/0` generator is only used to test the `is_queue/1` function.  Using `is_queue/1` in the generator then really doesn't test anything since we are sure of the result.
I'd rather keep it as it is and add a comment about the potential brittleness of the test.

Maria-12648430: Oh, yeah, I forgot that this is the only place where that generator is used :sweat_smile: What In meant was to put the generator as it is in place in `prop_is_queue`, and rewrite the generator to use `queue:is_queue/1` for when it is used in other properties that require non-queues. But since there are no such places, there is no point in doing that.

OTOH, we could add more tests that provide invalid arguments for each function. The `queue` implementation has guards for that on every function, eg https://github.com/erlang/otp/blob/f01ea0ab3d0517fbda145a32346a6cd3310f862a/lib/stdlib/src/queue.erl#L74-L79, something which is currently not covered by the test suite. For that, the `non_queue` generator would be useful.

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: See latest commit for my suggestion. As I said, it is IMO more complex and way more verbose than the previous implementation.

lib/stdlib/test/property_test/queue_prop.erl
lisztspace: Yes, using a generator to generate terms that are _not_ queues and testing the resilience of the functions is a good idea.  It might this be a good idea to extend the current tests as you suggest.

lib/stdlib/test/property_test/queue_prop.erl
lisztspace: I really don't agree that it is more complex.  It has the advantage of clearly separating the different parts, i.e., driver, model and implementation.  As for verbosity, possibly yes, but that isn't necessarily bad, especially when it leads to better clarity when reading.
In `2` above I suggested a mapping from generated operations to `queue` operations.  Instead of doing this with a number of clauses and `when` expressions, one can do it directly with a map, i.e.,
```
#{peek => fun queue:peek/1,
   get => fun queue:get/1,
   in => fun queue:in/2,
  ... etc
}
```
Preserve the return values and exceptions and have the list model have the same semantics and return values instead of writing scaffolding around the functions in `queue` or write a general conversation for the case when `error:empty` is raised.  You'll need some additional code to take care of the difference when operations have 1 or two arguments, but that should be quite simple.  The main point is making the mapping from generated operations to `queue` operations so simple that one does not question them and then have a simple list model for confirmation.

I do think that what you have now is better than your original implementation, but I also think that it can be made even better.

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: Ok, I added tests for invalid arguments for all functions, take a look :smile:

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: I'll have to think about this a bit and will come back to it next week  :sweat_smile:

Maria-12648430: Hm, I have been thinking about this time and again, but I can't get to a satisfactory result. I guess I'm stuck (or dense :sweat:)
To be clear, I take it that you think of running the list of generated operations with the queue and with the model, then compare the collected return values and final states of each, right?

The mapping approach for (2) is problematic since not all queue functions have the same argument order. Specifically, `snoc` is an exception because it takes the queue as first argument, while all other have it as last. That can be solved by handling that single exception specially, no biggie.

What is more problematic is that without knowing the operation (function) we're running, we don't know how to interpret the return value and what to pass into the next function.
For some functions (`get`, `get_r`, `head` etc), the return value is an item from the queue (or an error exception) and the queue itself is unchanged and what is to be passed on is the queue we already have.
For other functions (`drop`, `drop_r`, `tail` etc), the return value is the changed queue (to be passed on) itself, _unless_ there is an error exception, _then_ the queue we already have needs to be passed on.
Still others (`out` and `out_r`) return _both_ an item _and_ the updated queue, in which case we have to take it apart in order to separate the item from the queue to pass on.

In a nutshell, no matter how go about it, I have to heavily rely on the operation, which leads me back to just another implementation of what I have already. I could strip away some of the wrappings I did, yes, but it always boils back down to the same or similar set of (function or case) clauses.

lisztspace: It's just me being lazy, describing in (too) abstract terms instead of writing code that shows what I was thinking about.  Below are two functions; one for constructing the mapping and a rewritten `do_op_queue/2` that uses the mapping.
I have preserved the call semantics that you had, so it is only the expression of the mapping that is different.  By setting up some higher order functions to facilitate the construction of the map, the actual mapping can then be expressed rather compact, hopefully making it easier to understand.
As stated in the comments, it could be made simpler, but that would affect the semantics, so further changes elsewhere, i.e., in `do_op_list/2` would be needed.  To be honest, I would actually rewrite that function as well not only for the reason of making the semantics simpler, but for clarity.  I think it would be easier if grouped by operation rather than argument.  Currently, several operations are described by two clauses far from each other, so to understand, e.g., `init`, one has to find the two clauses where `init` is in the guard instead of finding one clause (or mapping) where all of `init` is described.  Yes, this is me being lazy again and describing in abstract terms :smile: .. sorry. 

```
%% Construct the mapping of an abstract operation Op | {Op, I} to a
%% function that always returns
%%  {{value, <value>} |  undefined | empty, <queue>}
%% Written this way to not rewrite do_op_list/2 (which also could be
%% simplified).
queue_map() ->
    %% Swap arguments
    Swap = fun(F) -> fun(I, Q) -> F(Q, I) end end,

    %% Affects only queue, returns no value, one or two args
    QOp1 = fun(F) -> fun(Q) -> {undefined, F(Q)} end end,
    QOp2 = fun(F) -> fun(I, Q) -> {undefined, F(I, Q)} end end,
    %% Returns value, does not affect queue.
    %% For functions that return {value, <value>} | empty
    IValue = fun(F) -> fun(Q) -> {F(Q), Q} end end,
    %% For functions thath just return the value
    Value = fun(F) -> fun(Q) -> {{value, F(Q)}, Q} end end,

    %% Catch error and map to {ErrTo, <unchanged queue>}
    %% If we mapped to {empty, ..} in both cases we could remove the
    %% second argument and have no need for both VCatch and UCatch
    Catch = fun(F, ErrTo) -> fun(Q) ->
                                     try F(Q)
                                     catch
                                         error:empty ->
                                             {ErrTo, Q}
                                     end
                             end
            end,

    VCatch = fun(F) -> Catch(F, empty) end,
    UCatch = fun(F) -> Catch(F, undefined) end,

    #{in => QOp2(fun queue:in/2),
      in_r => QOp2(fun queue:in_r/2),
      cons => QOp2(fun queue:cons/2),
      snoc => QOp2(Swap(fun queue:snoc/2)),
      out => fun queue:out/1,
      out_r => fun queue:out_r/1,
      peek => IValue(fun queue:peek/1),
      peek_r => IValue(fun queue:peek_r/1),

      get => VCatch(Value(fun queue:get/1)),
      get_r => VCatch(Value(fun queue:get_r/1)),
      head => VCatch(Value(fun queue:head/1)),
      daeh => VCatch(Value(fun queue:daeh/1)),
      last => VCatch(Value(fun queue:last/1)),

      init => UCatch(QOp1(fun queue:init/1)),
      drop => UCatch(QOp1(fun queue:drop/1)),
      drop_r => UCatch(QOp1(fun queue:drop_r/1)),
      tail => UCatch(QOp1(fun queue:tail/1)),
      liat => UCatch(QOp1(fun queue:liat/1))
     }.

do_op_queue({Op, I}, Q) ->
    F = maps:get(Op, queue_map()),
    F(I, Q);
do_op_queue(Op, Q) ->
    F = maps:get(Op, queue_map()),
    F(Q).
```

Maria-12648430: :smiley_cat: Ok, I'll try to work with that, thanks :wink:

Maria-12648430: Ok, please see my last commit, is that better? :blush: I took most of your example and made some minor changes. I'll add an explanatory comment if you say that it's to your liking :wink:

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -685,51 +680,45 @@ do_op_list(Op, [_|L1])
...
-    end.
+queue_op_map() ->
+    SwapArgs = fun (F) -> fun (X, Y) -> F(Y, X) end end,
+    QueueOp1 = fun (F) -> fun (Q) -> {undefined, F(Q)} end end,
+    QueueOp2 = fun (F) -> fun (I, Q) -> {undefined, F(I, Q)} end end,
+    WrappedReturn = fun (F) -> fun (Q) -> {F(Q), Q} end end,
+    PlainReturn = fun (F) -> fun (Q) -> {{value, F(Q)}, Q} end end,
+    CatchEmpty = fun (F) ->
+        fun (Q) ->
+            try F(Q) catch error:empty -> {empty, Q} end
```
lisztspace, line 691: This and the above is much better after changing the semantics of `do_op_list/2`.  Much better names than I propposed as well.
Minor point on formatting: in the OTP code base, about 90% of the existing anonymous are written as `fun(..) -> .. `, i.e., with no space after `fun`.

lib/stdlib/test/property_test/queue_prop.erl
```diff
@@ -636,22 +636,17 @@ do_op_list({Op, I}, L0)
        Op =:= cons->
     {undefined, [I|L0]};
 %% when the model list is empty empty, out/1, out_r/1, peek/1, peek_r/1,
-%% get/1, get_r/1, head/1, daeh/1 and last/1 return `empty' and do not
-%% change the model list
+%% get/1, get_r/1, head/1, daeh/1, init/1, last/1, drop/1, drop_r/1,
+%% tail/1 and liat/1 return `empty' and do not change the model list
 do_op_list(Op, [] = L0)
```
lisztspace, line 641: This function still suffers from the semantics of operations being spread out and can be rewritten in a similar way as `do_op_queue`.  This both adds clarity and consistency.

lib/stdlib/test/property_test/queue_prop.erl
lisztspace: Yes, this is much better!

lib/stdlib/test/property_test/queue_prop.erl
lisztspace: In the spirit of not being lazy, I include a suggested rewrite of `do_op_list/2` as well.
```

list_op_map() ->
    QueueOp1 = fun(F) -> fun(Q) -> {undefined, F(Q)} end end,
    QueueOp2 = fun(F) -> fun(I, Q) -> {undefined, F(I, Q)} end end,

    CatchEmpty = fun(F) ->
                         fun([]) -> {empty, []};
                            (L) -> F(L)
                         end
                 end,

    Wrap = fun(F) -> fun(L) -> {{value, F(L)}, L} end end,

    %% Add element to end of queue
    App = fun(I, L) -> L ++ [I] end,
    %% Add element at head of queue
    Cons = fun(I, L) -> [I|L] end,

    %% Remove and return first element of queue
    RemoveFirst = fun([H|T]) -> {{value, H}, T} end,
    %% Remove and return last element of queue
    RemoveLast = fun(L) -> {{value, lists:last(L)}, lists:droplast(L)} end,
    %% Get first element of queue, but do not remove it
    First = Wrap(fun erlang:hd/1),
    %% Get last element of queue, but do not remove it
    Last = Wrap(fun lists:last/1),

    %% Remove first element of queue without returning it
    DropFirst = QueueOp1(fun erlang:tl/1),
    %% Remove last element of queue without returning it
    DropLast = QueueOp1(fun lists:droplast/1),

    #{
      in => QueueOp2(App),
      snoc => QueueOp2(App),
      in_r => QueueOp2(Cons),
      cons => QueueOp2(Cons),

      out => CatchEmpty(RemoveFirst),
      out_r => CatchEmpty(RemoveLast),

      peek => CatchEmpty(First),
      get => CatchEmpty(First),
      head => CatchEmpty(First),

      peek_r => CatchEmpty(Last),
      get_r => CatchEmpty(Last),
      daeh => CatchEmpty(Last),
      last => CatchEmpty(Last),

      drop_r => CatchEmpty(DropLast),
      init => CatchEmpty(DropLast),
      liat => CatchEmpty(DropLast),

      drop => CatchEmpty(DropFirst),
      tail => CatchEmpty(DropFirst)
     }.

do_op_list({Op, I}, L) ->
    F = maps:get(Op, list_op_map()),
    F(I, L);
do_op_list(Op, L) ->
    F = maps:get(Op, list_op_map()),
    F(L).
```

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: Ok, I changed that in my last commit.
Personally, I prefer a more condensed style with fewer spaces. When I'm working in other code bases that use a more spacey style, I seem to over-adapt and sprinkle spaces all over the place :sweat_smile:

lib/stdlib/test/property_test/queue_prop.erl
Maria-12648430: I changed the model functions to use an operations map, like for the queue functions, so they look pretty similar now.

Maria-12648430: Oh, I noticed your suggestion only just now, ie after having it done by myself already :sweat_smile:  Thanks, though :hugs:

Maria-12648430: I just took a closer look at your suggestion, and I'm amazed how much of it happens to be the same as in mine, especially `CatchEmpty`/`MaybeEmpty` 😄

lisztspace: No surprise, really, considering the history of the conversation.

Maria-12648430:
@lisztspace well now... I think we have ironed out all the issues you originally pointed out, right? :smile_cat:

lisztspace:
> @lisztspace well now... I think we have ironed out all the issues you originally pointed out, right? smile_cat

Yes, I think so.  Do you want to squash history before we merge?

PR #4726, opened 2021-04-10
Fix typespec in gen.erl

### LINKED ISSUE #4725: start_ret() type definition does not take into account start_monitor
I think 

`-type start_ret()  :: {'ok', pid()} | 'ignore' | {'error', term()}.`

should be replaced by 

`-type start_ret()  :: {'ok', pid()} | {'ok', {pid(), reference()}} | 'ignore' | {'error', term()}.` 

in gen.erl, so dialyzer don't complain on start_monitor

Sebastian

paulo-ferraz-oliveira:
Closes #4725... as per the issue's description and analysis.

### DIFF
--- a/lib/stdlib/src/gen.erl
+++ b/lib/stdlib/src/gen.erl
@@ -46,7 +46,7 @@
                     | {'global', term()}
                     | {'via', Module :: module(), Name :: term()}.
 
--type start_ret()  :: {'ok', pid()} | 'ignore' | {'error', term()}.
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
@rickard-green, I'm still unsure where the `ignore` comes from, since I didn't go down the rabbit hole.

rickard-green:
The `init()` function of the callback module may return `ignore` which in turn also will be returned by `start*()` functions.

paulo-ferraz-oliveira:
Ah, you mean `gen_server` and family 👍, I get it. There's still some references to `gen_fsm` there, though. I could clean up in a different pull request.

PR #4727, opened 2021-04-11
Reenable tls prf test

NelsonVides:
As far as I could discover, it seems to me that in the commit
ec0f8d69a98b08b2effd07121b5f87e327d370ee, when moving around many test
suites, the prf test was left out, and it hasn't been running tests for
a while. Here I re-enable it, fixing also any other changes that seemed
necessary to make the test pass.

@IngelaAndin @peterdmv 

edit: there's a second commit, 8e19af3, removing some duplicated code as well, the prf code was copy&pasted and duplicated.

### DIFF
--- a/lib/ssl/src/ssl_gen_statem.erl
+++ b/lib/ssl/src/ssl_gen_statem.erl
@@ -774,33 +774,9 @@ handle_call({set_opts, Opts0}, From, StateName,
 handle_call(renegotiate, From, StateName, _) when StateName =/= connection ->
     {keep_state_and_data, [{reply, From, {error, already_renegotiating}}]};
 
-handle_call({prf, Secret, Label, Seed, WantedLength}, From, _,
-	    #state{connection_states = ConnectionStates,
-		   connection_env = #connection_env{negotiated_version = Version}}) ->
-    #{security_parameters := SecParams} =
-	ssl_record:current_connection_state(ConnectionStates, read),
-    #security_parameters{master_secret = MasterSecret,
-			 client_random = ClientRandom,
-			 server_random = ServerRandom,
-			 prf_algorithm = PRFAlgorithm} = SecParams,
-    Reply = try
-		SecretToUse = case Secret of
-				  _ when is_binary(Secret) -> Secret;
-				  master_secret -> MasterSecret
-			      end,
-		SeedToUse = lists:reverse(
-			      lists:foldl(fun(X, Acc) when is_binary(X) -> [X|Acc];
-					     (client_random, Acc) -> [ClientRandom|Acc];
-					     (server_random, Acc) -> [ServerRandom|Acc]
-					  end, [], Seed)),
-		ssl_handshake:prf(ssl:tls_version(Version), PRFAlgorithm, SecretToUse, Label, SeedToUse, WantedLength)
-	    catch
-		exit:_ -> {error, badarg};
-		error:Reason -> {error, Reason}
-	    end,
-    {keep_state_and_data, [{reply, From, Reply}]};
 handle_call(_,_,_,_) ->
     {keep_state_and_data, [postpone]}.
+
 handle_info({ErrorTag, Socket, econnaborted}, StateName,
 	    #state{static_env = #static_env{role = Role,
                                             host = Host,
--- a/lib/ssl/src/tls_dtls_connection.erl
+++ b/lib/ssl/src/tls_dtls_connection.erl
@@ -646,6 +646,8 @@ connection({call, From}, negotiated_protocol,
                                                  negotiated_protocol = undefined}} = State) ->
     ssl_gen_statem:hibernate_after(?FUNCTION_NAME, State,
                                        [{reply, From, {ok, SelectedProtocol}}]);
+connection({call, From}, Msg, State) when element(1, Msg) =:= prf ->
+    handle_call(Msg, From, ?FUNCTION_NAME, State);
 connection(cast, {internal_renegotiate, WriteState}, #state{static_env = #static_env{protocol_cb = tls_gen_connection},
                                                             handshake_env = HsEnv,
                                                             connection_states = ConnectionStates}
--- a/lib/ssl/test/ssl_api_SUITE.erl
+++ b/lib/ssl/test/ssl_api_SUITE.erl
@@ -247,7 +247,8 @@ since_1_2() ->
 pre_1_3() ->
     [
      default_reject_anonymous,
-     connection_information_with_srp
+     connection_information_with_srp,
+     prf
     ].
 
 gen_api_tests() ->
@@ -376,7 +377,7 @@ init_per_testcase(prf, Config) ->
          {sha512, <<145,8,98,38,243,96,42,94,163,33,53,49,241,4,127,28>>},
          %% TLS 1.0 and 1.1 PRF:
          {md5sha, <<63,136,3,217,205,123,200,177,251,211,17,229,132,4,173,80>>}],
-    TestPlan = prf_create_plan([Version], PRFS, ExpectedPrfResults),
+    TestPlan = prf_create_plan(Version, PRFS, ExpectedPrfResults),
     [{prf_test_plan, TestPlan} | Config];
 init_per_testcase(handshake_continue_tls13_client, Config) ->
     case ssl_test_lib:sufficient_crypto_support('tlsv1.3') of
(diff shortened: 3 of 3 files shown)


IngelaAndin:
When testing this I see interop problems on some of our machines that run quite old OpenSSL versions (it might be part of a test cuddling problem that lead to the test accidently being excluded in the first place). I think that the following diff could help. Would you mind including it in your PR? 

```
- diff --git a/lib/ssl/test/ssl_api_SUITE.erl b/lib/ssl/test/ssl_api_SUITE.erl
- index 15cd418d88..aabbc28890 100644
- --- a/lib/ssl/test/ssl_api_SUITE.erl
- +++ b/lib/ssl/test/ssl_api_SUITE.erl
- @@ -2665,7 +2665,7 @@ prf_ciphers_and_expected(TlsVer, PRFs, Results) ->
-      case TlsVer of
-          TlsVer when TlsVer == tlsv1
-                      orelse TlsVer == 'tlsv1.1' orelse TlsVer == 'dtlsv1' ->
- -            Ciphers = ssl:cipher_suites(default, TlsVer),
- +            Ciphers = ssl:cipher_suites(all, TlsVer),
-              Expected = [Expect#{prf := md5sha} || Expect <- Results],
-              [[{tls_ver, TlsVer}, {ciphers, Ciphers}, {expected, Expected}, {prf, md5sha}]];
-          TlsVer when  TlsVer == 'tlsv1.2' orelse  TlsVer == 'dtlsv1.2'->
- @@ -2693,10 +2693,10 @@ prf_get_ciphers(TlsVer, PRF) ->
-                                  false
-                          end
-                  end,
- -    ssl:filter_cipher_suites(ssl:cipher_suites(default, TlsVer), [{prf, PrfFilter}]).
- +    ssl:filter_cipher_suites(ssl:cipher_suites(all, TlsVer), [{prf, PrfFilter}]).
-  
-  prf_run_test(_, TlsVer, [], _, Prf) ->
- -    ct:fail({error, cipher_list_empty, TlsVer, Prf});
- +    ct:comment(lists:flatten(io_lib:format("cipher_list_empty Ver: ~p  PRF: ~p", [TlsVer, Prf])));
-  prf_run_test(Config, TlsVer, Ciphers, Expected, Prf) ->
-      {ClientNode, ServerNode, Hostname} = ssl_test_lib:run_where(Config),
-      BaseOpts = [{active, true}, {versions, [TlsVer]}, {ciphers, Ciphers}, {protocol, tls_or_dtls(TlsVer)}],
- 
```

NelsonVides:
Hi @IngelaAndin 
I've applied the patch as you mentioned, and I've tested it building against up to openssl 1.0.2, but I couldn't easily build older versions of openssl locally to test, I'll try to get that working for the future. In the meantime I hope the patch fixes the issue :)

IngelaAndin:
I will enable it in our test after OTP-24-RC3 (on Wednesday) and we will probably know at the end of the week. Thank you.

IngelaAndin:
Still not passing on those old OpenSSL versions. We will need to investigate.

NelsonVides:
> Still not passing on those old OpenSSL versions. We will need to investigate.

Can you give me a bit more detail? Which OpenSSL versions, maybe what system is running on (OS, kernel, cpu, etc)? I can give it a try as well :)

IngelaAndin:
OpenSSL 0.9.8h, OpenSSL 0.9.8k,  linux 2.6.16 and 2.6.27 

TLS-1.2 will fail with TLS client: In state hello received SERVER ALERT: Fatal - Insufficient Security
and previous versions with TLS client: In state hello received SERVER ALERT: Fatal - Handshake Failure

IngelaAndin:
Any luck @NelsonVides ? My coworker has updated cryptos configure which seems to have removed some other strange test case problems, so I will enable these tests again to see if it could be the problem here too.

IngelaAndin:
It did not help, still failing!

IngelaAndin:
Ok, so I think I found the problem, the following patch ought to fix it. Can you include it too. Note the patch is based on master.

```
index 15cd418d88..d907f5a2c8 100644
--- a/lib/ssl/test/ssl_api_SUITE.erl
+++ b/lib/ssl/test/ssl_api_SUITE.erl
@@ -2693,7 +2693,17 @@ prf_get_ciphers(TlsVer, PRF) ->
                                 false
                         end
                 end,
-    ssl:filter_cipher_suites(ssl:cipher_suites(default, TlsVer), [{prf, PrfFilter}]).
+    RSACertNoSpecialConf = fun(rsa) ->
+                                   true;
+                              (ecdhe_rsa) ->
+                                   lists:member(ecdh,crypto:supports(public_keys)); 
+                              (dhe_rsa) ->
+                                   true;
+                              (_) ->
+                                   false
+                           end,
+    ssl:filter_cipher_suites(ssl:cipher_suites(default, TlsVer), [{key_exchange, RSACertNoSpecialConf},
+                                                                  {prf, PrfFilter}]).
 
 prf_run_test(_, TlsVer, [], _, Prf) ->
     ct:fail({error, cipher_list_empty, TlsVer, Prf});

```

NelsonVides:
@IngelaAndin hi there, thanks for having a look at this, I was off for a few days and then got a big workload at the job on my return. Taking action soon 👍🏽

NelsonVides:
Right, passing the regular tests in my system, but for the love of Cthulhu I cannot build any version of OpenSSL before 1.0 to test those, still to find out why 🤷🏽‍♂️

IngelaAndin:
Well it is an ancient version no longer supported by OpenSSL, but we have some pretty old test platforms that primarily tests other things. But anyhow it is good for testcases to check preconditions and create interoperable configurations.

IngelaAndin:
Sorry if I was unclear but we need both my suggested patches.

NelsonVides:
> Sorry if I was unclear but we need both my suggested patches.

Not sure I understand you this time. So you mean rebase on master, apply the `key_exchange` filter, and make `ssl:cipher_suites/2` use `all` instead of `default`? 🤔 

If code from master is needed, can you point to me to the change (PR or so) that explained why this was an issue? I'm trying to study these cipher filters now 🙂

IngelaAndin:
Well yes, because now I see this failure again ``` {test_case_failed,{error,empty_prf_test_plan}}```
for TLS-1.2 on the old platforms, which is fixed by 

```
-  prf_run_test(_, TlsVer, [], _, Prf) ->
- -    ct:fail({error, cipher_list_empty, TlsVer, Prf});
- +    ct:comment(lists:flatten(io_lib:format("cipher_list_empty Ver: ~p  PRF: ~p", [TlsVer, Prf])));
```
The all vs default is because legacy cipher suites from TLS-1.0 and TLS-1.1 that may be needed to run this test against legacy OpenSSL servers are not part of the default as they are consider weak.

NelsonVides:
> empty_prf_test_plan

Right, the error `empty_prf_test_plan` is thrown elsewhere, by `prf_run_test` caller, at `prf/1`, that's the one that fails for empty plans. The patch you mention here is already applied and it didn't fix it, I'm fixing that one before as well 👍🏽

NelsonVides:
@IngelaAndin hi there, any news on this PR? :)

IngelaAndin:
I re-enabled it in the test now! We had public holiday in Sweden in the end of last week so not so much has happened after the release. And just before was the code freeze.

IngelaAndin:
It now passes TLS-1.2 with OpenSSL-0.9.8h but TLS-1.1 and TLS-1.0 still
fail  with:
 
```
TLS server: In state hello at tls_dtls_connection.erl:686 generated SERVER ALERT: Fatal - Handshake Failure
 - malformed_handshake_data

```

NelsonVides:
Rebasing onto maint, easy, will do later today. The OpenSSL issue, can't say how long would that take me 😔 

But about maint, that reminds me, that one of the goals I wanted to achieve originally, was to have TLS exporters for v1.3, and this prf test was something I encountered when basically trying to do that other thing, so I guess it's about time I ask you for advise about it. Is it in your release plan already? Onto which branch should I base the work? Maybe we should start a discussion about that in the erlang mailing lists?

NelsonVides:
Rebased onto maint 👍🏽

IngelaAndin:
We have no current plan to add any additional exporters (than the existing keylog exporters)  for TLS-1.3.
The are available with the  keep_secrets option enabled. Could you elaborate of what you had in mind?

IngelaAndin:
`@NelsonVides`  I think I found the last update needed for this PR.

```
prf_create_plan(TlsVer, _PRFs, Results) when TlsVer == tlsv1
		                             orelse TlsVer == 'tlsv1.1'
                                             orelse TlsVer == 'dtlsv1' ->
    Ciphers = prf_get_ciphers(TlsVer, default_prf),
    {_, Expected} = lists:keyfind(md5sha, 1, Results),
    [[{ciphers, Ciphers}, {expected, Expected}, {prf, md5sha}]];

```
If you change the function clause above in ssl_api_SUITE to look like above the tests will pass. I actually run the on one of our old machines like this and it passed. Also note my last comment if you have not seen it yet!

NelsonVides:
@IngelaAndin Applied this patch, `default_prf`, I think that makes sense, how did you realise this one? 😄 

About the exporters, my main goal was to achieve [this draft](https://datatracker.ietf.org/doc/html/draft-ietf-kitten-tls-channel-bindings-for-tls13-08), but that requires the TLS implementation to expose [exporters](https://datatracker.ietf.org/doc/html/rfc8446#section-7.5). I haven't seen the `exporter_master_secret` being calculated in the ssl libraries, so that's the first place I wanted to edit, to calculate and store that value somewhere for future usage. `keep_secrets` option, didn't notice that one, will look after it 🙂

IngelaAndin:
> @IngelaAndin Applied this patch, `default_prf`, I think that makes sense, how did you realise this one? 

I code to review the test case trying to understand why my previous suggestions fixed the test case for TLS-1.2 but not for previous versions.  

> About the exporters, my main goal was to achieve [this draft](https://datatracker.ietf.org/doc/html/draft-ietf-kitten-tls-channel-bindings-for-tls13-08), but that requires the TLS implementation to expose [exporters](https://datatracker.ietf.org/doc/html/rfc8446#section-7.5). I haven't seen the `exporter_master_secret` being calculated in the ssl libraries, so that's the first place I wanted to edit, to calculate and store that value somewhere for future usage. `keep_secrets` option, didn't notice that one, will look after it

Ah, I see. Without digging too deep I think it sounds like we should extend the prf to work for TLS-1.3. We are still working on TLS-1.3 features. It might be able to reuse some of the things done for keylogging, which has a debugging purpose and can be used with Wireshark for a system that you have control over.

PR #4730, opened 2021-04-13
Improve the 'Last calls' section included in crash dumps

michalslaski:
The **Last calls** section of a crash dump is present, if the `save_calls` process flag is set. If the process sends or receives messages, the section lists events like `send`, `'receive'` or `timeout`. This PR adds '\n' at the end of the line for such events. It helps with parsing it.

In the current version there is [a workaround](https://github.com/erlang/otp/blob/master/lib/observer/src/crashdump_viewer.erl#L1288) for the missing \n characters. After the fix it will only be needed for older versions of the crash dump file.

### DIFF
--- a/erts/emulator/beam/break.c
+++ b/erts/emulator/beam/break.c
@@ -348,11 +348,11 @@ print_process_info(fmtfn_t to, void *to_arg, Process *p, ErtsProcLocks orig_lock
 	     if (j < 0)
 		j += scb->len;
 	     if (scb->ct[j] == &exp_send)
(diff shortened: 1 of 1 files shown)


michalslaski:
On Ubuntu tests pass:
```
$ cd $ERL_TOP/lib/observer/
$ make test
[...]
The test(s) ran successfully (ct_run returned a success code)

64 Ok, 0 Failed of 64
```

garazdawi:
thanks!

PR #4734, opened 2021-04-13
Correct the `Object` description in `erlang:monitor/2` doc

michalwski:
The `Object` is always a pid when the target entity is a pid.
It doesn't matter if the pid is local or remote, the `DOWN` message will
always contain the original pid as the `Object` element.


garazdawi:
thanks!

PR #4737, opened 2021-04-14
ssl: avoid decoding certs several times

dgud:
The asn.1 decoding can be expansive avoid it by keeping both der and the
decoded cert available.

### DIFF
--- a/lib/ssl/src/ssl_certificate.erl
+++ b/lib/ssl/src/ssl_certificate.erl
@@ -63,7 +63,7 @@
 %%--------------------------------------------------------------------
 trusted_cert_and_paths([Peer],  CertDbHandle, CertDbRef, PartialChainHandler) ->
     OtpCert = public_key:pkix_decode_cert(Peer, otp),
-    Chain = [{Peer, OtpCert}],
+    Chain = [#cert{der=Peer, otp=OtpCert}],
     case public_key:pkix_is_self_signed(OtpCert) of
         true ->
             [{selfsigned_peer, [Peer]}];
@@ -76,7 +76,7 @@ trusted_cert_and_paths(Chain0,  CertDbHandle, CertDbRef, PartialChainHandler) ->
     %% If the chain contains extraneous certificates there could be
     %% more than one possible path such chains might be used to phase out
     %% an old certificate.
-    Chain = [{Der,public_key:pkix_decode_cert(Der, otp)} || Der <- Chain0],
+    Chain = [#cert{der=Der,otp=public_key:pkix_decode_cert(Der, otp)} || Der <- Chain0],
     Paths = paths(Chain, CertDbHandle),
     lists:map(fun(Path) ->
                       case handle_partial_chain(Path, PartialChainHandler, CertDbHandle, CertDbRef) of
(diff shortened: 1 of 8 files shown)


lib/ssl/src/ssl_certificate.erl
```diff
@@ -518,37 +534,39 @@ maybe_shorten_path(Path, PartialChainHandler, Default) ->
...
 
 extraneous_chains(Certs) ->
     %% If some certs claim to be the same cert that is have the same
     %% subject field we should create a list of possible chain certs
     %% for each such cert. Only one chain, if any, should be
     %% verifiable using available ROOT certs.
-    Subjects = [{subject(Cert), Cert} || Cert <- Certs],
+    Subjects = [{subject(OTP), Cert} || {_,OTP} = Cert <- Certs],
     Duplicates = find_duplicates(Subjects),
+    io:format("~P~n", [Duplicates, 20]),
```
IngelaAndin, line 569: I think you forgot to remove a debug printout?!

dgud: I think so too :-)

lib/ssl/test/ssl_cert_SUITE.erl
```diff
@@ -222,7 +222,8 @@ all_version_tests() ->
      extended_key_usage_client_auth,
      cert_expired,
      no_auth_key_identifier_ext,
-     no_auth_key_identifier_ext_keyEncipherment
+     no_auth_key_identifier_ext_keyEncipherment,
+     duplicate_chain
```
IngelaAndin, line 226: I guess you added the test earlier but forgot to list it?!

dgud: Also true, found this by strange coverage when I ran test suite.

lib/ssl/src/ssl_certificate.erl
```diff
@@ -61,20 +61,22 @@
 %% Note: Path = lists:reverse(Chain) -- Root, that is on the peer cert 
 %% always comes first in the chain but last in the path.
 %%--------------------------------------------------------------------
-trusted_cert_and_paths([Peer] = Chain,  CertDbHandle, CertDbRef, PartialChainHandler) ->
+trusted_cert_and_paths([Peer],  CertDbHandle, CertDbRef, PartialChainHandler) ->
     OtpCert = public_key:pkix_decode_cert(Peer, otp),
+    Chain = [{Peer, OtpCert}],
```
KennethL, line 66: What about using a record here like -record(cert, {der, decoded}) wouldn't that be easier to match and easier to detect errors in compile time with? Never mind the names, they could perhaps be better.

lib/ssl/src/ssl_certificate.erl
```diff
@@ -509,7 +524,8 @@ maybe_shorten_path(Path, PartialChainHandler, Default) ->
     %% public_key:pkix_path_validation by letting
     %% the user put its trust in an intermidate cert
     %% from the certifcate chain sent by the peer.
-    try PartialChainHandler(Path) of
+    DerCerts = [Der || {Der, _} <- Path],
+    try PartialChainHandler(DerCerts) of
```
KennethL, line 528: Why can't PartialChainHandler take the Path as is? Is that because it is in the public API?
Does this mean that the `PartialChainHandler` need to decode the certificates although the are already decoded?
I suppose the PartialChainHandler is provided by the user?

IngelaAndin: That would be an API change.

dgud: Yes

lib/ssl/src/ssl_certificate.erl
dgud: I thought about that and I was lazy, you get so long lines when you pick them apart, but maybe we should change
that.

KennethL: I think it can be hard to remember which part of the tuple is the decoded part and which one is the other and it could be mixed up when matching other tuples in function heads.

IngelaAndin: The principal have been to call the decode cert OtpCert as in reference to the ASN-1 type of the decode certs. Although I see that it has in some places been called ElrCert instead so it could be a good idea to clarify this in an other way.

PR #4739, opened 2021-04-15
Add the hello module as a test case

bjorng:
When developing a JIT, or a new implementation of Erlang, or a
new compiler for Erlang, one usually writes multiple small test cases
to use during development. Usually all that testing code is discarded
afterwards.

To avoid having to come up with those tests from scratch the next time
such a task is attempted, this commit adds a `hello` module that tests
most constructs that an Erlang implemenation needs to handle. To ensure
that it keeps working, we will embed it into a test suite for the
ERTS application.

### DIFF
--- a/erts/emulator/test/hello_SUITE_data/hello.erl
+++ b/erts/emulator/test/hello_SUITE_data/hello.erl
@@ -1,0 +1,2768 @@
+%% %CopyrightBegin%
+%%
+%% Copyright Ericsson AB 2021. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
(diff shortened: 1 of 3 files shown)


erts/emulator/test/hello_SUITE_data/hello.erl
```diff
@@ -0,0 +1,2762 @@
...
+
+    true = id(is_tuple(Tuple)),
+    false = id(is_list(Tuple)),
+
+    true = id(Tuple0 =:= Tuple),
+    false = id(Tuple =:= 42),
+    ok.
+
+test_catches() ->
+    thrown = (catch do_throw(id(thrown))),
```
lhoguin, line 484: I'm sure you know this so feel free to ignore me, but after https://github.com/erlang/otp/pull/4736 gets merged you can drop the parentheses (4 instances of this in the file).

bjorng: Removed the parentheses. (#4736 is not needed because it fixes pretty-printing and documentation, while the actual change to `erl_parse` was done in a previous commit.)

PR #4741, opened 2021-04-15
Add enif_dynamic_resource_call

sverker:
As a "safe" way to call NIF code in another module.

Also add `enif_init_resource_type` as a new (and hopefully the final) way to create a resource type with the new callback function `dyncall`.

### DIFF
--- a/erts/emulator/test/nif_SUITE_data/nif_mod.c
+++ b/erts/emulator/test/nif_SUITE_data/nif_mod.c
@@ -26,6 +26,9 @@
 #if ERL_NIF_MAJOR_VERSION*100 + ERL_NIF_MINOR_VERSION >= 215
 # define HAVE_ENIF_MONITOR_PROCESS
 #endif
+#if ERL_NIF_MAJOR_VERSION*100 + ERL_NIF_MINOR_VERSION >= 216
+# define HAVE_ENIF_DYNAMIC_RESOURCE_CALL
+#endif
 
 #define CHECK(X) ((void)((X) || (check_abort(__LINE__),1)))
 #ifdef __GNUC__
@@ -47,6 +50,7 @@ static ERL_NIF_TERM am_resource_type;
 static ERL_NIF_TERM am_resource_dtor_A;
 static ERL_NIF_TERM am_resource_dtor_B;
 static ERL_NIF_TERM am_resource_down_D;
+static ERL_NIF_TERM am_resource_dyncall;
 static ERL_NIF_TERM am_return;
 
 static NifModPrivData* priv_data(ErlNifEnv* env)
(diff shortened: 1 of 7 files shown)


erts/emulator/test/nif_SUITE_data/nif_mod.c
```diff
@@ -161,15 +175,31 @@ static void open_resource_type(ErlNifEnv* env, int arity, const ERL_NIF_TERM* ar
...
-            init.down = resource_down_D;
+	init.down = NULL;
+
+#  ifdef HAVE_ENIF_DYNAMIC_RESOURCE_CALL
+	init.members = 0xdead;
+	init.dyncall = (ErlNifResourceDynCall*) 0xdeadbeaf;
+
+	if (enif_is_identical(arr[6], am_resource_dyncall)) {
+            init.dyncall = resource_dyncall;
+	    init.members = 4;
```
garazdawi, line 186: Should we provide a macro that does this initialization? i.e. something 

```c
#define ERL_NIF_INIT_RESOURCE_TYPE(init) \
  memset(init,0,sizeof(ErlNifResourceTypeInit)); \
  init->members = CURRENT_MAX_MEMBERS
```

sverker: Yes maybe, but that doesn't help if you have a static initialized struct

`ErlNifResourceTypeInit my_resource_init = { my_dtor, ... };
`

garazdawi: true...

josevalim:
Is the goal here to implement some sort of callback? As a pseudo example, the inet module could say it works with different types of NIF resources and it expects them to implement certain dyncall?

RaimoNiskanen:
@josevalim: The origin of this feature is the need to get a file descriptor from the file NIF to be used in sendfile from the socket NIF, in an atomic way so to not leak it nor give it a chance to change target during the operation.

I foresee that some other NIF also might want to get a file descriptor from the file NIF, so I sketched on an API there, thinking about forward/backward compatibility.  I should/will create a PR of the ongoing sendfile work, now that I missed 24.0-rc3, so we can discuss how dynamic calls should be used...

lhoguin:
Back when I started working on `esdl2` (SDL2 NIF), I wanted to have a separate NIF for each of the SDL2 and related components (SDL2 itself, SDL_Image, SDL_ttf, OpenGL and others). This way someone that only wants SDL2 could avoid including OpenGL for example. _Bonus points if I could use the OpenGL available with `wx` instead of writing it, as well._

This required passing a resource to the other NIF and being able to extract information from it (some kind of handle, window, image and such). But it did not seem possible at the time. It sounds like this PR would allow doing exactly that.

RaimoNiskanen (commented):
Gr8!

PR #4746, opened 2021-04-19
supervisor: refine error reasons for bad combinations of options

juhlig:
This PR is a follow-up for #4638 (@garazdawi @HansN @IngelaAndin)

It is invalid to combine the child spec flag `significant => true` with the sup flag `auto_shutdown => never` or with the child spec flag `restart => permanent`. The current implementation returns `{invalid_significant, true}` in those cases, which is ambiguous because a similar reason is returned when invalid (ie, non-boolean) values are given for `significant`.

This PR changes and refines the error reasons for the aforementioned cases, as requested in #4638:
* `{bad_combination, [{auto_shutdown, never}, {significant, true}]}` for the first case
* `{bad_combination, [{restart, permanent}, {significant, true}]}` for the second case
* (if both `auto_shutdown => never` _AND_ `restart => permanent` are combined with `significant => true`, the rule for the first case applies)
* `{invalid_significant, Significant}` if a non-boolean value is given in `Significant`, as before

The tuple-list format for describing the options in question has been chosen over the other suggested possibilities (maps or strings) because it seems to be the smallest deviation from the practice overall common in the `supervisor` module, while still containing all the relevant information.

### DIFF
--- a/lib/stdlib/src/supervisor.erl
+++ b/lib/stdlib/src/supervisor.erl
@@ -1444,10 +1444,10 @@ validRestartType(temporary)   -> true;
 validRestartType(transient)   -> true;
 validRestartType(RestartType) -> throw({invalid_restart_type, RestartType}).
 
-validSignificant(true, permanent, _AutoShutdown) ->
-    throw({invalid_significant, true});
 validSignificant(true, _RestartType, never) ->
-    throw({invalid_significant, true});
+    throw({bad_combination, [{auto_shutdown, never}, {significant, true}]});
+validSignificant(true, permanent, _AutoShutdown) ->
+    throw({bad_combination, [{restart, permanent}, {significant, true}]});
 validSignificant(Significant, _RestartType, _AutoShutdown)
   when is_boolean(Significant) ->
     true;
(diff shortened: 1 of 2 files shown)


garazdawi:
looks good to me. If Ingela and Hans do not object and tests pass I'll merge next week.

HansN (approved):

IngelaAndin (approved):

garazdawi:
thanks!

KennethL:
What is so special with setting an invalid value for `significant` that it deserves to be called `invalid_significant`.
`{invalid_significant, Significant}` if a non-boolean value is given in Significant, as before
I would suggest a generic error like `{invalid_value, {significant, blahablaha}}`, I suppose the value set for `auto_shutdown` can also be invalid and can be reported in the same way.

juhlig:
> What is so special with setting an invalid value for `significant` that it deserves to be called `invalid_significant`.

Nothing is, but that is the way all other invalid settings are reported, like `invalid_restart_type`, `invalid_shutdown`, etc.

PR #4747, opened 2021-04-19
supervisor: allow 0 for shutdown

Maria-12648430:
The shutdown flag of a supervisor child spec is typed (and documented) as `shutdown()`, which in turn is `brutal_kill | timeout()`. The `timeout()` type includes `0`. However, supervisors will reject child specs with a shutdown value of `0`.

This PR is really just plain cosmetics. There probably is no use case for a shutdown value of `0`, at least I can't imagine one. But rejecting `0` violates the spec.

The relevant test cases now include a check for `0` shutdown, too. While I was at it, I remodeled the `child_specs` test case (for the tuple format) a bit to test all possible combinations of valid child spec flags, like the recent `child_specs_map` test case (for the map format) does.

### DIFF
--- a/lib/stdlib/src/supervisor.erl
+++ b/lib/stdlib/src/supervisor.erl
@@ -1455,7 +1455,7 @@ validSignificant(Significant, _RestartType, _AutoShutdown) ->
     throw({invalid_significant, Significant}).
 
 validShutdown(Shutdown)
-  when is_integer(Shutdown), Shutdown > 0 -> true;
+  when is_integer(Shutdown), Shutdown >= 0 -> true;
 validShutdown(infinity)             -> true;
 validShutdown(brutal_kill)          -> true;
 validShutdown(Shutdown)             -> throw({invalid_shutdown, Shutdown}).
--- a/lib/stdlib/test/supervisor_SUITE.erl
+++ b/lib/stdlib/test/supervisor_SUITE.erl
@@ -702,15 +702,6 @@ child_specs(Config) when is_list(Config) ->
     B5 = {child, {m,f,[a]}, permanent, 1000, worker, dy},
     B6 = {child, {m,f,[a]}, permanent, 1000, worker, [1,2,3]},
 
-    %% Correct child specs!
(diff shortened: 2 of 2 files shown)


Leeeooo19888:
https://github.com/erlang/otp/pull/4747#issue-617966383

Maria-12648430:
> [#4747 (comment)](https://github.com/erlang/otp/pull/4747#issue-617966383)

@Leeeooo19888 ... huh? ^^;;;

lhoguin (approved):
There's as much use cases for a shutdown of 0 as there is for a shutdown of 1...

Maria-12648430:
> There's as many use cases for a shutdown of 0 as there is for a shutdown of 1...

"Proof?!", the Haskeller retorted 😁

HansN:
Hi!
I have some problem with the semantics of shutdown = 0. It would mean
 "_terminate_ [the child process] _by calling exit(Child,shutdown) and then wait for an exit signal with reason shutdown back from the child process_" ... "_If no exit signal is received within the specified number of milliseconds_ [that is, 0]_, the child process is unconditionally terminated using exit(Child,kill)_".

Calling exit(C,shutdown) and the immediately an exit(C,kill) seems a bit like overkill (no pun intended).

Is there some other possible interpretation that I have missed?

lhoguin:
The problem is that the spec/documentation allows 0 but the code doesn't. There's probably no use case for having 0 there. But there's probably also no use case for having 1, 2, 3 and other very low values for the timeout, and those are allowed by the code. Perhaps instead of just allowing 0, a 0 could be converted into a 1 in order to allow the supervisor to be scheduled out (not sure this happens with 0).

Maria-12648430:
I wouldn't treat shutdown=0 any special, that is, just treat them the same as any other (integer) shutdown value, like 1. As @lhoguin  said, there is no real use case for any of those, anyway, and the change is just for the reason that the spec allows something that the code doesn't.

HansN:
Thanks for the PR!

PR #4748, opened 2021-04-20
Sendfile over NIF dyncall (OTP-17154)

RaimoNiskanen:
This is an implementation of sendfile for the socket api, that utilizes NIF dyncall in PR-4741, for public review.

The particulars about how I have used dyncall may be of interest...


erts/emulator/nifs/common/prim_file_nif.c
```diff
@@ -414,6 +419,66 @@ static void owner_death_callback(ErlNifEnv* env, void* obj, ErlNifPid* pid, ErlN
...
+
+        dc_dup->error = efile_get_handle(env, d, do_dup, &dc_dup->handle);
+
+        previous_state = erts_atomic32_cmpxchg_relb(&d->state,
+            EFILE_STATE_IDLE, EFILE_STATE_BUSY);
+
+        ASSERT(previous_state != EFILE_STATE_IDLE);
+
+        if(previous_state == EFILE_STATE_CLOSE_PENDING) {
+            /* This is the only point where a change from CLOSE_PENDING is
```
jhogberg, line 440: Nitpick: this is done in `file_handle_wrapper` too, so both comments need to be updated (there's nothing wrong with the code though).

RaimoNiskanen: Oh.  So now neither place is the only point... Right!

RaimoNiskanen: It would be nice to merge those two functions, to remove code duplication, but I did not see an elegant way.

RaimoNiskanen:
### Dyncall implementation
The idea is that dyncalls to the file nif should be somewhat backwards and forwards compatible.
The caller creates a request with backwards compatibility alternatives for older NIF implementations, and the callee processes the first request alternative it recognizes.

There is a header file prim_file_nif_dyncall.h, the question is where to place that file.
It defines operation numbers and structs.  In this case prim_file_nif_dyncall_dup.

#### Caller
The caller creates a requests and maybe alternatives in a data buffer, and expects one of them to be completed, preferably the first one; the later ones are fallbacks, in case we happen to talk to an older NIF.  The last request alternative is a zero length end marker.

The caller calls enif_dynamic_resource_call().

The structs in the data buffer is checked after the dyncall returns, to see which request alternative is completed.  That one is checked for the result.

The code for this is in prim_socket_nif.c, function esock_sendfile_start() lines 7363 through 7402.

#### Callee
The callee traverses the request buffer and processes the first alternative it recognizes, marking it as completed.  A request alternative it does not recognized is assumed to be too new and is ignored.  This may result in no change of the data buffer i.e ignoring all request alternatives

The code for this is in prim_file_nif.c functions dyncall() (and dyncall_dup())

### Overengineering
Is this overengineering or a joust approach?

lhoguin:
I am wondering in what scenario would a backward/forward compatible call be necessary? Shouldn't the other NIFs in OTP be compatible without this? Is it to allow non-OTP NIFs to use dyncalls as well?

Assuming there's a good reason to do this, this sounds like a good plan. Perhaps some macros can also be added to create the requests without having to worry about including all the alternatives.

sverker:
I can think it is a bit over engineered. If someone really needs their client NIF to be able to talk to different versions of prim_file_nif then that could be achieved by retrying.That is, try call enif_dynamic_resource_call with different operations until it succeeds (or you are out of options). And then remember the successful operation for next time.

RaimoNiskanen:
@lhoguin: People have a tendency to use OTP internal stuff, and getting a file descriptor to do funny things with sounds like an attractive target for non-OTP NIFs

erts/emulator/nifs/common/prim_file_nif_dyncall.h
```diff
@@ -0,0 +1,36 @@
...
+    size_t size;
+    enum prim_file_nif_dyncall_op op;
+    int completed;
+};
+
+struct prim_file_nif_dyncall_dup {
+    struct prim_file_nif_dyncall common;
+
+    int error;
+    ERL_NIF_TERM handle;
```
sverker, line 35: Why is the file descriptor not returned as-is. Why the binary term packaging?

RaimoNiskanen:
@sverker: Is there a theoretical possibility that the next call to the other NIF will be to a different version, due to dynamic reload.  If not, then looping over enif_dynamic_resource_call is a simpler approach that moves the problem to some future code that might as well be the code that should solve the problem.

erts/emulator/nifs/common/prim_file_nif_dyncall.h
RaimoNiskanen: I tried to find a way to declare a type for it that would work both on all Unixes and Windows, but could not figure out how to make prim_socket_nif.c, prim_file_nif_dyncall.h, unix_prim_file.c and win_prim_file.c agree on it.  If that would be feasible I would prefer such a solution.

Now it is only prim_socket_nif.c and unix_prim_file.c that needs to agree, and in both those files the type of a file descriptor is obvious.

sverker:
@RaimoNiskanen Yes, the next call to enif_dynamic_resource_call may theoretically land in an upgraded version of the NIF. But is that a problem? To be safe(er) if failing with a previously succesful cached operation you could assume upgrade has happened and fall back to the top of the retry loop again.

You will not detect a newer (and better) upgraded operation of course. Unless you invalidate the cached operation once in a while.

RaimoNiskanen:
@sverker: I guess it is really not a problem.  Each call is an atomic operation, so you will get the result, or a failure, and just get on with it.

So it would be sufficient to just have a header file with an enum set and operation structs, and then only add operation numbers and structs in the future. (?)

erts/emulator/nifs/common/prim_file_nif_dyncall.h
RaimoNiskanen: The binary packaging is just a way to handle the unknown size of the file descriptor, since it solves memory allocation and de-allocation

RaimoNiskanen:
I have simplified the dynamic call according to feedback

PR #4749, opened 2021-04-20
erl_docgen: fix ordered list rendering in HTML output

Maria-12648430:
The tag `<list type="ordered">` was rendered as an unordered (bulleted instead of numbered) list in the HTML output due to a missing XSL template, as can be seen in [the documentation for the `<list>` tag](http://erlang.org/doc/apps/erl_docgen/block_tags.html#%3Clist%3E---list) itself =^^=


lthor (approved):

lthor:
Thank you for the fix, it looked nice in the nightly build documentation so I merge it.

PR #4752, opened 2021-04-20
Add documentation and some handling about undocumented pki_asn1_type {no_asn1, new_openssh}

GuillaumeMilan:
Hello,
Here was the old way to do fully encode and decode of key genrated by `ssh-keygen`
Old way to treat complete path
```elixir
algo = :ed25519
type = :eddsa
{pub, priv} = :crypto.generate_key(type, algo)
special = :pubkey_ssh.new_openssh_encode({:ed_pri, algo, pub, priv})
 :public_key.pem_encode([{{:no_asn1, :new_openssh}, (special), :not_encrypted}])
```

New way to treat complete path
```elixir
algo = :ed25519
type = :eddsa
pki_asn1_type = {:no_asn1, :new_openssh}
{pub, priv} = :crypto.generate_key(type, algo)
special = :public_key.pem_entry_encode(pki_asn1_type, {:ed_pri, algo, pub, priv})
:public_key.pem_encode([{{:no_asn1, :new_openssh}, (special), :not_encrypted}])
```

You should have a carefull look to these changes as I am not very familiar with erlang

### DIFF
--- a/lib/public_key/src/pubkey_ssh.erl
+++ b/lib/public_key/src/pubkey_ssh.erl
@@ -78,6 +78,8 @@ decode(Bin, Type) ->
 %%--------------------------------------------------------------------
 encode(Bin, ssh2_pubkey) ->
     ssh2_pubkey_encode(Bin);
+encode(Bin, new_openssh) ->
+    new_openssh_encode(Bin);
 encode(Entries, Type) ->
     iolist_to_binary(lists:map(fun({Key, Attributes}) ->
 					      do_encode(Type, Key, Attributes)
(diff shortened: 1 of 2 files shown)


HansN:
I close this by reasons below.

You are welcome to open a new PR, but please follow the guidelines.  For example, what do you want to achive and what is the problem you want to solve?  You say "_encode and decode of key genrated by ssh-keygen_" but then you have only examples using crypto-generated keys to generate a format supported by ssh-keygen. A PR should generally also include test cases and documentation changes or additions.

The `ed_pri `(and `ed_pub`)  has a warning in the documentation that it is temporary and may change without prior notice. In fact we have discussed a change for the 24.0 release, but we don't have time for that right now, but might have in a near future.

Similarly the `pubkey_ssh:new_openssh_decode` and `pubkey_ssh:new_openssh_encode` are marked as experimental in the code and are not documented. The ` {no_asn1, new_openssh}` has a similar warning and will be removed from the 24.0 release.

In fact the whole `pubkey_ssh` module is for internal use since it is not documented, and has not been used since the 23.0 release.  The `public_key:ssh_decode/2` and the `public_key:ssh_encode/2` functions will be deprecated from the 24.0 release.

PR #4756, opened 2021-04-21
Ingela/ssl/eddsa/gh 4637/gh 4650/otp 17142

IngelaAndin:
Add EDDSA support for TLS connections


lib/crypto/src/crypto.erl
```diff
@@ -2362,6 +2362,9 @@ format_pkey(rsa, Key) ->
     map_ensure_int_as_bin(Key);
 format_pkey(ecdsa, [Key, Curve]) ->
     {nif_curve_params(Curve), ensure_int_as_bin(Key)};
+format_pkey(eddsa, [PubKey, Curve]) when  Curve == ed25519;
+                                          Curve == ed448 ->
+    [ensure_int_as_bin(PubKey), Curve];
```
KennethL, line 2367: Why is format_pkey returning a two-tuple for ecdsa and a list with 2 elements for eddsa, why not the same datatype always? And the order is different as well in the tuple Key is last and in the list Key is first?

lib/public_key/asn1/ECPrivateKey.asn1
```diff
@@ -12,13 +12,19 @@ IMPORTS
...
 
-EcpkParameters FROM PKIX1Algorithms88;
+EcpkParameters FROM PKIX1Algorithms88
+
+Attributes FROM PKCS-8;
 
 ECPrivateKey ::= SEQUENCE {
   version        INTEGER,
-  privateKey     OCTET STRING,
+  privateKey     CurvePrivateKey,
```
KennethL, line 21: Why is ECPrivateKey.asn1 changed? The changes has no impact on the code it is still exactly the same SEQUENCE.

lib/public_key/src/pubkey_cert.erl
```diff
@@ -1362,6 +1369,20 @@ public_key(#'DSAPrivateKey'{p=P, q=Q, g=G, y=Y}, _) ->
     Algo = #'PublicKeyAlgorithm'{algorithm= ?'id-dsa', 
 				 parameters={params, #'Dss-Parms'{p=P, q=Q, g=G}}},
     #'OTPSubjectPublicKeyInfo'{algorithm = Algo, subjectPublicKey = Y};
+public_key(#'ECPrivateKey'{version = _Version,
+			  privateKey = _PrivKey,
```
KennethL, line 1373: strange indentation for privateKey here and some lines below

lib/crypto/src/crypto.erl
IngelaAndin: @HansN  can you comment to this, I just added what was missing to make it work.

lib/public_key/asn1/ECPrivateKey.asn1
IngelaAndin: At least to me it feels clearer what the OCTET STRING actually represents logically, this is what newer versions of ASN-1 specs for PKIX calls it too. (Bringing our specs up to date with those is another and bigger job).

lib/public_key/src/pubkey_cert.erl
IngelaAndin: Humm .. wonder how that happened, I will fix this.

IngelaAndin:
Rearrange the commits a bit to be able to use the already existing digest type none for EDDSA  instead of undefined as it is a better API.

lib/crypto/src/crypto.erl
HansN: @KennethL asked:
> Why is format_pkey returning a two-tuple for ecdsa and a list with 2 elements for eddsa, why not the same datatype always? And the order is different as well in the tuple Key is last and in the list Key is first?

It is because the format of the Key down in the C-code is different for different algorithms.   That part of the crypto nif is a bit hairy, and would probably have been different if it had been designed from the very beginning to be extensible with by then unknown algorithms.  Now it is just a sort of most-easy-at-both-sides.

As soon as it is unpacked in C (pkey.c  static int get_pkey_private_key for example) it is quite well-structured and easy to follow. I also think that the amount of testing down in C is quite low.

What the added clause to format_pkey/2  (lines 2365-2367) adds, is that the PubKey now can be an large integer and not only an binary.  The reason is that in SSH the Key was already formatted as a list of a binary and an atom and returned untouched by the last clause of format_pkey/2, while the unpacking of stuff (certificates and such) in SSL may return a large integer.

HansN (approved):

lib/crypto/src/crypto.erl
```diff
@@ -1411,7 +1411,7 @@ verify(Algorithm, Type, Data, Signature, Key) ->
                         Result :: boolean().
 
 verify(Algorithm0, Type0, Data, Signature, Key, Options) ->
-    {Algorithm, Type} = sign_verify_compatibility(Algorithm0, Type0, Data),
+    {Algorithm, Type} = handled_digest_type(Algorithm0, Type0, Data),
```
KennethL, line 1414: I find the function name confusing, what does the function do? Can it have a better name describing what it does?

lib/public_key/src/public_key.erl
```diff
@@ -226,6 +227,18 @@ pem_entry_encode('SubjectPublicKeyInfo',
     Spki = {'SubjectPublicKeyInfo',
             {'AlgorithmIdentifier', ?'id-dsa', ParamDer}, KeyDer},
     pem_entry_encode('SubjectPublicKeyInfo', Spki);
+pem_entry_encode('SubjectPublicKeyInfo',
```
KennethL, line 230: There are 2 function clauses that are exactly the same here could have been written as one clause like this:
```erlang
pem_entry_encode('SubjectPublicKeyInfo',
	 {#'ECPoint'{point = Key}, {namedCurve, ID}}) when is_binary(Key) andalso (ID == ?'id-Ed25519' orelse ?'id-Ed488' ) ->
    Spki = {'SubjectPublicKeyInfo',
	    {'AlgorithmIdentifier', ID, asn1_NOVALUE},
	    Key},
    pem_entry_encode('SubjectPublicKeyInfo', Spki);
```

lib/public_key/src/public_key.erl
```diff
@@ -333,17 +370,37 @@ der_encode('PrivateKeyInfo', #'RSAPrivateKey'{} = PrivKey) ->
...
                  {asn1_OPENTYPE, der_encode('EcpkParameters', Parameters)}},
                 der_encode('ECPrivateKey', PrivKey#'ECPrivateKey'{parameters = asn1_NOVALUE}), 
                 asn1_NOVALUE});
-der_encode(Asn1Type, Entity) when (Asn1Type == 'PrivateKeyInfo') or
+der_encode('OneAsymmetricKey', #'ECPrivateKey'{parameters = {namedCurve, CurveOId},
+                                               privateKey = PrivKey,
+                                               attributes = Attr,
+                                               publicKey = PubKey}) ->
+    der_encode('OneAsymmetricKey',
+               {'OneAsymmetricKey', 1,
```
KennethL, line 396: Why is not the record generated from the ASN.1 spec used here? It will handle the asn1_novalue automatically and assure the right format of the tuple.

lib/public_key/src/public_key.erl
```diff
@@ -587,6 +644,12 @@ generate_key({rsa, ModulusSize, PublicExponent}) ->
                              when OthersECDHkey :: #'ECPoint'{},
                                   MyECDHkey :: #'ECPrivateKey'{},
                                   SharedSecret :: binary().
+compute_key(#'ECPoint'{point = Point}, #'ECPrivateKey'{privateKey = PrivKey,
+						       parameters = {namedCurve, Curve} = Param})
+  when (Curve == 'id-X25519') orelse
+       (Curve == 'id-X448') ->
```
KennethL, line 650: When is 'id-X448' used and when is ?'id-X448' to be used? What is the distinction? Is this really correct here , to not use ?'id-X448'  ??

lib/crypto/src/crypto.erl
IngelaAndin: The function handles the documented digest type and converts it to what crypto nif internally uses.
The argument is called DigestType in the spec and documentation. Before a conversion was only needed for backwards compatibility handling of the DSA algorithm and was symmetrical for sign and verify. For some reason it is asymmetrical what is needed for sign and verify for the EDDSA algorithm.  This however is internal to the crypto application and could perhaps be part of a future enhancement of crypto. In the API we use the already existing type of none as EDDSA uses the identity function as digest here (in math terms return the input unchanged) .

lib/public_key/src/public_key.erl
IngelaAndin: Good catch this is a mistake! I will fix it.

lib/public_key/src/public_key.erl
IngelaAndin: Which record are you referring to? I can not find a record type for  `OneAsymmetricKey`

lib/public_key/src/public_key.erl
IngelaAndin: It could! In this case we avoid a slight code duplication by adding a guard instead of,  in my opinion, an easier to read direct match. We could add a function subject_public_info(ID) function also to avoid code duplication.

lib/public_key/src/public_key.erl
KennethL: The SEQUENCE OneAssymetricKey is defined in PKCS8.asn1 , you added it in this PR. What do you do with the generates from PKCS8?

lib/crypto/src/crypto.erl
KennethL: I still think that `handled_digest_type` feels like a spelling error or a bad name.
I think it is transforming or converting or normalizing a digest type. If you like to keep handle then I propose `handle_digest_type` not **`handled`**.

lib/public_key/src/public_key.erl
IngelaAndin: Ok, I found it was confused by the different sets. I will write a ticket to make a general improvement of this type of thing in public_key. In an upcoming release.

lib/crypto/src/crypto.erl
IngelaAndin: You are right the last d was a typo, but lets call it internal_verify_alg_types instead! Hope that feels clearer to you.

KennethL: ok

lib/crypto/src/crypto.erl
```diff
@@ -1420,11 +1420,21 @@ verify(Algorithm0, Type0, Data, Signature, Key, Options) ->
...
     {sha, {digest, Digest}};
-sign_verify_compatibility(Algorithm0, Type0, _Digest) ->
+sign_compatibility(Algorithm0, Type0, _Digest) ->
     {Algorithm0, Type0}.
 
+internal_verify_alg_types(dss, none, Digest) ->
+    %% Backwards compatible
+    {sha, {digest, Digest}};
+internal_verify_alg_types(eddsa, none, _) ->
+    {eddsa, undefined};
```
HansN, line 1432: The change from 'none' to 'undefined' should not be needed afaik, because if the sign/verify algorithm name is 'eddsa' the hash type argument ( none,undefined ) not checked.

HansN: I also think that it is not necessary with two different functions for the same purpose, that is, adapting the sha and digest for old programs to the new format. And it is the same C-function that handles those parameter in the nif (pkey.c: get_pkey_sign_digest)

IngelaAndin: Humm ... I tried removing it altogether and it seems to work as you suggest @HansN . I am sure I needed it at some point during the development but it might have been intermediate. There was many steps and rebases before this started to feel like the solution.

lib/public_key/src/public_key.erl
IngelaAndin: As there was other changes that creates the needed for this PR to be run for another night, I included a commit for improving usage of generated records in public_key.

lib/public_key/src/public_key.erl
IngelaAndin: I also removed the code duplication with help of a function when fixing the records usage discussed below.

PR #4757, opened 2021-04-21
Add error info to io:format and friends

garazdawi:
This PR adds error information to `io:format`, `io:fwrite`, `io:put_chars`, `io:write` and `io:nl` and is aimed to be released in OTP-24.1.

Initially, this was triggered by `io:format(FileDev,"test",[])` trying to write to a full disk and io:format then only throws a `badarg` error. Now it throws this:

```erlang
1> proc_lib:spawn(fun() -> io:format(FullDevice,"test",[]) end).
<0.87.0>
=CRASH REPORT==== 21-Apr-2021::14:26:14.701154 ===
  crasher:
    initial call: erl_eval:'-expr/5-fun-3-'/0
....
    exception error: bad argument
      in function  io:format/3
         called as io:format(<0.86.0>,"test",[])
         *** argument 1: no space left on device (enospc)
    ancestors: [<0.84.0>]
....
```

while at it we decided to implement error info for many more of the io:format errors. For example:

```erlang
1> io:format("as",[a]).
     in function  io:format/2
        called as io:format("as",[a])
        *** argument 1: wrong number of arguments
2> io:format("~p",[]).
     in function  io:format/2
        called as io:format("~p",[])
        *** argument 1: wrong number of arguments
3> io:format("~t",[]).
     in function  io:format/2
        called as io:format("~t",[])
        *** argument 1: format string invalid (invalid control ~t)
4> io:format("~.*f",[1.0,1]).
     in function  io:format/2
        called as io:format("~.*f",[1.0,1])
        *** argument 2: element 1 must be of type integer 
                        element 2 must be of type float
5> io:format("~",[]). % "
     in function  io:format/2
        called as io:format("~",[]) % "
        *** argument 1: format string invalid (truncated)
```

This PR extends the possible values returned from a `Mod:format_error/3` callback to include `general` which is to be returned when an error happens that is not related to an argument. Thus EEP-54 will also have to be updated.

Also, it changes the formatting a bit so that multiline error info strings are correctly indented.

### DIFF
--- a/lib/stdlib/src/erl_error.erl
+++ b/lib/stdlib/src/erl_error.erl
@@ -360,9 +360,9 @@ format_arg_errors(ArgNum, [_|As], ErrorMap) ->
             [FirstLine | Lines] = string:lexemes(Err, "\r\n"),
             ArgStr = io_lib:format(<<"*** argument ~w: ">>, [ArgNum]),
             [io_lib:format(<<"~ts~ts">>, [ArgStr, FirstLine])] ++
-                [io_lib:format(<<"~*ts~ts">>,[string:length(ArgStr), "", Line]) ||
-                    Line <- Lines] ++
-             format_arg_errors(ArgNum + 1, As, ErrorMap);
+                [io_lib:format(<<"~*ts~ts">>,[string:length(ArgStr), "", Line])
+                 || Line <- Lines] ++
+                format_arg_errors(ArgNum + 1, As, ErrorMap);
         #{} ->
             format_arg_errors(ArgNum + 1, As, ErrorMap)
     end;
--- a/lib/stdlib/test/io_SUITE.erl
+++ b/lib/stdlib/test/io_SUITE.erl
@@ -33,7 +33,7 @@
 	 maps/1, coverage/1, otp_14178_unicode_atoms/1, otp_14175/1,
          otp_14285/1, limit_term/1, otp_14983/1, otp_15103/1, otp_15076/1,
          otp_15159/1, otp_15639/1, otp_15705/1, otp_15847/1, otp_15875/1,
-				 github_4801/1]).
+         error_info/1,github_4801/1]).
 
 -export([pretty/2, trf/3]).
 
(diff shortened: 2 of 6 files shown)


lib/stdlib/src/erl_error.erl
```diff
@@ -347,8 +347,13 @@ format_arg_errors(ArgNum, [_|As], ErrorMap) ->
         #{} ->
             format_arg_errors(ArgNum + 1, As, ErrorMap)
     end;
-format_arg_errors(_, _, _) ->
-    [].
+format_arg_errors(_, _, ErrorMap) ->
+    case ErrorMap of
+        #{ general := Err } ->
```
bjorng, line 359: Nitpick for consistency with the formatting in the rest of the module.

```suggestion
        #{general := Err} ->
```

lib/stdlib/test/io_SUITE.erl
```diff
@@ -2947,3 +2948,108 @@ otp_15847(_Config) ->
 otp_15875(_Config) ->
     S = io_lib:format("~tp", [[{0, [<<"00">>]}]], [{chars_limit, 18}]),
     "[{0,[<<48,...>>]}]" = lists:flatten(S).
+
+error_info(_Config) ->
+
+    FullDev =
+        fun() ->
+                {ok, Dev} = file:open("/dev/full",[read, write]),
```
bjorng, line 2956: This might only work on Linux. It certainly does not work on macOS or Windows.

I think that it would be better to create a fake IO device (as an Erlang process that implements the IO protocol). We might have code for a fake IO devices in some other test suites.

bjorng (approved):
Looks good in general, but `io_SUITE:error_info/1` needs to be fixed to work on more platforms.

lib/stdlib/test/io_SUITE.erl
garazdawi: > This might only work on Linux. It certainly does not work on macOS or Windows.

Indeed, I need to add a check for that.

> I think that it would be better to create a fake IO device (as an Erlang process that implements the IO protocol). We might have code for a fake IO devices in some other test suites.

Yes and no. If we change how file_io_server returns errors, we should have a test for that. But your idea about creating a dummy io server could help test some other faults that I was unable to test with the built-in versions.

garazdawi:
I've opened #4764 that contains the extensions to error_info needed by this PR. Then I've rebased this PR on top of #4764.

garazdawi:
Merged to maint a23364516485d7320b18c1285ac3a9df551c93b5

PR #4760, opened 2021-04-22
Windows CI build specify nuget source explicity

dominicletz:
The windows CI worker of github actions changed and does not come with the list of default sources anymore. Adding the source flag to nuget explicitly fixes this issue.


garazdawi:
thanks a bunch!

PR #4762, opened 2021-04-23
Add types nonempty_{binary,bitstring}()

### LINKED ISSUE #4636: New type nonempty_binary
Just throwing this in here to see if fruitful discussion revolves around it. Feel free to close if deemed too silly.

We get a few warnings (via `dialyzer`) stating that a given `binary()` could be `-spec(_).`ed as `<<_:8, _:_*8>>` (for example). This represents a "nonempty_binary" (not the only type, ofc). Since that name (`nonempty_binary`) is a much more representative then whatever notation `dialyzer` was able to infer, would it be good to have a type representing this in Erlang, directly?

(rationale: - close but not the same thing, ofc - we already have `nonempty_list`, `nonempty_string`)

---

We have done stuff like declaring a type named `nonempty_binary` with that definition and then using that, but it's in a shared lib. and I'd like to avoid depending on that for something as simple as a type declaration, while also not wanting to force all our lib.s to declare their own type.

Also note that even in OTP's base there's a couple of occurrences for this or even `<<_:32,_:_*8>>`, `<<_:64,_:_*8>>`.

uabboli:
Closes #4636.

### DIFF
--- a/lib/dialyzer/src/erl_types.erl
+++ b/lib/dialyzer/src/erl_types.erl
@@ -167,6 +167,8 @@
 	 t_nil/0,
 	 t_node/0,
 	 t_none/0,
+	 t_nonempty_binary/0,
+	 t_nonempty_bitstring/0,
 	 t_nonempty_list/0,
 	 t_nonempty_list/1,
 	 t_nonempty_string/0,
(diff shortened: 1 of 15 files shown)


paulo-ferraz-oliveira:
Very cool stuff. Thanks for this, @uabboli.

bjorng (approved):

PR #4764, opened 2021-04-23
Extend and document exceptions and error_info

garazdawi:
This PR does four things:

1. Adds a new type called `erlang:stacktrace/0`
2. Adds a new return key from format_error/2 called `general` that allows error information about something that is not an argument to be returned.
3. Adds a new return key from format_error/2 called `reason` that overrides the default `reason` printed by erl_error.
4. Properly aligns the strings returned by format_error/2 if they contain multiple lines.

It also attempts to document all of this in a way that hopefully makes sense.

#4757 builds on things in this PR.


erts/doc/src/erlang.xml
```diff
@@ -1791,24 +1792,34 @@ example_fun(A1, A2) ->
...
+          provide the arguments for the current function in the stack
+          back-trace. As evaluating this function causes an exception to be
+	  raised, it has no return value.</p>
+        <p>If the <c>error_info</c> option is given, the <c><anno>ErrorInfoMap</anno></c>
+          will be inserted into the stacktrace. The information given in the
+          <c><anno>ErrorInfoMap</anno></c> is used to be used by error formatters such as
+          <seemfa marker="stdlib:erl_error#format_exception/4"><c>erl_error</c></seemfa>
+          to provide more context around an error.</p>
+        <p>The default <c>module</c> of the <c><anno>ErrorInfoMap</anno></c> is the
+          module that the call to <seemfa marker="#error/3"><c>erlang:error/3</c></seemfa>
```
bjorng, line 1814:
```suggestion
module that the call to <c>error/3</c>
```

`error/3` can be called without the `erlang` prefix.

It should not be a link to itself.

system/doc/reference_manual/errors.xml
```diff
@@ -130,7 +130,11 @@
           <tag><c>error_info</c></tag>
           <item>The second element of the tuple is a map
 	  providing additional information about what caused the
-	  exception.
+	  exception. This information can be created by calling
+          <seemfa marker="erts:erlang#error/3"><c>erlang:error/3</c></seemfa>
```
bjorng, line 134:
```suggestion
<seemfa marker="erts:erlang#error/3"><c>error/3</c></seemfa>
```

`error/3` does not need any `erlang` prefix. Not sure whether the marker part should be `"erts:erlang#error/3"` or "erts:#error/3".

erts/doc/src/erlang.xml
```diff
@@ -1791,24 +1792,34 @@ example_fun(A1, A2) ->
...
-        into the stacktrace.</p>
-	<p>As evaluating this function causes an exception to be
-	raised, it has no return value.  The intent of the exception
-	class <c>error</c> is to signal that an unexpected error has
-	happened (for example, a function is called with a parameter
-	that has an incorrect type). See the guide about <seeguide
-	marker="system/reference_manual:errors">errors and error
-	handling</seeguide> for additional information.</p>
+          <c><anno>Reason</anno></c>, where <c><anno>Reason</anno></c>
+          is any term. <c><anno>Args</anno></c> is expected to be the
```
bjorng, line 1800: Could you document here that if `Args` is not a list, the arity of the calling function will be used in the stacktrace? Or perhaps we should say that if `Args` is the atom `none`, the arity of the calling function will be used?

bjorng (commented):
Overall looks good.

In the commit message for b2d3b50c158f12344682af0f6ceec335b551a974, I think you meant to say that functions without **arguments** will call `format_error/2`.

erts/doc/src/erlang.xml
garazdawi: > Or perhaps we should say that if Args is the atom none, the arity of the calling function will be used?

I don't know which way is better. I suppose that none of the ways are documented at the moment?

sverker (commented):
Looks good.

PR #4766, opened 2021-04-23
Handle when cover doesn't have any clause data

mrallen1:
When cover runs against an erlang module that implements functions for a NIF, cover doesn't have any clause data to emit in its report. Currently, the code expects clauses always exist for a module, but in this particular they don't and it causes cover to crash when run in the context of rebar3.

You can test this by cloning the repo at [erlang-h3](https://github.com/helium/erlang-h3) and running `rebar3 cover` . An error similar to

```erlang
{{badmatch,[]},
 [{cover,write_module_data,2,[{file,"cover.erl"},{line,2789}]},
  {cover,do_export,4,[{file,"cover.erl"},{line,2745}]}]}
```

(line numbers are different because this error was generated against 23.3.1 instead of against HEAD)

### DIFF
--- a/lib/tools/src/cover.erl
+++ b/lib/tools/src/cover.erl
@@ -2990,8 +2990,12 @@ merge([],ModuleList) ->
 
 write_module_data([{Module,File}|ModList],Fd) ->
     write({file,Module,File},Fd),
-    [Clauses] = ets:lookup(?COLLECTION_CLAUSE_TABLE,Module),
-    write(Clauses,Fd),
+    case ets:lookup(?COLLECTION_CLAUSE_TABLE,Module) of
+        [] ->
+	    ok;
+        [Clauses] ->
+            write(Clauses,Fd)
+    end,
     ModuleData = ets:match_object(?COLLECTION_TABLE,{#bump{module=Module},'_'}),
     do_write_module_data(ModuleData,Fd),
     write_module_data(ModList,Fd);


bjorng:
I can't reproduce your crash, and therefore I can't verify that your PR solves the problem in the correct way. Here is the end of the run of `rebar3 cover` on my computer:

```
===> Performing cover analysis...
===> No coverdata found
```

I used OTP 23.3 and my rebar3 version is:

```
rebar 3.4.1+build.3823.ref83b41523 on Erlang/OTP 23 Erts 11.2
```

Can you reproduce the problem on your computer with those versions?

bjorng:
I am closing this PR due to inactivity.

yordis:
@bjorng wouldn't be a good idea to add stale integration?

That way you don't have to be on top of inactive tickets, here an example: https://github.com/straw-hat-team/adr/blob/master/.github/stale.yml

You can add some labels to exclude it from marking as stale, example: https://github.com/straw-hat-team/adr/blob/fedfb4690cc18055433f64c5607ef4ef29a79df7/.github/stale.yml#L4-L5

mrallen1:
OK, I haven't had time to follow up on this since I opened it. Maybe it's some weird thing with my environment

bjorng:
Feel free to reopen the PR if you'll find out more about the issue.

PR #4767, opened 2021-04-24
Floats for ct multiply timetraps

tomas-abrahamsson:
For the `ct_run` option `-multiply_timetraps N`, allow `N` to be a float as well as previously an integer.  This makes it possible to scale down time traps, which can be useful for instance when running locally and the time traps are long to avoid false positives on slower hosts.

Also fix a doc typo while at it.

### DIFF
--- a/lib/common_test/test/ct_error_SUITE_data/error/test/timetrap_9_SUITE.erl
+++ b/lib/common_test/test/ct_error_SUITE_data/error/test/timetrap_9_SUITE.erl
@@ -1,0 +1,122 @@
+%%
+%% %CopyrightBegin%
+%%
+%% Copyright Ericsson AB 2021. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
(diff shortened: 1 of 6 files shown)


lib/common_test/test/ct_error_SUITE_data/error/test/timetrap_9_SUITE.erl
```diff
@@ -0,0 +1,130 @@
...
+%% Config0 = Config1 = [tuple()]
+%% Reason = term()
+%%--------------------------------------------------------------------
+init_per_testcase(tc1, Config) ->
+    ct:timetrap({seconds,1}),
+    Config;
+
+init_per_testcase(tc2, Config) ->
+    ct:timetrap(250),
+    Config;
```
u3s, line 82: This looks like unwanted copy/paste. If not needed, please remove.

tomas-abrahamsson: Oh, indeed! Fixed by removing the lines. I updated the branch by force-pushing (hoping this was the right thing to do).

u3s:
The test result are OK, I will merge it after release of OTP 24. I hope it's not a problem.

While testing I noticed, float multipliers seem to work in testspecs, as there is less parsing done and number with point makes it to code doing the timetrap math.

PR #4775, opened 2021-04-27
fix problem with unordered SCTP mode in DIAMETER

RoadRunnr:


RoadRunnr:
updated with a test in diameter_gen_sctp_SUITE and fix for long message delays in the same suite.

vkatsuba:
Hi All,

Do you have any updates about this fix? Does this fix can be included in the near future in some eg `OTP-24.0-rcX` release?

Best Regards,
--V

otp-anders:
Sorry, I haven't seen this until now. 'ordered' was a typo that I put a fix for in out nightly tests after receiving a mail about it last week, but the other commits are interesting. I have some questions though.

Regarding the semantics, they are documented in diameter_sctp(3), in the discussion of the unordered option, but where is assoc_id = 0 as "all future associations" coming from? gen_sctp(3) documents the default as 0 on an sctp_sndrcvinfo record (which is being passed to inet:setopts/2) as the "whole endpoint", so using that default value is wrong, but that would seem to apply to existing associations as well. Also, where is 1 as current association coming from? It seems to me that we should be putting the association id that's been received from gen_sctp in the unordered request.

otp-anders:
The sctp_nodelay fix is very interesting since why SCTP-related testcases have been failing sporadically has been a mystery for some time. In fact, diameter_gen_sctp_SUITE was written because of them, and if the explanation is sctp_nodelay then there's no point to that suite at all. There are also other testcases that are affected, but all of this really needs better explanation in the commit comments. I can do that once I've tested some more myself, but the delays in this case are not small at all, so it's still a bit surprising.

RoadRunnr:
So magic assoc Ids come from [RFC 6458, Sect. 7.2.](https://tools.ietf.org/html/rfc6458#section-7.2). The RFC only defined symbolic names, but so far all SCTP headers I have checked use the same mapping:

```
#define SCTP_FUTURE_ASSOC       0
#define SCTP_CURRENT_ASSOC      1
#define SCTP_ALL_ASSOC          2
```

The correct fix would be to have the inet driver map from atoms to those values. But that would require more work. So I opted for the quick fix for the moment.

The nodelay fix might be specific to localhost connections, MSS on localhost is much bigger that on remote connections. But for a message oriented protocol, disabling NAGLE might be the prefered setting anyway.

otp-anders:
Thanks for that. Have you tried just using the association id that gen_sctp communicates? That should work (if it doesn't then that's another problem :), but I don't want to hardcode any values if there's no reason to.

If you don't mind, I'll rework the PR a bit to take care of this and the other test issues I mentioned, and write some more details in commit comments.

RoadRunnr:
I tried it now and it works (on Linux).

Do you want me to update the PR or do you want to it yourself?

And of course, I don't mind if you rework the PR!

otp-anders:
Thanks, I'll do it. Like I said, I want to look through my other test code and see what sctp_nodelay actually solves. I'm still seeing failures in diameter_gen_sctp_SUITE even with the {sctp_nodelay, true} so that doesn't seem to be the whole story.

It also seems that 1 has no special meaning in gen_sctp: running your original commit gets me {error, einval} on inet:setopts/2 when 1 isn't my association id. Also, gen_sctp(3) documents ids as opaque, but gives special meaning to 0, so 1 looks to mean nothing unless that happens to be the association id gen_sctp gave you. :)

otp-anders:
I actually have another branch where I fixed the typo (before I knew about the inappropriate assoc_id) so I might just continue there. I'll get the fixes in the pipe in some fashion though. Not sure it'll make 24.0 since our code stop is today and I need at least one run though our nightly tests, but if not 24.0 then likely the first patch release. I'll get it into the next 23 patch as well.

RoadRunnr:
> It also seems that 1 has no special meaning in gen_sctp: running your original commit gets me {error, einval} on inet:setopts/2 when 1 isn't my association id. Also, gen_sctp(3) documents ids as opaque, but gives special meaning to 0, so 1 looks to mean nothing unless that happens to be the association id gen_sctp gave you. :)

As far as I can tell, Linux 5.1 seems to be the earliest version that supports the SCTP_CURRENT_ASSOC (1) and SCTP_ALL_ASSOC(2) special ids. On earlier versions using the real assoc id would indeed be the only option.

otp-anders:
Yes, that can explain what I see on my crusty corporate laptop.

otp-anders:
I've put a patch in the queue for 23.3.4, which I think will be released today, and it'll also be in 24.0, planned for Wednesday. I only included the changes to diameter_sctp itself. I still see testcases in diameter_gen_sctp_SUITE timing out even with {sctp_nodelay, true}, so I rewrote the suite a bit to have the client and server running on different hosts, and then it looks to work fine with or without Nagle. The problem may just be with the loopback address, but I'll have to look more closely after the 24 release.

otp-anders:
23.3.4 has been released. I'll close the pull request.

Issue #4774, opened 2021-04-27
got "Internal consistency check failed - please report this bug"


halturin:
Trying to build the project i'm working on and got this error.

```
===> Compiling apps/arweave/src/ar_network.erl failed
ar_network: function request/3+85:
  Internal consistency check failed - please report this bug.
  Instruction: {call_last,3,{f,444},2}
  Error:       {bad_arg_type,
                   {x,0},
                   {t_union,none,none,none,
                       [{{2,{t_atom,[get_block]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_block]}}}},
                        {{2,{t_atom,[get_block_index]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_block_index]}}}},
                        {{2,{t_atom,[get_chunk]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_chunk]}}}},
                        {{2,{t_atom,[get_sync_record]}},
                         {t_tuple,2,true,
                             #{1 => {t_atom,[get_sync_record]},2 => nil}}},
                        {{2,{t_atom,[get_tx]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_tx]}}}},
                        {{2,{t_atom,[get_wallet_list]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_wallet_list]}}}},
                        {{2,{t_atom,[get_wallet_list_chunk]}},
                         {t_tuple,2,true,
                             #{1 => {t_atom,[get_wallet_list_chunk]}}}},
                        {{2,{t_atom,[post_block]}},
                         {t_tuple,2,true,#{1 => {t_atom,[post_block]}}}},
                        {{2,{t_atom,[post_tx]}},
                         {t_tuple,2,true,#{1 => {t_atom,[post_tx]}}}}],
                       none},
                   {t_union,none,none,none,
                       [{{2,{t_atom,[get_block]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_block]}}}},
                        {{2,{t_atom,[get_block_index]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_block_index]}}}},
                        {{2,{t_atom,[get_chunk]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_chunk]}}}},
                        {{2,{t_atom,[get_tx]}},
                         {t_tuple,2,true,#{1 => {t_atom,[get_tx]}}}},
[...]

PR #4776, opened 2021-04-28
beam_validator: Support subtraction in update_container_type/4

jhogberg:
Fixes https://github.com/erlang/otp/issues/4774


jhogberg:
Thank you, I'll look into it. :-)

jhogberg:
I've decided to revert this PR for now as a deficiency in `beam_types:subtract/2` made it cause more problems than it solves. I'll reintroduce it once the latter is fixed.

PR #4777, opened 2021-04-28
Fix inserting list of terms in non-existing ETS table bug

kjellwinblad:
The macro DB_GET_TABLE was used inside a function that was yielded by
YCF. DB_GET_TABLE contains a conditional return statement which is
"hidden" from YCF. As the return statement is "hidden", YCF did not
insert the appropriate instructions before the return which sometimes
caused a core dump crash of the VM when inserting lists of incorrect
terms into a table that has been deleted.


bjorng (approved):

kjellwinblad:
Merged https://github.com/erlang/otp/commit/e62fda674fd1ccd2a5e66ce32c2687105dfee5d0

PR #4778, opened 2021-04-28
OTP-17355: socket select return value unification

RaimoNiskanen:
Oops!

I just figured out that the return values from all socket functions really would benefit from a unification of the select return value.
Example:
``` erlang
send(Socket, Data, SendRef) ->
        ok |
        {ok, {RestData, SelectInfo}} |
        {select, SelectInfo} |
        {error, Reason}
```
The two values with `SelectInfo` have tags `ok` and `select`, respectively.  This gets in the way because for both values with `SelectInfo` you want to do a receive, and you want to do it in the function that calls creates `SendRef` and calls `send/3` to get the receive optimization for the `SendRef`.

So better would be:
``` erlang
send(Socket, Data, SendRef) ->
        ok |
        {select, SelectInfo} |
        {select, {SelectInfo, RestData}} |
        {error, Reason}
```
Because then you can:
``` erlang
    SendRef = make_ref(),
    case send(Socket, Data, SendRef) of
        {select, Select} ->
            receive
                {'$socket', Socket, select, SendRef} ->
                     case Select of
                         {SelectInfo, RestData} ->
                             loop(Socket, SelectInfo, RestData);
                         SelectInfo ->
                             loop(Socket, SelectInfo, [])
                     end;
```
and hence not have to duplicate the receive statement for both the `{ok, {RestData, SelectInfo}}` and the `{select, SelectInfo}` return values.

Therefore this panic change before 24.0...

### DIFF
--- a/lib/kernel/src/gen_tcp_socket.erl
+++ b/lib/kernel/src/gen_tcp_socket.erl
@@ -1755,12 +1755,12 @@ handle_recv_peek(P, D, ActionsR, Packet) ->
                     handle_recv_peek(
                       P, D, ActionsR, Packet,
                       <<ShortData/binary, FinalData/binary>>);
-                {ok, {_, SelectInfo}} ->
+                {select, SelectInfo} ->
                     {next_state,
                      #recv{info = SelectInfo},
                      {P, D},
                      reverse(ActionsR)};
-                {select, SelectInfo} ->
+                {select, {SelectInfo, _}} ->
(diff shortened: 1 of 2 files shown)


RaimoNiskanen:
I threw this in just before the code freeze and will henceforth call all future adjustments of this branch "bug fixes"...

So this code is not entirely frozen just because we have a code freeze. ;-)

You are welcome!

PR #4783, opened 2021-04-29
Make flaky tests less flaky

kjellwinblad:
This pull-request fixes some test cases that gave false-positive results on the Erlang/OTP team's PowerPC machine with 64 cores and 32 GB memory.

### DIFF
--- a/erts/emulator/test/driver_SUITE.erl
+++ b/erts/emulator/test/driver_SUITE.erl
@@ -78,6 +78,7 @@
          otp_9302/1,
          thr_free_drv/1,
          async_blast/1,
+         thr_msg_blast/0,
          thr_msg_blast/1,
          consume_timeslice/1,
          env/1,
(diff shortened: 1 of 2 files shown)


erts/emulator/test/driver_SUITE.erl
```diff
@@ -124,7 +124,7 @@
 
 suite() ->
     [{ct_hooks,[ts_install_cth]},
-     {timetrap, {minutes, 1}}].
+     {timetrap, {minutes, 10}}].
```
garazdawi, line 127: You can increase the timeout just for that specific testcase, no need to do it for the entire suite.

garazdawi (approved):
lgtm except for a minor detail.

erts/emulator/test/driver_SUITE.erl
kjellwinblad: Thanks, fixed and merged!

PR #4785, opened 2021-04-29
erts: Hibernate system procs to save memory

garazdawi:
The code checker and literal checker sometimes use a lot of memory so their heaps will become pretty large. So we hibernate them when they have not done anything for 60 seconds in order to shrink the heap size back to 0.


erts/preloaded/src/erts_literal_area_collector.erl
```diff
@@ -81,12 +89,13 @@ msg_loop(Area, Outstnd, GcOutstnd, NeedGC) ->
 	%% Unexpected garbage message. Get rid of it...
 	_Ignore ->
 	    msg_loop(Area, Outstnd, GcOutstnd, NeedGC)
-
+    after HibernateTmo ->
+            %% We hibernate in order to clear the heap completely.
+            erlang:hibernate(?MODULE, start, [])
```
rickard-green, line 94: This one is problematic. We cannot hibernate unless `Outsntd == 0`, i.e. we have no outstanding replies. If we hibernate with outstanding replies we might release the literal area we are working on too early.

garazdawi: It will never do that as then `HibernateTmo` will be `infinity`.

rickard-green: Ahh missed that

rickard-green (approved):

PR #4787, opened 2021-04-30
Do not bind sockets created from FDs

### LINKED ISSUE #4680: gen_tcp socket based implementation do not work with FDs passed from systemd
**Describe the bug**

`gen_tcp` socket do not receive any messages when passing file descriptor as an `inet6` socket.

**To Reproduce**

[Example application can be found in `systemd` repo](https://github.com/hauleth/erlang-systemd/tree/895eecfd2c649f0f8f8735c2a1529a46537ebc2d/examples/echo).

To be able to test it out you need to be running systemd-enabled Linux distribution. You can use [Multipass](https://multipass.run) to get one (Ubuntu) if you do not have one.

1. Run application with socket activation
    ```shell
    $ systemd-socket-activate -l 8888 --fdname=tcp -E ERL_FLAGS='-kernel inet_backend socket' rebar3 shell
    ```
2. In separate shell connect to the echo service `nc localhost 8888`
3. No message is delivered to the `echo_tcp` process.

For comparing with default implementation works as expected. Steps:

1. Run application with socket activation
    ```shell
    $ systemd-socket-activate -l 8888 --fdname=tcp rebar3 shell
    ```
2. In separate shell connect to the echo service `nc localhost 8888`
3. Messages are delivered to listening process and are responded to.

**Expected behavior**

When running process with `-kernel inet_backend socket` messages should be delivered to listening process and such process should respond to the client.

**Affected versions**

I have tested on OTP 23.3, but I assume that all 23.x line is affected.

**Additional context**

This come out when I was trying to test whether `socket` based implementation of `gen_tcp` will allow fixing problem with dropping the connections when using systemd socket activation.

Ref https://github.com/hauleth/erlang-systemd/pull/24#issuecomment-807484539

The goal there is to allow full VM restarts without dropping socket connections.

hauleth:
These sockets should be already bound for us, otherwise it is pointless to pass them. In the end we are (almost) always able to create sockets on our own (while we can not be able to bind it to some values). This will check if socket is created form existing FD, and if so, it will silently skip bind, assuming that it is already bound.

Fix #4680

### DIFF
--- a/lib/kernel/src/gen_tcp_socket.erl
+++ b/lib/kernel/src/gen_tcp_socket.erl
@@ -121,32 +121,25 @@ connect_lookup(Address, Port, Opts, Timer) ->
             opts = ConnectOpts}} ->
             %%
             %% ?DBG({Domain, BindIP}),
-            BindAddr = bind_addr(Domain, BindIP, BindPort),
+            BindAddr = bind_addr(Domain, BindIP, BindPort, Fd),
+            ExtraOpts = extra_opts(Fd),
             connect_open(
-              Addrs, Domain, ConnectOpts, StartOpts, Fd, Timer, BindAddr)
+              Addrs, Domain, ConnectOpts, StartOpts, ExtraOpts,
+              Timer, BindAddr)
     catch
         throw : {ErrRef, Reason} ->
             ?badarg_exit({error, Reason})
     end.
 
-connect_open(Addrs, Domain, ConnectOpts, Opts, Fd, Timer, BindAddr) ->
-    %% ?DBG({Addrs, Domain, ConnectOpts, Opts, Fd, Timer, BindAddr}),
+connect_open(Addrs, Domain, ConnectOpts, Opts, ExtraOpts, Timer, BindAddr) ->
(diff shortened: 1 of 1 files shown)


hauleth:
Only question can be whether we shouldn't call `bind` at all, or call it, but ignore the result if there is file descriptor passed.

bmk:
You also need to check that Fd > 0 (to avoid -1).

hauleth:
@bmk changed the check to use `ExtraOpts` built above, so it will reuse the same checks instead of copying them over.

bmk:
> @bmk changed the check to use `ExtraOpts` built above, so it will reuse the same checks instead of copying them over.

Good idea.

RaimoNiskanen:
I pushed a suggestion.  This is better handled in bind_addr(), I think, and in a new helper function extra_opts().
Also, I took the opportunity to fix how ExtraOpts are massaged just before socket open into a more general and correct way since the {netns,NS} option was mishandled, from what I could see.

RaimoNiskanen:
Forgot to remove redundant code, so I force pushed

RaimoNiskanen:

PR #4791, opened 2021-05-04
Signal doc improvements

rickard-green:
Documentation improvements of [signals](https://erlang.org/~rickard/signal-doc-improvements/doc/reference_manual/processes.html#signals) and [process aliases](https://erlang.org/~rickard/signal-doc-improvements/doc/reference_manual/processes.html#process-aliases) in the reference manual.

### DIFF
--- a/system/doc/reference_manual/data_types.xml
+++ b/system/doc/reference_manual/data_types.xml
@@ -163,8 +163,15 @@ phone_number
 
   <section>
     <title>Reference</title>
-    <p>A reference is a term that is unique in an Erlang runtime
-      system, created by calling <c>make_ref/0</c>.</p>
+    <p>
+      A term that is
+      <seeguide marker="system/efficiency_guide:advanced#unique_references">unique</seeguide>
+      among connected nodes. A reference can be created by calling the
+      <seemfa marker="erts:erlang#make_ref/0"><c>make_ref/0</c></seemfa>
+      BIF. The
+      <seemfa marker="erts:erlang#is_reference/1"><c>is_reference/1</c></seemfa>
+      BIF can be used to test if a term is a reference.
+    </p>
   </section>
 
   <section>
--- a/system/doc/reference_manual/data_types.xml
+++ b/system/doc/reference_manual/data_types.xml
@@ -200,21 +200,28 @@ phone_number
   </section>
 
   <section>
-    <title>Pid</title>
-    <p>A process identifier, pid, identifies a process.</p>
-    <p>The following BIFs, which are used to create processes, return
-      values of this data type:</p>
-    <list type="bulleted">
-      <item><c>spawn/1,2,3,4</c></item>
-      <item><c>spawn_link/1,2,3,4</c></item>
-      <item><c>spawn_opt/4</c></item>
-    </list>
+    <title>PID</title>
+    <p>
+      PID is an abbreviation for process identifier. Each process has
+      a PID which identifies the process. PIDs are unique among processes
+      that are alive on connected nodes. However, a PID of a terminated
+      process may be reused as a PID for a new process after a while.
+    </p>
+    <p>
+      The BIF <seemfa marker="erts:erlang#self/0"><c>self/0</c></seemfa>
+      returns the PID of the calling process. When
+      <seeguide marker="processes#process-creation">creating a new
+      process</seeguide>, the parent process will be able to get the PID
+      of the child process either via the return value, as is the case
+      when calling the <seemfa marker="erts:erlang#spawn/3"><c>spawn/3</c></seemfa>
+      BIF, or via a message, which is the case when calling the
+      <seemfa marker="erts:erlang#spawn_request/5"><c>spawn_request/5</c></seemfa>
+      BIF. A PID is typically used when when sending a process a
+      <seeguide marker="processes#signals">signal</seeguide>. The
+      <seemfa marker="erts:erlang#is_pid/1"><c>is_pid/1</c></seemfa>
+      BIF can be used to test whether a term is a PID. 
+    </p>
     <p><em>Example:</em></p>
-    <pre>
-1> <input>spawn(m, f, []).</input>
-&lt;0.51.0></pre>
-    <p>In the following example, the BIF <c>self()</c> returns
-      the pid of the calling process:</p>
     <pre>
 -module(m).
 -export([loop/0]).
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -430,14 +430,15 @@ receive
     PatternN [when GuardSeqN] ->
         BodyN
 end</pre>
-    <p>Receives messages sent to the process using the send operator
-      (!). The patterns <c>Pattern</c> are sequentially matched
-      against the first message in time order in the mailbox, then
+    <p>
+      Fetches a received message present in the message queue of
+      the process. The patterns <c>Pattern</c> are sequentially matched
+      against the first message in time order in the message queue, then
       the second, and so on. If a match succeeds and the optional
       guard sequence <c>GuardSeq</c> is true, the corresponding
       <c>Body</c> is evaluated. The matching message is consumed, that
-      is, removed from the mailbox, while any other messages in
-      the mailbox remain unchanged.</p>
+      is, removed from the mail queue, while any other messages in
+      the mail queue remain unchanged.</p>
     <p>The return value of <c>Body</c> is the return value of
       the <c>receive</c> expression.</p>
     <p><c>receive</c> never fails. The execution is suspended, possibly
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -433,9 +433,11 @@ end</pre>
     <p>
       Fetches a received message present in the message queue of
       the process. The patterns <c>Pattern</c> are sequentially matched
-      against the first message in time order in the message queue, then
-      the second, and so on. If a match succeeds and the optional
-      guard sequence <c>GuardSeq</c> is true, the corresponding
+      against the first message in the message queue, then the second,
+      and so on. Messages in the message queue are
+      <seeguide marker="processes#message-queue-order">ordered in the
+      order they were received</seeguide>. If a match succeeds and the
+      optional guard sequence <c>GuardSeq</c> is true, the corresponding
       <c>Body</c> is evaluated. The matching message is consumed, that
       is, removed from the mail queue, while any other messages in
       the mail queue remain unchanged.</p>
--- a/system/doc/reference_manual/processes.xml
+++ b/system/doc/reference_manual/processes.xml
@@ -185,15 +185,507 @@ spawn(Module, Name, Args) -> pid()
   </section>
 
   <section>
-    <title>Message Sending</title>
-    <p>Processes communicate by sending and receiving messages.
-      Messages are sent by using
-      the <seeguide marker="expressions#send">send operator !</seeguide>
-      and received by calling
-      <seeguide marker="expressions#receive">receive</seeguide>.</p>
-    <p>Message sending is asynchronous and safe, the message is
-      guaranteed to eventually reach the recipient, provided that
-      the recipient exists.</p>
+    <title>Signals</title>
+    <p>
+      <marker id="message-sending"/>
+      All communication between Erlang processes and Erlang ports is done by
+      sending and receiving asynchronous signals. The most common signals are
+      Erlang message signals. A message signal can be sent using the
+      <seeguide marker="expressions#send">send operator <c>!</c></seeguide>.
+      A received message can be fetched from the message queue by the
+      receiving process using the
+      <seeguide marker="expressions#receive"><c>receive</c></seeguide>
+      expression.
+    </p>
+
+    <marker id="sync-comm"/>
+    <p>
+      Synchronous communication can be broken down into multiple asynchronous
+      signals. An example of such a synchronous communication is a call to the
+      <seemfa marker="erts:erlang#process_info/2"><c>erlang:process_info/2</c></seemfa>
+      BIF when the first argument does not equal the process identifier of
+      the calling process. The caller sends an asynchronous signal requesting
+      information, and then blocks waiting for the reply signal containing
+      the requested information. When the request signal reaches its
+      destination, the destination process replies with the requested
+      information.
+    </p>
+
+    <section>
+      <title>Sending Signals</title>
+
+      <p>
+	There are many signals that processes and ports use to communicate. The list below
+	contains the most important signals. In all the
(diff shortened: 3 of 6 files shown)


system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+	communicating over the distribution, signals can be lost if the
+	distribution channel goes down.
+      </p>
+
+      <p>
+	The only signal ordering guarantee given is the following: if an
+	entity sends multiple signals to the same destination entity, the
+	order is preserved; that is, if <c>A</c> sends a signal <c>S1</c> to
+	<c>B</c>, and later sends signal <c>S2</c> to <c>B</c>, <c>S1</c> is
+	guaranteed not to arrive after <c>S2</c>.
```
garazdawi, line 606: Should we also mention that in the scenario above, it is not guaranteed that S1 will arrive before S2? that is that S1 can be lost?

system/doc/reference_manual/expressions.xml
```diff
@@ -430,14 +430,15 @@ receive
     PatternN [when GuardSeqN] ->
         BodyN
 end</pre>
-    <p>Receives messages sent to the process using the send operator
-      (!). The patterns <c>Pattern</c> are sequentially matched
-      against the first message in time order in the mailbox, then
+    <p>
+      Fetches a received message present in the message queue of
+      the process. The patterns <c>Pattern</c> are sequentially matched
+      against the first message in time order in the message queue, then
```
garazdawi, line 436: Maybe add a link from here to the signal order guarantees?

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+	  </item>
+	  <item>
+	    <seemfa marker="erts:erlang#link/1"><c>erlang:link/1</c></seemfa>
+	  </item>
+	  <item>
+	    <seemfa marker="erts:erlang#group_leader/2"><c>erlang:group_leader/2</c></seemfa>
+	  </item>
+	</list>
+	<p>
+	  This has not been changed since it has been present so long.
```
garazdawi, line 655:
```suggestion
	  The irregularities mentioned above cannot be fixed as they have been part of Erlang too long and it would break a lot of existing code.
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+	</p>
+      </section>
+      <section>
+	<title>Blocking signaling over distribution</title>
+	<p>
+	  When sending a signal over a distribution channel, the sending
+	  process may unfortunately be suspended even though the signal is
```
garazdawi, line 662:
```suggestion
	  process may be suspended even though the signal is
```

garazdawi (approved):

system/doc/reference_manual/data_types.xml
```diff
@@ -193,22 +200,29 @@ phone_number
...
-      <item><c>spawn/1,2,3,4</c></item>
-      <item><c>spawn_link/1,2,3,4</c></item>
-      <item><c>spawn_opt/4</c></item>
-    </list>
+    <title>PID</title>
+    <p>
+      PID is an abbreviation for process identifier. Each process has
+      a PID which identifies the process. PIDs are unique among processes
+      that are alive on connected nodes. A PID of a terminated process may
+      however be reused as a PID for a new process after a while.
```
bjorng, line 208:
```suggestion
      that are alive on connected nodes. However, a PID of a terminated process may
      be reused as a PID for a new process after a while.
```

system/doc/reference_manual/data_types.xml
```diff
@@ -193,22 +200,29 @@ phone_number
...
+    </p>
+    <p>
+      The BIF <seemfa marker="erts:erlang#self/0"><c>self/0</c></seemfa>
+      returns the PID of the calling process. When
+      <seeguide marker="processes#process-creation">creating a new
+      process</seeguide>, the parent process will be able to get the PID
+      of the child process either via the return value as is the case
+      when calling the <seemfa marker="erts:erlang#spawn/3"><c>spawn/3</c></seemfa>
+      BIF or via a message which is the case when calling the
```
bjorng, line 217:
```suggestion
      of the child process either via the return value, as is the case
      when calling the <seemfa marker="erts:erlang#spawn/3"><c>spawn/3</c></seemfa>
      BIF, or via a message, which is the case when calling the
```

system/doc/reference_manual/data_types.xml
```diff
@@ -193,22 +200,29 @@ phone_number
...
+      <seemfa marker="erts:erlang#spawn_request/5"><c>spawn_request/5</c></seemfa>
+      BIF. A PID is typically used when when sending a process a
+      <seeguide marker="processes#signals">signal</seeguide>. The
+      <seemfa marker="erts:erlang#is_pid/1"><c>is_pid/1</c></seemfa>
+      BIF can be used to test if a term is a PID.
```
bjorng, line 222:
```suggestion
      BIF can be used to test whether a term is a PID. 
```

system/doc/reference_manual/processes.xml
```diff
@@ -83,6 +89,76 @@ spawn(Module, Name, Args) -> pid()
...
+      <item>create an alias identifying another process than the caller.</item>
+      <item>deactivate an alias unless it identifies the caller.</item>
+      <item>look up an alias.</item>
+      <item>look up the process identified by an alias.</item>
+      <item>check if an alias is active or not.</item>
+      <item>check if a reference is an alias.</item>
+    </list>
+    <p>
+      These are all intentional design decisions relating to
+      performance, scalability and distribution transparency.
```
bjorng, line 158:
```suggestion
      performance, scalability, and distribution transparency.
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+      receiving process using the
+      <seeguide marker="expressions#receive"><c>receive</c></seeguide>
+      expression.
+    </p>
+    <section>
+      <title>Sending Signals</title>
+
+      <p>
+	Besides message signals there are a lot of other signals that can be
+	sent. The set of signals used may vary. A signal may be replaced by
```
bjorng, line 205: "Vary" in which way? Over time? Between releases? Something else? I think you should clarify or remove this sentence.

"may be replaced by": Be replaced by whom? By the runtime system? The user?

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+	another type of signal as long as it does not change the API. The
+	important piece of information here is that <em>all</em> communication
+	between Erlang processes and Erlang ports is done via asynchronous
+	signaling of some kind, and in a lot of cases the specific type of
```
bjorng, line 209:
```suggestion
	signaling of some kind, and in many cases the specific type of
```

bjorng, line 209:
```suggestion
	signaling of some kind, and in many cases the specific type of
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+	  BIF.
+	</item>
+
+	<tag><c>gc_request</c></tag>
+	<item>
+	  Sent to a process when requesting a garbage collection of another
+	  process via the
+	  <seemfa marker="erts:erlang#garbage_collect/1">garbage_collect/1,2</seemfa>
+	  BIFs. This signal is also sent to the caller itself when requesting
+	  the garbage collection to be performed asynchronously when using
```
bjorng, line 329:
```suggestion
	  the garbage collection to be performed asynchronously using
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+	<seemfa marker="erts:erlang#send_nosuspend/2"><c>erlang:send_nosuspend/2,3</c></seemfa>
+	BIFs instead of using the
+	<seeguide marker="expressions#send">send operator <c>!</c></seeguide>.
+      </p>
+
+      <p><marker id="runtime-service"/>
+	The clock service, the timer service, and the spawn service mentioned
+	above are services provided by the runtime system. Each of these
+	services consists of multiple independently executing entities. Such a
+	service can be viewed as a group of processes, and actually could be
```
bjorng, line 436:
```suggestion
	service can be viewed as a group of processes, and could actually be
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,508 @@ spawn(Module, Name, Args) -> pid()
...
+      </p>
+      <section>
+	<title>The Excluded Resources</title>
+	<p>
+	  The process identifier of the process cannot be released for
+	  reuse until everything regarding the process has been released.
+	</p>
+	<p>
+	  A process executing dirty native code in a NIF when it receives
+	  an exit signal will be set into an exiting state even though it is
```
bjorng, line 561:
```suggestion
	  an exit signal will be set into an exiting state even if it is
```

system/doc/reference_manual/processes.xml
rickard-green: Yes, this should be clarified.

This was what I was thinking about when I wrote this:
* A signal may be replaced by another type of signal in a patch. For example, a reply signal as a response to a `gc_request` signal is currently always a message signal, but can in the synchronous case be replaced by a non-message signal specially crafted for this reply. Such a replacement if fully compatible and can be introduced in any patch.
* A signal type may change at runtime depending on the state of the system. For example, a reply signal to an `alive_request` may currently be a message signal in some cases, but in other cases not, depending on the states of the involved processes.

However, the signals are are still `gc_reply` and `alive_reply` regardless of whether they are realized using message signals or not. I'll do some rewriting here...

system/doc/reference_manual/processes.xml
rickard-green: I think the meaning of this sentence changes to something it is not supposed to say when removing "when". What about:

This signal is also sent to the caller itself when it requests the garbage collection to be performed asynchronously by passing the `{async, RequestId}` option to `garbage_collect/2`.

rickard-green:
@bjorng & @garazdawi all comments should have been addressed now. Not always exactly as suggested, but similar. The "sending signals" part also got changed quite a bit.

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+      destination, the destination process replies with the requested
+      information.
+    </p>
+
+    <section>
+      <title>Sending Signals</title>
+
+      <p>
+	Besides message signals there are a lot of other signals that can be
+	sent. Below follows a list of the most important signals. In all the
```
garazdawi, line 219:
```suggestion
	There are many signals that processes and ports use to communicate. The list below
	contains the most important signals. In all the
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+	cases below of request/reply signal pairs, the request signal is sent
```
garazdawi, line 220:
```suggestion
	cases of request/reply signal pairs, the request signal is sent
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+	that this does <em>not</em> violate the
+	<seeguide marker="#signal-delivery">signal ordering
+	guarantee</seeguide> of the language.
+      </p>
+
+      <p>
+	The realization of the signals described above may change both at
+	runtime and due to changes in implementation. You may be able to
+	detect such changes using <c>receive</c> tracing or by inspecting
+	message queues. However, this is internal implementation details of
```
garazdawi, line 404:
```suggestion
	message queues. However, these are internal implementation details of
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+	the runtime system that you should <em>not</em> rely on. As an example,
+	a large part of the reply signals above are ordinary message signals.
```
garazdawi, line 406:
```suggestion
	many of the reply signals above are ordinary message signals.
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+	When the operation is synchronous, the reply signals do not have to be
+	message signals. The current implementation also take advantage of this
```
garazdawi, line 408:
```suggestion
	message signals. The current implementation takes advantage of this
```

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+
+	<tag><c>demonitor</c></tag>
+	<item>
+	  Sent when calling one of the
+	  <seemfa marker="erts:erlang#demonitor/1">demonitor/1,2</seemfa>
+	  BIFs, or when a process monitoring another process terminates.
+	</item>
+
+	<tag><c>down</c></tag>
+	<item>
```
garazdawi, line 274: Do we have any guarantee about the order of `down` signal vs `exit`?

system/doc/reference_manual/processes.xml
```diff
@@ -109,23 +185,503 @@ spawn(Module, Name, Args) -> pid()
...
+	    <item>
+	      <seemfa marker="erts:erlang#link/1"><c>erlang:link/1</c></seemfa>
+	    </item>
+	    <item>
+	      <seemfa marker="erts:erlang#group_leader/2"><c>erlang:group_leader/2</c></seemfa>
+	    </item>
+	  </list>
+	  <p></p>
+	</item>
+	<tag>Unexpected Behaviour of Exit Signal With Exit Reason <c>normal</c></tag>
```
garazdawi, line 637: Should we mention this scenario:

```erlang
Eshell V11.2.1  (abort with ^G)
1> process_flag(trap_exit, true).
false
2> spawn_link(fun() -> exit(kill) end).
<0.87.0>
3> flush().
Shell got {'EXIT',<0.87.0>,kill}
ok
```

i.e. that when a link breaks with reason `kill` it is not the same as calling `erlang:exit(Pid,kill)`?

system/doc/reference_manual/processes.xml
rickard-green: No

PR #4793, opened 2021-05-04
Adjust refc binary growth

garazdawi:
When a refc binary is very large, it does not make sense to double the size of it anymore when preallocating for more writes. So when it is larger than 64 MB we just increase it by 10%.

@sverker @kjellwinblad what do you think about these limits? I just pulled these ones out guessing what might be reasonable. The Erlang heap grows with 20%, maybe should we do the same for binaries?

### DIFF
--- a/erts/emulator/beam/erl_bits.c
+++ b/erts/emulator/beam/erl_bits.c
@@ -78,6 +78,11 @@ typedef Uint16 erlfp16;
 
 static byte get_bit(byte b, size_t a_offs); 
 
+/* If the proc bin is larger than 16 MB,
+   we only increase by 20% instead of doubling */
+#define GROW_PROC_BIN_SIZE(size) \
+    (((size) > (1ull << 24)) ? 1.2*(size) : 2*(size))
+
 /* the state resides in the current process' scheduler data */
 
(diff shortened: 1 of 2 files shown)


erts/emulator/beam/erl_bits.c
```diff
@@ -78,6 +78,11 @@ typedef Uint16 erlfp16;
 
 static byte get_bit(byte b, size_t a_offs); 
 
+/* If the proc bin is larger than 64 MB,
+   we only increase by 10% instead of doubling */
+#define NEW_PROC_BIN_SIZE(size) \
+    ((size) > (1ull << 26)) ? 1.1*(size) : 2*(size)
+
```
sverker, line 85: Best to put parentheses around the entire macro.

and naming is hard... "NEW" gives me the wrong associations, may I suggest something like GROW_PROC_BIN_SIZE.

To me 64MB sounds a bit large. What about 8MB and 20%

zuiderkwast: ... or PROC_BIN_ALLOC_SIZE or PROC_BIN_CAPACITY?

garazdawi: I did 16MB and 20% as a compromise :)

Issue #4796, opened 2021-05-04
Cover compilation does not work with `+tuple_calls` in OTP 22


jesperes:
**Describe the bug**
On OTP 22, cover-compilation does not work on parameterized modules, presumably because the `tuple_calls` compiler option is now mandatory, but removed by `cover:filter_options/1`.

**To Reproduce**

```shell
#!/bin/bash

cat >cover_pmod.erl <<EOF
-module(cover_pmod, [GlobalVar]).

-include("pmod.hrl").

-export([main/1]).

main(Arg) ->
  GlobalVar + Arg.
EOF

cat >cover_pmod_test.escript <<EOF
#!/usr/bin/env escript

-mode(compile).

-export([main/1]).

main(_) ->
  {ok, _} = cover:compile_beam("cover_pmod.beam"),
  Instance = cover_pmod:new(1),
  io:format("Parameterized module: ~p~n", [Instance]),
  3 = Instance:main(2),
  erlang:halt(0).
EOF

# insert path to pmod_transform application here:
PMOD_TRANSFORM=[...]/pmod_transform
cp "$PMOD_TRANSFORM"/src/pmod_pt.erl .
cp "$PMOD_TRANSFORM"/include/pmod.hrl .

erlc pmod_pt.erl
erlc -pa . +tuple_calls +debug_info cover_pmod.erl

chmod +x cover_pmod_test.escript
if ./cover_pmod_test.escript; then
    echo ok
else
    echo "fail!"
fi
```

**Expected behavior**
Tuple calls should work in cover-compiled code, and the above shell script should print `ok`.

**Affected versions**
OTP 21

**Additional context**
Add any other context about the problem here. If you wish to attach Erlang code you can either write it directly in the post using [code tags](https://guides.github.com/features/mastering-markdown/), create a [gist](https://gist.github.com), or attach it as a zip file to this post.

mikpe:
Broken in OTP-21 and OTP-23 too, works with OTP-20.

mikpe:
Regressed with f2b332186a Make tuple calls opt-in, but that just exposed the issue that cover strips now-required compiler options.

mikpe:
Here's a simpler self-contained test case that doesn't depend on any parse transform.

```
> cat bug4796.erl
-module(bug4796).
-export([test/0, foo/1]).

test() ->
  PMod = new(42),
  PMod:foo().

new(X) ->
  {?MODULE, X}.

foo({?MODULE, 42}) -> ok.
```
[...]

PR #4797, opened 2021-05-04
cover:filter_options/1: preserve tuple_calls

mikpe:
Fixes https://github.com/erlang/otp/issues/4796

### DIFF
--- a/lib/tools/src/cover.erl
+++ b/lib/tools/src/cover.erl
@@ -335,6 +335,7 @@ filter_options(Options) ->
                              {d, _Macro} -> true;
                              {d, _Macro, _Value} -> true;
                              export_all -> true;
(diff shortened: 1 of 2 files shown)


bjorng:
Could you add a test case?

mikpe:
> Could you add a test case?

I'll give it a try.

mikpe:
Test case added, verified to fail without the fix and pass with it. Squashed. Had to rebase due to a merge conflict.

bjorng:

> Question: how do you want references to issues on github to look? "otp_NNNNN" are for your OTP-private bugs, "erl_NNNN" for the old non-github bug tracker, so I chose "GH-NNNN" as a general reference and "gh_NNNN" for function/module names.

Yes, that is the convention that we use.

PR #4806, opened 2021-05-06
Fix doc for ets:select_count/2

Maria-12648430:
`select_count/2` is described as `match_delete/2` without deleting, but should be `select_delete/2`.


sverker:
Thanks for the correction, but someone already fixed the same typo on master branch for OTP-24.0.

PR #4807, opened 2021-05-06
Improve receive docs

sverker:
Clarify message matching order; first all patterns then next message
in queue.

Rename "mailbox" and "mail queue" to message queue.

receive after limit is almost 50 days.

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -432,15 +432,17 @@ receive
 end</pre>
     <p>
       Fetches a received message present in the message queue of
-      the process. The patterns <c>Pattern</c> are sequentially matched
-      against the first message in the message queue, then the second,
-      and so on. Messages in the message queue are
-      <seeguide marker="processes#message-queue-order">ordered in the
-      order they were received</seeguide>. If a match succeeds and the
-      optional guard sequence <c>GuardSeq</c> is true, the corresponding
-      <c>Body</c> is evaluated. The matching message is consumed, that
-      is, removed from the mail queue, while any other messages in
-      the mail queue remain unchanged.</p>
+      the process. The first message in the message queue is matched
+      sequentially against the patterns from top to bottom. If no
+      match was found, the matching sequence is repeated for the second
+      message in the queue, and so on. Messages are queued in the
+      <seeguide marker="processes#message-queue-order">order they were
+      received</seeguide>. If a match succeeds, that
+      is, if the <c>Pattern</c> matches and the optional guard sequence
+      <c>GuardSeq</c> is true, then the message is removed from the message
+      queue and the corresponding <c>Body</c> is evaluated. All other
+      messages in the message queue remain unchanged.
+    </p>
     <p>The return value of <c>Body</c> is the return value of
       the <c>receive</c> expression.</p>
     <p><c>receive</c> never fails. The execution is suspended, possibly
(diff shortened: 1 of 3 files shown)


system/doc/reference_manual/expressions.xml
```diff
@@ -432,15 +432,17 @@ receive
 end</pre>
     <p>
       Fetches a received message present in the message queue of
-      the process. The patterns <c>Pattern</c> are sequentially matched
-      against the first message in the message queue, then the second,
-      and so on. Messages in the message queue are
+      the process. The first message in the message queue is matched
+      sequentially against the patterns from top to bottom. If no
+      match was found the matching sequence is repeated for the second
```
rickard-green, line 437:
```suggestion
      match was found, the matching sequence is repeated for the second
```

system/doc/reference_manual/expressions.xml
```diff
@@ -470,13 +472,15 @@ after
...
-    <p><c>ExprT</c> is to evaluate to an integer. The highest allowed
-      value is 16#FFFFFFFF, that is, the value must fit in 32 bits.
-      <c>receive..after</c> works exactly as <c>receive</c>, except
+    <p><c>receive..after</c> works exactly as <c>receive</c>, except
       that if no matching message has arrived within <c>ExprT</c>
       milliseconds, then <c>BodyT</c> is evaluated instead. The
       return value of <c>BodyT</c> then becomes the return value
-      of the <c>receive..after</c> expression.</p>
+      of the <c>receive..after</c> expression. <c>ExprT</c> is to
+      evaluate to an integer. The highest allowed value is
```
rickard-green, line 480:
```suggestion
      evaluate to an integer, or the atom <c>infinity</c>. If it evaluates to an integer, the highest allowed value is
```

rickard-green (approved):

system/doc/reference_manual/expressions.xml
rickard-green: Hmm, that's weird. I think it is quite confusing to first claim that it should evaluate to an integer and then half a page below add to it. It should probably also describe the allowed range of integers, `i.e.` 0..4294967295`, not just mention the largest value allowed.

sverker:
@rickard-green Added a paragraph about the "asynchronousness" of signal reception.

system/doc/reference_manual/expressions.xml
```diff
@@ -432,15 +432,17 @@ receive
...
     <p>
       Fetches a received message present in the message queue of
-      the process. The patterns <c>Pattern</c> are sequentially matched
-      against the first message in the message queue, then the second,
-      and so on. Messages in the message queue are
+      the process. The first message in the message queue is matched
+      sequentially against the patterns from top to bottom. If no
+      match was found, the matching sequence is repeated for the second
+      message in the queue, and so on. Messages in the message queue are
       <seeguide marker="processes#message-queue-order">ordered in the
```
paulo-ferraz-oliveira, line 439: Would `delivered` be better than `ordered`, here?

sverker: I don't know. The text talks about the message queue as passive and the receive expression as active, how it matches and picks messages from the queue. So I think "delivered" is a bit confusing; from who to who?

Is it the "ordered in the order" phrasing that you don't like?

paulo-ferraz-oliveira: > Is it the "ordered in the order" phrasing that you don't like?

Yes, it's odd :) but I get your point.

paulo-ferraz-oliveira: Alternatives:
* "ordered as they were received"
* "kept in the order they were received"
* (longer form) "Messages are queued in the order they were received"

PR #4811, opened 2021-05-07
timer: modernization and improvements

Maria-12648430:
This started as a minor improvement, and gradually grew into, well, _this_... ^^; The `timer` module seems to be getting out of fashion these days, having a reputation for getting clogged when there are many timers. And no wonder, the most recent changes being 6 years ago, but most of it more than 10 years old. But the functionality is actually pretty handy (and I like old stuff ;)), so I went to dust it off a bit.

This is what has been done:
* All validations have been moved into the exported accessor functions, meaning they happen in the calling process and don't bother the timer server.
* An `apply_after` call (and thus, other functions using it, like `send_after`) do the apply (call `do_apply`) immediately without going through the timer server if the timeout is `0`.
* ~The `ensure_started/0` function, to start the timer server dynamically when needed, has been removed. It checked if `timer_server` was registered, on every call, and started it if not. Now, a call is always tried, and if it results in an exit exception with reason `{noproc, _}`, the timer server is started.~
* The backing `ets` tables are not named any more, instead their `tid`s are carried in a state field.
* ~The backing `ets` tables were made `public` so they can be accessed from other processes to enable parallelism.~
* Where possible, some of the newer `ets` functions are used, like `ets:take/2` to replace lookup-and-delete.
* ~The processing of the timers is now done in a spawned process.~
* ~The cancellation of timers is also done in spawned processes.~
* Interval timers are not linked to processes any more, but instead they are monitored. Linking, in the old implementation, had two downsides: processes that were linked once were never unlinked even when the interval timers were all cancelled; and, the tables needed to be scanned with `select` and subsequent `delete`s to find and remove all timers linked to a pid. While, with links, there would only be a single `EXIT` message, now with monitors there are `DOWN` messages for every interval timer tied to a process, ~but as cancellation is done in spawned processes, the timer server is not blocked while they pour in~.

~Doing timeout processing and cancellation in spawned processes has the potential for race conditions when it comes to interval timers, where two tables are involved and the scheduled timeouts are taken and re-inserted. For this case, the order of operations has been slightly changed. @juhlig and I, independently, played out every possible order of events on paper, and don't see a way where a race condition could lead to orphaned timers. But we may have missed something nevertheless, so it would be good if someone took another look at it.~

The test suite(s) have not been touched much and still pass (with only some very small alterations). The `get_status` function which is only for test purposes is now obsolete and replaced by retrieving the state and reading the table info. ~I left it in, though, if you think it should remain.~

~There was a special case that we don't know how to address properly, so it is still there. An interval timer with an interval of `0` will stall the timer server, as the `timer_timeout` function will hit it over and over again, that is, the timer does not "move". Ideas, anyone?~

~Another thing where we are unsure is the handling of manual cancellation. As mentioned, the actual operation is done in a spawned process (so it won't hinder the timer server), but the answer to the caller is kept back until it has finished. Not sure if this is necessary, after all, `{ok, cancelled}` is returned in any case, even if the timer didn't exist in the first place. So it would be possible to reply immediately. Opinions?~

[EDIT]: See https://github.com/erlang/otp/pull/4811#issuecomment-839833211 below.

### DIFF
--- a/lib/stdlib/doc/src/timer.xml
+++ b/lib/stdlib/doc/src/timer.xml
@@ -50,7 +50,9 @@
     <p>Creating timers using
       <seemfa marker="erts:erlang#send_after/3">erlang:send_after/3</seemfa> and
       <seemfa marker="erts:erlang#start_timer/3">erlang:start_timer/3</seemfa>
-      is much more efficient than using the timers provided by this module. See
+      is more efficient than using the timers provided by this module. However,
+      the timer module has been improved in OTP 25, making it more efficient and
+      less susceptible to being overloaded. See
       <seeguide marker="system/efficiency_guide:commoncaveats#timer-module">the
       Timer Module section in the Efficiency Guide</seeguide>.</p>
   </description>
@@ -116,8 +118,9 @@
           <c>exit_after(<anno>Time</anno>, self(),
           <anno>Reason1</anno>)</c>.</p>
         <p><c>exit_after/3</c> sends an exit signal with reason
-          <c><anno>Reason1</anno></c> to
-          pid <c><anno>Pid</anno></c>. Returns <c>{ok, <anno>TRef</anno>}</c>
+          <c><anno>Reason1</anno></c> to <c><anno>Target</anno></c>,
+          which can be a local process identifier or an atom of a registered
+          name. Returns <c>{ok, <anno>TRef</anno>}</c>
           or <c>{error, <anno>Reason2</anno>}</c>.</p>
       </desc>
     </func>
@@ -149,7 +152,7 @@
         <p><c>kill_after/1</c> is the same as
           <c>exit_after(<anno>Time</anno>, self(), kill)</c>.</p>
         <p><c>kill_after/2</c> is the same as
-          <c>exit_after(<anno>Time</anno>, <anno>Pid</anno>, kill)</c>.</p>
+          <c>exit_after(<anno>Time</anno>, <anno>Target</anno>, kill)</c>.</p>
       </desc>
     </func>
 
@@ -190,15 +193,16 @@
     <func>
       <name name="send_after" arity="2" since=""/>
       <name name="send_after" arity="3" since=""/>
-      <fsummary>Send <c>Message</c> to <c>Pid</c> after a specified
+      <fsummary>Send <c>Message</c> to <c>Destination</c> after a specified
         <c>Time</c>.</fsummary>
       <desc>
         <taglist>
           <tag><c>send_after/3</c></tag>
           <item>
-            <p>Evaluates <c><anno>Pid</anno> ! <anno>Message</anno></c> after
-              <c><anno>Time</anno></c> milliseconds. (<c><anno>Pid</anno></c>
-              can also be an atom of a registered name.)</p>
+            <p>Evaluates <c><anno>Destination</anno> ! <anno>Message</anno></c> after
+              <c><anno>Time</anno></c> milliseconds. (<c><anno>Destination</anno></c> can
+              be a remote or local process identifier, an atom of a registered name or
+              a tuple <c>{RegName, Node}</c> for a registered name at another node.)</p>
             <p>Returns <c>{ok, <anno>TRef</anno>}</c> or
               <c>{error, <anno>Reason</anno>}</c>.</p>
             <p>See also
@@ -223,10 +227,11 @@
         <taglist>
           <tag><c>send_interval/3</c></tag>
           <item>
-            <p>Evaluates <c><anno>Pid</anno> ! <anno>Message</anno></c>
+            <p>Evaluates <c><anno>Destination</anno> ! <anno>Message</anno></c>
               repeatedly after <c><anno>Time</anno></c> milliseconds.
-              (<c><anno>Pid</anno></c> can also be
-              an atom of a registered name.)</p>
+              (<c><anno>Destination</anno></c> can be a remote or local process
+              identifier, an atom of a registered name or a tuple <c>{RegName, Node}</c>
+              for a registered name at another node.)</p>
             <p>Returns <c>{ok, <anno>TRef</anno>}</c> or
               <c>{error, <anno>Reason</anno>}</c>.</p>
           </item>
@@ -249,6 +254,11 @@
           or suspends the process forever if <c><anno>Time</anno></c> is the
           atom <c>infinity</c>. Naturally, this
           function does <em>not</em> return immediately.</p>
+        <note>
+          <p>Before OTP 25, <c>timer:sleep/1</c> did not accept integer
+            timeout values greater than <c>16#ffffffff</c>, that is, <c>2^32-1</c>.
+            Since OTP 25, arbitrarily high integer values are accepted.</p>
+        </note>
       </desc>
     </func>
 
@@ -350,4 +360,3 @@ timer:cancel(R),
       <seemfa marker="#cancel/1"><c>cancel/1</c></seemfa>.</p>
   </section>
 </erlref>
-
--- a/lib/stdlib/src/timer.erl
+++ b/lib/stdlib/src/timer.erl
@@ -26,129 +26,198 @@
 	 cancel/1, sleep/1, tc/1, tc/2, tc/3, now_diff/2,
 	 seconds/1, minutes/1, hours/1, hms/3]).
 
--export([start_link/0, start/0, 
-	 handle_call/3,  handle_info/2,  
+-export([start_link/0, start/0,
+	 handle_call/3,  handle_info/2,
 	 init/1,
 	 code_change/3, handle_cast/2, terminate/2]).
 
-%% internal exports for test purposes only
--export([get_status/0]).
-
-%% types which can be used by other modules
+%% Types which can be used by other modules
 -export_type([tref/0]).
 
-%% Max
--define(MAX_TIMEOUT, 16#0800000).
--define(TIMER_TAB, timer_tab).
--define(INTERVAL_TAB, timer_interval_tab).
+%% Max value for a receive's after clause.
+-define(MAX_RECEIVE_AFTER, 16#ffffffff).
+
+%% Validations
+-define(valid_time(T), is_integer(T), T >= 0).
+-define(valid_mfa(M, F, A), is_atom(M), is_atom(F), is_list(A)).
 
 %%
 %% Time is in milliseconds.
 %%
--opaque tref()    :: {integer(), reference()}.
+-opaque tref()    :: {type(), reference()}.
+-type type()      :: 'once' | 'interval' | 'instant' | 'send_local'.
 -type time()      :: non_neg_integer().
 
 %%
 %% Interface functions
 %%
 -spec apply_after(Time, Module, Function, Arguments) ->
-                         {'ok', TRef} | {'error', Reason} when
-      Time :: time(),
-      Module :: module(),
-      Function :: atom(),
-      Arguments :: [term()],
-      TRef :: tref(),
-      Reason :: term().
+          {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Module :: module(),
+                   Function :: atom(),
+                   Arguments :: [term()],
+                   TRef :: tref(),
+                   Reason :: term().
+apply_after(0, M, F, A)
+  when ?valid_mfa(M, F, A) ->
+    do_apply({M, F, A}),
+    {ok, {instant, make_ref()}};
+apply_after(Time, M, F, A)
+  when ?valid_time(Time),
+       ?valid_mfa(M, F, A) ->
+    req(apply_once, {system_time(), Time, {M, F, A}});
+apply_after(_Time, _M, _F, _A) ->
+    {error, badarg}.
 
-apply_after(Time, M, F, A) ->
-    req(apply_after, {Time, {M, F, A}}).
+-spec send_after(Time, Destination, Message) -> {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Destination :: pid() | (RegName :: atom()) | {RegName :: atom(), Node :: node()},
+                   Message :: term(),
+                   TRef :: tref(),
+                   Reason :: term().
+send_after(0, PidOrRegName, Message)
+  when is_pid(PidOrRegName);
+       is_atom(PidOrRegName) ->
+    PidOrRegName ! Message,
+    {ok, {instant, make_ref()}};
+send_after(0, {RegName, Node} = Dest, Message)
+  when is_atom(RegName),
+       is_atom(Node) ->
+    Dest ! Message,
+    {ok, {instant, make_ref()}};
+send_after(Time, Pid, Message)
+  when ?valid_time(Time),
+       is_pid(Pid),
+       node(Pid) =:= node() ->
+    TRef = erlang:send_after(Time, Pid, Message),
+    {ok, {send_local, TRef}};
+send_after(Time, Pid, Message)
+  when is_pid(Pid) ->
+    apply_after(Time, ?MODULE, send, [Pid, Message]);
+send_after(Time, RegName, Message)
+  when is_atom(RegName) ->
+    apply_after(Time, ?MODULE, send, [RegName, Message]);
+send_after(Time, {RegName, Node} = Dest, Message)
+  when is_atom(RegName),
+       is_atom(Node) ->
+    apply_after(Time, ?MODULE, send, [Dest, Message]);
+send_after(_Time, _PidOrRegName, _Message) ->
+    {error, badarg}.
 
--spec send_after(Time, Pid, Message) -> {'ok', TRef} | {'error', Reason} when
-      Time :: time(),
-      Pid :: pid() | (RegName :: atom()),
-      Message :: term(),
-      TRef :: tref(),
-      Reason :: term().
-send_after(Time, Pid, Message) ->
-    req(apply_after, {Time, {?MODULE, send, [Pid, Message]}}).
-
--spec send_after(Time, Message) -> {'ok', TRef} | {'error', Reason} when
-      Time :: time(),
-      Message :: term(),
-      TRef :: tref(),
-      Reason :: term().
+-spec send_after(Time, Message) -> {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Message :: term(),
+                   TRef :: tref(),
+                   Reason :: term().
 send_after(Time, Message) ->
     send_after(Time, self(), Message).
 
--spec exit_after(Time, Pid, Reason1) -> {'ok', TRef} | {'error', Reason2} when
-      Time :: time(),
-      Pid :: pid() | (RegName :: atom()),
-      TRef :: tref(),
-      Reason1 :: term(),
-      Reason2 :: term().
+-spec exit_after(Time, Target, Reason1) -> {'ok', TRef} | {'error', Reason2}
+              when Time :: time(),
+                   Target :: pid() | (RegName :: atom()),
+                   TRef :: tref(),
+                   Reason1 :: term(),
+                   Reason2 :: term().
 exit_after(Time, Pid, Reason) ->
-    req(apply_after, {Time, {erlang, exit, [Pid, Reason]}}).
+    apply_after(Time, erlang, exit, [Pid, Reason]).
 
--spec exit_after(Time, Reason1) -> {'ok', TRef} | {'error', Reason2} when
-      Time :: time(),
-      TRef :: tref(),
-      Reason1 :: term(),
-      Reason2 :: term().
+-spec exit_after(Time, Reason1) -> {'ok', TRef} | {'error', Reason2}
+              when Time :: time(),
+                   TRef :: tref(),
+                   Reason1 :: term(),
+                   Reason2 :: term().
 exit_after(Time, Reason) ->
     exit_after(Time, self(), Reason).
 
--spec kill_after(Time, Pid) -> {'ok', TRef} | {'error', Reason2} when
-      Time :: time(),
-      Pid :: pid() | (RegName :: atom()),
-      TRef :: tref(),
-      Reason2 :: term().
+-spec kill_after(Time, Target) -> {'ok', TRef} | {'error', Reason2}
+              when Time :: time(),
+                   Target :: pid() | (RegName :: atom()),
+                   TRef :: tref(),
+                   Reason2 :: term().
 kill_after(Time, Pid) ->
     exit_after(Time, Pid, kill).
 
--spec kill_after(Time) -> {'ok', TRef} | {'error', Reason2} when
-      Time :: time(),
-      TRef :: tref(),
-      Reason2 :: term().
+-spec kill_after(Time) -> {'ok', TRef} | {'error', Reason2}
+              when Time :: time(),
+                   TRef :: tref(),
+                   Reason2 :: term().
 kill_after(Time) ->
     exit_after(Time, self(), kill).
 
 -spec apply_interval(Time, Module, Function, Arguments) ->
-                            {'ok', TRef} | {'error', Reason} when
-      Time :: time(),
-      Module :: module(),
-      Function :: atom(),
-      Arguments :: [term()],
-      TRef :: tref(),
-      Reason :: term().
-apply_interval(Time, M, F, A) ->
-    req(apply_interval, {Time, self(), {M, F, A}}).
+          {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Module :: module(),
+                   Function :: atom(),
+                   Arguments :: [term()],
+                   TRef :: tref(),
+                   Reason :: term().
+apply_interval(Time, M, F, A)
+  when ?valid_time(Time),
+       ?valid_mfa(M, F, A) ->
+    req(apply_interval, {system_time(), Time, self(), {M, F, A}});
+apply_interval(_Time, _M, _F, _A) ->
+    {error, badarg}.
 
--spec send_interval(Time, Pid, Message) ->
-                           {'ok', TRef} | {'error', Reason} when
-      Time :: time(),
-      Pid :: pid() | (RegName :: atom()),
-      Message :: term(),
-      TRef :: tref(),
-      Reason :: term().
-send_interval(Time, Pid, Message) ->
-    req(apply_interval, {Time, Pid, {?MODULE, send, [Pid, Message]}}).
+-spec send_interval(Time, Destination, Message) -> {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Destination :: pid() | (RegName :: atom()) | {RegName :: atom(), Node :: node()},
+                   Message :: term(),
+                   TRef :: tref(),
+                   Reason :: term().
+send_interval(Time, Pid, Message)
+  when ?valid_time(Time),
+       is_pid(Pid) ->
+    req(apply_interval, {system_time(), Time, Pid, {?MODULE, send, [Pid, Message]}});
+send_interval(Time, RegName, Message)
+  when ?valid_time(Time),
+       is_atom(RegName) ->
+    req(apply_interval, {system_time(), Time, RegName, {?MODULE, send, [RegName, Message]}});
+send_interval(Time, Dest = {RegName, Node}, Message)
+  when ?valid_time(Time),
+       is_atom(RegName),
+       is_atom(Node) ->
+    req(apply_interval, {system_time(), Time, Dest, {?MODULE, send, [Dest, Message]}});
+send_interval(_Time, _Pid, _Message) ->
+    {error, badarg}.
 
--spec send_interval(Time, Message) -> {'ok', TRef} | {'error', Reason} when
-      Time :: time(),
-      Message :: term(),
-      TRef :: tref(),
-      Reason :: term().
+-spec send_interval(Time, Message) -> {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Message :: term(),
+                   TRef :: tref(),
+                   Reason :: term().
 send_interval(Time, Message) ->
     send_interval(Time, self(), Message).
 
--spec cancel(TRef) -> {'ok', 'cancel'} | {'error', Reason} when
-      TRef :: tref(),
-      Reason :: term().
-cancel(BRef) ->
-    req(cancel, BRef).
+-spec cancel(TRef) -> {'ok', 'cancel'} | {'error', Reason}
+              when TRef :: tref(),
+                   Reason :: term().
+cancel({instant, Ref})
+  when is_reference(Ref) ->
+    {ok, cancel};
+cancel({send_local, Ref})
+  when is_reference(Ref) ->
+    _ = erlang:cancel_timer(Ref),
+    {ok, cancel};
+cancel({once, Ref} = TRef)
+  when is_reference(Ref) ->
+    req(cancel, TRef);
+cancel({interval, Ref} = TRef)
+  when is_reference(Ref) ->
+    req(cancel, TRef);
+cancel(_TRef) ->
+    {error, badarg}.
 
--spec sleep(Time) -> 'ok' when
-      Time :: timeout().
+-spec sleep(Time) -> 'ok'
+              when Time :: timeout().
+sleep(T)
+  when is_integer(T),
+       T > ?MAX_RECEIVE_AFTER ->
+    receive
+    after ?MAX_RECEIVE_AFTER ->
+            sleep(T - ?MAX_RECEIVE_AFTER)
+    end;
 sleep(T) ->
     receive
     after T -> ok
@@ -157,10 +226,10 @@ sleep(T) ->
 %%
 %% Measure the execution time (in microseconds) for Fun().
 %%
--spec tc(Fun) -> {Time, Value} when
-      Fun :: function(),
-      Time :: integer(),
-      Value :: term().
+-spec tc(Fun) -> {Time, Value}
+              when Fun :: function(),
+                   Time :: integer(),
+                   Value :: term().
 tc(F) ->
     T1 = erlang:monotonic_time(),
     Val = F(),
@@ -171,11 +240,11 @@ tc(F) ->
 %%
 %% Measure the execution time (in microseconds) for Fun(Args).
 %%
--spec tc(Fun, Arguments) -> {Time, Value} when
-      Fun :: function(),
-      Arguments :: [term()],
-      Time :: integer(),
-      Value :: term().
+-spec tc(Fun, Arguments) -> {Time, Value}
+              when Fun :: function(),
+                   Arguments :: [term()],
+                   Time :: integer(),
+                   Value :: term().
 tc(F, A) ->
     T1 = erlang:monotonic_time(),
     Val = apply(F, A),
@@ -186,12 +255,12 @@ tc(F, A) ->
 %%
 %% Measure the execution time (in microseconds) for an MFA.
 %%
--spec tc(Module, Function, Arguments) -> {Time, Value} when
-      Module :: module(),
-      Function :: atom(),
-      Arguments :: [term()],
-      Time :: integer(),
-      Value :: term().
+-spec tc(Module, Function, Arguments) -> {Time, Value}
+              when Module :: module(),
+                   Function :: atom(),
+                   Arguments :: [term()],
+                   Time :: integer(),
+                   Value :: term().
 tc(M, F, A) ->
     T1 = erlang:monotonic_time(),
     Val = apply(M, F, A),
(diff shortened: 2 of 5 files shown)


Maria-12648430:
Oh, one thing I forgot to mention: `timer:sleep/1` now works with arbitrary timeouts, that is, it can sleep for ages ;) This is just a cosmetic change, nobody should _want_ to wait for more than ~50 days. But it was the _only_ function in the `timer` module that didn't accept arbitrary timeouts.

lib/stdlib/src/timer.erl
```diff
@@ -130,8 +148,21 @@ apply_interval(Time, M, F, A) ->
...
-    req(apply_interval, {Time, Pid, {?MODULE, send, [Pid, Message]}}).
+send_interval(Time, Pid, Message)
+  when ?valid_time(Time),
+       is_pid(Pid) ->
+    req(apply_interval, {Time, Pid, {?MODULE, send, [Pid, Message]}});
+send_interval(Time, RegName, Message)
+  when ?valid_time(Time),
+       is_atom(RegName) ->
+    case get_pid(RegName) of
+	Pid when is_pid(Pid) ->
```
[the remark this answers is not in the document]
josevalim: Ah, please ignore me. We were calling `get_pid` early before as well, the only difference is that it was happening on the server instead of the client. :)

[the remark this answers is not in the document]
Maria-12648430: Right =^^= But anyway, the general plan was to not have changes in outside behavior, so if you spot anything that may have an impact there, please report :)

lib/stdlib/src/timer.erl
```diff
@@ -236,106 +278,128 @@ hours(Hours) ->
...
+%% Start an interval timer.
+handle_call({apply_interval, {Time, Pid, MFA}, Started}, _From, State) ->
+    Interval = Time * 1000,
+    StartTime = Started + Interval,
+    Ref = monitor(process, Pid),
+    TRef = {StartTime, Ref},
+    Timer = {TRef, {repeat, Interval}, MFA},
+    ets:insert(State#state.interval_tab, {Ref, TRef}),
+    ets:insert(State#state.timer_tab, Timer),
+    {reply, {ok, {interval, Ref}}, State, next_timeout(State)};
```
josevalim, line 348: I wonder if this implementation is trying to be too smart because it was implemented before timer wheels were added to the VM. Maybe the simplest and faster approach now is to use send_after instead of getting the `next_timeout`. This would have the benefit that the management of the timer for operations like `apply_after` could be done completely in the client (the server would effectively only receive the MFA on timeout).

juhlig: Hm, not a bad idea... I'll have to think this through a bit more, but in theory, this could eliminate the need for the `ordered_set` timer table altogether. Interval timers would be just timers that reschedule the send, too. It will probably still all have to go through the timer server, so we know what timers we _actually_ have and for things like `cancel` etc, but... yeah :)

@Maria-12648430 any objections, in terms of keeping outward behavior specifically?

Maria-12648430: > @Maria-12648430 any objections, in terms of keeping outward behavior specifically?

Not off the top of my head, no. But I will check in more depth, after lunch :)

Maria-12648430: No objections, at least I don't see any problems ^^;

rickard-green: By using the `{abs, true}` option of `erlang:send_after/4` or `erlang:start_timer/4` you can prevent that interval timers slowly drifts as well.

rickard-green: Had a quick look at the code and slow drift of intervals is perhaps not an issue in current implementation. Anyway, the `{abs, true}` option may help with intervals if rewritten using `send_after()`/`start_timer()`.

Maria-12648430:
The last commit takes into account the suggestions by @josevalim and @rickard-green. By doing so, the implementation of `timer` has deviated quite a bit from the current implementation, and also from the one we proposed before.

* One-shot timers (generally the ones started via the `timer:*_after` functions) are not stored in the timer server any more. They start an (`erlang:`) timer running in the client process via `erlang:send_after/3`, which just sends the `MFA` to the timer server to execute after the given timeout (@josevalim's suggestion).
* Interval timers (the ones started via `timer:*_interval` functions) are started by the timer server. It keeps a simple lookup table for them, mapping the `TRef` that the client starting the timer receives to the reference of the actual (`erlang:`) timer and updates them accordingly each time a interval is triggered and a new (`erlang:`) timer needs to be started. Interval timers rely on `{abs, true}` to avoid drift (@rickard-green's suggestion)
* `timer:cancel` works in two ways:
  * if the timer to cancel is a one-shot timer, it simply calls `erlang:cancel_timer` with the given reference
  * if it is an interval timer, it tells the timer server to cancel the timer, which looks up (and removes) the timer by the (`timer:`) `TRef` and cancels the associated (`erlang:`) timer.
* Any use of spawned processes introduced by the previous commit have been removed again, for three reasons:
  * Instructions to apply an `MFA` are not actually _applied_ (in the sense of `apply/3`) but used in a `spawn/3` call, so they already are run in separate processes.
  * Aside from that, the timer server only manages interval timers, in an `ets` table. All operations on that table involve the key, so I think we can't get faster than that, not much anyway.
  * Processing requests (namely, `apply_interval` tasks) serially also works as a throttling mechanism, preventing interval timers with very low (maybe even `0`) intervals from flooding the message queue. Put differently, at any time, for an interval timer there is either 1 `erlang` timer running, or 1 apply instruction message in the timer server's queue. Never both, never more than one.

You might notice some peculiarities in our implementation, they are usually there to keep the new `timer` implementation's behavior outwardly the same as the old one's. As far as we are aware, there is really only one difference in behavior, which is that the time values cannot be really arbitrarily high any more, but are subject to the same restrictions as outlined in the documentation for `erlang:start_timer/4`.

Aside from the changes outlined above, we extended the test suite, and reorganized it to a degree. Changes to existing tests, however, are just marginal.

garazdawi:
> if the timer to cancel is a one-shot timer, it simply calls erlang:cancel_timer with the given reference

I think this introduce a race condition that was not there before. Imagine this:

```erlang
start() ->
  Ref = make_ref(),
  Tmr = timer:apply_after(1000, ?MODULE, echo, [self(), Ref]),
  timer:sleep(1000),
  timer:cancel(Tmr),
  %% Flush potential message from mail-box
  receive Ref -> ok after 0 -> ok.

echo(Pid, Ref) -> Pid ! Ref.
```

After the code above has executed, in the old implementation we know that the message sent in `echo` will not be in the mailbox of the caller. However, with the proposed solution, the message can arrive later as the timer might have triggered just before we try to cancel it.

I don't know if the above is something that the `timer` API should guarantee. Just thought that I would mention it so that we keep it in mind.

Maria-12648430:
>   Tmr = timer:apply_after(1000, ?MODULE, echo, [self(), Ref]),

(it should be `{ok, Tmr} = ...`, just nitpicking ;))

> After the code above has executed, in the old implementation we know that the message sent in `echo` will not be in the mailbox of the caller. However, with the proposed solution, the message can arrive later as the timer might have triggered just before we try to cancel it.

Hm, I'm still trying to wrap my head around this, but...

It seems to me that this race condition is present in the old implementation as well, when using `apply_after` at least (not when using `send_after`, `exit_after` or `kill_after`), because the MFAs are passed into `spawn/3` (except `{timer, send, [To, Msg]}` and `{erlang, exit, [Pid, Reason]}`, they are executed in the timer process directly). So it _may_ happen that `timer:cancel` gets called while the timer server is processing the timers. The call will return only after it has finished that, so yes, the timer is definitely gone, but the action is performed in a spawned process, and so the message sent by `echo` in this case may arrive later, after the `receive`. Right?

The only difference with the new implementation would be that the same may happen with `send_after`, `exit_after` and `kill_after`, which wasn't the case with the old implementation.

Maria-12648430:
> ```erlang
> echo(Pid, Ref) -> Pid ! Ref.
> ```

Put a `timer:sleep` before the send operation to see what i mean ;)

Maria-12648430:
I'm aware that the new implementation adds another layer to this, namely that the (`erlang:`) timer has sent the apply instruction to the timer server just the moment we try to cancel it. There may then be a small delay before the timer server receives and processes it.

I think that at the expense of more complexity, we could probably bring back the old behavior for `send_after`, `exit_after` and `kill_after`. However, I guess we cannot reasonably prevent side effects from `apply_after` calls of being late.

Maria-12648430:
I just realized, this practice of fire-and-forget spawning might turn out a hidden trapdoor when short intervals meet long execution times in `apply_interval`...
On average, there will be about `execution_time / interval` processes running for a given interval timer. When the system gets more load, execution times tend to drag out, but the interval is fixed and the timer accordingly will happily spawn process after process, putting even more stress on the system.

As we want to keep the old behavior, I don't think we should try to change anything here. But a warning in the docs might be appropriate.

lib/stdlib/src/timer.erl
```diff
@@ -26,129 +26,192 @@
...
+	    {'ok', TRef} | {'error', Reason}
+  when Time :: time(),
+       Module :: module(),
+       Function :: atom(),
+       Arguments :: [term()],
+       TRef :: tref(),
+       Reason :: term().
+apply_after(0, M, F, A)
+  when ?valid_mfa(M, F, A) ->
+    do_apply({M, F, A});
```
bjorng, line 64: The return value will not be correct:

```
1> timer:apply_after(0, erlang, abs, [-42]).
<0.85.0>
2> timer:apply_after(0, erlang, exit, [[], whatever]).
{'EXIT',{badarg,[{erlang,exit,
                         [undefined,whatever],
                         [{error_info,#{module => erl_erts_errors}}]},
                 {timer,do_apply,1,[{file,"timer.erl"},{line,468}]},
                 {erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,689}]},
```

Please update the test suite to test the return values.

(The reason that Dialyzer did not complain is probably because `do_apply/1` uses an old-style `catch`. An old-style `catch` can hide type violations for Dialyzer.)

lib/stdlib/src/timer.erl
```diff
@@ -26,129 +26,192 @@
...
+  when Time :: time(),
+       Pid :: pid() | (RegName :: atom()),
+       Message :: term(),
+       TRef :: tref(),
+       Reason :: term().
+send_after(0, PidOrRegName, Message)
+  when is_pid(PidOrRegName);
+       is_atom(PidOrRegName) ->
+    PidOrRegName ! Message;
+send_after(0, Dest = {RegName, Node}, Message) % Undocumented; for backwards compatibility.
```
bjorng, line 91: I don't see any reason to not document the `{Name, Node}` and remove the comment. Do you?

lib/stdlib/test/timer_simple_SUITE.erl
```diff
@@ -89,11 +194,24 @@ init_per_testcase(_, Config) when is_list(Config) ->
 
 %% Testing timer interface!!
 
-%% Test of apply_after, with sending of message.
-apply_after(Config) when is_list(Config) ->
+%% Test of apply_after with time = 0, with sending of message.
+apply_after1(Config) when is_list(Config) ->
+    timer:apply_after(0, ?MODULE, send, [self(), ok_apply]),
```
bjorng, line 199: Please test all return values to ensure that they follow the documentation.

bjorng (commented):
Thanks for making the effort to clean up and improve the `timer` module. I have posted a few initial comments, but I will come back with more...

lib/stdlib/src/timer.erl
Maria-12648430: Funny thing is, I believe we had this right in one of the earlier commits =^^=

lib/stdlib/src/timer.erl
Maria-12648430: It's not working with `send_interval` and up to now I thought this couldn't be done, but I just realized that it actually is possible now.
I also see that there is actually a slight (unintended, @juhlig?) change in behavior in `send_interval`.
I'll see to it.

lib/stdlib/test/timer_simple_SUITE.erl
Maria-12648430: Ok, will do :)

KennethL:
One reason for storing the timers in the timer_server was that it supports larger integer values for the timeouts (bignums) than the BIFs like `erlang:send_after` so it is not that easy to remove the storage of timers in the timer_server (or its ets -tables) if we want to keep the current functionality. I also think (have not checked the C-code yet) that the timeoutvalue might have a different limit on Erlang systems built for 32-bit and 64-bit. The timer module is implemented with the intention that there should be no difference between 32 and 64 bit systems. At least that is what I remember from when I worked with the timer module a long time ago. Tested in the shell and for sure erlang:send_after has a lower limit than timer:send_after. Probably the supported timeout value is big enough on 64 bit systems, but it is not good to have a different limit on 32 and 64 bit systems.

lib/stdlib/src/timer.erl
juhlig: @Maria-12648430... so is it still Easter or what? ^^;;;

Maria-12648430: @juhlig yes 😁 (Hint: check your mail)

KennethL:
Ok, you can forget my comment, the timer BIFs can handle big enough timeouts (since OTP 18)

lib/stdlib/src/timer.erl
Maria-12648430: Fixed and added return values to tests in latest commit.

lib/stdlib/src/timer.erl
Maria-12648430: Added to docs in last commit, also changed `send_interval` to match.

Just noticed that I forgot to remove the comment, though. Will do that later.

Now, the only functions that do not accept `{RegName, Node}` tuples are `exit_after` and `kill_after`, because `exit/2`  only accepts pids, and TBH I don't know how to resolve the tuple to a pid 😅

lib/stdlib/test/timer_simple_SUITE.erl
Maria-12648430: Done in `timer_SUITE` and `timer_simple_SUITE`.

lib/stdlib/src/timer.erl
Maria-12648430: ... or even do the `exit` itself via `rpc:cast/4`?

lib/stdlib/src/timer.erl
```diff
@@ -203,250 +269,216 @@ tc(M, F, A) ->
...
+do_apply({?MODULE, send, A}) ->
+    catch send(A);
+%% If exit op. resolve registered name
+do_apply({erlang, exit, [Name, Reason]}) ->
+    catch exit(get_pid(Name), Reason);
+do_apply({M,F,A}) ->
+    catch spawn(M,F,A).
 
-%%
-%%  system_time() -> time in microseconds
```
bjorng, line 442: I suggest that you add a comment that explains what this function does. It is not immediately obvious from the code what is returned.

lib/stdlib/test/timer_SUITE.erl
```diff
@@ -168,7 +168,7 @@ s_a_t(C, TimeOut) ->
 a_t(C, TimeOut) ->
     start_watchdog(self(), TimeOut),
     Start = system_time(),
-    timer:send_after(TimeOut, self(), now),
+    {ok, {once, _}} = timer:send_after(TimeOut, self(), now),
```
bjorng, line 171: Testing undocumented implementation details (`{ok, {once, _}}` instead of `{ok, _}`) is OK and may find more bugs which is always good, but I think you should add a comment at the beginning of the module to mention that you do that and that the test suite might need to be updated if those implementation details are changed.

bjorng (commented):
Here follows my nitpicks.

bjorng:
Here follows some thoughts on what needs to be updated in the documentation.

It should be documented that `timer:sleep/1` now accepts as long timeout value as all the other functions in the `timer` module.

The current documentation says:

> Creating timers using erlang:send_after/3 and erlang:start_timer/3 is much more efficient than using the timers provided by this module.

Would it be appropriate to delete "much"?

The Efficiency Guide says:

> The `timer` module uses a separate process to manage the timers. That process can easily become overloaded if many processes create and cancel timers frequently (especially when using the SMP emulator).

(The part about the SMP emulator should be deleted because it is irrelevant and has been for several years.)

Do we think that the timer process can still be "easily" overloaded?

In any case, both the `timer` documentation and Efficiency Guide should have notes that mention that the implementation has been improved in OTP 25.

lib/stdlib/src/timer.erl
Maria-12648430: Added in last commit.

lib/stdlib/test/timer_SUITE.erl
Maria-12648430: Added in last commit.

Maria-12648430:
> > Creating timers using erlang:send_after/3 and erlang:start_timer/3 is much more efficient than using the timers provided by this module.
> 
> Would it be appropriate to delete "much"?

Definitely.

For one-shot timers, the new implementation does not manage the timers at all, the actual timers are related to the caller process and the server timer process is just told to execute (send message/exit process/spawn MFA) the actions when they expire. Cancellations also just cancels the timers related to the caller process and don't touch the timer server.

Interval timers are managed (ie, the timer server knows them so they can be cancelled, and monitors the related process to automatically cancel timers in case of that process dying) and the associated timers are related to the timer server process, so they are a bit more demanding, but not much.

In any case, the actual process of _timing_ is not part of the timer server any more.

> The Efficiency Guide says:
> 
> > The `timer` module uses a separate process to manage the timers. That process can easily become overloaded if many processes create and cancel timers frequently (especially when using the SMP emulator).
> 
> (The part about the SMP emulator should be deleted because it is irrelevant and has been for several years.)

I don't see it in the current documentation (?)

> Do we think that the timer process can still be "easily" overloaded?

I wouldn't say so, no ;) The actual timer server is now, give or take, only concerned with handling its message queue, which can contain
* one-shot timer expirations, all resulting in fast operations, either send a message, `exit/2` a process, or `spawn/3` an MFA
* interval timer expirations, as above, but followed by starting another `erlang:send_after` timer and updating the reference in the timer table
* starting of interval timers, meaning inserting them in the timer table and starting an `erlang:send_after` timer
* cancelling of interval timers, meaning looking up and removing (taking) the reference from the timer table and cancelling the `erlang:send_after` timer
* `DOWN` messages from a monitored (interval-timer related) process, resulting in the same as cancelling an interval timer

The death of a monitored process may result in a surge of `DOWN` messages, cancelling interval timers one by one, if the process was the destination of many `timer:send_interval` timers and/or if it was running many `timer:apply_interval` timers.

josevalim:
For what is worth, when we got Phoenix to 2 million connections, the first bottleneck that we found was in the `timer` module. We were using `timer:send_interval/3` to send pings to clients. The server had 40 (or 48 cores) and the bottleneck was reached at around 30k simultaneous connections. Luckily, a very simple bottleneck to address.

While the proposed improvements are more efficient and really welcome, I still think it would be a bottleneck in the use case above. Nowadays, I generally struggle to see use cases for `send_interval/3`, because the majority of use cases can be easily implemented with `send_after/3`. If I see `timer:send_interval/3` in a code review, I'd most likely suggest it to be rewritten. :)

In any case, a disclaimer that says that the `timer` module is backed by a single process and therefore it can become a bottleneck is enough? As any other `gen_server`.

bjorng:
> > (The part about the SMP emulator should be deleted because it is irrelevant and has been for several years.)
> I don't see it in the current documentation (?)

Somehow I must have looked at an older version. It is deleted in OTP 24.

> In any case, a disclaimer that says that the timer module is backed by a single process and therefore it can become a bottleneck is enough?

Yes, I think that is a good idea, along with a mention that single process does less in OTP 24.

Maria-12648430:
> For what is worth, when we got Phoenix to 2 million connections, the first bottleneck that we found was in the `timer` module. We were using `timer:send_interval/3` to send pings to clients. The server had 40 (or 48 cores) and the bottleneck was reached at around 30k simultaneous connections. Luckily, a very simple bottleneck to address.

The timer management overhead mainly contributed to that, I guess... With 30k timers and, say, a ping interval of 30s, that would amount to 1000 timers firing (and being re-scheduled) per second, with operations in two tables, removal of stale timers using a select on a non-key and subsequent deletes in two tables again, etcetc... Meaning the timer server would spend more and more time just managing its timers and have less and less time for actual requests like starting new ones or cancelling.

> While the proposed improvements are more efficient and really welcome, I still think it would be a bottleneck in the use case above.

Just put enough pressure on it I guess ;) But with most of the expensive management gone, I'd expect the point of serious performance degradation to be pushed quite a way back. Feel free to share experiences when you get any ;)

> In any case, a disclaimer that says that the `timer` module is backed by a single process and therefore it can become a bottleneck is enough? As any other `gen_server`.

@bjorng
> ... that single process does less in OTP 24.

What do you mean? ^^;;;

bjorng:
> > ... that single process does less in OTP 24.
> What do you mean? ^^;;;

That the single process does less work and has less overhead.

lib/stdlib/src/timer.erl
Maria-12648430: Hm, now one thing came to mind regarding the original question about documenting `{RegName, Node}`: `Node` will be automatically connected by the send operation. This is no change of behavior, though, it would have done that before, too.

Maria-12648430:
@bjorng the last commit adds the suggested documentation changes. Please review :)

lib/stdlib/doc/src/timer.xml
```diff
@@ -249,6 +254,11 @@
           or suspends the process forever if <c><anno>Time</anno></c> is the
           atom <c>infinity</c>. Naturally, this
           function does <em>not</em> return immediately.</p>
+        <note>
+          <p>Before OTP 25, <c>timer:sleep/1</c> did not accept integer
+            timeout values greater than <c>16#ffffffff</c>, that is, <c>2^64-1</c>.
```
bjorng, line 259:
```suggestion
            timeout values greater than <c>16#ffffffff</c>, that is, <c>2^32-1</c>.
```

bjorng: Otherwise the documentation looks good to me.

Maria-12648430: Oh, of course 😅

bjorng:
This pull request is almost ready to be merged.

It seems that you have not indented the code using Emacs. Most code in OTP is indented using the Erlang mode in Emacs. I have pushed a commit that reindents the `timer` module using the Erlang mode in Emacs. If you agree to this change, you can just include it while squashing the commits.

I suggest squashing all commits to a single commit. In the commit message I would like you to summarize the changes. I think that your description in this PR would be a great starting point. We want this information in the commit message itself, so that someone who finds this commit a year later won't have to locate the pull request to find out more information. (See [Writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages).)

When you have squashed the commits, we will add the PR to our daily builds, and unless any problems are discovered, we'll expect to merge this PR within a few days or so.

Again, thanks for all your work on this pull request.

lib/stdlib/src/timer.erl
```diff
@@ -358,8 +358,8 @@ init([]) ->
 
 %% server calls
 
-% Try sending a call. If it fails with reason noproc,
-% try starting the timer server and try once again.
+                                                % Try sending a call. If it fails with reason noproc,
+                                                % try starting the timer server and try once again.
```
lhoguin, line 362: That one's probably not right. Emacs likely wants `%%` instead of `%`.

Maria-12648430:
> It seems that you have not indented the code using Emacs.

Indeed. `vim` is my weapon of choice, and usually I like to indent manually (and in a different fashion) 😉

>I have pushed a commit that reindents the `timer` module using the Erlang mode in Emacs. If you agree to this change, you can just include it while squashing the commits.

Sure, thanks :)

> I suggest squashing all commits to a single commit. In the commit message I would like you to summarize the changes. I think that your description in this PR would be a great starting point.

Okay, I'll try summarizing as best as I can, but it may still turn out a bit verbose ;)

> We want this information in the commit message itself, so that someone who finds this commit a year later won't have to locate the pull request to find out more information. (See [Writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages).)

Btw, I noticed an error there, in the quote from https://git-scm.com/book/ch5-2.html:
```text
  - Typically a hyphen or asterisk is used for the bullet, preceded by a
    single space, with blank lines in between, but conventions vary here
```
It should be "_followed_ by a single space", which is what it says in the quoted document (might have been different at the time when it was quoted, though).

> When you have squashed the commits, we will add the PR to our daily builds, and unless any problems are discovered, we'll expect to merge this PR within a few days or so.

Exciting :)

> Again, thanks for all your work on this pull request.

You're welcome 🤗

lib/stdlib/src/timer.erl
Maria-12648430: Keen eyes :) Thanks for pointing it out, I'll change.

bjorng:
> It should be "followed by a single space", which is what it says in the quoted document (might have been different at the time when it was quoted, though).

Thanks! Fixed.

Maria-12648430:
@bjorng okaaay now, there it is, all squashed up and with my best-shot summary in the commit message.

bjorng:
Thanks! Looks good. Added to our daily builds.

bjorng:
There were some test cases in our daily builds that failed with a `timetrap` timeout, with a stack backtrace that looked like this:

```
=== Location: [{erts_internal,await_result,0},
              {timer,cancel,201},
              {ssh_options_SUITE,try_to_connect,1444},
              {ssh_options_SUITE,max_sessions_sftp_start_channel_sequential,last_expr},
              {test_server,ts_tc,1783},
              {test_server,run_test_case_eval1,1292},
              {test_server,run_test_case_eval,1224}]
=== Reason: timetrap timeout
```

To reproduce:

```
ts:run(ssh, ssh_options_SUITE, [batch]).
```

or:

```
ts:run(ssh, ssh_renegotiate_SUITE, [batch]).
```

One or more test cases should fail with a `timetrap` timeout.

Maria-12648430:
@bjorng huh  🤔 we'll investigate

garazdawi:
Seems like you have stumbled upon an emulator bug when using `[{info,false}]` that has been there unnoticed for a long time. We are trying to figure out what is going on.

Maria-12648430:
> Seems like you have stumbled upon an emulator bug when using `[{info,false}]` that has been there unnoticed for a long time. We are trying to figure out what is going on.

Two birds with one stone then :)

rickard-green:
Fix for the `erlang:cancel_timer(_, [{info, false}])` bug can be found in #4932

lib/stdlib/src/timer.erl
```diff
@@ -26,129 +26,195 @@
...
-    req(cancel, BRef).
-
--spec sleep(Time) -> 'ok' when
-      Time :: timeout().
+-spec cancel(TRef) -> {'ok', 'cancel'} | {'error', Reason}
+              when TRef :: tref(),
+                   Reason :: term().
+cancel({once, TRef})
+  when is_reference(TRef) ->
+    _ = erlang:cancel_timer(TRef, [{info, false}]),
```
rickard-green, line 201: How come you use don't use `{async, true}` here as well? There is no guarantee that the timer has not already been triggered when `timer:cancel()` has returned, so the user must be prepared for a possible timeout message anyway.

`[{async, true}, {info, false}]` gives a real performance gain in case the timer was created on another scheduler while the gain from `[{info, false}]` is minimal (avoids creating an integer).

Maria-12648430: You do have a point :) I'll see to it after lunch.

rickard-green: The combo `[{async, true}, {info, false}]` is also not effected by the `[{info, false}]` bug which is nice :-)

juhlig: I don't know if that was me... if so, I just forgot or was distracted or something.

bjorng:
@Maria-12648430 I've pushed a commit that fixes two failed test cases in the dialyzer test suite. (The updated `timer` module has more complete types than before.) Please squash it.

lib/stdlib/src/timer.erl
Maria-12648430: Added `{async, true}` in last commit.

Maria-12648430:
> @Maria-12648430 I've pushed a commit that fixes two failed test cases in the dialyzer test suite. (The updated `timer` module has more complete types than before.) Please squash it.

Done in last commit, thanks :)

lib/stdlib/src/timer.erl
Maria-12648430: Oh, I just remembered... I did that, and for a reason. I also realized now that I didn't do it properly, though...

The scenario concerns a `send_after` to `self()` timer. When `cancel` is called on such a timer and the timer has not expired yet, it will be stopped and no message will be sent. If it has expired however, the message should also be in the inbox, so it can be instantly flushed out with `receive ... after 0 -> ... end`. For that reason a synchronous (implicit `{async, false}`) call to `erlang:cancel_timer`, which will return only when the timer has really been cancelled (or doesn't exist any more).

I noticed that this falls short when the timer has just expired and told the timer server to execute the send, but the timer server has not done so yet, so a message may still arrive later and be missed by the flush. So we can as well use `{async, true}` for that matter.

Question now is, what shall we do? ;) Keep it like it is now (with `{async, true}`), or do we have to guarantee what the previous implementation (`{async, false}`) tried to achieve?

bjorng: > Oh, I just remembered... I did that, and for a reason.

@rickard-green and I discussed this. We arrived at the following:

With the current timer implementation, a message from the timer process can arrive after canceling the timer, but it is guaranteed that the message can be discarded like this:

```
receive
   <pattern that matches the timer message> -> ok
after 0 ->
   ok
end
```

With the new timer implementation, that guarantee no longer holds.

We think it is necessary to retain the guarantee that a `receive ... after 0 -> ok` will discard the timer message. To do that, the timer needs to be canceled in the timer server process (synchronously using `erlang:cancel_timer/1`). That is, the client process send a `cancel` request to the timer server process, the timer server process cancels the timer, and sends a reply to to client process.

bjorng: I forgot:

The timer server also needs to discard its timer message after canceling the timer. (`timer:apply_after()` should use `erlang:start_timer()` instead of `erlang:send_after()` to include the TimerRef in the timer message.)

Maria-12648430: Hm... synchronous cancelling and flushing out a possible apply message in the timer server is a bit unfortunate. When many frequent cancels happen, the performance of the timer server may degrade.

What do you think of this alternative approach:
* The `erlang:send_after` timer is still cancelled in the client process, but with `[{async, false}, {info, true}]`.
  * If the result is `<some-integer>`, the timer still existed and has been stopped before it expired. Nothing will be sent to the timer server, so we are done here.
  * If the result is `false`, we know that the timer has expired and that the apply message is in the timer server and cannot be stopped any more; but we _don't_ know if the apply message has already been processed. So the client sends a call like `sync` to the timer server, to which the timer server will simply reply with `ok`.
When the server receives such a `sync` message, all apply messages that came before it (from a just-expired timer) will have been processed because they arrived in the mailbox before the `sync`. When the reply arrives at the client, we know the action we could not stop any more has been performed, and so cancel is done here.

This approach would keep the timer server almost entirely out of the picture, all it may have to do is unconditionally reply to the `sync` message, and even that will only happen if the erlang timer could not be cancelled by the client.

rickard-green: @Maria-12648430 Sounds like a great solution!

bjorng: @Maria-12648430 I have replaced `%` with `%%` in your updated cancel code, re-indented, and force-pushed the updated commit. I intend to merge this pull request on Monday if everything goes well in our daily builds.

Maria-12648430: Oops, sorry, forgot about that 😅

rickard-green: Sorry I responded to fast :-( The BIF timer message is not guaranteed to be delivered to the server before the sync message is delivered since they originate from two different entities (see [signal doc](https://erlang.org/doc/reference_manual/processes.html#signal-delivery)) which will be a problem since the timer action might trigger after the cancel operation has returned.

juhlig: That complicates things... But I have an idea how it can be done. @Maria-12648430, meet on IRC tomorrow to discuss?

Maria-12648430: Not sure when I'm available... just mail me the details?

juhlig: Ok :)

Maria-12648430: @rickard-green @bjorng I discussed with @juhlig and implemented it like this:
* one-shot timers are now started from the timer server much like interval timers, and like those, they are registered in the timer table
* only timeout messages from timers in that table cause an action. Timeouts of one-shot timers cause removal from the table, timeouts of interval timers cause rescheduling and update.
* when a timer, one-shot or interval, is cancelled, it is removed from the table (if it still exists) and the erlang timer is stopped (if it still exists), so either the action has been performed already, or it will never be performed, no matter if and when the timeout message from the timer arrives

What do you think?

(And I hope I did the formatting right this time =^^=)

rickard-green: @Maria-12648430 Looks good!

I guess the special case where the timer is aimed at the caller of `timer:send_after()` could be optimized by handling setup/cancel in the client though.

Maria-12648430: Hm, I don't see how, off the top of my head ^^; What do you suggest? =^^=

Maria-12648430: Are this what you are suggesting?

```erlang
send_after(Time, Pid, Msg) when Pid =:= self() ->
    TRef = erlang:send_after(Time, Pid, Msg),
    {ok, {detached, TRef}};
...
cancel({detached, TRef}) ->
    _ = erlang:cancel_timer(TRef),
    {ok, cancel};
```

If yes, two things:
* there is a tiny change in behavior, don't know if it matters: if the timer server gets stopped or crashes, naturally, all timers will be stopped; _except_ those detached timers, they will keep running.
* why not do that for _all_ local pids (`when node(Pid) =:= node()`), not just for `self()`?

rickard-green: I'm wrong, it doesn't matter if it is aimed at yourself or not, but rather if it is a `send_after()` operation or not. Since the `tref()` type is opaque, no one is allowed to peek into it and the implementation is allowed to change it. Using a different tag, like for example `once_message`, when setting up a `send_after()` timer instead of `once`, `cancel()` can distinguish between `send_after()` timers and other timers and cancel the appropriate way.

rickard-green: I didn't see your latest reply before replying. About the behavior change, yes that is true. Without having put much thought at it: Perhaps the timer server should be placed under the kernel supervisor so that it brings down the whole node if crash? Whan do you think @bjorng?

Have to do some errands, so I wont be able to reply for a few hours.

rickard-green: Ugh, might the timer server be stopped :-( thats annoying...

Maria-12648430: @rickard-green It already is under `kernel_safe_sup`, or rather, it starts itself under it... and it can't be stopped in a easy straightforward way. So that's ok then?

Maria-12648430: @rickard-green well, it is not meant to be stopped. So that's probably ok ;)

bjorng: @rickard-green I think it's reasonable to place the timer server under the Kernel supervisor. I can't see any reason for why starting and stopping of the timer server should be supported.

Maria-12648430: I take that as a yes :)

juhlig: Should it be moved into the `kernel` application then?

bjorng: @juhlig We avoid moving modules from stdlib to kernel because it could cause various problems. The dets server is now under the kernel supervisor, but the dets modules are still in stdlib.

Maria-12648430: Last commit adds optimized `send_after` (and according `cancel`) for local pids.

juhlig: @bjorng I see, less work :) Then just change the `kernel_safe_sup` in `timer.erl` to `kernel_sup`?
Actually, `kernel_sup` is where it is placed if the timer server is started via kernel env param `start_timer` set to `true`. But otherwise, when the timer server is started dynamically, it is placed under `kernel_safe_sup`.

rickard-green: @Maria-12648430 yes, the module doesn't even have a `stop()` function, so I guess it should be rather uncontroversial change to not support stop of the server :-) @juhlig yes, using `kernel_sup` instead of `kernel_safe_sup` should be enough.

Maria-12648430: Good morning :) With the last commit, the timer server is started under `kernel_sup`.

@rickard-green @bjorng if you think everything is fine now, tell me and I'll squash it all together and augment the commit message accordingly.

bjorng: Looks good to *me*. I suggest that you wait for Rickard's answer before you squash.

Maria-12648430: Of course :)

rickard-green: Looks good to me too!

Maria-12648430:
The last commit squashes again the previous commits and updates the commit message.

bjorng:
I force-pushed a correction of the dialyzer test suite before the merge. (Reverting the previous changes, because the types are back to what they were.)

Thanks again for all your work on this pull request!

Maria-12648430:
Yay :) You're welcome

PR #4837, opened 2021-05-16
jinterface: Fix a NullPointerException in AbstractConnection

JeromeDeBretagne:
After a call to close(), `socket` is set to null which can lead to a NullPointerException in the `receive_loop` still handling tick messages [in AbstractConnection line 568](https://github.com/erlang/otp/blob/master/lib/jinterface/java_src/com/ericsson/otp/erlang/AbstractConnection.java#L568):
```
567                        synchronized (this) {
568                            OutputStream out = socket.getOutputStream();
569                            out.write(tock);
570                            out.flush();
571                        }
```

Cleanly end the `receive_loop` instead in this case. 

The crash was detected through future changes to the https://github.com/JeromeDeBretagne/erlanglauncher/ project, I have confirmed this change is fixing the issue as expected.

Thanks a lot

### DIFF
--- a/lib/jinterface/java_src/com/ericsson/otp/erlang/AbstractConnection.java
+++ b/lib/jinterface/java_src/com/ericsson/otp/erlang/AbstractConnection.java
@@ -1,7 +1,7 @@
 /*
  * %CopyrightBegin%
  *
- * Copyright Ericsson AB 2000-2017. All Rights Reserved.
+ * Copyright Ericsson AB 2000-2021. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -565,6 +565,11 @@ public abstract class AbstractConnection extends Thread {
                     // received tick? send tock!
                     if (len == 0) {
                         synchronized (this) {
+                            if (socket == null) {
+                                // protect from a potential thin race when the
+                                // connection to the remote node is closed
+                                throw new IOException("socket was closed");
+                            }
                             OutputStream out = socket.getOutputStream();
                             out.write(tock);
                             out.flush();


lib/jinterface/java_src/com/ericsson/otp/erlang/AbstractConnection.java
```diff
@@ -565,6 +565,11 @@ public void run() {
                     // received tick? send tock!
                     if (len == 0) {
                         synchronized (this) {
+                            if (socket == null) {
+                                // end the receive_loop once the connection
+                                // to the remote node is closed
+                                return;
```
sverker, line 571: It seems all other exit paths from the run() function either calls deliver() with an exception or throws an exception.

JeromeDeBretagne: The exceptions are even caught at the end of the `run()` method to call `deliver()` with an exception too. I had indeed seen that pattern, either with `Remote is sending garbarge` or `Remote has closed connection`.

The `deliver()` abstract method is described like this:
```
    /**
     * Deliver communication exceptions to the recipient.
     */
    public abstract void deliver(Exception e);
```
and the implementation in OtpCookedConnection like this:
```
    // pass the error to the node
    @Override public void deliver(final Exception e) {
```

This last case is different, this is a normal end triggered by a regular call to `close()` on the local node (and not an event triggered by the remote node as covered by all the other exit paths). I don't see any reason to report an error back to the node itself, while it is being closed cleanly, hence my proposition to simply return to end `run()` normally.

sverker: The new return added by this PR does not seem to be a "normal" case, but rather a thin race where the socket is closed _after_ a tick has been successfully read but _before_ attempting to send the reply "tock".

Wouldn't the normal local close case cause readSock() to fail with an exception.

JeromeDeBretagne: > The new return added by this PR does not seem to be a "normal" case, but rather a thin race 

I don't know if this is a race, but I can share from my (limited) experience that it is 100% reproducible once one hits it. In my code, I've got this NullPointerException when simply trying to add a final one-liner (OtpNode) `node.close()` call, and that's the only change I've made to trigger the crash.

> where the socket is closed after a tick has been successfully read but before attempting to send the reply "tock".

I see your assumption, but that's not what I'm detecting in my testing, it looks like the socket is first closed **then** a tick is still received successfully and finally it crashes. I can say this because the crash doesn't happen right away on close() but only after a 15-20 seconds for the tick message to arrive later on. 

To me, it seems the call to `getInputStream().read()` below simply blocks and succeeds in reading 4 bytes even after the call to `socket.close()` in the close() method:

```
     while (got < len) {
        i = s.getInputStream().read(b, got, len - got);

        if (i < 0) {
            throw new IOException("expected " + len
                    + " bytes, got EOF after " + got + " bytes");
        } else if (i == 0 && len != 0) {
            /*
            * This is a corner case. According to
            * http://java.sun.com/j2se/1.4.2/docs/api/ class InputStream
            * is.read(,,l) can only return 0 if l==0. In other words it
            * should not happen, but apparently did.
            */
            throw new IOException("Remote connection closed");
        } else {
            got += i;
        }
     }
     return got;
```

So in this case, readSock doesn't fail with an exception and simply return the received tick message.  

> Wouldn't the normal local close case cause readSock() to fail with an exception.

That's a good question, I guess it can depend on the types of sockets that are used and they may have different semantics, especially about how the above blocking read() call behaves. Let's keep in mind that the type of sockets (the transport classes in Jinterface) can be chosen to use a custom kind, I'm playing with Unix domain sockets for instance in addition to usual TCP ones. Is there a specific semantics expected or to be enforced ?

Anyway, if readSock was failing with an IOException as you suggest, it would then reach this catch block in run():
```
        } catch (final IOException e) {
            deliver(new OtpErlangExit("Remote has closed connection"));
```
which seems wrong as the connection was **not** closed by the remote node but by the local one. I still don't see any reason to send an exception back to a node that is closing itself cleanly, there should be a normal return in such a case. Or at least I would find this much more logical and easier to read. At the moment, there is no exit path in `run()` without an exception, that looks ... surprising.  

Do you have an approach to suggest to avoid this NullPointerException differently?

JeromeDeBretagne: @sverker I had a doubt earlier today, I've checked tonight : I'm getting the NullPointerException only when using my custom Unix Domain Sockets but not when using TCP sockets with the builtin OtpTransportFactory from Jinterface.

I've compared the 2 cases, now I better understand what's going on. With the builtin OtpTransportFactory, the receive loop ends with a "java.net.SocketException: Socket closed" triggered on line 1006 from :
```
        i = s.getInputStream().read(b, got, len - got);
```
This behavior is indeed the default for transports using subclasses of java.net.Socket class, I'm not sure if this is necessarily guaranteed for all other stream-oriented transports when calling close().

That was not the default behavior in my case using [LocalSocket from Android](https://developer.android.com/reference/android/net/LocalSocket) when calling [close()](https://developer.android.com/reference/android/net/LocalSocket#close()) but luckily enough there are two methods [shutdownInput()](https://developer.android.com/reference/android/net/LocalSocket#shutdownInput()) and [shutdownOutput()](https://developer.android.com/reference/android/net/LocalSocket#shutdownOutput()) to achieve the same behavior / semantics so I've been able to avoid the NullPointerException by updating my own custom transport `close()` method.

> Wouldn't the normal local close case cause readSock() to fail with an exception.

Thanks for this pointer, it was really useful. What do you think about this issue now, should I still try to make the code more robust somehow against potential a NullPointerException? Or is it a mandatory semantic for all transports so that read() should always raise an IOException when the socket is closed?

sverker: > I'm getting the NullPointerException only when using my custom Unix Domain Sockets but not when using TCP sockets with the builtin OtpTransportFactory from Jinterface.

That makes sense.

> What do you think about this issue now, should I still try to make the code more robust somehow against a potential NullPointerException? Or is it a mandatory semantic for all transports anyway so that read() should always raise an IOException when the socket is closed?

Well, the NullPointerException bug is still there even for the default TCP transport as a race. So I think you should keep your fix but change it to throw IOException to get the same behaviour as a failure in readSock().

About the issue "why deliver exception on a local close?"; I think that is a different topic and potential PR. But I think there could be a good argument about keeping backward "bug" compatilibity, as there could be exisiting user code waiting for that exception message. To troubleshoot events that does not happen is not fun.

JeromeDeBretagne: > So I think you should keep your fix but change it to throw IOException to get the same behaviour as a failure in readSock().

Sounds good to me, I will send an updated version.

> About the issue "why deliver exception on a local close?"; I think that is a different topic and potential PR. But I think there could be a good argument about keeping backward "bug" compatilibity

Indeed, backward compatibility is important, and this is not even a bug, more of a design choice.

> as there could be exisiting user code waiting for that exception message.

Yesterday I was thinking about keeping the exception but updating the message to something more generic like : "Connection closed" but I will refrain from proposing such a change then.

sverker:
Thank you, @JeromeDeBretagne.

JeromeDeBretagne:
> Thank you, @JeromeDeBretagne.

Thank you, @sverker

PR #4840, opened 2021-05-17
minor memory leaks fixes

max-au:
Running "erl -emu_type asan -eval "erlang:halt()." shows several minor issues straight away. First, init_misc_aux_work is called twice (and first allocation is made orphan), second, when shutting down, async threads don't clean up thread progress data, and - an expected leak, even documented - signal stack.

### DIFF
--- a/erts/emulator/asan/suppress
+++ b/erts/emulator/asan/suppress
@@ -1,8 +1,3 @@
-leak:erts_alloc_permanent_cache_aligned
-
-# Harmless leak of ErtsThrPrgrData from async threads in exiting emulator
-leak:erts_thr_progress_register_unmanaged_thread
-
 # Block passed to sigaltstack()
 leak:sys_thread_init_signal_stack
 
--- a/erts/emulator/beam/erl_async.c
+++ b/erts/emulator/beam/erl_async.c
@@ -368,7 +368,12 @@ static ERTS_INLINE void async_reply(ErtsAsync *a, ErtsThrQPrepEnQ_t *prep_enq)
 static void
 async_wakeup(void *vtse)
 {
-    erts_tse_set((erts_tse_t *) vtse);
+    /*
+     * 'vtse' might be NULL if we are called after an async thread
+     * has unregistered from thread progress prior to termination.
+     */
+    if (vtse)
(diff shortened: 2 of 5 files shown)


erts/emulator/sys/unix/sys_signal_stack.c
```diff
@@ -273,10 +273,9 @@ static void sys_sigaltstack(void *ss_sp) {
 /*
  * Set up alternate signal stack for an Erlang process scheduler thread.
  */
+static char sys_signalstack[SIGSTKSZ];
 void sys_thread_init_signal_stack(void) {
-    /* This will never be freed. */
-    char *stack = malloc(SIGSTKSZ);
-    sys_sigaltstack(stack);
+    sys_sigaltstack(sys_signalstack);
 }
```
rickard-green, line 279: This will lead to multiple threads (e.g. all schedulers) sharing the same signal stack which may cause a crash

max-au:
Would it work now, with `alloca` avoiding sharing signal buffer?

max-au:
I kept the suppression in sys_signal_stack.c, let's address it separately. I think now the PR is ready for review (I also rebased it on top of master branch for merge convenience).

max-au:
Is there any interest in merging this PR? If not, I'm fine with just closing it.

rickard-green:
> Is there any interest in merging this PR? If not, I'm fine with just closing it.

Yes, there is interest. I just forgot about this one, sorry.

rickard-green:
@max-au

I've added a commit that doesn't remove the `wakup` callback when unregistering from thread progress as well as making the async threads `wakup` callback prepared for a `NULL` argument. This since, otherwise an outstanding wakup request could trigger after the unregistration and cause an emulator crash. An alternative could be to be prepared for `NULL` wakeup callbacks instead of requiring the `wakup` callbacks of potentially unregistered threads being prepared for a `NULL` argument. `wakup` callbacks are however executed to a **much** higher degree than the async thread callback; therefore, I don't like such an approach.

Preferably you squash these commits.

max-au:
Thanks for noticing that, I completely missed the problem (due to lack of async threads usage, I think).
Squashed the commits and rebased to fresh maint branch.

PR #4843, opened 2021-05-17
Seal memory leak of jit_code_entry

sverker:
by making the `__jit_debug_descriptor` linked list thread safe with a mutex.

(@garazdawi @jhogberg What is `__jit_debug_descriptor` used for anyway?)


garazdawi:
It is used to inform gdb jit-reader about metadata of the erlang process stack. With the introduction of framepointers I think we can remove this code as it gdb then can do the unwinding for us.

jhogberg:
> It is used to inform gdb jit-reader about metadata of the erlang process stack. With the introduction of framepointers I think we can remove this code as it gdb then can do the unwinding for us.

I don't think we can, doesn't `gdb` need the symbols either way?

garazdawi:
> I don't think we can, doesn't gdb need the symbols either way?

Good point. Yes, it does, so no we cannot remove it.

PR #4850, opened 2021-05-19
Skip one extra uri_string:parse for httpc URL

sirihansen:
From profiling I see we are spending surprisingly much time in uri_string, and to me it seems that the current implementation of parsing and normalizing the URL in httpc does the parse+recompose twice. I cannot see any reason for this, so trying here to just return the map directly from uri_string:normalize.

All existing tests in httpc_SUITE pass, but I don't know if there are any tests that should be added? Any cornercases that might break from this change?

### DIFF
--- a/lib/inets/src/http_client/httpc.erl
+++ b/lib/inets/src/http_client/httpc.erl
@@ -501,10 +501,7 @@ service_info(Pid) ->
 %%% Internal functions
 %%%========================================================================
 normalize_and_parse_url(Url) ->
-    case uri_string:normalize(Url) of
-        {error, _, _} = Error -> Error;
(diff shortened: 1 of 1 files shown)


sirihansen:
A related question to this: would you consider a PR that allows the `Url` parameter to `httpc:request/x` even being a `uri_string:uri_map()`? From experiments it seems this already works, at least for the simple case, but it is not completely supported through the code and documentation.

The reasoning behind this would also be performance, since we today are spending time composing the Url from it's components, only to have httpc splitting and recomposing it again.

IngelaAndin:
@sirihansen sure as long as you also update the test and documentation ;)

PR #4854, opened 2021-05-20
Improve handling of exceptions from parse transforms

bjorng:
746fb27c38c7e1 (#4528) refactored handling of internal compiler
errors. An unintended consequence was that `exit` and `throw` exceptions
raised by parse transforms or core transforms would be reported as
internal compiler errors. As well as being confusing, the return value
from `compile:file/2` would be the atom `error` even when the `return_errors`
option was given. (The `return_errors` and `return_warnings` options are
intentionally ignored when the compiler has crashed in an unexpected way.)

Change the error handling so that exceptions in parse and core
transforms are reported as an error in the given transform. While at
it, also improve the formatting of the exception reasons and
stacktraces for transforms.

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -324,12 +324,14 @@ format_error({write_error, Error}) ->
 format_error({rename,From,To,Error}) ->
     io_lib:format("failed to rename ~ts to ~ts: ~ts",
 		  [From,To,file:format_error(Error)]);
-format_error({parse_transform,M,R}) ->
-    io_lib:format("error in parse transform '~ts': ~tp", [M, R]);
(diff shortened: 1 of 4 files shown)


zuiderkwast (approved):
LGTM

paulo-ferraz-oliveira:
Thanks, @bjorng.

In the meantime, my workaround had been to use `error` instead of `exit` and slightly tweak pattern matching. In any case, I didn't pursue this further since I saw in the official doc.s that parse_transforms are unsupported 😢: ```Programmers are strongly advised not to engage in parse transformations. No support is offered for problems encountered.``` (in https://erlang.org/doc/man/erl_id_trans.html#parse-transformations)

FYI: @elbrujohalcon, this is related to the recent `mixer` update.

PR #4863, opened 2021-05-20
Keep `compiler_generated` annotation in `sub_get_var`

Kuroneer:
In snippets like
```erlang
 foo(X) ->
     Y = ".",
     fun() -> X = Y ++ "." end(),
     ok.
```
the expression `Y ++ "."` builds a term that's unused when substituted
as the return expression in core erlang's compiler-generated clause
body.
The `compiler_generated` annotation guarantees that this substitution
won't generate a `"A term is constructed but never used"` warning.


bjorng:
Thanks for finding this bug and suggesting how to fix it. I have fixed the bug in #4899 because I wanted to do some refactoring to avoid code duplication and I also wanted the test case to be written in a different way (we avoid writing test cases using Core Erlang code, unless there is no other way to do it).

PR #4875, opened 2021-05-23
fix internal hash for external pids

max-au:
When 64-bit external pids were introduced in e9ab9a75a, internal
hash function was not adjusted accordingly. This led to a large
amount of hash collisions, ultimately making maps:from_list and
maps:from_keys to crash the emulator (stack overflow, or assertion
in a debug build).

### DIFF
--- a/erts/doc/src/erl_ext_dist.xml
+++ b/erts/doc/src/erl_ext_dist.xml
@@ -657,12 +657,12 @@
 	    <cell align="center">4</cell>
 	  </row>
 	  <row>
-	    <cell align="center"><c>89</c></cell>
+	    <cell align="center"><c>120</c></cell>
 	    <cell align="center"><c>Node</c></cell>
 	    <cell align="center"><c>ID</c></cell>
 	    <cell align="center"><c>Creation</c></cell>
 	  </row>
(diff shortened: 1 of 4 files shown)


sverker:
Good catch, @max-au. This is definitely a bug as the HAMT relies on the hashing using **every significant bit** into the mix for the hash value. Otherwise the hash collision of two different keys (external pids in this case) may never be resolved, no matter how much salt is poured.

sverker:
And we got the same bug for 64-bit ports a few code lines further down (EXTERNAL_PORT_SUBTAG).

sverker:
I'll fix the port bug and push to this PR branch, if that's ok.

max-au:
Sure, feel free to. Sorry for not noticing it, it was a separate commit I missed.

sverker:
Added fix for ports.
Also did correction for pids: External pids are 64-bit even on a 32-bit node as it may connect to 64-bit-nodes.
Moved tests from node_container_SUITE to map_SUITE.

sverker:
Squashed and rebased on OPT-24.0 for inclusion in next OTP-24.0.x.

sverker:
Merged to maint/master and put into pipe for OTP-24.0.x.

PR #4876, opened 2021-05-23
Remove autoconf from compiling instructions

v0idpwn:
This step is not needed as of #4835

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,6 @@ To compile Erlang from source, run the following commands. The complete building
 ```
 git clone https://github.com/erlang/otp.git
 cd otp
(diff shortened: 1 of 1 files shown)


garazdawi:
I think that since this step is needed for all old releases and it does not hurt when done in new releases, we should leave it for a while. That readme is what is shown when you navigate to the github page so I think a lot of people will use that as a guide on how to compile Erlang/OTP.

Should we add a comment saying that for recent versions that step is no longer needed?

v0idpwn:
While I presonally think that the master README should reflect the current version, I find your point 100% reasonable, @garazdawi. Closing this for now.

PR #4878, opened 2021-05-24
Clarify the rules for variable shadowing in comprehensions

### LINKED ISSUE #4862: Variable bindings in comprehensions
**Describe the bug**
I was playing with comprehensions involving multiple generators and I got some list/binary comprehension results which were news to me.  I am posting it here to initiate some discussion.

**To Reproduce**
```erl
Eshell V10.3.4  (abort with ^G)
1> [X || X <- [1,2,3], X <- [3,4,5]].
[3,4,5,3,4,5,3,4,5]
2> [X || X <- [1,2,3], X <- [a,b,c]].
[a,b,c,a,b,c,a,b,c]
3> [X || X <- [1,2,3], Y=X <- [a,b,c]].
[a,b,c,a,b,c,a,b,c]
4> [{X,Y} || X <- [1,2,3], X=Y <- [a,b,c]].
[{a,a},{b,b},{c,c},{a,a},{b,b},{c,c},{a,a},{b,b},{c,c}]
```

**Expected behavior**
The above surprised me; I understand why it happens, of course, but I still find it surprising, esp. since in Erlang it's allowed for the term X in the "variable" position of a generator (X <- L) to be an already bound term and thus use it as a filter.

**Affected versions**
Probably all.

**Additional context**
The same behavior exists in compiled code, but at least there one gets warnings (a weird one about "an unused variable" and another one about shadowing that at least makes some sense).

bjorng:
Closes #4862

### DIFF
--- a/system/doc/programming_examples/list_comprehensions.xml
+++ b/system/doc/programming_examples/list_comprehensions.xml
@@ -190,6 +190,12 @@ select(X, L) ->  [Y || {X1, Y} <- L, X == X1].]]></code>
     <pre>
 > <input>select(b,[{a,1},{b,2},{c,3},{b,7}]).</input>
 [2,7]</pre>
+    <p>Also note that a variable in a generator pattern will shadow
+    a variable with the same name bound in a previous generator
+    pattern. For example:</p>
+    <pre>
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
LGTM

PR #4880, opened 2021-05-24
debug emulator build: clean up unused assertions

max-au:
Two asserts can be safely removed. One asserts that code size is
greater than zero (which is not necessarily so), and another one
verifies that dual mapping is enabled (and it's not if we enable
Linux Perf support in either dump or map mode)

### DIFF
--- a/erts/emulator/beam/jit/x86/beam_asm.cpp
+++ b/erts/emulator/beam/jit/x86/beam_asm.cpp
@@ -809,8 +809,6 @@ extern "C"
 
     void beamasm_purge_module(const void *native_module_exec,
                               void *native_module_rw) {
-        ASSERT(native_module_exec != native_module_rw);
-
         jit_allocator->release(const_cast<void *>(native_module_exec));
     }
 
--- a/erts/emulator/beam/jit/x86/beam_asm_perf.cpp
+++ b/erts/emulator/beam/jit/x86/beam_asm_perf.cpp
@@ -137,7 +137,7 @@ public:
         for (BeamAssembler::AsmRange &r : ranges) {
             size_t nameLen = r.name.size();
             ptrdiff_t codeSize = (char *)r.stop - (char *)r.start;
-            ASSERT(codeSize > 0);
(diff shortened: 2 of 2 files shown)


erts/emulator/beam/jit/x86/beam_asm_perf.cpp
```diff
@@ -137,7 +137,6 @@ class JitPerfDump {
         for (BeamAssembler::AsmRange &r : ranges) {
             size_t nameLen = r.name.size();
             ptrdiff_t codeSize = (char *)r.stop - (char *)r.start;
-            ASSERT(codeSize > 0);
```
garazdawi, line 140: This assert was added because perf does not like it when the codeSize is a negative number and we had mistakes where it was. Have you found any cases where this assert triggers for negative numbers, or only for when codeSize == 0?

I think it would be better to just change it to `codeSize >= 0`.

max-au: I've only seen zero. Changing to "codeSize >=0" works for me too, I'll update the PR.

erts/emulator/beam/jit/x86/beam_asm.cpp
```diff
@@ -800,8 +800,6 @@ extern "C"
 
     void beamasm_purge_module(const void *native_module_exec,
                               void *native_module_rw) {
-        ASSERT(native_module_exec != native_module_rw);
```
max-au, line 803: For  this one, I can also wrap it in ifdef/endif for `HAVE_LINUX_PERF_SUPPORT` and dump/map support enabled, although it seem to complicate the assert unnecessarily.

garazdawi: I agree that we can remove it, maybe it will be added later if we get into problems with it again.

garazdawi:
Would you mind rebasing on maint?

garazdawi:
Thanks!

PR #4886, opened 2021-05-26
etp-commands: various small fixes

max-au:
Fixed etp-cons and etpf-cons printing.
Added dirty CPU/IO schedulers and poll thread to etp-thr.py.
Fixed outdated comments

### DIFF
--- a/erts/etc/unix/etp-commands.in
+++ b/erts/etc/unix/etp-commands.in
@@ -67,7 +67,7 @@ document etp-help
 %   etp-overlapped-heaps, etp-chart, etp-chart-start, etp-chart-end
 %
 % System inspection
-%   etp-system-info, etp-schedulers, etp-process, etp-ports, etp-lc-dump,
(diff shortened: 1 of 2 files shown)


max-au:
It appears that a few more commands do not work:
* etp-msgq: "Attempt to take address of value not located in memory."
* etp-ets-tabledump 23% Dumping ETS table 23:No symbol "meta_main_tab" in current context.
* etp-disasm: <net_kernel:handle_info/2+0x17a8>: No symbol "beam_ops" in current context.

garazdawi:
Maybe it is time to start writing tests for the `etp` commands...

> etp-disasm: <net_kernel:handle_info/2+0x17a8>: No symbol "beam_ops" in current context.

This one has always been a bit unstable and now that the JIT is here it is completely broken, so I think we can remove it.

garazdawi:
thanks!

PR #4830, opened 2021-05-14
Init `socket` module on `kernel` start

hauleth:
`persistent_term` is cleaned on restart which causes the failures when using `kernel.inet_backend = socket` as the `erl_epmd` module cannot create connection. While this can slow down subsequent startups of `kernel` application it ensures that the proper set of the protocols and options is loaded to the `persistent_term` after restarts.

Close #4827

PR #4888, opened 2021-05-26
Re-init socket tables during init:restart

RaimoNiskanen:
A counter-PR for #4830.

I moved code around a bit so the re-initialization gets done as early in system restart as in boot to make sockets usable early.  I also made the call to `prim_socket:init/0` conditional as the one to `prim_init:on_load/0`.


RaimoNiskanen:
I have rebased to be able to release this as a patch before OTP-24.1, and therefore force pushed.

PR #4890, opened 2021-05-26
Fix lost line numbers in Dialyzer warnings for Elixir code

michallepicki:
Elixir code often generates Erlang forms, e.g. literals
with Location set to 0. A change in Dialyzer warning generation
caused those 0 Locations to be picked up and shown
to the user even if there are valid line numbers elsewhere
in the tree. As a fix, when reading Location from the tree,
ignore all line numbers lower than 1.

### DIFF
--- a/lib/dialyzer/src/dialyzer_utils.erl
+++ b/lib/dialyzer/src/dialyzer_utils.erl
@@ -1008,7 +1008,7 @@ get_all_locations(Tree) ->
   ++
   lists:append([get_all_locations(T) || T <- SubTrees]).
 
-maybe_get_location([Line|_]) when is_integer(Line) ->
+maybe_get_location([Line|_]) when is_integer(Line), Line > 0 ->
   Line;
 maybe_get_location([{Line, Column}|_Tail]) when is_integer(Line),
                                                 is_integer(Column) ->


michallepicki:
From my knowledge, Erlang 24+ is affected. Here is an example warning that "lost" the line number when switching to OTP 24.0.1, with some more details: https://github.com/michallepicki/elixir-lang-dialyzer-runs/pull/11#issuecomment-848044785

michallepicki:
@uabboli I ran the `dialyzer` application tests suite locally and it passed, hopefully that is all. I have not added a test for this change since a proper integration test with Elixir compiler would be hard to set up, but do you think it is a good idea to add a "unit" test that calls `dialyzer_utils:get_location` with some forms with a `0` somewhere inside the tree?

Overall, I am also not sure if this is the correct "fix". The [`location()` type](https://github.com/erlang/otp/blob/009f17d4b7cbcf4e21ec28a8c66a0ca1adbaf986/lib/stdlib/src/erl_anno.erl#L65) is a `non_neg_integer()`, so 0 should be valid, but for the user, I think files are 1-indexed. Was the intention to have `0` be a value that can be set by generated code, e.g. by Elixir macros? This shows up for Elixir code that is not necessarily generated, so maybe the Elixir compiler should do a better job setting proper location for all forms?

uabboli:
> I ran the dialyzer application tests suite locally and it
> passed, hopefully that is all.

Yes, I think so.

> The location() type is a non_neg_integer()

Note that the location() type is used in erl_anno, which is tightly
coupled with erl_scan, erl_parse and the abstract format. Dialyzer analyzes the
Core Erlang format, where annotations are not of type erl_anno:anno().

I think a proper fix is to move
dialyzer_dataflow:is_compiler_generated() (and get_line()) to
dialyzer_utils and use it in dialyzer_utils:get_all_locations().

Thanks for pointing this out. I'll create a ticket, label it as a
'bug', close your PR, and create a new PR for you to try out.

michallepicki:
> Thanks for pointing this out. I'll create a ticket, label it as a 'bug', close your PR, and create a new PR for you to try out.

Sounds great, thank you!

michallepicki:
> I think a proper fix is to move dialyzer_dataflow:is_compiler_generated() (and get_line()) to dialyzer_utils and use it in dialyzer_utils:get_all_locations().

I am not sure it will fix it since AFAIK [this atom](https://github.com/elixir-lang/elixir/blob/6cacc65d7df5c0f3f006581071c6a5cfedff1b4c/lib/elixir/lib/dynamic_supervisor.ex#L351) is not marked as generated, it just has 0 instead of a proper location, but I may be wrong here.

PR #4891, opened 2021-05-26
ssl: Remove duplicate option documentation

### LINKED ISSUE #4873: ssl: doc `honor_cipher_order` mentioned twice in the same options list
Here: https://erlang.org/doc/man/ssl.html#type-server_option.

As both `boolean()` and `honor_cipher_order()` (which is `boolean()`).

IngelaAndin:
Closes #4873

### DIFF
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -445,7 +445,6 @@
                                 {alpn_preferred_protocols, server_alpn()} |
                                 {next_protocols_advertised, server_next_protocol()} |
                                 {psk_identity, server_psk_identity()} |
-                                {honor_cipher_order, boolean()} |
                                 {sni_hosts, sni_hosts()} |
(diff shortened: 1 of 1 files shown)


lib/ssl/src/ssl.erl
```diff
@@ -445,7 +445,6 @@
                                 {alpn_preferred_protocols, server_alpn()} |
                                 {next_protocols_advertised, server_next_protocol()} |
                                 {psk_identity, server_psk_identity()} |
-                                {honor_cipher_order, boolean()} |
```
paulo-ferraz-oliveira, line 448: I was unsure this change alone would remake the doc. which is why I didn't do it myself 😄 Didn't want to go after doc. generation issues, as I have in the past 😄

IngelaAndin: Sympathize with that ;)

PR #4893, opened 2021-05-26
dialyzer: Do not expose line number 0 in message locations

uabboli:
See also GH-4890: https://github.com/erlang/otp/pull/4890.

Commit 04b905d assigned better start locations to messages, but also
sometimes let line number 0 slip through. This is now corrected.

Closes GH-4890.


michallepicki (approved):

michallepicki:
Am I right that this was not released in 24.0.2? Can I ask why?

uabboli:
I didn't think of including the fix in a patch.

We tend to keep patches as small as possible and displaying wrong line number is a minor thing. But the modification is harmless, and I'll queue the fix for the next patch of Erlang/OTP 24.0.

Issue #4846, opened 2021-05-18
OTP-24.0 is miscompiled by gcc's LTO


mikpe:
**Describe the bug**
When `gcc`'s link-time optimizations are enabled, via `-flto`, OTP-24.0 and current master are miscompiled, which shows itself as
an internal error in beam_kernel_to_ssa as the new VM invokes its `erlc` for the first time:

> make[1]: Leaving directory '/tmp/otp.git/erts'
> cd lib && \
>   ERL_TOP=/tmp/otp.git PATH=/tmp/otp.git/bootstrap/bin:"${PATH}" \
>         make opt SECONDARY_BOOTSTRAP=true ERL_COMPILE_WARNINGS_AS_ERRORS=yes
> make[1]: Entering directory '/tmp/otp.git/lib'
> make[2]: Entering directory '/tmp/otp.git/lib/parsetools'
> === Entering application parsetools
> make[3]: Entering directory '/tmp/otp.git/lib/parsetools/src'
> erlc -W  -Werror +debug_info -DUSE_ESOCK=true -I/tmp/otp.git/lib/stdlib/include -Werror -o../ebin leex.erl
> Function: compile/3
> leex.erl: internal error in pass beam_kernel_to_ssa:
> exception error: bad key: 1
>   in function  map_get/2
> ...

**To Reproduce**
Build current maint-24 or master with `CFLAGS="-O2 -flto"` and gcc-10.3. I see the error with `gcc version 10.3.1 20210422 (Red Hat 10.3.1-1) (GCC)` on Fedora 32 and Fedora 33. I have not tested other GCC versions yet, but it is persistent with this one.

**Expected behavior**
I expected the newly built VM to manage to run its compiler to build its Erlang modules.

**Affected versions**
OTP-24.0 and current master are affected. Current versions of OTP-23 and OTP-22 are _not_ affected.

**Additional context**
We build Erlang/OTP as an RPM, and `-flto` is added by default from `/usr/lib/rpm/redhat/macros`. We can work around the error by adding `%define _lto_cflags %{nil}` to our RPM `.spec` file, presumably trading a performance loss for correctness.

The build with `-flto` shows a scary warning:

> beam/erl_driver.h:331:12: warning: type of 'driver_select' does not match original declaration [-Wlto-type-mismatch]
  331 | EXTERN int driver_select(ErlDrvPort port, ErlDrvEvent event, int mode, int on);
      |            ^
[...]

PR #4894, opened 2021-05-26
Fix maps internal sorting bug for gcc flag "-flto"

sverker:
Closes GH-4846.

### DIFF
--- a/erts/emulator/beam/beam_debug.c
+++ b/erts/emulator/beam/beam_debug.c
@@ -1313,27 +1313,39 @@ ms_wait(Process *c_p, Eterm etimeout, int busy)
  */
 
 UWord
-erts_check_stack_recursion_downwards(char *start_c)
+erts_check_stack_recursion_downwards(char *start_c, char* prev_c)
 {
     char *limit = ERTS_STACK_LIMIT;
     char c;
     UWord res;
+
+    if (prev_c == &c) {
+         /* Protect against eternal loop if C compiler do tail call
+            optimization. This may also prevent such optimizations. */
+        return 0;
+    }
     if (erts_check_below_limit(&c, limit + 1024))
         return (char *) erts_ptr_id(start_c) - (char *) erts_ptr_id(&c);
-    res = erts_check_stack_recursion_downwards(start_c);
+    res = erts_check_stack_recursion_downwards(start_c, &c);
     erts_ptr_id(&c);
     return res;
 }
 
 UWord
-erts_check_stack_recursion_upwards(char *start_c)
+erts_check_stack_recursion_upwards(char *start_c, char* prev_c)
 {
     char *limit = ERTS_STACK_LIMIT;
     char c;
     UWord res;
+
+    if (prev_c == &c) {
+         /* Protect against eternal loop if C compiler do tail call
+            optimization. This may also prevent such optimizations. */
+        return 0;
+    }
     if (erts_check_above_limit(&c, limit - 1024))
         return (char *) erts_ptr_id(&c) - (char *) erts_ptr_id(start_c);
-    res = erts_check_stack_recursion_upwards(start_c);
+    res = erts_check_stack_recursion_upwards(start_c, &c);
     erts_ptr_id(&c);
     return res;
 }
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -4102,9 +4102,9 @@ BIF_RETTYPE erts_debug_get_internal_state_1(BIF_ALIST_1)
             UWord size;
             char c;
             if (erts_is_above_stack_limit(&c))
-                size = erts_check_stack_recursion_downwards(&c);
+                size = erts_check_stack_recursion_downwards(&c, &c);
             else
-                size = erts_check_stack_recursion_upwards(&c);
+                size = erts_check_stack_recursion_upwards(&c, &c);
 	    if (IS_SSMALL(size))
 		BIF_RET(make_small(size));
 	    else {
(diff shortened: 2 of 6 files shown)


mikpe:
I've tested this branch (merged into maint) with gcc-10.3 (Fedora 32) and gcc-11.1 (Fedora 34). The good news is that with this branch the build with `-flto` succeeds. The bad news is that the `emulator` suite hangs in tight CPU-consuming loops (`gdb` backtrace points to `select()`).

sverker:
I just tried the same on Ubuntu 18 with gcc-10.3 and it hangs spinning in test case erts_debug_SUITE:stack_check. Is is the same for you?

sverker:
Ok, the test case `stack_check` is suppose to do body recursive calls until native stack limit is reached. It does an inter-object-file function call to dummy function erts_ptr_id() to prevent last call optimization. But with -flto gcc is not fooled by that, making it an endless loop as the stack pointer never changes.

sverker:
Test case `erts_debug_SUITE:stack_check` was introduced in its current state in OTP-20.0.

mikpe:
> I just tried the same on Ubuntu 18 with gcc-10.3 and it hangs spinning in test case erts_debug_SUITE:stack_check. Is is the same for you?

I _think_ so. Since it hangs I have to ^C it, so I get incomplete results, but it did seem to point to erts_debug_SUITE as being the last thing run.

mikpe:
> Ok, the test case `stack_check` is suppose to do body recursive calls until native stack limit is reached. It does an inter-object-file function call to dummy function `erts_ptr_id()` to prevent last call optimization. But with `-flto` gcc is not fooled by that, making it an endless loop as the stack pointer never changes.

Good find. I'll take a look tomorrow, but a strategically-placed `asm()` might cure that, or an `attribute(noinline,noclone)` (but I don't know how well that one works with `-flto`).

mikpe:
This fixed the `emulator` suite for me:
```
diff --git a/erts/emulator/beam/utils.c b/erts/emulator/beam/utils.c
index 1657431875..c5d36949d1 100644
--- a/erts/emulator/beam/utils.c
+++ b/erts/emulator/beam/utils.c
@@ -5104,9 +5104,12 @@ erts_check_above_limit(char *ptr, char *limit)
     return ptr > limit;
 }
 
-void *
+void * ERTS_NOINLINE
 erts_ptr_id(void *ptr)
 {
+#ifdef __GNUC__
+    asm("" : "+r" (ptr), "+m" (* (char *) ptr) : "0" (ptr));
+#endif
     return ptr;
 }
 
```

sverker:
So, the asm is an opaque dummy thing just make gcc to back off and don't optimize away the function?

sverker:
I get the following from gcc-10.3:
```
beam/utils.c: In function ‘erts_ptr_id’:
beam/utils.c:5121:5: error: inconsistent operand constraints in an ‘asm’
 5121 |     asm("" : "+r" (ptr), "+m" (* (char *) ptr) : "0" (ptr));
      |     ^~~
```

mikpe:
> So, the asm is an opaque dummy thing just make gcc to back off and don't optimize away the function?

It's there to ensure there is a USE of the `&c` being passed in, and a DEF for the return value, without gcc realizing it's the identity function.

mikpe:
> I get the following from gcc-10.3:
> 
> ```
> beam/utils.c: In function ‘erts_ptr_id’:
> beam/utils.c:5121:5: error: inconsistent operand constraints in an ‘asm’
>  5121 |     asm("" : "+r" (ptr), "+m" (* (char *) ptr) : "0" (ptr));
>       |     ^~~
> ``

What platform is that on, and with what compiler flags?

sverker:
> 
> What platform is that on, and with what compiler flags?

It's Ubuntu 18 gcc-10.3.0. It looks like I get it only for debug build, when passing -O0. Here is my full compile command line:

`gcc-10 -O0 -Werror=undef -Werror=implicit -Werror=return-type  -fno-common -g -fomit-frame-pointer -I/home/sverker/src/otp4/erts/x86_64-pc-linux-gnu    -D_GNU_SOURCE  -DHAVE_CONFIG_H -Wall -Wstrict-prototypes -Wpointer-arith -Wmissing-prototypes -Wdeclaration-after-statement -DUSE_THREADS -D_THREAD_SAFE -D_REENTRANT -DPOSIX_THREADS -D_POSIX_THREAD_SAFE_FUNCTIONS  -DERLANG_GIT_VERSION="\"d8d345b390\"" -DBEAMASM=1 -Ix86_64-pc-linux-gnu/opt/jit -Ibeam -Isys/unix -Isys/common -Ix86_64-pc-linux-gnu -Izlib  -Ipcre -I../include -I../include/x86_64-pc-linux-gnu -I../include/internal -I../include/internal/x86_64-pc-linux-gnu -Ibeam/jit -Ibeam/jit/x86 -c beam/utils.c -o obj/x86_64-pc-linux-gnu/opt/jit/utils.o`

If I change `-O0` to `-O2` (or anything at all) it goes away. Strange.

frej:
Gcc won't do the tail call optimization if the function has the `__attribute__((noreturn))` attribute, see [this GCC wontfix bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=10837). If Clang does the same, this could be an alternative to the inline assembly.

sverker:
Added some more C code that protects against tail call optimization but also seem to prevent it.

mikpe:
This one (merged onto maint) passes the emulator suite fine for me when compiled with `CFLAGS="-O2 -flto"` (Fedora 33, gcc-10.3.1). No `asm()` needed apparently :-)

PR #4895, opened 2021-05-26
Tune delays in user_sup

richcarl:
Discovered while investigating startup and shutdown times:
 - The shell is delayed 100 ms because of the sleep time in the loop that checks if the user driver is started. By checking every 1 ms instead, this is more or less eliminated. Also, waiting a full 10 seconds until giving up seems way too much, so I lowered it to 5.
 - Waiting a whole second before killing the user driver seems excessive, since this holds up the whole shutdown sequence. I'm not sure what's a good number these days or if there are better solutions, but lowering it to 10 ms makes it less of a bother.
 - It seems brutal to terminate the system with halt() if running as slave node when the master user driver cannot be found. Changed to returning error just like in the normal case.

These patches are mainly meant as suggestions.

### DIFF
--- a/lib/kernel/src/user_sup.erl
+++ b/lib/kernel/src/user_sup.erl
@@ -43,8 +43,12 @@ init([]) ->
 	nouser ->
 	    ignore;
 	{master, Master} ->
-	    Pid = start_slave(Master),
-	    {ok, Pid, Pid};
+	    case start_slave(Master) of
+		{ok, Pid} ->
+		    {ok, Pid, Pid};
+		Error ->
+		    Error
+	    end;
 	{M, F, A} ->
 	    case start_user(M, F, A) of
 		{ok, Pid} ->
@@ -57,11 +61,10 @@ init([]) ->
 start_slave(Master) ->
     case rpc:call(Master, erlang, whereis, [user]) of
 	User when is_pid(User) ->
-	    spawn(?MODULE, relay, [User]);
+	    {ok, spawn(?MODULE, relay, [User])};
 	_ ->
 	    error_logger:error_msg("Cannot get remote user", []),
-	    receive after 1000 -> true end,
-	    halt()
+	    {error, nouser}
     end.
 
 -spec relay(pid()) -> no_return().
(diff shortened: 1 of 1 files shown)


lib/kernel/src/user_sup.erl
```diff
@@ -108,7 +111,8 @@ wait_for_user_p(N) ->
 	    link(Pid),
 	    {ok, Pid};
 	_ ->
-	    receive after 100 -> ok end,
+            %% minimal sleep so the shell is not delayed more than needed
+	    receive after 1 -> ok end,
```
max-au, line 115: I wonder if there is a way to make 'user' send a reply back instead of busy-wait loop.

richcarl: Yeah, I don't like the "wait for name to become registered" method. Adding proper synchronization would be more elegant, but obviously a bigger change, and I wasn't sure if there was a reason somewhere for not relying on the user process to send an ack.
I'm not particularly worried about the busy-waiting though. A millisecond is still a fairly long time in this context.

garazdawi: When `user` is called (i.e. when a non-interactive shell is started) the name should always be registered when the `apply` returns as far as I can tell. However, for interactive shells `user_drv` is called which registers the `user` name asynchronously which is why the spinning is needed. At least that is how I read the code.

@richcarl Did you find the delay also in non-interactive shells?

richcarl: No, there is no delay when using -noshell. This is still worth doing something about, because you can feel those extra 100 ms between typing 'erl' and getting the interactive shell prompt.

richcarl: @garazdawi And yes, the user_drv process is spawned off and does its own registration, whereas the user.erl start code spawns the process and does the registration before returning. If there's no particular reason for letting these behave as they like, it would be straighforward to make both do a synchronized start.

garazdawi: > If there's no particular reason for letting these behave as they like, it would be straighforward to make both do a synchronized start.

No reason that I can think of. We already synchronize today by spinning on the name, so might as well make the sync more explicit by making `user_drv` send an ack when it knows that `user` exists.

garazdawi:
Fixed in #6144

PR #4896, opened 2021-05-26
lists: takeuntil/2 and dropuntil/2

Maria-12648430:
This PR provides two functions similar to `takewhile/2` and `dropwhile/2` for the `lists` module, but with the following differences:

`takewhile/2` will return a list of all elements from the beginning of the input list up to the point where the predicate returns `false` for the first time, _excluding this element_ (it is not taken)
`takeuntil/2` will return a list of all elements from the beginning of the input list up to the point where the predicate returns `true` for the first time, _including this element_ (it is taken)

`dropwhile/2` will return a list with all elements removed from the beginning of the input list up to the point where the predicate returns `false` for the first time, _including this element_ (it is not dropped)
`dropuntil/2` will return a list with all elements removed from the beginning of the input list up to the point where the predicate returns `true` for the first time, _excluding this element_ (it is dropped)

The reversal of the predicate result is necessary to "chime" with the chosen name(s), ie `takewhile`/`dropwhile` --> "while ... is true" vs `takeuntil`/`dropuntil` --> "until ... is true".

Another function that could use a companion of this sort is `splitwith/2`, but here I'm really at a loss for a good name 😓

A use case I often come across is taking/dropping things from a list until some sort of special item is found, and the need to get (or remove) this item, too. Removing another item from a `dropwhile/2` result is easy via `tl/1`, but one has to make sure that the list is not empty. Adding another to the result from a `takewhile/2` is hard.

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -38,8 +38,8 @@
 
 -export([all/2,any/2,map/2,flatmap/2,foldl/3,foldr/3,filter/2,
 	 partition/2,zf/2,filtermap/2,
-	 mapfoldl/3,mapfoldr/3,foreach/2,takewhile/2,dropwhile/2,
-         search/2, splitwith/2,split/2,
+	 mapfoldl/3,mapfoldr/3,foreach/2,takewhile/2,takeuntil/2,
+	 dropwhile/2, dropuntil/2, search/2, splitwith/2,split/2,
 	 join/2]).
 
 %%% BIFs
@@ -1390,6 +1390,19 @@ takewhile(Pred, [Hd|Tail]) ->
     end;
 takewhile(Pred, []) when is_function(Pred, 1) -> [].
 
+-spec takeuntil(Pred, List1) -> List2 when
+      Pred :: fun((Elem :: T) -> boolean()),
+      List1 :: [T],
+      List2 :: [T],
+      T :: term().
+
+takeuntil(Pred, [Hd|Tail]) ->
+    case Pred(Hd) of
+	true -> [Hd];
+	false -> [Hd|takeuntil(Pred, Tail)]
+    end;
+takeuntil(Pred, []) when is_function(Pred, 1) -> [].
+
 -spec dropwhile(Pred, List1) -> List2 when
       Pred :: fun((Elem :: T) -> boolean()),
       List1 :: [T],
@@ -1402,6 +1415,19 @@ dropwhile(Pred, [Hd|Tail]=Rest) ->
 	false -> Rest
     end;
 dropwhile(Pred, []) when is_function(Pred, 1) -> [].
+
+-spec dropuntil(Pred, List1) -> List2 when
+      Pred :: fun((Elem :: T) -> boolean()),
+      List1 :: [T],
+      List2 :: [T],
+      T :: term().
+
+dropuntil(Pred, [Hd|Tail]) ->
+    case Pred(Hd) of
+	true -> Tail;
+	false -> dropuntil(Pred, Tail)
+    end;
+dropuntil(Pred, []) when is_function(Pred, 1) -> [].
 
 -spec search(Pred, List) -> {value, Value} | false when
       Pred :: fun((T) -> boolean()),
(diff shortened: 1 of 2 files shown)


Maria-12648430:
On second thought, I have another idea that would serve my purpose just as well if not better 🙂 I'll close this PR and be back with another tomorrow.

zuiderkwast:
It looks a bit like Perl (where you have *until* and *do-until* loops, which are simply negated *while*).

Why do the *until* versions include the first mismatching element? I think including it is orthogonal to negating the predicate. It's like if Perl's until loop would execute one more time after the condition becomes true... which it doesn't. So the analogy can become a source of confusion for anyone coming from Perl.

> Adding another to the result from a `takewhile/2` is hard.

```Erlang
inclusive_takewhile(Pred, List) ->
    case lists:splitwith(Pred, List) of
        {_, []} -> List;
        {Take, [FirstDrop|_]} -> Take ++ [FirstDrop]
    end.
```

The inclusive takewhile above is less efficient due to the ++ of course.

If inclusive versions are really wanted, how about instead adding `takewhile(Pred, List, Options)` where `inclusive` is the only option?

Maria-12648430:
> It looks a bit like Perl (where you have _until_ and _do-until_ loops, which are simply negated _while_).

I had to look it up, not using Perl on a regular basis ;) `takeuntil` and `dropuntil` were meant in the sense of `do... until` (with "do" being a specific action here, `take` or `drop`), ie first do the action (`take`/`drop`), _then_ perform the check.

> Why do the _until_ versions include the first mismatching element? I think including it is orthogonal to negating the predicate. It's like if Perl's until loop would execute one more time after the condition becomes true... which it doesn't. So the analogy can become a source of confusion for anyone coming from Perl.

If read as `do... until` (vs just `until`), the analogy fits I think ;)

> > Adding another to the result from a `takewhile/2` is hard.
> 
> ```erlang
> inclusive_takewhile(Pred, List) ->
>     case lists:splitwith(Pred, List) of
>         {_, []} -> List;
>         {Take, [FirstDrop|_]} -> Take ++ [FirstDrop]
>     end.
> ```
> 
> The inclusive takewhile above is less efficient due to the ++ of course.

Ok, probably "hard" wasn't the right word for what I wanted to say ;) What I meant was sth along the lines of "... unless resorting to stunts like the above" ;)

> If inclusive versions are really wanted, how about instead adding `takewhile(Pred, List, Options)` where `inclusive` is the only option?

I refrained from that because all the other functions in the `lists` module come without processing options variants like this, they are pretty fixed in their behavior. And the use cases for this are probably too few to justify breaking the pattern ;)

Anyway, I realized that this is not what I really wanted... `splitwith` probably suffices for most cases. I think what I would like somewhat on top of `splitwith` is to not split in two, but split into multiple lists whenever the predicate returns `true` to say "cut here". This can be emulated with successive calls of `splitwith` on the right lists, only you have to remove the first element each time to use the same predicate. There might be a more general function that can do this, a kind of fold with look-ahead/-behind ability, that may be generally more useful still. I'm still undecided =^^=

zuiderkwast:
A *do-until* and *do-while* loop (like do-while in C) always runs the first time, but it stops when the condition is met. It doesn't run one extra round after the condition is met, so the analogy is close but not exactly right. Also, it's the *do* that does this, not the *until* or *while*. :-)

I guess `lists:partition/2` is also not what you need.

If you want look ahead and look behind, be aware that this too comes from Perl (at least the regexp ones). I think it's a sign. You may become a [Perl monger](https://www.pm.org/) or a [Perl monk](https://www.perlmonks.org/). The Perl community is funny. They have proverbs like "There is More Than One Way To Do It (TIMTOWTDI)" or ["There’s more than one way to do it, but sometimes consistency is not a bad thing either (TIMTOWTDIBSCINABTE)", pronounced Tim Toady Bicarbonate](https://en.wikipedia.org/wiki/TIMTOWTDIBSCINABTE).

Jokes aside, fold is the most generic of list functions, but even more generic is to just do what you want in a recursive function. Then you can stop traversing the list whenever you want, etc.

PR #4899, opened 2021-05-27
Eliminate incorrect warning for useless building

bjorng:
In OTP 24, there will be a warning for the following code:

    foo(X) ->
        Y = ".",
        fun() ->
                X = Y ++ "."  %Warning: a term is constructed, but never used
        end(),
        ok.

The warning is incorrect, because the built term will be matched to
the previous value of `X`. This commit eliminates the warning.

To explain the bug, I will illustrate how the example program is transformed
by the Core Erlang passes. My illustrative examples will use Erlang instead
of Core Erlang.

The first transformation makes the match `X = Y ++ "."` explicit:

    foo(X) ->
        Y = ".",
        fun() ->
                X1 = Y ++ ".",
                case X1 =:= X of
                    true ->
                        X1;
                    false ->
                        error({badmatch,X})
                end
        end(),
        ok.

The variable `X1` is only needed for the matching. It will be
annotated with a `compiler_generated` annotation to ensure that no
compiler warnings will be generated if it would later turn out that it
would never be used.

Next constants are propagated and constant expressions are evaluated.
Since the fun is only used once, it will be eliminated and its body
inlined. Now we have:

    foo(X) ->
        case ".." =:= X of
            true ->
                "..";
            false ->
                error({badmatch,X})
        end,
        ok.

The final optimization for this example is simplification of expressions
whose values are never used:

    foo(X) ->
        case ".." =:= X of
            true ->
                ok;  %Warning: a term is constructed, but never used
            false ->
                error({badmatch,X})
        end,
        ok.

The `".."` string will never be used, so it will be replaced with `ok`.
At the same time, a warning will be emitted.

The reason for the warning is that the `".."` string does not have a
`compiler_generated` annotation to indicate that it was introduced by
the compiler. The `X1` variable had a `compiler_generated` annotation,
but it was lost when `X1` was replaced with `".."`.

To eliminate the unwanted warning, the `compiler_generated` annotation
must be propagated from the variable to the substituted value.

Thanks to Jose Maria Perez Ramos (@Kuroneer) for noticing this bug and
suggesting a way to fix it.

### DIFF
--- a/lib/compiler/src/sys_core_fold.erl
+++ b/lib/compiler/src/sys_core_fold.erl
@@ -1074,23 +1074,13 @@ let_substs_1(Vs, #c_values{es=As}, Sub) ->
 let_substs_1([V], A, Sub) -> let_subst_list([V], [A], Sub);
 let_substs_1(Vs, A, _) -> {Vs,A,[]}.
 
-let_subst_list([V|Vs0], [A0|As0], Sub) ->
+let_subst_list([V|Vs0], [A|As0], Sub) ->
     {Vs1,As1,Ss} = let_subst_list(Vs0, As0, Sub),
-    case is_subst(A0) of
+    case is_subst(A) of
 	true ->
-	    A = case is_compiler_generated(V) andalso
-		    not is_compiler_generated(A0) of
-		    true ->
-			%% Propagate the 'compiler_generated' annotation
-			%% along with the value.
-			Ann = [compiler_generated|cerl:get_ann(A0)],
-			cerl:set_ann(A0, Ann);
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
Hm... I wonder if this fix will have an impact on what @michaelklishin describes in https://github.com/erlang-lager/lager/pull/547.

bjorng:
> Hm... I wonder if this fix will have an impact on what @michaelklishin describes in erlang-lager/lager#547.

Yes, this bug could explain that there were still warnings even though `compiler_generated` was used.

PR #4903, opened 2021-05-27
Fix busy port handling in port_command()

### LINKED ISSUE #4898: Scheduler falls into a busy loop handling bif_handle_signals_return trap
We recently attempted an upgrade to OTP-22.3.4.15 to OTP-22.3.4.19. After the upgrade we observed the following behavior. Pretty consistently processes dealing with ports would cease making any progress and would get continuously scheduled/rescheduled.

Connecting gdb to beam.smp shows one of the schedulers constantly in bif_handle_signals_return:

```
(gdb) bt
#0  bif_handle_signals_return (A__p=0x7fed060c0618, BIF__ARGS=0x7fed06c40200, A__I=<optimized out>) at beam/bif.c:4981
#1  0x0000563a6f6c40af in process_main (x_reg_array=0x7fed06c40200, f_reg_array=0x7fed0417eb84) at x86_64-unknown-linux-gnu/opt/smp/beam_cold.h:59
#2  0x0000563a6f6d01a5 in sched_thread_func (vesdp=0x7fed05144e00) at beam/erl_process.c:8512
#3  0x0000563a6f914b1c in thr_wrapper (vtwd=0x7ffe82962e60) at pthread/ethread.c:118
#4  0x00007fed48dfd299 in start_thread () from /usr/lib/libpthread.so.0
#5  0x00007fed48d1b053 in clone () from /usr/lib/libc.so.6
```

The process that is being executed is always the same:

```
(gdb) etp-process-info A__p
  Pid: <0.158.0>
  State: running | active | prq-prio-normal | usr-prio-normal | act-prio-normal

  Flags: trap-exit 
  Registered name: 'sink-stderr'
  Current function: erlang:bif_handle_signals_return/1
  CP: #Cp<erlang:port_command/2+0x68>
  I: #Cp<0x563a6fa1d5e0>
  Heap size: 1598
  Old-heap size: 376
  Mbuf size: 38
  Msgq len: 29 (inner=29, outer=0)
  Msgq Flags: on-heap 
  Parent: <0.157.0>
  Pointer: (Process*)0x7fed060c0618
```

Stepping through this function shows that no signals are actually handled, but the function returns `THE_NON_VALUE`:

```
(gdb) etp BIF__ARGS[0]
<the-non-value>.
```

This appears to tell the scheduler to immediately call into `bif_handle_signals_return` again.

Looking at the changes between OTP-22.3.4.15 and OTP-22.3.4.19 the following caught our attention:

```
commit 15ac634091f25df3a46c7f354711773167602994
Author:     Rickard Green <rickard@erlang.org>
AuthorDate: Tue Apr 20 16:45:54 2021 +0200
Commit:     Rickard Green <rickard@erlang.org>
CommitDate: Thu Apr 22 18:46:08 2021 +0200

    Ensure reply signals from ports respect signal order
    
    The reply signal from a port to a process, when the process called one of the
    'port_*' BIFs, could pass before other signals sent from the port to the
    process. That is, the reply signal could be delivered too early.
```

And indeed applying the following patch makes the problem disappear:

```
---
 erts/emulator/beam/erl_bif_port.c | 5 -----
 1 file changed, 5 deletions(-)
diff --git a/erts/emulator/beam/erl_bif_port.c b/erts/emulator/beam/erl_bif_port.c
index 11bd3729cd..5bcef284a5 100644
--- a/erts/emulator/beam/erl_bif_port.c
+++ b/erts/emulator/beam/erl_bif_port.c
@@ -236,11 +236,6 @@ BIF_RETTYPE erts_internal_port_command_3(BIF_ALIST_3)
 	KILL_CATCHES(BIF_P);	/* Must exit */
 	ERTS_BIF_PREP_ERROR(res, BIF_P, EXC_ERROR);
     }
-    else {
-        /* Ensure signal order is preserved... */
-        if (state & (ERTS_PSFLG_SIG_Q|ERTS_PSFLG_SIG_IN_Q))
-            ERTS_BIF_PREP_HANDLE_SIGNALS_RETURN(res, BIF_P, res);
-    }
     return res;
 }
--
2.24.3 (Apple Git-128)
```

Looking at `erts_internal_port_command_3` it seems that if the port is busy and the process has got outstanding signals, first a trap will be set up to yield (https://github.com/erlang/otp/blob/OTP-22.3.4.19/erts/emulator/beam/erl_bif_port.c#L214-L215), and then this trap will be overwritten by a trap to handle outstanding signals (https://github.com/erlang/otp/blob/OTP-22.3.4.19/erts/emulator/beam/erl_bif_port.c#L242). The value of `res` in this case will be `THE_NON_VALUE`. It will get passed to `bif_handle_signals_return` and returned from it (https://github.com/erlang/otp/blob/OTP-22.3.4.19/erts/emulator/beam/bif.c#L5003) causing the observed behavior. The original trap will never get executed.

If the explanation above is not correct, please let us know what more information we could provide: we can reproduce the issue pretty easily.

rickard-green:
A call to `port_command()` could cause a scheduler to end up in an eternal loop if the port was busy and the calling process had incoming signals at the time of the call. This bug was introduced in OTP 23.3.2 (ERTS version 11.2.1), OTP 22.3.4.18 (ERTS version 10.7.2.10), and OTP 21.3.8.23 (ERTS version 10.3.5.18).

Closes #4898

### DIFF
--- a/erts/emulator/beam/erl_bif_port.c
+++ b/erts/emulator/beam/erl_bif_port.c
@@ -211,8 +211,8 @@ BIF_RETTYPE erts_internal_port_command_3(BIF_ALIST_3)
                 ERTS_BIF_PREP_RET(res, am_false);
             else {
                 erts_suspend(BIF_P, ERTS_PROC_LOCK_MAIN, prt);
-                ERTS_BIF_PREP_YIELD3(res, bif_export[BIF_erts_internal_port_command_3],
-                                     BIF_P, BIF_ARG_1, BIF_ARG_2, BIF_ARG_3);
+                ERTS_BIF_YIELD3(bif_export[BIF_erts_internal_port_command_3],
+                                BIF_P, BIF_ARG_1, BIF_ARG_2, BIF_ARG_3);
             }
             break;
         case ERTS_PORT_OP_BUSY_SCHEDULED:
--- a/erts/emulator/test/busy_port_SUITE.erl
+++ b/erts/emulator/test/busy_port_SUITE.erl
@@ -27,7 +27,8 @@
 	 hard_busy_driver/1, soft_busy_driver/1,
          scheduling_delay_busy/1,
          scheduling_delay_busy_nosuspend/1,
-         scheduling_busy_link/1]).
+         scheduling_busy_link/1,
+         busy_with_signals/1]).
 
 -include_lib("common_test/include/ct.hrl").
 
@@ -43,7 +44,7 @@ all() ->
      no_trap_exit, no_trap_exit_unlinked, trap_exit,
      multiple_writers, hard_busy_driver, soft_busy_driver,
      scheduling_delay_busy,scheduling_delay_busy_nosuspend,
-     scheduling_busy_link].
+     scheduling_busy_link, busy_with_signals].
 
 init_per_testcase(_Case, Config) when is_list(Config) ->
     Killer = spawn(fun() -> killer_loop([]) end),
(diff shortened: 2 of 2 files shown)


aartamonau:
Apologies if my question is trivial. But I was wondering if there are any detrimental effects of bypassing the check for an exiting process when the port is busy [here](https://github.com/erlang/otp/blob/master/erts/emulator/beam/erl_bif_port.c#L234-L238)?

rickard-green:
@aartamonau No it should not become exiting in that state. The process is alive when calling the `port_command()` BIF and the operation immediately backs off when it detects that the port is busy without communicating with the port, so it cannot become exiting due to that.

PR #4906, opened 2021-05-28
Correct error messages for timer BIFs

bjorng:
The extended error information for timer BIFs would be incorrect if a
time value that exceeded the maximum allowed value was given:

    1> erlang:send_after(1 bsl 50, self(), whatever).
         in function  erlang:send_after/3
            called as erlang:send_after(1125899906842624,<0.83.0>,whatever)
    2> erlang:send_after(1 bsl 50, self(), whatever, []).
         in function  erlang:send_after/4
            called as erlang:send_after(1125899906842624,<0.85.0>,whatever,[])
            *** argument 4: invalid option in list

For the call to `erlang:send_after/3`, no extended errors information would be
reported, while for the call to `erlang:send_after/4` there would be an incorrect
error message for the fourth argument.

This commit correct the messages, producing:

    1> erlang:send_after(1 bsl 50, self(), whatever).
         in function  erlang:send_after/3
            called as erlang:send_after(1125899906842624,<0.83.0>,whatever)
            *** argument 1: exceeds the maximum supported time value
    2> erlang:send_after(1 bsl 50, self(), whatever, []).
         in function  erlang:send_after/4
            called as erlang:send_after(1125899906842624,<0.85.0>,whatever,[])
            *** argument 1: exceeds the maximum supported time value

### DIFF
--- a/erts/emulator/test/exception_SUITE.erl
+++ b/erts/emulator/test/exception_SUITE.erl
@@ -681,6 +681,7 @@ do_error_3(Reason, Args, Options) ->
 error_info(_Config) ->
     DeadProcess = dead_process(),
     NewAtom = non_existing_atom(),
+    Eons = 1 bsl 50,
 
     %% We'll need an incorrect memory type for erlang:memory/1. We want to test an
     %% incorrect atom if our own allocators are enabled, but if they are disabled,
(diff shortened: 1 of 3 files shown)


erts/emulator/test/exception_SUITE.erl
```diff
@@ -1119,7 +1120,13 @@ error_info(_Config) ->
          {send, [[bad,dest], message]},
          {send, [[bad,dest], message, bad]},
 
+         {send_after, [Eons, self(), message]},
```
garazdawi, line 1123: In #4764 I added in the stdlib error_info_lib so that you can validate which arguments should cause an error and also that the correct description is printed. For example: https://github.com/erlang/otp/pull/4757/files#diff-17817455d8726ece362654904cc419cfe59a2c0e5237bda32e3dd98cf1fd5f5fR3000

Do we want to use that for these tests as well?

garazdawi (approved):
lgtm

erts/emulator/test/exception_SUITE.erl
bjorng: Yes, I think that is a good idea. We'll better do that in a separate pull request.

PR #4909, opened 2021-05-28
fix ./configure script handling of expr(1) on BSD

dch:
Without this patch, the following error occurs repeatedly:

```
expr: illegal option -- -
expr: usage: expr [-e] expression
```

FreeBSD ports tree has carried a local patch addressing this for a while.

I haven't tested this on other OS but I believe -- is a GNU-ism already.

### DIFF
--- a/configure
+++ b/configure
@@ -55,7 +55,7 @@ skip_applications=
 while test $# != 0; do
     case $1 in
 	-srcdir=* | --srcdir=*)
-	    user_srcdir=`expr "$1" : '[^=]*=\(.*\)'`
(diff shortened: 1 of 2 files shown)


tianon:
Unfortunately, this breaks with BusyBox's `expr`, which thus breaks on Alpine Linux: :disappointed:

```console
$ docker run -it --rm alpine:3.14
/ # expr -- --without-stuff : '--without-\(.*\)'
expr: syntax error
/ # expr --without-stuff : '--without-\(.*\)'
stuff
```

rickard-green:
@tianon :( please write a bug report at https://github.com/erlang/otp/issues or if you have a solution (for both or rather all platforms) a pull request is welcome

Issue #4448, opened 2020-11-20
ERL-1414: global:unregister_name(order_processor) takes too long to execute


OTP-Maintainer:
Original reporter: `sonicoder`
Affected version: `OTP-23.1.1`
Component: `kernel`
Migrated from: https://bugs.erlang.org/browse/ERL-1414

---

```
Since we upgraded from 20.3.8 to 23.1.1, we are seeing major slowdown because:

Line 958: global:unregister_name(order_processor) takes too long to execute.

!image-2020-11-20-21-51-31-768.png|width=592,height=96!

!image-2020-11-20-21-52-40-081.png|width=1037,height=103!

Note: the name order_processor is being registered only by 1 process so no fancy parallel contention could be happening .. I have noticed that it might take over 10 secs for this call to finish whereas other times it finishes very fast
```

uabboli:
I cannot say if GH-4912 fixes this particular problem, but it corrects
one bug, and now the test suite hasn't failed in our daily builds for
quite some time. A patch for Erlang/OTP 23 is upcoming.

Issue #3923, opened 2019-03-14
ERL-885: Calls to global:register_name hang in whole cluster for hours


OTP-Maintainer:
Original reporter: `rumataestor`
Affected version: `OTP-19.3.6`
Component: `kernel`
Migrated from: https://bugs.erlang.org/browse/ERL-885

---

```
We are developing a distributed platform that consists a number of logically and physically separated applications.
Some of those applications use {{global:register_name}} to ensure only a single process associated with a particular value exists in a logically related part of the cluster.
We also use {{net_kernel:hidden_connect_node}} to prevent the global registry of one logical application from leaking to the rest of the cluster - to reduce synchronisation overhead and to encourage process discovery using RPC based API.

It worked very well for quite some time until a few days ago we experienced a serious outage when no node in the cluster could register a new process using {{global:register_name}}.


The first thing we noticed was that the "isolated to be" logical application did actually connect to the rest of the cluster due to a small bug – a connection to a node of a different application was done visibly.
I think that it shouldn’t be a problem and couldn’t cause the disaster as the names don't clash.


So I tried to find out why registration was not progressing, which I discovered by checking the {{registrar}} process attached to the {{global_name_server}}:

{noformat}

iex(pcp_agent@HCCPCPNODE326)22> pid(0,1935,0) |> Process.info(:backtrace) |> elem(1) |> IO.puts
Program counter: 0x00007fac8caf4d50 (global:random_sleep/1 + 392)
CP: 0x0000000000000000 (invalid)
arity = 0

0x00007fac8a8cea70 Return addr 0x00007fac8cae9260 (global:set_lock_known/2 + 488)
y(0)     6420

0x00007fac8a8cea80 Return addr 0x00007fac8cae8f68 (global:trans_all_known/1 + 144)
y(0)     4046
y(1)     {global,<0.1935.0>}

0x00007fac8a8cea98 Return addr 0x00007fac8caf5c68 (global:loop_the_registrar/0 + 160)
y(0)     []
y(1)     []
y(2)     {global,<0.1935.0>}
y(3)     #Fun<global.1.112826559>

0x00007fac8a8ceac0 Return addr 0x00000000009383e8 (<terminate process normally>)
y(0)     {<0.15261.568>,#Ref<0.0.15990785.211922>}

{noformat}

[...]

PR #4912, opened 2021-05-28
kernel: Fix a race condition in Global

uabboli:
A race between locker processes on different nodes has been resolved
by using global_name_server as proxy.

Possibly fixes GH-4448 and possibly fixes GH-3923.

The intention is to include this fix in patches of 22, 23, and 24 shortly.


uabboli:
The branch has been rebased onto OTP-22.3.4. Patches are upcoming for
Erlang/OTP 22, 23 and 24. No patches are created for releases before
22, but it should be possible to apply the modifications of the branch
on 19, 20 and 21.

Issue #4824, opened 2021-05-13
io_lib:format/3 with specific chars_limit falls into a loop


hareen-kancharla:
**Describe the bug**
For specific nested terms and chars_limit, like the one mentioned below io_lib:format/3 falls into a loop. 
**To Reproduce**
```
1> f(F), F = fun R(I) -> case I =:= 0 of true -> 0; false -> [I, R(I-1)] end end.
#Fun<erl_eval.31.126501267>
2> io_lib:format("~p", [F(1000)], [{chars_limit, 1000}]). 
```
Similar bugs exists for nested tuples and nested maps. 
```
f(F), F = fun R(I) -> case I =:= 0 of true -> 0; false -> {I, R(I-1)} end end.
f(F), F = fun R(I) -> case I =:= 0 of true -> 0; false -> #{I => R(I-1)} end end.
```
**Expected behavior**
The list should be trimmed, to print the term until the soft limit of chars_limit is printed.
**Affected versions**
Tested in on the following version - Erlang/OTP 22 [erts-10.7.2.7] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [hipe]
**Additional context**

I think the bug is in the function io_lib_pretty:print_length_list1, where the length needed to print a comma as a delimiter between elements in a list is counted twice. T is decremented twice at line 698 (tsub(T,1))and line 699 (tsub(T,Len+1)). But the length calculation done via list_length does it correctly by accounting it for only once.

Edit 1: More accurately the bug is this I think - when you have a tail-nested term like, {1,{2,{3 ...}}}} we reserve space for a comma even for the tail element (in line 698) which will never be used. I think there should be stronger checks to check if E is a tail element or not and only then decrement T by 1 and not decrement T at all in line 699.

And therefore in the function clause:
`print_length_list1(Term, D, T, RF, Enc, Str) when D =:= 1; T =:= 0
` when T =:= 0 and the walk terminates, the Length accumulated in Len via list_length (in print_length_list) is less than chars_limit.

Now, In find_upper (on finding Len < T) we exponentially keep increasing the depth and end up looping.

Find outputs from examples runs here - https://gist.github.com/hareen-kancharla/43480e5070a2a8c7498be69e18cfe644

I wish to make a pull request for this with the following change and understand if replacing tsub(T,1) wouldn't regress other cases. Let me know. 
```
--- a/lib/stdlib/src/io_lib_pretty.erl
[...]

Issue #4307, opened 2019-06-06
ERL-967: io_lib:format/3 can fall into an infinite loop when chars_limit is set


OTP-Maintainer:
Original reporter: `avitko`
Affected versions: `OTP-22.0`, `OTP-21.3`
Fixed in versions: `OTP-22.0.3`, `OTP-21.3.8.4`
Component: `stdlib`
Migrated from: https://bugs.erlang.org/browse/ERL-967

---

```
*io_lib:format/3* can fall into an infinite loop when we have nested data structures (probably only with binaries)
and particular value for *chars_limit*.

The infinite loop appears in *io_lib_pretty:find_upper/9* when *io_lib_pretty:expand/3* starts to return
same values. *Len* returned by *io_lib_pretty:expand/3* is equal to *T* which leads to a re-call of *io_lib_pretty:find_upper/9*.

I'm not sure that condition *Len =< T* is wrong. Probably *io_lib_pretty:expand/3* or some function inside of has a mistake.

Example to reproduce error:

{code}
io_lib:format("~tp", [[{0, [<<"00">>]}]], [{chars_limit, 18}]).
{code}
```

PR #4913, opened 2021-05-28
stdlib: Fix a loop when using io_lib option 'chars_limit'

uabboli:
See GH-4824: https://github.com/erlang/otp/pull/4824.

The fix of ERL-967, see https://github.com/erlang/otp/issues/4307,
proved to be inaccurate.

Thanks to Hareen (kvsrh) for tracking down the bug. The fix in
https://github.com/erlang/otp/pull/4842 is included.

A clause that breaks the infinite loop is also added. It will hide
further bugs, but it is deemed more important not to loop.

Fixes GH-4824. Closes GH-4842.

### DIFF
--- a/lib/stdlib/src/io_lib_pretty.erl
+++ b/lib/stdlib/src/io_lib_pretty.erl
@@ -448,11 +448,11 @@ intermediate(Term, D, T, RF, Enc, Str) when T > 0 ->
     case If of
         {_, Len, Dots, _} when Dots =:= 0; Len > T; D =:= 1 ->
             If;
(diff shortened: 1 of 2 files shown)


uabboli:
The fix for this bug has been released in the following patches:
* [OTP 24.0.2](https://erlang.org/download/OTP-24.0.2.README)
* [OTP 23.3.4.2](https://erlang.org/download/OTP-23.3.4.2.README)
* [OTP 22.3.4.20](https://erlang.org/download/OTP-22.3.4.20.README)

PR #4915, opened 2021-05-29
Fully support maps in ms_transform

gomoripeti:
Before this change only map patterns worked and only in the shell via a
hack in `normalise/1` (It converted a map pattern AST to map, but it did
not convert a map expressions AST).

Now the transformation takes care of map patterns in MS head and map
expressions in MS guards/body and `normalise/1` only does what
`erl_parse:normalise/1`. Checking restrictions on map keys and values is
left to the PAM machine.

### DIFF
--- a/lib/stdlib/src/ms_transform.erl
+++ b/lib/stdlib/src/ms_transform.erl
@@ -756,7 +756,12 @@ tg({bin_element,Anno,X,Y,Z},B) ->
 
 tg({bin,Anno,List},B) ->
     {bin,Anno,[tg(X,B) || X <- List]};
-    
+
+tg({map_field_assoc, Anno, Field, Value}, B) ->
+    {map_field_assoc, Anno, tg(Field, B), tg(Value, B)};
+tg({map, Anno, List}, B) ->
+    {map, Anno, [tg(X, B) || X <- List]};
+
 tg(T,B) when is_tuple(T), tuple_size(T) >= 2 ->
     Element = element(1,T),
     Anno = element(2,T),
@@ -858,6 +863,9 @@ th({var,Anno,Name},B,OB) ->
 	Trans ->
 	    {{atom,Anno,Trans},B}
     end;
+th({map_field_exact,Anno,Field,Value},B,OB) ->
+    {[NField, NValue], NB} = th([Field, Value], B, OB),
+    {{map_field_assoc,Anno,NField,NValue}, NB};
 th([H|T],B,OB) ->
     {NH,NB} = th(H,B,OB),
     {NT,NNB} = th(T,NB,OB),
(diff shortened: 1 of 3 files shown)


gomoripeti:
hi Hans, thanks for the feedback

(oops, the erlang:error is indeed just a copy-paste, will remove)

> If I understand correctly, it's a bug that => can be used in the
head? 

Do you mean that the below code works, even before this patch?
```
f() ->
  ets:fun2ms(
    fun({key, #{foo => V}}) -> %% using => in head
      V
    end).
```
(I did not realize this is possible, as it does not work in the shell, cannot create a shell fun with this illegal pattern. But this indeed works in compiled, parse_transformed code)
I think it still works, as a map_field_assoc gets the same generic handling in `th/3` as before, but I will add a testcase to document this.

uabboli:
> Do you mean that the below code works, even before this patch?

Yes, as you write, it works in compiled code.

A suggestion is to modify ms_transform_SUITE:do_eval() (and affected
test(s) in the suite) in order to bypass the linter check. Like this:

    {value,Res,[]} =  erl_eval:exprs(Tree, [], none, none),

Then it's easier to see that:

    do_eval("ets:fun2ms(fun({key, #{foo => V}}) -> V end)").

works with your patch. It used to crash with a function_clause.

gomoripeti:
I updated the PR according to your suggestions.

How much bug compatibility OTP needs to maintain?
- allowing => in head in compiled code could be considered an existing bug
- allowing the same in interpreted code could be considered a new bug
Would it be an option to "fix" them in this PR, by emitting a proper parse_transform error if they occur?
I can also create a separate PR for master with the "fix" (and converting the map_expr_in_head test cases to expect error)

uabboli:
I've talked with Kenneth about this, and we think that we need to do
this in two steps, as you suggest:

 - make ':=' work on maint;

 - do not allow '=>' on master.

I think it's OK if '=>' works with interpreted code on maint; it's not
something that will be mentioned in the release note, though.

gomoripeti:
thanks, I squashed the test cases for => and will submit a separate PR for master.

uabboli:
Sorry for the delay.

Your fixes have exposed a more general bug which is caught by
qlc_SUITE:info/1. To me it seems that the 'const' tag inserted by
fun2ms is not always handled correctly by the matchspec compiler.
You're of course welcome to try to fix the bug, but otherwise someone
at OTP will have a look when they are back from vacations.

Internal note: GH-4915 is handled by ticket OTP-17518.

gomoripeti:
thanks for the heads up. after a quick look at the qlc suite it is very likely I will have to leave the fix to the OTP team.

gomoripeti:
thanks, will do

bjorng:
I have changed the base to `master` as it's too late for OTP 24.3. But it is still time to include this PR in the next release candidate for OTP 25.

Please amend your commit to include this correction of `qlc_SUITE`:

```
diff --git a/lib/stdlib/test/qlc_SUITE.erl b/lib/stdlib/test/qlc_SUITE.erl
index 827af0bb18..ce0b0b96e0 100644
--- a/lib/stdlib/test/qlc_SUITE.erl
+++ b/lib/stdlib/test/qlc_SUITE.erl
@@ -2636,9 +2636,9 @@ info(Config) when is_list(Config) ->
              L = [{#{k => #{v => Fun}}, Fun}],
              H = qlc:q([Q || Q <- L, Q =:= {#{k => #{v => Fun}}, Fun}]),
              L = qlc:e(H),
-             {call,_,_,[{lc,_,{var,_,'Q'},
-                         [{generate,_,_,_},
-                          {op,_,_,_,_}]}]} =
+             {call,_,{remote,_,{atom,_,ets},{atom,_,match_spec_run}},
+                [_,
+                 {call,_,{remote,_,{atom,_,ets},{atom,_,match_spec_compile}},[_]}]} =
                 qlc:info(H, [{format,abstract_code}])">>
 
        ],
```

When you have done that, I will add the PR to our daily builds.

gomoripeti:
thanks Bjorn, I amended the commit.

TBH I completely forgot about this PR and the context. The discussion talks about removing the hidden behaviour that allows '=>' in function head on master. Do you think this should be done for OTP 25?

bjorng:

> The discussion talks about removing the hidden behaviour that allows '=>' in function head on master. Do you think this should be done for OTP 25?

It seems to be safer to wait to at least OTP 26. If I correctly understood the conversation above, `:=` did not work before this PR, so it would be impossible to write map matching code that would work both in OTP 24 and OTP 25.

bjorng:
Thanks for your pull request.

PR #4917, opened 2021-05-30
Make os:cmd compatible with Android

JeromeDeBretagne:
Support when the shell is located in /system/bin/sh on Android, otherwise a call to os:cmd would fail like this:

```
$ erl
Eshell V12.0  (abort with ^G)
1> os:cmd("ls").
"sh: /bin/sh: not found\n"

```

Let me know what you think and if you have suggestions for improvements. Thanks a lot

### DIFF
--- a/lib/kernel/src/os.erl
+++ b/lib/kernel/src/os.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 1997-2018. All Rights Reserved.
+%% Copyright Ericsson AB 1997-2021. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -313,7 +313,22 @@ mk_cmd(_,Cmd) ->
     %% We use an absolute path here because we do not want the path to be
     %% searched in case a stale NFS handle is somewhere in the path before
     %% the sh command.
-    {"/bin/sh -s unix:cmd", [out],
+    %%
+    %% Check if the default shell is located in /bin/sh as expected usually
+    %% or in /system/bin/sh as implemented on Android. The raw option is
+    %% used to bypass the file server and speed up the file access.
+    Shell = case file:read_file_info("/bin/sh",[raw]) of
+                {ok,#file_info{type=regular}} ->
+                    "/bin/sh";
+                _ ->
+                    case file:read_file_info("/system/bin/sh",[raw]) of
+                        {ok,#file_info{type=regular}} ->
+                            "/system/bin/sh";
+                        _ ->
+                            "/bin/sh"
+                    end
+            end,
+    {Shell ++ " -s unix:cmd", [out],
      %% We insert a new line after the command, in case the command
      %% contains a comment character.
      %%


lib/kernel/src/os.erl
```diff
@@ -313,7 +313,15 @@ mk_cmd(_,Cmd) ->
     %% We use an absolute path here because we do not want the path to be
     %% searched in case a stale NFS handle is somewhere in the path before
     %% the sh command.
-    {"/bin/sh -s unix:cmd", [out],
+    Command = case {filelib:is_regular("/bin/sh"),
```
garazdawi, line 316: `mk_cmd` can be called in a loop, so I think it makes sense to optimize this code a bit. If this is only for Android, then I would prefer if we could have a branch similar to windows for Android. That is:

```erlang
mk_cmd({win32,Wtype},Cmd) ->
  ...
mk_cmd({unix,android},Cmd) ->
  mk_unix_cmd("/system/bin/sh");
mk_cmd(_,Cmd) ->
  mk_unix_cmd("/bin/sh").
```

If it is not only for Android, then break it up into two case statements, so that only one file access has to be done in the normal case. That is:

```erlang
case filelib:is_regular("/bin/sh") of
  true -> "/bin/sh";
  false -> case filelib:is_regular("/system/bin/sh") of
             true -> "/system/bin/sh";
             false -> "/bin/sh";
           end.
```

JeromeDeBretagne: > `mk_cmd` can be called in a loop, so I think it makes sense to optimize this code a bit.

Sure, let me give it a try.

> If this is only for Android, then I would prefer if we could have a branch similar to Windows for Android. That is:
> 
> ```erlang
> mk_cmd({unix,android},Cmd) ->
>   mk_unix_cmd("/system/bin/sh");
> ```

That's exactly what I had in mind at first but `os:type()` returns `{unix,linux}` on Android. In fact, this is expected as the documentation states it gets the Osname part from `uname -s`, which indeed returns `Linux` on Android:
```
$ uname -s
Linux
```

> so that only one file access has to be done in the normal case. That is:
> 
> ```erlang
> case filelib:is_regular("/bin/sh") of
>   true -> "/bin/sh";
>   false -> case filelib:is_regular("/system/bin/sh") of
>              true -> "/system/bin/sh";
>              false -> "/bin/sh";
>            end.
> ```

I've re-implemented it this way to have only one file access as suggested. Let me know if you see any other improvements to take into account.

lib/kernel/src/os.erl
```diff
@@ -313,7 +313,15 @@ mk_cmd(_,Cmd) ->
     %% We use an absolute path here because we do not want the path to be
     %% searched in case a stale NFS handle is somewhere in the path before
     %% the sh command.
-    {"/bin/sh -s unix:cmd", [out],
+    Shell = case filelib:is_regular("/bin/sh") of
```
garazdawi, line 316:
```suggestion
    Shell = case file:read_file_info("/bin/sh",[raw]) of
         {ok,#file_info{type=regular}} -> "/bin/sh";
```
Having thought about it a bit more, let us read the first regular like this instead so that we do not have to jump via the file server.

JeromeDeBretagne: Only for the first one in the regular case? Or for both to provide a consistent implementation?

garazdawi: It is only really important for the first one, but may as well do it for both and add a comment about why we are doing it,

JeromeDeBretagne: > and add a comment about why we are doing it

Which I guess is ... a (much?) faster access ?

garazdawi: Yes

JeromeDeBretagne: > Having thought about it a bit more, let us read the first regular like this instead so that we do not have to jump via the file server.
> ```Erlang
> -    Shell = case filelib:is_regular("/bin/sh") of
> +    Shell = case file:read_file_info("/bin/sh",[raw]) of
> +         {ok,#file_info{type=regular}} -> "/bin/sh";
> ```

I've reimplemented it this way and I've added a comment as suggested. 

As a side feedback, I was really confused at first when reading the doc of read_file_info. I was expecting to also have to support the case when `{ok,#file_info{type=symlink}}` is returned since /bin/sh is a symlink in many cases. However, my local testing has confirmed that it is not needed indeed.

The `file` documentation should be improved to describe this implementation detail as this is far from obvious and quite important in terms of semantics. I'm still not even sure to understand in which case `{ok,#file_info{type=symlink}}` can be returned when calling `read_file_info`.

garazdawi: > I'm still not even sure to understand in which case {ok,#file_info{type=symlink}} can be returned when calling read_file_info.

I do not think it can. But the `file:read_link_info` documentation points to the same docs and when you use that you can get it. I'll see if I can make the docs clearer.

JeromeDeBretagne:
Thank you @garazdawi for the improved `file:read_file_info` doc in 965d1d4 and for the merge of this PR.

PR #4919, opened 2021-05-31
Fix crash in c:ls/1

### LINKED ISSUE #4916: c:ls/1 throws an error when called with an atom pointing to a file
**Describe the bug**
[c:ls/1](https://erlang.org/doc/man/c.html#ls-1) throws an error when called with an `atom` pointing to a `file`.

**To Reproduce**
```
$ erl
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit]

Eshell V12.0  (abort with ^G)
1> c:ls("/bin/sh").
/bin/sh     
ok
2> c:ls('/bin/sh').
     in function  length/1
        called as length('/bin/sh')
        *** argument 1: not a list
     in call from c:lengths/2 (c.erl, line 1083)
     in call from c:ls_print/1 (c.erl, line 1071)
```

**Expected behavior**
According to the spec, the function should work similarly with both string and atom parameters:

```
$ erl
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit]

Eshell V12.0  (abort with ^G)
1> c:ls("/bin/sh").
/bin/sh     
ok
2> c:ls('/bin/sh').
/bin/sh     
ok
```

**Affected versions**
Detected with Erlang/OTP 24.0, it may affect older versions.

**Additional context**

Here is the [c:ls/1](https://erlang.org/doc/man/c.html#ls-1) spec:

```
ls(Dir) -> ok
   Types
      Dir = file:name()
   Lists files in directory Dir or, if Dir is a file, only lists it.
```

and the [file:name()](https://erlang.org/doc/man/file.html#type-name) type definition:

```
name() = string() | atom() | deep_list()
```

When called with an `atom` pointing to a `directory`, the function works:

```
$ erl
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit]

Eshell V12.0  (abort with ^G)
1> c:ls("/bin").
bash                               bunzip2                            
busybox                            bzcat 
...
ok
2> c:ls('/bin').
bash                               bunzip2                            
busybox                            bzcat 
...
ok
```
so the behavior is currently not consistent if `Dir` is a directory or a file.

bjorng:
A call to `c:ls/1` with a filename given as an atom would crash.
Example:

    1> c:ls('/bin/sh').
         in function  length/1
            called as length('/bin/sh')
            *** argument 1: not a list
         in call from c:lengths/2 (c.erl, line 1083)
         in call from c:ls_print/1 (c.erl, line 1071)

Closes #4916

### DIFF
--- a/lib/stdlib/src/c.erl
+++ b/lib/stdlib/src/c.erl
@@ -38,7 +38,7 @@
 -export([appcall/4]).
 
 -import(lists, [reverse/1,flatten/1,sublist/3,sort/1,keysort/2,
-		max/1,min/1,foreach/2,foldl/3,flatmap/2]).
+		foreach/2,foldl/3,flatmap/2]).
 -import(io, [format/1, format/2]).
 
 %%-----------------------------------------------------------------------
@@ -1056,11 +1056,15 @@ ls() ->
 -spec ls(Dir) -> 'ok' when
       Dir :: file:name().
 
-ls(Dir) ->
-    case file:list_dir(Dir) of
+ls(Dir0) ->
+    case file:list_dir(Dir0) of
 	{ok, Entries} ->
 	    ls_print(sort(Entries));
 	{error, enotdir} ->
+            Dir = if
+                      is_list(Dir0) -> lists:flatten(Dir0);
+                      true -> Dir0
+                  end,
 	    ls_print([Dir]);
 	{error, Error} ->
 	    format("~ts\n", [file:format_error(Error)])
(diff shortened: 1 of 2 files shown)


lib/stdlib/src/c.erl
```diff
@@ -1080,8 +1080,12 @@ ls_print([H|T], Width, Len) ->
 ls_print([], _, _) ->
     io:nl().
 
-lengths([H|T], L) -> lengths(T, [length(H)|L]);
-lengths([], L)    -> L.
+max_length([H|T], L) when is_atom(H) ->
```
garazdawi, line 1083: How deep can a `deep_list` (http://erlang.org/doc/man/file.html#type-name) be? Is the same depth as an iolist allowed?

bjorng: Any depth, which `file` will handle, but not `c:ls/1`. I better fix that, too.

garazdawi (approved):

PR #4923, opened 2021-06-01
Make max_transfer_size parameter configurable to optimise mnesia table loading time

ssmyczynski:
When a table is sent from active node to another one it is done in chunks of N records each, where the N is calculated once based on size in bytes of first record and hard coded `MAX_TRANSFER_SIZE` (`7500`).
```erlang
    (?MAX_TRANSFER_SIZE div BinSize) + 1.
```
In practice it means that we are sending a huge number of super small chunks.
It is worth noting that value of `MAX_TRANSFER_SIZE` was not changed at least since 2009 :)

In our production setup by simply changing this hard coded value to `128000` we were able to reduce loading time of all tables from remote active node from 184s to 114s. Combining it with setting `send_compressed` parameter we were able to reduce it further to 70s.

The important thing is that with default  small chunks of max 7500 bytes using `send_compressed` option made loading time worse (220s). It is because compression of small chunks is ineffective (compared to atom dictionary compression in distribution protocol when sending raw terms).

In the proposed PR we are changing the default value of max transfer size to 64000 and make it configurable.

BTW. Since introducing fragmented packets in distribution protocol in OTP-22 sending big messages between nodes is no more a problem so setting this max transfer size to even bigger values is not a bad idea.

### DIFF
--- a/lib/mnesia/doc/src/mnesia.xml
+++ b/lib/mnesia/doc/src/mnesia.xml
@@ -3038,6 +3038,11 @@ raise(Name, Amount) ->
           understand this configuration.</p>
       </item>
       <item>
+        <p><c>-mnesia max_transfer_size Number</c>. Specifies the estimated size
+          in bytes of a single packet of data to be used when copying a table from the local
+          node to another one. Default is <c>64000</c>.</p>
+      </item>
+      <item>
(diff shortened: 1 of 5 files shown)


dgud:
Looks good to me, will run the tests for a couple days.

PR #4925, opened 2021-06-02
Correct printing of Unicode atoms in crash dumps

bjorng:
Atoms containing Unicode code points greater than 255 would not
be printed correctly in crash dumps.

This commit corrects the printing of Unicode atoms done by the
runtime-internal function `erts_printf()` (used when producing crash
dumps) and also updates `crashdump_viewer` to correctly handle such
atoms in function names in stack backtraces.

### DIFF
--- a/erts/emulator/beam/erl_printf_term.c
+++ b/erts/emulator/beam/erl_printf_term.c
@@ -139,8 +139,6 @@ do {                                                        \
 #define IS_CNTRL(c)  ((c) < ' ' || (c) >= 127)
 #endif
 
-#define IS_PRINT(c)  (!IS_CNTRL(c))
-
 /* return 0 if list is not a non-empty flat list of printable characters */
 
 static int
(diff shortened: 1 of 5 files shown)


erts/emulator/beam/erl_printf_term.c
```diff
@@ -185,7 +183,36 @@ static int is_printable_ascii(byte* bytep, Uint bytesize, Uint bitoffs)
...
+/*
+ * Helper function for print_atom_name(). Not generally useful.
+ */
+static ERTS_INLINE int latin1_char(int c1, int c2)
+{
+    if ((c1 & 0x80) == 0) {
+        /* Plain old 7-bit ASCII. */
+        return c1;
+    } else if ((c1 & 0xE0) == 0xC0) {
+        /* Unicode code points from 0x80 through 0xfFF. */
```
kpy3, line 195:
```suggestion
        /* Unicode code points from 0x80 through 0xFF. */
```

garazdawi:
The place where I originally found this problem was when dumping a process with `save_calls` on and a function name was unicode. Would you mind adding a testcase for that?

Besides that this looks good.

erts/emulator/beam/erl_printf_term.c
```diff
@@ -185,7 +183,36 @@ static int is_printable_ascii(byte* bytep, Uint bytesize, Uint bitoffs)
...
+ */
+static ERTS_INLINE int latin1_char(int c1, int c2)
+{
+    if ((c1 & 0x80) == 0) {
+        /* Plain old 7-bit ASCII. */
+        return c1;
+    } else if ((c1 & 0xE0) == 0xC0) {
+        /* Unicode code points from 0x80 through 0xFF. */
+        ASSERT((c2 & 0xC0) == 0x80);
+        return (c1 & 0x1F) << 6 | (c2 & 0x3F);
```
sverker, line 197: Code and comment still doesn't match.
The code decodes a full 2-byte code point, that is 5+6=11 bits up to 0x7FF.

To only decode the "latin1" interval up to 0xFF the three most significant payload bits must be zero:

`} else if ((c1 & 0xFC) == 0xC0) { `

bjorng: > Code and comment still doesn't match.

Thanks. I will update the comment. (The characters 0x100 - 0x7FF will be handled by the caller.)

kpy3: Sorry that was my fault, forgot about higher bits 😅 👍

garazdawi (approved):

dgud (approved):

PR #4926, opened 2021-06-02
gen_statem: Make it possible to spec data

attah:
Add /2 variants of types returning data, so a user can use them to
spec their implementation (parameterized with their own data type)
and thus get validation for it.

**NB**: There is one problem though, the docs doesn't work. [Bug report](https://github.com/erlang/otp/issues/4927)
There seems to be code in the xsl to handle selection with arity, but nobody is using it for types.
Putting both arities in the same datatype block is inspired by function documentation, where it is somewhat commonly used.

Regardless of if i use one of both names, with arity, i get errors like:
`gen_statem:init_result/1: unknown spec`

### DIFF
--- a/lib/stdlib/doc/src/gen_statem.xml
+++ b/lib/stdlib/doc/src/gen_statem.xml
@@ -1452,7 +1452,8 @@ handle_event(_, _, State, Data) ->
       </desc>
     </datatype>
     <datatype>
-      <name name="init_result"/>
+      <name name="init_result" n_vars="1"/>
+      <name name="init_result" n_vars="2"/>
       <desc>
 	<p>
 	  For a succesful initialization,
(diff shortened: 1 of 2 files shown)


attah:
Updated with correct arity selector for types - now the docs build too.

RaimoNiskanen:

attah:
Any chance of seeing this released?

RaimoNiskanen:
The PR was on 'master' and it is merged into that branch, so it will be released in OTP-25.0 in May (2022), release candidates from February...

attah:
So you want a new PR towards maint-24? Or would it not have any chance to get accepted?

RaimoNiskanen:
No worries.  I can rebase it - would like to run it in our daily tests just for the sake of it...

RaimoNiskanen:
Targetting OTP-24.2 to be released mid-December.

RaimoNiskanen:
Now merged to 'maint'; everything should hereafter just float to the finish...

PR #4928, opened 2021-06-02
Add lists:enumerate/1 and lists:enumerate/2

radrow:
This PR introduces `enumerate/1` and `enumerate/2` functions to
the `lists` module in the standard library. Simple unit tests and
documentation for the introduced functions are included.

### Motivation

I am working a lot with ASTs in Erlang and I very often run into a
situation where I need to do something with a list of arguments of
some function definition knowing their index and each time I am doing
so I need to write something like

```
[ ...
  || {I, Arg} <- lists:zip(lists:seq(0, length(Args)), Args)
]
```

This is really annoying and can be expressed with a single function
`enumerate` which does exactly that. The mentioned use case is not
the only one, of course. I run into this pattern really commonly.

Also, many modern languages support this already. The first that come
to my mind are Rust and Python. In Haskell one may use `zip [0..]` for
that.

Next thing is... why not? 

----------------------------

This is my first contribution to OTP. If it lacks something or misses
some guidelines, please let me know so I can fix it.

### DIFF
--- a/lib/stdlib/doc/src/lists.xml
+++ b/lib/stdlib/doc/src/lists.xml
@@ -175,6 +175,48 @@
     </func>
 
     <func>
+      <name name="enumerate" arity="1" since="OTP @OTP-17523@"/>
+      <fsummary>Annotates elements with their index.</fsummary>
+      <desc>
+        <p>Returns <c><anno>List1</anno></c> with each element
+        <c>H</c> replaced by a tuple of form <c>{I, H}</c> where
+        <c>I</c> is the position of <c>H</c> in
+        <c><anno>List1</anno></c>. The enumeration starts with 1 and
+        increases by 1 in each step.</p>
+        <p>That is, <c>enumerate/1</c> behaves as if it had been defined as follows:</p>
+        <code type="erl">
+enumerate(List) ->
+  {List1, _ } = lists:mapfoldl(fun(T, Acc) -> {{Acc, T}, Acc+1} end, 1, List),
+  List1.</code>
+        <p><em>Example:</em></p>
+        <pre>
+> <input>lists:enumerate([a,b,c]).</input>
+[{1,a},{2,b},{3,c}]</pre>
+      </desc>
+    </func>
+    
+    <func>
+      <name name="enumerate" arity="2" since="OTP @OTP-17523@"/>
+      <fsummary>Annotates elements with their index.</fsummary>
+      <desc>
+        <p>Returns <c><anno>List1</anno></c> with each element
+        <c>H</c> replaced by a tuple of form <c>{I, H}</c> where
+        <c>I</c> is the position of <c>H</c> in
+        <c><anno>List1</anno></c>. The enumeration starts with
+        <c><anno>Index</anno></c> and increases by 1 in each step.</p>
+        <p>That is, <c>enumerate/2</c> behaves as if it had been defined as follows:</p>
+        <code type="erl">
+enumerate(I, List) ->
+  {List1, _ } = lists:mapfoldl(fun(T, Acc) -> {{Acc, T}, Acc+1} end, I, List),
+  List1.</code>
+        <p><em>Example:</em></p>
+        <pre>
+> <input>lists:enumerate(10, [a,b,c]).</input>
+[{10,a},{11,b},{12,c}]</pre>
+      </desc>
+    </func>
+    
+    <func>
       <name name="filter" arity="2" since=""/>
       <fsummary>Select elements that satisfy a predicate.</fsummary>
       <desc>
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -32,7 +32,7 @@
 	 concat/1, flatten/1, flatten/2, flatlength/1,
 	 keydelete/3, keyreplace/4, keytake/3, keystore/4,
 	 keysort/2, keymerge/3, rkeymerge/3, rukeymerge/3, 
-	 ukeysort/2, ukeymerge/3, keymap/3]).
+	 ukeysort/2, ukeymerge/3, keymap/3, enumerate/1, enumerate/2]).
 
 -export([merge/3, rmerge/3, sort/2, umerge/3, rumerge/3, usort/2]).
 
@@ -956,6 +956,24 @@ keymap(Fun, Index, [Tup|Tail]) ->
    [setelement(Index, Tup, Fun(element(Index, Tup)))|keymap(Fun, Index, Tail)];
 keymap(Fun, Index, []) when is_integer(Index), Index >= 1, 
                             is_function(Fun, 1) -> [].
+
+-spec enumerate(List1) -> List2 when
+      List1 :: [T],
+      List2 :: [{Index, T}],
+      Index :: integer(),
+      T :: term().
+enumerate(List1) ->
+    enumerate(1, List1).
+
+-spec enumerate(Index, List1) -> List2 when
+      List1 :: [T],
+      List2 :: [{Index, T}],
+      Index :: integer(),
+      T :: term().
+enumerate(Index, [H|T]) when is_integer(Index) ->
+    [{Index, H}|enumerate(Index + 1, T)];
+enumerate(Index, []) when is_integer(Index) ->
+    [].
 
 %%% Suggestion from OTP-2948: sort and merge with Fun.
 
--- a/lib/stdlib/test/lists_SUITE.erl
+++ b/lib/stdlib/test/lists_SUITE.erl
@@ -58,7 +58,7 @@
 	 join/1,
 	 otp_5939/1, otp_6023/1, otp_6606/1, otp_7230/1,
 	 suffix/1, subtract/1, droplast/1, search/1, hof/1,
-         error_info/1]).
+         enumerate/1, error_info/1]).
 
 %% Sort randomized lists until stopped.
 %%
@@ -122,7 +122,7 @@ groups() ->
      {zip, [parallel], [zip_unzip, zip_unzip3, zipwith, zipwith3]},
      {misc, [parallel], [reverse, member, dropwhile, takewhile,
 			 filter_partition, suffix, subtract, join,
-			 hof, droplast, search, error_info]}
+			 hof, droplast, search, enumerate, error_info]}
     ].
 
 init_per_suite(Config) ->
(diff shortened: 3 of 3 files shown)


lib/stdlib/doc/src/lists.xml
```diff
@@ -171,6 +171,38 @@
...
+      <fsummary>Annotates elements with their index.</fsummary>
+      <desc>
+        <p>Returns <c><anno>List1</anno></c> with each element
+        <c>H</c> replaced by a tuple of form <c>{I, H}</c> where
+        <c>I</c> is the position of <c>H</c> in
+        <c><anno>List1</anno></c>. The enumeration starts with
+        <c><anno>Index</anno></c> and increases by 1 in each step.</p>
+        <p><em>Example:</em></p>
+      <pre>
+> <input>lists:enumerate([a,b,c]).</input>
```
lhoguin, line 201: Example doesn't use Index.

radrow: Fixed, thanks!

lib/stdlib/src/lists.erl
```diff
@@ -957,6 +957,24 @@ keymap(Fun, Index, [Tup|Tail]) ->
 keymap(Fun, Index, []) when is_integer(Index), Index >= 1, 
                             is_function(Fun, 1) -> [].
 
+-spec enumerate(List1) -> List2 when
+      List1 :: [T],
+      List2 :: [{Index, T}],
+      Index :: integer(),
+      T :: term().
+enumerate(List1) ->
+    enumerate(List1, 0).
```
attah, line 966: Hmm. Erlang normally does indexing starting from 1, is this really a good default?

juhlig, line 966: Indexing usually starts at `1`, not `0`.

juhlig: What @attah said =^^=

lib/stdlib/src/lists.erl
```diff
@@ -957,6 +957,24 @@ keymap(Fun, Index, [Tup|Tail]) ->
...
+
+-spec enumerate(List1, Index) -> List2 when
```
lhoguin, line 968: Should it be `Index, List1` instead perhaps? The list tends to be after other arguments in the rest of the module.

lib/stdlib/src/lists.erl
radrow: I was sure that `lists:nth` starts with 0. But you're right, changed to 1

lib/stdlib/src/lists.erl
radrow: Yeah, this is also more consistent with the result type

lib/stdlib/src/lists.erl
radrow: Addressed

dgud:
Looks good.
We also usages of other variants of this but decided that those could be implemented
with the use of a lists:mapfoldl/3.
But it would be good to hint the user reading the docs of that.
So we want you to add to the docs that this function is equivalent to using
` {ResList, _} = lists:mapfoldl(fun(T,Acc) -> {{Acc,T}, Acc+1} end, 1, List),`

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -2743,6 +2743,13 @@ hof(Config) when is_list(Config) ->
 
     ok.
 
+%% Test lists:enumerate/1 and lists:enumerate/2
+enumerate(Config) when is_list(Config) ->
+    [] = lists:enumerate([]),
+    [] = lists:enumerate(10, []),
+    [{1,a},{2,b},{3,c}] = lists:enumerate([a,b,c]),
+    [{10,a},{11,b},{12,c}] = lists:enumerate(10, [a,b,c])
```
dgud, line 2751: This can't compile, there is a missing dot here,
and the test case should return ok.
You might want to run the test-code :-)

Add some tests for bad input as well while you are at it.

dgud: You can run the lists suite with: `make test ARGS='-suite lists_SUITE'`

radrow: Uh, I got fooled by the passing checks in the PR. Any idea why they didn't catch it?

garazdawi: The github actions checks only compile and run a small subset of tests. The lists SUITE is not one of those.

dgud:
Can you fix the docs as I wrote above as well.

radrow:
Done. I also found worth mentioning that `enumerate` preserves length. Is that okay?

lib/stdlib/doc/src/lists.xml
```diff
@@ -196,6 +200,10 @@
         <c>I</c> is the position of <c>H</c> in
         <c><anno>List1</anno></c>. The enumeration starts with
         <c><anno>Index</anno></c> and increases by 1 in each step.</p>
+        <p>The following equalities hold for all enumerations:</p>
+        <code type="none">
+length(lists:enumerate(I, List)) =:= length(List)
+lists:enumerate(I, List) =:= element(1, lists:mapfoldl(fun(T,Acc) -> {{Acc,T}, Acc+1} end, I, List))</code>
```
garazdawi, line 206:
```suggestion
        <p>That is, <c>enumerate/2</c> behaves as if it had been defined as follows:</p>
        <code type="erl">
enumerate(I, List) ->
  {List1, _ } = lists:mapfoldl(fun(T, Acc) -> {{Acc, T}, Acc+1} end, I, List),
  List1.</code>
```

`lists:filtermap/2` uses the this style to describe what the implementation does, so I suggest that we mimic it.

radrow: I got my inspiration from `lists:seq/2`. Also, I think if we are giving a definition example, why not relate to `zip` and `seq`? It is more explicit to my eyes than `mapfoldl`.

garazdawi: Yes, good idea. Use lists:seq + lists:zip as an example instead.

Having a definition example helps when someone comes across this and needs it but with a small variation. They can then just copy the example and adapt it to their needs.

dgud: The idea with using lists:mapfold/3 was that you could change the fun to add a third element,  
or step 10 in each increment,  lists:mapfoldl can do arbitrary change that lists:seq and lists:zip can't do.

radrow: Check it out

radrow: I updated it to use zip&seq, please check it out

radrow:
Do you think `enumerate/3` is worth having? The third argument could be
 - step if `integer()`
 - index update if `function(1)` (assuming `integer() -> integer()`)

dgud:
No we want to keep the lists module as small as possible, 
the lists:mapfold/3 example was intended to avoiding all different variants, 
as you can easily write it, i.e. if you want anything else do it like this. :-)

We searched and found other variants such as order changed '{Element, Index}' and came up with others ideas when we 
discussed this, in the end we decided that we want add the ones you did and add an example of how to do anything else.

radrow:
So do you want me to move back to `mapfoldl` in the example? 

Btw, what is the reason for keeping that lib tight? I am a bit surprised coming from the Haskell land where `Data.List` is very extensive which I find quite convenient in many situations.

lib/stdlib/doc/src/lists.xml
dgud: Ok, it is smaller but that also shows how to write the function in the worst way, from a performance perspective.

The definition is pretty clear from the usage example, so we have already decided that we want the
mapfold example to avoid other variants of this function.

```
<code>
enumerate(I, List) ->
  {List1, _} = lists:mapfoldl(fun(T, Index) -> {{Index, T}, Index+1} end, I, List),
  List1.
</code>

```

dgud:
> So do you want me to move back to mapfoldl in the example?
Yes

> Btw, what is the reason for keeping that lib tight? I am a bit surprised coming from the Haskell land where Data.List is very extensive which I find quite convenient in many situations.

Some want to keep the stdlib as minimal as possible, easier to find functions I guess, most of them are trivial to write
and you can't have all variants because then it gets really hard to give the functions a clear and descriptive name and finding the right one. On the other hand, IMO stdlib should have functions that we write often despite being trivial.

radrow:
Done. 

So please just allow me to alleviate my frustration for lacking such functions by submitting PRs that add them ;)

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -2743,6 +2743,20 @@ hof(Config) when is_list(Config) ->
...
+    [] = lists:enumerate(10, []),
+    [{1,a},{2,b},{3,c}] = lists:enumerate([a,b,c]),
+    [{10,a},{11,b},{12,c}] = lists:enumerate(10, [a,b,c]),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(0)),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(0, 10)),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(1.0, [])),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(1.0, [a,b,c])),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(<<1>>, []),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(<<1>>, [a,b,c]),
+    {'EXIT', {function_clause, _}} = (catch lists:enumerate(1, <<1,2,3>>)).
```
garazdawi, line 2758:
```suggestion
    {'EXIT', {function_clause, _}} = catch lists:enumerate(0),
    {'EXIT', {function_clause, _}} = catch lists:enumerate(0, 10),
    {'EXIT', {function_clause, _}} = catch lists:enumerate(1.0, []),
    {'EXIT', {function_clause, _}} = catch lists:enumerate(1.0, [a,b,c]),
    {'EXIT', {function_clause, _}} = catch lists:enumerate(<<1>>, []),
    {'EXIT', {function_clause, _}} = catch lists:enumerate(<<1>>, [a,b,c]),
    {'EXIT', {function_clause, _}} = catch lists:enumerate(1, <<1,2,3>>).
```
You forgot some ending parenthesis on line 2756 and 2757. `catch` should no longer need the parenthesis, so I suggest you just remove them.

dgud:
Have you run the testcase yet?

Because it fails, probably since it returns an catch error.

radrow:
I just tried to rerun tests on freshly recloned branch but I get an error:
```
make[1]: Entering directory '/home/radek/Programowanie/Erlang/otp/lib/common_test/test_server'
Makefile:85: /make/otp_release_targets.mk: No such file or directory
make[1]: *** No rule to make target '/make/otp_release_targets.mk'.  Stop.
make[1]: Leaving directory '/home/radek/Programowanie/Erlang/otp/lib/common_test/test_server'
make: *** [Makefile:1097: lib/common_test/test_server] Error 2
```

I just did `./configure`, `make` and then `make test ARGS='-suite lists_SUITE'`. Did I miss something?

garazdawi:
You need to set the `ERL_TOP` environment variable. i.e.

```sh
export ERL_TOP=`pwd`
./configure
make
cd lib/stdlib
make test ARGS='-suite lists_SUITE'
```

How to build and run tests is documented here: https://github.com/erlang/otp/blob/master/HOWTO/TESTING.md and more specifically running it using the makefiles is documented here: https://github.com/erlang/otp/blob/master/HOWTO/TESTING.md#convenience-for-running-tests-without-the-release-and-configuration-steps

radrow:
It looks like I forgot that the tests need to return `ok`. Now it works on my side.

dgud:
Looks good now can you squash the commits to one.

radrow:
I think you should be able to squash and merge with one click, but here you are

dgud:
It  also rebases the branch and makes a fast forward merge if I do that and we don't want that, so can't do that.
Thanks

dgud:
I forgot the since tags, sorry can you squash again, so I can merge this.

radrow:
Done

dgud:
Thanks

PR #4929, opened 2021-06-02
lists: add extended function variants

Maria-12648430:
This PR is intended as a proposal, to see if there is actually any interest in the matter.
As such, it provides just the bare implementation, but no changes to specs, docs and tests.
**_Please react with thumbs up/down if you like/dislike the proposal._**

This PR equips any function in `lists` which takes a function (predicates and others) with a second variant on top of the current ones, which takes a function with two more arguments: a list of elements before (`Left`) and another with elements after (`Right`)  the current element (`Cur`).
This allows the function to not only consider the current element, but also the adjacent elements.

The elements in `Left` are in reverse and the elements in `Right` are in the same order as in the original list. `Cur` can thus be said to be the pivot. Usually, you need to consider the elements adjacent or near the pivot element, not those at the beginning.

The only functions not equipped with the extended behavior are the family of sorts, merges, and zipwith. I can't imagine a (likely) use case, and I'm not sure it would even be possible to implement.

#### `map` example
Smoothen a time series by replacing all values with the averages of the previous, current and next values (the first and last values are special, in this implementation they are left as is).

```erlang
lists:map(
    fun
        ([], First, _Right) -> First;
        (_Left, Last, []) -> Last;
        ([Prev|_], Cur, [Next|_]) -> (Prev + Cur + Next) / 3
    end,
    [1, 5, 7, 3, 9, 2]
).

[1, 4.333333333333333, 5.0, 6.333333333333333, 4.666666666666667, 2]
```

#### `filter` example
Collapse a list of elements that may occur multiple times one after another by replacing such multiples with a single one.

```erlang
lists:filter(
    fun
        (_Left, Cur, [Cur|_]) -> false;
        (_Left, _Cur, _Right) -> true
    end,
    [a, a, a, b, c, c, d, d, d, a, e, f, f, b]
).

[a,b,c,d,a,e,f,b]
```

#### `any` example
Are there any duplicate elements in the list, next to each other?

```erlang
lists:any(
    fun
        (_Left, Cur, [Cur|_]) -> true;
        (_Left, _Cur, _Right) -> false
    end,
    [a, b, c, c, d, e]
).

true
```

#### `all` example
Is a list sorted, ie is each element less than or equal to the next?

```erlang
lists:all(
    fun
        (_Left, Cur, [Next|_]) -> Cur =< Next;
        (_Left, _Last, []) -> true;
        (_Left, _Cur, _Right) -> false
    end,
    [1, 2, 4, 8, 16, 0, 32]
).

false
```

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -952,10 +952,21 @@ rukeymerge(Index, T1, L2) when is_integer(Index), Index > 0 ->
       TupleList2 :: [Tuple],
       Tuple :: tuple().
 
-keymap(Fun, Index, [Tup|Tail]) ->
-   [setelement(Index, Tup, Fun(element(Index, Tup)))|keymap(Fun, Index, Tail)];
-keymap(Fun, Index, []) when is_integer(Index), Index >= 1, 
-                            is_function(Fun, 1) -> [].
+keymap(Fun, Index, List) when is_function(Fun, 1), is_integer(Index), Index >= 1 ->
+    keymap_simple(Fun, Index, List);
+keymap(Fun, Index, List) when is_function(Fun, 3), is_integer(Index), Index >= 1 ->
+    keymap_extended(Fun, Index, [], List).
+
+keymap_simple(Fun, Index, [Tup|Tail]) ->
+    [setelement(Index, Tup, Fun(element(Index, Tup)))|keymap_simple(Fun, Index, Tail)];
+keymap_simple(_Fun, _Index, []) ->
+    [].
+
+keymap_extended(Fun, Index, Left, [Cur|Right]) ->
+    [setelement(Index, Cur, Fun(Left, element(Index, Cur), Right))|keymap_extended(Fun, Index, [Cur|Left], Right)];
+keymap_extended(_Fun, _Index, _Left, []) ->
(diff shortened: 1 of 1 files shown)


josevalim:
I am not sure if those additions are worth it. In all of those cases, it feels like reaching out to recursion would be clearer. For instance, the filter example:

```erlang
dedup([Cur, Cur | Rest]) -> dedup([Cur | Rest]);
dedup([Cur | Rest]) -> [Cur | dedup(Rest)];
dedup([]) -> [].
```

I also think in the recursion case the code is also clearer, as it doesn't require understanding exactly how the three arguments in the function work. But I concede this is totally a matter of taste. :)

Maria-12648430:
Ok, so that makes it 1 for and 2 against at this point I guess.

I added a sentence to the first message encouraging to indicate if it is liked or disliked by thumbs up or down reactions, so it can be better tracked =^^=

lhoguin:
Yeah I don't think it brings anything super useful to the table.

Maria-12648430:
Ok, let's see... 😅

```text
👍 : 1
👎 : 5
👀 : 1 (^^?)
```

Looks like a pretty clear outcome, so I'm closing this one.

PR #4932, opened 2021-06-03
Fix erlang:cancel_timer(_, [{info, false}])

rickard-green:


benbro:
Is this bug only in OTP-24 or also previous versions?
What is the implication of using erlang:cancel_timer(Ref, [{async, true}, {info, false}]) in affected version?

rickard-green:
> Is this bug only in OTP-24 or also in previous versions?

All versions since the introduction of the `{info, false}` option in OTP 18.0 is effected by this bug.

> What is the implication of using erlang:cancel_timer(Ref, [{async, true}, {info, false}]) in affected version?

The combo `[{async, true}, {info, false}]` is not effected by this bug.

PR #4933, opened 2021-06-03
bmk/kernel/20210525/reuseaddr_on_windows/otp-17447

bmk:
The socket option 'reuseaddr' is "unsafe" on Windows, therefor it is "ignored" in the inet-driver.

We do not want to change the behaviour of the current implementation, since the behaviour
of the 'reuseaddr' socket option is a bit "iffy" on windows (and Windows was "excluded" for
a reason).
Therefor the current 'reuseaddr' option will remain as is. Instead we have implemented
a 'win_reuseaddr' option. Both operate on the same socket option, but:
'reuseaddr' - is ignored on Windows
'win_reuseaddr' - is ignored on non-windows.
Ṕlease review and comment.


juhlig:
Hm... if you enable thsi option, you should probably also provide means for the `SO_EXCLUSIVEADDRUSE` option, maybe even imply it (in the sense of "you have to explicitly turn it _off_") when `win_reuseaddr` is used. As I understand it, without this additional option `SO_REUSEADDR` is dangerous to use on windows. See [Using SO_REUSEADDR and SO_EXCLUSIVEADDRUSE](https://docs.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse).

bmk:
Good point.

bmk:
Ignore

PR #4937, opened 2021-06-07
Consolidate mandatory distribution flags across applications

bjorng:
The three applications that supports Erlang distribution (erts,
erl_interface, and jinterface), do not agree which the mandatory
distribution flags are.

For example, DFLAG_NEW_FLOATS (support for the "new" external format
for floats introduced in R11B-5) is considered mandatory in
erl_interface, but not in the Erlang emulator. jinterface does not
test the flag, but encodes all floats in the new external format.

Update erts, erl_interface, and jinterface to all use the same set of
mandatory distribution flags.

UPDATE:

Also make the following distribution flags mandatory:
    
* DFLAG_EXPORT_PTR_TAG
* DFLAG_BIT_BINARIES
* DFLAG_MAP_TAG
    
Support for them in erl_interface and jinterface was added in OTP 22
or earlier. Making them mandatory allows us to remove ugly fallback
code.


erts/emulator/beam/external.c
```diff
@@ -3775,123 +3705,6 @@ store_in_vec(TTBEncodeContext *ctx,
     }
 }
 
-static byte *
-begin_hopefull_data(TTBEncodeContext *ctx, byte *ep)
```
sverker, line 3779: You should also be able to delete the actual transcoding of EXPORT_EXT and BIT_BINARY_EXT in `transcode_dist_obuf()`. I think you can remove the entire
`if (~dflags & hopefull_flags) {
   ...
}
`

bjorng: Done.

sverker (approved):

bjorng:
Superseded by #4972.

PR #4938, opened 2021-06-07
RFC: diameter: allow `{error, Cause}` return from pick_peer/4

RoadRunnr:
Currently, when pick_peer/4 can not pick a peer, it return false.
This is translated to a `{error, no_connection}` from diameter:call/4.

The are instance when the inability to choose a peer does not
mean that we don't have a connection to that peer. The main reason
is that we have somehow determined that all peers are overload or close
to beeing overload and want to moderate load.

This kind of load control can not be done the level of diameter:call/4
because at that level only the type of the call is know, not the
possible peers nor can it account for the failover to other peers.

By allowing errors returned from pick_peer/4, it can inform the
application that the request was actually dropped for other (e.g.
load) reasons.

### DIFF
--- a/lib/diameter/src/base/diameter_service.erl
+++ b/lib/diameter/src/base/diameter_service.erl
@@ -331,6 +331,8 @@ pick(#state{options = SvcOpts}
     App = App0#diameter_app{module = ModX ++ Xtra},
     [_,_] = RealmAndHost = diameter_lib:eval([DestF, Dict]),
     case pick_peer(App, RealmAndHost, [Filter | TPids], S) of
+        {error, _} = No ->
+            No;
         {_TPid, _Caps} = TC ->
             {{TC, App}, SvcOpts};
         false = No ->
@@ -1568,8 +1570,8 @@ pick_peer(Local, Remote, Pid, _SvcName, #diameter_app{mutable = true} = App)
             T;
         false = No ->
             No;
-        {error, _} ->
-            false
+        {error, _} = No ->
+            No
     end;
 
 %% App state isn't mutable or it is and we're in the service process:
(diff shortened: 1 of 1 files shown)


RoadRunnr:
@otp-anders this is a request for comments, I'll add the needed doc changes and maybe a test if the general idea is acceptable.

otp-anders:
diameter(3) documents no_connection as being returned if "there are no suitable peers, or if pick_peer/4 rejects them by returning false", so it's already no connection in the sense of not selected rather than not existing.

You're right that there's no way to communicate why a selection was not made though, but is there any compelling reason (other than that it's possible :) for distinguishing between no connections available for a pick_peer callback (ie. no connections to choose from) and rejected by the callback?

If we were to distinguish between the cases then it should also be done in the case when the callback returns NewState, and it's a rejection rather than an error, so maybe 'error' isn't the right way to communicate it. It's been a while since I looked at this interface at all, but I'll take a look to see if it brings back any memories.

RoadRunnr:
The reason for touching this is that I want to implement DIAMETER load and overload control on a per peer level (no on the more generic per request level).

What I need is:
 a) choose a peer based of relative load,
 b) account for the number of outstanding requests per peer
 c) communicate back to the caller if a call failed for overload reasons

This change tries to address c). The primary reason for that are DIAMETER APIs (e.g. Gy) where a explicit alternative peer can be communicated, but the failover to this hosts is not only bound to the existence of a connection, but most generic errors (including overload) can trigger a failover to the alternate host. That means that the existing failover mechanism can not be used, because it only the handles the connection error case. Something at the application has to actually decide that failover is need and possible, and then resent the requests. For that it needs every bit of information it can get about the problem cause. 

Something that I wanted to address separately is that the existing API is also not suitable for a) and b). Picking a peer needs to alter the load indicators for the chosen peer atomically (e.g. the number of outstanding requests), if the request to the picked peer is terminated or abandoned (for whatever reason) that metric needs to be adjusted again atomically. The current API is lacking a reliable indicator/callback for that.  Especially the failover logic is bypassing every return or error handler for the chosen peer, only the final result is passed to the handler.

otp-anders:
You can use the {extra, _} option on call/4 to pass some identifying term into callbacks that they can use to store/retrieve information that has a lifetime beyond a single invocation. Don't know if that's something you've looked at, but that's been used to solve a few communication problems.

RoadRunnr:
Yes, I've already locked at that and I doesn't help with determining when a request to a peer is done. The problem is mostly with the failover logic not invoking any kind of end-of-request handler.

otp-anders:
How do you mean with the failover logic? If a peer is selected then one of handle_answer or handle_error will follow.

otp-anders:
In the case of failover there's always a new pick_peer, so that can be used to keep track of outstanding requests if you use {extra, _} to identify the call/4 invocation. Without that it's impossible to deduce that the new pick_peer is a consequence of failover.

otp-anders:
Closing.

PR #4939, opened 2021-06-07
Add process flag 'heap_growth'

sverker:
New process flag `heap_growth` to enable more **memory conservative garbage collections**.

- `normal` is the default old behavior. When allocating heap capacity during GC it takes into account the amount of live terms that survived the last GC. A process continuously generating live data will thus grow its heap capacity over time and get less frequent but larger GCs.
- `low` is the new memory conservative option. The GC maintains a bounded small heap capacity which results in more frequent but smaller GCs.

In both cases, terms that survive two GCs will be moved to the old-heap whose size calculation is not affected by this PR.

heap_growth can be set/read per process with
- `process_flag(heap_growth, low)`
- `process_info(Pid, heap_growth)`
- `spawn_opt(_, [{heap_growth}])`

and node wise with
- `system_flag(heap_growth, low)`
- `system_info(heap_growth)`
- `erl -hg low`

### DIFF
--- a/erts/emulator/beam/atom.names
+++ b/erts/emulator/beam/atom.names
@@ -331,6 +331,7 @@ atom group_leader
 atom handle
 atom have_dt_utag
 atom heap_block_size
+atom heap_growth
 atom heap_size
 atom heap_sizes
 atom heap_type
--- a/erts/emulator/beam/bif.c
+++ b/erts/emulator/beam/bif.c
@@ -1918,6 +1918,20 @@ BIF_RETTYPE process_flag_2(BIF_ALIST_2)
        MAX_HEAP_SIZE_FLAGS_SET(BIF_P, max_heap_flags);
        BIF_RET(old_value);
    }
+   else if (BIF_ARG_1 == am_heap_growth) {
+       old_value = (BIF_P->flags & F_HEAP_GROWTH_LOW) ? am_low : am_normal;
+       switch (BIF_ARG_2) {
+       case am_low:
+           BIF_P->flags |= F_HEAP_GROWTH_LOW;
+           break;
+       case am_normal:
+           BIF_P->flags &= ~F_HEAP_GROWTH_LOW;
+           break;
+       default:
+           goto error;
+       }
+       BIF_RET(old_value);
+   }
    else if (BIF_ARG_1 == am_message_queue_data) {
        old_value = erts_change_message_queue_management(BIF_P, BIF_ARG_2);
        if (is_non_value(old_value))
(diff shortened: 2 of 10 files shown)


KennethL:
Should this:
`spawn_opt(_, [{heap_growth}])`
be
`spawn_opt(_, [{heap_growth, low | normal}])`
??

lhoguin:
Can we have some context as to why you are adding this feature? In what scenarios will it provide benefits?

starbelly:
Hey @sverker , this seems terribly useful. If I'm grokking correctly, this is terribly useful in the context of long lived processes which may have to consume lots of messages but do hold them in stay (i.e., proc heap). Is that right?

sverker:
The idea of this PR was to optimize process heap memory consumption at the expense of CPU load caused by more frequent GCs when a process **creates lots of live data**.

With the current (normal) GC strategy, we exponentially grow the size of the new-heap based on the amount of terms that survived the _last_ GC. This is done before we know how much data will survive _this_ GC. We assume the process will continue to create lots of live data and we want to give it more heap capacity to do that. A larger heap will prolong the time until next GC and that will also give the data more time to die.

If the process then stops creating live data and enters a more passive period, then it might get "stuck" with a large inflated new-heap with lots of capacity that it does not use.

The `low` heap growth option introduced by this PR will disable that exponential growth for anticipated heap use and instead lets the process work with a more or less constant sized new-heap.

We had a real use case where memory consumption was a high priority concern. However, they gained almost nothing from this PR when tested. That's why it has not been merged.

starbelly:
> The idea of this PR was to optimize process heap memory consumption at the expense of CPU load caused by more frequent GCs when a process **creates lots of live data**.
> 
> With the current (normal) GC strategy, we exponentially grow the size of the new-heap based on the amount of terms that survived the _last_ GC. This is done before we know how much data will survive _this_ GC. We assume the process will continue to create lots of live data and we want to give it more heap capacity to do that. A larger heap will prolong the time until next GC and that will also give the data more time to die.
> 
> If the process then stops creating live data and enters a more passive period, then it might get "stuck" with a large inflated new-heap with lots of capacity that it does not use.
> 
> The `low` heap growth option introduced by this PR will disable that exponential growth for anticipated heap use and instead lets the process work with a more or less constant sized new-heap.

Precisely. We have processes in our system at work that suffer because of this, such that we have to set fullsweep_after to a very low value (or zero in some cases), in addition to some more expensive work arounds. This is mainly per passing complex terms to processes and then putting them in ets. We are actually going to make changes to one such set of procs that gets around all this via passing and storing serialized binaries via term_to_binary/1 and friends. 

That said, we have some other procs that could maybe benefit from this.

> 
> We had a real use case where memory consumption was a high priority concern. However, they gained almost nothing from this PR when tested. That's why it has not been merged.

 I'm happy to try a patch out if there's still interest in this.

starbelly:
> > I'm happy to try a patch out if there's still interest in this.
> 
> This branch is currently based on OTP-24.3, what do you need? You can always try merge forward to newer versions. I don't think the merge conflicts you might get will be particularly hard to solve.

I can merge it forward I'm quite sure, I'll report back soon.

starbelly:
Hey @sverker it took me a minute on this, but I can say I saw no substantial improvement.

Issue #4877, opened 2021-05-23
TLS handshake issues once DST Root CA X3 expires


voltone:
**Describe the bug**

Let's Encrypt, probably the most popular certificate authority these days, is now issuing certificates with a chain that includes a cross-signed certificate (details further below). Once the *DST Root CA X3* root certificate expires on September 30th, TLS clients using `ssl` will fail to connect to servers with such certificates. 

If users are caught unaware of this issue, I expect a lot of production environments are going to break on September 30th due to the widespread use of Let's Encrypt certificates. The only two options available as it stands are:

* Configure the server to not send the cross-signed CA certificate; this does not work for servers that are not under one's control
* Remove the *DST Root CA X3* certificate from the trust store prior to September 30th; this will work with clients that use `partial_chain` functions to handle alternate chains, but requires editing/forking the CA trust store

**To Reproduce**

The [future.voltone.net](future.voltone.net) server simulates the situation after the *DST Root CA X3* root certificate expires, using Let's Encypt staging certificates. The CA certificates and trust store for this environment can be found [here](https://gist.github.com/voltone/b0ff56227f342833900765e99dbee1a0).

```erlang
ssl:start().

ssl:connect("future.voltone.net", 443, [
  {verify, verify_peer},
  %% Trust store with both root CA certificates:
  {cacertfile, "trust_store.pem"},
  {depth, 3}
]).
%% --> {error,{tls_alert,{certificate_expired, ...}}}

ssl:connect("future.voltone.net", 443, [
  {verify, verify_peer},
  %% Only the new (short chain) root CA:
  {cacertfile, "new_root_ca.pem"},
  {depth, 3}
]).
%% --> {error,{tls_alert,{unknown_ca, ...}}}
```

**Expected behavior**

Ideally `ssl` would itself look for alternate ways of completing the chain using one of the certificates in its trust store, which would allow both scenarios above to pass without a `partial_chain` option. That would match how modern browsers and TLS implementations in other languages handle cross-signed certificates.

[...]

PR #4941, opened 2021-06-07
Ingela/ssl/alt certs/gh 4877/otp 17475

IngelaAndin:
This PR makes ssl application work the same as OpenSSL s_client  for the provided example in #4877 
Regression tests and possible code documentation is still to be added.

### DIFF
--- a/lib/public_key/src/public_key.erl
+++ b/lib/public_key/src/public_key.erl
@@ -1110,18 +1110,18 @@ pkix_path_validation(TrustedCert, CertChain, Options)
 pkix_path_validation(#'OTPCertificate'{} = TrustedCert, CertChain, Options)
   when is_list(CertChain), is_list(Options) ->
     MaxPathDefault = length(CertChain),
-    {VerifyFun, Userstat0} =
+    {VerifyFun, UserState0} =
 	proplists:get_value(verify_fun, Options, ?DEFAULT_VERIFYFUN),
-    try pubkey_cert:validate_time(TrustedCert, Userstat0, VerifyFun) of
-        Userstate1 -> 
+    try pubkey_cert:validate_time(TrustedCert, UserState0, VerifyFun) of
+        UserState1 -> 
             ValidationState = pubkey_cert:init_validation_state(TrustedCert, 
                                                                 MaxPathDefault, 
-                                                                [{verify_fun, {VerifyFun, Userstate1}} | 
+                                                                [{verify_fun, {VerifyFun, UserState1}} | 
                                                                  proplists:delete(verify_fun, Options)]),
             path_validation(CertChain, ValidationState)
     catch
-        throw:{bad_cert, cert_expired} = Reason ->
-            {error, Reason}    
+        throw:{bad_cert, _} = Result ->
+            {error, Result}
     end.
 
 %--------------------------------------------------------------------
--- a/lib/ssl/src/ssl_certificate.erl
+++ b/lib/ssl/src/ssl_certificate.erl
@@ -22,7 +22,43 @@
 %% Purpose: Help funtions for handling certificat verification.
 %% The path validation defined in ssl_handshake.erl that mainly
 %% calls functions in this module is described in RFC 3280. 
-%%----------------------------------------------------------------------
+%% The basic verification checks are done by
+%% public_key:pkix_path_validation/3
+%%
+%% TLS code handles construction of alternative certificate paths
+%% that can be used as input to public_key:pkix_path_validation/3
+%% to try and find one path that is considerd valid.
+%%
+%% The TLS protocol will send certificate chains that should consist
+%% of [PeerCert, CA0 ... CAN, ROOTCert]. The path will be the reverse of the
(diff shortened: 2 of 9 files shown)


voltone:
I can confirm that this enables support for the kind of cross-signing that Let's Encrypt are doing, and that we might see elsewhere as various root CAs are expiring later this year (starting with DST Root CA X3). It handles scenarios where the old root CA is present but expired, or where it is absent from the CA bundle.

In my testing, clients that use a `partial_chain` callback to handle cross-signed certificates today will continue to work, and in the future they will be able to drop that option altogether on OTP versions that include this PR.

Thanks! 🎉 If we can get this into users' hands before September 30th I think it will prevent a lot of breakage that day. I will do my best to raise awareness in the community once this PR makes it into a release.

IngelaAndin:
Added test case and some documentation. Still working on some further code enhancements before merging this to main and master. We are considering backporting it to the OTP-23 track, but that probably would need some further commits being backported to keep things simple.

lib/public_key/src/public_key.erl
```diff
@@ -1110,18 +1110,18 @@ pkix_path_validation(TrustedCert, CertChain, Options)
 pkix_path_validation(#'OTPCertificate'{} = TrustedCert, CertChain, Options)
   when is_list(CertChain), is_list(Options) ->
     MaxPathDefault = length(CertChain),
-    {VerifyFun, Userstat0} =
+    {VerifyFun, UserState0} =
 	proplists:get_value(verify_fun, Options, ?DEFAULT_VERIFYFUN),
-    try pubkey_cert:validate_time(TrustedCert, Userstat0, VerifyFun) of
+    try pubkey_cert:validate_time(TrustedCert, UserState0, VerifyFun) of
         Userstate1 ->
```
zuiderkwast, line 1116: Userstate1 is not camel-case. Perhaps change this one too?

lib/ssl/src/ssl_certificate.erl
```diff
@@ -295,13 +347,33 @@ do_certificate_chain(CertDbHandle, CertsDbRef, Chain, SerialNr, Issuer, _, ListD
 	    {ok, undefined, lists:reverse(Chain)}
     end.
 
+find_alternative_root([Cert | _], CertDbHandle, CertDbRef, InvalidatedList) ->
+    OtpCert = public_key:pkix_decode_cert(Cert, otp),
+    case find_issuer(#cert{der=Cert, otp=OtpCert}, CertDbHandle, CertDbRef, [], InvalidatedList) of
+        {error, issuer_not_found} ->
+            unknown_ca;
+        {ok, {SerialNr, IssuerId}} ->
+                case ssl_manager:lookup_trusted_cert(CertDbHandle, CertDbRef, SerialNr, IssuerId) of
```
zuiderkwast, line 356: Too much indentation here?

zuiderkwast (commented):
Spotted minor style issues only. I didn't check what the PR actually does.

lib/public_key/src/public_key.erl
IngelaAndin: Well sure why not polish this too

lib/ssl/src/ssl_certificate.erl
IngelaAndin: Strange Emacs agrees and that is what I use, will fix!

IngelaAndin:
Updated after a code review, expect to merge tomorrow.  Patches can be expected during July. Some delays are expected due to summer vacations.

lib/ssl/test/ssl_cert_SUITE.erl
```diff
@@ -98,8 +98,10 @@
          unsupported_sign_algo_cert_client_auth/1,
          longer_chain/0,
          longer_chain/1,
-         duplicate_chain/0,
-         duplicate_chain/1,
+         cross_singned_chain/0,
+         cross_singned_chain/1,
```
u3s, line 102: type-o, should be cross_signed in lines above?

u3s (approved):

lib/ssl/test/ssl_cert_SUITE.erl
IngelaAndin: Fixed the typo.

lib/ssl/src/ssl_certificate.erl
```diff
@@ -42,7 +42,8 @@
 	 select_extension/2,
 	 extensions_list/1,
 	 public_key_type/1,
-	 foldl_db/3
+	 foldl_db/3,
+         find_alternative_paths/5
```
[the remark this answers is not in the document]
zuiderkwast: It's spaces/tabs. `find_alternative_paths` has spaces while the old lines have tabs. Looks good in emacs but not in GitHub.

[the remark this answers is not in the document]
IngelaAndin: @zuiderkwast  Interesting. I "folded" the export list and "re-tabed" it in emacs and now it looks good on github too.

PR #4947, opened 2021-06-08
Fix inconsistency in defining ?FILE macro 

pergu:
This happens when the source_name is explicitly defined as an option to epp:parse_file/2. Before an include it got expanded to the actual file name, after an include it was defined as the source_name. With this change it will always be the source_name.

This come up when using --deterministic mode to compile causing the output to be less deterministic if you use the ?FILE macro before any includes.

### DIFF
--- a/lib/mnesia/src/mnesia_monitor.erl
+++ b/lib/mnesia/src/mnesia_monitor.erl
@@ -136,7 +136,7 @@ unsafe_create_external(Tab, Alias, Mod, Cs) ->
 disconnect(Node) ->
     cast({disconnect, Node}).
 
-%% Returns GoodNoodes
+%% Returns GoodNodes
 %% Creates a link to each compatible monitor and
 %% protocol_version to agreed version upon success
 


lib/stdlib/test/epp_SUITE.erl
```diff
@@ -113,6 +113,16 @@ include_local(Config) when is_list(Config) ->
...
     ok.
 
+file_macro(Config) when is_list(Config) ->
+    DataDir = proplists:get_value(data_dir, Config),
+    File = filename:join(DataDir, "file_macro.erl"),
+    {ok, List} = epp:parse_file(File, [DataDir], [{source_name, "Other source"}]),
+    %% Both attribute a and b are defined as ?FILE, they should be the same
+    {value, {attribute,_,a,FileA}} = lists:keysearch(a,3,List),
+    {value, {attribute,_,b,FileB}} = lists:keysearch(b,3,List),
+    FileA = FileB,
```
zuiderkwast, line 123: Why only check that they're the same? More explicit would be to also check that they're the expected value:

```suggestion
    "Other source" = FileA = FileB,
```

pergu: That's a great idea!

lib/stdlib/test/epp_SUITE.erl
```diff
@@ -113,6 +113,16 @@ include_local(Config) when is_list(Config) ->
         [ FileLine || {attribute,_,file,FileLine} <- List ],
     ok.
 
+file_macro(Config) when is_list(Config) ->
+    DataDir = proplists:get_value(data_dir, Config),
+    File = filename:join(DataDir, "file_macro.erl"),
+    {ok, List} = epp:parse_file(File, [DataDir], [{source_name, "Other source"}]),
+    %% Both attribute a and b are defined as ?FILE, they should be the same
+    {value, {attribute,_,a,FileA}} = lists:keysearch(a,3,List),
+    {value, {attribute,_,b,FileB}} = lists:keysearch(b,3,List),
```
bjorng, line 122:
```suggestion
    {attribute,_,a,FileA} = lists:keyfind(a, 3, List),
    {attribute,_,b,FileB} = lists:keyfind(b, 3, List),
```

lib/stdlib/test/epp_SUITE.erl
```diff
@@ -63,16 +63,16 @@ suite() ->
...
-all() -> 
+all() ->
     [rec_1, {group, upcase_mac}, include_local, predef_mac,
      {group, variable}, otp_4870, otp_4871, otp_5362, pmod,
      not_circular, skip_header, otp_6277, otp_7702, otp_8130,
      overload_mac, otp_8388, otp_8470, otp_8562,
      otp_8665, otp_8911, otp_10302, otp_10820, otp_11728,
      encoding, extends, function_macro, test_error, test_warning,
-     otp_14285, test_if, source_name, otp_16978, otp_16824, scan_file].
+     otp_14285, test_if, source_name, otp_16978, otp_16824, scan_file, file_macro].
```
bjorng, line 73: Have you tried to run this test suite? `file_macro/1` needs to be exported or the test case will fail.

lib/stdlib/test/epp_SUITE.erl
```diff
@@ -113,6 +113,16 @@ include_local(Config) when is_list(Config) ->
         [ FileLine || {attribute,_,file,FileLine} <- List ],
     ok.
 
+file_macro(Config) when is_list(Config) ->
+    DataDir = proplists:get_value(data_dir, Config),
+    File = filename:join(DataDir, "file_macro.erl"),
+    {ok, List} = epp:parse_file(File, [DataDir], [{source_name, "Other source"}]),
```
bjorng, line 119: The third argument for `epp:parse_file/3` is not the option list, but the predefined macros. You will need to use `epp:parse_file/2`.

```suggestion
    {ok, List} = epp:parse_file(File, [{includes, [DataDir]},
                                       {source_name, "Other source"}]),
```

bjorng (changes requested):
Thanks for your pull request.

The correction is correct, but the test case is broken (in two ways). It seems that you have never tried to run the test case.

Here are the [instructions for running tests](https://github.com/erlang/otp/wiki/Running-tests).

When you fixed the test case, please squash the commits and follow the guidelines for [writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages). Also include the following sentence from your description of the pull request in your commit message:

> This come up when using --deterministic mode to compile causing the output to be less deterministic if you use the ?FILE macro before any includes.

Such information is useful to have in the commit message if someone wonders in ten years what prompted this change.

lib/stdlib/test/epp_SUITE.erl
pergu: Exported file_macro, and have run the test now and it passes, sorry about that.

lib/stdlib/test/epp_SUITE.erl
pergu: Acknowledged

PR #4952, opened 2021-06-09
stdlib: Add gen_server:format_status/1

### LINKED ISSUE #4673: a crashing gen_server should be able to control last message formatting
**Describe the bug**
When a gen_server crashes, a crash report is output which combines the last message received ("Last message in was") with the server's state. The latter is passed through the server's optional `format_status/2` callback, which allows it to limit the amount of detail logged. Unfortunately, the server's callback module has no say in how to format the last message. The only control I've found is the `kernel` application's `error_logger_format_depth` variable, but that applies to much more than just the last message. The end result is that if gen_servers crash while receiving large messages (and the crashes may be due to unrelated issues such as destinations being down), considerable resources may be spent formatting and enqueuing crash reports, ultimately causing nodes to run out of memory.

**To Reproduce**
Send a large message to a gen_server that crashes. Observe the message being included in the crash report.

**Expected behavior**
A gen_server should be able to control, via a callback, how the last message is formatted in crash reports.

**Affected versions**
n/a

**Additional context**
We may end up using `error_logger_format_depth`, but it's a blunt instrument.

garazdawi:
This new format_status callback can also be used to format the messages in the sys logger or crash reports. It is
built-in order to make it possible to extend in the future with more values.

I've opened this PR in order to get some feedback on the new API before starting working on gen_statem/gen_event and tests for all of this.

Example usage: 

```erlang
-spec format_status(Status) -> NewStatus when
  Status :: { state => term(),
              message => term(),
              reason => term(),
              logs => [sys:system_event()] },
  NewStatus :: Status.
format_status(Status) ->
  maps:map(
    fun(state,State) ->
            maps:remove(private_key, State);
       (message,{password, _Pass}) ->
            {password, removed};
       (_,Value) ->
            Value
    end, Status).
```

Fixes #4673

### DIFF
--- a/lib/stdlib/doc/src/gen_event.xml
+++ b/lib/stdlib/doc/src/gen_event.xml
@@ -835,6 +835,77 @@ gen_event:stop             ----->  Module:terminate/2
     </func>
 
     <func>
+      <name since="OTP @OTP-17351@">Module:format_status(Status) -> NewStatus</name>
+      <fsummary>Optional function for providing a term describing the
+        current event handler state.</fsummary>
+      <type>
+        <v>Status = #{ state => State,</v>
+        <v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        message => Message,</v>
+        <v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        reason => Reason,</v>
+        <v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        log => Log }</v>
+        <v>NewStatus = Status</v>
+        <v>State = Message = Reason = term()</v>
+        <v>Log = [<seetype marker="sys#system_event">sys:system_event()</seetype>]</v>
+      </type>
+      <desc>
+        <note>
+          <p>This callback is optional, so event handler modules need
+            not export it. If a handler does not export this function,
+            the <c>gen_event</c> module uses the handler state directly for
+            the purposes described below.</p>
+          <p>
+	    If this callback is exported but fails, to hide possibly sensitive
+            data, the default function will instead return the fact that
+	    <c>format_status/1</c> has crashed.</p>
+        </note>
+        <p>This function is called by a <c>gen_event</c> process in the
+          following situations:</p>
+        <list type="bulleted">
+          <item>One of <seemfa marker="sys#get_status/1">
+            <c>sys:get_status/1,2</c></seemfa>
+            is invoked to get the <c>gen_event</c> status.</item>
+          <item>The event handler terminates abnormally and <c>gen_event</c>
+            logs an error.</item>
+        </list>
+        <p>This callback is to be used to limit the status of the event handler before
+          returned by <seemfa marker="sys#get_status/1"><c>sys:get_status/1,2</c></seemfa> or
+          sent to <seeerl marker="kernel:logger"><c>logger</c></seeerl>. The status
+          of the event handler is passed as a map with the following associations:
+        </p>
+        <taglist>
+          <tag><c>state => term()</c></tag>
+          <item>The internal state of the event handler.</item>
+          <tag><c>message => term()</c></tag>
+          <item>The message that caused the event handler to terminate.</item>
+          <tag><c>reason => term()</c></tag>
+          <item>The reason that caused the event handler to terminate.</item>
+          <tag><c>log => [</c><seetype marker="sys#system_event"><c>sys:system_event/0</c></seetype><c>]</c></tag>
+          <item>The <seemfa marker="sys#log/2">sys log</seemfa> of the server.</item>
+        </taglist>
+        <p>New associations may be added into the status map without prior notice.</p>
+        <p>The function is to return <c>NewStatus</c>, a map containing the same
+          associations as the input map.</p>
+        <p>Two possible use cases for this callback is to either remove
+          sensitive information from the state so that it not printed in log files,
+          or to remove large irrelevant terms of state that clutter the logs.</p>
+        <code type="erl"><![CDATA[
+format_status(Status) ->
+  maps:map(
+    fun(state,State) ->
+            maps:remove(private_key, State);
+       (message,{password, _Pass}) ->
+            {password, removed};
+       (_,Value) ->
+            Value
+    end, Status).
+]]></code>
+      </desc>
+    </func>
+    
+    <func>
       <name since="OTP R14B">Module:format_status(Opt, [PDict, State]) -> Status</name>
       <fsummary>Optional function for providing a term describing the
         current event handler state.</fsummary>
@@ -845,6 +916,11 @@ gen_event:stop             ----->  Module:terminate/2
         <v>Status = term()</v>
       </type>
       <desc>
+        <warning>
+          <p>This callback is deprecated, in new code use <seemfa marker="#Module:format_status/1">
+          format_status/1</seemfa>. If a <seemfa marker="#Module:format_status/1">format_status/1</seemfa>
+            callback exists, then this function will never be called.</p>
+        </warning>
         <note>
           <p>This callback is optional, so event handler modules need
             not export it. If a handler does not export this function,
--- a/lib/stdlib/doc/src/gen_server.xml
+++ b/lib/stdlib/doc/src/gen_server.xml
@@ -788,6 +788,79 @@ gen_server:abcast     -----> Module:handle_cast/2
     </func>
 
     <func>
+      <name since="OTP @OTP-17351@">Module:format_status(Status) -> NewStatus</name>
+      <fsummary>Optional function for providing a term describing the
+        current <c>gen_server</c> status.</fsummary>
+      <type>
+        <v>Status = #{ state => State,</v>
+        <v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        message => Message,</v>
+        <v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        reason => Reason,</v>
+        <v>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        log => Log }</v>
+        <v>NewStatus = Status</v>
+        <v>State = Message = Reason = term()</v>
+        <v>Log = [<seetype marker="sys#system_event">sys:system_event()</seetype>]</v>
+      </type>
+      <desc>
+        <note>
+          <p>This callback is optional, so callback modules need not
+            export it. The <c>gen_server</c> module provides a default
+            implementation of this function that returns the callback
+            module state.</p>
+          <p>
+	    If this callback is exported but fails, to hide possibly sensitive
+            data, the default function will instead return the fact that
+	    <c>format_status/1</c> has crashed.</p>
+        </note>
+        <p>This function is called by a <c>gen_server</c> process in the following situations:</p>
+        <list type="bulleted">
+          <item>
+            <p><seemfa marker="sys#get_status/1"><c>sys:get_status/1,2</c></seemfa>
+              is invoked to get the <c>gen_server</c> status.</p>
+          </item>
+          <item>
+            <p>The <c>gen_server</c> process terminates abnormally and logs an error.</p>
+          </item>
+        </list>
+        <p>This callback is to be used to limit the status of the process before
+          returned by <seemfa marker="sys#get_status/1"><c>sys:get_status/1,2</c></seemfa> or
+          sent to <seeerl marker="kernel:logger"><c>logger</c></seeerl>. The status
+          of the server is passed as a map with the following associations:
+        </p>
+        <taglist>
+          <tag><c>state => term()</c></tag>
+          <item>The internal state of the <c>gen_server</c> process.</item>
+          <tag><c>message => term()</c></tag>
+          <item>The message that caused the server to terminate.</item>
+          <tag><c>reason => term()</c></tag>
+          <item>The reason that caused the server to terminate.</item>
+          <tag><c>log => [</c><seetype marker="sys#system_event"><c>sys:system_event/0</c></seetype><c>]</c></tag>
+          <item>The <seemfa marker="sys#log/2">sys log</seemfa> of the server.</item>
+        </taglist>
+        <p>New associations may be added into the status map without prior notice.</p>
+        <p>The function is to return <c>NewStatus</c>, a map containing the same
+          associations as the input map.</p>
+        <p>Two possible use cases for this callback is to either remove
+          sensitive information from the state so that it not printed in log files,
+          or to remove large irrelevant terms of state that do clutter the logs.</p>
+        <p>Example:</p>
+        <code type="erl"><![CDATA[
+format_status(Status) ->
+  maps:map(
+    fun(state,State) ->
+            maps:remove(private_key, State);
+       (message,{password, _Pass}) ->
+            {password, removed};
+       (_,Value) ->
+            Value
+    end, Status).
+]]></code>
+      </desc>
+    </func>
+
+    <func>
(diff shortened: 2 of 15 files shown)


garazdawi:
Another question that just came to me, do we want to be able to do the same thing for `reason`?

lib/stdlib/src/gen_server.erl
```diff
@@ -1226,23 +1233,53 @@ format_status(Opt, StatusData) ->
...
+-type format_status() :: #{ opt := terminate | normal,
+                            state => term(),
+                            dictionary => [{term(),term()}],
+                            message => term() }.
+-spec limit_status(Mod :: module(), Status :: format_status()) -> format_status().
+limit_status(Mod, Status) ->
+    DefStatus =
+        case maps:get(opt, Status) of
+            terminate -> Status;
+            _ -> Status#{ state := [{data, [{"State", maps:get(state, Status, undefined)}]}]}
```
iilyak, line 1261: Status  includes the copy of a process dictionary which could contain sensitive data.
In previous version of the code the `DefaultStatus` (on line 1238) didn't include PDict. Now it does.

garazdawi: If you check [this line](https://github.com/erlang/otp/pull/4952/files#diff-1795d892ea0ec0b373b04026a3c449d436b22b1684eb5e69844efba22e86e305R1224), it is only the `state` that is actually returned when you do `sys:get_status`.

The `PDict` is included in the incoming status as it was available to the old implementation. Maybe I should just remove it from the `Status` as the user can always fetch it using `get()` themselves.

josevalim:
> Another question that just came to me, do we want to be able to do the same thing for reason?

Regardless of the answer, it may be worth noting that new keys may be added to the map, and therefore users should have a catch-all clause?

garazdawi:
> Regardless of the answer, it may be worth noting that new keys may be added to the map, and therefore users should have a catch-all clause? This would allow us to decide that and revisit keys in the future.

Yes. The reason why we cannot really re-use the current callback is that we would break all currently working code. So this new alternative must be extensible so that we don't end up in the same situation again.

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -787,6 +787,83 @@ gen_server:abcast     -----> Module:handle_cast/2
...
+        <p>One use for this function is to return compact alternative
+          state representations to avoid that large state terms are
+          printed in log files.</p>
+        <p>Example:</p>
+        <code type="erl"><![CDATA[
+format_status(Status) ->
+  maps:map(
+    fun(state,State) ->
+            case maps:get(opt,Status) of
+                normal -> [{data, [{"State", State}]}];
```
josevalim, line 855: I find the `opt` option and how it affects the output a bit hard to explain and reason about.

I thought about a couple different approaches and what do you think about this: in the output `Status` map, there may be a key called `sys`. If this key is available, `sys:get_status/1` will use it, otherwise it will use the `state` key wrapped in `{data, [{"State", ...}]}`. This means the `format_status/1` message handling code could be:

```erlang
format_status(Status) ->
  maps:map(
       (message,{password, _Pass}) ->
            {password, removed};
       (_,Value) ->
            Value
    end, Status).
```

If someone is interested in modifying the `sys` output, then they can add the key. And if someone is interested in modifying the `sys` output but it is expensive to compute, they can use the `opt` key to know if the `sys` key should be computed or not.

What do you think?

garazdawi: hmm, yes, I like this approach. Good idea. However, I think maybe I will name the value returned `status`.

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -834,6 +834,77 @@ gen_event:stop             ----->  Module:terminate/2
...
+          <item>The message that caused the event handler to terminate.</item>
+          <tag><c>reason => term()</c></tag>
+          <item>The reason that caused the event handler to terminate.</item>
+          <tag><c>log => [</c><seetype marker="sys#system_event"><c>sys:system_event/0</c></seetype><c>]</c></tag>
+          <item>The <seemfa marker="sys#log/2">sys log</seemfa> of the server.</item>
+        </taglist>
+        <p>New associations may be added into the status map without prior notice.</p>
+        <p>The function is to return <c>NewStatus</c>, a map containing the same
+          associations as the input map.</p>
+        <p>Two possible usecases for this callback is to either remove
```
bjorng, line 891:
```suggestion
        <p>Two possible use cases for this callback is to either remove
```

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -834,6 +834,77 @@ gen_event:stop             ----->  Module:terminate/2
...
+          sensitive information from the state so that it not printed in log files,
+          or to remove large irrelevant terms of state that do clutter the logs.</p>
```
bjorng, line 893:
```suggestion
          or to remove large irrelevant terms of state that clutter the logs.</p>
```

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -787,6 +787,79 @@ gen_server:abcast     -----> Module:handle_cast/2
...
+          <item>The message that caused the server to terminate.</item>
+          <tag><c>reason => term()</c></tag>
+          <item>The reason that caused the server to terminate.</item>
+          <tag><c>log => [</c><seetype marker="sys#system_event"><c>sys:system_event/0</c></seetype><c>]</c></tag>
+          <item>The <seemfa marker="sys#log/2">sys log</seemfa> of the server.</item>
+        </taglist>
+        <p>New associations may be added into the status map without prior notice.</p>
+        <p>The function is to return <c>NewStatus</c>, a map containing the same
+          associations as the input map.</p>
+        <p>Two possible usecases for this callback is to either remove
```
bjorng, line 845:
```suggestion
        <p>Two possible use cases for this callback is to either remove
```

lib/stdlib/doc/src/gen_statem.xml
```diff
@@ -2475,6 +2475,102 @@ init(Args) -> erlang:error(not_implemented, [Args]).</pre>
...
+          <tag><c>postponed => [{</c><seetype marker="#event_type"><c>event_type()</c></seetype><c>,term()}]</c></tag>
+          <item>The <seetype marker="#postpone">postponed</seetype> events queue of the <c>gen_statem</c> process.</item>
+          <tag><c>log => [</c><seetype marker="sys#system_event"><c>sys:system_event/0</c></seetype><c>]</c></tag>
+          <item>The <seemfa marker="sys#log/2">sys log</seemfa> of the server.</item>
+        </taglist>
+        <p>New associations may be added into the status map without prior notice.</p>
+        <p>The function is to return <c>NewStatus</c>, a map containing the same
+          associations as the input map.</p>
+	<p>
+	  One use for this function is to return compact alternative
```
bjorng, line 2554:
```suggestion
	  One use case for this function is to return compact alternative
```

lib/stdlib/src/gen.erl
```diff
@@ -546,3 +548,41 @@ format_status_header(TagLine, RegName) when is_atom(RegName) ->
...
+      ReturnStatus :: #{ atom() => term(), '$status' => term()},
+      Args :: list(term()) | undefined.
+format_status(Mod, Opt, Status, Args) ->
+    case {erlang:function_exported(Mod, format_status, 1),
+          erlang:function_exported(Mod, format_status, 2)} of
+        {true, _} ->
+            try Mod:format_status(Status) of
+                NewStatus when is_map(NewStatus) ->
+                    MergedStatus = maps:merge(Status, NewStatus),
+                    case maps:size(MergedStatus) == maps:size(NewStatus) of
```
bjorng, line 564:
```suggestion
                    case maps:size(MergedStatus) =:= maps:size(NewStatus) of
```

lib/stdlib/src/gen.erl
```diff
@@ -546,3 +548,41 @@ format_status_header(TagLine, RegName) when is_atom(RegName) ->
...
+                _ ->
+                    Status#{ 'EXIT' => atom_to_list(Mod) ++ ":format_status/1 did not return a map" }
+            catch
+                E:R:ST ->
+                    ?LOG_ERROR("Calling ~p:format_status/1 failed: ~tp:~tp~n~tp",
+                               [Mod, E, R, ST]),
+                    Status#{ 'EXIT' => atom_to_list(Mod) ++ ":format_status/1 crashed" }
+            end;
+        {false, true} when is_list(Args) ->
+            case catch Mod:format_status(Opt, Args) of
```
bjorng, line 581: I suggest using `try` instead of `catch` here.

lib/stdlib/src/gen_server.erl
```diff
@@ -939,19 +947,37 @@ terminate(Class, Reason, Stacktrace, ReportReason, Name, From, Msg, Mod, State,
...
+                             message => Msg,
+                             log => Log },
+                          [get(),State]),
+    ReportReason =
+        if ST == undefined ->
+                %% When ST is undefined, it should not be included in the
+                %% reported reason for the crash as it is then caused
+                %% by an invalid return from a callback and thus thus the
+                %% stacktrace is irrelevant.
+                maps:get(reason,Status);
```
bjorng, line 970:
```suggestion
                maps:get(reason, Status);
```

lib/stdlib/src/gen_server.erl
```diff
@@ -939,19 +947,37 @@ terminate(Class, Reason, Stacktrace, ReportReason, Name, From, Msg, Mod, State,
...
+           true ->
+                {maps:get(reason,Status), ST}
```
bjorng, line 972:
```suggestion
                {maps:get(reason, Status), ST}
```

lib/stdlib/src/gen_server.erl
```diff
@@ -1212,37 +1238,42 @@ mod(_) -> "t".
...
-	true ->
-	    case catch Mod:format_status(Opt, [PDict, State]) of
-		{'EXIT', _} -> DefStatus;
-		Else -> Else
-	    end;
-	_ ->
-	    DefStatus
+    %% If format_status/1 was exported, the log has already been handled by
+    %% that call, so we should not pass all log events into the callback again.
+    case erlang:function_exported(Mod,format_status,1) of
```
bjorng, line 1262:
```suggestion
    case erlang:function_exported(Mod, format_status, 1) of
```

lib/stdlib/src/gen_statem.erl
```diff
@@ -2741,34 +2791,34 @@ single(false) -> "".
...
+%%     end.
+
+%% %% The default Module:format_status/3
+%% format_status_default(Opt, State_Data) ->
+%%     case Opt of
+%% 	terminate ->
+%% 	    State_Data;
+%% 	_ ->
+%% 	    [{data,[{"State",State_Data}]}]
+%%     end.
```
bjorng, line 2821: Remove this code instead of commenting it out.

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -798,6 +871,10 @@ gen_server:abcast     -----> Module:handle_cast/2
         <v>Status = term()</v>
       </type>
       <desc>
+        <warning>
+          <p>This callback is deprecated, in new code use <seemfa marker="#Module:format_status/1">
+          format_status/1</seemfa>.</p>
```
bjorng, line 876: Add a note that the this old callback function will not be called if the new one is present.

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -845,6 +916,10 @@ gen_event:stop             ----->  Module:terminate/2
         <v>Status = term()</v>
       </type>
       <desc>
+        <warning>
+          <p>This callback is deprecated, in new code use <seemfa marker="#Module:format_status/1">
+          format_status/1</seemfa>.</p>
```
bjorng, line 921: Add a note that the this old callback function will not be called if the new one is present.

bjorng (commented):
Looks good to me.

lib/stdlib/src/gen_statem.erl
```diff
@@ -346,8 +348,20 @@
...
 
+%% Format the callback module status in some sensible that is
+%% often condensed way.
+-callback format_status(Status) -> NewStatus when
+      Status :: #{ state => state(),
+                   data => data(),
+                   reason => term(),
+                   queue => [{event_type(), term()}],
+                   postponed => [{event_type(), term()}],
+                   log => [sys:system_event()] },
```
RaimoNiskanen, line 359: Is not `timeouts => Something,` missing here?

lib/stdlib/src/gen.erl
```diff
@@ -546,3 +548,41 @@ format_status_header(TagLine, RegName) when is_atom(RegName) ->
...
+                        false ->
+                            ?LOG_ERROR("~p:format_status(~p) returned an invalid map",
+                                       [Mod, Status]),
+                            Status#{ 'EXIT' => atom_to_list(Mod) ++ ":format_status/1 returned an invalid map" }
+                    end;
+                _ ->
+                    Status#{ 'EXIT' => atom_to_list(Mod) ++ ":format_status/1 did not return a map" }
+            catch
+                E:R:ST ->
+                    ?LOG_ERROR("Calling ~p:format_status/1 failed: ~tp:~tp~n~tp",
```
garazdawi, line 576: We should remove this ?LOG_ERROR in order to not accidentally leak any state. The same with the invalid map log

lib/stdlib/src/gen_statem.erl
garazdawi: Yes, good catch. Missing both here and in the docs. Fixed.

lib/stdlib/src/gen_statem.erl
garazdawi: Fixed

lib/stdlib/src/gen.erl
garazdawi: fixed.

lib/stdlib/src/gen.erl
garazdawi: fixed.

lib/stdlib/src/gen.erl
```diff
@@ -546,3 +548,41 @@ format_status_header(TagLine, RegName) when is_atom(RegName) ->
...
+-spec format_status(Mod :: module(), Opt :: terminate | normal, Status, Args) ->
+          ReturnStatus when
+      Status :: #{ atom() => term() },
+      ReturnStatus :: #{ atom() => term(), '$status' => term()},
+      Args :: list(term()) | undefined.
+format_status(Mod, Opt, Status, Args) ->
+    case {erlang:function_exported(Mod, format_status, 1),
+          erlang:function_exported(Mod, format_status, 2)} of
+        {true, _} ->
+            try Mod:format_status(Status) of
```
iilyak, line 561: The PDict is not passed to format_status. Which makes following [code pattern](https://github.com/apache/couchdb/blob/0059b8f90e58e10b199a4b768a06a762d12a30d3/src/couch/src/couch_file.erl#L552:L554) impossible.

```
format_status(_Opt, [PDict, #file{} = File]) ->
    {_Fd, FilePath} = couch_util:get_value(couch_file_fd, PDict),
    [{data, [{"State", File}, {"InitialFilePath", FilePath}]}].
```

garazdawi: If you use this code:

```erlang
format_status(_Opt, [PDict, #file{} = File]) ->
    {_Fd, FilePath} = couch_util:get_value(couch_file_fd, get()),
    [{data, [{"State", File}, {"InitialFilePath", FilePath}]}].
```

you will get the result that you are after. Or am I missing some detail where that is not possible?

iilyak: Yes this would work, but according to the modified docs in this PR it would mean relying on [deprecated](https://github.com/erlang/otp/pull/4952/files#diff-72eeebfa525d4fd117bb50d045278a11e3695660abf9bc5cbe38581690241de3R920) functionality.

iilyak: Actually I didn't notice a slight change you did. Let me review again.

lib/stdlib/src/gen.erl
```diff
@@ -546,3 +548,41 @@ format_status_header(TagLine, RegName) when is_atom(RegName) ->
...
+                NewStatus when is_map(NewStatus) ->
+                    MergedStatus = maps:merge(Status, NewStatus),
+                    case maps:size(MergedStatus) =:= maps:size(NewStatus) of
+                        true ->
+                            MergedStatus;
+                        false ->
+                            Status#{ 'EXIT' => atom_to_list(Mod) ++ ":format_status/1 returned an map with unknown keys" }
```
iilyak, line 568: I think it should be `a map` instead of `an map`

```
":format_status/1 returned an map with unknown keys"
```

iilyak (approved):
Looks good to me

PR #4954, opened 2021-06-09
Added workflow_dispatch to main CI

vkatsuba:
Added [workflow_dispatch](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) to `main.yaml`

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -22,6 +22,7 @@ name: Build and check Erlang/OTP
 on:
   push:
   pull_request:
+  workflow_dispatch:
 
 jobs:
 


garazdawi:
Hello! Why would you like to add that?

vkatsuba:
> Hello! Why would you like to add that?

Hi. I was add it for ease of run of workflows for different branches manually. But since the launch occurs automatically for each PR, it is possible that this changes is unnecessary and you should not pay attention to it. So, up to you.

garazdawi:
I think I'll leave it as it for now. Thanks!

PR #4955, opened 2021-06-09
Refactoring of logger_disk_log_h

vkatsuba:
Minor changes:
* Remove `get_default_config/0` and merging with empty `map`
* Update copyright year of `logger_disk_log_h`
* Remove white spaces to try to keep one code style

### DIFF
--- a/lib/kernel/src/logger_disk_log_h.erl
+++ b/lib/kernel/src/logger_disk_log_h.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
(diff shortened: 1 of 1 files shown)


garazdawi:
Hello! Thanks for the PR!

The uniform coding style for `logger` is to *not* use spaces. So if you want to help make the styling more uniform, that is the direction you should take.

I'm not really a fan of the no-spaces style myself, but it is what the original author chose so we will stick with that for now.

vkatsuba:
> The uniform coding style for logger is to not use spaces. 

Done.

garazdawi:
thanks!

PR #4960, opened 2021-06-11
Fix A Bug That Causes The Test System To Not Run

skynet1:
The test system doesn't run because subsystem tests
don't know where ERL_TOP is.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -1095,7 +1095,7 @@ tests release_tests: $(TEST_DIRS)
 
 $(TEST_DIRS):
 	if test -f $@/Makefile; then \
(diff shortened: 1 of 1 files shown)


skynet1:
I hope I did this right. I had a little bit of trouble understanding the contribution rules, but I think I've got it right.

kjellwinblad:
Hi,

Thanks for the PR. I think you get this issue because the `ERL_TOP` variable has not been exported by running the following from the root of source code tree:

    export ERL_TOP=`pwd`

This is documented in [`HOWTO/TESTING.md`](https://github.com/erlang/otp/blob/master/HOWTO/TESTING.md). However, I don't think this fix can cause any problem so I think we can merge it after it has gone through our nightly tests.

PR #4961, opened 2021-06-11
supervisor: optimize restart intensity calculation

Maria-12648430:
This PR is just a micro-optimization and some cleaning =^^= It should be pretty obvious what has been done.

`add_restart/3` was just a customized re-invention of `lists:takewhile/2`, and as such even a misnomer, as it didn't _add_ anything. And by putting the new restart on the list (in `add_restart/1`) _before_ passing it to `add_restart/3` to be filtered caused an unnecessary filter step, as the new restart would never be removed anyway. Finally, the `inPeriod/3` function (used for checking if a restart was in or out of period) did a calculation that could really be replaced by a pre-calculated constant value, as the `Now` and `Period` arguments were always the same over a run of `add_restart/3`, and `Now =< Then + Period` could be rewritten as `Now - Period =< Then`.

This PR pre-calculates the treshold time beyond which restarts are considered as out of period, filters the old list of restarts by means of `lists:takewhile/2` using this treshold, and only then adds the new restart. This way, even the name `add_restart` starts to fit, and the `inPeriod/3` function became obsolete.

### DIFF
--- a/lib/stdlib/src/supervisor.erl
+++ b/lib/stdlib/src/supervisor.erl
@@ -1500,7 +1500,7 @@ add_restart(State) ->
     P = State#state.period,
     R = State#state.restarts,
     Now = erlang:monotonic_time(1),
-    R1 = add_restart([Now|R], Now, P),
+    R1 = add_restart(R, Now, P),
     State1 = State#state{restarts = R1},
     case length(R1) of
 	CurI when CurI  =< I ->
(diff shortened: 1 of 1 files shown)


bjorng:
Looks good to me.

Please add the explanation in the PR description to the commit message.

Maria-12648430:
@bjorng I did, ok now? :)

lib/stdlib/src/supervisor.erl
```diff
@@ -1509,18 +1509,13 @@ add_restart(State) ->
...
-	    []
-    end;
-add_restart([], _, _) ->
-    [].
-
-inPeriod(Then, Now, Period) ->
-    Now =< Then + Period.
+add_restart(Restarts0, Now, Period) ->
+    Treshold = Now - Period,
+    Restarts1 = lists:takewhile(
```
vkatsuba, line 1514: Just wondering if makes sense to replace `lists:takewhile/2` to a list comprehensions:
```erlang
...
Restarts1 = [R || R <- Restarts0, R >= Treshold],
...
```

Maria-12648430: The restarts in the list are naturally ordered newest to oldest, so we may stop processing early when we hit the first out-of-period timestamp, because anything beyond it will be even older. This is just what `lists:takewhile/2` does. A list comprehension would traverse the list in full, like `lists:filter/2` would.

bjorng:
> @bjorng I did, ok now? :)

I prefer your explanation in the description of the PR. It is longer, but it describes what is wrong with the old way and how you fixed it.

Maria-12648430:
Okay, I'll rewrite :)

Maria-12648430:
@bjorng the last commit has a more detailed description, mostly taken from the PR description.

bjorng:

bjorng:
Thanks for your pull request!

PR #4962, opened 2021-06-11
observer: add button to trace all processes and ports

legoscia:
This is more convenient than finding the right application and tracing its process tree.

Not sure about the button label: `Add all Processes and Ports` is quite wide and might make the buttons overflow the visible part of the window.

Also not sure what to do about trace options. Right now the new button pops up the trace option dialog for processes, which is not strictly correct as `all` includes ports as well.

image


dgud:
Hmm, I don't see anything better.
Making one button with another option dialog seems worse to me.

Maybe just name the button 'Add All' and have a tooltip that says "Trace all processes and ports" ?

legoscia:
Makes sense. I pushed a change for that.

PR #4977, opened 2021-06-17
ssl: Use new socket:monitor/1

IngelaAndin:
To be able to fully work with new gen_tcp socket backend.


bmk:
Why don't you use inet:monitor/1?

IngelaAndin:
@bmk as I did not know it existed, that is why you should review the code! Do I understand it correctly that it will work for both new and old sockets and that I then can skip the is_port guard in the original code?

bmk:
Yes, it should work for "all" kinds of sockets.
The function socket:monitor/1 *is* official, but is intended for when you *only* work with 'socket' sockets.

IngelaAndin:
Replaced by #5050

PR #4979, opened 2021-06-17
ssl: Fix signature algorithm selection for ECDSA

### LINKED ISSUE #4958: TLS 1.3: Server sends incorrect CertificateVerify for ECDSA certificates
**Describe the bug**
When the Erlang/OTP server authenticates with an ECDSA certificate, it will populate its CertificateVerify message with the first EDCSA signature algorithm in the intersection of client and server supported signature algorithms, without considering the curve of the certificate's public key algorithm.

This is strongly suspected to be due to how [select_sign_algo](https://github.com/erlang/otp/blob/master/lib/ssl/src/tls_handshake_1_3.erl#L2329) is defined. When `PublicKeyAlgo` is `ecdsa`, It returns the first algorithm in `PeerSignAlgs` whose scheme is `ecdsa`, since `validate_key_compatbility` doesn't perform any validation for non-RSA algorithms. 

Based on the code, this issue likely extends to EDDSA certificates as well, though I haven't tested this. 

**To Reproduce**
*Setup*
Run an Erlang/OTP server with an ECDSA certificate. I discovered this issue on a server with an ECDSA certificate with an `secp256r1` curve.

*Scenario 1*
Client sends ClientHello with the following in the ~~signature_algorithms_cert/~~ signature_algorithms extension: `ecdsa_secp521r1_sha512, ecdsa_secp384r1_sha384, ecdsa_secp256r1_sha256`.
Server sends CertificateVerify with `ecdsa_secp521r1_sha512` as the algorithm. 

*Scenario 2*
Client sends ClientHello with the following in the ~~signature_algorithms_cert/~~ signature_algorithms extension: `ecdsa_secp384r1_sha384, ecdsa_secp256r1_sha256, ecdsa_secp521r1_sha512`.
Server sends CertificateVerify with `ecdsa_secp384r1_sha384` as the algorithm. 

*Scenario 3*
Client sends ClientHello with the following in the ~~signature_algorithms_cert/~~ signature_algorithms extension: `ecdsa_secp256r1_sha256, ecdsa_secp521r1_sha512, ecdsa_secp384r1_sha384`.
Server sends CertificateVerify with `ecdsa_secp256r1_sha256` as the algorithm. 

**Expected behavior**
~~Server should send CertificateVerify with the `ecdsa_secp256r1_sha256` algorithm in all three scenarios, corresponding to the certificate.~~
Server should send CertificateVerify with an `algorithm` value that corresponds to the actual signature algorithm used for the `signature` in CertificateVerify. In the scenarios above, because the server signs with the `ecdsa_secp256r1_sha256` algorithm, `algorithm` should be `ecdsa_secp256r1_sha256`. 

**Affected versions**
This issue was discovered and reproduced on 23.1.1, but the code that would cause this is present in the current master branch. 

Edit: Remove mistaken inclusion of `signature_algorithms_cert` extension. Clarify expected behavior.

peterdmv:
The signature selection algorithm has been changed to also verify
if the client supports signatures using the elliptic curve of the
server's public/private key pair. This change fixes #4958.

The old algorithm only verified the signature algorithm and the
hash algorithm, not taking into account the elliptic curve of
the key in the server's certificate. This could result in
successful TLS connections with non-standard algorithm combinations
in the CertificateVerify message such as ECDSA with secp256r1 and
SHA512, when the server's certificate had a public key on an
unsupported elliptic curve, signed with a non-ECDSA signature
algorithm and hash algorithm combination e.g. RSA with SHA256.

TODO:

- [x] Handle explicit curve: secp384r1
- [x] Handle explicit curve: secp512r1

### DIFF
--- a/lib/ssl/src/ssl_handshake.erl
+++ b/lib/ssl/src/ssl_handshake.erl
@@ -1512,7 +1512,7 @@ select_hashsign({#hash_sign_algos{hash_sign_algos = ClientHashSigns},
                 Cert, KeyExAlgo, SupportedHashSigns, {Major, Minor})
   when Major >= 3 andalso Minor >= 3 ->
     ClientSignatureSchemes = get_signature_scheme(ClientSignatureSchemes0),
-    {SignAlgo0, Param, PublicKeyAlgo0, _} = get_cert_params(Cert),
+    {SignAlgo0, Param, PublicKeyAlgo0, _, _} = get_cert_params(Cert),
     SignAlgo = sign_algo(SignAlgo0),
     PublicKeyAlgo = public_key_algo(PublicKeyAlgo0),
 
@@ -1566,7 +1566,7 @@ select_hashsign(#certificate_request{
                 Cert,
                 SupportedHashSigns,
 		{Major, Minor})  when Major >= 3 andalso Minor >= 3->
-    {SignAlgo0, Param, PublicKeyAlgo0, _} = get_cert_params(Cert),
+    {SignAlgo0, Param, PublicKeyAlgo0, _, _} = get_cert_params(Cert),
     SignAlgo = sign_algo(SignAlgo0),
     PublicKeyAlgo = public_key_algo(PublicKeyAlgo0),
 
@@ -1589,7 +1589,7 @@ select_hashsign(#certificate_request{
 	    ?ALERT_REC(?FATAL, ?INSUFFICIENT_SECURITY, no_suitable_signature_algorithm)
     end;
 select_hashsign(#certificate_request{certificate_types = Types}, Cert, _, Version) ->
-    {_, _, PublicKeyAlgo0, _} = get_cert_params(Cert),
+    {_, _, PublicKeyAlgo0, _, _} = get_cert_params(Cert),
     PublicKeyAlgo = public_key_algo(PublicKeyAlgo0),
 
     %% Check cert even for TLS 1.0/1.1
@@ -1606,6 +1606,7 @@ select_hashsign(#certificate_request{certificate_types = Types}, Cert, _, Versio
 %% - parameters of the signature algorithm
 %% - public key algorithm (key type)
 %% - RSA key size in bytes
+%% - Elliptic Curve (public key)
 get_cert_params(Cert) ->
     #'OTPCertificate'{tbsCertificate = TBSCert,
 		      signatureAlgorithm =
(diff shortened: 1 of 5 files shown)


lib/ssl/src/ssl_handshake.erl
```diff
@@ -1621,7 +1622,101 @@ get_cert_params(Cert) ->
             _ ->
                 undefined
         end,
-    {SignAlgo, Param, PublicKeyAlgo, RSAKeySize}.
+    Curve = get_ec_curve(TBSCert#'OTPTBSCertificate'.subjectPublicKeyInfo),
+    {SignAlgo, Param, PublicKeyAlgo, RSAKeySize, Curve}.
+
+get_ec_curve(#'OTPSubjectPublicKeyInfo'{
+                algorithm = #'PublicKeyAlgorithm'{
+                               algorithm = {1,2,840,10045,2,1},
```
IngelaAndin, line 1630: public_key application provides macros for these OIDs  such as ?'id-ecPublicKey' 
you can find them in OTP-PUB-KEY.hrl in public_key's include directory

peterdmv: Done.

IngelaAndin:
@peterdmv  I believe this PR is breaking some of the property based tests due to the setup of test not taking all things into consideration. I believe that the following change should fix things as it will select a ECC-curve  that will be always supported by default. If you agree please add it to your PR. 

```
diff --git a/lib/ssl/test/property_test/ssl_eqc_chain.erl b/lib/ssl/test/property_test/ssl_eqc_chain.erl
index c063ef09f7..47288ed7d6 100644
--- a/lib/ssl/test/property_test/ssl_eqc_chain.erl
+++ b/lib/ssl/test/property_test/ssl_eqc_chain.erl
@@ -411,13 +411,13 @@ create_extraneous_and_unorded([Client, _CCA0, _CCA1, CCA2, _CCA3], [Client, OCCA
     {[Client, OCCA0, CCA2, OCCA2, OCROOT, OCCA1], [Server, OSCA0, SCA2, OSCA2, OSROOT, OSCA1]}.
                               
 root_key(ecdsa) ->
-    []; %% Just generate one
+    [{namedCurve, ?secp256r1}]; %% Use a curve that will be default supported in all TLS versions
 root_key(rsa) ->
     %% As rsa keygen is not guaranteed to be fast
     [{key, ssl_test_lib:hardcode_rsa_key(6)}].
 
 peer_key(ecdsa) ->
-    []; %% Just generate one
+    [{namedCurve, ?secp256r1}]; %% Use a curve that will be default supported in all TLS versions
 peer_key(rsa) ->
     %% As rsa keygen is not guaranteed to be fast
     [{key, ssl_test_lib:hardcode_rsa_key(6)}].

```

peterdmv:
@IngelaAndin Good catch, I didn't have Proper on my machine. :) I have tried the patch but the same test cases are failing in ssl_eqc_SUITE. Both the client and server stop with an internal error when signing their respective CertificateVerify messages with a key on an unsupported elliptic curve (OID: 1.3.132.0.9 - secp160k1) and sha1. I have pushed a fix for the handling of unsupported elliptic curves but I think there is a bug in ssl_eqc_SUITE as it can generate unsupported certs for TLS 1.3.

Server is trying to sign with illegal parameters (sha1, secp160k1):
```erlang
TLS client: In state wait_sh received SERVER ALERT: Fatal - Internal Error

(<0.5301.0>) call tls_handshake_1_3:sign(<<181,107,231,85,86,250,197,78,222,97,192,2,158,100,219,201,57,138,115,93,15,
  161,176,69,226,246,117,86,45,18,98,214,111,14,166,148,8,94,138,169,50,87,193,
  80,7,122,57,123>>,<<"TLS 1.3, server CertificateVerify">>,sha1,{'ECPrivateKey',1,
                <<0,37,225,108,88,72,125,189,112,123,174,164,96,185,123,159,
                  159,103,157,120,255>>,
                {namedCurve,{1,3,132,0,9}},
                <<4,60,22,198,209,231,166,111,126,203,195,163,43,9,56,145,208,
                  228,170,111,237,108,225,51,236,23,6,129,84,114,226,4,176,208,
                  20,238,228,125,202,158,44>>,
                asn1_NOVALUE},ecdsa) ({tls_handshake_1_3,certificate_verify,4})
=NOTICE REPORT==== 8-Jul-2021::18:26:04.782842 ===
TLS server: In state negotiated at tls_handshake_1_3.erl:579 generated SERVER ALERT: Fatal - Internal Error
 - badarg
(<0.5301.0>) returned from tls_handshake_1_3:sign/5 -> {error,
                                                        {alert,2,80,
                                                         #{file =>
                                                            "tls_handshake_1_3.erl",
                                                           line => 579,
                                                           mfa => 
                                                            {tls_handshake_1_3,
                                                             sign,5}},
                                                         undefined,badarg}}

```

ssl_eqc_SUITE generates secp160k1 key:
```erlang
(<0.7021.0>) call public_key:generate_key({namedCurve,{1,3,132,0,9}}) ({pubkey_cert,
                                                                        root_cert,
                                                                        2})
(<0.7021.0>) returned from public_key:generate_key/1 -> {'ECPrivateKey',1,
                                                         <<0,253,48,29,233,100,
                                                           29,158,93,169,166,
                                                           77,124,83,85,69,192,
                                                           227,121,110,188>>,
                                                         {namedCurve,
                                                          {1,3,132,0,9}},
                                                         <<4,220,56,146,228,86,
                                                           64,178,188,115,122,
                                                           37,127,230,191,96,
                                                           189,76,207,250,201,
                                                           61,94,243,185,231,
                                                           138,69,51,253,126,
                                                           87,59,97,221,147,
                                                           243,122,222,110,122>>,
                                                         asn1_NOVALUE}
```

Legacy signature algorithm selected:
```erlang
(<0.7095.0>) returned from tls_handshake_1_3:validate_key_compatibility/5 -> {ok,
                                                                              ecdsa_sha1}
(<0.7095.0>) returned from tls_handshake_1_3:select_sign_algo/5 -> {ok,
                                                                    ecdsa_sha1}
```

IngelaAndin:
@peterdmv Are you sure you ran the correct version of the test code?  My patch should fix that secp256r1 is chosen instead of secp160k1. Although we might need another fix to avoid the selection of sha1 too. I will look closer when I am back at work again.

peterdmv:
I have force pushed a fix for the property tests. In case of unsupported curves both client and server will return an "insufficient_security" alert.
With this latest patch, ```tls_handshake_1_3:validate_key_compatibility/5``` can only select a supported signature algorithm when using ECDSA: ```ecdsa_secp256r1_sha256```, ```ecdsa_secp384r1_sha384```, ```ecdsa_secp521r1_sha512```. ```sha1```cannot be selected.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -650,13 +650,15 @@ do_start(#client_hello{cipher_suites = ClientCiphers,
         Cipher = Maybe(select_cipher_suite(HonorCipherOrder, ClientCiphers, ServerCiphers)),
         Groups = Maybe(select_common_groups(ServerGroups, ClientGroups)),
         Maybe(validate_client_key_share(ClientGroups, ClientShares)),
-        {PublicKeyAlgo, SignAlgo, SignHash, RSAKeySize} = get_certificate_params(Cert),
+        {PublicKeyAlgo, SignAlgo, SignHash, RSAKeySize, Curve} = get_certificate_params(Cert),
 
         %% Check if client supports signature algorithm of server certificate
+        %% TODO: We do validate the signature algorithm and signature hash but we could check
```
IngelaAndin, line 656: You have a TODO comment here, I suppose we want to fix that?!

IngelaAndin (commented):
Looking good in nightly builds now :)

lib/ssl/src/tls_handshake_1_3.erl
peterdmv: Yes, I believe this can happen if a client connects to a faulty/malicious server that won't respect the signature_algorithms/signature_algorithms_cert extensions.

IngelaAndin:
We decided we need to merge this PR as we want to back port  it for OTP-23 with or without the last TODO enhancement. 
So either you make a new PR for the last enhancement or I (we) will implement it later. @peterdmv  we can sync it in some other forum!

PR #4982, opened 2021-06-18
Improve the readability of .asm-dumps

frej:
This is a set of patches which improve the readability of the
.asm-dumps, produced when running the emulator with the `+JDump true`
flag, by pretty-printing the labels marking the start of a function
using its <name>/<arity> Erlang name.

If this PR is accepted I can do an Arm-version of the last patch, but
I will require a volunteer to test it.

### DIFF
--- a/erts/emulator/beam/beam_file.c
+++ b/erts/emulator/beam/beam_file.c
@@ -323,24 +323,23 @@ static int parse_import_chunk(BeamFile *beam, IFF_Chunk *chunk) {
     return 1;
 }
 
-static int parse_export_chunk(BeamFile *beam, IFF_Chunk *chunk) {
-    BeamFile_ExportTable *exports;
+static int parse_export_table(BeamFile_ExportTable *dest,
+			      BeamFile *beam, IFF_Chunk *chunk) {
     BeamFile_AtomTable *atoms;
     BeamReader reader;
     Sint32 count;
     int i;
 
-    exports = &beam->exports;
-    ASSERT(exports->entries == NULL);
+    ASSERT(dest->entries == NULL);
 
     beamreader_init(chunk->data, chunk->size, &reader);
 
     LoadAssert(beamreader_read_i32(&reader, &count));
-    LoadAssert(CHECK_ITEM_COUNT(count, 0, sizeof(exports->entries[0])));
+    LoadAssert(CHECK_ITEM_COUNT(count, 0, sizeof(dest->entries[0])));
 
-    exports->entries = erts_alloc(ERTS_ALC_T_PREPARED_CODE,
-                                  count * sizeof(exports->entries[0]));
-    exports->count = count;
+    dest->entries = erts_alloc(ERTS_ALC_T_PREPARED_CODE,
+                                  count * sizeof(dest->entries[0]));
+    dest->count = count;
 
     atoms = &beam->atoms;
 
@@ -355,12 +354,16 @@ static int parse_export_chunk(BeamFile *beam, IFF_Chunk *chunk) {
         LoadAssert(arity >= 0 && arity <= MAX_ARG);
         LoadAssert(label >= 0);
 
-        exports->entries[i].function = atoms->entries[atom_index];
-        exports->entries[i].arity = arity;
-        exports->entries[i].label = label;
+        dest->entries[i].function = atoms->entries[atom_index];
+        dest->entries[i].arity = arity;
+        dest->entries[i].label = label;
     }
 
     return 1;
+}
+
+static int parse_export_chunk(BeamFile *beam, IFF_Chunk *chunk) {
+    return parse_export_table(&beam->exports, beam, chunk);
 }
 
 static int parse_lambda_chunk(BeamFile *beam, IFF_Chunk *chunk) {
--- a/erts/emulator/beam/beam_file.c
+++ b/erts/emulator/beam/beam_file.c
@@ -366,6 +366,12 @@ static int parse_export_chunk(BeamFile *beam, IFF_Chunk *chunk) {
     return parse_export_table(&beam->exports, beam, chunk);
 }
 
+#ifdef BEAMASM
+static int parse_locals_chunk(BeamFile *beam, IFF_Chunk *chunk) {
+    return parse_export_table(&beam->locals, beam, chunk);
+}
+#endif
+
 static int parse_lambda_chunk(BeamFile *beam, IFF_Chunk *chunk) {
     BeamFile_LambdaTable *lambdas;
     BeamFile_AtomTable *atoms;
@@ -743,6 +749,7 @@ beamfile_read(const byte *data, size_t size, BeamFile *beam) {
         MakeIffId('C', 'I', 'n', 'f'), /* 8 */
         MakeIffId('L', 'i', 'n', 'e'), /* 9 */
         MakeIffId('A', 't', 'U', '8'), /* 10 */
+        MakeIffId('L', 'o', 'c', 'T'), /* 11 */
     };
 
     static const int ATOM_CHUNK = 0;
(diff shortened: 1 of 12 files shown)


erts/emulator/beam/beam_file.c
```diff
@@ -740,6 +749,7 @@ beamfile_read(const byte *data, size_t size, BeamFile *beam) {
         MakeIffId('C', 'I', 'n', 'f'), /* 8 */
         MakeIffId('L', 'i', 'n', 'e'), /* 9 */
         MakeIffId('A', 't', 'U', '8'), /* 10 */
+	MakeIffId('L', 'o', 'c', 'T'), /* 11 */
```
bjorng, line 752: Nitpick: Nowadays we intend with space, not tabs.

bjorng (commented):
The nicer labels for local function does not seem to work for me. I get plain labels (such as `labels_42`). (macOS 11.4/Big Sur, Intel processor)

When adding a debug printout to `BeamModuleAssembler::BeamModuleAssembler()`, which is supposed to generate the nicer label, I see that the `named_labels` argument is always a NULL pointer.

bjorng:
By the way, I can test on Arm when you have added that code.

erts/emulator/beam/beam_file.c
frej: Fixed.

frej:
> The nicer labels for local function does not seem to work for me.

Never assume you can do an obvious clean up without retesting and without errors, I had dropped a line. :)

frej:
> By the way, I can test on Arm when you have added that code.

I added a completely untested arm-version.

garazdawi:
Would you mind renaming the suite `erts/emulator/test/perf_SUITE.erl` to `jit_SUITE.erl` and adding a basic testcase that just makes sure that this works?

frej:
@garazdawi :
> Would you mind renaming the suite `erts/emulator/test/perf_SUITE.erl` to `jit_SUITE.erl` and adding a basic testcase that just makes sure that this works?

I can do that. I had a look at `perf_SUITE.erl`, when running with `+JDump true` we will produce a lot of `.asm`-files in cwd. The existing `init/exit_per_suite()`-functions have logic to remove files named `jitted-*.so` which could be adapted to handle the `.asm`-files. But I wonder, do we still produce these `jitted-*.so`-files, isn't that left over from the LLVM-based JIT?

garazdawi:
> But I wonder, do we still produce these jitted-*.so-files, isn't that left over from the LLVM-based JIT?

Older perf versions generate this type of files, so yes they are created.

frej:
> Older perf versions generate this type of files, so yes they are created.

Kk, then I'll leave that code alone.

bjorng:
I have pushed a `fixup` commit with a small bug fix to your branch.

Unfortunately, it is not enough to get the prettier labels work on AArch64. There is a mechanism implemented by @jhogberg that will extend the range of branches, and it seems that mechanism interferes with pretty labels. (Basically, it seems that the labels given in the BEAM file is seldom or never actually used.)  I'll have a look tomorrow to see if I can figure it out.

frej:
@bjorng: I'll integrate your fixup when I've implemented @garazdawi's suggestion, so be sure to refresh your copy of the branch before you start work tomorrow.

frej:
@garazdawi, is there a way to only run the pretty-print test case on x86 until we know if it will be possible to implement on AArch64?

garazdawi:
Maybe you can use the return of `erlang:system_info(system_architecture).` in init_per_testcase?

frej:
@garazdawi, @bjorng: All all review comments are addressed, so feel free to give it a second look,

erts/emulator/test/jit_SUITE.erl
```diff
@@ -18,16 +18,17 @@
...
+-export([all/0, suite/0, init_per_suite/1, init_per_testcase/2,
+	 end_per_suite/1]).
+-export([symbols/1, annotate/1, named_labels/1]).
 
 suite() ->
     [{timetrap, {minutes, 4}}].
 
 all() -> 
-    [symbols, annotate].
+    [symbols, annotate, named_labels].
```
garazdawi, line 31: Put the perf related testcases in a perf group and then change the init_per_suite to be a init_per_group so that the correct testcases are skipped on the correct platforms.

```suggestion
    [{group, perf}, named_labels].
    
groups() ->
  [{perf,[symbols, annotate]}].
```

erts/emulator/test/jit_SUITE.erl
```diff
@@ -128,4 +150,37 @@ annotate(Config) ->
...
+    %% Check that pretty printing of named labels is working. We do
+    %% that by loading this module in an emulator running with +JDdump
+    %% true and then checking that the produced jit_SUITE.asm contains
+    %% a label for each exported function. We also check that the
+    %% label for the non-exported function get_tmp_asm_files/0, which
+    %% is used by this test, is present.
+    Exports = proplists:get_value(exports, ?MODULE:module_info()),
+    ModName = atom_to_list(?MODULE),
+    ModulePath = filename:dirname(code:which(?MODULE)),
+    Cmd = "erl +JDdump true -noshell -pa " ++ ModulePath ++ " -eval '"
```
garazdawi, line 164: I think this `Cmd` will fail on Windows because of the quoting. Use " instead.

frej: done

erts/emulator/test/jit_SUITE.erl
frej: Done.

bjorng:
I have pushed two `fixup` commits.

The first one names the label used for the trampolines as `@Mod:Name/Arity-SequenceNumber`. This would have been easier if asmjit had supported reading out the name of a named label and to find out the number of labels defined so far. As it is now, I have to keep track label names for entry labels and a sequence number myself.

The second commit ensures that the test case is run on AArch64.

If you see something that can be done better, feel free to change.

bjorng:
There is one remaining issue with the test suite. When running without the JIT enabled (for example, on a 32-bit system), the `named_labels/1` test case fails.

frej:
> I have pushed two `fixup` commits.

@bjorng: I'll take a look. Unless you tell me not to, I'll amend the commit for the arm-part to make you the author as you have done the work (and run clang-format).

> There is one remaining issue with the test suite. When running without the JIT enabled (for example, on a 32-bit system), the named_labels/1 test case fails.

Isn't that a pre-existing failure? As the perf-tests run `erl +JPperf true` that must have failed since before I started hacking this suite, or am I missing something? Anyway it won't be hard to skip the whole suite in `init_per_suite` if `erlang:system_info(emu_flavor) =/= jit`. I'll do that in a separate commit.

As for the failures in the CI 32-bit x86 build, my diagnosis is that it isn't the fault of these patches, as other PRs trigger it too.

garazdawi:
> As for the failures in the CI 32-bit x86 build, my diagnosis is that it isn't the fault of these patches, as other PRs trigger it too.

#4991 solves these problems.

bjorng:
> Isn't that a pre-existing failure? 

Yes, it is. We didn't notice it in our daily builds, because all test machines that don't support the JIT don't have `perf` installed, so the entire test suite happens to be skipped.

frej:
@bjorng check that you're happy with the commit message I wrote in your name for the Arm part.

I moved the detection of a JIT emulator to the refactoring commit. In the test case, there is completely untested code for extracting the assembly labels used on Arm which could use a test run.

Right now connection failures to Docker seems to randomly trigger failures in the CI test, I will try to retrigger them in the morning.

bjorng:
> @bjorng check that you're happy with the commit message I wrote in your name for the Arm part.

Yes, looks good.

> I moved the detection of a JIT emulator to the refactoring commit. In the test case, there is completely untested code for extracting the assembly labels used on Arm which could use a test run.

It succeeds. I have also that verified that the test case fails if I disable the creation of the nicer named labels.

Added to our daily builds.

PR #4986, opened 2021-06-19
Clarify the Pred function in lists must return booleans

### LINKED ISSUE #4985: lists:search/2 documentation and implementation disagree about non-true return values of Pred function
**Describe the bug**

The man page for `lists` (`lib/stdlib/doc/src/lists.xml`) describes the `lists:search/2` function as

  * `search(Pred, List) -> {value, Value} | false`
    If there is a `Value` in `List` such that `Pred(Value)` returns `true`, returns `{value, Value}` for the first such `Value`, otherwise returns `false`.

This makes no mention whatsoever about what the `Pred(Value)` call needs to return if it does not return `true`, so all values which are not the atom `true` should be valid according to that description.

However, the implementation of `lists:search/2` in `lib/stdlib/src/lists.erl` is more strict and insists that `Pred(Value)` MUST return either the atom `true` or the atom `false`, which is also reflected by its `-spec` specifying a `boolean()` return value.

```
-spec search(Pred, List) -> {value, Value} | false when
      Pred :: fun((T) -> boolean()),
      List :: [T],
      Value :: T.

search(Pred, [Hd|Tail]) ->
    case Pred(Hd) of
        true -> {value, Hd};
        false -> search(Pred, Tail)
    end;
search(Pred, []) when is_function(Pred, 1) ->
    false.
```

**To Reproduce**

Run this script:

```
#!/usr/bin/env escript
%% -*- erlang -*-

main([]) ->
    lists:search(fun(X) ->
			 case X rem 2 of
			     0 -> true;
			     _ -> odd
			 end
		 end,
		 [5,1,3,2,9,14]),
    ok.
```

```
[user@host ~]$ escript lists-search-pred-values
escript: exception error: no case clause matching odd
  in function  lists:search/2 (lists.erl, line 1412)
  in call from erl_eval:do_apply/6 (erl_eval.erl, line 680)
  in call from escript:eval_exprs/5 (escript.erl, line 872)
  in call from erl_eval:local_func/6 (erl_eval.erl, line 567)
  in call from escript:interpret/4 (escript.erl, line 788)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_em/1 
  in call from init:do_boot/3 
[user@host ~]$ 
```

Yes, this example `fun` may look a bit contrived, but with a number of OTP functions such as `lists:search/2` or `lists:keysearch/3` returning either `{value, Value}` or `false`, there are certainly precedents for a function returning either a boolean value or something that is not a boolean value at all not being very unusual.

**Expected behavior**

If the documentation is correct, the escript should run without throwing an exception.

**Affected versions**

All OTP versions since the introduction of `lists:search/2` which, according to `lists.xml`, was *OTP 21.0*. In my case, *OTP 23*.

**Additional context**

I see two ways to fix this:

  * either fix the documentation to explicitly demand that `Pred(Value)` needs to return either `true` or `false`
  * or fix the implementation (and the `-spec`) to allow for any non-`true` value to be interpreted as currently `false` is

There are arguments for both ways.

Fixing the implementation to allow all non-`true` values appears to be more Erlang-y to me (by making good use of pattern matching), which is good, does not change much in the code, and maintains compatibility in that `Pred` functions returning `boolean()` will continue to work. However, if existing code relies on a non-`boolean()` `Pred` value throwing an exception, this change would break that. BTW, I am not sure whether the `-spec` for `Pred` is a good way of expressing this:

```
-spec search(Pred, List) -> {value, Value} | false when
      Pred :: fun((T) -> true | term()),
      List :: [T],
      Value :: T.

search(Pred, [Hd|Tail]) ->
    case Pred(Hd) of
        true -> {value, Hd};
        _ -> search(Pred, Tail)
    end;
search(Pred, []) when is_function(Pred, 1) ->
    false.
```

However, that would mean changing a public API, which is bad, but not very much, which could be OK.

josevalim:
Closes #4985.

### DIFF
--- a/lib/stdlib/doc/src/lists.xml
+++ b/lib/stdlib/doc/src/lists.xml
@@ -69,7 +69,8 @@
       <desc>
         <p>Returns <c>true</c> if <c><anno>Pred</anno>(<anno>Elem</anno>)</c>
           returns <c>true</c> for all elements <c><anno>Elem</anno></c> in
-          <c><anno>List</anno></c>, otherwise <c>false</c>.</p>
+          <c><anno>List</anno></c>, otherwise <c>false</c>. The <c>Pred</c>
+          function must return a boolean.</p>
       </desc>
     </func>
 
@@ -80,7 +81,8 @@
       <desc>
         <p>Returns <c>true</c> if <c><anno>Pred</anno>(<anno>Elem</anno>)</c>
           returns <c>true</c> for at least one element <c><anno>Elem</anno></c>
-          in <c><anno>List</anno></c>.</p>
+          in <c><anno>List</anno></c>. The <c>Pred</c> function must return a
+          boolean.</p>
       </desc>
     </func>
 
@@ -154,7 +156,8 @@
         <p>Drops elements <c><anno>Elem</anno></c> from
           <c><anno>List1</anno></c> while
           <c><anno>Pred</anno>(<anno>Elem</anno>)</c> returns <c>true</c> and
-          returns the remaining list.</p>
+          returns the remaining list. The <c>Pred</c> function must return a
+          boolean.</p>
       </desc>
     </func>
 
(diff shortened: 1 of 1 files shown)


bjorng:
Could you base this PR on maint so that we can include it in OTP 24.1?

josevalim:
@bjorng done!

PR #4987, opened 2021-06-19
Enable horizontal scrolling in docs code block samples

brunojppb:
While reading the [sys and proc_lib docs](http://erlang.org/doc/design_principles/spec_proc.html#id80464), I realised that the code blocks wrapped in a `<pre>` tag were overflowing its container. This prevents scrolling the container and seeing the text on smaller screens. 

This PR should fix that by enabling horizontal scrolling within code block samples.
Here is some screenshots and video of the before and after this PR.

## Before
Screenshot 2021-06-19 at 20 45 02

## After
Screenshot 2021-06-19 at 20 49 18

## Video of the current solution being applied

https://user-images.githubusercontent.com/2049560/122652658-2bed0d80-d140-11eb-8237-ff062b4fa6a5.mp4

### DIFF
--- a/lib/erl_docgen/priv/css/otp_doc.css
+++ b/lib/erl_docgen/priv/css/otp_doc.css
@@ -67,8 +67,8 @@ a:visited      { color: #1b6ec2; text-decoration: none }
 
 .innertube 
 {
-    margin-left: 15px; /* Magins for inner DIV inside each DIV (to provide padding) */
-    margin-right: 11em;
+    margin: 0 15px; /* Magins for inner DIV inside each DIV (to provide padding) */
+    max-width: 67ex;
 }
 
 .footer 
@@ -180,7 +180,7 @@ a:visited      { color: #1b6ec2; text-decoration: none }
   padding: 0.5em 1em;
   margin: 1em 0;
   font-size: 0.7em;
-  overflow-x: scroll;
+  overflow-x: auto;
 }
 .extrafrontpageinfo {
   color: #C00;
@@ -204,7 +204,8 @@ pre {
 }
 .exports-tube
 {
-    margin-right: 11em;
+    margin-right: 15px;
+    max-width: 67ex;
 }
 
 footer       { }
--- a/lib/erl_docgen/priv/css/otp_doc.css
+++ b/lib/erl_docgen/priv/css/otp_doc.css
@@ -180,6 +180,7 @@ a:visited      { color: #1b6ec2; text-decoration: none }
   padding: 0.5em 1em;
   margin: 1em 0;
   font-size: 0.7em;
+  overflow-x: scroll;
 }
 .extrafrontpageinfo {
   color: #C00;


zuiderkwast:
That's nice!

This is mostly a problem in a narrow window (not full-screen, or mobile device), as the content shrinks but the *large fixed margin of 11em to the right* is kept. Perhaps you want to fix that too? Here's a suggestion using max-width instead:

Replace the `margin-right: 11em` in `.innertube {...}` and `.exports-tube {...}` blocks with `margin-right: 15px; maxwidth: 67ex`.

The result looks the same on a full-screen window but uses the space better on a narrower window.



| Before | After |
|------|------|
| before | after |

brunojppb:
> That's nice!
> 
> This is mostly a problem in a narrow window (not full-screen, or mobile device), as the content shrinks but the _large fixed margin of 11em to the right_ is kept. Perhaps you want to fix that too? Here's a suggestion using max-width instead:
> 
> Replace the `margin-right: 11em` in `.innertube {...}` and `.exports-tube {...}` blocks with `margin-right: 15px; maxwidth: 67ex`.

Very good point @zuiderkwast 

I can certainly provide a fix for that as well. I realised that we don't need the `max-width` there since the `#content` selector already contains a `max-width: 52em;` which seems great already 👍

brunojppb:
It indeed renders nicely on narrower screens. Here is a little video including the changes from my latest commit while resizing the window

https://user-images.githubusercontent.com/2049560/122672559-6a300e80-d1cc-11eb-94c6-5f03bb7c4403.mp4

lib/erl_docgen/priv/css/otp_doc.css
```diff
@@ -67,8 +67,7 @@ a:visited      { color: #1b6ec2; text-decoration: none }
 
 .innertube 
 {
-    margin-left: 15px; /* Magins for inner DIV inside each DIV (to provide padding) */
-    margin-right: 11em;
+    margin: 0 15px; /* Magins for inner DIV inside each DIV (to provide padding) */
```
brunojppb, line 70: short-hand for: 
- top and bottom: 0px
- left and right: 15px

lib/erl_docgen/priv/css/otp_doc.css
```diff
@@ -180,6 +179,7 @@ a:visited      { color: #1b6ec2; text-decoration: none }
   padding: 0.5em 1em;
   margin: 1em 0;
   font-size: 0.7em;
+  overflow-x: auto;
```
brunojppb, line 182: `auto` will let browsers provide scrollbars if content overflows.

zuiderkwast:
> I realised that we don't need the `max-width` there since the `#content` selector already contains a `max-width: 52em;` which seems great already

It's not exactly the same max-width though. Some contents are wider than others, e.g. `.func-table` is designed to be wider than the `.exports-tube`. Check pages with the "since OTP-version" marker, such as here: https://erlang.org/doc/man/gen_sctp.html#listen-2

brunojppb:
> It's not exactly the same max-width though. Some contents are wider than others, e.g. `.func-table` is designed to be wider than the `.exports-tube`. Check pages with the "since OTP-version" marker, such as here: https://erlang.org/doc/man/gen_sctp.html#listen-2

Hoo now I get it. Thanks for the context. pushing a commit with that change now.

zuiderkwast (approved):
LGTM

garazdawi:
I think that all of the issues in this PR have been addressed in #5197. For a (somewhat up-to-date) preview see https://beta.erlang.org/doc.

PR #4990, opened 2021-06-21
Add a "Coming Potential Incompatibilities" page to the documentation

bjorng:
Some upcoming changes can be awkward to fit into the "Deprecations" and "Scheduled for Removal" pages. Add a new "Coming Potential Incompatibilities" pages to address that issue.

To that page, add a list of the distribution capabilities that will become mandatory in OTP 25 (see #4972).

Also copy with edits some upcoming changes to the distribution now described in "Scheduled for Removal".

### DIFF
--- a/erts/doc/src/erl_dist_protocol.xml
+++ b/erts/doc/src/erl_dist_protocol.xml
@@ -988,7 +988,7 @@ DiB == gen_digest(ChA, ICA)?
         <tag><c>-define(DFLAG_EXTENDED_REFERENCES,16#4).</c></tag>
         <item>
           <p>The node implements extended (3 &times; 32 bits) references. This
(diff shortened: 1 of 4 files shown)


zuiderkwast (commented):
Good info!

How about "upcoming" vs "coming"?

"Upcoming" warns the reader that it's in the relative near future. Also, it sounds better IMO, but what do I know(?)...

bjorng:
> How about "upcoming" vs "coming"?
> 
> "Upcoming" warns the reader that it's in the relative near future. Also, it sounds better IMO, but what do I know(?)...

"Upcoming" sounds better to me, too. I'll change.

rickard-green (approved):

PR #4996, opened 2021-06-23
fix ssl:close with nif sockets

### LINKED ISSUE #4993: ssl:close fails with nif sockets
**Describe the bug**

`ssl:close` fails with `function_clause` error when using nif sockets api

**To reproduce**

1. Start erlang with `erl -kernel inet_backend socket`.
2. Type

```
application:ensure_all_started(ssl).
{ok,S}=ssl:connect("erlang.com", 443, [], 5000).
ssl:close(S).
```

**Full log**

https://gist.github.com/petrkozorezov/d2376e15fac0687bf0ee5e22be3ae809

petrkozorezov:
Closes #4993

### DIFF
--- a/lib/kernel/src/inet.erl
+++ b/lib/kernel/src/inet.erl
@@ -202,6 +202,8 @@ get_rc() ->
 -spec close(Socket) -> 'ok' when
       Socket :: socket().
 
+close(?module_socket(GenSocketMod, _) = Socket) when is_atom(GenSocketMod) ->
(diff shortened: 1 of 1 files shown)


bmk:
Completely missed that one. Close is almost always called from gen_tcp or ssl or similar.
*But* for completeness sake, it *should* also handle the new compat-sockets.

bmk (approved):

PR #4997, opened 2021-06-23
epp: Fix inconsistent handling of undefined pre-defined macros

### LINKED ISSUE #4995: epp doesn't handle -ifdef(MODULE) in .hrl file
**Describe the bug**
An `.hrl` file contains a fragment which needs to expand to use `?MODULE` when included from an `.erl` file, but to something else when pre-processed in isolation. When pre-processed in isolation, `-ifdef(MODULE)` erroneously selects the `?MODULE`-using fragment, which however fails to pre-process because `MODULE` is in fact not defined.

The error seems to be that `epp:predef_macros/1` binds `MODULE` to `undefined` in the map, `scan_ifdef/4` only checks that the name is bound (which it is), and `expand_macro1/4` throws an error if the macro name is unbound or bound to `undefined`. So there is a discrepancy between the logic in `scan_ifdef/4` and `expand_macro1/4` with regards to predefined macros like `MODULE`.

**To Reproduce**
```
> cat bug.hrl
-ifdef(MODULE).
-record(r, {x = ?MODULE}).
-endif.
> erl
Erlang/OTP 24 [erts-12.0.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V12.0.2  (abort with ^G)
1> epp:parse_file("bug.hrl", []).
{ok,[{attribute,1,file,{"bug.hrl",1}},
     {error,{2,epp,{undefined,'MODULE',none}}},
     {eof,4}]}
2>
```
So the entire `-record` form is gone.

**Expected behavior**
I expected the file to pre-process, either substituting `undefined` for `?MODULE`, or skipping the `-ifdef(MODULE)` path.

**Affected versions**
At least OTP-20/21/22/23/24. Have not checked older versions.

**Additional context**
This was discovered in the context of `erlavro`. It seems Elixir is pre-processing `.hrl` files containing record declarations in order to build helper functions. In our case the `.hrl` file used errors mentioning `?MODULE` to mark record fields which must not have default values. This works when compiling in Erlang, but failed when Elixir pre-processed the `.hrl` file in isolation. Trying to work around the issue with `-ifdef(MODULE)` did not work.

bjorng:
The `?MODULE` pre-defined macro would be considered defined even before
a `-module()` definition had been seen. For example:

    -ifdef(MODULE).
    -error("?MODULE is incorrectly considered defined").
    -endif.
    -module(t).

When attempting to compile this module, the following error would
be produced:

    t.erl:2:2: -error("?MODULE is incorrectly considered defined").
    %    2| -error("?MODULE is incorrectly considered defined").
    %     |  ^

The behavior is inconsistent with how an attempt to use ?MODULE before
the `-module()` definition has been seen is handled (there will be
an "undefined macro" error).

This bug matters in practice because Elixir is pre-processing `.hrl`
files containing record declarations in order to build helper
functions. Pre-processing the following header file will fail when
pre-processed in isolation by Elixir:

    -ifdef(MODULE).
    -record(r, {x = ?MODULE}).
    -endif.

Fixes #4995.

### DIFF
--- a/lib/stdlib/src/epp.erl
+++ b/lib/stdlib/src/epp.erl
@@ -1088,17 +1088,17 @@ scan_include_lib1(Toks, Inc, From, St) ->
 %%  Report a badly formed if[n]def test and then treat as undefined macro.
 
 scan_ifdef([{'(',_Alp},{atom,_Am,M},{')',_Arp},{dot,_Ad}], _IfD, From, St) ->
-    case St#epp.macs of
-	#{M:=_Def} ->
+    case is_macro_defined(M, St) of
+        true ->
 	    scan_toks(From, St#epp{istk=[ifdef|St#epp.istk]});
-	_ ->
+        false ->
 	    skip_toks(From, St, [ifdef])
     end;
 scan_ifdef([{'(',_Alp},{var,_Am,M},{')',_Arp},{dot,_Ad}], _IfD, From, St) ->
-    case St#epp.macs of
-	#{M:=_Def} ->
+    case is_macro_defined(M, St) of
+        true ->
 	    scan_toks(From, St#epp{istk=[ifdef|St#epp.istk]});
-	_ ->
+        false ->
 	    skip_toks(From, St, [ifdef])
     end;
 scan_ifdef(Toks, IfDef, From, St) ->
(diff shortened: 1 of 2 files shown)


mikpe (approved):

lib/stdlib/src/epp.erl
```diff
@@ -1107,24 +1107,31 @@ scan_ifdef(Toks, IfDef, From, St) ->
...
 	    scan_toks(From, St#epp{istk=[ifndef|St#epp.istk]})
     end;
 scan_ifndef(Toks, IfnDef, From, St) ->
     T = find_mismatch(['(',var_or_atom,')',dot], Toks, IfnDef),
     epp_reply(From, {error,{loc(T),epp,{bad,ifndef}}}),
     wait_req_skip(St, [ifndef]).
 
+is_macro_defined(Name, #epp{macs=Macs}) ->
+    case Macs of
+        #{Name := undefined} -> false;
```
zuiderkwast, line 1130: I hope this doesn't cause macros defined to `undefined` to be considered undefined. Does it?

```Erlang
-define(dummy, undefined).
-ifndef(dummy).
-error("Defined to undefined").
-endif.

bjorng: No, the value for your `dummy` macro would be represented as `[{atom,Anno,undefined}]`, which is different from `undefined` used to mark a pre-defined macro currently without a value.

mikpe: It's my understanding that defined macros are bound to `{none,[Expr]}` in that map, so there should be no issue.

zuiderkwast: Good, so this can be resolved.

PR #5002, opened 2021-06-24
erts/emulator: use single type for ErlDrvEvent

trofi:
In https://bugs.gentoo.org/797886 I noticed that LTO build
complains about inconsistent type within a single final binary.

The simplest way to find it is to use `-Werror=lto-type-mismatch`:

```
 LD     otp/bin/x86_64-pc-linux-gnu/beam.jit
  beam/erl_driver.h:331:12: error:
    type of 'driver_select' does not match original declaration [-Werror=lto-type-mismatch]
  331 | EXTERN int driver_select(ErlDrvPort port, ErlDrvEvent event, int mode, int on);
      |            ^
sys/common/erl_check_io.c:796:1: note: type mismatch in parameter 2
  796 | driver_select(ErlDrvPort ix, ErlDrvEvent e, int mode, int on)
      | ^
sys/common/erl_check_io.c:796:1: note: type 'ErlDrvEvent' should match type 'struct _erl_drv_event *'
sys/common/erl_check_io.c:796:1: note: 'driver_select' was previously declared here
sys/common/erl_check_io.c:796:1: note: code may be misoptimized unless '-fno-strict-aliasing' is used
lto1: some warnings being treated as errors
...
```

Multiple definitions are located in:

```
$ git grep ErlDrvEvent | fgrep typedef
erts/emulator/beam/erl_driver.h:typedef struct _erl_drv_event* ErlDrvEvent; /* An event to be selected on. */
erts/emulator/beam/erl_sys_driver.h:typedef SWord ErlDrvEvent; /* An event to be selected on. */
```

The change uses the first `typedef struct _erl_drv_event* ErlDrvEvent;` everywhere.

Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>

### DIFF
--- a/erts/emulator/sys/common/erl_check_io.c
+++ b/erts/emulator/sys/common/erl_check_io.c
@@ -43,6 +43,9 @@
 #define ERTS_WANT_TIMER_WHEEL_API
 #include "erl_time.h"
 
+#define fd2event(fd) (ErlDrvEvent)(SWord)(fd)
+#define event2fd(e) (int)(SWord)(e)
+
 #if 0
 #define DEBUG_PRINT(FMT, ...) do { erts_printf(FMT "\r\n", ##__VA_ARGS__); fflush(stdout); } while(0)
 #define DEBUG_PRINT_FD(FMT, STATE, ...)                                 \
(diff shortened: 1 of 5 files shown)


erts/emulator/sys/common/erl_check_io.c
```diff
@@ -798,7 +801,7 @@ driver_select(ErlDrvPort ix, ErlDrvEvent e, int mode, int on)
     void (*stop_select_fn)(ErlDrvEvent, void*) = NULL;
     Port *prt = erts_drvport2port(ix);
     Eterm id = erts_drvport2id(ix);
-    ErtsSysFdType fd = (ErtsSysFdType) e;
+    ErtsSysFdType fd = (ErtsSysFdType) event2fd(e);
```
sverker, line 804: This all looks good for Unix where `ErtsSysFdType` is `int`, but on **Windows** it's defined as `HANDLE` which in turn is a pointer I think.

trofi: Ah, good point!

trofi:
I unlikely be able to amend it for windows in a short term. Ideally type system should help us here. Initial state is less problematic in this regard.

As the change is fairly mechanical (and error prone) I'll abandon this patch set.

PR #5008, opened 2021-06-27
Fail early when calling_self

### LINKED ISSUE #5005: Application freezes during start failure when logging to the group leader
# Issue
An OTP application whose `init/2` callback fails is supposed to cause the node to exit (the usual "Kernel pid terminated (application_controller)…"). While this error is being handled, OTP uses `logger` to emit error reports. If a log handler uses the group leader as IO device, the node will not exit; in console mode, the prompt will not be printed. Worse, no error message is being printed. It seems something is stuck.

# Context
Following a discussion on erlang-questions regarding the fact that logger messages are not captured by EUnit (so that they can be printed only when a test case fails), I start using a log handler which writes to the group leader, as suggested in [this email](https://erlang.org/pipermail/erlang-questions/2021-May/101010.html) by @garazdawi (you may find it interesting). I ended up discovering this problem because my application `start/2` callback failed and I could not get any error message, with the node staying up.

# Example
The following application reproduces the problem every time: https://github.com/galdor/erl-gl-test.
You can clone it, and run:
```
rebar3 release
./_build/default/rel/gl_test/bin/gl_test console
```

In `src/gl_test_app.erl`, using `standard_error` for `Device` fixes the problem.

# Environment
- Erlang/OTP 24.
- Rebar 3.16.1.

josevalim:
This provides early feedback in gen behaviours if a process
attempts to call itself, which is especially important on
infinity timeouts.

The functionality was also implemented for the io module,
which also errors with reason "calling_self" and is handled
by the error formatter accordingly:

    1> io:put_chars(self(), "123").
    ** exception error: calling_self
         in function  io:put_chars/2
            called as io:put_chars(<0.83.0>,"123")
            *** argument 1: process sent a blocking request to itself

Closes #5005.

### DIFF
--- a/lib/stdlib/src/erl_stdlib_errors.erl
+++ b/lib/stdlib/src/erl_stdlib_errors.erl
@@ -417,6 +417,11 @@ format_io_error(_, _, {io, arguments}, true) ->
     [device_arguments];
 format_io_error(_, _, {io, arguments}, false) ->
     [{general,device_arguments}];
+%% calling_self, Io =:= self()
+format_io_error(_, _, {io, calling_self}, true) ->
+    [calling_self];
+format_io_error(_, _, {io, calling_self}, false) ->
+    [{general,calling_self}];
 %% terminated, monitor(Io) failed
 format_io_error(_, _, {io, terminated}, true) ->
     [device_terminated];
@@ -988,6 +993,8 @@ expand_error(bad_update_op) ->
     <<"not a valid update operation">>;
 expand_error(bitstring) ->
     <<"is a bitstring (expected a binary)">>;
+expand_error(calling_self) ->
+    <<"the device is not allowed to be the current process">>;
 expand_error(counter_not_integer) ->
     <<"the value in the given position, in the object, is not an integer">>;
 expand_error(dead_process) ->
--- a/lib/stdlib/test/io_SUITE.erl
+++ b/lib/stdlib/test/io_SUITE.erl
@@ -28,7 +28,7 @@
 	 io_fread_newlines/1, otp_8989/1, io_lib_fread_literal/1,
 	 printable_range/1, bad_printable_range/1,
 	 io_lib_print_binary_depth_one/1, otp_10302/1, otp_10755/1,
-         otp_10836/1, io_lib_width_too_small/1,
+         otp_10836/1, io_lib_width_too_small/1, calling_self/1,
          io_with_huge_message_queue/1, format_string/1, format_neg_zero/1,
 	 maps/1, coverage/1, otp_14178_unicode_atoms/1, otp_14175/1,
          otp_14285/1, limit_term/1, otp_14983/1, otp_15103/1, otp_15076/1,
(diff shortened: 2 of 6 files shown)


lib/stdlib/src/erl_stdlib_errors.erl
```diff
@@ -988,6 +993,8 @@ expand_error(bad_update_op) ->
     <<"not a valid update operation">>;
 expand_error(bitstring) ->
     <<"is a bitstring (expected a binary)">>;
+expand_error(calling_self) ->
+    <<"process sent a blocking request to itself">>;
```
garazdawi, line 997:
```suggestion
    <<"the device is not allowed to be the calling process">>;
```
Would this be more helpful to the user?

lib/stdlib/test/io_SUITE.erl
```diff
@@ -210,6 +210,10 @@ float_w(Config) when is_list(Config) ->
 
     ok.
 
+calling_self(Config) when is_list(Config) ->
```
garazdawi, line 213: There is a testcase in this SUITE called `error_info` that should also be updated.

garazdawi (commented):
For `io` I think that this makes a lot of sense.

It also makes sense for gen, but there we need to think a lot more about backward compatibility. I hope that no one relies on doing a `gen_server:call(self(),msg,1000)` that triggers a timeout and then handle the call themselves... but I have seen a lot of odd code doing things that I did not expect...

lib/stdlib/src/erl_stdlib_errors.erl
josevalim: Nice! Maybe “the device cannot be the current process”?

josevalim:
@garazdawi we could for now raise calling_self for infinity timeout exclusively, as the timeout one will eventually fail. Although I would prefer to raise early, I understand the compatibility concern. Let me know how you prefer to move forward and I will amend. :)

garazdawi:
Leave the gen parts as they are for now and I'll get back to you later about it.

garazdawi:
> Leave the gen parts as they are for now and I'll get back to you later about it.

I've talked about it with the team and we agree with you that given an error early is better, so we'll include gen as proposed.

`erl_error:explain_reason/1` should also be expanded to contains a better description of the error.

josevalim:
@garazdawi regarding erl_error:explain_reason/7, I have added this locally:

```erl
explain_reason(calling_self, exit, [], _PF, _S, _Enc, _CL) ->
    <<"the current process attempted to call itself">>;
```

However, this won't be shown by `gen_server:call(self(), oops).` because the exit reason is actually wrapped in a tuple:

```
1> gen_server:call(self(), oops).
** exception exit: {calling_self,{gen_server,call,[<0.83.0>,oops]}}
     in function  gen_server:call/2 (gen_server.erl, line 239)
```

I could match on the tuple, but then it means the actual call is no longer in the error message. Am I missing something? I think in Elixir we unpack said tuples in case the second element looks like MFArgs before handling the "explain_reason" but that's probably a separate discussion.

josevalim:
Except by the pending comment above, which may require further changes, this should be good to go!

garazdawi:
> I could match on the tuple, but then it means the actual call is no longer in the error message. Am I missing something?

I wasn't aware that this was done, I suppose it tries to make sure the arguments to the gen:call is printed in the error.
Leave that as is for now and we can come back to it if we want.

garazdawi:
> Except by the pending comment above, which may require further changes, this should be good to go!

lgtm, putting in test.

garazdawi:
thanks!

PR #5010, opened 2021-06-28
Fix code generation in guard for not (X =:= true)

### LINKED ISSUE #5007: Compiler emits invalid code when using a guard with not (X =:= true)
**To Reproduce**

Take this code:

```erl
-module(foo).
-export([bar/1]).

bar(X) when not (X =:= true) -> ok;
bar(_) -> error.
```

Then the following session:

```
$ erl
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V12.0  (abort with ^G)
1> foo:bar(true).
error
2> foo:bar(foo).
error
```

The last call should have returned `ok` instead.

**Expected behavior**

```
1> foo:bar(true).
error
2> foo:bar(foo).
ok
```

**Affected versions**

I have verified it on Erlang/OTP 24, compiler-8.0. I will be glad to check other versions if necessary.

bjorng:
In a guard, `not (X =:= true)` would incorrectly evaluate to `false`
for non-boolean values of `X`.

The is because of an unsafe transformation in the `v3_core` that
transforms `not (X =:= true)` to `X =:= false`. Unfortunately, that
transformation is only safe if `X` is known to be a boolean. (The
bug was introduced in OTP R13.)

Fixes #5007.

### DIFF
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -380,7 +380,7 @@ gexpr({call,Line,{remote,_,{atom,_,erlang},{atom,_,'not'}},[A]}, Bools, St) ->
 gexpr(E0, Bools, St0) ->
     gexpr_test(E0, Bools, St0).
 
-%% gexpr_not(L, R, Bools, State) -> {Cexpr,[PreExp],Bools,State}.
+%% gexpr_bool(L, R, Bools, State) -> {Cexpr,[PreExp],Bools,State}.
 %%  Generate a guard for boolean operators
 
 gexpr_bool(Op, L, R, Bools0, St0, Line) ->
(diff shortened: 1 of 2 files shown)


garazdawi:
Will the beam_ssa type pass restore the optimization for the cases when we actually know that it is a boolean?

bjorng:
> Will the beam_ssa type pass restore the optimization for the cases when we actually know that it is a boolean?

Yes.

mikpe:
In the issue you stated that you did not want to backport the codegen fix to older releases, so I modified our compiler (OTP-22 at the moment) to warn if the optimization you're removing triggers. That caused a _lot_ of warnings, all of them for cases where the compiler itself has converted `not X` guards into `not (X =:= true)`, which then would become `X =:= false`. Adjusting the warning to not be issued for compiler-generated `X =:= true` silenced them all.

This tells me the optimization triggers a lot for compiler-generated intermediate code. Will other parts of the compiler recover these optimizations?

For the record, this is what I have right now:
```
diff --git a/lib/compiler/src/v3_core.erl b/lib/compiler/src/v3_core.erl
index 4ad9167714..d67949075b 100644
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -396,7 +396,12 @@ gexpr_not(A, Bools0, St0, Line) ->
             %%
             Ae = EqCall#icall{args=[E,#c_literal{val=false}]},
             {Al,Alps,St2} = force_safe(Ae, St1),
-            {Al,Aps ++ Alps,Bools,St2};
+            St3 =
+                case lists:member(compiler_generated, (EqCall#icall.anno)#a.anno) of
+                  true -> St2;
+                  false -> add_warning(Line, unsafe_not_true, St2)
+                end,
+            {Al,Aps ++ Alps,Bools,St3};
         Ae ->
             {Al,Alps,St2} = force_safe(Ae, St1),
             Anno = lineno_anno(Line, St2),
@@ -2714,6 +2719,8 @@ is_simple_list(Es) -> lists:all(fun is_simple/1, Es).
 
 -spec format_error(err_desc()) -> nonempty_string().
 
+format_error(unsafe_not_true) -> % GH-5007
+    "potentially unsafe optimization triggered; avoid not(X =:= true)";
 format_error(nomatch) ->
     "pattern cannot possibly match";
 format_error(bad_binary) ->
```

bjorng:
> This tells me the optimization triggers a lot for compiler-generated intermediate code. Will other parts of the compiler recover these optimizations?

Most of them, but not all. When I ran [scripts/diffable](https://github.com/erlang/otp/blob/e11a30663afb645b99595b1ada41127ddc97a5fd/scripts/diffable) and compared, only three modules out of a sample 1000+ modules from OTP and Elixir had worse code.

But your experiments with testing `compiler_generated` got me thinking. It will probably be safe to keep the optimization, but only apply it to compiler-generated code. I will think some more about that.

It would be interesting to see how many modules in your code base would be changed by this PR. Perhaps you could modify `scripts/diffable` to run on a representative sample of your code base?

bjorng:
Superseded by #5012.

PR #5012, opened 2021-06-29
Fix code generation in guard for not (X =:= true)

### LINKED ISSUE #5007: Compiler emits invalid code when using a guard with not (X =:= true)

bjorng:
In a guard, `not (X =:= true)` would incorrectly evaluate to `false`
for non-boolean values of `X`.

The reason for the bug is an unsafe transformation in the `v3_core`
pass that transforms `not (X =:= true)` to `X =:= false`. That
transformation is safe only if `not (X =:= true)` was transformed from
`not X` by the compiler itself. (This bug was introduced in OTP R13.)

Eliminate the problem by only doing the transformation when `X =:= true`
is known to have been expanded from `not X` by the compiler.

While at it, also improve the comments.

Fixes #5007.

### DIFF
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -380,7 +380,7 @@ gexpr({call,Line,{remote,_,{atom,_,erlang},{atom,_,'not'}},[A]}, Bools, St) ->
 gexpr(E0, Bools, St0) ->
     gexpr_test(E0, Bools, St0).
 
-%% gexpr_not(L, R, Bools, State) -> {Cexpr,[PreExp],Bools,State}.
+%% gexpr_bool(L, R, Bools, State) -> {Cexpr,[PreExp],Bools,State}.
 %%  Generate a guard for boolean operators
 
 gexpr_bool(Op, L, R, Bools0, St0, Line) ->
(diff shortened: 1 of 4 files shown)


lib/compiler/src/v3_core.erl
```diff
@@ -400,13 +400,32 @@ gexpr_bool(Op, L, R, Bools0, St0, Line) ->
...
+            %%
+            %%    not Expr  ==>  Expr =:= false
+            %%
+            %% which will produce the same result, but may eliminate
+            %% redundant is_boolean/1 tests (see unforce/3).
+            %%
+            %% Note that this tranformation would not be safe if the
+            %% original expression had been:
+            %%
+            %%    Expr =:= true
```
mikpe, line 428: nit: `not(Expr =:= true)`

mikpe (approved):

PR #5015, opened 2021-06-30
Fix for DistEntry write-after-free bug

### LINKED ISSUE #4964: core dump with automatic async dist connection
VM dumps the core (or aborts in a debug build) when dist_auto_connect it enabled (which is the default).

This appears to be a race condition, when erl_node_tables collects garbage (every 120 sec by default, but can be configured down to 1 second), and `{auto_connect, Node, DHandle}` messages are still in the net_kernel message queue, pointing to DistEntry that has been already freed.

**To Reproduce**

Use emulator with option "+zdntgc 1" to simplify the repro

```
-module(dist_crash).
-export([main/1]).

-mode(compile).

main([]) ->
    inet_db:set_lookup([file]),
    Nodes = [list_to_atom("none@host" ++ integer_to_list(Seq)) || Seq <- lists:seq(1, 200)],
    Self = self(),
    [spawn(fun () -> sender(Self, Nodes) end) || _ <- lists:seq(1, 200)],
    count(0).

-define(PRINT_EVERY, 100).

count(Total) ->
    receive
        count when Total rem ?PRINT_EVERY =:= 0 ->
            io:format("Total attempts: ~bk~n", [Total div ?PRINT_EVERY]),
            count(Total + 1);
        count ->
            count(Total + 1)
    end.

sender(Main, Nodes) ->
    [{none, Node} ! msg || Node <- Nodes],
    Main ! count,
    sender(Main, Nodes).
```

sverker:
Fixes #4964

### DIFF
--- a/erts/emulator/beam/erl_bif_unique.h
+++ b/erts/emulator/beam/erl_bif_unique.h
@@ -91,6 +91,7 @@ ERTS_GLB_INLINE void erts_sched_make_magic_ref_in_array(ErtsSchedulerData *esdp,
 							Uint32 ref[ERTS_REF_NUMBERS]);
 ERTS_GLB_INLINE Eterm erts_sched_make_ref_in_buffer(ErtsSchedulerData *esdp,
 						    Eterm buffer[ERTS_REF_THING_SIZE]);
+ERTS_GLB_INLINE Eterm erts_mk_magic_ref_get_refc(Eterm * *hpp, ErlOffHeap * ohp, Binary*, erts_aint_t*);
 ERTS_GLB_INLINE Eterm erts_mk_magic_ref(Eterm **hpp, ErlOffHeap *ohp, Binary *mbp);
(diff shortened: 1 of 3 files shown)


dist_crash.erl
```diff
@@ -0,0 +1,27 @@
+-module(dist_crash).
+-export([main/1]).
+
+-mode(compile).
+
+main([]) ->
+    inet_db:set_lookup([file]),
```
sverker, line 7: @max-au Do you remember if this `inet_db:set_lookup([file])` does anything useful for the test?

rickard-green (approved):

dist_crash.erl
max-au: Iit makes the connection attempt to fail fast. Otherwise DNS resolution kicks in, taking a long time, and since connections are all serialised through net_kernel, the race isn't triggered.

Side note, I think it's worth making hostname resolution async, to avoid head-of-the-line issue when a single delayed DNS request may trigger long wait time.

PR #5017, opened 2021-07-01
Do not attempt to start children of running apps

josevalim:
_Note: PR #2675 introduced a performance regression on application:ensure_all_started/2. This pull request address it by adding the same check found in `application_controller:start_application/2` This applies to the `maint` branch, as you may want to include this on 24.1._

When optional applications were added in 79146711, ensure_all_started/2 had to take explicit control over the initialization cycle so available optional applications were started properly. However, in doing so, we lost one important optimization, which was to not attempt to start children of a given application if said application was already running.

This commit address this by exporting is_running/1 and skipping the work in case the given app is already running. This change makes it so the time taken to start a deep dependency tree goes from 58s to 4s.

### DIFF
--- a/lib/kernel/src/application_controller.erl
+++ b/lib/kernel/src/application_controller.erl
@@ -23,7 +23,7 @@
 -export([start/1, 
 	 load_application/1, unload_application/1, 
 	 start_application/2, start_boot_application/2, stop_application/1,
-	 control_application/1,
+	 control_application/1, is_running/1,
 	 change_application_data/2, prep_config_change/0, config_change/1,
 	 which_applications/0, which_applications/1,
 	 loaded_applications/0, info/0, set_env/2,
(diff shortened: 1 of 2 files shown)


lib/kernel/src/application_controller.erl
```diff
@@ -237,6 +237,17 @@ unload_application(AppName) ->
 start_application(AppName, RestartType) ->
     gen_server:call(?AC, {start_application, AppName, RestartType}, infinity).
 
+%%-----------------------------------------------------------------
+%% Func: is_running/1
+%% Args: Application = atom()
+%% Purpose: Checks if an application is running.
+%%          This is used by application to avoid traversing an application
```
kjellwinblad, line 244: application child -> application's child ?

josevalim: Fixed.

PR #5019, opened 2021-07-01
Add no_return() spec to gen_statem:init/1 note

Maria-12648430:
If no spec is given, dialyzer will complain that `init/1` only terminates with explicit exception.

### DIFF
--- a/lib/stdlib/doc/src/gen_statem.xml
+++ b/lib/stdlib/doc/src/gen_statem.xml
@@ -2469,6 +2469,7 @@ handle_event(_, _, State, Data) ->
 	    in that case be implemented as:
 	  </p>
 	  <pre>
+-spec init(_) -> no_return().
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
It's in an example in the docs. (I didn't realize at first.)

Maria-12648430:
> It's in an example in the docs. (I didn't realize at first.)

Not in this corner of the docs, though ;) Where is the one you mentioned?

zuiderkwast:
@Maria-12648430 your change is in the docs, in a piece of code, in a note. That's what I meant. Reading the PR title without care I thought it's a spec in the actual `gen_statem` module.

Maria-12648430:
@zuiderkwast oic =^^=

RaimoNiskanen:
I wonder how many of our documentation code examples that consider Dialyzer?
A bigger review might be needed...

PR #5020, opened 2021-07-02
Optimize signal sending to processes with message_queue_data=off_heap enabled

kjellwinblad:
Erlang guarantees that signals (i.e., message signals and non-message signals) sent from a single process to another process are ordered in send order. However, there are no ordering guarantees for signals sent from different processes to a particular process. Therefore, several processes can send signals in parallel to a specific process without synchronizing with each other. However, such signal sending was previously always serialized as the senders had to acquire the lock for the outer signal queue of the receiving process. This commit makes it possible for several processes to send signals to a process with the message_queue_data=off_heap setting(1) activated in parallel and without interfering with each other. This parallel signal sending optimization yields much better scalability for signal sending than what was previously possible(2).

(1) Information about how to enable the message_queue_data=off_heap setting can be found in the documentation of the functions [erlang:process_flag/2](https://erlang.org/doc/man/erlang.html#process_flag-2) and [erlang:spawn_opt/4](https://erlang.org/doc/man/erlang.html#spawn_opt-4).

(2) http://winsh.me/bench/erlang_sig_q/sigq_bench_result.html

Implementation
--------------

The parallel signal sending optimization works only on processes with the message_queue_data=off_heap setting enabled. For processes with the message_queue_data=off_heap setting enabled, the new optimization is activated and deactivated on demand based on heuristics to give a small overhead when the optimization is unnecessary. The optimization is activated when the contention on the lock for the outer message queue is high. It is deactivated when the number of enqueued messages per fetch operation (that fetch messages from the outer message queue to the inner) is low.

When the optimization is active, the outer message queue has an array of signal buffers where sending processes enqueue signals. When the receiving process needs to fetch messages from the outer message queue, the contents of the non-empty buffers are append to the outer message queue. Each process is assigned a particular slot in the buffer array (the process ID is used to hash to a particular slot). That way, the system can preserve the send order between messages coming from the same process.

### DIFF
--- a/erts/emulator/beam/erl_process.h
+++ b/erts/emulator/beam/erl_process.h
@@ -1116,8 +1116,9 @@ struct process {
 
     erts_atomic32_t state;  /* Process state flags (see ERTS_PSFLG_*) */
     erts_atomic32_t dirty_state; /* Process dirty state flags (see ERTS_PDSFLG_*) */
-
+    Uint sig_inq_contention_counter;
     ErtsSignalInQueue sig_inq;
+    erts_atomic_t sig_inq_buffers;
     ErlTraceMessageQueue *trace_msg_q;
     erts_proc_lock_t lock;
     ErtsSchedulerData *scheduler_data;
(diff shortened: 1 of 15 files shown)


ranjanified:
Nice.. And Brilliant step :+1: ... Hope it doesn't affect anything else...

erts/emulator/beam/erl_process.h
```diff
@@ -1116,8 +1116,16 @@ struct process {
...
     erts_atomic32_t dirty_state; /* Process dirty state flags (see ERTS_PDSFLG_*) */
-
+    Uint sig_inq_contention_counter;
     ErtsSignalInQueue sig_inq;
+    erts_atomic_t sig_inq_buffers;
+    erts_atomic64_t sig_inq_buffers_refc;
+    /* 
+     * sig_inq_buffers_refc is incremented by dirty schedulers that access
+     * sig_inq_buffers to prevent deallocation while they are
+     * accessing the buffer array. This is needed since durty
```
garazdawi, line 1126:
```suggestion
     * accessing the buffer array. This is needed since dirty
```

kjellwinblad: Thanks, fixed in commit below.

PR #5022, opened 2021-07-02
test suites: replace ?t: with test_server:

max-au:
Removing `?t` previously retained for backward compatibility allows to
search for test_server callsites easier.

### DIFF
--- a/lib/common_test/include/ct.hrl
+++ b/lib/common_test/include/ct.hrl
@@ -39,5 +39,4 @@
 %% Backward compatibility for test_server test suites.
 %% DO NOT USE IN NEW TEST SUITES.
 -define(line,).
--define(t,test_server).
 -define(config,test_server:lookup_config).
--- a/lib/snmp/test/exp/snmp_agent_bl_test.erl
+++ b/lib/snmp/test/exp/snmp_agent_bl_test.erl
@@ -87,12 +87,12 @@ all(suite) -> {req,
 	       [{conf, init_all, cases(), finish_all}]}.
 
 init_per_testcase(_Case, Config) when list(Config) ->
-    Dog = ?t:timetrap(?t:minutes(6)),
+    Dog = test_server:timetrap(test_server:minutes(6)),
     [{watchdog, Dog}|Config].
 
 end_per_testcase(_Case, Config) when list(Config) ->
     Dog = ?config(watchdog, Config),
-    ?t:timetrap_cancel(Dog),
+    test_server:timetrap_cancel(Dog),
     Config.
 
 cases() ->
--- a/lib/stdlib/test/gen_statem_SUITE.erl
+++ b/lib/stdlib/test/gen_statem_SUITE.erl
@@ -541,7 +541,7 @@ abnormal1(Config) ->
 	   gen_statem:call(Name, {delayed_answer,2000}, 100),
 	   Reason),
     ok = gen_statem:stop(Name),
-    ?t:sleep(1100),
+    ct:sleep(1100),
     ok = verify_empty_msgq().
 
 %% Check that time outs in calls work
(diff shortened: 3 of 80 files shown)


lib/snmp/test/exp/snmp_agent_bl_test.erl
```diff
@@ -4993,14 +4993,14 @@ otp_4394_config(AgentDir, MgrDir, Ip0) ->
     ?line write_community_conf(AgentDir, [C1, C2]),
     ?line update_vacm(Vsn, AgentDir),
     Ta1 = {"shelob v1", 
-	   [134,138,177,177], 5000, 1500, 3, %% Anv�nd Ip och modda
+	   [134,138,177,177], 5000, 1500, 3, %% Anv�nd Ip och modda
```
u3s, line 4996: Do you know what happened with swedish character here?

lib/stdlib/test/gen_statem_SUITE.erl
```diff
@@ -2368,7 +2368,7 @@ init(stop) ->
 init(stop_shutdown) ->
     {stop,shutdown};
 init(sleep) ->
-    ?t:sleep(1000),
+    ct:sleep(1000),
```
u3s, line 2371: is replacement with ct and not test_server intended? functions has a different API and also calling ct:sleep will invoke test_server:adjusted_sleep providing other logic than test_server:sleep...

lib/common_test/include/ct.hrl
```diff
@@ -39,5 +39,4 @@
 %% Backward compatibility for test_server test suites.
 %% DO NOT USE IN NEW TEST SUITES.
 -define(line,).
--define(t,test_server).
```
u3s, line 42: such definition is copy/pasted(?) in several other files (e.g. disk_log_SUITE.erl) and seems unused. can you remove them as well in this PR?

lib/snmp/test/exp/snmp_agent_bl_test.erl
KennethL: If swedish is used in a comment please translate it to english at least if the line with the comment is touched in a commit.
We should not have swedish in the comments.

max-au: It was unintended, I will drop accidental changes caused by improper editor that broke non-ASCII characters. I think it might be better to clean up comments in a separate patch for review convenience.

lib/stdlib/test/gen_statem_SUITE.erl
max-au: Yes, I'd like to suggest this change. If my understanding it correct, this test suite needs "at least X" milliseconds, so adjusting that X should not hurt.
BTW are there official guidelines which method to use, `receive after 1000 -> ok end`, or `timer:sleep`, or `ct:sleep`, or `test_server:sleep`?

IngelaAndin: test_server:sleep is legacy, e.i it is from the original internal tool that now only exists as an internal part of the official common_test application. So it should not be used. I would think ct:sleep is best in the cases that there is no better alternative to sleep in a test suite as it adopts the sleep period for slower test runs such as when you run cover for instance.

IngelaAndin:
@max-au I think it looks good. However, I am thinking this is a test suite refresh, has no impact on functionality but changes a lot of test code, perhaps we ought to rebase it on maint to reduce the maintenance burden.  Also when you are at it maybe change  ?t:fail to ct:fail for the same reason as sleep.

max-au:
Cleaning spree. 
As suggested, rebased to maint, and replaced all (except those in test_server_SUITE itself) invocations of `test_server:fail` with corresponding `ct:fail` calls.

IngelaAndin:
Thank you for your work on this.

PR #5023, opened 2021-07-03
edoc: Add source file to the warning on skipped tag

wojtekmach:
Before this patch:

    edoc: warning: 'blockquote' is not allowed - skipping tag, extracting content

After:

    telemetry_poller.erl: warning: 'blockquote' is not allowed - skipping tag, extracting content


garazdawi:
So much code just to carry the `Opts` all the way down there... looks good to me. Putting into test.

wojtekmach:
> So much code just to carry the Opts all the way down there

tell me about it :D

garazdawi:
thanks!

PR #5025, opened 2021-07-05
Lock free atom table

mikpe:
This converts the atom table to be lock-free. The central change is to use a lock-free hash table, based on the so-called "split-ordered lists" by Shalev and Shavit (Journal of the ACM, May 2006). The index table and text buffer have also been made lock-free, but those were straight-forward.

Even read-only workloads, e.g. `list_to_existing_atom/1` or the initial lookup in `list_to_atom/1`, need to take a shared lock around the atom table, causing cache-line ping-pongs between cores. By making the atom table lock-free, this is avoided. On a 32-core Xeon Platinum 8175M, a benchmark that repeatedly calls `binary_to_existing_atom/1` for all known atoms concurrently on all cores gets 9% faster with this lock-free atom table.

I did not convert the common hash table to be lock-free, since it needs to support deletions, and that gets tricky in a lock-free setting (not the unlinking of the object from the structure but the actual deallocation). The atom table doesn't need deletions. And the common index table is too tied to the common hash table to be useful here.

I couldn't find atomics support for pointers (mainly CAS, but then one also needs atomic read/write) so I implemented them as thin layers on top of the regular atomics. Please let me know if there's a cleaner way of doing this without losing proper typing.

As `atom.c` is almost entirely rewritten I recommend reading the individual commits.

Some TODOs:
* relax atomics to use less than full memory barriers, where applicable
* find a faster way to do the bit-reversal, even is that means using `asm()`
* likewise find a faster way to compute "parent bucket" (clear highest set bit)

### DIFF
--- a/erts/emulator/beam/atom.c
+++ b/erts/emulator/beam/atom.c
@@ -113,7 +113,12 @@ static Uint atom_space;		/* Amount of atom text space used */
 static void atom_index_info(fmtfn_t to, void *arg)
 {
     AtomIndexTable *t = &erts_atom_table;
+    int lock = !ERTS_IS_CRASH_DUMPING;
+    if (lock)
+	atom_read_lock();
     hash_info(to, arg, &t->htable);
+    if (lock)
+	atom_read_unlock();
     erts_print(to, arg, "=index_table:%s\n", t->htable.name);
     erts_print(to, arg, "size: %d\n", (int) erts_atomic_read_mb(&t->size));
     erts_print(to, arg, "limit: %d\n", t->limit);
@@ -126,10 +131,18 @@ static void atom_index_info(fmtfn_t to, void *arg)
 static int atom_index_table_sz(void)
 {
     AtomIndexTable *t = &erts_atom_table;
+    int lock = !ERTS_IS_CRASH_DUMPING;
+    int hsz;
+
+    if (lock)
+	atom_read_lock();
+    hsz = hash_table_sz(&t->htable);
+    if (lock)
+	atom_read_unlock();
     return (sizeof(AtomIndexTable)
 	    - sizeof(Hash)
 	    + (int) erts_atomic_read_mb(&t->size) * sizeof(AtomInt *)
-	    + hash_table_sz(&t->htable));
+	    + hsz);
 }
 
 /*
@@ -161,7 +174,9 @@ static int atom_index_put(AtomInt *tmpl)
     int size, size0;
     AtomInt **segment;
 
+    atom_write_lock();
     p = (AtomInt *) hash_put(&t->htable, tmpl);
+    atom_write_unlock();
     ix = (int) erts_atomic_read_mb(&p->index);
     if (ix >= 0) {
 	return ix;
@@ -208,7 +223,11 @@ static int atom_index_put(AtomInt *tmpl)
 static int atom_index_get(AtomInt *tmpl)
 {
     AtomIndexTable *t = &erts_atom_table;
-    AtomInt *p = (AtomInt *) hash_get(&t->htable, tmpl);
+    AtomInt *p;
+
+    atom_read_lock();
+    p = (AtomInt *) hash_get(&t->htable, tmpl);
+    atom_read_unlock();
     return p ? (int) erts_atomic_read_mb(&p->index) : -1;
 }
 
@@ -223,17 +242,11 @@ static AtomInt *atom_index_lookup(Uint ix)
  */
 void atom_info(fmtfn_t to, void *to_arg)
 {
-    int lock = !ERTS_IS_CRASH_DUMPING;
-    if (lock)
-	atom_read_lock();
     atom_index_info(to, to_arg);
 #ifdef ERTS_ATOM_PUT_OPS_STAT
     erts_print(to, to_arg, "atom_put_ops: %ld\n",
 	       erts_atomic_read_nob(&atom_put_ops));
 #endif
-
-    if (lock)
-	atom_read_unlock();
 }
 
 /*
@@ -450,9 +463,7 @@ erts_atom_put_index(const byte *name, Sint len, ErtsAtomEncoding enc, int trunc)
 
     a.atom.len = tlen;
     a.atom.name = (byte *) text;
-    atom_read_lock();
     aix = atom_index_get(&a);
-    atom_read_unlock();
     if (aix >= 0) {
 	/* Already in table no need to verify it */
 	return aix;
@@ -490,9 +501,7 @@ erts_atom_put_index(const byte *name, Sint len, ErtsAtomEncoding enc, int trunc)
     a.atom.len = tlen;
     a.atom.latin1_chars = (Sint16) no_latin1_chars;
     a.atom.name = (byte *) text;
-    atom_write_lock();
     aix = atom_index_put(&a);
-    atom_write_unlock();
     return aix;
 }
 
@@ -518,26 +527,12 @@ am_atom_put(const char* name, Sint len)
 
 int atom_table_size(void)
 {
-    int ret;
-    int lock = !ERTS_IS_CRASH_DUMPING;
-    if (lock)
-	atom_read_lock();
-    ret = (int) erts_atomic_read_mb(&erts_atom_table.entries);
-    if (lock)
-	atom_read_unlock();
-    return ret;
+    return (int) erts_atomic_read_mb(&erts_atom_table.entries);
 }
 
 int atom_table_sz(void)
 {
-    int ret;
-    int lock = !ERTS_IS_CRASH_DUMPING;
-    if (lock)
-	atom_read_lock();
-    ret = atom_index_table_sz();
-    if (lock)
-	atom_read_unlock();
-    return ret;
+    return atom_index_table_sz();
 }
 
 Eterm
@@ -586,9 +581,7 @@ erts_atom_get(const char *name, Uint len, ErtsAtomEncoding enc)
         break;
     }
 
-    atom_read_lock();
     i = atom_index_get(&a);
-    atom_read_unlock();
 
     return (i >= 0) ? make_atom(i) : THE_NON_VALUE;
 }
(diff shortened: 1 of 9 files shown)


kjellwinblad:
Thanks for the PR!

Is this work motivated by a real-world application that performs better due to these changes?

> Even read-only workloads, e.g. `list_to_existing_atom/1` or the initial lookup in `list_to_atom/1`, need to take a shared lock around the atom table, causing cache-line ping-pongs between cores. By making the atom table lock-free, this is avoided. On a 32-core Xeon Platinum 8175M, a benchmark that repeatedly calls `binary_to_existing_atom/1` for all known atoms concurrently on all cores gets 9% faster with this lock-free atom table.

The shared lock for the atom table is a readers-writer lock optimized for frequent reads. A frequent read optimized lock means that if the number of schedulers is less than the reader groups limit (configurable(1)), each scheduler gets its own cache line in the lock data structure. When the lock is acquired for reading, only the current scheduler's own cache line needs to be modified. Therefore, the cache-line ping-pongs that you describe should not happen. As long as the atom table is not locked for writing, reading from it should be very cheap. Locking the table for writing is more expensive as the writer has to look at all reader groups, and all access to the table is blocked while the lock is held. However, writing to the atom table should be rare as Erlang users are discouraged from creating atoms dynamically.

I think it is surprising that the lock-free atom table is faster in the read-only scenario you describe. The lock-free table might (I have not had a deep look at it yet) introduce more complex logic and more memory indirection. Can you change the ETHR_CACHE_LINE_SIZE constant from 64 to 128, recompile and rerun the benchmark? I think some Intel CPUs sometimes fetches two cache lines instead of one (just in case the next cache-line will be needed later), so changing the cache-line size constant may have some impact on this benchmark.

The current atom table has a fixed size, and the lock-free table can expand dynamically, so the lock-free table might have an advantage if the atom table is too full. Can you also duplicate the size of the atom table(2) and rerun your benchmark to see if that makes some difference?

(1) See +rg in https://erlang.org/doc/man/erl.html

(2) One can configure the size of the atom table with the +t parameter (https://erlang.org/doc/man/erl.html), and the current size can be found by executing the function `erlang:system_info(atom_limit)`.

mikpe:
> Thanks for the PR!
> 
> Is this work motivated by a real-world application that performs better due to these changes?

It was motivated by a thread in erlang-questions about 4 weeks ago, about the performance of `binary_to_term/1`. At one point someone argued that for trivial atoms like `'a'`, `atom_to_binary/1` and `binary_to_atom/1` ought to have similar costs, but observed that they hadn't, to which I replied that it was because of the cost for "interning" atoms.

That made me investigate the possibility of making the atom table lock-free.

In terms of real-world applications I'll just note that at Klarna we have a lot of data being read from DB tables or parsed from network messages, and there's usually a lot of atoms in that data, so speeding that up makes a lot of sense.

> > Even read-only workloads, e.g. `list_to_existing_atom/1` or the initial lookup in `list_to_atom/1`, need to take a shared lock around the atom table, causing cache-line ping-pongs between cores. By making the atom table lock-free, this is avoided. On a 32-core Xeon Platinum 8175M, a benchmark that repeatedly calls `binary_to_existing_atom/1` for all known atoms concurrently on all cores gets 9% faster with this lock-free atom table.
> 
> The shared lock for the atom table is a readers-writer lock optimized for frequent reads. A frequent read optimized lock means that if the number of schedulers is less than the reader groups limit (configurable(1)), each scheduler gets its own cache line in the lock data structure. When the lock is acquired for reading, only the current scheduler's own cache line needs to be modified. Therefore, the cache-line ping-pongs that you describe should not happen. As long as the atom table is not locked for writing, reading from it should be very cheap. Locking the table for writing is more expensive as the writer has to look at all reader groups, and all access to the table is blocked while the lock is held. However, writing to the atom table should be rare as Erlang users are discouraged from creating atoms dynamically.

Thanks, I didn't know about this one-lock-cache-line-per-scheduler thing.
 
> I think it is surprising that the lock-free atom table is faster in the read-only scenario you describe. The lock-free table might (I have not had a deep look at it yet) introduce more complex logic and more memory indirection. Can you change the ETHR_CACHE_LINE_SIZE constant from 64 to 128, recompile and rerun the benchmark? I think some Intel CPUs sometimes fetches two cache lines instead of one (just in case the next cache-line will be needed later), so changing the cache-line size constant may have some impact on this benchmark.

This looks likely. I rebuild my vanilla baseline OTP-24 with `--with-assumed-cache-line-size=128` and that sped the benchmark up by 25%.

> The current atom table has a fixed size, and the lock-free table can expand dynamically, so the lock-free table might have an advantage if the atom table is too full. Can you also duplicate the size of the atom table(2) and rerun your benchmark to see if that makes some difference?

I don't think this is it. I run the benchmark in a freshly started VM which only has about 9K atoms, far less than the 1M limit.

kjellwinblad:
> It was motivated by a thread in erlang-questions about 4 weeks ago, about the performance of `binary_to_term/1`. At one point someone argued that for trivial atoms like `'a'`, `atom_to_binary/1` and `binary_to_atom/1` ought to have similar costs, but observed that they hadn't, to which I replied that it was because of the cost for "interning" atoms.
> 
> That made me investigate the possibility of making the atom table lock-free.
> 
> In terms of real-world applications I'll just note that at Klarna we have a lot of data being read from DB tables or parsed from network messages, and there's usually a lot of atoms in that data, so speeding that up makes a lot of sense.

The Erlang/OTP system is currently not designed for creating an unbounded number of atoms. This limitation gives freedom to do optimizations that would otherwise not be possible (e.g., having a fixed-sized atom table highly optimized for reads and slow for writes and not ever removing created atoms). Creating atoms from data coming from databases and network messages is fine as long as one can ensure a limit on the total number of atoms that can be created. Is this the case for your use case at Karna? If that is the case, I don't think you will get a performance benefit from the look-free atom table as most of the atom table accesses are likely read-accesses. You can also avoid the risk of long latency due to a locked atom table by pre-creating the atoms that the system will use when the system is starting. 

I think it should be considered bad practice to use `erlang:list_to_atom` and `erlang:binary_to_atom` in a long-running system for the abovementioned reason. The functions `erlang:list_to_existing_atom` and `binary_to_existing_atom` were added to avoid the risk of creating too many atoms. We currently have notes in the documentation of [`erlang:list_to_atom`](https://erlang.org/doc/man/erlang.html#list_to_atom-1) and [`erlang:binary_to_atom`](https://erlang.org/doc/man/erlang.html#binary_to_atom-2) about that one should consider using the `erlang:*_to_existing_atom` functions instead.

I think the main advantages of your lock-free atom table implementation is that it makes creating new atoms faster and that it does not block the whole atom table while inserting new atoms. However, I'm still not convinced that this is a powerful motivation as creating new atoms should be relatively rare. As I have tried to argue above, read-accesses to the atom table might get slower with the lock-free implementation (at least with an optimally configured assumed-cache-line-size). That being said, I have not looked at your implementation in detail yet, so I don't know if it can do some read tasks faster than the current atom table. For example, I can imagine that having a dynamically expandable table can make the table more compact, which can have some cache-locality benefits in read-only scenarios. However, one could also implement an expandable table in a blocking way. You also mentioned that the memory barriers are not optimal yet, so there are possibilities for improvement.

> This looks likely. I rebuilt my vanilla baseline OTP-24 with `--with-assumed-cache-line-size=128` and that sped the benchmark up by 25%.

 This means that the current atom-table with `--with-assumed-cache-line-size=128` is faster than your lock-free atom table implementation with about 15% in your benchmark, right? It is very interesting that the change of `assumed-cache-line-size` had so big impact. I think we should consider changing the `assumed-cache-line-size`'s default value from 64 to 128. That constant and frequent read optimized readers-writer locks are used in many other places in the Erlang/OTP system (e.g., ETS tables) so that change might benefit other parts of the system as well (at least on some Intel CPUs).

kjellwinblad:
@mikpe FYI, due to vacations, we from the Erlang/OTP team at Ericsson will probably not have time to look more at your PR in the coming ~3-4 weeks.

ssmyczynski:
> find a faster way to do the bit-reversal, even if that means using `asm()`

I am quite sure you can skip that point from your TODO list. 
gcc compiler is smart enough to compress the first two lines of `reverse` function:
```c
    w = ((w & 0x0000FFFF) << 16) |  (w >> 16);
    w = ((w & 0x00FF00FF) << 8)  | ((w >> 8) & 0x00FF00FF);
```
into single assembler instruction (at least for x86 architecture):
```asm
    bswap %edi
```

kjellwinblad:
@mikpe Do you have any real use case that would benefit from the lock-free atom table instead of the current atom table with `--with-assumed-cache-line-size=128`? If not, I don't currently think there is good enough motivation to take this in, but I would be happy to be convinced otherwise.

mikpe:
> @mikpe Do you have any real use case that would benefit from the lock-free atom table instead of the current atom table with `--with-assumed-cache-line-size=128`? If not, I don't currently think there is good enough motivation to take this in, but I would be happy to be convinced otherwise.

I'll have to re-run the benchmark to see how the larger cache line size affects the lock-free atom table, but my suspicion is that it's the false sharing of the per-scheduler locks that's the culprit, and enlarging the cache line size is the appropriate fix. On vacation now, will check in two weeks.

kjellwinblad:
@mikpe, Is there any news regarding the benchmark? Otherwise, I will close this for the time being, and we can open it again if we get measurements that indicate that this PR can improve performance.

Btw, we are planning to do a small investigation to decide if we should use `--with-assumed-cache-line-size=128` as default. Do you use `--with-assumed-cache-line-size=128` at Klarna now? If so, have you noticed any change in performance?

mikpe:
> @mikpe, Is there any news regarding the benchmark? Otherwise, I will close this for the time being, and we can open it again if we get measurements that indicate that this PR can improve performance.
> 
> Btw, we are planning to do a small investigation to decide if we should use `--with-assumed-cache-line-size=128` as default. Do you use `--with-assumed-cache-line-size=128` at Klarna now? If so, have you noticed any change in performance?

No news, been too busy since vacation. I'll test (our version of) OTP-22 with --with-assumed-cache-line-size=128 next week to see if that performs well, and if so I'll try to get that into prod well before BF week.

I agree that the current PR is complex and probably not necessary, so I'll close it (but I'll keep the branch for the time being).

Issue #5009, opened 2021-06-28
Error downgrading TLS/DTLS just after connection


garthk:
Using `ssl:close/2` to hand over to a new controller process just after `ssl:connect/3` fails with `{error, closed}`.

**To Reproduce**
With Erlang/OTP 24.0.2:

```erlang
#!/usr/bin/env escript
%% -*- erlang -*-
% @format({formatter, default_formatter}, #{paper => 98}).

main([]) ->
    application:ensure_all_started(ssl),
    % The apparent need to call  ssl:close/2 to renegotiate is in jamdb_oracle_conn
    % but I can reproduce the unexpected {error, _} response using erlang.org:
    TcpHost = "www.erlang.org",
    TcpPort = 443,
    Ciphers =
        ssl:filter_cipher_suites(
            ssl:cipher_suites(all, 'tlsv1.3'), []),
    SslOptions =
        [{cacertfile, "_build/dev/lib/castore/priv/cacerts.pem"},
         {certfile, "wallet/clcerts.pem"},
         {ciphers, Ciphers},
         {keyfile, "wallet/priv.pem"},
         {server_name_indication, TcpHost},
         {verify, verify_peer}],
    {ok, Port} =
        gen_tcp:connect(TcpHost,
                        TcpPort,
                        [binary,
                         {active, false},
                         {packet, raw},
                         {nodelay, true},
                         {keepalive, true}],
                        15000),
    {ok, {sslsocket, {gen_tcp, Port, tls_connection, undefined}, [_, _]} = Socket} =
        ssl:connect(Port, SslOptions, 1000),
    % Insert delay to prevent the failure:
    % timer:sleep(1000),
    {ok, Port} = ssl:close(Socket, {self(), 15000}),
    gen_tcp:close(Port).
```

**Expected behavior**
This line to successfully match, even if there's been no delay:
```erlang
    {ok, Port} = ssl:close(Socket, {self(), 15000}),
```

**Affected versions**
Erlang/OTP 24.0.2 on macOS 11.4 (20F71) installed with `asdf` and compiled with:
```bash
KERL_BUILD_DOCS="yes"
KERL_CONFIGURE_OPTIONS="--without-javac --with-ssl=/usr/local/opt/openssl@1.1"
```

**Additional context**
[...]

PR #5030, opened 2021-07-07
Ingela/ssl/downgrade/gh 5009/otp 17517

IngelaAndin:
Solve at least some issues discovered due to  #5009

### DIFF
--- a/lib/ssl/src/ssl_connection.hrl
+++ b/lib/ssl/src/ssl_connection.hrl
@@ -92,8 +92,9 @@
 
 -record(connection_env, { 
                           user_application      :: {Monitor::reference(), User::pid()},
-                          downgrade,
-                          terminated = false                          ::boolean() | closed,  
+                          downgrade             :: {NewController::pid(), From::gen_statem:from()} | 'undefined',
+                          socket_terminated = false                          ::boolean(),
+                          socket_tls_closed = false                          ::boolean(),
                           negotiated_version    :: ssl_record:ssl_version() | 'undefined',
                           erl_dist_handle = undefined :: erlang:dist_handle() | 'undefined',
                           private_key          :: public_key:private_key() | secret_printout() | 'undefined'
--- a/lib/ssl/src/ssl_gen_statem.erl
+++ b/lib/ssl/src/ssl_gen_statem.erl
@@ -594,40 +594,36 @@ connection({call, From}, negotiated_protocol,
                                                  negotiated_protocol = undefined}} = State) ->
     hibernate_after(?FUNCTION_NAME, State,
 		    [{reply, From, {ok, SelectedProtocol}}]);
-connection({call, From},
-           {close, {Pid, _Timeout}},
-           #state{connection_env = #connection_env{terminated = closed} = CEnv,
-                 protocol_specific = PS} = State) ->
-    {next_state, downgrade, State#state{connection_env =
-                                            CEnv#connection_env{terminated = true,
-                                                                downgrade = {Pid, From}},
-                                        protocol_specific = PS#{active_n_toggle => true,
-                                                                active_n => 1}
-                                       },
-     [{next_event, internal, ?ALERT_REC(?WARNING, ?CLOSE_NOTIFY)}]};
-connection({call, From},
-           {close,{Pid, Timeout}},
+connection({call, From}, 
+           {close,{NewController, Timeout}},
            #state{connection_states = ConnectionStates,
                   static_env = #static_env{protocol_cb = Connection},
                   protocol_specific = #{sender := Sender} = PS,
-                  connection_env = CEnv
+                  connection_env = #connection_env{socket_tls_closed = PeerClosedTLS} = CEnv
                  } = State0) ->
+    Action = case PeerClosedTLS of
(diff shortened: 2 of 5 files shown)


lib/ssl/src/ssl_connection.hrl
```diff
@@ -93,7 +93,8 @@
 -record(connection_env, { 
                           user_application      :: {Monitor::reference(), User::pid()},
                           downgrade,
```
KennethL, line 95: Why not have a type spec for the downgrade field as well. Should be `{Pid::pid(),From:pid()}` i think. The question is what to name the first element with a better name than Pid.

u3s (commented):
I've left some minor comments inline.

lib/ssl/src/ssl_gen_statem.erl
```diff
@@ -1051,13 +1046,13 @@ maybe_invalidate_session({false, first}, server = Role, Host, Port, Session) ->
...
-    %% when ssl:close/1 returns unless it is a downgrade where
-    %% we want to guarantee that close alert is received before
-    %% returning. In both cases terminate has been run manually
-    %% before run by gen_statem which will end up here
+terminate({shutdown, downgrade}, downgrade, State) ->
+    %% Socket shall not be closed as it sgould be returned to user
+    handle_trusted_certs_db(State);
+terminate(_, _, #state{connection_env = #connection_env{socket_terminated = true}}) ->
+    %% Happens when user closes the connection using ssl:close/1 e.i. terminate
+    %% is called explicitly beforhand to to guarantee that Transport:close has been called
```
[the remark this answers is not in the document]
IngelaAndin: fixed

lib/ssl/src/ssl_gen_statem.erl
```diff
@@ -1051,13 +1046,13 @@ maybe_invalidate_session({false, first}, server = Role, Host, Port, Session) ->
...
 
-terminate(_, _, #state{connection_env = #connection_env{terminated = true}}) ->
-    %% Happens when user closes the connection using ssl:close/1
-    %% we want to guarantee that Transport:close has been called
-    %% when ssl:close/1 returns unless it is a downgrade where
-    %% we want to guarantee that close alert is received before
-    %% returning. In both cases terminate has been run manually
-    %% before run by gen_statem which will end up here
+terminate({shutdown, downgrade}, downgrade, State) ->
+    %% Socket shall not be closed as it sgould be returned to user
```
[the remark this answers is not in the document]
IngelaAndin: fixed

lib/ssl/src/ssl_connection.hrl
IngelaAndin: I made it {NewController::pid(), Caller::gen_statem:from()}

lib/ssl/src/ssl_connection.hrl
```diff
@@ -92,8 +92,9 @@
 
 -record(connection_env, { 
                           user_application      :: {Monitor::reference(), User::pid()},
-                          downgrade,
-                          terminated = false                          ::boolean() | closed,  
+                          downgrade             :: {NewController::pid(), Caller::gen_statem:from()} | 'undefined',
```
KennethL, line 95: If the element in the tuple are called NewController and Caller in the type spec why are the variable names not named the same? Now when matching the variables are named Pid and From but why not NewController and Caller instead?
Just a thought.

lib/ssl/src/ssl_gen_statem.erl
```diff
@@ -596,16 +596,11 @@ connection({call, From}, negotiated_protocol,
...
-                                                                downgrade = {Pid, From}},
-                                        protocol_specific = PS#{active_n_toggle => true,
-                                                                active_n => 1}
-                                       },
-     [{next_event, internal, ?ALERT_REC(?WARNING, ?CLOSE_NOTIFY)}]};
-connection({call, From},
+           #state{connection_env = #connection_env{socket_tls_closed = true} = CEnv} = State) ->
+    %%Close Alert already received from peer, "replay" in downgrade state.
+    {next_state, downgrade, State#state{connection_env = CEnv#connection_env{downgrade = {Pid, From}}},
+                                        [{next_event, internal, ?ALERT_REC(?WARNING, ?CLOSE_NOTIFY)}]};
```
u3s, line 603: Should the tls_sender:downgrade/2 call go here as well or not?

lib/ssl/src/ssl_connection.hrl
IngelaAndin: Well, to make it consistent I changed it again to  NewController and From as From is very commonly used to refer to the caller when using the gen_* behaviours.  And changed the code accordingly!

lib/ssl/src/ssl_gen_statem.erl
IngelaAndin: Humm ... good question I actually think it should! I made a new version.

IngelaAndin:
This is merged now as it is an improvement of downgrade handling, although I am not sure if it completely solves the reported issue. We have a known race condition that we are working on independently of this that involves the downgrade scenario. Merging this will facilitate the continued work on that.

PR #5032, opened 2021-07-07
Do not unload modules during stop

### LINKED ISSUE #5031: init:stop() is slow to exit due to unloading modules
**Describe the bug**
Currently init:stop() unloads all modules before exiting and this can take a really long time on some platforms. For example, on a Raspberry Pi Zero unloading modules can take nearly 2 minutes even though loading all of them in took 10-20 seconds. 

This [mailing list thread](https://erlang.org/pipermail/erlang-questions/2021-July/101209.html) contains more information.

**To Reproduce**
I would imagine that init:stop() is slow on all platforms, but just not as obvious as it is on a Raspberry Pi Zero.

**Expected behavior**
init:stop() exits the VM in a couple seconds on a RPi Zero. Responses to the mailing list thread suggest removing the code that unloads modules, and based on what I've seen, that would fix the issue.

**Affected versions**
I only tested OTP 24.0.2.

josevalim:
Closes #5031.

I did not update the bootstrap code under the assumption you likely
don't want external contributors to change it, regardless of how
trustworthy they are. :)

### DIFF
--- a/erts/preloaded/src/init.erl
+++ b/erts/preloaded/src/init.erl
@@ -463,7 +463,7 @@ new_kernelpid({_Name,ignore},BootPid,State) ->
     State;
 new_kernelpid({Name,What},BootPid,State) ->
     erlang:display({"could not start kernel pid",Name,What}),
-    clear_system(BootPid,State),
+    clear_system(false,BootPid,State),
     crash("could not start kernel pid", [Name, What]).
 
 %% Here is the main loop after the system has booted.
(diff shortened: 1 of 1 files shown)


fhunleth:
I just confirmed that this patch reduces init:stop() from around 2 minutes to just under 2 seconds for the Raspberry Pi Zero in #5031 (see mailing list thread).

bjorng (changes requested):
Any reason that this PR shouldn't be based on maint? I think that this change can be considered a bug fix and that it therefore should be included in OTP 24.1.

josevalim:
Applied changes and rebased to maint.

bjorng (approved):

bjorng:

bjorng:
Thanks for your pull request.

PR #5035, opened 2021-07-08
Dgud/debugger/record fixes

dgud:


uabboli:
If I have a record #r{] and writes '#r{}' in 'Evaluator:' it says undefined_record. Also, if R is set to #{}, the evaluated value of R is displayed as {r,...}.

uabboli:
It's a pity that records from other modules are not printed nicely. And maybe it'd useful if the pretty-printing of tuples could be turned off. Do you plan to do more, or is this a finished job?

dgud:
My thought was that it was finished for now, but alas, something is broken in message handling and mnesia no longer starts up
if interpreted so I have introduced some bug somewhere.

The evaluator doesn't currently support records I have not looked at how to handle that, neither editing data/bindings, 
but I thought this was a good improvement to start with.

dgud:
I have improved the handling of records in the 'Evalutor', it now prints records and you can access fields of an variable.

lib/debugger/src/dbg_iload.erl
```diff
@@ -538,23 +664,259 @@ make_bit_type(_Line, Size, Type0) ->            %Integer or 'all'
...
+not_a_tuple({float,_,_}) -> true;
+not_a_tuple({nil,_}) -> true;
+not_a_tuple({cons,_,_,_}) -> true;
+not_a_tuple({char,_,_}) -> true;
+not_a_tuple({string,_,_}) -> true;
+not_a_tuple({record_index,_,_,_}) -> true;
+not_a_tuple({bin,_,_}) -> true;
+not_a_tuple({op,_,_,_}) -> true;
+not_a_tuple({op,_,_,_,_}) -> true;
+not_a_tuple(_) -> false.
```
bjorng, line 740: There is no need for the evaluator to handle non-tuples as a special case.

```suggestion
    Fs = record_fields(Name, Anno, St),
    expr({call,Anno,{remote,Anno,{atom,Anno,erlang},{atom,Anno,is_record}},
          [Term,{atom,Anno,Name},{integer,Anno,length(Fs)+1}]}, false, St).
```

bjorng (changes requested):
Looks good to me, except for a bug and a possibility for further simplifications.

bjorng (approved):

dgud:
Merged

Issue #4702, opened 2021-04-01
win32reg:delete_key() does not work


b-winkler:
The function `win32reg:delete_key()` does not and can never work.
It does not delete the current registry key and always returns `{error,einval}` instead of `ok`.

Can easily be reproduced on Windows:
```
{ok, H}  = win32reg:open([write]).
ok       = win32reg:change_key_create(H, "\\hkcu\\software\\my-test-key").
{ok, []} = win32reg:sub_keys(H). % no subkeys that would prevent deletion
ok       = win32reg:delete_key(H).  % <-- badmatch with {error,einval}
```

Implementation seems to be `registry_drv.c` with the following [line](https://github.com/erlang/otp/blob/master/erts/emulator/drivers/win32/registry_drv.c#L321):
```
result = RegDeleteKey(rp->hkey, NULL);
```

However, the [MS documentation](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regdeletekeya) of `RegDeleteKey()` explicitly states conditions for `lpSubKey` (the 2nd argument):
> This parameter cannot be **NULL**.

Since the implementation stems from R13B03 I doubt that it has ever worked.

Kind regards,
Benjamin

b-winkler:
My take:
The promised functionality never worked (needs to be confirmed), so the function could be abandoned and replaced by
`win32reg:delete_subkey(Handle, Subkey)` which mirrors the win32 interface more closely.

Benjamin

garazdawi:
Thanks for the bug report. This is not an area that we prioritize, so if you want a fix, the fastest way would be to submit a PR with a test case.

sverker:
Fixed in OTP 24.1.

PR #5027, opened 2021-07-05
Allow binary bifs in match_spec and trace pattern guards

stolen:
This commit adds byte_size/2, binary_part/2 and binary_part/3 to
a valid guard lists, extending functionality of ets select and tracing.

These calls are now valid:
```
> dbg:fun2ms(fun([B, _]) when byte_size(B) == 4 -> return_trace() end).
> ets:fun2ms(fun({B}) when byte_size(B) == 4 -> ok end).

> dbg:fun2ms(fun([B, _]) when binary_part(B, {2, 2}) == <<"ll">> -> return_trace() end).
> ets:fun2ms(fun({B}) when binary_part(B, {2, 2}) == <<"ll">> -> ok end).

> dbg:fun2ms(fun([B, _]) when binary_part(B, 2, 2) == <<"ll">> -> return_trace() end).
> ets:fun2ms(fun({B}) when binary_part(B, 2, 2) == <<"ll">> -> ok end).
```

Also `erts_internal:trace_pattern` and `ets:select` work with these match specs.

Fixes #5026

PR #5038, opened 2021-07-09
Fix win32reg issues #5037 and #4702

### LINKED ISSUE #5037: win32reg:value/2 fails to get default value ('default' or "")
The function win32reg:value/2 cannot be used to get the default value of a registry key.  
- Passing 'default' results in `Bad value on output port 'registry__drv__ wr'`
- Passing "" results in `error:{case_clause,{value,{default,"test"}}} at [{win32reg,value,2,[{file,"win32reg.erl"},{line,175}]},`

The expected outcome is `{ok, Value}` when a value exists.

This affects OTP 24.0

I have a pull request #5038 which should fix this issue, as well as #4702 and I also noticed a weakness in change_key/2 when passed a relative path of .. and at the "top" of the tree.

reganheath:
Pull request to resolve #5037 and #4702 (issues with win32reg).  I have to be honest, I found it very hard to debug and test this due to the complexity of building OTP on windows.  I managed to build it in WSL, but of course this doesn't help as the code in Q is windows only.

### DIFF
--- a/erts/emulator/drivers/win32/registry_drv.c
+++ b/erts/emulator/drivers/win32/registry_drv.c
@@ -318,8 +318,12 @@ reg_from_erlang(ErlDrvData clientData, char* buf, ErlDrvSizeT count)
 	}
 	break;
     case CMD_DELETE_KEY:
-	result = RegDeleteKey(rp->hkey, NULL);
-	reply(rp, result);
+	{
+	    char* key = buf;
+
+	    result = RegDeleteKey(rp->hkey, key);
+	    reply(rp, result);
+	}
 	break;
     case CMD_DELETE_VALUE:
 	result = RegDeleteValue(rp->hkey, buf);
--- a/lib/stdlib/src/win32reg.erl
+++ b/lib/stdlib/src/win32reg.erl
@@ -180,10 +180,17 @@ set_value({win32reg, Reg}, Name0, Value) when is_port(Reg) ->
       Name :: name(),
       ReturnValue :: {'ok', Value :: value()} | {'error', ErrorId :: atom()}.
 
-value({win32reg, Reg}, Name) when is_port(Reg) ->
+value({win32reg, Reg}, Name0) when is_port(Reg) ->
+    Name =
+	case Name0 of
+	    default -> [];
+	    _ -> Name0
+	end,
     Cmd = [?cmd_get_value, Name, 0],
     Reg ! {self(), {command, Cmd}},
     case get_result(Reg) of
+	{value, {default, Value}} when Name =:= [] ->
+	    {ok, Value};
 	{value, {Name, Value}} ->
 	    {ok, Value};
 	{error, Reason} ->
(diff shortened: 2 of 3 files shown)


garazdawi:
Hello! Github actions build windows installers that you can download and test your changes with. If you click on the [actions](https://github.com/erlang/otp/pull/5038/checks) tabs there is an artifacts dropdown where you can download `otp_prebuilt_win32`.

reganheath:
Awesome.  I have downloaded the windows installer, installed it on a clean VM, then run my test cases in an escript.  I found one mistake, corrected it, and I believe the changes are good.

erts/emulator/drivers/win32/registry_drv.c
```diff
@@ -318,8 +318,12 @@ reg_from_erlang(ErlDrvData clientData, char* buf, ErlDrvSizeT count)
...
 	break;
     case CMD_DELETE_KEY:
-	result = RegDeleteKey(rp->hkey, NULL);
-	reply(rp, result);
+	{
+	    char* key = buf;
+
+	    result = RegDeleteKey(rp->hkey, key);
+	    reply(rp, result);
+	}
```
sverker, line 326: I don't get how this could work (other than pure luck). The data buffer sent for CMD_DELETE_KEY does only contain one byte (the value 7). The pointer `buf` has been incremented (before the `switch`) to point just beyond that first and only byte into unknown territory. Passing it as argument to any function is just wrong.

According to `RegDeleteKey` documentation, the second argument `lpSubKey` should be the name of the subkey to be deleted.

sverker: Maybe better (as @b-winkler proposed) to replace the broken `delete_key` with a new `delete_subkey` which is simpler to implement and does not have the quirk to invalidate the current key.

reganheath: This necessitates a change in the documentation as well, which is why I avoided it initially.  If we're certain we prefer it, I can make that change.  Let me know :)

sverker: Well, if you have done a `delete_key` that actually works I'm fine with keeping it.

sverker:
@reganheath I would prefer if you could rearrange the commits with one for each issue, `value` and `delete_key`.

Also you should describe the actual fix in the commit headline, like "Fix win32reg:value/2 for default value". If referring the github issue or pull request do that with GH-5027 and PR-5038.

If you want, you could also rebase on maint ("git rebase --onto maint master") to get these fixes into OTP-24.1.

sverker:
... and if so, also click Edit and change this PR to be based on maint.

sverker:
Rebased on maint and put into test.

sverker:
Added `win32reg:change_key(Reg, "\\hkcu")` in the new tests to try fix failures with {error, eacces} on our test machines.

reganheath:
Sorry!  I've been swamped with stuff recently.

sverker:
Merged to maint for 24.1.
Thanks for the contribution.

PR #5041, opened 2021-07-09
ci: Cache wxWidgets on Windows

wojtekmach:
This shaves off ~7min from build time.

We put `~/wxWidgets`, as opposed to `/mnt/c/opt/local64/pgm/wxWidgets-*`, because with the latter restoring the cache doesn't work, it silently fails.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -7,9 +7,6 @@
 ## not possible so we need to rebuild all of Erlang/OTP multiple
 ## times.
 ##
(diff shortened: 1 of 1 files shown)


paulo-ferraz-oliveira:
I guess this is better than not having it, but FYI cache only lasts 7 days.

wojtekmach:
I have found even more bugs with actions/cache on windows. I'm going to close this for the time being. Sorry for the noise.

PR #5042, opened 2021-07-10
Doc typo: build_flavor was renamed to emu_flavor

michaelklishin:
but one doc guide was not updated accordingly.


garazdawi:
Thanks! Would you mind rebasing on maint so that we can get this fix in a bit earlier?

michaelklishin:
@garazdawi I will submit another PR for `maint`.

garazdawi:
Thanks! It should have been possible to reuse this PR by changing the base and force pushing to your branch. But opening a new PR works as well.

PR #5043, opened 2021-07-10
ci: Cache wxWidgets on Windows

wojtekmach:
This shaves off ~7min from build time.

We cache `~/wxWidgets` instead of `/mnt/c/opt/local64/pgm/wxWidgets-*`. With the latter, restoring the cache silently fails.


dominicletz:
Sweet!

garazdawi:
Great! Would you mind rebasing to maint so that we can get this on both maint and master?

wojtekmach:
@garazdawi done!

PR #5055, opened 2021-07-15
Optimize functions in erl_types to make Dialyzer a bit faster

uabboli:


uabboli:
Merged commit a40b5380e05a8b390c776556521b6e94e9e11b73 into erlang:master.

PR #5057, opened 2021-07-16
Fixing #5054, public_key:pkix_sign for RSASSA-PSS signature

### LINKED ISSUE #5054: public_key:pkix_sign generates incorrect signature for RSASSA-PSS algorithm
When you sign a certificate with RSASSA-PSS private key using function public_key:pkix_sign/2 it generates incorrect signature. Resulting certificate fails path validation.

**To Reproduce**

I've attached simple test to demonstrate issue: [test.zip](https://github.com/erlang/otp/files/6819964/test.zip)

Just execute script `test3` and it returns: `Certificate verification result: {error,{bad_cert,invalid_signature}}`

Script load certificate `cert.pem` (self-signed) and private key `key.pem` (RSASSA-PSS), both created in openssl:

~~~
openssl genpkey -algorithm rsa-pss -pkeyopt rsa_keygen_bits:2048 -pkeyopt rsa_keygen_pubexp:65537 -pkeyopt rsa_pss_keygen_md:sha256 -pkeyopt rsa_pss_keygen_saltlen:20 -out key.pem
openssl req -new -x509 -batch -days 1000 -key key.pem -subj '/C=AQ/O=Penguin Co./CN=Demo Cert' -out cert.pem
~~~

Then it signs hardcoded tbs certificate, writes it to `signed-cert.pem` and executes path validation function `public_key:pkix_path_validation/3`.

**Expected behavior**

Signed certificate must pass path validation.

**Affected versions**
OTP24

**Possible fix**

~~~~
diff --git a/lib/public_key/src/pubkey_cert.erl b/lib/public_key/src/pubkey_cert.erl
index 66d37ebf96..5310785328 100644
--- a/lib/public_key/src/pubkey_cert.erl
+++ b/lib/public_key/src/pubkey_cert.erl
@@ -537,10 +537,10 @@ gen_test_certs(
 
 
 x509_pkix_sign_types(#'SignatureAlgorithm'{algorithm = ?'id-RSASSA-PSS',
-                                           parameters = #'RSASSA-PSS-params'{hashAlgorithm = #'HashAlgorithm'{algorithm = Alg}}}) ->
+                                           parameters = #'RSASSA-PSS-params'{saltLength = SaltLen, hashAlgorithm = #'HashAlgorithm'{algorithm = Alg}}}) ->
     Hash = public_key:pkix_hash_type(Alg),
     {Hash, rsa_pss_pss, [{rsa_padding, rsa_pkcs1_pss_padding},
-                         {rsa_pss_saltlen, -1},
+                         {rsa_pss_saltlen, SaltLen},
                          {rsa_mgf1_md, Hash}]};
 x509_pkix_sign_types(#'SignatureAlgorithm'{algorithm = Alg}) ->
     {Hash, Sign} = public_key:pkix_sign_types(Alg),
~~~~

After that patch signed ceritficate passes path validation in erlang and in openssl.

sigsergv:
fixes #5054

### DIFF
--- a/lib/public_key/test/public_key_SUITE.erl
+++ b/lib/public_key/test/public_key_SUITE.erl
@@ -1373,6 +1373,6 @@ pss_params(sha256) ->
        maskGenAlgorithm = #'MaskGenAlgorithm'{algorithm = ?'id-mgf1',
                                               parameters = #'HashAlgorithm'{algorithm = ?'id-sha256'}
                                              },
-       saltLength = 20,
+       saltLength = 32,
        trailerField = 1}.
    
--- a/lib/public_key/test/public_key_SUITE.erl
+++ b/lib/public_key/test/public_key_SUITE.erl
@@ -776,7 +776,25 @@ pkix_path_validation(Config) when is_list(Config) ->
 
     {error, custom_reason} =
         public_key:pkix_path_validation(selfsigned_peer, [Trusted], [{verify_fun,
-                                                                      VerifyFunAndState2}]).
+                                                                      VerifyFunAndState2}]),
+    % check RSASSA-PSS key
+    % RsaPssKey = {public_key:generate_key({rsa, 1024, 65537}), pss_params(sha256)},
+    RsaPssKey = {hardcode_rsa_key(1), pss_params(sha256)},
+
+    CaKPSS = {TrustedPSSCert,_} = erl_make_certs:make_cert([{key, RsaPssKey},
+                 {subject, [
+                    {name, "RSASSA-PSS Public Key"},
+                    {?'id-at-name', {printableString, "public_key"}},
+                    {?'id-at-pseudonym', {printableString, "pubkey"}},
+                    {city, "Stockholm"},
+                    {country, "SE"},
+                    {org, "erlang"},
+                    {org_unit, "testing dep"}
+                       ]}
+                ]),
+    ChainPSSCert = {CertPSS, _} = erl_make_certs:make_cert([{issuer, {TrustedPSSCert,RsaPssKey}}]),
+    {ok, _} = public_key:pkix_path_validation(TrustedPSSCert, [CertPSS], []).
+
 pkix_path_validation_root_expired() ->
     [{doc, "Test root expiration so that it does not fall between chairs"}].
 pkix_path_validation_root_expired(Config) when is_list(Config) ->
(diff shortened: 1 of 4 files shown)


sigsergv:
Proposed fix for issue #5054

IngelaAndin:
Thanks, could you base it on maint instead of master? As it is a bug fix it would be nice to have it in the OTP-24 track. It looks like it is not breaking any existing test cases,  which I think is expected. Although the existing test cases all use a salt length that is also compliant with the old implementation, so please consider adding a test case using a different salt length.

sigsergv:
BTW, looks like test `rsa_pss_sign_verify` in `public_key_SUITE.erl` does nothing, there is no actual check:

```
public_key:pkix_verify(Cert, {#'RSAPublicKey'{modulus=Mod, publicExponent=Exp}, Parms}).
```

It should be:

```
true = public_key:pkix_verify(Cert, {#'RSAPublicKey'{modulus=Mod, publicExponent=Exp}, Parms}).
```

And it fails then because this certificate is created incorrectly, it's an EC certificate, not rsa-pss. That is not directly related to this PR or the original bug, so I created a separate PR: https://github.com/erlang/otp/pull/5064

IngelaAndin:
Good catch!  This PR  does not fail our TLS tests that use correct RSA-PSS certs :) But of course, we want the public_key test to be correct too. It would still be nice if you could rebase both your PRs so that they are based on the maint branch and will make it for OTP-24.1 and not on the master branch that will become OTP-25 (much later). Of course, maint will be merged to master so that it is fixed in both tracks.

sigsergv:
I made #5064 based on main branch. And still working on #5057.

IngelaAndin:
Ok, thank you!

sigsergv:
Added additional check into public_key_SUITE:pkix_path_validation/1

Also found and fixed the same issue in public_key:pkix_verify

Commits are rebased on maint.

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -1355,6 +1373,6 @@ pss_params(sha256) ->
        maskGenAlgorithm = #'MaskGenAlgorithm'{algorithm = ?'id-mgf1',
                                               parameters = #'HashAlgorithm'{algorithm = ?'id-sha256'}
                                              },
-       saltLength = 32,
+       saltLength = 20,
```
IngelaAndin, line 1376: It is recommended that the MGF hash function be the same as the scheme hash algorithm/function and that the salt length is hLen, the length of the output of the hash function.

sigsergv:
Yup, forgot to restore it after playing with non-standard parameters.

Fixed.

IngelaAndin:
Added lates to the test cycle again

IngelaAndin:
Thank you for the PR.

PR #5059, opened 2021-07-17
'extprg' needs to be properly addressed

yaobao1993:
Without using "./extprg", an error occurs saying:
`sh: line 0: exec: extprg: not found`


RaimoNiskanen:

This must be a remnant of the times when it was more common than not to have "." in one's path.

PR #5060, opened 2021-07-20
corrected spelling

massivefermion:
"protocol not suppored" to "protocol not supported"


RaimoNiskanen:
Thank you for the correction!

PR #5061, opened 2021-07-20
inet_res validate reply OTP-17323

RaimoNiskanen:
This PR rewrites how the DNS resolver `inet_res` handles the reply from the DNS server.
* More fields in the DNS reply are verified before accepting it.
* The reply's Resource Record is verified to be a reply to our query.
* Character case insensitive domain name comparison has been optimized.
* Database table key has been changed to directly allow case insensitive lookup.
* CNAME lookup has been rewritten to ensure not mixing new and cached values.
* Handling of IPv4-compatible IPv6 PTR lookups has been tightened.
* TTL = 0 records are no longer cached, according to standards.

These changes may have performance impacts, in either direction, and are rather extensive, so community feedback is much appreciated to iron out bugs before merging into the master branch for OTP-25.0.

@josevalim @sneako might find this interesting...


RaimoNiskanen:
Have anyone tested this rewrite in some kind of demanding environment?

sneako:
Sorry @RaimoNiskanen I have had my hands full, but I should be able to try it out sometime next week for you. Really appreciate your work here!

RaimoNiskanen:
Great @sneako!

sneako:
Hi @RaimoNiskanen, I'm sorry but a few things came up at work that I had to handle first. I am wrapping that up today, and I'll try to get to this next week.

sneako:
Good morning @RaimoNiskanen !

I have just begun testing this. Currently I have most of my nodes running OTP 24.0.6, and one node running [the same OTP version plus this patch](https://github.com/sneako/otp/tree/OTP-24.0.6-inet_validate).

I slowly ramped up the traffic levels until all nodes were receiving similar amounts of traffic. Each of these nodes is sending out ~11k HTTP requests per second across a dozen or so different hosts.

So far the nodes have been running at this level of traffic for ~45min and I have not seen any errors or strange changes in my application metrics, everything is looking good!

I will leave this node with the patched OTP version running for a few hours and report back.

In the meantime, if you have anything specific you would like me to check, please let me know.

sneako:
This has been running now for a over 4 hours. Traffic has increased to >17k outgoing HTTP requests per second per node. All of my application metrics are consistent across the two version and look good. There have been 0 errors during this time. As far as I am concerned this PR is good :+1: 

I will end the test now.

Thanks for the great work!

RaimoNiskanen:
Superb!  I wanted to get some verification that the increased result checking did not affect performance too much, and this is a good indicator!

Thank you for testing!

PR #5062, opened 2021-07-20
Fix distribution handshake challenge bug

jovdipp:
When Multiple nodes would try to connect to each other, after having
each used erlang:set_cookie( OtherNode, OtherCookie ) the dist_util
handshake would fail.

This was in part due to auth:get_cookie(OtherNode) default behaviour
to return it's own Nodes cookie when a OtherNode was not in the cookie
ETS table.

The behaviour was as follows :

- NodeA uses NodeB's Cookie to establish connection:

  - NodeA challenge used NodeB's cookie, NodeB attempts to use NodeA's
    cookie but gets it's own cookie instead ( as NodeB did not use
    set_cookie for NodeA ), all cookies match in challenges successfully.

- NodeA & NodeB try to each establish connection to the other:

  - NodeA having set_cookie( NodeB, NodeBCookie )

  - NodeB having set_cookie( NodeA, NodeACookie )

  - Either Node attempts connection, i.e :

    - NodeA challenge used NodeB's cookie, NodeB used NodeA's cookie
      leading to mismatched challenges, failing to match.

The bug arose in dist_util.erl from mis-use/set of 'MyCookie' &
'OtherCookie'.

The behaviour now appropriately uses & sets the 'OtherCookie' to the Other Node
and 'MyCookie' to it's own node using erlang:get_cookie().

### DIFF
--- a/lib/kernel/src/dist_util.erl
+++ b/lib/kernel/src/dist_util.erl
@@ -28,10 +28,10 @@
 %%-compile(export_all).
 -export([handshake_we_started/1, handshake_other_started/1,
          strict_order_flags/0, rejectable_flags/0,
-	 start_timer/1, setup_timer/2, 
-	 reset_timer/1, cancel_timer/1,
+         start_timer/1, setup_timer/2,
+         reset_timer/1, cancel_timer/1,
          is_node_name/1, split_node/1, is_allowed/2,
-	 shutdown/3, shutdown/4]).
+         shutdown/3, shutdown/4]).
 
 -import(error_logger,[error_msg/2]).
 
@@ -45,38 +45,38 @@
 -endif.
 
 -define(to_port(FSend, Socket, Data),
-	case FSend(Socket, Data) of
-	    {error, closed} ->
-		self() ! {tcp_closed, Socket},
-	        {error, closed};
-	    R ->
-	        R
-        end).
+    case FSend(Socket, Data) of
+        {error, closed} ->
+            self() ! {tcp_closed, Socket},
+            {error, closed};
+        R ->
+            R
+    end).
 
 
 -define(int16(X), [((X) bsr 8) band 16#ff, (X) band 16#ff]).
 
--define(int32(X), 
-	[((X) bsr 24) band 16#ff, ((X) bsr 16) band 16#ff,
-	 ((X) bsr 8) band 16#ff, (X) band 16#ff]).
+-define(int32(X),
+    [((X) bsr 24) band 16#ff, ((X) bsr 16) band 16#ff,
+     ((X) bsr 8) band 16#ff, (X) band 16#ff]).
 
 -define(i16(X1,X0),
-        (?u16(X1,X0) - 
+    (?u16(X1,X0) -
          (if (X1) > 127 -> 16#10000; true -> 0 end))).
 
 -define(u16(X1,X0),
-        (((X1) bsl 8) bor (X0))).
+    (((X1) bsl 8) bor (X0))).
 
 -define(u32(X3,X2,X1,X0),
-        (((X3) bsl 24) bor ((X2) bsl 16) bor ((X1) bsl 8) bor (X0))).
+    (((X3) bsl 24) bor ((X2) bsl 16) bor ((X1) bsl 8) bor (X0))).
 
 -define(CREATION_UNKNOWN,0).
 
 -record(tick, {read = 0,
-	       write = 0,
-	       tick = 0,
-	       ticked = 0
-	       }).
+               write = 0,
+               tick = 0,
+               ticked = 0
+}).
 
 dflag2str(?DFLAG_PUBLISHED) ->
     "PUBLISHED";
@@ -145,10 +145,10 @@ publish_flag(hidden, _, _) ->
     0;
 publish_flag(_, _, OtherNode) when is_atom(OtherNode) ->
     case net_kernel:publish_on_node(OtherNode) of
-	true ->
-	    ?DFLAG_PUBLISHED;
-	_ ->
-	    0
+        true ->
+            ?DFLAG_PUBLISHED;
+        _ ->
+            0
     end.
 
 name_type(Flags) ->
@@ -161,12 +161,12 @@ name_type(Flags) ->
 
 %% Sync with dist.c
 -record(erts_dflags, {
-          default,      % flags erts prefers
-          mandatory,    % flags erts needs
-          addable,      % flags local dist implementation is allowed to add
-          rejectable,   % flags local dist implementation is allowed to reject
-          strict_order  % flags for features needing strict order delivery
-}).
+    default,      % flags erts prefers
+    mandatory,    % flags erts needs
+    addable,      % flags local dist implementation is allowed to add
+    rejectable,   % flags local dist implementation is allowed to reject
+    strict_order  % flags for features needing strict order delivery
+                     }).
 
 -spec strict_order_flags() -> integer().
 strict_order_flags() ->
@@ -227,8 +227,8 @@ handshake_other_started(#hs_data{request_type=ReqType,
                                HSData3#hs_data.other_flags),
     HSData4 = HSData3#hs_data{this_flags = ChosenFlags,
                               other_flags = ChosenFlags},
-    ChallengeB = recv_challenge_reply(HSData4, ChallengeA, MyCookie),
-    send_challenge_ack(HSData4, gen_digest(ChallengeB, HisCookie)),
+    { ChallengeB, ChallengeCookie } = recv_challenge_reply( HSData4, ChallengeA, MyCookie, HisCookie ),
+    send_challenge_ack(HSData4, gen_digest(ChallengeB, ChallengeCookie)),
     ?debug({dist_util, self(), accept_connection, Node}),
     connection(HSData4);
 
@@ -244,26 +244,26 @@ check_dflags(#hs_data{other_node = Node,
                       other_started = OtherStarted,
                       require_flags = RequiredFlags} = HSData,
              #erts_dflags{}=EDF) ->
-
+    
     Mandatory = (EDF#erts_dflags.mandatory bor RequiredFlags),
     Missing = check_mandatory(Mandatory, OtherFlags, []),
     case Missing of
         [] ->
             ok;
         _ ->
-	    case OtherStarted of
-		true ->
-		    send_status(HSData, not_allowed),
-		    Dir = "from",
-		    How = "rejected";
-	        _ ->
-		    Dir = "to",
-		    How = "aborted"
-	    end,
-	    error_msg("** ~w: Connection attempt ~s node ~w ~s "
-		      "since it cannot handle ~p."
-		      "**~n", [node(), Dir, Node, How, Missing]),
-	    ?shutdown2(Node, {check_dflags_failed, Missing})
+            case OtherStarted of
+                true ->
+                    send_status(HSData, not_allowed),
+                    Dir = "from",
+                    How = "rejected";
+                _ ->
+                    Dir = "to",
+                    How = "aborted"
+            end,
+            error_msg("** ~w: Connection attempt ~s node ~w ~s "
+                      "since it cannot handle ~p."
+                      "**~n", [node(), Dir, Node, How, Missing]),
+            ?shutdown2(Node, {check_dflags_failed, Missing})
     end.
 
 check_mandatory(0, _OtherFlags, Missing) ->
@@ -280,93 +280,93 @@ check_mandatory(Mandatory, OtherFlags, Missing) ->
                          Missing
                  end,
     check_mandatory(Left, OtherFlags, NewMissing).
-                    
+
 
 %% No nodedown will be sent if we fail before this process has
 %% succeeded to mark the node as pending.
 
 mark_pending(#hs_data{kernel_pid=Kernel,
-		      other_node=Node,
-		      this_node=MyNode}=HSData) ->
+                      other_node=Node,
+                      this_node=MyNode}=HSData) ->
     case do_mark_pending(Kernel, MyNode, Node,
-			 HSData#hs_data.other_flags) of
-	ok ->
-	    send_status(HSData, ok),
-	    reset_timer(HSData#hs_data.timer),
+                         HSData#hs_data.other_flags) of
+        ok ->
+            send_status(HSData, ok),
+            reset_timer(HSData#hs_data.timer),
             HSData;
-
+        
         {ok, DynNodeName, Creation} ->
             HSData1 = HSData#hs_data{other_node = DynNodeName,
                                      other_creation = Creation},
             send_status(HSData1, named),
-	    reset_timer(HSData1#hs_data.timer),
+            reset_timer(HSData1#hs_data.timer),
             HSData1;
-
-	ok_pending ->
-	    send_status(HSData, ok_simultaneous),
-	    reset_timer(HSData#hs_data.timer),
+        
+        ok_pending ->
+            send_status(HSData, ok_simultaneous),
+            reset_timer(HSData#hs_data.timer),
             HSData;
-
-	nok_pending ->
-	    send_status(HSData, nok),
-	    ?shutdown(Node);
-	    
-	up_pending ->
-	    %% Check if connection is still alive, no
-	    %% implies that the connection is no longer pending
-	    %% due to simultaneous connect
-	    do_alive(HSData),
-
-	    %% This can happen if the other node goes down,
-	    %% and goes up again and contact us before we have
-	    %% detected that the socket was closed. 
-	    wait_pending(Kernel),
-	    reset_timer(HSData#hs_data.timer),
+        
+        nok_pending ->
+            send_status(HSData, nok),
+            ?shutdown(Node);
+        
+        up_pending ->
+            %% Check if connection is still alive, no
+            %% implies that the connection is no longer pending
+            %% due to simultaneous connect
+            do_alive(HSData),
+            
+            %% This can happen if the other node goes down,
+            %% and goes up again and contact us before we have
+            %% detected that the socket was closed.
+            wait_pending(Kernel),
+            reset_timer(HSData#hs_data.timer),
             HSData;
-
-	already_pending ->
-	    %% FIXME: is this a case ?
-	    ?debug({dist_util,self(),mark_pending,already_pending,Node}),
-	    ?shutdown(Node)
+        
+        already_pending ->
+            %% FIXME: is this a case ?
+            ?debug({dist_util,self(),mark_pending,already_pending,Node}),
+            ?shutdown(Node)
     end.
 
 
 %%
-%% Marking pending and negotiating away 
+%% Marking pending and negotiating away
 %% simultaneous connection problems
 %%
 
 wait_pending(Kernel) ->
     receive
-	{Kernel, pending} ->
-	    ?trace("wait_pending returned for pid ~p.~n", 
-		   [self()]),
-	    ok
+        {Kernel, pending} ->
+            ?trace("wait_pending returned for pid ~p.~n",
+                   [self()]),
+            ok
     end.
 
 do_alive(#hs_data{other_node = Node} = HSData) ->
     send_status(HSData, alive),
     case recv_status_reply(HSData) of
-	true  -> true;
-	false -> ?shutdown(Node)
+        true  -> true;
+        false -> ?shutdown(Node)
     end.
-    
+
 do_mark_pending(Kernel, MyNode, Node, Flags) ->
     Kernel ! {self(), {accept_pending,MyNode,Node,
-		       publish_type(Flags)}},
+                       publish_type(Flags)}},
     receive
-	{Kernel,{accept_pending,Ret}} ->
-	    ?trace("do_mark_pending(~p,~p,~p,~p) -> ~p~n",
-		   [Kernel, MyNode, Node, Flags, Ret]),
-	    Ret
+        {Kernel,{accept_pending,Ret}} ->
+            ?trace("do_mark_pending(~p,~p,~p,~p) -> ~p~n",
+                   [Kernel, MyNode, Node, Flags, Ret]),
+            Ret
     end.
 
 is_pending(Kernel, Node) ->
     Kernel ! {self(), {is_pending, Node}},
     receive
-	{Kernel, {is_pending, Reply}} -> Reply
+        {Kernel, {is_pending, Reply}} -> Reply
     end.
-    
+
 %%
 %% This will tell the net_kernel about the nodedown as it
 %% recognizes the exit signal.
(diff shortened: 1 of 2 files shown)


RaimoNiskanen:
I think this is a bug, but that you have the wrong solution to the problem, and that the code in `dist_util.c` is mostly right.

Your suggestion changes the protocol in that it uses a different cookie in the challenge ack than was supposed to be used before.  This change probably botches connections with; erlang nodes of older version, C nodes, Java nodes.
I think the actual bug is that `dist_util:get_cookies(Node)` does not return `{MyCookie, OtherCookie}` as variable names suggest.

Oddly enough that code has been the same since we entered Git (R13B03)...

I have tried that change, and came up with this diff:
``` erlang
diff --git a/lib/kernel/src/dist_util.erl b/lib/kernel/src/dist_util.erl
index d8a6371d9c..7af9d51757 100644
--- a/lib/kernel/src/dist_util.erl
+++ b/lib/kernel/src/dist_util.erl
@@ -525,14 +527,16 @@ gen_challenge() ->
 %% Get the cookies for a node from auth
 %%    
 get_cookies(Node) ->
-    case auth:get_cookie(Node) of
-       X when is_atom(X) ->
-           {X,X}
-%      {Y,Z} when is_atom(Y), is_atom(Z) ->
-%          {Y,Z};
-%      _ ->
-%          erlang:error("Corrupt cookie database")
-    end.    
+    case {auth:get_cookie(),auth:get_cookie(Node)} of
+        {nocookie,Z} when is_atom(Z) ->
+            %% This just hides a problem.  The right way
+            %% should be to have auth: reply with the own
+            %% cookie for a dynamic node, not to use the
+            %% other nodes' cookie.
+            {Z, Z};
+       {Y,Z} = Cookies when is_atom(Y), is_atom(Z) ->
+           Cookies
+    end.
 
 %% No error return; either succeeds or terminates the process.
 do_setnode(#hs_data{other_node = Node, socket = Socket, 
```
Unfortunately I do not believe in this code, as you see in the comment,
although works for a hand test with a nodeA and nodeB with different cookies.

I think that `auth:` should be augmented with a function `get_cookies(Node)` that does the right thing,
also for dynamically named nodes (`nonode@nohost`, then `net_kernel:connect_node/1`), see
`erl_distribution_SUITE:dyn_node_name/1`.

I would like to wait for the people that implemented dynamically named nodes to come back from their
vacations to figure out exactly what is the right solution for this problem.


Regarding your test case `simultaneous_remote_linking/1`;   I think the start of it does not work since for a connection to work between the test node and nodeA/nodeB, the latters would need to have the test node cookie configured, and I have not figured out yet how to accomplish that.

jovdipp:
Have you run the test case ?

NodeA & NodeB do not need to know the test node's cookie.

The reason the test case runs successfully, is that test node starts test_server peer nodes. It then tests the concept of one node ( Test node ) connecting another node using other node's cookie ( NodeA & NodeB respectively ), whereby the another nodes ( NodeA & NodeB ) do not know the connecting ( Test node ) node's cookie.

Having established the above, the test case tests one node ( NodeA ) connecting using other node's cookie ( NodeB ) where the other node ( NodeB ) does know the connecting ( NodeA ) node's cookie. Lastly ( mostly redundantly ) NodeB pings NodeA.

jovdipp:
Regarding the challenge ack, their are two use cases that make sense in connecting nodes :
1. NodeA ( initiator node ) knows NodeB's cookie and NodeB ( the receiving node ) does not know NodeA's cookie. lets call this OIK ( only initiator knows )
2. NodeA ( initiator node ) knows NodeB's cookie and NodeB ( the receiving node ) knows NodeA's cookie. lets call this 
BNK ( both nodes know )

In the OIK case ( which is prob the most common ) NodeB ( the receiver ) does not know the cookie of NodeB ( the initiator ) to use in generating the digest to send a challenge ack, hence why it would only ever handle challenges with it's own cookie. 

The onus is upon NodeA ( the initiator ) to already know NodeB's cookie hence why the two can match challenges to prove to NodeB that NodeA knows the NodeB's cookie without plainly sending it the handshake. 

BNK use case works in the same manner, as it does not matter whether NodeB ( the receiver ) knows NodeA's cookie.

jovdipp:
Regarding botching connections to "erlang nodes of older version, C nodes, Java nodes" could you expand on why you think that ? is there a test to reveal that

RaimoNiskanen:
I am arguing that it may well be so that the code in `dist_util.erl` indicates that for a connection to succede, both nodes need to know the other node's cookie, and this is supposed to be enforced in the challenge exchange.  Your test case may be based on incorrect assumptions.

That this seems to have been broken since R13B01 does not make the property invalid.  Other implementations that have used the source in `dist_util.erl` as reference, such as maybe `ei`, `erl_interface` and `jinterface`, or other external implementations, may successfully implement such a challenge exchange.

We need to figure out how this is supposed to work.  C nodes, dynamic nodes, hidden nodes, etc. are all newer creatures than use the Erlang distribution protocol, and they should not misuse it.  Therefore we need to have all people that have worked on these features back from vacation, to be able to sort this out.

jovdipp:
Seems jinterfaces' AbstractConnection.java only uses it's localNode.cookie() in the challenge digest. This is as per the documentation which states 

> Connections are only permitted by nodes using the same security cookie

So jinterface should work with the pull request bug fix.

RaimoNiskanen:
I have created an alternative pull request, based on this, with an alternative fix (that I believe to be correct) and slightly modified test case. PR #5082

jovdipp:
I think my solution is valid & tested, it would be better to extend this PR test case with any concerns that either prove or disprove the remedy proposed. Creating a new PR for your own solution seems odd. I don't agree thus far with your reasoning for your proposed fix as a substitute for this PR. 
I do agree with testing of dynamic nodes, hidden nodes and so forth with consultation of the team you stated is on leave would be wise. I also think this PR would pass the extending testing.

RaimoNiskanen:
Well, the reason I created a different PR for my solution is so both solutions could be easily compared side by side.  Both PR:s solve the same problem but in fundamentally different ways.

I do not doubt that your solution is tested and works in existing tests and use cases, because my suspicion is that this code rotted 20 years ago, and since then nobody has bothered to investigate why Erlang Distribution with different cookies does not work.  Therefore the only thing that is really tested and used today is having the same cookie for all nodes in a cluster; and the current code, your solution, my solution, `erl_interface`, `jinterface` and whatnot all work well for that case.

##### Concerning extending this PR:s test case
Your proposed test case `simultaneous_remote_linking/1` begins with starting `nodeA@host` with cookie `nodeAcookie` and `nodeB@host` with cookie `nodeBcookie`.  Then it, in node `test_server@host`; calls `erlang:set_cookie(nodeA@host, nodeAcookie)` and connects to the node with `net_adm:ping(nodeA@host)`.  The comment says: "`This should work as only the Test node has used set_cookie ETS`".

On the contrary I say that this should **not** work since `nodeA@host` does not have the correct configuration of the cookie for node `test_server@host`, so that connection should fail (and there should be a test case that verifies that it fails).  `nodeA@host` has to call `erlang:set_cookie(test_server@host, TestServerCookie)` before the connection should work, and before that it has to get `TestServerCookie` somehow.  I can write such a test case, but I do not think you will agree that it is a valid test, since we apparently do not agree on the fundamentals of how the distribution handshake challenge should work.

##### What now?
Therefore I think we should now wait for all stakeholders (I hate using business lingo!) to be back from their vacation, and then have a discussion comparing the two different solutions in these PR:s...

jovdipp:
@RaimoNiskanen thank you kindly for a well balanced & respectful explanation of your thinking.

I too am surprised at the handshake bug existing for so many years, it caused many problems with systems I designed based on what I thought was a foundation and could never have been an issue.

I believe I have seen the bug indirectly appearing  in forums over the many years, perhaps with the inevitable "solution" of setting the cookie on all nodes to be the same.

The reason the bug is tricky is not that nodes with different cookie could not connect but simply not connect if both nodes knew each others respective cookies before an initial connection attempt ( and/or perhaps re-connects ).

Allow me to get on the same page as you with regards to the testing.

> ...Therefore the only thing that is really tested and used today is having the same cookie for all nodes in a cluster...

Re the above :  the use cases that are being used today are not only nodes with the same cookie but also nodes with different cookies. Nodes with different cookies connect fine as long as only one of the nodes
initiate the connection to a receiving node that does not already "know" the connecting nodes cookie.

So there are 3 use cases for connecting nodes :
 1. Nodes with the same cookies ( Works without this PR ) 👍 
 2. Nodes with different cookies, splits into two use cases :
     2.1. Connecting a node to a receiver node, where the receiving node has NOT set the connecting nodes cookie.  ( Works without this PR ) 👍 
     2.2. Connecting a node to a receiver node, where the receiving node set the connecting nodes cookie. ( Doesn't work without this PR ) 👎 

> Your proposed test case `simultaneous_remote_linking/1` begins with starting `nodeA@host` with cookie `nodeAcookie` and `nodeB@host` with cookie `nodeBcookie`. Then it, in node `test_server@host`; calls `erlang:set_cookie(nodeA@host, nodeAcookie)` and connects to the node with `net_adm:ping(nodeA@host)`. The comment says: "`This should work as only the Test node has used set_cookie ETS`".

Re the above : Yes, correct as this tests use case 2.1. Which demonstrates existing functionality of connecting node for over 20 years still works with this PR.

> On the contrary I say that this should **not** work since `nodeA@host` does not have the correct configuration of the cookie for node `test_server@host`, so that connection should fail (and there should be a test case that verifies that it fails). `nodeA@host` has to call `erlang:set_cookie(test_server@host, TestServerCookie)` before the connection should work, and before that it has to get `TestServerCookie` somehow.

Re the above : The fail aspect to test in the test case, comes straight after by connecting NodeA to NodeB. This is what tests 2.2 which without the fix in PR would fail. This PR has the fix and hence it passes.

Further more stating "...this should **not** work" would be in direct contention with use case 2.1.
As well as the implication of your statement would mean you intend/assume that nodes only allow connecting to each other if they both are initially set up to know about each other beforehand. That would mean a new node started in future would never be able to join others as they would not have known of the future node.
That would fundamentally break the last 20 or so years of nodes connecting.

> I can write such a test case, but I do not think you will agree that it is a valid test, since we apparently do not agree on the fundamentals of how the distribution handshake challenge should work.

Re the above : Hopefully what is stated in the above "Re..." clears up the need for a different test case.

A further explanation :  
- extending the test case to explicitly test use case 1, would be a good idea.
- my fundamental understanding of the handshake and its original intent is guided by :  
       the key concept that to implement the use cases in order stated above, that use case 2.1 would never be 
       able to work as the RECEIVING node could NEVER know THE connecting nodes COOKIE. Hence the idea of 
       the connecting node knowing some secret information ( receiving nodes cookie ) and receiving node 
       knowing the same secret information ( it's own cookie ) allows as unique successful handshake.
       Hence any intent on requiring the connecting nodes' cookie is invalid, as illustrated by the actual execution
       of the code never using the connecting nodes cookie. The get_cookie(Node) function mislead the  
       implementor from seeing this, as the default behaviour returns it's own cookie if it does not know the 
       answer to the question "What is the other Nodes cookie".

So you can see that in this PR changing the variable names and using the correct "secret" as should be
intended. This is proved in the fact that the PR test case passes for existing connection use cases and the 
previously failing use case.

RaimoNiskanen:
Thank you, very enlightening!

So, since about 20 years back we have a bug, that because we have lived with it for so long, maybe we should call it "unintended behaviour" instead...

Nevertheless, today, before any of our PR:s, we have a situation where the `dist_util` code uses misleading variable names OurCookie/OtherCookie (MyCookie/HisCookie) since only one cookie is used during the handshake.

I checked, and it seems to be working better than i thought...  If I start nodeB with cookie `cB` and nodeA with cookie `cA`, on nodeA call `erlang:set_cookie(NodeB, cB)`; then connections work in both directions between nodeA and nodeB.  The cookie `cB` is used both for connections nodeA -> nodeB as well as nodeB -> nodeA.

This might be viewed as that the link between nodeA and nodeB, or that the node pair (nodeA, nodeB); uses cookie `cB`.

The configuration feels asymmetrical, but is that an actual problem?  This is a possible point of view:
Both nodes has a default cookie for their default cluster, and they might have exceptions for the _links / entry points_ to other cluster(s).  So one has to decide if the link between nodeA and nodeB should belong to cluster A or cluster B.  The configuration above makes the link belong to cluster B, and then it would be an error to on nodeB call `erlang:set_cookie(NodeA, cA)` because then the nodes would not agree about which cluster the link belongs to.

If nodeA should need to participate fully in cluster B then it needs to have cookie `cB` set for all possible node names in the cluster, or else it can not set up connections to all nodes in the cluster as `global` requires.  That would be awkward or impossible.  Therefore this case is more suitable for e.g hidden nodes that dodges `global`.

That is the situation since 20+ years, and we might just call it a documentation weakness, and misleading code in `dist_util`, that both should be fixed.

Fixing it the way I have suggested would break existing use cases by requiring that both nodes of a link has the other node's cookie configured before the connection is established.  I added the possibility to specify other nodes' cookies on the command line which makes this solution more usable, but it is still backwards incompatible. It might be closer to the intentions before the 20+ years old "unintended behaviour", as the variable names in `dist_util` indicates.  It would also not need the idea that a link must belong to one cluster.

(Having the possibility to specify other nodes' cookies on the command line might be a useful feature in itself, though)

Fixing it the way this PR suggests would actually also break existing working configurations.  The one I described above, for instance.  Such an asymmetrical configuration works now, in both directions, but with this PR it would be required that nodeB configured nodeA's cookie, otherwise connections nodeA -> nodeB would work, but not nodeB -> nodeA, because then the cookie of the receiving node would always be used.  This would be a very unfortunate situation with asymmetrical connectivity, in my opinion.

The bottom line: I am starting to think that we maybe instead should declare the current behaviour to be intended, document it, (posibly add the feature to specify other nodes' cookies on the command line, at least to facilitate testing, ) and rework the misleading variable names and functions in `dist_util`.

jovdipp:
> Thank you, very enlightening!

> So, since about 20 years back we have a bug, that because we have lived with it for so long, maybe we should call it "unintended behaviour" instead...

Re the above : hmmm, okay.. doing that might just mean there is no such thing as a bug in software ;) Unless we define a bug life expectancy.

> Nevertheless, today, before any of our PR:s, we have a situation where the `dist_util` code uses misleading variable names OurCookie/OtherCookie (MyCookie/HisCookie) since only one cookie is used during the handshake.

Re the above : yes, the receiving nodes' cookie. It's not just misleading variable names as the use of the connecting nodes cookie by the receiving node in the challenge, is a bug which breaks what would otherwise be a successful connection handshake. 

> I checked, and it seems to be working better than i thought... If I start nodeB with cookie `cB` and nodeA with cookie `cA`, on nodeA call `erlang:set_cookie(NodeB, cB)`; then connections work in both directions between nodeA and nodeB. The cookie `cB` is used both for connections nodeA -> nodeB as well as nodeB -> nodeA.

Re the above :  The issue is based on initialisation of connections ( where the handshake takes place ) not the on going link after a successful connection. What you have described is use case 2.1.
Connection initiated by nodeB wouldn't work unless given `cA` by calling `erlang:set_cookie(NodeA, cA)`. Which doesn't work If 'set_cookie' is done on both nodes before the connection initialisation. ( Hence the fix in this PR )

> This might be viewed as that the link between nodeA and nodeB, or that the node pair (nodeA, nodeB); uses cookie `cB`.

Re the above :  only on initial connection handshake when nodeB is the receiving node and does not have nodeA's cookie set.

> The configuration feels asymmetrical, but is that an actual problem? 

Re the above :  yes, as I have a large scale micro services architecture, which provides different services ( authentication, analytics, notifications... so on ). I need to connect some nodes to some services without needing to connect to others. As a well as foundational apps in all services which ensure certain services ( nodes/clusters ) maintain connection to each other by automatically retrying to connect to each if a connection is lost ( node down, network split.. so on ). So nodeB runs a service for which nodeA connects as well as nodeC, nodeC runs a service for which nodeB needs to connect.

I really want to avoid detailing my specific use cases as it distracts from the simplest point that nodes with different cookie can connect to each other, this is a feature in erlang. NodeA can initiate a connection to nodeB with a different cookie or nodeB can initiate a connection to nodeA. Both are perfectly fine and have been used by us all ( the community ) for decades.
The bug which prevents either node successfully initiating a connection as above, if they both simultaneously know each others cookie before hand cannot be an intended behaviour. It is very simply a bug which this PR fix's without backward compatibility issues or inventing new behaviour.

> That is the situation since 20+ years, and we might just call it a documentation weakness, and misleading code in `dist_util`, that both should be fixed.

Re the above :  it's a bug not a documentation weakness.

> Fixing it the way I have suggested would break existing use cases by requiring that both nodes of a link has the other node's cookie configured before the connection is established. I added the possibility to specify other nodes' cookies on the command line which makes this solution more usable, but it is still backwards incompatible.
> 

Re the above : correct. 

> Fixing it the way this PR suggests would actually also break existing working configurations. The one I described above, for instance. Such an asymmetrical configuration works now, in both directions, but with this PR it would be required that nodeB configured nodeA's cookie, otherwise connections nodeA -> nodeB would work, but not nodeB -> nodeA, because then the cookie of the receiving node would always be used. This would be a very unfortunate situation with asymmetrical connectivity, in my opinion.

Re the above : This PR does work in existing configurations. The above assumptions you're making are incorrect. As stated before and seen in the test case, the test node connects to both nodeA & nodeB without nodeA or nodeB setting the test nodes cookie. This PR does not require the receiving node to know the connecting nodes cookie.

> The bottom line: I am starting to think that we maybe instead should declare the current behaviour to be intended, document it, (posibly add the feature to specify other nodes' cookies on the command line, at least to facilitate testing, ) and rework the misleading variable names and functions in `dist_util`.

Re the above : I don't agree, base on all the points above.

This PR works and is tested for existing functionality of connecting node as well as fixing the use case 2.1.

RaimoNiskanen:
We have to agree on some facts here...

Node A and node B have different default cookies.
* Your use case 2.1: node A sets node B's cookie, then connects to node B; works today.
* Use case 2.1.b: node A sets node B's cookie then _node B_ connects to _node A_; works today.

Do we agree so far?

* Your use case 2.2: node A sets node B's cookie, node B sets node A's cookie, then either node connects to the other node; neither does not work today.

Your proposed change fixes use case 2.2, but makes use case 2.1.b not work, which is a backwards incompatible change.

Do we still agree?

Both you and I have assumed that 2.2 is a valid use case, and the correct configuration, but why should that be the correct configuration?  You have assumed that only the receiving node's cookie should be used in the handshake, and I have assumed that both nodes' cookies should be used.  Neither of those assumptions have to be correct.

Why cannot the correct configuration be that node A sets node B's cookie and that node B should not set node A's cookie?  Because then connections in both directions between node A and node B works, today?  What is the problem with the current behaviour and such a configuration requirement?  This, I presume, has worked the last 20+ years. (I just verified it with OTP 20.2)

RaimoNiskanen:
Furthermore, this will have to go through the OTP Technical Board.

So Then we would need to know more about the use case for connecting a network with different cookies, since another "solution" would be to remove support for that feature.

Why is a network with more than one distinct cookie a useful feature?

jovdipp:
@RaimoNiskanen thanks again for a detailed reply.

I tested the use case you presented as 2.1.b, in which a node connection handshake is successful even when the receiving node knows the initiating nodes' cookie. I believe we are the same page now regarding this use case. Thank you once again for taking the time to elaborate further. 

I see the use case as perhaps an unintended/unexpected 'for list', similar to 'spam mechanisms' in which contacts existing in a receivers 'contact book' are given trust. The validity of the behaviour needs no debate, as its existed for so long it needs to be backward compatible. ( I believe this in line with what you have said )

To resolve the bug of both nodes having the other nodes' in its 'allowed list', as well ensuring backward compatibility. I have updated dist_util to check challenges for both cookies as its impossible for each node to know which cookie was used to generate the handshake.

The tests have been updated as well to better reflect all the discussed use cases.

Please have look at the update, for your insights.

RaimoNiskanen:
I am pleased to have brought some clarity.  This anomaly is a great finding by you!

I believe that the current behaviour more or less "has happened", probably in a distribution protocol rewrite for OTP R6B during the previous millennium, but we need to be backwards compatible, agreed.

The view that a node allows connection from nodes "in its contact list" is rather nice.  I tried to sell another view, that is; a node that knows another nodes cookie can act as a member of that cluster when communicating with that node.
A cookie defines a cluster. nodeA is a member of cluster1 (has default cookie cluster1).  nodeB is a member of cluster2.  We then, in nodeA, set cookie cluster2 for nodeB.  Now nodeA acts like a member of cluster2 when communicating with nodeB.

I have some comments on your new code suggestion, but first; some more fundamental questions:
1) What are the use cases for running a network with different cookies (clusters)?  Yes, it has worked for 20+ years, but if we removed the feature totally and only allowed one cookie, one cluster, for a node; how bad would that be?  Is this feature essential and if so; why?  To remove this feature _is_ an option that the OTP Technical Board will consider so _why_ would that be bad?
2) Why is the current behaviour to be considered incorrect?  nodeA acts as a member of cluster2 when communicating with nodeB.  Then you can obviously not configure nodeB to act as a member of cluster1 when it communicates with nodeA.  The configuration of both nodes has to be consistent on which cluster they belong to when communicating!  Right?!  Is this a bad way of reasoning and if so; why?

Your current code suggestion tries both known cookies in the handshake.  That is a backwards compatible solution to the problem, _if_ it is to be considered a problem...

##### Code comments
It would be a wee bit easier to evaluate the code if there were no just-white-space changes.  But more importantly; unneeded white space changes increase the risk of future bogus merge conflicts.  Please avoid them.

In my counter-PR I added a function `auth:get_cookies(Node)` and used it instead of `dist_util:get_cookies(Node)` that does two calls; `erlang:get_cookie(), auth:get_cookie(Node`.  The change avoids code race and improves efficiency.  I think it is worth to consider:
https://github.com/erlang/otp/pull/5082/commits/b860e7a2c9578dc92c61c5924297c562290556f4#diff-92ddd47d70acc66fc6e03bc74961bcef420e8c82cf9fbc1cbdb16f4839e60e9c File:auth.erl

When validating the sum and its alternative the code calculates both sums and then matches on both.  I think the code should be reworked to regard one of the sums as the preferred one and first prepare the received sum, then calculate the preferred sum and compare, if incorrect then calculate the other sum and compare.  This to avoid always calculating one sum in vain.  The code could also check if both cookies are the same before calculating the other sum; it would optimize a bad (hopefully rare) case, but it might be nice to not do unnecessary work even for incorrect configurations.

##### Request for code comment
In my counter-PR I have a commit that, in addition to the existing `-setcookie OurCookie`, adds the node start command line argument: `-setcookie Node NodeCookie` that can be used multiple times, so you can start a node with a complete cookie configuration from the inception.  I think it would be a good feature at least for testing, but what do you think?  Would it be good for your use case?
https://github.com/erlang/otp/commit/2eaeac3fbadeb89645811bd4b30a3443e226a7bd#diff-9b388e03a9038a603ba9d7c1d1cb04d38bbdd341f4ef7b4a8452306d8fe8fb79

PR #5064, opened 2021-07-21
Fixing test public_key_SUITE:rsa_pss_sign_verify()

sigsergv:
Original test behaves incorrectly: it doesn't check result and testing certificate created for this test is not a proper rsa-pss one.


IngelaAndin:
Thank you for fixing this test case.

PR #5065, opened 2021-07-21
xmerl: Fix the XPath normalize-space() function

axelf4:
This pull request fixes two issues with `normalize-space()` where:
1. Evaluation of the function call caused the `badarith` error to be thrown; and
1. The call to `xmerl_scan:accumulate_whitespace/4` in the implementation of `normalize/1` passed the argument `preserve` instead of the correct atom `normalize`.

### DIFF
--- a/lib/xmerl/src/xmerl_xpath_pred.erl
+++ b/lib/xmerl/src/xmerl_xpath_pred.erl
@@ -758,8 +758,8 @@ mk_boolean(C, Expr) ->
 normalize([H|T]) when ?whitespace(H) ->
     normalize(T);
 normalize(Str) ->
-    ContF = fun(_ContF, RetF, _S) ->
-		    RetF()
+    ContF = fun(_ContF, RetF, S) ->
+		    RetF(S)
 	    end,
     normalize(Str,
 	      #xmerl_scanner{acc_fun = fun() -> exit(acc_fun) end,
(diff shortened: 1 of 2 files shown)


otp-anders:
This and much more is fixed on a branch I've been working on for a while to fix all the problems and limitations in the current XPath support. Not sure what release this will be in yet, but you can see the branch here for now:

https://github.com/erlang/otp/compare/maint...otp-anders:anders/xmerl/xpath/OTP-17333

There's usable (hopefully) documentation now, but the formatting of type specs produced by edoc in the OTP build is fairly unreadable, and if you build with OTP 24 then a number of non-public specs also leak into the documentation as well; two more problems yet to fix.

If anyone is willing to give the branch a whirl then I'd be interested in any feedback. It should be mostly backwards compatible except where the old implementation simply got the evaluation wrong, and the many cases in which the old implementation simply crashed should now be fixed. The only known problem I can think of offhand is that the sign of zero in IEEE 754 arithmetic can be lost in OTP 23 (but not 24).

Anders, Erlang/OTP

otp-anders:
One more limitation: the core library function id() doesn't really do what the XPath 1.0 specification says it should.

The spec says to search for attributes whose type is declared as ID in the DTD in question, but the old implementation instead searched for attributes whose name was id, which is something else. However, there's no easy way to access attribute declarations in the case that there's a DTD, so the new implementation still searches for attributes named id.

axelf4:
Neat, feel free to close this then.

otp-anders:
Thanks, closing.

otp-anders:
Just an update, but an updated xmerl_xpath won't make it into OTP 25.0 because of problems in xmerl_xsd that we want to fix first; both with its use of xmerl_xpath and quite a bit more. Hopefully it'll all be done for 25.1.

otp-anders:
25.1 obviously hasn't happened either since the problems in xmerl_xsd are
quite a bit bigger than I realized. Right now I'm not sure how we'll do it,
if we'll release a working XPath implementation beside the current one
until xsd is sorted out or something else.

PR #5069, opened 2021-07-23
Add dist_util_spawn_options in kernel

zzydxm:
This change allows user to specify spawn options when dist protocol
starts dist_util:handshake_other_started or handshake_we_started
processes.

These processes will be alive until the connections are down. With
thousands of ssl dist connections it spends quite amount of memory,
and it would be great if we can change garbage collection, as well
as other spawn options for dist_util processes.


rickard-green:
I think a better parameter name would be `net_ticker_spawn_options` or similar, since these are the processes enforcing the value of the `net_ticktime` kernel configuration parameter.

I'm not sure I think it is a good idea to allow the user to lower the priority. Perhaps it should be forced just like the `link` option, but as I said I'm not sure.

zzydxm:
Thanks for the comment! I've changed the option name and default option to force priority max.

PR #5070, opened 2021-07-23
Early draft: showing non-unique labels in observer

nathanl:
This PR is nowhere close to complete. I'm posting it in the hopes of getting guidance and help to carry it through.

In observer, we sometimes have a number of "generic" processes, like database connections in a pool.
Because they do not have registered names, they are labeled only with a pid, which is not informative.
The process architecture would be more understandable if we could use non-unique labels, like `database_connection`. 

I brought this up on the erlang-questions mailing list a while back (see "Support for non-unique process labels?") and people seemed to like the idea.

I am inexperienced with Erlang (I mostly use Elixir), and I was not at all familiar with the code behind observer, but I've stumbled far enough to get a tiny demo working:

non_unique_labels

Here, processes which do not have a registered name, but *do* have a `'$label'` entry in their dictionary, are labeled accordingly in observer.

There are probably numerous things wrong with my approach so far.
And manually adding a label to the process dictionary is not very discoverable - ideally we could specify this label when (for example) starting a `gen_server`.

Does this seem like a worthwhile effort? If so, any pointers about how to proceed?

### DIFF
--- a/lib/observer/src/observer_app_wx.erl
+++ b/lib/observer/src/observer_app_wx.erl
@@ -497,13 +497,21 @@ middle([{#box{y=Y0},_}|List], _) ->
 
 box(Str0, N2P, FontW) ->
     Pid = gb_trees:get(Str0, N2P),
-    Str = if hd(Str0) =:= $< -> lists:append(io_lib:format("~w", [Pid]));
+    %% This string looks like a pid
+    Str = if hd(Str0) =:= $< -> pid_or_label(Pid);
+       %% Probably a global name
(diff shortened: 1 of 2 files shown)


RaimoNiskanen:
Is this a duplicate of PR #5039?

nathanl:
@RaimoNiskanen yes it is! I was the one who suggested this and didn't see that someone else was working on it. Closing.

RaimoNiskanen:
Great!  Best solution - somebody else just do it...

PR #5075, opened 2021-07-27
Force atom format strings to strings

hauleth:
This is meant to prevent situations when user will fire messages using

    logger:notice(string, [$a]).

That would be accidentally treated the same as:

    logger:notice("a").

As earlier both produced message in form of `{string, "a"}`. Now it simply cast format strings to never be an atom. This isn't breaking change as message was defined to be `{io:format(), [term()]}`, which mean that both forms should be treated identically.


garazdawi:
Thanks!

Issue #5063, opened 2021-07-20
Distribution handshake error ( Connection attempt from node .. rejected. Invalid challenge reply )


jovdipp:
**Describe the bug**
Two nodes with different cookies trying to connect to each other after both nodes have previously 
use set_cookie(OtherNode, OtherCookie)

**To Reproduce**
1. Start two nodes with different cookies.
2. On each node call erlang:set_cookie( Node, Cookie ), replacing Node & Cookie with the Other Nodes details
3. attempt connecting the nodes from either node ( i.e net_adm:ping )
4. Error appears "Connection attempt from node .. rejected. Invalid challenge reply "

**Expected behavior**
I expect the nodes to connect successfully.

**Affected versions**
23.3.3 is what I have tested on

**Additional context**
I created a Pull request with Tests and Fix.
( https://github.com/erlang/otp/pull/5062 )

RaimoNiskanen:
I have created an alternative pull request, based on yours, with an alternative fix (that I believe to be correct) and slightly modified test case. PR #5082

jovdipp:
> I have created an alternative pull request, based on yours, with an alternative fix (that I believe to be correct) and slightly modified test case. PR #5082

@RaimoNiskanen to keep in sync for others finding the issue, with our comments in PR #5062. 
As you stated the additional PR #5082 created would not be backward compatible.

mikpe:
You told both nodes to use the other node's cookie when communicating with that other node. This creates an inconsistent view of which cookie to use for that connection, which causes the rejections.

The correct procedure is to tell _one_ node which cookie to use for the _other_ node, as [documented](https://erlang.org/doc/reference_manual/distributed.html#security).

I don't see a bug here.

jovdipp:
@mikpe perhaps read the conversation on the PR #5062 to better your understand the bug.

https://github.com/erlang/otp/pull/5062

Both @RaimoNiskanen and I discuss and breakdown the use cases for communicating to other nodes in detail.

Also reading the documentation link provided states for a node "...connect to, or accept a connection from..." to use set_cookie for the other node. 
it doesn't state this should exclusively or only be done by one node. Which is correct from a general understanding of intent but not correct in terms of the code implementation. 

Due to the allowing connection to or from one node to another, the mechanism shouldn't break simply because both nodes want to allow the other to connect at any time.

RaimoNiskanen:
[...]

PR #5082, opened 2021-07-30
Correct distribution handshake challenge cookie handling

RaimoNiskanen:
This is an alternative correction (to the pull request #5062) of issue #5063, that I believe to be correct.

It is based on PR #5062 and reverts the protocol change in `dist_util.erl` but keeps the test case.  The test case is modified in a later commit to use hidden slave nodes, which is a harder problem and a better test, hopefully.

The functional change is instead in `auth.erl` to return `{OurCookie,OtherCookie}` as appropriate for the connection, and a possibility to specify other nodes' cookies on the command line has been added.

This correction uses `OurCookie` and `OtherCookie` in the handshake, both for incoming and for outgoing connections. `OtherCookie` is used for generating the challenge message to the other party, and `OurCookie` when verifying incoming challenge message.  This way both sides' cookies are used in all connection setups, and both sides' cookies has to be configured correctly on both sides before connection setup.

The command line argument `-setcookie` is augmented to allow arity 2 arguments, so `-setcookie OurCookie` as before sets the starting nodes' cookie, and `-setcookie OtherNode OtherCookie` sets the cookie to be used for `OtherNode` using `erlang:set_cookie/2`.  This facilitates starting a node with usable cookies without having to run code before starting the distribution so utilities like `slave:start/*` works.

That a node needs to have the other nodes' cookie for both incoming and outgoing connections to succeed has been documented since OTP_R11B, but it seems to have been broken since at least OTP_R12B.  In OTP_R7B (the oldest release I have access to) it was possible to configure a cookie pair `{CookieIn,CookieOut}` for a specific node (Using `erlang:set_cookie/2`), although a comment in `erlang.erl` claims that that feature is discarded, indicating that our cookie should be `CookieIn` and the other node's cookie should be `CookieOut`.  Getting a such a cookie pair was disabled in OTP_R12B, maybe by accident, thinking that `OurCookie` and `OtherCookie` had replaced the cookie pair, despite that detail had never been implemented.  Setting a cookie pair was disabled in OTP_R14B01.

So, using a cookie pair has not worked since R12B, and using our for incoming and the other's for outgoing as a cookie pair has never been implemented.  Until in this pull request.  Now it will finally be possible to connect nodes with different cookies as it was documented in OTP_R10B and clarified in OTP_R12B.  The same explanation remains today in http://erlang.org/doc/reference_manual/distributed.html#security, fifth paragraph:
> For a node Node1 with magic cookie Cookie to be able to connect to, or accept a connection from, another node Node2 with a different cookie DiffCookie, the function erlang:set_cookie(Node2, DiffCookie) must first be called at Node1. Distributed systems with multiple user IDs can be handled in this way.

### DIFF
--- a/erts/doc/src/erl_cmd.xml
+++ b/erts/doc/src/erl_cmd.xml
@@ -4,7 +4,7 @@
 <comref>
   <header>
     <copyright>
-      <year>1996</year><year>2020</year>
+      <year>1996</year><year>2021</year>
       <holder>Ericsson AB. All Rights Reserved.</holder>
     </copyright>
     <legalnotice>
@@ -626,6 +626,12 @@ $ <input>erl \
       <tag><c><![CDATA[-setcookie Cookie]]></c></tag>
       <item>
         <p>Sets the magic cookie of the node to <c><![CDATA[Cookie]]></c>; see
+          <seemfa marker="erlang#set_cookie/2">
+          <c>erlang:set_cookie/2</c></seemfa>.</p>
+      </item>
+      <tag><c><![CDATA[-setcookie Node Cookie]]></c></tag>
+      <item>
+        <p>Sets the magic cookie of Node to <c><![CDATA[Cookie]]></c>; see
           <seemfa marker="erlang#set_cookie/2">
           <c>erlang:set_cookie/2</c></seemfa>.</p>
       </item>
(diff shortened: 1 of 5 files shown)


RaimoNiskanen:
I am closing this PR since it is a too large change in the current behaviour

Issue #5088, opened 2021-08-03
Common_test : grammar in doc for test_spec doesn't match actual behavior.


acw224:
Here is a subset of the current grammar as described in https://erlang.org/doc/apps/common_test/run_test_chapter.html#test_specifications

 Groups          = GroupPath | [GroupPath] | GroupSpec | [GroupSpec] 
 GroupPath       = [GroupName]
 GroupSpec       = GroupName | {GroupName,Properties,GroupSpec}
 GroupName       = atom()

This grammar is ambiguous : [group1, group2] can either be understood as a GroupPath or as a [GroupSpec]. In practice, it is understood as the latter, so the former parsing path should be removed.
The second problem is that the item {GroupName,Properties,GroupSpec} is also not recognised, but only {GroupName,Properties,[GroupSpec]} is.

Hence as a first step, we propose to update the Grammar to be:

 Groups          = GroupPath | GroupSpec | [GroupSpec] 
 GroupPath       = [[GroupName]]
 GroupSpec       = GroupName | {GroupName,Properties,[GroupSpec]}
 GroupName       = atom()

which removes the ambiguity and replaces the problematic case with the one working.

From that, it seems that since there is support for group specifications and group-paths, the group-path should be able to handle group-specifications instead of just group-names.

Expected Behaviour: 

Ideally, we would like to target a grammar of the form: 

 Groups          = GroupPath | GroupSpec | [GroupSpec] 
 GroupPath       = [[GroupSpec]]
 GroupSpec       = GroupName | {GroupName,Properties,[GroupSpec]}
 GroupName       = atom()

Affected versions
All.

acw224:
Associated Pull Request : https://github.com/erlang/otp/pull/5089

PR #5090, opened 2021-08-04
Common_test : Updated grammar for tests specifications

acw224:
CommonTest : Update the grammar for test_spec and add relevant tests as documented in #5088

### DIFF
--- a/lib/common_test/src/ct_run.erl
+++ b/lib/common_test/src/ct_run.erl
@@ -2050,8 +2050,23 @@ final_tests1([{TestDir,Suite,GrsOrCs}|Tests], Final, Skip, Bad) when
 			  [ct_groups:make_conf(TestDir, Suite,
 					       GroupName, Props, TCs)];
 		     ({GroupOrGroups,TCs}) ->
-			  [ct_groups:make_conf(TestDir, Suite,
-					       GroupOrGroups, [], TCs)];
+			  case GroupOrGroups of
+			   [GroupList] when is_list(GroupList) ->
+				   {GrpNames, Props} = lists:foldl(
+				   fun({GrpName,_} = GrSpec, {GrpNames, Props}) -> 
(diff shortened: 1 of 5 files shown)


IngelaAndin:
I think this sounds like a good idea. We are still working to get up to full speed again after summer vacations so, please have some patience. We will start by testing your branch and hopefully have some more feedback fairly soon.

lib/common_test/src/ct_run.erl
```diff
@@ -2050,8 +2050,29 @@ final_tests1([{TestDir,Suite,GrsOrCs}|Tests], Final, Skip, Bad) when
...
+				 				({GroupName,_,_}, Groups)
+							 -> [GroupName | Groups];
+							    (GroupName, Groups)
+							 -> [GroupName | Groups] end,
+													 [], GroupList),
+						 [ct_groups:make_conf(TestDir, Suite,
+				 				[GroupNames], Props, TCs)];
+				 _ -> [ct_groups:make_conf(TestDir, Suite,
+				 GroupOrGroups, [], TCs)]
+			end;
```
IngelaAndin, line 2075: This is rather hard to read. For instance if you format the fun like 
```
fun({_,_} = GrSpec, Props) -> 
		[GrSpec | Props];
	   ({_,_,_} = GrSpec, Props) -> 
		[GrSpec | Props];
	   (_, Props) -> 
		Props 
	end,
```     it becomes easier to parse.  Also moving GroupNames first and then refering to it in "Props construction" as well as later, I think would make the code easier to read. Even without my suggest changes the indentation does not follow the normal indentation rules.

u3s:
- please fix the build error
- you might consider branch renaming, to make it similar to OTP convention - e.g. "acw224/ct/group_parsing_modification"

acw224:
Updated with new branch name in #5242

Issue #3533, opened 2021-01-17
ERL-1460: Mention state copy in efficiency guide


OTP-Maintainer:
Original reporter: `quickcheck`
Affected version: `Not Specified`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1460

---

```
I don't know where to add this knowledge, but bumped into a performance issue caused by misunderstanding the compiler. I think the efficiency-guide/advanced.html under memory is a good place to write something about this. But there probably is a better place.

The issue is:

when the programmer writes:
{code:java}
handle_call(f, _From, State) ->
    Fun = fun() -> State#state.small == 12 end,
    {reply, Fun, State}.{code}
then the complete state is the "environment" of the function. This complete state is copied and therefore this operation may take a lot of time when other field of the state is huge (see example attached).

So, one should advice users to bind only the part of state needed (or optimise the compiler):
{code:java}
handle_call(f, _From, State) ->
    X = State#state.small,
    Fun = fun() -> X == 12 end,
    {reply, Fun, State}.
{code}
```

PR #5099, opened 2021-08-06
Add a chapter about maps to the Efficiency Guide

bjorng:


system/doc/efficiency_guide/maps.xml
```diff
@@ -0,0 +1,667 @@
...
+          <cell align="center"><em>foo</em></cell>
+          <cell align="center"><em>baz</em></cell>
+          <cell align="center"><em>bird</em></cell>
+        </row>
+        <tcaption>#{a => foo, q => baz, z => bird}</tcaption>
+      </table>
+
+      <p>When the value for an existing key is updated, the key tuple is not
+      be updated, allowing the key tuple to be shared with instances of the
+      map that have the same keys. In fact, the key tuple can be shared between all
```
zuiderkwast, line 286: Typo "be" and suggestion "other".

```suggestion
      <p>When the value for an existing key is updated, the key tuple is not
      updated, allowing the key tuple to be shared with other instances of the
      map that have the same keys. In fact, the key tuple can be shared between all
```

system/doc/efficiency_guide/maps.xml
```diff
@@ -0,0 +1,667 @@
...
+      <p>Defined like this, the key tuple <c>{a,b,c}</c> will be a
+      global literal. To ensure that the key tuple is shared
+      when creating an instance of the map, always call
+      <c>new()</c> and modify the returned map:</p>
+
+      <code type="erl"><![CDATA[
+    (SOME_MODULE:new())#{a := 42}.]]></code>
+
+      <p>Using the map syntax with small maps is particularly
+      efficient.  As long as the keys are known at compile-time, the
```
zuiderkwast, line 303: Emacs space.

```suggestion
      efficient. As long as the keys are known at compile-time, the
```

system/doc/efficiency_guide/maps.xml
```diff
@@ -0,0 +1,667 @@
...
+      more efficient, and it helps catching mispellings of
+      keys.</p></item>
+
+      <item><p>Whenever possible, match multiple map elements at
+      once.</p></item>
+
+      <item><p>Whenever possible, update multiple map elements at
+      once.</p></item>
+
+      <item><p>Avoid default values and the <seemfa
```
garazdawi, line 105: Maybe add an explanation why maps:get/3 should be avoided?

garazdawi (commented):
Should there be a link from the maps module documentation to this document?

system/doc/efficiency_guide/maps.xml
bjorng: I have added a link from the `maps` module documentation and added an explanation as to why `maps:get/3` should be avoided.

system/doc/efficiency_guide/maps.xml
```diff
@@ -0,0 +1,681 @@
...
+      marker="stdlib:maps#get/3">maps:get/3</seemfa> function. If
+      there are default values, sharing of keys between different
+      instances of the map will be less effective, and it is not
+      possible to match multiple elements having default values in one
+      go. The <c>maps:get/3</c> function is <seecref
+      marker="#maps_get_3">implemented in Erlang</seecref>, making it
```
garazdawi, line 111:
```suggestion
      go. The <c>maps:get/3</c> function is <seeguide
      marker="#maps_get_3">implemented in Erlang</seeguide>, making it
```

josevalim:
Excellent read! 💯 

One common pitfall, that applies to both records and maps, and I see people do from time to time is the use of maps/records in closures where you only need certain fields. For example, someone using Cowboy might do:

    spawn_link(fun() -> log(map_get(path, CowboyReq)) end)

Which could be quite problematic as it has to copy the whole CowboyReq. Of course, it would be better written as:

    Path = map_get(path, CowboyReq),
    spawn_link(fun() -> log(Path) end)

I am not sure if this belongs to the efficiency guide - but I thought I would mention it.

bjorng:
> One common pitfall, that applies to both records and maps, and I see people do from time to time is the use of maps/records in closures where you only need certain fields.

Yes, I have helped people debug that kind of problem a few times. It has also been reported: #3533. I haven't got around to write something about it yet. I do think it belongs in the Efficiency Guide, because it is a reason for mysterious huge memory consumption leading to bad performance and/or crashes.

PR #5101, opened 2021-08-07
Add is_map_key/2 as a guard BIF to the reference documentation

tomas-abrahamsson:
I noticed [`is_map_key/2`](https://erlang.org/doc/man/erlang.html#is_map_key-2) is allowed in guard tests, but the reference documentation currently does not list it, neither the [Maps in Guards](https://erlang.org/doc/reference_manual/expressions.html#maps-in-guards) section, nor the [Guard Expressions](https://erlang.org/doc/reference_manual/expressions.html#guard-expressions) section. This PR adds it to the reference documentation. I added it to the table of type test BIFs since this is where the rest of the `is_...` functions are listed.

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1085,6 +1085,10 @@ handle_call(change, From, #{ state := start } = S) ->
 			   in the <c>erlang</c> module
 		  </item>
 		  <item>
+			  <seemfa marker="erts:erlang#is_map_key/2">is_map_key/2</seemfa>
+			   in the <c>erlang</c> module
+		  </item>
+		  <item>
 			  <seemfa marker="erts:erlang#map_get/2">map_get/2</seemfa>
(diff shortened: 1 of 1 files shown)


zuiderkwast:
Nice, but it's not really a type test bif, so I think it's better to put it in the table of Other BIFs allowed in guards.

tomas-abrahamsson:
Yes, you have a point. Should I move it? It would then be the only `is_` function in that table, and also the only function returning a boolean. The table it is in now is full of those. But maybe that is secondary.

zuiderkwast:
I think move it, but let's leave it to the maintaimers to decide.

bjorng:
> I think move it, but let's leave it to the maintaimers to decide.

I also think it should be moved.

tomas-abrahamsson:
I moved it to the other table and force-pushed it.

bjorng:
Thanks for your pull request.

PR #5102, opened 2021-08-07
tools: emacs: add regexp for Edoc errors

sirikid:
<details>
<summary>Compilation example</summary>

image

</details>

### DIFF
--- a/lib/tools/emacs/erlang.el
+++ b/lib/tools/emacs/erlang.el
@@ -5581,6 +5581,15 @@ future, a new shell on an already running host will be started."
   (interactive)
   (call-interactively erlang-next-error-function))
 
+(defconst erlang-edoc-error-regexp
+  "^\\(.*\\), function \\(.*\\): at line \\([0-9]+\\):"
+  "Regexp matching Edoc error message.")
+
+(add-to-list 'compilation-error-regexp-alist-alist
+             `(erlang-edoc
+               ,erlang-edoc-error-regexp
+               1 3 nil 1 nil (2 font-lock-function-name-face)))
+
 
 
 ;;;


kjellwinblad:
Hi,

I tried your change and did not get it to work.

I then googled for `compilation-error-regexp-alist-alist` and found that there is also an `compilation-error-regexp-alist`. After adding:

```
(add-to-list 'compilation-error-regexp-alist 'erlang-edoc)
```

it started to work.

Is it possible that the line I added is missing in your commit?

Perhaps it works for you but not for me because we have different setups. I'm using 25.2.2 and start the compilation by typing "M-x compile ENTER rebar3 edoc ENTER". If it works for you even without `(add-to-list 'compilation-error-regexp-alist 'erlang-edoc)` and adding that line does not break anything for you, you can just add this line so it works in both of our setups.

sirikid:
> Is it possible that the line I added is missing in your commit?

No, `compilation-error-regexp-alist` is a [customizable variable][1] and should not be changed when the library is loaded.

[1]: https://www.gnu.org/software/emacs/manual/html_node/emacs/Easy-Customization.html

kjellwinblad:
> No, compilation-error-regexp-alist is a customizable variable and should not be changed when the library is loaded.

Ok, but is this feature not expected to work without any extra configuration?

Is there anyway to get it to work out-of-the-box without adding to `compilation-error-regexp-alist`?

legoscia:
FWIW, I think it would be better to change Edoc to output error messages according to the [GNU standard](https://www.gnu.org/prep/standards/html_node/Errors.html) like the Erlang compiler does already, e.g.:

```
edoc_test.erl:5: function foo/0: warning: tag @docx not recognized
```

erszcz:
> I think it would be better to change Edoc to output error messages according to the GNU standard like the Erlang compiler does already

Indeed, EDoc errors are not very structured and could use some tidying up :+1:

sirikid:
> > No, compilation-error-regexp-alist is a customizable variable and should not be changed when the library is loaded.
> 
> Ok, but is this feature not expected to work without any extra configuration?
> 
> Is there anyway to get it to work out-of-the-box without adding to `compilation-error-regexp-alist`?

I can submit this patch to Emacs :smile: 

> FWIW, I think it would be better to change Edoc to output error messages according to the [GNU standard](https://www.gnu.org/prep/standards/html_node/Errors.html) like the Erlang compiler does already, e.g.:
> 
> ```
> edoc_test.erl:5: function foo/0: warning: tag @docx not recognized
> ```

That would be great. But swap function and warning:
```
edoc_test.erl:5: warning: function foo/0: tag @docx not recognized
```

kjellwinblad:
I noticed that this PR makes some of our nightly tests fail. @sirikid can you try to reproduce and look into if you can see why this is happening:

Here is how to run the tests:
```
export ERL_TOP=/path/to/opt/repo
export TESTSERVER="$ERL_TOP"/release/tests/test_server
./configure
make
(make release_tests && cd $TESTSERVER && cerl -eval "ts:install(), ts:run(tools, emacs_SUITE, [batch]),erlang:halt()")
```
Note that if you change the checked-in test case, you will have to rerun `make release_tests` so that the update is copied to the $TESTSERVER dir.

sirikid:
I am working on adopting GNU style into edoc.

PR #5104, opened 2021-08-09
Reduce the size of the JITted code for try_case

bjorng:
This commit reduces X register shuffling after an exception has
occurred. This will slightly reduce the size of the JITted code, but
is unlikely to improve speed because is usually rare that execeptions
are raised.

### DIFF
--- a/erts/emulator/beam/emu/instrs.tab
+++ b/erts/emulator/beam/emu/instrs.tab
@@ -1086,28 +1086,39 @@ catch(Y, Fail) {
 }
 
 catch_end(Y) {
+    /*
+     * At entry:
+     *
+     *    x0 = Result of expression or THE_NON_VALUE
+     *
+     * If x0 is THE_NON_VALUE, the following registers are also set:
+     *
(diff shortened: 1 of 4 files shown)


erts/emulator/beam/emu/instrs.tab
```diff
@@ -1086,28 +1086,36 @@ catch(Y, Fail) {
 }
 
 catch_end(Y) {
+    /*
+     * At entry:
+     *
+     *    x0 = THE_NON_VALUE
```
garazdawi, line 1092: `x0` can be a value here as well, right?

garazdawi (approved):

erts/emulator/beam/emu/instrs.tab
bjorng: Yes, you are right. I have force-pushed a correction.

bjorng:
I have pushed a correction of exception traces.

jhogberg (approved):

PR #5125, opened 2021-08-17
Add a section about accidental copying and loss of sharing

### LINKED ISSUE #3533: ERL-1460: Mention state copy in efficiency guide

bjorng:
Add a section about accidental copying and loss of sharing to the
Common Caveats chapter.

While at it, also update information about copying in the Processes
chapter as follows:

* Lift up several sub sections to the top level to make them visible
in the left sidebar.

* Rename the "Constant Pool" to "Literal Pool" (which is the correct
technical term), and also update information about copying and
persistent terms.

Closes #3533.

### DIFF
--- a/system/doc/efficiency_guide/processes.xmlsrc
+++ b/system/doc/efficiency_guide/processes.xmlsrc
@@ -11,7 +11,7 @@
       Licensed under the Apache License, Version 2.0 (the "License");
       you may not use this file except in compliance with the License.
       You may obtain a copy of the License at
- 
+
           http://www.apache.org/licenses/LICENSE-2.0
 
       Unless required by applicable law or agreed to in writing, software
@@ -19,7 +19,7 @@
       WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
       See the License for the specific language governing permissions and
       limitations under the License.
-    
+
     </legalnotice>
 
     <title>Processes</title>
@@ -27,7 +27,7 @@
     <docno></docno>
     <date>2007-11-21</date>
     <rev></rev>
-    <file>processes.xml</file>
+    <file>processes.xmlsrc</file>
   </header>
 
   <section>
@@ -49,7 +49,7 @@ Eshell V5.6  (abort with ^G)
 {memory,1232}
 3> <input>Bytes div erlang:system_info(wordsize).</input>
 309</pre>
-    
+
     <p>The size includes 233 words for the heap area (which includes the
     stack). The garbage collector increases the heap as needed.</p>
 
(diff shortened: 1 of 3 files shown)


system/doc/efficiency_guide/processes.xmlsrc
```diff
@@ -268,87 +271,98 @@ cross_function_receive(Ref) ->
...
+    <pre>
+    erl +MIscs 2048</pre>
+  </section>
+
+  <section>
+    <marker id="loss-of-sharing"></marker>
+    <title>Loss of Sharing</title>
+
+    <p>Shared subterms are <em>not</em> preserved in the following
+    cases:</p>
```
sverker, line 324: Here I think it would be nice to start off with a really simple example of shared subterms to explain what it is. Like `{SubTerm, SubTerm}`.

system/doc/efficiency_guide/processes.xmlsrc
```diff
@@ -358,13 +372,12 @@ true
...
-      return the same value. Sharing has been lost.</p>
+    <p>When the data has passed through an Ets table,
+    <c>erts_debug:size/1</c> and <c>erts_debug:flat_size/1</c>
+    return the same value. Sharing has been lost.</p>
 
-      <p>In a future Erlang/OTP release, it might be implemented a
-      way to (optionally) preserve sharing.</p>
-    </section>
+    <p>In a future Erlang/OTP release, it might be implemented a
+    way to (optionally) preserve sharing.</p>
```
sverker, line 380: Should we mention configure option  `--enable-sharing-preserving`?

system/doc/efficiency_guide/processes.xmlsrc
bjorng: Done.

system/doc/efficiency_guide/processes.xmlsrc
bjorng: Done.

Issue #4920, opened 2021-05-31
erlang (erl.exe) executable built with –enable-fips option generates core while shutting down rabbitMQ(3.8.14) on windows


peculiar-pooja:
**Describe the bug**

While stopping rabbitmq on windows host below core is generated by erl.exe
========== Stack trace from Coredump ===========

CONTEXT:  (.ecxr)
rax=0000000000000000 rbx=000000820ec4bd20 rcx=0000000000000000
rdx=0000000000000000 rsi=0000008209facd30 rdi=000000000001330b
rip=00007ffe120a4520 rsp=000000820b99f570 rbp=00000082099801c0
r8=0000008209facdc8  r9=0000000000000000 r10=0000000000000000
r11=00007ffe123abad0 r12=00007ffe120aa8aa r13=00007ffe120a9e5c
r14=00007ffe120a9e0b r15=0000008209facd50
iopl=0         nv up ei pl zr na po nc
cs=0033  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00010244
beam_smp!build_stacktrace+0xc40:
00007ffe`120a4520 488b11          mov     rdx,qword ptr [rcx] ds:00000000`00000000=????????????????
Resetting default scope

FAULTING_IP:
beam_smp!build_stacktrace+c40
00007ffe`120a4520 488b11          mov     rdx,qword ptr [rcx]

EXCEPTION_RECORD:  (.exr -1)
ExceptionAddress: 00007ffe120a4520 (beam_smp!build_stacktrace+0x0000000000000c40)
   ExceptionCode: c0000005 (Access violation)
  ExceptionFlags: 00000000
NumberParameters: 2
   Parameter[0]: 0000000000000000
   Parameter[1]: 0000000000000000
Attempt to read from address 0000000000000000 

DEFAULT_BUCKET_ID:  NULL_POINTER_READ
PROCESS_NAME:  erl.exe
FOLLOWUP_IP:
beam_smp!build_stacktrace+c40
00007ffe`120a4520 488b11          mov     rdx,qword ptr [rcx]

READ_ADDRESS:  0000000000000000
ERROR_CODE: (NTSTATUS) 0xc0000005 - The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.
EXCEPTION_CODE: (NTSTATUS) 0xc0000005 - The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.
[...]

PR #5126, opened 2021-08-17
Fixing core dump caused by enable/disable FIPS

graorane:
FIPS enable and disable or vice versa just to check curve count, sometime leads to core dump at the time of process exit.  So this fix  avoids enable/disable FIPS.  It delays getting curve count during FIPS mode. Simultaneously it keeps the optimization added to cache of supported curves in the NIF at application

### DIFF
--- a/lib/crypto/c_src/algorithms.c
+++ b/lib/crypto/c_src/algorithms.c
@@ -31,9 +31,10 @@ static unsigned int algo_pubkey_cnt, algo_pubkey_fips_cnt;
 static ERL_NIF_TERM algo_pubkey[12]; /* increase when extending the list */
 void init_pubkey_types(ErlNifEnv* env);
 
-static unsigned int algo_curve_cnt, algo_curve_fips_cnt;
 static ERL_NIF_TERM algo_curve[2][89]; /* increase when extending the list */
+static ErlNifMutex* mtx_init_curve_types;
 void init_curve_types(ErlNifEnv* env);
+int get_curve_cnt(ErlNifEnv* env, int fips);
 
 static unsigned int algo_rsa_opts_cnt, algo_rsa_opts_fips_cnt;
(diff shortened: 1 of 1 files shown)


lib/crypto/c_src/algorithms.c
```diff
@@ -44,14 +45,18 @@ void init_rsa_opts_types(ErlNifEnv* env);
...
+    mtx_init_curve_types =  enif_mutex_create("init_curve_types");
     init_hash_types(env);
     init_pubkey_types(env);
     init_curve_types(env);
     init_rsa_opts_types(env);
     /* ciphers and macs are initiated statically */
 }
 
-
+void cleanup_algorithms_types(ErlNifEnv* env)
```
HansN, line 56: Warning from our internal test build since we always run with -Wmissing-prototypes in our builds:

no previous prototype for ‘cleanup_algorithms_types’ [-Wmissing-prototypes] void cleanup_algorithms_types(ErlNifEnv* env)

lib/crypto/c_src/algorithms.c
```diff
@@ -190,43 +195,79 @@ ERL_NIF_TERM mac_algorithms(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+        algo_curve_cnt_initialized = 1;
+    }
+    enif_mutex_unlock(mtx_init_curve_types);
+
+    return cnt;
+}
+
 void init_curve_types(ErlNifEnv* env) {
-    algo_curve_cnt = algo_curve_fips_cnt = 0;
+    int curve_cnt = 0;
```
HansN, line 253: Here we get the warning:

variable ‘curve_cnt’ set but not used [-Wunused-but-set-variable] int curve_cnt = 0;

HansN:
Do you think you will have some time to fix this soon?

HansN:
Merged to maint for release in OTP-24.1
Resolves #4920

Issue #5040, opened 2021-07-09
mnesia:delete_object/1 breaks index


mikpe:
**Describe the bug**
If `mnesia:delete_object/1` is called to remove a non-existing record from a table with an index, the index is updated incorrectly, causing future `mnesia:dirty_index_read/3` calls to not find records present in the table.

**To Reproduce**
```
> cat mnesia_bug.erl 
-module(mnesia_bug).
-export([test/0]).

-record(bug, {key, ix, val}).

test() ->
  init_mnesia(),
  init_table(),
  {atomic, ok} = mnesia:clear_table(bug),
  %% Populate with a known record and check that the index finds it.
  Record = #bug{key = 2, ix = 4, val = stuff},
  {atomic, ok} = mnesia:transaction(fun() -> mnesia:write(Record) end),
  [Record] = mnesia:dirty_index_read(bug, 4, ix),
  %% This delete_object is for a record which isn't in the table.
  NonexistingRecord = #bug{key = 2, ix = 4, val = nonexisting},
  {atomic, ok} = mnesia:transaction(fun() -> mnesia:delete_object(NonexistingRecord) end),
  %% The index should still find the record, but it doesn't.
  case mnesia:dirty_index_read(bug, 4, ix) of
    [Record] -> ok;
    [] -> error(delete_object_broke_index)
  end.

init_table() ->
  CreateOptions =
    [ {disc_copies, [node()]}
    , {attributes, record_info(fields, bug)}
    , {index, [ix]}
    ],
  {atomic, ok} = mnesia:create_table(bug, CreateOptions),
  ok = mnesia:wait_for_tables([bug], timer:minutes(1)).

init_mnesia() ->
  os:cmd("rm -rf " ++ mnesia_dir()),
  file:make_dir(mnesia_dir()),
  init_mnesia_env(),
  ok = mnesia:create_schema([node()]),
  ok = application:start(mnesia).

init_mnesia_env() ->
  Env =
    [ {dir, mnesia_dir()}
    ],
  [application:set_env(mnesia, Var, Val) || {Var, Val} <- Env],
  ok.

mnesia_dir() ->
  "/tmp/mnesia_bug-db".
> erl
Erlang/OTP 24 [erts-12.0.3] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Eshell V12.0.3  (abort with ^G)
1> c(mnesia_bug).
{ok,mnesia_bug}
2> mnesia_bug:test().
[...]

PR #5131, opened 2021-08-18
mnesia: Do not unconditionally delete index keys

dgud:
delete_object(NonExistingRecord) deleted index keys even if the record
didn't exist, in set tables, and nothing was deleted from the real table.

Solve this by doing the same checks as was done for bag tables.

Fixes GH-5040


dgud:
Does this look ok?
@mikpe @uwiger

mikpe:
> Does this look ok?
> @mikpe @uwiger

I'll take a look on Monday when I'm back to work.

uwiger:
LGTM /Ulf

Issue #5128, opened 2021-08-17
fts_rand_float_decimals test fail in some rare case


DianaOlympos:
[Link to the test code](https://github.com/erlang/otp/blob/master/erts/emulator/test/num_bif_SUITE.erl#L283)
Reproduction at the end

This was found as part of #4719 

Here is what i understand of the problem here.

We are basically testing `printf` from the libc. The problem is that we try to compute the maximum difference when limiting the number of decimals from first principles... but this is heavily influenced by _what type of rounding does `printf` use_.

Sadly, i went diving into the C spec and the spec says

> The low-order digit shall be rounded in an implementation-defined manner. 

What that means is that we are incapable of predicting the maximum difference in a way that is highly useful, cause we are forced to expect it to round up or down by a whole unit.

So i think this test should be deleted and would be happy to offer a PR for this.

Additional support for this deletion:
- Some of this test is already not run on windows
- Some of this test is already not run for more than 10 decimals

At this point, there is so much edge cases we cannot protect on for a test that basically test `printf` that it seems we should just delete it.

```
F0 = -0.48224313707914995 <<191,222,221,18,81,158,108,219>>

L1 = -0.4822431370792

F1 = -0.4822431370792 <<191,222,221,18,81,158,112,97>>

Diff = 5.007105841059456e-14, MaxDiff = 5.005551115123126e-14


*** CT Error Notification 2021-08-11 17:29:43.936 ***
num_bif_SUITE:'-fts_rand_float_decimals/1-lc$^0/1-0-' failed on line 309
Reason: {badmatch,error}

Full error description and stacktrace




=== Ended at 2021-08-11 17:29:43
=== Location: [{num_bif_SUITE,'-fts_rand_float_decimals/1-lc$^0/1-0-',309},
              {num_bif_SUITE,fts_rand_float_decimals,319},
              {num_bif_SUITE,t_float_to_string,203},
              {test_server,ts_tc,1783},
              {test_server,run_test_case_eval1,1292},
              {test_server,run_test_case_eval,1224}]
=== === Reason: no match of right hand side value error
  in function  num_bif_SUITE:'-fts_rand_float_decimals/1-lc$^0/1-0-'/1 (num_bif_SUITE.erl, line 309)
  in call from num_bif_SUITE:fts_rand_float_decimals/1 (num_bif_SUITE.erl, line 319)
  in call from num_bif_SUITE:t_float_to_string/1 (num_bif_SUITE.erl, line 203)
[...]

PR #5133, opened 2021-08-18
delete part of fts_rand_float_decimals that depend on libc rounding

DianaOlympos:
This is a fix for #5128 

I am happy to hear for solution that are not just about deleting it, but i do not see how else to handle the fact that the rounding of the last decimal digit is unspecified in the spec and let to the implementers

### DIFF
--- a/erts/emulator/test/num_bif_SUITE.erl
+++ b/erts/emulator/test/num_bif_SUITE.erl
@@ -305,7 +305,7 @@ fts_rand_float_decimals(N) ->
               end,
          F1 = list_to_float(L1),
          Diff = abs(F0-F1),
-         MaxDiff = max_diff_decimals(F0, D),
+         MaxDiff = max_diff_decimals(F0, D-1),
          ok = case Diff =< MaxDiff of
                   true -> ok;
                   false ->


kjellwinblad:
Hi @DianaOlympos and thanks for the PR!

Is the problem that, for example, the following:

`erlang:float_to_list(0.115, [{decimals, 2}]).`

can on some platforms give:

"0.12"

but on other platforms it can give:

"0.11"

?

Is that correct?

Could we just change the line `max_diff_decimals(F0, D)` to `max_diff_decimals(F0, D-1)`? That way, the test will at least check that the result is not obviously wrong.

DianaOlympos:
Yep that is the problem ! And there is no way to know. I mean we can check that, but i think we kinda already check that with the checking of the formatting above ? Up to you, i can change it. I think this whole test feel really icky and unstable already tbh.

kjellwinblad:
> but i think we kinda already check that with the checking of the formatting above

Do you mean the `L0 =:= IOL` check? Have you checked if `lists:flatten(io_lib:format("~.*f", [D, F0]))` is not just using `erlang:float_to_list` to get the string?

DianaOlympos:
I just checked this, finally.

So `io_lib:format` end up calling `io_lib_format:fwrite_f` which does `float_to_list` then do all kind of change on it. So yeah i suppose we can try to just keep it as `D-1` for now. I still think the whole test stink but i do not have the energy nor time to find a better test

kjellwinblad:
Merged a83541527625fae5d7f849718a052cf6f2833809

PR #5134, opened 2021-08-18
delete all use of -Wdeclaration-after-statement in configure

DianaOlympos:
Following a discussion with @garazdawi in #4719 , this delete all the checks to add `-Wdeclaration-after-statement` in the CFLAGS.

This is legal in C99 and at this point all compiler supported on all supported platform should be able to deal with it.

### DIFF
--- a/erts/configure
+++ b/erts/configure
@@ -7984,31 +7984,6 @@ fi
       WFLAGS="$WFLAGS -Wmissing-prototypes";;
   esac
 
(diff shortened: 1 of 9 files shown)


garazdawi:
We have decided that we want to continue to force declarations to be at the top of the scope, even if C99 allows it to be otherwise.

The reason for this is that all code we have now is written using that style (so we are used to it that way) and we believe it forces the coder into making the scoping of variables clearer.

PR #5141, opened 2021-08-20
Fix inconsistency in defining ?FILE macro

pergu:
This happens when the source_name is explicitly defined as an option to epp:parse_file/2. Before an include it got expanded to the actual file name, after an include it was defined as the source_name. With this change it will always be the source_name.

This come up when using --deterministic mode to compile causing the output to be less deterministic if you use the ?FILE macro before any includes.

This is really an update to https://github.com/erlang/otp/pull/4947 put somehow that PR lost my commits. This incorporates all the changes discussed in that PR.


bjorng:
Thanks! Looks good.

Since this is a bug fix, I think that it should be based on maint so that we can include it in OTP 24.1. Could you rebase it on maint and edit the PR to make maint the base?

pergu:
Ok, I have rebased it on maint now.

bjorng:
Thanks. Added to our daily builds.

PR #5145, opened 2021-08-23
Malformed DNS RRs shall result in parse errors

snar:
For now, when DNS RR is malformed it results in 'parse ok,  however, some records are not parsed':

`3> inet_res:resolve("kinolupa.net", in, aaaa, [], 5000).`
`{ok,#dns_rec{header = #dns_header{id = 1,qr = true,
                                  opcode = query,aa = false,tc = false,
                                  rd = true,ra = true,pr = false,rcode = 0},
             qdlist = [#dns_query{domain = "kinolupa.net",type = aaaa,
                                  class = in}],
             anlist = [#dns_rr{domain = "kinolupa.net",type = aaaa,
                               class = in,cnt = 0,ttl = 600,
                               data = <<171,22,127,201>>,
                               tm = undefined,bm = [],func = false},
                       #dns_rr{domain = "kinolupa.net",type = aaaa,class = in,
                               cnt = 0,ttl = 600,
                               data = <<171,22,127,201>>,
                               tm = undefined,bm = [],func = false}],
             nslist = [],arlist = []}}
`

As you can see, some broken DNS server returned two malformed (too short) AAAA records that 
erlang was not able to parse and so they are returned as binary <<171,22,127,201>>. 
With this pull applied, such broken records results in `{error, formerr}`.

### DIFF
--- a/lib/kernel/src/inet_dns.erl
+++ b/lib/kernel/src/inet_dns.erl
@@ -500,6 +500,13 @@ decode_data(<<Flags:8,Data0/binary>>, _, ?S_CAA, _) ->
         orelse throw(?DECODE_ERROR),
     Value = binary_to_list(Data1),
     {Flags,inet_db:tolower(Tag),Value};
+%% malformed known RR in inet domain
+decode_data(_, in, T, _) when T == ?S_A; T == ?S_AAAA; T == ?S_WKS ->
+    throw(?DECODE_ERROR);
+%% malormed known RR in any domain
+decode_data(_, _, T, _) when T == ?S_HINFO; T == ?S_MX; T == ?S_SRV;
+    T == ?S_NAPTR; T == ?S_URI; T == ?S_CAA ->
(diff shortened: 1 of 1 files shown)


RaimoNiskanen:
@snar: Can you test the code with my rewrite to verify that it seems to work as expected?

snar:
On Mon, Aug 30, 2021 at 05:35:31AM -0700, Raimo Niskanen wrote:
> @snar: Can you test the code with my rewrite to verify that it seems to work as
> expected?

Sure, already done: known 'bad' query results in {error, formerr} as
expected:

17> inet_res:resolve("kinolupa.net", in, aaaa, [{nameservers, [{{47,74,31,16}, 53}]}], 5000).
{error,formerr}

smoke tests for 'good' queries results in perfectly normal answers.

RaimoNiskanen:
Good to know!  Thank you for your pull request!

Issue #5122, opened 2021-08-16
raw option doesn't work gen_tcp when the socket backend is enabled (otp 24)


dogtapdsd:
**Describe the bug**
[https://bugs.erlang.org/browse/ERL-1287](url)
this bug was fixed in otp 23.1, but happen again in otp 24,   It cann't  listening multiple sockets  on the same port.

**To Reproduce**

> Eshell V12.0.3  (abort with ^G)
1> 
1>  gen_tcp:listen(0, [{inet_backend, socket}, {raw, 1, 15, <<1:32/native>>}, {port, 8888}, {ip, any}]).
{ok,{'$inet',gen_tcp_socket,
             {<0.82.0>,
              {'$socket',#Ref<0.1220593668.1297743873.200397>}}}}
2> 
2>  gen_tcp:listen(0, [{inet_backend, socket}, {raw, 1, 15, <<1:32/native>>}, {port, 8888}, {ip, any}]).
{error,eaddrinuse}

**Expected behavior**
I test in otp 23.3 with no error

>Eshell V11.2  (abort with ^G)
1> gen_tcp:listen(0, [{inet_backend, socket}, {raw, 1, 15, <<1:32/native>>}, {port, 8888}, {ip, any}]).
{ok,{'$inet',gen_tcp_socket,
             {<0.83.0>,
              {'$socket',#Ref<0.3300202661.3446013953.155967>}}}}
2> 
2> gen_tcp:listen(0, [{inet_backend, socket}, {raw, 1, 15, <<1:32/native>>}, {port, 8888}, {ip, any}]).
{ok,{'$inet',gen_tcp_socket,
             {<0.85.0>,
              {'$socket',#Ref<0.3300202661.3446013953.155982>}}}}
3> 

**Affected versions**
Erlang/OTP 24 [erts-12.0.3]

RaimoNiskanen:
It seems that I refactored code when implementing `fdopen/2` in `gen_tcp_socket`, and then setting level socket options was moved to after the `bind(2)` call.  Now I will have to figure out if that was purely accidental or if there was some reason behind it.

That code refactoring causes your raw option (that by the way exists in non-raw form too `{reuseaddr, true}`), to be set after the address has been bound.  Hence it has no effect.  The same applies to the non-raw option too...  So this is formally not the same bug that was fixed in ERL-1287.

I want to take a more thorough look at option handling in gen_tcp_socket, to see how to make it less messy...

RaimoNiskanen:
[...]

PR #5151, opened 2021-08-25
Improve inet_backend = socket option parsing for gen_tcp (OTP-17580)

RaimoNiskanen:
This is a rewrite of the option parsing in `gen_tcp_socket`, a.k.a the implementation module for `inet_backend = socket`.

The intention is to make it a little less messy, and to fix the bug in #5122.

### DIFF
--- a/lib/kernel/src/gen_tcp_socket.erl
+++ b/lib/kernel/src/gen_tcp_socket.erl
@@ -316,8 +316,7 @@ accept(?MODULE_socket(ListenServer, ListenSocket), Timeout) ->
         Server =
             val(ErrRef,
                 start_server(
-                 ServerData,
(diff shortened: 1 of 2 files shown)


bmk:
filtermap_opts_1: list comprehension?

RaimoNiskanen:
`filtermap_opts()` -> list comprehension...  That could work!

One perk of this implementation of `filtermap` is that it speculates to not change the result list, and if the result is not changed; no new list nor any garbage is created.  But I realize that it should at least follow the spec of `lists:filtermap/2`.  And premature optimization is known to be one root cause of evil...

RaimoNiskanen:
I changed to list comprehensions - it worked nicely

bmk:
Other then that, it looks good. Simplifying the option handling was neeeeded.

bmk (approved):

PR #5153, opened 2021-08-26
Ensure reproducible builds of Erlang code

bjorng:
The `deterministic` option for the compiler is not always sufficient
to make builds reproducible. There remains the problems with map
literals having more than 32 elements. Literals are converted to the
external term format by term_to_binary/2 before embedding them in the
BEAM file. However, the order in which individual map elements are
encoded for a map with more than 32 elements can vary.

To fix this issue, implement the option `deterministic` for
term_to_binary/2. When `deterministic` is given, term_to_binary/2
sorts map elements into a total term order before encoding them.

Also update the compiler to use `deterministic` when encoding
literals, and also to stop evaluating term_to_binary/1 with a literal
argument.


bjorng:
I have force-pushed a commit that:

* Eliminates a potential memory leak
* Fixes a crash for the DEBUG build of the runtime system

erts/emulator/beam/external.c
```diff
@@ -3023,7 +3054,51 @@ dec_pid(ErtsDistExternal *edep, ErtsHeapFactory* factory, const byte* ep,
...
+        return 0;
+    }
+}
+
+#ifdef INCLUDE_YCF_TRANSFORMED_ONLY_FUNCTIONS
+static void yielding_qsort(Eterm *map_array, Uint n)
+{
+    erts_qsort(map_array, n/2, 2*sizeof(Eterm),
+               (int (*)(const void *, const void *)) map_key_compare);
+}
```
kjellwinblad, line 3086: One could call `erts_qsort_ycf_gen_yielding` dirctly so this wrapper is not needed. `erts_qsort_ycf_gen_yielding` has a prototype in  [global.h](https://github.com/erlang/otp/blob/master/erts/emulator/beam/global.h#L1408). Having the wrapper is of course a matter of taste and I don't see any reason not to have it other than that YCF does not need to do any work for this file if the rapper does not exist.

erts/emulator/beam/external.c
```diff
@@ -3142,6 +3224,87 @@ enc_term_int(TTBEncodeContext* ctx, ErtsAtomCacheMap *acmp, Eterm obj, byte* ep,
...
+                    ctx->ycf_yield_state = NULL;
+                    yielding_qsort_ycf_gen_yielding(&num_reductions,
+                                                    &ctx->ycf_yield_state,
+                                                    NULL,
+                                                    ycf_yield_alloc,
+                                                    ycf_yield_free,
+                                                    NULL,
+                                                    0,
+                                                    NULL,
+                                                    map_array, n);
```
kjellwinblad, line 3264: I think there is a potential for a memory leak as I don't think `yielding_qsort_ycf_gen_destroy` is called anywhere. A memory leak can happen if the process calling term_to_binary/2 is killed when yielded in the `yielding_qsort_ycf_gen_yielding` call. Then `yielding_qsort_ycf_gen_yielding` has allocated memory that will never be freed.

kjellwinblad (commented):
Looks good to me except the memory leak issue that is mentioned in a comment.

bjorng:
Pushed a fixup commit that eliminates the wrapper and the memory leak.

PR #5160, opened 2021-08-29
diameter: Change callback anchors from `Mod:` to `Module:`

wojtekmach:
- diameter: Change callback anchors from `Mod:` to standard `Module:`

`#Module:` is the more commonly used way of referring to callbacks in
the generated links.

By standardising on this, we improve interopability for external tools.
For example, ExDoc can generate URLs using the uniform pattern. (As
opposed to, for example, maintain a list of per-module exceptions.)

- ~diameter: Use `<seetype>` for types~

### DIFF
--- a/lib/diameter/doc/src/diameter_app.xml
+++ b/lib/diameter/doc/src/diameter_app.xml
@@ -180,7 +180,7 @@ process.</p>
 <funcs>
 
 <func>
-<name since="OTP R14B03">Mod:peer_up(SvcName, Peer, State) -> NewState</name>
+<name since="OTP R14B03">Module:peer_up(SvcName, Peer, State) -> NewState</name>
 <fsummary>Invoked when a transport connection has been established</fsummary>
 <type>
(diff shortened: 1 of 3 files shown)


garazdawi:
The `seeerl` tags in the diameter docs are correct the way they are. This is because diameter does not use `-type` as the source of documentation type information, so the datatype section from erl_docgen's point of view is just like any other `<section>` in the documentation, and thus `<seeerl>` should be used to point to the manually inserted `<marker>`s.

wojtekmach:
Got it, makes sense. I removed that commit from the PR and updated the description accordingly.

KennethL:
The reason is to get a uniform naming and linking to callback functions all over OTP.

otp-anders:
That's fine, but that should be explained in the commit comment.

wojtekmach:
I've updated the PR & commit description and rebased with the latest master.

wojtekmach:
No longer relevant!

PR #5162, opened 2021-08-30
peer: new module to replace 'slave'

max-au:
New 'peer' module is designed to replace multiple implementations present in the codebase, mostly in tests. The changeset implements 'peer' with related documentation, replaces usage in test suites, removes 'loose_node' implementing a similar concept, and then marks all functions of old 'slave' as deprecated.

### DIFF
--- a/erts/emulator/test/distribution_SUITE.erl
+++ b/erts/emulator/test/distribution_SUITE.erl
@@ -34,9 +34,6 @@
 
 -include_lib("common_test/include/ct.hrl").
 
-%-define(Line, erlang:display({line,?LINE}),).
--define(Line,).
-
 -export([all/0, suite/0, groups/0,
          init_per_suite/1, end_per_suite/1,
          init_per_group/2, end_per_group/2,
@@ -82,7 +79,7 @@
 -export([sender/3, receiver2/2, dummy_waiter/0, dead_process/0,
          group_leader_1/1,
          optimistic_dflags_echo/0, optimistic_dflags_sender/1,
-         roundtrip/1, bounce/1, do_dist_auto_connect/1, inet_rpc_server/1,
+         roundtrip/1, bounce/1,
          dist_parallel_sender/3, dist_parallel_receiver/0,
          derr_run/1,
          dist_evil_parallel_receiver/0, make_busy/2]).
@@ -168,10 +165,10 @@ ping(Config) when is_list(Config) ->
 
     %% Pings another node.
 
-    {ok, OtherNode} = start_node(distribution_SUITE_other),
+    {ok, Peer, OtherNode} = ?CT_PEER(),
     io:format("Pinging ~s (assumed to exist)", [OtherNode]),
     test_server:do_times(Times, fun() -> pong = net_adm:ping(OtherNode) end),
-    stop_node(OtherNode),
+    peer_stop(Peer, OtherNode),
 
     %% Pings our own node many times.
 
@@ -183,54 +180,44 @@ ping(Config) when is_list(Config) ->
 
 %% Test erlang:group_leader(_, ExternalPid), i.e. DOP_GROUP_LEADER
 group_leader(Config) when is_list(Config) ->
-    ?Line Sock = start_relay_node(group_leader_1, []),
-    ?Line Sock2 = start_relay_node(group_leader_2, []),
-    try
-        ?Line Node2 = inet_rpc_nodename(Sock2),
-        ?Line {ok, ok} = do_inet_rpc(Sock, ?MODULE, group_leader_1, [Node2])
-    after
-        ?Line stop_relay_node(Sock),
-        ?Line stop_relay_node(Sock2)
-    end,
-    ok.
+    {ok, Peer1, Node1} = ?CT_PEER(#{connection => 0, args => ["-setcookie", "NONE"]}),
+    {ok, Peer2, Node2} = ?CT_PEER(#{connection => 0, args => ["-setcookie", "NONE"]}),
+    pang = net_adm:ping(Node1), pang = net_adm:ping(Node2), %% extra check, may be skipped
+    ok = peer:call(Peer1, ?MODULE, group_leader_1, [Node2]),
+    peer:stop(Peer1), %% verify_nc() will fail because peer is not dist-connected
+    peer:stop(Peer2).
 
 group_leader_1(Node2) ->
-    ?Line ExtPid = spawn(Node2, fun F() ->
+    ExtPid = spawn(Node2, fun F() ->
                                         receive {From, group_leader} ->
                                                 From ! {self(), group_leader, group_leader()}
                                         end,
                                         F()
                                 end),
-    ?Line GL1 = self(),
-    ?Line group_leader(GL1, ExtPid),
-    ?Line ExtPid ! {self(), group_leader},
-    ?Line {ExtPid, group_leader, GL1} = receive_one(),
+    GL1 = self(),
+    group_leader(GL1, ExtPid),
+    ExtPid ! {self(), group_leader},
+    {ExtPid, group_leader, GL1} = receive_one(),
 
     %% Kill connection and repeat test when group_leader/2 triggers auto-connect
-    ?Line net_kernel:monitor_nodes(true),
-    ?Line net_kernel:disconnect(Node2),
-    ?Line {nodedown, Node2} = receive_one(),
-    ?Line GL2 = spawn(fun() -> dummy end),
-    ?Line group_leader(GL2, ExtPid),
-    ?Line {nodeup, Node2} = receive_one(),
-    ?Line ExtPid ! {self(), group_leader},
-    ?Line {ExtPid, group_leader, GL2} = receive_one(),
+    net_kernel:monitor_nodes(true),
+    net_kernel:disconnect(Node2),
+    {nodedown, Node2} = receive_one(),
+    GL2 = spawn(fun() -> dummy end),
+    group_leader(GL2, ExtPid),
+    {nodeup, Node2} = receive_one(),
+    ExtPid ! {self(), group_leader},
+    {ExtPid, group_leader, GL2} = receive_one(),
     ok.
 
 %% Test optimistic distribution flags toward pending connections (DFLAG_DIST_HOPEFULLY)
 optimistic_dflags(Config) when is_list(Config) ->
-    ?Line Sender = start_relay_node(optimistic_dflags_sender, []),
-    ?Line Echo = start_relay_node(optimistic_dflags_echo, []),
-    try
-        ?Line {ok, ok} = do_inet_rpc(Echo, ?MODULE, optimistic_dflags_echo, []),
-
-        ?Line EchoNode = inet_rpc_nodename(Echo),
-        ?Line {ok, ok} = do_inet_rpc(Sender, ?MODULE, optimistic_dflags_sender, [EchoNode])
-    after
-        ?Line stop_relay_node(Sender),
-        ?Line stop_relay_node(Echo)
-    end,
-    ok.
+    {ok, PeerSender, _Sender} = ?CT_PEER(#{connection => 0, args => ["-setcookie", "NONE"]}),
+    {ok, PeerEcho, Echo} = ?CT_PEER(#{connection => 0, args => ["-setcookie", "NONE"]}),
+    ok = peer:call(PeerEcho, ?MODULE, optimistic_dflags_echo, []),
+    ok = peer:call(PeerSender, ?MODULE, optimistic_dflags_sender, [Echo]),
+    peer:stop(PeerSender),
+    peer:stop(PeerEcho).
 
 optimistic_dflags_echo() ->
     P = spawn(fun F() ->
@@ -245,25 +232,25 @@ optimistic_dflags_echo() ->
     ok.
 
 optimistic_dflags_sender(EchoNode) ->
-    ?Line net_kernel:monitor_nodes(true),
+    net_kernel:monitor_nodes(true),
 
     optimistic_dflags_do(EchoNode, <<1:1>>),
     optimistic_dflags_do(EchoNode, fun lists:map/2),
     ok.
 
 optimistic_dflags_do(EchoNode, Term) ->
-    ?Line {optimistic_dflags_echo, EchoNode} ! {self(), Term},
-    ?Line {nodeup, EchoNode} = receive_one(),
-    ?Line {EchoPid, Term} = receive_one(),
+    {optimistic_dflags_echo, EchoNode} ! {self(), Term},
+    {nodeup, EchoNode} = receive_one(),
+    {EchoPid, Term} = receive_one(),
     %% repeat with pid destination
-    ?Line net_kernel:disconnect(EchoNode),
-    ?Line {nodedown, EchoNode} = receive_one(),
-    ?Line EchoPid ! {self(), Term},
-    ?Line {nodeup, EchoNode} = receive_one(),
-    ?Line {EchoPid, Term} = receive_one(),
+    net_kernel:disconnect(EchoNode),
+    {nodedown, EchoNode} = receive_one(),
+    EchoPid ! {self(), Term},
+    {nodeup, EchoNode} = receive_one(),
+    {EchoPid, Term} = receive_one(),
 
-    ?Line net_kernel:disconnect(EchoNode),
-    ?Line {nodedown, EchoNode} = receive_one(),
+    net_kernel:disconnect(EchoNode),
+    {nodedown, EchoNode} = receive_one(),
     ok.
 
 
@@ -281,13 +268,13 @@ bulk_send(Terms, BinSize) ->
     ct:timetrap({seconds, 30}),
 
     io:format("Sending ~w binaries, each of size ~w K", [Terms, BinSize]),
-    {ok, Node} = start_node(bulk_receiver),
+    {ok, Peer, Node} = ?CT_PEER(),
     Recv = spawn(Node, erlang, apply, [fun receiver/2, [0, 0]]),
     Bin = binary:copy(<<253>>, BinSize*1024),
     Size = Terms*size(Bin),
     {Elapsed, {Terms, Size}} = test_server:timecall(?MODULE, sender,
                                                     [Recv, Bin, Terms]),
-    stop_node(Node),
+    peer_stop(Peer, Node),
     {comment, integer_to_list(round(Size/1024/max(1,Elapsed))) ++ " K/s"}.
 
 sender(To, _Bin, 0) ->
@@ -321,7 +308,7 @@ bulk_sendsend2(Terms, BinSize, BusyBufSize) ->
 
     io:format("\nSending ~w binaries, each of size ~w K",
               [Terms, BinSize]),
-    {ok, NodeRecv} = start_node(bulk_receiver),
+    {ok, RecvPeer, NodeRecv} = ?CT_PEER(),
     Recv = spawn(NodeRecv, erlang, apply, [fun receiver/2, [0, 0]]),
     Bin = binary:copy(<<253>>, BinSize*1024),
 
@@ -334,8 +321,7 @@ bulk_sendsend2(Terms, BinSize, BusyBufSize) ->
     %% default busy size and "+zdbbl 5", and if the 5 case gets
     %% "many many more" monitor messages, then we know we're working.
 
-    {ok, NodeSend} = start_node(bulk_sender, "+zdbbl " ++
-                                    integer_to_list(BusyBufSize)),
+    {ok, SendPeer, NodeSend} = ?CT_PEER(["+zdbbl", integer_to_list(BusyBufSize)]),
     _Send = spawn(NodeSend, erlang, apply,
                   [fun sendersender/4, [self(), Recv, Bin, Terms]]),
     {Elapsed, {_TermsN, SizeN}, MonitorCount} =
@@ -347,8 +333,8 @@ bulk_sendsend2(Terms, BinSize, BusyBufSize) ->
             {sendersender, BigRes} ->
                 BigRes
         end,
-    stop_node(NodeRecv),
-    stop_node(NodeSend),
+    peer_stop(RecvPeer, NodeRecv),
+    peer_stop(SendPeer, NodeSend),
     {round(SizeN/1024/Elapsed), MonitorCount}.
 
 %% Sender process to be run on a slave node
@@ -449,7 +435,7 @@ receiver2(Num, TotSize) ->
 %% Test that link/1 to a busy distribution port works.
 link_to_busy(Config) when is_list(Config) ->
     ct:timetrap({seconds, 60}),
-    {ok, Node} = start_node(link_to_busy),
+    {ok, Peer, Node} = ?CT_PEER(),
     Recv = spawn(Node, erlang, apply, [fun sink/1, [link_to_busy_sink]]),
 
     Tracer = case os:getenv("TRACE_BUSY_DIST_PORT") of
@@ -475,7 +461,7 @@ link_to_busy(Config) when is_list(Config) ->
     do_busy_test(Node, fun () -> tail_applied_linker(Recv) end),
 
     %% Done.
-    stop_node(Node),
+    peer_stop(Peer, Node),
     stop_busy_dist_port_tracer(Tracer),
     ok.
 
@@ -495,7 +481,7 @@ tail_applied_linker(Pid) ->
 %% Test that exit/2 to a busy distribution port works.
 exit_to_busy(Config) when is_list(Config) ->
     ct:timetrap({seconds, 60}),
-    {ok, Node} = start_node(exit_to_busy),
+    {ok, Peer, Node} = ?CT_PEER(),
 
     Tracer = case os:getenv("TRACE_BUSY_DIST_PORT") of
                  "true" -> start_busy_dist_port_tracer();
@@ -545,7 +531,7 @@ exit_to_busy(Config) when is_list(Config) ->
     end,
 
     %% Done.
-    stop_node(Node),
+    peer_stop(Peer, Node),
     stop_busy_dist_port_tracer(Tracer),
     ok.
 
@@ -664,7 +650,7 @@ sink1() ->
 %% Test that EXIT and DOWN messages send to another node are not lost if
 %% the distribution port is busy.
 lost_exit(Config) when is_list(Config) ->
-    {ok, Node} = start_node(lost_exit),
+    {ok, Peer, Node} = ?CT_PEER(),
 
     Tracer = case os:getenv("TRACE_BUSY_DIST_PORT") of
                  "true" -> start_busy_dist_port_tracer();
@@ -717,7 +703,7 @@ lost_exit(Config) when is_list(Config) ->
 
     %% Done.
     stop_busy_dist_port_tracer(Tracer),
-    stop_node(Node),
+    peer_stop(Peer, Node),
     ok.
 
 dummy_waiter() ->
@@ -730,9 +716,7 @@ dummy_waiter() ->
 %% AND that the link is teared down.
 link_to_dead(Config) when is_list(Config) ->
     process_flag(trap_exit, true),
-    {ok, Node} = start_node(link_to_dead),
-    %    monitor_node(Node, true),
-    net_adm:ping(Node), %% Ts_cross_server workaround.
+    {ok, Peer, Node} = ?CT_PEER(),
     Pid = spawn(Node, ?MODULE, dead_process, []),
     receive
     after 5000 -> ok
@@ -749,13 +733,14 @@ link_to_dead(Config) when is_list(Config) ->
     {links, Links} = process_info(self(), links),
     io:format("Pid=~p, links=~p", [Pid, Links]),
     false = lists:member(Pid, Links),
-    stop_node(Node),
     receive
         Message ->
             ct:fail({unexpected_message, Message})
     after 3000 ->
               ok
     end,
+    process_flag(trap_exit, false),
+    peer_stop(Peer, Node),
     ok.
 
 dead_process() ->
@@ -767,18 +752,24 @@ link_to_dead_new_node(Config) when is_list(Config) ->
     process_flag(trap_exit, true),
 
     %% Start the node, get a Pid and stop the node again.
-    {ok, Node} = start_node(link_to_dead_new_node),
+    Name = ?CT_PEER_NAME(?FUNCTION_NAME),
+    {ok, Peer, Node} = ?CT_PEER(#{name => Name}),
     Pid = spawn(Node, ?MODULE, dead_process, []),
-    stop_node(Node),
+
+    peer_stop(Peer, Node),
+    %% since exits are trapped, need to catch exiting peer
+    receive {'EXIT', Peer, normal} -> ok end,
 
     %% Start a new node with the same name.
-    {ok, Node} = start_node(link_to_dead_new_node),
+    {ok, Peer2, Node} = ?CT_PEER(#{name => Name}),
     link(Pid),
+
     receive
         {'EXIT', Pid, noproc} ->
             ok;
         Other ->
-            stop_node(Node),
+            process_flag(trap_exit, false),
+            peer_stop(Peer2, Node),
             ct:fail({unexpected_message, Other})
     after 5000 ->
               ct:fail(nothing_received)
@@ -788,13 +779,16 @@ link_to_dead_new_node(Config) when is_list(Config) ->
     {links, Links} = process_info(self(), links),
     io:format("Pid=~p, links=~p", [Pid, Links]),
     false = lists:member(Pid, Links),
-    stop_node(Node),
     receive
         Message ->
+            process_flag(trap_exit, false),
             ct:fail({unexpected_message, Message})
     after 3000 ->
               ok
     end,
+    %% stop trapping exits, and let the peer stop
+    process_flag(trap_exit, false),
+    peer_stop(Peer2, Node),
     ok.
 
 %% Test that sending a port or reference to another node and back again
@@ -803,13 +797,13 @@ ref_port_roundtrip(Config) when is_list(Config) ->
     process_flag(trap_exit, true),
     Port = make_port(),
     Ref = make_ref(),
-    {ok, Node} = start_node(ref_port_roundtrip),
+    {ok, Peer, Node} = ?CT_PEER(),
     net_adm:ping(Node),
     Term = {Port, Ref},
     io:format("Term before: ~p", [show_term(Term)]),
     Pid = spawn_link(Node, ?MODULE, roundtrip, [Term]),
     receive after 5000 -> ok end,
-    stop_node(Node),
+    peer_stop(Peer, Node),
     receive
         {'EXIT', Pid, {Port, Ref}} ->
             io:format("Term after: ~p", [show_term(Term)]),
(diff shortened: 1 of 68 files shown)


max-au:
Future work:
 - add a convenience macro to `ct.hrl` for common usage pattern:
 ```
-define(CT_PEER(Args), peer:start_link(#{
    name => peer:random_name(?MODULE_STRING ++ "__" ++ atom_to_list(?FUNCTION_NAME)),
    args => ["-pa", filename:dirname(code:which(?MODULE)) | Args]})
).
```
- sweep test suites again to switch to this macro when appropriate
- deprecate `ct_slave.erl` (fortunately it's almost unused and can easily be swapped out)
- replace a few more implementations of peer node (e.g. one in distribution_SUITE, long_timer_test. sensitive_SUITE, erlexec_SUITE, ethread_SUITE, upgrade_SUITE, ts_run, emem_SUITE and likely more)
- replace 'slave' in comments, atoms, string constants
-

File: system/doc/general_info/DEPRECATIONS [hunk 1]
```diff
@@ -17,6 +17,11 @@
 # is scheduled to be removed in OTP 25.
 #
 
+#
+# Added in OTP 25.
+#
+slave:_/_ since=25 remove=26
```
bjorng, line 23: The OTP Technical Board will have 

```suggestion
slave:_/_ since=25 remove=27
```

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,319 @@
...
+  <modulesummary>Start and control linked Erlang nodes.
+  </modulesummary>
+  <description>
+    <p>This module provides functions for starting linked Erlang nodes.
+      When origin node (the one started linked peers) terminates, all
+      linked peers are also terminated.
+      Depending on the start mode, terminal output and file I/O
+      of linked peers may go through the origin.</p>
+
+    <p>Peer nodes on other hosts than the current one are started with
```
bjorng, line 44: This seems to be contradicted later in that `rsh` is the default. In my opinion, support for `rsh` should be removed, but the OTB will have to decide.

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,319 @@
...
+  </header>
+  <module since="OTP 25.0">peer</module>
+  <modulesummary>Start and control linked Erlang nodes.
+  </modulesummary>
+  <description>
+    <p>This module provides functions for starting linked Erlang nodes.
+      When origin node (the one started linked peers) terminates, all
+      linked peers are also terminated.
+      Depending on the start mode, terminal output and file I/O
+      of linked peers may go through the origin.</p>
```
bjorng, line 42: Does "start mode" refer to the `connection` option? This should be clarified.

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,319 @@
...
+        controlling process on the origin node exits with corresponding reason, effectively
+        providing two-way link.</p>
+        <p>When <c>connection</c> is set to a port number, origin starts listening on
+          the requested TCP port, and peer node connects to the port. When it is set to
+          an <c>{IP, Port}</c> tuple, origin listens only on the specified IP. Port can
+          be set to 0 for automatic selection.
+        </p>
+        <p>Using <c>standard_io</c> alternative connection starts the peer attached to
+          the origin (other connections use <c>-detached</c> flag to erl). In this mode
+          peer and origin communicate via stdin/stdout.
```
bjorng, line 161: Is this similar to using the `slave` module?

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,319 @@
...
+      <name name="connection"/>
+      <desc><p>Alternative connection between the origin and the peer. When the connection
+        closes, peer node is terminated. If <c>report_reason</c> startup flag is set to true,
+        controlling process on the origin node exits with corresponding reason, effectively
+        providing two-way link.</p>
+        <p>When <c>connection</c> is set to a port number, origin starts listening on
+          the requested TCP port, and peer node connects to the port. When it is set to
+          an <c>{IP, Port}</c> tuple, origin listens only on the specified IP. Port can
+          be set to 0 for automatic selection.
+        </p>
```
bjorng, line 158: Would this be similar to using a peer node started by `test_server:start_node()` with the `peer` option?

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,319 @@
...
+            </p>
+            <p>
+              Useful for testing backwards compatibility with previous releases,
+              or when Erlang installation location is missing from the <c>PATH</c>,
+              or when starting peers on a host with different operating system.
+            </p>
+          </item>
+          <tag><c>connection</c></tag>
+          <item>
+            <p>Alternative connection specification.</p>
```
bjorng, line 117: What is the default value?

lib/stdlib/test/dets_SUITE.erl
```diff
@@ -376,9 +376,9 @@ dirty_mark(Config) when is_list(Config) ->
     {ok, Node} = test_server:start_node(dets_dirty_mark,
                                         slave,
                                         [{linked, false},
-                                         {args, "-pa " ++
+                                         {args, ["-pa",
                                               filename:dirname
-						(code:which(?MODULE))}]),
+						(code:which(?MODULE))]}]),
```
bjorng, line 381: Why isn't this call replaced with a call to `peer`? Isn't the `peer` module powerful enough?

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,319 @@
...
+          </p>
+        </note>
+      </desc>
+    </func>
+
+    <func>
+      <name name="start" arity="1" since="OTP 25.0"/>
+      <fsummary>Start a peer node.</fsummary>
+      <desc>
+        <p>Starts a peer node with specified <seetype marker="#start_options"><c>start_options()</c></seetype>.</p>
```
bjorng, line 278: I think we will need a short summary of the start options. Which are mandatory? What are the defaults for the others? Will the behavior be similar to the `slave` module or a peer node started with `test_server:start_node()`?

bjorng (commented):
Nice initiative for cleaning up the existing mess of different way to start peer and slave nodes. I have some questions:

* Is `peer` intended to replace both `slave` and peer nodes as started by `test_server:start_node()` with the `peer` option? (The name `peer` would suggest that.)
* If so, why are there still calls to `test_server:start_node()` in the test suites? Is the current implementation of `peer` not powerful enough yet or is there some other reason?

I have some comments:

* There is no reason to update the primary bootstrap with `peep.beam`. Please remove it. We have a policy that only OTP team members should do such updates since it is difficult to review the changes in a binary file.
* I think that the documentation should have some examples on how to use `peer` in different scenarios (e.g. similar to the old `slave` module and as a peer node and so on). The interaction between the different start options is not obvious.

lib/stdlib/doc/src/peer.xml
max-au: Yes, `connection` option. Let me clarify this behaviour.

lib/stdlib/doc/src/peer.xml
max-au: Default (when no option is specified) is Erlang Distribution connection only, just like existing `slave` implementation.

lib/stdlib/test/dets_SUITE.erl
max-au: I (incorrectly) assumed that `test_server:start_node` is used by OTP test suites to channel `start_node` calls through some Test Server that runs somewhere else, not on the machine where test cases are actually executed. 

From my reading of `test_server:start_node`, it calls group leader (which may reside anywhere in the connected cluster) to start a new ERTS instance. So technically it is possible that test cases are running on `runner@host_a`, but group leader is running on node `test_server@host_b`. In this case new nodes will be started on `host_b`. In my change I retained this behaviour.

If OTP test cases are always running in the same host with `test_server`, it significantly simplifies everything! And lets me to simply use `peer:start_link` instead of `test_server`. Would you be ok with this approach? If yes, let me just replace all `test_server:start_node` with `peer:start_link`, it will be so much easier to debug.

lib/stdlib/doc/src/peer.xml
max-au: All options are optional. I will make a few clarifications regarding the behaviour. In short, when options do not contain alternative connection specification, behaviour is similar to `slave` (because Erlang Distribution connection is providing the heartbeat, as soon as it drops, peer terminates).
When alternative `connection`, behaviour is similar to `test_server:start_node(peer)`.

lib/stdlib/doc/src/peer.xml
max-au: That'd be great if `rsh` support can be dropped. However I will still keep some way to amend/process path to executable and command line, because it's needed to support virtualised environments. I will provide examples in the documentation demonstrating intended usage, and add corresponding test cases. In short, instead of ssh host_b erl -name node2`, I use it to run `docker run -it otp -name node_in_docker`.

lib/stdlib/test/dets_SUITE.erl
bjorng: > If OTP test cases are always running in the same host with test_server, it significantly simplifies everything! 

Yes, that should be the case as far as I know. Some of the code you saw probably remains from the days when we supported VxWorks. Only one Erlang node could run on each VxWorks host, so all slave nodes were remote.

However, when discussing this PR with @garazdawi, he pointed out that `test_server:start_node()` sets up locations for crash dumps so that they can be copied to the test case log. Perhaps we should invent a new module `ct_peer` that is a thin wrapper on top of `peer` that can handle such details. Your suggested `?CT_PEER` macro could call `ct_peer`. (I like your `?CT_PEER` macro because it invents a node name that is unlikely to collide with other node names; we have had a lot of problems with nodes remaining after a failed test case and other test cases that used the same name also failed.)

max-au:
Quick update, I am re-working this PR with better documentation and improved existing code compatibility. I managed to get rid of implicit "RSH" parameter, and fix "old releases" compatibility issues (since it's not documented how to test "22_latest" or "24_latest" releases, I had to do some reverse engineering first).

max-au:
I reworked the PR, keeping 4 separate commits for easier review:
 - peer node implementation 8bc41af - adds new module to stdlib application, contains documentation, code and tests. Existing code is not changed in any way.
 -  ?CT_PEER macro b3cd13e - convenience primitives mentioned in "peer" module documentation (with usage examples)
 - emulator: replace test_server:start_node with ?CT_PEER 65f2ef0 - this commit clean up test cases for the "emulator" application, replacing all test_server:start_node invocations and removing ad-hoc "slave" implementations from distribution_SUITE, driver_SUITE. Essentially it is a cleanup patch keeping existing logic intact. The intent is to demonstrate migration
 - slave module deprecation 13692f4 - no binary files checked in

Several changes and decisions made from the original iteration:
* no binaries checked-in (peer.beam, otp_internal.beam are to be generated by the OTP team)
* explicit `rsh` support has been removed from `peer`, but examples section demonstrates how it's possible
* documentation updated with test cases
* only one application (emulator) has been re-worked to provide a demo/proof-of-concept

If this approach is accepted, I will:
* convert the remaining applications (kernel and stdlib require some efforts, they do contain quite a number of ad-hoc peer node implementations)
* remove `test_server:start_node` and all related code 

Let me know if you want this in the same PR, or a separate "cleanup" PR to facilitate reviews.

===

I also have number of questions regarding existing `slave` usage:

1. `pool.erl` uses it to start extra nodes on a number of hosts. I wonder if there are many customers using `pool`, or it can also be deprecated and removed together with `slave`. Some context; `pool:start/2` accepts a concatenated command line. It is not easy to do given variations in escaping rules on different OSes. So `peer` accepts a list of command line args instead. If `pool` cannot be deprecated altogether, I'd suggest to deprecate `pool:start/2` accepting command line, and add a new `pool:start_args/2` that accepts a list of command line args. Until the decision is made, I put a suppression into `pool` to not warn about deprecated `slave` module usage.
2. `eunit_data` uses `slave:start/2` with concatenated command line. I am not aware of any method to warn  on that one. Potentially I can make a regex to split the supplied command line back to a list of arguments, but I'd appreciate if there is a way to avoid OS-dependent quoting in eunit as well.

bjorng:
Nice updates!

The `test_server` and `test_server_node` modules will need suppressions for use of the `slave` modul; otherwise the build will fail when we'll update the primary bootstrap. The `asn1_SUITE` module in the ASN.1 test suite also needs suppressions (or `slave` to be replaced with `peer`).

Otherwise, it looks good so far I have looked. We will take a closer look, but I have already added the branch to our daily builds (with suppressions in the modules mentioned above and an updated primary bootstrap).

max-au:
@bjorng thanks for kind words!
Indeed I missed `test_server_node`, as I avoided building deprecations and bootstrap. Fixed and force-pushed!

I also pushed the update `asn1` tests, following `emulator` pattern, and refactored two `diameter` previously using `ct_slave` to ensure that a new `peer` module is indeed capable of replacing all variants of slave/peer node.

bjorng:
Here are some results from our daily builds:

The test `peer_SUITE:old_release/1` fails on most platforms. It seems that the test case expects `none` when no old release is available, but the actual value is `not_available`.

The test case `driver_SUITE: lots_of_used_fds_on_boot/1` fails on many platforms, including on my own Mac. Here is part of the log:

```
=== Location: [{peer,start_it,453},
              {driver_SUITE,lots_of_used_fds_on_boot_test,1880},
              {test_server,ts_tc,1784},
              {test_server,run_test_case_eval1,1293},
              {test_server,run_test_case_eval,1225}]
```

And then there are many failed test case in `peer_SUITE` that fails on Windows and FreeBSD (12.2 and 13.0). Specifically, on those platforms, all test cases in the `tcp` group fail. There are also some failures in the `standard_io` group; on Windows, the following test cases fail: `init_debug`, `basic`, `detached`, `io_redirect`, and `duplicate_name`; on FreeBSD, `duplicate_names` fails. The `init_per_group` test case for the `remote` group also fails. Most of them fails with log similar to the on shown above. The test cases in the `dist` group succeed.

max-au:
I miserably failed to set up Windows build, official build guide let me progress to "configure" step when "chmod" was failing (due to NTFS on /mnt/c/users/max-au/source/repos/otp not supporting Linux permissions). So I tried to perform the setup that github workflow follows, but it failed "make" step. 
Hence I could not reproduce Windows failures. But I managed to figure out a way to reproduce `tcp` group failures, and fix was quite trivial. I force-pushed the updated variant which definitely passed FreeBSD 12 and 13.

I could not make `driver_SUITE: lots_of_used_fds_on_boot` to fail, likely that fixes I made to prefer localhost IP removed that failure mode.

lib/common_test/src/test_server_node.erl
```diff
@@ -390,6 +390,9 @@ start_node_slave(SlaveName, OptList, From, _TI) ->
 	end,
     gen_server:reply(From,Ret).
 
+%% Temporary suppression, to avoid a warning calling undocumented
+%%  but deprecated function.
+-compile([{nowarn_deprecated_function,[{slave,start,5}]}]).
```
bjorng, line 395:
```suggestion
-compile([{nowarn_deprecated_function,[{slave,start,5},{slave,stop/1}]}]).
```

bjorng:
Yes, building OTP for Windows can be tricky.

Your branch does not have any suppression for `slave:stop/1` in the `test_server` module.

I have added your updated branch (with the missing suppression added and updated primary bootstrap) to our daily builds.

bjorng:
@max-au I have added a section to the Wiki page on how to run tests with `cover`: https://github.com/erlang/otp/wiki/Running-tests#running-with-the-cover-tool

bjorng:
`driver_SUITE:lots_of_used_fds_on_boot/1` fails for me when it is not possible to open more than 1024 files at once. When there are insufficient file descriptors, the following lines are meant to skip the test case:

https://github.com/erlang/otp/blob/1ed066cb89807ef6dcefe56c1fda5e00e1d884e8/erts/emulator/test/driver_SUITE.erl#L1906-L1907

With your rewritten test case, there will be a timeout and the test case will fail.

I can think of several ways of fixing the test case:

* Enhancing the `peer` module to allow retrieving the exit code.
* Modifying the wrapper to communicate the inability to create many files in another way, for example by writing to a file.
* Restoring the original test case code.

max-au:
Now I understand that I missed the case when `ulimit` was set to a small amount, which should just skip the case instead of failing.
`peer` supports exit code retrieval, I should document that (will do in the next iteration). I fixed `lots_of_used_fds_on_boot` to skip the test if wrapper binary exits with code 17, as you suggested.

lib/stdlib/src/peer.erl
```diff
@@ -0,0 +1,846 @@
...
+        _ ->
+            %% if progname is not known, use `erlexec` from the same ERTS version we're currently running
+            %% BINDIR environment variable is already set
+            %% EMU variable it also set
+            Root = code:root_dir(),
+            Erts = filename:join(Root, lists:concat(["erts-", erlang:system_info(version)])),
+            BinDir = filename:join(Erts, "bin"),
+            {filename:join(BinDir, "erlexec"), []}
+    end.
+
```
bjorng, line 670: Here is hack to handle our special builds such as address sanitizer, gcov and so on. It has been tested in our daily builds.

```suggestion
            find_executable_and_args(Prog);
        _ ->
            %% if progname is not known, use `erlexec` from the same ERTS version we're currently running
            %% BINDIR environment variable is already set
            %% EMU variable it also set
            %% (NOTE: this will not work in a source code tree)
            Root = code:root_dir(),
            Erts = filename:join(Root, lists:concat(["erts-", erlang:system_info(version)])),
            BinDir = filename:join(Erts, "bin"),
            {filename:join(BinDir, "erlexec"), []}
    end.

find_executable_and_args(Prog) ->
    case os:find_executable(Prog) of
        false ->
            case binary:split(list_to_binary(Prog), <<"/cerl ">>) of
                [Prefix, Arg] ->
                    %% This is a hack to handle the 'cerl' script used
                    %% by the Erlang/OTP test suites. When 'cerl'
                    %% starts the runtime system, it typically sets
                    %% 'progname' to the path of the 'cerl' script,
                    %% followed by an argument. For example:
                    %%
                    %%     /<full_path_to>/cerl -debug
                    %%     /<full_path_to>/cerl -asan
                    %%     /<full_path_to>/cerl -gcov
                    %%
                    %% We should find a better way to handle this, for
                    %% example by passing the emulator type and flavor
                    %% using the '-emu_type' and '-emu_flavor'
                    %% options. However, this is not without
                    %% complications as those option does not really
                    %% work for an installed system. Also, it is
                    %% probably a good idea to stop using 'slave'
                    %% before attempting to do this.

                    {binary_to_list(Prefix) ++ "/cerl", [binary_to_list(Arg)]};
                _ ->
                    %% FIXME: Handle this error in a better way.
                    error(failed_to_find_executable)
            end;
        Executable ->
            {Executable, []}
    end.
```

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -0,0 +1,459 @@
...
+    peer:stop(Peer).
+
+%% -------------------------------------------------------------------
+%% Compatibility: old releases
+old_release() ->
+    [{doc, "Verity running with previous OTP release"}].
+
+old_release(Config) when is_list(Config) ->
+    PrivDir = proplists:get_value(priv_dir, Config),
+    ?assertEqual(not_available, ?CT_PEER([], "1", PrivDir)),
```
bjorng, line 387: A bug in `test_server_node` results in the pattern `"1"` finding an OTP 19 release (which will not work because it does not have `is_map_key/2`). I don't have to go down the rabbit hole of fixing bus in `test_server_node`, so here is a workaround that avoids the bug: 

```suggestion
    ?assertEqual(not_available, ?CT_PEER([], "10", PrivDir)),
```

erts/emulator/test/distribution_SUITE.erl
```diff
@@ -2543,43 +2420,47 @@ dmsg_bad_tag() ->  %% Will fail early at heap size calculation
...
 
 no_epmd(Config) when is_list(Config) ->
     %% Trying to start a node with -no_epmd but without passing the
     %% --proto_dist option should fail.
-    {error, timeout} = start_node(no_epmd, "-no_epmd").
+    try
+        ?CT_PEER(#{connection => standard_io, args => ["-no_epmd"]}),
+        ct:fail(unexpected_no_epmd_start)
+    catch
+        exit:{boot_failed, normal} ->
```
bjorng, line 2437: This test failed started to failed after your last update:

```suggestion
        exit:{boot_failed, {exit_status, 1}} ->
```

bjorng (commented):
Here comes more suggested fixes to eliminate failed test cases.

max-au:
Thank you!
I updated PR with following changes:
 - fixed distribution_SUITE and peer_SUITE with changes you suggested
 - updated peer.xml documentation to mention `exit_code` support
 - supported `$ERL_TOP/bin/cerl -debug -another_option`

For the latter, since it's possible that multiple options are passed to `cerl`, I added a small command line parser to correctly split the line into list of argv. It is however arguable if all these args should be passed back to `cerl`, or only selected. For now I went with all arguments, but happy to change this.

bjorng:
I will come back with more information about failed test cases, but here are the test cases in `peer_SUITE` that always fails on Windows: `basic`, `io_redirect`, `detached`.

max-au:
I reproduced 3 mentioned failures with peer_SUITE on Windows. Root cause puzzled me a bit. Turns out that on Windows sending `erlang:port_command(Port, <<13>>)` is somehow transformed into message `{Port, {data, <<10>>}}`. Meaning, CR is replaced with LF.
I was not expecting this behaviour.

Now I pushed the updated implementation which uses the same encoding rules for incoming and outgoing stdin/stdout communications, and it seems to pass Windows tests on my machine.

bjorng:
> Interesting catch. How do I reproduce it? The fix is trivial, but I need some way to confirm it.

First build a special runtime system, for example:

    make emulator TYPE=debug

Then start the emulator like this:

    ERL_AFLAGS="-emu_type debug" $ERL_TOP/bin/erl

> I also found that "cover" reporting does not work correctly with `peer`. I am adding automatic "cover" startup to `?CT_PEER` macro, but the real problem comes from `cover` itself which wants `cover:flush` to be executed on a cover master node. When slave nodes are stopped via `test_server:stop_node`, there is explicit call. But peer nodes halt when linked process exits, and I can't put the call in `peer` itself. One way to solve it would be to get back to stopping peers via `test_server:stop_node`, but I'd rather prefer this to be automatically done by the `cover` itself somehow.

It would probably be possible to implement functionality for the remote cover nodes to initiate a push of its collected data to the main cover node.

Assuming that it is possible, could it be arranged so that `peer` will notify the cover process running on the remote node so that it could push its data before the node goes down?

lib/stdlib/src/peer.erl
```diff
@@ -0,0 +1,909 @@
...
+                    erlang:demonitor(Mref, [flush]),
+                    erlang:exit({boot_failed, Reason});
+                {'DOWN', Mref, _, _, Reason} ->
+                    erlang:exit(Reason)
+            after
+                WaitBoot ->
+                    _ = gen_server:stop(Pid),
+                    erlang:demonitor(Mref, [flush]),
+                    receive
+                        {[alias | Mref], Reply} ->
```
rickard-green, line 455: This `receive` clause should never be able to match anything since `Mref` isn't communicated with any other processes (unless `Mref` should be picked up via `process_info()` which would be a bit strange). Since `Mref` is not communicated to anyone, there is also no reason to make it an alias since noone will be able to use it as an alias.

max-au:
I updated the PR with following changes:
 - addressed @bjorng comment about unsetting ERL_AFLAGS (also ERL_FLAGS, kept ERL_ZFLAGS however)
 - addressed @rickard-green comment on unnecessary `alias` usage (leftovers from early implementation)
 - added a (horrible) hack to keep `cover` test coverage support when peers are started via `?CT_PEER` (will be looking further into `cover` implementation, I think it can be significantly improved).

I also have a patch that converts `common_test` usage of `ct_slave` into `?CT_PEER`. I think for now it might be better to keep it a separate PR. To address @rickard-green  question, I do appreciate any help with peerifying, especially for old or complex suites (like erl_distribution_SUITE and SASL tests). Ideally I'd like to replace all mentions of `slave` with `peer` before the next major release, and clean up `test_server` to remove old `start_node` code.

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -0,0 +1,461 @@
...
+%% -------------------------------------------------------------------
+%% Compatibility: old releases
+old_release() ->
+    [{doc, "Verity running with previous OTP release"}].
+
+old_release(Config) when is_list(Config) ->
+    PrivDir = proplists:get_value(priv_dir, Config),
+    %% don't expect OTP 10 to be available
+    ?assertEqual(not_available, ?CT_PEER([], "10", PrivDir)),
+    PrevRel = "22",
```
bjorng, line 390: `rpc` in OTP 25 will not support communicating with OTP 22 nodes. @rickard-green suggest avoiding hard-coding OTP release numbers and instead calculating it as the current release minus 2:

```suggestion
    PrevRel = integer_to_list(list_to_integer(erlang:system_info(otp_release)) - 2),
```

max-au:
Makes sense. Pushed the update. However there are at least 20 more callsites that would need to be updated. Pretty much every single one that uses old release has hardcoded "22". I can sweep those in my diff too, while peerifying.

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -0,0 +1,461 @@
...
+dist_up_down() ->
+    [{doc, "Test that Erlang distribution can go up and down (with TCP alternative)"}].
+
+dist_up_down(Config) when is_list(Config) ->
+    {ok, Peer, Node} = peer:start_link(#{name => peer:random_name(?FUNCTION_NAME),
+        connection => {{127, 0, 0, 1}, 0}}),
+    ?assertEqual(true, net_kernel:connect_node(Node)),
+    ?assertEqual(true, net_kernel:disconnect(Node)),
+    ?assertEqual(true, net_kernel:connect_node(Node)),
+    peer:stop(Peer).
```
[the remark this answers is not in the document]
max-au: That's good new, thank you! Will be waiting for the OTP board review decision.

bjorng:
We found some more failing test cases. To reproduce these failing test cases, you will need to combine your branch with #5307 (which will peerify some test cases). With this setup, the following test case will always fail:

```
ts:run(kernel, rpc_SUITE, called_node_dies, [batch,cover]).
```

Also, if you have an OTP 23 node installed and available, `rpc_SUITE:cast_old_against_new/1` will also fail. (It seems that the OTP 23 system for some reason attempts to load BEAM files compiled for OTP 25.)

max-au:
Pushed a tiny fix for it, now #5307 passes. The issue was, when `cover` is used, but peer node shuts down abruptly, `peer` was still attempting to fetch coverage counters. With this fix, it won't.

I have an impression that existing `cover` tool might not work well with spawned nodes, especially those not using distribution (and using alternative connection instead). Is there any reasoning/context why `cover` requires distribution to send the coverage information? Would OTP be interested in a contribution that changes `cover` so it uses file system to dump coverage counters? (also I'd really like cover to be done as compile-time parse transform, as current implementation simulates parse_transform via reading AST from debug chunks, which is somewhat slow, and has to be repeated for every spawned child node).

bjorng:
> Pushed a tiny fix for it, now #5307 passes.

Thanks.

> I have an impression that existing `cover` tool might not work well with spawned nodes, especially those not using distribution (and using alternative connection instead). Is there any reasoning/context why `cover` requires distribution to send the coverage information? Would OTP be interested in a contribution that changes `cover` so it uses file system to dump coverage counters?

Yes. It seems that it would improve the reliability and accuracy of `cover` in our daily builds. For backward compatibly reasons we would probably need to have the new way as an option and keep the old way by default (I wrote "probably" because we are slightly less concerned about backward compatibility for tools).

> (also I'd really like cover to be done as compile-time parse transform, as current implementation simulates parse_transform via reading AST from debug chunks, which is somewhat slow, and has to be repeated for every spawned child node).

Wouldn't that mean that you would have re-compile every time you want to run `cover` and re-compile again when you want normal code? That would be inconvenient. When I develop features in the compiler, I tend to alternate cover runs with non-cover runs of the compiler test suite. It would also need changes in how we do our daily builds (we use the same tar ball with pre-built BEAM files for all test runs on all hosts). Couldn't the same goal be achieved by caching the transformed code in the cover server process?

lib/stdlib/src/peer.erl
```diff
@@ -0,0 +1,906 @@
...
+    Total = <<Size:32, Bytes/binary, Crc:32>>,
+    <<<<3:2, Upper:4, 3:2, 3:2, Lower:4, 3:2>> || <<Upper:4, Lower:4>> <= Total>>.
+
+%% convert terminal codes to Erlang term, printing everything that
+%%  was detected as text
+decode_port_data(<<>>, Str, Bin) ->
+    {Str, Bin};
+decode_port_data(<<3:2, Quad:4, 3:2, Rest/binary>>, Str, Bin) ->
+    decode_port_data(Rest, Str, quad_concat(Bin, Quad, erlang:bit_size(Bin)));
+decode_port_data(<<Char:8, Rest/binary>>, Str, Bin) ->
```
bjorng, line 552: When will this happen?

I think that this protocol allowing both binary data and text to be sent deserves some comments explaining what is going on.

bjorng (commented):
Preparing for the OTB meeting today, I took a closer look at the code and wrote some comments.

Your code is not formatted using the Erlang mode for Emacs. Most of the OTP team members use Emacs. Would you mind reformatting the `peer.erl` source file in Emacs, or letting me do it?

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,538 @@
...
+          </item>
+          <tag><c>connection</c></tag>
+          <item>
+            <p>Alternative connection specification. See the
+              <seetype marker="#connection"><c>connection</c> datatype</seetype>.</p>
+          </item>
+          <tag><c>args</c></tag>
+          <item>
+            <p>Extra command line arguments to append to the "erl" command. Arguments are
+              passed as is, no escaping or quiting is needed or accepted.</p>
```
bjorng, line 305:
```suggestion
              passed as is, no escaping or quoting is needed or accepted.</p>
```

bjorng:
OTB approved this PR!

One item that come up was whether `peer` belongs in STDLIB. We discussed other possible applications such as Kernel or Tools where it might fit in better. In the end we decided that we should merge the PR as is (with `peer` in STDLIB), and perhaps move `peer` later, but before the release of OTP 25.

When you have addressed the minor issues that I commented about today,  we will run the updated PR in the daily builds and then merge, probably at the beginning of next week.

max-au:
Thanks for the good news! I updated PR with the suggestions you made, except for "emacs" formatting, which I'm not sure how to apply (so please go ahead with fixup/reformat). I added a very short description of standard_io escaping, if necessary, I can make a note of it in the documentation, and apply a few more follow-ups (ideally after merging the diff, as it will make further conversions easier, and will allow me to make a few more conversion PRs).

I also found one failure running `ts:run(stdlib, peer_SUITE, [batch, cover])`, will look into it.

bjorng:
Thanks! I have pushed two fixup commits, one with the indentation fix and one with a very minor optimization.

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,538 @@
...
+          connection => standard_io}),
+    </code>
+
+    <p>
+      The following Common Test case demonstrates Docker integration, starting two
+      containers with hostnames "one" and "two". In this example Erlang nodes
+      running inside containers form an Erlang cluster.
+    </p>
+    <code type="erl">
+      docker(Config) when is_list(Config) ->
```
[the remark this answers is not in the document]
max-au: Today I learnt! Is there any way to point erl_docgen to test suite code? Just moving the code into a separate file won't be convenient.

max-au:
I pushed the updated branch that:
 - incorporates emacs formatting fixups into original commit
 - fixes `ts:run(stdlib, peer_SUITE, [cover, batch])` test failure running old_release test case
 - incorporated @garazdawi suggestion for documentation

bjorng:
@max-au I see only a very small update to `peer.xml`. I don't see any use of `codeinclude`, or any change to `peer.erl`. Did you forget to push the latest version of the branch?

lib/stdlib/doc/src/peer.xml
```diff
@@ -0,0 +1,538 @@
...
+      these arguments</item>
+      <item>multiple test cases can run concurrently speeding up overall testing
+      process, peer node names are unique even when there are multiple instances
+      of the same test suite running in parallel</item>
+    </list>
+    <code type="erl">
+      -module(my_SUITE).
+      -behaviour(ct_suite).
+      -export([all/0, groups/0]).
+      -export([basic/1, args/1, named/1, restart_node/1, multi_node/1, old/1]).
```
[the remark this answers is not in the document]
max-au: good catch! Fixed and force-pushed.

max-au:
@bjorng I could not find an easy way to do `codeinclude` from the test suite. If my understanding is correct, it allows me to have `examples` directory in the documentation, and make references to it, but not to the actual test case. Also, I'm not sure that it will be easier to maintain, so I just learnt that it's possible but did not introduce additional `example` directory, I think it's not yet necessary.

Change fixing `cover` issue is in dd26cca (see a clause added for `cover_compiled` return by `code:which`).

@u3s Technically I don't see any issues with using `peer:start_link` instead of `CT_PEER`, the latter is convenience primitive implementing Common Test conventions. Also, `peer:stop` is just an alias for `gen:stop`. Also, it's not a big deal if `peer:stop` is not called at all, when the rest runner process terminates, all linked peers will also terminate (causing lots of logs printed, hence I prefer to have explicit `peer:stop` to avoid those crash reports).
I don't mind having `?CT_PEER_STOP` or something similar, but I don't see any value in it either. Here I'd just follow OTP team guidance, - happy to hear from @bjorng, @rickard-green, @garazdawi.

bjorng:
@max-au I think that one way to use `codeinclude` would be to put the code in an `examples` directory that would be included in the release, and the test suite could then access the code in the `examples` directory. Not sure that it is worth doing.

> I don't mind having `?CT_PEER_STOP` or something similar, but I don't see any value in it either.

I agree.

I will add the updated branch to our daily builds. If everything goes well, I will merge the branch tomorrow.

bjorng:
Thanks for your pull request!

bjorng:
I suggest organizing further peerifying of test suites into one pull request per application.

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -0,0 +1,463 @@
...
+ssh(Config) when is_list(Config) ->
+    %% figure out path to 'erl' locally ('erl' may not be in path via SSH)
+    case {os:find_executable("erl"), os:find_executable("ssh")} of
+        {false, _} ->
+            {skip, "erl not found"};
+        {ErlPath, SshPath} ->
+            Name = peer:random_name(?FUNCTION_NAME),
+            {ok, Peer, _Node} = peer:start_link(#{exec => {SshPath, ["localhost", ErlPath]},
+                connection => standard_io, name => Name,
+                args => ["-env", "BINDIR", os:getenv("BINDIR")], host => "localhost"}),
```
garazdawi, line 416: @max-au Do you recall why you needed to have the BINDIR here? This testcase fails on windows as BINDIR is not set there so I'm thinking of removing the `-env` bit of this testcase.

max-au: Ouch, I think it's an accidental leftover from my original attempt to make this test case work without any Erlang installation (when `os:get_executable("erl")` returns `false`). Initial version used a technique similar to `peer:default_erts`, starting `erlexec` with BINDIR transitively set via SSH. Sure, go ahead and remove it.

PR #5166, opened 2021-08-30
Make re copy small subbinaries 

### LINKED ISSUE #5150: The same input + regex for re:run/3 in a tight loop leads to inconsistent results
**Describe the bug**

Take this module:

```erlang
-module(trim).
-export([trim/1, sorted_loop/2]).

trim(Entry) when is_binary(Entry) -> do_trim(Entry).
do_trim(<<Rest/bits>>) -> Rest.

sorted_loop(Fun, Attempts) ->
  Results = lists:map(Fun, lists:seq(1, Attempts)),
  io:format("~p~n", [lists:usort(Results)]).
```

Now compile this module with `erlc trim.erl` and then run this:

    erl -noshell -eval 'trim:sorted_loop(fun(_) -> re:run(trim:trim(<<"EmployeeEmailAddress">>), <<"([a-z])([A-Z])">>, [global]) end, 1000000), erlang:halt().'

The expected result is this (to match on both eE and lA):

`` `erlang
[{match,[[{7,2},{7,1},{8,1}],[{12,2},{12,1},{13,1}]]}]
```

But on Erlang/OTP 23, you might find the following results:

```erlang
[{match,[[{7,2},{7,1},{8,1}]]},
 {match,[[{7,2},{7,1},{8,1}],[{12,2},{12,1},{13,1}]]}]
```

or this:

```erlang
[nomatch,
 {match,[[{7,2},{7,1},{8,1}]]},
 {match,[[{7,2},{7,1},{8,1}],[{12,2},{12,1},{13,1}]]}]
```

Which means the same inputs is leading to different results.

Some important notes:

  1.  I can reproduce it with 100_000 attempts but the example above uses a million to be sure

  2. If I remove the trim:trim/1 call, I can no longer reproduce the issue

  3. If I compile all code (instead of using -eval), I can no longer reproduce the issue

  4. Removing the global option also reproduces the bug as it returns `nomatch` in some cases

**Affected versions**

I can reproduce the bug on Erlang/OTP 23.1 but not on Erlang/OTP 24. I don't know if it is because the bug has been fixed or a side-effect, so I thought I would report it anyway.

Thank you!

garazdawi:
Because small subbinaries can be converted to heapbinaries
during GC, we do not want to have references to those in the
re state, so we copy any binary that is small enough to fit
on the heap.

Closes #5150

### DIFF
--- a/erts/emulator/beam/erl_bif_re.c
+++ b/erts/emulator/beam/erl_bif_re.c
@@ -1292,7 +1292,7 @@ re_run(Process *p, Eterm arg1, Eterm arg2, Eterm arg3)
     }
 
     /*  Optimized - if already in binary off heap, keep that and avoid
-       copying, also binary returns can be sub binaries in that case */
+        copying, also binary returns can be sub binaries in that case. */
 
     restart.flags = 0;
     if (is_binary(arg1)) {
@@ -1307,7 +1307,10 @@ re_run(Process *p, Eterm arg1, Eterm arg2, Eterm arg3)
 
 	slength = binary_size(arg1);
 	bptr = binary_val(real_bin);
-	if (bitsize != 0 || bitoffs != 0 ||  (*bptr != HEADER_PROC_BIN)) {
+	if (bitsize != 0 || bitoffs != 0 || slength <= ERL_ONHEAP_BIN_LIMIT) {
+            /* If this is an unaligned subbinary,
+               or the binary is smaller than the ERL_ONHEAP_BIN_LIMIT
+               we make a copy of the binary. */
 	    goto handle_iolist;
 	}
 	pb = (ProcBin *) bptr;


sverker:
I don't understand the assumption that re_run is making. If it is a heap binary from the start then GC will move the data anyway.

garazdawi:
hmm, that is a good point... I wonder if this has ever worked? I'll rework this fix so that pcre behaves plays nice if the location of the binary changes.

I wonder if we can then skip the GC change... what do you think @sverker?

garazdawi:
I found why it works for heap binaries: https://github.com/erlang/otp/blob/82f855a/erts/emulator/beam/erl_bif_re.c#L1303

If a heap binary is encountered, the logic that makes a copy of the subject is triggered. So maybe the solution is just to trigger that when the size of the binary may trigger a subbinar compaction.

sverker:
or bump the refc to keep it from being deallocated.

If we don't know of any yielding bifs that makes such assumption, then I personally don't think we should complicate the GC code.

bjorng (approved):

jhogberg (approved):

PR #5170, opened 2021-08-30
Fix ip_mreq spec to match implementation

fhunleth:
This changes the `address` key to `interface` which matches what is
wanted for the `add_membership` option.


RaimoNiskanen:
You are quite right!  Thank you for your pull request!

PR #5176, opened 2021-09-01
Fix Chunk Element Inline/Block Types

eksperimental:
`<br>` is an inline element, not a block level type.

### DIFF
--- a/lib/erl_docgen/doc/src/doc_storage.xml
+++ b/lib/erl_docgen/doc/src/doc_storage.xml
@@ -59,8 +59,8 @@
 -type chunk_element_attrs() :: [chunk_element_attr()].
 -type chunk_element_attr() :: {atom(),unicode:unicode_binary()}.
 -type chunk_element_type() :: chunk_element_inline_type() | chunk_element_block_type().
--type chunk_element_inline_type() :: a | code | strong | b | em | i.
--type chunk_element_block_type() :: p | 'div' | br | pre | ul |
+-type chunk_element_inline_type() :: a | code | strong | b | em | i | br.
+-type chunk_element_block_type() :: p | 'div' | pre | ul |
                                     ol | li | dl | dt | dd |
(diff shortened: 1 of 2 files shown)


garazdawi:
hmm, while you are correct from an HTML perspective, shell_docs treats `<br/>` as a block element at the moment. 

So for this fix to be correct, we would also need to change the way that the rendering and validation is done so that you can put a `<br/>` within another inline element.

eksperimental:
> hmm, while you are correct from an HTML perspective, shell_docs treats `<br/>` as a block element at the moment.
> 
> So for this fix to be correct, we would also need to change the way that the rendering and validation is done so that you can put a `<br/>` within another inline element.

Could you please give me a hint on where to find this? I am be able to contribute a fix as well.

garazdawi:
You need to update the `BLOCK` and `INLINE` macros in shell_docs.erl. The current shell_docs docs adds a newline after each block construct and by moving `br` to inline you need to add a clause that adds a newline when it encounters a `br`.

There is a testcase called "render" in shell_docs_SUITE that tests that the output of `shell_docs:render` is what it should be, so make sure that it passes. You can run the test like this:

```
> cd lib/stdlib
> make test ARGS="-suite shell_docs_SUITE"
```

Just make sure you have the correct erlang in the path with the documentation built.

garazdawi:
Did you make any progress? If not, I may give it a crack "soon".

eksperimental:
> Did you make any progress? If not, I may give it a crack "soon".

Sorry, I haven't it looked into it yet.
If you start working on it, please ping me here so i don't.
Thank you

garazdawi:
Closing this PR as there as be no progress in a while.

eksperimental:
Sorry, I have been super-busy lately. I have saved this pull request, and I will reopen it whenever I have some spare time for it. Thank you @garazdawi

PR #5184, opened 2021-09-03
Fix doc typo

qrede:
Fix doc typo


IngelaAndin:
Looks good could you please base the fix on the maint branch instead of the master branch to get the correction into the OTP-24 track?  It needs to be cherry-picked/rebased --onto and force pushed, the edit option will not work in this case.

PR #5186, opened 2021-09-03
Allow reconfiguration of logger

hauleth:
This adds function `logger:reconfigure/0` that allows to reset logger configuration to what is configured in `kernel` application configuration. This function is meant to be used by build/run tools like Rebar3 or Mix to provide "production-like" way to configure logging in reliable way.

This function is obviously not safe to be called during application runtime, as it may cause losing some log messages, but it is by far better solution than what we have now, where Rebar3 tries to reimplement such functionality on their own (risking incompatibilities if `logger` configuration will be expanded one day) or just ignore it, and requires users to configure logging on their own (like Mix does, that just relies that users will use Elixir's Logger). Lack of such functionality prevents Elixir from further integration with Erlang's Logger.

### DIFF
--- a/lib/kernel/doc/src/logger.xml
+++ b/lib/kernel/doc/src/logger.xml
@@ -1189,6 +1189,18 @@ logger:set_proxy_config(maps:merge(Old, Config)).
       </desc>
     </func>
 
+    <func>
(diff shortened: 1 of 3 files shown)


hauleth:
It is followup #4663 as my Git-fu screwed the previous MR

garazdawi:
Thanks!

Issue #4728, opened 2021-04-11
ets:lookup_element/3 allocation bottleneck if called from multiple processes


vans163:
EDIT: TLDR, :ets.lookup_element for a large term (~60kb) causes memory allocations to spike and the dirty_schedulers to acquire locks trying to reserve ram from the kernel.  The system becomes stalled. This happens when over 7000 processes call into the same table (different rows but similar sized terms).

```elixir
    opt = [
      :public,
      :named_table,
      :ordered_set,
      {:write_concurrency, true},
      {:read_concurrency, true}
    ]
```

**Describe the bug**
We have been noticing large performance regressions when scaling out workloads.  Scaling being non-linear and exponentially poor.  Example 1000 sessions 5% cpu usage, 2000 10%, 5000 50% (critical level), 7000 100% (emulator is over 100%, many processes dont get scheduled).

I have drilled it down to the allocator.  I am not sure why, but allocations are global and it seems 1 (erlang) process is responsible for them, as soon the workload scales beyond the critical level, the global allocator cannot respond in time.  And everything exponentially slows down.

I am not sure on the solution for this, its very hard to find the cause of this as it does not appear on profiling tools when tracing a process.  Also allocator on msacc is pretty low,  the usage gets roilled into emulator/allocator.  

We have EPYC boxes with 64 cores and a high clocked 8core CPU outperforms the EPYC, due to it being able to process the allocations 2x faster which are the bottleneck.

min_heap_size does not seem to be a real solution when the allocation is huge like 2GB+. We have bumped min_heap_size on certain processes and indeed it did help get the allocator portion of msacc down.

**To Reproduce**
A simple way is create a large ets table of like 5000 records of total table size 4gb.  And call :ets.tab2list(),  this will cause a large allocation.   Then reduce the list into a new one.  Probably also would need to simulate a semi-real workload so all the schedulers are busy.

**Expected behavior**
What I would expect to happen is 1 scheduler out of say 64, to be maxed at 100%.

What happens is the entire VM on all 64 schedulers is 100% emulator usage across 64 cores. Starving all the other processes for scheduler time, and the lock seems deep too because the runqueue does not spike too much, but its clear the processes are not getting any scheduler time.

**Affected versions**
Erlang/OTP 23 [erts-11.2] [source] [64-bit] [smp:64:64] [ds:64:64:10] [async-threads:1] [hipe]

Interactive Elixir (1.11.4)

**Additional context**
[...]

PR #1790, opened 2018-04-19
Improve memory instrumentation

jhogberg:
This PR replaces the old memory instrumentation with a new implementation that scans carriers instead of wrapping `erts_alloc`/`erts_free`. The old implementation could not extract information without halting the emulator, had considerable runtime overhead, and the memory maps it produced were noisy and lacked critical information.

Since the new implementation walks through existing data structures there's no longer a need to start the emulator with special flags to get information about carrier utilization/fragmentation. Memory fragmentation is also easier to diagnose as it's presented on a per-carrier basis which eliminates the need to account for "holes" between `mmap` segments.

To help track allocations, each allocation can now be tagged with what it is and who allocated it at the cost of one extra word per allocation. This is controlled on a per-allocator basis with the `+M<S>atags` option, and is enabled by default for `binary_alloc` and `driver_alloc` (which is also used by NIFs).

The old interface has been removed since it can't be adapted to the new implementation and the module has always been experimental. It could coexist with the new one if we kept the old implementation around, but it will be removed unless someone makes a strong case for keeping it.

There's still some work left on the documentation, but with rc1 coming soon I figured it's best to get feedback as soon as possible.

---
#### Examples:

`instrument:carriers/0-1` return a list with information about each carrier's total size, combined allocation size, allocation count, whether it's in the migration pool, and a histogram of *free* block sizes (log2, the first interval ends at 512 by default).

In the example below, `ll_alloc` carrier has no free blocks at all, the  `binary_alloc` and `eheap_alloc` ones look healthy with a few very large blocks, and the `fix_alloc` carrier is somewhat fragmented with 22 free blocks smaller than 512 bytes (although this is not a problem for that allocator type).

```erlang
1> instrument:carriers().
{ok,{512,
     [{ll_alloc,1048576,0,1048344,71,false,{0,0,0,0,0,0,0,0,0,0,0,0,0,0}},
      {binary_alloc,1048576,0,324640,13,false,{3,0,0,1,0,0,0,2,0,0,0,0,0,0}},
      {eheap_alloc,2097152,0,1037200,45,false,{2,1,1,3,4,3,2,2,0,0,0,0,0,0}},
      {fix_alloc,32768,0,29544,82,false,{22,0,0,0,0,0,0,0,0,0,0,0,0,0}},
      {...}|...]}}
```

`instrument:allocations/0-1` return an allocation summary with histograms of *allocated* block sizes grouped by their origin and allocation type. The example below is taken with `+Muatags true` to track allocations on all allocator types.

[...]

PR #5187, opened 2021-09-06
Allocator instances for dirty schedulers

rickard-green:
Enabled using the following command line argument:

`+Mdai max|<amount>`

Set amount of dirty allocator instances used. Defaults to 0. That is, by default no instances will be used. The maximum amount of instances equals the amount of dirty CPU schedulers on the system.

By default, each normal scheduler thread has its own allocator instance for each allocator. All other threads in the system, including dirty schedulers, share one instance for each allocator. By enabling dirty allocator instances, dirty schedulers will get and use their own set of allocator instances. Note that these instances are not exclusive to each dirty scheduler, but instead shared among dirty schedulers. The more instances used the less risk of lock contention on these allocator instances. Memory consumption do however increase with increased amount of dirty allocator instances.


erts/emulator/beam/erl_msacc.c
```diff
@@ -364,20 +364,18 @@ erts_msacc_request(Process *c_p, int action, Eterm *threads)
...
 
-    *threads = erts_no_schedulers;
-    *threads += 1; /* aux thread */
+    *threads = erts_no_aux_work_threads;
 
     erts_atomic32_init_nob(&msaccrp->refc,(erts_aint32_t)*threads);
 
     erts_proc_add_refc(c_p, *threads);
 
     if (erts_no_schedulers > 1)
```
garazdawi, line 373: This condition is incorrect. I think it should be removed all together?

As it is now the emulator -> statistics_SUITE -> msacc testcase breaks when running using `+S 1`.

rickard-green: Thanks! Fixed now.

rickard-green: > Seems to me like the entire if should be removed?

Yes you are right, the whole if should be removed.

> Won't the fix you made break the case where we have only 1 aux thread and more than 1 schedulers?

No `erts_no_aux_work_threads` is always greater than 1 (threads handling aux work including schedulers) so it is an unnecessary test.

max-au:
Amount of allocations and deallocations stays roughly the same. `lcnt` complains about `alcu_allocator`, although I now suspect that problem happens due to general inefficiency of processes running large heaps. 
With default settings, single carrier block threshold (512k) is lower than 1 Mb. So every time GC happens, it means a new mmap, allocating many pages but not committing. Then, during GC, old heap is copied onto newly allocated heap, causing large amount of minor page faults (one per 4k page).
I am trying to figure out a test case for this PR, proving and explaining specific case when multiple dirty allocators make significant different, but so far unsuccessful.

I can paste lcnt/MSACC output with and without this PR (if it helps).

rickard-green:
Yes, please paste the lcnt output.

max-au:
Control node (single allocator instance):
```

        Thread    alloc      aux      bifbusy_wait check_io emulator      ets       gc  gc_full      nif    other     port     send    sleep   timers
Stats per type:
         async    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%  100.00%    0.00%
           aux   27.00%    3.38%    0.00%    0.00%    0.21%    0.00%    0.00%    0.00%    0.00%    0.00%    0.06%    0.00%    0.00%   69.37%    0.00%
dirty_cpu_sche    7.86%    0.00%    0.00%    0.00%    0.00%    0.04%    0.00%    7.47%    5.27%    0.00%    0.01%    0.00%    0.00%   79.34%    0.00%
dirty_io_sched    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.03%    0.00%    0.00%    0.00%   99.97%    0.00%
          poll    0.52%    0.00%    0.00%    0.00%    2.83%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%   96.65%    0.00%
     scheduler    6.79%    0.37%   23.09%    0.00%    0.10%   41.76%   12.83%    7.54%    0.96%    1.58%    0.46%    1.07%    2.31%    1.10%    0.05%
```
LCNT (relevant part)
```
                       lock    id   #tries  #collisions  collisions [%]  time [us]  duration [%]
                      -----   ---  ------- ------------ --------------- ---------- -------------
             alcu_allocator    10    96707        12916         13.3558   32133181      771.8113
                  run_queue    38  4425716        14124          0.3191     468189       11.2455
```

Same node, same workload (~same performance), now with `+Mdai max`:
```

        Thread    alloc      aux      bifbusy_wait check_io emulator      ets       gc  gc_full      nif    other     port     send    sleep   timers
Stats per type:
         async    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%  100.00%    0.00%
           aux    4.47%   12.50%    0.00%    0.00%    0.01%    0.00%    0.00%    0.00%    0.00%    0.00%    3.10%    0.00%    0.00%   79.92%    0.00%
dirty_cpu_sche   21.88%    0.00%    0.00%    0.00%    0.00%    0.03%    0.00%    8.17%    5.95%    0.00%    0.01%    0.00%    0.00%   63.96%    0.00%
dirty_io_sched    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%  100.00%    0.00%
          poll    0.54%    0.00%    0.00%    0.00%    2.47%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%   96.99%    0.00%
     scheduler    6.32%    0.34%   26.11%    0.00%    0.09%   41.23%   12.03%    8.06%    1.00%    1.28%    0.48%    1.04%    1.94%    0.00%    0.06%
```

LCNT:
```
                       lock    id   #tries  #collisions  collisions [%]  time [us]  duration [%]
                      -----   ---  ------- ------------ --------------- ---------- -------------
                       mseg    37    23764        16190         68.1283  107373242     3229.7736
             alcu_allocator   334   173617         2210          1.2729   46857250     1409.4602
                  run_queue    38  3639436         5648          0.1552     148024        4.4525
```

Having multiple aux threads appears to trigger significant contention for mseg_allocator.
If I limit number of dirty allocators to 2, `mseg_alloc` is not contended, but `aux` threads are still utilised.

Results (again, same workload):

LCNT:
```
                       lock    id   #tries  #collisions  collisions [%]  time [us]  duration [%]
                      -----   ---  ------- ------------ --------------- ---------- -------------
             alcu_allocator    28   144129        17534         12.1655  224679296     7249.6766
                       mseg     3    36593        15014         41.0297    8116618      261.8971
                  run_queue    54  4570185        16518          0.3614     178884        5.7720
```

Inspecting it shows expected contention:
```
           lock            id  #tries  #collisions  collisions [%]  time [us]  duration [%] histogram [log2(us)]
          -----           --- ------- ------------ --------------- ---------- ------------- ---------------------
 alcu_allocator   eheap_alloc   11906         8924         74.9538  114779075     3703.5508 |      ..Xx.. ......xxxxXXx..  |
 alcu_allocator   eheap_alloc   11132         8219         73.8322  109890725     3545.8195 |      ..X..........xxxxXXx..  |
 alcu_allocator      sl_alloc   39338          365          0.9279       4024        0.1298 |      X.............          |
```

rickard-green:
Thanks! The mapping to `mseg` instance to use by dirty alloc instances was wrong, so all dirty allocator instances used the `mseg` instance for arbitrary threads instead of their own `mseg` instances. I'll push a fix later today.

rickard-green:
I've pushed the fix, but I haven't tested it yet, though.

rickard-green:
I've done some testing of the fix now and it seems to solve the issue.

max-au:
Confirming, `mseg` allocator lock contention is no longer visible. `alcu_allocator` lock contention went down:

```
                       lock    id  #tries  #collisions  collisions [%]  time [us]  duration [%]
                      -----   --- ------- ------------ --------------- ---------- -------------
             alcu_allocator   478  242331          603          0.2488    2373413       80.8481
                  run_queue    54 4395045         6657          0.1515     130074        4.4308
```

LCNT:
```
        Thread    alloc      aux      bifbusy_wait check_io emulator      ets       gc  gc_full      nif    other     port     send    sleep   timers
Stats per type:
         async    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%  100.00%    0.00%
           aux    0.92%    2.87%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%   13.71%    0.00%    0.00%   82.49%    0.00%
dirty_cpu_sche    1.32%    0.00%    0.00%    0.00%    0.00%    0.02%    0.00%    7.94%    5.90%    0.00%    0.01%    0.00%    0.00%   84.81%    0.00%
dirty_io_sched    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%  100.00%    0.00%
          poll    0.49%    0.00%    0.00%    0.00%    2.18%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%    0.00%   97.33%    0.00%
     scheduler   10.12%    0.44%   20.69%    0.00%    0.05%   45.39%   10.13%    7.97%    0.85%    1.04%    0.37%    0.73%    2.18%    0.01%    0.04%
```

Side note, it hasn't helped this specific workload at all. The workload runs processes with large heaps (>1 Mb), and every GC turns to be extra expensive. It means jumps from normal to dirty scheduler and back. Dirty scheduler makes expensive allocation (of a single block carrier). Then GC copying data to the new heap is even more expensive due to large amount of minor page faults (currently BEAM supports only 4k memory pages, and it appears to lead to lock contention somewhere else (in kernel?)). Superpages (2Mb pages) could help to a degree. But GC for processes containing large amount of data on heap is very expensive.
So far we decided to utilise supercarrier to avoid minor page faults. It helps but has implications...

vans163:
This issue here does not seem to be fixed, https://github.com/erlang/otp/issues/4728#issuecomment-821292654 but its a huge step forward definitely.

Checking out this branch and running with `iex --sname test --erl '+Mdai max'`
image

The emulator is able to get 95% usage now (its not waiting for alloc on the dirties) and the dirties spend 45% of their time in GC.

The total system cpu is still stuck waiting for kernel locks (total system CPU utilization is below 30% but beam is at 100%), which is the same as without using the `+Mdai max`, its just the cpu usage moved from the `alloc on dirties` to the `emulator on schedulers` .

----

`iex --sname test --erl '+Mdai max +MMscs 96000'`

Using a large supercarrier makes things interesting now while checking the run_queue it has bursts of brilliance so the run queue is at 0-100 for some 5 seconds, then spikes for 3000+ (there are 10k processes running in the example linked) then clears up back to 0-100
```elixir
:erlang.statistics(:total_run_queue_lengths)
```
Virt is at 130gb (the system has 126G ram) so it seems it keeps hitting a ceiling on the maximum memory it can address.

```elixir
Enum.each 1..30, fn(_)->
  Process.sleep(1000)
  runq = :erlang.statistics(:total_run_queue_lengths)
  IO.inspect {DateTime.utc_now(), runq}
end
```
```
{~U[2021-11-03 22:34:50.652249Z], 127}
{~U[2021-11-03 22:34:51.712489Z], 96}
{~U[2021-11-03 22:34:52.776392Z], 10}
{~U[2021-11-03 22:34:53.781926Z], 0}
{~U[2021-11-03 22:34:54.784176Z], 0}
{~U[2021-11-03 22:34:55.831824Z], 569}
{~U[2021-11-03 22:34:57.476933Z], 4106}
{~U[2021-11-03 22:34:58.918308Z], 3863}
{~U[2021-11-03 22:35:00.548597Z], 3455}
{~U[2021-11-03 22:35:01.963742Z], 1941}
{~U[2021-11-03 22:35:03.628905Z], 1828}
{~U[2021-11-03 22:35:04.974563Z], 1151}
{~U[2021-11-03 22:35:06.485571Z], 1153}
{~U[2021-11-03 22:35:08.286411Z], 949}
{~U[2021-11-03 22:35:09.793041Z], 720}
{~U[2021-11-03 22:35:10.834515Z], 119}
{~U[2021-11-03 22:35:11.861284Z], 2}
{~U[2021-11-03 22:35:12.869329Z], 5}
{~U[2021-11-03 22:35:13.871627Z], 0}
{~U[2021-11-03 22:35:15.071407Z], 1592}
{~U[2021-11-03 22:35:16.716128Z], 3855}
{~U[2021-11-03 22:35:18.331870Z], 3472}
{~U[2021-11-03 22:35:19.821004Z], 2145}
{~U[2021-11-03 22:35:20.993760Z], 1509}
{~U[2021-11-03 22:35:22.276291Z], 1497}
{~U[2021-11-03 22:35:23.670082Z], 1041}
{~U[2021-11-03 22:35:24.908241Z], 1023}
{~U[2021-11-03 22:35:26.117832Z], 993}
{~U[2021-11-03 22:35:27.272381Z], 433}
{~U[2021-11-03 22:35:28.390729Z], 419}
```
image
We can see virt always pushing to reach 141G before dropping to around 126G. Res stayed fixed at 64G.

70-75% of the system itself is idle while the emu is at 0% idle.
image

No pressure on the CPU or memory
```
root@a ~ # cat /proc/pressure/cpu
some avg10=6.34 avg60=5.72 avg300=5.18 total=1154577495
full avg10=0.00 avg60=0.00 avg300=0.00 total=718162792
root@a ~ # cat /proc/pressure/memory 
some avg10=0.00 avg60=0.00 avg300=0.00 total=0
full avg10=0.00 avg60=0.00 avg300=0.00 total=0
```
-------
 
So I think that this PR takes the load off the allocator on the dirties, but there are underlying issues, its still mapping a ton of virtual memory and waiting for kernel locks (when there is over 65gb RES available to use, why is it trying to map more virt?)

The benchmark itself is very simple 10,000 processes call `map = :ets.lookup_element(Bench, "1", 2)` where the term returned is a size 63 binary, once per second.

rickard-green:
Current implementation of garbage collection has not been optimized for processes with large heaps. Moving garbage collection of such processes to dirty schedulers was not in order to improve performance of such processes, but instead to prevent such processes from destroying system characteristics of the system as a whole. Unfortunately there is no quick fix that will improve performance and memory consumption of such processes. It will be quite a large task to implement a garbage collection optimized for such processes and currently we have not prioritized this. The general design recommendation has for ever (or at least a very long time) been to design systems with processes not consuming large amounts of memory. That recommendation will still remain as such until we get the time to optimize this. I've written an issue #5396 about this. At some point we will prioritize this, but I cannot say when.

rickard-green:
@max-au wrote:
> So far we decided to utilise supercarrier to avoid minor page faults. It helps but has implications...

Please list issues you've encountered by using the super carrier. We know of a few, like it being protected by a global lock that might cause scaling issues and that sizing of it might not be easy for the user, but you might have encountered issues we are not aware of. If so, it would be good for us to know about them.

max-au:
> Before there was a decent module to do this (it would flag each term allocated if beam was compiled with a special flag) but it was removed in OTP 21 or somewhere around there.

Do you happen to remember that module? Since we face this problem (memory profiling) quite often, it'd be interesting to find some inspiration.

vans163:
> > Before there was a decent module to do this (it would flag each term allocated if beam was compiled with a special flag) but it was removed in OTP 21 or somewhere around there.
> 
> Do you happen to remember that module? Since we face this problem (memory profiling) quite often, it'd be interesting to find some inspiration.

Its the instrument module but for OTP20, in OTP21 it was rewritten and most of the features got removed. 

Here is a link to the OTP20 docs for it http://erlang.org/documentation/doc-9.3/lib/tools-2.11.2/doc/html/instrument.html

I believe I used it like..

```
MemData = instrument:memory_data()
ComplexTerm = instrument:descr(MemData)
```
Then ComplexTerm is very nice, it has the # of refs to the term the pid holding the term and the terms location in memory (I dont know of any other tool that does this after OTP21, excluding GDB as its hard to wire that up to a live node)

You can then read the terms location in memory to get the actual erlang term.

```
cat /proc/{beam_pid}/mem
OR
file api to read/seek
```

rickard-green:
That instrumentation was replaced by a new instrumentation implemented in #1790. The old instrumentation was written before the smp support and suffered from severe lock contention issues, so it could typically not be run on any real systems. We never found it that useful and especially not the memory map of the whole system. We think the new instrumentation gives you better abilities to find where things originates from and you are able to use it without the system choking on locks.

vans163:
> That instrumentation was replaced by a new instrumentation implemented in #1790. The old instrumentation was written before the smp support and suffered from severe lock contention issues, so it could typically not be run on any real systems. We never found it that useful and especially not the memory map of the whole system. We think the new instrumentation gives you better abilities to find where things originates from and you are able to use it without the system choking on locks.

But the new module does not allow you to pinpoint the exact terms that are problematic (holding refs or too many), it just gives you an overall summary which is not very helpful.  Having the exact term and what process created it allows one to very quickly narrow down where the term came from.  For example we quickly found issues where too many json strings were sitting on the binheap because 1 value from the json was written to ets and the process that created the was still alive.  A `binary:copy/1` for the value fixed that quickly. Another case was way too many terms in the heap that came from a ets table, the process in question was reading from the ets table in a loop, making it do a `lookup_element` for a primitive term solved that one.

The current instrument module would not allow you to get this kind of information unless I am missing some of its features.

PR #5190, opened 2021-09-07
Eliminate redundant tests

bjorng:
This pull request reimplements existing optimizations as optimizations on SSA code, making them more effective, and eliminates the `beam_peep` pass.

The main optimization in this pull request is elimination of redundant tests. Redundant tests can be introduced by inlining, macros, or complex guards such as:

    is_head(M, S) when M =:= <<1>>, S =:= <<2>> ->
        true;
    is_head(M, S) when M =:= <<1>>, S =:= <<3>> ->
        false.

Here the `M =:= <<1>>` test is repeated.

Another source of redundant test is when a test is following by its inverse test, such as:

    if
        A < B -> . . . ;
        A >= B -> . . .
    end

and:

    if
        A < B -> . . . ;
        A > B -> . . . ;
        A == B -> . . .
    end

The last test in each `if` is redundant and will be removed.

The module that this optimization had the biggest impact on is the Elixir module `Enum`. The `Enum` module inlines its `reduce/3` function in many functions. In about 20 functions (such as `drop/2`), the inlining would result in unreachable code for folding a list. The redundant test optimization removes that unreachable code.


lib/compiler/src/beam_jump.erl
```diff
@@ -629,6 +629,28 @@ opt([{test,_,{f,_}=Lbl,_}=I|Is], Acc, St) ->
     opt(Is, [I|Acc], label_used(Lbl, St));
 opt([{test,_,{f,_}=Lbl,_,_,_}=I|Is], Acc, St) ->
     opt(Is, [I|Acc], label_used(Lbl, St));
+opt([{select,select_val,R,F,Vls0}|Is], Acc, St) ->
+    case prune_redundant_values(Vls0, F) of
+	[] ->
+	    %% No values left. Must convert to plain jump.
```
jhogberg, line 623: Nitpick: there seems to be a mix of tabs and spaces in this section, maybe convert them all to spaces?

jhogberg (approved):

lib/compiler/src/beam_jump.erl
bjorng: Done.

Issue #5118, opened 2021-08-13
Opacity check misses violations in unions


mheiber:
**Describe the bug**

In some cases when an opaque type is part of a union opacity violations are not caught.

**To Reproduce**

Given this module:

```erl
-module(sample).
-compile([export_all, nowarn_export_all]).

% No opacity violation error
-spec test1() -> {term(), term()} | a.
test1() ->
    case rand:uniform(2) of
        1 -> gb_sets:new();
        2 -> a
    end.


% (correct) opacity violation error
-spec test2() -> {term(), term()} | a.
test2() ->
    case rand:uniform(2) of
        1 -> gb_sets:new()
    end.
```

Run `dialyzer sample.erl`

**Expected behavior**

Opacity violation for `test1/0`. The error should be the same as for `test2/0`.

My reasoning is baed on [the definition of opacity from the OTP docs on type specs](https://erlang.org/doc/reference_manual/typespec.html):

> Types declared as opaque represent sets of terms whose structure is not supposed to be visible from outside of their defining module. That is, only the module defining them is allowed to depend on their term structure.

`test1/0` depends on the structure of `gb_sets:set/1` being a two-tuple, which is why this looks like an opacity violation to me.

**Affected versions**

OTP 25 and (probably) earlier versions

jhogberg:
This is a fun one, but it's neither a bug nor does it have anything to do with opaque types. Let's tweak your example a little bit:
```erlang
%% No error here either!
-spec test1() -> a | b.
test1() ->
    case rand:uniform(2) of
        1 -> neither_a_nor_b;
        2 -> a
    end.

%% Invalid type specification
-spec test2() -> a | b.
test2() ->
    case rand:uniform(2) of
        1 -> neither_a_nor_b
    end.

%% No union, but still no warning!?
-spec test3() -> a.
test3() ->
    case rand:uniform(2) of
        1 -> definitely_not_a;
        2 -> a
    end.

%% Error here, despite the spec for test1/0 saying `a | b`!
test4() ->
      b = test1().
```

What's going on?

The spec for `test1/0` says it returns a subtype of `a | b`, and the implementation can return `a | neither_a_nor_b`. Since `neither_a_nor_b` is not a subtype of `a | b`, `dialyzer` concludes that case is impossible and thus the function returns `a`. Your example had an opaque type in place of `neither_a_nor_b`, but the principle is the same.

[...]

PR #5191, opened 2021-09-07
dialyzer: Fix opaque type handling for `underspecs`/`overspecs`

jhogberg:
Most of the warning checks would look straight through opaque types regardless of whether they belonged to the current module or not, resulting in wonky messages or missing warnings.

https://github.com/erlang/otp/issues/5118

### DIFF
--- a/lib/dialyzer/src/dialyzer_contracts.erl
+++ b/lib/dialyzer/src/dialyzer_contracts.erl
@@ -249,14 +249,14 @@ check_contract(#contract{contracts = Contracts}, SuccType, Opaques) ->
 	case check_contract_inf_list(InfList, SuccType, Opaques) of
 	  {error, _} = Invalid -> Invalid;
           ok ->
-            case check_extraneous(Contracts2, SuccType) of
+            case check_extraneous(Contracts2, SuccType, Opaques) of
               {error, invalid_contract} = Err ->
                 Err;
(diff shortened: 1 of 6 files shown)


lib/dialyzer/src/dialyzer_contracts.erl
```diff
@@ -933,28 +936,34 @@ t_from_forms_without_remote(_Forms, _MFA, _File, _RecDict) ->
...
+        true  -> false;
+        false -> blame_remote_list(CArgs, NRArgs, SArgs, Opaques)
       end
   end.
+
+%% As erl_types:t_is_subtype/2 but without looking into opaque types that
+%% aren't known to us.
+is_subtype(T1, T2, Opaques) ->
+    Inf = erl_types:t_inf(T1, T2, Opaques),
+    erl_types:t_is_equal(T1, Inf).
```
bjorng, line 969: Nitpick: dialyzer source files should be indented with 2 spaces, not 4. (There is an `-*- erlang-indent-level: 2 -*-` comment at the top of all files.)

bjorng (approved):

lib/dialyzer/src/dialyzer_contracts.erl
jhogberg: Thanks, fixed locally.

PR #5195, opened 2021-09-12
Further optimize off-heap traversal during minor GC

### LINKED ISSUE #4424: ERL-1347: Performance issue with file:read_line and ref binaries
Original reporter: `JIRAUSER16604`
Affected version: `OTP-23`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1347

---

```
Hi, I have been experiencing some strange performance issue while parsing large CSV files (in the order of millions of lines) when some of the fields are larger than 64 bytes, and it looks like I am not the only one (see [File.stream! too slow reading from text file|https://github.com/elixir-lang/elixir/issues/9956]).

This is what I am trying (and please excuse my Erlang :)):
{code:erlang}
start(_StartType, _StartArgs) ->
    lists:foreach(fun(Name) ->
        {T, _} = timer:tc(fun() ->
            {ok, FD} = file:open(Name, [raw, read_ahead, binary]),
            parse(FD)
        end),
        io:format("~s: ~fs~n", [Name, T/1000000])
    end, ["100k_64b.csv", "100k_65b.csv"]),
    readtest_sup:start_link().

parse(FD) ->
    parse(FD, []).
parse(FD, Acc) ->
    case file:read_line(FD) of
        {ok, Line} ->
            parse(FD, [string:trim(Line) | Acc]);
        eof ->
            file:close(FD),
            lists:reverse(Acc)
    end.
{code}
The file _100k_64b.csv_ contains 100000 lines where each line is exactly 64 bytes long (plus new line) and _100k_65b.csv_ the same but with lines of 65 bytes (again plus new line). My idea was to test the difference in performance between refc binaries and heap binaries while parsing files.

The results are pretty similar to the ones I get in Elixir:
 100k_64b.csv: 0.780010s
 100k_65b.csv: 31.810716s

I understand why refc binaries should be slower, but I was not expecting it to be this much.

Thanks!
```

sverker:
Only traverse first part of off-heap list that reside on the new-heap, except for writable binaries that are now kept in a separate list.

Fixes #4424.
    
Earlier during minor GC, traversing the off-heap list was continued to the end over terms residing on the old-heap. This for two reasons:
1. The BIN_OLD_VHEAP value was calculated by adding the sizes of all `ProcBin`'s on the old-heap.
2. All writable binaries are handled by shrinking (reallocating) binaries that have not been written to since last GC. Even binaries residing on the old-heap.

To optimize and only traverse the first part the off-heap list that reside on the new-heap, we now
1. maintain a correct value of BIN_OLD_VHEAP so it does not have to be recalculated during GC.
2. keep all writable binaries in a separate list `Process.wrt_bins` that is traversed to the end during every GC.

The aim of this PR has been to maintain the overall GC semantics and especially not alter the value of BIN_OLD_VHEAP or how binaries have been reallocated at the end of a GC.

### DIFF
--- a/erts/emulator/beam/erl_gc.c
+++ b/erts/emulator/beam/erl_gc.c
@@ -463,14 +463,11 @@ erts_gc_after_bif_call(Process* p, Eterm result, Eterm* regs, Uint arity)
 static ERTS_INLINE void assert_no_active_writers(Process *p)
 {
 #ifdef DEBUG
-    struct erl_off_heap_header* ptr;
-    ptr = MSO(p).first;
-    while (ptr) {
-	if (ptr->thing_word == HEADER_PROC_BIN) {	
-	    ProcBin *pbp = (ProcBin*) ptr;
-	    ERTS_ASSERT(!(pbp->flags & PB_ACTIVE_WRITER));
-	}
-	ptr = ptr->next;
+    ProcBin *pb = (ProcBin*) p->wrt_bins;
+    while (pb) {
+	ASSERT(pb->thing_word == HEADER_PROC_BIN);
+        ERTS_ASSERT(!(pb->flags & PB_ACTIVE_WRITER));
+	pb = (ProcBin*) pb->next;
     }
 #endif
 }
@@ -1744,7 +1741,7 @@ do_minor(Process *p, ErlHeapFragment *live_hf_end,
     OLD_HTOP(p) = old_htop;
     HIGH_WATER(p) = n_htop;
 
-    if (MSO(p).first) {
+    if (MSO(p).first || p->wrt_bins) {
 	sweep_off_heap(p, 0);
     }
 
@@ -1979,7 +1976,7 @@ full_sweep_heaps(Process *p,
 
     n_htop = sweep_heaps(n_heap, n_htop, oh, oh_size);
 
-    if (MSO(p).first) {
+    if (MSO(p).first || p->wrt_bins) {
 	sweep_off_heap(p, 1);
     }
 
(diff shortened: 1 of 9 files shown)


erts/emulator/beam/erl_gc.c
```diff
@@ -3049,94 +3021,151 @@ sweep_off_heap(Process *p, int fullsweep)
...
+                   less than 8 bytes will have no real effect */
+                if (unused >= 8) { /* A shrink candidate */
+                    if (shrink_unresolved_end) {
+                        shrink_writable_bin(pb, 0);
+                    }
+                    else if (++shrink_ncandidates > ERTS_INACT_WR_PB_LEAVE_LIMIT) {
+                        shrink_unresolved_end = pb;
+                        shrink_writable_bin(pb, 0);
+                    }
+                    /*else unresolved, handle in second traversal below */
```
bjorng, line 3116:
```suggestion
                    /* else unresolved, handle in second traversal below */
```

erts/emulator/beam/erl_gc.c
```diff
@@ -3049,94 +3021,151 @@ sweep_off_heap(Process *p, int fullsweep)
...
+            on_old_heap = ErtsInArea(pb, oheap, oheap_sz);
+            if (!on_old_heap) {
+                /* garbage */
+                erts_bin_release(pb->val);
+                pb = (ProcBin*) pb->next;
+                *prev = (struct erl_off_heap_header*) pb;
+                continue;
+            }
+        }
+        if (pb->flags) {
```
bjorng, line 3089:
```suggestion
        /*
         * Now see if we should shrink writable binaries. There are
         * two distinct cases:
         *
         * + There are one or more active writers. We will shrink all writable
         *   binaries without active writers down to their original sizes.
         *
         * + There are no active writers. We will shrink all writable
         *   binaries, but not fully. How much margin we will leave
         *   depends on the number of writable binaries.
         */
        if (pb->flags) {
```

bjorng (approved):
Looks good to me.

PR #5197, opened 2021-09-13
Make erl_docgen docs more responsive

garazdawi:
This PR does various improvements in the erl_docgen HTML docs.

1. This PR makes padding and margins smaller when the screen size becomes smaller and when it is small enough it moves the navigation to a collapsable topbar instead.  
  To achieve this the top-level layout has been changed from using and fixed placed div for the navbar to use css grid instead.  
  When on a small screen the index pages' navigation is permanently expanded so that the user does not have to expand it.
2. Moves all title links (i.e. links to section titles, function names etc) to be an `id` in the header instead of a `<a name>` tag.
3. Change the title links to use flexbox instead of tables for styling.
4. Use more semantic HTML elements in the documentation.

These changes are a small step towards making the documentation more accessible to everyone.

You can view the new documentation here: https://erlang.org/~lukas/responsive_docs/doc/index.html


lib/erl_docgen/priv/css/otp_doc.css
```diff
@@ -31,266 +35,446 @@ a:active       { color: #1c7cd6; text-decoration: none }
...
+    #leftnav.show-permanent {
+        max-height: none;
+    }
+    .show-permanent > .leftnav-tube {
+        overflow: initial;
+        max-height: none;
+        height: inherit;
+    }
+    .hide-mobile {
+        obacity: 0;
```
michallepicki, line 182:
```suggestion
        opacity: 0;
```
? (haven't tested, just noticed a potential typo)

garazdawi: Good catch. Thanks.

ferd:
Bit of a funny effect opening the menu on FF on iOS 15.0, where the bottom part of the menu seems to be missing a bit of height and hides content.

At first I thought it was thoroughly broken as the image below shows:

image

Clicking the 'expand all' button does show more menus but you still can't scroll to the bottom:

image

I'm not exactly sure if the search endpoint is supposed to be supported (couldn't find it), it's for sure the thing that looks the goofiest on mobile today.

garazdawi:
> Bit of a funny effect opening the menu on FF on iOS 15.0, where the bottom part of the menu seems to be missing a bit of height and hides content.

Can you check https://beta.erlang.org/doc and see if you have the same problem there? I've done some small updates and not bothered to update the docs in the link in this PR.

I can't reproduce it using FF on Android, but maybe that is just be

> I'm not exactly sure if the search endpoint is supposed to be supported (couldn't find it), it's for sure the thing that looks the goofiest on mobile today.

By search, you mean this page: https://erlang.org/doc/search/? If so, then the plan is to replace it with the search found on https://beta.erlang.org/doc which should look better in mobile.

ferd:
Yep the same exact thing is happening on the beta site, so my comment is irrelevant to this PR

garazdawi:
> Yep the same exact thing is happening on the beta site, so my comment is irrelevant to this PR

The beta website runs the latest version of this PR, so the problem remains... I'll need to find someone with an iOS device and see if I can reproduce the behaviour.

garazdawi:
Should be fixed now!

PR #5205, opened 2021-09-16
EDoc: Adjust default/private/hidden doc visibility in chunks

erszcz:
Prior to these adjustments when using ExDoc to generate docs for Erlang projects modules were required to have some top-level docs to be listed in the generated HTML. They could still have function-level docs, but the module wouldn't be listed anyway, so these docs wouldn't be accessible.

With these changes a module will be listed even if it does not have a top-level doc, so that its function docs are available in the generated HTML. This means a module has to be explicitly marked with EDoc `@private` or `@hidden` not to be listed.

See also https://github.com/elixir-lang/ex_doc/issues/1393.

### DIFF
--- a/lib/edoc/src/edoc_doclet_chunks.erl
+++ b/lib/edoc/src/edoc_doclet_chunks.erl
@@ -89,11 +89,8 @@ gen(Sources, _App, Modules, Ctxt) ->
 
 sources(Sources, Dir, Modules, Env, Options) ->
     Suffix = proplists:get_value(file_suffix, Options, ?DEFAULT_FILE_SUFFIX),
-    Private = proplists:get_bool(private, Options),
-    Hidden = proplists:get_bool(hidden, Options),
     {Ms, E} = lists:foldl(fun (Src, {Set, Error}) ->
-				  source(Src, Dir, Suffix, Env, Set,
-					 Private, Hidden, Error, Options)
+				  source(Src, Dir, Suffix, Env, Set, Error, Options)
 			  end,
 			  {sets:new(), false}, Sources),
     {[M || M <- Modules, sets:is_element(M, Ms)], E}.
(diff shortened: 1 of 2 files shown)


erszcz:
The Windows build failure seems to be caused by the environment, not the PR itself.

michallepicki:
I don't think it's in line with the change proposed in https://github.com/erlang/eep/pull/33/files ?
> the atom `hidden` must be given whenever documentation has been explicitly disabled for this entry

That sounds like both `@private` and `@hidden`

erszcz:
@michallepicki Good point, thanks :+1: I'll adjust the PR according to the new proposal.

michallepicki:
(If empty map is allowed, I think the format could be simpler with one less possible value. So IMO either the empty map or `none` could be removed)

erszcz:
I've aligned the PR with https://github.com/erlang/eep/pull/33:
- the default documentation value is the empty map, so even with no documentation present, a module will be listed
- EDoc `@private` maps to either `hidden` if the documentation is supposed to be hidden, or to the (possibly empty) map, if we're generating "private documentation" as requested by the user by specifying the `private` option to EDoc
- EDoc `@hidden` is always mapped to `none` in the chunk, as [that's what the EDoc documentation says about `@hidden`](https://erlang.org/doc/apps/edoc/chapter.html): "Marks the module so that it will not appear in the documentation (even if "private" documentation is generated)."

erszcz:
https://github.com/elixir-lang/ex_doc/commit/46947ac075bedc8dc18dd050e4ad43626ba10479 following the discussion in https://github.com/erlang/eep/pull/33 fixes the glitch in ExDoc and EDoc interop, so this part of the diff:

```
 -spec doc_content(_, _) -> doc().
-doc_content([], _Opts) -> none;
+doc_content([], _Opts) -> #{};
```

is not necessary anymore (but it doesn't hurt either). The rest of this PR is still useful, as it enables to use EDoc `private` option to generate private docs in chunks, and adds some comments / cleans things up.

garazdawi:
The code looks good and tests pass. Do we need a testcase for this?

PR #5206, opened 2021-09-16
shell_docs: Fix signature rendering

erszcz:
Prior to this change the rendered signature looks like this:

```
1> h(certifi, cacertfile).

\240\240cacertfile()

  CACertFile gives the path to the file with an X.509 certificate list containing the Mozilla CA
  ...
```

With this fix, i.e. by using an ordinary space for padding instead of a non-breaking space, the signature is rendered as expected:

```
1> h(certifi, cacertfile).

  cacertfile()

  CACertFile gives the path to the file with an X.509 certificate list containing the Mozilla CA
  ...
```

My env is iterm2 Build 3.4.9beta1 on macOS Catalina 10.15.7 (19H1323) running Erlang master (85530f5fb3), but the problem appeared at least since 24.0 RC1 (see https://github.com/erlef/build-and-packaging-wg/issues/25). I'm not sure if the NBSP handling behaviour is the same in other environments.

### DIFF
--- a/lib/stdlib/src/shell_docs.erl
+++ b/lib/stdlib/src/shell_docs.erl
@@ -590,6 +590,8 @@ normalize_format(Docs, #docs_v1{ format = <<"text/", _/binary>> }) when is_binar
 %%% Functions for rendering reference documentation
 render_function([], _D, _Config) ->
     {error,function_missing};
+render_function(FDocs, D, Config) when is_map(Config) ->
+    render_function(FDocs, D, init_config(D, Config));
 render_function(FDocs, #docs_v1{ docs = Docs } = D, Config) ->
     Grouping =
         lists:foldl(
@@ -602,7 +604,7 @@ render_function(FDocs, #docs_v1{ docs = Docs } = D, Config) ->
           end, #{}, lists:sort(FDocs)),
     lists:map(
       fun({Group,Members}) ->
-              Signatures = lists:flatmap(fun render_signature/1,lists:reverse(Members)),
+              Signatures = lists:flatmap(fun render_signature/1, lists:reverse(Members)),
               case lists:search(fun({_,_,_,Doc,_}) ->
                                         Doc =/= #{}
                                 end, Members) of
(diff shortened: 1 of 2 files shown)


garazdawi:
The reason it looks like that is because your (and my) terminal reports that it only supports latin1 characters. So what we need to do is to look at the encoding in the `#config` record and if latin1 emit 32 or else emit 160.

The reason why I would like to continue emitting 160 when possible is that it behaves better if for any reason someone has a small terminal.

I'll also have to check why my terminal reports that it only supports latin1... it should be able to print unicode...

garazdawi:
I pushed changes to this PR that makes shell_docs translate all `nbsp` to `sp` for all text. There is a number of places in the Erlang/OTP docs where nbsp is used, so this fixes the original issue and then some more.

erszcz:
> The reason it looks like that is because your (and my) terminal reports that it only supports latin1 characters.

Indeed, I compared macOS Terminal and iTerm2 (which I actually use daily). Terminal:

```
1> io:getopts().
[{expand_fun,#Fun<group.0.82824323>},
 {echo,true},
 {binary,false},
 {encoding,unicode}]
```

iterm2:

```
(search)`io:': io:getopts().
[{expand_fun,#Fun<group.0.82824323>},
 {echo,true},
 {binary,false},
 {encoding,latin1}]
```

I tried googling how to make iTerm2 behave, but no luck so far :/

> The reason why I would like to continue emitting 160 when possible is that it behaves better if for any reason someone has a small terminal.

Good point, but I think it wasn't consistent anyway. For example, even with latin1 in iTerm2, I got:

```
1> h(certifi, cacerts).

  -spec cacerts() -> [binary(), ...].

  CACerts builds an X.509 certificate list containing the Mozilla CA Certificate that can then be used via the cacerts setting in ssl options passed to the connect function.
ok
2> h(certifi, cacertfile).

\240\240cacertfile()

  CACertFile gives the path to the file with an X.509 certificate list containing the Mozilla CA Certificate that can then be used via the cacertfile setting in ssl options passed to the connect function.
ok
```

With the "universal translation" from nbsp to sp if latin1 is in use the problem should go away, so I think it's a good idea 👍

PR #5214, opened 2021-09-20
dialyzer: Add options for missing/extra range warnings

jhogberg:
These warnings are useful on their own, and I think they're worth breaking out from the more chatty `-Wunderspecs`/`-Woverspecs`.

```erlang
-module(t).
-export([t/1, q/1]).

-spec t(integer()) -> odd.
t(N) when N rem 2 =/= 0 ->
    odd;
t(_) ->
    even.

-spec q(integer()) -> positive | negative | zero.
q(N) when N > 0 ->
    positive;
q(N) when N < 0 ->
    negative.
```
```
# (The first message is an extra_range warning, the second is a missing_range warning.)
$ dialyzer t.erl -Wextra_range -Wmissing_range
  Checking whether the PLT /home/ejohogb/.dialyzer_plt is up-to-date... yes
  Proceeding with analysis...
t.erl:4:2: The success typing for t:t/1 implies that the function might also return 
          'even' but the specification return is 
          'odd'
t.erl:10:2: The specification for t:q/1 states that the function might also return 
          'zero' but the inferred return is 
          'negative' | 'positive'
 done in 0m0.59s
done (warnings were emitted)
```

I'm not a huge fan of the names for these options though, so suggestions are very welcome.

### DIFF
--- a/lib/dialyzer/doc/src/dialyzer.xml
+++ b/lib/dialyzer/doc/src/dialyzer.xml
@@ -316,6 +316,16 @@ dialyzer --plts plt_1 ... plt_n -- files_to_analyze</code>
       <item>
         <p>Include warnings for functions that only return by an exception.</p>
       </item>
+      <tag><c>-Wextra_return</c> (***)</tag>
+      <item>
+        <p>Warn about functions that return values that are not part of the
+          specification.</p>
+      </item>
+      <tag><c>-Wmissing_return</c> (***)</tag>
+      <item>
+        <p>Warn about functions whose specification includes types that the
+          function cannot return.</p>
+      </item>
       <tag><c>-Wno_behaviours</c></tag>
       <item>
         <p>Suppress warnings about behavior callbacks that drift from the
@@ -426,6 +436,16 @@ dialyzer --plts plt_1 ... plt_n -- files_to_analyze</code>
         <p>Suppress warnings about underspecified functions (the
           specification is strictly more allowing than the success typing).
           </p>
+      </item>
+      <tag><c>-Wno_extra_return</c></tag>
+      <item>
+        <p>Suppress warnings about functions that return values that are not
+          part of the specification.</p>
+      </item>
+      <tag><c>-Wno_missing_return</c></tag>
+      <item>
+        <p>Suppress warnings about functions whose specification includes types
+          that the function cannot return.</p>
       </item>
     </taglist>
 
--- a/lib/dialyzer/src/dialyzer.hrl
+++ b/lib/dialyzer/src/dialyzer.hrl
@@ -46,7 +46,9 @@
 -define(WARN_CONTRACT_SYNTAX, warn_contract_syntax).
 -define(WARN_CONTRACT_NOT_EQUAL, warn_contract_not_equal).
 -define(WARN_CONTRACT_SUBTYPE, warn_contract_subtype).
+-define(WARN_CONTRACT_MISSING_RETURN, warn_contract_missing_return).
 -define(WARN_CONTRACT_SUPERTYPE, warn_contract_supertype).
+-define(WARN_CONTRACT_EXTRA_RETURN, warn_contract_extra_return).
 -define(WARN_CONTRACT_RANGE, warn_contract_range).
 -define(WARN_CALLGRAPH, warn_callgraph).
 -define(WARN_UNMATCHED_RETURN, warn_umatched_return).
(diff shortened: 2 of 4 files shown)


josevalim:
Bikeshedding on the name only because it was requested. :)

I personally find the term `range` not immediately clear and also a bit overloaded as typespecs have integer ranges (1..3). If the warning applies only to return values, you could consider `missing_return` and `extra_return`. If you want to keep the same name as `overspecs` and `underspecs`, then there is also `over_return` and `under_return`. FWIW I prefer `missing_return` and `extra_return`.

bjorng:
There are five failed test cases in the dialyzer test suite.

KennethL:
I think the error printouts are too verbose and not as to the point as they ought to be.
I also agree with Josè that `missing_return` and `extra_return` would be better. By the way why do we want/need two options? Why not one option `check_return` that turns the checking of return values (both missing and extra).

t.erl:4:2: The success typing for t:t/1 implies that the function might also return 
          'even' but the specification return is 
          'odd'
Alternative text:
.... t:t/1 seems to return 'even' which is not part of the specified return types 

t.erl:10:2: The specification for t:q/1 states that the function might also return 
          'zero' but the inferred return is 
          'negative' | 'positive'
Alternative text:
... t:q/1 seems to not return 'zero' which is part of the specified return types

Could also have the strategy of marking every warning message with a category, in this case
extra_return: ..... text
and
missing_return: ......text

This would probably have implications for all warning messages so it is probably not for this PR

josevalim:
> By the way why do we want/need two options? Why not one option check_return that turns the checking of return values (both missing and extra).

I think having them being separate options is a good idea because having the `@spec` return a broader type than the code is not necessarily a bad thing. For example, you could have a function that returns `{ok, _} | {error, _}` in its spec but, in the first version of the code, the `{error, _}` value is never returned. You may still want callers of the function to potentially handle the error tuple though to avoid breaking changes in the future.

jhogberg:
> Bikeshedding on the name only because it was requested. :)
> 
> I personally find the term `range` not immediately clear and also a bit overloaded as typespecs have integer ranges (1..3). If the warning applies only to return values, you could consider `missing_return` and `extra_return`. If you want to keep the same name as `overspecs` and `underspecs`, then there is also `over_return` and `under_return`. FWIW I prefer `missing_return` and `extra_return`.

Thanks, I prefer them too. :)

> There are five failed test cases in the dialyzer test suite.

Thanks, I forgot to turn these on when `-Wspecdiffs` was provided.

jhogberg:
Sure, I'll add a ticket to our backlog :)

TD5:
@jhogberg, do you have a link to a follow up on @losvedir's question/issue?

Someone came to me asking about enabling these new options, and here was my response:

> It's not clear to me how this avoids the pitfalls of overspec/underspec.
> 
> For example, Dialyzer can infer more return types than are actually possible in the code, because Dialyzer reserves the right to "simplify" a list of values into a single, broader type. For example, Dialyzer might reduce 1 | 2 | 3 | 4 into non_neg_integer(). But if your spec correctly said your function returned 1 | 2 | 3 | 4 and then you'd presumably get a warning that you're missing all the other values in non_neg_integer()?
> 
> Similarly, in the other case, your function may return only a | b, but you might spec it as a | b | c for future-proofing, making sure callers handle the c case even if it can't strictly be generated yet. This is technically valid from the point of view that returning a more specific type that causes callers to handle more cases can't result in a crash - at worse it results in dead code - but it can have some practical benefits during the evolution or migration of an API.

I don't think this PR actually claims to resolve these points of confusion, and I know over-/under-specs have their uses for debugging Dialyzer itself, but it might be useful to make these pitfalls/usecases more clear in the documentation. I am happy to create a new PR that elaborates on this in the docs, if that would be of use.

jhogberg:
> @jhogberg, do you have a link to a follow up on @losvedir's question/issue?

It's tracked in our venerable internal issue tracker, I can't link to it but it's basically just a reminder to "see comment on #5214".

> For example, Dialyzer can infer more return types than are actually possible in the code, because Dialyzer reserves the right to "simplify" a list of values into a single, broader type. For example, Dialyzer might reduce 1 | 2 | 3 | 4 into non_neg_integer(). But if your spec correctly said your function returned 1 | 2 | 3 | 4 and then you'd presumably get a warning that you're missing all the other values in non_neg_integer()?

More or less, yeah.

> Similarly, in the other case, your function may return only a | b, but you might spec it as a | b | c for future-proofing, making sure callers handle the c case even if it can't strictly be generated yet. This is technically valid from the point of view that returning a more specific type that causes callers to handle more cases can't result in a crash - at worse it results in dead code - but it can have some practical benefits during the evolution or migration of an API.

Unfortunately a spec is not a type signature but a set of _additional constraints_ on top of what `dialyzer` has figured out on its own. Speccing the aforementioned function as returning `a | b | c` still makes it `a | b` as far as `dialyzer` is concerned, so any users trying to handle `c` will get a warning saying it's impossible.

> I don't think this PR actually claims to resolve these points of confusion, and I know over-/under-specs have their uses for debugging Dialyzer itself, but it might be useful to make these pitfalls/usecases more clear in the documentation. I am happy to create a new PR that elaborates on this in the docs, if that would be of use.

I'm not sure how helpful that would be given how under-documented specs are to begin with, I think we need to start in that end first.

PR #5215, opened 2021-09-20
Fix doc typo

qrede:
Fix doc typo


KennethL:
If the comment should be changed at all it should rather be
% to support debugging

PR #5222, opened 2021-09-21
ssl: Refactor to avoid re-encoding of decoded certs

IngelaAndin:
Let public key define the record #cert{} that can be used to provide
the original DER certs received from the peer and the decoded version
together as input to certificate handling functions such as
public:pkix_path_validation/3
    
It will be more efficient as we can avoid conversions and also a way for
applications such as ssl to tolerate certs with some ASN1-encoding errors that can be worked around.

### DIFF
--- a/lib/public_key/include/public_key.hrl
+++ b/lib/public_key/include/public_key.hrl
@@ -78,6 +78,8 @@
 	  point
 	 }).
 
+-record(cert, {der, otp}).
+
 -define(unspecified, 0).
 -define(keyCompromise, 1).
 -define(cACompromise, 2).
--- a/lib/ssl/src/ssl_certificate.erl
+++ b/lib/ssl/src/ssl_certificate.erl
@@ -88,7 +88,7 @@
 
 %%--------------------------------------------------------------------
 -spec trusted_cert_and_paths([der_cert()], db_handle(), certdb_ref(), fun()) ->
-          [{der_cert() | unknown_ca | invalid_issuer | selfsigned_peer, [der_cert()]}].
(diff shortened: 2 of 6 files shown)


lib/ssl/src/ssl_certificate.erl
```diff
@@ -103,7 +104,7 @@ trusted_cert_and_paths([Peer],  CertDbHandle, CertDbRef, PartialChainHandler) ->
     Chain = [#cert{der=Peer, otp=OtpCert}],
     case public_key:pkix_is_self_signed(OtpCert) of
         true ->
-            [{selfsigned_peer, [Peer]}];
+            [{selfsigned_peer, [#cert{der=Peer, otp=OtpCert}]}];
```
dgud, line 107: Use Chain created two lines above

IngelaAndin: Will do

lib/public_key/include/public_key.hrl
```diff
@@ -78,6 +78,8 @@
 	  point
 	 }).
 
+-record(cert, {der, otp}).
+
 -define(unspecified, 0).
 -define(keyCompromise, 1).
 -define(cACompromise, 2).
```
KennethL, line 85: Why cACompromise and not caCompromise?

IngelaAndin: These are enumeration names from an RFC and defined in an PKIX ASN-1 spec, and also not part of my PR!

IngelaAndin:
Closing this to replace it with a more appropriate branch name.

PR #5223, opened 2021-09-21
dialyzer: Miscellaneous bug fixes and cleanups

jhogberg:
This PR cleans up some cruft and fixes various bugs. Most of it is rather straightforward but there's a bug in the type lattice that I can't figure out how to fix without drastic measures, and I'd really appreciate it if someone more familiar with `dialyzer` (@kostis?) could have a look.

The gist of it is that opaque types are invariant on their type parameters: an opaque type `foo:bar(gurka | gaffel)` _is not_ a subtype of `foo:bar(atom())` nor is it a supertype of `foo:bar(gurka)`. For an opaque type to be compatible with another instance of the same type, their type arguments have to be exactly equal.

This may seem a bit strict but is required for correctness: while we can see that one can pass an `[integer()]` to a function expecting `[number()]` as lists are covariant on the element type, we don't have a way to specify the variance of an opaque type parameter, nor can we infer it beyond toy cases.

The easiest way to keep the type system sound in the face of that is to treat all type parameters as invariant. It's perhaps a bit surprising but there's nothing wrong with it. So far so good.

Then we hit this marvel in `gb_sets`:

```erlang
-spec add(Element, Set1) -> Set2 when
      Set1 :: set(Element),
      Set2 :: set(Element).
```

A strict reading of this says that it accepts a set containing a singleton `Element` and produces a set containing `Element` alone. If we were to pass a set produced by this function to one that expects (say) `gb_sets:set(atom())`, we should get a type error unless `Element` had already collapsed to `atom()`. Other funny consequences are left as an exercise to the reader.

Yet that works at the moment. Huh?!

1. `dialyzer` throws away type variables in specs, replacing them with `any()` and removing all relations between them.
    
   Trying to improve that is how I noticed this quirk.
2. `dialyzer` considers `any()` in type variables to be compatible with all other types. That is, `gb_sets:set(any())` is okay in functions  that expect (say) `gb_sets:set(atom())` and vice versa.

The former is awkward but the latter is where things go off the rails. We may pass `any()`'d and more explicitly typed opaques around interchangeably, but since type parameters are invariant there's no way for us to tell which is the subtype and which is the supertype (note that saying that "`any()` is always the supertype" will be wrong if the type parameter happens to be contravariant). This breaks the type lattice and sends `dialyzer` into infinite loops often enough that adding a test case for this bug is rather pointless: you can hardly even begin building a PLT of OTP before hitting it. The only reason it works at present is because we hide the error by detecting and breaking out of loops, accepting subtly wrong results.

I can't see a way around this without inventing a new notation and requiring people to rewrite their specs with it, so I've decided to treat all type arguments as `any()` in this PR. Unfortunately this makes `dialyzer` less useful and I'd be very grateful if someone could propose an alternative solution or point out where I had a brainfart.

### DIFF
--- a/lib/dialyzer/src/dialyzer.erl
+++ b/lib/dialyzer/src/dialyzer.erl
@@ -725,14 +725,14 @@ pp_type(Type) ->
   Form = {attribute, erl_anno:new(0), type, {t, Type, []}},
   TypeDef = erl_pp:form(Form, [{quote_singleton_atom_types, true}]),
   {match, [S]} = re:run(TypeDef, <<"::\\s*(.*)\\.\\n*">>,
-                        [{capture, all_but_first, list}, dotall]),
+                        [{capture, all_but_first, list}, dotall, unicode]),
   S.
 
 pp_spec(Spec) ->
   Form = {attribute, erl_anno:new(0), spec, {{a,b,0}, Spec}},
   Sig = erl_pp:form(Form, [{quote_singleton_atom_types, true}]),
   {match, [S]} = re:run(Sig, <<"-spec a:b\\s*(.*)\\.\\n*">>,
-                        [{capture, all_but_first, list}, dotall]),
+                        [{capture, all_but_first, list}, dotall, unicode]),
   S.
 
 parse_types_and_literals(Src) ->
--- a/lib/dialyzer/src/dialyzer_typesig.erl
+++ b/lib/dialyzer/src/dialyzer_typesig.erl
@@ -133,10 +133,8 @@
 %%-define(DEBUG_CONSTRAINTS, true).
 -ifdef(DEBUG).
 -define(DEBUG_NAME_MAP, true).
--define(DEBUG_LOOP_DETECTION, true).
 -endif.
 %%-define(DEBUG_NAME_MAP, true).
-%%-define(DEBUG_LOOP_DETECTION, true).
 
 -ifdef(DEBUG).
 -define(debug(__String, __Args), io:format(__String, __Args)).
@@ -1884,19 +1882,23 @@ scc_fold_fun(F, FunMap, State) ->
                                                     format_type(NewType)]),
   NewFunMap.
 
-solve(Fun, FunMap, State) ->
-  {ok, Map} = v2_solve_ref(Fun, FunMap, State),
-  Map.
-
 %% Solver v2
 
 -record(v2_state, {constr_data = maps:new() :: map(),
 		   state :: state()}).
 
-v2_solve_ref(Fun, Map, State) ->
+solve(Fun, FunMap, State) ->
   V2State = #v2_state{state = State},
-  {ok, NewMap, _, _} = v2_solve_reference(Fun, Map, V2State),
-  {ok, NewMap}.
+  try v2_solve_reference(Fun, FunMap, V2State) of
+    {ok, NewMap, _, _} -> NewMap
+  catch
+    throw:infinite_loop:Stack ->
+        erlang:raise(error,
+                     {"Infinite loop detected, please report this bug.",
+                      Fun,
+                      State#state.module},
+                     Stack)
+  end.
 
 v2_solve(#constraint{}=C, Map, V2State) ->
   case solve_one_c(C, Map) of
@@ -2137,10 +2139,10 @@ v2_solve_conj([], _Cs, _I, Map, Conj, IsFlat, V2State, UL, NewFs, VarsUp,
       ?debug("conjunct finished Id=~w\n", [Conj#constraint_list.id]),
       {ok, Map, V2State, lists:umerge([U|VarsUp])};
     NewFlags when NewFlags =:= LastFlags, Map =:= LastMap ->
-      %% A loop was detected! The cause is some bug, possibly in erl_types.
-      %% The evaluation continues, but the results can be wrong.
-      report_detected_loop(Conj),
-      {ok, Map, V2State, lists:umerge([U|VarsUp])};
+      %% We're stuck in an infinite loop, so we'll crash in the hopes of
+      %% getting a report. Trying to return anyway will yield potentially
+      %% misleading results.
+      throw(infinite_loop);
     NewFlags ->
       ?debug("conjunct restart Id=~w\n", [Conj#constraint_list.id]),
       #constraint_list{type = conj, list = Cs} = Conj,
@@ -2159,14 +2161,6 @@ v2_solve_conj(Is, [_|Tail], I, Map, Conj, IsFlat, V2State, UL, NewFs, VarsUp,
              LastMap, LastFlags) ->
   v2_solve_conj(Is, Tail, I+1, Map, Conj, IsFlat, V2State, UL, NewFs, VarsUp,
                LastMap, LastFlags).
-
--ifdef(DEBUG_LOOP_DETECTION).
-report_detected_loop(Conj) ->
-  io:format("A loop was detected in ~w\n", [Conj#constraint_list.id]).
--else.
-report_detected_loop(_) ->
-  ok.
--endif.
 
 add_mask_to_flags(Flags, [Im|M], I, L) when I > Im ->
   add_mask_to_flags(Flags, M, I, [Im|L]);
--- a/lib/dialyzer/src/dialyzer_typesig.erl
+++ b/lib/dialyzer/src/dialyzer_typesig.erl
@@ -2130,7 +2130,7 @@ v2_solve_conj([I|Is], [Cs|Tail], I, Map0, Conj, IsFlat, V2State0,
                     [U|UL], NewFs, VarsUp, LastMap, LastFlags)
   end;
 v2_solve_conj([], _Cs, _I, Map, Conj, IsFlat, V2State, UL, NewFs, VarsUp,
-             LastMap, LastFlags) ->
+              LastMap, LastFlags) ->
   U = lists:umerge(UL),
   case lists:umerge(NewFs) of
     [] ->
@@ -2142,6 +2142,7 @@ v2_solve_conj([], _Cs, _I, Map, Conj, IsFlat, V2State, UL, NewFs, VarsUp,
       report_detected_loop(Conj),
       {ok, Map, V2State, lists:umerge([U|VarsUp])};
     NewFlags ->
+      ?debug("conjunct restart Id=~w\n", [Conj#constraint_list.id]),
       #constraint_list{type = conj, list = Cs} = Conj,
       v2_solve_conj(NewFlags, Cs, 1, Map, Conj, IsFlat, V2State,
                     [], [], [U|VarsUp], Map, NewFlags)
--- a/lib/dialyzer/src/dialyzer_typesig.erl
+++ b/lib/dialyzer/src/dialyzer_typesig.erl
@@ -1801,12 +1801,12 @@ get_bif_test_constr(Dst, Arg, Type, _State) ->
 solve([Fun], State) ->
   ?debug("============ Analyzing Fun: ~tw ===========\n",
 	 [debug_lookup_name(Fun)]),
-  solve_fun(Fun, map_new(), State);
+  solve_fun(Fun, #{}, State);
 solve([_|_] = SCC, State) ->
   ?debug("============ Analyzing SCC: ~tw ===========\n",
 	 [[debug_lookup_name(F) || F <- SCC]]),
   Users = comp_users(SCC, State),
-  solve_scc(SCC, map_new(), State, Users, _ToSolve=SCC, false).
+  solve_scc(SCC, #{}, State, Users, _ToSolve=SCC, false).
 
 comp_users(SCC, State) ->
   Vars0 = [{Fun, state__get_rec_var(Fun, State)} || Fun <- SCC],
@@ -1817,11 +1817,8 @@ comp_users(SCC, State) ->
                                      Vars)]).
 
 solve_fun(Fun, FunMap, State) ->
-  Cs = state__get_cs(Fun, State),
-  Deps = get_deps(Cs),
-  Ref = mk_constraint_ref(Fun, Deps),
   %% Note that functions are always considered to succeed.
-  NewMap = solve(Fun, Ref, FunMap, State),
+  NewMap = solve(Fun, FunMap, State),
   NewType = lookup_type(Fun, NewMap),
   NewFunMap1 = case state__get_rec_var(Fun, State) of
 		 error -> FunMap;
@@ -1873,10 +1870,8 @@ affected(Updated, Users) ->
                 end || V <- Updated]).
 
 scc_fold_fun(F, FunMap, State) ->
-  Deps = get_deps(state__get_cs(F, State)),
-  Cs = mk_constraint_ref(F, Deps),
   %% Note that functions are always considered to succeed.
-  Map = solve(F, Cs, FunMap, State),
+  Map = solve(F, FunMap, State),
   NewType0 = unsafe_lookup_type(F, Map),
   NewType = t_limit(NewType0, ?TYPE_LIMIT),
   NewFunMap = case state__get_rec_var(F, State) of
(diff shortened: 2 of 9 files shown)


bjorng (approved):
Looks good to me.

bjorng:
> Unfortunately this makes dialyzer less useful and I'd be very grateful if someone could propose an alternative solution or point out where I had a brainfart.

I would say that dialyzer will still warn for the most likely misuses of opaque types. For example, consider this module:

```
-module(opaque).

confused(E) ->
    S = gb_trees:empty(),
    gb_sets:add_element(E, S).

breaking_abstraction(E) ->
    {1,_} = gb_sets:singleton(E).
```

In the `confused/1` function, the `gb_trees` and `gb_sets` modules are mixed. In the `breaking_abstraction/1` function, we try to match an opaque type. Dialyzer will still issue warnings:

```
opaque.erl:4:1: Function confused/1 has no local return
opaque.erl:6:28: The call gb_sets:add_element
         (E :: any(),
          S :: gb_trees:tree(_, _)) does not have an opaque term of type 
          gb_sets:set(_) as 2nd argument
opaque.erl:8:1: Function breaking_abstraction/1 has no local return
opaque.erl:9:5: The attempt to match a term of type 
          gb_sets:set(_) against the pattern 
          {1, _} breaks the opacity of the term
```

bjorng:
I tried to come up with an example that dialyzer in OTP 24 would warn for, but dialyzer with this PR applied would not warn for,
but I failed to come up with an example where dialyzer would issue warnings. Here is one of examples I tried:

```
test() ->
    S = gb_sets:add(99, int_set()),
    add_to_atom_set(foobar, S).

-spec int_set() -> gb_sets:set(integer()).
int_set() ->
    gb_sets:singleton(42).

-spec add_to_atom_set(atom(), gb_sets:set(atom())) -> gb_sets:set(atom()).

add_to_atom_set(E, S) ->
    gb_sets:add(E, S).
```

Dialyzer in OTP 24 did not issue any warnings. Therefore, it could be that this PR in practice does not remove much functionality from dialyzer.

mheiber:
I'm working on a type checker for Erlang and our solution was to enforce that all opaque type aliases are **covariant**.
This accommodate safe and idiomatic use of parameterised opaques such as gb_sets:set/1, sets:set/1, maps:iterator/1, etc.

Searching in OTP code, I couldn't find a single example of a contravariant opaque, so this rule doesn't seem to significantly limit expressivity. Which is not surprising, since (iuc) the only source of contravariance in Erlang is function types, and it's incredibly rare to have function types in aliases.

I'm glad you're looking into the issue of variance in opaques, but worry that the invariance choice could back the Erlang type system into a corner: Dialyzer can use approximations, but there is an increasing amount of tooling built around the type language that attempts to be sound (such as Gradualizer statically-typed BEAM languages). If there were a guideline in the docs about avoiding contravariant aliases it will make it easier to adopt tooling with more safety guarantees than Dialyzer, and could also make Dialyzer's behavior more predictable (as pointed out in the PR description).

jhogberg:
> I'm working on a similar problem for typed Erlang and our solution was to enforce that all opaque type aliases are covariant.
This accommodates safe and idiomatic use of parameterised opaques such as gb_sets:set/1, sets:set/1, maps:iterator/1, etc. Would it be possible to use this approach in Dialyzer as well?

Doing so was tempting but not backwards compatible, I agree that the overwhelming majority of uses would be just fine with covariance alone but there may be code out in the wild that's not. I don't feel comfortable breaking it just to fix this bug, but perhaps I'm too much of a stickler for backwards compatibility. :smile: 

It's a shame the type language didn't have variance annotations to begin with. Having slept on it I'm leaning towards keeping the current behavior of assuming invariance and adding explicit variance annotations (_implicit_ invariance would need to follow the all-vars-are-`any()` behavior as in this PR). I'll take it up with the technical board in the near future.

mheiber:
> I'm too much of a stickler for backwards compatibility. 😄

fwiw, I agree it's good to not break people with changes in this PR.


I'm more concerned about the direction for the type language rather than specific behavior of Dialyzer for these cases, because, as [@bjorng pointed out](https://github.com/erlang/otp/pull/5223#issuecomment-924656957) this PR probably doesn't change any user-facing behavior.

From the programmer's point of view, assuming *invariance* is backwards-incompatible:

```erl
sets:add_element(1, sets:new()). % ill-typed, since `1` is not a subtype of `none()`
```

I'm hoping there can be guidance for users that type specs are to be read such that type variables are covariant. Invariance seems like a non-starter because then things like `sets:set/1`, `maps:iterator/1` `gb_sets:set/1`, etc. become really difficult to use. Explicit annotations don't seem worth it, imo, since there are almost no contravariant or invariant type parameters in the wild afaict.

> Please let me know if this PR is the wrong place to bring this up, and I'll migrate these comments to the mailing list.

jhogberg:
> Please let me know if this PR is the wrong place to bring this up, and I'll migrate these comments to the mailing list.

I'm happy to take it here, it's easier to have it all in one place. Thanks for the comments by the way :)

> I'm hoping there can be guidance for users that type specs are to be read such that type variables are covariant. Invariance seems like a non-starter because then things like sets:set/1, maps:iterator/1 gb_sets:set/1, etc. become really difficult to use. 

They won't become difficult to use if you assume that all type arguments are `any()` in the absence of variance annotations. It would be more or less equal to what we have now, and in time we'll have annotated all the common types anyway.

> Explicit annotations don't seem worth it, imo, since there are almost no contravariant or invariant type parameters in the wild afaict.

Without annotations, how would you express a type that has callbacks which depend on a type parameter? It's common enough to be a concern and I don't want to needlessly restrict the expressiveness of the type language.

> From the programmer's point of view, assuming invariance is backwards-incompatible:

If you'll allow me to go off on a tangent, this would be an error if we assumed covariance too. Type variables in function specs are [explicitly documented as referring to _the same value_](http://erlang.org/doc/reference_manual/typespec.html#specifications-for-functions), so the current spec:

```erlang
-spec add(Element, Set1) -> Set2 when
      Set1 :: set(Element),
      Set2 :: set(Element).
```

Requires that `Set1` already is the singleton set of `Element`, and produces the singleton set of `Element`, which is not quite what you'd expect from that function. The spec is simply buggy. :(

The correct spec should be somewhere along the lines of:

```erlang
-spec add(Element, Set1) -> Set2 when
      Set1 :: set(Old),
      Set2 :: set(Old | Element).
```

Which will happily produce a `sets:set(1)` from `sets:add_element(1, sets:new())`, or a `sets:set(foo | bar)` from `sets:add_element(bar, sets:add_element(foo, sets:new()))`. `sets:del/2` is a wee bit harder to express though. :)

Either way, I agree that covariance is that we'd want in this case: not being able to pass a `sets:set(foobar)` to a function expecting `sets:set(atom())` is silly.

mheiber:
@jhogberg replying to [your most recent comment](https://github.com/erlang/otp/pull/5223#issuecomment-925144066)

Sticking with the example:

```erl
sets:add_element(1, sets:new())
```

You wrote:

> If you'll allow me to go off on a tangent, this would be an error if we assumed covariance too.


```erl
-spec add(Element, Set1) -> Set2 when
      Set1 :: set(Element),
      Set2 :: set(Element).
```

I think the spec is OK as-is. A constraint solver [like the one in "Local Type Inference"](https://dl.acm.org/doi/pdf/10.1145/345099.345100?casa_token=zZIBD4DJawoAAAAA:KWiWHm-iV-LY-EUSkSGTfShlnvFOplmPdw9w_Zs8-jmGn73-5inOdHlOwdikZ6tdbX0MxO123dk) will infer the smallest supertype for params in covariant position: In this case, `none() | 1`, which is equivalent to `1`.

Typed Racket follows this algorithm and behaves as expected:

```racket
#lang typed/racket
(: add (All (T) (-> T (Listof T) (Listof T))))
(define (add t ts) (cons t ts))

(add 1 '())        ; OK, T is inferred to be`(Listof 1)`
(add 1 '(true)) ; OK, T is inferred to be `(Listof (U True 1))`
```

TypeScript handles this case similarly ([playground link](https://www.typescriptlang.org/play?#code/FAEwpgxgNghgTmABAMwK4DsIBcCWB7dRGEEAHgBUA+ACiwC5FyAaRLAZwYGUwsLKBKLjz7AIBNlkQAPRAF4iJagEYW6MAHdE3LNX79EAegOIAtnlRswAWjwA3MHEQADKU4ZYAngAckONohx0ZAcEECJ-J21SdFQTACMHSicgA)), but TS has intricate behavior for more complex cases so is harder to compare to:

```ts
declare function add<T>(t: T, ts: Set<T>): Set<T>
const x = add(1, new Set()) // hover over `x`: type is inferred as `Set<number>`
```

jhogberg:
> I think the spec is OK as-is. A constraint solver like the one in "Local Type Inference" will infer the smallest supertype for params in covariant position: In this case, none() | 1, which is equivalent to 1.

I'm not so sure. According to the spec `Set1` must already be a subtype of `set(Element)`, which the empty set would be if covariant, but it would break as soon as you add another element, e.g. `sets:add_element(2, sets:add_element(1, sets:new()))` since `sets:set(1)` is not a subtype of `sets:set(2)`.

mheiber:
Re your comment
 
> I'm not so sure. According to the spec `Set1` must already be a subtype of `set(Element)`, which the empty set would be if covariant, but it would break as soon as you add another element, e.g. `sets:add_element(2, sets:add_element(1, sets:new()))` since `sets:set(1)` is not a subtype of `sets:set(2)`.

mheiber:
Replying to [comment](https://github.com/erlang/otp/pull/5223#issuecomment-925346169)

> I'm not so sure. According to the spec Set1 must already be a subtype of set(Element).

One way of looking at the example is that we are finding a substitution for the invisible `T` at the application site:

```erl
sets:add_element<T=???>(1, sets:new())
```

Once we pick a type for `T`, type checking reduces to the problem of type-checking monomorphic functions, so it's like `add_element` becomes: `-spec add_element(Elem :: 1 | none(), set(1 | none()))`

> According to the spec Set1 must already be a subtype of set(Element), 

It sounds like you're saying that `sets:set(none())` is not a subtype of `sets:set(number())`. But I think it is, given that sets are covariant and `none()` is a subtype of every type.

> it would break as soon as you add another element, e.g. sets:add_element(2, sets:add_element(1, sets:new()))

The type of that expression looks like `sets:set(1 | 2)` to me. With explicit type applications:

```erl
sets:add_element<T=1 | 2>(sets:add_element<T=1>(sets:new()))
 ```

**Comment by jhogberg** [comment]
> It sounds like you're saying that sets:set(none()) is not a subtype of sets:set(number()). But I think it is, given that sets are covariant and none() is a subtype of every type.

Yes, that's what I meant by "which the empty set would be if covariant." It breaks as soon as you go beyond that though. This is what the documentation says about type variables in function specs:

> Type variables can be used in specifications to specify relations for the input and output arguments of a function. For example, the following specification defines the type of a polymorphic identity function: 
>
> `  -spec id(X) -> X.`
>
>  Notice that the above specification does not restrict the input and output type in any way. These types can be constrained by guard-like subtype constraints and provide bounded quantification: 
>
>  `-spec id(X) -> X when X :: tuple().`
>
>  Currently, the :: constraint (read as «is a subtype of») is the only guard constraint that can be used in the when part of a -spec attribute. 
>
>  The above function specification uses multiple occurrences of the same type variable. That provides more type information than the following function specification, where the type variables are missing: 
>
> `  -spec id(tuple()) -> tuple().`
>
>  The latter specification says that the function takes some tuple and returns some tuple. The specification with the X type variable specifies that the function takes a tuple and returns the **same tuple**.

The last sentence is what breaks it: since `Element` has the same _value_ everywhere in the spec, `Set1 :: set(Element)` means that `Set1` is a subtype of the singleton set of the _value_ passed as `Element`. Hence, `sets:add_element(2, sets:add_element(1, sets:new()))` will be ill-typed regardless of whether a set is covariant or not.

**Comment by mheiber** [comment]
Re
> The last sentence is what breaks it: since `Element` refers to the _same value_

I'm saying `Element` can refer to the type `1 | 2`, and the values `1` and `2` can be ascribed the type `1 | 2`.

 `Element` is a type variable. It's safe to instantiate the type variable with whatever, and then proceed normally as if `add_element` is monomorphic.

Here are reasonable instantiations of the type variable s.t. the expression is well-typed:

```erl
sets:add_element<Element=1 | 2>(sets:add_element<Elem=1>(sets:new()))
```

The algorithm for finding these instantiates that I linked to above does the following: for covariant position `Elem` and argument types for that position `A` and `B`, instantiate `T` to `A | B`. The Typed Racket code that I linked shows that this is what Typed Racket does as well. Any type checker for a language with subtyping, generics, and type  is likely to also accept code like the above if `set` is covariant: I gave the examples of TypeScript and Racket above. Here's a Scala example as well: [scastie](https://scastie.scala-lang.org/O7DSEEKjS96Y9Jc0sYrueA). (I couldn't use `Set` in Scala because `Set` is invariant in Scala).

michalmuskala:
While TBH I'm not sure about the discussed issue of `sets:add(1, sets:new())` being well- or ill-typed under the current definition, I think the biggest issue with treating type parameters invariantly is what has been hinted at earlier.

In particular given I have a function:
```erlang
-spec has_number(sets:set(integer()), integer()) -> boolean().
```

Call I call such function with a value obtained from the call to `sets:add(1, sets:new())`? If we treat types invariantly it would seem like we can't - this seems like a huge practical issue.

If we say in this case type inference would type the call `sets:add(1, sets:new())` as `sets:set(integer())`, this still leaves the issue of having a function that, for example, is spec'ed to return something like `sets:set(pos_integer())` or a number of other issues like this.

jhogberg:
> Element can be instantiated to the type `1 | 2`, and the values `1` and `2` can be ascribed the type `1` | `2`.

This is not a matter of how we find types but what the spec says. The example in question should be an error, and giving `Element` a wider type than we already know does nothing but hide it.

Let's take a simpler example:

```erlang
-spec contrived(X, X) -> ok.
```

This says that both arguments must be _exactly equal in value_. When passed `contrived(1, 2)`, should we accept `1 | 2` here as well as the type for `X`?

> Call I call such function with a value obtained from the call to sets:add(1, sets:new())? If we treat types invariantly it would seem like we can't - this seems like a huge practical issue.
>
> If we say in this case type inference would type the call sets:add(1, sets:new()) as sets:set(integer()), this still leaves the issue of having a function that, for example, is spec'ed to return something like sets:set(pos_integer()) or a number of other issues like this.

I'm not saying we should make `set(T)` invariant, I propose adding annotations to let us to make it covariant on `T`. All opaque types are currently invariant with an undocumented corner case around `any()` that happens to make ordinary uses work by accident. This PR preserves that behavior in a consistent way that no longer breaks the type lattice.

michalmuskala:
You're right. It does look like dialyzer today treat opaque type params as invariant. For example, given the following program:
```erlang
-module(test).

-export([run/0]).

-spec value() -> sets:set(pos_integer()).
value() ->
   sets:add_element(1, sets:new()).

-spec consume(sets:set(integer())) -> boolean().
consume(Set) ->
    sets:is_element(1, Set).

run() ->
    consume(value()).
```

we get following errors:

```
test.erl:9:2: Invalid type specification for function test:consume/1. The success typing is
          (sets:set(pos_integer())) -> boolean()
test.erl:13:1: Function run/0 has no local return
test.erl:14:13: The call test:consume
         (sets:set(pos_integer())) does not have an opaque term of type
          sets:set(integer()) as 1st argument
```

Indicating that indeed dialyzer does not allow unifying `sets:set(pos_integer())` with `sets:set(integer())`.

mheiber:
> You're right. It does look like dialyzer today treat opaque type params as invariant.

How about I open a separate issue for treating opaques as covariant?

The reason I think treating opaques as invariant is a non-starter is that (imo) the main benefit of opaques is to enable encapsulation of containers. For example, this is what enabled OTP team to change the implementation of sets:set/1 to use maps instead of tuples.

jhogberg:
> The variance-annotations solution is a possibility, but it's a big burden to place on users for not much benefit, given that non-covariant opaques are so rare: I have not been able to find a single example. And it can be argued that users expect aliases to be covariant (Dart team at least argued this, and didn't support variance annotations).

As long as things keep working as they are I'm not sure it would be a big burden in practice, opaques are rare to begin with and parameterized ones rarer still.

> How about I open a separate issue for treating opaques as covariant?

Sounds good, we can continue the discussion there. :)

mheiber:
> Sounds good, we can continue the discussion there. :)
Thanks for our discussion! Here is the issue about covariant opaques: https://github.com/erlang/otp/issues/5240

jhogberg:
Merged as 008f44fefeb950bc4bfbe6a13922f2a8f5af49c9, breaking out the unicode fix to `maint`.

PR #5233, opened 2021-09-23
No longer support loading BEAM files compiled before OTP 22

bjorng:
Some things we are planning to do in OTP 25 would be complicated by keeping compatibility with OTP 21 and earlier.


sverker:
(Just saw that you pressed Merge, a couple of seconds too late)

When trying to load a beam file compiled with OTP-18 I got the following:
```
=ERROR REPORT==== 28-Sep-2021::11:52:16.725269 ===
beam/beam_load.c(130): Error loading module example:
  missing atom table


{error,badfile}
=ERROR REPORT==== 28-Sep-2021::11:52:16.725275 ===
Loading of /home/.../example.beam failed: badfile
```

I think it would be nice get a more informative error indicating compiler vs runtime version missmatch.

bjorng:
@sverker OK, I will fix that.

PR #5236, opened 2021-09-23
Replace "expr --" usage with (POSIX) parameter expansion

tianon:
See https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

(Whitespace throughout this file is very inconsistent -- I've used purely tabs on the new lines I wrote, but I'm happy to adjust them to match what the lines they're replacing used, or any other adjustment desired. :heart:)

Easiest way to test/validate these:

```console
$ a='--srcdir=foo-bar=baz-buzz'
$ echo "${a#*=}"
foo-bar=baz-buzz
$ a='-srcdir=foo-bar=baz-buzz'
$ echo "${a#*=}"
foo-bar=baz-buzz

$ a='CFLAGS=foo=bar=baz'
$ var="${a%%=*}"
$ val="${a#$var=}"
$ echo " var='$var'  val='$val' "
 var='CFLAGS'  val='foo=bar=baz' 

$ a='--without-foo-bar-baz'
$ echo "${a#--without-}"
foo-bar-baz
```

(This should fix the issue I noted in https://github.com/erlang/otp/pull/4909#issuecomment-925425681)

### DIFF
--- a/configure
+++ b/configure
@@ -55,7 +55,7 @@ skip_applications=
 while test $# != 0; do
     case $1 in
 	-srcdir=* | --srcdir=*)
-	    user_srcdir=`expr -- "$1" : '[^=]*=\(.*\)'`
+		user_srcdir="${1#*=}"
 	    if test "$ERL_TOP" != ""; then
 		echo "WARNING: Overriding ERL_TOP with $user_srcdir" 1>&2
 		echo "" 1>&2
@@ -95,7 +95,7 @@ while test $# != 0; do
 	    echo "" 1>&2
 	    ;;
 	-cache-file=* | --cache-file=* )
-	    static_cache=`expr -- "$1" : '[^=]*=\(.*\)'`
+		static_cache="${1#*=}"
 	    if test "$static_cache" != "/dev/null"; then
 		echo "WARNING: Only using config cache file '$static_cache' as static cache" 1>&2
 		echo "" 1>&2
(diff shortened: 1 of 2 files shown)


tianon:
I have applied this patch to a build of 24.1 which was failing on Alpine Linux before (due to not parsing my `--without-xxx` flags correctly) and now succeeds. :smile:

rickard-green:
Thanks for the pull request!

It would be nice to use these parameter expansion features. However, the [autoconf documentation](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Shell-Substitutions.html) say this in the chapter "portable shell programming" section "shell substitutions":

> ${#var}
> ${var%word}
> ${var%%word}
> ${var#word}
> ${var##word}
>
>  Posix requires support for these usages, but they do not work with many traditional shells, e.g., Solaris 10 /bin/sh.
>
> Also, pdksh 5.2.14 mishandles some word forms. For example if ‘$1’ is ‘a/b’ and ‘$2’ is ‘a’, then ‘${1#$2}’ should yield ‘/b’, but with pdksh it yields the empty string.

That is, this is not portable enough for us to use in `configure`. I would have gone for using `sed`, but there may be better alternatives.

A fix for this issue is wanted as a patch for OTP 24, so if you change this PR or do a new one, please base you changes on the top of the `maint-24` branch.

Regarding whitespace in files. We used to mix tabs and spaces, but have changed to only use spaces. We only change whitespace on lines changed for another reason.

tianon:
Ok, fair enough -- I guess `echo '...' | sed -e 's/.../.../'` is the desired pattern then?

I'll work on a second attempt (targeting `maint-24` and using spaces instead of tabs). :smile:

tianon:
Closing in favor of https://github.com/erlang/otp/pull/5251. :+1:

PR #5237, opened 2021-09-23
Improved inline multiplication in JIT for x86

potatosalad:
This moves multiplication for small integers inline with overflow checking, along with specialization for when the left or right hand side are immediates.

The original comments related to clobbering `RDX`, I _think_, are incorrect based on how we're using `imul` in 2-operand form instead of 1-operand form (see [reference docs for `imul`](https://www.felixcloutier.com/x86/imul)):

```
We avoid using ARG2 and ARG3 because multiplication clobbers RDX, which is
ARG2 on Windows and ARG3 on SystemV.
```

As a very non-scientific benchmark: based on some tests involving a pure Erlang version of `MurmurHash3_x86_32` with a 1MB input key, this change alone roughly doubled performance for me (30ms originally down to 15ms).

### DIFF
--- a/erts/emulator/beam/jit/x86/instr_arith.cpp
+++ b/erts/emulator/beam/jit/x86/instr_arith.cpp
@@ -653,106 +653,48 @@ void BeamModuleAssembler::emit_i_m_div(const ArgVal &Fail,
     mov_arg(Dst, RET);
 }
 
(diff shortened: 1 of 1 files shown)


garazdawi:
Thanks!

PR #5245, opened 2021-09-27
Fix crypto unload memory leak

sverker:
Not that serious leak but will silence valgrind and LeakSanitizer.


HansN (approved):
Looks good to me. Good catch!

PR #5252, opened 2021-09-30
Provide more helpful error messages when loading old BEAM files

bjorng:
After the changes in c7ddbcbe7d95 (#5233), attempting to load a BEAM
file compiled with OTP 20 or earlier would result in the following
error report:

    =ERROR REPORT==== 30-Sep-2021::09:28:25.626580 ===
    beam/beam_load.c(130): Error loading module t:
      missing atom table

Provide a more helpful error report if the BEAM file contains
an old-style atom table:

    =ERROR REPORT==== 30-Sep-2021::09:22:24.235402 ===
    beam/beam_load.c(133): Error loading module t:
      please re-compile this module with an OTP 25 compiler

While we are at it, also update all similar messages by the
loader to say "OTP 25 compiler" instead of "25 compiler".
(The missing "OTP" made more sense when releases were named
with an "R" prefix, such as R17.)


erts/emulator/beam/beam_load.c
```diff
@@ -164,7 +167,7 @@ erts_prepare_loading(Binary* magic, Process *c_p, Eterm group_leader,
         BeamLoadError2(stp,
                        "This BEAM file was compiled for a later version"
                        " of the run-time system than " ERLANG_OTP_RELEASE ".\n"
-                       "  To fix this, please recompile this module with an "
+                       "  To fix this, please recompile this module with an OTP "
```
garazdawi, line 170:
```suggestion
                       "  To fix this, please recompile this module with an Erlang/OTP "
```

I think it is better to say Erlang/OTP to be clearer.

sverker:
Would it not be good to get information about the minimum requirement. An OTP 22 compiler in this case. Or is it only we (the Erlang/OTP team) that runs the same beam files on different OTP versions.
I discovered this in our valgrind tests that run the same beam files on maint (24) and master (25).

bjorng:
@garazdawi I have rephrased the messages to use "Erlang/OTP 25".

bjorng:
> Would it not be good to get information about the minimum requirement.

That would not be trivial to implement correctly. A BEAM file does not contain any indication of which Erlang/OTP version that was used to compile it.

> An OTP 22 compiler in this case.

Actually, some OTP 21 modules can be loaded by OTP 25. It depends on which instructions they use.

PR #5254, opened 2021-09-30
erts: Optimize fun calls

jhogberg:
By making local and external funs share the same structure, we can remove a substantial amount of code from fun-related instructions.

As a result fun calls are now about 15% faster with the interpreter on my machine, 10% faster with the x64 JIT on that same machine, and 15% faster with the AArch64 JIT on my M1 Mac. The improvement is a bit less pronounced for funs with free variables, and a bit more for those without (including external funs).

This PR is partially reliant on some upcoming work by @kjellwinblad, which will guarantee that the second word of a boxed term is always readable. It only affects the AArch64 JIT and is not a practical hindrance in testing as things can only go awry when calling the empty tuple as a fun, so I've taken the liberty to post this for review now.


bjorng (approved):
Looks good to me.

PR #5259, opened 2021-10-01
arm jit opt: emit_i_is_tagged_tuple and emit_i_is_tagged_tuple_ff 

kjellwinblad:
This PR contains two main changes:

* Make sure that all boxed terms have allocated memory after the arity word. This is archived by padding all literal areas by one word and changing all places that previously created empty tuples on processes' heaps so that they instead create pointers to a constant empty tuple literal. (fafe55f, 7dc999b)
* Optimize the instructions `emit_i_is_tagged_tuple` and `emit_i_is_tagged_tuple_ff` in the ARM JIT. These optimization are enabled by the empty tuple changes described above. (abf3acd)

Commit 0fbbe22 is a bug fix that will be removed from this PR before it is merged as it will be merged to the maint branch first.


erts/emulator/beam/erl_map.c
```diff
@@ -368,10 +369,14 @@ static Eterm flatmap_from_validated_list(Process *p, Eterm list, Eterm fill_valu
     Sint  idx = 0;
 
 
-    hp    = HAlloc(p, 3 + 1 + (2 * size));
+    hp    = HAlloc(p, 3 + (size == 0 ? 0 : 1) + (2 * size));
```
michalmuskala, line 372: If I understand correctly this is to safely handle empty maps. If that's the case, would it make sense to also apply the same treatment to empty maps and have just one empty map literal that could be returned?

erts/emulator/beam/erl_gc.c
```diff
@@ -3821,6 +3823,165 @@ erts_dbg_within_proc(Eterm *ptr, Process *p, Eterm *real_htop)
...
+              have an arity of 0. This can change in the feature and
+              the condition above needs to be changed if it does.
+             */
+            erts_exit(ERTS_ABORT_EXIT,
+                      "Non-literal empty tuple found in heap.\n"
+                      "This is not allowed due to an optimization\n"
+                      "that assumes that the word after the arity\n"
+                      "word is allocated.\n");
+        }
+    }
```
jhogberg, line 3970: I found this hard to follow, how about something like:
```suggestion
   if (is_boxed(term)) {
        Uint arity = arityval(*boxed_val(term));

        /* Maps can have 0 arity even though they have something after the
         * arity word. */
        if (arity == 0 && !is_map(term)) {
            if (term != erts_get_global_literal(ERTS_LIT_EMPTY_TUPLE)) {
                /* Empty tuples are the only type of boxed value that can
                 * have an arity of 0. This can change in the feature and
                 * the condition above needs to be changed if it does. */
                erts_exit(ERTS_ABORT_EXIT,
                          "Non-literal empty tuple found in heap.\n"
                          "This is not allowed due to an optimization\n"
                          "that assumes that the word after the arity\n"
                          "word is allocated.\n");
            }
        }
    }
```

erts/emulator/beam/erl_gc.h
```diff
@@ -193,6 +193,13 @@ int erts_dbg_within_proc(Eterm *ptr, Process *p, Eterm* real_htop);
 /* Validates the frame chain, ensuring that it always points within the stack
  * and that no frames are skipped. */
 void erts_validate_stack(Process *p, Eterm *frame_ptr, Eterm *stack_top);
+int
+erts_dbg_check_heap_terms(int (*check_eterm)(Eterm),
+                          Process *p,
+                          Eterm *real_htop);
```
jhogberg, line 199: If we always abort on failure, why return `int`?

erts/emulator/beam/erl_global_literals.h
```diff
@@ -24,8 +24,12 @@
 #define ERTS_LIT_OS_TYPE         0
 #define ERTS_LIT_OS_VERSION      1
 #define ERTS_LIT_DFLAGS_RECORD   2
+#define ERTS_LIT_EMPTY_TUPLE     3
```
jhogberg, line 27: Calling `erts_get_global_literal(ERTS_LIT_EMPTY_TUPLE)` all the time seems a bit odd, perhaps we should make it `extern Eterm ERTS_LIT_XYZ` instead?

jhogberg (commented):
LGTM.

erts/emulator/beam/erl_map.c
kjellwinblad: Yes, we would save some memory and perhaps utilize the processor caches better by doing that. The drawback would be that, in some places, we would need a little bit more code when creating maps in BIF's to check if the size is 0 and then return the empty map (e.g., the [maps:remove/2 BIF](https://github.com/erlang/otp/blob/7079bda1d36707788149197ec75238cbc8c7b6b2/erts/emulator/beam/erl_map.c#L1834)). 

The empty tuple optimization was primary motivated by the fact that it (together with padding literal areas with 1 word) enables us to ensure that the word after the arity word of boxed terms always is allocated memory which in turn enables some optimizations in the ARM JIT and for invocations of funs.

However, an empty map takes up more space than an empty tuple and empty maps are probably much more common than empty tuples so I think it can be worth it just because it gives better memory properties. An alternative (suggested by @jhogberg ) to creating a constant empty map is to introduce a new term type for the empty map similar to how we do with the empty list.

erts/emulator/beam/erl_gc.h
kjellwinblad: It is to make the debug function more general and useful for other things one might want to use it for in the future. If the check_eterm function provided as an argument return 0 for any of the term on the heap, the whole function returns 0 instead of 1. This could be useful, for example, if one wants to log something when a particular term is found instead of aborting.

erts/emulator/beam/erl_global_literals.h
kjellwinblad: Yes, that might be better. I can do that change.

erts/emulator/beam/erl_gc.c
kjellwinblad: Thanks, your code is easier to follow in my opinion as well.

erts/emulator/beam/erl_global_literals.h
sverker: One disadvantage with referring to a global variable like ERTS_LIT_EMPTY_TUPLE in a lot of places for reading is that it may be harder to find where the variable is written to.

erts/emulator/beam/copy.c
```diff
@@ -116,10 +117,14 @@ Uint size_object_x(Eterm obj, erts_literal_area_t *litopt)
 		case ARITYVAL_SUBTAG:
 		    ptr = tuple_val(obj);
 		    arity = header_arity(hdr);
-		    sum += arity + 1;
 		    if (arity == 0) { /* Empty tuple -- unusual. */
+                        /*
+                           The empty tuple is always a literal so
+                           it does not take up any extra space
+                         */
 			goto pop_next;
```
sverker, line 125: Add `ASSERT(erts_is_literal(obj,ptr));`

and move down `ptr = tuple_val(obj);` after `if`.

erts/emulator/beam/erl_gc.h
jhogberg: I think it would be better to add that when we need it, but I don't have a strong opinion and leave it up to you. It's a small nit anyway :)

erts/emulator/beam/erl_global_literals.h
jhogberg: I don't think that will be much of a problem in practice since there's so few of them and they're only written once. Searching for `erts_register_global_literal` will point you right to them.

I'll leave it up to @kjellwinblad though. :)

erts/emulator/beam/copy.c
```diff
@@ -1909,7 +1924,11 @@ copy_shallow_x(Eterm* ERTS_RESTRICT ptr, Uint sz, Eterm** hpp, ErlOffHeap* off_h
 	    break;
 	case TAG_PRIMARY_LIST:
 	case TAG_PRIMARY_BOXED:
-	    *hp++ = byte_offset_ptr(val, offs);
+            if (empty_tuple_literal == val) {
```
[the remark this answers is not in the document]
jhogberg: `empty_tuple_literal` is const, so this pattern will catch errors like `empty_tuple_literal = val`. I'm pretty sure we've enabled warnings for assignments-in-ifs though, so I don't think we gain much by doing this.

https://en.wikipedia.org/wiki/Yoda_conditions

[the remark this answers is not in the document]
sverker: Yes, I think the compiler will catch assignments-in-ifs. Not a fan of Yoda conditions.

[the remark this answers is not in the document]
kjellwinblad: Thanks, fixed!

erts/emulator/beam/copy.c
kjellwinblad: Thanks, fixed!

erts/emulator/beam/erl_global_literals.h
kjellwinblad: I think I will keep it as it is since I agree with @sverker's comment and I don't think making it `extern Eterm ERTS_LIT_XYZ` will have any performance advantage in production builds (correct me if I'm wrong).

sverker: `erts_get_global_literal` should be an inline function to not cost any performance. We don't have link-time-optimization enabled by default.

kjellwinblad: I see, thanks. I have created a global variable for the empty tuple literal (1f5f661). The extern declaration is in `erl_global_literals.h` and the normal declaration is in `erl_init.c` as it contains many other global variables. Please let me know if you think there is a better place to place these.

kjellwinblad: Good point! Fixed in 76843a8

PR #5263, opened 2021-10-05
Fix syntax for inlining a function

dszoboszlay:
The compiler (unfortunately) doesn't understand the `-inline(...).` attribute, the proper syntax for inlining a function is `-compile({inline, ...}).`. 

This small patch updates the code generator of the `unicode_util` module to emit correct code.


dgud:
Wow, now I feel stupid.
How did you find that?

dszoboszlay:
I wasn't sure about the syntax for inlining functions, so I grepped for `^-inline` in our codebase. Surprisingly, there were only two hits:
* A module based on this code in https://github.com/benoitc/unicode_util_compat
* A line I added myself some time ago :rofl: 

Then I went on and checked in the OTP repo too. I feel equally stupid for making the same mistake myself.

PR #5267, opened 2021-10-06
ci: Use prebuilt archive on Windows build

wojtekmach:
This cuts the build time from ~30min to ~20min. On the flip side, we're now doing it serially (first "build" job and then the "windows" job) whereas previously the windows build run in parallel.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -44,7 +44,7 @@ jobs:
       WXWIDGETS_VERSION: 3.1.4
     name: Build Erlang/OTP on Windows
     runs-on: windows-latest
-    needs: pack
+    needs: build
     steps:
       - uses: Vampire/setup-wsl@v1
         with:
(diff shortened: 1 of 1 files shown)


garazdawi:
Why do you want to do this change?

garazdawi:
> > Why do you want to do this change
> 
> just trying to avoid unnecessary work. As I mentioned initially there might be a little slowdown actually given this job needs to wait for the other job. So if you’d rather not pursue it, it is fine.

The total time goes from [40 min](https://github.com/erlang/otp/actions/runs/1312977808) to [53 min](https://github.com/erlang/otp/actions/runs/1313826518).

I've tried to do similar optimizations for the different linux builds in the past and while it consumed less CPU time, it does increase lead-time which I think is more important.

I'll prepare a PR for the things I and @dgud talked about. That should speed up the windows builds a bit.

PR #5272, opened 2021-10-07
Fix #5257 - Clarify gen_event:add_sup_handler Re 'EXIT'

### LINKED ISSUE #5257: gen_event:add_sup_handler docs omit the passing on of the `'EXIT'` message on handler supervisor death
The documentation of `gen_event` does not mention deliberate passing-on of `{'EXIT', ...}` messages via `handle_info`. (It does mention that it `trap_exit`s, so it's technically possible to infer it with careful reading of that section and the `handle_info` docs -- I'm not a fan of keeping behaviours documented in the bottom of a locked filing cabinet stuck in a disused lavatory with a sign on the door saying ‘Beware of the Leopard', though...)

Test case:

```erlang
-module(test).

-export([s/1, w/2, init/1, handle_info/2]).

s(R) ->
        {ok, E} = gen_event:start_link(),
        spawn(fun () -> w(E, R) end),
        gen_event:add_handler(E, ?MODULE, 's'),
        timer:sleep(2000).

w(E, R) ->
        gen_event:add_sup_handler(E, ?MODULE, 'w'),
        timer:sleep(1000),
        exit(R).

init(X) -> {ok, X}.

handle_info(M, S) -> io:format("~p ~p~n", [M, S]), {ok, S}.
```

```erlang
4> c(test), test:c(blah).
** exception error: undefined function test:c/1
5> c(test), test:s(blah).
{'EXIT',<0.101.0>,blah} s
ok
6> c(test), test:s(normal).
{'EXIT',<0.108.0>,normal} s
ok
```

This `'EXIT'` message came as a surprise to our developers, and isn't mentioned directly in the [documentation](https://erlang.org/doc/man/gen_event.html#add_sup_handler-3). Assuming the documentation were the spec, the `{'EXIT', ...}` messages above are superfluous and hence in error -- and confuse code which intentionally sends and processes messages of that shape, expecting to be the owner of it. However, I imagine there is much code at this point relying on it. The code in `gen_event.erl` says:

```erlang
%% First terminate the supervised (if exists) handlers and
%% then inform other handlers.
%% We do not know if any handler really is interested but it
%% may be so !
handle_exit(From, Reason, MSL, SName) ->
    MSL1 = terminate_supervised(From, Reason, MSL, SName),
    {_,MSL2}=server_notify({'EXIT', From, Reason}, handle_info, MSL1, SName),
    MSL2.
```

It would be grand if the documentation could be updated to cover this behaviour -- I'd recommend adding a note to both:

* `add_sup_handler` (indicating that the mechanism of the supervision is a link, and hence that death of the process will be handled in `handle_info`), and
* `handle_info` (indicating that an example message would be the death of a handler supervisor added via `add_sup_handler`).

bucko909:
Make clear the semantics of `gen_event:add_sup_handler/3` as regards links and `'EXIT'` messages.

### DIFF
--- a/lib/stdlib/doc/src/gen_event.xml
+++ b/lib/stdlib/doc/src/gen_event.xml
@@ -207,14 +207,17 @@ gen_event:stop             ----->  Module:terminate/2
       <desc>
         <p>Adds a new event handler in the same way as
           <seemfa marker="#add_handler/3"><c>add_handler/3</c></seemfa>,
-          but also supervises the connection between the event handler
+          but also supervises the connection by linking the event handler
           and the calling process.</p>
         <list type="bulleted">
           <item>If the calling process later terminates with <c>Reason</c>,
-           the event manager deletes the event handler by calling
+           the event manager deletes any supervised event handlers by calling
            <seemfa marker="#Module:terminate/2">
-           <c>Module:terminate/2</c></seemfa>
-           with <c>{stop,Reason}</c> as argument.
+           <c>Module:terminate/2</c></seemfa>,
+           then calls
+           <seemfa marker="#Module:handle_info/2">
+           <c>Module:handle_info/2</c></seemfa>
+           for each remaining handler.
           </item>
           <item>
             <p>If the event handler is deleted later, the event manager
@@ -1005,6 +1008,14 @@ gen_event:stop             ----->  Module:terminate/2
           an event manager receives any other message than an event or
           a synchronous request (or a system message).</p>
         <p><c>Info</c> is the received message.</p>
+        <p>In particular, this callback will be made when a process terminated
+          after calling
+          <seemfa marker="#add_sup_handler/3">
+          <c>add_sup_handler/3</c></seemfa>. Any event handler
+          attached to an event manager which in turn has a supervised handler
+          should expect callbacks of the shape
+          <seemfa marker="#Module:handle_info/2">
+          <c>Module:handle_info({'EXIT', Pid, Reason}, State)</c></seemfa>.</p>
         <p>For a description of <c>State</c> and possible return values, see
           <seemfa marker="#Module:handle_event/2">
           <c>Module:handle_event/2</c></seemfa>.</p>
--- a/system/doc/design_principles/events.xml
+++ b/system/doc/design_principles/events.xml
@@ -227,7 +227,8 @@ ok</pre>
       <c>handle_info(Info, State)</c>
       must be implemented to handle them. Examples of other
       messages are exit messages, if the <c>gen_event</c> is linked to
-      other processes (than the supervisor) and trapping exit signals.</p>
+      other processes (than the supervisor, for example via
+      <c>add_sup_handler</c>) and trapping exit signals.</p>
     <code type="none">
 handle_info({'EXIT', Pid, Reason}, State) ->
     ..code to handle exits here..


lib/stdlib/doc/src/gen_event.xml
```diff
@@ -245,6 +245,16 @@ gen_event:stop             ----->  Module:terminate/2
                   Which term depends on the error.</p></item>
             </list>
           </item>
+          <item>
+              <p>If the calling process exits later, the event manager will
+              delete all handlers the calling process supervises (and call
+              <seemfa marker="#Module:terminate/2">
+              <c>Module:terminate({'EXIT', Pid, Reason}, State)</c></seemfa>),
+              then call
+              <seemfa market="#Module:handle_info/2">
```
u3s, line 254: using _marker_ attribute instead of _market_ will fix the docs build error.
you can check "make docs" command yourself before pushing

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -1005,6 +1015,12 @@ gen_event:stop             ----->  Module:terminate/2
           an event manager receives any other message than an event or
           a synchronous request (or a system message).</p>
         <p><c>Info</c> is the received message.</p>
+        <p>This callback will receive <c>'EXIT'</c> messages from processes
+          which terminated after calling
+          <seemfa market="#add_sup_handler/3">
```
u3s, line 1020: using _marker_ attribute instead of _market_ will fix the docs build error.
you can check "make docs" command yourself before pushing

bucko909: Build seems broken for reasons I'm not clear on (I just did `./configure && make docs`), so I'll let the Github CI tell me off:

```
 ERLC   ../ebin/edoc_layout_chunks.beam
edoc_layout_chunks.erl:58: can't find include lib "kernel/include/eep48.hrl"
edoc_layout_chunks.erl:61: record docs_v1 undefined
edoc_layout_chunks.erl:458: record docs_v1 undefined
edoc_layout_chunks.erl:457: Warning: variable 'Anno' is unused
edoc_layout_chunks.erl:457: Warning: variable 'Docs' is unused
edoc_layout_chunks.erl:457: Warning: variable 'Metadata' is unused
edoc_layout_chunks.erl:457: Warning: variable 'ModuleDoc' is unused
```

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -1005,6 +1015,12 @@ gen_event:stop             ----->  Module:terminate/2
           an event manager receives any other message than an event or
           a synchronous request (or a system message).</p>
         <p><c>Info</c> is the received message.</p>
+        <p>This callback will receive <c>'EXIT'</c> messages from processes
```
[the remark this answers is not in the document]
bucko909: I went with something like the above.

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -245,6 +245,16 @@ gen_event:stop             ----->  Module:terminate/2
...
+          <item>
+              <p>If the calling process exits later, the event manager will
+              delete all handlers the calling process supervises (and call
+              <seemfa marker="#Module:terminate/2">
+              <c>Module:terminate({'EXIT', Pid, Reason}, State)</c></seemfa>),
+              then call
+              <seemfa marker="#Module:handle_info/2">
+              <c>Module:handle_info({'EXIT', Pid, Reason}, State)</c></seemfa> for each
+              remaining handler.</p>
+          </item>
```
u3s, line 257: Can you please combine this bullet with bullet above starting with "<item>If the calling process later terminates" ?
I think they're describing same scenario and there is no point in splitting them.

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -245,6 +245,16 @@ gen_event:stop             ----->  Module:terminate/2
                   Which term depends on the error.</p></item>
             </list>
           </item>
+          <item>
+              <p>If the calling process exits later, the event manager will
+              delete all handlers the calling process supervises (and call
+              <seemfa marker="#Module:terminate/2">
+              <c>Module:terminate({'EXIT', Pid, Reason}, State)</c></seemfa>),
```
u3s, line 252: Module:terminate/2 callback will not receive {'EXIT', Pid, Reason} but {stop, Reason} instead.

I would suggest skipping argument examples in links - because arguments are described in detail for both Module:terminate/2 and Module:handle_info/2 in the linked sections. Repeated descriptions are harder to maintain.

lib/stdlib/doc/src/gen_event.xml
bucko909: You're right, I missed that.

bucko909:
OK, have combined bullets and removed duplicate references to argument shapes (including the one that was already present).

PR #5283, opened 2021-10-12
JOB_OBJECT_LIMIT_BREAKAWAY_OK

obi458:
Sometimes we want to break out, for example to start programs in another windows session.

https://docs.microsoft.com/en-us/windows/win32/api/jobapi2/nf-jobapi2-createjobobjectw

All processes associated with a job must run in the same session. 
A job is associated with the session of the first process to be assigned to the job.

### DIFF
--- a/erts/etc/win32/erlsrv/erlsrv_service.c
+++ b/erts/etc/win32/erlsrv/erlsrv_service.c
@@ -535,10 +535,14 @@ static BOOL start_a_service(ServerInfo *srvi){
 	  HANDLE hJob = CreateJobObject(NULL, NULL);
 	  JOBOBJECT_EXTENDED_LIMIT_INFORMATION jeli = { 0 };
 	  /*
-	   * Causes all processes associated with the job to terminate when the
-	   * last handle to the job is closed.
-	   */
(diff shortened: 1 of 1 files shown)


sverker:
@obi458 Sorry for letting this PR brew for so long. It would be good however if it could be complemented with

- Documentation. In https://www.erlang.org/doc/man/erlsrv.html#port-programs maybe.
- Test case. Ideally.
- A release note. I think you would formulate that much better than me.

sverker:
@obi458 Could you contribute something to write in the release note.

obi458:
> @obi458 Could you contribute something to write in the release note.

The erlang service in Windows is running as locale system and runs on session 0. If you want so start a new child process in Erlang with a different session number than 0, e.g. session 1 which belongs to a logged in user in Windows itself, the JOB_OBJECT_LIMIT_BREAKAWAY_OK is needed in the erlang service.
Therefore it is necessary to apply the JOB_OBJECT_LIMIT_BREAKAWAY_OK in the erlang service to be able to start a child process with a different session number in windows.
Without applying JOB_OBJECT_LIMIT_BREAKAWAY_OK you are not able to start a child process, the create process would fail.

PR #5286, opened 2021-10-13
jit: Improve gdb/perf support

jhogberg:
This PR helps external tools like `gdb` and `perf` describe where things happen in the _original source code_, which is very helpful when debugging the JIT or profiling code.

image

Since this is an emulator feature it works out of the box for all BEAM languages. The screenshot below is from `perf annotate` after profiling some Elixir code:
image


erts/emulator/beam/jit/beam_jit_common.cpp
```diff
@@ -313,6 +253,120 @@ void BeamModuleAssembler::codegen(char *buff, size_t len) {
                            CodeHolder::kCopyPadSectionBuffer);
 }
 
+void BeamModuleAssembler::register_metadata(const BeamCodeHeader *header) {
+#ifndef WIN32
+    const BeamCodeLineTab *line_table = header->line_table;
+
+    char name_buffer[MAX_ATOM_SZ_LIMIT];
+    std::string module_name = getAtom(mod);
+    std::vector<AsmRange> ranges;
```
garazdawi, line 262: Should we add a check here to return if no metadata is to be updated? So that we don't build the AsmRange structure in vain. My original thinking was that there was no need as when profiling this function did not show up, but maybe it is time to do something about it?

jhogberg: Ranges and lines improve core dumps, so they're more or less always needed. I'm not sure it's worth adding an option to control it.

garazdawi:
The perf section of the internal docs should also be updated: https://www.erlang.org/doc/apps/erts/BeamAsm.html#linux-perf-support

PR #5288, opened 2021-10-14
ssl: Fix testcase

dgud:
Test what it is documented to test.

### DIFF
--- a/lib/ssl/test/ssl_test_lib.erl
+++ b/lib/ssl/test/ssl_test_lib.erl
@@ -1710,16 +1710,20 @@ make_rsa_cert(Config) ->
             ServerChain = proplists:get_value(server_chain, Config, default_cert_chain_conf()),
             CertChainConf = gen_conf(rsa, rsa, ClientChain, ServerChain),
             GenCertData = public_key:pkix_test_data(CertChainConf),
+            #{client_config := ClientDerConf, server_config := ServerDerConf} = GenCertData,
+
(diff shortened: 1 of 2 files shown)


lib/ssl/test/ssl_test_lib.erl
```diff
@@ -1710,16 +1710,20 @@ make_rsa_cert(Config) ->
...
+	     {server_rsa_verify_opts, [{reuseaddr, true}, {verify, verify_peer} | ServerConf]},
 	     {client_rsa_opts, ClientConf},
-             {client_rsa_verify_opts,  [{verify, verify_peer} |ClientConf]}
-	     | Config];
+             {client_rsa_verify_opts,  [{verify, verify_peer} |ClientConf]},
+             {server_rsa_der_opts, [{reuseaddr, true} | ServerDerConf]},
+	     {server_rsa_der_verify_opts, [{reuseaddr, true}, {verify, verify_peer} | ServerDerConf]},
+	     {client_rsa_der_opts, ClientDerConf},
+             {client_rsa_der_verify_opts,  [{verify, verify_peer} |ClientDerConf]}
+            | Config];
```
u3s, line 1726: did this table became complex to read/verify/(maintain?) with this addition?
once I've re-ordered elements, so similar tuples are grouped, it became obvious (only for me?) that list contains pairs of almost duplicates.
if you agree, then maybe readability could be improved here and in make_rsa_1024_cert in some way.

PR #5290, opened 2021-10-15
Add [32-bit] to the Erlang shell title row

sverker:
As 32-bit is nowadays the more divergent one.

```
> erl
Erlang/OTP 25 [erts-12.1.2] [32-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]

Eshell V12.1.2  (abort with ^G)
1> 
```

### DIFF
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -92,8 +92,12 @@ static char erts_system_version[] = ("Erlang/OTP " ERLANG_OTP_RELEASE
 				     " [source]"
 #endif
 #endif	
-#ifdef ARCH_64
+#if defined(ARCH_64)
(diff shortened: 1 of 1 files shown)


erts/emulator/beam/erl_bif_info.c
```diff
@@ -92,8 +92,12 @@ static char erts_system_version[] = ("Erlang/OTP " ERLANG_OTP_RELEASE
 				     " [source]"
 #endif
 #endif	
-#ifdef ARCH_64
+#if defined(ARCH_64)
 				     " [64-bit]"
+#elif defined(ARCH_32)
+                                     " [32-bit]"
```
JeromeDeBretagne, line 98: Nitpick: this file is using a whitespace format based on tabs, it would be great to keep it consistent on L98

PR #5293, opened 2021-10-18
compiler: Add ASM form to compile:forms/2 spec

dumbbell:
This was missing so far because the assembly form was never specified.

This patch adds a type spec for the assembly form to the `beam_disasm` module and exports it. This location was chosen because it also defines the `#function{}` record which is used in the assembly form.

I'm not sure if this qualifies as a new feature or a bug fix. The pull request is against the `master` branch currently.


dumbbell:
This change was reverted in Erlang/OTP 27.0 by erlang/otp@b324095fc2c3a840ec1aa49f08abcbed081dd57e. The commit doesn't explain the reason. I take a note of this revert here for completeness.

PR #5295, opened 2021-10-19
Correct binary assignment notice

Maria-12648430:
While "`B=<<1>>` is interpreted as `B =<<1>>`" is kinda correct in a sense, it doesn't make the point clear, namely that `B=<<1>>` is interpreted as `B =< <1>>`. In https://erlang.org/doc/programming_examples/bit_syntax.html#lexical-note, it is correct.

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1265,7 +1265,7 @@ Ei = Value |
     <p>Notice that bit string patterns cannot be nested.</p>
     <p>Notice also that "<c><![CDATA[B=<<1>>]]></c>" is interpreted as
       "<c><![CDATA[B =< <1>>]]></c>" which is a syntax error. The correct way is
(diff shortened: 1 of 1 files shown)


system/doc/reference_manual/expressions.xml
```diff
@@ -1264,7 +1264,7 @@ Ei = Value |
 </pre>
     <p>Notice that bit string patterns cannot be nested.</p>
     <p>Notice also that "<c><![CDATA[B=<<1>>]]></c>" is interpreted as
-      "<c><![CDATA[B =<<1>>]]></c>" which is a syntax error. The correct way is
+      "<c><![CDATA[B =< <1>>]]></c>" which is a syntax error. The correct way is
       to write a space after '=': "<c><![CDATA[B= <<1>>]]></c>.</p>
```
vkatsuba, line 1268:
```suggestion
      to write a space after '=': "<c><![CDATA[B = <<1>>]]></c>.</p>
```

PR #5302, opened 2021-10-21
Add recursion in systools_make module checks

ferd:
In the current OTP release mechanisms, systools can be configured to do
a source file check to make sure any built module has a matching source
file. When such a source file is not found, a warning is emitted.

The current mechanism only supports looking in two directories: $app/src/
and $app/e_src/. This causes issues with existing OTP applications, such
as Wx, which uses $app/src/gen/ for auto-generated modules, for example,
meaning that any release relying on sub-directories as part of its build
process will not pass any of the sanity checks.

This patch changes the behaviour to look in all the subdirectories of
$app/src and $app/e_src (on top of these directories specifically)
for source files. This makes release creation warning-free for most
applications.

The only potential incompatibilities for this change would be:

- An OTP application which has various source files with the same name,
  where there is specifically one of them in a subdirectory of src/ and
  the desired one at the top-level of e_src/, where a new file is now
  going to be favoured

- Any subdirectory structure that loops onto itself or throws
  `filelib:wildcard/2` into exponential behaviour but somehow was not an
  issue when compiling the application normally might now be triggered.

These two possible incompatibilities are something we should consider
less likely than trying to build releases that contain the built-in Wx
OTP application, for example.


garazdawi:
Thanks!

Issue #5297, opened 2021-10-20
Get source code location in trace `call` messages


sidkshatriya:
trace `call` messages have the following format

`{trace, Pid, call, {M, F, Args}}`

See https://erlang.org/doc/man/erlang.html#trace-3 

This allows you to get the parameters of the function call easily. By using `return_trace` in a call match specification you can also get the return value of the function. These are very useful. But sometimes a function may have many invocations of a particular function (with the same arity) and it would be useful to disambiguate them.

__One feature that would be very useful is where exactly is the function call coming from in source code?__

So something like this would be useful 

`{trace, Pid, call_with_loc, {M, F, Args}, CodeLocation}`

Where CodeLocation could be something like `cowboy_http.erl:45` or `phoenix.ex:900`.

Questions:
- Can we accomplish something similar like this today?
- If not, would this be a useful feature to implement? Does it require a lot of changes to the Erlang/OTP code?
- Would the core team be interested in implementing something like this?
- How could one go about implementing such a feature?

jhogberg:
> I'll obviously need to get into the C-code to understand how things work. I'm not familiar with beam internals. Are there some good resources I can read to get a grip on the fundamentals?

The documentation we have mostly concerns how breakpoints are set (e.g. [Tracing.md](https://github.com/erlang/otp/blob/master/erts/emulator/internal_doc/Tracing.md) or the [tracing/nif loading section in BeamAsm.md](https://github.com/erlang/otp/blob/master/erts/emulator/internal_doc/BeamAsm.md#tracing-and-nif-loading)) since that's the trickiest part, but that's not very relevant here. I'll make an effort to expand the documentation.

In the meantime, I'll be happy to answer your questions. :)

> Could the following work (pseudo-code) ? ... snip ...

It would work but it would have a slight overhead on _all_ calls since it's impossible to know whether the callee is traced. The main benefit of our current method is that is has no overhead when tracing is inactive, and that's something we don't want to lose.

sidkshatriya:
Thanks. 

> It would work but it would have a slight overhead on all calls since it's impossible to know whether the callee is traced.

[...]

PR #5305, opened 2021-10-22
Tracing: Add new option `caller_line` for fun match spec body

sidkshatriya:
```
   Tracing: Add new option `caller_line` for fun match spec body
    
    This is similar to `caller`. `caller` gives you {M, F, Arity} while
    `caller_line` gives you {M, F, Arity, {File, Line}} where
    File:Line is the source location of the function call site.
    
    If the location cannot be found you will get {M, F, Arity, undefined}.
```
-------
(The description below is the outdated one and can be ignored)
```
Tracing: Add new option `caller_line` for fun match spec body

This is similar to `caller`. `caller` gives you {M, F, Arity} while
`caller_line` gives you {M, F, Arity, Line} where Line is the source
code line number in which module M is located.
```

This is my first attempt at #5297 -- please suggest improvements. 

**BTW thanks for the pointers and tips on how to approach this feature**. 

I was able to create the new option `caller_line` by using the code for `caller` which is quite similar.

Continuing with the example mentioned here: https://github.com/erlang/otp/issues/5297#issuecomment-948587395

```erlang
-module(test_prog).
-export([foo/0]).

foo() ->
    lots(),
    ok.

lots() ->
    'of'().

'of'() ->
    indirections().

indirections() ->
    bar(10).

bar(0) ->
    done;
bar(N) ->
    bar(N - 1).
```

**And using the trace pattern:**

```elixir
 :erlang.trace_pattern(
      {:test_prog, :_, :_},
      [{:_, [], [{:return_trace}, {:message, {:caller_line}}]}],
      [:local]
    )
```
Notice `caller_line` (which is what this PR adds) instead of `caller` in the `:erlang.trace_pattern` 

**The trace from my custom tracer when invoking `test_prog:foo()` looks like this:**

```
1 call: :test_prog.foo() (from {erl_eval,do_apply,6,685})
2 call: :test_prog.lots() (from {test_prog,foo,0,16777221})
3 call: :test_prog.of() (from {test_prog,foo,0,16777221})
4 call: :test_prog.indirections() (from {test_prog,foo,0,16777221})
5 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
6 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
7 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
8 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
9 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
10 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
11 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
12 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
13 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
14 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
15 call: :test_prog.bar() (from {test_prog,foo,0,16777221})
```

- The `caller_line` option sends `{M, F, Arity, Line}`. **The first line 685 is correct**
- The rest of the line numbers are wrong, I'm guessing this is because of tail recursion
- For some reason, running traces on Elixir code also always gives junk line numbers 16777XXX though I need to explore this some more and see if it will work in some situations
 
cc: @jhogberg @garazdawi

### DIFF
--- a/erts/doc/src/match_spec.xml
+++ b/erts/doc/src/match_spec.xml
@@ -140,6 +140,7 @@
         <c><![CDATA[process_dump]]></c> | <c><![CDATA[enable_trace]]></c> |
         <c><![CDATA[disable_trace]]></c> | <c><![CDATA[trace]]></c> |
         <c><![CDATA[display]]></c> | <c><![CDATA[caller]]></c> |
+        <c><![CDATA[caller_line]]></c> |
         <c><![CDATA[set_tcw]]></c> | <c><![CDATA[silent]]></c>
       </item>
     </list>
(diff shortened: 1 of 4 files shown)


garazdawi:
Thanks!

arcusfelis:
Dear random reader.
If you wondering how to use it with recon:
```
recon_trace:calls({queue, new, [{'_', [], [{return_trace}, {message, {caller_line}}]}]}, 1000, []).
1
(mongooseim2@localhost)26> queue:new().

9:50:09.362967 <0.1645.0> queue:new() {erl_eval,do_apply,7,{"erl_eval.erl",750}}
{[],[]}

9:50:09.363141 <0.1645.0> queue:new/0 --> {[],[]}
```

(I am leaving it here because it is not documented).

PR #5316, opened 2021-10-25
jit: Optimize instructions based on operand types

jhogberg:
This PR lets the JIT optimize instructions by their operand types. The current implementation settles for removing redundant tag tests and the like, but we plan to add more aggressive optimizations in the near future, such as reducing arithmetic to single machine instructions when we've determined they can't overflow.


itssundeep:
This diff breaks the following example.
```
-module(testcode).
-compile(warn_missing_spec).

%% API
-export([
  usec_to_time/1
]).

%% Converts Usec to erlang 3-tuple time
-spec usec_to_time(integer()) -> erlang:timestamp().
usec_to_time(Usec) ->
  Us = Usec rem 1000000,
  Sec = (Usec div 1000000) rem 1000000,
  MegaSec = (Usec div 1000000) div 1000000,
  {MegaSec, Sec, Us}.
  ```
  ```
  16> testcode:usec_to_time(1530342000*1000).
{1,0,0}
```

Can someone please look into this. I think it is failing for larger integers.

bjorng:
@itssundeep Thanks, I have created an issue to make sure that we don't forget to fix the bug.

Issue #5314, opened 2021-10-24
erlang-mode is broken with GNU Emacs HEAD


wahjava:
**Describe the bug**

Since emacs-mirror/emacs@86da812afb2572c7fead2bb07570b976bffd7c55 (guessing), `erlang-mode` is broken. Following is a helpful backtrace:

```
Debugger entered--Lisp error: (error "Given parent class xref-file-location is not a cla...")
  signal(error ("Given parent class xref-file-location is not a cla..."))
  error("Given parent class %S is not a class" xref-file-location)
  eieio-defclass-internal(erlang-xref-location (xref-file-location) ((arity :type fixnum :initarg :arity :reader erlang-xref-location-arity)) (:documentation "An erlang location is a file location plus arity."))
  (progn (defalias 'erlang-xref-location-p (eieio-make-class-predicate 'erlang-xref-location)) (defalias 'erlang-xref-location--eieio-childp (eieio-make-child-predicate 'erlang-xref-location)) (defalias 'erlang-xref-location-child-p #'erlang-xref-location--eieio-childp) (make-obsolete 'erlang-xref-location-child-p "use (cl-typep ... \\='erlang-xref-location) instead" "25.1") (define-symbol-prop 'erlang-xref-location 'cl-deftype-satisfies #'erlang-xref-location--eieio-childp) (eieio-defclass-internal 'erlang-xref-location '(xref-file-location) '((arity :type fixnum :initarg :arity :reader erlang-xref-location-arity)) '(:documentation "An erlang location is a file location plus arity.")) (cl-defmethod erlang-xref-location-arity ((this erlang-xref-location)) "Access the slot `arity' from object of class `erla..." (slot-value this 'arity)) (defun erlang-xref-location (&rest slots) "Create a new object of class type `erlang-xref-loc..." (declare (compiler-macro (lambda (whole) (if (not (stringp ...)) whole (macroexp-warn-and-return (format "Obsolete name arg %S to constructor %S" ... ...) `...))))) (apply #'make-instance 'erlang-xref-location slots)))
  (defclass erlang-xref-location (xref-file-location) ((arity :type fixnum :initarg :arity :reader erlang-xref-location-arity)) :documentation "An erlang location is a file location plus arity.")
[...]

PR #5324, opened 2021-10-26
Fix the emacs erlang-mode: port EIEIO to CL-lib

juster:
This is a backwards-incompatible fix for issue #5314. More work will be
necessary to be compatible with both new and old versions of xref.

Xref is a package which is also bundled with emacs:
http://elpa.gnu.org/packages/xref.html

Xref switched its own internally defined classes from EIEIO to CL-lib at
emacs-mirror/emacs@86da812afb2572c7fead2bb07570b976bffd7c55.

Modifications:
- Changes erlang-xref-location to be a struct which "includes"
  xref-file-location instead of a class which inherits.
- Avoids calling make-instance to create an object from a class.
- Updates comments that make references to classes.

Testing:
1. Install the latest version of the Xref package in Emacs:
   (in emacs) M-x package-install RET xref RET
2. Verify that opening any erlang source file will cause an error.
3. Update erlang-mode with this patch and reload erlang-mode (i.e.
   restart emacs).
4. Verify that opening any erlang file will not cause an error.
5. Generate a TAGS file for the project with Exuberant Tags:
   (in shell) ctags -e *.erl
6. Verify xref works. Move the cursor to a function call and find
   the definition:
   (in emacs) M-x xref-find-definitions

### DIFF
--- a/lib/tools/emacs/erlang.el
+++ b/lib/tools/emacs/erlang.el
@@ -5120,14 +5120,19 @@ about Erlang modules."
 ;;
 ;; As mentioned this xref implementation is based on the etags xref
 ;; implementation.  But in the cases where arity is considered the
-;; etags information structures (class xref-etags-location) will be
-;; translated to our own structures which include arity (class
+;; etags information structures (struct xref-etags-location) will be
+;; translated to our own structures which include arity (struct
 ;; erlang-xref-location).  This translation is started in the function
 ;; `erlang-refine-xrefs'.
 
 ;; I mention this as a head up that some of the functions below deal
 ;; with xref items with xref-etags-location and some deal with xref
 ;; items with erlang-xref-location.
+
+;; NOTE: Around Sept 2021, the xrefs package changed all of its defined types
+;; (i.e.  xref-location, xref-file-location) from EIEIO classes to CL-Lib
+;; structures. These are both supported. Older Emacsen with earlier versions of
+;; xref will continue to use defclass. Newer Emacsen will use cl-defstruct.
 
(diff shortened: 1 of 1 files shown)


juster:
Ok I think it is finished. If there is an EIEIO class bound to _xref-file-location_ then we use EIEIO's _defclass_ to subclass it. Otherwise we use _cl-defstruct_. In order for either case to be equivalent the code now avoids checking/calling anything other than the constructor, accessor, and predicate functions.

I added testing instruction in the commit message. I used the automated tests in the _erlang-test.el_ file in the same directory. The test succeeded with both new and old xref on Emacs 27.2 and 26.1.

juster:
I added a two-line change and I successfully ran `make test ARGS='-suite emacs_SUITE'`.

That is a byte code compile error (or warning in more recent emacs versions). One can see it when trying to compile the erlang.el file:
```
emacs --batch --execute '(byte-compile-file "erlang.el")'
```

I now see this is why there are many checks for bound functions (with _fboundp_). I have added a check to ensure the _xref-make_ function is bound at line 5375:

```
  (when (and (fboundp 'make-erlang-xref-location)
             (fboundp 'xref-make))
```

This wraps the call to _xref-make_ and should hide the warning/error.

rudi:
I confirm that the patch fixes the problem for me (Emacs 27.2, xref 1.3.2).  Remark that xref might be updated for emacs versions >= 26.1, so the bug might start appearing on previously-unaffected Emacs installations when the user installs a package that pulls in a newer version of xref than the one released with Emacs.

PR #5329, opened 2021-10-28
Mark FUN_EXT as deprecated

michalmuskala:
Since OTP 23 the `FUN_EXT` encoding is [not decoded](https://github.com/erlang/otp/blob/f0cf38a32b957d956be4170e24ad72f8fca3e3f6/erts/emulator/beam/external.c#L5667-L5672) by `binary_to_term`. It should be marked deprecated as other no longer used tags.

### DIFF
--- a/erts/doc/src/erl_ext_dist.xml
+++ b/erts/doc/src/erl_ext_dist.xml
@@ -1106,7 +1106,7 @@
 
     <section>
       <marker id="FUN_EXT"/>
-      <title>FUN_EXT</title>
+      <title>FUN_EXT (removed)</title>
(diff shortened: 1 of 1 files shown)


sverker:
Suggestion:
- Change _deprecated_ to _removed_.
- Add note to say not encoded since OTP R8 and not decoded since OTP 23.
- Remove the description text but keep the table.
- Remove word "new" in "This is the new encoding..." for NEW_FUN_EXT.

sverker (approved):

sverker:
Thanks for the contribution.

PR #5335, opened 2021-10-29
common_test: fix spec

attah:
Add missing option to spec for ct_gen_conn:start/4.
Without it, dialyzer thinks ct_telnet:open would always fail
since it supplies this option.

CC: @tomas-abrahamsson @edvreu


attah:
Hmm, turns out this should have been towards the maint branch. Let's see if i can fix.

attah:
Okay, maybe now it is finally correct.

PR #5337, opened 2021-11-01
kernel: Fix erl_epmd:address_please

### LINKED ISSUE #5334: inet_tcp_dist / inet_tls_dist ignores custom EPMD name resolution mechanism
**Describe the bug**
When using custom EPMD module with address_please/3 defined, ip resolving of a dist connection should always go through this function. However right now in inet_tcp_dist:select/1 and inet_tls_dist:select/1, it can end up calling inet:gethostbyname_tm, which is not desired.

**To Reproduce**
Terminal 1:
erl -sname "a@a" -epmd_module my_epmd

Terminal 2:
erl -sname "b@a" -epmd_module my_epmd
net_kernel:connect_node('a@a').
This will return false

However, if we do this, it will success
Terminal 1:
erl -sname "a@localhost" -epmd_module my_epmd

Terminal 2:
erl -sname "b@localhost" -epmd_module my_epmd
net_kernel:connect_node('a@localhost').
true

[my_epmd.zip](https://github.com/erlang/otp/files/7437488/my_epmd.zip)

**Expected behavior**
inet_tcp_dist:select/1 and inet_tls_dist:select/1 should respect EpmdModule:address_please/3 if it is defined

net_kernel:connect_node('a@a') should success

**Affected versions**

garazdawi:
The initial check for ipv4 vs ipv6 did not use the
erl_epmd interface to lookup the address which caused
invalid addresses to not be resolvable by erl_epmd callback
modules.

Closes #5334

### DIFF
--- a/erts/emulator/test/distribution_SUITE.erl
+++ b/erts/emulator/test/distribution_SUITE.erl
@@ -2558,17 +2558,23 @@ no_epmd(Config) when is_list(Config) ->
 epmd_module(Config) when is_list(Config) ->
     %% We need a relay node to test this, since the test node uses the
     %% standard epmd module.
-    Sock1 = start_relay_node(epmd_module_node1, "-epmd_module " ++ ?MODULE_STRING),
-    Node1 = inet_rpc_nodename(Sock1),
+    {N1,_,S1} = Sock1 = start_relay_node(
+                          epmd_module_node1,
+                          "-epmd_module " ++ ?MODULE_STRING,
+                         [{"ERL_AFLAGS","-sname epmd_module_node1@dummy " ++ os:getenv("ERL_AFLAGS","")}]),
+    Node1 = inet_rpc_nodename({N1,"dummy",S1}),
     %% Ask what port it's listening on - it won't have registered with
     %% epmd.
     {ok, {ok, Port1}} = do_inet_rpc(Sock1, application, get_env, [kernel, dist_listen_port]),
 
     %% Start a second node, passing the port number as a secret
     %% argument.
-    Sock2 = start_relay_node(epmd_module_node2, "-epmd_module " ++ ?MODULE_STRING
-			     ++ " -other_node_port " ++ integer_to_list(Port1)),
-    Node2 = inet_rpc_nodename(Sock2),
+    {N2,_,S2} = Sock2 = start_relay_node(
+                          epmd_module_node2,
+                          "-epmd_module " ++ ?MODULE_STRING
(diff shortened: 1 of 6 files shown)


lib/ssl/src/inet_tls_dist.erl
```diff
@@ -582,8 +574,8 @@ do_setup_connect(Driver, Kernel, Node, Address, Ip, TcpPort, Version, Type, MyNo
     Opts =  trace(connect_options(get_ssl_options(client))),
     dist_util:reset_timer(Timer),
     case ssl:connect(
-        Address, TcpPort,
-        [binary, {active, false}, {packet, 4},
+        Ip, TcpPort,
+        [binary, {active, false}, {packet, 4}, {server_name_indication, Address},
```
u3s, line 578: why this adjustment is needed?

lib/ssl/test/ssl_dist_SUITE.erl
```diff
@@ -58,7 +58,9 @@
          verify_fun_fail/0,
          verify_fun_fail/1,
          verify_fun_pass/0,
-         verify_fun_pass/1
+         verify_fun_pass/1,
+         epmd_module/0,
```
u3s, line 62: any reason to not add epmd_module testcase to all?
its not executed when whole SUITE is run

lib/ssl/src/inet_tls_dist.erl
garazdawi: It is needed because erl_epmd:address_please/3 will return an IP for the address it is given and the original address does not have to have a valid DNS record.

lib/ssl/test/ssl_dist_SUITE.erl
garazdawi: I forgot! Will fix!

erts/emulator/test/distribution_SUITE.erl
```diff
@@ -2608,7 +2614,7 @@ port_please(_Name, _Ip) ->
 	    {port, Port, Version}
     end.
 
-address_please(_Name, _Address, _AddressFamily) ->
+address_please(_Name, "dummy", _AddressFamily) ->
```
IngelaAndin, line 2617: Would it not make sense to make the fallback consider the AdressFamily? Maybe this test is never run over ipv6?

garazdawi: hmm, good question.

Issue #5279, opened 2021-10-12
gen_udp:connect/3 changes local port number


benbro:
**Describe the bug**
When opening a udp socket with port=0 and calling gen_udp:connect/3 the local port changes:

**To Reproduce**
```erlang
{ok, S} = gen_udp:open(0).
{ok, Port1} = inet:port(S).
gen_udp:connect(S, {127,0,0,1}, 10000).
{ok, Port2} = inet:port(S).
Port1 =/= Port2.
```
**Expected behavior**
The local port should not change.

**Affected versions**
OTP 24.1.2

**Additional context**
Maybe related to PR #5120

benbro:
I can't reproduce it with otp 24.1.2 running on ubuntu 20.04.
Still trying.
```erlang
{ok,S1}=gen_udp:open(10000, [{ip, {127,0,0,1}}]).
{ok,S2}=gen_udp:open(20000, [{ip, {127,0,0,1}}]).
gen_udp:send(S1,{127,0,0,1},20000,<<"test 1">>).
flush().
gen_udp:connect(S2, {127,0,0,1}, 10000).
gen_udp:send(S1,{127,0,0,1},20000,<<"test 2">>).
flush().
```

RaimoNiskanen:
I am for documenting `gen_udp:connect/3`!

benbro:
@RaimoNiskanen I understand not wanting to workaround a Linux kernel bug.
I need the source port to be constant because I advertise it to the client as part of DTLS handshake. Instead of using port 0 for open I could choose a random port in the ephemeral port range and retry on failure but it can get inefficient when large number of ports are already used. Is there other option?

RaimoNiskanen:
Then I guess we will have another look at this...

RaimoNiskanen:
@benbro: I have created a fix that it would be very nice if you could evaluate, in PR #5343

benbro:
Thank you for the PR.
Does the socket backend considered stable enough to be used in production?
I've verified that the local port doesn't change after the first call to gen_udp:connect/3 with both backends:
```erlang
{ok, S} = gen_udp:open(0).
{ok, Port1} = inet:port(S).
gen_udp:connect(S, {127,0,0,1}, 10000).
{ok, Port2} = inet:port(S).
Port1 =:= Port2.
```
```erlang
{ok, S} = gen_udp:open(0, [{inet_backend, socket}]).
{ok, Port1} = inet:port(S).
[...]

PR #5343, opened 2021-11-03
Linux dgram reconnect

RaimoNiskanen:
For Linux, when connecting; dissolve the current association only if connected, that is; not for the socket's first connection.

I used a simple Erlang-only solution for the `gen_udp_socket` legacy compatibility module, so now a `gen_udp:connect/3` with `inet_backend = socket` calls `os:type/0`, for Linux `socket:peername/1`, and finally `socket:connect/2`.  A faster solution could use a new level `otp` socket option to automatically dissolve before reconnect and then do without the two prelude calls.  But we can take that path if it turns out to be actually needed...

Solves GH #5279.


RaimoNiskanen:
It seems I had the wrong target branch for this PR, so yes, it is merged to 'master', but do not fear - it is also merged into 'maint', and targeted for 'maint-24' i.e to be released in OTP-24.1.5.

PR #5344, opened 2021-11-03
supervisor: Add missing shutdown clauses

lhoguin:
This prevents annoying shutdown_error logs from being printed
in cases both the process and its supervisor started shutting
down at the same time.

See https://github.com/erlang/otp/pull/5201#issuecomment-951004125


lhoguin:
Wait I think that's not enough.

lhoguin:
This comment has the rationale behind the "not is_permanent": https://github.com/erlang/otp/pull/5201#discussion_r710896389

I am still trying to figure out why we get some logs in some cases where we shouldn't (`shutdown_error` with reason `shutdown`).

lhoguin:
Ready.

Maria-12648430 (approved):

juhlig (approved):

PR #5348, opened 2021-11-04
Do not bind on FD with undefined ifaddr

### LINKED ISSUE #4680: gen_tcp socket based implementation do not work with FDs passed from systemd

hauleth:
This is behaviour exposed by the `inet` backend and also the expected behaviour, as it allows to use `socket` backend together with systemd socket activation where the application receives bound socket with messages already in queue.

Close #4680

### DIFF
--- a/lib/kernel/src/gen_tcp_socket.erl
+++ b/lib/kernel/src/gen_tcp_socket.erl
@@ -198,7 +198,9 @@ extra_opts(OpenOpts) when is_list(OpenOpts) ->
    maps:from_list(OpenOpts).
 
 
-default_any(Domain, undefined = Undefined) ->
+default_any(_Domain, undefined, #{fd := _}) ->
+    undefined;
+default_any(Domain, undefined, _Opts) ->
     if
         Domain =:= inet;
         Domain =:= inet6 ->
@@ -206,9 +208,9 @@ default_any(Domain, undefined = Undefined) ->
               addr   => any,
               port   => 0};
         true ->
-            Undefined
+            undefined
     end;
-default_any(_Domain, BindAddr) ->
+default_any(_Domain, BindAddr, _Opts) ->
     BindAddr.
 
 bind_addr(_Domain, BindIP, BindPort)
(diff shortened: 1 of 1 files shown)


RaimoNiskanen:
I tried to move this to 'maint' but that failed miserably.  I think this correction should be in an earlier release.
Can you rebase it to 'maint' or maybe OTP-24.1.4 where you spotted the problem?

hauleth:
Sure

hauleth:
@RaimoNiskanen done

RaimoNiskanen:

PR #5349, opened 2021-11-05
Implement the compiler part of EEP 49

bjorng:
This pull request implements the compiler part of EEP 49 (Value-Based Error
Handling Mechanisms).

There is one difference compared to the current version of the EEP 49
document, namely that the matching operator now is `?=` instead of
`<~`. Thus the rewritten Mnesia code from the EEP will now look like:

    commit_write(OpaqueData) ->
        maybe
            ok ?= disk_log:sync(OpaqueData#backup.file_desc),
            ok ?= disk_log:close(OpaqueData#backup.file_desc),
            ok ?= file:rename(OpaqueData#backup.tmp_file, OpaqueData#backup.file),
            {ok, OpaqueData#backup.file}
        end.

The reason for the change is that `<-` and `<~` look too similar to
each other (especially with some fonts and text sizes) and could
easily be confused. `?=` does not have that problem, and its similarity
to `=` suggests that it is a matching operator.

The preprocessor uses `?` to signal the start of a macro invocation.
Although `=` is a valid macro name, that does not cause any ambiguity,
because macro names that are not atoms or variables must be
single-quoted. Therefore, `?=` is not a valid invocation of the `=`
macro; the invocation of the `=` macro must be written like `?'='`.

The `maybe` ... `else` ... `end` construct is only allowed to be used
if the compiler option `{enable_feature,maybe}` has been given. If
not given, `maybe` and `else` will be interpreted as atoms, and code
that use those atoms can be compiled. As currently implemented, the
`enable_feature` option will not be recognized if given in the source
code for the module. This limitation will be lifted later (there is
ongoing work on a more robust and general implementation of
`enable_feature`).

This PR supersedes #5340.

### DIFF
--- a/lib/compiler/src/beam_kernel_to_ssa.erl
+++ b/lib/compiler/src/beam_kernel_to_ssa.erl
@@ -133,20 +133,31 @@ cg(#k_return{args=[Ret0]}, St) ->
 cg(#k_break{args=Bs}, #cg{break=Br}=St) ->
     Args = ssa_args(Bs, St),
     {[#cg_break{args=Args,phi=Br}],St};
-cg(#k_letrec_goto{label=Label,first=First,then=Then,ret=Rs},
+cg(#k_letrec_goto{label=Label,vars=Vs0,first=First,then=Then,ret=Rs},
    #cg{break=OldBreak,labels=Labels0}=St0) ->
     {Tf,St1} = new_label(St0),
     {B,St2} = new_label(St1),
     Labels = Labels0#{Label=>Tf},
-    {Fis,St3} = cg(First, St2#cg{labels=Labels,break=B}),
-    {Sis,St4} = cg(Then, St3),
-    St5 = St4#cg{labels=Labels0},
-    {BreakVars,St} = new_ssa_vars(Rs, St5),
-    Phi = #cg_phi{vars=BreakVars},
-    {Fis ++ [{label,Tf}] ++ Sis ++ [{label,B},Phi],St#cg{break=OldBreak}};
-cg(#k_goto{label=Label}, #cg{labels=Labels}=St) ->
+    {Vs,St3} = new_ssa_vars(Vs0, St2),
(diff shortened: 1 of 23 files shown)


leoliu:
I have been thinking (wishfully) if reusing try ... end can be more general.

We allow condition (exception) classes to have default handler. The existing `throw`, `exit`, and `error` has no default handler so must be `catch`ed or crash.

We then introduce new condition class `nomatch` which is triggered by the new operator `?=`. The default handler will return the unmatched value if no custom handler is provided. So the example can be written as:

```erlang
commit_write(OpaqueData) ->
    try
        ok ?= disk_log:sync(OpaqueData#backup.file_desc),
        ok ?= disk_log:close(OpaqueData#backup.file_desc),
        ok ?= file:rename(OpaqueData#backup.tmp_file, OpaqueData#backup.file),
        {ok, OpaqueData#backup.file}
    end.
```

Users can override the default handler like so:
```erlang
commit_write(OpaqueData) ->
    try
        ok ?= disk_log:sync(OpaqueData#backup.file_desc),
        ok ?= disk_log:close(OpaqueData#backup.file_desc),
        ok ?= file:rename(OpaqueData#backup.tmp_file, OpaqueData#backup.file),
        {ok, OpaqueData#backup.file}
    catch
      nomatch:{error, disk_full} -> crash;
      nomatch:{error, no_permission} -> ok
    end.
```

In future there may be new operators or primitives that trigger new conditions and we are ready.

ferd:
The weakness of using exceptions is that they are not properly scoped; if it is not caught locally, it can be caught by the next exception handler on top, which loses locality of the value. By comparison, the match mechanism can be local but if it doesn't match, it _transforms_ into a `badmatch` error, which puts a scope boundary to where it can be confused for a local return value when it isn't.

There was a brief (and more detailed) discussion of it back in 2018 in the eeps-erlang mailing list when considering potential designs: http://erlang.org/pipermail/eeps/2018-December/000612.html

bjorng:
Superseded by #5405.

PR #5355, opened 2021-11-08
Fix forced distributed signals

### LINKED ISSUE #4989: process waits spawn_opt/5 more than 4 hours
A gen_server process is stuck when executing spawn/4 call. Process info shows spawn_opt/5 as a current function and status is waiting:

```
> process_info(P).
[{current_function,{erlang,spawn_opt,5}},
 {status,waiting},
 {message_queue_len,13},
 {trap_exit,false},
 {priority,normal},
 ...]
```

I checked 13 messages of the gen_server, they were all specific to my application - no "{spawn_reply, Ref, ok|error, Pid|Error}", just usual '$gen_cast' and 'system' tagged messages. Judging from messages, the caller was blocked for about 4 hours when I noticed that. I haven't increased net_ticktime.

Current stacktrace looks like:

```
> process_info(P, current_stacktrace).
{current_stacktrace,[{erlang,spawn_opt,5,[]},
                     {erlang,spawn,4,[]},
                     ...
                     {my_gen_server,handle_cast,2,
                                    [{file,"..."},
                                     {line,...}]},
                     {gen_server,try_dispatch,4,
[{file,"gen_server.erl"},{line,695}]},
                     {gen_server,handle_msg,6,
[{file,"gen_server.erl"},{line,771}]},
                     {proc_lib,init_p_do_apply,3,
[{file,"proc_lib.erl"},{line,226}]}]}
```

The code is simple: gen_server just spawns a function on several connected nodes. The function doesn't interact with the gen_server that calls spawn/4. The code worked for a couple of years: I can't be 100% sure if this happened before or not, but in any case I see this situation for the first time. A recent change of my system is upgrade to OTP24.

```
handle_cast(do_start, State) ->
    try
        my_start()
    catch
        What:Why:Stacktrace ->
            logger:warning("~p:~p~n~p", [What, Why, Stacktrace], #{mod => ?MODULE, line => ?LINE})
    end,
    {noreply, State};

my_start() ->
    [spawn(Node, ?MODULE, my_action, [erlang:system_time(1)]) || Node <- [node() | my_cluster_nodes()]].

my_cluster_nodes() ->
    {L, _} = lists:unzip(ets:tab2list(?TBL)),
    L.

my_action(Now) ->
    % work with ETS, use an external database service
```

All nodes of the cluster are ordinary Erlang node, and each node runs the same version of Erlang/OTP 24.0.2:

`Erlang/OTP 24 [erts-12.0.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]`

OS is Ubuntu 20 LTS.

The environment is Kubernetes with istio used for networking, that means that the code runs inside docker container, and networking is complicated. It's quite possible that one of nodes of the cluster was restarted abruptly, and may be it was related to version upgrade of istio networking, so we have either restart of a node or a possible glitch of networking to break connection.

### LINKED ISSUE #5346: waiting for spawn_reply indefinitely in rpc:multicall/5
**Describe the bug**

`rpc:multicall/5` with `infinity` timeout got stuck waiting for `spawn_reply`

**To Reproduce**

It happened twice to us during load tests, but we are so far unable to find a minimal environment to reproduce.

**Expected behavior**

`spawn_request` should always send back `reply_tag`

**Affected versions**

Our test was done on OTP23.2.7.2

**Additional context**

The application code makes an `rpc:multicall`, towards 3 nodes, timeout=`infinity`
And gets stuck in states like below

```
process_info(<0.9.1000>, current_stacktrace).
{current_stacktrace,[{rpc,mc_spawn_replies,5,
                          [{file,"rpc.erl"},{line,833}]},
                     {rpc,multicall,5,[{file,"rpc.erl"},{line,774}]},
...
```

And the process has `DOWN` messages in its mailbox

```
process_info(<0.9.1000>, messages).          
{messages,[{'DOWN',#Ref<0.3171699583.3710648329.31880>,
                   process,<0.16590.952>,
                   {#Ref<0.3171699583.3710648329.31879>,return,
                    {true,['emqx@10.*.*.241']}}},
           {'DOWN',#Ref<0.3171699583.3710648329.31882>,process,
                   <60658.18472.957>,
                   {#Ref<0.3171699583.3710648329.31879>,return,
                    {true,['emqx@10.*.*.143']}}},
           {'DOWN',#Ref<0.3171699583.3710648329.31881>,process,
                   <60657.19060.942>,
                   {#Ref<0.3171699583.3710648329.31879>,return,
                    {true,['emqx@10.*.*.9']}}},
```

There were around 100K such processes as a result of a load test.

EDIT:
`{true,['emqx@10.*.*.241']}` is the 'normal' return value from the remote call (passed as a DOWN message to the caller).
meaning the remote call was successful, but for some reason `spawn_reply` was either not sent or lost.

rickard-green:
Fixes #5346
Fixes #4989

### DIFF
--- a/erts/emulator/test/signal_SUITE.erl
+++ b/erts/emulator/test/signal_SUITE.erl
@@ -38,6 +38,7 @@
          kill2killed/1,
          contended_signal_handling/1,
          busy_dist_exit_signal/1,
+         busy_dist_demonitor_signal/1,
          busy_dist_down_signal/1,
          busy_dist_spawn_reply_signal/1,
          busy_dist_unlink_ack_signal/1]).
@@ -63,6 +64,7 @@ all() ->
      kill2killed,
      contended_signal_handling,
      busy_dist_exit_signal,
+     busy_dist_demonitor_signal,
      busy_dist_down_signal,
      busy_dist_spawn_reply_signal,
      busy_dist_unlink_ack_signal].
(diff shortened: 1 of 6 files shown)


max-au:
That's a good catch! I imagine that to test this, an alternative distribution controller would be needed (so reliably trigger busy dist port).

rickard-green:
Yes, my plan is to use the [gen_tcp_dist](https://github.com/erlang/otp/blob/maint-24/lib/kernel/examples/gen_tcp_dist/src/gen_tcp_dist.erl) example from kernel and suspend the input controller on the node initiating the `spawn_request()`.

erts/emulator/test/signal_SUITE.erl
```diff
@@ -255,6 +257,55 @@ busy_dist_exit_signal(Config) when is_list(Config) ->
     stop_node(OtherNode),
     ok.
 
+busy_dist_demonitor_signal(Config) when is_list(Config) ->
+    BusyTime = 1000,
+    {ok, BusyChannelNode} = start_node(Config),
+    {ok, OtherNode} = start_node(Config, "-proto_dist gen_tcp"),
```
max-au, line 263: Now with `peer` landed, may I suggest to use `{ok, Peer, Node} = ?CT_PEER(["-proto_dist", "gen_tcp"]),` instead?

rickard-green: Yes, I've already rewritten it like that in a branch made for merging into master. All of these branches that I've pushed to github are for merging into branches preceding master which lacks peer support, so they need to use the old functionality.

PR #5357, opened 2021-11-09
Fix typos for doc in erts

kianmeng:


erts/doc/src/notes.xml
```diff
@@ -16200,7 +16200,7 @@
           <p>
 	    The effect of the deprecated environment variable
 	    ERL_MAX_PORTS had been removed premeturely. It has now
-	    been readded. Note that this is still scheduled to be
+	    been read. Note that this is still scheduled to be
```
garazdawi, line 16203:
```suggestion
	    been re-added. Note that this is still scheduled to be
```

garazdawi (changes requested):
Thanks for this! I just found a small mistake that you did, when you fix that we can merge this.

erts/doc/src/notes.xml
sverker: While at it, fix misspelled "prem**a**turely" in the line above.

garazdawi:
Thanks!

PR #5358, opened 2021-11-09
stdlib: Add helper to support round tripping using erl_pp:form/{1,2}

frej:
The Erlang compiler will, when expanding records to tuples, introduce
new variables in the abstract representation. As the expansion is done
on the abstract representation, the compiler can safely name the new
variables with names that are not syntactically valid in Erlang source
code (The name starts with a lowercase letter), thus ensuring the
uniqueness of the new names.

The above strategy leads to problems if a user uses the compiler to
produce the abstract representation (using options `E` and `binary`),
transforms it and intends to use the functions in erl_pp to produce
source code, as the generated variable names will be printed as atoms.

This patch introduces an utility function, `erl_pp:legalize_vars/1`,
which when given the abstract representation of a function, will
return an equivalent function where all variables will have
syntactically valid names. As outputting Erlang source code from the
abstract representation is a niche application, the solution with an
utility function, instead of modifying compiler passes, was chosen to
not slow down the common use case.

### DIFF
--- a/lib/stdlib/doc/src/erl_pp.xml
+++ b/lib/stdlib/doc/src/erl_pp.xml
@@ -45,6 +45,14 @@
 
     <p>All functions can have an optional argument, which specifies a hook
       that is called if an attempt is made to print an unknown form.</p>
+
+    <p>Note that if the functions in this module are used to convert
+      abstract code back to Erlang source code, the enclosing function
+      should first be processed by <seemfa
+      marker="#legalize_vars/1"><c>legalize_vars/1</c></seemfa> in order
+      to ensure that the output is semantically equivalent to the
+      abstract code.</p>
+
   </description>
 
   <datatypes>
@@ -144,6 +152,29 @@
       <desc>
         <p>Same as <seemfa marker="#form/1"><c>form/1,2</c></seemfa>,
           but only for the guard test <c><anno>Guard</anno></c>.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="legalize_vars" arity="1" since=""/>
+      <fsummary>Ensure all variable names are valid.</fsummary>
+      <desc>
+        <p>The Erlang compiler will, when expanding records to tuples,
+          introduce new variables in the abstract representation. As the
+          expansion is done on the abstract representation, the compiler
+          can safely name the new variables with names that are not
+          syntactically valid in Erlang source code (the name starts
+          with a lowercase letter), thus ensuring the uniqueness of the
+          new names.</p>
+
+        <p>The above strategy leads to problems if a user wants to
+          convert the abstract representation, using the functions of
+          this module back to Erlang source code. Typically, pattern
+          variables are output as atoms thus changing the sematics of
+          the program. To solve this problem <c>legalize_vars/1</c>,
+          when run on the abstract representation of a function, will
+          return an equivalent function where all variables will have
+          syntactically valid names.</p>
       </desc>
     </func>
   </funcs>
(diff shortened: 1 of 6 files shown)


lib/stdlib/doc/src/erl_pp.xml
```diff
@@ -146,6 +152,28 @@
...
+
+    <func>
+      <name name="legalize_vars" arity="1" since=""/>
+      <fsummary>Ensure all variable names are valid.</fsummary>
+      <desc>
+        <p>The Erlang compiler will, when expanding records to tuples,
+          introduce new variables in the abstract representation. As the
+          expansion is done on the abstract representation, the compiler
+          can safely name the new variables with names that are not
+          syntactically valid in Erlang source code (The name starts
```
bjorng, line 164:
```suggestion
          syntactically valid in Erlang source code (the name starts
```

(You might want do the same correction in the commit message.)

lib/stdlib/src/erl_pp.erl
```diff
@@ -1352,3 +1381,36 @@ word('when', WT) -> element(16, WT);
...
+
+fold_vars(F, Acc0, Forms) when is_list(Forms) ->
+    lists:foldl(fun(Elem, Acc) -> fold_vars(F, Acc, Elem) end, Acc0, Forms);
+fold_vars(F, Acc0, V={var,_,_}) ->
+    F(V, Acc0);
+fold_vars(F, Acc0, Form) when is_tuple(Form) ->
+    lists:foldl(fun(Elem, Acc) -> fold_vars(F, Acc, Elem) end,
+                Acc0, tuple_to_list(Form));
+fold_vars(_, Acc, Form) when is_atom(Form); is_integer(Form) ->
+    Acc.
```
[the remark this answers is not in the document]
frej: I thought about that, I think we are safe in this application. The only case where something user supplied isn't wrapped in a `Rep(X)` (to use the notation in the abstract format documentation) I could find, is for attributes and those can't occur inside a function (`legalize_vars/1` is ony defined for functions).

lib/stdlib/doc/src/erl_pp.xml
```diff
@@ -45,6 +45,12 @@
 
     <p>All functions can have an optional argument, which specifies a hook
       that is called if an attempt is made to print an unknown form.</p>
+
+    <p>Note that if the functions in this module are used to produce
+      Erlang source code from forms produced by the Erlang compiler, the
```
bjorng, line 50:
```suggestion
    <p>Note that if the functions in this module are used to convert abstract
    code back to Erlang source code that can be compiled, the
```

(Other compilers than the Erlang compiler may produce variable names that are not syntactically correct.)

bjorng (commented):
I think that it would be nice if the `E` option would directly produce compilable code. What about adding an optional compiler pass that calls `legalize_vars/1` for all function when the `E` option is given?

lib/stdlib/doc/src/erl_pp.xml
frej: fixed.

lib/stdlib/doc/src/erl_pp.xml
frej: I don't like the qualification of "source code that can be compiled", my experience with Bugerl is that most output can be compiled, it just doesn't have the same semantics as the input. I have reformulated it to talk about output that is semantically equivalent to the abstract code.

frej:
> I think that it would be nice if the `E` option would directly produce compilable code. 

I have such I patch, although I haven't made a test case for it.

frej:
Corrected too strict guards in `fold_vars` and /`map_vars`.
Pushed an additional commit which runs `erl_pp:legalize_vars/1` when running `erl -E`.

bjorng:
Added to our daily builds. Will merge tomorrow if everything goes well.

bjorng:
Thanks for your pull request.

PR #5366, opened 2021-11-10
stdlib: Fix erl_error to print re compile errors

garazdawi:
When re:run and friends fail because the regexp is invalid it would print in the error message that the regexp was of
an invalid type. Now it will instead print the compilation error message gotten from running re:compile.

Before:

    > re:run("asd",")").
         in function  re:run/2
            called as re:run("asd",")")
            *** argument 2: neither an iodata term nor a compiled regular expression

After:

    > re:run("asd",")").
         in function  re:run/2
            called as re:run("asd",")")
            *** argument 2: could not parse regular expression
                            unmatched parentheses on character 0


bjorng (approved):
Looks good to me.

PR #5371, opened 2021-11-11
yecc: Avoid compilation errors if atoms become reserved words

bjorng:
In code generated by `yecc`, only atoms known to be reserved words
would be quoted. If a generated module is later compiled with a
version of the compiler that has made one of the used atoms a
reserved word, compilation would fail.

Avoid the problem by having `yecc` code quote all atoms in the
generated code.

### DIFF
--- a/lib/parsetools/src/yecc.erl
+++ b/lib/parsetools/src/yecc.erl
@@ -2130,7 +2130,7 @@ output_goto_fini(F, NT, #yecc{includefile_version = {1,1}}=St0) ->
                 [F]),
     fwrite(St, 
            ?YECC_BUG(<<"{~ts, State, missing_in_goto_table}">>,
-                     [quoted_atom(St0, NT)]),
+                     [quoted_atom(NT)]),
            []);
 output_goto_fini(_F, _NT, St) ->
     fwrite(St, <<".\n\n">>, []).
@@ -2255,13 +2255,13 @@ output_action(St, State, Terminal, #reduce{}=Action, IsFirst, SI) ->
 output_action(St0, State, Terminal, #shift{state = NewState}, IsFirst, _SI) ->
     St10 = delim(St0, IsFirst),
     St = fwrite(St10, <<"yeccpars2_~w(S, ~ts, Ss, Stack, T, Ts, Tzr) ->\n">>,
-                [State, quoted_atom(St10, Terminal)]),
+                [State, quoted_atom(Terminal)]),
     output_call_to_includefile(NewState, St);
 output_action(St0, State, Terminal, accept, IsFirst, _SI) ->
     St10 = delim(St0, IsFirst),
     St = fwrite(St10, 
                 <<"yeccpars2_~w(_S, ~ts, _Ss, Stack, _T, _Ts, _Tzr) ->\n">>,
-                [State, quoted_atom(St10, Terminal)]),
+                [State, quoted_atom(Terminal)]),
     fwrite(St, <<" {ok, hd(Stack)}">>, []);
 output_action(St, _State, _Terminal, nonassoc, _IsFirst, _SI) ->
     St.
@@ -2287,7 +2287,7 @@ output_reduce(St0, State, Terminal,
               IsFirst, StateInfo) ->
     St10 = delim(St0, IsFirst),
     QuotedTerminal = if 
-                         is_atom(Terminal) -> quoted_atom(St10, Terminal);
+                         is_atom(Terminal) -> quoted_atom(Terminal);
                          true -> Terminal
                      end,
     St20 = fwrite(St10,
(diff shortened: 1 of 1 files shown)


lisztspace:
Reasonable change since the generated code will be precise and consistent rather than trying to be terse when possible.  Making it future safe is a good property as it will not force recompiles just because a used atoms become reserved.

Testing this on a small grammar described pleasant S-expressions.

```
Nonterminals list elements element.
Terminals atom '(' ')'.
Rootsymbol list.
list -> '(' ')' : nil.
list -> '(' elements ')' : '$2'.
elements -> element : {cons, '$1', nil}.
elements -> element elements : {cons, '$1', '$2'}.
element -> atom : '$1'.
element -> list : '$1'.
```

I found, e.g., a function previously generated as `yeccgoto_element` now being generated as `yeccgoto_101108101109101110116`.  While technically not a problem (and at least "readable") that was perhaps not the intent of the change.  This is consistent for all functions related to non terminals.

```
-dialyzer({nowarn_function, yeccgoto_101108101109101110116/7}).
-compile({nowarn_unused_function,  yeccgoto_101108101109101110116/7}).
yeccgoto_101108101109101110116(2, Cat, Ss, Stack, T, Ts, Tzr) ->
 yeccpars2_5(5, Cat, Ss, Stack, T, Ts, Tzr);
yeccgoto_101108101109101110116(5, Cat, Ss, Stack, T, Ts, Tzr) ->
 yeccpars2_5(5, Cat, Ss, Stack, T, Ts, Tzr).
```

I also expected my included atoms, i.e., `cons` and `nil` to be quoted, but I see

```
-compile({inline,yeccpars2_5_/1}).
-dialyzer({nowarn_function, yeccpars2_5_/1}).
-compile({nowarn_unused_function,  yeccpars2_5_/1}).
-file("sexpr.yrl", 2).
yeccpars2_5_(__Stack0) ->
 [___1 | __Stack] = __Stack0,
 [begin
                      {cons, ___1, nil}
  end | __Stack].
```

There are still a large number of unquoted atoms in the generated file and any inlined Erlang code (after `Erlang code.`) seems to be untouched as well.  Have I misunderstood the intent?  Also, to be really paranoid about this, shouldn't even module names and function names be quoted?  If they are not, having a function named `maybe` will cause problems when EEP49 is introduced, although this problem can be circumvented by compiling the file with `--disable-feature=maybe` (thereby showing that it makes sense to have that ability).

lisztspace:
> There are still a large number of unquoted atoms in the generated file and any inlined Erlang code (after `Erlang code.`) seems to be untouched as well.

The inlined code is of course difficult to do anything about since it is just copied verbatim (Syntax errors and all) to the output.  Fixing that can be done, but is a major change in the workings of `yecc` since it would require reading the verbatim code printing it in _quoting atoms mode_.

bjorng:
> I found, e.g., a function previously generated as `yeccgoto_element` now being generated as `yeccgoto_101108101109101110116`.

That was unintentional. Fixed now.

bjorng:
> > There are still a large number of unquoted atoms in the generated file and any inlined Erlang code (after `Erlang code.`) seems to be untouched as well.
> 
> The inlined code is of course difficult to do anything about since it is just copied verbatim (Syntax errors and all) to the output. Fixing that can be done, but is a major change in the workings of `yecc` since it would require reading the verbatim code printing it in _quoting atoms mode_.

Yes, I don't want to change the reading of the code after `Erlang code.`. That could cause unforeseen problems.

However, it is easy to quote atoms in embedded actions, but I don't think that would be appropriate. To show why, consider this example:

```
maybe_expr -> 'maybe' maybe_match_exprs 'end' :
	{maybe,?anno('$1'),'$2'}.
```

If `yecc` were to quote atoms in the generated Erlang code, everything would work **now**. But in some release in the future, `maybe` will probably become a reserved word, and when that happens, `yecc` will read the unquoted `maybe` as a keyword and output it as an unquoted keyword. That will result in a syntax error when trying to compile the generated module. Therefore, it is better to force the implementer of an experimental new feature to quote all atoms that are likely to become reserved words:

```
maybe_expr -> 'maybe' maybe_match_exprs 'end' :
	{'maybe',?anno('$1'),'$2'}.
```

bjorng:
To clarify the intent of this PR: In the code generated by `yecc`, all terminals in the grammar (tokens such `maybe` or `else`) would be used as atoms. Those atoms would be quoted when they were symbols (such as `?=`) or previously known reserved words (such as `case`), but would otherwise be unquoted. This PR quotes all atoms used in that way.

lisztspace:
> To clarify the intent of this PR: In the code generated by `yecc`, all terminals in the grammar (tokens such `maybe` or `else`) would be used as atoms. Those atoms would be quoted when they were symbols (such as `?=`) or previously known reserved words (such as `case`), but would otherwise be unquoted. This PR quotes all atoms used in that way.

Good with this clarification, which then makes the scope appropriate.

lisztspace (approved):
Approved after fixing the mangling of names and the scope clarification.

PR #5372, opened 2021-11-11
erts: Fix integer overflow problem in WSTACK

garazdawi:
In the testcase binary_SUITE:t2b_deterministic a very large hash map is created and encoded. This lead to a WSTACK_RESERVE call with an N value of 4 000 000 which made the pointer wrap around zero and thus the stack was not grown as it should.

As far as I can tell this can only happen in the new deterministic term_to_binary.

### DIFF
--- a/erts/emulator/beam/global.h
+++ b/erts/emulator/beam/global.h
@@ -642,28 +642,29 @@ do {						\
 
 #define WSTACK_IS_STATIC(s) (s.wstart == WSTK_DEF_STACK(s))
 
-#define WSTACK_PUSH(s, x)				\
-do {							\
-    if (s.wsp == s.wend) {				\
-	erl_grow_wstack(&s, 1); 	                \
-    }							\
-    *s.wsp++ = (x);					\
+#define WSTACK_RESERVE(s, push_cnt)             \
+do {						\
+    if (s.wend - s.wsp < (Sint)(push_cnt)) {    \
+	erl_grow_wstack(&s, (push_cnt));        \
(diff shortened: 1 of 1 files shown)


erts/emulator/beam/global.h
```diff
@@ -642,6 +642,14 @@ do {						\
 
 #define WSTACK_IS_STATIC(s) (s.wstart == WSTK_DEF_STACK(s))
 
+#define WSTACK_RESERVE(s, push_cnt)             \
+do {						\
+    UWord *wend = s.wend - (push_cnt);          \
+    if (s.wsp > wend || wend > s.wend) {        \
+	erl_grow_wstack(&s, (push_cnt));        \
```
sverker, line 649: Is it not even better with
`if (s.wend - s.wsp < (SWord)push_cnt)
`
Pointer arithmetic outside allocated areas is undefined.

Also, same bug in ESTACK_RESERVE.
We should probably merge the two some time. WSTACK is a residue from halfword emulator.

bjorng: I agree with @sverker, but sholdn't it be:

```
if (s.wend - s.wsp < (UWord)push_cnt)
```

because the difference between two pointers is unsigned?

sverker: Pointer subtraction is signed.
If wsp is larger than wend (which I hope it never can be) then a signed comparison could save the day and trigger a stack growth.

sverker (approved):

PR #5375, opened 2021-11-12
common_test: add anchor links to log

attah:
This adds "anchor links" to the common_test log output. This enables linking to a particular log entry.
A-tags with an ever-incremented id, and href to that same #id are added after 'hd' parts of logged data.
Screenshot from 2021-11-12 13-01-20

As suggested/inspired by: @tomas-abrahamsson

### DIFF
--- a/lib/common_test/priv/ct_default.css
+++ b/lib/common_test/priv/ct_default.css
@@ -207,3 +207,8 @@ th + td {
     color: #fff;
     background-color: #809FFF;
 }
+
+a.link-to-entry {
+    float: right;
+    text-decoration: none;
+}
--- a/lib/common_test/src/ct_logs.erl
+++ b/lib/common_test/src/ct_logs.erl
@@ -655,7 +655,8 @@ log_timestamp({MS,S,US}) ->
 		      tc_groupleaders,
 		      stylesheet,
 		      async_print_jobs,
-		      tc_esc_chars}).
+		      tc_esc_chars,
+		      log_index}).
 
 logger(Parent, Mode, Verbosity) ->
     register(?MODULE,self()),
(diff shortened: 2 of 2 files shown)


u3s:
nice idea. might be handy for longer logs.

u3s:
would it be possible to adjust the proposal a little so it is similar to what we have in documentation?

for example [erlang page](https://www.erlang.org/doc/man/erlang.html#type-any) displays similar anchor icon - however it appears on mouse over event and is aligned to left.

it would be great if we could have similar look and feel.

attah:
I will try and see what happens.
The reasoning for right-aligning was to not compete for space or affect alignment.
I don't think we have margins on the left to put it in, like in the documentation... i guess i could add some, but that will probably look strange.

u3s:
Thanks, if mimicking exactly same layout becomes problematic please share arguments - maybe some other hybrid approach could be taken (e.g auto-hiding icon but right aligned to header - to by pass lack of margin).

attah:
Okay, some quick observations:
left-aligned inside the div will not be possible, as the opening tags and header it is constructed out-of-process (and the counter naturally needs to be in process).
Left of the div should be doable, but the {int,div}_{header,footer} functions are abusing what they do and wrapping everything in-between in `<pre></pre>`. I.e. header starts with `</pre>` and footer ends with `<pre>` - so i can't also inject code here.
I guess i could try to move that hack into the process, but honestly i'm apprehensive about restructuring code that looks this fragile.
Having seen it to the left now, i'm not convinced that sacrificing prime real-estate for this is a good idea.
Noting also that https://www.erlang.org/eeps/eep-0054.html#extending-the-stacktrace has the more common layout, right after the headers (and permanently visible- perhaps a bit too visible).

u3s:
Thanks for considering alternatives! I agree we should merge the original proposal.

PR #5384, opened 2021-11-13
[erts] fix crash dump caused by killing logger

max-au:
Init process is linked to logger. Killing logger while still being
linked to it causes `{'EXIT',<0.42.0>,killed}` message back to init
which has already entered erts_halt, and yielded this way:

    ERTS_BIF_YIELD2(BIF_TRAP_EXPORT(BIF_halt_2), BIF_P, am_undefined, am_undefined);

This yields just as if process called `erlang:halt(undefined, undefined)`, which
is invalid, and halt throws `badarg`, causing ERTS to terminate with crash dump
and a Slogan "System process <0.0.0> terminated: {badarg,[{erlang,halt,[undefined,undefined],[.."

Related refactoring made the code slightly more readable.

### DIFF
--- a/erts/preloaded/src/init.erl
+++ b/erts/preloaded/src/init.erl
@@ -641,8 +641,9 @@ clear_system(Unload,BootPid,State) ->
     Logger = get_logger(State#state.kernel),
     shutdown_pids(Heart,Logger,BootPid,State),
     Unload andalso unload(Heart),
-    kill_em([Logger]),
-    do_unload([logger_server]).
+    unlink(Logger),
+    exit(Logger, kill),
+    Unload andalso do_unload([logger_server]).
 
 flush() ->
     receive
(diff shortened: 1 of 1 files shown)


garazdawi:
hmm, after the process has yielded to `erlang:halt(undefined, undefined)` it is not supposed to run again as the run-queue should enter the halting state and never schedule any processes again. So I don't think the problem is the linking, but rather that the init process is scheduled when it should not be.

max-au:
This might be some old standing race. I suspect that #5032 changed halt sequence. Probably before merging it some processes that were running in the modules being unloaded were brutally killed, but with #5032 processes keep running.
At least I think this behaviour changed after we merged that PR (and a bunch of others too).

max-au:
I no longer observe these crashes (although I think the problem is still there).

PR #5389, opened 2021-11-15
sasl: Be even more resilient of exiting processes

### LINKED ISSUE #5387: Failing app upgrade in release_hander_1 because of process shutdown
**Describe the bug**
In https://github.com/erlang/otp/issues/3696 there were two places where release handler failed due to a process shutdown. The issue is fixed in https://github.com/erlang/otp/pull/2666

We encountered another place where the same improvement is needed. During release upgrade:
```
release_handler: {'EXIT',{{shutdown,{socket_error,closed,'The socket has been closed.'}},{gen_server,call,[<0.27413.4942>,which_children,infinity]}}}
error during a which_children call to cowboy_tls (<0.27413.4942>). [State: running] Exiting ... 
```

This points to : https://github.com/erlang/otp/blob/master/lib/sasl/src/release_handler_1.erl#L643 where in case the Result is not a list an error is reported and the release upgrade fails.

**Expected behavior**
Release upgrade does not fail when a process shuts down during an upgrade as that process shall not be treated furthermore.

**Affected versions**
All

sverker:
Fixes #5387.

Same kind of bug fix in sasl release_handler as #2666.

@zsoci Would be good if you could give this patch a thorough look.
Contributions to test these kinds of fail scenarios are also welcome.

### DIFF
--- a/lib/sasl/src/release_handler_1.erl
+++ b/lib/sasl/src/release_handler_1.erl
@@ -637,14 +637,19 @@ maybe_supervisor_which_children(Proc, Name, Pid) ->
             error(suspended_supervisor);
 
         running ->
-            case catch supervisor:which_children(Pid) of
+            try supervisor:which_children(Pid) of
                 Res when is_list(Res) ->
-                    Res;
-                Other ->
+                    Res
+            catch
+                exit:Reason when Reason =/= timeout andalso
+                                 not (is_tuple(Reason) andalso
+                                      element(1,Reason) =:= nodedown) ->
+                    [];
+                exit:Other ->
                     error_logger:error_msg("release_handler: ~p~nerror during"
                                            " a which_children call to ~p (~w)."
                                            " [State: running] Exiting ... ~n",
-                                           [Other, Name, Pid]),
+                                           [{'EXIT',Other}, Name, Pid]),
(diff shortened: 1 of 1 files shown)


zsoci:
Please give me a couple of days to check it through. I will be able to take a look at it this week. @sverker

lib/sasl/src/release_handler_1.erl
```diff
@@ -637,14 +637,19 @@ maybe_supervisor_which_children(Proc, Name, Pid) ->
...
+            try supervisor:which_children(Pid) of
                 Res when is_list(Res) ->
-                    Res;
-                Other ->
+                    Res
+            catch
+                exit:Reason when Reason =/= timeout andalso
+                                 not (is_tuple(Reason) andalso
+                                      element(1,Reason) =:= nodedown) ->
+                    noproc;
```
zsoci, line 647: Should not this be [] instead of noproc?

zsoci:
Seems to be OK! Thanks.

lib/sasl/src/release_handler_1.erl
sverker: Agree, [] is more consistent. Fixed.

Issue #5311, opened 2021-10-23
README.md – remove mention of "./otp_build autoconf"


marcellanz:
**Describe the bug**
README.md mentions to run `./otp_build autoconf` to compile from sources but is superfluous as per otp_build.

**To Reproduce**
Read the [README.md](https://github.com/erlang/otp#compiling-from-source) and run:
```
> ./otp_build autoconf
```

**Expected behavior**
I would expect that autoconf is run, but a message is printed explaining to run `./otp_build autoconf` is superfluous now.

**Affected versions**
OTP-24.1.2

PR #5390, opened 2021-11-15
Remove superfluous "otp_build autoconf" and improve README build instructions

sverker:
One commit from @marcellanz (#5311) removing `otp_build autoconf` from various scripts and docs (except README.md). Also thanks to @v0idpwn and @kpy3 for earlier stabs at `otp_build autoconf` in README.md.

Added some comments and a "git checkout" step to the build instructions in README.md,
which can be viewed here:
https://github.com/sverker/otp/tree/sverker/otp_build-autoconf-cleanup

### DIFF
--- a/.github/scripts/init-pre-release.sh
+++ b/.github/scripts/init-pre-release.sh
@@ -1,26 +1,6 @@
 #!/bin/sh
 
-## We create a tar ball that has generated configure
-## (old versions might not have these included)
-## This is used later by build-otp-tar to create
-## the pre-built tar ball
+## We create a tar ball that is used later by build-otp-tar
+## to create the pre-built tar ball
 
-if [ -f ./configure ]; then
-    git archive --prefix otp/ -o otp_src.tar.gz HEAD
-else
-    ERL_TOP=`pwd`
-    ./otp_build autoconf
-    find . -name aclocal.m4 | xargs git add -f
-    find . -name configure | xargs git add -f
-    find . -name config.h.in | xargs git add -f
-    find . -name config.guess | xargs git add -f
-    find . -name config.sub | xargs git add -f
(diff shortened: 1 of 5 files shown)


lib/megaco/Makefile
```diff
@@ -115,7 +115,6 @@ include $(ERL_TOP)/make/otp_subdir.mk
 
 reconf:
 	(cd $(ERL_TOP) && \
-		./otp_build autoconf && \
 		./otp_build configure && \
 		cd $(ERL_TOP)/../libraries/megaco)
```
rickard-green, line 119: @bmk this last `cd` (which in it self is a bit strange since it is the last command in a subshell) seems to be a leftover since the clearcase days when megaco was in a different VOB than erts.

.github/scripts/init-pre-release.sh
```diff
@@ -9,7 +9,6 @@ if [ -f ./configure ]; then
     git archive --prefix otp/ -o otp_src.tar.gz HEAD
```
rickard-green, line 9: I think it was I who added this `if`-statement when I committed the `configure` scripts into the repo, but I'm not sure why. This line should be the only one needed (apart the #!/bin/sh line) in this file since `./configure` always should be present. Either I'm missing something now or I missed something then...

rickard-green (approved):

bjorng (approved):

garazdawi (approved):

marcellanz:
Love to see this being merged :)
Thanks @sverker and also thanks @v0idpwn and @kpy3 for your initial work.

.github/scripts/init-pre-release.sh
sverker: Fixed. Removed `if `and entire `else` block.

PR #5036, opened 2021-07-08
gh-actions: Add macOS builds

wojtekmach:
Another stab at https://github.com/erlang/otp/issues/4487.

We statically link OpenSSL:

    % otool -L otp/otp/lib/crypto-5.0.2/priv/lib/crypto.so
    otp/otp/lib/crypto-5.0.2/priv/lib/crypto.so:
            /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)

See https://rentzsch.tumblr.com/post/33696323211/wherein-i-write-apples-technote-about-openssl-on for rationale.

We also statically link wxWidgets:

    % otool -L otp/otp/lib/wx-2.0.1/priv/wxe_driver.so
    otp/otp/lib/wx-2.0.1/priv/wxe_driver.so:
            /System/Library/Frameworks/IOKit.framework/Versions/A/IOKit (compatibility version 1.0.0, current version 275.0.0)
            /System/Library/Frameworks/Carbon.framework/Versions/A/Carbon (compatibility version 2.0.0, current version 164.0.0)
            /System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 23.0.0)
            /System/Library/Frameworks/QuartzCore.framework/Versions/A/QuartzCore (compatibility version 1.2.0, current version 1.11.0)
            /System/Library/Frameworks/AudioToolbox.framework/Versions/A/AudioToolbox (compatibility version 1.0.0, current version 1000.0.0)
            /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.100.5)
            /System/Library/Frameworks/OpenGL.framework/Versions/A/OpenGL (compatibility version 1.0.0, current version 1.0.0)
            /System/Library/Frameworks/AGL.framework/Versions/A/AGL (compatibility version 1.0.0, current version 1.0.0)
            /System/Library/Frameworks/WebKit.framework/Versions/A/WebKit (compatibility version 1.0.0, current version 611.1.21)
            /usr/lib/libexpat.1.dylib (compatibility version 7.0.0, current version 8.0.0)
            /usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
            /System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 59754.100.106)
            /usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
[...]

PR #5393, opened 2021-11-16
erts: Remove `CFBundle*` fields from `Info.plist`

wojtekmach:
I'm building an `.app` bundle with it's own `Info.plist`. In my plist,
the values I'm putting in seem to be ignored, when running the app,
the app name in the menu bar is always "Erlang".

With this patch, the `CFBundle*` from my app's plist take effect.

One should be able to re-produce this issue by executing this shell
script:

```sh
#!/bin/bash
set -e

rm -rf Example.app
mkdir -p Example.app/Contents/MacOS

cat <<EOF > Example.app/Contents/Info.plist
<plist version="1.0">
<dict>
  <key>CFBundleName</key>
  <string>Example</string>
  <key>CFBundleDisplayName</key>
  <string>Example</string>
</dict>
</plist>
EOF

cat <<EOF > Example.app/Contents/MacOS/Example
#!/bin/bash
erl -noshell -eval '
Wx=wx:new(), F=wxFrame:new(Wx,-1,"Example",[]), wxFrame:show(F),

%% listen for menubar events, in particular cmd+q so we can quit.
MenuBar = wxMenuBar:new(), wxFrame:setMenuBar(F,MenuBar),
wxFrame:connect(F,command_menu_selected),
receive Event ->
    io:format("~p~n", [Event])
end,
halt().
' > ~/Example.stdout 2> ~/Example.stderr
EOF

chmod +x Example.app/Contents/MacOS/Example
open Example.app
```

### DIFF
--- a/erts/etc/darwin/Info.plist
+++ b/erts/etc/darwin/Info.plist
@@ -2,14 +2,6 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>CFBundleName</key>
-	<string>Erlang</string>
-	<key>CFBundleDisplayname</key>
-	<string>Erlang</string>
-	<key>CFBundleDevelopmentRegion</key>
-	<string>English</string>
-	<key>CFBundleIdentifier</key>
-	<string>org.erlang.beam</string>
 	<key>NSAppTransportSecurity</key>
 	<dict>
 		<key>NSExceptionDomains</key>


wojtekmach:
@dominicletz feedback appreciated!

leoliu:
I experimented with wx 2 days ago and overall I really enjoyed it.

What frustrated me the most is how to distribute it i.e. package all deps (wxwidgests) in one app bundle. I also encountered this issue while manual making an app bundle. A partial solution is to set the `WX_APP_TITLE` envvar unfortunately the menubar title is still "Erlang" though every in it is changed to the value of `WX_APP_TITLE`.

The other issue to how to teach wx to load dylib from within the app bundle. I used `install_name_tool` to change the dylib paths but no effect at all. `wxe_driver.so` insists loading from the path it was compiled with.

wojtekmach:
@leoliu you want to statically link wx, you need wx to be built with `--disable-shared` flag. See https://github.com/erlang/otp/pull/5036/files#diff-b71abeb0efb0f9fb4825e58d03055c35878b327ccba0b5e3110199caf19ea662R17.

dominicletz:
> @leoliu you want to statically link wx, you need wx to be built with `--disable-shared` flag. See https://github.com/erlang/otp/pull/5036/files#diff-b71abeb0efb0f9fb4825e58d03055c35878b327ccba0b5e3110199caf19ea662R17.

Another option @leoliu is to provide a DYLD_LIBRARY_PATH e.g. from your env.sh and whitelist the environment variable:

in env.sh.eex
```
export DYLD_LIBRARY_PATH="$RELEASE_ROOT/lib/<%= @release.name %>-<%= @release.version %>/priv/"
```

and in yourapp.entitlements
```
<plist version="1.0">
<dict>
	<key>com.apple.security.cs.allow-jit</key>
        <true/>
        <key>com.apple.security.cs.allow-dyld-environment-variables</key>
        <true/>
</dict>
</plist>
```

and then ensure that all needed libraries are in that /priv/ dir when shipping. This is how we at Diode.io are notarizing and shipping our mac app at the moment. Same approach is used by wings3d mac app https://github.com/dgud/wings/tree/master/macosx

dominicletz:
> I'm building an `.app` bundle with it's own `Info.plist`. In my plist, the values I'm putting in seem to be ignored, when running the app, the app name in the menu bar is always "Erlang".
> 
> With this patch, the `CFBundle*` from my app's plist take effect.
> 
> One should be able to re-produce this issue by executing this shell script:
> 
> ```shell
> set -e
> 
> rm -rf Example.app
> mkdir -p Example.app/Contents/MacOS
> 
> cat <<EOF > Example.app/Contents/Info.plist
> <?xml version="1.0" encoding="UTF-8"?>
> <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
> <plist version="1.0">
> <dict>
>   <key>CFBundleName</key>
>   <string>Example</string>
>   <key>CFBundleDisplayName</key>
>   <string>Example</string>
> </dict>
> </plist>
> EOF
> 
> cat <<EOF > Example.app/Contents/MacOS/Example
> erl -noshell -eval '
> Wx=wx:new(), F=wxFrame:new(Wx,-1,"Example",[]), wxFrame:show(F),
> 
> %% listen for menubar events, in particular cmd+q so we can quit.
> MenuBar = wxMenuBar:new(), wxFrame:setMenuBar(F,MenuBar),
> wxFrame:connect(F,command_menu_selected),
> receive Event ->
>     io:format("~p~n", [Event])
> end,
> halt().
> ' > ~/Example.stdout 2> ~/Example.stderr
> EOF
> 
> chmod +x Example.app/Contents/MacOS/Example
> open Example.app
> ```

Agreed, the commit looks good to me. I did not find any other way to suppress the embedded `CFDisplayBundleName` either. All other places seem to obey `WX_APP_TITLE` or use the process name (which is set through `WX_APP_TITLE`) there is some documentation I could find here: http://mirror.macintosharchive.org/developer.apple.com/documentation/Cocoa/Reference/ApplicationKit/ObjC_classic/AppKit.pdf on how the ApplicationName is determined (and that seems to be used here)
Selection_760
 
It's pretty strange though that the `Info.plist` value embedded in `beam.smp` is overriding the `Info.plist` value you're providing in the apps `Info.plist`. Another option I guess is to strip the embedded `Info.plist` completely from the beam.smp binary before bundling in a new app.


P.S.
I had to add `#!/bin/bash` to the front of the Example script to make your example work for me.

wojtekmach:
@dominicletz thank you for the extra context and testing it out! (Thanks for the `#!/bin/bash` bit too, it got stripped when I committed it, re-added it back)

If beam.smp Info.plist always takes precedence over app's Info.plist and the only way to override it is to strip it out then maybe there should be no beam.smp Info.plist at all? Or not having it would be a worse experience, i.e. stuff wouldn't work out of the box?

dominicletz:
I've added `Info.plist` it to make the webview example in `wx:demo().` work with `http://localhost:xxx` addresses. Without the localhost exception for it won't work out of the box.

wojtekmach:
Oh, that makes sense. Then perhaps stripping Info.plist to the absolute minimum is a better idea.

leoliu:
Another argument for completely striping Info.plist from beam is if you get info from Activity Monitor you can see `Library/Saved Application State/org.erlang.beam.savedState`, it looks like all apps are using the bundle ID `org.erlang.beam`.

leoliu:
> Another option @leoliu is to provide a DYLD_LIBRARY_PATH e.g. from your env.sh and whitelist the environment variable:
> 
> in env.sh.eex
> 
> ```
> export DYLD_LIBRARY_PATH="$RELEASE_ROOT/lib/<%= @release.name %>-<%= @release.version %>/priv/"
> ```
> 
> and in yourapp.entitlements....

@dominicletz amazingly I got it to work ;) Many thanks for the hints.

wojtekmach:
If this is accepted, I think it would be nice if it lands in OTP 24.x so please let me know if I should change the target to a maint branch.

bjorng:
> If this is accepted, I think it would be nice if it lands in OTP 24.x so please let me know if I should change the target to a maint branch.

Yes, please target the maint branch.

leoliu:
> Another argument for completely striping Info.plist from beam is if you get info from Activity Monitor you can see `Library/Saved Application State/org.erlang.beam.savedState`, it looks like all apps are using the bundle ID `org.erlang.beam`.

My comment here suggested that `CFBundleIdentifier` is also overridden to be `org.erlang.beam` which means stripping Info.plist is the only way forward. WDYT?

bjorng:
> My comment here suggested that `CFBundleIdentifier` is also overridden to be `org.erlang.beam` which means stripping Info.plist is the only way forward.

Do you mean like this? Will that work?

```
<plist version="1.0">
<dict>
	<dict>
		<key>NSExceptionDomains</key>
		<dict>
			<key>localhost</key>
			<dict>
				<key>NSExceptionAllowsInsecureHTTPLoads</key>
				<true/>
				<key>NSIncludesSubdomains</key>
				<true/>
			</dict>
		</dict>
	</dict>
</dict>
</plist>
```

leoliu:
> Do you mean like this? Will that work?
> 
> ```
> <?xml version="1.0" encoding="UTF-8"?>
> <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
> <plist version="1.0">
> <dict>
> 	<dict>
> 		<key>NSExceptionDomains</key>
> 		<dict>
> 			<key>localhost</key>
> 			<dict>
> 				<key>NSExceptionAllowsInsecureHTTPLoads</key>
> 				<true/>
> 				<key>NSIncludesSubdomains</key>
> 				<true/>
> 			</dict>
> 		</dict>
> 	</dict>
> </dict>
> </plist>
> ```

Not exactly. I mean with Info.plist like this

```
    ...
    <key>CFBundleIdentifier</key>
    <string>com.example.testapp</string>
    ...
```

and after open the App with the `CFBundleIdentifier` above. I have this:

```
$ defaults read org.erlang.beam
{
    NSTreatUnknownArgumentsAsOpen = NO;
}

defaults read com.example.testapp
2021-11-22 23:49:17.155 defaults[23340:2132773] 
Domain com.example.testapp does not exist
```

It seems to suggest that every App is using `org.erlang.beam` as `CFBundleIdentifier` regardless of the setting in the user's Info.plist file.

wojtekmach:
@leoliu I believe @bjorng's suggestion alleviates your concern and I just committed it. Please let me know otherwise!

bjorng:
Thanks! I have tested Wings3D with this pull request and it works fine.

leoliu:
It looks good :) Thanks to all for getting this merged.

PR #5398, opened 2021-11-16
vxworks: remove remaining traces

max-au:
Erlang/OTP no longer support VxWorks. Cleanup mentions to avoid confusion.


erts/doc/src/inet_cfg.xml
```diff
@@ -79,7 +79,7 @@
       <c><![CDATA['file']]></c> or <c><![CDATA['dns']]></c>, see below.</p>
 
     <p>Native lookup (system calls) is always the default resolver method.
-      This is true for all platforms, except VxWorks and OSE Delta where
+      This is true for all platforms, except OSE Delta where
```
garazdawi, line 82: You can remove this entire sentence since OSE Delta is also not supported.

max-au:
Rebased to master branch.

PR #5400, opened 2021-11-17
[ssl] fix SSL tests run in specific environment

max-au:
Before this patch, run_where and node_to_hostip were returning
different IP addresses, failing large number of tests when /etc/hosts
file was containing
    host 127.0.0.1
    host.domain 127.0.0.2

It was happening because `node()` was `test_server@host`, while
`net_adm:localhost()` was returning `host.domain`, so they were
returning different IP addresses, subsequently failing in verification
in check_result/4.


IngelaAndin:
@max-au looking good, could you base it on maint instead so that we can get the test case improvement for the OTP-24 track ?

IngelaAndin:
Thank you for the PR

PR #5408, opened 2021-11-17
Allow some configuration to be read from XDG config

### LINKED ISSUE #5016: Support XDG or platform-specific paths for `.erlang` and `.erlang.cookie` files

Due to unfortunate decision of early UNIX developers to use `path[0] == '.'` instead of `strcmp(path, ".") || strcmp(path, "..")` in `ls` when filtering out the default paths, people started to using "dotfiles" stored in the `$HOME` of the current user to provide configuration and runtime commands for various files. Erlang currently follows that unfortunate trend in form of `.erlang` and `.erlang.cookie` files. To solve that mess that is all the dotfiles in home directory and all of mixing of configuration, caches, logs, data, etc. in the same directories and/or more files in home directory. Erlang already uses platform-specific paths for shell history, but `.erlang` and `.erlang.cookie` are still where these were before.

Use either [XDG Base Directory](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) specification on all OSes or platform-specific approach for all files. This could be implemented in a way that "old" paths are fallbacks if the "new" files do not exists.

Do nothing.

garazdawi:
This PR changes so that the location of some system global files can be configured using XDG. The files are:

* $XDG_CONFIG_HOME/erlang
  * `.erlang`
  * `.erlang.cookie`
  * `.erlang.crypt`
  * `debugger`
* $XDG_CACHE_HOME/erlang
  * `.dialyzer_plt`

In order to keep backward compatibility, the config files are first read from $HOME, and then if they don't exist the $XDG_CONFIG_HOME/erlang directory is checked.

Since the dialyzer plt is a cache than a config file I decided to change the default location for it.

I decided to not do anything about `.erlang.hosts` as that file is hardly used by anyone.

Closes #5016

### DIFF
--- a/erts/doc/src/erl_cmd.xml
+++ b/erts/doc/src/erl_cmd.xml
@@ -461,10 +461,12 @@ $ <input>erl \
         <p>Makes the Erlang runtime system into a distributed node.
           This flag invokes all network servers necessary for a node to
           become distributed; see <seeerl marker="kernel:net_kernel">
-          <c>net_kernel(3)</c></seeerl>. It is also ensured that
+          <c>net_kernel(3)</c></seeerl>. It also ensures that
           <c><![CDATA[epmd]]></c> runs on the current host before Erlang is
-          started; see <seecom marker="epmd"><c>epmd(1)</c></seecom> and the
-          <seecom marker="#start_epmd"><c>-start_epmd</c></seecom> option.</p>
+          started (see <seecom marker="epmd"><c>epmd(1)</c></seecom> and the
+          <seecom marker="#start_epmd"><c>-start_epmd</c></seecom> option) and
+          that a magic cookie has been set
+          (see <seecom marker="#setcookie">-setcookie</seecom>).</p>
         <p>The node name will be <c><![CDATA[Name@Host]]></c>, where
           <c><![CDATA[Host]]></c> is the fully qualified host name of the
           current host. For short names, use flag <c><![CDATA[-sname]]></c>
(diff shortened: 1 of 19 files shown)


bjorng (approved):

JeromeDeBretagne:
> I think that in order to keep backward compatibility, it would need to be placed in ~/.erlang.cookie.That way jinterface, erl_interface and old erlang versions would still use the same default created cookie.

So you've used the above approach in the PR at the moment.

What about reading in both Home and XDGPath as introduced but creating new erlang.cookie files in XDGPath on supported platforms as soon as OTP 25, as a well documented transition?

The reading of an existing ~/.erlang.cookie file should bring backward compatibility for many (most?) cases, such as in-place upgrades of existing code, while new code / new installations based on OTP 25 would switch to the new XDG-based location (in one go for jinterface and erl_interface too).

Do you have other backward compatibility scenarios you want to cover, that would break with this direct transition? If yes, do you already have a later transition path in mind (in OTP 26) to move away from ~/.erlang.cookie? Thanks

garazdawi:
> in one go for jinterface and erl_interface too

The problem is this. Neither C nor Java have built-in implementations of XDG and vendoring in libraries with support for it is (IMO) more hassle than it is worth.

If anyone can think of a simple way to do this for jinterface and erl_interface I'm open to suggestions.

> If yes, do you already have a later transition path in mind (in OTP 26) to move away from ~/.erlang.cookie? Thanks

I don't know. What we do not want to do is break the experience of someone just starting with Erlang and setting it up to play with Erlang distribution. Normally you would just do it with a couple of Erlang nodes, but a surprising amount of people start by trying to connect it with their favorite Java/C application.

erts/doc/src/erl_cmd.xml
```diff
@@ -1870,14 +1875,15 @@ $ <input>erl \
       <tag>The <c>.erlang</c> startup file</tag>
       <item>
         <p>When Erlang/OTP is started, the system searches for a file named
-        <c>.erlang</c> in the user&apos;s home directory.</p>
+          <c>.erlang</c> in the user&apos;s home directory or
+          <seemfa marker="stdlib:filename#basedir/3"><c>$XDG_CONFIG_HOME/erlang</c></seemfa>.</p>
```
sverker, line 1879: What if `.erlang` files exist both places? Should the search order be documented?

system/doc/reference_manual/distributed.xml
```diff
@@ -181,8 +181,11 @@ dilbert@uab</pre>
       magic cookie and the cookie of other nodes is assumed to be
       <c>nocookie</c>. The first action of the Erlang network
       authentication server (<c>auth</c>) is then to read a file named
-      <c>$HOME/.erlang.cookie</c>. If the file does not exist, it is
-      created. The UNIX permissions mode of the file is set to octal
+      <c>$HOME/.erlang.cookie</c> or the
+      <seemfa marker="stdlib:filename#basedir/3">XDG path</seemfa>
+      <c>$XDG_CONFIG_HOME/erlang/.erlang.cookie</c>. If none of the
+      files exist, a <c>$HOME/.erlang.cooke</c> file is created.
+      The UNIX permissions mode of the file is set to octal
```
sverker, line 188: More clear about file search order here also?

sverker (approved):

sverker (approved):

PR #5411, opened 2021-11-17
Implement EEP 49 (Value-Based Error Handling Mechanisms) 

bjorng:
This is the complete implementation of EEP 49, but it is still a draft because the implementation will have to be revised when the enable_feature EEP has been published and implemented

### DIFF
--- a/erts/doc/src/absform.xml
+++ b/erts/doc/src/absform.xml
@@ -430,6 +430,21 @@
           <c>{match,ANNO,Rep(P),Rep(E_0)}</c>.</p>
       </item>
       <item>
+        <p>If E is a conditional match operator expression <c>P ?= E_0</c>,
+          where <c>P</c> is a pattern, then Rep(E) =
+          <c>{maybe_match,ANNO,Rep(P),Rep(E_0)}</c>.</p>
+      </item>
+      <item>
+        <p>If E is a maybe expression <c>maybe B end</c>,
+          where <c>B</c> is a body then Rep(E) =
+          <c>{'maybe',ANNO,Rep(B)}</c>.</p>
+      </item>
+      <item>
+        <p>If E is a maybe expression <c>maybe B else Ec_1 ; ... ; Ec_k end</c>,
+          where <c>B</c> is a body and each <c>Ec_i</c> is an else clause then Rep(E) =
+          <c>{'maybe',ANNO,Rep(B),{'else',ANNO,[Rep(Ec_1), ..., Rep(Tc_k)]}}</c>.</p>
+      </item>
+      <item>
         <p>If E is nil, <c>[]</c>, then Rep(E) = <c>{nil,ANNO}</c>.</p>
       </item>
       <item>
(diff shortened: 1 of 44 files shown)


valleyofblackpanther:
What do you do? Like I know you are doing something with Erlang but what specifically...?

bjorng:
> What do you do? Like I know you are doing something with Erlang but what specifically...?

The PR is an implementation of https://www.erlang.org/eeps/eep-0049

erts/doc/src/absform.xml
```diff
@@ -429,6 +429,21 @@
...
+          where <c>P</c> is a pattern, then Rep(E) =
+          <c>{maybe_match,ANNO,Rep(P),Rep(E_0)}</c>.</p>
+      </item>
+      <item>
+        <p>If E is a maybe expression <c>maybe B end</c>,
+          where <c>B</c> is a body then Rep(E) =
+          <c>{'maybe',ANNO,Rep(B)}</c>.</p>
+      </item>
+      <item>
+        <p>If E is a maybe expression <c>maybe B else Ec_1 ; ... ; Ec_k end end</c>,
```
gomoripeti, line 443: the double `end end` might be a typo

bjorng: Thanks! Corrected.

lib/tools/test/emacs_SUITE_data/icr
```diff
@@ -155,3 +155,42 @@ indent_receive() ->
...
+    begin
+        maybe_should_be_indented_as_begin,
+    end,
+    maybe
+        1 = foo(X),
+        2 ?= asd(X),
+        line_with_break =
+            foo(X+1),
+        line_with_break ?=
+             foo(X+1)
```
bjorng, line 169:
```suggestion
            foo(X+1)
```

bjorng:
The enable_feature EEP and implementation is not yet ready.

I have merged this PR so that it can be included in the first release candidate for OTP 25. Because enable_feature is not fully implemented, it is not possible to enable the `maybe` feature by putting an option inside the module that uses `maybe`. Instead, the option must be given as option to the compiler like this:

```
1> c(some_module, [{enable_feature,maybe_expr}]).
```

or by using `erlc`:

```
$ erlc '+{enable_feature,maybe_expr}' some_module.erl
```

attah:
Any chance of a minimal working example?
I'm on a newly-built OTP-25 and compiling with erlc '+{enable_feature,maybe_expr}'
No matter what snippet i take from the EEP or the SUITE, i get "syntax error before" whatever comes after maybe.

garazdawi:
```erlang
-module(t).
-feature(maybe_expr,enable).
-export([test/0]).
test() ->
    maybe
        ok ?= io:format("hello")
    end.
```

This works for me. According to the [docs](https://www.erlang.org/doc/reference_manual/features.html#enabling-and-disabling-features) you should use `erlc -enable-feature maybe_expr` to enable it using the command line. The command line option for enabling features was changes since the RC that @bjorng mentioned above.

PR #5412, opened 2021-11-17
Correct hostent_addr h_addrtype typespec

varnerac:
Fixes a broken spec for the `hostent` `h_addrtype` spec. It covers the results from [this](https://github.com/erlang/otp/blob/master/lib/kernel/src/inet_db.erl#L701) function head.

### DIFF
--- a/lib/kernel/include/inet.hrl
+++ b/lib/kernel/include/inet.hrl
@@ -22,9 +22,11 @@
 
 -record(hostent,
 	{
-	 h_name		  :: inet:hostname(),	%% official name of host
-	 h_aliases = []   :: [inet:hostname()],	%% alias list
-	 h_addrtype	  :: 'inet' | 'inet6',	%% host address type
-	 h_length	  :: non_neg_integer(),	%% length of address
-	 h_addr_list = [] :: [inet:ip_address()]%% list of addresses from name server
+	 h_name		  :: inet:hostname(),	 %% official name of host
+	 h_aliases = []   :: [inet:hostname()],	 %% alias list
+	 h_addrtype	  :: inet | inet6 | caa | cname | gid | hinfo | ns | mb | md | mg
+	               | mf | minfo | mx | naptr | null | ptr | soa | spf | srv | txt
+	               | uid | uinfo | unspec | uri | wks, %% host address type
+	 h_length	  :: non_neg_integer(),	 %% length of address
+	 h_addr_list = [] :: [inet:ip_address()] %% list of addresses from name server
 	}).


lib/kernel/include/inet.hrl
```diff
@@ -22,9 +22,11 @@
...
-	 h_name		  :: inet:hostname(),	%% official name of host
-	 h_aliases = []   :: [inet:hostname()],	%% alias list
-	 h_addrtype	  :: 'inet' | 'inet6',	%% host address type
-	 h_length	  :: non_neg_integer(),	%% length of address
-	 h_addr_list = [] :: [inet:ip_address()]%% list of addresses from name server
+	 h_name		  :: inet:hostname(),	 %% official name of host
+	 h_aliases = []   :: [inet:hostname()],	 %% alias list
+	 h_addrtype	  :: inet | inet6 | caa | cname | gid | hinfo | ns | mb | md | mg
+	               | mf | minfo | mx | naptr | null | ptr | soa | spf | srv | txt
+	               | uid | uinfo | unspec | uri | wks, %% host address type
```
RaimoNiskanen, line 29: I suggest this should be:
``` erlang
h_addrtype :: inet | inet6 | inet |_res:rr_type(),
```
and that `inet_res:rr_type()` should be exported.
This would be a bit incorrect since `a` and `aaaa` today cannot be in `#hostent.h_addrtype`, but future safer since it uses the type that the rest of `inet_res` uses for DNS record types, and future additions will be only to `inet_res:rr_type()`.

IngelaAndin: Would it perhaps make sense to have a `inet_res:dns_sockaddr_record` that could be exported and  make the type` h_addrtype :: inet | inet6 | inet_res:dns_sockaddr_record`  and make `inet_res:rr_type `be
`rr_type :: a | aaaa | dns_sockaddr_record()` or for even more clarity
 `dns_ip_record::  a | aaaa` and  `rr_type :: dns_ip_record() | dns_sockaddr_record()  It could also be more clear to call rr_type dns_record_type, longer but much easier to understand!

RaimoNiskanen: `inet_res:rr_type()` should probably rather be named `inet_res:dns_rr_type()` similar to `inet_res:dns_name()` and `inet_res:dns_class()`...

Other than that I think that subdividing `rr_type()` would be artificial since it simply specifies the flat set of RR types.

The name `dns_sockaddr_record` is misleading since:
*   It is not a record - it is a record type or rather a Resource Record type, which within DNS is abbreviated RR, hence "RR Type".  So "record" should be "resource_record_type" or rather an artificial subset thereof.
*   It has very little to do with a socket address ("sockaddr").  Normally, when a `#hostent` record is used, the field `h_addrtype` contains `inet` or `inet6`, which both are socket address types, a.k.a address families, so there is a connection between this field and a socket address type.  But when the field is abused by `inet_res:getbyname/2,3`, it contains a DNS RR Type, and the content of a RR need not be an address - it can be a CNAME string, a TXT string and whatnot.

The problem is `inet_res:getbyname/2,3` that returns the result as a `#hostent` record.  That return format is only suitable for RR types `a` and `aaaa` that have the translations into socket address types `inet` and `inet6`.  For all other RR types the `#hostent` record return value is unsuitable and is abused.

So `inet_res:getbyname/2,3` creates the type problem and I see no good solution.  The least bad I can think of is to type specify `#hostent.h_addrtype` to be the union of the address families that can be returned from gethosbyname(), i.e `inet | inet6`, and the DNS RR types, i.e `inet_res:rr_type()`.  That two RR types are translated into socket address types by the function so they can never occur in the result, I think is a minor problem.

But, maybe we should change `inet_res:rr_type()` into `inet_res:dns_rr_type()` before exporting it...

IngelaAndin: For the record, I see I misunderstood. We have discussed this internally and will make a different solution to try to make the best of it.

RaimoNiskanen:
Fixed by PR #5803

PR #5415, opened 2021-11-17
[parsetools] peer-ify parsetools application

max-au:
Replace test_server:start_node with `?CT_PEER`, providing
test resilience and even allowing running the same test
concurrently on the same host.


bjorng:

PR #5420, opened 2021-11-18
Change default unspecified ssh line ending to CRLF

fhunleth:
If an SSH client does not specify what line endings to use, use CRLF.
This fixes the built-in Windows ssh client that currently reports as
"OpenSSH_for_Windows_8.1p1, LibreSSL 3.0.2".

The reason for changing the default rather than adding another special
case is that OpenSSH defaults to sending CR characters when unspecified.
See the use of ONLCR at [1]. This mirrors that behavior.

Interestingly, I found ONLCR is normally on with my testing with OpenSSH
SSH to connect between MacOS and Linux, so this may not be an unusual
setting anyway for non-Windows computers. Verification was done by
capturing output and checking `stty -a | grep onlcr`.

Tested on Windows (Putty and commandline ssh in Powershell), Linux and
MacOS.

[1] https://github.com/openssh/openssh-portable/blob/2dc328023f60212cd29504fc05d849133ae47355/openbsd-compat/bsd-openpty.c#L194

### DIFF
--- a/lib/ssh/src/ssh_connection.erl
+++ b/lib/ssh/src/ssh_connection.erl
@@ -811,26 +811,20 @@ handle_msg(#ssh_msg_channel_request{recipient_channel = ChannelId,
 				    request_type = "pty-req",
 				    want_reply = WantReply,
 				    data = Data},
-	   Connection, server, SSH) ->
+	   Connection, server, _SSH) ->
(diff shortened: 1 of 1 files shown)


lisztspace:
Thanks for the contribution!

PR #5421, opened 2021-11-18
Add crypto:pbkdf2_hmac/5 function

gearnode:
I propose to add the pbkdf2_hmac/2 function to the crypto module.

PBKD2 with HMAC is used by a lot of applications and libraries in Erlang,
currently, most of it uses the basho one or a fork of the basho one. 

As Erlang already binds OpenSSL and because this function is available
even on the OpenSSL 0.9.8c, I think it can be a great UX improvement to
have it from the OTP instead of an external library.

It's why I suggest enriching the crypto module with this function.

I tried as much as possible to follow the contribution guide. Don't
hesitate to tell me if I made a mistake somewhere.


max-au:
That'd be really great to have. We are also using pbkdf2+HMAC extensively, and pure Erlang implementation seems ~50x slower.

HansN:
Our nightly tests went well.  But - the function PKCS5_PBKDF2_HMAC is not present in OpenSSL 0.9.8.  Only PKCS5_PBKDF2_HMAC_SHA1 is present in thoose early libs.  So an 

```
ERL_NIF_TERM pbkdf2_hmac_nif(ErlNifEnv* env, int argc,
                             const ERL_NIF_TERM argv[])
{
#ifdef HAS_ PKCS5_PBKDF2_HMAC
.... % your function body
#else
   return atom_notsup;
#endif
}
```
or similar is needed.  The definition of `HAS_PKCS5_PBKDF2_HMAC` should be put in openssl_config.h  There are plenty of examples there. Something like
```
#if OPENSSL_VERSION_NUMBER >= PACKED_OPENSSL_VERSION_PLAIN(1,0,0)
# define HAS_PKCS5_PBKDF2_HMAC
#endif
```
should do it.

An other thing: if you also rebase your PR onto maint, you could have this released in the 24.2 and not have to wait for 25.0.  But please do all changes asap, we have a code stop in one week.

I also have some comments on the code, but I'll put them inline.

lib/crypto/c_src/pbkdf2_hmac.c
```diff
@@ -0,0 +1,99 @@
...
+    const EVP_MD* digest;
+
+    ASSERT(argc == 5);
+
+    if ((digp = get_digest_type(argv[0])) == NULL)
+        goto bad_arg;
+    if (digp->md.p == NULL)
+        goto bad_arg;
+
+    switch (EVP_MD_type(digp->md.p))
```
HansN, line 41: This `switch .... { ... }` is a bit too complicated.  The `digest = EVP_sha1();` etc are not needed. The EVP_MD* is already present in `digp->md.p`
The check of that the EVP_MD is sha1 or sha2 could be done by adding a new flag in digest.h, add it to the definitions of digests in digest.c and finally as a test in your code that it is present. See for instance cipher.c, cipher.h.
If it seem too complicated I can add do it later.

gearnode:
> Our nightly tests went well. But - the function PKCS5_PBKDF2_HMAC is not present in OpenSSL 0.9.8. Only PKCS5_PBKDF2_HMAC_SHA1 is present in thoose early libs.

True, my bad... I've updated the patch.

> An other thing: if you also rebase your PR onto maint, you could have this released in the 24.2 and not have to wait for 25.0. But please do all changes asap, we have a code stop in one week.
Done.

lib/crypto/c_src/pbkdf2_hmac.c
gearnode: @HansN I've done something don't hesitate if it's not why you have it in mind.

lib/crypto/c_src/pbkdf2_hmac.c
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#include "common.h"
+#include "pbkdf2_hmac.h"
+#include "digest.h"
+
+ERL_NIF_TERM pbkdf2_hmac_nif(ErlNifEnv* env, int argc,
+                             const ERL_NIF_TERM argv[])
+{
+#ifdef HAS_PKCS5_PBKDF2_HMAC
```
HansN, line 29: You have forgot to push an addition to openssl_config.h with
```
#if OPENSSL_VERSION_NUMBER >= PACKED_OPENSSL_VERSION_PLAIN(1,0,0)
# define HAS_PKCS5_PBKDF2_HMAC
#endif

```

HansN (changes requested):
Otherwise it looks very good!

lib/crypto/c_src/pbkdf2_hmac.c
HansN: Your change is exactly what I tried to describe!

lib/crypto/c_src/pbkdf2_hmac.c
gearnode: Good catch :)

HansN (approved):

lib/crypto/c_src/crypto.c
```diff
@@ -90,6 +91,7 @@ static ErlNifFunc nif_funcs[] = {
     {"rand_uniform_nif", 2, rand_uniform_nif, 0},
     {"mod_exp_nif", 4, mod_exp_nif, 0},
     {"do_exor", 2, do_exor, 0},
+    {"pbkdf2_hmac_nif", 5, pbkdf2_hmac_nif, 0},
```
NelsonVides, line 94: I'm actually surprised that such a function is not using timeslices nor it is run in a dirty scheduler. Actually that makes me wonder of all other CPU hungry functions here that aren't run on dirty scheduler either, am I getting something wrong?

NelsonVides (commented):
I just have one comment, but overall I didn't imagine such a function might get into OTP, I thought this was rather an optimisation belonging to libraries, of which I made plenty of NIFs in the past. My comment is from my experience there.

HansN:
Merged to maint and master.  Thanks for the PR!

PR #5424, opened 2021-11-19
[stdlib] peer-ify application tests

max-au:
Replace usage of `test_server:start_node` with `?CT_PEER`.
Additionally, disable local_time_to_universal_time_dst when running in
Pacific Time (as it inevitably fails), and fix tar_SUITE test that
fails when `compile` module is taken from the installed system (in
that case compile.beam would be read-only, and after un-tarring
mode will be read-write, making this test fail).


bjorng:
Thanks!

PR #5432, opened 2021-11-19
[jinterface] peer-ify application

max-au:
Replace test_server:start_node with ?CT_PEER.
Cleanup unused lines while I'm at it.


bjorng:
Thanks!

PR #5436, opened 2021-11-22
Fix typos in erts/emulator/drivers

kianmeng:


erts/emulator/drivers/common/inet_drv.c
```diff
@@ -12268,7 +12268,7 @@ static int should_use_so_bsdcompat(void)
 	    fprintf(stderr, "uname: %s\r\n", strerror(sock_errno()));
 	    return 1;
 	}
-	/* Format is <version>.<patchlevel>.<sublevel><extraversion>
+	/* Format is <version>.<patchlevel>.<sublevel><extroversion>
```
bjorng, line 12271: This change does not seem to be correct. "extraversion" is not a misspelling of "extroversion", but the words "extra" and "version" concatenated.

bjorng (changes requested):

bjorng (approved):

bjorng:
Thanks!

PR #5438, opened 2021-11-22
Somewhat improve pretty printing of binaries in Dialyzer

### LINKED ISSUE #5429: Dialyzer error messages render binary literals/patterns/types differently to the original source
**Describe the bug**
Dialyzer error messages print snippets of binary literals/patterns/types in a normalized way, rather than the format in which it was given in the source code. This can lead to confusion as to whether the error message really does refer to the code snippet in question, since the error message snippet's formatting doesn't match the source.

**To Reproduce**

When Dialyzer analyzes this code:
```
-spec foo() -> ok.
foo() ->
    case foo of
      <<"bar">> -> ok;
      _ -> ok
    end.
```

it gives this error:
```
The pattern <<6447474:24>> can never match the type 'foo'
```

Notice how the binary appears as `<<"bar">>` in the source, but appears as `<<6447474:24>>` in the snippet in the error message. I think these both correspond to the same binary, so in some sense this is "technically correct", but potentially quite confusing since they don't _look_ the same.

**Expected behavior**
I'd expect the error message to use the same format as the source code, i.e. the error would be something like this:

```
The pattern <<"bar">> can never match the type 'foo'
```

**Affected versions**
HEAD of erlang/OTP right now, but I think this has been around for a fair while.

**Additional context**
Arguably this is as much a feature request as it is a bug, in that the current output is not strictly "wrong". However, in my experience, Dialyzer users see a different rendering of the binary, and wonder whether they're looking at the correct source code or not. Happy to relabel this as a feature request or otherwise as people deem appropriate.

In terms of implementing a fix, as I understand it, the issue stems from [pretty printing the core Erlang](https://github.com/erlang/otp/blob/master/lib/dialyzer/src/dialyzer.erl#L724), where knowledge of the way in which the binary pattern was expressed in the source code has been lost. `pattern_match` errors are detected in [`dialyzer_dataflow.erl`](https://github.com/erlang/otp/blob/master/lib/dialyzer/src/dialyzer_dataflow.erl#L1360), so we'd need accurate source location information or an annotation with a high-fidelity representation of the original source here. With recent changes to improve location information in terms, perhaps this is more feasible now than it was in the past?

bjorng:
When running Dialyzer on this example:

    -spec foo() -> ok.
    foo() ->
        case foo of
          <<"bar">> -> ok;
          _ -> ok
        end.

the following error would be generated:

    The pattern <<6447474:24>> can never match the type 'foo'

That message is very confusing. Unfortunately, Dialyzer has only
access to Core Erlang version of the code, so there is no easy way to
exactly pretty print code fragments. However, the optimization of
strings that produces those huge numbers can be disabled when Dialyzer
invokes the compiler, producing the following message:

    The pattern <<98,97,114>> can never match the type 'foo'

Closes #5429.

### DIFF
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -1932,11 +1932,12 @@ map_sort_key(Key, KeyMap) ->
 %% pat_bin([BinElement], State) -> [BinSeg].
 
 pat_bin(Ps0, St) ->
-    Ps = pat_bin_expand_strings(Ps0),
+    Ps = pat_bin_expand_strings(Ps0, St),
     pat_segments(Ps, St).
 
-pat_bin_expand_strings(Es0) ->
-    foldr(fun ({bin_element,Line,{string,_,[_|_]=S},default,default}, Es1) ->
+pat_bin_expand_strings(Es0, #core{dialyzer=Dialyzer}) ->
+    foldr(fun ({bin_element,Line,{string,_,[_|_]=S},default,default}, Es1)
+                when not Dialyzer ->
                   bin_expand_string(S, Line, 0, 0, Es1);
               ({bin_element,Line,{string,_,S},Sz,Ts}, Es1) ->
                   foldr(


lhoguin:
If the shell can do:

```
1> <<98,97,114>>.
<<"bar">>
```

Then should Dialyzer do that as well when pretty printing? I suspect there's more cases where the printable string was an actual binary string in Dialyzer reports, than a non-string binary.

bjorng:
Yes, it should be possible to do better in Dialyzer. I am not sure why Dialyzer pretty prints terms the way it does, and I would want to understand more about it before I try to fix it.

TD5:
> If the shell can do:
> 
> ```
> 1> <<98,97,114>>.
> <<"bar">>
> ```
> 
> Then should Dialyzer do that as well when pretty printing? I suspect there's more cases where the printable string was an actual binary string in Dialyzer reports, than a non-string binary.

At least with respect to the original issue, I think the ideal case would be where the error message reflected the format used in the original source code. If the user wrote `<<"bar">>`, then that would be the format that appears in the error message, but if they wrote `<<98,97,114>>`, then this would be the format used.

In this way, I think my expectations of the rendering in the shell and Dialyzer would be different: For me, the shell would print in a standardised, canonical format, but ideally Dialyzer would always respect the format used in the source, since the aim with Dialyzer is to help the user understand the issue that was detected in their code, rather than the runtime value generated in the case of the shell. I am not sure if my view is widely held here or not, though, and I suspect it's more technically difficult to achieve that than just adopting the shell's logic.

bjorng:
> In this way, I think my expectations of the rendering in the shell and Dialyzer would be different: For me, the shell would print in a standardised, canonical format, but ideally Dialyzer would always respect the format used in the source, since the aim with Dialyzer is to help the user understand the issue that was detected in their code, rather than the runtime value generated in the case of the shell.

I agree. Ideally what Dialyzer prints should reflect the source code being analyzed.

> I am not sure if my view is widely held here or not, though, and I suspect it's more technically difficult to achieve that than just adopting the shell's logic.

Yes, it is more difficult (I hinted how it could be done in a comment to #5429). Until we can do it properly, we might update Dialyzer to do it like the shell does.

PR #5441, opened 2021-11-23
compiler: Preserve extended error information

### LINKED ISSUE #5440: Inconsistent error messages when errors are optimized away by the compiler
**Describe the bug**

Take this code:

```erlang
-module(foo).
-compile(export_all).

example1() ->
  try
     atom_to_binary(11111)
  catch
     error:Error:Stacktrace ->
        erlang:display({Error, Stacktrace})
  end.

example2(Binary) ->
  try
     atom_to_binary(Binary)
  catch
     error:Error:Stacktrace ->
        erlang:display({Error, Stacktrace})
  end.
```

The functions above will raise different error messages in the shell. I assume this happens because the actual function call is eliminated by the compiler in `example1`:

```erlang
1> foo:example1().
{badarg,[{foo,example1,0,[{file,"foo.erl"},{line,6}]},{erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,689}]},{shell,exprs,7,[{file,"shell.erl"},{line,686}]},{shell,eval_exprs,7,[{file,"shell.erl"},{line,642}]},{shell,eval_loop,3,[{file,"shell.erl"},{line,627}]}]}
true
2> foo:example2(1111).
{badarg,[{erlang,atom_to_binary,[1111],[{error_info,#{module=>erl_erts_errors}}]},{foo,example2,1,[{file,"foo.erl"},{line,14}]},{erl_eval,do_apply,6,[{file,"erl_eval.erl"},{line,689}]},{shell,exprs,7,[{file,"shell.erl"},{line,686}]},{shell,eval_exprs,7,[{file,"shell.erl"},{line,642}]},{shell,eval_loop,3,[{file,"shell.erl"},{line,627}]}]}
```

**Expected behavior**

I thought I would open up the issue for discussion as it would be nice if both messages were the same for consistency. Perhaps the simplest way is to not remove the problematic call, even if the error is found at compile time? Another idea is to inline the relevant `error_info` in the call site too.

However, I can also see that the consistency between the messages may not really be a strong requirement, so feel free to close this. :)

**Affected versions**

Since the fantastic EEP 54 was added. :)

bjorng:
The compiler currently replaces known failing calls with a call to
`error(badarg)`. Example:

    foobar() ->
        atom_to_binary(42).

The compiler rewrites that to:

    foobar() ->
        error(badarg).

With the introduction of extended error information (EEP 54), that
"optimization" results in no extended error information being shown:

    1> t:foobar().
         in function  t:foobar/0 (t.erl, line 5)

To ensure that extended error information is always shown, update
the compiler so that it will keep the known failing call (but still
emit a warning). Thus:

    2> t:foobar().
         in function  atom_to_binary/1
            called as atom_to_binary(42)
            *** argument 1: not an atom

Closes #5440.


garazdawi:
Is there any need to add a testcase to exception_SUITE:error_info in erts?

bjorng:
> Is there any need to add a testcase to exception_SUITE:error_info in erts?

Not really. Such test case would probably never ever fail.

eksperimental:
Awesome. Thank you @bjorng

PR #5444, opened 2021-11-24
Improve runtime_tools scheduler (utilization) module

### LINKED ISSUE #5425: `scheduler:sample/0` reports inconsistent values from child processes
**Describe the bug**

When calling `scheduler:sample/0` from a spawned process, time values for the `normal` schedulers will only be reported starting from the moment the child process was started. This makes values reported from different processes inconsistent with each other. When passing samples from different child processes to `scheduler:utilization`, the resulting utilization values will be meaningless.

However, this behaviour only happens if the parent process has not itself called `scheduler:sample/0` before -- after the parent process calls `scheduler:sample/0`, the values reported from the child processes will be consistent with each other.

[Note: I'm describing this in terms of `scheduler:sample/0`, but the same issue can be observed by calling `erlang:statistics(scheduler_wall_time)` directly]

**To Reproduce**

To simplify the resulting output, we'll call `element(4, hd(element(2, scheduler:sample())))`, obtaining the total time utilization of the first normal scheduler. Note that the behaviour can be observed for the total and active time values of all schedulers.

From an ERL shell, without any previous calls to `scheduler:sample()`:

```
spawn(fun() -> erlang:display(element(4, hd(element(2, scheduler:sample())))) end).
timer:sleep(1000).
spawn(fun() -> erlang:display(element(4, hd(element(2, scheduler:sample())))) end).
```

**Expected behavior**

The time values returned by these two functions should be at least `1000000` microseconds apart, so that calculating the utilization from them returns meaningful values.

Alternatively, if the value is meant to have a per-process scope, then it should be documented as such, and then the values of parent processes should not leak to child processes. Note that the following **does** correctly return values from the spawned processes that are at least `1000000` microseconds apart:

```
scheduler:sample().
spawn(fun() -> erlang:display(element(4, hd(element(2, scheduler:sample())))) end).
timer:sleep(1000).
spawn(fun() -> erlang:display(element(4, hd(element(2, scheduler:sample())))) end).
```

That is, the moment you call `scheduler:sample/0` on the parent process, values obtained by calling it from its child processes have time values that are consistent with the parent process. My hunch is that calling `scheduler:sample/0` from the parent process initialises some scheduler time counters that get passed to the child process.

EDIT: it seems that calling `erlang:system_flag(scheduler_wall_time, true)` from the parent process has the same effect. Calling `scheduler:sample()` implicitly sets that flag, but when it is called in a child process, the flag is not propagated to the parent.

**Affected versions**
Tested on OTP 24. Older versions may be affected as well.

**Additional context**

I've found this issue when fetching samples from independent `Elixir.Task` processes and calculating the utilization.

sverker:
- Add `scheduler:get_sample/0` and `scheduler:get_sample_all/0`
- Clarify docs of `scheduler` module regarding system flag `scheduler_wall_time`.
- Clarify docs of `erlang:system_flag(scheduler_wall_time,_)`
- Fix bug in `scheduler:utilization(Seconds)` not always restoring `scheduler_wall_time`.

Closes #5425

### DIFF
--- a/lib/runtime_tools/doc/src/scheduler.xml
+++ b/lib/runtime_tools/doc/src/scheduler.xml
@@ -35,12 +35,52 @@
   <module since="OTP 21.0">scheduler</module>
   <modulesummary>Measure scheduler utilization</modulesummary>
   <description>
-    <p>This module contains utility functions for easier measurement and
-    calculation of scheduler utilization, otherwise obtained from calling the
-    more primitive <seeerl marker="erts:erlang#statistics_scheduler_wall_time">
-    <c>statistics(scheduler_wall_time)</c></seeerl>.</p>
-    <p>The simplest usage is to call <seemfa marker="#utilization/1">
-    <c>scheduler:utilization(Seconds)</c></seemfa>.</p>
+    <p>
+      This module contains utility functions for easy measurement and
+      calculation of scheduler utilization. It act as a wrapper around the more
+      primitive API <seeerl marker="erts:erlang#statistics_scheduler_wall_time">
+      <c>erlang:statistics(scheduler_wall_time)</c></seeerl>.
+    </p>
+    <p>
+      The simplest usage is to call the blocking <seemfa marker="#utilization/1">
+      <c>scheduler:utilization(Seconds)</c></seemfa>.
+    </p>
+    <p>
+      For non blocking and/or continuous calculation of scheduler utilization,
+      the recommended usage is:
+    </p>
+    <list>
+      <item><p>
+	First call <seeerl marker="erts:erlang#system_flag_scheduler_wall_time">
+	<c>erlang:system_flag(scheduler_wall_time,true)</c></seeerl> to enable
+	scheduler wall time measurements.
+      </p></item>
+      <item><p>
+	Call <seemfa marker="#get_sample/0"><c>get_sample/0</c></seemfa> to
+	collect samples with some time in between.
+      </p></item>
+      <item><p>
+	Call <seemfa marker="#utilization/2"><c>utilization/2</c></seemfa> to
+	calculate the scheduler utilization in the interval between two
+	samples.
+      </p></item>
+      <item><p>
+	When done call
+	<seeerl marker="erts:erlang#system_flag_scheduler_wall_time">
+	<c>erlang:system_flag(scheduler_wall_time,false)</c></seeerl> to disable
+	scheduler wall time measurements and avoid unecessary cpu overhead.
+      </p></item>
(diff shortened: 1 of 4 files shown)


sverker:
@unflxw Would be good if you could review the documentation changes:

- the scheduler modulehttps://erlang.org/~sverker/pr-5444/lib/runtime_tools-1.17/doc/html/scheduler.html
https://erlang.org/~sverker/pr-5444//erts-12.1.5/doc/html/erlang.html#system_flag_scheduler_wall_time

lib/runtime_tools/doc/src/scheduler.xml
```diff
@@ -35,12 +35,53 @@
...
+	<seeerl marker="erts:erlang#system_flag_scheduler_wall_time">
+	<c>erlang:system_flag(scheduler_wall_time,false)</c></seeerl> to disable
+	scheduler wall time measurements and avoid unecessary cpu overhead.
+      </p></item>
+      </list>
+      <p>
+	Note scheduler wall time measurements will automatically be disabled if
+	the process that called
+	<seeerl marker="erts:erlang#system_flag_scheduler_wall_time">
+	<c>erlang:system_flag(scheduler_wall_time,true)</c></seeerl> terminates.
```
unflxw, line 74: It might be good, here or at some other point in this introduction, to explain what the consequences of the flag being set to false between samples is -- specifically, that an utilization measurement taken from two samples won't have meaningful values if `scheduler_wall_time` was set to `false` (or the process that set it to true was terminated) between samples.

unflxw (approved):
Apologies for not getting back to you earlier about these changes, life got in the way. The documentation changes look good to me -- the relationship between the scheduler module and the system flag is clear now, and the recommended usage pattern (along with the new `get_sample/0`) makes it harder to accidentally misuse it.

I understand the drive behind the initial implementation of this module, wanting to abstract away the existence of this flag entirely. But given the way the flag works, I think that the approach you've taken, dealing with the flag's existence at the documentation's level and asking the module's user to enable it, makes it harder for the user to use this module incorrectly.

Additionally, the note about CPU overhead is valuable, and has helped me realise there's a flaw with my always-keep-the-flag-on workaround, so thank you for that as well.

lib/runtime_tools/doc/src/scheduler.xml
sverker: In note box for sample/0 I wrote: "_If scheduler_wall_time has been disabled between two samples, passing them to utilization/2 will yield invalid results._".  But maybe that should be mentioned up in the description (also)?

PR #5447, opened 2021-11-25
Fix typos in erts/emulator/beam

kianmeng:


erts/emulator/beam/erl_process.c
```diff
@@ -14123,7 +14123,7 @@ erts_continue_exit_process(Process *p)
         /* Set state to not active as we don't want this process
            to be scheduled in again after this. */
         {
-            /* Inactivate */
+            /* Deactivate */
```
garazdawi, line 14126: Whatever tool it is that you are using should make sure not to do this one :)

garazdawi (changes requested):

bjorng:
Please squash the commits.

PR #5448, opened 2021-11-26
Update some vars (line to position) in yecc docs

gomoripeti:
The text already mentions "position (eg. line number)" and the spec also refers to location in 
https://github.com/erlang/otp/blob/master/lib/parsetools/include/yeccpre.hrl#L50,
so this change is only to avoid any misunderstanding that the position must be a line number.


bjorng:
Thanks!

PR #5451, opened 2021-11-26
Fix iOS build

dominicletz:
iOS doesn't ship <net/if_arp.h> so that file can't be included on that platform. This fix removes the include on iOS.

It seems it was anyway only included to check for additionally defined constants, so no further change here. 

FYI @garazdawi


bmk (approved):

garazdawi:
Thanks!

PR #5452, opened 2021-11-27
Fixed OTP_RELEASE macro

czakop:
One of the examples (OTP_RELEASE macro) had a logical mistake.

### DIFF
--- a/system/doc/reference_manual/macros.xml
+++ b/system/doc/reference_manual/macros.xml
@@ -255,9 +255,11 @@ or
     %% Code that will work in OTP 22 or higher
   -elif(?OTP_RELEASE >= 21).
     %% Code that will work in OTP 21 or higher
+  -else.
+    %% OTP 20 or lower.
   -endif.
 -else.
-  %% OTP 20 or lower.
+  %% OTP_RELEASE is not defined.
 -endif.
 ...</code>
     <p>The code uses the <c>OTP_RELEASE</c> macro to conditionally


system/doc/reference_manual/macros.xml
```diff
@@ -255,9 +255,11 @@ or
     %% Code that will work in OTP 22 or higher
   -elif(?OTP_RELEASE >= 21).
     %% Code that will work in OTP 21 or higher
+  -else.
+    %% OTP 20 or lower.
   -endif.
 -else.
-  %% OTP 20 or lower.
+  %% OTP_RELEASE is not defined.
```
mikpe, line 262: This change is wrong. The `-else.` _implies_ that `OTP_RELASE` isn't defined, but the _implication_ is that we're on OTP-20 or older, which is the information we want.

system/doc/reference_manual/macros.xml
```diff
@@ -255,9 +255,11 @@ or
     %% Code that will work in OTP 22 or higher
   -elif(?OTP_RELEASE >= 21).
     %% Code that will work in OTP 21 or higher
+  -else.
+    %% OTP 20 or lower.
```
mikpe, line 259: This `-else.` is dead code. Read line 253...

bjorng:
Thanks for your pull request.

The documentation is correct. The `OTP_RELEASE` macro was introduced in OTP 21; thus, as @mikpe pointed out, if it is not defined it implies that the code is running on OTP 20 or lower.

PR #5457, opened 2021-11-29
ssl: Allow binary keyfile passwords

lhoguin:
A user has reported that while `{password, <<"pass">>}` doesn't work, `{password, [<<"pass">>]}` does. It doesn't seem to be useful to restrict the password to a list string.

For context: RabbitMQ has an option that allows encrypting passwords, but decrypted passwords are always returned as binary. It would be good to avoid special casing this.

This is a very early draft of a potential PR. There's a couple more options that could be improved by allowing binary strings as well.

@IngelaAndin What do you think?

### DIFF
--- a/lib/public_key/src/pubkey_pbe.erl
+++ b/lib/public_key/src/pubkey_pbe.erl
@@ -33,7 +33,7 @@
 %%====================================================================
 
 %%--------------------------------------------------------------------
--spec encode(binary(), string(), string(), term()) -> binary().
+-spec encode(binary(), iodata(), string(), term()) -> binary().
 %%
 %% Description: Performs password based encoding
 %%--------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


IngelaAndin:
That option was first introduced a very long ago, so I think this is an improvement to the newer circumstances that would be beneficial :)

lhoguin:
On the one hand users are highly unlikely to give an iolist for the password. On the other hand the crypto function accepts it. Up to you.

IngelaAndin:
After considering this in our team we think that probably the best is to change ssl and public_key specs to iodata() since that will keep the elements to octets. And perhaps write something enlightening in the documentation about it.

lhoguin:
Sounds good. I will work on this next week as time allows.

lhoguin:
I have updated and pushed. I have not run Dialyzer, how do I do that?

IngelaAndin:
make dialyzer in ssl directory I think should be the simplest way

lhoguin:
Understood, thanks. I got confused because my auto complete didn't catch any commands. Turns out they're not there until some files are built, and I had a perfectly clean checkout.

lhoguin:
All good from my end.

IngelaAndin:
FYI, we are waiting for us to complete PR-5320 (we are working on the tests). We would like to merge that first and see if we need any adjustments for this one after that. Otherwise, it is looking good in the test.

IngelaAndin:
@lhoguin we are almost ready for merge, could you rebase your branch to get rid of the merge conflicts.

lhoguin:
Rebased.

Issue #3064, opened 2016-02-23
ERL-97: eunit run twice tests when both module and module_tests given


OTP-Maintainer:
Original reporter: `crownedgrouse`
Affected version: `Not Specified`
Component: `eunit`
Migrated from: https://bugs.erlang.org/browse/ERL-97

---

```
Hi,
I have a module 'xxxx.erl' (no tests in it) and 3 eunit tests in 'xxxx_tests.erl' .

```
1> eunit:test(['xxxx_tests'],[]).
  All 3 tests passed.
ok
2> eunit:test(['xxxx'],[]).      
  All 3 tests passed.
3> eunit:test(['xxxx_tests','xxxx'],[]).
  All 6 tests passed.```

So giving 'xxxx' launch tests in 'xxxx_tests'  (documentation explain this ) but giving both modules run twice the tests... 

```

PR #5461, opened 2021-11-29
Delete duplicates of unit tests

vkatsuba:
Remove duplicates in tests before run unit.
This is need for case if in tests will be used duplicates
with/without prefix, example:
* Input  - `[issue_tests, issue]`
* Output - `[issue]`

Related to https://github.com/erlang/otp/issues/3064

### DIFF
--- a/lib/eunit/src/eunit.erl
+++ b/lib/eunit/src/eunit.erl
@@ -145,7 +145,8 @@ test(Tests, Options) ->
 
 %% @private
 %% @doc See {@link test/2}.
-test(Server, Tests, Options) ->
+test(Server, Tests0, Options) ->
+    Tests = delete_duplicates(Tests0),
     Listeners = listeners(Options),
     Serial = eunit_serial:start(Listeners),
     case eunit_server:start_test(Server, Serial, Tests, Options) of
@@ -219,7 +220,7 @@ listeners(Options) ->
 start_listeners([P | Ps]) when is_pid(P) ; is_atom(P) ->
     [P | start_listeners(Ps)];
 start_listeners([{Mod, Opts} | Ps]) when is_atom(Mod) ->
-    [Mod:start(Opts) | start_listeners(Ps)];	    
+    [Mod:start(Opts) | start_listeners(Ps)];
 start_listeners([]) ->
     [].
 
@@ -264,4 +265,40 @@ all_options(Opts) ->
 	    end
     catch
 	_:_ -> Opts
+    end.
+
+%% Delete duplicates of tests
+
+delete_duplicates([]) ->
+    [];
+delete_duplicates([H|T]) when is_atom(H) ->
+    case is_module(H) of
+        true ->
+            C = prepare_candidate(H),
+            [C | [X || X <- delete_duplicates(T), C /= X]];
+        false ->
+            [H | delete_duplicates(T)]
+    end;
+delete_duplicates([H|T]) ->
+    [H | delete_duplicates(T)];
+delete_duplicates(Tests) ->
+    Tests.
+
+is_module(M) ->
+    try M:module_info(exports) of
+        _ ->
+            true
+    catch
+        error:undef ->
+            false
+    end.
+
+prepare_candidate(Test) ->
+    Name = atom_to_list(Test),
+    case lists:suffix(?DEFAULT_TESTMODULE_SUFFIX, Name) of
+        true ->
+            Index = length(Name) - length(?DEFAULT_TESTMODULE_SUFFIX),
+            list_to_atom(lists:sublist(Name, Index));
+        false ->
+            list_to_atom(Name)
     end.
--- a/lib/eunit/test/eunit_SUITE.erl
+++ b/lib/eunit/test/eunit_SUITE.erl
@@ -23,7 +23,7 @@
 	 init_per_group/2,end_per_group/2,
 	 app_test/1,appup_test/1,eunit_test/1,surefire_utf8_test/1,surefire_latin_test/1,
 	 surefire_c0_test/1, surefire_ensure_dir_test/1,
-	 stacktrace_at_timeout_test/1]).
+	 stacktrace_at_timeout_test/1, eunit_duplicates_test/1]).
 
 -include_lib("common_test/include/ct.hrl").
 
@@ -31,7 +31,8 @@ suite() -> [{ct_hooks,[ts_install_cth]}].
 
 all() ->
     [app_test, appup_test, eunit_test, surefire_utf8_test, surefire_latin_test,
-     surefire_c0_test, surefire_ensure_dir_test, stacktrace_at_timeout_test].
+     surefire_c0_test, surefire_ensure_dir_test, stacktrace_at_timeout_test,
+     eunit_duplicates_test].
 
 groups() ->
     [].
@@ -90,6 +91,32 @@ stacktrace_at_timeout_test(Config) when is_list(Config) ->
             ct:fail(missing_stacktrace_in_surefire);
         _ ->
             ok
+    end.
+
+eunit_duplicates_test(Config) when is_list(Config) ->
+    LogFile = "eunit-events-duplicates.log",
+    ok = file:set_cwd(code:lib_dir(eunit)),
+    Tests = [[eunit, eunit_tests, eunit], eunit, eunit_tests, [eunit],
+             [eunit_tests], [eunit, eunit], [eunit, eunit_tests],
+             [eunit_tests, eunit], [eunit_tests, eunit_tests]],
+    case eunit_duplicates(LogFile, Tests) of
+    ok ->
+        ok;
+    {error, Test, Log} ->
+        ct:fail("Test point ~p must have only one event in:~n~p", [Test, Log])
+    end.
+
+eunit_duplicates(LogFile, []) ->
+    ok = file:delete(LogFile),
+    ok;
+eunit_duplicates(LogFile, [H|T]) ->
+    ok = eunit:test(H, [{event_log, LogFile}]),
+    {ok, Events} = file:consult(LogFile),
+    case [Event || Event = {_, [Count], _} <- Events, Count /= 1] of
+    [] ->
+        eunit_duplicates(LogFile, T);
+    [_|_] ->
+        {error, H, Events}
     end.
 
 check_surefire(Module) ->


lib/eunit/src/eunit.erl
```diff
@@ -265,3 +266,49 @@ all_options(Opts) ->
...
+
+%% Delete duplicates of tests
+
+delete_duplicates([_|_] = Tests) ->
+    Candidates = [atom_to_list(T) || T <- Tests, is_atom(T), is_mod_exist(T)],
+    delete_duplicates(Candidates, Tests);
+delete_duplicates(Tests) ->
+    Tests.
+
+is_mod_exist(M) ->
```
bjorng, line 278: I suggest changing the name to `does_module_exist` or `is_module`.

lib/eunit/src/eunit.erl
```diff
@@ -265,3 +266,49 @@ all_options(Opts) ->
...
+    try M:module_info(exports) of
+    _ ->
+        true
+    catch
+    _:_ ->
+        false
+    end.
```
bjorng, line 285:
```suggestion
    try M:module_info(module) of
        _ ->
            true
    catch
        error:undef ->
            false
    end.

```

lib/eunit/src/eunit.erl
```diff
@@ -265,3 +266,49 @@ all_options(Opts) ->
...
+                    Tests
+            end;
+        false ->
+            Candidate = list_to_atom(Name ++ ?DEFAULT_TESTMODULE_SUFFIX),
+            case lists:member(Candidate, Tests) of
+                true ->
+                    lists:delete(Candidate, Tests);
+                false ->
+                    Tests
+            end
```
bjorng, line 313: It seems unnecessary to test for membership before deleting since the `Tests` lists unlikely to be very long.

```suggestion
            Tests -- [Candidate]
```

lib/eunit/src/eunit.erl
```diff
@@ -265,3 +266,49 @@ all_options(Opts) ->
...
+delete_duplicates([], _, Tests) ->
+    Tests;
+delete_duplicates([H|T], Candidates, Tests) ->
+    delete_duplicates(T, Candidates, delete_candidate(H, Candidates, Tests)).
+
+delete_candidate(Name, Candidates, Tests) ->
+    case lists:suffix(?DEFAULT_TESTMODULE_SUFFIX, Name) of
+        true ->
+            Delta = length(Name) - length(?DEFAULT_TESTMODULE_SUFFIX),
+            CatName = lists:sublist(Name, 1, Delta),
```
bjorng, line 299:
```suggestion
            CatName = lists:sublist(Name, Delta),
```

bjorng (commented):
Thanks for your pull request.

This is not a complete review, just a few comments about some things that I think could be improved in the code.

I think that you'll need to add a test case to ensure that your duplication removal works correctly and keeps working.

lib/eunit/src/eunit.erl
vkatsuba: Done.

lib/eunit/src/eunit.erl
vkatsuba: Done.

lib/eunit/src/eunit.erl
vkatsuba: Done.

lib/eunit/src/eunit.erl
vkatsuba: Done.

IngelaAndin:
This does not seem to break anything but as @bjorng pointed out we would like to have a test to ensure that it works and keeps working.

vkatsuba:
One test case is added. Could you please help me understand - what other tests you expected from my side?

IngelaAndin:
Oh, now I see the test.  I expect the test to be in the test suite and not in the src code.

`eunit/test/eunit_SUITE.erl `

vkatsuba:
Should I move this test to eunit_SUITE.erl or we can keep it as is?

IngelaAndin:
Please move it.

vkatsuba:
@IngelaAndin done.

u3s:
i browsed the code related to this functionality and seems to me that it is not tested that well.
could you extend the test case a bit further, so we can get a better coverage and control over change introduced?

I'm thinking about following inputs to eunit:test calls (apart from input already proposed):

1. eunit
2. eunit_tests
3. [eunit]
4. [eunit_tests]
5. [eunit,eunit]
6. [eunit,eunit_tests]
7. [eunit_tests, eunit]
8. [eunit_tests, eunit_tests]

I assume that should be quite simple, if some generic code is brought into test case. Maybe list above could be a list of tuples combining {TestInput, ExpectedExecutionResults} and then consumed by test case ...

vkatsuba:
@u3s many thanks! Yep, I'm  miss of some test points. Fixed.

u3s:
1. please check shell output below - code from your PR and the base commit was used
2. eunit_tests contains 7 tests and eunit contains 1 test (which I added manually
   for both commits)
3. It seems that test from eunit module is not executed with your proposal
4. could eunit_duplicates_test be improved so it controls this behavior?
5. regarding changes in eunit module - I plan to discuss with the team if we're
   fine with the proposal or would like some different approach. I will let you know
   next week.
   However, improving the test case is something that would be good no matter
   what discussion output will.

```
**** test added to eunit module
git diff
diff --git a/lib/eunit/src/eunit.erl b/lib/eunit/src/eunit.erl
index 5b9ead74e9..db6dc20103 100644
--- a/lib/eunit/src/eunit.erl
+++ b/lib/eunit/src/eunit.erl
@@ -302,3 +302,6 @@ prepare_candidate(Test) ->
         false ->
             list_to_atom(Name)
     end.
+
+dummy_test() ->
+    ok.

**** 32d401265d (with test added to eunit)
1> eunit:test(eunit).
  All 8 tests passed.
ok
2> eunit:test(eunit_tests).
  All 7 tests passed.
ok
3> eunit:test([eunit]).
  All 8 tests passed.
ok
4> eunit:test([eunit_tests]).
  All 7 tests passed.
ok
5> eunit:test([eunit,eunit]).
  All 16 tests passed.
ok
6> eunit:test([eunit,eunit_tests]).
  All 15 tests passed.
ok
7> eunit:test([eunit_tests, eunit]).
  All 15 tests passed.
ok
8> eunit:test([eunit_tests, eunit_tests]).
  All 14 tests passed.
ok

**** 4bf53f3e10 (with test added to eunit)
1> eunit:test(eunit).
  All 7 tests passed.
ok
2> eunit:test(eunit_tests).
  All 7 tests passed.
ok
3> eunit:test([eunit]).
  All 7 tests passed.
ok
4> eunit:test([eunit_tests]).
  All 7 tests passed.
ok
5> eunit:test([eunit,eunit]).
  All 7 tests passed.
ok
6> eunit:test([eunit,eunit_tests]).
  All 7 tests passed.
ok
7> eunit:test([eunit_tests, eunit]).
  All 7 tests passed.
ok
8> eunit:test([eunit_tests, eunit_tests]).
  All 7 tests passed.
ok
9>
```

vkatsuba:
The output looks exactly as I expected(as it was difficult for me to choose what to leave on the list - so I left without a prefix - the main issue was - what name should be prioritized with or without prefix during duplication?). But maybe this is not what your team expects, so, I will waiting for feedback in next week and maybe it might be worth implementing a different approach. Thanks.

**P.S.** To be honest, I don't really like this approach either which is provided in current PR, maybe need change it as:
1 - Detect duplicates and remove all duplicates with and without prefixes
2 - Check if module with and without prefixes is have test cases
3 - if one of the modules contains tests and the other does not, give priority to module with tests
4 - If two modules have tests - collect them in one place and run it

u3s:
Summary of discussion we had about this change.
Currently PR proposes 2 adjustments to become new default execution flow in Eunit framework.
For the discussion purpose, I will describe each of them separately - please comment if I'm missing something.

OPTION 1
1. is about reverting side effects of EUnit heuristic resulting with execution of tests found in modules `fred`, `fred_tests` if `eunit:test(fred)` or `eunit:test([fred])` is called
2. is probably the root problem reported in #3064
3. I'm presuming some build systems or libraries might prefer call eunit:test on all modules they manage to find without analyzing their content or relation between filenames. And not expecting execution containing tests from `fred`, `fred_tests`, `fred_tests`  if they call `eunit:test([fred, fred_tests])`
4. instead execution containing tests from `fred`, `fred_tests` if they call `eunit:test([fred, fred_tests])` could be preferred

OPTION 2
1. is about removing duplicate items on the input list - so if `eunit:test([eunit, eunit, eunit])` is called execution flow happens as if user called `eunit:test([eunit])`

Proposal for OPTION 1
1. leave legacy execution flow unchanged
2. add new option for eunit:test/2 - so that users who don't like the heuristic can disable it
3. when used, new option should result with _not_ executing related code in `eunit_data:get_module_tests1/2`
4. it would be nice to have it documented
5. it would be nice to have a test(or tests) covering heuristic enabled (default) and disabled(when new option used)

Proposal for OPTION 2
1. we are having doubts about it - maybe it was not described precisely enough
2. firstly, some of the users might see a purpose in having duplicates in the input list
3. secondly, we are not sure if it is a problem reported in #3064, although maybe some of the comments could suggest that direction
4. we would like to skip it for now - unless you have a clear view on what the functionality should be and would like to discuss that

vkatsuba:
@u3s many thanks for feedback! I need spend some time to take a look into it a little big closer. As soon as I have any updates, I will share them here or update the PR with separate commits.

u3s:
any updates?

vkatsuba:
> any updates?

Was need to switch to other stuff. I hope will try to improve this PR till end of next week.

u3s:
please let us know if you plan to work on that one or you have different priorities right now and would like us to take over this one.

vkatsuba:
> please let us know if you plan to work on that one or you have different priorities right now and would like us to take over this one.

Yes, I want to work on it. But at the same time I have some important tasks which has become a little more lately than I expected ... So, if you feel that it would be more productive if your team took on this task - please go ahead. If by some reason we will provide implementations at the same time, I will not insist on the adoption of my PR - because it doesn't matter who fixes it, what matters is that the fix will provided :upside_down_face:.

u3s:
OK. I will let you know if I start working on it, to reduce risk of unnecessary duplicated work.

u3s:
This PR was replaced with #6157 which implements description from section "Proposal for OPTION 1" above.
Thanks for input and discussions which lead to improving eunit behavior. Sorry for delays.

vkatsuba:
Great news! Many thanks for let me know!

Issue #5463, opened 2021-11-30
diameter Address type only handles IP-addresses


sebastiw:
**Describe the bug**

Diameter crashes when receiving an E.164 address in an AVP of type `Address`.

**To Reproduce**
E.g. Receiving an [Ro/CCR](https://www.etsi.org/deliver/etsi_ts/132200_132299/132299/16.02.00_60/ts_132299v160200p.pdf) with Client-Address set to an E.164 address.

**Expected behavior**
No crash, but return the octet string/decoded address.

**Affected versions**
All OTP versions since R14 (since the introduction of the diameter application)

**Additional context**

According to [RFC3588 section 4.3](https://datatracker.ietf.org/doc/html/rfc3588#section-4.3) the Address type is of OctetString where the first octet represents the Address Type.

> The first two octets of the Address AVP represents the Address
> Type, which contains an Address Family defined in [IANAADFAM]. 
> The AddressType is used to discriminate the content and format of the remaining octets.

And according to [IANAADFAM](https://www.iana.org/assignments/address-family-numbers/address-family-numbers.xhtml) there is a lot of different Address Types, but only IPv4 and IPv6 are handled by diameter application.

From [diameter_types.erl](https://github.com/erlang/otp/blob/master/lib/diameter/src/base/diameter_types.erl#L245) only IP-addresses are allowed.

```
'Address'(encode, zero, _) ->
    <<0:48>>;

'Address'(decode, <<A:16, B/binary>>, _)
  when 1 == A,  4 == size(B);
       2 == A, 16 == size(B) ->
    list_to_tuple([N || <<N:A/unit:8>> <= B]);

'Address'(decode, B, _) ->
    ?INVALID_LENGTH(B);

'Address'(encode, T, _) ->
    Ns = tuple_to_list(diameter_lib:ipaddr(T)),  %% length 4 or 8
    A = length(Ns) div 4,                        %% 1 or 2
    B = << <<N:A/unit:8>> || N <- Ns >>,
    <<A:16, B/binary>>.
```

Best regards

otp-anders:
You're right that support for other address types is missing (which I'll fix), but where is the crash? Reception of It should just result in a decode error that's handled like any other.

sebastiw:
sorry it was most likely a decode error (I got the issue from a colleague)

otp-anders:
Addressed in 25.0. (Missed closing this.)

PR #5468, opened 2021-12-01
Diameter Address types

sebastiw:
Support full list of IANA family numbers

Suggestion for solving #5463. I tried to validate byte sizes of some of the types, but there is a lot to improve here.

### DIFF
--- a/lib/diameter/src/base/diameter_types.erl
+++ b/lib/diameter/src/base/diameter_types.erl
@@ -242,19 +242,335 @@
 'Address'(encode, zero, _) ->
     <<0:48>>;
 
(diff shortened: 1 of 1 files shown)


otp-anders:
Sorry for the slow response.

Thanks, I think I'll close this and rework it a bit. I don't think we want to introduce atoms for all of these types, and if there's nothing well-defined to decode to (eg. NSAP seems to have encoding possibilities of its own) then the value can just as well be left as binary() for the user to decode. The main problem now is that everything other than two address types is treated as a decode error.

I'll put some solution into OTP 25.

PR #5477, opened 2021-12-02
Add support for static Elixir NIF modules

sverker:
or any modules with "exotic" characters in the name (like '.').

* Add STATIC_ERLANG_NIF_LIBNAME macro which is used both to
  identify the archive name and create a unique C identifier for
  the _nif_init function.

* Run all static *_nif_init functions unconditionally at VM boot to
  get the corresponding module names. erlang:load_nif/2 can then
  match only against module names and does not have to care about
  archive filenames for static nifs.


dominicletz:
Just late confirmation here, this works well with the static compilation for iOS binaries as well and we've updated the libraries we're using from the previous macro hack to this new `STATIC_ERLANG_NIF_LIBNAME` macro in 25.0.4 . Including exqlite https://github.com/elixir-sqlite/exqlite/pull/213/files

Thanks @sverker

PR #5479, opened 2021-12-03
Add -nifs() attribute

sverker:
New `nifs` attribute.

```
-module(m).
-export([foo/0, bar/1, baz/2]).
-nifs([foo/0, baz/2, mylocal/3]).
```

The purpose it to give compiler and loader information about which functions that may be overridden as NIFs by `erlang:load_nif/2`, and thereby empower them to make better decisions, for example regarding optimizations.

To maintain **backward compatibility**, if no `-nifs()` in module, allow any functions as before. If `-nifs()` are declared, fail `erlang:load_nif/2` if undeclared functions are specified as NIFs in the dynamic lib.

This PR does

- Add new beam instruction `nif_start` at the top of every function declared as nif. Only to be used by compiler itself and loader to identify potential NIFs.
- Restrict the function size padding we do, to only declared nifs. Function size padding is done during load of Erlang functions to ensure minimal size needed for `call_nif` instruction overwritten by `erlang:load_nif`.
- Fix the "match state passed to NIF" bug by letting compiler disable match state optimizations for declared nifs.

Other potential use:
- Let compiler avoid inlining for functions declares as nifs.
- Add padding with no-ops before recursive call instructions early in functions declared as nifs. This will make sure the instruction after a recursive call (at the return address) can never be destroyed when writing the call_nif instruction.

### DIFF
--- a/erts/doc/src/erl_nif.xml
+++ b/erts/doc/src/erl_nif.xml
@@ -100,6 +100,8 @@ ERL_NIF_INIT(niftest,nif_funcs,NULL,NULL,NULL,NULL)</code>
 
 -export([init/0, hello/0]).
 
+-nifs([hello/0]).
+
 -on_load(init/0).
 
 init() ->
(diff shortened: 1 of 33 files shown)


asabil:
Given that the existing attributes `export`, `export_type`, `import`... are all in singular form. Would it make sense to use `nif` instead of `nifs` here?

bjorng:
I have always read `export` and `import` as verbs, not nouns. `export_type` is clearly a verb.

Since `nifs` is a noun, it makes sense. `nif` does not make sense.

michalmuskala:
Is there some compiler optimisations that are prevented by functions possibly being nifs? Would it mean we'd need to declare `-nifs([]).` in every module without NIFs to enable them?

sverker:
@michalmuskala No, `nifs([])` will not be needed. Modules not containing an explicit call to `erlang:load_nif/2` are guaranteed to not load any NIFs.

PR #5486, opened 2021-12-06
Add lists:transpose/1, useful for list of lists transposition

manuel-rubio:
When I was developing the Advent of Code 2021, using Elixir and Erlang, I just realized the rotating of a matrix like this:

```erlang
L = [
    [1, 2, 3],
    [4, 5, 6]
],
```

Could be rotated with `lists:zip/2` if we pass the lists as:

```erlang
L1 = [1, 2, 3],
L2 = [4, 5, 6],
lists:zip(L1, L2).
% [{1, 4}, {2, 5}, {3, 6}]
```

But, what if we have the need to rotate a matrix of arbitrary size? Then I've implemented `lists:zip/1` which is accepting the `L` parameter as a list of lists (matrix) and it's returning the rotated matrix:

```erlang
L = [
    [1, 2, 3],
    [4, 5, 6]
],
lists:zip(L).
% [[1, 4], [2, 5], [3, 6]]
```

And of course, we could use not only a matrix of 3x2 but also 9x7 or other completely different sizes.

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -27,6 +27,7 @@
 	 seq/2, seq/3, sum/1, duplicate/2, min/1, max/1, sublist/2, sublist/3,
 	 delete/2,
 	 unzip/1, unzip3/1, zip/2, zip3/3, zipwith/3, zipwith3/4,
+	 transpose/1,
 	 sort/1, merge/1, merge/2, rmerge/2, merge3/3, rmerge3/3,
 	 usort/1, umerge/1, umerge3/3, umerge/2, rumerge3/3, rumerge/2,
 	 concat/1, flatten/1, flatten/2, flatlength/1,
@@ -390,6 +391,35 @@ delete(_, []) -> [].
 
 zip([X | Xs], [Y | Ys]) -> [{X, Y} | zip(Xs, Ys)];
 zip([], []) -> [].
+
+%% Return [{A0, ..., Z0}, ..., {An, ..., Zn}] for lists [A0, ..., An],
+%% [B0, ..., Bn], ..., [Z0, ..., Zn].
+
+-spec transpose(Lists) -> Lists when
+      Lists :: [[A]],
+      A :: term().
+
+transpose([]) -> [];
+transpose([List|Tail] = Lists) ->
+    ListLen = length(List),
+    case all(fun(ListN) -> ListLen =:= length(ListN) end, Tail) of
+        true -> transpose(Lists, true, []);
+        false -> erlang:error(badarg, Lists)
+    end.
+
+transpose([[] | _], _MustReverse, Acc) -> reverse(Acc);
+transpose(Lists, MustReverse, Acc) ->
+    {Row, NewLists} = transpose_lists(Lists, [], []),
+    NewRow = case MustReverse of
+                 true -> reverse(Row);
+                 false -> Row
+             end,
+    transpose(NewLists, not MustReverse, [NewRow|Acc]).
+
+transpose_lists([[H|T]|List], Row, Acc) ->
+    transpose_lists(List, [H|Row], [T|Acc]);
+transpose_lists([], Row, Acc) ->
+    {Row, Acc}.
 
 %% Return {[X0, X1, ..., Xn], [Y0, Y1, ..., Yn]}, for a list [{X0, Y0},
 %% {X1, Y1}, ..., {Xn, Yn}].
--- a/lib/stdlib/test/lists_SUITE.erl
+++ b/lib/stdlib/test/lists_SUITE.erl
@@ -54,6 +54,7 @@
 	 ufunsort_1/1, ufunsort_stable/1, ufunsort_rand/1,
 	 ufunsort_error/1,
 	 zip_unzip/1, zip_unzip3/1, zipwith/1, zipwith3/1,
+	 transpose/1,
 	 filter_partition/1, 
 	 join/1,
 	 otp_5939/1, otp_6023/1, otp_6606/1, otp_7230/1,
@@ -122,7 +123,8 @@ groups() ->
      {zip, [parallel], [zip_unzip, zip_unzip3, zipwith, zipwith3]},
      {misc, [parallel], [reverse, member, dropwhile, takewhile,
 			 filter_partition, suffix, subtract, join,
-			 hof, droplast, search, enumerate, error_info]}
+			 hof, droplast, search, enumerate, error_info,
+			 transpose]}
     ].
 
 init_per_suite(Config) ->
@@ -2332,6 +2334,27 @@ flatten_2(Config) when is_list(Config) ->
 %% flatten/2 error cases.
 flatten_2_e(Config) when is_list(Config) ->
     ok.
+
+%% Test lists:transpose/1.
+transpose(Config) when is_list(Config) ->
+    [] = lists:transpose([]),
+    [] = lists:transpose([[], [], [], []]),
+    [[a0, b0, c0], [a1, b1, c1], [a2, b2, c2]] =
+        lists:transpose([[a0, a1, a2], [b0, b1, b2], [c0, c1, c2]]),
+    [[a0, b0], [a1, b1], [a2, b2]] =
+        lists:transpose([[a0, a1, a2], [b0, b1, b2]]),
+    _ = [begin
+             Matrix = [[make_ref() || _ <- lists:seq(1, M)] ||
+                          _ <- lists:seq(1, N)],
+             Matrix = lists:transpose(lists:transpose(Matrix))
+         end || N <- lists:seq(1, 10), M <- lists:seq(1, 10)],
+
+    %% Error cases.
+    {'EXIT',{badarg,_}} = (catch lists:transpose([[], [b]])),
+    {'EXIT',{badarg,_}} = (catch lists:transpose([[a], []])),
+    {'EXIT',{badarg,_}} = (catch lists:transpose([[a], [b,c]])),
+    ok.
+
 
 %% Test lists:zip/2, lists:unzip/1.
 zip_unzip(Config) when is_list(Config) ->
--- a/lib/stdlib/doc/src/lists.xml
+++ b/lib/stdlib/doc/src/lists.xml
@@ -1068,6 +1068,23 @@ splitwith(Pred, List) ->
     </func>
 
     <func>
+      <name name="transpose" arity="1" since=""/>
+      <fsummary>Transpose a list of lists.</fsummary>
+      <desc>
+        <p>Transposes a matrix (list of lists) into one list of lists
+          where the first element of each resulting list is taken from the first
+          list and the rest of elements are taken from the first element of the
+          following lists.</p>
+        <p>That is, if we have a matrix of 3x2 and we transpose it, we
+          obtain a matrix of 2x3.</p>
+        <p><em>Example:</em></p>
+        <pre>
+> <input>lists:transpose([[1,2,3], [4,5,6]]).</input>
+[[1,4],[2,5],[3,6]]</pre>
+      </desc>
+    </func>
+
+    <func>
       <name name="zip" arity="2" since=""/>
       <fsummary>Zip two lists into a list of two-tuples.</fsummary>
       <desc>
@@ -1075,6 +1092,9 @@ splitwith(Pred, List) ->
           where the first element of each tuple is taken from the first
           list and the second element is taken from the corresponding
           element in the second list.</p>
+        <p>If you need to transpose a list of lists instead of "zips"
+          together two lists, see
+          <seemfa marker="#transpose/1"><c>transpose/1</c></seemfa>.</p>
       </desc>
     </func>
 
@@ -1087,6 +1107,9 @@ splitwith(Pred, List) ->
           from the first list, the second element is taken from
           the corresponding element in the second list, and the third
           element is taken from the corresponding element in the third list.</p>
+        <p>If you need to transpose a list of lists instead of "zips"
+          together three lists, see
+          <seemfa marker="#transpose/1"><c>transpose/1</c></seemfa>.</p>
       </desc>
     </func>
 


attah:
Should this maybe be under another name since the output is not tuples? (and cannot be unzipped again).

manuel-rubio:
> Should this maybe be under another name since the output is not tuples? (and cannot be unzipped again).

You're right, I've changed the name of the function.

lauramcastro:
Good one, Manuel!

> You're right, I've changed the name of the function.

If I may make a suggestion I'd use transposition instead of rotatio.n

manuel-rubio:
> Good one, Manuel!
> 
> > You're right, I've changed the name of the function.
> 
> If I may make a suggestion I'd use transposition instead of rotation.

could be better "transposition" or "transpose"?

lauramcastro:
> could be better "transposition" or "transpose"?

The second is more coherent with the names of other functions in the module!

bjorng (commented):
Thanks for your pull request.

I have written some suggestions for improvements. Whether we will include this function in OTP or not will be decided by the OTP Technical Board.

Regarding the naming, I think that best of the suggested names is `transpose`. The name `rotate` suggests that it should take an angle argument to tell it how much to rotate.

lib/stdlib/src/lists.erl
```diff
@@ -391,6 +392,26 @@ delete(_, []) -> [].
...
+      Lists :: [[A]],
+      A :: term().
+
+rotate(Lists) -> rotate(Lists, []).
+
+rotate([[] | _], Acc) -> lists:reverse(Acc);
+
+rotate(Lists, Acc) ->
+    {Row, NewLists} = rotate_lists(Lists, {[], []}),
+    rotate(lists:reverse(NewLists), [lists:reverse(Row)|Acc]).
```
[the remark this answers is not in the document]
manuel-rubio: @bjorng I'm not getting exactly what you propose

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -2333,6 +2334,14 @@ flatten_2(Config) when is_list(Config) ->
 flatten_2_e(Config) when is_list(Config) ->
     ok.
 
+%% Test lists:rotate/1.
+rotate(Config) when is_list(Config) ->
```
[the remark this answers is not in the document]
manuel-rubio: I've added new test and error tests.

HansN:
Just a thought.
If I see lists:rotate or lists:transpose I have no idea what it is about.  Matrix:transpose however gives me an immediate understanding.
Also, probably later someone will propose lists:inverse, lists:add, lists:determinant or any of all the matrix operations which have no sense for general lists.

Wouldn't it be better to start a new stdlib module called matrix?

manuel-rubio:
> Just a thought. If I see lists:rotate or lists:transpose I have no idea what it is about. Matrix:transpose however gives me an immediate understanding. Also, probably later someone will propose lists:inverse, lists:add, lists:determinant or any of all the matrix operations which have no sense for general lists.
> 
> Wouldn't it be better to start a new stdlib module called matrix?

I think it could be a great initiative and we should do it. At the moment I guess the main point could be to determine which functions could be provided for matrix handling. For me, it is clear transposition is a good start, but creating the module only providing only one function is very poor. However, there is no rush.

I guess we could create:

- `matrices:add/2`
- `matrices:subtract/2`
- `matrices:divide/2`
- `matrices:multiply/2`
- `matrices:dot/2` (as the product of two matrices)
- `matrices:sqrt/2` (square root of each value inside of the matrix)
- `matrices:sum/1`
- `matrices:sum_axis_y/1`
- `matrices:sum_axis_x/1`
- `matrices:inverse/1`
- `matrices:transpose/1`

And the definition of the type:

```
-type matrix() :: [[number()]].
```

However, the transpose could be applied to data which isn't need to be a number. As shown in the example for the documentation, I'm using atoms there. Therefore, it could exist a transpose in general terms for a list (no matter the content it has) and one specific for matrices.

What do you think?

bjorng:
> Wouldn't it be better to start a new stdlib module called matrix?

I think that `transpose` can be useful on lists of lists that are not matrices. It is a generalization of `zip` and `zip3`.

I am not sure that we ever want to have a `matrix` module in OTP. If we would want one, we might want a different representation than lists of lists, perhaps even different possible representations depending on the nature of the matrix.

lib/stdlib/src/lists.erl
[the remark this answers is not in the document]
bjorng: My idea was to minimize the number of calls to `reverse/1`, but it turned out to be a little bit trickier than I naively thought it would be. If I'll get it to work and it's not too messy, I'll post the code.

lib/stdlib/src/lists.erl
```diff
@@ -391,6 +392,25 @@ delete(_, []) -> [].
...
+-spec transpose(Lists) -> Lists when
+      Lists :: [[A]],
+      A :: term().
+
+transpose(Lists) -> transpose(Lists, []).
+
+transpose([[] | _], Acc) -> reverse(Acc);
+transpose(Lists, Acc) ->
+    {Row, NewLists} = transpose_lists(Lists, [], []),
+    transpose(reverse(NewLists), [reverse(Row)|Acc]).
```
bjorng, line 407: Here is how many calls to `reverse/1` can be avoided:

```suggestion
transpose(Lists) -> transpose(Lists, true, []).

transpose([[] | _], _MustReverse, Acc) ->
    reverse(Acc);
transpose(Lists, MustReverse, Acc) ->
    {Row, NewLists} = transpose_lists(Lists, [], []),
    NewRow = case MustReverse of
                 true -> reverse(Row);
                 false -> Row
             end,
    transpose(NewLists, not MustReverse, [NewRow|Acc]).
```

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -2333,6 +2334,21 @@ flatten_2(Config) when is_list(Config) ->
...
     ok.
 
+%% Test lists:transpose/1.
+transpose(Config) when is_list(Config) ->
+    [] = lists:transpose([[], [], [], []]),
+    [[a0, b0, c0], [a1, b1, c1], [a2, b2, c2]] =
+        lists:transpose([[a0, a1, a2], [b0, b1, b2], [c0, c1, c2]]),
+    [[a0, b0], [a1, b1], [a2, b2]] =
+        lists:transpose([[a0, a1, a2], [b0, b1, b2]]),
+
```
bjorng, line 2344: To thoroughly test the new implementation, here is an exhaustive test of all matrices from 1x1 up to 10x10:

```suggestion
    _ = [begin
             Matrix = [[make_ref() || _ <- lists:seq(1, M)] ||
                          _ <- lists:seq(1, N)],
             Matrix = lists:transpose(lists:transpose(Matrix))
         end || N <- lists:seq(1, 10), M <- lists:seq(1, 10)],
 
```

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -2333,6 +2334,21 @@ flatten_2(Config) when is_list(Config) ->
...
+    %% Error cases.
+    {'EXIT',{function_clause,_}} = (catch lists:transpose([], [b])),
+    {'EXIT',{function_clause,_}} = (catch lists:transpose([a], [])),
+    {'EXIT',{function_clause,_}} = (catch lists:transpose([a], [b,c])),
```
bjorng, line 2348: These test cases fail. Did you actually run the test suite?

manuel-rubio: @bjorng  I don't know why the tests cases take ages on my laptop and they are failing in different parts :-/ I was checking with Actions here... and looks like all tests are passing, right?

bjorng: The Gihub actions only run a very small subset of test cases, essentially a smoke test.

Have you looked at https://github.com/erlang/otp/wiki/Running-tests and how you can run individual tests? So after the initial building, you can do:

```
1> ts:run(stdlib, lists_SUITE, [batch]).
```

or even:

```
1> ts:run(stdlib, lists_SUITE, transpose, [batch]).
```

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -119,7 +120,7 @@ groups() ->
      {flatten, [parallel],
       [flatten_1, flatten_2, flatten_1_e, flatten_2_e]},
      {tickets, [parallel], [otp_5939, otp_6023, otp_6606, otp_7230]},
-     {zip, [parallel], [zip_unzip, zip_unzip3, zipwith, zipwith3]},
+     {zip, [parallel], [zip, zip_unzip, zip_unzip3, zipwith, zipwith3]},
```
bjorng, line 123: You forgot to remove `zip` when you changed name:

```suggestion
     {zip, [parallel], [zip_unzip, zip_unzip3, zipwith, zipwith3]},
```

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -119,7 +120,7 @@ groups() ->
...
      {misc, [parallel], [reverse, member, dropwhile, takewhile,
 			 filter_partition, suffix, subtract, join,
 			 hof, droplast, search, enumerate, error_info]}
```
bjorng, line 126:
```suggestion
			 hof, droplast, search, enumerate, error_info,
			 transpose]}
```

lib/stdlib/test/lists_SUITE.erl
manuel-rubio: done :-)

lib/stdlib/test/lists_SUITE.erl
```diff
@@ -122,7 +123,8 @@ groups() ->
      {zip, [parallel], [zip_unzip, zip_unzip3, zipwith, zipwith3]},
      {misc, [parallel], [reverse, member, dropwhile, takewhile,
 			 filter_partition, suffix, subtract, join,
-			 hof, droplast, search, enumerate, error_info]}
+			 hof, droplast, search, enumerate, error_info,
+             transpose]}
```
bjorng, line 127: Indentation is off.

```suggestion
                         transpose]}
```

manuel-rubio: Done.

manuel-rubio:
> Before introducing a new function like transpose there is a need to look over the already existing functions. In particular the functions `zip/2, zip3/3, zipwith/3, zipwith3/4, unzip/1, unzip3/1`.

I can understand `lists:zip/2` and `lists:zip3/3` makes sense in the terms of the return `[{any(), any()}]` or `[{any(), any(), any()}]` respectively are good to be obtained, but `lists:transpose/1` is working over `[[any()]]` which is different from the zip/unzip functions.

> The name `lists:transpose` is not obviously a good one (depending how we do with the zip/unzip functions). An alternative could be to introduce a `zipwithn(Combine,ListOfLists) Combine = fun((Elements) -> T)`, the `fun(Elements) -> Elements end`, would give the same result as `lists:transpose/1`. Another possibility would be to define the Combine function in `zipwithn` as `fun(Tuple) -> T end` where `Tuple` is a tuple of the same size as `length(ListOfLists)`.

We were on that previously and finally, we thought it's not a good idea because `unzip` is the opposite operation of `zip`.

Using what you suggest, and assuming `Elements` is a list `[any()]` containing the first elements for each list contained in the main list:

```erlang
Combine = fun(Elements) -> Elements end,
L = lists:zipwithn(Combine, lists:zipwithn(Combine, L)).
```

It sounds to me a bit difficult because we had to specify `zip`, an extra function (the `with` part) to get what we want and even the combination function (and even use the extra `n` to specify a different input in comparison with `lists:zipwith/3`) isn't clear enough, and finally, use it with the list. Not very clean. On the other hand:

```erlang
L = lists:transpose(lists:transpose(L)).
```

It's clear enough about what do you expect because the transposition of a list of lists it's a mathematical concept. In my opinion is clear enough and it doesn't need explaining.

Even, if you ask me, in my opinion `lists:zip3/3` and `lists:zipwith3/4` is a workaround. I guess in someplace it could be used and it's useful, but if you develop directly `lists:zip/1` using a list of lists, it's orthogonal including both:

```erlang
L1 = lists:seq(10, 100, 10),
L2 = lists:seq(15, 105, 10),
L3 = lists:seq(13, 103, 10),
lists:zip([L1, L2]),
lists:zip([L1, L2, L3]),
lists:zip([L1, L2, L1, L3]).
```

The only negative point is the definition of the lists isn't as clear as the tuples, but in my opinion (and keeping in mind we are working in a module called `lists`), makes sense to return a list of lists and make it possible to do:

```erlang
[L1, L2, L1, L3] = lists:zip(lists:zip([L1, L2, L1, L3])).
```

This way, the use of `lists:unzip/1` isn't needed.

lib/stdlib/src/lists.erl
```diff
@@ -391,6 +392,35 @@ delete(_, []) -> [].
...
+%% [B0, ..., Bn], ..., [Z0, ..., Zn].
+
+-spec transpose(Lists) -> Lists when
+      Lists :: [[A]],
+      A :: term().
+
+transpose([]) -> [];
+transpose([List|_] = Lists) ->
+    ListLen = length(List),
+    case all(fun(ListN) -> ListLen =:= length(ListN) end, Lists) of
```
seriyps, line 405: I think there is no need to calculte and check the length of the 1st list twice?

```suggestion
transpose([List|Tail] = Lists) ->
    ListLen = length(List),
    case all(fun(ListN) -> ListLen =:= length(ListN) end, Tail) of
```

Also, can this check be skipped entirely and error detected at some later stage? `length(ListN)` is O(n), so, might be noticeable on long lists.

manuel-rubio: Added. Thanks @seriyps

bjorng:
After further consideration, the OTP Technical Board decided to reject this proposed extension of the `lists` module. See [this post](https://erlangforums.com/t/otb-decision-on-four-proposed-extensions-to-the-lists-module/1146?u=bjorng) in the Erlang Forums for the motivation.

PR #5492, opened 2021-12-07
Improve error messages for lists:seq/3

nixxquality:
Right now, lists:seq performs as documented if you provide bad input.

> Failures:
> - If To < From - Incr and Incr > 0.
> - If To > From - Incr and Incr < 0.
> - If Incr =:= 0 and From =/= To.

However, the current error message is really confusing when it can be a lot better with just a little bit of rejigging.

```
% currently:
> lists:seq(1,5,-1).
** exception error: no true branch found when evaluating an if expression
     in function  lists:seq/3 (lists.erl, line 264)
% with this patch:
> seq:seq(1,5,-1).   
** exception error: no function clause matching seq:seq(1,5,-1) (seq.erl, line 32)
```

This error brings your attention to the incorrect input you've provided rather than pointing to some possible internal misbehavings.

The code still behaves as it should with all inputs I quickly tested.
```
14> seq:seq(1,5,0).       
** exception error: no function clause matching seq:seq(1,5,0) (seq.erl, line 32)
15> seq:seq(1,5,1). 
[1,2,3,4,5]
16> seq:seq(1,5,-1). 
** exception error: no function clause matching seq:seq(1,5,-1) (seq.erl, line 32)
17> seq:seq(5,1,-1). 
[5,4,3,2,1]
18> seq:seq(5,1,1).  
** exception error: no function clause matching seq:seq(5,1,1) (seq.erl, line 32)
19> seq:seq(1,1,1). 
[1]
21> seq:seq(1,1,0). 
[1]
22> seq:seq(1,2,0).
** exception error: no function clause matching seq:seq(1,2,0) (seq.erl, line 32)
```

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -259,16 +259,16 @@ seq_loop(0, _, L) ->
       Incr :: integer(),
       Seq :: [integer()].
 
-seq(First, Last, Inc) 
-    when is_integer(First), is_integer(Last), is_integer(Inc) -> 
-    if
-        Inc > 0, First - Inc =< Last;
-        Inc < 0, First - Inc >= Last ->
-            N = (Last - First + Inc) div Inc,
-            seq_loop(N, Inc*(N-1)+First, Inc, []);
-        Inc =:= 0, First =:= Last ->
-            seq_loop(1, First, Inc, [])
-    end.
+seq(First, Last, Inc)
+    when is_integer(First), is_integer(Last), is_integer(Inc),
+        (Inc > 0 andalso First - Inc =< Last) orelse
+        (Inc < 0 andalso First - Inc >= Last) ->
+    N = (Last - First + Inc) div Inc,
+    seq_loop(N, Inc * (N - 1) + First, Inc, []);
+seq(Same, Same, 0) when is_integer(Same) ->
+    [Same];
+seq(First, Last, Inc) ->
+    erlang:error(badarg, [First, Last, Inc], [{error_info, #{module => erl_stdlib_errors}}]).
 
 seq_loop(N, X, D, L) when N >= 4 ->
      Y = X-D, Z = Y-D, W = Z-D,
--- a/lib/stdlib/src/erl_stdlib_errors.erl
+++ b/lib/stdlib/src/erl_stdlib_errors.erl
@@ -189,7 +189,20 @@ format_lists_error(keysearch, Args) ->
 format_lists_error(member, [_Key, List]) ->
     [[], must_be_list(List)];
 format_lists_error(reverse, [List, _Acc]) ->
-    [must_be_list(List)].
+    [must_be_list(List)];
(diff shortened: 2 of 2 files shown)


jhogberg:
Thanks for the PR, but we think it would be better to follow EEP 54 (https://github.com/erlang/otp/pull/2849) than giving people a slightly more useful `function_clause` error. Do you think you can do that?

nixxquality:
I wasn't aware that existed until now.
Is something like this good?

I honestly don't understand the purpose of format_error (it didn't seem to affect anything in my version of erlang) so I just left a basic one there.

jhogberg:
> I wasn't aware that existed until now.
> Is something like this good?

It's better, but it's a bit wasteful to build that information in `lists:seq/3` itself since it'll be generated on every error, and we don't need it until someone wants the extended error information.

The `format_error/2` function knows which function crashed and what its arguments were, so you can figure out what exactly went wrong in there instead.

> I honestly don't understand the purpose of format_error (it didn't seem to affect anything in my version of erlang) so I just left a basic one there.

The shell (or any other) calls it to get better error information. It should be visible when you provoke the error in the shell.

lib/stdlib/src/lists.erl
```diff
@@ -260,15 +260,19 @@ seq_loop(0, _, L) ->
...
+        Inc =/= 0 ->
             N = (Last - First + Inc) div Inc,
             seq_loop(N, Inc*(N-1)+First, Inc, []);
-        Inc =:= 0, First =:= Last ->
+        Inc =:= 0 ->
             seq_loop(1, First, Inc, [])
-    end.
+    end;
+seq(First, Last, Inc) ->
+    erlang:error(badarg, [First, Last, Inc], [{error_info, #{module => erl_stdlib_errors}}]).
```
jhogberg, line 275: After checking the disassembly, it seems the compiler doesn't manage to optimize the `if` very well. Can you hoist it up to a function clause instead? e.g.

```erlang
seq(First, Last, Inc)
    when is_integer(First), is_integer(Last), is_integer(Inc),
        (Inc > 0 andalso First - Inc =< Last) orelse
        (Inc < 0 andalso First - Inc >= Last) ->
    N = (Last - First + Inc) div Inc,
    seq_loop(N, Inc * (N - 1) + First, Inc, []);
seq(Same, Same, 0) when is_integer(Same) ->
    [Same];
seq(First, Last, Inc) ->
    erlang:error(badarg, [First, Last, Inc],
                 [{error_info, #{module => erl_stdlib_errors}}]).
```

jhogberg (commented):
Thanks, the tests ran fine and we're almost ready to merge it. Can you squash the PR down to one commit?

nixxquality:
There, I squashed it and fixed that random missing newline while I was at it.
Apologies about the delay, I'm not able to checkout the repository on Windows so I had to jump through some hoops.

jhogberg:
Merged, thanks for the PR! :)

PR #5493, opened 2021-12-07
Remove undocumented {value, Value, Ann} in erl_eval

josevalim:
The shell relies on an implicit contract
with erl_eval where a value triplet is allowed.
Unfortunately, to support this, evaluating
anonymous functions would require a look-ahead
traversal to hide any value triplet from
erl_lint and this affected performance negatively.

This patch removes this undocumented contract
and removes the look-ahead. To address this,
we took different approaches:

  1. v(N) was expanded to a value before evaluated.
     Now, we process and validate the arguments,
     but communicate with the shell process to get
     the value, similar to how history() works

  2. References, Ports, and PIDs were converted to
     variables and then replaced to values. Now,
     we expand `#PID<0.13.0>` into an expression
     such as `erlang:list_to_pid("#PID<0.13.0>").`

  3. Finally, the prompt evaluation would use values,
     but it only had to pass `[{history,integer()}]`,
     which can be manually converted to AST

With this change, evaluating anonymous functions in
a loop gets considerably faster, up to 10% in a
macro benchmark that is doing much more than just
dispatch functions.

### DIFF
--- a/lib/debugger/src/dbg_icmd.erl
+++ b/lib/debugger/src/dbg_icmd.erl
@@ -382,19 +382,14 @@ eval_restricted({From,_Mod,Cmd,SP}, Bs) ->
     case catch parse_cmd(Cmd, 1) of
 	{'EXIT', _Reason} ->
 	    From ! {self(), {eval_rsp, 'Parse error'}};
-	{[{var,_,Var}], XBs} ->
+	[{var,_,Var}] ->
 	    Bs2 = bindings(Bs, SP),
 	    Res = case get_binding(Var, Bs2) of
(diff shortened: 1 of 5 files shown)


josevalim:
There is currently some dead code in the `eval` module. It still handles `{value, ..., ...}` after this PR and it still handles tuple calls (aka `({Mod, Fun})(Arg)`). I decided to not remove them for now, but it could be done in a future chap.

jhogberg:
`shell_SUITE:otp_14296/1` fails with this PR, can you look into it?

jhogberg:
Merged, thanks for the PR!

PR #5495, opened 2021-12-07
Optimize erl_lint:is_guard_test/2-3

josevalim:
Only compute erl_lint:start() if there are records.


josevalim:
Note that `is_guard_test/1` also bypasses `start()`. I thought about changing `is_guard_test/1` to call `is_guard_test/3`, but I noticed that `is_guard_test/1` does not call `set_file(Expression, "nofile")`, and it is unclear if it should call or not. For this reason, I decided to not change `is_guard_test/1`.

jhogberg:
Merged, thanks for the PR!

PR #5496, opened 2021-12-07
Traverse eval forms for function used vars once

josevalim:
Prior to this implementation, every time an
anonymous function was defined via eval, we
would traverse its tree to find its used vars.

However, if there are nested anonymous functions,
they are traversed once by their parent and then
traversed again each time they are executed.
This means any form of nested loops are extremely
expensive in the shell.

With this patch, we traverse the tree once and
then we keep the tree of used vars for all of its
anonymous functions inside. This drastically speeds
up those scenarios.

Take this Elixir code:

    lines
    |> Enum.reduce(%{}, fn
      [x, y1, x, y2], grid ->
        Enum.reduce(y1..y2, grid, fn y, grid ->
          Map.update(grid, {x, y}, 1, &(&1 + 1))
        end)

      [x1, y, x2, y], grid ->
        Enum.reduce(x1..x2, grid, fn x, grid ->
          Map.update(grid, {x, y}, 1, &(&1 + 1))
        end)

      [x1, y1, x2, y2], grid ->
        Enum.reduce(Enum.zip(x1..x2, y1..y2), grid, fn point, grid ->
          Map.update(grid, point, 1, &(&1 + 1))
        end)
    end)
    |> Enum.count(fn {_, v} -> v > 1 end)

In the code above, there are three levels of anonymous function,
which then become very expensive to traverse on every iteration.
With this patch, the time to evaluate the code above goes from
2s to 0.6s.


lib/stdlib/src/erl_eval.erl
```diff
@@ -503,6 +511,28 @@ find_maxline(LC) ->
...
+       case Fuv of
+          %% If this clause is in our fun used vars tree,
+          %% then we do not need to compute to traverse it again.
+           #{Cs := UsedAndFuv} ->
+               UsedAndFuv;
+
+           #{} ->
+                %% Save only used variables in the function environment.
+                %% {value,L,V} are hidden while lint finds used variables.
+                {Fun1, _} = hide_calls(Fun, 0),
```
josevalim, line 525: Because we still have to hide calls, if the `Fun` has `{value, _, _}` in it, we won't be able to lookup its clauses. If we merge #5493, then we will improve the "cache hit".

jhogberg (commented):
Thanks for the PR! I've left a few comments inline.

josevalim:
Thank you! I will wait for the other PR, then rebase and apply your suggestion.

I also want to send another PR that gives control over how eval exceptions are raised. If so, I will convert all of those arguments into a record and give them proper names. :D

I guess it may be worth using a record even if we don't add a new argument, right after this PR is merged?

lib/stdlib/src/erl_eval.erl
```diff
@@ -88,8 +90,8 @@
       NewBindings :: binding_struct()).
 exprs(Exprs, Bs) ->
     case check_command(Exprs, Bs) of
-        ok -> 
-            exprs(Exprs, Bs, none, none, none);
+        ok ->
+            exprs(Exprs, Bs, none, none, none, fuv());
```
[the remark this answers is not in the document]
josevalim: I went with `empty_fun_used_vars()` because I have been bitten in the past when I had to change the default value of something and that led to me chasing which of all empty maps in a file were the empty maps I was looking for. :D But if you prefer `#{}`, then I will gladly change it.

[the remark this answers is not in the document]
jhogberg: Fine with me :)

josevalim:
Rebased and all feedback addressed! 💯

lib/stdlib/src/erl_eval.erl
```diff
@@ -496,6 +523,26 @@ find_maxline(LC) ->
     _ = erl_parse:map_anno(F, LC),
     erase('$erl_eval_max_line').
 
+fun_used_bindings(Fun, Cs, Bs, FUVs) ->
+    {Used,InnerFUVs} =
+	case FUVs of
+	   %% If this clause is in our fun used vars tree,
```
jhogberg, line 529: Nitpick: this comment is a bit misaligned, and there's a mix of tabs and spaces throughout the function.

jhogberg (commented):
Thanks, I've added it to our nightly builds.

lib/stdlib/src/erl_eval.erl
josevalim: Fixed. I keep forgetting OTP is no longer doing emacs mode indentation!

jhogberg:
Sorry, it collided with https://github.com/erlang/otp/pull/5411 so I had to take it out of the nightly runs. I'll resolve the conflict on Monday.

jhogberg:
`ms_transform_SUITE:from_shell/1`, `shell_SUITE:otp_7232/1`, and `shell_SUITE:otp_6554/1` fail with this PR. :-/

josevalim:
@jhogberg apparently I forgot to change `erl_eval:fun_data/1`. All of those tests should pass now.

jhogberg:

itssundeep:
@josevalim This is breaking maps:fold nested in lists:foldl in an interesting way, infact for example in file
A.erl
```
func1() ->
  maps:fold(
      fun(A, B, C) ->
          // something
      end,
      [],
      SomeMap
   )
   
 func2() ->
     lists:foldl(
        fun(A, B) ->
            map:fold(
                fun(K, V, Acc) ->
                   // something
                end,
                B,
                SomeMap
           ),
       end,
       Acc0,
       SomeList
     )
    ```
   
During the call to func2() , we are somehow calling map:fold in func1(). Reverting this change doesn't reproduce the issue, so I am thinking somehow this change is triggering this.

**Comment by josevalim** [comment]
@itssundeep interesting. Can you please provide complete steps for reproduction? Like:

1. create this file
2. compile the file
3. start `erl` and call functions with these arguments

Thanks!

**Comment by itssundeep** [comment]
I looked into this a little more and this change breaks [power_shell](https://github.com/WhatsApp/power_shell) code and will be handled in power_shell, so this change is good.

File: lib/stdlib/src/erl_lint.erl [hunk 4]
```diff
@@ -529,10 +534,9 @@ used_vars(Exprs, BindingsList) ->
 		  ({V,_Val}, Vs0) -> [{V,{bound,unused,[]}} | Vs0]
 	       end, [], BindingsList),
     Vt = orddict:from_list(Vs),
-    {Evt,_St} = exprs(set_file(Exprs, "nofile"), Vt, start()),
-    {ok, foldl(fun({V,{_,used,_}}, L) -> [V | L];
-                  (_, L) -> L
-	       end, [], Evt)}.
+    St0 = (start())#lint{fun_used_vars=maps:new()},
+    {_Evt,St1} = exprs(Exprs, Vt, St0),
+    St1#lint.fun_used_vars.
```
max-au, line 539: This changed the return type spec from `{ok, map()}` to just `map()`, hence some (hacky, just like `power_shell`) software started to fail.

josevalim: I see. It was private API but I can figure something out if the Erlang/OTP team wants to bring it back.

max-au: It is private API and my laziness, I forked `erl_eval` instead of making a small upstream fix (which I will do now).
I already pushed a fixed power_shell, so unless there are other usages of this function in the wild, I'd rather keep this change as is.

PR #5498, opened 2021-12-08
dialyzer: Track type & behaviour dependencies

TD5:
Dialyzer now tracks the dependencies between modules induced by
type-level information, such as type definitions & specs, as well
as behaviour callbacks.

Previously, only function calls were considered when computing dependencies.
This was incomplete because changes to types and behaviours can also affect
the result of a Dialyzer analysis.

Aside from the immediate impact of this change, it also serves as a prelude to
adding full incrementality to Dialyzer.

### DIFF
--- a/lib/dialyzer/src/dialyzer_analysis_callgraph.erl
+++ b/lib/dialyzer/src/dialyzer_analysis_callgraph.erl
@@ -131,7 +131,7 @@ analysis_start(Parent, Analysis, LegalWarnings) ->
                           solvers = Analysis#analysis.solvers
 			 },
   Files = ordsets:from_list(Analysis#analysis.files),
-  {Callgraph, TmpCServer0} = compile_and_store(Files, State),
+  {Callgraph, ModCallDeps, Modules, TmpCServer0} = compile_and_store(Files, State),
   %% Remote type postprocessing
   Args = {Plt, Analysis, Parent},
   NewCServer = remote_type_postprocessing(TmpCServer0, Args),
@@ -145,6 +145,9 @@ analysis_start(Parent, Analysis, LegalWarnings) ->
   NonExports = sets:subtract(sets:from_list(AllNodes, [{version, 2}]), Exports),
   NonExportsList = sets:to_list(NonExports),
   State2 = analyze_callgraph(Callgraph, State1),
+  ModTypeDeps = dict:from_list(maps:to_list(dialyzer_typegraph:module_type_deps(Analysis#analysis.use_contracts, CServer, Modules))),
+  ModDeps = dialyzer_callgraph:merge_module_deps(ModCallDeps, ModTypeDeps),
+  send_mod_deps(Parent, ModDeps),
   #analysis_state{plt = Plt2,
                   doc_plt = DocPlt,
                   codeserver = Codeserver0} = State2,
@@ -227,6 +230,7 @@ analyze_callgraph(Callgraph, #analysis_state{codeserver = Codeserver,
 					     doc_plt = DocPlt,
                                              plt = Plt,
 					     timing_server = TimingServer,
+					     parent = Parent,
                                              solvers = Solvers} = State) ->
   case State#analysis_state.analysis_type of
     plt_build ->
@@ -238,10 +242,10 @@ analyze_callgraph(Callgraph, #analysis_state{codeserver = Codeserver,
     succ_typings ->
       {Warnings, NewPlt, NewDocPlt} =
         dialyzer_succ_typings:get_warnings(Callgraph, Plt, DocPlt, Codeserver,
-                                           TimingServer, Solvers),
+                                          TimingServer, Solvers),
       dialyzer_callgraph:delete(Callgraph),
       Warnings1 = filter_warnings(Warnings, Codeserver),
-      send_warnings(State#analysis_state.parent, Warnings1),
+      send_warnings(Parent, Warnings1),
       State#analysis_state{plt = NewPlt, doc_plt = NewDocPlt}
     end.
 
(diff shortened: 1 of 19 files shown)


lib/dialyzer/src/dialyzer_typegraph.erl
```diff
@@ -0,0 +1,105 @@
...
+    ModulesMentionedInContracts ++
+    ModulesMentionedInTypeDefinitions ++
+    ModulesMentionedAsBehaviours,
+
+  %% Find the union of module dependencies from all sources, removing self-references
+  AllDeps =
+    lists:foldl(
+      fun ({M,Deps}, Acc) ->
+        NewDepsSetWithoutSelf = gb_sets:delete_any(M, Deps),
+        maps:update_with(
```
jhogberg, line 83: You're dealing with a set of sets, why not use `sofs` instead? Both this part and transposition would become one-liners, check out `sofs:family_union/2` and `sofs:converse/1`.

lib/dialyzer/src/erl_types.erl
```diff
@@ -5705,3 +5707,128 @@ handle_base(Unit, Neg) ->
...
+%%
+%% Utilities for finding a module's type dependencies
+%%
+%%=============================================================================
+
+
+-spec module_type_deps_of_type_defs(type_table()) -> gb_sets:set(module()).
+
+module_type_deps_of_type_defs(TypeTable) ->
+  ModuleTypeDependencies = [module_type_deps_of_entry(TypeTableEntry) || TypeTableEntry <- maps:to_list(TypeTable)],
```
jhogberg, line 5721: Nitpick: this line is too long. Try keeping lines below 80 characters.

jhogberg (changes requested):
Thanks for the PR! It looks good overall, and I've made a few comments inline.

lib/dialyzer/src/dialyzer_callgraph.erl
```diff
@@ -314,6 +317,17 @@ module_deps(#callgraph{digraph = DG}) ->
   digraph_delete(MDG),
   dict:from_list(Deps).
 
+-spec merge_module_deps(mod_deps(), mod_deps()) -> mod_deps().
+merge_module_deps(Left, Right) ->
+    dict:merge(
+      fun (_Mod, L, R) ->
+        gb_sets:to_list(gb_sets:union(
```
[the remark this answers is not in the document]
TD5: Similarly to above, the answer is probably "yes", but I tried to stay consistent with existing types where I could.

lib/dialyzer/src/dialyzer_typegraph.erl
TD5: I haven't really considered `sofs` - I'll take a look.

lib/dialyzer/src/dialyzer_typegraph.erl
```diff
@@ -0,0 +1,105 @@
...
+
+  AllDepsRaw =
+    ModulesMentionedInContracts ++
+    ModulesMentionedInTypeDefinitions ++
+    ModulesMentionedAsBehaviours,
+
+  %% Find the union of module dependencies from all sources, removing self-references
+  AllDeps =
+    lists:foldl(
+      fun ({M,Deps}, Acc) ->
```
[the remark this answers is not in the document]
TD5: If I recall correctly, this was purely due to my subjective opinion that, of the things that came to mind, the original was simplest to understand. I'll take another look at this and see whether this could be reformulated.

lib/dialyzer/src/dialyzer_analysis_callgraph.erl
```diff
@@ -151,6 +151,9 @@ analysis_start(Parent, Analysis, LegalWarnings) ->
       false -> Callgraph
     end,
   State2 = analyze_callgraph(NewCallgraph, State1),
+  ModTypeDeps = dict:from_list(maps:to_list(dialyzer_typegraph:module_type_deps(Analysis#analysis.use_contracts, CServer, Modules))),
```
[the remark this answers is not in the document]
jhogberg: Fair enough, we can always do it later on.

jhogberg:
The `dialyzer` checks still fail, try running `scripts/run-dialyzer` locally before pushing again.

TD5:
> The `dialyzer` checks still fail, try running `scripts/run-dialyzer` locally before pushing again.

Thanks! I wasn't aware of this script and have been running the analysis for this change with a slightly different set of options. I can repro now, so I'll get the build fixed up.

lib/dialyzer/src/dialyzer_typegraph.erl
```diff
@@ -0,0 +1,105 @@
...
+          NewDepsSetWithoutSelf,
+          Acc)
+      end,
+      maps:new(),
+      AllDepsRaw),
+
+  ModuleToWhatItDependsOn = maps:map(fun (_Mod, Deps) -> gb_sets:to_list(Deps) end, AllDeps),
+
+  ModuleToThoseModulesThatDependOnIt = dialyzer_utils:transpose_map_list(ModuleToWhatItDependsOn),
+
```
bjorng, line 97: The `sofs` module can be a little intimidating, so here is an example how you can use `sofs`:

```suggestion
  AllDepsRaw1 = [{M, gb_sets:to_list(Deps)} || {M, Deps} <- AllDepsRaw],
  S0 = sofs:relation(AllDepsRaw1, [{atom,[atom]}]),
  S1 = sofs:relation_to_family(S0),
  S2 = sofs:family_union(S1),
  S3 = sofs:family_to_relation(S2),
  S4 = sofs:converse(S3),
  S5 = sofs:relation_to_family(S4),
  S6 = [{M, Ds -- [M]} || {M, Ds} <- sofs:to_external(S5)],
  ModuleToThoseModulesThatDependOnIt = maps:from_list(S6),

```

(I leave it to you to avoid creating the `gb_sets` in the first place, so that you can eliminate the list comprehension in first line.)

bjorng (commented):
I have taken a quick look and found a few things.

(The test case `small_SUITE:fun_named/1` fails.) UPDATED: Not it, doesn't. Local problem. I ran test suites from the latest `master`, which had a test case for the infinite loop problem.

TD5:
Okay, I think that's everything I can see resolved now

jhogberg (approved):

jhogberg:
Thanks, I've added it to our nightly test runs. :)

jhogberg:

Issue #4671, opened 2021-03-24
erl_syntax_lib:analyze_attribute/1 is not works as document described.


slepher:
**Describe the bug**
erl_syntax_lib:analyze_wild_attribute/1 is not works as document described:
https://erlang.org/doc/man/erl_syntax_lib.html#analyze_wild_attribute-1
Returns the name and value of a "wild" attribute. The result is the pair {Name, Value}, if Node represents "-Name(Value)".

**To Reproduce**
```erlang
Analyzed = erl_syntax_lib:analyze_attribute({attribute,17,mark,mark_1}).
?assertEqual({mark, mark_1}, Analyzed).
```
it actually returns
```erlang
{mark, {mark, mark_1}}
```

**Affected versions**
OTP-23 and master branch is tested.

PR #5499, opened 2021-12-08
Fixes analysing a wild attribute to return Info

hazbo:
As per #4671:

`erl_syntax_lib:analyze_attribute({attribute,17,mark,mark_1}).`

returns `{mark, {mark, mark_1}}` due to the attribute being wrapped on
line 1320:

`{A, analyze_attribute(A, Node)}`

when what would be expected as per the documentation is:

`{mark, mark1}`

Returning *just* the Info from `{_, Info} = analyze_wild_attribute(Node)`
within `analyze_attribute/2`, this double wrapping does not occur, fixing
the original issue.


hazbo:
Closing for now as this commit is currently causing an issue in building the docs. Apologies for opening this PR while the actions were still being ran on my branch.

PR #5500, opened 2021-12-08
Fix typos in lib/dialyzer

kianmeng:


lib/dialyzer/test/opaque_SUITE_data/src/big_external_type.erl
```diff
@@ -351,7 +351,7 @@
 -define(line(Tup), element(2, Tup)).
```
KennethL, line 351: `.erl` files in the */test/*_SUITE_data/ directories are mostly copies (as is) from other internal and external Erlang applications and there is no point correcting spelling errors here, the files are supposed to be used as is.

KennethL (changes requested):

bjorng:
Thanks!

PR #5502, opened 2021-12-09
dialyzer: Eliminate the `race_conditions` option

bjorng:
The `race_conditions` option has not matured despite having been in
Dialyzer since 2008. Quoting its documentation:

> Notice that the analysis that finds data races performs
  intra-procedural data flow analysis and can sometimes explode in
  time. Enable it at your own risk.

Removing the `race_conditions` option simplifies the Dialyzer code base;
especially the `dialyzer_dataflow` module.


jhogberg (approved):

apoorv-2204:
why did we have it in the first place?
If there is any race condition how do we determine it?

PR #5509, opened 2021-12-12
Fixes analysing a wild attribute to return Info

hazbo:
As per issue #4671:

`erl_syntax_lib:analyze_attribute({attribute,17,mark,mark_1}).`

returns `{mark, {mark, mark_1}}` due to the attribute being wrapped on line 1320:

`{A, analyze_attribute(A, Node)}`

when what would be expected as per the documentation is:

`{mark, mark1}`

Returning *just* the Info from `{_, Info} = analyze_wild_attribute(Node)`
within analyze_attribute/2, this double wrapping does not occur, fixing
the original issue.

This PR is basically a reopening of https://github.com/erlang/otp/pull/5499 (was unable to actually reopen that PR) including a fix that was missed.

### DIFF
--- a/lib/syntax_tools/src/erl_syntax_lib.erl
+++ b/lib/syntax_tools/src/erl_syntax_lib.erl
@@ -1106,7 +1106,7 @@ collect_attribute(file, _, Info) ->
     Info;
 collect_attribute(record, {R, L}, Info) ->
     finfo_add_record(R, L, Info);
-collect_attribute(_, {N, V}, Info) ->
+collect_attribute(N, V, Info) ->
     finfo_add_attribute(N, V, Info).
 
 %% Abstract datatype for collecting module information.
@@ -1335,7 +1335,8 @@ analyze_attribute(record, Node) ->
     analyze_record_attribute(Node);
 analyze_attribute(_, Node) ->
     %% A "wild" attribute (such as e.g. a `compile' directive).
-    analyze_wild_attribute(Node).
+    {_, Info} = analyze_wild_attribute(Node),
+    Info.
 
 
 %% =====================================================================


lib/syntax_tools/src/erl_syntax_lib.erl
```diff
@@ -1106,7 +1106,7 @@ collect_attribute(file, _, Info) ->
     Info;
 collect_attribute(record, {R, L}, Info) ->
     finfo_add_record(R, L, Info);
-collect_attribute(_, {N, V}, Info) ->
+collect_attribute(N, V, Info) ->
```
bjorng, line 1109: Please explain this change.

bjorng (commented):
Your bug fix in `analyze_attribute/2` seems to be correct.

But I don't understand the other change. Please explain it.

hazbo:
Thanks for taking a look at this @bjorng

In my original PR, #5499, we can see [this build failure][1] in the GitHub actions, which is what occurs when the change to `collect_attribute` is not in place. In short, the fix to `analyze_attribute` is a breaking API change and `collect_attribute` is on the receiving end of that change. This failure occurs when building the documentation.

To expand on that a little bit, the error message in the build points to `edoc_extract.erl` line 301 which makes a call to `erl_syntax_lib:analyze_forms/1` which returns `finfo_to_list(lists:foldl(fun collect_form/2, new_finfo(), Forms))`. 

`collect_form` passes the `Node` to `analyze_form/1`, which in the case of the `Node` type being of the type `attribute` makes the call to `analyze_attribute`. The resulting value of `analyze_form(Node)` is matched against ` {attribute, {Name, Data}}`, where the `Name`, `Data` and `Info` are passed to `collect_attribute/3`.

With the change I've made to `analyze_attribute`, the value of `N` should now be passed as the first argument to `collect_attribute/3` rather than it being part of the tuple in the second argument.

[1]: https://github.com/erlang/otp/runs/4462452939?check_suite_focus=true#step:8:6603

bjorng:
Thanks!

Since this bug fix is a breaking change, I think it will be better to include it in OTP 25 (instead of OTP 24.x).

I have changed the base of the pull request to `master` and added it to our daily builds.

bjorng:
Thanks for your pull request.

PR #5511, opened 2021-12-13
socket_SUITE: do not fail on IPv6-only nodes

max-au:
Before this change, socket_SUITE fails most of IPv4 tests
with IPv4 address available only on loopback adapter. With this
patch:
 - IPv4-specific tests will be skipped on IPv6-only nodes
 - non-specific tests will utilise either v4 or v6, depending
   on availability (with v4 preferred)
 - minor cleanup (removed unused suite/doc functions)

### DIFF
--- a/lib/kernel/test/socket_SUITE.erl
+++ b/lib/kernel/test/socket_SUITE.erl
@@ -197,7 +197,7 @@
          api_opt_sock_sndtimeo_udp4/1,
          api_opt_sock_timestamp_udp4/1,
          api_opt_sock_timestamp_tcp4/1,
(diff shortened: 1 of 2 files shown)


bmk:
It looks ok, but there are now merge conflicts. Could you please rebase on the lastest master.

max-au:
Rebased to master.

One more note, I removed the commend stating that `peer` is not able to run nodes on a different host. This isn't true, `peer` is supports starting nodes that way, using `host` parameter. There is an example of doing that over SSH and even an example with docker container (which is technically separate host).

bmk:
Yes, I know that but the changes where not made such (the host value was just ignored).

PR #5512, opened 2021-12-13
inets: fix typo in httpc error

attah:


u3s:
- sorry for holding this sooo long. the change itself looks OK, but I planned to read more surrounding code before merging it and this the reason for delay - maybe I will manage do it next week
- can you change target branch to master? it seems potentially incompatible (although not documented anywhere)

Issue #5437, opened 2021-11-22
otp24 asmjit time-consuming recursion is greater than otp22


leonlee2007:
**Describe the bug**
This is not a bug but otp24 asmjit time-consuming recursion is greater than otp22. it's very incredible！

**To Reproduce**
Call `test_calc/0, test_calc_tail/0` with otp22 and otp24 asmjit to get the following results: 

**`test_calc/0` test data: (output maximum, minimum and average time consumption)**

| function_name | otp_version | compile_version | argument  | min(us) | max(us) | avg(us) |
| ------------- | ----------- | --------------- | --------- | ------- | ------- | ------- |
| calc          | 22          | 7.4.5           | 1,000     | 17      | 122     | 20      |
| calc          | 22          | 7.4.5           | 10,000    | 184     | 887     | 291     |
| calc          | 22          | 7.4.5           | 50,000    | 2,311   | 5,281   | 3,649   |
| calc          | 22          | 7.4.5           | 500,000   | 18,893  | 105,137 | 23,008  |
| calc          | 22          | 7.4.5           | 1,000,000 | 42,922  | 64,701  | 47,986  |
| calc          | 22          | 7.4.5           | 2,000,000 | 86,144  | 134,785 | 95,996  |

| function_name | otp_version | compile_version | argument  | min(us) | max(us) | avg(us) |
| ------------- | ----------- | --------------- | --------- | ------- | ------- | ------- |
| calc          | 24          | 8.0.3           | 1,000     | 17      | 205     | 23      |
| calc          | 24          | 8.0.3           | 10,000    | 196     | 741     | 305     |
| calc          | 24          | 8.0.3           | 50,000    | 2,982   | 4,996   | 3,795   |
| calc          | 24          | 8.0.3           | 500,000   | 118,755 | 170,838 | 132,329 |
| calc          | 24          | 8.0.3           | 1,000,000 | 224,605 | 315,486 | 243,924 |
| calc          | 24          | 8.0.3           | 2,000,000 | 488,299 | 660,719 | 548,853 |

**`test_calc_tail/0` test data: (output maximum, minimum and average time consumption)** 

| function_name | otp_version | compile_version | argument   | min(us) | max(us) | avg(us) |
| ------------- | ----------- | --------------- | ---------- | ------- | ------- | ------- |
| calc_tail     | 22          | 7.4.5           | 1,000      | 10      | 2,114   | 17      |
[...]

PR #5514, opened 2021-12-13
Minor JIT optimizations

jhogberg:
This PR fixes some minor inefficiencies discovered while investigating #5437.

### DIFF
--- a/erts/emulator/beam/jit/arm/instr_float.cpp
+++ b/erts/emulator/beam/jit/arm/instr_float.cpp
@@ -88,42 +88,97 @@ void BeamModuleAssembler::emit_fstore(const ArgVal &Src, const ArgVal &Dst) {
     flush_var(dst);
 }
 
-static double handle_fconv(Eterm src) {
-    if (is_small(src)) {
-        return (double)signed_val(src);
-    } else if (is_float(src)) {
-        double res;
-        GET_DOUBLE(src, *(FloatDef *)&res);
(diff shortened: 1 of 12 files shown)


erts/emulator/beam/jit/arm/instr_float.cpp
```diff
@@ -88,42 +88,97 @@ void BeamModuleAssembler::emit_fstore(const ArgVal &Src, const ArgVal &Dst) {
...
+    Label error = a.newLabel();
+
+    /* Is the source a bignum? */
+    {
+        emit_is_boxed(error, ARG1);
+
+        emit_untag_ptr(TMP1, ARG1);
+        a.ldr(TMP1, arm::Mem(TMP1));
+
+        /* The mask (0b111011) cannot be encoded as an immediate an immediate
```
bjorng, line 102:
```suggestion
        /* The mask (0b111011) cannot be encoded as an immediate
```

bjorng (commented):
Looks good to me, but I missed some information in the commit messages:

* "aarch64: Break out breakpoint trampoline to a shared fragment" - Why? Presumably to reduce the size of the loaded code. Will there be a slight performance penalty when a breakpoint is taken?
* "jit: Optimize float instructions" - Perhaps mention that you have optimized fconv by using type information and by inlining it.

jhogberg:
Thanks, I've updated the commit messages.

PR #5517, opened 2021-12-15
Fix typos in lib/et

kianmeng:


garazdawi:
The application is called `et`, not `etc`. Can you fix the commit comment to be correct?

PR #5519, opened 2021-12-15
stdlib: accept empty list of patterns in binary matching functions

v0idpwn:
This changes the binary_match_compile BIF so that it throws badarg only on empty binaries, and not on empty lists.
This change cascades into all functions that rely on binary matching, such as `binary:replace`. This is coming from discussion on https://github.com/elixir-lang/elixir/issues/11485

The general idea of this PR is that an empty list of patterns will never match, but shouldn't throw an error, only follow the usual flow.

### DIFF
--- a/erts/emulator/beam/erl_bif_binary.c
+++ b/erts/emulator/beam/erl_bif_binary.c
@@ -1039,9 +1039,10 @@ static int do_binary_match_compile(Eterm argument, Eterm *tag, Binary **binp)
 	characters = binary_size(argument);
     }
 
-    if (characters == 0) {
-	goto badarg;
-    }
+	if (!(argument == THE_NON_VALUE || argument == NIL) && characters == 0) {
+	    goto badarg;
+	}
+
     ASSERT(words > 0);
 
     if (words == 1) {
--- a/lib/stdlib/test/binary_module_SUITE.erl
+++ b/lib/stdlib/test/binary_module_SUITE.erl
@@ -154,11 +154,11 @@ badargs(Config) when is_list(Config) ->
 	   binary:bin_to_list([1,2,3])),
 
     nomatch =
+	?MASK_ERROR(binary:match(<<1,2,3>>,[])),
+    nomatch =
 	?MASK_ERROR(binary:match(<<1,2,3>>,<<1>>,[{scope,{0,0}}])),
     badarg =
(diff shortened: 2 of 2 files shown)


erts/emulator/beam/erl_bif_binary.c
```diff
@@ -1037,10 +1037,9 @@ static int do_binary_match_compile(Eterm argument, Eterm *tag, Binary **binp)
 	words = 1;
 	comp_term = argument;
 	characters = binary_size(argument);
-    }
-
-    if (characters == 0) {
-	goto badarg;
+	if (characters == 0) {
+	    goto badarg;
+	}
```
v0idpwn, line 1042: I think I probably need to add another check here to substitute this one

v0idpwn: Yep, some error cases that previously fell there won't anymore. Example: `binary:compile_pattern(abc)`

v0idpwn: Fixed in ammended commit

v0idpwn:
Amended fixing documentation entry

josevalim:
Should we then allow an empty list on `binary:compile_pattern/1`? Because to me that would be the main blocker for this change. Maybe it should just return an empty list, given the compiled pattern is opaque anyway?

jhogberg:
Thanks for the PR!

I'm not sure we want this behavior, why call these functions if you don't want them to do anything?

v0idpwn:
I think this is probably the logically expected behaviour when you accept `[pattern()]`.

Having the binary:* functions working like this would also avoid the need to program defensively in some cases. If you don't think this fits, I think the documentation should mention this corner case.

jhogberg:
> Having the binary:* functions working like this would also avoid the need to program defensively in some cases.

On the other hand, if they don't crash when given garbage, then I need to be paranoid and verify the inputs myself.

Doing nothing when handed no patterns may be an okay thing to do for some applications, but there's a lot of code out there and it's impossible to know the effects of suddenly limping along when passed an empty list of patterns.

> If you don't think this fits, I think the documentation should mention this corner case.

I agree, change the spec so that they accept `[pattern(), ...]` instead.

v0idpwn:
Had forgotten about this, will send a fix soon :)

PR #5521, opened 2021-12-15
[ct_slave] deprecate the module completely

max-au:
There is not a single usage for ct_slave within OTP codebase,
proving that replacement is complete.


bjorng:

bjorng:
Thanks!

victorolinasc:
Sorry for the noise here but that would be a great time to switch the branch master to main. Timing is everything.

PR #5524, opened 2021-12-17
Update Install Location and Registry Keys

Trenly:
This PR should solve two issues - 

1) When installing a newer version, the old version remains, but uninstalling the older version removes the newer version from the start menu. This is solved by always updating in-place and using a single folder

2) This should write the version and the publisher in the uninstall entry

### DIFF
--- a/erts/etc/win32/nsis/erlang20.nsi
+++ b/erts/etc/win32/nsis/erlang20.nsi
@@ -56,9 +56,9 @@ Var STARTMENU_FOLDER
 
 ;Folder selection page
 !if ${WINTYPE} == "win64"
-  	InstallDir "$PROGRAMFILES64\erl-${OTP_RELEASE_VERSION}"
+  	InstallDir "$PROGRAMFILES64\Erlang OTP"
 !else
-  	InstallDir "$PROGRAMFILES\erl-${OTP_RELEASE_VERSION}"
+  	InstallDir "$PROGRAMFILES\Erlang OTP"
 !endif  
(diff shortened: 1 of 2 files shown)


garazdawi:
> I haven't tested it, since I don't have the tools to compile it

It is compiled as part of github actions, so if you enable github actions on your repository you can download the windows installer from there and then test it.

Trenly:
> > I haven't tested it, since I don't have the tools to compile it
> 
> It is compiled as part of github actions, so if you enable github actions on your repository you can download the windows installer from there and then test it.

Ohhhhh, Thank you for the advice!

Trenly:
After checking, the Uninstall entry does reflect the version and publisher as expected, and upgrading updates the single uninstall entry
image

I've also confirmed that installing an update does overwrite the files to perform the in-place upgrade as expected
image

The app also seems to run properly after installation

inikulshin:
@Trenly and what happens now after installing a newer version without uninstalling the old one(s)?

Information about multiple installed Erlang version is lost, because the uninstall registry in `SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Erlang OTP` is overwritten and thus contains information only about latest installed Erlang version.

In case of default `INSTDIR` (`c:\Program Files\Erlang OTP`), uninstall of the latest will at least remove all the files from previously installed Erlang versions, but in case of non-default `INSTDIR`, all the files from previously installed Erlang versions will be just lost.

PR #5528, opened 2021-12-17
erts: Fix memory leak in efile_read_handle_info on Windows

jhogberg:
Fixes https://github.com/erlang/otp/issues/5527


lhoguin:
For the record this solves this question: https://erlang.org/pipermail/erlang-questions/2021-September/101523.html

PR #5532, opened 2021-12-18
Support Ed25519/Ed448 SSH host keys in old format

jjcarstens:
If you use a host OS to generate a key via `ssh-keygen` for elliptic curves ed25519 and ed448, they would be generated in OpenSSH format which is currently supported in OTP.

However, if you use another generator, (i.e. `openssl`) or create a private key via Erlang functions (i.e. `public_key:generate_key({namedCurve, ed25519})`), it will format the key with `---- BEGIN EC PRIVATE KEY ----` which is not currently supported.

This PR adds in support for  Ed25519/Ed448 SSH host keys in this old format to allow for key generation and storage via Erlang functions.


HansN:
Thanks for the PR - indeed a good addition to ssh!

PR #5533, opened 2021-12-18
Document opaques - ready for review

### LINKED ISSUE #5240: Support Covariant Opaques (documentation)
@Feature: support covariant opaques (documentation)

I think it's beneficial if the community has a shared understanding of what `-opaque` means. In particular, a variety of tools consume Erlang type annotations, and ideally the tools would be compatible.

I think the language ideally would have an opinion on what the variance of `T` in `-opaque foo(T) :: .....` is, in order for opaques to actually be opaque, based on [the definition in the docs](https://erlang.org/doc/reference_manual/typespec.html)

> "Types declared as opaque represent sets of terms whose structure is not supposed to be visible from outside of their defining module. That is, only the module defining them is allowed to depend on their term structure. "

Currently, any covariant uses of opaques are "depending on [the opaque's] term structure": if the definition changes such that the type is no longer covariant, then users' code breaks.

The best use case for parameterized opaques, imo, is covariant containers like `sets:set/1`, `maps:iterator/1`, `gb_sets:set/1`, etc. In practice all opaques are covariant: I have yet to see a non-covariant opaque. 

- Document that all opaque definitions are assumed to be covariant. In practice, this just means that the docs discourage users from defining opaques where a type parameter appears in function type parameter position: `-opaque bad(T) :: fun((T) -> term()).
- Follow-on issue: support covariant opaques in Dialyzer. Using opaques as if they are covariant works most of the time in Dialyzer, but there are cases where there are spurious errors due to Dialyzer assuming opaques are invariant. [Example and discussion](https://github.com/erlang/otp/pull/5223#issuecomment-925696731)

1. Treat opaques as invariant by default, allow annotations for covariant and contravariant parameters

This idea was suggested by @jhogberg in [a comment on #5223](https://github.com/erlang/otp/pull/5223#issuecomment-924917745). I think it would look something like:

`-opaque foo(InvariantParam, +CovariantParam, -ContravariantParam)`

Requiring such variance annotations seems like a high burden for users: not just at the use site, but at the call site as well, since one has to look at the source code of the opaque type to see what the variance is in order to know whether one is using the type correctly. The burden would be worth it if Erlang were a mutation-heavy language with lots of compound data structures, but Erlang is not that kind of language and I haven't seen any contravariant opaques. Type variables are already [pretty difficult for some devs to understand](https://youtu.be/_u1NDuFsW2A?t=1528), and I worry variance annotations would make the problem worse.

> An open question is whether plain type aliases (`-type`) would also require annotations for variance.

2. Opaques covariant by default, allow annotations for invariant and contravariant parameters

This is consistent with my proposal and could be done as optional follow-on work.

**Additional context**

This issue grew out of discussion on https://github.com/erlang/otp/pull/5223

mheiber:
close  #5240

## Pre-land checklist

- [x] Verify the xml with ~~`make release_docs`~~ `make xmllint`

> I wasn't able to verify this, since the command fails for me locally: make[3]: `*** No rule to make target `../chunks/array.chunk', needed by `chunks'.  Stop.`

- [x] Format the XML better

### DIFF
--- a/system/doc/reference_manual/opaques.xml
+++ b/system/doc/reference_manual/opaques.xml
@@ -22,7 +22,7 @@
 
     </legalnotice>
 
-    <title>Types and Function Specifications</title>
+    <title>Opaques</title>
     <prepared>Max Heiber</prepared>
     <docno></docno>
     <date></date>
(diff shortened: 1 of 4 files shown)


system/doc/reference_manual/opaques.xml
```diff
@@ -0,0 +1,69 @@
...
+    <prepared>Max Heiber</prepared>
+    <docno></docno>
+    <date></date>
+    <rev></rev>
+    <file>opaques.xml</file>
+  </header>
+
+  <section>
+    <title>Opaque Type Aliases</title>
+    <p>The main use case for opacity in Erlang is to hide the implementation of a data type, enabling evolving the API while minimizing risk of breaking consumers. The runtime does not check opacity. Dialyzer provides some opacit-checking, but the rest is up to convention.
```
elbrujohalcon, line 35:
```suggestion
    <p>The main use case for opacity in Erlang is to hide the implementation of a data type, enabling evolving the API while minimizing the risk of breaking consumers. The runtime does not check opacity. Dialyzer provides some opacity-checking, but the rest is up to convention.
```

system/doc/reference_manual/opaques.xml
```diff
@@ -0,0 +1,69 @@
...
+
+        <pre>-opaque set(Element) :: #set{segs :: segs(Element)}.</pre>
+            OTP 24  changed the definition to the following, in <a href="https://github.com/erlang/otp/commit/e66941e8d7c47b973dff94c0308ea85a6be1958e">this commit</a>.
+        <pre>-opaque set(Element) :: #set{segs :: segs(Element)} | #{Element => ?VALUE}.</pre>
+<p>And this change was safer and more backwards-compatible than if the type had been defined with <pre>-type</pre> instead of <pre>-opaque</pre>. Here's why: when a module defines an <pre>-opaque</pre>, the contract is that only the defining module should rely on the definition of the type: no other modules should rely on the definition.</p>
+<p>This means that code that pattern-matched on <pre>set</pre> as a record/tuple technically broke the contract, and opted in to being potentially broken when the definition of <pre>set()</pre> changed. Before OTP 24, this code printed <pre>ok</pre>. In OTP 24 it may error:</p>
+    <pre>
+    Set = sets:new(),
+caseof
+    X when is_tuple(Set) -&gt;
```
elbrujohalcon, line 49:
```suggestion
case sets:new() of
    Set when is_tuple(Set) -&gt;
```

system/doc/reference_manual/opaques.xml
```diff
@@ -0,0 +1,69 @@
...
+<item><pre>sets:set(a)</pre> is a subtype of <pre>sets:set(a | b)</pre> and not the other way around. Generally, you can rely on the property that <pre>the_opaque(T)</pre> is a subtype of <pre>the_opaque(U)</pre> when T is a subtype of U.</item>
+</list>
+<p><strong>When defining your own opaques, here are some recommendations:</strong></p>
+<list type="bulleted">
+<item>Since consumers are expected to not rely on the definition of the opaque type, you must provide functions for constructing and querying/deconstructing intances of your opaque type. For example, sets can be constructed with <pre>sets:new/0</pre>, <pre>sets:from_list/1</pre>, <pre>sets:add/2</pre>, queried with <pre>sets:is_element/2</pre>, and deconstructed with <pre>sets:to_list/1</pre>.</item>
+<item>Don't define an opaque with a type variable in parameter position. This breaks the normal and expected behavior that (for example) <pre>my_type(a)</pre> is a subtype of <pre>my_type(a | b)</pre></item>
+</list>
```
elbrujohalcon, line 64: I would add an extra item to this list noting that, if you define an opaque in a module, you *must* add specs to all the functions that use it. In other words: https://medium.com/erlang-battleground/help-dialyzer-help-you-94db66bfbc5a

elbrujohalcon (changes requested):

jhogberg (commented):
Thanks for the PR!

> Is there a tool that auto-formats or validates the formatting of the XML? I know it's a little wonky in my PR now

Hop down into the source folder and run `make xmllint`

mheiber:
> > Is there a tool that auto-formats or validates the formatting of the XML? I know it's a little wonky in my PR now
> 
> Hop down into the source folder and run `make xmllint`

Looks promising, but `make xmllint` errors for me with `no funciton clause matching docgen_xml_to_chunk:find_type_arity(ascii_string,[])`: I pasted the full stack in the PR description.  I get the same error on master, so there must be something up with my environment.

system/doc/reference_manual/opaques.xml
```diff
@@ -0,0 +1,66 @@
...
+
+  <section><title>Opaque Type Aliases</title>
+    <p>The main use case for opacity in Erlang is to hide the implementation of a data type, enabling evolving the API while minimizing the risk of breaking consumers. The runtime does not check opacity. Dialyzer provides some opacity-checking, but the rest is up to convention.
+    </p>
+    <p>
+This document explains what Erlang opacity is (and the trade-offs involved) via the example of OTP's <c>sets:set()</c> data type. This type <em>was</em> defined in `sets` module like this:
+    </p>
+
+        <code type="erl">-opaque set(Element) :: #set{segs :: segs(Element)}.</code>
+            OTP 24  changed the definition to the following, in <url href="https://github.com/erlang/otp/commit/e66941e8d7c47b973dff94c0308ea85a6be1958e">this commit</url>.
```
jhogberg, line 41: Wrap this line in a `<p>`

garazdawi:
> UDPATE I wiped my environment pretty clean, and now am getting this error when I do make xmllint on master: make[3]: *** No rule to make target ../chunks/array.chunk', needed by chunks'. Stop.

The reason why you get that is that the doc build needs a built repository to start with. So:

```sh
./configure && make && make xmllint
```

mheiber:
> > UDPATE I wiped my environment pretty clean, and now am getting this error when I do make xmllint on master: make[3]: *** No rule to make target ../chunks/array.chunk', needed by chunks'. Stop.
> 
> The reason why you get that is that the doc build needs a built repository to start with. So:
> 
> ```shell
> ./configure && make && make xmllint
> ```

Thanks, I tried this after `./otp_build autoconf`. Now I'm getting   `pkey.c:76:14: error: implicit declaration of function 'FIPS_mode' is invalid in C99 [-Werror,-Wimplicit-function-declaration]`. On Mac. OTP built fine for me on this machine a few months ago, but little luck now.

mheiber:
I got xmllint to pass, so I think this is ready for review again.

I had to build on Linux, xmllint didn't work for me on Mac.

system/doc/reference_manual/opaques.xml
```diff
@@ -0,0 +1,141 @@
...
+    <p>
+      <strong>When defining your own opaques, here are some recommendations:</strong>
+    </p>
+    <list type="bulleted">
+      <item>
+        Since consumers are expected to not rely on the definition of the opaque type, you must provide functions for constructing and querying/deconstructing intances of your opaque type. For example, sets can be constructed with
+        <c>sets:new/0</c>, <c>sets:from_list/1</c>, <c>sets:add/2</c>, queried with <c>sets:is_element/2</c>, and deconstructed with<c>sets:to_list/1</c>.
+      </item>
+      <item>
+        Don't define an opaque with a type variable in parameter position. This breaks the normal and expected behavior that (for example)
```
seriyps, line 115: > opaque with a type variable in parameter position

what does it mean?

system/doc/reference_manual/opaques.xml
```diff
@@ -0,0 +1,141 @@
...
+        <c>sets:set(a)</c>
+        is a subtype of
+        <c>sets:set(a | b)</c>
+        and not the other way around. Generally, you can rely on the property that
+        <c>the_opaque(T)</c>
+        is a subtype of
+        <c>the_opaque(U)</c>
+        when T is a subtype of U.
+      </item>
+    </list>
```
seriyps, line 105: Another interesting pattern is when one stores opaque type in, eg, optional record field (or generally in some container) and then have some code that checks if record's field have been initialized:

```erlang
-record(lazy_set, {the_set :: sets:set() | undefined, ...}).

add(Val, #lazy_set{the_set = undefined} = LS) ->
   add(Val, add(#lazy_set{the_set = sets:new()}));
add(Val, #lazy_set{the_set = Set)) ->
  ...
```
or similar with maps
```erlang
-type my_type() :: #{my_key => sets:set()}.

-spec my_fun(..., my_type()) -> ...
my_fun(..., Map) ->
  case maps:get(my_key, Map, undefined) of
    undefined ->  ...;
    Set -> ...
  end
```

So, the recomendation, I believe, is to wrap the opaque value to some enum-style tuple, or have some flag in a separate field

```
-record(lazy_set, {the_set :: {set, sets:set()} | undefined}).
```

system/doc/reference_manual/opaques.xml
mheiber:
```erl
-opaque bad(T) :: fun(T) -> ok.
```

I've never seen one of these in the wild.

What's terrible about such an opaque is that the implementation leaks. `sets:set(a)` is a subtype of `sets:set(a | b)` (of course) but `bad(a)` is not a subtype of `bad(a |  b)`. Instead `bad(a | b)` is a subtype of `bad(a)`.

jhogberg:
Sorry about taking so long to respond, the winter holidays made it fall through the cracks. D:

I think it looks good and we're ready to merge it. Are there any more changes you'd like to make?

mheiber:
> Sorry about taking so long to respond, the winter holidays made it fall through the cracks. D:

No problem, and thanks for taking another look!

> I think it looks good and we're ready to merge it. Are there any more changes you'd like to make?

Looks good to me, I'd love to get this change in so I can share links to the new docs when people ask me about opaques!

jhogberg:

PR #5534, opened 2021-12-19
Fix typos in lib/kernel

kianmeng:


lib/kernel/examples/gen_tcp_dist/src/gen_tcp_dist.erl
```diff
@@ -700,7 +700,7 @@ dist_cntrlr_setup_loop(Socket, TickHandler, Sup) ->
 
 dist_cntrlr_input_setup(DHandle, Socket, Sup) ->
     link(Sup),
-    %% Ensure we don't try to put data before registerd
+    %% Ensure we don't try to put data before registers
```
bjorng, line 703:
```suggestion
    %% Ensure we don't try to put data before we are registered
```

lib/kernel/src/gen_tcp_socket.erl
```diff
@@ -1439,7 +1439,7 @@ handle_event({call, From}, {setopts, Opts}, State, {P, D}) ->
     %% we are in state 'recv' but are inactive is simply stored in the buffer.
     %% If activated: active: false -> once | true | N > 0
     %% We need to check if there is something in our buffers, and maybe deliver
-    %% it to its owner. This is what we do here. This should only occure
+    %% it to its owner. This is what we do here. This should only occurred
```
bjorng, line 1442:
```suggestion
    %% it to its owner. This is what we do here. This should only occur
```

lib/kernel/src/logger_h_common.erl
```diff
@@ -396,7 +396,7 @@ do_log_to_binary(Log,Config) ->
                                      {config,FormatterConfig},
                                      {log_event,Log},
                                      {bad_return_value,String},
-                                     {catched,{C2,R2,S2}}]),
+                                     {caught,{C2,R2,S2}}]),
```
bjorng, line 399: Don't change this spelling since it is code.

lib/kernel/test/logger_std_h_SUITE.erl
```diff
@@ -1780,7 +1780,7 @@ rotation_opts_restart_handler(Config) ->
     {ok,#file_info{size=29}} = file:read_file_info(Log++".0.gz"),
     [_] = filelib:wildcard(Log++".*"),
 
-    %% Stop/start handler and turn off compression. Check that achives
+    %% Stop/start handler and turn off compression. Check that achieves
```
bjorng, line 1783:
```suggestion
    %% Stop/start handler and turn off compression. Check that archives
```

lib/kernel/test/socket_SUITE.erl
```diff
@@ -22045,7 +22045,7 @@ api_opt_ipv6_recvpktinfo_udp(InitState) ->
 %% There seem to be some weirdness with the definition of this
 %% option, so its defined in an include file we don't include
 %% (directly or indirectly). And since some of the defines
-%% are occure in a file we *do* include (via netinet/in.h), we
+%% are occurred in a file we *do* include (via netinet/in.h), we
```
bjorng, line 22048:
```suggestion
%% occurred in a file we *do* include (via netinet/in.h), we
```

bjorng (changes requested):

lib/kernel/src/logger_h_common.erl
kianmeng: Should this variable name tally up with the `caught_call_exception` type here, like everywhere use `caught` instead of `catched`?

See https://github.com/erlang/otp/blob/master/lib/kernel/src/erpc.erl

```erlang
-type caught_call_exception() ::
        {throw, Throw :: term()}
      | {exit, {exception, Reason :: term()}}
      | {error, {exception, Reason :: term(), StackTrace :: [stack_item()]}}
      | {exit, {signal, Reason :: term()}}
      | {error, {?MODULE, Reason :: term()}}.
```

bjorng: I don't think this instance of `catched` has something to do with `erpc`.

@garazdawi Do you think it safe to change this instance of `catched` to `caught` or is the code (or documentation) that depends on it remaining `catched`?

garazdawi: Should be fine to change. The code is only used when a formatter crashes and the exact keys are not documented.

lib/kernel/test/socket_SUITE.erl
bjorng: Please remove the "are".

kianmeng: > Please remove the "are".

@bjorng Done.

bjorng (approved):
Added to our daily builds.

lib/kernel/test/socket_SUITE.erl
zuiderkwast: Shound't it be "occur" instead? Present tense...

bjorng: @zuiderkwast You are right. Present tense makes more sense.

@kianmeng Please change.

kianmeng: @bjorng @zuiderkwast :ok_hand:

lib/kernel/test/zlib_SUITE_data/zipdoc
```diff
@@ -767,7 +767,7 @@ General Format of a ZIP file
             Currently, date-time width is 32-bit, but future version may
             support be 64-bit times (see flags)
 
-          CrGMTOffs     Long(signed!)   difference "local Creat. time - UTC"
+          CrGMTOffs     Long(signed!)   difference "local Create. time - UTC"
```
bjorng, line 770: This change causes three test cases in `zlib_SUITE` to fail. Please undo the change.

bjorng (approved):
Thanks!

kianmeng:
:tada: :tada: :tada: :tada: :tada:

PR #5536, opened 2021-12-21
.gitignore: cleanup no longer relevant ignores

max-au:
OTP no longer contains HiPE or GS, and no longer supports OSE.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,11 +46,8 @@ i386-pc-solaris[0-9]*.[0-9]*
 i386-unknown-freebsd[0-9]*.[0-9]*
 x86_64-unknown-freebsd[0-9]*.[0-9]*
 x86_64-unknown-openbsd[0-9]*.[0-9]*
-tile-tilera-linux-gnu
 powerpc-unknown-linux-gnu
 aarch64-unknown-linux-gnu
-i386-elf-ose
-powerpc-unknown-ose
 
 # Mac OS X
 a.out.dSYM/
@@ -92,8 +89,6 @@ lib/crypto/priv/obj/
 lib/erl_interface/obj.md/
 lib/erl_interface/obj.mdd/
 lib/erl_interface/src/win32/
-lib/gs/priv/tcl/
-lib/gs/tcl/binaries/
 lib/megaco/src/flex/win32/
 lib/odbc/c_src/win32/
 lib/odbc/priv/bin/odbcserver.exe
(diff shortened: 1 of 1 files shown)


max-au:
I wonder if there is a way to separate source tree from build artifacts, so they can be put in a single directory that can then be safely ignored in bulk (similar to "release" subdirectory).

bjorng:
Thanks!

bjorng:
> I wonder if there is a way to separate source tree from build artifacts

I don't see how that could be done if we still want to be able to run Erlang directly from the Git repository without first doing `make install`.

max-au:
If I understand it right, operating from Git repository still requires at least `make` to complete. What if `make` writes artifacts into "_build" folder, similar to how `rebar3` does it? Re-creating original structure
```
_build
  opt-jit
    bin
      erl
      ...
    erts
      x86_64-pc-linux-gnu
      aarm64
    lib
      kernel
        ebin
        doc
```

This way it's possible to separate OTP types+flavours as well (no files overwritten)

bjorng:
Do you suggest that the installed system should have the same structure?

If so, there is probably some code (outside of OTP) that makes assumptions about the layout of directories and would fail. The question is how much code.

On the other hand, some users would like that they can have one installation that would work for all architectures.

To go forward with this idea (if that is your idea), I think we would need an EEP with a motivation and a discussion about backward compatibility. The Erlang Forums could be a good place for initial discussions to see what users think.

Or do you suggest to use the new layout only in the git repository?

In that case the code server, other part of OTP, and the test suites would have to be modified to work with both OTP in the git repository and installed. There should not be any major backward compatibility issues for users.

max-au:
I was hoping for (2), keeping the released OTP the way it is now, only changing where C/C++ compiler build artifacts are put. Currently, they are spread all over the repo. For example, `erts/obj/x86_64-pc-linux-gnu`, then `erts/lib_src/yielding_c_fun/x86_64-pc-linux-gnu`, and `lib/wx/c_src/x86_64-pc-linux-gnu` etc.. 

They can be put into, say, `_build/opt-jit/x86_64_pc-linux-gnu/[erts|yielding_c_fun|wx|crypto]`. The success criteria would be having nearly empty `.gitignore` file. It would only be necessary to ignore `_build` and `release` directories. I can even imagine not having `_build`, or putting it under `release`.

bjorng:
> I was hoping for (2), keeping the released OTP the way it is now, only changing where C/C++ compiler build artifacts are put.

Yes, seems to be possible without that much work.

PR #5540, opened 2021-12-21
[erts] reduce console output for Common Test

max-au:
Some test cases produce 20+ Mb of debug output which does
not help to figure out a problem. This change suppresses
logs, leaving a chance to remove suppression for local runs.

### DIFF
--- a/erts/emulator/sys/common/erl_check_io.c
+++ b/erts/emulator/sys/common/erl_check_io.c
@@ -2955,12 +2955,8 @@ erts_check_io_debug(ErtsCheckIoDebugInfo *ciodip)
     erts_dsprintf(dsbufp, "internal fds=%d\n", counters.internal_fds);
 #endif
     erts_dsprintf(dsbufp, "---------------------------------------------------------\n");
-    if (counters.num_errors > 0)
-        erts_send_error_to_logger_nogl(dsbufp);
-    else {
-        erts_free(ERTS_ALC_T_LOGGER_DSBUF, (void *) dsbufp->str);
-        erts_free(ERTS_ALC_T_LOGGER_DSBUF, (void *) dsbufp);
-    }
+
+    erts_send_error_to_logger_nogl(dsbufp);
 #ifdef ERTS_SYS_CONTINOUS_FD_NUMBERS
     erts_free(ERTS_ALC_T_TMP, (void *) counters.epep);
 #endif
--- a/erts/emulator/sys/common/erl_check_io.c
+++ b/erts/emulator/sys/common/erl_check_io.c
@@ -2955,7 +2955,12 @@ erts_check_io_debug(ErtsCheckIoDebugInfo *ciodip)
     erts_dsprintf(dsbufp, "internal fds=%d\n", counters.internal_fds);
 #endif
     erts_dsprintf(dsbufp, "---------------------------------------------------------\n");
-    erts_send_error_to_logger_nogl(dsbufp);
+    if (counters.num_errors > 0)
+        erts_send_error_to_logger_nogl(dsbufp);
+    else {
+        erts_free(ERTS_ALC_T_LOGGER_DSBUF, (void *) dsbufp->str);
+        erts_free(ERTS_ALC_T_LOGGER_DSBUF, (void *) dsbufp);
+    }
 #ifdef ERTS_SYS_CONTINOUS_FD_NUMBERS
     erts_free(ERTS_ALC_T_TMP, (void *) counters.epep);
 #endif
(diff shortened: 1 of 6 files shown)


erts/emulator/sys/common/erl_check_io.c
```diff
@@ -2955,7 +2955,7 @@ erts_check_io_debug(ErtsCheckIoDebugInfo *ciodip)
     erts_dsprintf(dsbufp, "internal fds=%d\n", counters.internal_fds);
 #endif
     erts_dsprintf(dsbufp, "---------------------------------------------------------\n");
-    erts_send_error_to_logger_nogl(dsbufp);
+    /* erts_send_error_to_logger_nogl(dsbufp); */
```
garazdawi, line 2958: I don't think we want to remove this error logger report completely as it helps a lot when trying to debug very rare race conditions in these testcases.

I wonder if we could use the `cth_log_redirect` on test suites that emit a lot of these warnings? That way they should end up in those testcase logs instead of stdout.

erts/emulator/test/busy_port_SUITE.erl
```diff
@@ -255,7 +255,7 @@ no_trap_exit(Config) when is_list(Config) ->
     Pid = fun_spawn(fun no_trap_exit_process/3, [self(), linked, Config]),
     receive
         {Pid, port_created, Port} ->
-            io:format("Process ~w created port ~w", [Pid, Port]),
+            ct:log("Process ~w created port ~w", [Pid, Port]),
```
garazdawi, line 258: Any reason why you made this change? The logs end up in the same place anyway.

bjorng (approved):
Looks good to me. @rickard-green, @garazdawi?

garazdawi (commented):
I wonder how much output could be avoided by just passing the cth_log_redirect hook inside the ct spec emulator?

erts/emulator/sys/common/erl_check_io.c
max-au: Maybe instead of printing this information it could be returned? So whenever test case fails, it gets printed, but otherwise it does not. My primary problem is that when `ts:run(emulator, [batch])` is used for testing, it generates >20Mb log files that aren't working well with things like GitLab.

erts/emulator/test/busy_port_SUITE.erl
max-au: I thought that test cases are expected to use `ct:log` (or, `print`, `pal`) to make the intent explicit. 
In my experience, I've been using `erlang:display` for hard debugging (when I want to avoid any log redirection), then `io:format(user, ...` for slightly softer debugging (and some convenience for argument format), but I could not see any value in having `io:format` in test suites. To my understanding, it should be one of `ct:pal/log/print`.

max-au:
What I would be really interested in is some built-in CT console output filter. Similar to what `cth_readable` CT hook does for tests running via rebar3. That is, by default don't print anything for successful test cases, but print when the test case fails.
That would keep CT output quite neat.

erts/emulator/sys/common/erl_check_io.c
garazdawi: Yes, doing it like that would also be an acceptable solution.

What is the limit that gitlab has? You cannot have too large files, or is it the console log that cannot be too big?

erts/emulator/test/busy_port_SUITE.erl
garazdawi: I agree with your reasoning and I do the same as you. I was only curious if you had any other reason.

garazdawi:
> What I would be really interested in is some built-in CT console output filter. Similar to what cth_readable CT hook does for tests running via rebar3. That is, by default don't print anything for successful test cases, but print when the test case fails.
That would keep CT output quite neat.

Yes, I agree. Though if we were to do something like that I think that it might be faster to just make it possible to use `rebar3 ct` to run the tests so that we get all of the bells and whistles of `rebar3 ct` and not just a subset. I made a quick attempt at trying to do this a while ago, but `rebar3 ct` expects to be able to build the main application before it runs the tests (which is normally a good thing), so getting it to work for us turned out to not be trivial.

erts/emulator/sys/common/erl_check_io.c
max-au: Hard limit is 20Mb, with 500kb displayed. Files can be of any size, but console output is limited.
I found that printing this debug info can only be useful when test case is about to fail, hence I changed the diff to send the error log only when error is actually detected.

With cth_surefire support this change may not be relevant, but I still think it'd be good to have it for manual test runs.

max-au: Works for me too, output still fits in 500kb displayed.

Issue #5264, opened 2021-10-06
erl -profile_boot true    vm  crash


xiake-1024:
**Describe the bug**
erl -profile_boot true  vm crash

**To Reproduce**
C:\Users\Administrator>erl -profile_boot true
{"init terminating in do_boot",{function_clause,[{init,collect_mfa,[{math,fmod,2
},undefined,0,0],[{file,"init.erl"},{line,1544}]},{init,collect_mfas,2,[{file,"i
nit.erl"},{line,1535}]},{init,do_boot,3,[{file,"init.erl"},{line,892}]}]}}
init terminating in do_boot ({function_clause,[{init,collect_mfa,[{_},undefined,
0,0],[{_},{_}]},{init,collect_mfas,2,[{_},{_}]},{init,do_boot,3,[{_},{_}]}]})

Crash dump is being written to: erl_crash.dump...done

**Affected versions**
OTP 24

**Additional context**
My suggest is in  init.erl
collect_mfas([MFA|MFAs],Info) ->
    case erlang:trace_info(MFA,call_time) of
        {call_time, []} ->
            collect_mfas(MFAs,Info);
        {call_time, undefined} ->
            collect_mfas(MFAs,Info);
        {call_time, false} ->
            collect_mfas(MFAs,Info);
        {call_time, Data} ->
            case collect_mfa(MFA,Data,0,0) of
                {{0,_},_} ->
                    %% ignore mfas with zero time
                    collect_mfas(MFAs,Info);
                MfaData ->
                    collect_mfas(MFAs,[MfaData|Info])
            end
    end.

garazdawi:
Fixed in #5546

PR #5546, opened 2021-12-23
Fix erl -profile_boot true crash

vkatsuba:
Inside of `init:collect_mfas/2` function is used `erlang:trace_info(_, call_time)` which can return
`{call_time, list()}` | `{call_time, boolean()}` | `{call_time, undefined}`
The issue was in missing clause matching for `{call_time, undefined}`

Related to https://github.com/erlang/otp/issues/5264.


vkatsuba:
This is my first contribution for `erts` and I not sure if I need also update preloaded and primary commits. If I need to do it, just let me know.

garazdawi:
> This is my first trying of contribution for erts and I not sure if I need also update preloaded and primary commits. If I need to do it, just let me know.

We will push any updates to preloaded and the primary bootstrap that are needed. So what you have done is correct.

Adding this change to tests, not that I expect that it will break anything but you never know...

PR #5552, opened 2021-12-27
jit: Optimize yield tests

jhogberg:
This PR makes yield tests slightly cheaper, improving `dialyzer` performance by about 1% on my x64 machine.

### DIFF
--- a/erts/emulator/beam/jit/x86/instr_common.cpp
+++ b/erts/emulator/beam/jit/x86/instr_common.cpp
@@ -1916,19 +1916,14 @@ void BeamModuleAssembler::emit_raw_raise() {
     a.mov(getXRef(0), imm(am_badarg));
 }
 
+#define TEST_YIELD_RETURN_OFFSET                                               \
+    (BEAM_ASM_FUNC_PROLOGUE_SIZE + 16 +                                        \
+     (erts_frame_layout == ERTS_FRAME_LAYOUT_FP_RA ? 4 : 0))
+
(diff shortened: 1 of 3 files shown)


erts/emulator/beam/jit/x86/instr_common.cpp
```diff
@@ -1938,20 +1933,19 @@ void BeamGlobalAssembler::emit_i_test_yield_shared() {
...
     emit_enter_frame();
 
-    a.bind(entry);
+    a.lea(ARG3, x86::qword_ptr(currLabel, TEST_YIELD_RETURN_OFFSET));
     a.dec(FCALLS);
-    a.short_().jg(next);
-    a.lea(ARG3, x86::qword_ptr(entry));
-    a.call(yieldEnter);
-    a.bind(next);
+    a.jle(yieldEnter);
```
garazdawi, line 1945: Am I correct in thinking that this will make the process not do a `dec FCALLS` when scheduled back in? Is that a problem?

garazdawi (approved):
Could you expand a bit in the commit message of `jit: Optimize yield tests` what the optimization does?

erts/emulator/beam/jit/x86/instr_common.cpp
jhogberg: Yes, it'll now match both the ARM JIT and the interpreter. I don't expect it to cause any problems.

PR #5559, opened 2021-12-30
Fix typos in lib/snmp

kianmeng:


bmk (approved):
My only question was for the changed (standard) mibs.
But since they where all located in the test dir and where used
in testing, I think we can let that slide.

kianmeng:
:tada: :tada: :tada: :tada: :tada:

PR #5563, opened 2021-12-31
tools/cover: Add method to generate summary file

avtobiff:
This commit adds the capability to cover:analyse_to_file/1,2 to generate
a coverage summary file. The summary file consists of a table where each
row contains module name, coverage percent, covered lines, and not
covered lines.

The summary file is generated by default if more than one module is
analyzed.

New options: {summary, boolean()} and {summaryfile, list()}.

By default the summary file is named "all.COVER.summary", or
"all.COVER.summary.html" if the option 'html' is enabled.

By default UTF-8 encoding is used for the summary file.

Signed-off-by: Per Andersson <avtobiff@foo.nu>

### DIFF
--- a/lib/tools/src/cover.erl
+++ b/lib/tools/src/cover.erl
@@ -547,7 +547,9 @@ analyse_to_file() ->
 
 -type analyse_option() :: 'html'
                         | {'outfile', OutFile :: file:filename()}
-                        | {'outdir', OutDir :: file:filename()}.
+                        | {'outdir', OutDir :: file:filename()}
+                        | {'summary', Summary :: boolean()}
+                        | {'summaryfile', SummaryFile :: file:filename()}.
 -type analyse_answer() :: {'ok', OutFile :: file:filename()} |
                           {'error', analyse_rsn()}.
 -type analyse_file_ok() :: [OutFile :: file:filename()].
@@ -635,7 +637,8 @@ is_options([html]) ->
     true; % this is not 100% safe - could be a module named html...
 is_options([html|Opts]) ->
     is_options(Opts);
-is_options([{Opt,_}|_]) when Opt==outfile; Opt==outdir ->
+is_options([{Opt,_}|_])
+  when Opt == outfile; Opt == outdir; Opt == summary; Opt == summaryfile ->
     true;
 is_options(_) ->
     false.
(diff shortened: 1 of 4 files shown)


lib/tools/src/cover.erl
```diff
@@ -2704,8 +2707,20 @@ analyse_list_to_file(Modules, Opts, State) ->
...
+    {AnalyseOk, AnalyseError} = split_ok_error(pmap(Fun, MFs), [], []),
+
+    %% Don't generate a summary file if only one module is analysed
+    {Ok, Error1} =
+        if length(MFs) > 1 ->
+               SummaryRes = do_summary(MFs, Opts, State),
+               split_ok_error(SummaryRes, AnalyseOk, AnalyseError);
+           true ->
+               {AnalyseOk, AnalyseError}
+        end,
```
avtobiff, line 2721: Regarding not emitting the summary report if explicitly analysing only one
file, perhaps it should be enabled by default instead of omitted.

The information in the summary is not readily available in any other way
without manually analysing with e.g. cover:analyse/0 to get covered,
not covered lines, and then calculating the coverage percentage.

lib/tools/src/cover.erl
```diff
@@ -2834,6 +2831,222 @@ do_analyse_to_file1(Module, OutFile, ErlFile, HTML) ->
...
+    {result, CovRes, CovErr} = analyse_list(Compiled, coverage, module, State),
+
+    %% Character encoding is hardcoded to UTF-8 for summary file
+    Enc = utf8,
+    HTML = lists:member(html, Opts),
+    SummaryFile= summaryfilename(Opts),
+
+    %% Try find application name by searching source dirs for .app.src and
+    %% .app. Iff a single application name is found, use it; otherwise use
+    %% the summary file name.
```
bjorng, line 2857: I suggest breaking out the code for generating the title to a helper function.

lib/tools/src/cover.erl
```diff
@@ -2834,6 +2831,222 @@ do_analyse_to_file1(Module, OutFile, ErlFile, HTML) ->
...
+
+
+do_summary1(Modules, Opts, State) ->
+    Compiled = get_modules(Modules),
+    {result, CovRes, CovErr} = analyse_list(Compiled, coverage, module, State),
+
+    %% Character encoding is hardcoded to UTF-8 for summary file
+    Enc = utf8,
+    HTML = lists:member(html, Opts),
+    SummaryFile= summaryfilename(Opts),
```
bjorng, line 2853:
```suggestion
    SummaryFile = summaryfilename(Opts),
```

lib/tools/src/cover.erl
```diff
@@ -2834,6 +2831,222 @@ do_analyse_to_file1(Module, OutFile, ErlFile, HTML) ->
...
+            {error, Reason} ->
+                {[], [{error, {file, SummaryFile, Reason}}]}
+        end,
+
+    FileOk ++ CovErr ++ FileErr.
+
+
+-spec summaryfilename(SummaryFilename :: list(), HTML :: boolean()) ->
+    SummaryFilenameRes :: list().
+summaryfilename(Opts) ->
```
bjorng, line 2934: Perhaps name this function `summary_filename`?

bjorng (changes requested):
I think that it would better to always generate the summary by default, even if there is only module.

It seems unnecessary to have decimals for the percentages. I suggest using only integers.

The columns in the output do not line up properly:

```
beam_block	 81.58%	93	21
beam_asm	 83.97%	199	38
core_scan	  0.00%	0	173
rec_env	  0.00%	0	96
sys_pre_attributes	  0.00%	0	58
core_pp	  0.00%	0	194
```

I suggest using spaces instead of tabs.

The modules are listed in random order. I suggest sorting by module name.

lib/tools/src/cover.erl
avtobiff: I prefer this style, I just followed how outfilename was named.

I'll change it.

avtobiff:
I have been busy with other commitments. Hope to push an update
with review fixes soon.

bjorng:
I'll close this PR for now. Feel free to re-open it when you addressed my comments above.

avtobiff:
Yes, I have not had time to finish up the last remaining test changes.

I'll return with a PR when I have time to address the comments.

PR #5566, opened 2022-01-01
[tools] replace test_server:start_node with ?CT_PEER()

max-au:
Make tests more robust - node names can never clash now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,9 @@ TAGS
 # vscode
 .vscode
 
+# IntelliJ IDEA
+.idea
+
 autom4te.cache
 *.beam
 *.asn1db
--- a/lib/common_test/src/test_server.erl
+++ b/lib/common_test/src/test_server.erl
@@ -2832,7 +2832,9 @@ start_peer(#{name := Name} = Opts, Module) ->
         end,
     FullArgs = CookieArg ++ ["-pa", filename:dirname(code:which(Module)),
         "-env", "ERL_CRASH_DUMP", CrashFile] ++ Args,
-    case test_server:is_cover() of
+    %% start_cover => false is intentionally undocumented, and is not
+    %%  expected to be used by anything but cover_SUITE test.
+    case maps:get(start_cover, Opts, true) andalso test_server:is_cover() of
         true ->
             %% when cover is active, node must shut down gracefully, otherwise
             %% coverage information won't be sent to cover master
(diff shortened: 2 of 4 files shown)


lib/tools/test/emem_SUITE.erl
```diff
@@ -138,8 +138,7 @@ end_per_suite(Config) when is_list(Config) ->
 
 live_node(Config) when is_list(Config) ->
     {ok, EmuFlag, Port} = start_emem(Config),
-    Nodename = mk_nodename(Config),
```
garazdawi, line 141: This SUITE does not compile as the `Nodename` variable is used further down in this function.

max-au: My bad, I did not intend to change this file at all, but asked @bjorng whether it can be removed altogether. I was not able to find any way to actually run these tests. They are always skipped, because `proplists:get_value(data_dir, Config)` (line 100) returns the name of the directory that does not exist.
I was also not able to find any mentions of "emem" utility except for `c_src` folder that builds it. I could not find any documentation what the tool does either.

How can I run these tests, and how do I run the tool? It felt abandoned, so I wonder whether I can leave test case as-is, and then make another PR with the tool removed completely, or documented/explained.

garazdawi: The emem tool is a "memory tracer" that was used a long time ago, but the code in the emulator is still there and it can be enabled using `+Mit IP:port`. Though since it seems like it has not been tested in a long time I would be very surprised if it will worked.

@rickard-green do you have any idea where the emem_SUITE_data has gone? It seems like it was not included in the initial import from clearcase to git? I suppose that we can remove it all without anyone missing it?

garazdawi: @rickard-green says that we can remove it, so a PR would be welcome.

max-au:
- kept emem_SUITE (to be removed in a separate PR)
- added .idea to .gitignore (can be a separate PR, but why not...)
- rebased to master to ensure no merge conflicts can happen

lib/tools/test/emem_SUITE.erl
max-au: My pleasure - https://github.com/erlang/otp/pull/5591

garazdawi:
The reason why this is not merged yet is that we are chasing down a bug that manifests in cover_SUITE when this PR is run together with #5611.

max-au:
Right, that diff still uses the `test_server:start_node` (mainly for the reason I haven't completed the big PR that peerifies kernel application). That is unfortunate, as I re-did global_SUITE significantly.

PR #5567, opened 2022-01-02
[runtime_tools] replace test_server:start_node with ?CT_PEER

max-au:
Also fix some test cases that may fail due to linked processes
going down when stopping the peer node.


garazdawi:
thanks!

PR #5577, opened 2022-01-04
Remove unused erl_eval code

josevalim:
* {value, Value, Binding} was an implicit contract
    with shell that is no longer used

  * Both record and tuple calls have been deprecated
    and removed for quite some time


jhogberg:
Merged, thanks for the PR!

PR #5581, opened 2022-01-05
ct: Add correct file and line to surefire report

garazdawi:
This PR adds file and line number info to the surefire report so that tools (such as [publish-unit-test-result-action](https://github.com/EnricoMi/publish-unit-test-result-action)) can locate where a failed test case is.

The algorithm for finding the source file is very naive, so most users will most likely need to post-process the XML file in order to get the correct path.

### DIFF
--- a/lib/common_test/test/ct_surefire_SUITE_data/surefire_SUITE.erl
+++ b/lib/common_test/test/ct_surefire_SUITE_data/surefire_SUITE.erl
@@ -44,6 +44,7 @@ groups() ->
 testcases() ->
     [tc_ok,
      tc_fail,
+     tc_badmatch,
      tc_skip,
      tc_autoskip_require].
 
@@ -76,11 +77,16 @@ break(_Config) ->
     test_server:break(""),
     ok.
 
-tc_ok(_Config) ->
-    ok.
+tc_ok(Config) when is_list(Config) ->
+    ok;
+tc_ok(_) ->
+    ct:fail("This should never fail").
 
 tc_fail(_Config) ->
     ct:fail("this test should fail").
+
+tc_badmatch(_Config) ->
+    ok = nok.
 
 tc_skip(_Config) ->
     {skip,"this test is skipped"}.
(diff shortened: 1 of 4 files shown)


u3s (approved):
please try to find the test case that is MIA :-)

lib/common_test/test/ct_surefire_SUITE_data/surefire_SUITE.erl
```diff
@@ -76,12 +77,17 @@ break(_Config) ->
     test_server:break(""),
     ok.
 
-tc_ok(_Config) ->
-    ok.
+tc_ok(Config) when is_list(Config) ->
+    ok;
+tc_ok(_) ->
```
u3s, line 82: Whats the purpose of this clause?

u3s (approved):

lib/common_test/test/ct_surefire_SUITE_data/surefire_SUITE.erl
garazdawi: To check that the line number reported is the first clause of a function and that multiple clauses actually work.

u3s: Does it also work if you reverse the clause order? So the 2nd matches?

max-au:
I wonder how this change is reflected in the report?

This is how junit_report.xml rendered with GitLab looks before the change:
before

And that's after the change:
after

(I'm interested in improving cth_surefire to a degree that it may be used for test debugging).

lib/common_test/test/ct_surefire_SUITE_data/surefire_SUITE.erl
garazdawi: No, it takes the first clause no matter what.

garazdawi:
I had a quick look at the [gitlab source](https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/parsers/test/junit.rb#L81-91) and it seems like they don't parse the `line` number, but they do parse the `file`. What they do with it is unclear though. Maybe the file has to exist? So if you make sure to rewrite the paths in the surefire XML file it will point to the correct place?

PR #5583, opened 2022-01-05
Introducing lists:foldlwhile

meox:
The goal of the PR is to introduce `lists:foldlwhile` taking inspiration from the Elixir `Enum:reduce_while`.
The motivation is to add an easy way to interrupt the folding action as soon as some condition has been detected.

### DIFF
--- a/lib/stdlib/doc/src/lists.xml
+++ b/lib/stdlib/doc/src/lists.xml
@@ -324,6 +324,28 @@ flatmap(Fun, List1) ->
     </func>
 
     <func>
+      <name name="foldlwhile" arity="3" since=""/>
+      <fsummary>Fold a function over a list until <anno>halt</anno> is returned</fsummary>
+      <desc>
+        <p>Calls <c><anno>Fun</anno>(<anno>Elem</anno>, <anno>AccIn</anno>)</c>
+          on successive elements <c>A</c> of <c><anno>List</anno></c>, starting
+          with <c><anno>AccIn</anno> == <anno>Acc0</anno></c>.
+          <c><anno>Fun</anno>/2</c> must return a tuple <c>{cont, Acc}</c>
+          to continue or <c>{halt, Acc}</c>
+          to stop the folding. The second element of the tuple is the accumulator, which is
+          passed to the next call. The function returns the final value of
+          the accumulator. <c><anno>Acc0</anno></c> is returned if the list is
+          empty.</p>
+        <p><em>Example:</em></p>
+        <pre>
+> <input>lists:foldlwhile(fun(X, Sum) when X=&lt;3 -> {cont, X + Sum}; (_X, Sum) -> {halt, Sum} end, 0, [1,2,3,4,5]).</input>
+6
+> <input>lists:foldlwhile(fun(X, Prod) when X=&lt;3 -> {cont, X * Prod}; (_X, Prod} -> {halt, Prod} end, 1, [1,2,3,4,5]).</input>
+6</pre>
+      </desc>
+    </func>
+
+    <func>
       <name name="foldr" arity="3" since=""/>
       <fsummary>Fold a function over a list.</fsummary>
       <desc>
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -39,7 +39,7 @@
 -export([all/2,any/2,map/2,flatmap/2,foldl/3,foldr/3,foldlwhile/3,filter/2,
 	 partition/2,zf/2,filtermap/2,
 	 mapfoldl/3,mapfoldr/3,foreach/2,takewhile/2,dropwhile/2,
-         search/2, splitwith/2,split/2,
+	 search/2, splitwith/2,split/2,
 	 join/2]).
 
 %%% BIFs
@@ -1347,11 +1347,11 @@ foldr_1(_F, Accu, []) ->
 %%  example:
 %%  lists:foldlwhile(
 %%    fun
-%%      (X, Acc) when X < 3 ->
-%%        {cont, Acc+X};
-%%      (X, Acc) ->
-%%        {halt, Acc+X}
-%%    end, 0, [1,2,3,4,5]).
+%%      (X, Acc) when X =< 3 ->
+%%        {cont, Acc + X};
+%%      (_X, Acc) ->
+%%        {halt, Acc}
+%%    end, 0, [1, 2, 3, 4, 5]).
 %%  > 6
 
 -spec foldlwhile(Fun, Acc0, List) -> Acc1 when
@@ -1363,7 +1363,6 @@ foldr_1(_F, Accu, []) ->
       List :: [T],
       T :: term().
 
-foldlwhile(_F, Accu, []) -> Accu;
 foldlwhile(F, Accu, List) when is_function(F, 2) ->
     foldlwhile_1(F, Accu, List).
 
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -36,7 +36,7 @@
 
 -export([merge/3, rmerge/3, sort/2, umerge/3, rumerge/3, usort/2]).
 
--export([all/2,any/2,map/2,flatmap/2,foldl/3,foldr/3,filter/2,
+-export([all/2,any/2,map/2,flatmap/2,foldl/3,foldr/3,foldlwhile/3,filter/2,
 	 partition/2,zf/2,filtermap/2,
 	 mapfoldl/3,mapfoldr/3,foreach/2,takewhile/2,dropwhile/2,
          search/2, splitwith/2,split/2,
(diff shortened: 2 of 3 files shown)


garazdawi:
If this function is to be included, I think the name should be `lists:foldlwhile/3` to keep the naming convention with the rest of the `lists` module.

There also needs to be tests + docs, but that can be completed when we have decided whether we want this or not.

meox:
Ok waiting for your final decision

lhoguin:
I think it would be very useful, I can't count the number of times I have written functions to do just that. The alternative is to do lists:fold+throw and that's not often super readable.

meox:
doc & tests added

bjorng:
The OTP Technical Board will discuss this pull request in an upcoming meeting.

lib/stdlib/doc/src/lists.xml
```diff
@@ -323,6 +323,28 @@ flatmap(Fun, List1) ->
...
+          <c><anno>Fun</anno>/2</c> must return a tuple <c>{<anno>cont</anno>, <anno>Acc</anno>}</c>
+          to continue or <c>{<anno>halt</anno>, <anno>Acc</anno>}</c>
+          to stop the folding. The second element of the tuple is the accumulator, which is
+          passed to the next call. The function returns the final value of
+          the accumulator. <c><anno>Acc0</anno></c> is returned if the list is
+          empty.</p>
+        <p><em>Example:</em></p>
+        <pre>
+> <input>lists:foldlwhile(fun(X, Sum) when X&lt;3 -> {cont, X + Sum}; (X, Sum) -> {halt, X + Sum} end, 0, [1,2,3,4,5]).</input>
+6
```
KennethL, line 342: I don't think this example is obvious. I would expect the iteration to return the sum of elements with a value less than 3. In this case 1+2. Since it halt on 3 I don't think it is obvious to add 3 to the result in the `{halt, X + Sum}`.
It would have been more intuitive if the fun would have been `fun(X,Sum) when X=<3 -> {cont, X + Sum}; (X, Sum) -> {halt, X + Sum} end`. In this case it feels natural that the value 3 is included in the sum.

meox: thanks for the suggestion, just fixed

meox:
I don't understand this error in the doc:

Error in lists:foldlwhile/-: unknown annotation cont GEN	../man3/math.3

meox:
Now the doc seems ok: https://github.com/meox/otp/actions/runs/1686771938

lib/stdlib/doc/src/lists.xml
```diff
@@ -323,6 +323,28 @@ flatmap(Fun, List1) ->
...
 
+    <func>
+      <name name="foldlwhile" arity="3" since=""/>
+      <fsummary>Fold a function over a list until <anno>halt</anno> is returned</fsummary>
+      <desc>
+        <p>Calls <c><anno>Fun</anno>(<anno>Elem</anno>, <anno>AccIn</anno>)</c>
+          on successive elements <c>A</c> of <c><anno>List</anno></c>, starting
+          with <c><anno>AccIn</anno> == <anno>Acc0</anno></c>.
+          <c><anno>Fun</anno>/2</c> must return a tuple <c>{cont, Acc}</c>
+          to continue or <c>{halt, Acc}</c>
```
Maria-12648430, line 334: `cont` and `halt` sound strange, I have never encountered something like that anywhere else, certainly not in the `lists` module.

Maria-12648430:
Shouldn't there be a `foldrwhile` accompanying `foldlwhile`, for symmetry?

lib/stdlib/doc/src/lists.xml
meox: any suggestion?

meox:
> Shouldn't there be a `foldrwhile` accompanying `foldlwhile`, for symmetry?

honestly I tried at the beginning but I don't any idea how to implement it.

lib/stdlib/doc/src/lists.xml
Maria-12648430: If it is to follow the "style" of the `lists` module, it probably boils down to `true` and `false` 😓 That is not very intuitive I admit, but OTOH the fold function _is_ halfway a predicate, so... Well, what do others think? 😅

meox: I take the terminology from Elixir. I think that an atom say a lot more than a simple boolean.

Maria-12648430:
> > Shouldn't there be a `foldrwhile` accompanying `foldlwhile`, for symmetry?
> 
> honestly I tried at the beginning but I don't have any idea how to implement it. Possibile implementation:
> 
> ```erlang
> foldrwhile(F, Acc, List) ->
>    foldlwhile(F, Acc, reverse(List)).
> ```

Hm, off the top of my head, without trying:

```erlang
foldrwhile(F, Acc0, L) when is_function(F, 2) ->
    {_, Acc1}=foldrwhile_1(F, Acc0, L),
    Acc1.

foldrwhile_1(F, Acc0, [H|T]) ->
    case foldrwhile_1(F, Acc0, T) of
        R={halt, _Acc1} -> R;
        {cont, Acc1} -> F(H, Acc1)
    end;
foldrwhile_1(_F, Acc0, []) ->
    {cont, Acc0}.
```

It may also be acceptable to use a `throw` in the case of `halt` to return the `Acc` without unwinding the recursion, but I don't know ;)

lib/stdlib/doc/src/lists.xml
meox: Is the same problem that we have with the gen_server or other part of the stdlib.
Remembering stuff is hard but {true, Acc} or {false, Acc} for me doesn't means nothing.

Probably is better `Acc` and `{stop, Acc}`?

meox:
> > > Shouldn't there be a `foldrwhile` accompanying `foldlwhile`, for symmetry?
> > 
> > 
> > honestly I tried at the beginning but I don't have any idea how to implement it. Possibile implementation:
> > ```erlang
> > foldrwhile(F, Acc, List) ->
> >    foldlwhile(F, Acc, reverse(List)).
> > ```
> 
> Hm, off the top of my head, without trying (I'll stay with `cont`/`halt` for the time being ;)):
> 
> ```erlang
> foldrwhile(F, Acc0, L) when is_function(F, 2) ->
>     {_, Acc1}=foldrwhile_1(F, Acc0, L),
>     Acc1.
> 
> foldrwhile_1(F, Acc0, [H|T]) ->
>     case foldrwhile_1(F, Acc0, T) of
>         R={halt, _Acc1} -> R;
>         {cont, Acc1} -> F(H, Acc1)
>     end;
> foldrwhile_1(_F, Acc0, []) ->
>     {cont, Acc0}.
> ```
> 
> It may also be acceptable to use a `throw` in the case of `halt` to return the `Acc` without unwinding the recursion, but I don't know ;)

Before proceeding I'll wait for an approval from the board meeting. I really open to any suggestion, for me the important part is to have this PR merged in this way we can avoid to reimplement it several times.

lib/stdlib/doc/src/lists.xml
Maria-12648430: > Is the same problem that we have with the gen_server or other part of the stdlib.

There it is justified, they are complex, you simply can't get away with booleans there of course ;)

> Remembering stuff is hard but {true, Acc} or {false, Acc} for me doesn't means nothing.

It's obvious from the `foldlwhile` context it is used in IMO (and the `...while` in the function name hints at `while ($some_boolean) {...}` loops as found in other languages), but I may be biased and have just gotten used to it.

> * `Acc` and `{stop, Acc}`

No, this is ambiguous. `Acc` could be a 2-tuple with `stop` as first element.

> * `{true, Acc}` and `{false, Acc}`

That is my favorite :) Btw, using booleans means you can also use any function that evaluates to a boolean there. Compare this fold function...
```erlang
fun
    (N, Acc) -> {N<3, [N|Acc]}
end
```
... with this...
```erlang
fun
    (N, Acc) ->
        ContOrHalt=case N<3 of
            true -> cont;
            false -> halt
        end,
        {ContOrHalt, [N|Acc]}
end
```
... or this...
```erlang
fun
    (N, Acc) when N<3 ->
        {cont, [N|Acc]};
    (N, Acc) ->
        {halt, [N|Acc]}
end
```

meox: Yes I see your point but this trick apply only in some simple case like when the discriminator return a boolean. But in all the other case when you have a function that returns `{error, Reason}` you have to fallback use `case ... of` or other helpers

zuiderkwast:
> I think it would be very useful, I can't count the number of times I have written functions to do just that. The alternative is to do lists:fold+throw and that's not often super readable.

@lhoguin The real alternative is to write recursive functions to instead of using foldl.

meox:
@bjorng any comment on this PR?

KennethL:
We had an OTP Technical Board meeting where this and other suggested additions to the lists module where discussed.

We need to set the principles for what functions we add to the `lists` module and to other modules in `stdlib`. We can not just add all functions suggested.

- There must be convincing use cases for the functions where they can be used to write shorter, nicer, more understandable and correct code.
- The API must match the other functions in the module which they are added to.
- They should be known from other languages libraries and kind of expected in a standard library 

When it comes to `lists:foldwhile/3`  another suggestion which we actually prefer is a `foldwhile/4` where there are 2 separate fun:s, one for the calculation and one for the stop criteria.
But before digging deeper into that discussion we really want to see a lot of concrete use cases where this function would be very beneficial.
And about other libraries for other languages how much use is there of `Enum:reduce_while` in Elixir.

PR #5588, opened 2022-01-06
Add maps:groups_from_list

meox:


michalmuskala:
What about a `/3` variant that also allows mapping the value? It's pretty useful in [Elixir](https://hexdocs.pm/elixir/Enum.html#group_by/3).

vkatsuba:
I can be wrong but it's looks like in current PR is missed few points:
* Tests
* Documentation

meox:
> I can be wrong but it's looks like in current PR is missed few points:
> 
>     * Tests
> 
>     * Documentation

yes I'll add soon if there is some interest. I've opened a discussion also in the forum:
https://erlangforums.com/t/lists-foldlwhile-lists-groupby/913

meox:
I'm reading https://github.com/erlang/otp/blob/master/CONTRIBUTING.md but it's no clear for me what is the best way to add the doc. Any suggestions?

meox:
https://github.com/meox/otp/actions/runs/1664149588

garazdawi:
> I'm reading https://github.com/erlang/otp/blob/master/CONTRIBUTING.md but it's no clear for me what is the best way to add the doc. Any suggestions?

You should add an entry to [lists.xml](https://github.com/erlang/otp/blob/master/lib/stdlib/doc/src/lists.xml)

meox:
> What about a `/3` variant that also allows mapping the value? It's pretty useful in [Elixir](https://hexdocs.pm/elixir/Enum.html#group_by/3).

group_by/3 added plus test and doc

filmor:
Shouldn't it be more efficient to prepend the values and do `lists:reverse` in the end (or skip that step, if the order is not important)?

meox:
> Shouldn't it be more efficient to prepend the values and do `lists:reverse` in the end (or skip that step, if the order is not important)?

Yes we can reverse the list at the end. I prefer to maintain the order of the elements but it's an open point!

bjorng:
The OTP Technical Board will discuss this pull request in an upcoming meeting.

KennethL:
Since it returns a map another possibility is to have the function in the `maps` module maybe called `maps:from_groups` of `maps:groups_from_list` or just `maps:group_by`.  I might think the maps module is a better fit since there is currently no function returning maps in the lists module.

meox:
> Since it returns a map another possibility is to have the function in the `maps` module maybe called `maps:from_groups` of `maps:groups_from_list` or just `maps:group_by`. I might think the maps module is a better fit since there is currently no function returning maps in the lists module.

I can move in the `maps` module. Any preference for the name?

KennethL:
You don't need to move it to maps right now. I think we are only discussing different possibilities for the moment and nothing is decided. Fell free to comment on the original suggestion and the alternative to have it in the maps module. Or come with other suggestions:)

meox:
> You don't need to move it to maps right now. I think we are only discussing different possibilities for the moment and nothing is decided. Feel free to comment on the original suggestion and the alternative to have it in the maps module. Or come with other suggestions:)

there is also another issue: what this function should returns? A `map` or a `proplists`?

max-au:
I'd assume that anything in `maps` module returns maps, not proplists.
The major difference between maps and proplists is that maps cannot have duplicate keys, while proplist can. For this case I think map is more appropriate.

lib/stdlib/doc/src/lists.xml
```diff
@@ -1126,6 +1126,34 @@ splitwith(Pred, List) ->
...
+    <func>
+      <name name="groupby" arity="3" since=""/>
+      <fsummary>Splits the list into groups using a function as discriminator.</fsummary>
+      <desc>
+        <p>The result is a map where each key is given by <anno>Fun</anno> and each value is a list of elements given by the <anno>ValueFun</anno>. The order of elements within each list is preserved from the list.</p>
+        <p><em>Examples:</em></p>
+        <pre>
+> <input>lists:groupby(fun(X) -> X rem 2 end, fun(X) -> X*X end, [1,2,3]).</input>
+#{0 => [4], 1 => [1, 9]}
+> <input>lists:groupby(fun erlang:length/1, fun([H | _]) -> H end, ["ant", "buffalo", "cat", "dingo"]).</input>
```
seriyps, line 1153: use of `fun lists:reverse/1` as 2nd argument may make this example a bit more clear I think

meox: lists:groupby(fun erlang:length/1, fun reverse/1, ["ant", "buffalo", "cat", "dingo"])?

seriyps: yes. `fun lists:reverse/1`. Just IMO.

seriyps: Also, your current example actually does not produce the output
```
#{3 => ["a", "c"], 5 => ["d"], 7 => ["b"]}
```
It would produce lists of characters, not lists of strings
```
#{3 => [$a, $c], 5 => [$d], 7 => [$b]}
```
unless you change it the way you did in the test or wrap `H` in a list: `fun([H | _]) -> [H] end`

zuiderkwast:
I think it belongs in the lists module. It operates on a list and returns multiple lists. It is similar to lists:partition/2. Why no lists function so far returns a map is just that almost all of these functions are from before maps existed. That's no reason to avoid it for new functions IMO.

meox:
any update for this PR?

meox:
> The OTP Technical Board will discuss this pull request in an upcoming meeting.

any update for this PR?

garazdawi:
> any update for this PR?

We have not taken any decision yet, unforeseen circumstances have prevented the OTP Technical Board from meeting the last couple of weeks.

meox:
> The OTP Technical Board approved this proposal, but requested that the functions should be placed in the `maps` module and named `groups_from_list`. See [this post](https://erlangforums.com/t/otb-decision-on-four-proposed-extensions-to-the-lists-module/1146?u=bjorng) in the Erlang Forums for the motivation.
> 
> When you have done the changes, please squash into a single commit.

done

lib/stdlib/src/maps.erl
```diff
@@ -497,6 +498,48 @@ with_1([K|Ks], Map) ->
...
+    MapOut :: #{Selected => List},
+    Selected :: term(),
+    List :: [T],
+    T :: term().
+
+groups_from_list(_Fun, []) ->
+    #{};
+groups_from_list(Fun, List) when is_function(Fun, 1) ->
+    groups_from_list_1(Fun, lists:reverse(List), #{}).
+groups_from_list_1(Fun, [H | Tail], Acc) ->
```
bjorng, line 514: Insert a blank line before this function (for better readability).

lib/stdlib/src/maps.erl
```diff
@@ -497,6 +498,48 @@ with_1([K|Ks], Map) ->
...
+    K = Fun(H),
+    NewAcc = update_with(K, fun(Vs) -> [H | Vs] end, [H], Acc),
```
bjorng, line 516:
```suggestion
    NewAcc = case Acc of
                 #{K := Vs} -> Acc#{K := [H | Vs]};
                 #{} -> Acc#{K => [H]}
             end,
```

lib/stdlib/src/maps.erl
```diff
@@ -497,6 +498,48 @@ with_1([K|Ks], Map) ->
...
+    ValOut :: term(),
+    List :: [T],
+    ListOut :: [ValOut],
+    T :: term().
+
+groups_from_list(_Fun, _ValueFun, []) ->
+    #{};
+groups_from_list(Fun, ValueFun, List) when is_function(Fun, 1), is_function(ValueFun, 1) ->
+    groups_from_list_2(Fun, ValueFun, lists:reverse(List), #{}).
+groups_from_list_2(Fun, ValueFun, [H | Tail], Acc) ->
```
bjorng, line 535: Add a blank line before this line.

lib/stdlib/src/maps.erl
```diff
@@ -497,6 +498,48 @@ with_1([K|Ks], Map) ->
...
+    K = Fun(H),
+    V = ValueFun(H),
+    NewAcc = update_with(K, fun(Vs) -> [V | Vs] end, [V], Acc),
```
bjorng, line 538:
```suggestion
    NewAcc = case Acc of
                 #{K := Vs} -> Acc#{K := [V | Vs]};
                 #{} -> Acc#{K => [V]}
             end,
```

lib/stdlib/src/maps.erl
```diff
@@ -497,6 +498,48 @@ with_1([K|Ks], Map) ->
...
+    Fun :: fun((Elem :: T) -> Selected),
+    MapOut :: #{Selected => List},
+    Selected :: term(),
+    List :: [T],
+    T :: term().
+
+groups_from_list(_Fun, []) ->
+    #{};
+groups_from_list(Fun, List) when is_function(Fun, 1) ->
+    groups_from_list_1(Fun, lists:reverse(List), #{}).
```
bjorng, line 513: In order to support extended error information, the following changes are necessary:

```suggestion
groups_from_list(Fun, List0) when is_function(Fun, 1) ->
    try lists:reverse(List0) of
        List ->
            groups_from_list_1(Fun, List, #{})
    catch
        error:_ ->
            badarg_with_info([Fun, List0])
    end;
groups_from_list(Fun, List) ->
    badarg_with_info([Fun, List]).
```

lib/stdlib/src/maps.erl
```diff
@@ -497,6 +498,48 @@ with_1([K|Ks], Map) ->
...
+    Key :: term(),
+    ValOut :: term(),
+    List :: [T],
+    ListOut :: [ValOut],
+    T :: term().
+
+groups_from_list(_Fun, _ValueFun, []) ->
+    #{};
+groups_from_list(Fun, ValueFun, List) when is_function(Fun, 1), is_function(ValueFun, 1) ->
+    groups_from_list_2(Fun, ValueFun, lists:reverse(List), #{}).
```
bjorng, line 534: In order to support extended error information, the following changes are necessary:

```suggestion
groups_from_list(Fun, ValueFun, List0) when is_function(Fun, 1),
                                            is_function(ValueFun, 1) ->
    try lists:reverse(List0) of
        List ->
            groups_from_list_2(Fun, ValueFun, List, #{})
    catch
        error:_ ->
            badarg_with_info([Fun, ValueFun, List0])
    end;
groups_from_list(Fun, ValueFun, List) ->
    badarg_with_info([Fun, ValueFun, List]).
```

bjorng (changes requested):
Thanks for your pull request.

The `maps` module now supports extended error information. Since it is probably not immediately obvious, how to add extended error information, I have included the necessary changes as suggestion in my inline comments. In addition to those changes, you will also need to do the following changes:

```diff --git a/lib/stdlib/src/erl_stdlib_errors.erl b/lib/stdlib/src/erl_stdlib_errors.erl
index dc1152cb39..2306c37bf1 100644
--- a/lib/stdlib/src/erl_stdlib_errors.erl
+++ b/lib/stdlib/src/erl_stdlib_errors.erl
@@ -225,6 +225,10 @@ format_maps_error(get, [_Key,Map]) ->
         true ->
             [[],not_map]
     end;
+format_maps_error(groups_from_list, [Fun, List]) ->
+    [must_be_fun(Fun, 1), must_be_list(List)];
+format_maps_error(groups_from_list, [Fun1, Fun2, List]) ->
+    [must_be_fun(Fun1, 1), must_be_fun(Fun2, 1), must_be_list(List)];
 format_maps_error(get, [_,_,_]) ->
     [[],not_map];
 format_maps_error(intersect, [Map1, Map2]) ->


diff --git a/lib/stdlib/test/maps_SUITE.erl b/lib/stdlib/test/maps_SUITE.erl
index e4ee347add..2492130cf1 100644
--- a/lib/stdlib/test/maps_SUITE.erl
+++ b/lib/stdlib/test/maps_SUITE.erl
@@ -818,6 +818,13 @@ error_info(_Config) ->
          {get, [key, {no,map}]},
          {get, [key, {no,map}, default]},
 
+         {groups_from_list, [not_a_fun, []]},
+         {groups_from_list, [fun hd/1, not_a_list]},
+
+         {groups_from_list, [not_a_fun, fun(_) -> ok end, []]},
+         {groups_from_list, [fun(_) -> ok end, not_a_fun, []]},
+         {groups_from_list, [fun(_) -> ok end, fun(_) -> ok end, not_a_list]},
+
          {intersect, [#{a => b}, y]},
          {intersect, [x, #{a => b}]},
          {intersect, [x, y],[{1,".*"},{2,".*"}]},
```

bjorng (changes requested):
Looks good, except that there is still a very long line in the documentation.

lib/stdlib/doc/src/maps.xml
```diff
@@ -657,5 +657,37 @@ error</code>
 #{1337 => "value two"}</code>
       </desc>
     </func>
+
+    <func>
+      <name name="groups_from_list" arity="2" since="OTP 26.0"/>
+      <fsummary>Splits the list into groups using a function as discriminator.</fsummary>
+      <desc>
+        <p>The result is a map where each key is given by <anno>Fun</anno> and each value is a list of elements. The order of elements within each list is preserved from the list.</p>
```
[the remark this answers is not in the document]
meox: right! I did only for the other function. Just fixed

bjorng (approved):

bjorng:

bjorng:
Thanks for your pull request.

PR #5589, opened 2022-01-06
Remove the ERLANG_COMPILE_DATE timestamp from erl_version.h 

JeromeDeBretagne:
Do not generate the `ERLANG_COMPILE_DATE` timestamp in `erl_version.h` with the `make_version` script to ensure reproducible builds.

While preparing this PR, I have found that `mkver.c` seems totally unused in the builds, so the first commit removes this file.

Linked to #4417


JeromeDeBretagne:
I've just pushed a minor change, simply updating the Copyright header date to 2022 in the modified files.

PR #5591, opened 2022-01-08
[erts, tools] remove "emem" tool support

max-au:
Test suites for the tool were lost long ago.
Removing allocator wrappers support in a separate commit to squeeze a tiny bit of performance.


max-au:
- rebased to master to fix merge conflicts (there were typos fixed in lib/tools/c_src/erl_memory.c which I removed completely, and that caused merge conflict).

PR #5593, opened 2022-01-08
Compile time no longer available

JeromeDeBretagne:
Here is a PR inspired by https://github.com/erlang/otp/commit/b58478b9d3e2433ebb20d9af362f174333631bec which removed an occurence of the old `compile time` info.

While preparing #5589 I detected a few more similar leftover cases so I'm proposing to remove those as well to complete the clean-up. I'm not sure if this is a good candidate for `maint` or for `master` though.

### DIFF
--- a/lib/diameter/src/info/diameter_dbg.erl
+++ b/lib/diameter/src/info/diameter_dbg.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 2010-2016. All Rights Reserved.
+%% Copyright Ericsson AB 2010-2022. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


bmk:
Not sure when compile time was removed, but surely it was before 24?
If so, as far as megaco and snmp goes, I think I would like it in maint (24.3?).
Its not a biggy, so it depends what the others say.

JeromeDeBretagne:
> Not sure when compile time was removed, but surely it was before 24?


Indeed, it was before 24 as the previous cleanup commit I've found was from more than 2 years ago

bjorng:
The changes in the `c` module looks good to me and are safe to include in `maint`.

IngelaAndin:
inets changes are ok for maint too, @otp-anders what about diameter?

JeromeDeBretagne:
I've just pushed a minor change, simply updating the Copyright header date to 2022 in all 6 modified files.

bmk:
(my parts; megaco and snmp) Still looks ok to me.

bjorng:
Thanks for your pull request.

PR #5597, opened 2022-01-10
Add DEVELOPMENT.md and expand gh actions to run tests

garazdawi:
This PR does a lot of different things, but the major points are:

1. Polish and test the `make test` support in Erlang/OTP build system
2. Add [DEVELOPMENT.md](https://github.com/garazdawi/otp/blob/lukas/add-gh-actions-testing/HOWTO/DEVELOPMENT.md), a guide to how to develop and test using the Erlang/OTP build system
3. Run tests on Ubuntu Linux for all PRs opened towards the Erlang/OTP build repo, you can see an example of this in my private repo: https://github.com/garazdawi/otp/pull/16

I may change some minor details before merging (such as the location of the test logs), but most of the changes should be done.

Any feedback would be appreciated, especially on the contents of the new development guide.


make/test_target_script.sh
```diff
@@ -182,60 +174,129 @@ then
...
+
+# Check that we are running a released erlang when we have to
+if [ "$TEST_NEEDS_RELEASE" = "true" ]; then
+    MSG=$(cat <<EOF
+This application needs to be tested using a released Erlang/OTP.
+Erlang/OTP will now be released to ${RELEASE_ROOT}.
+EOF
+       )
+    print_highlighted_msg $YELLOW "${MSG}"
+    release_erlang "${RELEASE_ROOT}" > "${RELEASE_LOG}" 2>&1
```
max-au, line 207: Now I see that you implemented both modes. So I guess the only reason to keep the mode that runs test in the source tree is the slowness of `make release` itself? Because the build system is not smart enough to detect that no changes were made between test invocations and skip the release part.

max-au (commented):
A great start of the year! Thank you!

lib/kernel/src/inet_db.erl
```diff
@@ -498,7 +498,7 @@ res_check_option(nameservers, NSs) ->
 res_check_option(alt_nameservers, NSs) ->
     res_check_list(NSs, fun res_check_ns/1);
 res_check_option(domain, Dom) ->
-    inet_parse:visible_string(Dom);
+    Dom =:= "" orelse inet_parse:visible_string(Dom);
```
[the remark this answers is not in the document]
garazdawi: Yes, it is intentional. It is an API change of an internal function only used by testing.

lib/diameter/Makefile
```diff
@@ -28,5 +28,6 @@ SPECIAL_TARGETS =
 include $(ERL_TOP)/make/otp_subdir.mk
 
 DIA_PLT_APPS=ssl runtime_tools syntax_tools
+TEST_NEEDS_RELEASE=true
```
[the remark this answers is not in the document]
garazdawi: One of the nice things about using `ct_run` to run tests instead of `ts` is that we get all the features of ct without having to update ts to support them. So, if anything I would like to see us converge on using `ct_run` instead of `ts` if possible.

lib/crypto/configure
```diff
@@ -1492,7 +1492,7 @@ Some influential environment variables:
   CXX         C++ compiler
   CXXFLAGS    C++ compiler flags
   LD          linker (is often overridden by configure)
-  LDFLAGS     linker flags (can be risky to set since LD may be overridden by
+  LDFLAGS     linker flags (can be risky to set since LD may be overriden by
```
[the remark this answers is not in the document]
garazdawi: indeed, thanks.

make/test_target_script.sh
garazdawi: Yes, indeed. Since we use recursive make it is not really possible to make `make release` much faster. Today it takes about 10 seconds (on my machine), which when you just want to run a single testcase adds up to a frustrating amount of wait time...

lib/ssl/test/ssl_gh.spec
```diff
@@ -0,0 +1,19 @@
+% {merge_tests,false}.
+{alias,dir,"../ssl_test"}.
+
+{suites,dir,all}.
+{skip_suites,dir,[openssl_ocsp_SUITE],"Unstable testcases"}.
+{skip_groups,all_nodes, dir, openssl_renegotiate_SUITE, 'dtlsv1.2',
```
IngelaAndin, line 6: Line 6 can be removed once PR-#5576 is merged.

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -51,7 +51,7 @@ alternative() ->
 
 groups() ->
     [
-        {dist, [parallel], [errors, dist, peer_down_crash, peer_down_continue, peer_down_boot,
+        {dist, [], [errors, dist, peer_down_crash, peer_down_continue, peer_down_boot,
```
max-au, line 54: I wonder what exactly breaks when tests are running in parallel? Is it timetrap too tight (so multiple concurrent nodes are not able to connect in a timely manner)? Or is it some node name clash (which would be very surprising, as the while idea of `peer` is to avoid such cases).

One failure mode for `peer` may be caused by setting `ERL_FLAGS="-noshell`. As I discovered it makes the started peer node to hang indefinitely (because no code from `peer.erl` is ever involved, it is starting via undocumented `-user` parameter which simply does not take effect when ERL_FLAGS specify something else).

garazdawi: I'll revert the change and we can see if the gh test run fails :) I don't recall exactly what the error was.

garazdawi: So, I/we remembered/figured out again what the problem was. Due to the changes in #5610 the `dist_io_redirect` test case fails when run in parallel as a new warning is printed to the io captured there. @rickard-green will move just that test case out from the parallel run to fix the problem.

garazdawi:
This PR is getting close to being ready to merge. There will probably have to be some adjustments to how this works in the future, but it is a good start.

You can have a look at https://github.com/garazdawi/otp/pull/19 to see what the comment will look like now. I've added so that you don't have to download the test results and docs to view them, but instead they are uploaded to erlang.github.io.

Once #5581 is merged, I will merge this.

PR #5604, opened 2022-01-13
Eliminate confusing `case_clause` exception

### LINKED ISSUE #5513: Confusing `case_clause` exception instead of `function_clause`
**Describe the bug**

When a fun is used immediately, the compiler will inline it, and if no clause in the fun matches, there will be a `case_clause` error instead of a `function_clause` error.

**To Reproduce**

Consider this module:

```
-module(bug).
-export([foo/0]).

foo() ->
    fun(a) -> ok end(b).
```

Compile and run it:

```
1> c(bug).
bug.erl:5:5: Warning: no clause will ever match
%    5|     fun(a) -> ok end(b).
%     |     ^

{ok,bug}
2> bug:foo().
** exception error: no case clause matching {b}
     in function  bug:foo/0 (bug.erl, line 5)
```

**Expected behavior**

```
2> bug:foo().      
** exception error: no function clause matching bug:'-foo/0-fun-0-'(b) (bug.erl, line 5)
```

**Affected versions**

OTP 24, which introduced automatic inlining of funs that are immediately used and discarded.

**Additional context**

While not strictly a bug, this behavior is confusing.

bjorng:
Consider this module:

    -module(bug).
    -export([foo/0]).

    foo() ->
        fun(a) -> ok end(b).

The call to the fun will always fail, which will be noted by
the compiler:

    1> c(bug).
    bug.erl:5:5: Warning: no clause will ever match
    %    5|     fun(a) -> ok end(b).
    %     |     ^

    {ok,bug}

What is unexpected is that the exception that is raised is not
a `function_clause` exception:

    2> bug:foo().
    ** exception error: no case clause matching {b}
         in function  bug:foo/0 (bug.erl, line 5)

This confusing `case_clause` exception started to appear in OTP 24
because of 72675baaa9fd30 (#4545) that inlines funs that are
immediately used.

Before OTP 24, when the fun was not inlined, the exception would be:

    2> bug:foo().
    ** exception error: no function clause matching bug:'-foo/0-fun-0-'(b) (bug.erl, line 5)

The reason that `function_clause` exceptions are rewritten to
`case_clause` exceptions in inlined code is to avoid the even more
confusing and misleading exception:

    2> bug:foo().
    ** exception error: no function clause matching bug:foo(b) (bug.erl, line 5)

This is confusing because it seems that `foo/0` was called with one argument.

To reduce the confusion, this pull request ensures that `function_clause` exceptions in
inlined code remains `function_clause` exceptions, but with a generated name that
makes it clear that the `function_clause` exception occurred in a fun:

    2> bug:foo().
    ** exception error: no function clause matching bug:'-foo/0-inlined-0-'(b) (bug.erl, line 5)

Fixes #5513

### DIFF
--- a/lib/compiler/src/beam_ssa_pre_codegen.erl
+++ b/lib/compiler/src/beam_ssa_pre_codegen.erl
@@ -73,14 +73,16 @@
 
 -import(lists, [all/2,any/2,append/1,duplicate/2,
                 foldl/3,last/1,member/2,partition/2,
-                reverse/1,reverse/2,sort/1,splitwith/2,zip/2]).
+                reverse/1,reverse/2,seq/2,sort/1,
+                splitwith/2,usort/1,zip/2]).
 
 -spec module(beam_ssa:b_module(), [compile:option()]) ->
                     {'ok',beam_ssa:b_module()}.
 
 module(#b_module{body=Fs0}=Module, Opts) ->
     Ps = passes(Opts),
-    Fs = functions(Fs0, Ps),
+    Fs1 = functions(Fs0, Ps),
+    Fs = fc_stubs(Fs1, Module),
     {ok,Module#b_module{body=Fs}}.
 
 functions([F|Fs], Ps) ->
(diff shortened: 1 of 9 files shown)


lib/compiler/src/beam_validator.erl
```diff
@@ -243,9 +243,15 @@ build_function_table([{function,Name,Arity,Entry,Code0}|Fs], Acc) ->
...
-                      parameter_info => find_parameter_info(Is, #{}) },
+            Info0 = #{ name => Name,
+                       arity => Arity,
+                       parameter_info => find_parameter_info(Is, #{}) },
+            Info = case Is of
+                       [{jump,_}|_] ->
+                           Info0#{always_fails => true};
+                       _ ->
+                           Info0
+                   end,
```
jhogberg, line 254: Nitpick: I prefer having these properties explicit, maybe break this out into another function like `find_parameter_info/2`?

```suggestion
                       parameter_info => find_parameter_info(Is, #{}),
                       always_fails => find_always_fails(Is) },
```

lib/compiler/src/beam_ssa_pre_codegen.erl
```diff
@@ -860,34 +864,29 @@ match_fail_instrs_blk([#b_set{op=call,anno=Anno,
...
 match_fail_instrs_blk([I|Is], Arity, Acc) ->
     match_fail_instrs_blk(Is, Arity, [I|Acc]);
 match_fail_instrs_blk(_, _, _) ->
     none.
 
-match_fail_instr(Call, Tag, Val, Is, Acc) ->
+match_fail_instr(Call, function_clause, Values, Is, Acc) ->
+    case beam_ssa:get_anno(inlined, Call, none) of
+        none ->
+            %% If there is no `inlined` annotatation, it implies that
```
jhogberg, line 877:
```suggestion
            %% If there is no `inlined` annotation, it implies that
```

lib/compiler/src/v3_kernel.erl
```diff
@@ -405,36 +405,51 @@ letrec_goto([{#c_var{name={Label,0}},Cfail}], Cb, Sub0,
...
         false ->
             %% The arguments in the `function_clause` exception don't
-            %% match the arguments for the current function because
-            %% of inlining. Keeping the `function_clause`
-            %% exception reason would be confusing. Rewrite it to
-            %% a `case_clause` exception with the arguments in a
-            %% tuple.
-	    [cerl:c_tuple([#c_literal{val=case_clause},
-                           cerl:c_tuple(As)])]
+            %% match the arguments for the current function because of
```
jhogberg, line 434: Nitpick:
```suggestion
            %% Match the arguments for the current function because of
```

jhogberg (approved):

PR #5607, opened 2022-01-14
public_key,ssl: add password fun for decoding keyfiles

u3s:
Implementing password fun option for ssl connect API

This change allows password option to accept as a value
either a string or a fun. Added specific tests and
documentation.

The motivation for this change is to better protect a
private key. If the private key is protected by password the
password may be retreived using user supplied function therefore
enabling storing password in a secure vault or something similar.

This replaces #5320


janciesla8818:
@u3s thanks for the effort and getting this one merged. Do we how which releases will include this change?

u3s:
no problem. currently it is planned to be included 24.3.

PR #5610, opened 2022-01-17
Prevent global inconsistency by preventing overlapping partitions

rickard-green:
The state of `global` could become inconsistent when connections between nodes where lost in such a way that nodes became connected in overlapping partitions. This inconsistency could remain indefinitely even after connections had been restored so that the original fully connected network of nodes had been formed again. In order to prevent this from happening `global` will now forcibly disconnect nodes that have lost connections to other nodes, so that the remaining connected nodes will form a smaller fully connected network.

One such scenario is described on the erlang questions mailing list: https://erlang.org/pipermail/erlang-questions/2020-October/100034.html


rickard-green:
Replaced by #5611

PR #5611, opened 2022-01-17
Prevent global inconsistency by preventing overlapping partitions

rickard-green:
By default `global` does *not* take any actions to restore a fully connected network when connections are lost due to network issues. This is problematic for all applications expecting a fully connected network to be provided, such as for example `mnesia`, but also for `global` itself. A network of overlapping partitions might cause the internal state of `global` to become inconsistent. Such an inconsistency can remain even after such partitions have been brought together to form a fully connected network again. The effect on other applications that expects that a fully connected network is maintained may vary, but they might misbehave in very subtle hard to detect ways during such a partitioning.

In order to prevent such issues, we have introduced a *prevent overlapping partitions* fix which can be enabled using the [`prevent_overlapping_partitions`](https://erlang.github.io/prs/5611/lib/kernel-8.2/doc/html/kernel_app.html#prevent_overlapping_partitions) kernel(6) parameter. When this fix has been enabled, `global` will actively disconnect from nodes that reports that they have lost connections to other nodes. This will cause fully connected partitions to form instead of leaving the network in a state with overlapping partitions. Note that this fix *has* to be enabled on *all* nodes in the network in order to work properly. Since this quite substantially changes the behavior, this fix is currently disabled by default. Since you might get hard to detect issues without this fix you are, however, *strongly* advised to enable this fix in order to avoid issues such as the ones described above. As of OTP 25 this fix  will become enabled by default.

One example of a `global` internal insconsistency is described on the erlang questions mailing list: https://erlang.org/pipermail/erlang-questions/2020-October/100034.html

#5687 contains this PR merged to the latest master (as of 2022-02-07) plus enabling of the kernel parameter `prevent_overlapping_partitions` by default.


lib/kernel/src/global.erl
```diff
@@ -473,6 +493,12 @@ init([]) ->
                trace = T0,
                the_registrar = start_the_registrar(),
                connect_all = Ca},
+    _ = rand:seed(default,
+                  (erlang:monotonic_time(nanosecond) rem 1000000000)
+                  + (erlang:system_time(nanosecond) rem 1000000000)),
+    CreX = ((rand:uniform(?MAX_64BIT_SMALL_INT - ?MIN_64BIT_SMALL_INT)
+             - 1) band (bnot ((1 bsl 32) -1))),
+    put(creation_extention, CreX),
```
jhogberg, line 501: Nitpick:
```suggestion
    put(creation_extension, CreX),
```

lib/kernel/src/global.erl
```diff
@@ -2070,15 +2189,124 @@ pid_locks(Ref) ->
...
+    end,
+    NewS#state{known = maps:remove(Node, S#state.known),
                synced = lists:delete(Node, Syncs)}.
 
+inform_connection_loss(_Node, #state{connect_all = false}) ->
+    ok;
+inform_connection_loss(Node, #state{connect_all = true} = S) ->
+    Msg = {lost_connection,
+           node(),
+           (get(creation_extention)
```
jhogberg, line 2217:
```suggestion
           (get(creation_extension)
```

lib/kernel/src/global.erl
```diff
@@ -2070,15 +2189,124 @@ pid_locks(Ref) ->
...
+                  (N) when is_atom(N) ->
+                      {N, node_vsn(N, S)};
+                  ({N, V} = NV) when is_atom(N),
+                                     is_integer(V) ->
+                      NV
+              end, NList).
+                          
+
+mk_known_list(Vsn, #state{known = Known}) when Vsn < 7 ->
+    lists:map(fun ({N, _V}) -> N end, maps:to_list(Known));
```
jhogberg, line 2280:
```suggestion
    maps:keys(Known);
```

erts/doc/src/erl_cmd.xml
```diff
@@ -567,6 +567,19 @@ $ <input>erl \
         <p>Replaces the path specified in the boot script; see
           <seefile marker="sasl:script"><c>script(4)</c></seefile>.</p>
       </item>
+      <tag><marker id="prevent_overlapping_partitions"/>
```
max-au, line 570: Would it be possible to use kernel application configuration (e.g. `-kernel prevent_overlapping_partitions true|false`) instead of a command line switch?
Unlike sys.config, command lines are harder to read and write correctly, and more difficult to test.

rickard-green: Yes. I just mimicked `-connect_all false` without thinking more about it.

max-au:
I attempted to include this PR in more tests of existing software. It fails a large amount of tests where distributed applications are involved, because of the change in the default behaviour, and the fact that by default `global` is enabled.

I wonder if there is a chance of keeping the compatibility behaviour by default, and allowing to explicitly enable `prevent_overlapping_partitions` for those who need this issue fixed?

rickard-green:
We will release this (with some modifications) as a patch on OTP 24, and then it will be disabled by default due to the change in behaviour. The intent is to have it enabled by default in OTP 25 since it will be present in the initial version where users expect larger changes to be made. To me it feels *really weird* having a bugfix disabled by default especially in the first version of a new release. I would rather disable `global` as such as default before disabling the bugfix as default (note that I'm not saying that we will do that).

Regardless, even if we were to have it disabled by default, all test cases in OTP have to be able to handle this fix being enabled. Otherwise, we would not be able to run test with it enabled. Testcases not depending on `global`, and not depending on any other applications depending on fully connected networks, may of course disable `global` using `-connect_all false`.

max-au:
Having `global` disabled by default sounds even better to me. Although it is also a significant behaviour change.

What I am really concerned about is `global` disconnecting nodes based on local information about the cluster state. This may be a subject to various race conditions, similar to all kinds of those I had with `pg`.
My worst nightmare would be cascading disconnects, e.g. when a node in the cluster loses connection to another, and then all other nodes of the cluster also decide to disconnect the node that had a small hiccup.

I wonder if there is any other solution that does not involve disconnection? Apart from https://erlang.org/pipermail/erlang-questions/2020-October/100034.html are there any other manifestations of this bug? I fear that the bugfix itself will cause more churn than leaving the bug there (or solving it differently, if you're open for discussions on that).

okeuday:
@rickard-green I would have expected something similar to a kernel app configuration parameter like `dist_auto_disconnect` (like [dist_auto_connect](https://www.erlang.org/doc/man/kernel_app#configuration)).  That would make it easier to disable for obtaining the desired behavior among specific source code.  That should also make misconfiguration less likely.  People would be making the decision based on their source code's use of global which can be separate from how the source code is executed.

rickard-green:
Copies of conversations in #5417 related to this PR


@rickard-green wrote:
> @max-au wrote:
> 
> > I verified that both workarounds keep tests passing, and can make any desired change. But to be honest I'm not sure whether global can make such decisions locally and disconnect nodes at will.
> 
> One of `global`'s main tasks is to maintain a fully connected network. Other applications such as `mnesia` relies on this. Currently `global` fails at providing this service in the presence of network failures. The only way *) to provide this service when connections are lost, is to remove other connections so you end up with fully connected partitions.
> 
> If `global` is disabled (`-connect_all false`), it won't provide a fully connected network and will also not take down connections like this.
> 
> Since this is quite a big change in behaviour compared to before, it will also be possible to disable this bugfix (`-prevent_overlapping_partitions false`), but then you might get intricate issues with applications relying on the fully connected network service being provided.
> 
> *) This is not completely true. You could also provide this service by implementing "virtual connections" routing signals over other connected nodes. We would, however, not be able to deliver such a solution in the near future. That is, that can only be a long term option. Note that I'm not saying whether or not "virtual connections" will be implemented, just that it could be an option for the future.

rickard-green:
@okeuday wrote:
> @rickard-green I would have expected something similar to a kernel app configuration parameter like `dist_auto_disconnect` (like [dist_auto_connect](https://www.erlang.org/doc/man/kernel_app#configuration)). That would make it easier to disable for obtaining the desired behavior among specific source code. That should also make misconfiguration less likely. People would be making the decision based on their source code's use of global which can be separate from how the source code is executed.

Yes, I've changed it to a kernel parameter, I will push that change soon. I've not changed `-connect_all` to a kernel parameter in this branch which I think we should do too, but I'll make that change in a separate branch.

rickard-green:
Copies of conversations in #5654 related to this PR

@rickard-green wrote:

rickard-green:
> Having `global` disabled by default sounds even better to me. Although it is also a significant behaviour change.
> 

I'm personally not that attached to `global`, but I think we probably not should disable `global` by default. However, compared to having a broken `global` by default I think it would be better.

> What I am really concerned about is `global` disconnecting nodes based on local information about the cluster state. This may be a subject to various race conditions, similar to all kinds of those I had with `pg`. My worst nightmare would be cascading disconnects, e.g. when a node in the cluster loses connection to another, and then all other nodes of the cluster also decide to disconnect the node that had a small hiccup.
> 
> I wonder if there is any other solution that does not involve disconnection? Apart from https://erlang.org/pipermail/erlang-questions/2020-October/100034.html are there any other manifestations of this bug? I fear that the bugfix itself will cause more churn than leaving the bug there (or solving it differently, if you're open for discussions on that).

There are unfortunately (as I wrote in the [comment](https://github.com/erlang/otp/pull/5611#issuecomment-1030956057) from #5417 above) only two alternatives if `global` (or anything else) are to provide a fully connected mesh, either take down connections or keep them up by other means.

I worked on another solution, simulating a `nodedown`/`nodeup` pair internally in `global` without actually taking down any connections, when I viewed this as a `global` internal issue. This should prevent global from getting internal inconsistencies. While working with that I realized that such a solution would just make things worse for other applications, such as `mnesia`, that rely on global providing a fully connected mesh, so I dropped that solution. This since `global` would then just hide information about these issues taking place preventing such applications from taking appropriate actions.

rickard-green:
@max-au #5687 contains this PR merged to the latest master plus enabling of the kernel parameter `prevent_overlapping_partitions` by default.

rickard-green:
@max-au

We've had an OTP Technical Board meeting discussing defaults for `global`. The decisions made were that `global` will remain enabled by default and as of OTP 25 the `prevent_overlapping_partitions` fix will also be enabled by default.

Regarding the issues we've seen with these defaults together with `peer`. `peer:stop()` will by default stop nodes by calling `erlang:disconnect_node(PeerNode)`. From `global`'s perspective this is problematic since `global` on `PeerNode` will react to this and potentially cause even more connections to be dropped. This problem is however not `global` specific. Other (current and/or future) applications executing on `PeerNode` may also react on the connection loss in undesirable ways. That is, `peer:stop()` at least needs to try to halt the node properly before removing the connection. It can as a last resort disconnect from it if it is unresponsive, though. I've made a pull request #5705  implementing this and fixing some other issues. @bmk and I tested a simpler version of this which seems to fix all `snmp` test issues. That version of this fix had other issues, though.

@max-au @okeuday

Regarding kernel parameters. In this PR there is now the `prevent_overlapping_partitions` kernel parameter. If the `-connect_all` switch is introduced as a kernel parameter as well, which I think should be named `global` since it enables/disables `global`, we get yet another kernel parameter for `global`. Do you think it is better with a `global` kernel parameter with a value containing a list of two tuples like: `[{enable, true | false}, {prevent_overlapping_partitions, true | false}]` or are two separate boolean parameters `global` and `prevent_overlapping_partitions` better? Or do you have any other suggestions?

okeuday:
@rickard-green My concern regarding the `prevent_overlapping_partitions` kernel parameter was focused on the ability to disable it in a config file to ensure `global` doesn't attempt to disconnect nodes.  I don't think I would disable `global`, though I avoid its use.  My view is that overlapping partitions should be able to exist without getting resolved, due to use of a separate process group (e.g., https://github.com/okeuday/cpg/ handles both `hidden` and `visible` node connections, so any network topology is allowed).  If `global` use blocked infinitely due to overlapping partitions, I would still consider that better than automatically disconnecting nodes (for my use), though I understand the `prevent_overlapping_partitions` feature is important for `global` use.

A `global` kernel parameter with a value of `[{enable, true | false}, {prevent_overlapping_partitions, true | false}]` is a good approach.  An alternative I was thinking about was `dist_auto_disconnect` with a value of `never | global` (possibly other values in the future) instead of a `prevent_overlapping_partitions` value.  The name `dist_auto_disconnect` makes the potential to cause node disconnects clearer (to avoid the reaction of "oh no, where did my nodes go, why is this happening, ahhhh!", the logging output may already make the circumstance clear but I would assume people would want to be aware of the potential disconnects in advance of the occurrence(s)).

A way to disable `global` with a kernel parameter is important and using a `global` parameter name for that with a list of options is the best solution.

max-au:
@rickard-green thank you for the comprehensive update!
As I commented in #5705, it was a design choice to have `brutal_kill` acting brutally. It is however questionable whether it should be the default mode or not (before #5611, I think it was a good choice as it sped up all tests, but now it can be reconsidered). If you believe that `peer` shutdown should be graceful by default, I won't object (but please leave the choice to have brutally fast stop for tests that want that).

As for `global` kernel parameter, I prefer two separate parameters, like `-kernel global_enabled true` and `-kernel prevent_overlapping_partitions true`. The reason is, if I ever need to override that via command line, it is simple to do with separate parameters. Figuring out bash/zsh/...sh escaping to pass `[{enable, true | false}, {prevent_overlapping_partitions, true | false}]` scares me off.

I also agree with @okeuday that a name clearly designating the actual action (disconnecting otherwise perfectly fine nodes) should have `disconnect` (and probably `force`) in its name. To me "prevent" means "do not allow to connect" which isn't exactly true, those nodes would successfully connect and disconnect in a rapid succession (e.g. "blink", go up/down/up/down).

rickard-green:
@max-au @okeuday

We are much closer to the first release candidate than I thought, and this needs to be present in it, so it is to late to change naming of parameters before the merge. The naming of the parameters and format of their values can, however, still be changed up until we've released it in a non release candidate version. We'll have to have a discussion about it internally here at OTP as well once the release candidate is out.

PR #5614, opened 2022-01-17
stdlib: normalize empty port in uri_string

u3s:
- adjust port normalize functionality
- remove port subcomponent if port value is empty or key in URIMap if undefined

### DIFF
--- a/lib/inets/test/httpc_SUITE.erl
+++ b/lib/inets/test/httpc_SUITE.erl
@@ -258,8 +258,8 @@ init_per_group(http_unix_socket = Group, Config0) ->
             file:delete(?UNIX_SOCKET),
             start_apps(Group),
             Config = proplists:delete(port, Config0),
-            Port = server_start(Group, server_config(Group, Config)),
-            [{port, Port} | Config]
+            {Pid, Port} = server_start(Group, server_config(Group, Config)),
+            lists:append([{dummy_server_pid, Pid}, {port, Port}], Config)
     end;
 init_per_group(http_ipv6 = Group, Config0) ->
     case is_ipv6_supported() of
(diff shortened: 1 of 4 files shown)


lib/stdlib/src/uri_string.erl
```diff
@@ -2387,6 +2389,15 @@ normalize_port(Map, Port, Default) ->
...
 
 
+normalize_undefined_port(Map) ->
+    case maps:get(port, Map, not_present) of
+        undefined ->
+            maps:remove(port, Map);
+        _Else ->
+            Map
+    end.
+
```
peterdmv, line 2399: I believe this function can be further simplified like this:
```erlang
normalize_undefined_port(#{port := undefined} = Map) ->
    maps:remove(port, Map);
normalize_undefined_port(#{} = Map) ->
    Map.
```

lib/inets/test/httpc_SUITE.erl
```diff
@@ -56,8 +56,10 @@ all() ->
      {group, http_ipv6},
      {group, sim_http},
      {group, http_internal},
-     {group, http_unix_socket},
      {group, https},
+     %% http_unix_socket group creates a persistent connection session in
+     %% httpc_manager for localhost:80 which may affect tests exected afterwards
+     {group, http_unix_socket},
```
peterdmv, line 62: This change probably belongs to another branch.

u3s: It is related in a way that `httpc_SUITE:undefined_port` started failing after adjusting normalize behavior in uri_string.
Another question could be: what `httpc_SUITE:undefined_port` is supposed to test at all and if it is formulated properly.

I will have a look once again.

PR #5618, opened 2022-01-18
Cons/eep/experimental features

lisztspace:


erts/test/lux/input/feature_directive_maybe_ifn.erl
```diff
@@ -0,0 +1,87 @@
+%% This module knows about the both features ifn and maybe
+%% These atoms are thus quoted
+
+-module(feature_directive_maybe_ifn).
+
+-export([foo/0,
+	 bar/0,
```
u3s, line 7: for some reason your exports contain tab characters (other files as well), while spaces are used in other places ...

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+    case init:get_argument('enable-feature') of
+        error ->
+            %% no features enabled
+            ok;
+        {ok, Ftrs} ->
+            %% FIXME Need to be paranoid about any failures here.
+            %% Only convert to an existing atom.  This will also catch
+            %% a too long atom.
+            F = fun(String) ->
+                        case catch list_to_existing_atom(String) of
```
bjorng, line 234: I suggest using try/catch. We try to avoid the old-fashioned catch in new code (except test suites).

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+          %% FIXME Should this rather be named dependencies?
+          options := list()
+         }.
+feature_info(ifn_expr) ->
+    #{description =>
+          "Inclusion of expression `ifn cond -> body end`, which "
+      "evaluates `body` when cond is false.  This is a truly "
+      "experimental feature, present only to show and use the "
+      "support for experimental features.  Not extensively tested.  "
+      "Implementated by a transofmration in the parser.",
```
bjorng, line 88:
```suggestion
      "Implementated by a transformation in the parser.",
```

lib/kernel/src/code_server.erl
```diff
@@ -1163,7 +1175,41 @@ load_file_1(Mod, From, St) ->
 	error ->
 	    {reply,{error,nofile},St};
 	{Mod,Binary,File} ->
-	    try_load_module_1(File, Mod, Binary, From, St)
+            case is_loaded(features, St#state.moddb) of
+                false ->
+                    Allowed = true,
+                    persistent_term:put(Mod, no_features);
```
bjorng, line 1181: This use of `persistent_term` can clash with the module's own use of persistent terms, since the documentation has the following recommendation:

> It is recommended to use keys like `?MODULE` or `{?MODULE,SubKey}` to avoid name collisions.

I suggest using an ETS table instead. (Not sure whether the code_sever should own it or if there should be a `features` process.)

lib/stdlib/src/Makefile
```diff
@@ -78,6 +78,7 @@ MODULES= \
 	escript \
 	ets \
 	eval_bits \
+	features \
```
bjorng, line 81: You will need to update `stdlib.app.src` as well.

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+      Description :: string(),
+      FeatureInfoMap ::
+        #{description := Description,
+          type := type(),
+          status := status(),
+          %% List of additional compiler options to be given to
+          %% activate this feature.  Useful when there is a dependency
+          %% to other features.
+          %% FIXME Should this rather be named dependencies?
+          options := list()
```
IngelaAndin, line 80: If it is compiler options only maybe it should be called erlc_options or compiler_options?! Or do we think there could be other types of options in which case options name is very addition friendly.  Calling it dependencies makes me think what kind of dependcies?

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+
+enabled_features() ->
+    persistent_term:get(enabled_features).
+
+reserved_words() ->
+    persistent_term:get(reserved_words).
+
+%% Return features used by module or beam file
+features_used(Module) ->
+    FileName =
```
IngelaAndin, line 286:
```suggestion
    FileName = module_to_filname(Module),
```

lib/stdlib/src/features.erl
bjorng: How will this list of options be used? Will the options be automatically used when the feature is enabled or is it informational?

lib/stdlib/src/features.erl
IngelaAndin, line 286:
```
module_to_filename(Module) when is_list(Module) ->
    Module;
module_to_filename(Module) ->
    case code:is_loaded(Module) of
        false ->
	    case code:get_object_code(Module) of
		error ->
		    not_found;
		{_Mod, _Bin, FName} ->
		    FName
	    end;
	{file, FName} ->
	    FName
    end.
```

IngelaAndin: Where the implementation of  module_to_filname is as shown below. I think it makes the code easier to read. Your code was not horribly v-shaped (as some code is) and it might be a little subjective, so this is my opinion for what it is worth.

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%
+%% %CopyrightEnd%
+%%
+-module(features).
```
michalmuskala, line 20: `features` is a fairly "common" word and it might easily clash with existing code. Additionally this module is not "crucial" and I'd expect it won't be used much in user code. Could this follow naming convention similar to other compiler-related modules with `erl_features` - following `erl_parse`, `erl_scan`, and friends.

lisztspace: Good point.  We'll discuss a better, more "internal" name for the module.

lib/kernel/src/code_server.erl
```diff
@@ -1104,6 +1104,18 @@ load_abs(File, Mod, From, St) ->
     FileName = absname(FileName0),
     case erl_prim_loader:get_file(FileName) of
 	{ok,Bin,_} ->
+            %% FIXME This is probably not needed!
```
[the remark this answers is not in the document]
lisztspace: The comment refers to the whole chunk, which is, indeed, not needed and will be removed.

lib/kernel/src/code_server.erl
lisztspace: The use of `persistent_term` is just old debug code, which will be removed.

erts/test/lux/input/feature_directive_maybe_ifn.erl
lisztspace: Strange.  I'll take a look at this and correct it.

lib/stdlib/src/features.erl
lisztspace: Thanks, but I'll rewrite the code anyway using the fact that the chunk can be extracted from the binary returned by `code:get_object_code/1`, which works for a module in the load path regardless of whether it is loaded (or allowed to be loaded).

lib/stdlib/src/features.erl
u3s: I agree.

IMHO, from shell experience perspective but also current EEP description `erl_features:F` is nicer than `features:F` because EEP is specific that it deals with "Erlang language features"

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+
+-export([features/0,
```
u3s, line 22: why `features:feature_info` and not `features:info` ?
why `features:enable_feature` and not `features:enable`

does word repetition serve any purpose here?

lib/stdlib/src/features.erl
```diff
@@ -0,0 +1,308 @@
...
+         feature_info/1,
+         enabled_features/0,
+         is_valid_feature/1,
+         reserved_words/0,
+         reserved_words/1,
+         resword_add_feature/2,
```
u3s, line 28: is resword a short name for reserved_word? it was not obvious for me at first.

- if yes, then maybe could be used in all function names (e.g. `reswords` rather than `reserved_words`) ?
- I'm wondering if there is a better short name than resword ... I thought about keyword, which might be more obvious, but I'm not sure if they're equivalent in this context. are they?

lib/stdlib/src/features.erl
lisztspace: Naming is difficult.  These will change to avoid the repetition.

lib/stdlib/src/features.erl
lisztspace: Yes, _keyword_ is both shorter and less cumbersome.  In the code both terms are used for the same concept, so using _keyword_ shouldn't be too controversial.  I'll do some archeology on terms.

lib/compiler/Makefile
```diff
@@ -39,3 +39,9 @@ include $(ERL_TOP)/make/otp_subdir.mk
 DIA_PLT_APPS=crypto
 
 include $(ERL_TOP)/make/app_targets.mk
+
+# Enable feature maybe_expr in runtime when running tests
+# Note: this overrides, and uses, test target in app_targets.mk
+.PHONY: test
+test:
+	$(MAKE) ERL_ARGS="-enable-feature maybe_expr" -f $(ERL_TOP)/make/app_targets.mk $@
```
garazdawi, line 47:
```suggestion
	$(MAKE) ERL_ARGS="-enable-feature maybe_expr $(ERL_ARGS)" -f $(ERL_TOP)/make/app_targets.mk $@
```

lib/stdlib/Makefile
```diff
@@ -40,3 +40,11 @@ DIA_PLT_APPS=compiler crypto
 TEST_NEEDS_RELEASE=true
 
 include $(ERL_TOP)/make/app_targets.mk
+
+# Enable feature maybe_expr in runtime when running tests.
+# This is very non local and somewhat ugly.
+.PHONY: test
+test:
+	$(MAKE) TEST_NEEDS_RELEASE=$(TEST_NEEDS_RELEASE) \
+		ERL_ARGS="-enable-feature maybe_expr"    \
```
garazdawi, line 49:
```suggestion
		ERL_ARGS="-enable-feature maybe_expr $(ERL_ARGS)"    \
```

garazdawi (commented):
In order for [the supported way of running tests in CI](https://github.com/erlang/otp/blob/master/HOWTO/INSTALL.md#testing) to work, you need to update ts_run to enable the maybe feature for test suites. I suggest adding the `-enable-feature maybe_expr` to [ts_run:make_command/1](https://github.com/erlang/otp/blob/master/lib/common_test/test_server/ts_run.erl#L262).

PR #5621, opened 2022-01-18
Add filelib:ensure_path/1

starbelly:
This commit adds `ensure_path/1` to `filelib`. Unlike `ensure_dir/1` this will always create the last path segment, `mkdir -p` style. 

In the wild you will find lots of implementations of this function, mostly using `ensure_dir/1` for a side effect to produce this behavior. Specifically, `filelib:ensure_dir("/some/path/" ++ "some_file_i_will_not_actually_create")`. 

I have left two implementations in the source code, one which is completely unrelated to `ensure_dir/1` and one which makes use of `ensure_dir/1` (this one is commented out).  I don't have strong feelings as to which implementation should be favored, but if it were up to me I'd go with the uncommitted version. It seems like duplicate code, but as mentioned, the functionality is different. 

I will glady add tests and docs if OTP would accept this PR. 

Edit:

Note I did do a version of `filelib:ensure_dir/2`, but I felt that made for a very strange API and would lead to confusing documentation. 

Edit:

We could also simply call `ensure_dir(filename:join(Path, "fake_file")))` but IMO it feels odd, but I would not be against it either. Likewise, an alternative name for the function `ensure_path/1` could be `mkdir_p/1`

### DIFF
--- a/lib/stdlib/doc/src/filelib.xml
+++ b/lib/stdlib/doc/src/filelib.xml
@@ -108,6 +108,20 @@
     </func>
 
     <func>
+      <name name="ensure_path" arity="1" since=""/>
+      <fsummary>Ensure that all parent directories for a target directory exist.</fsummary>
+      <desc>
+        <p>Ensures that all parent directories for the specified path <c><anno>Path</anno></c>
+          exist, trying to create them if necessary.</p>
+	    <p>Unlike <seemfa marker="#ensure_dir/1"><c>ensure_dir/1</c></seemfa>, this function will attempt
+        to create all path segments as a directory, including the last segment.</p>
+        <p>Returns <c>ok</c> if all parent directories already exist
+          or can be created. Returns <c>{error, <anno>Reason</anno>}</c> if
+          some parent directory does not exist and cannot be created.</p>
+      </desc>
+    </func>
+
+    <func>
       <name name="file_size" arity="1" since=""/>
       <fsummary>Return the size in bytes of a file.</fsummary>
       <desc>
--- a/lib/stdlib/src/filelib.erl
+++ b/lib/stdlib/src/filelib.erl
@@ -21,7 +21,7 @@
 
 %% File utilities.
 -export([wildcard/1, wildcard/2, is_dir/1, is_file/1, is_regular/1]).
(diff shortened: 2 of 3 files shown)


peerst:
I like the name `ensure_path`

max-au:
Seems a worthy addition to me, but why the implementation is more complex than just `ensure_dir(filename:join(Path, "x")))`?

starbelly:
> Seems a worthy addition to me, but why the implementation is more complex than just `ensure_dir(filename:join(Path, "x")))`?


My rationale was getting away from using `ensure_dir/1` that way anymore, but it's also about scratching a common itch as well, so I'd be fine changing it to that too.

josevalim:
Probably the opposite makes sense: change `ensure_dir` to directly use this new function.

josevalim:
What i mean is that `ensure_dir` internally could use this function. Something like this: `ensure_dir(F) -> ensure_path(filename:dirname(F)).` There are probably some caveats though. :)

starbelly:
> What i mean is that `ensure_dir` internally could use this function. Something like this: `ensure_dir(F) -> ensure_path(filename:dirname(F)).` There are probably some caveats though. :)

That would work, can't think of any caveats atm since ensure_dir already assumes the last segment is a regular file.

max-au:
Either way sounds good (also, I'd think of a testcase or even a bunch here).

starbelly:
> Either way sounds good (also, I'd think of a testcase or even a bunch here).

Just to re-iterate. I'll add test cases and docs once OTP team gives a 👍

bjorng:
Looks good! Please add test cases and documentation, and squash to a single commit.

starbelly:
> Looks good! Please add test cases and documentation, and squash to a single commit.

Will do at some point this week. Thanks!

starbelly:
@bjorng  This should be good to go I believe. Happy to refactor tests or update docs if desired. I do think we should consider deprecating ensure_dir/1, but I'll open a separate issue for that post-merge.

bjorng:

bjorng:
I amended your commit to fix a few cosmetic issues, such as incorrect indentation of exported functions in `filelib_SUITE`, and then merged to `master`.

Thanks for your pull request!

PR #5623, opened 2022-01-19
Dynamic node name improvements

sverker:
- Let `erlang:is_alive/0` return true when dynamic node name not yet received.
- Add `net_kernel:get_state`.

Edit: set_cookie fix broken out to #5670 for maint branch.
Edit2: Documentation improvements broken out and merged to maint.


lib/kernel/src/net_kernel.erl
```diff
@@ -707,6 +721,25 @@ handle_call({getopts, Node, Opts}, From, State) ->
...
+              end,
+    {NameType,Name} = case {persistent_term:get(net_kernel, undefined), node()} of
+                          {undefined, Node} ->
+                              {static, Node};
+                          {dynamic_node_name, nonode@nohost} ->
+                              {dynamic, undefined};
+                          {dynamic_node_name, Node} ->
+                              {dynamic, Node}
+                      end,
+    Return = #{started => Started,
```
garazdawi, line 738:
```suggestion
    Return = #{start_type => Started,
```
maybe?

lib/kernel/src/net_kernel.erl
```diff
@@ -707,6 +721,25 @@ handle_call({getopts, Node, Opts}, From, State) ->
...
+               name_type => NameType,
```
garazdawi, line 739: Can we add a field for long/short domain name? possibly called `domain_type` that has values `short | long`?

lib/kernel/src/net_kernel.erl
```diff
@@ -190,13 +192,25 @@ longnames() ->                 request(longnames).
 
 nodename() ->                  request(nodename).
 
+is_alive() ->
+    whereis(net_kernel) =/= undefined.
+
+get_state() ->
```
garazdawi, line 198: I think that this function would be useful to use, so it should be documented.

lib/kernel/src/net_kernel.erl
sverker: done

PR #5628, opened 2022-01-21
ssl: Do not exclude legacy suite

IngelaAndin:
Before TLS-1.3 cipher suites from previous versions are inclusive.
No rsa suites are included as default values so this function
should list all still possible rsa suites.


samwar (commented):
👍🏻

PR #5631, opened 2022-01-21
Pass Anno to ExternalFunctionHandler in erl_eval

### LINKED ISSUE #5526: Provide better stacktraces for errors in erl_eval
Today, if there is an error on the expression being evaluated, there is very little information about where said exception happened and I would like to improve this.

A simple way we could achieve this is to wrap each external call in a try/catch and reraise any exception with the additional stacktrace empty. However, I have some concerns with this approach:

  1. `erl_eval` currently does not have the file, which makes it impossible to assemble any kind of stacktrace.

  2. Wrapping all external function applications in try/catch may make code that does not care about the stacktrace slower.

  3. `erl_eval` still raises its own exceptions (case clauses errors, match errors, etc) that are outside of the external function handler.

We could add the `File` as an argument to `erl_eval`, and maybe additional flags to control how the stacktrace would look like, but it may be simpler to allow users to control this behaviour instead. This could be done with two relatively small changes:

  1. Allow the `ExternalFunction` to be a function of arity 4, in the shape of `Fun(Ann, Mod, Fun, Args)`. This way, as a user of `erl_eval`, I can pass my own `ExternalFunction` and add the relevant try/catch. `Ann` is the `erl_anno:anno()` of the node from where I can access additional information such as line in case I need to build a stacktrace.

  2. Change `erl_eval` to dispatch all of its internal errors (case clause error, match error, etc) through the external handler too (which, per the item above, will also include `Ann`). In other words, instead of `erlang:error(error, Reason, ?STACKTRACE)`, we would do something similar to: `do_apply(Ann, erlang, error, [error, Reason, ?STACKTRACE], Ef, RBn, ...)`.

With those two changes, I should be able to add my stacktrace with a simple function external function as this:

```erlang
File = "this/file/comes/from/somewhere",

ExternalFunction = fun(Ann, Mod, Fun, Args) ->
  try
    apply(Mod, Fun, Args)
  catch
    Kind:Reason ->
      Entry = {Mod, Fun, length(Args), [{file, File}, {line, erl_anno:line(Ann)}]},
      erlang:raise(Kind, Reason, [Entry | ?STACKTRACE])
  end
end.
```

The stacktrace may need some munging but that's precisely the point, I will have the flexibility to do so. :D

**Additional context**

I will be glad to implement this. I am looking for an initial thumbs up if this direction is acceptable. Thank you! 🎄

josevalim:
This allows code using `erl_eval` to provide better
stacktraces if desired. For this purpose, we also
make sure all errors raised from `erl_eval` pass
through the ExternalFunctionHandler.

This PR is broken in two commits to facilitate review.
The first focuses on `erl_eval` and the second also
modifies `eval_bits`. I can squash if desired.

Closes #5526.

### DIFF
--- a/lib/stdlib/test/erl_eval_SUITE.erl
+++ b/lib/stdlib/test/erl_eval_SUITE.erl
@@ -43,6 +43,7 @@
          otp_13228/1,
          otp_14826/1,
          funs/1,
+         custom_stacktrace/1,
 	 try_catch/1,
 	 eval_expr_5/1,
 	 zero_width/1,
@@ -91,7 +92,7 @@ all() ->
      simple_cases, unary_plus, apply_atom, otp_5269,
      otp_6539, otp_6543, otp_6787, otp_6977, otp_7550,
      otp_8133, otp_10622, otp_13228, otp_14826,
-     funs, try_catch, eval_expr_5, zero_width,
+     funs, custom_stacktrace, try_catch, eval_expr_5, zero_width,
      eep37, eep43, otp_15035, otp_16439, otp_14708, otp_16545, otp_16865].
 
 groups() -> 
@@ -1000,7 +1001,7 @@ otp_14826(_Config) ->
     backtrace_check("fun(P) when is_pid(P) -> true end(a).",
                     function_clause,
                     [{erl_eval,'-inside-an-interpreted-fun-',[a],[]},
-                     {erl_eval,eval_fun,7},
+                     {erl_eval,eval_fun,8},
                      ?MODULE]),
     backtrace_check("B.",
                     {unbound_var, 'B'},
@@ -1012,11 +1013,11 @@ otp_14826(_Config) ->
     backtrace_check("1/0.",
                     badarith,
                     [{erlang,'/',[1,0],[]},
-                     {erl_eval,do_apply,6}]),
+                     {erl_eval,do_apply,7}]),
     backtrace_catch("catch 1/0.",
                     badarith,
                     [{erlang,'/',[1,0],[]},
-                     {erl_eval,do_apply,6}]),
+                     {erl_eval,do_apply,7}]),
     check(fun() -> catch exit(foo) end,
           "catch exit(foo).",
           {'EXIT', foo}),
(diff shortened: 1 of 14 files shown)


lib/stdlib/test/erl_eval_SUITE.erl
```diff
@@ -1172,13 +1173,143 @@ simple1() ->
...
+                  (_) -> true
+                end, Stacktrace),
+            %% Now we prune any shared code path from erl_eval
+            {current_stacktrace, Current} =
+                erlang:process_info(self(), current_stacktrace),
+            Reversed = drop_common(lists:reverse(Current), lists:reverse(Pruned)),
+            Location = [{file, "evaluator"}, {line, erl_anno:line(Ann)}],
+            %% Add our file+line information at the bottom
+            Custom = lists:reverse([{my_module, my_function, 0, Location} | Reversed]),
+            erlang:raise(Kind, Reason, Custom)
```
josevalim, line 1294: For the tests, I have tried to emulate what I could do to get more precise stacktraces on eval.

lib/stdlib/test/shell_SUITE.erl
```diff
@@ -256,6 +256,8 @@ restricted_local(Config) when is_list(Config) ->
                   local_allowed(_,_,State) ->
                       {false,State}.
 
+                  non_local_allowed({erlang,raise},[_, _, _],State) ->
+                      {true,State};
```
josevalim, line 260: @bjorng this is a potential backwards incompatible change. Now all errors are raised through the evaluator, which means a restricted shell needs to allow them by allowing `erlang:raise/3` as above. Not allowing this means that accessing an undefined variable now says the operation is not restricted instead of raising.

This can be seen as a feature, but it can also be breaking. Other options are:

1. Have `erl_eval:raise/3` instead, with the `Anno`, `Reason`, and `Stacktrace`, to work as a clearer flag to check in the restricted shell, although ultimately it doesn't change much

2. Automatically raise for `erlang:raise/3`, instead of asking the restricted shell

If any of those changes are not desired, then the only option is to add a new callback to `erl_eval`.

max-au:
Interesting suggestion! I has an external implementation of this in `power_shell`, the PR may significantly simplify `power_shell` now (https://github.com/WhatsApp/power_shell). This PR will also solve an open issue I have there (when interpreted function is called as a callback, for example, from `lists:foldl`).

One other thing you might find useful. 
Currently, `erl_eval` gives up when a local function call is evaluated (https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_eval.erl#L310). In the forked evaluator,  `power_shell` calls local_handler: https://github.com/WhatsApp/power_shell/blob/main/src/power_shell_eval.erl#L326 that returns clauses for further evaluation.

With that change `power_shell` is able to evaluate all Erlang code. I can attempt a PR to move this to erl_eval. But it seems logical to include that code into your PR for completeness.

lib/stdlib/test/shell_SUITE.erl
bjorng: > Not allowing this means that accessing an undefined variable now says the operation is not restricted instead of raising.

Did you mean?

> ... is restricted instead of raising.

Also, it seems that accessing undefined variables will usually be catched by `erl_lint`:

```
Restricted Eshell V12.2  (abort with ^G)
1> Undefined.
* 1:1: variable 'Undefined' is unbound
```

For other types of errors, such various record errors and failure to match, there will be an ugly error message:

```
2> a = b.
** exception exit: restricted shell does not allow erlang:raise(error,
                                                                {badmatch,b},
                                                                [{erl_eval,expr,6,[{file,"erl_eval.erl"},{line,470}]},
                                                                 {shell,exprs,7,[{file,"shell.erl"},{line,689}]},
                                                                 {shell,eval_exprs,7,[{file,"shell.erl"},{line,645}]},
                                                                 {shell,restricted_eval_loop,4,
                                                                        [{file,"shell.erl"},{line,639}]}])
```

In OTP 24, the message would be:

```
1> a = b.
** exception error: no match of right hand side value b
```

I think that this change is acceptable from a backward compatibility perspective. Everything will still work if a restrictive shell does not let through `erlang:raise/3` but error messages will be harder to read.

@KennethL and @jhogberg, what do you think?

lib/stdlib/test/shell_SUITE.erl
```diff
@@ -256,6 +256,8 @@ restricted_local(Config) when is_list(Config) ->
                   local_allowed(_,_,State) ->
                       {false,State}.
 
+                  non_local_allowed({erlang,raise},[_, _, _],State) ->
```
bjorng, line 259: Shouldn't it be sufficient to just let `erlang:raise(error, Reason, Stk)` through?

```suggestion
                  non_local_allowed({erlang,raise},[error, _, _],State) ->
```

Also, the documentation of restricted shell in the documentation of `shell` needs to be updated.

lib/stdlib/test/shell_SUITE.erl
bjorng: Additional idea after discussion with @jhogberg: When an exception is thrown because the restricted shell does not allow it, the catcher of the exception (the module `erl_error`, I think), should be able to figure out that the original call originated from `erl_eval` and reconstruct the original error message.

josevalim: @bjorng do you mean we could look at the stacktrace top and see if it comes from `erl_eval`?

bjorng: Yes.

josevalim: Doing that or keeping it as is both work for me. Let me know if you have a preference. :)

josevalim: Perhaps I would keep it as is, to make it clear for those in restricted shells that they may want to handle it, and consider adding the mentioned fallback based on feedback.

bjorng: > Perhaps I would keep it as is, to make it clear for those in restricted shells that they may want to handle it, and consider adding the mentioned fallback based on feedback.

OK. Let's keep it and wait for feedback.

We'll need an update of the documentation for restricted shell.

josevalim:
I matched on `error`, updated the restricted shell docs, and squashed!

bjorng:
Thanks, added to our daily builds.

Issue #5608, opened 2022-01-15
Add GC/hibernation configurations for supervisors, or even set it to be default option


zzydxm:
**Is your feature request related to a problem? Please describe.**
Supervisors are often mostly idle after being started. Every garbage generated during the start process of the supervisor will stay the whole lifetime of the supervisor.

Especially, supervisor has a progress report of start_child, which calls customized log filters and handlers if it was defined, and that could cause large memory footprint.

After tls_dyn_connection_sup was introduced. Every TLS connection starts a separate supervisor, this caused an issue on us while we have 50k+ connections. Each tls_dyn_connection_sup supervisor holds a 300KB heap, and the summation is 15GB, which makes our service OOM-looping.

**To Reproduce**
erl
logger:set_primary_config(level, info).
logger:add_primary_filter(test, {fun(_, _) -> lists:seq(1, 1000000), ignore end, ok}).
application:ensure_all_started(os_mon).
process_info(whereis(os_mon_sup),[total_heap_size]).

return:
[{total_heap_size,4712239}]

**Expected behavior**
After some time total_heap_size should shrink to negligible.

**Affected versions**
This supervisor behavior existed for a quite long time, but the newly added tls_dyn_connection_sup might easily cause memory issue because of this behavior
https://github.com/erlang/otp/commit/3f8d48a6fa0dd5ccdf6c0d401efca52228af5ca8

**Describe the solution you'd like**

While we can do something on our logging to solve this issue. I think it might be good to add support in supervisor itself to support hibernation with a new API start_link/4, and ideally change most non-dynamic supervisors to hibernate by default.

max-au:
I would be interested in suppressing supervisor progress reports altogether, especially for processes that are expected to be spawned with high intensity (e.g. accepting TLS connections). There is not much value in having those logged anyway.

IngelaAndin:
Note that some supervisor shutdown logs are avoided by adjustment b0f82ee6e721a7bab80cd9b034061459c4b36dc7
that has not been released yet, but will be included in the next OTP-24 patch.  I also agree that supervisor progress reports for the dynamic supervisor processes are not what you want. I think that supervisor progress reports in common ought to be debug-level logging, but this would be changing legacy behavior so it would take some consideration when and if to change it.

garazdawi:
> While we can do something on our logging to solve this issue. I think it might be good to add support in supervisor itself to support hibernation with a new API start_link/4, and ideally change most non-dynamic supervisors to hibernate by default.

[...]

PR #5632, opened 2022-01-21
Garbage collect supervisors after initialization

josevalim:
Supervisor may generate garbage during initialization,
especially if logging is enabled, and this garbage may
never be collected, especially if a supervisor does not
perform additional work.

This commit forces the supervisor to immediately
garbage collect after initialization.

See #5608.

### DIFF
--- a/lib/stdlib/src/supervisor.erl
+++ b/lib/stdlib/src/supervisor.erl
@@ -349,6 +349,10 @@ init_children(State, StartSpec) ->
         {ok, Children} ->
             case start_children(Children, SupName) of
                 {ok, NChildren} ->
+                    %% Force a garbage collection since initialization may generate
+                    %% garbage, especially if logging, but the supervisor may not work
+                    %% frequently enough to trigger a GC any time soon.
+                    _ = erlang:garbage_collect(),
                     {ok, State#state{children = NChildren}};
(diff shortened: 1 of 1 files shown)


lhoguin:
The same should probably be done on the top-level process (application behavior) because work that is done in `start/2` may never be freed. We can trigger the GC manually in this case so it's less of an issue.

ferd:
Would post-initialization hibernation make more sense, since it GCs and compacts memory at once and we know there's likely a bunch of other processes ready to do work (as part of initialization)?

josevalim:
@ferd I think you could argue for hibernation too but I went with GC because it is the minimal amount of work. Not sure if we could end-up doing "too much".

IngelaAndin:
Probably hibernation would be better except for simple_one_for_one supervisors.

josevalim:
@IngelaAndin no worries. The PR is only 7 LOC and if you would prefer to go with another approach, then feel free to close this. hibernation may be even fewer lines :D

IngelaAndin:
@josevalim ok I do not have time to do this for RC1 as deadline is today. I will leave this open, as a placeholder, until I have time to make an other PR that can replace it.

IngelaAndin:
Replaced by #6895

Issue #5140, opened 2021-08-20
Incredibly slow recompilation in repl due to needless optimizations


drathier:
**Describe the bug**
Some files are incredibly slow to recompile because the optimizer wants to optimize code that doesn't need optimizing. This makes iterative work in the repl much much slower.

**To Reproduce**
Build a huge file which the compiler wants to optimize.

**Expected behavior**
Sub-second recompiles without optimizations, for fast development. 

**Affected versions**
OTP 24

**Additional context**

Manually rewriting code to avoid some optimization passes to trigger on a specific function took me from 227s to 117s. If I could avoid `beam_trim` too, that would save me another 75s, down to 42s. I don't want release builds when recompiling, iteration speed is super important.

I went from this:
```
core                          :      7.465 s  257290.3 kB
 sys_core_fold                 :      1.041 s  232849.0 kB
 sys_core_alias                :      0.048 s  232849.0 kB
 core_transforms               :      0.000 s  232849.0 kB
 sys_core_bsm                  :      0.020 s  232849.0 kB
 v3_kernel                     :      0.461 s  238360.5 kB
 beam_kernel_to_ssa            :      0.134 s   91849.2 kB
 beam_ssa_bool                 :     18.652 s   91849.2 kB
 beam_ssa_share                :      0.071 s   91849.2 kB
 beam_ssa_recv                 :      0.002 s   91849.2 kB
 beam_ssa_bsm                  :      0.147 s   92004.6 kB
    %% Sub passes of beam_ssa_bsm from slowest to fastest:
    allow_context_passthrough  :      0.137 s  97 %
    annotate_context_parameters:      0.002 s   2 %
    combine_matches            :      0.001 s   1 %
    accept_context_args        :      0.001 s   1 %
    skip_outgoing_tail_extracti:      0.000 s   0 %
 beam_ssa_funs                 :      0.053 s   92004.7 kB
 beam_ssa_opt                  :     80.944 s   92113.3 kB
    %% Sub passes of beam_ssa_opt from slowest to fastest:
    ssa_opt_sink               :     74.789 s  93 %
    ssa_opt_live               :      1.690 s   2 %
    ssa_opt_type_continue      :      1.305 s   2 %
    ssa_opt_dead               :      1.011 s   1 %
    ssa_opt_type_start         :      0.981 s   1 %
    ssa_opt_cse                :      0.237 s   0 %
    ssa_opt_tail_phis          :      0.165 s   0 %
[...]

PR #5637, opened 2022-01-24
beam_trim: Reduce compile time for huge straight-line functions

bjorng:
For a huge function with straight-line code, the `beam_trim` pass
could take forever, mainly because of the ineffcient way of
calculating the use information in the `frame_layout()` function.

This pull request reduces the time for the `beam_trim` of the sample code
in #5140 to about 1 second on my computer (down from at least
5 minutes).

While at it, also add support for new instructions in OTP 25, and
fix a minor bug resulting in some trimming opportunities were being
ignored.

Co-authored-by: John Högberg <john@erlang.org>

### DIFF
--- a/lib/compiler/src/beam_trim.erl
+++ b/lib/compiler/src/beam_trim.erl
@@ -21,12 +21,13 @@
 -module(beam_trim).
 -export([module/2]).
 
--import(lists, [any/2,reverse/1,reverse/2,sort/1]).
+-import(lists, [any/2,reverse/1,reverse/2,seq/2,sort/1]).
 
 -include("beam_asm.hrl").
 
 -record(st,
-	{safe :: sets:set(beam_asm:label()) %Safe labels.
+        {safe :: sets:set(beam_asm:label()),    %Safe labels.
+         fsz :: non_neg_integer()
         }).
 
 -spec module(beam_utils:module_code(), [compile:option()]) ->
@@ -38,8 +39,9 @@ module({Mod,Exp,Attr,Fs0,Lc}, _Opts) ->
 
 function({function,Name,Arity,CLabel,Is0}) ->
     try
-        St = #st{safe=safe_labels(Is0, [])},
-        Is = trim(Is0, St, []),
+        St = #st{safe=safe_labels(Is0, []),fsz=0},
+        Usage = none,
+        Is = trim(Is0, Usage, St),
         {function,Name,Arity,CLabel,Is}
     catch
         Class:Error:Stack ->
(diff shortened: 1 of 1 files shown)


lib/compiler/src/beam_trim.erl
```diff
@@ -385,177 +387,210 @@ is_safe_label_block([]) -> true.
...
+            U = {Regs,Ns},
+            do_usage(Is, Safe, Regs, Ns, [U|Acc]);
+        false ->
+            none
+    end;
+do_usage([{get_map_elements,Fail,S,{list,List}}|Is], Safe, Regs0, Ns0, Acc) ->
+    case is_safe_branch(Fail, Safe) of
+        true ->
+            {Ss,Ds1} = beam_utils:split_even(List),
+            Ds = yregs(Ds1),
```
jhogberg, line 510: Nitpick:
```suggestion
            {Ss,Ds0} = beam_utils:split_even(List),
            Ds = yregs(Ds0),
```

lib/compiler/src/beam_trim.erl
```diff
@@ -212,12 +222,8 @@ expand_recipe({Layout,Trim,Moves}, FrameSize) ->
     end.
 
 create_map(Trim, []) ->
```
jhogberg, line 224: Why remap through a fun, shouldn't it be enough to pass `Trim` and the move map to `remap`?

lib/compiler/src/beam_trim.erl
```diff
@@ -38,51 +39,55 @@ module({Mod,Exp,Attr,Fs0,Lc}, _Opts) ->
...
+        none ->
             %% No recipe worked out. Use the original init_yregs/1
             %% instruction.
-	    trim(Is0, St, [I|Acc])
+	    trim(Is0, Us, St, [I|Acc]);
+        R ->
+            %% Apply the recipe.
+            {TrimInstr,Map} = expand_recipe(R, FrameSize),
+            Is = remap(Is0, Map, []),
+	    trim(Is, none, St, reverse(TrimInstr) ++ Acc)
```
jhogberg, line 82: I think this would be clearer as a body-recursive function. `reverse(TrimInstr) ++ Acc` feels a bit awkward.

lib/compiler/src/beam_trim.erl
```diff
@@ -257,14 +250,19 @@ create_map(Trim, Moves) ->
 
 remap([{'%',Comment}=I|Is], Map, Acc) ->
```
jhogberg, line 251: I think this would be clearer as a body-recursive function.

jhogberg (approved):

lib/compiler/src/beam_trim.erl
bjorng: Will fix.

lib/compiler/src/beam_trim.erl
bjorng: Yes, make sense. Will fix.

PR #5640, opened 2022-01-24
jinterface: Fix encoding bugs for pids, ports and refs

sverker:
in class `OtpOutputStream`
for functions
```
void write_pid(String node, int id, int serial, int creation)
void write_port(String node, int id, int creation)
void write_ref(String node, int[] ids, int creation)
```

### DIFF
--- a/lib/jinterface/java_src/com/ericsson/otp/erlang/OtpOutputStream.java
+++ b/lib/jinterface/java_src/com/ericsson/otp/erlang/OtpOutputStream.java
@@ -700,24 +700,22 @@ public class OtpOutputStream extends ByteArrayOutputStream {
      *            the nodename.
      *
      * @param id
-     *            an arbitrary number. Only the low order 15 bits will be used.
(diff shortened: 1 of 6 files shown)


lib/jinterface/java_src/com/ericsson/otp/erlang/OtpOutputStream.java
```diff
@@ -818,13 +815,10 @@ public void write_ref(final String node, final int[] ids, final int creation) {
 
 	write_atom(node);
 
-	write1(creation & 0x3); // 2 bits
-
-	// first int gets truncated to 18 bits
-	write4BE(ids[0] & 0x3ffff);
+	write4BE(creation);
 
 	// remaining ones are left as is
```
michalmuskala, line 820: This comment should probably be updated as well

sverker: Yes. Thank you.

PR #5646, opened 2022-01-25
Add IP checking functions to the inet module

Maria-12648430:
This PR was promted by [this and the following posts](https://erlangforums.com/t/what-do-you-think-should-be-in-otp/556/78?u=maria-12648430) on the Erlang Forum. There are currently no functions to check if a given term qualifies as an IP address (v4, v6, or both) that would be accepted by other `inet` (and related) functions.

The OP asked for IP-checking _guards_, but IMO this is not feasible given IPs are something that is restricted to the context of `inet` and friends. Having functions that can be used for IP checking, however, removes the need to implement such functions or macros in user code.


meox:
Thanks @Maria-12648430 for this PR. The purpose of having as a macro was to use it as a function guard.
For instance in our codebase we have a lot of checks like this one:

`erlang
check_pkt(IP) when IS_IPv4() ->
.... do stuff for IPV4
check_pkt(IP) when IS_IPv6() ->
.... do stuff for IPV6
`

Maria-12648430:
> Thanks @Maria-12648430 for this PR. The purpose of having as a macro was to use it as a function guard. For instance in our codebase we have a lot of checks like this one:
> 
> ```erlang
> check_pkt(IP) when IS_IPv4(IP) ->
> .... do stuff for IPV4
> check_pkt(IP) when IS_IPv6(IP) ->
> .... do stuff for IPV6
> ```

Yes... with the additions in this PR, you won't be able to do it as nice and convenient as if you had guards, but at least you don't have to implement IP checking.
```erlang
check_pkt(IP) ->
    case {inet:is_ipv4_address(IP), inet:is_ipv6_address(IP)} of
        {true, false} ->
            do_stuff_for_IPv4();
        {false, true} ->
            do_stuff_for_IPv6();
        {false, false} ->
            error(not_an_ip);
        {true, true} ->
            exit(this_should_never_happen)
    end
```

IngelaAndin:
FYI: It was decided that we want a technical board discussion for this. That is why it will be stalled until then.

Maria-12648430:
@IngelaAndin ok :) Didn't think it was such a biggie, but no problem :)

IngelaAndin:
@Maria-12648430 well, even things that feel uncontroversial at first can turn out not to be. I prefer to postpone further explanations to promote a constructive dialog after the meeting.

Maria-12648430:
Ok :+1: :smile:

RaimoNiskanen:
We (the OTB (Otp Technical Board)) has decided to approve this PR.
I just labeled it Testing and do not anticipate any problems.

The discussion was mainly about the macros alternative:
since there _is_ an include file kernel/include/inet.hrl (that defines the `#hostent{}` record),
it _would_ be possible to introduce macros a'la:
``` erlang
-define(is_ipv4_address(A),
        (is_tuple(A) andalso tuple_size(A) =:= 4 andalso
         ((element(1, (A)) bor element(2, (A)) bor element(3, (A)) bor element(4, (A)))
          band (bnot 16#ff)) =:= 0)
```
Such macros would work in guard context but evaluate the argument
6 times for IPv4 and 10 times for IPv6 (6 or 11 times for IPv4+IPv6) before returning `true`.

The global macro name space would be polluted by 2 or 3 obvious names, but only for those
including inet.hrl, and those who want the macros would have to include the file.
Alas, Elixir cannot use Erlang include files (to our knowledge).

Given that we do not know how big interest there remains in such macros given the flip sides and that
the functions in this PR are introduced, we leave the possible macros to be a future problem,
and accept this PR.

Maria-12648430:
Hi @RaimoNiskanen, thanks for the explanation 🙂
One last question, though: what shall I put in the `since` attributes in the docs for the new functions?

RaimoNiskanen:
I have created an internal ticket for this to get a release note.
Use `since="OTP @OTP-17923@"`.

RaimoNiskanen:
We have code freeze for OTP-25.rc1 now, if you are fast I may be able to merge this PR for that...

Maria-12648430:
> We have code freeze for OTP-25.rc1 now, if you are fast I may be able to merge this PR for that...

There! 5 mins, how's that? 😁

RaimoNiskanen:
Excellent!

RaimoNiskanen:
Finally smoke test success.

PR #5649, opened 2022-01-26
ets: Rename and export some types

garazdawi:
Rename some ets types to clearer name:

* access() -> table_access()
* tab() -> table()
* type() -> table_type()
* comp_match_spec() -> compiled_match_spec()

The exported types have their old variants kept for backward compatibility.

The types `table_access` and `table_type` have been exported.

For consistency, I also renamed all usage of `Tab` to `Table` within the ets module itself.

### DIFF
--- a/erts/preloaded/src/erts_internal.erl
+++ b/erts/preloaded/src/erts_internal.erl
@@ -797,7 +797,7 @@ spawn_system_process(_Mod, _Func, _Args) ->
 %%
 
 -spec ets_lookup_binary_info(Tab, Key) -> BinInfo when
-      Tab :: ets:tab(),
+      Tab :: ets:table(),
       Key :: term(),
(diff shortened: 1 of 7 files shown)


paulo-ferraz-oliveira:
If I'm looking at this right, it means in the future (when this comes out), `dialyzer` (via `unknown`) will start complaining about stuff.

I like the new names, in any case, and will surely pull request changes to lib.s when I find these interrupt our CI (we consider types part of our "interface").

I just want to make sure this is the intended behaviour.

sverker:
`git grep "ets:tab("` gave some more hits apart from `dets` and `comp_match_spec` is used in `dets`.

KennethL:
I think this is good

KennethL (approved):
I think this is good, and as far as I can see there is no incompatibility introduced (old code using the exported types before this change should not get new Dialyzer warnings)

PR #5654, opened 2022-01-26
[kernel] peer-ify application

max-au:
Replace test_server:start_node with ?CT_PEER.


max-au:
This PR leaves `erl_distribution_SUITE` and `global_SUITE` out to avoid merge conflicts with #5611. 
Also, `interactive_shell_SUITE` is not included, for it appears to use one more technique for starting peer node which may not be supported by `?CT_PEER` yet.

It also improves `peer_SUITE` in presence of rogue nodes left from not-yet-converted test cases.

lib/kernel/test/pg_SUITE.erl
```diff
@@ -317,228 +294,228 @@ thundering_herd(Config) when is_list(Config) ->
...
     sync(?FUNCTION_NAME),
 
-    stop_node(Peer1, Socket1),
-    stop_node(Peer2, Socket2),
+    peer:stop(Peer1),
+    peer:stop(Peer2),
     ok.
 
 nolocal(Config) when is_list(Config) ->
-    {Peer, Socket} = spawn_node(?FUNCTION_NAME, ?FUNCTION_NAME),
```
garazdawi, line 466: I get why you want to rename the variable `Socket` to `Node` here, but could you in the future make that type of change into a separate commit? It makes it very hard to review the changes when 90% of all changes are renamed variables.

lib/kernel/test/pg_SUITE.erl
```diff
@@ -766,49 +669,13 @@ forever() ->
...
-                        {error, {'EXIT', Reason}};
-                    error:Reason ->
-                        {error, {'EXIT', Reason}}
-                end,
-            ok = gen_tcp:send(Sock, term_to_binary(Ret)),
-            server_loop(Sock);
-        {tcp_closed, Sock} ->
-            erlang:halt(1)
-    end.
+spawn_node(Scope, TestCase) ->
```
garazdawi, line 677: Would it make sense to call this `spawn_disconnected_node`?

lib/stdlib/src/peer.erl
```diff
@@ -649,17 +649,19 @@ command_line(Listen, Options) ->
     StartCmd =
         case Listen of
             undefined when map_get(connection, Options) =:= standard_io ->
-                ["-user", atom_to_list(?MODULE)];
```
garazdawi, line 652: Please break the changes in peer out into a separate commit.

garazdawi: Also, please add a testcase to peer_SUITE that tests the `-noshell`/`-user` interaction.

garazdawi (changes requested):

lib/stdlib/src/peer.erl
max-au: I will separate this change to a separate PR. The whole interaction is rather intricate and potentially error-prone, so I need to document it better (or even improve to a point where it's not an issue).

lib/kernel/test/pg_SUITE.erl
max-au: My bad, I had a number of commits (one per test suite in fact), but then habit took over and I squashed the thing. For the future changes I will separate commits containing refactoring from logic changes.

lib/kernel/test/heart_SUITE.erl
```diff
@@ -21,159 +21,131 @@
...
 end_per_suite(Config) when is_list(Config) ->
-    ignore_cores:fini(Config).
+    Config.
 
+%%-----------------------------------------------------------------
+%% Test suite for heart
+
+heart_args() ->
+    Common = ["-heart"],
+    case test_server:os_type() of
```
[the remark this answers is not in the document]
max-au: Would you accept a PR that codemod's `test_server:os_type()` for `os:type()` everywhere? I hope that will modernise codebase and make it less cryptic.

lib/kernel/test/heart_SUITE.erl
```diff
@@ -187,114 +159,84 @@ reboot(Config) when is_list(Config) ->
...
-    try
-	node_start_soon_after_crash_test(Config2)
-    after
-	ignore_cores:restore(Config2)
-    end.
- 
-node_start_soon_after_crash_test(Config) when is_list(Config) ->
-    {ok, Node} = start_check(loose, ?UNIQ_NODE_NAME,
-			     [{"ERL_CRASH_DUMP_SECONDS", "10"}]),
+    {ok, _Peer, Node} = ?CT_PEER(#{connection => standard_io,
```
garazdawi, line 206: This TC also produces a core and needs to be ignored.

lib/kernel/test/heart_SUITE.erl
```diff
@@ -21,159 +21,131 @@
...
+%% Test suite for heart
 
-init_per_suite(Config) when is_list(Config) ->
+heart_args() ->
+    Common = ["-heart"],
     case os:type() of
-	{win32, windows} ->
-	    {skipped, "No use to run on Windows 95/98"};
-	_ ->
-	    ignore_cores:init(Config)
```
garazdawi, line 96: The ignire_cores call in init_per_suite and end_per_suite needs to be there for the core ignore things to work.

lib/kernel/test/heart_SUITE.erl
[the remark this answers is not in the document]
garazdawi: Yes, that should be fine.

garazdawi:
These testcase are failing:

* global_group_SUITE:no_gg_proc_sync
* global_group_SUITE:no_gg_proc
* global_group_SUITE:compatible
* erl_prim_loader:multiple_slaves
* pg_SUITE:group_leave

I suspect that it is (again) interactions with #5611 that make them fail. So I will put this PR on hold until that PR is merged.

max-au:
I reproduced `pg` failures with #5611 using master branch (before this PR) with `prevent_overlapping_partitions` set to true. I think that new implementation of `global` and `pg`  have contradicting assumptions: `pg` is designed to work with missing links in the mesh (tolerating netsplits), while `global` wants to disconnect all loose ends and keep only the strongly connected component.

`global_group_SUITE` has that issue too, but is ignores slave shutdown issues, see https://github.com/erlang/otp/blob/master/lib/kernel/test/global_group_SUITE.erl#L1214

I'll wait for OTP board decision on #5611, as changing the default behaviour that way may render significant amount of tests broken, or requiring to disable `global` via `-connect_all false` for all peer nodes.

rickard-green:
> I reproduced `pg` failures with #5611 using master branch (before this PR) with `prevent_overlapping_partitions` set to true.

The `pg_SUITE` has since its introduction tried to disable `global` but failed to do so by concatenating the cookie with the `-connect_all false` argument due to a missing space:
```
 -spec controller(atom(), atom(), pid()) -> ok.
 controller(Name, Scope, Self) ->
     Pa = filename:dirname(code:which(?MODULE)),
     Pa2 = filename:dirname(code:which(pg)),
     Args = lists:concat(["-setcookie ", erlang:get_cookie(),
-            "-connect_all false -kernel dist_auto_connect never -noshell -pa ", Pa, " -pa ", Pa2]),
+            " -connect_all false -kernel dist_auto_connect never -noshell -pa ", Pa, " -pa ", Pa2]),
     {ok, Node} = test_server:start_node(Name, peer, [{args, Args}]),
     case rpc:call(Node, ?MODULE, control, [Scope], 5000) of
         {badrpc, nodedown} ->
             Self ! {badrpc, Node},
             ok;
```

When I made the above fix (to an unmodified `pg_SUITE` on `master`) `pg_SUITE` does not fail more with #5611 than it did before #5611 (it has never run completely clean on all platforms).

> I think that new implementation of `global` and `pg` have contradicting assumptions: `pg` is designed to work with missing links in the mesh (tolerating netsplits), while `global` wants to disconnect all loose ends and keep only the strongly connected component.
> 

I don't think `pg` and `global` contradicts each other anymore now with #5611 than what they did when `pg` was introduced. `global` has always had the idea of a fully connected mesh (although failed in providing that).

> `global_group_SUITE` has that issue too, but is ignores slave shutdown issues, see https://github.com/erlang/otp/blob/master/lib/kernel/test/global_group_SUITE.erl#L1214
> 

As I wrote in #5417 I think this should be solved by using `connection => 0`.

> I'll wait for OTP board decision on #5611, as changing the default behaviour that way may render significant amount of tests broken, or requiring to disable `global` via `-connect_all false` for all peer nodes.

Using `-connect_all false` for all peer nodes will not be acceptable, since it then would not be possible to test anything using `global`. The only tests effected by #5611 should be test making explicit disconnect. There are a huge amount of tests having no issue what so ever with having `prevent_overlapping_partitions` enabled.

Copying this #5611 since it relates to that PR

max-au:
> The pg_SUITE has since its introduction tried to disable global but failed to do so by concatenating the cookie with the -connect_all false argument due to a missing space:

This very PR actually has the problem fixed - unlike `slave`, `peer` accepts a list of command line arguments, making it impossible to forget a space.
I will re-run the tests with the updated `global` and ensure there are no failures.

max-au:
Rebased to updated master branch and fixed merge conflicts.

PR #5656, opened 2022-01-26
SCTP connectx API

dotsimon:
Introduce gen_sctp:connectx_init/* for initiating multi-homed SCTP connections using sctp_connectx(3).

The advantage of sctp_connectx is that the underlying OS can initiate the connection using all available paths.

### DIFF
--- a/erts/emulator/drivers/common/inet_drv.c
+++ b/erts/emulator/drivers/common/inet_drv.c
@@ -2337,6 +2337,14 @@ static int async_ok_assoc_id(inet_descriptor* desc, sctp_assoc_t assoc_id)
 
     return erl_drv_send_term(desc->dport, caller, spec, i);
 }
+
+static int async_ok_maybe_assoc_id(inet_descriptor* desc, sctp_assoc_t *p_assoc_id)
+{
+  if (p_assoc_id)
+    return async_ok_assoc_id(desc, *p_assoc_id);
+  else
+    return async_ok(desc);
+}
 #endif
 
 static int async_ok_port(inet_descriptor* desc, ErlDrvTermData Port2)
@@ -12643,6 +12651,7 @@ static ErlDrvSSizeT packet_inet_ctl(ErlDrvData e, unsigned int cmd, char* buf,
 #ifdef HAVE_SCTP
 	unsigned timeout;
 	sctp_assoc_t assoc_id = 0;
+	sctp_assoc_t *p_assoc_id = NULL;
 #endif
 	DEBUGF(("packet_inet_ctl(%p): CONNECT\r\n",
                 desc->port)); 
@@ -12691,7 +12700,8 @@ static ErlDrvSSizeT packet_inet_ctl(ErlDrvData e, unsigned int cmd, char* buf,
 		  memcpy(rabuf, &remote, remote_size);
 		  rabuf += remote_size;
 		}
-		code = p_sctp_connectx(desc->s, addrs, addrcnt, &assoc_id);
+		p_assoc_id = &assoc_id;
+		code = p_sctp_connectx(desc->s, addrs, addrcnt, p_assoc_id);
                 FREE((char *)addrs);
 	      } else {
 		return ctl_error(EINVAL, rbuf, rsize);
@@ -12710,15 +12720,11 @@ static ErlDrvSSizeT packet_inet_ctl(ErlDrvData e, unsigned int cmd, char* buf,
 		if (timeout != INET_INFINITY)
 		    driver_set_timer(desc->port, timeout);
 		enq_async(desc, tbuf, INET_REQ_CONNECT);
-		if (assoc_id) {
-		  async_ok_assoc_id(desc, assoc_id);
-		} else {
-		  async_ok(desc);
-		}
+		async_ok_maybe_assoc_id(desc, p_assoc_id);
 	    }
 	    else if (code == 0) { /* OK we are connected */
 		enq_async(desc, tbuf, INET_REQ_CONNECT);
-		async_ok(desc);
+		async_ok_maybe_assoc_id(desc, p_assoc_id);
 	    }
 	    else {
 		return ctl_error(sock_errno(), rbuf, rsize);
--- a/lib/kernel/test/gen_sctp_SUITE.erl
+++ b/lib/kernel/test/gen_sctp_SUITE.erl
@@ -844,10 +844,12 @@ api_connectx_init(Config) when is_list(Config) ->
     {ok,Sb} = gen_sctp:open(Pb),
     {ok,Sa} = gen_sctp:open(),
     {ok, A1} = gen_sctp:connectx_init(Sa, [localhost], Pb, []),
+    true = (A1 =/= 0),
     {Localhost,Pb,#sctp_assoc_change{state=cant_assoc, assoc_id = A1}} =
 	recv_event(log_ok(gen_sctp:recv(Sa, infinity))),
     ok = gen_sctp:listen(Sb, true),
     {ok, A2} = gen_sctp:connectx_init(Sa, [localhost], Pb, []),
+    true = (A2 =/= 0),
     {Localhost,Pb,#sctp_assoc_change{state=comm_up, assoc_id = A2}} =
 	recv_event(log_ok(gen_sctp:recv(Sa, infinity))),
     ok = gen_sctp:close(Sa),
@@ -2406,10 +2408,12 @@ t_simple_local_sockaddr_in_connectx_init(Config) when is_list(Config) ->
                  port     => Pb},
     case gen_sctp:connectx_init(Sa, [SockAddr], []) of
         {ok, A1} ->
+	    true = (A1 =/= 0),
             {Localhost,Pb,#sctp_assoc_change{state=cant_assoc, assoc_id = A1}} =
                 recv_event(log_ok(gen_sctp:recv(Sa, infinity))),
             ok = gen_sctp:listen(Sb, true),
             {ok, A2} = gen_sctp:connectx_init(Sa, [SockAddr], []),
+	    true = (A2 =/= 0),
             {Localhost,Pb,#sctp_assoc_change{state=comm_up, assoc_id = A2}} =
                 recv_event(log_ok(gen_sctp:recv(Sa, infinity))),
             ok = gen_sctp:close(Sa),


dotsimon:
I fixed up the commit authoring. Do I need to do anything else to make the CLA assistant happy?

erts/emulator/drivers/common/inet_drv.c
```diff
@@ -517,6 +517,13 @@ static typeof(sctp_freepaddrs) *p_sctp_freepaddrs = NULL;
 static void (*p_sctp_freepaddrs)(struct sockaddr *addrs) = NULL;
 #endif
 
+#if defined(__GNUC__) && defined(HAVE_SCTP_CONNECTX)
+static typeof(sctp_connectx) *p_sctp_connectx = NULL;
+#else
+static int (*p_sctp_connectx)
+        (int sd, struct sockaddr * addrs, int addrcnt, sctp_assoc_t * assoc_id) = NULL;
+#endif
+
```
RaimoNiskanen, line 526: Shouldn't this be:
``` C
#ifdef HAVE_SCTP_CONNECTX
#ifdef __GNUC__
// GNU typeof declaration of sctp_connectx
#else
// Fallback declaration of sctp_connectx
#endif
#endif

File: erts/emulator/drivers/common/inet_drv.c [hunk 2]
```diff
@@ -12606,18 +12653,52 @@ static ErlDrvSSizeT packet_inet_ctl(ErlDrvData e, unsigned int cmd, char* buf,
 
 	    /* For SCTP, we do not set the peer's addr in desc->remote, as
 	       multiple peers are possible: */
-	    if ((xerror = inet_set_faddress
-		 (desc->sfamily, &remote, &buf, &len)) != NULL)
+	    if ((*buf) == INET_AF_LIST) {
+	      /* called as gen_sctp:connectx... */
+	      int addrcnt = 0;
+	      buf++; len--;
+	      addrcnt = get_int8(buf),
```
RaimoNiskanen, line 12660: Why is the length field a byte / is max 256 addresses an arbitrary limitation, or has it got some reasoning behind it?

erts/emulator/drivers/common/inet_drv.c
```diff
@@ -12606,18 +12653,52 @@ static ErlDrvSSizeT packet_inet_ctl(ErlDrvData e, unsigned int cmd, char* buf,
...
+	      buf++; len--;
+	      if (! p_sctp_connectx)
+		return ctl_error(ENOTSUP, rbuf, rsize);
+	      if (addrcnt > 0) {
+		inet_address ras[addrcnt];
```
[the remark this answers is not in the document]
RaimoNiskanen: We do not have room for a stack allocated array of 255 `inet_address` - it is a future stack overflow

RaimoNiskanen:
It seems FreeBSD 10 according to its man page for `sctp_bindx` only allows a list of one (1) addresses.
I will have to check later versions...

dotsimon:
> It seems FreeBSD 10 according to its man page for `sctp_bindx` only allows a list of one (1) addresses. I will have to check later versions...

Except this is connectx ?

RaimoNiskanen:
> Except this is connectx ?

Oh...  In too much of a hurry.  Sorry about that.  I find no such limitation on `sctp_connectx()`.

erts/emulator/drivers/common/inet_drv.c
RaimoNiskanen: Nope.  It is just like for the other functions.  Looks fine.

erts/emulator/drivers/common/inet_drv.c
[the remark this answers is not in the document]
RaimoNiskanen: Furthermore; the Linux `sctp_bindx()` API accepts an "array" of mixed IPv4 and IPv6 socket addresses:
> addrs is a pointer to an array of one or more socket addresses. Each address is con‐
> tained in its appropriate structure(i.e. struct sockaddr_in or struct sockaddr_in6).
> The family of the address type must be used to distinguish the address  length.  The
> caller specifies the number of addresses in the array with addrcnt.

Therefore I'd say it is incorrect to have an array of `inet_address`.

It should be a `malloc`():ed `char` buffer into which either a `struct sockaddr_in` or a `struct sockaddr_in6`
is copied, while advancing the destination pointer `sizeof()` the struct just copied.  We can count on
the `sctp_bindx()` call to fail if an address is not supported.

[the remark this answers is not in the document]
RaimoNiskanen: I now found that you are using `inet_set_faddress()` and its return len, so the code should be correct, but it is confusing that allocates as an array of `inet_address` and uses it as a `char buf[]` without a comment about that.

erts/preloaded/src/prim_inet.erl
```diff
@@ -1936,6 +1936,16 @@ type_value_2(addr, {local,Addr}) ->
...
 	    end
     end;
+type_value_2(addr, {connectx, [SockAddr0|_]=SockAddrs})
+  when is_map(SockAddr0) andalso length(SockAddrs) < 255 ->
+    lists:all(fun(SockAddr) -> type_value_2(addr, SockAddr) end, SockAddrs);
+type_value_2(addr, {{connectx, Addrs}, Port})
+  when length(Addrs) > 0 andalso length(Addrs) < 255 ->
+    lists:all(fun({A,B,C,D}) when ?ip(A,B,C,D) -> true;
+                 ({A,B,C,D,E,F,G,H}) when ?ip6(A,B,C,D,E,F,G,H) -> true;
+                 (_) -> falae
```
RaimoNiskanen, line 1946: `(_) -> false`

RaimoNiskanen:
I do not like that the `Addr` argument of `prim_inet:connect` is misused with `{connectx,SockAddrs} | {{connectX,Addrs},Port}`.  That argument is abused enough already with `IPv4 | IPv6 | {IPv4|IPv6,Port} | {Family,AddrWithPort}`.  A 2-tuple means it is a complete address containing Port.

Better would be, I think, to add `prim_inet:connectx(S, Addrs, Time)` or to piggy back on `prim_inet_connect(S, Addrs, Time)` along the lines of `prim_inet:bindx(S, Flag, Addrs)`.  It could do address validation the same way, or it might be a good idea to add a type `addrs` (or `addr_list` to `prim_inet:type_value_2/2` and `prim_inet:enc_value_2/2` that simply validates a list of `addr` types.
(In the process I am eager to clean up `prim_inet:connect/3,4` and `prim_inet:async_connect/4`;
 there is so much code duplication there so it is hard to see what the difference is between them...)

I do not see much point in having a length field first in the buffer down to `inet_drv.c`.  Why not just loop use a distinct operation code `SCTP_REQ_CONNECTX` and loop until out of addresses, like `SCTP_REQ_BINDX` does?

Is a new API function `gen_sctp:connectx_init/3,4,5` needed, or can we instead allow a list of addresses to `gen_sctp:connect/3,4,5` and `gen_sctp:connect_init/3,4,5`?

Should we allow mixing SockAddr (map())  addresses with IPv4/IPv6 addresses in the same list?  In `gen_udp` we have added the possibility to use `{Family,AddrWithPort}` addresses, and that would be a welcome feature in `gen_tcp` and `gen_sctp` too.  Then we get the problem of how to handle a distinct `Port` argument in combination with e.g an `Addr = {IPv4,OtherPort} | {inet,{IPv4,OtherPort}} | #{family=>inet, addr => IPv4}` and the map with or without a `port` field...

I think it would be strange to separate map() addresses from all the other possible (not added yet) address forms into different API functions.

In any case - thank you for this PR.  This is an old can of worms that is tedious to poke around in to not make things too much worse.  I hope we can evolve it as far as needed...

dotsimon:
Hi Raimo,
Thanks for your thoughtful comments, I am recreating for a few days so I won't be making any code changes until next week.
I agree the {connectx, ...} construct is ugly. I developed this on 23.1, before the sockaddr stuff and even though it seemed a bit of a kludge then, it was less much ugly!
I added the sockaddr forms of the API in order to keep "in the spirit" of the new changes but I can't think of a case where it would be actually useful. Although, at least with the underlying Linux API you could have both IPv4 & IPv6 addresses **iff** the socket was opened as IPv6. However, in OTP it feels like all the handling is **either** IPv4 or IPv6 (e.g. inet_scto vs inet6_sctp) so I'm not sure how that fits in.
The sockaddr approach also has another problem which my patch doesn't (but should) check - that the port number has to be same in all the addrs.
I deliberately did not make a `gen_sctp:connectx/3,4,5` because the code that matches the result for `gen_sctp:connect/3,4,5` is basically wrong (or, can do the wrong thing) which is the whole reason why I wrote the original patches for `gen_sctp:connect_init/4,5`. However in this case since the `assoc_id` is returned from the driver, it should be possible to implement a blocking call that doesn't get confused by other socket messages, but who would use that API?
I'm ok with the idea of dropping the length and looping over the addresses. If you feel like that's more the OTP style I can do that.
Regarding if the sctp_connectx functionality should be hidden from the user by allowing `connect_init` to accept a list of addresses, I see the attractiveness of having only one API but on the other hand it hides the underlying behaviour from the user. In the particular case that inspired these patches, I needed `sctp_connectx(3)` and if it were not available then a different strategy was used for certain peers.
I'm interested to hear your thoughts on this so we can work toward a patchset that is going to get this API implemented without creating more stumbles for the future development you mention.

erts/emulator/drivers/common/inet_drv.c
dotsimon: Yes, I have to admit I copied the pattern from `sctp_bindx` without thinking about it too hard!

RaimoNiskanen:
Hi Simon.
Great to know that you will not work on this until next week, and for a nice reason :-)

To use separate sockets for IPv4 and IPv6 is what we are drawn to in OTP because it is the least common denominator.  IIRC FreeBSD (among others) strongly recommends that, for robustness / security reasons (to avoid many firewall configuration mistakes), but it is still possible, in particular on Linux, to use an IPv6 socket also for IPv4, so we need to support that.  And yes, it does not fit well with the module division between `inet_sctp` and `inet6_sctp`; the latter should be able to handle IPv4 cases too, at as IPv6-mapped-IPv4-addresses...

> The sockaddr approach also has another problem which my patch doesn't (but should) check - that the port number has to be same in all the addrs.

Oops!  There will be some cases to handle, then...: the `Port` argument - `0` or not, and a list of `{IP,P} | {IP, 0} | #{family:=inet, addr:=A} | #{family:=inet, addr:= A,port:=0} | #{family:=inet, addr:=A, port:=P}`.  Port numbers anywhere and all needs to be consistent in some way...

Should they have to contain the same port?
Is `0` fine if there is a `Port` argument that is not `0`?
Would it be OK if the first `Port` in the address the list is `=/= 0` and then the others may be the same, `0`, or not present?
Should we maybe not have `connectx_init/*` functions without a `Port` argument...?
Perhaps just use the `Port` argument for completing IPv4 and IPv6 addresses, then pass all addresses to the OS call, and let it barf as it likes?


> I'm ok with the idea of dropping the length and looping over the addresses. If you feel like that's more the OTP style I can do that.

I looked at it again, and bindx does not need to allocate - it just loops over `sctp_bindx` with one address at the time, so it is not a completely relevant example.

In this case, if `inet_drv.c` gets the number of addresses, it can over-allocate for N structs of the largest size, just as you have suggested.  And a new operation number is not needed, as you have suggested.


I do not have a strong opinion about `gen_sctp:connect/3,4,5`. We have been meaning to fix it maybe by passing a ref back and fort to `inet_drv.c`, but I think there are other messages that may come before that are interesting, so perhaps it is better to forget about that function and deprecate it.


Today, `gen_sctp:open/1,2` accepts multiple `{ifaddr, IP|SockAddr}` options, and aggregates them into a `bind()` call and a loop of `bindx(add)` calls.  So there the list of addresses is accepted by the existing call.
I do not know if that is sufficient argument to have `sctp:connect_init/*` use `sctp_connectx()` when given a list of addresses (also a list of 1), and normal `connect()` when not a list.
Is there much semantic difference between `connect()` and `sctp_connectx()`, or is it just that it allows multiple addresses? 
I do not have a very firm opinion here, but so far lean towards allowing a list in the existing `connect_init` function.  It would not be very hard to get me to lean the other way...

dotsimon:
Hi Raimo,
I will answer your other concerns separately later, but regarding this:
> Is there much semantic difference between connect() and sctp_connectx(), or is it just that it allows multiple addresses?

In Linux the connect ends up in the same function regardless of how the user called (`sctp_connect` or any of the `sctp_connectx` variants). In FreeBSD they have duplicated the code in two functions (grrr) but the body of code that handles the first address looks the same.

Therefore, I am liking your idea of always using `sctp_connectx(3)` if it is available and extending `connect_init/*` to take a list of addresses. In my specific case, I would not need to know at runtime which is being used because the build & deployment environments are tightly controlled. Maybe you can think of a case where the user might want/need this info, and if so what would be an appropriate API to discover that?

Another option is to hide it completely from the user - so, in `inet_drv` to use `sctp_connectx` if available for >1 address and `sctp_connect` otherwise using only the first address.

It's also worth pointing out that the user still needs to rotate the addresses in case of error e.g.  `#sctp_assoc_change{state /= comm_up,...`. At least under Linux certain errors cause the kernel to return `cant_assoc` immediately rather than trying the next path. Without turning the OTP documentation into an SCTP essay, I'm not sure the best way to convey this. To me, if you're using SCTP you should already have a basic understanding of multihoming & a simple sentence like "_The user should rotate the addresses between calls in case of error_" could be sufficient.

/Simon

dotsimon:
One more thing, I left two commits on this branch - one for the code and the other for updating `preloaded`.
Is this the way the OTP team prefer this? Or not update the preload at all to make merging easier?

RaimoNiskanen:
> Therefore, I am liking your idea of always using sctp_connectx(3) if it is available and extending connect_init/* to take a list of addresses. In my specific case, I would not need to know at runtime which is being used because the build & deployment environments are tightly controlled. Maybe you can think of a case where the user might want/need this info, and if so what would be an appropriate API to discover that?

My suggestion is to always use connect(2) for an address, and always sctp_connectx(3) for a list of addresses (length >= 1).  Minimal change, easy to explain, and control over which backend function that will be used.


Regarding documentation of platform peculiarities: we should not have to document them, but it would be nice to the user to mention that some platforms bail out early so it might be worth rotating the list.


Regarding preloaded: yes, it is good to have the preloaded as the latest commit.  We will remove that and build our own, for security principals. When we merge we will not merge that commit.
In fact, you may remove the preloaded commit from the PR since then I can mark the PR as "Testing" and our integration script will automatically rebuild preloaded files if needed.

dotsimon:
The newly pushed branch covers the following points from previous discussion:

- allocation in `inet_drv.c`

> Why is the length field a byte / is max 256 addresses an arbitrary limitation, or has it got some reasoning behind it?
> We do not have room for a stack allocated array of 255 inet_address - it is a stack overflow waiting to happen

Agreed, the max 256 addresses an arbitrary limitation however it makes handling in the driver easier since an array of inet_address must be allocated.
Your original observation about the stack allocation was correct and I have changed it to use ALLOC()/FREE()
I don't believe the limitation is a practical one - it is absurd to think even a clustered SCTP peer would have 256 *distinct* addresses.

- `(_) -> falae` in `prim_inet.erl`
Fixed

- `gen_sctp:connectx_init/*` vs overloaded `gen_sctp:connect_init/*`

In the end I decided to keep the APIs separate because of the return value. The underlying `sctp_connectx` API returns the association ID which is extremely useful to get back. It very extremely messy to allow `connect_init` to take a list of addresses but then have the return value be maybe `ok` or maybe `{ok, AssocId}`

- (ab)use of `addr` type in `prim_inet.erl`

I added new exports `prim_inet:connectx/2,3` and a new type `addr_list` with hopefully enough comments!

- mixed IPv4 & IPv6 sockaddr

Since it is not currently the OTP way, I did not make any changes. FWIW, both Linux & FreeBSD allow IPv4 & IPv6 addresses for a socket that has been opened as IPv6. FreeBSD 12 specifically disallows IPv4-mapped IPv6 addresses presumably because the caller should pass them as IPv4 (family = inet). Other FreeBSD & Linux may enforce the same, I did not check.

- removed the preloaded commit

dotsimon:
> removed the preloaded commit

But now GitHub's automatic build and check action fails because the preload is not updated?

Let me know what, if anything, you'd like me to do about that!

RaimoNiskanen:
> >     removed the preloaded commit
> But now GitHub's automatic build and check action fails because the preload is not updated?

Alright.  There was the reason to have the preloaded commit. Better add it again to get GitHub's build and check working again...
The "Testing" label will not work, but I can add it manually to our daily tests - no biggie.

RaimoNiskanen:
I have pushed an update proposal, and run it in our daily tests.
The problem with assoc_id = 0 in the test suite remains.

We have code freeze for OTP-24.3 in a few days, so if we can find a solution before that this might get in, otherwise we will have to aim for OTP-25.0.

RaimoNiskanen:
In our daily builds it seems there are 3 machines on which the test passes, i.e they return an association ID /= 0, but there are 50 machines that the test fails on, where I have looked because they return an association ID == 0.

We also have a Solaris 11 machine that returns ENOTSUP for connectx

Should we modify the test to pass for `assoc_id = 0`, or do you suspect some other problem?

dotsimon:
I guess it will be OTP-25.0 then since I probably can't look at this until later this week.

> It seems on my machine (Ubuntu 18.04.6 LTS with libsctp1-1.0.17) sctp_connectx does not return an Association ID

`sctp_connectx` has been through a few implementation revisions but I'm pretty sure the API always had the association ID set on success. I have a VM with Ubuntu 18.04.6 LTS that I can test your case.

Do you have some OS info for those other failing machines?

> We also have a OpenIndiana machine that returns ENOTSUP for connectx

`enotsup` should be handled in the test since it is a valid response or otherwise used to skip those cases

> Should we modify the test to pass for assoc_id = 0, or do you suspect some other problem?

My recollection was that assoc_id = 0 is not valid under Linux  but I will dig into that more and also check the bsd kernel

lib/kernel/test/gen_sctp_SUITE.erl
```diff
@@ -798,6 +800,39 @@ api_connect_init(Config) when is_list(Config) ->
...
+    {ok,Pb} = inet:port(S),
+    try gen_sctp:connectx_init(S, Localhost, 12345, [])
+    catch error:badarg -> ok
+    end,
+    try gen_sctp:connectx_init(S, [Localhost], not_allowed_for_port, [])
+    catch error:badarg -> ok
+    end,
+    try gen_sctp:connectx_init(S, [Localhost], 12345, not_allowed_for_opts)
+    catch error:badarg -> ok
+    end,
```
RaimoNiskanen, line 817: These do not test that `gen_sctp:connectx_init/4` throws a `badarg`.
Should it return something; that value is simply returned and ignored.
I think the test should verify that it *shall* throw a badarg and *not* return...

lib/kernel/test/gen_sctp_SUITE.erl
```diff
@@ -2339,6 +2374,26 @@ do_simple_sockaddr_send_recv(#{family := _Fam} = SockAddr, _) ->
...
+%% Test the sockaddr version of connectx_init/4.
+t_simple_local_sockaddr_in_connectx_init(Config) when is_list(Config) ->
+    Localhost = {127,0,0,1},
+    {ok,Sb} = gen_sctp:open(),
+    {ok,Pb} = inet:port(Sb),
+    {ok,Sa} = gen_sctp:open(),
+    SockAddr = #{family   => inet,
+                 addr     => Localhost,
+                 port     => Pb},
+    {ok, A1} = gen_sctp:connectx_init(Sa, [SockAddr], []),
```
RaimoNiskanen, line 2386: On OpenIndiana `gen_sctp:connectx_init/3` here returns `{error,enotsup}` which does not match and the test fails.
I think that should cause the test case to be skipped instead.

dotsimon: Agreed. In my comment I meant that the test should _be changed_ to handle that case. Sorry for the confusion

lib/kernel/test/gen_sctp_SUITE.erl
dotsimon: Thanks for pointing that out, I had not considered an actual return.
Of course, the bad news is that this whole suite uses that pattern in many pre-existing test cases...

RaimoNiskanen:
> Do you have some OS info for those other failing machines?

We have Ubuntu 20.04.3 LTS, Ubuntu 18.04.5 LTS, SLES 12-0, SLES 12-2, Linux Mint 19.3, Linux Mint 20.2, Fedora 34.
libsctp-dev/focal 1.0.18+dfsg-1 amd64, libsctp-dev/bionic 1.0.17+dfsg-2 amd64, lksctp-tools 1.0.15, lksctp-tools 1.0.18

lib/kernel/test/gen_sctp_SUITE.erl
RaimoNiskanen: Oops...

dotsimon: How about wrapping the funtion to fail in `exit()` resulting in a pattern like this?
```erlang
try exit(gen_sctp:connectx_init(S, [Localhost], 12345, not_allowed_for_opts))
catch error:badarg -> ok
end,
```

I could make a separate PR to clean up the rest of the suite after this one is done.

RaimoNiskanen: Sounds good, but I do not like using `exit(_)` for this.  Maybe this trick:
``` erlang
try error({unexpected, gen_sctp:connectx_init(S, [Localhost], 12345, not_allowed_for_opts)})
catch error : badarg -> ok
end
```

Or create a wrapper function:
``` erlang
verify_badarg(Fun) ->
    try Fun() of
        Result -> ct:fail({unexpected, Result})
    catch error : badarg -> ok
    end.

    :
    verify_badarg(
        fun () ->
            gen_sctp:connectx_init(S, [Localhost], 12345, not_allowed_for_opts)
        end),
    :
```

dotsimon:
The Linux problem of no returned association ID will always occur when configuring with `--enable-sctp` however using `--enable-sctp=lib` or FreeBSD will work.

The underlying cause is that `libscp` contains versioned symbols for `sctp_connectx`, and long story short - the linker picks up the newest _VERS_3_ but `dlsym` gets the first version that did not return an association ID.

The irritating part is that _VERS_3_ has been fully working since 2013 (lksctp 1.0.13 & kernel 2.6.33) but it's not the default symbol version.

I pushed the branch now to return `enotsup` for the dynamically loaded library case and fixed up the tests to handle that.

In doing so, your commit that returns `{ok, 0}` is reverted. I could not see the usefulness of this. It would make a special case that has to be tested for by the caller separately, and then more special case code to handle the later `sctp_assoc_change` that comes with the real association ID.

Finally, your code cleanup commit [8f1b2ae](https://github.com/erlang/otp/pull/5656/commits/8f1b2ae79c096c2ed42eabe4feddce9df441c20c) changes the behaviour of an invalid port number. I adapted the test case, but maybe this wasn't what you intended?

Would you consider this PR enough for now? The sort of people wanting to use `sctp_connectx` probably aren't using an off-the-shelf distro build of OTP that dynamically loads the sctp library.

If so, I can make a separate PR to use `dlvsym` to pick the "right" `sctp_connectx` in a couple of weeks.

lib/kernel/test/gen_sctp_SUITE.erl
dotsimon: In the end I made a macro since the no_return case could be useful for other tests that may not necessarily be catching `badarg`

dotsimon:
I stopped being lazy and fixed the dynamic loading.
My expectation is that `gen_sctp:connectx_init` should now work on all platforms that have `sctp_connectx` regardless of how OTP is configured.
I left the last two separate commits for update_preloaded and update_configure since I assume you will want to regenerate these.

RaimoNiskanen:
Good work indeed!

Since the 'maint' window has passed, this branch will have to be rebased to 'master'; probably just change the target branch in the GitHub web interface at the top.

I will insert it in our daily tests when that is done (but will test it manuallly right now).

>In doing so, your commit that returns `{ok, 0}` is reverted. I could not see the usefulness of this. It would make a special case that has to be tested for by the caller separately, and then more special case code to handle the later `sctp_assoc_change` that comes with the real association ID.

The purpose of that change was to avoid having `connectx/*` return `ok | {ok, AssocID} | {error,Reason}` since then there are two OK values that will need different case clauses.  The signature should be `{ok,AssocID} | {error,Reason}`.  Then it is possible to have Let It Crash code that directly matches the return value to `{ok,AssocID}`.

Either it is an error that sctp_connectx(3) returns `AssocID =:= 0` and then that should be converted to `{error,FabricatedReason}`, or it is a platform peculiarity and then`{ok,0}` should be returned that the caller needs to handle by using a guard to create a case clause or some other way.  inet_drv.c should not accidentally return `ok`.

dotsimon:
If sctp_connectx succeeds then the association id should always be set to a value >= 3.
The values < 3 are special (see https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/sctp.h#L60) 

The reason for previously getting zero was the dynamic symbol resolution to the "original" sctp_connect which has this prototype:
```
int sctp_connectx(int fd, struct sockaddr *addrs, int addrcnt);
```

When using this old/obsolete version of the API, the `assoc_id` variable (passed by reference expecting the current API) is not modified and retains it's initialised value of zero.

Now using `dlvsym` the code only sets p_sctp_connectx iff VERS_3 is resolved, meaning `enotsup` is returned on any subsequent calls if the current API is not available. Therefore we should expect sctp_connectx/* to return `{ok,AssocID} | {error,Reason}` and never `ok`.

I see just changing the base branch causes conflicts on configure and the preloaded, so I will rebase, recreate those and push again shortly.

RaimoNiskanen:
Thank you for the rebase!

I will include it in our daily builds for 'master', and ask the VM team for review of `erts_sys_ddll_vsym2()`.

> Now using dlvsym the code only sets p_sctp_connectx iff VERS_3 is resolved, meaning enotsup is returned on any subsequent calls if the current API is not available. Therefore we should expect sctp_connectx/* to return {ok,AssocID} | {error,Reason} and never ok.

That is all good and well.  But now it is technically possible for the code in inet_drv.c to return `ok` if for some yet unknown reason the OS would choose to not assign `*assoc_id`, or assign it to `0`.  To rule out that possibility (for future debugging of some strange support case) the code _should_ be changed to always return `{ok,AssocID}` (or `{error,_}`) from `gen_tcp:connectx/*` whatever the OS returns, so we can exclude `ok` from the return type.

dotsimon:
Ok I get what you are saying... but since it would be an invalid response from the SCTP API then the state of the association is unknown.
In this case to return `{ok, 0}` is clearly wrong, however to synthesise an error implies the connection has failed and we can't know what the state is.
If you have something specific in mind, let me know and I can make that change. I am not going to lose any sleep over this case that should never happen™️

RaimoNiskanen:
I have discussed with the VM guys and we decided that we do not want to rely on a GNU extension for resolving dynamic library function versions, and we do not want to tie a resolved function to a hardcoded version - it smells like future problems when there comes a later version.

Therefore your previous approach seems like the least bad, i.e to use static linking to get the right function (version), and to require static linking for sctp_connectx. That would be a bit annoying for FreeBSD users, since from what I understand it works there, but probably avoids misuse from Linux users.  When Linux updates the default version we can change to allowing dynamic linking.

> In this case to return `{ok, 0}` is clearly wrong, however to synthesise an error implies the connection has failed and we can't know what the state is.

Yes. Good point.

Always return `{ok,_}` from `gen_sctp:connectx/*` seems to be the only option, then.  A flag is needed in inet_drv.c to remember `if ((*buf) == INET_AF_LIST)` and that can be done as I suggested with an `sctp_assoc_t *`, or with a boolean.

The test case should regard `{ok,0}` as an error e.g:

dotsimon:
I'd really like you/they to reconsider that decision:

> we do not want to tie a resolved function to a hardcoded version

This is **exactly** what you do want - to tie the resolved function to a version that works with the code using it.

> it smells like future problems when there comes a later version

Again, the code works with VERS_3 so a later version in the library will not be used. The only way this changes is if libsctp get a completely new version (eg. libsctp.so.2) but then there are no guarantees of compatibility and all configure, code, etc need to be rewritten to use the new library. This is the whole point of symbol versioning - to avoid having to completely rev the library.

> we do not want to rely on a GNU extension for resolving dynamic library function versions

To get a working sctp_connectx on Linux with a dynamic library then it must rely on `dlvsym` and resolving the correct `sctp_connectx` out of `libsctp`. But that working function is there and can be tested for iff `dlvsym` is available. Why would you not want to use functionality if it is there? If the platform is so old that it doesn't have `dlvsym` or `sctp_connectx` VERS_3 this is detected and the user ultimately gets `enotsup`

All of this only applies to Linux - in FreeBSD all the SCTP API are provided in libc. On that platform configure with `--enable-sctp=lib` has the exact same effect as `--enable-sctp` and will find `sctp_connectx` I think since FreeBSD 9 or 10.

So please think about this - I don't see how using a particular version can be bad when this is how symbol versioning is supposed to work. 

In the meantime, I will fix the return `{ok, AssocId}` from the driver and update the tests.

Thanks/Simon

RaimoNiskanen:
I had another discussion with the VM guys, and they accepted (muttering about bad choices in Linux)...

The branch is now running in our daily tests.

RaimoNiskanen:

PR #5658, opened 2022-01-27
docs: note that file lists have to be passed to typer before recursive directories

maco:
I ran into this today:

```
❯ typer -r src foo/bar.erl
typer: error involving a use of file:list_dir/1
```

It works fine if I do `typer foo/bar.erl -r src`, so I'm updating the docs to reflect that.

### DIFF
--- a/lib/dialyzer/doc/src/typer_cmd.xml
+++ b/lib/dialyzer/doc/src/typer_cmd.xml
@@ -54,7 +54,7 @@ typer --help</code>
 typer [--help] [--version] [--plt PLT] [--edoc]
       [--show | --show-exported | --annotate | --annotate-inc-files]
       [-Ddefine]* [-I include_dir]* [-pa dir]* [-pz dir]*
-      [-T application]* [-r] file*</code>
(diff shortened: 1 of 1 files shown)


lib/dialyzer/doc/src/typer_cmd.xml
```diff
@@ -54,7 +54,7 @@ typer --help</code>
 typer [--help] [--version] [--plt PLT] [--edoc]
       [--show | --show-exported | --annotate | --annotate-inc-files]
       [-Ddefine]* [-I include_dir]* [-pa dir]* [-pz dir]*
-      [-T application]* [-r] file*</code>
+      [-T application]* file* [-r]</code>
```
bjorng, line 57: Shouldn't it be like like this?

```suggestion
      [-T application]* file* [-r directory*]</code>
```

bjorng (commented):
Thanks for your pull request. I have one small question/comment.

bjorng:
Thanks! I made the change and merged to master.

PR #5175, opened 2021-09-01
Union, intersection and subtraction of character classes

rustkas:
Character classes have ability to use union, intersection and subtraction.

This functionality expands the possibilities for describing the required data.

PR #5659, opened 2022-01-28
Tune documentation texts

rustkas:


jhogberg:
The first two commits look like a duplicate of #5175, which was rejected. Can you drop them?

rustkas:
Yes. Please, do it. I can not do it by myself.

jhogberg:

PR #5669, opened 2022-01-31
Fix filelib:find_source for asn1 files

garazdawi:
This PR fixes `filelib:find_source/1` to work for asn1 source files and also fixes some minor issues (in public_key and eunit) where the source file for an Erlang/OTP module could not be found by `filelib:find_source/1`.


KennethL:
I think is strange that filelib should have to know how the catalogues and makefiles for public_key is organized, which is definitely not according to any standard.

In the asn1 directory there are soure for asn.1 files but that does not say that the generated .erl files should land in the asn1 directory. Perhaps better to gave den under src or in src/asn1_generates

garazdawi:
> In the asn1 directory there are soure for asn.1 files but that does not say that the generated .erl files should land in the asn1 directory. Perhaps better to gave den under src or in src/asn1_generates

Yes, this is exactly what this PR does.

The change in filelib is there for the .asn1 files that are directly compiled to .beam without any intermediary generated .erl files.

lib/public_key/asn1/Makefile
```diff
@@ -18,6 +18,9 @@
 # %CopyrightEnd%
 #
 
+EGEN = ../src
```
u3s, line 21: is the EGEN needed for something? does not seem to be used in current Makefile...

lib/public_key/src/.gitignore
```diff
@@ -0,0 +1,2 @@
+OTP-PUB-KEY.*
+PKCS-FRAME.*
```
u3s, line 2: Maybe something related to asn1 folder content is not needed anymore in `./lib/public_key/.gitignore` and can be removed?

lib/public_key/asn1/Makefile
garazdawi: it is used in `otp.mk` to know where to find/put generated erlang files.

lib/public_key/src/.gitignore
garazdawi: hmm, yes. will fix.

garazdawi: I removed the new .gitignore and changed the rules in `lib/public_key/.gitignore` instead.

u3s (approved):

IngelaAndin (approved):

PR #5677, opened 2022-02-03
[erts] enable Visual Studio 2022 for Windows

max-au:
Visual Studio default path has changes in 2022, it is no
longer a 32-bit application.


dgud:
I guess this would be useful on maint as well?

max-au:
Makes sense, thanks! Rebased.

PR #5678, opened 2022-02-03
beam_makeops: Make syntax nicer

bjorng:
This pull request updates the syntax of `*..tab` files to make it nicer and more explicit.

Repeating variables in patterns is no longer allowed. For example:

```
move R R =>
```

This must now be written:

```
move R1 R2 | equal(R1, R2) => _
```

This examples also illustrates another change. When instructions are to be removed (rather than replaced), the substitution on the right-hand side must now be written as '_'. That makes it clearer, and also enabling automatic continuation of lines that end in `|` or `=>`. For example, a complex transformation such as:

```
is_tuple NotTupleFail Src | \
  is_tagged_tuple WrongRecordFail Tuple Arity Atom | \
  equal(Src, Tuple) => \
   i_is_tagged_tuple_ff NotTupleFail WrongRecordFail Src Arity Atom | \
   current_tuple Src
```

can now be written simply as:

```
is_tuple NotTupleFail Src |
  is_tagged_tuple WrongRecordFail Tuple Arity Atom |
  equal(Src, Tuple) =>
   i_is_tagged_tuple_ff NotTupleFail WrongRecordFail Src Arity Atom |
   current_tuple Src
```

### DIFF
--- a/erts/emulator/beam/emu/ops.tab
+++ b/erts/emulator/beam/emu/ops.tab
@@ -37,7 +37,7 @@ FORBIDDEN_TYPES=h
 # needs to be re-compiled with a modern compiler.
 
 too_old_compiler/0
-too_old_compiler | never() =>
+too_old_compiler | never() => _
 
 # In R9C and earlier, the loader used to insert special instructions inside
 # the module_info/0,1 functions. (In R10B and later, the compiler inserts
@@ -82,7 +82,7 @@ int_func_start Lbl Line M F A => label Lbl | i_func_info u M F A | line Line
 # The end of a function.
 int_func_end/2
 int_func_end Func Entry | needs_nif_padding() => i_nif_padding
-int_func_end Func Entry =>
+int_func_end Func Entry => _
 
 i_nif_padding
 
@@ -117,7 +117,7 @@ move S X0=x==0 | line Loc | call_only Ar Func => \
 move S X0=x==0 | line Loc => line Loc | move S X0
 
 # The line number in int_func_start/5 can be NIL.
-line n =>
+line n => _
 line I
 
 # For the JIT, the init_yregs/1 instruction allows generation of better code.
@@ -258,7 +258,8 @@ i_get_tuple_element3 x P x
 is_number f? xy
 %hot
 
-is_number Fail=f i =>
+is_number Fail=f i => _
+
 is_number Fail=f na => jump Fail
 is_number Fail Literal=q => move Literal x | is_number Fail x
 
@@ -288,7 +289,7 @@ i_raise
 
 # Workaround the limitation that generators must always return at least one instruction.
 delete_me/0
-delete_me =>
+delete_me => _
 
 system_limit/1
 system_limit p => system_limit_body
@@ -479,7 +480,8 @@ send
 # Optimized comparisons with one immediate/literal operand.
 #
 
-is_eq_exact Lbl S S =>
+is_eq_exact Lbl S S => _
+
 is_eq_exact Lbl C1=c C2=c => move C1 x | is_eq_exact Lbl x C2
 is_eq_exact Lbl C=c R=xy => is_eq_exact Lbl R C
 
@@ -697,7 +699,7 @@ get_tuple_element Reg=x P1 D1=y | get_tuple_element Reg=x P2 D2=y | \
 
 get_tuple_element Reg P Dst => i_get_tuple_element Reg P Dst
 
-is_integer Fail=f i =>
+is_integer Fail=f i => _
 is_integer Fail=f an => jump Fail
 is_integer Fail Literal=q => move Literal x | is_integer Fail x
 
@@ -707,7 +709,7 @@ is_integer_allocate f? x t t
 
 is_integer f? xy
 
-is_list Fail=f n =>
+is_list Fail=f n => _
 is_list Fail Literal=q => move Literal x | is_list Fail x
 is_list Fail=f c => jump Fail
 is_list f? x
@@ -738,7 +740,7 @@ is_atom f? x
 %cold
 is_atom f? y
 %hot
-is_atom Fail=f a =>
+is_atom Fail=f a => _
 is_atom Fail=f niq => jump Fail
 
 is_float f? x
@@ -748,7 +750,7 @@ is_float f? y
 is_float Fail=f nai => jump Fail
 is_float Fail Literal=q => move Literal x | is_float Fail x
 
-is_nil Fail=f n =>
+is_nil Fail=f n => _
 is_nil Fail=f qia => jump Fail
 
 is_nil f? xy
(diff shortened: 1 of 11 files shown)


erts/emulator/beam/emu/ops.tab
```diff
@@ -479,15 +502,16 @@ send
 # Optimized comparisons with one immediate/literal operand.
 #
 
-is_eq_exact Lbl S S =>
+is_eq_exact Lbl S S2 | equal(S, S2) => _
```
jhogberg, line 505: Nitpick (same for ARM/x86):
```suggestion
is_eq_exact Lbl LHS RHS | equal(LHS, RHS) => _
```

erts/emulator/beam/emu/ops.tab
```diff
@@ -559,7 +583,7 @@ put_tuple2 xy I *
 #
 put_list Const=c n Dst => move Const x | put_list x n Dst
 
-put_list Src Dst=x Dst => update_list Src Dst
+put_list Src Dst=x Dst2 | equal(Dst, Dst2) => update_list Src Dst
```
jhogberg, line 586: Nitpick:
```suggestion
put_list Head Tail=x Dst | equal(Tail, Dst) => update_list Head Dst
```

erts/emulator/beam/jit/arm/instr_common.cpp
```diff
@@ -1079,12 +1079,8 @@ void BeamModuleAssembler::emit_i_is_tuple(const ArgVal &Fail,
...
         comment("skipped header test since we know it's a tuple when boxed");
 
-        /* Note: ARG1 will NOT be set. This is OK since the operand
-         * for `current_tuple` has a type; that operand will not match
-         * the type-less operand for `get_tuple_element`. Thus, there
-         * will always be a `load_tuple_ptr` instruction emitted if
-         * this instruction is immediately followed by a
-         * `get_tuple_element` instruction. */
+        /* Must still set ARG1. */
+        emit_untag_ptr(ARG1, src.reg);
```
jhogberg, line 1083: This is identical on both branches, maybe hoist it?

erts/emulator/beam/jit/x86/instr_common.cpp
```diff
@@ -1187,15 +1187,11 @@ void BeamModuleAssembler::emit_i_is_tuple(const ArgVal &Fail,
...
-
-        /* Note: ARG2 will NOT be set. This is OK since the operand
-         * for `current_tuple` has a type; that operand will not match
-         * the type-less operand for `get_tuple_element`. Thus, there
-         * will always be a `load_tuple_ptr` instruction emitted if
-         * this instruction is immediately followed by a
-         * `get_tuple_element` instruction. */
+        /* We must be careful to still leave the pointer to the tuple
+         * in ARG2. */
+        mov_arg(ARG2, Src);
```
jhogberg, line 1192: Maybe hoist this move, and change the comment to explain why we're going through `ARG2`?

erts/emulator/beam/jit/x86/instr_common.cpp
```diff
@@ -1187,15 +1187,11 @@ void BeamModuleAssembler::emit_i_is_tuple(const ArgVal &Fail,
...
+        (void)emit_ptr_val(ARG2, ARG2);
+        a.test(ARG2.r8(), imm(_TAG_PRIMARY_MASK - TAG_PRIMARY_BOXED));
```
jhogberg, line 1194: Swapping these improves latency a teeny tiny bit:
```suggestion
        a.test(ARG2.r8(), imm(_TAG_PRIMARY_MASK - TAG_PRIMARY_BOXED));
        (void)emit_ptr_val(ARG2, ARG2);
```

erts/emulator/beam/jit/x86/ops.tab
```diff
@@ -432,8 +456,13 @@ put_tuple2 S A *
 
 store_cons/2
 put_list Hd Tl Dst => put_cons Hd Tl | store_cons u=1 Dst
-store_cons Len Dst | put_list Hd Dst Dst | distinct(Dst, Hd) => combine_conses(Len, Dst, Hd)
-store_cons Len Dst | put_list Hd Dst OtherDst => store_cons Len Dst | append_cons u Hd | store_cons u=1 OtherDst
+
+store_cons Len Dst | put_list Hd Dst2 Dst3 |
+  equal(Dst, Dst2) | equal(Dst, Dst3) | distinct(Dst, Hd) =>
+    combine_conses(Len, Dst, Hd)
```
jhogberg, line 462: Nitpick:
```suggestion
store_cons Len Dst | put_list Head Tail Dst2 |
  equal(Dst, Tail) | equal(Dst, Dst2) | distinct(Dst, Head)  =>
    combine_conses(Len, Dst, Head)
```
(same on the lines below)

erts/emulator/beam/predicates.tab
```diff
@@ -215,27 +215,33 @@ pred.map_key_sort(Size, Rest) {
...
+pred.distinct(Val1, Val2) {
+    if (Val1.type != Val2.type) {
         return 1;
-    } else if (Reg1.type == TAG_x || Reg1.type == TAG_y) {
+    } else if (Val1.type == TAG_x || Val1.type == TAG_y) {
         /* We must not compare the type indices (if any). */
-        return (Reg1.val & REG_MASK) != (Reg2.val & REG_MASK);
+        return (Val1.val & REG_MASK) != (Val2.val & REG_MASK);
+    } else if (Val1.type == TAG_n) {
+        /* NIL has no associated value. */
```
jhogberg, line 226: Do you think it would be better if we normalized the value of `TAG_n` in `beam_file` instead?

jhogberg (approved):

erts/emulator/beam/predicates.tab
bjorng: Yes, will do.

erts/emulator/beam/jit/arm/instr_common.cpp
bjorng: Will do.

erts/emulator/beam/jit/x86/instr_common.cpp
bjorng: But `emit_ptr_val()` will do an `and` when literals are tagged (always on Windows), which will disturb ZF.

erts/emulator/beam/predicates.tab
bjorng: I take it back. I think it's safeer as it is. `beam_file` already sets the value to NIL, but any generator could set the type to `TAG_n` and forget to set the value (or set it to a different value). In fact, [beam_file itself](https://github.com/erlang/otp/blob/07b78d31b1d9edb78aa4a089cc733634ff06e541/erts/emulator/beam/beam_file.c#L1755) sets the type to `TAG_n` without setting the value.

garazdawi (approved):

PR #5679, opened 2022-02-04
Improper tag for private ED keys when encoding

hauleth:
[Documentation states that the tag is `ed_pri` not `ed_priv` for legacy ED private keys][1].

[1]: https://erlang.org/doc/man/public_key.html#type-ed_private_key


hauleth:
Tests failed due to some errors on CI (C compiler cannot produce output).

HansN:
Good catch! I've added it to our internal tests, but it should be no problem I think.

The error in the Github CI seem to be something in the Erlang CI part itself since it's being re-made (I've reported it).

HansN:
@hauleth  Could you please rebase it onto maint so we also get it on OTP-24 and not only on the OTP-25 not yet released?

hauleth:
@HansN done

HansN:
Thanks for the PR!

PR #5683, opened 2022-02-05
fix: handle invalid timeout values returned from gen_server callback functions

pouriya:
`gen_server` module currently does not check returned timeout from `init` and all other `handle_*` callback functions. IMO in this case it should terminate with a `{bad_return_value, _}` too.  

### Example
```erlang
-module(m).
-compile(export_all).
-compile(nowarn_export_all).

start_link(Timeout) ->
    gen_server:start_link({local, ?MODULE}, ?MODULE, Timeout, []).

init(Timeout) ->
    {ok, undefined, Timeout}.

handle_info(_, S) ->
    % Max 'receive' timeout value + 1
    {noreply, S, 16#FFFFFFFF+1}.

terminate(_, _) ->
    ok.
```

#### Before this change
```erlang
1> erlang:process_flag(trap_exit, true).
false

2> m:start_link(-1).
{ok,<0.80.0>} % Dies right after entering the gen_server loop
=CRASH REPORT==== 5-Feb-2022::03:18:40.013997 ===
  crasher:
    initial call: m:init/1
    pid: <0.80.0>
    registered_name: m
    exception error: bad receive timeout value
      in function  gen_server:loop/7 (gen_server.erl, line 411)
    ancestors: [<0.77.0>]
    message_queue_len: 0
    messages: []
    links: [<0.77.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 376
    stack_size: 27
    reductions: 207
  neighbours:

3> flush().
Shell got {'EXIT',<0.80.0>,
              {timeout_value,
                  [{gen_server,loop,7,[{file,"gen_server.erl"},{line,411}]},
                   {proc_lib,init_p_do_apply,3,
                       [{file,"proc_lib.erl"},{line,249}]}]}}
ok


% let it start and live:
4> m:start_link(infinity).
{ok,<0.83.0>}

5> m ! oops.
oops
=CRASH REPORT==== 5-Feb-2022::03:19:09.405236 ===
  crasher:
    initial call: m:init/1
    pid: <0.83.0>
    registered_name: m
    exception error: bad receive timeout value
      in function  gen_server:loop/7 (gen_server.erl, line 411)
    ancestors: [<0.77.0>]
    message_queue_len: 0
    messages: []
    links: [<0.77.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 376
    stack_size: 27
    reductions: 199
  neighbours:

6> flush().
Shell got {'EXIT',<0.83.0>,
              {timeout_value,
                  [{gen_server,loop,7,[{file,"gen_server.erl"},{line,411}]},
                   {proc_lib,init_p_do_apply,3,
                       [{file,"proc_lib.erl"},{line,249}]}]}}
ok
```

#### After this change
```erlang
1> erlang:process_flag(trap_exit, true).
false

2> m:start_link(-1).
{error,{bad_return_value,{ok,undefined,-1}}}

=CRASH REPORT==== 5-Feb-2022::03:24:44.849737 ===
  crasher:
    initial call: m:init/1
    pid: <0.81.0>
    registered_name: m
    exception exit: {bad_return_value,{ok,undefined,-1}}
      in function  gen_server:init_it/6 (gen_server.erl, line 403)
    ancestors: [<0.77.0>]
    message_queue_len: 0
    messages: []
    links: [<0.77.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 376
    stack_size: 27
    reductions: 203
  neighbours:


3> flush().
Shell got {'EXIT',<0.81.0>,{bad_return_value,{ok,undefined,-1}}}
ok


% let it start and live:
4> m:start_link(infinity).
{ok,<0.84.0>}

5> m ! oops.
oops

=ERROR REPORT==== 5-Feb-2022::03:25:09.760708 ===
** Generic server m terminating 
** Last message in was oops
** When Server state == undefined
** Reason for termination ==
** {bad_return_value,{noreply,undefined,4294967296}}

=CRASH REPORT==== 5-Feb-2022::03:25:09.761283 ===
  crasher:
    initial call: m:init/1
    pid: <0.84.0>
    registered_name: m
    exception exit: {bad_return_value,{noreply,undefined,4294967296}}
      in function  gen_server:handle_common_reply/8 (gen_server.erl, line 825)
    ancestors: [<0.77.0>]
    message_queue_len: 0
    messages: []
    links: [<0.77.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 2586
    stack_size: 27
    reductions: 4652
  neighbours:


6> flush().
Shell got {'EXIT',<0.84.0>,{bad_return_value,{noreply,undefined,4294967296}}}
ok
```

### DIFF
--- a/lib/stdlib/src/gen_server.erl
+++ b/lib/stdlib/src/gen_server.erl
@@ -137,6 +137,11 @@
    STACKTRACE(),
    element(2, erlang:process_info(self(), current_stacktrace))).
 
+-define(
+	is_timeout(X),
+	( (X) =:= infinity orelse ( is_integer(X) andalso (X) >= 0 ) )
+).
+
 %%%=========================================================================
 %%%  API
 %%%=========================================================================
@@ -557,7 +562,7 @@ multi_call(Nodes, Name, Request, Timeout)
        ) ->
                         no_return().
 %%
-enter_loop(Mod, Options, State) ->
+enter_loop(Mod, Options, State) when is_atom(Mod) andalso is_list(Options) ->
     enter_loop(Mod, Options, State, self(), infinity).
 
 -spec enter_loop(
@@ -573,33 +578,81 @@ enter_loop(Mod, Options, State) ->
          State   :: term(),
          Timeout :: timeout()
        ) ->
+                        no_return();
+       (
+           Module  :: module(),
+           Options :: [enter_loop_opt()],
+           State   :: term(),
+           Hibernate :: hibernate
+       ) ->
+                        no_return();
+       (
+           Module  :: module(),
+           Options :: [enter_loop_opt()],
+           State   :: term(),
+           Continue :: {continue, term()}
+       ) ->
                         no_return().
 %%
 enter_loop(Mod, Options, State, ServerName = {Scope, _})
-  when Scope == local; Scope == global ->
+    when is_atom(Mod) andalso
+         is_list(Options) andalso
+         (Scope == local orelse Scope == global) ->
     enter_loop(Mod, Options, State, ServerName, infinity);
 %%
-enter_loop(Mod, Options, State, ServerName = {via, _, _}) ->
+enter_loop(Mod, Options, State, ServerName = {via, _, _}) when is_atom(Mod) andalso is_list(Options) ->
     enter_loop(Mod, Options, State, ServerName, infinity);
 %%
-enter_loop(Mod, Options, State, Timeout) ->
-    enter_loop(Mod, Options, State, self(), Timeout).
+enter_loop(Mod, Options, State, TimeoutOrHibernate)
+    when is_atom(Mod) andalso
+         is_list(Options) andalso
+         (?is_timeout(TimeoutOrHibernate) orelse TimeoutOrHibernate =:= hibernate) ->
+    enter_loop(Mod, Options, State, self(), TimeoutOrHibernate);
+%%
+enter_loop(Mod, Options, State, {continue, _}=Continue) when is_atom(Mod) andalso is_list(Options) ->
+    enter_loop(Mod, Options, State, self(), Continue).
 
 -spec enter_loop(
         Module     :: module(),
         Options    :: [enter_loop_opt()],
         State      :: term(),
         ServerName :: server_name() | pid(),
-        Timeout    :: timeout()
+        Timeout    :: timeout() | hibernate | {continue, term()}
+       ) ->
+                        no_return();
+       (
+           Module     :: module(),
+           Options    :: [enter_loop_opt()],
+           State      :: term(),
+           ServerName :: server_name() | pid(),
+           Hibernate    :: hibernate
+       ) ->
+                        no_return();
+       (
+           Module     :: module(),
+           Options    :: [enter_loop_opt()],
+           State      :: term(),
+           ServerName :: server_name() | pid(),
+           Continue    :: {continue, term()}
        ) ->
                         no_return().
 %%
-enter_loop(Mod, Options, State, ServerName, Timeout) ->
+enter_loop(Mod, Options, State, ServerName, TimeoutOrHibernate)
+    when is_atom(Mod) andalso
+         is_list(Options) andalso
+         (?is_timeout(TimeoutOrHibernate) orelse TimeoutOrHibernate =:= hibernate) ->
     Name = gen:get_proc_name(ServerName),
     Parent = gen:get_parent(),
     Debug = gen:debug_options(Name, Options),
     HibernateAfterTimeout = gen:hibernate_after(Options),
-    loop(Parent, Name, State, Mod, Timeout, HibernateAfterTimeout, Debug).
+    loop(Parent, Name, State, Mod, TimeoutOrHibernate, HibernateAfterTimeout, Debug);
+%%
+enter_loop(Mod, Options, State, ServerName, {continue, _}=Continue) when is_atom(Mod) andalso is_list(Options) ->
+    Name = gen:get_proc_name(ServerName),
+    Parent = gen:get_parent(),
+    Debug = gen:debug_options(Name, Options),
+    HibernateAfterTimeout = gen:hibernate_after(Options),
+    loop(Parent, Name, State, Mod, Continue, HibernateAfterTimeout, Debug).
 
 %%%========================================================================
 %%% Gen-callback functions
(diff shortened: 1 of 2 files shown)


pouriya:
> Checking for the basic type and range catches practically all errors.  

What if some users calculate timeout dynamically (e.g. read the value from config file)?  
<br/><br/>


> Better to let it crash in the receive statement is that upper limit is violated.

In any case (With or without these changes) if we violate the limit the code crashes. The only difference is that with these changes we have a normal crash with complete report:  
```erlang
=ERROR REPORT==== 5-Feb-2022::03:25:09.760708 ===
** Generic server m terminating 
** Last message in was oops
** When Server state == undefined
** Reason for termination ==
** {bad_return_value,{noreply,undefined,4294967296}}
```
<br/><br/>
 
> it seems that your changes in handle_msg/6 and in handle_common_reply/8 does not handle the hibernate and {continue,_} cases.  

Yes you're right! I think that the variable name `Time1` confused me!
<br/><br/>


### Questions for next changes
We should just go for `((X) =:= infinity orelse (is_integer(X) andalso (X) >= 0))` and do not check upper bound. Am I right?  
We need new clauses in `handle_msg/6` and in `handle_common_reply/8` to check for `hibernate` and `{continue, _}`. Do we need to check more details too?
For example:  
```erlang
% inside handle_msg/6 or `handle_common_reply/8 checks:
    {noreply, S, Timeout} when ?is_timeout(Timeout) ->
        ...;
    {noreply, S, hibernate} ->
        ...;
    {noreply, S, {continue, _}} ->
        ...;

% or
-define(
    is_timeout_or_hibernate_or_continue(X)
    (((X) =:= infinity orelse (is_integer(X) andalso (X) >= 0)) orelse X =:= hibernate orelse (tuple_size(X) == 2 andalso element(1, X) == continue))
).
% ...
% inside handle_msg/6 or `handle_common_reply/8 checks:
    {noreply, S, TimeoutOrHibernateOrContinue} when ?is_timeout_or_hibernate_or_continue(TimeoutOrHibernateOrContinue) ->
        ...;

```
Should `enter_loop/4-5` also check the Timeout argument? I think yes they should but I forgot to add the check. Should I add it?

RaimoNiskanen:
> What if some users calculate timeout dynamically (e.g. read the value from config file)?

If it violates the type (`integer() >= 0 | 'infinity'`) an informative error is in order.  But it is (theoretically) very hard or impossible to predict what the upper integer limit is since it is in fact a system limit and how that behaves is subject to change.  For instance; a reasonable future extension would be to allow absolute time-outs, and then we have a dynamically calculated time-out value that for a specific absolute time-out time might or might not  be accepted by `receive`.  A static type check of the value can not know if it is ok, and then I think a `{bad_return_value,_}` error exception is misguiding.

The simplest solution is to let `receive` do its thing and crash for a too large time-out if that should happen in run time.


> Do we need to check more details too?

Checking for `timeout() | 'hibernate' | {'continue',_}` would be sufficient.

I would prefer to not have one gigantic macro for "is timeout or hibernate or continue", and instead separate them and/or use multiple clauses.  I think `{continue,_}` is more readable than `is_continue(Time)`.  `is_timeout(Time) orelse Time =:= hibernate` could be on one clause.

A warning: `tuple_size(T), element(1, T) =:= continue` is a neat trick in a guard, but if you write a macro `is_continue(T)` as `(tuple_size(T) =:= 2 andalso element(1, T) =:= continue)` and later use it as `is_continue(T) orelse T =:= hibernate` I think you would have a bug; `tuple_size(hibernate)` should crash and the whole guard fail before testing `T =:= hibernate`. But `T =:= hibernate orelse is_continue(T)` would accidentally work.  So a general guard macro using `andalso`/`orelse` should not skip the `is_tuple(T)` test before `tuple_size(T)`...

> Should enter_loop/4-5 also check the Timeout argument? I think yes they should but I forgot to add the check. Should I add it?

I think that would be nice to have.  What should `enter_loop/5` check, then...?
That `Mod` is an atom, that `Options` is a list, and that `Timeout` is a time-out, `'hibernate'` or `{continue,_}`, I'd say.  `ServerName` will cause a crash in `gen:get_proc_name/1` if incorrect; that should be ok.

The type spec now only allows `Timeout :: timeout()`but that is not what the code does...  There is no apparent reason to prohibit `hibernate` - one might want to enter hibernation immediately, and I do not realize why it should be impossible to go right into `handle_continue/2`.

pouriya:
Another question: For `{'$gen_call', _, _}` message checks we have [these](https://github.com/erlang/otp/blob/OTP-24.3/lib/stdlib/src/gen_server.erl#L758-L761) checks and we have the same checks in `handle_common_reply` [here](https://github.com/erlang/otp/blob/OTP-24.3/lib/stdlib/src/gen_server.erl#L806-L809). Why do we need them in `handle_msg({'$gen_call', From, Msg} ...)` function when this function also calls `handle_common_reply` [here](https://github.com/erlang/otp/blob/OTP-24.3/lib/stdlib/src/gen_server.erl#L768)?

RaimoNiskanen:
> Why do we need them in handle_msg({'$gen_call', From, Msg} ...) function when this function also calls handle_common_reply [here](https://github.com/erlang/otp/blob/OTP-24.3/lib/stdlib/src/gen_server.erl#L768)?

I see no point.  It might be an optimization.  I do not know how much of a performance gain it might give...  It sure is tempting to remove the clauses from handle_msg, to improve clarity!

RaimoNiskanen:
Unfortunately, now when the last 24 maintenance release is out, this PR will have to be rebased to 'master'...  After that we have code freeze for 25.rc2 on Friday.  Hopefully we can get this in.

pouriya:
Here (in Iran) we're getting close to [Nowruz](https://en.wikipedia.org/wiki/Nowruz) (It's celebrated on Saturday). So l can change things over the next two days and before the holidays. Anything else?

RaimoNiskanen:
I can not think of anything else to fix.

Regarding optimization I sense that the argument order of `handle_common_reply/8,9` could be improved to minimize register moves, but it is better to see what our benchmarks say before changing that, and I can do it myself on top of your branch.

So, go ahead and rebase...

lib/stdlib/src/gen_server.erl
```diff
@@ -617,23 +617,23 @@ enter_loop(Mod, Options, State, {continue, _}=Continue) when is_atom(Mod) andals
         Options    :: [enter_loop_opt()],
         State      :: term(),
         ServerName :: server_name() | pid(),
-        Timeout    :: timeout() | hibernate | {continue, term()}
+        Timeout    :: timeout()
```
pouriya, line 620: Oops! Thank you 👍🏼

RaimoNiskanen:
I forgot to mention that the branch base should also be changed from 'maint' to 'master' in GitHubs GUI, so it looked like a mess for a while...

Also, documentation build failed due to `-spec` changes, which is a blocking issue for our daily builds.

pouriya:
I think that `andalso` and `orelse` are simpler to understand. What is your opinion? What if we update  all traditional comma-and-semicolon separated guards?

RaimoNiskanen:
My opinion is that comma-and-semicolon guards are easier to read, when line broken consistently.  Probably since I have used Erlang since before `andalso` and `orelse` were usable in guards...

So this is a matter of taste, therefore we prefer to keep the style of the module.

RaimoNiskanen:

PR #5684, opened 2022-02-05
Fix url [doc] [ci skip]

marcandre:


rickard-green:
Please do not type [ci skip] in commit messages in the future, since it removes build tests we want to have executed on all PRs. You do not have to change this PR or #5685, though.

PR #5688, opened 2022-02-07
Further strengthen the type-based optimizations

bjorng:
Building on #5316 and #5664, this pull request annotates more arithmetic instructions with type information to allow the JIT to generate better code.

With this PR, the compiler and the JIT can now optimize a tuple used as a lookup table such as in the following example (similar to real code in the `base64` module):

```
foo(N0) ->
    N1 = N0 band 3,
    N = N1 + 1,
    element(N, {zero,one,two,three}).
```

Because `N` is known to be an integer in the range 1 through 4, the JIT will omit all type and range checks and generate inlined code to fetch the `N`th element of the tuple.

### DIFF
--- a/erts/emulator/beam/jit/x86/instr_guard_bifs.cpp
+++ b/erts/emulator/beam/jit/x86/instr_guard_bifs.cpp
@@ -162,6 +162,72 @@ void BeamModuleAssembler::emit_bif_element(const ArgVal &Fail,
                      signed_val(Pos.getValue()) > 0 &&
                      signed_val(Pos.getValue()) <= (Sint)MAX_ARITYVAL;
 
+    /*
+     * Try to optimize the use of a tuple as a lookup table.
+     */
+    if (exact_type(Pos, BEAM_TYPE_INTEGER) && Tuple.isLiteral()) {
+        Eterm tuple = beamfile_get_literal(beam, Tuple.getValue());
+        if (is_tuple(tuple)) {
+            Label error = a.newLabel(), next = a.newLabel();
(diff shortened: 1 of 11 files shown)


erts/emulator/beam/jit/x86/instr_guard_bifs.cpp
```diff
@@ -162,6 +162,72 @@ void BeamModuleAssembler::emit_bif_element(const ArgVal &Fail,
...
+            auto [min, max] = getIntRange(Pos);
+            bool is_bounded = min <= max;
+            bool can_fail = is_bounded || min < 1 || size < max;
+
+            comment("skipped tuple test since source is always a literal "
+                    "tuple");
+            mov_arg(ARG2, Tuple);
+            mov_arg(ARG1, Pos);
+            x86::Gp boxed_ptr = emit_ptr_val(ARG3, ARG2);
+            a.lea(ARG4, emit_boxed_val(boxed_ptr));
```
garazdawi, line 182: Can't we just load the untagged tuple pointer directly into ARG4?

bjorng: There does not seem to be an easy way to do that.

The literal is not at its final position, so the instruction that loads the tuple pointer into `ARG2` will be patched later. To load an untagged pointer, we would have to extend the patch machinery in the loader to support untagging a pointer while patching.

garazdawi: Ah, yes of course. I had forgotten about that.

PR #5693, opened 2022-02-07
erts: Fix proc bin writable trace fixup 

garazdawi:
When creating a new proc bin because tracing has removed the
writable flag from the proc bin, we must make sure to also
create a new sub binary if the sub binary is on the mature
or old heap. When this is not done, the subbinary can be
promoted to the old heap without the proc bin also being there.

Bug was introduced in #5195, aka e16e545

### DIFF
--- a/erts/emulator/beam/erl_bits.c
+++ b/erts/emulator/beam/erl_bits.c
@@ -1729,18 +1729,31 @@ erts_bs_private_append_checked(Process* p, Eterm bin, Uint build_size_in_bits, U
 	     */
 	    Binary* bptr = erts_bin_nrml_alloc(new_size);
             ProcBin* new_pb;
+            Uint sz = PROC_BIN_SIZE;
 
             sys_memcpy(bptr->orig_bytes, binp->orig_bytes, binp->orig_size);
 
-            new_pb = (ProcBin*) HeapFragOnlyAlloc(p, PROC_BIN_SIZE);
+            /* If the subbinary is on the mature or old heap, we need to also move it */
+            if (ErtsInArea(sb, OLD_HEAP(p), ((char *)OLD_HTOP(p)) - ((char *)OLD_HEAP(p))) ||
+                ErtsInArea(sb, HEAP_START(p), ((char *)HIGH_WATER(p)) - ((char *)HEAP_START(p)))) {
+                sz += ERL_SUB_BIN_SIZE;
+            }
+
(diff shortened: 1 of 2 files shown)


erts/emulator/beam/erl_bits.c
```diff
@@ -1729,18 +1729,31 @@ erts_bs_private_append_checked(Process* p, Eterm bin, Uint build_size_in_bits, U
...
 	    Binary* bptr = erts_bin_nrml_alloc(new_size);
             ProcBin* new_pb;
+            Uint sz = PROC_BIN_SIZE;
 
             sys_memcpy(bptr->orig_bytes, binp->orig_bytes, binp->orig_size);
 
-            new_pb = (ProcBin*) HeapFragOnlyAlloc(p, PROC_BIN_SIZE);
+            /* If the subbinary is on the mature or old heap, we need to also move it */
+            if (ErtsInArea(sb, OLD_HEAP(p), OLD_HTOP(p) - OLD_HEAP(p)) ||
+                ErtsInArea(sb, HEAP_START(p), HIGH_WATER(p) - HEAP_START(p))) {
```
sverker, line 1738: ErtsInArea wants nr of **bytes**, not Eterm's.

May we need another macro ErtsInBetween(ptr, start, end).

erts/emulator/beam/erl_bits.c
```diff
@@ -1729,18 +1729,31 @@ erts_bs_private_append_checked(Process* p, Eterm bin, Uint build_size_in_bits, U
...
             new_pb->bytes = (byte *) bptr->orig_bytes;
             new_pb->next = p->wrt_bins;
             p->wrt_bins = (struct erl_off_heap_header*) new_pb;
-            sb->orig = make_binary(new_pb);
             pb = new_pb;
+            if (sz != PROC_BIN_SIZE) {
+                ErlSubBin *new_sb = (ErlSubBin*)(new_pb+1);
+                sys_memcpy(new_sb, sb, sizeof(*new_sb));
+                sb = new_sb;
+                bin = make_binary(sb);
```
sverker, line 1754: This makes callers of erts_bs_private_append_checked to actually have to take care of the return value even if it's != THE_NON_VALUE.
But, from what I can see, they do.

garazdawi: Yes, at first I thought something would break when I did this but seems like every caller has done the correct thing (tm).

sverker (approved):

PR #5694, opened 2022-02-08
Provide more information for failed record operations

bjorng:
When a record operation fails, a `{badrecord,Tag}` error exception
is raised, where `Tag` is the name of the record. For example, if
we have this function:

    -record(foo, {a,b,c,d}).

    bar(R) ->
        R#foo.a.

we will get the following exception if we call `bar/1` with something
else than a #foo{} record:

    1> t:bar({x,y,z}).
    ** exception error: {badrecord,foo}
         in function  t:bar/1 (t.erl, line 7)

When records were first introduced, there were no line numbers in
exceptions, so if a record operation failed in a large function
knowing the record name could be useful.

Nowadays, the line number usually makes it obvious which kind of
record the failed operation expected. It would be more useful to
know the actual value passed to the record operation.

This pull request changes the exception for a failed record operation to
`{badrecord,ActualValue}`. For example:

    1> t:bar({x,y,z}).
    ** exception error: {badrecord,{x,y,z}}
         in function  t:bar/1 (t.erl, line 7)
    2> t:bar(lists:seq(1, 1000)).
    ** exception error: {badrecord,[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,
                                    20,21,22,23,24,25,26,27|...]}
         in function  t:bar/1 (t.erl, line 7)

Note that it is not documented what the exception for a failed record
operation is supposed to look like. Furthermore, it is unlikely that
there is existing code that catches expressions involving records
operations and tries to match the exact error reason. Therefore, we
don't consider this an incompatible change.


michalmuskala:
What is the reason for introducing a new instruction rather than having `erlang:error({badrecord, Value})`. Is it to avoid the instruction sequence to construct the error tuple?

bjorng:
> Is it to avoid the instruction sequence to construct the error tuple?

Yes, especially for the JIT. For x86_64, the `badrecord` instruction consists of 5 instructions, while the sequence to build the tuple and call error/1 is about 20 instructions.

michalmuskala:
Out of curiosity - what is the main source of the size of those instructions? Is this the GC check to allocate the tuple?

bjorng:
Given this BEAM code:

```
    {test_heap,3,1}.
    {put_tuple2,{x,0},{list,[{atom,badrecord},{x,0}]}}.
    {line,[{location,"t.erl",34}]}.
    {call_ext_only,1,{extfunc,erlang,error,1}}.
```
The following code is generated:

```
# test_heap_It
    lea rdx, qword ptr [r15+56]
    cmp rdx, rsp
    short jbe L14
    mov ecx, 1
    call 4312877104
L14:
# put_tuple2_SA
# Move arity word
    mov edi, 128
    mov qword ptr [r15], rdi
# Move tuple data
    mov qword ptr [r15+8], 91211
    mov rdi, qword ptr [rbx]
    mov qword ptr [r15+16], rdi
# Create boxed ptr
    lea rdi, qword ptr [r15+2]
    add r15, 24
    mov qword ptr [rbx], rdi
# line_I
# allocate_tt
    lea rdx, qword ptr [r15+32]
    cmp rdx, rsp
    short jbe L15
    mov ecx, 1
.db 0x66, 0x90
    call 4312877104
L15:
# call_light_bif_be
    align 4
L16:
L17:
    long mov rcx, 9223372036854775807
    mov rax, 4302763136
    lea rdx, qword ptr [L16]
.db 0x0F, 0x1F, 0x00
    call 4312875968
```

The `allocate_tt` instruction was needed in this case because there was no a stack frame previously. It would not have been needed if there have already been a stack frame.

Issue #5368, opened 2021-11-10
No appropriate replacement for http_uri:encode/1 in uri_string


samwar:
**Is your feature request related to a problem? Please describe.**
OTP 25 is set to remove the `http_uri:encode/1` function without having an appropriate replacement in the `uri_string` module. I've seen discussion around using the `compose_query` function for encoding, but that does not work for encoding parts of the path, and another [closed issue](https://github.com/erlang/otp/issues/4607) that doesn't get to the heart of the problem.

The API I work on is used to manage network appliances, and our paths include ids that can have `/` in them along with other characters that need to be encoded. For instance, if we want to get information about a specific interface on a switch (say Ethernet0/4), a user would navigate to `https://internal.api.com/switches/Ethernet0%2F4`, which would display information about how that interface is configured and a set of links with other operations the user can perform with the interface id. We generate those links using the `http_uri:encode/1` function. I have tried using the `url_string:normalize`, `url_string:resolve`, and `url_string:transcode` functions in the hopes they would encode the interface id correctly, but the Ethernet0/4 interface id is not interpreted as id, but as two separate path elements, `Ethernet0` and `4`. We also use the `http_uri:encode/1` function to build URIs so as to test our API out in our CT integration tests.

```erlang
1> InterfaceId = "Ethernet0/4".
"Ethernet0/4"
2> uri_string:normalize(InterfaceId).
"Ethernet0/4"
3> uri_string:transcode(InterfaceId, [{in_encodeing, utf8}, {out_encoding, utf8}]).
"Ethernet0/4"
4> uri_string:resolve("/switches/Ethernet0/4", "http://localhost/a/b/c?q").
"http://localhost/switches/Ethernet0/4"
5> uri_string:resolve("/switches/Ethernet0/4", "http://localhost/a/b").    
"http://localhost/switches/Ethernet0/4"
6> uri_string:resolve(#{path => InterfaceId}, "http://localhost/switches/b/c").  
"http://localhost/switches/b/Ethernet0/4"
7> uri_string:resolve(#{path => InterfaceId}, "http://localhost/switches/a/b").  
"http://localhost/switches/a/Ethernet0/4"
8> uri_string:resolve(#{path => InterfaceId}, "http://localhost/switches/a/b/c/d").
"http://localhost/switches/a/b/c/Ethernet0/4"
9> http_uri:encode(InterfaceId).   
"Ethernet0%2F4"
10> uri_string:percent_decode(http_uri:encode(InterfaceId)).
[...]

PR #5700, opened 2022-02-09
stdlib: add quote, unquote functions in uri_string

u3s:
- functions for percent encoding and decoding user data
- intended as utility functions for preparing URI components


samwar:
🎊  It's great to see a partial solution for #5368. Also interesting that the stdlib was going to run into a problem when `http_uri:encode/decode` were removed.

samwar:
Any thought to writing a set of tests to show that `http_uri:encode/decode` work the same way as 'uri_string:quote/unquote`? It would give us all a higher degree of confidence that the new functions are a drop-in replacement.

u3s:
> Any thought to writing a set of tests to show that `http_uri:encode/decode` work the same/similar way as 'uri_string:quote/unquote`? It would give us all a higher degree of confidence that the new functions are a drop-in replacement.

Sure why not, I can add it for transition period.

leoliu:
I am curious why not call them `encode/decode` as they are usually called? What to do about the api `uri_string:percent_decode/1`?

u3s:
> I am curious why not call them `encode/decode` as they are usually called? What to do about the api `uri_string:percent_decode/1`?

that's a good question and I think maybe should be somehow addressed in a doc update as well.
I guess it's an attempt of choosing less confusing name in general (probably not perfect).

My thinking is that, uri_string also today contains 2 categories of functions
1. those which take URIMap, URIString types as input
2. those which don't

1st group - can act in a component wise manner and preferably implement RFC3986;

for example if path component provided by user contains characters not allowed in that specific component and `uri_string:recompose` is called, they will be automatically percent encoded.

`uri_string:percent_decode` belongs to 1st group as well, I think its primary intention was to revert automatic percent encoding (although its implementation can be used for general percent decode and overlap with `unquote/1` which I don't like :-/)

2nd group - utility functions, to be used in preparations before calling function from 1st group. `quote, unquote` would be member of it as well.
intention of not using verb "encode" or "percent_encode" was to make that characteristic easier to notice.

see also #5368 for more background

PR #5705, opened 2022-02-11
Peer fixes

rickard-green:
* `peer:stop()` currently by default stop a peer node, which is supervised over an erlang distribution connection, by closing the connection. This can cause other code on the peer node to react in undesirable ways to broken links, `'DOWN'` messages, and `nodedown` messages before the `peer` code on the peer node is able to take down the node. This PR makes `peer:stop()` first try to stop the node properly before, as a last resort, disconnecting from it.
* `peer:stop()` could return before a potential connection to the peer node had been closed which could cause issues for some test cases. This PR ensures that `peer:stop()` doesn't return until the connection has been closed.
* Added missing license/copyright header in `peer.erl` (reused the same as used on `pg`).

### DIFF
--- a/lib/stdlib/doc/src/peer.xml
+++ b/lib/stdlib/doc/src/peer.xml
@@ -379,6 +379,11 @@
         is specified instead of a timeout, the peer will send <c>Tag</c> to the
         requested process.</p></desc>
     </datatype>
+    <datatype>
+      <name name="disconnect_timeout"/>
+      <desc><p>Disconnect timeout. See
+      <seemfa marker="#stop/1"><c>stop()</c></seemfa>.</p></desc>
+    </datatype>
   </datatypes>
 
   <funcs>
@@ -525,12 +530,76 @@
     <func>
       <name name="stop" arity="1" since="OTP @OTP-17720@"/>
       <fsummary>Stop controlling process and terminate peer node.</fsummary>
+      <type name="disconnect_timeout"/>
       <desc>
-        <p>Depending on the <c>shutdown</c> option used to start the peer,
-        stopping may be graceful (via <c>init:stop()</c>) or less so,
-        by just closing the control connection, which triggers <c>erlang:halt()</c>
-        executed by the peer.</p>
-        <p>A graceful shutdown can be slower, or even block indefinitely.</p>
+        <p>
+          Stops a peer node. How the node is stopped depends on the
+          <seetype marker="#start_options"><c>shutdown</c></seetype>
+          option passed when starting the peer node. Currently the
+          following <c>shutdown</c> options are supported:
+        </p>
+        <taglist>
+          <tag><c>halt</c></tag>
+          <item><p>
+            This is the default shutdown behavior. It behaves as <c>shutdown</c>
+            option <c>{halt, DefaultTimeout}</c> where <c>DefaultTimeout</c>
+            currently equals <c>5000</c>.
+          </p></item>
+          <tag><c>{halt, Timeout :: disconnect_timeout()}</c></tag>
+          <item><p>
+            Triggers a call to
+            <seemfa marker="erts:erlang#halt/0"><c>erlang:halt()</c></seemfa>
+            on the peer node and then waits for the Erlang distribution
+            connection to the peer node to be taken down. If this connection
+            has not been taken down after <c>Timeout</c> milliseconds, it will
+            forcefully be taken down by <c>peer:stop/1</c>. See the
+            <seeerl marker="#dist_connection_close">warning</seeerl> below for
+            more info about this.
+          </p></item>
+          <tag><c>Timeout :: disconnect_timeout()</c></tag>
+          <item><p>
+            Triggers a call to
+            <seemfa marker="erts:init#stop/0"><c>init:stop()</c></seemfa>
+            on the peer node and then waits for the Erlang distribution
+            connection to the peer node to be taken down. If this connection
+            has not been taken down after <c>Timeout</c> milliseconds, it will
+            forcefully be taken down by <c>peer:stop/1</c>. See the
+            <seeerl marker="#dist_connection_close">warning</seeerl> below for
+            more info about this.
+          </p></item>
+          <tag><c>close</c></tag>
+          <item><p>
+            Close the <i>control connection</i> to the peer node and
+            return. This is the fastest way for the caller of
+            <c>peer:stop/1</c> to stop a peer node.
+          </p>
+          <p>
+            Note that if the Erlang distribution connection is not used as
+            control connection it might not have been taken down when
+            <c>peer:stop/1</c> returns. Also note that the
+            <seeerl marker="#dist_connection_close">warning</seeerl> below
+            applies when the Erlang distribution connection is used as control
+            connection.
+          </p>
+          </item>
+        </taglist>
+
+        <marker id="dist_connection_close"/>
+        <warning>
+          <p>
+            In the cases where the Erlang distribution connection is taken
+            down by <c>peer:stop/1</c>, other code independent of the peer
+            code might react to the connection loss before the peer node is
+            stopped which might cause undesirable effects. For example,
+            <seeerl marker="kernel:global#prevent_overlapping_partitions"><c>global</c></seeerl>
+            might trigger even more Erlang distribution connections to other
+            nodes to be taken down. The potential undesirable effects are,
+            however, not limited to this. It is hard to say what the effects
+            will be since these effects can be caused by any code with links
+            or monitors to something on the origin node, or code monitoring
+            the connection to the origin node. 
+          </p>
+        </warning>
       </desc>
     </func>
 
--- a/lib/stdlib/src/peer.erl
+++ b/lib/stdlib/src/peer.erl
@@ -112,8 +112,10 @@
           args => [string()],                 %% additional command line parameters to append
           env => [{string(), string()}],      %% additional environment variables
           wait_boot => wait_boot(),           %% default is synchronous start with 15 sec timeout
-          shutdown => brutal_kill |           %% halt the peer brutally (default)
-          timeout()                       %% send init:stop() request and wait up to specified timeout
+          shutdown => close |                 %% close supervision channel
+          halt | %% The default...            %% stop node using erlang:halt() wait default timeout for nodedown
+          {halt, disconnect_timeout()} |      %% stop node using erlang:halt() wait timeout() for nodedown
+          disconnect_timeout()                %% send init:stop() request and wait up to specified timeout for nodedown
          }.
 
 %% Peer node states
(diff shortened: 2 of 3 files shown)


lib/stdlib/src/peer.erl
```diff
@@ -372,28 +390,71 @@ handle_info({tcp_closed, Sock}, #peer_state{connection = Sock} = State) ->
...
+            end;
         {Shutdown, error} ->
             Timeout = shutdown(dist, undefined, Node, Shutdown),
-            receive {nodedown, Node} -> ok after Timeout -> ok end;
+            receive {nodedown, Node} -> ok
+            after Timeout -> _ = erlang:disconnect_node(Node), ok
+            end;
         {Shutdown, {ok, standard_io}} ->
             Timeout = shutdown(port, Port, Node, Shutdown),
+            Deadline = erlang:monotonic_time(millisecond) + Timeout,
```
rickard-green, line 410: Hmm, I need to take care of timeout being `infinity` as well I guess...

lib/stdlib/src/peer.erl
```diff
@@ -372,28 +390,76 @@ handle_info({tcp_closed, Sock}, #peer_state{connection = Sock} = State) ->
...
+                                     true -> TL
+                                  end
+                          end,
+                    receive {nodedown, Node, _} -> ok
+                    after Tmo ->  _ = erlang:disconnect_node(Node), ok
+                    end
+            end
+    end.
+
 %% This hack is a temporary workaround for test coverage reports
```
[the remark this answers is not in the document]
max-au: It absolutely is! Thank you for fixing.

max-au:
My bad, I haven't documented `shutdown` option properly, so my intent around it wasn't clear.

The idea is, `brutal_kill` option is exactly designed to halt the node in the most brutal way. The purpose of it is to execute unclean shutdown, and if there is alternative connection, without waiting for the distribution connection to go down. The primary benefit of this is significantly faster action.  Due to some `sleep` calls on the shutdown path, graceful shutdown takes significantly longer (IIRC there is at least 500 ms sleep somewhere in graceful shutdown via `init:stop`).
With `brutal_kill`, code coverage reports aren't passed to the `cover` server (intentionally - this is Sparta!)

So if test case needs a graceful shutdown (which is always true for `cover` runs), I'd suggest using `?CT_PEER(#{name => ?CT_PEER_NAME(), shutdown => 5000})`. Unlike `brutal_kill`, this mode is designed to do what you were after.

I would really love to keep it that way, because otherwise there is no way to have fast test cases, especially property-based test for `pg` (because of how slow the shutdown becomes).

There are two ways forward I see. One is to explicitly add `shutdown => 5000` to all cases where it is important that peer shuts down gracefully. Another is to change the default shutdown mode from `brutal_kill` to 5000 (then I will update test cases that don't use `global` or don't care about links state so they run faster, e.g. `pg`, especially SPBT one).

max-au:
I realised there is a third way as well, add `peer:stop/2` accepting Timeout overriding initial `shutdown` setting and requesting graceful shutdown. Maybe even add `?CT_PEER_STOP(P)` macro to make API symmetrical.

watsy0007:
OTP-25.0 ? [OTP-25.0](https://github.com/erlang/otp/milestone/152)

rickard-green:
My main concern is peer:stop() using disconnect as a way of stopping the node in the default case. This since not just global, but also other code executing on the peer node might cause trouble when it is just disconnected and letting that code be exposed to broken links, DOWNs, and nodedowns. `erpc:call(Node,erlang,halt,[],Timeout)` should also be very quick in the normal case. It just terminates the runtime and doesn't try to be graceful as init:stop().

Waiting for connections to be taken down before returning from peer:stop() is another issue, and should perhaps be handled in a different way. Perhaps using different configuration parameter.

I'm too tired to think more about it right now, though. Haven't slept all night and have a meeting in about two hours, so I think I'll have to try to get some sleep and get back to you later on this.

BTW are you happy with that license/copyright commit or do you want something else as copyright?

max-au:
I had a sleepless night too, so I think now I understand what you mean. The `brutal_kill` may be indeed too brutal for default case. I agree that one more shutdown method (`halt`?) makes more sense to be the default.
Would that work for your case too?

rickard-green:
Yes, I think that a `halt` version as default with `brutal_kill` as optional would be ok assuming that `halt` work as `brutal_kill` currently is working in this PR. Not waiting for the connection to go down in the default case broke a lot of tests. We however have to postpone work and testing of this until at least some time late next week. We are about to release a release candidate, so we cannot afford breaking tests with new stuff right now, and I also have a bunch of other things I need to take care of right now.

rickard-green:
Regarding the `brutal_kill` shutdown option. It could at least on unix like platforms send the `SIGKILL` signal to the VM of the peer node. The name would then, at least to me, better represent what it does. It would also prevent buggy, e.g. deadlocked, VMs hanging around.

max-au:
It's been my original implementation of `brutal_kill`, but I could not make it beautiful enough. 
It was remembering OS pid returned by `open_port` in `peer_state` and then using `os:find_executable("kill")` to do `open_port(KillCmd, [{args, ["-9", OSPid]}])` (on Windows it was looking for "taskkill"). The code looked ugly and repetitive, so I tried looking for some BIF that can send a signal to OS process. I failed to find it, is there an implementation for that? I think it would be great to add `os:signal(OSPid, sigkill)`, and then use it in `peer` for brutal_kill.

PS. Thank you for fixing the copyright header, I somehow missed it.

rickard-green:
@max-au What about this?

Compared to what is in `master` now:
* Default to the new shutdown type `halt` which performs `erlang:halt()` and waits for dist connection to go down.
* Graceful shutdown using `init:stop()` also waits for dist connection to go down.
* Old shutdown type `brutal_kill` renamed to `close`, since it just close connection used for supervision and then return. This makes it possible to introduce a `brutal_kill` option in the future which sends `SIGKILL` to the peer VM.

max-au:
Looks good to me!
One suggestion I'd like to make is to add a few test cases verifying specific properties of a selected shutdown mode. Otherwise it'd be easy to forget why these specific choices were made.

rickard-green:
I've added test cases and updated the documentation.

rickard-green:
I've measured a couple of milliseconds longer for `halt` compared to `close`, but the difference will depend on the load of the system.

lib/stdlib/doc/src/peer.xml
```diff
@@ -526,11 +526,81 @@
...
+            <seeerl marker="#dist_connection_close">warning</seeerl> below
+            applies when the Erlang distribution connection is used as control
+            connection.
+          </p>
+          </item>
+        </taglist>
+
+        <p>
+          Currently integer <c>Timeout</c> values less than <c>1000</c> will
+          silently be increased to <c>1000</c>, and integer <c>Timeout</c>
```
max-au, line 583: To be honest, I'd rather have `verify_args` to throw when an invalid (<1000 or > max 32 bit int) is passed. Silently correcting invalid argument tends to swallow errors instead of surfacing them early (in `peer:start_link`).

Leaving the decision to you, but I introduced `verify_args` specifically to address that (since it was too easy to pass some invalid node names and then waste a lot of time trying to figure out what's wrong with that peer).

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -328,6 +336,59 @@ stop_peer(Config) when is_list(Config) ->
...
+                      erts_debug:set_internal_state(available_internal_state, true),
+                      erts_debug:set_internal_state(block, BlockTime+100)
+              end),
+    receive after 100 -> ok end,
+    Start = erlang:monotonic_time(millisecond),
+    peer:stop(Peer),
+    End = erlang:monotonic_time(millisecond),
+    WaitTime = End - Start,
+    Res = lists:member(Node, nodes(connected)),
+    io:format("Connected = ~p~nWaitTime = ~p~n", [Res, WaitTime]),
```
max-au, line 387: To my understanding, `ct:print` (or, `pal`, `log`) is the preferred way to print/log something within a test suite.
This allows to add fine-grained filters (e.g. print only when test case fails, because in most cases developers aren't interested in the output of successful cases).

max-au (commented):
Looks good in general, the only (non-internal) question to me is whether to fail `start_link` calls for invalid timeout, or silently correct.

lib/stdlib/doc/src/peer.xml
rickard-green: I've changed it to fail in `verify_args`

lib/stdlib/test/peer_SUITE.erl
rickard-green: Using `ct:pal` now

lib/stdlib/src/peer.erl
```diff
@@ -253,7 +286,16 @@ init([Notify, Options]) ->
                 receive {'EXIT', Port, _} -> undefined end
         end,
 
-    State = #peer_state{options = Options, notify = Notify, args = Args, exec = Exec},
+    %% Remvoe the default 'halt' shutdown option if present; the default is
```
max-au, line 289: Typo - "Remvoe".

lib/stdlib/src/peer.erl
```diff
@@ -370,38 +412,101 @@ handle_info({tcp_closed, Sock}, #peer_state{connection = Sock} = State) ->
...
+                          end,
+                    receive {nodedown, Node, _} -> ok
+                    after Tmo -> force_disconnect_node(Node)
+                    end
+            end
+    end.
+
+force_disconnect_node(Node) ->
+    _ = erlang:disconnect_node(Node),
+    error_logger:warning_msg("peer:stop() timed out waiting for disconnect from "
```
max-au, line 497: For my education, is `error_logger` still recommended to use? With new `logger` and `?LOG_WARNING` macro available, I'd think of using those instead of `error_logger` (which I thought was deprecated).

Also, it'd be great to have some guidelines on logging within OTP itself. I'd think of adding `otp` domain to all messages emitted within OTP itself (for easier filtering).

BTW is it necessary to log this event?

max-au (approved):
I'm not sure about extra logging (especially via `error_logger`, but the rest feels good.

lib/stdlib/src/peer.erl
rickard-green: Fixed

lib/stdlib/src/peer.erl
rickard-green: > For my education, is `error_logger` still recommended to use? With new `logger` and `?LOG_WARNING` macro available, I'd think of using those instead of `error_logger` (which I thought was deprecated).
> 

I guess `logger` is supposed to be used. Fixed it.

> Also, it'd be great to have some guidelines on logging within OTP itself. I'd think of adding `otp` domain to all messages emitted within OTP itself (for easier filtering).
> 

I'm not the right person to answer this. @garazdawi can probably answer this when he is back (currently out of office).

> BTW is it necessary to log this event?

Yes I *really* think so. It is really annoying trying to find the reason to why a test case failed when there is no information about what happened. This information is also only logged when something has gone wrong and should normally never happen.

BTW regarding logging. I really miss the old style `test_server:start_node()` logs where you could see in the test log that the node had come up and how it was started with the exact command line arguments. Those logs have helped me numerous times in the past.

PR #5709, opened 2022-02-14
Update Unicode version to 14.0.0 in README-UPDATE.txt

maintenance-beam-app:
These changes were picked up by the Maintenance project:
-------

This is an automated commit created by the Maintenance project
https://github.com/eksperimental/maintenance

Before merging, please read the realease notes by visiting
<http://www.unicode.org/versions/Unicode14.0.0/>
and assess if additional changes are necessary in the code base.

### DIFF
--- a/lib/stdlib/uc_spec/README-UPDATE.txt
+++ b/lib/stdlib/uc_spec/README-UPDATE.txt
@@ -1,7 +1,7 @@
 Unicode 14.0.0 was updated from:
-- https://www.unicode.org/Public/13.0.0/ucd/
-- https://www.unicode.org/Public/13.0.0/ucd/auxiliary/
-- https://www.unicode.org/Public/13.0.0/ucd/emoji/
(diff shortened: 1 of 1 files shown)


dgud:
@eksperimental  you (and/or me) also did not update the version in the script file, 
at least is isn't there now, don't really.

eksperimental:
> @eksperimental you (and/or me) also did not update the version in the script file, at least it isn't there now.

Yes, there was a bug in the Maintenance app.  Thank you for spotting it. It has been fixed.
https://github.com/eksperimental/maintenance/commit/a4bd81da38ae62a6c41dccdb594c7db81b0e9267#diff-4a6021848be8c0057c770d1cdef7b6b8192d67eb0fbb62dc8b08d8958c382c29L306

I have updated the PR to include this change.

dgud:
Ok, great will merge later this week, thought that I made something stupid, and couldn't understand what.

PR #5712, opened 2022-02-14
make Erlang more tooling-friendly: allow custom attributes everywhere

### LINKED ISSUE #5689: Attributes mixed with functions
### Is your feature request related to a problem? Please describe.
I would like to implement AdRoll/rebar3_format#244 using attributes to surround code blocks that the formatter should ignore. To do that, I would like to be able to write code as…

```erlang
-module my_mod.

-export [my_fun/0].

my_fun() ->
    my_fun(something, something).

-format_block ignore.
my_fun(something, something) -> [
  this
  , function
    , should
      , not be:formatted()
        ].
-format_bock.
```

But the compiler now tells me that…

```erlang
my_mod.erl:#:#: attribute format_block after function definitions
```

### Describe the solution you'd like
I would like to have a compiler option available to disable that check.

### Describe alternatives you've considered
A brutal alternative would be to remove that check entirely, but I don't really know if there are scenarios where it's really needed.
Another one would be to allow _some_ (i.e. a list of) attributes to be interleaved, but that list should either be configurable or otherwise properly maintained.
Another alternative is to preserve the check but only for OTP attributes… user attributes can be interleaved at will. That would be my favorite choice, but again… I'm not really aware of the consequences.

ilya-klyuchnikov:
This makes Erlang surface syntax more friendly to tooling
like formatters, linters, analyzers.
Currently, the `-dialyzer(...)` attribute is the only annotating
(that is, not affecting semantics) attribute that can be placed
everywhere.

There is no reason to ban other custom attributes
from being placed between functions.

The logic that certain `erlc` attributes cannot be placed
after functions is preserved

This also closes #5689.

### DIFF
--- a/lib/stdlib/src/erl_lint.erl
+++ b/lib/stdlib/src/erl_lint.erl
@@ -900,10 +900,18 @@ function_state({attribute,A,opaque,{TypeName,TypeDef,Args}}, St) ->
     type_def(opaque, A, TypeName, TypeDef, Args, St);
 function_state({attribute,A,spec,{Fun,Types}}, St) ->
     spec_decl(A, Fun, Types, St);
-function_state({attribute,_A,dialyzer,_Val}, St) ->
-    St;
-function_state({attribute,Aa,Attr,_Val}, St) ->
+function_state({attribute,Aa,Attr,_Val}, St) when Attr =:= module;
+                                                  Attr =:= export;
+                                                  Attr =:= export_type;
+                                                  Attr =:= import;
+                                                  Attr =:= behaviour;
+                                                  Attr =:= behavior;
+                                                  Attr =:= callback;
+                                                  Attr =:= optional_callbacks;
+                                                  Attr =:= on_load ->
     add_error(Aa, {attribute,Attr}, St);
+function_state({attribute,_A,_Attr,_Val}, St) ->
+    St;
 function_state({function,Anno,N,A,Cs}, St) ->
     function(Anno, N, A, Cs, St);
 function_state({eof,Location}, St) -> eof(Location, St).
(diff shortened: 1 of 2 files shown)


ilya-klyuchnikov:
Hello @garazdawi 

Is there any feedback that I should address for this PR?

Thanks

KennethL:
I cannot promise that we have time to look at this and take a decision
before the release of OTP 25.0 in May.
But I think it is a good idea worth considering. Have to find out if there
are any possible drawbacks or incompatibilities which can hit the users or
other internal tools if this is introduced.

bjorng:
Oops! I intended to close this pull request, but pressed the wrong button and merged it instead. I have now reverted the merge.

elbrujohalcon:
Oh, man… I was *so happy* for a second! 😢

paulo-ferraz-oliveira:
@TD5, same (just a few days ago)... I wanted to use `-elvis` attributes to prevent warnings for specific functions and had to add them to the top of the module, far from the function definitions 😢 (while at the same time I can have `-dialyzer` directives added basically everywhere, especially next to the functions - so basically the same use case).

... so I have the same question as you 😄

---

I could even see, as I was used to doing when C#ing, editors being able to visually fold function groups with something like

```
-group("API functions for ...").
...
-endgroup.
```

GitHub does it for `details/summary` (Markdown), for example...

```
<details>
<summary>API functions for ...</summary>
Hurrah, you did it!
</details>
```

which is

<details>
<summary>API functions for ...</summary>
Hurrah, you did it!
</details>

duncanatt:
Is there an update regarding these custom attributes that can be used to annotate functions? Is this something you plan to support in future releases?

garazdawi:
Right now there are no plans to support it.

PR #5722, opened 2022-02-16
Update guidance on spec "when" for consistency

mheiber:
*For now*, standardize on the simple substitution
interpretation of `::` in constraints. Bounded
quantification could always be added later with
dedicated syntax (`<:` or `subtype_of`, etc.).

## Standardization

Previously, guidance for `when` in specs was that tools could *decide* whether to implement bounded quantificaiton or not.

That's a dangerous path in a world where we want users and OTP team to be able to write specs *once* and not split the ecosystem.

So if we're going for consistency, then we have a choice of whether to say `when` in specs is for bounded quantification or simple substitution.

## Two Interpretations: Subsitution vs. Bounds

- Substitution interpretation: `-spec id_tup(X) -> X when X :: tuple()`
is equivalent to `-spec id_tup(tuple()) -> tuple().`
- Bounded quantification interpretation:
`-spec id_tuple(X) -> X :: tuple()` is generic and bounds
`X` to be a `tuple()`. This is the kind of thing that might
be written in made-up explicit syntax as
`-spec id_tuple<T extends tuple()>(T) -> T`.
`id_tuple(an_atom)` would lead to a type error and
`id_tuple({a, b})` would have return type `{a, b}`.

There are both pragmatic and theoretical reasons in favor of
the substitution interpretation.

## Reasons in Favor of The Substitution Interpretation

1. The substitution interpretation **reflects Erlang reality**.
   - Dialyzer, Gradualizer, and mystery-type-checker all use
   the substitution interpretation.
   - Most OTP specs don't make much sense on the bounded quanitification
   interpretation. `:: term()` appears over 2000 times in this repo,
   but why would someone bound something to be a subtype of `term()`?
2. The substitution interpretation makes things easier for implementers,
including contributors to this repo. Dialyzer currently (iuc) doesn't
handle generics as generics at all. It sounds easier to add unbounded
genericity first rather than committing to the more complex bounded
quantification.
3. The substitution interpretation is more syntactically consistent. In `-type ty() :: atom()`, for example, the `::` means "is", not  "some type that is a subtype of atom()".
4. I sometimes see Erlang devs confused about the difference between
type variables and `term()` ([example in Q&A](https://youtu.be/_u1NDuFsW2A?t=1528)). Bounded quantification is yet another rung of complexity for users.
5. Bounded quantification is a complex feature that ideally would be
incorporated into the language in a holistic way:
    - For example, it would be an odd barrier to composability if
    specs could express bounds but type aliases and opaques could not.
    Such an asymmetry would prevent seemingly trivial refactors such
    as extracting out parts of a long spec into type aliases.
    - To take another example, if bounds really are part of types,
    then error reporting for static analysis ideally would include
    bounds as well. It takes some work to do this in a way that
    is both informative and not befuddling.
    - Variance interacts with bounded quantification,
    but the variance story is still being written.
    [@jhogberg is investigating variance annotations for aliases and opaques](https://github.com/erlang/otp/pull/5223#issuecomment-924917745).

## Note on Backwards-Compatibility

This is a minor change in official guidance.

This change in guidance won't break anything as far as I can tell,
so I don't think we tarnish Erlang's great reputation as a stable, backwards-compatible language.

And this change improves the situation going forward: standardizing will help preventing the type-writing community from splitting in the future.

> Thanks for your time in considering this docs update

### DIFF
--- a/system/doc/reference_manual/typespec.xml
+++ b/system/doc/reference_manual/typespec.xml
@@ -550,40 +550,22 @@
       For example, the following specification defines the type of a
       polymorphic identity function:
     </p>
-    <pre>
-  -spec id(X) -> X.</pre>
+    <pre> -spec id(X) -> X.</pre>
     <p>
-      Notice that the above specification does not restrict the input
-      and output type in any way.
-      These types can be constrained by guard-like subtype constraints
-      and provide bounded quantification:
+    Notice that the above specification restricts the output type
+    to be the same as the input type, but imposes no other requirements.
+    So `id(atom())` is expected to return `atom()` and `id({a, b})` is
+    expected to return `{a, b}`.
     </p>
-    <pre>  -spec id(X) -> X when X :: tuple().</pre>
-    <p>
-      Currently, the <c>::</c> constraint
-      (read as &laquo;is a subtype of&raquo;) is
-      the only guard constraint that can be used in the <c>when</c>
-      part of a <c>-spec</c> attribute.
+    <p>Constraint syntax is permitted in function specifications.
+    The meaning of such constraints is the same as substituting the
+    right-hand side of <c>::</c> for the left-hand side. The following
+    two definitions are equivalent
     </p>
-    <note>
-      <p>
-      The above function specification uses multiple occurrences of
-      the same type variable. That provides more type information than the
-      following function specification, where the type variables are missing:
-      </p>
-      <pre>  -spec id(tuple()) -> tuple().</pre>
-      <p>
-	The latter specification says that the function takes some tuple
-	and returns some tuple. The specification with the <c>X</c> type
-	variable specifies that the function takes a tuple and returns
-	<em>the same</em> tuple.
-      </p>
-      <p>
-	However, it is up to the tools that process the specifications
-        to choose whether to take this extra information into account
-        or not.
-      </p>
-    </note>
+    <pre>  -spec id_atom(X) -> X when X :: atom().
+           % equivalent to
+           -spec id_atom(atom()) -> atom().
+    </pre>
     <p>
       The scope of a <c>::</c> constraint is the
       <c>(...) -> RetType</c>


kostis:
The PR, at some point, has a title which reads:
> Two Interpretations: Subsitution vs. Bounds

Besides the typo, this PR gives (to me, at least) the impression that its writer thinks that these two interpretations are mutually exclusive.  They are not.   Substitution is just a special case of bounded quantification.

Once upon a time, `when` clauses had (exclusively) an `is_subtype/2` constraint, which is what this is all about.  Somebody thought it was _cute_ that `::` was used as a shorthand of `is_subtype` and went ahead and used it all over the place (often substituting the `is_subtype` constraints), probably without (fully) realizing the consequences of this change.

In any case, `::` should really be read _is subtype (subset) of _.  It's really bounded quantification.  But yes, in the majority of examples, substitution suffices; not in all cases though and (IMO) it is **wrong** if the docs give this impression to their reader(s).  Take `append` as an example.  It's perfectly OK -arguably it's the right thing to do- to spec it as:

```erlang
-spec append([X], [Y]) -> [Z] when X :: Z, Y :: Z.
```
and have an analysis that warns that the following code:

```erlang
foo(L1, L2) ->
    N = length(L1),
    R = [X || X <- L2, X < 0],
    case append([N], L2)  of
        R -> ...
    end. 
```

will always fail (based on just type information; not based on some more complex reasoning).


By the way, it is not true that a type declaration like
```erlang
-type ty() :: atom().
```
declares that `ty()` _is_ the set of all atoms.  It's perfectly alright to think of this declaration as "ty() is some subset (subtype) of the set of all atoms"  (and, as an Erlang programmer, I am too lazy to tell you _exactly_ which one(s)).

Botttomline: This PR should be dropped.

mheiber:
> Botttomline: This PR should be dropped.

That's a bit strong, isn't it?

I appreciate your reply and it does help clarify some of the history and intent.

mheiber:
I changed the list of reasons in favor of the substitution interpretation from a numbered list to a bulleted list, so we can point to particular ones.

Kostis engaged with (4), which is the weakest point I gave in favor of the substitution approach. I wrote:

> The substitution interpretation is more syntactically consistent. In `-type ty() :: atom()`, for example, the `::` means "is", not  "some type that is a subtype of atom()".

Kostis [replied](https://github.com/erlang/otp/pull/5722#issuecomment-1042387730):

>  It's perfectly alright to think of this declaration as "ty() is some subset (subtype) of the set of all atoms" (and, as an Erlang programmer, I am too lazy to tell you exactly which one(s)).

A type checker should probably reject this code:

-spec id(atom()) -> a.
id(X) -> X. % Error: expected a but got atom().

Wouldn't the subtype reading of `::` in `-type ty() :: atom()` imply that a type checker should reject the following code?

```
-type ty() :: atom().

-spec id(atom()) -> ty().
id(X) -> X.   % Error: expected ty(), but got atom()
```

kostis:
> A type checker should probably reject this code:
> 
> -spec id(atom()) -> a. id(X) -> X. % Error: expected a but got atom().
> 
> Wouldn't the subtype reading of `::` in `-type ty() :: atom()` imply that a type checker should reject the following code?
> 
> ```
> -type ty() :: atom().
> 
> -spec id(atom()) -> ty().
> id(X) -> X.   % Error: expected ty(), but got atom()
> ```

What does this PR have to do with what _some_ type checker would or would not do?  Of course, it can do what it pleases (whatever it thinks it is best for its preferred interpretation of types and specs).

This is _exactly_ what's being discussed here.  You seem to be arguing about _imposing_ some particular interpretation to what types and specs mean, while I am arguing for leaving all possible (sensible) interpretations open for tools to decide which one(s) to employ.

Hence my objection to this PR being merged.

mheiber:
point taken re not editing previous messages, sorry about that

mheiber:
>  You seem to be arguing about imposing some particular interpretation to what types and specs mean, while I am arguing for leaving all possible (sensible) interpretations open for tools to decide which one(s) to employ.

I agree this is the heart of the matter. I was assuming that we'd want everyone to agree on what specs mean, the same way we agree about what Erlang expressions mean. But you've obviously thought about this more than me and want to leave specs open to interpretation. How can I learn more about this? Could  we discuss 1:1?

jhogberg:
Thanks for the PR!

> Previously, guidance for when in specs was that tools could decide whether to implement bounded quantificaiton or not.

Tools can decide whether to ignore the connection between `X` in specs such as  `id(X) -> X when X :: tuple()`, but they must still interpret `::` as "subtype of," much like how the argument to `func(tuple()) -> ok` should be read as "subtype of `tuple()`".

> I agree this is the heart of the matter. I was assuming that we'd want everyone to agree on what specs mean, the same way we agree about what Erlang expressions mean.

We agree, it's important to have a well-defined type language. However -- while the documentation could be clearer -- we consider our type language to be well-defined already.

While we're open to standardizing on the substitution interpretation at some point in the future, we're hesitant to take steps in this direction without having decided that's where we want to go. We want to be reasonably sure that it will help tools find more problems than the current interpretation before we go ahead.

mheiber:
@jhogberg thanks for your thoughts on this issue. To check my understanding, is this an accurate summary?
1. The team wants to provide clear, consistent guidance on what specs and aliases mean.
2. The current guidance is intended to be that `when` clauses have the bounded quantification interpretation.
3. The guidance *may* change in future to the substitution interpretation.


If this is right, what's the best way to track (3)? Shall I leave this issue open or close it and reopen in 6 months so we can see where things are then?

jhogberg:
> If this is right, I'm interested in following (3). How about I close this issue and reopen in 6 months so we can see where things are then?

Yes, that's right. I'll try to bump it if we have an update too. :)

ilya-klyuchnikov:
> In any case, :: should really be read _is subtype (subset) of _.

@kostis - does it mean that dialyzer misbehaves wrt this example?

```erlang
-module(interface).

-callback run() -> {ok, atom()} | {error, term()}.
```

```erlang
-module(implementation).
-behaviour(interface).

%% API
-export([run/0]).

-spec run() -> Res when Res :: tuple().
run() ->
  {ok, ok}.
```

```
dialyzer interface.erl implementation.erl
  Proceeding with analysis...
implementation.erl:7:2: The return type 
          tuple() in the specification of run/0 is not a subtype of 
          {'error', _} | {'ok', atom()}, which is the expected return type for the callback of the interface behaviour
 done in 0m0.12s
```

Staying on a position that `Res when Res :: tuple()` means some `Res` that is a subtype of `tuple()`, - why does dialyzer even mention tuple() here?

jhogberg:
It complains because the specification for `run/0` is more allowing than that of the callback, which means it's either redundant or incorrect (in this case the former because `tuple()` is a strict superset of the actual return).

jhogberg:
`Res :: tuple()` means that `Res` has some specific value that can be found in the set of all tuples.

`Res :: {foobar, term()}` means that `Res` has some specific value that can be found in the set of all 2-tuples whose first element is `foobar`.

Taken together, the first subtype constraint (`-spec run ...`) is redundant because it completely covers the latter (`-callback run ...`).

PR #5723, opened 2022-02-16
gh-actions: Upload macOS builds to releases

### LINKED ISSUE #4487: ERL-1428: OTP binary releases for macOS
Original reporter: `wojtekmach`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1428

---

```
On the last EEF build & package WG meeting we briefly discussed automatically building macOS binary releases (and in the future possibly other ones too like for popular linux distros) and so I'd like to formally propose that as a new feature.

Here are example use cases for such feature:
- https://github.com/actions/setup-elixir/issues/23
- https://github.com/asdf-vm/asdf-erlang/issues/165

And here's a proof-of-concept: https://github.com/wojtekmach/otp_builds. I wrote some notes in the README. In particular please notice that the action https://github.com/wojtekmach/otp_builds/runs/1479456475?check_suite_focus=true automatically uploaded the tarball to the release: https://github.com/wojtekmach/otp_builds/releases/tag/OTP-23.1.4.

The OTP GitHub setup is quite different, but hopefully the prototype proves that adding such feature is feasible at a first glance. I hit an issue with statically linking openssl (https://bugs.erlang.org/browse/ERL-1408?focusedCommentId=20003&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20003) but hopefully that can be resolved soon.

Please let me know if that's something you'd be interested in and I'll be happy to work on incorporating this into the erlang/otp repo (any feedback how best to do it would be very appreciated too).
```

wojtekmach:
Closes #4487

- example job run: https://github.com/wojtekmach/otp/runs/5222519364?check_suite_focus=true
- example release: https://github.com/wojtekmach/otp/releases/tag/OTP-25.0-rc1

### DIFF
--- a/.github/scripts/create-artifacts.sh
+++ b/.github/scripts/create-artifacts.sh
@@ -10,3 +10,5 @@ mv otp otp_src_${VSN}
 tar -czf ${DIR}/otp_src_${VSN}.tar.gz otp_src_${VSN}
 mv otp_doc_man.tar.gz ${DIR}/otp_doc_man_${VSN}.tar.gz
 mv otp_doc_html.tar.gz ${DIR}/otp_doc_html_${VSN}.tar.gz
+mv otp_doc_html.tar.gz ${DIR}/otp_doc_html_${VSN}.tar.gz
+mv otp_macos_x86-64.tar.gz ${DIR}/otp_macos_x86-64_${VSN}.tar.gz
--- a/.github/scripts/create-artifacts.sh
+++ b/.github/scripts/create-artifacts.sh
@@ -10,5 +10,4 @@ mv otp otp_src_${VSN}
 tar -czf ${DIR}/otp_src_${VSN}.tar.gz otp_src_${VSN}
 mv otp_doc_man.tar.gz ${DIR}/otp_doc_man_${VSN}.tar.gz
 mv otp_doc_html.tar.gz ${DIR}/otp_doc_html_${VSN}.tar.gz
-mv otp_doc_html.tar.gz ${DIR}/otp_doc_html_${VSN}.tar.gz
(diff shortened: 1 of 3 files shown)


max-au:
@paulo-ferraz-oliveira EEF (Build&Packaging WG) is attempting a prototype of CI/CD based on GitLab (see ci.erlef.org and corresponding discussions on Slack channel). I think there is more sense to host and support all community releases (including MacOS, iOS, ...) using that capacity.

paulo-ferraz-oliveira:
@max-au, thanks. That might just be a solution. I guess you need to talk to @wojtekmach, then 😄 Are the solutions compatible? ie. is the effort to set up that CI already taking @wojtekmach's solution into account (or would otherwise easily integrate it?).

Having a central place where all is available makes for easier searching of artefacts (something erlef/setup-beam can benefit from).

garazdawi:
> would we be able to pull it off using only public resources, like GitHub, for example?

If all we want to do is to build binary packages for various OSs, then what github provides should be plenty. If we also want to test that the packages actually work, then we probably have to step outside what github provides and into what @max-au is working on.

The main problem IMO is finding someone (person or organization) that is willing to spend the time to work out all the problems that are sure to arise as Erlang/MacOS/openssl/wx versions change. I expect there to be a lot of issues popping up, but maybe I'm over pessimistic.

max-au:
@paulo-ferraz-oliveira Storage might be an issue if we need to produce too many versions of various packages. But in the end I'd like to have the CI/CD integrated into popular package management systems (e.g. Homebrew, Choco, ...)

PR #5726, opened 2022-02-17
Fix UDP multicast IPv6 support

binaryseed:
Support for UDP `multicast_if` appears to be broken with IPv6. This PR is an attempt to fix that.

Problem noted some time ago here: https://stackoverflow.com/questions/38376640/cant-add-multicast-group

### DIFF
--- a/lib/kernel/test/inet_SUITE.erl
+++ b/lib/kernel/test/inet_SUITE.erl
@@ -48,7 +48,7 @@
 	 getif/1,
 	 getif_ifr_name_overflow/1,getservbyname_overflow/1, getifaddrs/1,
 	 parse_strict_address/1, ipv4_mapped_ipv6_address/1, ntoa/1,
-         simple_netns/1, simple_netns_open/1,
+         simple_netns/1, simple_netns_open/1, udp_multicast_ipv6/1,
          add_del_host/1, add_del_host_v6/1,
          simple_bind_to_device/1, simple_bind_to_device_open/1,
 	 socknames_sctp/1, socknames_tcp/1, socknames_udp/1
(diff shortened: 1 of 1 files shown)


binaryseed:
I was hoping that the tests would run for the PR, but perhaps they aren't setup to do so since I'm based on top of `maint` and that was only enabled with OTP 25?

I added a commit that I think will fix the issue, but I'm happy to change whatever is needed / hand this PR over...

erts/preloaded/src/prim_inet.erl
```diff
@@ -1884,6 +1884,8 @@ type_value_2(uint8, X)  when X band 16#ff =:= X       -> true;
 type_value_2(time, infinity)                          -> true;
 type_value_2(time, X) when is_integer(X), X >= 0      -> true;
 type_value_2(ip,{A,B,C,D}) when ?ip(A,B,C,D)          -> true;
+type_value_2(ip,{A,B,C,D,E,F,G,H})
+  when ?ip6(A,B,C,D,E,F,G,H)                          -> true;
```
RaimoNiskanen, line 1888: `ip` today means IPv4 address.  It should not be extended to allow IPv6 address.
(Maybe it should be renamed to `ip4` then we could use `ip` for `ip4` | `ip6`)

in `enc_value_3` a type `ip` value is encoded from `{A,B,C,D}` to a IPv4 binary address.  Your change allows to pass an IPv6 address to this point that fails this encoding.

There is also no corresponding code in `inet_drv.c` that handles an IPv6 address, so that would probably segfault if you send down an encoded IPv6 address.  This missing corresponding code in `inet_drv.c` should look at the socket type to figure which address to expect, whether to use IPV6_ADD_MEMBERSHIP or IP_ADD_MEMBERSHIP, and recognize when it gets an address on the wrong format.

RaimoNiskanen:
This problem stems from the old before-IPv6-design.  There is almost no options that are allowed for both IPv4 and IPv6 sockets, and these ones need to handle that.

We need a much more thorough change for this problem.

RaimoNiskanen:
Sorry, not thorough enough...

binaryseed:
Thanks, I didn't realize it went that deep... I will submit an Issue instead

binaryseed:
Opened https://github.com/erlang/otp/issues/5789

PR #5729, opened 2022-02-18
Generate class BeamGlobalAssembler with m4

sverker:
instead of C preprocessor macros, to make it easier to browse the code.

The second commit is an example of a generated `beam_asm_global.hpp` for illustration. It should not be merged.

However, we may not want this PR as it reintroduces build dependency to m4, which was removed together with HiPE.


sverker:
Superseded by #5873.

PR #5730, opened 2022-02-19
Net tick intensity

rickard-green:
Introduce a `net_tickintensity` kernel parameter with which one can control the amount of ticks during `net_ticktime`.

Also improve the `net_kernel:start/1` options making it possible to easier to add new options.


rickard-green:
Replaced by #5740

PR #5737, opened 2022-02-21
jit: Refactor ArgVal to catch errors earlier

jhogberg:
By breaking `ArgVal` into subtypes and only letting us inspect the value when we have a concrete type, we reduce the risk of errors like #5727 happening again.


erts/emulator/utils/beam_makeops
```diff
@@ -3366,3 +3454,25 @@ sub include_files() {
     print '#include "beam_load.h"', "\n";
     print "\n";
 }
+
+# Converts a sign string to its most specific argument class, e.g "an" becomes
```
bjorng, line 3458:
```suggestion
# Converts a sign string to its most specific argument class, e.g. "an" becomes
```

erts/emulator/beam/sys.h
```diff
@@ -176,6 +176,28 @@ typedef ERTS_SYS_FD_TYPE ErtsSysFdType;
 #  define ERTS_UNLIKELY(BOOL) (BOOL)
 #endif
 
+#if defined(DEBUG)
```
bjorng, line 179: I suggest adding a comment explaining what `ERTS_ASSUME()` and `ERTS_PURE` do.

bjorng (approved):

PR #5740, opened 2022-02-21
Net tick intensity

rickard-green:
Introduce a [`net_tickintensity`](https://erlang.github.io/prs/5740/lib/kernel-8.2/doc/html/kernel_app.html#net_tickintensity) kernel parameter with which one can control the amount of ticks during [`net_ticktime`](https://erlang.github.io/prs/5740/lib/kernel-8.2/doc/html/kernel_app.html#net_ticktime).

Also introduce the [`net_kernel:start/2`](https://erlang.github.io/prs/5740/lib/kernel-8.2/doc/html/net_kernel.html#start-2) function making it possible to easier to add new options.

### DIFF
--- a/lib/kernel/doc/src/net_kernel.xml
+++ b/lib/kernel/doc/src/net_kernel.xml
@@ -40,10 +40,11 @@
       <c>-name</c> or <c>-sname</c>:</p>
      <pre>
 $ <input>erl -sname foobar</input></pre>
-    <p>It is also possible to call <c>net_kernel:start([foobar])</c>
+    <p>It is also possible to call
+      <seemfa marker="#start/2"><c>net_kernel:start(foobar, #{})</c></seemfa>
       directly from the normal Erlang shell prompt:</p>
     <pre>
-1> <input>net_kernel:start([foobar, shortnames]).</input>
+1> <input>net_kernel:start(foobar, #{name_domain => shortnames}).</input>
 {ok,&lt;0.64.0>}
 (foobar@gringotts)2></pre>
     <p>If the node is started with command-line flag <c>-sname</c>,
@@ -113,8 +114,11 @@ $ <input>erl -sname foobar</input></pre>
       <name name="get_net_ticktime" arity="0" since=""/>
       <fsummary>Get <c>net_ticktime</c>.</fsummary>
       <desc>
-        <p>Gets <c>net_ticktime</c> (see
-          <seeapp marker="kernel_app"><c>kernel(6)</c></seeapp>).</p>
+        <p>Returns currently used net tick time in seconds. For more information
+        see the
+        <seeapp marker="kernel_app#net_ticktime"><c>net_ticktime</c></seeapp>
+        <c>kernel(6)</c> parameter.</p>
+
         <p>Defined return values (<c><anno>Res</anno></c>):</p>
         <taglist>
           <tag><c><anno>NetTicktime</anno></c></tag>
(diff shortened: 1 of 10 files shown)


lib/kernel/src/dist_util.erl
```diff
@@ -464,6 +464,17 @@ flags_to_version(Flags) ->
 %% The connection has been established.
 %% --------------------------------------------------------------
 
+-record(state, {kernel,
```
max-au, line 467: While at it, would you mind adding types to this structure?

lib/kernel/src/net_kernel.erl
```diff
@@ -151,14 +151,18 @@
...
--record(tick, {ticker,        %% ticker                     : pid()
-	       time           %% Ticktime in milli seconds  : integer()
-	      }).
-
--record(tick_change, {ticker, %% Ticker                     : pid()
-		      time,   %% Ticktime in milli seconds  : integer()
-		      how     %% What type of change        : atom()
-		     }).
+-record(tick,
+        {ticker,        %% ticker                      : pid()
```
max-au, line 155: Since you know the types, maybe put them as actual types instead of comments?

max-au (commented):
Looks interesting!
Sharing a couple of related effects we had in the past.
1. Single ticker process tend to cause large memory spikes when there are thousands of nodes connected via Erlang distribution. It happens because all ticks are sent at the same time (waking all thousands of otherwise sleeping processes). Obvious solution for us was to spread the ticks over NET_TICKTIME.
2. We are looking to add some useful signal to ticks (namely, scheduler/CPU/RAM utilisation), that allows node on the other end of the pipe to make a more informed decision on where to send requests. This, however, is better be done by a separate singleton process, hence I'd be interested in making net_kernel ticks optional.

rickard-green:
> Looks interesting! Sharing a couple of related effects we had in the past.
> 
>     1. Single ticker process tend to cause large memory spikes when there are thousands of nodes connected via Erlang distribution. It happens because all ticks are sent at the same time (waking all thousands of otherwise sleeping processes). Obvious solution for us was to spread the ticks over NET_TICKTIME.
> 

I thought about setting a timer in each ticker process and removing the tick orchestration process (which also sends the orchestration ticks via the `net_kernel` process), but it would have made this change even larger (it already got quite a bit larger than I initially expected) and it is also probably not as efficient CPU wise since there will be more reading of monotonic time than with current solution.

As of OTP 25 there will be a [`net_ticker_spawn_options`](https://erlang.org/documentation/doc-13.0-rc1/lib/kernel-8.3/doc/html/kernel_app.html#net_ticker_spawn_options) kernel parameter available (#5069) which was introduced in order to alter GC options of tickers which perhaps could be useful.

>     2. We are looking to add some useful signal to ticks (namely, scheduler/CPU/RAM utilisation), that allows node on the other end of the pipe to make a more informed decision on where to send requests. This, however, is better be done by a separate singleton process, hence I'd be interested in making net_kernel ticks optional.

We've had on the todo-list for quite some time to make it possible for the distribution module to provide its own alternate supervision of the channel, but it has never gotten enough priority to be implemented.

lib/kernel/src/net_kernel.erl
```diff
@@ -362,21 +366,112 @@ retry_request_maybe(Req) ->
 %% This function is used to dynamically start the
 %% distribution.
 
+-spec start(Args) -> {ok, pid()} | {error, Reason} when
+      Args :: nonempty_list(Arg),
+      Arg :: {name, Name}
+           | {name_domain, NameDomain}
+           | {net_ticktime, NetTickTime}
+           | {net_tickintensity, NetTickIntensity},
```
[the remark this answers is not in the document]
rickard-green: I implemented a version with options in a map instead of the 2-tuple list and I liked it, so I think I lean towards using option maps instead of option lists now.

lib/kernel/doc/src/net_kernel.xml
```diff
@@ -345,28 +349,114 @@ $ <input>erl -sname foobar</input></pre>
...
+            Dynamic Node Name</seeguide>. The only characters allowed in <c>Name</c>
+            are:
+        </p>
+        <list>
+          <item><c>a-z</c></item>
+          <item><c>A-Z</c></item>
+          <item><c>0-9</c></item>
+          <item><c>-</c></item>
+          <item><c>_</c></item>
+        </list>
```
KennethL, line 375: I propose the following alternative text. 
The allowed characters in <c>Name</c> are <c>[a-z] | [A-Z] | [0-9] '-' | '_'</c>.
Which when rendered could look something like: 
The allowed characters in `Name` are `[a-z] | [A-Z] | [0-9] '-' | '_'`.

PR #5741, opened 2022-02-21
Define list() + list(ContentType) using built-ins

mheiber:
Define list() + list(ContentType) using built-ins

Change the implementation of the `list/0` and `list/1` types exposed in the `erlang`
module to be as documented in https://www.erlang.org/doc/reference_manual/typespec.html
- My motivation for this change is to make it easier
to implement a type checker for Erlang specs,
since then we can remove some hard-coding for
the meaning of `list()` and rely on `erlang.erl`
as the source of truth.


bjorng:
I suggest that you amend your commit to also do the suggested change of `list(ContentType)`.

When you have done that, we will test the PR in our daily builds to ensure that it does not break anything.

PR #5751, opened 2022-02-25
gen_server types (OTP-17915)

RaimoNiskanen:
This pull request exports `gen_server` types as suggested in PR #2375 and PR #2690, fixes a few type violations now found in Kernel, updates the documentation to utilize the type specs, and hopefully gets all type details right, at least after a few rounds of review...

Is this the right set of types to export from `gen_server`?
Are the right types opaque in the right way?
Is the documentation update an improvement?
Are there any obvious things to fix in the documentation while at it?


bjorng:
This pull request causes the `behaviour_SUITE:gen_server_incorrect_args/1` and `small_SUITE:gencall/1` test cases for the `dialyzer` application to fail.

RaimoNiskanen:
To facilitate documentation review, here are some links of what I feel would be worth looking at:
* [Data Types](https://erlang.github.io/prs/5751/lib/stdlib-3.17/doc/html/gen_server.html#data-types)
* [`start_link/3,4`](https://erlang.github.io/prs/5751/lib/stdlib-3.17/doc/html/gen_server.html#start_link-3) that refers to type `start_ret()` for its return value instead of describing them, as does all `start*/*` functions.
* A perhaps new way of describing the return values for [`Module:handle_call/3`](https://erlang.github.io/prs/5751/lib/stdlib-3.17/doc/html/gen_server.html#Module:handle_call-3)
* [Module:init/1](https://erlang.github.io/prs/5751/lib/stdlib-3.17/doc/html/gen_server.html#Module:init-1) that refers to `Module:handle_call/3` for describing the `Timeout` values.

And, of course, If anyone would proofread the whole new [`gen_server`](https://erlang.github.io/prs/5751/lib/stdlib-3.17/doc/html/gen_server.html) documentation that would be appreciated...

IngelaAndin:
In the documentation of the function gen_server:call  it says:

` If no reply is received within the specified time, the function call fails. If the caller catches the failure and continues running, and the server is just late with the reply, it can arrive at any time later into the message queue of the caller. `

I never liked this wording. I think  "the function call fails" is too vague and it is not obvious what is meant by that.  I think it would be clearer to say the function calls fails by exiting the process that is calling the function, and if the process catches this exit and continues to run it must be prepared to receive a late answer, as the server is not affected by the client timeout.  This should be done by disregarding any such late messages that are two-element tuples with a reference as the first element.

Also, I think it would be an improvement if we could make an exit type for the most common exit reasons instead of just saying

 "The call can fail for many reasons, including time-out and the called gen_server process dying before or during the call."

IngelaAndin:
Also, enter_loop could specify its failing by exiting better.

RaimoNiskanen:
> Also, enter_loop could specify its failing by exiting better.

That is discussed in PR #5683.  I think we should handle it there.

RaimoNiskanen:
> Also, I think it would be an improvement if we could make an exit type for the most common exit reasons instead of just saying

Does it really merit a type of its own?
But I agree we could take the opportunity to document it.

IngelaAndin:
> > Also, enter_loop could specify its failing by exiting better.
> 
> That is discussed in PR #5683. I think we should handle it there.

Sure, you will be the driver either way ;)

IngelaAndin:
> > Also, I think it would be an improvement if we could make an exit type for the most common exit reasons instead of just saying
> 
> Does it really merit a type of its own? But I agree we could take the opportunity to document it.

Humm ...  maybe, maybe not (it is not a return-type after all). Make a suggestion and we can take it from there.

dgud (approved):

IngelaAndin (approved):

PR #5753, opened 2022-02-25
Add 'Erlang OTP Version' to the Erlang shell title row

henrythebuilder:
Expose the _Erlang full version_ string stored in the `ERLANG_OTP_VERSION` constant (`erl_version.h` header) into the Erlang shell title row.

In this way the full '_Erlang OTP Version_' would be also available from the 'code' by `erlang:system_info(system_version)`:

```
> erl
Erlang/OTP 25 [RELEASE CANDIDATE 1] [erts-12.2.1] [source-88e3472555] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns] [otp-version:25.0-rc1]

Eshell V12.2.1  (abort with ^G)
1> erlang:system_info(system_version).
"Erlang/OTP 25 [RELEASE CANDIDATE 1] [erts-12.2.1] [source-88e3472555] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns] [otp-version:25.0-rc1]\n"
2> 

```

Note: this value is already exposed in the 'Windows' version of the 'Erlang shell' through the 'About' function.

### DIFF
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -150,6 +150,9 @@ static char erts_system_version[] = ("Erlang/OTP " ERLANG_OTP_RELEASE
 #ifdef SHCOPY
 				     " [sharing-preserving]"
 #endif
+#ifdef ERLANG_OTP_VERSION
+             " [otp-version:" ERLANG_OTP_VERSION "]"
+#endif
 				     "\n");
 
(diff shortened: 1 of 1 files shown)


rickard-green:
We do not want to do this since an OTP installation may be a combination of multiple OTP versions. In such a case, faulty information would be presented.

henrythebuilder:
Thanks anyway.

PR #5754, opened 2022-02-28
Reorganize the exports of the lists module

bjorng:
Make it easer to get an overview over the function of the `lists`
module by grouping them into (I hope) logical groups.

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -22,28 +22,56 @@
 -compile({no_auto_import,[max/2]}).
 -compile({no_auto_import,[min/2]}).
 
--export([append/2, append/1, subtract/2, reverse/1,
-	 nth/2, nthtail/2, prefix/2, suffix/2, droplast/1, last/1,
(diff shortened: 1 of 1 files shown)


lib/stdlib/src/lists.erl
```diff
@@ -22,29 +22,57 @@
...
+         enumerate/1, enumerate/2,
+         flatlength/1, flatten/1, flatten/2,
+         join/2, last/1, min/1, max/1,
+         nth/2, nthtail/2,
+         prefix/2, reverse/1, seq/2, seq/3,
+         split/2, sublist/2, sublist/3,
+         subtract/2, suffix/2, sum/1,
+         unzip/1, unzip3/1,
+         zip/2, zip3/3,
+         zipwith/3, zipwith3/4]).
```
KennethL, line 41: zipwith/3 and zipwith3/4 take a fun as argument so they should either be in its own group or the comment about don't take funs should be changed.

KennethL (changes requested):

lib/stdlib/src/lists.erl
bjorng: Thanks! I have moved them into the groups of functions that take funs.

PR #5759, opened 2022-03-01
time_SUITE: skip tz-specific tests when running out of Stockholm

max-au:
These tests always fail when running in any timezone other
than Stockholm.


bjorng:
Thanks for pull request! Added to our daily builds.

max-au:
I think this PR does not fix `y2k_SUITE` issue, as it does not touch it at all.

bjorng:
Yes, you are right. I have unlinked the PR.

PR #5761, opened 2022-03-02
Remove case that filters all logs with defined domains

filmor:
The code as it was before resulted in `cth_log_redirect` hiding all log messages from the ct log that included a (custom) `domain`.

### DIFF
--- a/lib/common_test/src/cth_log_redirect.erl
+++ b/lib/common_test/src/cth_log_redirect.erl
@@ -171,8 +171,6 @@ log(#{msg:={report,Msg},meta:=#{domain:=[otp,sasl]}}=Log,Config) ->
     end;
 log(#{meta:=#{domain:=[otp]}}=Log,Config) ->
     do_log(add_log_category(Log,error_logger),Config);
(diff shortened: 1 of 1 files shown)


IngelaAndin:
I think that logs with their own defined domain usually have their own handlers, and I am not sure if this makes sense or not.
Our logger expert will be back from leave fairly soon and so I think we can wait for him to help resolve this.

garazdawi:
According to the [Common Test docs](https://www.erlang.org/doc/apps/common_test/ct_hooks_chapter.html#built-in-cths) this hook should:

> Captures all log events that would normally be printed by the default logger handler, and prints them to the current test case log.

The default behaviour of `logger` is to ignore all log messages from unknown domains, so this is what the cth_hook also does.

I think that the best way to handle this problem is to make the hook copy the filter configuration from the default handler. That way when if you change the default handler filter configuration, you also change the config for the hook.

Another option would also be to allow the hook to be configured with the desired filter config.

garazdawi:
I'm closing this PR due to inactivity. Feel free to reopen if you want to continue working on this.

PR #5762, opened 2022-03-02
epmd: don't fail if non-critical interfaces fail to bind

jeckersb:
When specifying bind address(es) via ERL_EPMD_ADDRESS or the -address
option, epmd will bind to the IPv4/IPv6 loopback intefaces in addition
to the user-supplied address(es).

However, it is possible that one of the loopback interfaces does not
exist.  For example, if the host system has ipv6 disabled via the
disable_ipv6 sysctl, there will be no loopback interface "::1", and
attmepting to bind to "::1" will fail.

Under these circumstances, log the failure but do not consider the
failure to bind to the loopback interface as fatal.

If any of the user-supplied addresses fail to bind, however, that is
still considered fatal and epmd will terminate.

### DIFF
--- a/erts/epmd/src/epmd_srv.c
+++ b/erts/epmd/src/epmd_srv.c
@@ -245,6 +245,7 @@ void run(EpmdVars *g)
   char socknamebuf[INET_ADDRSTRLEN];
 #endif
   int num_sockets = 0;
+  int nonfatal_sockets = 0;
   int i;
   int opt;
   unsigned short sport = g->port;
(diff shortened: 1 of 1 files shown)


sverker:
Question: If neither ERL_EPMD_ADDRESS nor -address is used, epmd bind sockets (v4 and v6) using address "any". Those are not included in your nonfatal_sockets. Is that by desing?

jeckersb:
> Question: If neither ERL_EPMD_ADDRESS nor -address is used, epmd bind sockets (v4 and v6) using address "any". Those are not included in your nonfatal_sockets. Is that by desing?

Yes, because the "any" address will still bind even if the protocol stack is disabled.  Although, I am making an assumption due to the way it works on Linux; I'm not 100% positive that it behaves the same way across other platforms.

```
# echo 1 > /proc/sys/net/ipv6/conf/all/disable_ipv6
# strace -e trace=bind epmd
bind(3, {sa_family=AF_INET, sin_port=htons(4369), sin_addr=inet_addr("0.0.0.0")}, 16) = 0
bind(4, {sa_family=AF_INET6, sin6_port=htons(4369), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::", &sin6_addr), sin6_scope_id=0}, 28) = 0
```

sverker:
@big-r81 
> Hi, will this fix be backported to older Erlang version like v23?

No such plan right now as the PR commit was based on master.

If you really want it: rebase e8c9e12 on OTP-23.x.y (not 23.x.y.z), create a new PR on maint-23 and make the case why this semantic change will not cause problems for existing users of OTP 23.

PR #5768, opened 2022-03-05
[erts] Expose parent process via process_info(_, parent) 

max-au:
Parent process information is available in the VM process structure,
but is not exposed via process_info. Sometimes it is very convenient
to know who spawned processes that are not supporting any OTP
behaviours (where this information is stored in a process dictionary).

In addition to counting reductions, it may also be important to keep
track of total number sent to a process. This allows various monitoring
scenarios. For example, when message_queue_len is draining slowly, is
it because it the process being slow, or is it because it gets more
messages enqueued. Having both message_queue_len and messages_enqueued
provides a way to calculate queue fill rate (or drain rate) over time.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -6672,9 +6672,9 @@ receive_replies(ReqId, N, Acc) ->
             <item>
                 <p><c><anno>Pid</anno></c> is the identifier
                 of the parent process, the one that spawned current
-                process. When the process does not have a local parent
-                (e.g. <c>init</c>, or process is spawned remotely),
-                <c>undefined</c> is returned.</p>
+                process. When the process does not have a parent
+                <c>undefined</c> is returned. Only the initial process
+                (<c>init</c>) on a node lacks a parent, though.</p>
             </item>
           <tag><c>{priority, <anno>Level</anno>}</c></tag>
           <item>
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -6668,6 +6668,14 @@ receive_replies(ReqId, N, Acc) ->
               <seeerl marker="#process_flag_message_queue_data">
               <c>process_flag(message_queue_data, MQD)</c></seeerl>.</p>
           </item>
+            <tag><c>{parent, <anno>Pid</anno>}</c></tag>
+            <item>
+                <p><c><anno>Pid</anno></c> is the identifier
+                of the parent process, the one that spawned current
+                process. When the process does not have a local parent
+                (e.g. <c>init</c>, or process is spawned remotely),
+                <c>undefined</c> is returned.</p>
+            </item>
           <tag><c>{priority, <anno>Level</anno>}</c></tag>
           <item>
             <p><c><anno>Level</anno></c> is the current priority level for
--- a/erts/emulator/beam/atom.names
+++ b/erts/emulator/beam/atom.names
@@ -530,6 +530,7 @@ atom owner
 atom packet
 atom packet_size
 atom parallelism
+atom parent
 atom Plus='+'
 atom PlusPlus='++'
 atom pause
--- a/erts/emulator/beam/break.c
+++ b/erts/emulator/beam/break.c
@@ -307,7 +307,8 @@ print_process_info(fmtfn_t to, void *to_arg, Process *p, ErtsProcLocks orig_lock
 		   p->current->arity);
     }
 
-    erts_print(to, to_arg, "Spawned by: %T\n", p->parent);
+    erts_print(to, to_arg, "Spawned by: %T\n",
+               p->parent == am_undefined ? NIL : p->parent);
 
     if (locks & ERTS_PROC_LOCK_MAIN) {
         erts_proc_lock(p, ERTS_PROC_LOCK_MSGQ);
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -769,6 +769,7 @@ collect_one_suspend_monitor(ErtsMonitor *mon, void *vsmicp, Sint reds)
 #define ERTS_PI_IX_GARBAGE_COLLECTION_INFO              33
 #define ERTS_PI_IX_MAGIC_REF                            34
 #define ERTS_PI_IX_FULLSWEEP_AFTER                      35
+#define ERTS_PI_IX_PARENT                               36
 
 #define ERTS_PI_FLAG_SINGELTON                          (1 << 0)
 #define ERTS_PI_FLAG_ALWAYS_WRAP                        (1 << 1)
@@ -824,7 +825,8 @@ static ErtsProcessInfoArgs pi_args[] = {
     {am_message_queue_data, 0, 0, ERTS_PROC_LOCK_MAIN},
     {am_garbage_collection_info, ERTS_PROCESS_GC_INFO_MAX_SIZE, 0, ERTS_PROC_LOCK_MAIN},
     {am_magic_ref, 0, ERTS_PI_FLAG_FORCE_SIG_SEND, ERTS_PROC_LOCK_MAIN},
-    {am_fullsweep_after, 0, 0, ERTS_PROC_LOCK_MAIN}
+    {am_fullsweep_after, 0, 0, ERTS_PROC_LOCK_MAIN},
+    {am_parent, 0, 0, ERTS_PROC_LOCK_MAIN}
 };
 
 #define ERTS_PI_ARGS ((int) (sizeof(pi_args)/sizeof(pi_args[0])))
@@ -943,6 +945,8 @@ pi_arg2ix(Eterm arg)
         return ERTS_PI_IX_MAGIC_REF;
     case am_fullsweep_after:
         return ERTS_PI_IX_FULLSWEEP_AFTER;
+    case am_parent:
+        return ERTS_PI_IX_PARENT;
     default:
         return -1;
     }
@@ -2024,6 +2028,10 @@ process_info_aux(Process *c_p,
 	    break;
 	}
 	break;
+
+    case ERTS_PI_IX_PARENT:
+        res = rp->parent == NIL ? am_undefined : rp->parent;
+        break;
 
     case ERTS_PI_IX_MAGIC_REF: {
 	Uint sz = 0;
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -2030,7 +2030,17 @@ process_info_aux(Process *c_p,
 	break;
 
     case ERTS_PI_IX_PARENT:
-        res = rp->parent == NIL ? am_undefined : rp->parent;
+        if (is_immed(rp->parent)) {
+            ASSERT(is_internal_pid(rp->parent) || rp->parent == am_undefined);
+            res = rp->parent;
+        }
+        else {
+            Uint sz;
+            ASSERT(is_external_pid(rp->parent));
+            sz = size_object(rp->parent);
+            hp = erts_produce_heap(hfact, sz, reserve_size);
+            res = copy_struct(rp->parent, sz, &hp, hfact->off_heap);
+        }
         break;
 
     case ERTS_PI_IX_MAGIC_REF: {
--- a/erts/emulator/beam/erl_gc.c
+++ b/erts/emulator/beam/erl_gc.c
@@ -2603,6 +2603,14 @@ setup_rootset(Process *p, Eterm *objv, int nobj, Rootset *rootset)
 	n++;
     }
 
+    ASSERT(p->parent == am_undefined
+           || is_pid(follow_moved(p->parent, (Eterm) 0)));
+    if (is_not_immed(p->parent)) {
+	roots[n].v  = &p->parent;
+	roots[n].sz = 1;
+	n++;
+    }
+
     /*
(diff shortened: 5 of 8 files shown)


erts/doc/src/erlang.xml
```diff
@@ -6625,6 +6625,11 @@ receive_replies(ReqId, N, Acc) ->
             <p><c><anno>MessageQueue</anno></c> is a list of the messages to
               the process, which have not yet been processed.</p>
           </item>
+          <tag><c>{messages, <anno>MessagesEnqueued</anno>}</c></tag>
```
potatosalad, line 6628: Typo here, I think: `messages -> messages_enqueued`

potatosalad:
@max-au [`erl_process.c` currently supports having `parent` set to `NIL`](https://github.com/erlang/otp/blob/master/erts/emulator/beam/erl_process.c#L12301-L12303), although right now it's admittedly abnormal for the majority of processes (maybe just the `init` process today?).

However, this "orphaned" process creation is used as part of [RFC: Erlang Dist Filtering](https://github.com/potatosalad/otp/pull/1) for the Erlang processes spawned to accept or reject the incoming dist command. There may be a different way to accomplish this, though, but I thought it worth mentioning that `parent` may not always be a `pid()`.

max-au:
Fixed a bunch of typos and updated documentation.

I see two ways to solve a problem with `init` (or other parent-less processes):
* return Pid of the process itself (as in, `init` is self-spawned). This is what Unix does and I like this more, so I updated the patch to do exactly this
* return `undefined` atom

rickard-green:
@max-au sorry for later answer. It has been really busy lately...

Regarding `process_info(Pid, messages_enqueued)`:
* I don't like to introduce things that breaks the signal order. One could perhaps get around that by documenting the value as being an approximate value. That way it can be considered as not breaking the signal order.
* I suspect that a lot of systems will never use it. All processes in all systems will have to carry the cost of the extra word in the process structure, though. It is very easy to add just another word, since it is just a word, but in the end up with a large amount of normally unused words in all processes structures, so we need to be a bit restrictive when it comes to adding stuff in the process structure. We have process specific data (search for PSD in `erl_process.h`) which is intended to be used for rarely used features. When used, an array, currently 10 words large, is allocated to hold such rarely used values. Once allocated for a process, it will stay there until the process terminates. If you should add this word to PSD and check for `message_enqueue` on all processes, all processes in your system would allocate 11 more words, though. Here it would have been nice with something in between PSD and putting it into the process structure...
* I also suspect that one actually would want a completely different approach than polling processes for information. If I understand correct, you poll processes repeatedly to fetch this information and then make decisions based on that. Frequently fetching information from processes this way is quite a heavy task. It is also just a snapshot, so you might miss peaks, for example in message queue length (although in this case the value is monotonically increasing). I suspect that functionality like this would fit much better in [erlang:system_monitor()](https://www.erlang.org/doc/man/erlang.html#system_monitor-2). It won't fit "as is", but, for example, by making it possible to set a limit in system monitor for fill rate and/or drain rate. I guess this would be much more efficient when looking at the total amount of work performed for the whole operation.

I don't see any major issues with `process_info(Pid, parent)`. One thing though is that processes that has a remote process as parent currently does not have it saved in the `parent` field. This would add a couple of words to the heap of that process by saving the remote pid, but the majority of processes on a system wont be remotely spawned, and I think it could be useful information. A bit funny that I both argue for not using more memory as well as using more memory in the same comment :-)

okeuday:
@max-au If a remote parent is not set, it seems best to have the parent in that case be `undefined` instead of self()

rickard-green:
> @max-au If a remote parent is not set, it seems best to have the parent in that case be `undefined` instead of self()

Yes, I agree

max-au:
I updated this PR:
 - removed `enqueued_messages` in favour of a separate PR for a more focused discussion  
 - rebased to master branch
 - changed to return `undefined` instead of `self()` for processes that have no local parent.

I explained my thought process for `self()` in the comment above (this is what Unix does). It also keeps the return spec simpler (there is always a `pid()` returned). But as I outlined above, I'm not against returning `undefined` either.

rickard-green:
> I explained my thought process for self() in the comment above (this is what Unix does). It also keeps the return spec simpler (there is always a pid() returned). But as I outlined above, I'm not against returning undefined either.

Yes, I think `self()` approach is nice, but I feel the `undefined` approach is more consistent with how we usually do.

I've pushed a commit which adds support for remote parents.

We are really soon about to release 25-rc2. This wont make it into rc2, but I'll put this into testing as soon as rc2 is out.

max-au:
Thanks!
I don't have a good benchmark for remote spawn/erpc, but just looking at code it should not cause significant regression.

max-au:
> I suspect that functionality like this would fit much better in [erlang:system_monitor()](https://www.erlang.org/doc/man/erlang.html#system_monitor-2).

I happened to have this PR in works as well. It introduces `long_queue` monitoring event (similar to `large_heap`, but triggered by total queue size). However the actual functionality we need (and internally still use) is this: https://github.com/max-au/otp/commit/a3d7f6026768650937cc99203963423d1148bbb9

Every N seconds we iterate over processes to find the queue fill/drain rate. This way we find processes that are building up queues over time. It can also be implemented as part of `system_monitor` flag (something like `{queue_fill_rate, 200, 1000}` meaning "send a monitoring message if process message queue grows by 200 messages in 1000 milliseconds"), but it would be even more expensive for all ERTS users. We also monitor process message queues (enqueued/dequeued messages) to see irregularities in system behaviour. It may be a good solution to have this optional (controlled by process flag), but that is also extra complexity in ERTS I was hoping to avoid. I should probably make a post on Erlang forums for a wider discussion, but I'd first appreciate if @rickard-green already has something to suggest.

rickard-green:
> > I suspect that functionality like this would fit much better in [erlang:system_monitor()](https://www.erlang.org/doc/man/erlang.html#system_monitor-2).
> 
> I happened to have this PR in works as well. It introduces `long_queue` monitoring event (similar to `large_heap`, but triggered by total queue size). However the actual functionality we need (and internally still use) is this: [max-au@a3d7f60](https://github.com/max-au/otp/commit/a3d7f6026768650937cc99203963423d1148bbb9)
> 

I guess you by "total queue size" mean the value in `sig_qs.len` field?

It is okay to base such functionality on this value as long as it does not claim it is measuring the message queue length. Calling it signal queue length is okay, though.

> Every N seconds we iterate over processes to find the queue fill/drain rate. This way we find processes that are building up queues over time. It can also be implemented as part of `system_monitor` flag (something like `{queue_fill_rate, 200, 1000}` meaning "send a monitoring message if process message queue grows by 200 messages in 1000 milliseconds"), but it would be even more expensive for all ERTS users.

What do you mean would be more expensive for all users? Repeated `process_info()` calls vs system monitor, or signal queue length vs queue fill rate?

> We also monitor process message queues (enqueued/dequeued messages) to see irregularities in system behaviour. It may be a good solution to have this optional (controlled by process flag), but that is also extra complexity in ERTS I was hoping to avoid. I should probably make a post on Erlang forums for a wider discussion, but I'd first appreciate if @rickard-green already has something to suggest.

I don't see it as acceptable that users not interested in this should pay an, other than negligible, performance/memory loss for this. If extra complexity is needed in order to achieve this, I see that as unavoidable.

Using PSD for holding data related to this type of monitoring would also not be as costly memory wise if only enabling this on a subset of processes.

rickard-green:
Exactly what type of system monitor limit one would want I think it is good to discuss at Erlang forums. My guess is that a system monitor limit on signal queue length would be appreciated by many and would also not require any extra process specific memory usage. This is probably something that should be added regardless of whether more even "fancier" system monitor limits are added or not.

rickard-green:
> > > I suspect that functionality like this would fit much better in [erlang:system_monitor()](https://www.erlang.org/doc/man/erlang.html#system_monitor-2).
> > 
> > 
> > I happened to have this PR in works as well. It introduces `long_queue` monitoring event (similar to `large_heap`, but triggered by total queue size). However the actual functionality we need (and internally still use) is this: [max-au@a3d7f60](https://github.com/max-au/otp/commit/a3d7f6026768650937cc99203963423d1148bbb9)
> 
> I guess you by "total queue size" mean the value in `sig_qs.len` field?
> 
> It is okay to base such functionality on this value as long as it does not claim it is measuring the message queue length. Calling it signal queue length is okay, though.
> 

@max-au I spoke without thinking... This is not signal queue length either, so one cannot document it as such. It actually is not possible to give a good description of what it represents except under certain very specific conditions. Its value is somewhere between message queue length and some definition of signal queue length. So by using it one will get some sloppy value of something that is a message-signal queueish lenghtish yuck... :-( At the moment I'm too tired to think of alternatives, but I'll try to figure something out later...

max-au:
I am thinking of it as a "total amount of messages" sent to the process (both in the queue and in flight). 

The caveat I see is that signals are also counted towards sig_qs.len and it would be expensive to keep track of non-message signal count (and in fact NM-signals are also affecting process performance/reductions).

KennethL:
I understand it that both signals and messages must be handled by the process so in practice they all add to the backlog of things the process has to handle. And it is processes beginning to get a too big backlog we want to catch with the system_monitor.

rickard-green:
@max-au wrote:
> I am thinking of it as a "total amount of messages" sent to the process (both in the queue and in flight).
> 
> The caveat I see is that signals are also counted towards sig_qs.len and it would be expensive to keep track of non-message signal count (and in fact NM-signals are also affecting process performance/reductions).

It is only message signals included in `sig_qs.len` but also message signals that have not yet been received. However, *not all* message signals that have not yet been received. There might be a large amount of message signals in `sig_inq` as well, especially in situations when we got a large `sig_qs`, since we only fetch from `sig_inq` when we have run out of things to handle in `sig_qs`. You, however, really do *not* want to mess with `sig_inq`. It is not only protected by locking, but might nowadays also consist of up to 64 separate queues.

When having a look at the code I also realized that in the case you got a process info request signal in the queue, updates of `sig_qs.len` are also delayed. That is, reading it at any arbitrary time may give you a value significantly smaller than even the message queue length.

That is, it is more or less impossible to describe what the `sig_qs.len` value represents at an arbitrary point in time...

@KennethL wrote:
> I understand it that both signals and messages must be handled by the process so in practice they all add to the backlog of things the process has to handle. And it is processes beginning to get a too big backlog we want to catch with the system_monitor.

Yes such a system monitor limit is preferably based on some definition of signal queue length. A gen server call implies 3 signals (`monitor,` `message,` `demonitor`) from the client, so message signals are even in minority for this operation from the servers point of view. Such a value is, however, currently not easily accessible. If one restrict it to the signal queue length of the private signal queue, and accept increased memory consumption and decreased performance it can, however, easily be implemented. By adding two 64-bit words on each process, a couple of additions when fetching and a lot of subtractions (at every removal of a signal) one would have a value representing the signal queue length private to the process. There can however still be an unlimited amount of signals in transit to the process that won't be accounted for. If one would try to account for signals in transit it would be a much greater performance and scalability loss, so I don't even see that as a viable option at all.

rickard-green:
If using `sig_qs.len` for a system monitor one needs to be really fuzzy about what it is. A system monitor called `{incoming_message, Low, High}` one could perhaps be document as:

*If* the system detects an amount of incoming messages higher then `High` it will send you a monitor message, where "incoming messages" is defined as messages in the message queue and in transit. No more monitor message will be sent until the system has interpreted the amount of incoming as lower than `Low`. Note that the system might misinterpret the situation.

max-au:
@rickard-green 
 I think that most usages of "long message queue" monitoring event would be to implement "circuit breaker". For example, call `pg:leave` for the process that had accumulated long message queue to provide overload protection (or to change `message_queue_data` to `off_heap`).

I wonder how long it might take for `sig_inq` to be moved to `sig_qs`. My reading suggests that it may not happen for quite some time. If that is true, it would be useful to have `sig_inq` taken into account. 

Implementation that does not take `sig_inq` into account is cheap. It only affects how `sig_inq` gets moved to `sig_qs` (updating `sig_qs.len`). Before OTP 25, inner queue could not contain any non-message signals and it was easy to document that as "message queue length". With receive markers in the inner queue it is no longer valid explanation.

rickard-green:
> @rickard-green I think that most usages of "long message queue" monitoring event would be to implement "circuit breaker". For example, call `pg:leave` for the process that had accumulated long message queue to provide overload protection (or to change `message_queue_data` to `off_heap`).
> 
> I wonder how long it might take for `sig_inq` to be moved to `sig_qs`.

It can take a *very* long time and it can contain a huge amount of signals. There is no hard upper limit.

> My reading suggests that it may not happen for quite some time. If that is true, it would be useful to have `sig_inq` taken into account.
>

I don't see it as viable to take `sig_inq` into account. Both since it already today would be a scalability and performance bottleneck, but also since it could prevent future optimizations if introduced. The fact that it could prevent future optimizations to me makes it even worse than the fact that it would be a bottleneck.

> Implementation that does not take `sig_inq` into account is cheap. It only affects how `sig_inq` gets moved to `sig_qs` (updating `sig_qs.len`). Before OTP 25, inner queue could not contain any non-message signals and it was easy to document that as "message queue length". With receive markers in the inner queue it is no longer valid explanation.

No it can not be described as message queue length. Not even before receive markers. It is a mixture of message queue length and the amount of some messages in transit, but with hickups which can give you values less than message queue length. If one were to use it, one needs to document it as something extremely fuzzy like in my comment above. Otherwise, someone will want a bug fix on it some day.

PR #5728, opened 2022-02-18
erts: Detect libdlpi for cross compilation

joaohf:
Hello,

This PR fixes an `unsafe for cross-compilation` detected by autoconf when I was trying to cross-compile Erlang/OTP. It seems the erts/configure.ac is trying to find libdlpi in places not used for cross-compilation. However libdlpi is only provided by Solaris (I guess, not 100% sure). So this PR tries to test the host OS before trying to check the libdlpi.

```
builder@8f0f4d978071:/build/tmp/work/arm1176jzfshf-vfp-poky-linux-gnueabi/erlang/25.0-rc1-r0/git$ find -name config.log                                                                                                                                                                                           
./erts/config.log                                                                                                                                                                                                                                                                                                 
./lib/crypto/config.log                                                                                                                                                                                                                                                                                           
./lib/common_test/config.log                                                                                                                                                                                                                                                                                      
./lib/wx/config.log                                                                                                                                                                                                                                                                                               
./lib/snmp/config.log                                                                                                                                                                                                                                                                                             
./lib/megaco/config.log                                                                                                                                                                                                                                                                                           
./lib/erl_interface/config.log                                                                                                                                                                                                                                                                                    
./make/config.log                                                                                                                                                                                                                                                                                                 

$ grep  'unsafe for cross-compilation' ./erts/config.log                                                                                                                                                          
/build/tmp/work/arm1176jzfshf-vfp-poky-linux-gnueabi/erlang/25.0-rc1-r0/recipe-sysroot-native/usr/bin/arm-poky-linux-gnueabi/../../libexec/arm-poky-linux-gnueabi/gcc/arm-poky-linux-gnueabi/11.2.0/ld: warning: library search path "/lib" is unsafe for cross-compilation                                       
$ grep  'unsafe for cross-compilation' ./lib/crypto/config.log                                                                                                                                                    
$ grep  'unsafe for cross-compilation' ./lib/common_test/config.log                                                                                                                                               
$ grep  'unsafe for cross-compilation' ./lib/wx/config.log                                                                                                                                                        
$ grep  'unsafe for cross-compilation' ./lib/snmp/config.log                                                                                                                                                      
$ grep  'unsafe for cross-compilation' ./lib/megaco/config.log                                                                                                                                                    
$ grep  'unsafe for cross-compilation' ./lib/erl_interface/config.log  
```

Some details from erts/config.log are:

```
It was created by configure, which was
generated by GNU Autoconf 2.71.  Invocation command line was

[...]

PR #5769, opened 2022-03-06
wx: Detect OpenGL for cross compilation

joaohf:
Hello,

This PR intend to fix cross compilation issues for wx. And uses the same strategy following https://github.com/erlang/otp/pull/5728.

Without this patch I got:

```
builder@f985825fa055:/build/tmp/work/cortexa57-poky-linux/erlang/25.0-rc1-r0/git$ find -name config.log
./erts/config.log
./lib/crypto/config.log
./lib/common_test/config.log
./lib/wx/config.log
./lib/snmp/config.log
./lib/megaco/config.log
./lib/erl_interface/config.log
./make/config.log
builder@f985825fa055:/build/tmp/work/cortexa57-poky-linux/erlang/25.0-rc1-r0/git$ grep  'unsafe for cross-compilation' ./lib/wx/config.log
cc1: warning: include location "/usr/local/include" is unsafe for cross-compilation [-Wpoison-system-directories]
```

The general OpenGL detection is right, the issue is when the configure tries to find it in the alternative folders like /usr/X11R6 and usr/local/.

### DIFF
--- a/lib/wx/configure.ac
+++ b/lib/wx/configure.ac
@@ -262,14 +262,14 @@ AS_IF([test X"$host_os" != X"win32"],
     AS_IF([test X"$ac_cv_header_GL_gl_h" != Xyes && test X"$ac_cv_header_OpenGL_gl_h" != Xyes],
       [
        	saved_CPPFLAGS="$CPPFLAGS"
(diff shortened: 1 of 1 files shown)


IngelaAndin:
Same as #5770

dgud:
Wrong solution, and nothing have happend so I'll close this.

PR #5770, opened 2022-03-06
wx: Enable wx cross compilation

joaohf:
Hello,

This PR enables wx to cross compilation. Well, it just removes the checks for cross compilation.


garazdawi:
I’m not the one that decides if we want to support this or not, but if we do, then there is a cross compile GitHub actions dockerfile that should be updated to also build using wx.

dgud:
Wrong solution, and nothing have happend so I'll close this.

PR #5771, opened 2022-03-07
logger: Don't use original meta data in internal log

sirihansen:
logger_backend emits internal debug logs if a handler or a filter
fails. This is done using the ?LOG_INTERNAL macro. Prior to this
commit, these logs would get the meta data from the original failing
log event. This commit instead uses the actual location of the
internal debug log.

The original meta data is still included as a part of the log report,
so no data is lost.


sirihansen:
How can I solve the failing cross-compile check?

garazdawi:
Looks good. I rebased to master and will add to testing.

You can ignore the cross-compile check as it seems to just have been a network problem in github's servers.

garazdawi:
thanks!

PR #5772, opened 2022-03-08
crypto:macN - fix specs

ilya-klyuchnikov:
both `macN/4` and `macN/5` finally call `erlang:binary_part/3` which can return ONLY `binary()` (or throw an exception).


HansN:
Thanks for the PR!
There are however more places that need fixes.  I fixed all of them and added some proper description synchronized with an other larger change of the crypto app. Therefor I close this now.
But thanks for highlighting this!

ilya-klyuchnikov:
For the record, the actual change is: https://github.com/erlang/otp/commit/85dce1ca7898b325abb9d896fc7aaa7b2b873e36

PR #5775, opened 2022-03-08
beam_ssa_type: Fix type inference on #b_switch{} fail case

jhogberg:


lib/compiler/src/beam_ssa_type.erl
```diff
@@ -1948,11 +1948,19 @@ update_switch([], V, FailType, Ts, Ds, Ls, IsTempVar, Acc) ->
                      %% we can infer types as if we matched it directly.
                      Lit = #b_literal{val=Value},
                      infer_types_switch(V, Lit, Ts, IsTempVar, Ds);
-                 error when IsTempVar ->
-                     ts_remove_var(V, Ts);
                  error ->
-                     Ts#{ V := FailType }
+                     %% There's more than one possible value, try a weaker
+                     %% variant of the above.
```
bjorng, line 1953:
```suggestion
                     %% variant of the inference for case labels.
```

bjorng (approved):

PR #5776, opened 2022-03-08
erl_tar: Slightly optimize extraction from tar files

bjorng:
Running eprof when extracting tar files revealed that the top CPU
time consumers in the erl_tar process itself (that is, not processes
doing file IO) were checksum calculation and parsing of octal numbers.

This commit optimizes those functions. It turned out that the
checksum is calculated in two different ways, although the second
signed checksum is only neeeded if the first checksum is incorrect
(when extracting very old tar files).


lib/stdlib/src/erl_tar.erl
```diff
@@ -1253,37 +1259,38 @@ parse_numeric(<<First, _/binary>> = Bin) ->
...
-    end;
-do_parse_octal(<<$\s,Rest/binary>>, Acc) ->
+    Acc;
+do_parse_octal(<<$\s, Rest/binary>>, Acc) ->
     do_parse_octal(Rest, Acc);
 do_parse_octal(<<0, Rest/binary>>, Acc) ->
     do_parse_octal(Rest, Acc);
 do_parse_octal(<<C, Rest/binary>>, Acc) ->
-    do_parse_octal(Rest, <<Acc/binary, C>>).
+    Digit = C - $0,
```
artman41, line 1271: is it worth instead changing this to be `integer_to_list([C])` to avoid parsing non-integers as integers?

lib/stdlib/src/erl_tar.erl
```diff
@@ -1253,37 +1259,38 @@ parse_numeric(<<First, _/binary>> = Bin) ->
...
+    if
```
artman41, line 1272: would the following be faster?

Rather than using guard functions, `band` to limit the range to `0-7` then match that the digit hasn't changed.

e.g
```erlang
case Digit band 7 of
    Digit ->
        do_parse_octal(Rest, Acc bsl 3 bor Digit);
    _ ->
        throw({error, invalid_tar_checksum})
end
```

lib/stdlib/src/erl_tar.erl
```diff
@@ -1253,37 +1259,38 @@ parse_numeric(<<First, _/binary>> = Bin) ->
...
         {error, _Str, _Rest} ->
             throw({error, {bad_header, invalid_string}})
-    end;
-do_parse_string(<<0, _/binary>>, Acc) ->
-    do_parse_string(<<>>, Acc);
-do_parse_string(<<C, Rest/binary>>, Acc) ->
-    do_parse_string(Rest, <<Acc/binary, C>>).
+    end.
+
+strlen(<<>>, N) -> N;
```
artman41, line 1291: probably worth renaming this function to be something like `length_until_zero` or something, I feel like `strlen` isn't really descriptive enough

lib/stdlib/src/erl_tar.erl
```diff
@@ -1253,37 +1259,38 @@ parse_numeric(<<First, _/binary>> = Bin) ->
...
-do_parse_string(<<>>, Acc) ->
-    case unicode:characters_to_list(Acc) of
+    N = strlen(Bin, 0),
+    <<Prefix:N/binary,_/binary>> = Bin,
+    case unicode:characters_to_list(Prefix) of
         Str when is_list(Str) ->
             Str;
         {incomplete, _Str, _Rest} ->
-            binary:bin_to_list(Acc);
+            binary_to_list(Bin);
```
artman41, line 1286: Does it make more sense here to use `Prefix` rather than `Bin` just so that it's clear we're returning bytes that are non-zero rather than anything that was passed in?

Functionally I don't think it'd make much of a difference so up to you

lib/stdlib/src/erl_tar.erl
```diff
@@ -1019,11 +1019,14 @@ do_get_format({error, _} = Err, _Bin) ->
 do_get_format(#header_v7{}=V7, Bin)
   when is_binary(Bin), byte_size(Bin) =:= ?BLOCK_SIZE ->
     Checksum = parse_octal(V7#header_v7.checksum),
-    Chk1 = compute_checksum(Bin),
-    Chk2 = compute_signed_checksum(Bin),
-    if Checksum =/= Chk1 andalso Checksum =/= Chk2 ->
+    BadChecksum = case compute_checksum(Bin) of
```
artman41, line 1022: I'd probably rename `BadChecksum` to `IsBadChecksum` just to infer that it's a boolean rather than a Checksum Binary

artman41:
@bjorng hope you don't mind, saw you reference the pull request on the forums so had a quick look through and left some comments 😃

lib/stdlib/src/erl_tar.erl
bjorng: Done.

lib/stdlib/src/erl_tar.erl
bjorng: No. That would be slower and require a `try`/`catch` and is unnecessary because any non-digits will be caught by the range check.

lib/stdlib/src/erl_tar.erl
bjorng: That would likely be slightly faster. Changed.

lib/stdlib/src/erl_tar.erl
bjorng: I kept the original handling of incomplete Unicode characters. I am not sure that it the right thing to do, but I don't want to change any behavior of `erl_tar` in this pull request.

lib/stdlib/src/erl_tar.erl
bjorng: Yeah, naming is one of hardest problems. I tried a few names myself, until I settled on `strlen/1` because it is very similar to C's `strlen()`. I think it is clear enough from context.

bjorng:
@artman41 Thanks for your comments!

PR #5785, opened 2022-03-12
stdlib: Keep the tail of the strings passed to string:next_grapheme/1

chrrasmussen:
I am not sure if this is considered a bug or not, but I was tripped up by the behaviour of `string:next_grapheme/1` (compared to `string:next_codepoint/1`).

The following table shows that `string:next_codepoint/1` returns a value `[H | T]` where `T` is the remainder of the input string, after the first codepoint is removed. I would expect `string:next_grapheme/1` to follow the same pattern. However, if I call `string:next_grapheme(<<"a">>)`, it returns the value `[$a|""]` instead of `[$a|<<>>]`. (I have desugared the results to highlight the inconsistency.)

| String           | string:next_codepoint/1 | string:next_grapheme/1 |
| ---------------- | ----------------------- | ---------------------- |
| `"a"`            | `[$a\|""]`               | `[$a\|""]`              |
| `"ab"`           | `[$a\|"b"]`              | `[$a\|"b"]`             |
| `<<"a">>`        | `[$a\|<<>>]`             | `[$a\|""]` **(Does not return the tail of the input string)** |
| `<<"ab">>`       | `[$a\|<<"b">>]`          | `[$a\|<<"b">>]`         |
| `[<<"a">>, "b"]` | `[$a\|[<<>>,"b"]]`       | `[$a\|[<<>>,"b"]]`      |

<details><summary>Erlang code to generate the results above</summary>

```erlang
lists:map(fun(X) -> {string:next_codepoint(X), string:next_grapheme(X)} end, ["a", "ab", <<"a">>, <<"ab">>, [<<"a">>, "b"]]).
```
</details>

Looking at the source code it seems that the intention is that the tail of the passed string should be kept: https://github.com/erlang/otp/blob/b21615d9719d530291418e91d5431998bbc2b8c1/lib/stdlib/uc_spec/gen_unicode_mod.escript#L665

This pull request fixes this inconsistency (if desirable) and adds new test cases.


dgud:
I am ambivalent on this, on one hand it is nice with consistency,
but it loses a tiny bit a performance (and garbage, creating 2 words extra) and I don't know for what purpose, 
you will still need to test [] = string:next_grapheme(<<>>) or use string:is_empty() for the general case.

The comment in code is referring to forward looking which picks out the next codepoint and looks for changes,
you could gain performance by doing [FirstGC,SecondCP|Rest] where SecondCP had already picked out of the binary,
but then string module would have to recreate the binary on return.

chrrasmussen:
I see 😊 Thanks for looking into it!

My motivation for this change originated from this commit: https://github.com/chrrasmussen/Idris2-Erlang/commit/698a73146a82dfbb148a0dad91c214649cc0c9ed

<details>
<summary>More details</summary>

The [Idris2-Erlang](https://github.com/chrrasmussen/Idris2-Erlang) repository is an Erlang code generator for [Idris 2](http://github.com/idris-lang/Idris2) (A purely functional programming language with first class types).
`StrTail` is one of the primitives in the Idris 2 compiler and it has has the following type signature:
```idris
partial
prim__strTail : String -> String
```

In the code generator I decided to represent Idris 2's `String` type as an Erlang binary. That means I need to make sure that I always pass around binaries (and not any of the other string variants in Erlang). As `prim__strTail` is a primitive, I can assume that it will always be called with a non-empty `String`. Regarding the implementation, it only means that the trivial solution (calling `string:next_grapheme/1` and returning the `tl` of the list) does not work in my case, but, as you say, there are many ways to go about this.
</details>

The changes in this pull request is not important to me. I was encouraged to try to make this "fix" because of the recent addition of the [`HOWTO/DEVELOPMENT.md`](https://github.com/erlang/otp/blob/083e2931d4cea6f63dfa2f6f776e2f897bd3e8e3/HOWTO/DEVELOPMENT.md) document. It was quite easy to get up and running. Thanks!

dgud:
An argument for your point from the string docs, which I have written.

> Unless otherwise specified the return value type is the same as the input type. That is, binary input returns binary output, list input returns a list output, and mixed input can return a mixed output.

It is a minor incompatibility though, but will add it testing after RC-2.

PR #5790, opened 2022-03-15
ssl_logger improvements

Flameeyes:
This includes a small change to fill in the location for all ssl_logger users.

`maybe_client_warn_no_verify` was the only caller that was missing location metadata.

### DIFF
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -26,6 +26,7 @@
 -module(ssl).
 
 -include_lib("public_key/include/public_key.hrl").
+-include_lib("kernel/include/logger.hrl").
 
 -include("ssl_internal.hrl").
 -include("ssl_api.hrl").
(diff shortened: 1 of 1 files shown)


IngelaAndin:
This change is not as small as it may seem. This would make the notice and warning alerts of the ssl application disappear from the default log. The messages that today uses the otp/ssl domain are debugging messages that will be handled by the ssl applications own logger handler. To be able to group all ssl logs under the same top domain sounds like something that could be desirable, but would probably require some new logger application support and some new subdomains in ssl. We will consider implementing such a possibility for the future, but we can not include this PR as it is right now.

Flameeyes:
@IngelaAndin what about the first of the two commits (providing location for `maybe_client_warn_no_verify`)?

The problem right now is that there s no way to select logs coming from the SSL app, because that one source does not have even module information :/

IngelaAndin:
The first commit could be added now. It's the domain part that is problematic. We have written an internal ticket to handle that part so if you want to change this PR to only provide the first commit we will consider it for inclusion for OTP-25

Flameeyes:
Done that, this is now a single commit pull request.

PR #5793, opened 2022-03-16
erl script: Support installation paths with spaces

bjorng:
Quote everything in the `erl` script so that it will be possible to
start Erlang if the installation has been moved to a path with
spaces in its name.

### DIFF
--- a/erts/etc/unix/erl.src.src
+++ b/erts/etc/unix/erl.src.src
@@ -19,7 +19,7 @@
 # %CopyrightEnd%
 #
 prog="$0"
-progdir="`dirname ${prog}`"
+progdir=`dirname "${prog}"`
 dyn_erl_path="${progdir}/%DYN_ERL_PATH%"
 if [ ! -f "$dyn_erl_path" ]
 then
@@ -28,9 +28,9 @@ fi
 
 if [ -f "$dyn_erl_path" ]
 then
-    dyn_rootdir=`${dyn_erl_path} --realpath`
-    dyn_rootdir=`dirname ${dyn_rootdir}`
-    dyn_rootdir=`dirname ${dyn_rootdir}`
+    dyn_rootdir=`"${dyn_erl_path}" --realpath`
+    dyn_rootdir=`dirname "${dyn_rootdir}"`
+    dyn_rootdir=`dirname "${dyn_rootdir}"`
     dyn_rootdir="${dyn_rootdir}%DYN_ROOTDIR_BASE_EXT%"
 else
     dyn_rootdir=""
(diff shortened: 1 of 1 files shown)


erts/etc/unix/erl.src.src
```diff
@@ -19,7 +19,7 @@
 # %CopyrightEnd%
 #
 prog="$0"
-progdir="`dirname ${prog}`"
+progdir="`dirname "${prog}"`"
```
rickard-green, line 22:
```suggestion
progdir=`dirname "${prog}"`
```
should work and by this avoiding ambiguous interpretation of the double quoting

erts/etc/unix/erl.src.src
```diff
@@ -28,9 +28,9 @@ fi
 
 if [ -f "$dyn_erl_path" ]
 then
-    dyn_rootdir=`${dyn_erl_path} --realpath`
-    dyn_rootdir=`dirname ${dyn_rootdir}`
-    dyn_rootdir=`dirname ${dyn_rootdir}`
+    dyn_rootdir="`"${dyn_erl_path}" --realpath`"
```
rickard-green, line 31: same here
```suggestion
    dyn_rootdir=`"${dyn_erl_path}" --realpath`
```

erts/etc/unix/erl.src.src
```diff
@@ -28,9 +28,9 @@ fi
...
+    dyn_rootdir="`dirname "${dyn_rootdir}"`"
+    dyn_rootdir="`dirname "${dyn_rootdir}"`"
```
rickard-green, line 33: same here
```suggestion
    dyn_rootdir=`dirname "${dyn_rootdir}"`
```

erts/etc/unix/erl.src.src
```diff
@@ -47,7 +47,7 @@ then
 else
     ROOTDIR="$ERL_ROOTDIR"
 fi
-BINDIR=$ROOTDIR/erts-%VSN%/bin
+BINDIR="$ROOTDIR/erts-%VSN%/bin"
 EMU=%EMULATOR%%EMULATOR_NUMBER%
 PROGNAME=`echo $0 | sed 's/.*\///'`
```
rickard-green, line 52:
```suggestion
PROGNAME=`basename "$0"`
```

rickard-green (changes requested):

erts/etc/unix/erl.src.src
rickard-green: same thing for the two lines below this as well

PR #5795, opened 2022-03-17
wx, observer: macOS menu bar improvements

wojtekmach:
### 2. observer: Call original event handlers for menu commands

This is important for macOS where by default we get the "App menu" with:

- Services
- Hide [app]
- Hide Others
- Show All
- Quit [app]

Before this patch, "Hide [app]", "Hide Others", and "Show All" are not
doing anything. This is what we get when clicking "Hide [app]":

    observer_sys_wx:238: Unhandled event {wx,5250,
                                          {wx_ref,35,wxFrame,[]},
                                          [],
                                          {wxCommand,command_menu_selected,[],-1,
                                           0}}

After this patch, the original (system-level) event handlers are fired
and on the Erlang side we can simply ignore these events.


dgud:
Nice, but "wx.hrl" is a generated file, so we need to fix the generator or its input to fix this, otherwise it will lost
the next time someone does something.

dgud:
Great, I can probably also take a look next week or so.

wojtekmach:
I tried running codegen but no luck. Here's what I did:

```
$ cd lib/wx/api_gen
$ make WX_DIR=$HOME/src/wxWidgets # my checkout is at v3.1.5
```

here's the log: https://gist.github.com/wojtekmach/ce3e9d8a0b1894b75b50b53a85fb356c

I installed doxygen with brew. I noticed my `lib/wx/api_docs/wx_xml` dir is empty.

dgud:
You will need to checkout  **wx_api_gen** branch from **git@github.com:erlang/wxWidgets.git**
I have not tested to generate the code on a mac (only linux or windows-wsl) but it should work :-)
You will need doxygen of a good version (i.e. what I have on ubuntu 1.8.17)..

dgud:
I added the defines in repository above and re-generated the code, and you had missed one place,
pushed that to your branch.

dgud:
Can you check the debugger as well?  (I don't have a mac)

wojtekmach:
I just checked debugger and the menu commands work well. Which is a bit surprising since it doesn't look like they don't pass `[{skip, true}]`.

wojtekmach:
@dgud if it's any help I tracked down the change to `defs.h` to this commit: https://github.com/wxWidgets/wxWidgets/commit/c46d050368659ba3afff87b8c0127ce2e77860b5#diff-e391c490e46ded5849800b8de65d0a1ea24e9b75bb38d2714a6a786a2d4a51f9

and it first show up in v2.9.2 release:

```
$ git tag --contains c46d050 | grep v | head -1
v2.9.2
```

dgud:
I think this is a good, can you squash my commit into yours.

dgud:
Also it works in the debugger, because the **wxXXX:connect(Menu, command_menu_selected, [{id, MenuId}])** 
is done on specific id, so the mac specific events will not be sent to erlang and handled by the default handler.

In observer we connect to all menu events.

wojtekmach:
Squashed! I also changed the branch this is opened against to `maint`. Please let me know if that's ok.

> Also it works in the debugger, because the `wxXXX:connect(Menu, command_menu_selected, [{id, MenuId}])` is done on specific id

Ah, that makes sense. Thank you!

dgud:
I changed it back to master, I don't think it is important enough to make an emergency patch and
we don't have any more maintenance patch package planned.

dgud:
Will merge after 25-rc2

wojtekmach:
sounds good, thanks!

PR #5660, opened 2022-01-28
Split TypEr core out of typer module

elbrujohalcon:
We (@pablocostass, @maco, and myself) are building [`rebar3_typer`](https://github.com/adRoll/rebar3_typer) and, as part of our efforts to integrate TypEr into `rebar3`, we found that `typer` (the module) didn't expose any interface that allowed us to use it from another Erlang module. That's why we decided to move some pieces of it into our project. In particular: we created `typer_core` with just the core functionality, detached from the parts that parse the command-line arguments and options.
Eventually, we came up with AdRoll/rebar3_typer#15, and we thought it would be great if we could contribute this back to OTP itself, so… for future versions of the plugin, we don't need to use _our own version of typer_ and we can just use what comes with OTP.
As a bonus, we added a new mode: `annotate_in_place`. It's like `annotate` but it replaces the original files instead of creating new ones.

In our project, we also added several [tests](https://github.com/AdRoll/rebar3_typer/blob/main/test/rebar3_typer_prv_SUITE.erl) for `typer_core`. I didn't move them here since I don't know if they match the test structure of this project. If you want to check them out and let us know if you like them, we can migrate them here of course.

PR #5802, opened 2022-03-21
Add annotate_in_place mode to TypEr

pablocostass:
This commit adds a new analysis mode to be ran with TypEr, `annotate_in_place`, which can be used to annotate the specs that the tool inferred directly on the source code files.

We (@maco, @elbrujohalcon and myself) developed this new mode during an internal hackathon at NextRoll. It was originally introduced in #5660, but we decided to remove it from that PR as we wanted to test it a bit more thoroughly, and to reduce the diff noise that said PR already had.


jhogberg:
Thanks for the PR! Can you update [the documentation](https://www.erlang.org/doc/man/typer.html) as well?

pablocostass:
@jhogberg Let me know if you need me to squash both commits into one

jhogberg:
> @jhogberg Let me know if you need me to squash both commits into one

Yes please.

> But I would avoid this warning entirely.

Seconding this, and it's probably a good idea to remove it from `--annotate-inc-files` too. Unless we know there's a bug that cannot be fixed within the bounds of the current interface, I don't feel it's right to warn about it.

lib/dialyzer/src/typer.erl
```diff
@@ -187,7 +189,7 @@ version_message() ->
 
 help_message() ->
   S = <<" Usage: typer [--help] [--version] [--plt PLT] [--edoc]
-              [--show | --show-exported | --annotate | --annotate-inc-files]
+              [--show | --show-exported | --annotate | --annotate-inc-files | --anotate-in-place]
```
jhogberg, line 192: Nitpick:
```suggestion
              [--show | --show-exported | --annotate | --annotate-inc-files | --annotate-in-place]
```

pablocostass: Oh gosh, I knew I would make that typo somewhere. Good catch! :)

jhogberg:

PR #5803, opened 2022-03-21
Correct typespec for inet_res:getbyname/2,3 (PR #5412, OTP-17986)

RaimoNiskanen:
Closes PR #5412.

Instead of changing the `#hostent{}` record type; correct the return type for `inet_res:getbyname/2,3` by creating a parallel tuple type that defines the peculiar use of the `#hostent{}` record.

### DIFF
--- a/lib/kernel/doc/src/inet_res.xml
+++ b/lib/kernel/doc/src/inet_res.xml
@@ -4,7 +4,7 @@
 <erlref>
   <header>
     <copyright>
-      <year>2009</year><year>2021</year>
+      <year>2009</year><year>2022</year>
       <holder>Ericsson AB. All Rights Reserved.</holder>
     </copyright>
     <legalnotice>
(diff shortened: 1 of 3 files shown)


bmk:
getbyname/2,3
I think it looks like the third section (starting with "This function uses resolver...") is the most
general description of the function and therefor should be moved up to start the description.
The second section has "h_addr_list = [dns_data()]" as part of the explanation, but since that is
what the type (hostent()) is, it is not unique.
Maybe make the alternative resolves into a bullet list?

RaimoNiskanen:
> I think it looks like the third section (starting with "This function uses resolver...") is the most
general description of the function and therefor should be moved up to start the description.

I think the first paragraph starting with "Resolves a DNS record of the specified Type for the specified host..." is the most general description.  That it uses the `search` list is a detail.

I will rephrase to reduce the cluttering when describing the `#hostent{}` fields.

PR #5806, opened 2022-03-22
erts: Fix crash dump assert when heart is set

garazdawi:
If a crash dump was generated while the VM was supervised
with heart, the debug build would assert because the heart
port was flushed.

This commit moves where erts_is_crash_dumping is set so that
the heart flush works as it should.

More importantly the commit also moves the
erts_thr_progress_fatal_error_block to be earlier so
that multiple threads do not race to create the crash dump.
The thr progress block was incorrectly moved by [#5315](https://github.com/erlang/otp/pull/5315) when
fixing a problem with the suspension of schedulers.


max-au (approved):
Thanks!
Took me a while to understand why there is `test_server` in the test case.

PR #5807, opened 2022-03-22
Make ?CATCH macro "syntactical"

michalmuskala:
Hiding a keyword behind a macro messes up syntax highlighting and
some tools that rely on analysing code without expanding macros.

Additionally the `catch` keyword has a fairly surprising precedence,
wrapping it in a macro with a proper argument makes it less error prone.

Debugging is still possible by changing the macro to just

    -define(CATCH(Expr), Expr).

### DIFF
--- a/erts/preloaded/src/prim_zip.erl
+++ b/erts/preloaded/src/prim_zip.erl
@@ -42,7 +42,7 @@
 -define(READ_BLOCK_SIZE, 16*1024).
 
 %% for debugging, to turn off catch
--define(CATCH, catch).
+-define(CATCH(Expr), (catch (Expr))).
 
 -record(primzip_file,
 	{name,
@@ -203,7 +203,7 @@ get_z_all(?DEFLATED, Compressed, Z, _F) ->
     ok = zlib:inflateInit(Z, -?MAX_WBITS),
     Uncompressed = zlib:inflate(Z, Compressed),
     %%_CRC = zlib:crc32(Z),
-    ?CATCH zlib:inflateEnd(Z),
+    _ = ?CATCH(zlib:inflateEnd(Z)),
     erlang:iolist_to_binary(Uncompressed); % {erlang:iolist_to_binary(Uncompressed), CRC}
 get_z_all(?STORED, Stored, _Z, _F) ->
     %%CRC0 = zlib:crc32(Z, <<>>),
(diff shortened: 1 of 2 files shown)


michalmuskala:
I was able to reproduce the test failures locally, but it also fails on master for me, so I don't think it's related

erts/preloaded/src/prim_zip.erl
```diff
@@ -350,7 +350,7 @@ prim_file_io({file_info, F}, _) ->
 	{error, E} -> throw(E)
     end;
 prim_file_io({open, FN, Opts}, _) ->
-    case ?CATCH prim_file:open(FN, Opts++[binary]) of
+    case ?CATCH(prim_file:open(FN, Opts++[binary])) of
```
bjorng, line 353:
```suggestion
    case prim_file:open(FN, Opts++[binary]) of
```

As far as I understand, `prim_file:open/2` can't raise an exception. @jhogberg, do you agree, or could there be some other exception that the `try`/`catch` in `prim_file:open/2` won't catch?

erts/preloaded/src/prim_zip.erl
```diff
@@ -203,7 +203,7 @@ get_z_all(?DEFLATED, Compressed, Z, _F) ->
     ok = zlib:inflateInit(Z, -?MAX_WBITS),
     Uncompressed = zlib:inflate(Z, Compressed),
     %%_CRC = zlib:crc32(Z),
-    ?CATCH zlib:inflateEnd(Z),
+    ?CATCH(zlib:inflateEnd(Z)),
```
bjorng, line 206:
```suggestion
    _ = ?CATCH(zlib:inflateEnd(Z)),
```

bjorng (approved):
I think this is a good and safe change for OTP 25.

For a future release, such as OTP 26, it would be good to replace all `case ?CATCH(...) of` with `try`/`catch`. That change will need some verification/testing, because it is not immediately clear whether the `?CATCH` is intended to catch both `throw` and `error` exceptions or only `throw` exceptions.

(I consider the use of the old-style `catch` fine when its return value is ignored, such as in `catch zlib:inflateEnd(Z)`.)

bjorng:
Please squash the commits.

bjorng:
Thanks for your pull request.

PR #5809, opened 2022-03-22
Add option +IOs to disable scheduler thread polling

sverker:
Default is true, scheduler thread polling enabled as before.

Scheduler polling has in some cases seen to cause longer read event latency, and can now be disabled at the expense of some increased cpu load.

### DIFF
--- a/erts/emulator/sys/common/erl_check_io.c
+++ b/erts/emulator/sys/common/erl_check_io.c
@@ -2693,9 +2693,17 @@ static int erts_debug_print_checkio_state(erts_dsprintf_buf_t *dsbufp,
                     err = 1;
             }
             else {
-                ErtsPollEvents ev = cio_events;
-                if (ev != ep_events && ep_events != ERTS_POLL_EV_NONE)
-                    err = 1;
+                if (ep_events != ERTS_POLL_EV_NONE) {
+                    if (!ERTS_POLL_USE_KERNEL_POLL
+                        || (!(state->flags & (ERTS_EV_FLAG_SCHEDULER|ERTS_EV_FLAG_FALLBACK))
+                            && ((cio_events ^ ep_events) & ep_events) != 0)) {
+                        err = 1;
+                    }
+                    /* else: Kernel poll with oneshot (used by poller threads)
+                     *       may cause a race where an event just triggered and
+                     *       thereby was cleared in the pollset (ep_events).
+                     */
+                }
                 erts_dsprintf(dsbufp, "cio_ev=");
                 print_events(dsbufp, cio_events);
                 erts_dsprintf(dsbufp, " ep_ev=");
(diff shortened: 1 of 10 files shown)


erts/emulator/sys/common/erl_check_io.c
```diff
@@ -93,24 +93,27 @@ typedef enum {
 #endif
     ERTS_EV_FLAG_WANT_ERROR    = 0x10,  /* ERL_NIF_SELECT_ERROR turned on */
 
-    /* Combinations */
-    ERTS_EV_FLAG_USED_FALLBACK = ERTS_EV_FLAG_USED | ERTS_EV_FLAG_FALLBACK,
-    ERTS_EV_FLAG_USED_SCHEDULER = ERTS_EV_FLAG_USED | ERTS_EV_FLAG_SCHEDULER,
-    ERTS_EV_FLAG_USED_IN_SCHEDULER = ERTS_EV_FLAG_USED | ERTS_EV_FLAG_SCHEDULER | ERTS_EV_FLAG_IN_SCHEDULER,
-    ERTS_EV_FLAG_UNUSED_SCHEDULER = ERTS_EV_FLAG_SCHEDULER,
-    ERTS_EV_FLAG_UNUSED_IN_SCHEDULER = ERTS_EV_FLAG_SCHEDULER | ERTS_EV_FLAG_IN_SCHEDULER
```
garazdawi, line 101: The nice thing about this enum was that gdb manages to print the flag combinations.

erts/emulator/sys/common/erl_check_io.h
```diff
@@ -112,6 +112,22 @@ typedef struct {
 } ErtsIoTask;
 
 
+ERTS_GLB_INLINE int erts_sched_poll_enabled(void);
+
+#if ERTS_GLB_INLINE_INCL_FUNC_DEF
+
+ERTS_GLB_INLINE int erts_sched_poll_enabled(void)
+{
+#if ERTS_POLL_USE_SCHEDULER_POLLING
```
garazdawi, line 121: Should we rename this define to `ERTS_POLL_ENABLE_SCHEDULER_POLLING` or something similar? I was confused at first as to why the define was needed at all.

garazdawi (commented):
In emulator there is the testcase scheduler_SUITE:poll_thread. I think we should add a testcase there that the option works.

Also, we ignore the option if ERTS_POLL_USE_SCHEDULER_POLLING is false, is that intentional?

sverker:
> we ignore the option if ERTS_POLL_USE_SCHEDULER_POLLING is false, is that intentional?

The easy way out I guess. I have pushed a commit 5e7aa243a031c0804572740c6e8c84345a014f6d that refuse +IOs true if ERTS_POLL_USE_SCHEDULER_POLLING is false.

sverker:
@garazdawi Three new commits, can you have a new look?

garazdawi (approved):
Looks good to me!

Issue #5782, opened 2022-03-10
httpc_client ignores connect_timeout when going through a proxy


max-au:
**Describe the bug**
From the [documentation](https://www.erlang.org/doc/man/httpc.html):

```
When https links need to go through a proxy, the CONNECT method extension to HTTP-1.1 is used to establish a tunnel and then the connection is upgraded to TLS. 
```

When `http_client` establishes the tunnel, it does not specify timeout to `ssl:connect` here: https://github.com/erlang/otp/blob/master/lib/inets/src/http_client/httpc_handler.erl#L1647

**To Reproduce**
Use `httpc_client:set_option([{https_proxy, _}], Profile)` to setup an HTTPS proxy, and try to make a `request` when proxy is up but the destination is not. Specify `connect_timeout` in HTTP options. The call appears to hang (and not honour the timeout)

**Expected behavior**
Request times out as specified by `connect_timeout`.

**Affected versions**
All versions since OTP16.

**Additional context**
I [made an attempt](https://github.com/erlang/otp/pull/1829) before, but could not come up with a good test case (in fact I failed to setup the test environment, it turned to be quite a tricky job for me at that point in time).

u3s:
`httpc_proxy_SUITE` is quite close to what we might need for testing this behavior.
unfortunately it seems it hasn't been run for a while - I'm working on re-enabling it.

u3s:
a proposal available here: https://github.com/erlang/otp/pull/5824

max-au:
works for me! Thank you!

PR #5810, opened 2022-03-23
gh: add tinyproxy to Dockerfile.ubuntu-base

u3s:


MessageMap:
Just wondering what is use case for tinyproxy?

max-au:
I'd hope to have some testing for #5782 with it

PR #5816, opened 2022-03-24
wx: Add MacGetCommonMenuBar and MacSetCommonMenuBar

wojtekmach:
This is especially useful for apps that manage multiple wxFrames where we can close the last window but want the application to keep running. Before this patch we'd be stuck with the default menu bar, created in `wxe_impl.cpp`, that we cannot interact with. After this patch we can, which means we can add new menu items, subscribe to events, etc.

Example:

```erlang
wx:new(), MB = wxMenuBar:macGetCommonMenuBar(), M = wxMenuBar:oSXGetAppleMenu(MB), wxMenu:setTitle(M, "Test").
```


lib/wx/c_src/gen/wxe_init.cpp
```diff
@@ -6664,6 +6664,12 @@ void WxeApp::init_consts(wxeMetaCommand& event) {
     { enif_make_atom(rt.env,"wxStandardID"), "wxID_MDI_WINDOW_PREV", rt.make_int(wxID_MDI_WINDOW_PREV) },
     { enif_make_atom(rt.env,"wxStandardID"), "wxID_MDI_WINDOW_NEXT", rt.make_int(wxID_MDI_WINDOW_NEXT) },
     { enif_make_atom(rt.env,"wxStandardID"), "wxID_MDI_WINDOW_LAST", rt.make_int(wxID_MDI_WINDOW_LAST) },
+    { enif_make_atom(rt.env,"wxStandardID"), "wxID_OSX_MENU_FIRST", rt.make_int(wxID_OSX_MENU_FIRST) },
+    { enif_make_atom(rt.env,"wxStandardID"), "wxID_OSX_HIDE", rt.make_int(wxID_OSX_HIDE) },
+    { enif_make_atom(rt.env,"wxStandardID"), "wxID_OSX_HIDEOTHERS", rt.make_int(wxID_OSX_HIDEOTHERS) },
+    { enif_make_atom(rt.env,"wxStandardID"), "wxID_OSX_SHOWALL", rt.make_int(wxID_OSX_SHOWALL) },
+    { enif_make_atom(rt.env,"wxStandardID"), "wxID_OSX_SERVICES", rt.make_int(wxID_OSX_SERVICES) },
+    { enif_make_atom(rt.env,"wxStandardID"), "wxID_OSX_MENU_LAST", rt.make_int(wxID_OSX_MENU_LAST) },
```
wojtekmach, line 6672: this change will land as part of #5795, need to wait for that one to land first.

lib/wx/src/gen/wxMenuBar.erl
```diff
@@ -253,6 +254,19 @@ oSXGetAppleMenu(#wx_ref{type=ThisT}=This) ->
   wxe_util:queue_cmd(This,?get_env(),?wxMenuBar_OSXGetAppleMenu),
   wxe_util:rec(?wxMenuBar_OSXGetAppleMenu).
 
+%% @doc See <a href="http://www.wxwidgets.org/manuals/2.8.12/wx_wxmenubar.html#wxmenubarmacgetcommonmenubar">external documentation</a>.
```
wojtekmach, line 257: this link is broken, there's no such function on wx 2.8, I think this is a wx 3.0+ addition. I assume it is ok though as wxMenuBar.xml contains the canonical docs.

leoliu:
Thanks for addressing another pain point on macOS.

In my last experiment I worked around this by using an invisible (size 0) frame as suggested in the wxWidgets wiki. It was tedious. For example, you need to take care of not having the invisible taking up space in Expose.

wojtekmach:
Rebased.

dgud:
Thanks

PR #5824, opened 2022-03-25
inets: re-use connect_timeout option in TLS upgrade

u3s:
- re-use connect_timeout during TLS upgrade of connection over proxy
    - connect_timeout test in inets
    - TLS upgrade client timeout test in ssl

### DIFF
--- a/lib/inets/src/http_client/httpc_handler.erl
+++ b/lib/inets/src/http_client/httpc_handler.erl
@@ -1637,14 +1637,14 @@ host_header(_, URI) ->
 tls_upgrade(#state{status = 
 		       {ssl_tunnel, 
 			#request{settings = 
-				     #http_options{ssl = {_, TLSOptions0} = SocketType},
+				     #http_options{ssl = {_, TLSOptions0} = SocketType,
+						   connect_timeout = ConnectTimeout},
 				     address = {Host, _} = Address} = Request},
 		   session = #session{socket = TCPSocket} = Session0,
 		   options = Options} = State) ->
 
     TLSOptions = maybe_add_sni(Host, TLSOptions0),
-
-    case ssl:connect(TCPSocket, TLSOptions) of
+    case ssl:connect(TCPSocket, TLSOptions, ConnectTimeout) of
 	{ok, TLSSocket} ->
 	    ClientClose = httpc_request:is_client_closing(Request#request.headers),
 	    SessionType = httpc_manager:session_type(Options),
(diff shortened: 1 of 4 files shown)


lib/inets/doc/src/httpc.xml
```diff
@@ -291,6 +291,7 @@
         <v>http_options() = [http_option()]</v>
         <v>http_option() = {timeout,         timeout()} | 
                            {connect_timeout, timeout()} | 
+                           {upgrade_timeout, timeout()} |
```
max-au, line 294: I wonder if there are any naming guidelines around this module. Should it be "upgrade_timeout" or "tls_upgrade_timeout"? it might not be easy to understand what the upgrade is.

u3s: I've no strong opinion here, on the other hand we've `connect_timeout` and not `tcp_connect_timeout` (maybe TCP and TLS are implicitly assumed here).
Maybe @IngelaAndin could share opinion as well ?

u3s: I've picked the proposal you've made earlier to have a dedicated timeout for upgrade for simplicity reasons.
In comparison to

1. re-using timeout
2. or making some timeout algebra (as suggested by Max in previous PR)

a dedicated timeout is easy to test (I can set only 2nd timeout in chain to zero) and also brings the biggest control to user.
Due to above I decided to start discussion with that proposal.

Timeout chain
=========
TCP TIMEOUT -> TLS UPGRADE TIMEOUT

if the same value is used for both, the 1st will be always triggered first in test ...

u3s:
we had a discussion and decided that maybe adding a new option is not a preferred solution.

i would like to propose something simple:
- re-using connect_timeout value for TLS upgrade step (so we don't introduce new option and possibly confusion)
- the simplest re-use of the value as is without any recalculation - my thinking is that if TCP connection times out then we have a single timeout, but if it succeeds then maybe it is pretty quick on average and taking it out from timeout budget is not worth doing
- test coverage is kind of implicit and combined from: test in inets for connect_timeout (but failing at TCP connect phase) + client TLS upgrade timeout tested in tls_api_SUITE

@max-au what do you think? is it good enough for you?

max-au (approved):

lib/inets/doc/src/httpc.xml
u3s: marking as resolved because we no longer have `upgrade_timeout` in PR at all :-)

PR #5831, opened 2022-03-28
stdlib: optimize gen_server by caching callback functions

frazze-jobb:
Cache callback functions handle_call, handle_cast and handle_info in a tuple when initializing the gen_server. This "should" give us a minor speed up by skipping lookups in export entry.

### DIFF
--- a/lib/stdlib/src/gen_server.erl
+++ b/lib/stdlib/src/gen_server.erl
@@ -147,6 +147,22 @@
 	( (X) =:= infinity orelse ( is_integer(X) andalso (X) >= 0 ) )
 ).
 
+-record(callback_cache,{module :: module(),
+                        handle_call :: fun((Request :: term(), From :: from(), State :: term()) ->
+                            {reply, Reply :: term(), NewState :: term()} |
+                            {reply, Reply :: term(), NewState :: term(), timeout() | hibernate | {continue, term()}} |
+                            {noreply, NewState :: term()} |
+                            {noreply, NewState :: term(), timeout() | hibernate | {continue, term()}} |
+                            {stop, Reason :: term(), Reply :: term(), NewState :: term()} |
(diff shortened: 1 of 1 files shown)


max-au:
Does it keep compatibility with existing hot code loading semantic?
Previously, executing `code:load` for module implementing `gen_server` behaviour resulted in an immediate change.

jhogberg:
> Does it keep compatibility with existing hot code loading semantic?
> Previously, executing `code:load` for module implementing `gen_server` behaviour resulted in an immediate change.

Yes, the funs refer to the MFAs and not the currently loaded code.

RaimoNiskanen:
So `Mod:handle_cast(Request, State)` is, after `HandleCast = fun Mod:handle_cast/2` slower than `HandleCast(Request, State)` ??

Is it the `fun()` creation time that it is worth the work to avoid?
Because the call will go through the export entry anyway...

jhogberg:
> So `Mod:handle_cast(Request, State)` is, after `HandleCast = fun Mod:handle_cast/2` slower than `HandleCast(Request, State)` ??

Yes.

> Is it the `fun()` creation time that it is worth the work to avoid? Because the call will go through the export entry anyway...

We want to avoid having to look up the export entry from MFA on every call. Remote funs do that once on creation, but `apply/3` (which `Mod:handle_cast/2` is syntactic sugar for) has to do that every time.

[Here's a small benchmark](https://gist.github.com/jhogberg/5d549f1e47e64b74560026b6c2d1fa71) demonstrating the difference.

essen:
> [Here's a small benchmark](https://gist.github.com/jhogberg/5d549f1e47e64b74560026b6c2d1fa71) demonstrating the difference.

Should this be mentioned in the efficiency guide?

jhogberg:
Sure, feel free to make a PR. :)

PR #5836, opened 2022-03-30
Implement fast niche PRNGs

RaimoNiskanen:
This PR adds two very fast but low quality niche algorithms to the `rand` module, not as part of the plug-in framework, to avoid the overhead.

The algorithms are two Linear Congruential Generators, one purely multiplicative - **`mcg35`**, with a prime modulus and one power of two based - **`lcg35`**.  Both have 35 bits, the multiplicative just short of.  According to state of the art today these are historical toy generators with plenty of statistical shortcomings, but an impressive speed, which is what still might make them relevant for niche applications such as "random" load balancing, routing, etc.

The reason for adding them to the `rand` module is so that users will not need to roll their own and accidentally do some unfortunate mistake that makes their implementation even worse.

This PR also exports the internal functions `exsp_next/1` and `exsp_jump/1` from the fastest generator in the `rand` module's normal repertoire, which enables users to use them without the plug-in framework overhead (and without the range facilities, `float()` generation and all the other conveniences of the plug-in framework).  This can cut down the number generation time to 1/3 of the framework's, provided that the application can make use of just a raw 58-bit number.

The function `splitmix64_next/1` is also exported to tap in to the generator internally used in the `rand` module for seeding other generators with large state arrays, which can be useful for customizing generator seeding.

#### Test improvements and optimizations
* Add a warm-up set to measure/1
* Compensate measure/1 numbers for overhead
* Measure the process dictionary variant of the default generator
* Remove redundant mask operations on the argument
  to splitmix64_next/1

#### Implement, document and measure niche PRNGs
* splitmix64_next
* exsp_next & exsp_jump
* mcg35
* lcg35

### DIFF
--- a/lib/stdlib/src/rand.erl
+++ b/lib/stdlib/src/rand.erl
@@ -832,7 +832,7 @@ exsss_seed({A1, A2, A3}) ->
 -define(
    exs_next(S0, S1, S1_b),
    begin
-       S1_b = S1 bxor ?BSL(58, S1, 24),
+       S1_b = ?MASK(58, S1) bxor ?BSL(58, S1, 24),
        S1_b bxor S0 bxor (S1_b bsr 11) bxor (S0 bsr 41)
    end).
 
@@ -842,9 +842,12 @@ exsss_seed({A1, A2, A3}) ->
        %% The multiply by add shifted trick avoids creating bignums
        %% which improves performance significantly
        %%
-       V_a = ?MASK(58, S + ?BSL(58, S, 2)), % V_a = S * 5
-       V_b = ?ROTL(58, V_a, 7),
-       ?MASK(58, V_b + ?BSL(58, V_b, 3)) % V_b * 9
+       %% Scramble ** (all operations modulo word size)
+       %% ((S * 5) rotl 7) * 9
+       %%
+       V_a = S + ?BSL(58, S, 2),                             % * 5
+       V_b = ?BSL(58, V_a, 7) bor ?MASK(7, V_a bsr (58-7)),  % rotl 7
+       ?MASK(58, V_b + ?BSL(58, V_b, 3))                     % * 9
    end).
 
 
@@ -855,16 +858,20 @@ exsss_seed({A1, A2, A3}) ->
                        {X :: uint58(), NewAlgState :: exsplus_state()}.
 exsp_next([S1|S0]) ->
     %% Note: members s0 and s1 are swapped here
-    NewS1 = ?exs_next(S0, S1, S1_1),
-    {?MASK(58, S0 + NewS1), [S0|NewS1]}.
+    S0_1 = ?MASK(58, S0),
+    NewS1 = ?exs_next(S0_1, S1, S1_b),
+    %% Scramble + (all operations modulo word size)
+    %% S0 + NewS1
+    {?MASK(58, S0_1 + NewS1), [S0_1|NewS1]}.
 
 -dialyzer({no_improper_lists, exsss_next/1}).
 
 -spec exsss_next(exsplus_state()) -> {uint58(), exsplus_state()}.
 exsss_next([S1|S0]) ->
     %% Note: members s0 and s1 are swapped here
-    NewS1 = ?exs_next(S0, S1, S1_1),
-    {?scramble_starstar(S0, V_0, V_1), [S0|NewS1]}.
+    S0_1 = ?MASK(58, S0),
+    NewS1 = ?exs_next(S0_1, S1, S1_b),
+    {?scramble_starstar(S0_1, V_1, V_2), [S0_1|NewS1]}.
 
 exsp_uniform({AlgHandler, R0}) ->
     {I, R1} = exsp_next(R0),
@@ -1168,13 +1175,6 @@ exro928ss_next({[S15,S0|Ss], Rs}) ->
     %% const uint64_t result_starstar = rotl(s0 * S, R) * T;
     %%
     {?scramble_starstar(S0, V_0, V_1), SR};
-%%    %% The multiply by add shifted trick avoids creating bignums
-%%    %% which improves performance significantly
-%%    %%
-%%    V0 = ?MASK(58, S0 + ?BSL(58, S0, 2)), % V0 = S0 * 5
-%%    V1 = ?ROTL(58, V0, 7),
-%%    V = ?MASK(58, V1 + ?BSL(58, V1, 3)), % V = V1 * 9
-%%    {V, SR};
 exro928ss_next({[S15], Rs}) ->
     exro928ss_next({[S15|lists:reverse(Rs)], []}).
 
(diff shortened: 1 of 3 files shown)


okeuday:
@RaimoNiskanen Though it may be dirty, you may want to compare to:

    (os:system_time(microsecond) rem N) + 1

when N is in the range (1..1000000).  I was hoping for something that is quicker than that.

lib/stdlib/src/rand.erl
```diff
@@ -1412,6 +1424,83 @@ dummy_seed({A1, A2, A3}) ->
...
+%% =====================================================================
+-define(LCG35_A, (15319397)).
+-define(LCG35_C, (15366142135)). % (1 bsl 35) / sqrt(5), odd
+-define(LCG35_B, 35). % Number of bits
+
+-type lcg35_state() :: 0..?MASK(?LCG35_B).
+
+-spec lcg35(X0 :: lcg35_state()) -> X1 :: lcg35_state().
+lcg35(X0) ->
+    ?MASK(?LCG35_B, ?LCG35_A * X0 + ?LCG35_C).
```
bjorng, line 1501: Here is an idea for possible improvement:

```suggestion
    ?MASK(?LCG35_B, ?LCG35_A * ?MASK(?LCG35_B, X0) + ?LCG35_C).
```

With this change, an unnecessary masking operation will be performed, but because of it, the type tests, overflow tests, and fallback code for bignum arithmetic will be omitted for the remaining three operations, resulting in smaller code with only a single branch instruction.

I don't know whether this will be faster. Perhaps you could re-run your benchmarks and find out.

If you want to inspect the native code, start the runtime system like this:

```
erl +JDdump true
```

The native code will be dumped to the `rand.asm` file.

bjorng (approved):
Looks good to me.

lib/stdlib/src/rand.erl
RaimoNiskanen: I tried and the benchmark maybe shows some improvement, it is hard to measure in the 2..4 nanoseconds range, but the generated native code gets significantly shorter, as you anticipated.  Using the same trick on the `mcg35` generator seems to give an even better improvement probably since there is a bit more code after the masking.

I also tried a better looking change i.e to add a guard `when is_integer(X0), 0 =< X0, X0 =< ?MASK(?LCG35_B)`, but apparently the JIT did not use that information to optimize the function body :-( Furthermore the 3-part range test itself needs much more code than a `band` operation which was a setback...

I dream of an integer_range guard test in the VM that the compiler could use in these cases, and that such a guard would be used by the JIT for function body optimization!

For both generators such a guard on the input would be user friendly anyway, especially for the `mcg35` generator where the allowed range is 1 =< X < 2^35 - 31 where masking with 2^35 - 1 is goofy.

bjorng: > I also tried a better looking change i.e to add a guard when `is_integer(X0), 0 =< X0, X0 =< ?MASK(?LCG35_B)`, but apparently the JIT did not use that information to optimize the function body :-( Furthermore the 3-part range test itself needs much more code than a `band` operation which was a setback...

No promises, but we hope to address both of those issues in OTP 26.

RaimoNiskanen:
@okeuday wrote:
> ``` erlang
> (os:system_time(microsecond) rem N) + 1
> ```
> when N is in the range (1..1000000). I was hoping for something that is quicker than that.

I added that to the benchmark bouquet, and it is a bit slower than the trick `erlang:phash2(erlang:unique_integer(), N)`, which both are about a factor two slower than `rand:lcg35(X0) rem N`.

I would advice against using `os:system_time/1` for this.  Getting the system time is unexpectedly awkward on many systems, especially virtualized, with possibly bottleneck all the way up in the hypervisor. `erlang:unique_integer/0,1` was created because what you actually often want is not the time, but a unique integer.

In this benchmark I would guess that `os:system_time(microsecond)` would return the same value about 50 times in a row, and changing to `nanosecond` may not give you actual nanoseconds one can not know how close in time two calls will give the same value...  But `erlang:unique_integer/0,1` will give you a different integer for every call, and `erlang:phash2/2` on that will give you statistical distribution provided that the range N is significantly lower than 2^32.  And `rand:lcg35/1` or `rand:mcg35/1` with `rem N` will give you both in half the time.

RaimoNiskanen:
Here is a typical `rand_SUITE:measure/1` result
```
RNG uniform integer full range performance

                   exsss:     53.3 ns (warm-up) [16#400000000000000]
                overhead:      3.4 ns      6.4% [16#1]
                   exsss:     51.1 ns    100.0% [16#400000000000000]
                   exrop:     49.5 ns     96.8% [16#400000000000000]
                    exsp:     46.6 ns     91.3% [16#400000000000000]
                exs1024s:    385.9 ns    755.4% [16#10000000000000000]
                   exs64:    356.6 ns    698.1% [16#10000000000000000]
                 exsplus:     72.4 ns    141.7% [16#400000000000000]
                 exs1024:    395.0 ns    773.3% [16#10000000000000000]
               exro928ss:     73.4 ns    143.7% [16#400000000000000]
                crypto64:    194.2 ns    380.1% [16#10000000000000000]
            crypto_cache:    101.0 ns    197.6% [16#100000000000000]
              crypto_aes:    167.9 ns    328.7% [16#400000000000000]
                  crypto:   2047.5 ns   4008.1% [16#10000000000000000]
                   dummy:     28.2 ns     55.3% [16#400000000000000]
            mcg35_inline:      6.1 ns     12.0% [16#7ffffffe0]
            lcg35_inline:      3.4 ns      6.7% [16#800000000]
             exsp_inline:     17.7 ns     34.7% [16#400000000000000]
           unique_phash2:     20.6 ns     40.4% [16#8000000]
                procdict:     73.0 ns    143.0% [16#400000000000000]
          lcg35_procdict:     16.0 ns     31.3% [16#800000000]
```

Selected results on a range:
```
RNG uniform integer range 10000 performance

                   exsss:     61.5 ns    100.0% [16#2710]
                    exsp:     52.9 ns     86.0% [16#2710]
            mcg35_inline:     15.0 ns     24.3% [16#2710]
            lcg35_inline:     13.8 ns     22.4% [16#2710]
             exsp_inline:     27.9 ns     45.3% [16#2710]
           unique_phash2:     24.0 ns     39.0% [16#2710]
         system_time_rem:     37.8 ns     61.3% [16#2710]
```

A mask is much faster than a `rem`:
```
RNG uniform integer 32 bit performance

                   exsss:     63.2 ns    100.0% [16#100000000]
                    exsp:     52.6 ns     83.3% [16#100000000]
            lcg35_inline:      4.3 ns      6.8% [16#100000000]
             exsp_inline:     18.0 ns     28.5% [16#100000000]
           unique_phash2:     23.3 ns     36.9% [16#100000000]
         system_time_rem:     27.6 ns     43.7% [16#100000000]
```

RaimoNiskanen:
I just want to point out that these are simply the results presented by `rand_SUITE:measure/1` so they can be run by anyone that can run an OTP test suite. :-)

dgud (approved):

PR #5838, opened 2022-03-30
Add support for erlang:statistics(dist) counters

potatosalad:
This is one part of the [RFC: Erlang Dist Security Filtering Prototype](https://erlangforums.com/t/rfc-erlang-dist-security-filtering-prototype/1002) (prototype PR [here](https://github.com/potatosalad/otp/pull/1)) effort that adds support for `erlang:statistics(dist)` which returns a list of message "types" which each have a list of "actions" taken by dist for the node (right now, the only action is `accept`).

Example (two nodes in separate terminals):

```bash
erl -name foo@0.0.0.0
erl -name bar@0.0.0.0
```

On node `foo@0.0.0.0`:

```erl
(foo@0.0.0.0)1> erlang:statistics(dist).
[{alias_send,[{accept,0}]},
 {demonitor,[{accept,0}]},
 {exit,[{accept,0}]},
 {exit2,[{accept,0}]},
 {group_leader,[{accept,0}]},
 {link,[{accept,0}]},
 {monitor,[{accept,0}]},
 {monitor_exit,[{accept,0}]},
 {reg_send,[{accept,0}]},
 {send,[{accept,0}]},
 {spawn_reply,[{accept,0}]},
 {spawn_request,[{accept,0}]},
 {unlink,[{accept,0}]},
 {unlink_ack,[{accept,0}]}]
(foo@0.0.0.0)2> net_adm:ping('bar@0.0.0.0').
pong
(foo@0.0.0.0)3> erlang:statistics(dist).
[{alias_send,[{accept,0}]},
 {demonitor,[{accept,2}]},
 {exit,[{accept,0}]},
 {exit2,[{accept,0}]},
 {group_leader,[{accept,0}]},
 {link,[{accept,0}]},
 {monitor,[{accept,2}]},
 {monitor_exit,[{accept,0}]},
 {reg_send,[{accept,7}]},
 {send,[{accept,2}]},
 {spawn_reply,[{accept,0}]},
 {spawn_request,[{accept,0}]},
 {unlink,[{accept,0}]},
 {unlink_ack,[{accept,0}]}]
```

On node `bar@0.0.0.0`:

```erl
(bar@0.0.0.0)1> erlang:statistics(dist).
[{alias_send,[{accept,0}]},
 {demonitor,[{accept,2}]},
 {exit,[{accept,0}]},
 {exit2,[{accept,0}]},
 {group_leader,[{accept,0}]},
 {link,[{accept,0}]},
 {monitor,[{accept,2}]},
 {monitor_exit,[{accept,0}]},
 {reg_send,[{accept,6}]},
 {send,[{accept,3}]},
 {spawn_reply,[{accept,0}]},
 {spawn_request,[{accept,0}]},
 {unlink,[{accept,0}]},
 {unlink_ack,[{accept,0}]}]
```


potatosalad:
Rebased, CI appears to be passing now 🎉

rickard-green:
This kind of statistics could very well be useful, but I see some issues with its current state.

The statistics counters are shared between all connections. Incoming signals on different channels modify the same set of counters which will lead to cache-line ping-ponging which is something we do not want. If they were channel specific this would not happen. I also guess that channel specific statistics also would be more interesting since different nodes might behave differently.

If we have counters for incoming signals, I also think we should have it for outgoing signals.

It is good to make the API extendable supporting future changes, but the `accept` tuple part kind of implies that something perhaps might not be accepted. Currently there are no such case and we don't want to commit to that there might be such a case either. Instead of `accept` it could be named `count`. In the case that we would introduce rejection of signals we could introduce a `rejected` count but still let `count` reflect the total count.

I also would have preferred a map of maps instead of a 2-tuple list of 2-tuple lists since it is much easier to fetch wanted information from such a term. Example:
```
#{alias_send => #{count => 0},
  demonitor => #{count => 2},
  exit => #{count => 0}, 
  exit2 => #{count => 0},
  group_leader => #{count => 0},
  link => #{count => 0},
  monitor => #{count => 2},
  monitor_exit => #{count => 0},
  reg_send => #{count => 6},
  send => #{count => 3},
  spawn_reply => #{count => 0},
  spawn_request => #{count => 0},
  unlink => #{count => 0},
  unlink_ack => #{count => 0}}
```

Depending on how expensive the updating of the statistics gets, there might also be a need for it to be optional so that you can chose to enable/disable this statistics or parts of it.

potatosalad:
@rickard-green Thank you for reviewing everything thus far, I appreciate it!

> The statistics counters are shared between all connections. Incoming signals on different channels modify the same set of counters which will lead to cache-line ping-ponging which is something we do not want. If they were channel specific this would not happen. I also guess that channel specific statistics also would be more interesting since different nodes might behave differently.

I [originally had this implemented](https://github.com/potatosalad/otp/pull/1/commits/09514decd3738366599dcc17106a0cd5fe4f0e4f) where both channels and global statistics were kept.  One issue there was that channel-level statistics were attached to the individual `dist_entry` and would therefore be dropped on disconnect.  I was also concerned about adding too much additional memory per-channel.

Would it be better to record things per-channel and then aggregate statistics by iterating over all the channels on read?  If so, do you have any recommendations on how this ought to work such that the global aggregates don't drop when a node disconnects?

> If we have counters for incoming signals, I also think we should have it for outgoing signals.

I agree, what do you think about `recv` and `send`?  For example: `#{alias_send => #{recv => 0, send => 0}, ...}`

> Depending on how expensive the updating of the statistics gets, there might also be a need for it to be optional so that you can chose to enable/disable this statistics or parts of it.

I agree, would a new `system_flag` be the right mechanism to enable/disable dist statistics?

rickard-green:
> @rickard-green Thank you for reviewing everything thus far, I appreciate it!
> 
> > The statistics counters are shared between all connections. Incoming signals on different channels modify the same set of counters which will lead to cache-line ping-ponging which is something we do not want. If they were channel specific this would not happen. I also guess that channel specific statistics also would be more interesting since different nodes might behave differently.
> 
> I [originally had this implemented](https://github.com/potatosalad/otp/pull/1/commits/09514decd3738366599dcc17106a0cd5fe4f0e4f) where both channels and global statistics were kept. One issue there was that channel-level statistics were attached to the individual `dist_entry` and would therefore be dropped on disconnect. I was also concerned about adding too much additional memory per-channel.
> 
> Would it be better to record things per-channel and then aggregate statistics by iterating over all the channels on read? 

Yes

> If so, do you have any recommendations on how this ought to work such that the global aggregates don't drop when a node disconnects?
> 

Dist entries are garbage collected using reference counting. Each remote pid, port, and reference refers to a dist entry (via a node entry) as well as each connection to a remote node refers to a dist entry. The dist entry is kept for a while even when the reference count reach zero to avoid excessive creation/destruction of dist entries.

You can increase the reference count on a dist entry by calling `erts_ref_dist_entry()` and decrease the reference count by calling `erts_deref_dist_entry()`. If you increase this reference count when enabling these statistics counters and decrease it when disabling these statistics counters, the dist entries will remain available until you have disabled statistics counters.

> > If we have counters for incoming signals, I also think we should have it for outgoing signals.
> 
> I agree, what do you think about `recv` and `send`? For example: `#{alias_send => #{recv => 0, send => 0}, ...}`
> 

Sounds good

> > Depending on how expensive the updating of the statistics gets, there might also be a need for it to be optional so that you can chose to enable/disable this statistics or parts of it.
> 
> I agree, would a new `system_flag` be the right mechanism to enable/disable dist statistics?

Yes

All currently existing dist entries can be found in the double linked lists:
* `erts_visible_dist_entries` (normal visible connected nodes)
* `erts_hidden_dist_entries` (hidden connected nodes)
* `erts_pending_dist_entries` (ongoing connect that has not finished yet)
* `erts_not_connected_dist_entries` (not connected)

These lists are protected by the `erts_dist_table_rwmtx` read/write lock. If you read lock it, no new dist entries can be registered in these lists and no new dist entries can be created until you release the lock. There are also counters for the amount of dist entries in each list. See the `nodes_1()` BIF implementation in `dist.c` for more info.

Each dist entry is protected by its own read/write lock. Read lock/unlock of dist entry `erts_de_rlock()`/`erts_de_runlock()` and write lock/unlock of dist entry `erts_de_rwlock()`/`erts_de_rwunlock()`. When installing the counters you want to write lock the dist entry. However, you are *not* allowed to lock a dist entry while you have `erts_dist_table_rwmtx` locked. Doing so could potentially cause a deadlock. Therefore, in order to install the counters you want to do something like this:
* Mark in a global variable that counting is enabled.
* Read lock `erts_dist_table_rwmtx`
* Go through each dist entry of the lists
    * Increment reference count on it which will prevent it from being removed (you don't need to lock it to do this)
    * Save dist entry in a local structure of dist entries to handle
* Read unlock `erts_dist_table_rwmtx`
* Go through each dist entry that you saved in your local structure
    * Write lock dist entry
    * install statistics counters
    * Write unlock dist entry

You also need to update `dist_table_alloc()` in `erl_node_tables.c` so that it installs statistics counters if your global variable indicates that statistics counting is enabled.

max-au:
@rickard-green 
Iterating over large number of channels may get expensive (especially when locks have to be taken for every single channel). There was a discussion in #2935 about a similar thing (process table is usually larger, but at least there is no lock-per-process taken).

rickard-green:
> @rickard-green Iterating over large number of channels may get expensive (especially when locks have to be taken for every single channel). There was a discussion in #2935 about a similar thing (process table is usually larger, but at least there is no lock-per-process taken).

Yes, if you have a huge amount of dist entries it might get problematic. Preferably one would yield in between installing counters if there are a large amount of dist entries. However, it is not comparable to #2935 where you blocked thread progress during inspection of the whole process table, which blocks execution for all normal schedulers regardless of what they are doing. In this case only schedulers creating/destroying dist entries will be blocked while incrementing reference counters, and only schedulers accessing specific dist entries while installing statistics counters.

rickard-green:
Closing this. If requested changes are implemented please reopen this PR.

PR #5839, opened 2022-03-30
erts: Flush all signals on monitor_nodes down signal

jhogberg:


garazdawi:
Would you mind adding a note in the commit message about why this is needed and which test case fails when it does not work?

rickard-green:
The test cases I will write will most likely not test this very well.

I think we need to write a few more test cases regarding parallel send/receive and signal order in general. Such test cases aren't especially quickly written though, so I think we need schedule and prioritize this work. I'll write a ticket for this.

PR #5849, opened 2022-04-01
JIT: Optimize `bsl` and `bxor` with known small operands

bjorng:
This pull request optimizes the `bsl` and `bxor` instructions when they have known small operands, and also fix a bug where unsafe code could would be generated for `band`, `bor`, and `bxor` in the x86 JIT.

### DIFF
--- a/erts/emulator/beam/jit/arm/beam_asm.hpp
+++ b/erts/emulator/beam/jit/arm/beam_asm.hpp
@@ -1313,6 +1313,26 @@ protected:
         }
     }
 
+    bool is_bsl_small(const ArgSource &LHS, const ArgSource &RHS) {
+        /*
+         * In the code compiled by scripts/diffable, there never
+         * seems to be any range information for the RHS. Therefore,
+         * don't bother unless RHS is an immediate small.
+         */
+        if (!(always_small(LHS) && RHS.isSmall())) {
+            return false;
+        } else {
+            auto [min1, max1] = getIntRange(LHS);
+            auto rhs_val = RHS.as<ArgSmall>().getSigned();
+
+            if (min1 < 0 || max1 == 0 || rhs_val < 0) {
+                return false;
+            }
+
+            return rhs_val < Support::clz(max1) - _TAG_IMMED1_SIZE;
+        }
+    }
+
     /* Helpers */
     void emit_gc_test(const ArgWord &Stack,
                       const ArgWord &Heap,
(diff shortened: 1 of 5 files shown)


erts/emulator/beam/jit/arm/beam_asm.hpp
```diff
@@ -1313,6 +1313,21 @@ class BeamModuleAssembler : public BeamAssembler {
...
+            return false;
+        } else {
+            auto [min1, max1] = getIntRange(LHS);
+            auto [min2, max2] = getIntRange(RHS);
+
+            if (min1 < 0 || max1 == 0 || min2 < 0 || min2 != max2) {
+                return false;
+            }
+
+            return max2 < Support::clz(max1)-4;
```
jhogberg, line 1327: Nitpick:
```suggestion
            return max2 < Support::clz(max1) - _TAG_IMMED1_SIZE;
```

erts/emulator/beam/jit/arm/instr_arith.cpp
```diff
@@ -1206,11 +1218,21 @@ void BeamModuleAssembler::emit_i_bsl(const ArgLabel &Fail,
...
                                      const ArgSource &RHS,
                                      const ArgRegister &Dst) {
-    Label generic = a.newLabel(), next = a.newLabel();
     auto [lhs, rhs] = load_sources(LHS, ARG2, RHS, ARG3);
     auto dst = init_destination(Dst, ARG1);
 
+    if (is_bsl_small(LHS, RHS)) {
+        comment("skipped tests because operands and result are always small");
+        a.and_(TMP1, lhs.reg, imm(~_TAG_IMMED1_MASK));
+        a.lsl(TMP1, TMP1, imm(RHS.as<ArgSmall>().getSigned()));
```
jhogberg, line 1227: Isn't it possible for `RHS` to be a register here?

```suggestion
    if (RHS.isSmall()) {
        a.lsl(TMP1, TMP1, imm(RHS.as<ArgSmall>().getSigned()));
    } else {
        a.lsl(TMP1, TMP1, rhs.reg);
    }
```

erts/emulator/beam/jit/arm/beam_asm.hpp
```diff
@@ -1313,6 +1313,21 @@ class BeamModuleAssembler : public BeamAssembler {
...
     }
 
+    bool is_bsl_small(const ArgSource &LHS, const ArgSource &RHS) {
+        if (!(always_small(LHS) && always_small(RHS))) {
+            return false;
+        } else {
+            auto [min1, max1] = getIntRange(LHS);
+            auto [min2, max2] = getIntRange(RHS);
+
+            if (min1 < 0 || max1 == 0 || min2 < 0 || min2 != max2) {
```
jhogberg, line 1323: `min2 != max2` looks a bit odd, won't this reject non-constant `RHS`? Is there any reason it can't work for, say `min2 = 3, max2 = 4`?

bjorng:
I now explicitly don't attempt the optimization in `bsl` if RHS is a variable and I've added a comment explaining why.

PR #5850, opened 2022-04-01
erts: Shrink ErlFunThing structure

jhogberg:
The sum of `arity` and `num_free` cannot exceed 255, so there's no point using a whole word for each.

Note that we will now badarg when passing an arity above 255 to `erlang:make_fun/3`, instead of creating a fun that can never be called and had iffy behavior when sent over the distribution (the arity got truncated down to 8 bits), but it ought to be okay as its spec has always said the valid range is 0..255.

### DIFF
--- a/erts/emulator/beam/bif.c
+++ b/erts/emulator/beam/bif.c
@@ -4417,7 +4417,7 @@ BIF_RETTYPE make_fun_3(BIF_ALIST_3)
     }
 
     arity = signed_val(BIF_ARG_3);
-    if (arity < 0) {
+    if (arity < 0 || arity > MAX_ARG) {
         BIF_ERROR(BIF_P, BADARG);
     }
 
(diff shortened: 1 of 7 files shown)


bjorng (approved):

erts/emulator/beam/bif.c
```diff
@@ -4417,7 +4417,7 @@ BIF_RETTYPE make_fun_3(BIF_ALIST_3)
     }
 
     arity = signed_val(BIF_ARG_3);
-    if (arity < 0) {
+    if (arity < 0 || arity > MAX_ARG) {
```
mikpe, line 4420: You might get better code from `if ((Uint)arity > MAX_ARG)`

jhogberg: Thanks for pointing it out, but I think it's clearer this way and I trust the compiler to do that optimization on its own.

PR #5859, opened 2022-04-05
Cons/features/essential meta chunk

lisztspace:


lib/compiler/src/beam_asm.erl
```diff
@@ -199,10 +199,16 @@ build_file(Code, Attr, Dict, NumLabels, NumFuncs, ExtraChunks, CompileInfo, Comp
                       <<?BEAM_TYPES_VERSION:32, NumTypes:32>>,
                       TypeTab),
 
+    %% Create the meta chunk
+    Meta = proplists:get_value(<<"Meta">>, ExtraChunks, term_to_binary([])),
+    MetaChunk = chunk(<<"Meta">>, Meta),
+    %% Remove Meta chunk from ExtraChunks since it is essential
+    ExtraChunks1 = ExtraChunks -- [{<<"Meta">>, Meta}],
```
bjorng, line 206: We use @rvirding's convention for numbering updated variables in the `compiler` application. Rename `ExtraChunks` in the function header for `build_files` to `ExtraChunks0`. Then:

```suggestion
    Meta = proplists:get_value(<<"Meta">>, ExtraChunks0, term_to_binary([])),
    MetaChunk = chunk(<<"Meta">>, Meta),
    %% Remove Meta chunk from ExtraChunks since it is essential
    ExtraChunks = ExtraChunks0 -- [{<<"Meta">>, Meta}],
```

lib/compiler/src/beam_asm.erl
```diff
@@ -212,7 +218,7 @@ build_file(Code, Attr, Dict, NumLabels, NumFuncs, ExtraChunks, CompileInfo, Comp
 
     %% Compile all extra chunks.
 
-    CheckedChunks = [chunk(Key, Value) || {Key, Value} <- ExtraChunks],
+    CheckedChunks = [chunk(Key, Value) || {Key, Value} <- ExtraChunks1],
```
bjorng, line 221:
```suggestion
    CheckedChunks = [chunk(Key, Value) || {Key, Value} <- ExtraChunks],
```

bjorng (changes requested):
Looks good to me, except that we have a special convention for numbering updated variables in the `compiler` application.

lib/compiler/src/beam_asm.erl
lisztspace: Neat, I'll fix it together with some comments from @jhogberg

bjorng (approved):

PR #5861, opened 2022-04-05
erts: Fix child_setup partial read

garazdawi:
When calling read it may return a partial result. So we make
sure that all the data is read before returning.

The problem has been observed when reading the ACK message,
but we use the new routine for all reads in the child so that
we know that it will work everywhere.

### DIFF
--- a/erts/emulator/sys/unix/erl_child_setup.c
+++ b/erts/emulator/sys/unix/erl_child_setup.c
@@ -136,6 +136,43 @@ void sys_sigrelease(int sig)
     sigprocmask(SIG_UNBLOCK, &mask, (sigset_t *)NULL);
 }
 
+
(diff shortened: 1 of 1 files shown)


sverker (commented):
In `main` there is a `read(sigchld_pipe[0], ibuff, sizeof(ibuff));`.
Should that also use `read_all`?

What about a `write_all`? write can also return partial results according to docs.

garazdawi:
> In main there is a read(sigchld_pipe[0], ibuff, sizeof(ibuff));.
> Should that also use read_all?

Yes, might as well change that one also.

> What about a `write_all`? write can also return partial results according to docs.

I've never seen a write do a partial write in practice... though I suppose it does not hurt to be extra safe.

garazdawi:
Fixed, please review again @sverker.

PR #5863, opened 2022-04-06
dialyzer: Improve deterministic builds

TD5:
Improves build determinism for non-native code by:
- Making codegen respect the +deterministic option
- Adding an ERL_DETERMINISTIC env var that controls OTP's build determinism

This addresses absolute paths being included in generated `.erl` files and compiled `.beam` files that resulted in builds from different source directories generating different artefacts (which is a component of the issue in https://github.com/erlang/otp/issues/4482, but is also useful in its own right, for example around caching builds and for the upcoming Dialyzer incremental mode).

OTP itself can now be built in deterministic mode by setting `ERL_DETERMINISTIC=yes` in the environment. I think it would make sense to make this the default at some stage, but I've put the change behind an option for now in an attempt to decouple making deterministic OTP build possible from making them the default.

Having `+deterministic` set also results in the compiler options being removed from the module info for modules where this options was used. This may have other implications for users of OTP.

For tests themselves, `+determinism` is disabled, since many test cases depend on accessing the test module's compilation options or other features not available in deterministic mode, in order to configure themselves. For tests of the determinism feature, `+determinism` is explicitly passed to the compiler within the relevant test cases.

> N.B. the bootstrap/preloaded `.beam` files need to be updated in order for this feature to work properly


TD5:
If there is particular enthusiasm for this change, we could also jump straight to making `+deterministic` the default for OTP, allowing us to remove a lot of the migratory entries in the Makefiles

garazdawi:
Could you please split the changes into separate commits so that it is easier to review? i.e. one commit for the makefile changes, one commit for the asn1 changes etc etc.

Also, in my opinion, this should be controlled through a configure option, or is there advantage I have missed why you have chosen to use an environment variable?

TD5:
Yep, I can split those bits up into separate commits.

The environment variable was just because I didn't really know the idiomatic way to do it in OTP, and I am definitely a general autoconf/makefile newbie. Still, I think I can move things around to put this behind a configure option instead 👍

lib/stdlib/src/epp.erl
```diff
@@ -781,14 +785,18 @@ enter_file(NewName, Inc, From, St) ->
...
          default_encoding = DefEncoding,
          in_prefix = InPrefix,
          erl_scan_opts = ScanOpts,
          else_reserved = ElseReserved,
          features = Ftrs} = St0,
-    Ms = Ms0#{'FILE':={none,[{string,Anno,Pname}]}},
+    SourceName = case St0#epp.deterministic of
+                     true -> filename:basename(Pname);
+                     false -> Pname
+                 end,
```
jhogberg, line 798: This gets repeated a lot, how about breaking it out?
```suggestion
    SourceName = source_name(St0),
```

jhogberg (changes requested):
Thanks for the PR!

I'd like to see the application changes broken out and the options documented. Once that's in place we can do the build changes ourselves, but if you're comfortable doing it we'd like it to be a configure flag (e.g. `--deterministic-build=yes|no`) that turns on the required options when needed.

As an aside, can you change the branch name from `master` to something more descriptive? "Merged master into master" will look rather funny in the history.

TD5:
Closed so I can put up a new PR with a better branch name. I'll apply the other suggested changes at the same time.

PR #5865, opened 2022-04-06
otp: Refactor esock build

jhogberg:
This PR makes the `socket` and `net` modules use NIF fallbacks when OTP is compiled without esock support, avoiding the need for `-ifdef`s and helping us simplify the build.


bmk:
One of the build checks (Emulator) fails.

bmk (approved):
The was a problem with one of the checks (Emulator).
But in principle I approve.

PR #5866, opened 2022-04-06
Add OFB mode support for AES

ausimian:
Add Output Feedback mode support for AES encryption / decryption for key
sizes of 128, 192 and 256 bits.


HansN:
There were no problems in our nightly tests, but the handling of the PR must wait till after the release of the release candidate 25.0-rc3.

HansN:
Thanks for the PR!

PR #5868, opened 2022-04-07
ssl: Handle TLS-1.3 defined legacy signature names in TLS-1.2

### LINKED ISSUE #5835: TLS Certificate Unknown in OTP 24.3
**Describe the bug**
After updating to OTP 24.3.2, we cannot establish an Elixir - HTTPoison connection to a server that requires a client certificate, instead we receive 

```elixir
{:error,
 %HTTPoison.Error{
   id: nil,
   reason: {:tls_alert,
    {:certificate_unknown,
     'TLS client: In state cipher received SERVER ALERT: Fatal - Certificate Unknown\n'}}
 }}
```

**To Reproduce**
Unfortunately this issues involves connecting to a private server of a partner that we cannot give access to; and we also do not have any access to the server in order to debug the requests. Please let me know about any debugging possibilities I should do to provide more information (see my attempts below)

**Expected behavior**
An established connection to the server.

**Affected versions**
(24.3,) 24.3.1 and 24.3.2, but not in 24.2.2

**Additional context**
When looking for an opportunity to debug the requests, I stumbled across `https://prod.idrix.eu/secure/`. I tested sending the requests against this server:

* OTP 24.2.2 worked as expected
* 24.3 returned `{:error, %HTTPoison.Error{id: nil, reason: :checkout_failure}}`
* 24.3.1 and 24.3.2 worked again (contrary to the server we are connecting to)

I then compared the outputs of requests of 24.2.2 and 24.3.2, and there are some whitespace related changes; however I do not know if is related in any way to the issue:

```diff
- [SSL_SECURE_RENEG] => true
+ [SSL_SECURE_RENEG] => true 

- [REQUEST_URI] => /secure/ 
+ [REQUEST_URI] => /secure/
```

Other differences are in `[SSL_SESSION_ID]`, `[REMOTE_PORT]`, `[REQUEST_TIME_FLOAT]` and `[REQUEST_TIME]`.

IngelaAndin:
Seems we need to handle some additional algorithm names in TLS-1.2 

Closes #5835

### DIFF
--- a/lib/ssl/src/ssl_cipher.erl
+++ b/lib/ssl/src/ssl_cipher.erl
@@ -651,6 +651,12 @@ signature_schemes_1_2(SigAlgs) ->
                                 [{Hash, Sign} | Acc];
                             {Hash, Sign = rsa_pss_rsae,_} ->
                                 [{Hash, Sign} | Acc];
+                            {sha1, rsa_pkcs1,_} ->
+                                [{sha, rsa} | Acc];
+                            {Hash, rsa_pkcs1,_} ->
+                                [{Hash, rsa} | Acc];
+                            {sha1, Sign = ecdsa,_} ->
+                                [{sha, Sign} | Acc];
                             {_, _, _} ->
                                 Acc
                         end;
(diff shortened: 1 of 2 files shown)


KennethL:
I added a commit with a correction so please try the new version of the PR

IngelaAndin:
Hi, now back from the Easter holiday, I think we are making progress. Now the correct algorithm is selected, however, we still need to make sure it is represented as a tuple and not an atom for the generic TLS-1.2 code. Please try this latest addition.

IngelaAndin:
Close in favor of #5919

PR #5879, opened 2022-04-12
Fix regex generic character type description

rustkas:
Replace "Delete" to "Digit".
Regex  character type `\d` means any decimal digit.


rustkas:
> Hello!
> 
> Have you tested that this is the case? In an Erlang string "\d" is the value 127, which is ascii delete and I think that list describes escape sequences allowed that should be the same as in Erlang.

You are right. 

`\d` do not mean a digit. I made a hasty conclusion.

```
1> d_lexer:string("0").
{error,{1,d_lexer,{illegal,"0"}},1}
```

garazdawi:
If you have any suggestions on how to make the documentation clearer so that others do not fall into the same trap a PR would be most welcome!

PR #5884, opened 2022-04-12
Change from using dicts in typer_core to using maps

maco:
This is (I think?) the last of the refactoring work we (@elbrujohalcon and @pablocostass) did on typer core in the course of making
the rebar3_typer plugin.

### DIFF
--- a/lib/dialyzer/src/typer_core.erl
+++ b/lib/dialyzer/src/typer_core.erl
@@ -17,9 +17,10 @@
 %% Author(s)   : The first version of typer was written by Bingwen He
 %%               with guidance from Kostis Sagonas and Tobias Lindahl.
 %%               Since June 2008 typer is maintained by Kostis Sagonas.
-%%               On 2022, Brujo, Pablo, and Mackenzie from NextRoll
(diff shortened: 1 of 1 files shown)


lib/dialyzer/src/typer_core.erl
```diff
@@ -384,20 +393,18 @@ clean_inc(Inc) ->
   normalize_obj(Inc1).
 
 remove_yecc_generated_file(#inc{filter = Filter} = Inc) ->
-  Fun = fun (Key, #inc{map = Map} = I) ->
-            I#inc{map = map__remove(Key, Map)}
-        end,
+  Fun = Fun = fun(Key, #inc{map = Map} = I) -> I#inc{map = maps:remove(Key, Map)} end,
```
elbrujohalcon, line 396:
```suggestion
  Fun = fun(Key, #inc{map = Map} = I) -> I#inc{map = maps:remove(Key, Map)} end,
```
We had too much fun here 🤪

maco:
squashed!

jhogberg:
Merged, thanks again for the PR! :slightly_smiling_face:

Issue #5732, opened 2022-02-20
Failed to build with Gitpod.


zeek0x:
**Describe the bug**

If I access gitpod from the gitpod link (https://gitpod.io/#https://github.com/erlang/otp) in otp/HOWTO/DEVELOPMENT.md, the build will start and fail.

**To Reproduce**

Open https://gitpod.io/#https://github.com/erlang/otp

**Expected behavior**

Successfully build an Erlang/OTP environment within Gitpod.

**Affected versions**

master branch version. (or PR based on master:  https://gitpod.io/#https://github.com/erlang/otp/pulls/xxx )

**Additional context**

I have tried to fix this problem. 

The first problem is running apt-get as `gitpod` user (≠ root).

https://github.com/erlang/otp/blob/9f87c568cd3cdb621cf4cae69ccce880be4ea1b6/.github/dockerfiles/Dockerfile.ubuntu-base#L14

```console
#5 1.238 E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
#5 1.238 E: Unable to lock directory /var/lib/apt/lists/
#5 ERROR: process "/bin/sh -c apt-get update && apt-get -y upgrade &&
...
```

I actually run it locally...

```console
$ docker run --rm gitpod/workspace-full whoami
gitpod
```

So I added `USER root` before `DEBIAN_FRONTEND=noninteractive`  (L10 of the following code). 

https://github.com/erlang/otp/blob/9f87c568cd3cdb621cf4cae69ccce880be4ea1b6/.github/dockerfiles/Dockerfile.ubuntu-base#L9-L14

Then I found a problem with the `apache2` installation. 

```console
Configuration file '/etc/apache2/apache2.conf'
 ==> File on system created by you or by a script.
 ==> File also in package provided by package maintainer.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** apache2.conf (Y/I/N/O/D/Z) [default=N] ?
```

Even though `DEBIAN_FRONTEND=noninteractive` is set, the installer still ask how to install.

I tried to connect the yes command to the apt-get pipeline, but ran into another problem.

```console
Setting up apache2 (2.4.41-4ubuntu3.9) ...

Configuration file '/etc/apache2/apache2.conf'
 ==> File on system created by you or by a script.
[...]

PR #5886, opened 2022-04-13
devcontainer: Fix depdendencies for otp_build configure

maennchen:
Add pkg-config to the apt dependencies so that
`./otp_build configure` runs successfully.

Fixes first part of https://github.com/erlang/otp/issues/5732#issuecomment-1097549304


.github/dockerfiles/Dockerfile.ubuntu-base
```diff
@@ -4,7 +4,7 @@
 ARG BASE=gitpod/workspace-full
 FROM $BASE
 
-ENV INSTALL_LIBS="zlib1g-dev libncurses5-dev libssl-dev unixodbc-dev libsctp-dev lksctp-tools libgmp3-dev libwxbase3.0-dev libwxgtk3.0-gtk3-dev libwxgtk-webview3.0-gtk3-dev"
```
garazdawi, line 7: Can you add the pkg-config to the list of tools on line 18 instead? The INSTALL_LIBS are meant to be the libraries that are needed at build and run-time.

maennchen: @garazdawi Done.

PR #5890, opened 2022-04-14
Clarify return order of filelib:wildcard/1|2

kjnilsson:
filelib:wildcard sorts the files using `lists:sort/1` before returning them to the caller.

### DIFF
--- a/lib/stdlib/doc/src/filelib.xml
+++ b/lib/stdlib/doc/src/filelib.xml
@@ -183,7 +183,7 @@
       <name name="wildcard" arity="1" since=""/>
       <fsummary>Match filenames using Unix-style wildcards.</fsummary>
       <desc>
-        <p>Returns a list of all files that match Unix-style wildcard string
+        <p>Returns a sorted list of all files that match Unix-style wildcard string
           <c><anno>Wildcard</anno></c>.</p>
         <p>The wildcard string looks like an ordinary filename, except
           that the following "wildcard characters" are interpreted in a special


kjnilsson:
By leaving it to the user do you mean removing the (expensive) `lists:sort/1` call before returning the files so that users don't have to pay the sorting performance hit twice (which is how I found it in the first place).

KennethL:
I think we should deliver in an undefined (or not guaranteed) order, which probably is the order delivered from the OS file system.
Why do we have the sorting? (Question to us the OTP team)

jhogberg:
> By leaving it to the user do you mean removing the (expensive) lists:sort/1 call before returning the files so that users don't have to pay the sorting performance hit twice (which is how I found it in the first place).

Either leaving it undocumented or removing it. I'm leaning towards the latter because I can't see any reason why it should be sorted, other than the output "looking nice" at a glance.

> Why do we have the sorting? (Question to us the OTP team)

From the commit logs it looks like historic baggage, there's nothing explaining the reasoning behind it.

kjnilsson:
Removing the sorting feels like the correct call, that said there may be code out there that relies on it being sorted.

jhogberg:
> Removing the sorting feels like the correct call, that said there may be code out there that relies on it being sorted.

There's only one way to find out :smile:

It's a bit late to do this in OTP 25 since the last release candidate is already out, but I think we ought to do it in OTP 26. Feel free to open a PR for that.

PR #5894, opened 2022-04-15
Ets take firts last

nyczol:
Add ets:take_last/1 and ets:take_first/1 func similar to ets:take/2, but return firts/last objects from ets.


nyczol:
Will be reopened due to CLA issue

PR #5896, opened 2022-04-16
add ets take_first/1 and take_last/1 functions

nyczol:
Add ets:take_last/1 and ets:take_first/1 func similar to ets:take/2, but return firts/last objects from ets.

### DIFF
--- a/erts/emulator/beam/erl_db.c
+++ b/erts/emulator/beam/erl_db.c
@@ -1190,6 +1190,58 @@ BIF_RETTYPE ets_take_2(BIF_ALIST_2)
     BIF_RET(ret);
 }
 
+/*
+** take_first(Tab)
+*/
+BIF_RETTYPE ets_take_first_1(BIF_ALIST_1)
+{
+    DbTable* tb;
+    int cret;
+    Eterm first,ret;
+    CHECK_TABLES();
+
+    DB_BIF_GET_TABLE(tb, DB_WRITE, LCK_WRITE_REC, BIF_ets_take_first_1);
+
+    cret = tb->common.meth->db_first(BIF_P, tb, &first);
+
+    if (cret != DB_ERROR_NONE) {
+        db_unlock(tb, LCK_WRITE_REC);
+        BIF_ERROR(BIF_P, BADARG);
+    }
+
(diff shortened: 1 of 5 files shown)


erts/emulator/beam/erl_db.c
```diff
@@ -1190,6 +1190,58 @@ BIF_RETTYPE ets_take_2(BIF_ALIST_2)
...
+    DB_BIF_GET_TABLE(tb, DB_WRITE, LCK_WRITE_REC, BIF_ets_take_first_1);
+
+    cret = tb->common.meth->db_first(BIF_P, tb, &first);
+
+    if (cret != DB_ERROR_NONE) {
+        db_unlock(tb, LCK_WRITE_REC);
+        BIF_ERROR(BIF_P, BADARG);
+    }
+
+    cret = tb->common.meth->db_take(BIF_P, tb, first, &ret);
```
sverker, line 1212: No decision has been made yet to accept this API addition, but I just noticed that

you don't treat `'$end_of_table'` (`first==am_EOT`). It just happens to work as long as there is no such key.

sverker: I realize now that if `first` returns am_EOT then no keys at all exist. But it could actually be a race where someone else inserts key `'$end_of_table'` and it would make the code more clear to treat `am_EOT`.

nyczol: Please note that '$end_of_table' atom can't be used as a key (https://www.erlang.org/doc/man/ets.html).
Functions take_first/1 and take_last/1 can be used to build very efficient FIFO queues using ets.

sverker:
I don't think we want to rush this into OTP 25. Especially as I discovered this implementation is not correct for table option `write_concurrency`. Two concurrent calls may get the same key from `db_first` and then race each other calling `db_take` where only one can succeed.

nyczol:
I've just fix concurrency issue.
Now table is locked in LCK_WRITE except LCK_WRITE_REC.

sverker:
It was fast and now it's correct. I think we should aim for both.

Keep LCK_WRITE_REC and implement new meth->take_first() functions for hash, tree and catree where we seize the lock once and release it when done.

PR #5897, opened 2022-04-16
Allow to change process name via env using ESCRIPT_NAME

colrack:
This patch allows us to change the name of the process after setting the env var `ESCRIPT_NAME`

`ESCRIPT_NAME` is already used in `erlexec` (see https://github.com/erlang/otp/blob/OTP-24.3/erts/etc/common/erlexec.c#L544 ) to change cmdline, setting argv[0] = `ESCRIPT_NAME`;
this is nice because you can easily identify your process when using standard unix tooling like `procps`.
For example, what you achieve with `ps auwx` is from something like this
```
root        8114  0.4  5.5 2650396 206664 pts/0  Ssl+ Apr12  23:56 /app/erts-12.3.1/bin/beam.smp -- -root /app -progname erl -- -home /root -- -start_epmd false -kernel inet_dist_use_interface {127,0,0,1} ...
```
to something like this
```
root       65386  0.5  5.4 2653464 205940 pts/0  Ssl+ Apr15   8:02 myapp -- -root /app -progname erl -- -home /root -- -start_epmd false -kernel inet_dist_use_interface {127,0,0,1} ...
```
However, the process name remains unchanged and in `Linux` tools like `iproute2` rely on `/proc` to grab the name of the process. 
```
~# cat /proc/71803/comm
beam.smp
~# cat /proc/71803/stat
71803 (beam.smp) S 71791 71803 71803 34816 71803 4210944 117512 1462 8460 0 35869 10731 1 0 20 0 24 0 26552034 2717884416 47557 18446744073709551615 94849568030720 94849571035741 140727584747344 0 0 0 0 69760 16902 0 0 0 17 2 0 0 0 0 0 94849572012368 94849572123400 94849572397056 140727584754120 140727584754519 140727584754519 140727584755674 0
```
This ends up having a lot of processes with the name `beam` and it is not easy to identify who is who, especially if you have a lot of beam processes in your host system.
`ps`
```
~# ps
    PID TTY          TIME CMD
  65386 pts/0    00:08:11 beam.smp
  71140 pts/0    00:05:49 beam.smp
  71467 pts/0    00:07:08 beam.smp
  71803 pts/0    00:07:43 beam.smp
  89285 pts/0    00:00:02 beam.smp
  89448 pts/0    00:00:00 sh
  89521 pts/0    00:00:00 ps
 ```
 `ss` from `iproute2`
```
~# ss -lupn
State             Recv-Q            Send-Q                       Local Address:Port                         Peer Address:Port            Process
UNCONN            0                 0                            172.17.177.95:2152                              0.0.0.0:*                users:(("beam.smp",pid=8114,fd=22))
UNCONN            0                 0                            172.17.177.95:8805                              0.0.0.0:*                users:(("beam.smp",pid=8114,fd=24))
```
With the patch, as long as you set the env var, the result is something like this:
```
~# ss -lupn
State             Recv-Q            Send-Q                       Local Address:Port                         Peer Address:Port            Process
UNCONN            0                 0                            172.17.177.95:2152                              0.0.0.0:*                users:(("myapp",pid=89285,fd=48))
UNCONN            0                 0                            172.17.177.95:8805                              0.0.0.0:*                users:(("myapp",pid=88285,fd=49))
```
I'm not sure about the place in the code where I put the patch, but it seems to me fine. 
To change process name, I'm using `ethr_setname` defined in `ethread.h` which is already used to name threads (using `pthread_setname_np`) and it should be portable across systems.

Any thought on this?


dch:
@rickard-green this seems a straightforward change, do you think it could do into OTP27 branch?

rickard-green:
Yes in a modified state.

The `ESCRIPT_NAME` environment variable is an internal variable not very well named for general use since it does not only affect `escripts`.

Another question is whether or not it should be an environment variable at all. We have since quite a long time ago rejected use of environment variables in favor of command line arguments, so it should probably be a command line argument (a + emulator flag)  instead. If one needs to pass information in an environment variable, it is always possible to do that by passing command line arguments in the general `ERL_FLAGS` environment variable.

The functionality should be documented.

A test case should also be present. It does not need to be very complicated.

Sorry @colrack for the long time to response.

colrack:
@rickard-green no worries, my initial aim was to retrieve some feedback.
The implementation was just few lines of code and helped me to describe the feature, and it was meant to be less pervasive as possible reusing existing stuff.
Could you please discuss this internally and decide about the implementation?
I would be happy to add a test case and docs; even happier if you takeover this :)

Issue #5074, opened 2021-07-27
httpc:request post hangs if a header is in binary format


nixxquality:
**Describe the bug**
The documentation states that only string() can be can be the value() in a header() but iolists work, so I thought binary would as well.
When I was building a reproducible piece of code I noticed that get requests (with the simpler request() form) aren't subject to this error.

The process hangs even if a timeout value is provided.

**To Reproduce**
```
> inets:start().
> ssl:start().
> httpc:request(get, {"https://erlang.org/", [{"header", <<"binary">>}]}, [], []).
{error,{headers_error,not_strings}}
> httpc:request(post, {"https://erlang.org/", [{"header", <<"binary">>}], "text/plain", ""}, [], []).
=WARNING REPORT==== 27-Jul-2021::11:50:56.346827 ===
Description: "Authenticity is not established by certificate path validation"
     Reason: "Option {verify, verify_peer} and cacertfile/cacerts is missing"

=CRASH REPORT==== 27-Jul-2021::11:50:56.603777 ===
  crasher:
    initial call: httpc_handler:init/1
    pid: <0.129.0>
    registered_name: []
    exception error: bad argument
      in operator  ++/2
         called as <<"binary">> ++ "\r\n"
      in call from http_request:headers_other/2 (http_request.erl, line 306)
      in call from http_request:http_headers/1 (http_request.erl, line 82)
      in call from httpc_request:send/4 (httpc_request.erl, line 103)
      in call from httpc_handler:connect_and_send_first_request/3 (httpc_handler.erl, line 831)
      in call from httpc_handler:init/1 (httpc_handler.erl, line 236)
    ancestors: [httpc_handler_sup,httpc_sup,inets_sup,<0.84.0>]
    message_queue_len: 0
    messages: []
    links: [<0.89.0>]
    dictionary: []
    trap_exit: true
    status: running
    heap_size: 2586
    stack_size: 29
    reductions: 17707
  neighbours:

=SUPERVISOR REPORT==== 27-Jul-2021::11:50:56.609809 ===
    supervisor: {local,httpc_handler_sup}
    errorContext: child_terminated
    reason: {badarg,[{erlang,'++',
                             [<<"binary">>,"\r\n"],
                             [{error_info,#{module => erl_erts_errors}}]},
                     {http_request,headers_other,2,
                                   [{file,"http_request.erl"},{line,306}]},
                     {http_request,http_headers,1,
                                   [{file,"http_request.erl"},{line,82}]},
[...]

PR #5907, opened 2022-04-19
Update httpc.xml

cmullaparthi:
Documentation states that HTTP header values can be binaries but code does not accept it.

```
10> httpc:request(get, {"http://www.google.com", [{"some header", <<"some value">>}]}, [], []).
{error,{headers_error,not_strings}}

11> httpc:request(get, {"http://www.google.com", [{"some header", "some value"}]}, [], []).    
{ok,{{"HTTP/1.0",400,"Bad Request"},
     [{"date","Tue, 19 Apr 2022 20:58:49 GMT"},
      {"content-length","1555"},
      {"content-type","text/html; charset=UTF-8"},
      {"referrer-policy","no-referrer"}],
     [60,33,68,79,67,84,89,80,69,32,104,116,109,108,62,10,60,104,
      116,109,108,32,108,97|...]}}
```

### DIFF
--- a/lib/inets/doc/src/httpc.xml
+++ b/lib/inets/doc/src/httpc.xml
@@ -109,7 +109,7 @@
     <p><c>headers() = [header()]</c></p>
     <p><c>header() = {field(), value()}</c></p>
     <p><c>field() = [byte()]</c></p>
-    <p><c>value() = binary() | iolist()</c></p>
+    <p><c>value() = iolist()</c></p>
     <taglist>
       <tag><c>body()</c></tag>
       <item><p>= <c>http_string() | binary()</c></p>


KennethL:
Thanks for the PR, but see #5074 and #5269 and the release note here:

## Inets 7.5

### Improvements and New Features

    Extend header values to httpc:request/5 to allow binary() as well. Make error detection of invalid arguments to httpc:request/5 be more precise so an error is returned in more cases instead of causing a hang or function_clause. Be more precise in documentation regarding the types of arguments being accepted.

    Own Id: OTP-17579 Aux Id: GH-5074

This was introduced in OTP 24.2.2

cmullaparthi:
Thanks @KennethL I thought I was using the latest httpc but alas. Apologies for wasting your time :-)

PR #5917, opened 2022-04-20
Fix rare bug in failing binary_to_term call

sverker:
Make sure `live_hf_end` is saved when the first BIF trap with disabled GC happens.

### DIFF
--- a/erts/emulator/beam/jit/arm/instr_bif.cpp
+++ b/erts/emulator/beam/jit/arm/instr_bif.cpp
@@ -455,13 +455,13 @@ void BeamGlobalAssembler::emit_call_light_bif_shared() {
         /* ERTS_IS_GC_DESIRED_INTERNAL */
         {
             a.ldr(TMP1.w(), arm::Mem(c_p, offsetof(Process, flags)));
-            a.and_(TMP1, TMP1, imm(F_FORCE_GC | F_DISABLE_GC));
+            a.tst(TMP1, imm(F_FORCE_GC | F_DISABLE_GC));
 
             a.ldr(TMP1, arm::Mem(c_p, offsetof(Process, bin_vheap_sz)));
             a.ldr(TMP2, arm::Mem(c_p, offsetof(Process, off_heap.overhead)));
 
-            /* If neither F_FORCE_GC nor F_DISABLE_GC were set, test whether
-             * whether binary heap size should trigger GC.
+            /* If neither F_FORCE_GC nor F_DISABLE_GC were set,
+             * test whether binary heap size should trigger GC.
              *
              * Otherwise, set the flags as if `off_heap.overhead > bin_vheap_sz`
              * to force a GC. */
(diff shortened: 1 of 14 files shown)


erts/emulator/beam/jit/arm/instr_bif.cpp
```diff
@@ -456,7 +456,9 @@ void BeamGlobalAssembler::emit_call_light_bif_shared() {
         {
             /* Test whether GC is forced. */
             a.ldr(TMP1.w(), arm::Mem(c_p, offsetof(Process, flags)));
-            a.tbnz(TMP1, Support::ctz(F_FORCE_GC), gc_after_bif_call);
+            a.tbnz(TMP1,
```
jhogberg, line 459: `tbnz` can only test a single bit. How about changing this sequence to:

```
            a.ldr(TMP1.w(), arm::Mem(c_p, offsetof(Process, flags)));
            a.and_(TMP1, TMP1, imm(F_FORCE_GC | F_DISABLE_GC));

            a.ldr(TMP1, arm::Mem(c_p, offsetof(Process, bin_vheap_sz)));
            a.ldr(TMP2, arm::Mem(c_p, offsetof(Process, off_heap.overhead)));

            /* If neither F_FORCE_GC nor F_DISABLE_GC were set, test whether
             * whether binary heap size should trigger GC.
             * 
             * Otherwise, set the flags as if `off_heap.overhead > bin_vheap_sz`
             * to force a GC. */
            a.ccmp(TMP2, TMP1, imm(NZCV::kCF), imm(arm::CondCode::kEQ));

            a.sub(TMP1, E, HTOP);
            a.asr(TMP1, TMP1, imm(3));
            a.ldr(TMP2, arm::Mem(c_p, offsetof(Process, mbuf_sz)));

            /* If our binary heap size was small enough not to need a GC, check
             * whether the heap fragment size is larger than the remaining heap
             * size.
             * 
             * Otherwise, set the flags as if it is to force a GC. */
            a.ccmp(TMP1, TMP2, imm(NZCV::kVF), imm(arm::CondCode::kLS));
            a.b_lt(gc_after_bif_call);
```

jhogberg (commented):
LGTM.

erts/emulator/beam/jit/arm/instr_bif.cpp
sverker: Thanks. I kind of understand what ccmp does.

erts/emulator/beam/jit/arm/instr_bif.cpp
```diff
@@ -454,21 +454,28 @@ void BeamGlobalAssembler::emit_call_light_bif_shared() {
...
+            a.and_(TMP1, TMP1, imm(F_FORCE_GC | F_DISABLE_GC));
 
-            /* Test whether binary heap size should trigger GC. */
             a.ldr(TMP1, arm::Mem(c_p, offsetof(Process, bin_vheap_sz)));
             a.ldr(TMP2, arm::Mem(c_p, offsetof(Process, off_heap.overhead)));
-            a.cmp(TMP2, TMP1);
 
-            /* Test if heap fragment size is larger than remaining heap size. */
+            /* If neither F_FORCE_GC nor F_DISABLE_GC were set, test whether
+             * whether binary heap size should trigger GC.
```
JeromeDeBretagne: Nit: duplicated "whether" here

sverker:
Merged to master at 61c4f8ede7d9b15b6f7f5dcadd6127c8d56e3e35.

sverker:
Released in OTP-23.3.4.14 and OTP-24.3.4.

PR #5918, opened 2022-04-21
erts: Fix a race in persistent_term:get/1,2

### LINKED ISSUE #5908: persistent_term:get could return random term when racing happens
**Describe the bug**
In very rare case, we see persistent_term:get/2 returns a random value that has never been set.

For example, we only have 2 write operations for key 'some_key': persistent_term:put('some_key', true) and persistent_term:erase('some_key'), but persistent_term:get('some_key', false) could return an integer or a reference or anything.

Our theory is, persistent_term_get_2 does not have locks, and  *hash_table can be changed during lookup and get_bucket operations.
https://github.com/erlang/otp/blob/master/erts/emulator/beam/erl_bif_persistent.c#L490

Basically, if some persistent_term write operations (erase, put, which changes the erts_atomic_read_nob(&the_hash_table)) happened when doing persistent_term_get_2, *hash_table is possibly freed and filled by anything.

If it also happens that get_bucket(hash_table, entry_index) is a tuple of {_, RandomTerm}, then persistent_term_get_2 will return RandomTerm as result.

**To Reproduce**
This is a really rare racing condition. We can only reproduce that erts_atomic_read_nob(&the_hash_table) can be changed to something different with hash_table while doing lookup and get_bucket operation
see https://github.com/potatosalad/otp/commit/7b460431c53a5715d4ca6d517ccb876470fb7b81

**Expected behavior**
persistent_term:get should never return a value that has never been set.

This might be achievable by adding a read write lock to hash_table: Basically, erts_atomic_read_nob(&the_hash_table) should not be changed when there is an on-going read operation.

If this hurts the performance a lot, we could at least add a lookup_term function which does an atomic operation instead of two steps: entry_index = lookup(hash_table, key); get_bucket(hash_table, entry_index);

**Affected versions**
erts-12.1.2

jhogberg:
Fixes #5908

### DIFF
--- a/erts/emulator/beam/erl_bif_persistent.c
+++ b/erts/emulator/beam/erl_bif_persistent.c
@@ -136,7 +136,7 @@ typedef struct {
     HashTable* old_table;
     HashTable* new_table;
     Uint entry_index;
-    Eterm old_term;
+    Eterm old_bucket;
     HashTable* tmp_table;
     int must_shrink;
     ErtsPersistentTermCpyTableCtx cpy_ctx;
@@ -147,7 +147,7 @@ typedef struct {
  */
 
 static HashTable* create_initial_table(void);
-static Uint lookup(HashTable* hash_table, Eterm key);
+static Uint lookup(HashTable* hash_table, Eterm key, Eterm *bucket);
 static int is_erasable(HashTable* hash_table, Uint idx);
 static HashTable* copy_table(ErtsPersistentTermCpyTableCtx* ctx);
 static int try_seize_update_permission(Process* c_p);
@@ -307,6 +307,7 @@ BIF_RETTYPE persistent_term_put_2(BIF_ALIST_2)
     static const Uint ITERATIONS_PER_RED = 32;
     ErtsPersistentTermPut2Context* ctx;
     Eterm state_mref = THE_NON_VALUE;
+    Eterm old_bucket;
     long iterations_until_trap;
     long max_iterations;
 #define PUT_TRAP_CODE                                                   \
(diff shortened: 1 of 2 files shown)


erts/emulator/beam/erl_bif_persistent.c
```diff
@@ -953,19 +960,25 @@ cleanup_trap_data(Binary *bp)
...
 
     while (1) {
         term = get_bucket(hash_table, idx & mask);
-        if (is_nil(term) || EQ(key, (tuple_val(term))[1]))
-            break;
+
+        if (is_nil(term) || EQ(key, (tuple_val(term))[1])) {
+            if (bucket) {
+                *bucket = term;
+            }
```
sverker, line 975: `bucket` argument is never NULL to `lookup()` what I can see

erts/emulator/beam/erl_bif_persistent.c
```diff
@@ -359,7 +360,7 @@ BIF_RETTYPE persistent_term_put_2(BIF_ALIST_2)
     ctx->key = BIF_ARG_1;
     ctx->term = BIF_ARG_2;
 
-    ctx->entry_index = lookup(ctx->hash_table, ctx->key);
+    ctx->entry_index = lookup(ctx->hash_table, ctx->key, &ctx->old_bucket);
```
sverker, line 363: The first `get_bucket()` call in `persistent_term_put_2` can be replaced with `ctx->old_bucket`:

`if (is_nil(get_bucket(ctx->hash_table, ctx->entry_index))) {
`

erts/emulator/beam/erl_bif_persistent.c
```diff
@@ -122,6 +122,7 @@ typedef struct {
     HashTable* hash_table;
     Eterm heap[3];
     Eterm tuple;
+    Eterm old_bucket;
```
sverker, line 125: `old_bucket` does not need to be in ErtsPersistentTermPut2Context as it's not used over traps, but it doesn't do any harm either so...

sverker (approved):

erts/emulator/beam/erl_bif_persistent.c
jhogberg: Thanks, fixed. :)

erts/emulator/beam/erl_bif_persistent.c
jhogberg: Good catch, thanks!

erts/emulator/beam/erl_bif_persistent.c
jhogberg: Good point, fixed. :)

max-au (approved):

PR #5924, opened 2022-04-22
stdlib: enhances the shell completion

frazze-jobb:
Have all functionality more or less in place, but I need to do some refactoring, and there are already 3 bugs that I know of.

I need to deal with c:memory/1 which have two type specs, I can use my fold function
but the fold function doesn't work. It can be seen in the new failing edlin_expand_SUITE testcases.
I also need to handle fun Mod:Fun/X, properly.

Please evaluate and see if the completions are helpful. I am thinking if a shell keyboard shortcut could show the whole type spec. Instead of just the current parameter of interest.
Also file:open("/filename",<tab> just gives ... because the suggestion is so long, file:open("/filename", [<tab> will work. We need to decide on some heuristics when to print all the concrete types  and when to print an abstract type.


lib/stdlib/src/edlin_expand.erl
```diff
@@ -20,43 +20,780 @@
...
 
 -import(lists, [reverse/1, prefix/2]).
 
-%% expand(CurrentBefore) ->
-%%	{yes, Expansion, Matches} | {no, Matches}
+%% Will remove this before delivery
+%% Questions:
+%% Types:
+%%  should we print all the leaf concrete types or stop at user_type and remote_type?
+%%   - file:open("/", )
```
garazdawi, line 32: I wonder if maybe it would be a good idea to expand one "level" of remote types per <TAB> you hit. I.e. in the example of `file:open("/", )` we first show:

```
directory    mode()     ram
```
then expand mode() to
```
directory read write append exclusive raw binary {delayed_write, non_neg_integer(), non_neg_integer()} delayed_write {read_ahead, pos_integer()} read_ahead compressed {encoding, unicode:encoding()} sync ram
```
It should however not expand the encoding as that is within another compound element, so leave that until the user does:

```
file:open("/",[{encoding, <TAB><TAB>
```

lib/stdlib/src/edlin_expand.erl
```diff
@@ -20,43 +20,780 @@
...
+%% Maps:
+%%  cant support #{hej:=T} = M because we cannot look ahead to see that its M we extract from
+%%      could try to complete with all keys of all maps in Bindings list,
+%%      but it could just confuse the user, and chances are that most of the time the user
+%%      is interested in extracting a key of a map that is returned from a function
+%% Hidden functions:
+%%  should we remove them from display, but still support autocompletion of them?
+%% Imported functions / shell commands (available in shell_default module):
+%%  should we make these visible for auto completion, is_atom/1, ht ...
+%%  should a banner for different suggestions be shown when suggestions are of different types?
```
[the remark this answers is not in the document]
garazdawi: Actually, maybe we can do it like this:

```
1> y<TAB>
y/1           y/2           yecc:
yeccparser:   yeccscan:
```

and then only autocomplete things in shell_default as top level expressions, so if we do:

```
1> element(y<TAB>
yecc:         yeccparser:   yeccscan:
```
it would only show the modules.

The difference between auto-imported functions (such as is_atom) and shell_default can be left for a potential syntax highlighter.

lib/stdlib/src/edlin_expand.erl
frazze-jobb: Just added this handling, I like it

lib/stdlib/src/edlin_expand.erl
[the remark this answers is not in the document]
frazze-jobb: It is done,

I have been thinking a little bit on formatting please have a look and let me know your preferences
https://gist.github.com/frazze-jobb/5ea36dd14a8115b06e670c7f1e79f18f

additionally,
should we special handle file:filename_all(), file:name_all() and just print filename()
currently date() has a range type printing 1..31
do we want to write non_neg_integer() as 0..
neg_integer() ..-1
positive_integer() 1..
or just keep them?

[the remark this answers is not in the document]
garazdawi: I also wrote some comments in the gist.

> should we special handle file:filename_all(), file:name_all() and just print filename()

Do you mean instead of expanding the type to its base types? If so, then yes I would say that we want that. If we cannot come up with a general method that does what we want.

> currently date() has a range type printing 1..31

Maybe we can update the type to include a name? i.e. 

```erlang
-type date()           :: {Year :: year(),Month :: month(),Day :: day()}.
```

Or do we want to also include the name of the expanded type. i.e. 

```bash
1> calendar:date_to_gregorian_days({<TAB>
year() :: 1970..10000
```

> do we want to write non_neg_integer() as 0..
> neg_integer() ..-1
> positive_integer() 1..

In erl_docgen we translate these to be `integer() >= 0`: https://github.com/erlang/otp/blob/master/lib/erl_docgen/src/docgen_otp_specs.erl#L279-L290. So to be consistent maybe we should do the same here?

lib/kernel/src/code.erl
```diff
@@ -906,21 +907,41 @@ get_doc_chunk_from_ast(Filename) ->
...
-                           docs = Docs }};
+                           metadata = #{ generated => true, otp_doc_vsn => ?CURR_DOC_VERSION},
+                           docs = Docs++Types }};
         {ok, {_Mod, [{abstract_code,no_abstract_code}]}} ->
             {error,missing};
         Error ->
             Error
     end.
 
+get_type_doc(Mod) ->
```
garazdawi, line 921: I think that we want to expand the `code:get_doc` API to be able to do we want here. Something like:

```erlang
%% Returns what we have today. That is eep48 is available, otherwise AST if available
> code:get_doc(Mod, #{ sources => [eep48, debug_info] }).
%% Returns documentation created from AST
> code:get_doc(Mod, #{ sources => [debug_info] }).
%% Returns documentation created from AST if available, otherwise eep48 if available
> code:get_doc(Mod, #{ sources => [debug_info, eep48] }).
```

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -77,16 +84,119 @@ normal(Config) when is_list(Config) ->
...
+type_completion(_Config) ->
+    %% test that edlin_expand:type_traverser works for all preloaded modules in the shell
+    ct:timetrap({minutes, 10}),
+    LoadedModules = [to_atom(M) || {M,_,_} <- code:all_available()],
+    ModExports = [{Mod, edlin_expand:get_exports(Mod)} || {ok, Mod} <- LoadedModules],
+        [[begin
+            Str = atom_to_list(Mod) ++ ":" ++ atom_to_list(Func) ++ "(",
+            io:format("~p~n", [Str]),
+            do_expand(Str)
+          end || {Func, _}<- Exports] || {Mod, Exports} <- ModExports],
```
garazdawi, line 108: Should the test also include all arguments to each function? That is something like this:

```suggestion
        [lists:foreach(fun({Func, Arity}) ->
          lists:foreach(fun(A) ->
            Str = io_lib:write_atom(Mod) ++ ":" ++ io_lib:write_atom(Func) ++ "(" ++ lists:duplicate(A, "ok,"),
            io:format("~p~n", [Str]),
            do_expand(Str)
          end, lists:seq(0, Arity))
         end, Exports) || {Mod, Exports} <- ModExports],
```

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -77,16 +84,119 @@ normal(Config) when is_list(Config) ->
...
+    ok.
+
+filename_completion(Config) ->
+    %% test that filenames can be completed
+    {yes,"ll\"", [{"null",""}]} = do_expand("\"/dev/nu"),
```
garazdawi, line 113: These filenames will not exist on Windows so something else has to be used there.

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -77,16 +84,119 @@ normal(Config) when is_list(Config) ->
...
+    LoadedModules = [to_atom(M) || {M,_,_} <- code:all_available()],
+    ModExports = [{Mod, edlin_expand:get_exports(Mod)} || {ok, Mod} <- LoadedModules],
+        [[begin
+            Str = atom_to_list(Mod) ++ ":" ++ atom_to_list(Func) ++ "(",
+            io:format("~p~n", [Str]),
+            do_expand(Str)
+          end || {Func, _}<- Exports] || {Mod, Exports} <- ModExports],
+    ok.
+
+filename_completion(Config) ->
```
garazdawi, line 111: Add a test that hidden files are shown when you auto-complete `.`. That is:

```
> "otp_src/.<TAB>
../               .devcontainer/    .dir-locals.el    .editorconfig     
.git/             .gitattributes    .github/          .gitignore        
.gitpod.yml       .mailmap          .vscode/          
```

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -18,12 +18,17 @@
 %% %CopyrightEnd%
```
garazdawi, line 18: Can you add tests for Unicode names, filenames, fields variables etc? Just in case anyone runs [beamoji](https://github.com/spawnfest/beamoji) on the source code and tries to autocomplete.

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -77,16 +84,119 @@ normal(Config) when is_list(Config) ->
...
+    %% test first and second parameter
+    %% test multiple arities with same type on first parameter
+    %% test multiple arities with different type on first parameter
+    %% test that recursive types does not trigger endless loop
+    %% test that getting type of out of bound parameter does not trigger crash
+    compile_and_load2(Config,complete_function_parameter),
+    {no, [], []} = do_expand("complete_function_parameter:an_untyped_fun("),
+    {no, [], [{"Start ::",[]},{"integer()",[]}]} = do_expand("complete_function_parameter:a_fun_name("),
+    {no, [], [{"End ::",[]},{"integer()",[]}]} = do_expand("complete_function_parameter:a_fun_name(1,"),
+    {no, [], [{"End ::",[]},{"integer()",[]}]} = do_expand("complete_function_parameter : a_fun_name ( 1 , "),
```
garazdawi, line 177: Should newlines work? i.e.

```
    {no, [], [{"End ::",[]},{"integer()",[]}]} = do_expand("complete_function_parameter:a_fun_name ( 1 ,\n"),
```

lib/stdlib/test/edlin_expand_SUITE.erl
frazze-jobb: The actual type traversal happens when we complete the first parameter, so all errors would already appear there, checking the other parameters dont give us much I think, if we would traverse down in nestings, that could give us something but then we would need to detect the type of each parameter to see if it takes tuples/maps/lists and add {,#{,[. This will probably take a lot of time though, maybe the function can be optimized further but the test already takes more than 10minutes locally.

garazdawi (commented):
Reviewed code and group so far.

lib/kernel/src/group.erl
```diff
@@ -20,10 +20,15 @@
 -module(group).
 
 %% A group leader process for user io.
-
--export([start/2, start/3, server/3]).
+-export([start/2, start/3, server/3, read_term/1, strip_types/1]).
```
garazdawi, line 23: Where are `read_term/1` and `strip_types/1` used?

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -60,98 +72,291 @@ cleanup() ->
...
+%     case erl_scan:string(Str) of
+%     {ok, [{atom,_,A}], _} ->
+%         {ok, A};
+%     _ ->
+%         error
+%     end.
+
+type_completion(_Config) ->
+    ok.
+    %% test that edlin_expand:type_traverser works for all preloaded modules in the shell
```
garazdawi, line 111: I ran this test case and it crashed because `sofs:from_sets(` does not work.

lib/stdlib/src/shell.erl
```diff
@@ -976,15 +1042,79 @@ local_func(h, [], Bs, Shell, RT, _Lf, _Ef) ->
...
+    ets:insert(FT, [{{function, {shell_default, FunName, Arity}}, Fun}]),
+    {value, FunName2, Bs};
+local_func(ft, _, Bs, _Shell, _RT, FT, _Lf, _Ef) ->
+    SecretArg = get(secret_arg),
+    put(secret_arg, []),
+    case erl_parse:parse_form(SecretArg) of
+        {ok, {attribute,_,spec,{{FunName, FunArity},_}}=AttrForm} ->
+            FunName2 = erlang:make_fun(shell_default, FunName, FunArity),
+            ets:insert(FT, [{{function_type, {shell_default, FunName, FunArity}}, AttrForm}]),
+            {value, FunName2, Bs};
```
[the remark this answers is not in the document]
frazze-jobb: What would be the most intuitive to return? "ok." ?
-record(). -type . -spec . : all end with . meaning that the return value cannot be used anyway. So maybe ok. makes sense for all of them?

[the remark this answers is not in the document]
garazdawi: yes, I think that makes sense.

garazdawi (changes requested):
Lots of great stuff in here now. Some details to polish though. For example:

#### No completions

If I type:

```erlang
1> lists:ab<TAB>
functions
```
It prints an empty `functions` heading. I think it should not print anything as that is what it does when it cannot find a module that matches. Possibly the same thing applies when it does not find an argument type that matches. Right now it prints the slogan of the function, but should it print nothing?

Or maybe both of these should print the first suggestion that appears by erasing the characters before the cursor one by one? That is `lists:ab` would show the suggestions for `lists:a` and `lists:all([{` would show the suggestions for `lists:all(`? Though I suppose that could potentially be very expensive to compute...

#### Multiple clauses

The completion for `ssh:connect(<TAB>` is not very intuitive:

<pre><code>
16> ssh:connect(&lt;TAB>
ssh:connect(<b>Host</b>, Port, Options, NegotiationTimeout)
ssh:connect(<b>Var1</b>, Var2, Var3)
<b>ssh:host()</b> :: {0..255, ...}      {0..65535, ...}    loopback           string()
ssh:connect(<b>OpenTcpSocket</b>, Options)
ssh:connect(<b>Var1</b>, Var2, Var3)
<b>ssh:open_socket()</b> :: {'$inet', ...}    port()
</code></pre>

Where is the Var1 etc coming from? `ssl:connect(<TAB>` looks much better.

<pre><code>
ssl:connect(<b>Host</b>, Port, TLSOptions)
ssl:connect(<b>Host</b>, Port, TLSOptions, Timeout)
<b>ssh:host()</b> :: {0..255, ...}      {0..65535, ...}    loopback           string()
ssl:connect(<b>TCPSocket</b>, TLSOptions)
ssl:connect(<b>TCPSocket</b>, TLSOptions, Timeout)
<b>ssl:socket()</b> :: {'$inet', ...}    port()
</code></pre>


#### system_info

erlang:system_info is one of the most complex functions we have to document, so I played around with it and found some bugs in the completion code.

1. `erlang:system_info(<TAB>` prints <code>erlang:system_<b>info</b>(info)</code>, i.e. the wrong `info` is highlighted.
2. `erlang:system_info({allocator,<TAB>` should show `atom()` as only possible completions, but shows a lot of other options as well.
3. `erlang:system_info(dist_<TAB>` shows this:  
    <pre><code><b>modules</b>
   dist_ac:     dist_util:
   erlang:system_info(<b>dist_buf_busy_limit</b>)
   erlang:system_info(<b>dist_ctrl</b>)</code></pre>
   I wonder if we should not show modules when there are matching function arguments? Should function name completion have its own header?

#### External fun completion

If I type `[fun ssl:connect/<TAB>` the autocompletion is `2,   3,   4,`. Why is there a comma at the end? I could want to use any of `,`, `]` or `|` at that position.

#### Fun completion

If I type `fun(a) when<TAB>` it completes as `when`, however it should not give any completions.
If I type `fun(A) -<TAB>` I get a bunch of incorrect completions.
If I type `fun(A) -> <TAB>` I get the same completions as above, should I really get any completions at all here?

lib/stdlib/src/shell_default.erl
```diff
@@ -154,3 +154,12 @@ iv()            -> calli(iv, []).
...
     c:appcall(debugger, i, F, Args).
+
+'$handle_undefined_function'(Func, Args) ->
+    {shell_state, _Bs, _RT, FT} = shell:get_state(),
+    try
+        {value, {_, Fun}} = lists:keysearch({function, {shell_default,Func,length(Args)}}, 1, FT),
+        apply(Fun, Args)
+    catch _:_ ->
+            error_handler:raise_undef_exception(?MODULE, Func, Args)
+    end.
```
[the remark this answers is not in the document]
frazze-jobb: I changed it, but I still need to account for the arity of the function when fetching it, in case the user defines both foo(A) and foo(A,B)

frazze-jobb:
> Lots of great stuff in here now. Some details to polish though. For example:
> #### No completions
> 
> If I type:
> 
> ```erlang
> 1> lists:ab<TAB>
> functions
> ```
> 
> It prints an empty `functions` heading. I think it should not print anything as that is what it does when it cannot find a module that matches. Possibly the same thing applies when it does not find an argument type that matches. Right now it prints the slogan of the function, but should it print nothing?
> 
I agree that it should not show the section title
if we would erase the last letter and get completions for that, then I think it could be easy to miss that you have typos on your prompt 

> Or maybe both of these should print the first suggestion that appears by erasing the characters before the cursor one by one? That is `lists:ab` would show the suggestions for `lists:a` and `lists:all([{` would show the suggestions for `lists:all(`? Though I suppose that could potentially be very expensive to compute...
> #### Multiple clauses
> 
> The completion for `ssh:connect(<TAB>` is not very intuitive:
> 
> ```
> 
> 16> ssh:connect(<TAB>
> ssh:connect(Host, Port, Options, NegotiationTimeout)
> ssh:connect(Var1, Var2, Var3)
> ssh:host() :: {0..255, ...}      {0..65535, ...}    loopback           string()
> ssh:connect(OpenTcpSocket, Options)
> ssh:connect(Var1, Var2, Var3)
> ssh:open_socket() :: {'$inet', ...}    port()
> ```
> 
> Where is the Var1 etc coming from? `ssl:connect(<TAB>` looks much better.
> 
> ```
> 
> ssl:connect(Host, Port, TLSOptions)
> ssl:connect(Host, Port, TLSOptions, Timeout)
> ssh:host() :: {0..255, ...}      {0..65535, ...}    loopback           string()
> ssl:connect(TCPSocket, TLSOptions)
> ssl:connect(TCPSocket, TLSOptions, Timeout)
> ssl:socket() :: {'$inet', ...}    port()
> ```

regarding multple clauses, the 'Var's appears if you have 'type()' in your function parameters in the spec, the reason I decided to it this way is because I expand the types, and putting the expanded types inside the function parameters would be nasty. But I could probably put the non expanded type in the function parameter somehow.
```
-spec connect(open_socket(), client_options(), timeout()) ->
                     {ok, connection_ref()} | {error, term()}
           ; (host(), inet:port_number(), client_options()) ->
                     {ok, connection_ref()} | {error, term()}.
```
so it would become:
```
ssh:connect(Host, Port, Options, NegotiationTimeout)
ssh:connect(host(), inet:port_number(), client_options())
ssh:host() :: {0..255, ...}      {0..65535, ...}    loopback           string()
ssh:connect(OpenTcpSocket, Options)
ssh:connect(open_socket(), client_options(), timeout())
ssh:open_socket() :: {'$inet', ...}    port()
```
> #### system_info
> 
> erlang:system_info is one of the most complex functions we have to document, so I played around with it and found some bugs in the completion code.
> 
>     1. `erlang:system_info(<TAB>` prints `erlang:system_info(info)`, i.e. the wrong `info` is highlighted.
I will have a look if I can make a smarter highlighting algorithm

>     2. `erlang:system_info({allocator,<TAB>` should show `atom()` as only possible completions, but shows a lot of other options as well.
> 
This is the purpose of edlin_expand:match_argument function, however, when the term is not completed (in this case an uncompleted tuple), then everything can match, but I agree it should filter the others out.

>     3. `erlang:system_info(dist_<TAB>` shows this:
>         ```
>        modules
>        dist_ac:     dist_util:
>        erlang:system_info(dist_buf_busy_limit)
>        erlang:system_info(dist_ctrl)
>        ```
>          
>        I wonder if we should not show modules when there are matching function arguments? Should function name completion have its own header?
> 
I can make a section title for that.
I don't know about your first idea, sometimes you want to have the possibility to get modules. I thought about if you could have a special keybind for module expansion vs the rest expansion. But not sure if its worth it. Often times module names do 'collide' with atoms and keywords. 

> 
> #### External fun completion
> 
> If I type `[fun ssl:connect/<TAB>` the autocompletion is `2, 3, 4,`. Why is there a comma at the end? I could want to use any of `,`, `]` or `|` at that position.

You could, my thought was that most of the time I want a comma when I specify my list and if I would want something else I would just erase and put the correct char. With the keybind for closing we could maybe erase the last comma and put the ]. But I don't have a good idea regarding '|'.

> #### Fun completion
> 
> If I type `fun(a) when<TAB>` it completes as `when`, however it should not give any completions. If I type `fun(A) -<TAB>` I get a bunch of incorrect completions. If I type `fun(A) -> <TAB>` I get the same completions as above, should I really get any completions at all here?

I agree, no completions

frazze-jobb:
After discussion offline with @garazdawi , we concluded that list elements should never autocomplete with , | or ]  and maps should not complete with , or }. With trailing comma support there might be a possibility that we revisit this.
For now it's easier to just type , instead of tab.

lib/kernel/test/interactive_shell_SUITE.erl
```diff
@@ -882,6 +889,93 @@ shell_invalid_ansi(_Config) ->
...
+    Peer = #{ name => Name,
+              post_process_args =>
+                        fun(["new-window","-n",_,"-d","--"|CmdAndArgs]) ->
+                                FlatCmdAndArgs =
+                                      lists:join(
+                                        " ",[[$',A,$'] || A <- CmdAndArgs]),
+                                ["send","-t",Name,lists:flatten(FlatCmdAndArgs),"Enter"]
+                        end
+            },
+    Term = start_tty([{peer, Peer}|Config]),
```
garazdawi, line 906:
```suggestion
    Term = start_tty(Config),
```

I don't think you need the part above, only `start_tty(Config)` should work.

lib/kernel/test/interactive_shell_SUITE.erl
```diff
@@ -882,6 +889,93 @@ shell_invalid_ansi(_Config) ->
...
+        send_tty(Term, "still "),
+        send_tty(Term, "C-O"), %% save in nano
+        send_tty(Term, "Enter"),
+        send_tty(Term, "C-X"), %% quit in nano
+        check_content(Term,"still some text with\\s*\n\\s*newline in it"),
+        ok
+    after
+        stop_tty(Term),
+        ok
+    end.
```
garazdawi, line 948:
```suggestion
external_editor(Config) ->
    Term = start_tty(Config),
    try
        send_tty(Term,"os:putenv(\"EDITOR\",\"nano\").\n"),
        send_tty(Term, "some text with\nnewline in it"),
        check_content(Term,"2> some text with\n\\s+newline in it"),
        send_tty(Term, "C-O"),
        check_content(Term,"GNU nano [\\d.]+\\s+.+\\s+some text with\\s*\n\\s*newline in it"),
        send_tty(Term, "still "),
        send_tty(Term, "C-O"), %% save in nano
        send_tty(Term, "Enter"),
        send_tty(Term, "C-X"), %% quit in nano
        check_content(Term,"still some text with\\s*\n\\s*newline in it"),
        ok
    after
        stop_tty(Term),
        ok
    end.
```

No need to have the initial code and we need to set the EDITOR in the Erlang shell. Also the nano version can contain atleast 3 digits, maybe more so I made the regexp more lax.

lib/kernel/src/code.erl
```diff
@@ -860,25 +861,38 @@ where_is_file(Tail, File, Path, Files) ->
...
     case which(Mod) of
         preloaded ->
-            Fn = filename:join([code:lib_dir(erts),"ebin",atom_to_list(Mod) ++ ".beam"]),
-            get_doc_chunk(Fn, Mod);
+            Ebin = case Source of
+                debug_info -> "preloaded/ebin";
+                eep48 -> "ebin"
+            end,
+            Fn = filename:join([code:lib_dir(erts),Ebin,atom_to_list(Mod) ++ ".beam"]),
+            GetDoc(Fn);
```
garazdawi, line 883:
```suggestion
            ErtsDir = code:lib_dir(erts),
            ErtsEbinDir =
                case filelib:is_dir(filename:join([ErtsDir,"ebin"])) of
                    true -> filename:join([ErtsDir,"ebin"]);
                    false -> filename:join([ErtsDir,"preloaded","ebin"])
                end,
            Fn = filename:join([ErtsEbinDir, atom_to_list(Mod) ++ ".beam"]),
            GetDoc(Fn);
```

Fetch code from correct dir when in source tree

lib/kernel/test/interactive_shell_SUITE.erl
```diff
@@ -1348,12 +1421,12 @@ test_exit_initial(new) ->
       [{putline, ""},
        {expect, "1> $"},
        {putline, "2."},
-       {expect, "2"},
-       {putline,"exit()."},
-       {expect, "Eshell"},
-       {expect, "1> $"},
+        {expect, "2"},
```
garazdawi, line 1424: The indentation in this file is still incorrect.

frazze-jobb: Reverted indentation to what it was before my changes

lib/stdlib/src/edlin_context.erl
```diff
@@ -0,0 +1,461 @@
...
+%% Recursively extracts nested keyword expressions
+%% Note: In the future we could autocomplete case expressions by looking at the
+%% return type of the expression.
+over_keyword_expression(Bef) ->
+    over_keyword_expression(Bef, []).
+over_keyword_expression("dne"++Bef, Expr)->
+    %% Nested expression
+    {Bef1, KWE}=over_keyword_expression(Bef),
+    over_keyword_expression(Bef1, KWE++"end"++Expr);
+over_keyword_expression("fi"++Bef, Expr) -> {Bef, "if" ++ Expr};
```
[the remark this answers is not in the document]
frazze-jobb: I checked with @bjorng and there is no such function. 

We are only interested in keywords that end with an 'end' atom, so we would have to keep a list of those as well. Maybe it would be more concise if we tried to read a word and check if it is one of our special keywords, instead of pattern matching on the reversed string.  Do you think its worth adding those two reserved words functions, and doing the refactoring?

lib/kernel/test/rtnode.erl
```diff
@@ -217,7 +218,21 @@ get_progs() ->
             {RunErl, ToErl, Erl};
         _ ->
             {error,"Not a Unix OS"}
-        end.
+    end.
+get_progs(Opts) ->
+    case get_progs() of
+        {RunErl, ToErl, Erl} ->
+            case proplists:get_value(release, Opts) of
+                none -> {RunErl, ToErl, Erl};
```
garazdawi, line 226:
```suggestion
                undefined -> {RunErl, ToErl, Erl};
```

lib/stdlib/test/edlin_expand_SUITE.erl
```diff
@@ -60,202 +73,600 @@ cleanup() ->
...
+    {no,[],[]} = do_expand("\" "),
+    {no,[],[]} = do_expand("\"\""),
+    {yes, "e\"", [{".hidden file",_}]} = do_expand("\".hidden\\ fil"),
+    {yes,"/", [{"../",_}]} = do_expand("\".."),
+    {yes,"ta/", _} = do_expand("\"../edlin_expand_SUITE_da"),
+    {yes,"erl\"",[{"complete_function_parameter.erl",_}]} = do_expand("\"complete_function_parameter."),
+    case os:type() of
+        {win32,_} -> {skip, "Unicode on filenames in windows are tricky"};
+        {win64,_} -> {skip, "Unicode on filenames in windows are tricky"};
+        _ ->
```
garazdawi, line 163:
```suggestion
    case {os:type(), file:native_name_encoding()} of
        {{win32,_},_} -> {skip, "Unicode on filenames in windows are tricky"};
        {_,latin1} -> {skip, "Cannot interpret unicode filenames when native_name_encoding is latin1"};
        _ ->
```

lib/kernel/test/rtnode.erl
```diff
@@ -217,7 +218,21 @@ get_progs() ->
...
+                Release ->
+                    case test_server_node:find_release(Release) of
+                        false -> {error, "Could not find release "++Release};
```
garazdawi, line 229:
```suggestion
                        none -> {error, "Could not find release "++Release};
```

garazdawi (approved):

PR #5926, opened 2022-04-22
[RFC] mnesia: Add post-commit hook

ieQu1:
Hello, OTP team. 

Our team has been working on improving horizontal scalability of mnesia by making replication partially eventually consistent. ["Whitepaper"](https://www.emqx.com/en/blog/mqtt-broker-clustering-part-3-challenges-and-solutions-of-emqx-horizontal-scalability#introducing-mria) [Source code](https://github.com/emqx/mria).
The results are quite promising, we've been able to run 20+ node mria/mnesia clusters with high write and read throughput.
However, in order to do this we had to add a hook point in `mnesia_tm`. Although the patch is really small, ideally we would like to merge it upstream.

Currently thsi PR is WIP, it lacks at least two things:
- Documentation
- Tests
If you are interested in accepting this patch, I will fix both issues.
Documentation is missing because I am not sure how to present this feature. It exposes some internal data structures, so it's not meant for use in the business logic. So the documentation should either go with a large disclaimer that the backward compatibility is not guaranteed, or, perhaps, it can be omitted altogether.
As for tests, we have some in mria repo, but I am not sure in which test suite in OTP can we put it. Some advice is welcome.

Thank you.

### DIFF
--- a/lib/mnesia/src/Makefile
+++ b/lib/mnesia/src/Makefile
@@ -72,7 +72,8 @@ MODULES= \
 	mnesia_sup \
 	mnesia_sp  \
 	mnesia_text \
-	mnesia_tm
+	mnesia_tm \
+	mnesia_hook
 
 HRL_FILES= mnesia.hrl
 
--- a/lib/mnesia/src/mnesia.app.src
+++ b/lib/mnesia/src/mnesia.app.src
@@ -32,7 +32,8 @@
 	     mnesia_sup,
 	     mnesia_sp,
 	     mnesia_text,
-	     mnesia_tm
+	     mnesia_tm,
+	     mnesia_hook
             ]},
   {registered, [
 		mnesia_dumper_load_regulator,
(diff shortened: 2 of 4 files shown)


ieQu1:
Hi, 

Sorry I missed that "waiting" label has been applied to the ticket. What kind of input can I provide?

dgud:
I think the wrong label was applied, sorry for taking some time, this is stuff that goes to the bottom
of the priority list for us :-/

And I have forgotten the answer (I'm old and vacation) to why this could not be in commit_work?

ieQu1:
P.S. There is another reason where it can be useful: other storage backends (such as rocksdb) can use this API for batch operations.

Mikaka27:
Hi, is this still relevant for you?

ieQu1:
Hi!

Yes, we are still using this hook.

mwieru:
Hi, our team also need post-commit hook feature. 

My proposal is to add new optional callback function to mnesia access module which can handle post-commit operations. 

We use several mnesia access modules in our project, so callback function is better for us to control when we really want 
post-commit hook to be executed.

PR #5927, opened 2022-04-22
Link to docs on heap sizing

nathanl:
I need some guidance on the proper syntax to make this link.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -6139,6 +6139,11 @@ receive_replies(ReqId, N, Acc) ->
           If <c><anno>MaxHeapSize</anno></c> is an integer, the system default
           values for <c>kill</c> and <c>error_logger</c> are used.
         </p>
+        <p>
+          For details on how the heap grows, see
+          <seeguide marker="erts:GarbageCollection#sizing-the-heap">Sizing
+          the heap</seeguide> in the ERTS internal documentation.
+        </p>
         <taglist>
           <tag><c>size</c></tag>
           <item>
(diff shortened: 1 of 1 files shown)


erts/doc/src/erlang.xml
```diff
@@ -6139,6 +6139,10 @@ receive_replies(ReqId, N, Acc) ->
           If <c><anno>MaxHeapSize</anno></c> is an integer, the system default
           values for <c>kill</c> and <c>error_logger</c> are used.
         </p>
+        <p>
+          For details on how the heap grows, see
+          https://www.erlang.org/doc/apps/erts/garbagecollection#sizing-the-heap
```
garazdawi, line 6144:
```suggestion
          <seeguide marker="erts:garbagecollection#sizing-the-heap">Sizing the heap</seeguide> in the ERTS internal documentation.
```

garazdawi:
Hello! The documentation for how to write links can be found here: https://www.erlang.org/doc/apps/erl_docgen/inline_tags.html#%3Csee*%3E---see-tags

garazdawi:
Hello!

I reproduced your example in `rr` using this script:

```bash
while cerl -rr -debug +S 8:8 -noshell -pa _build/default/lib/*/ebin -eval "segfault:start(), timer:sleep(1000), erlang:halt()."; do
  echo again
done
```

The problem seems to be that if a GC happens while a process is about to do a switch between `on_heap` to `off_heap` message queue data and there is an `on_heap` message in the external message queue, the GC does not include the message in the rootset and thus the VM segfaults later. Can you try the fix in #5934 and see if that solves the problem?

garazdawi:
I pushed an update as the name in the link has to have the same casing as the file it links to. Will merge once GH actions pass.

garazdawi:
Thanks!

PR #5936, opened 2022-04-27
reltool: Add option for selective beam stripping

attah:
The debug_info option can now take a list of chunk ids to keep.
Chunk ids are as per beam_lib:chunkid().

For example, {debug_info, ["Attr"]} will keep only the attributes
in the beam files and strip the rest.

---
Why?
We have some code that introspects itself for some declarative use-cases (mainly "wild" attributes), and recently we started needing to be able to deliver stripped binaries.

### DIFF
--- a/lib/reltool/doc/src/reltool.xml
+++ b/lib/reltool/doc/src/reltool.xml
@@ -226,10 +226,10 @@
 
       <tag><c>debug_info</c></tag>
       <item>
-        <p>The <c>debug_info</c> parameter controls whether the debug
-        information in the beam file should be kept (<c>keep</c>) or
-        stripped <c>strip</c> when the file is copied to the target
-        system.</p>
+        <p>The <c>debug_info</c> parameter controls what debug
+        information in the beam file should be kept or stripped.
+        <c>keep</c> keeps all debug info, <c>strip</c> strips all debug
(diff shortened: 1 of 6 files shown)


attah:
Hmm... neither exporting the type in beam_lib, nor adding it to DIA_PLT_APPS in reltool Makefile fixes the dialyzer warning.
(Unknown types: beam_lib:chunkid/0 (reltool.hrl:31:45))
BTW, since it is a type in a public API, shouldn't it have been exported already?
Or should i just put in a local type which also resolves to nonepty_string() and only refer to it in text?

garazdawi:
> exporting the type in beam_lib

Are you sure this did not work? I tried it locally and it worked for me.

> BTW, since it is a type in a public API, shouldn't it have been exported already?

Not all documented types are public. Some are just in the docs for documentation purposes.

attah:
Thanks for looking into it. I must have done something wrong (rebuild those two apps and run-dialyzer). I read you as the Makefile change not being needed, so i omitted that.

tsloughter:
Note also that `Attr` needs to remain for the release to work with `release_handler`. In relx the default for stripping a release is now to keep the attributes: https://github.com/erlware/relx/pull/906

garazdawi:
Thanks!

PR #5939, opened 2022-04-27
SCTP: Fix type specs for inet:getopts/2 and inet:setopts/2

dotsimon:
SCTP Socket options have some weird[^1] semantics that this PR fixes.

Specifically:

- two options are read-only (see `prim_inet:type_opt/2`)
- many options for getopts require a  `#sctp_...` record to specify the association

Previously, the same type specification was used for both getopts and setopts which caused dialyzer to barf on perfectly valid code like this:
```erlang
    inet:getopts(Sock, [{sctp_status, #sctp_status{assoc_id = AssocId}}])
```

[^1]: Weird, as in not like TCP or UDP but perfectly fine for SCTP!

### DIFF
--- a/lib/kernel/doc/src/gen_sctp.xml
+++ b/lib/kernel/doc/src/gen_sctp.xml
@@ -4,7 +4,7 @@
 <erlref>
   <header>
     <copyright>
-      <year>2007</year><year>2021</year>
+      <year>2007</year><year>2022</year>
       <holder>Ericsson AB. All Rights Reserved.</holder>
     </copyright>
     <legalnotice>
@@ -69,6 +69,7 @@
   </description>
 
   <datatypes>
+    <datatype_title>Exported data types</datatype_title>
     <datatype>
       <name>assoc_id()</name>
       <desc>
@@ -79,26 +80,47 @@
       </desc>
     </datatype>
     <datatype>
-      <name name="setoption"/>
+      <name name="option"/>
       <desc>
-        <p>One of the
-          <seeerl marker="#options">SCTP Socket Options</seeerl> used to set an option.</p>
+        <p>
+          One of the
+          <seeerl marker="#options">SCTP Socket Options</seeerl>
+          used to set an option.
+        </p>
       </desc>
     </datatype>
     <datatype>
-      <name name="getoption"/>
+      <name name="option_name"/>
       <desc>
-        <p>One of the
-          <seeerl marker="#options">SCTP Socket Options</seeerl> or name used to get an option.</p>
+        <p>
+          An option name or one of the
+          <seeerl marker="#options">SCTP Socket Options</seeerl>
+          used to get an option.
+        </p>
       </desc>
     </datatype>
     <datatype>
-      <name name="optionval"/>
+      <name name="option_value"/>
       <desc>
-        <p>One of the
-          <seeerl marker="#options">SCTP Socket Options</seeerl>.</p>
+        <p>
+          One of the
+          <seeerl marker="#options">SCTP Socket Options</seeerl>
+          as returned when getting an option.
+        </p>
       </desc>
     </datatype>
+    <datatype>
+      <name>sctp_socket()</name>
+      <desc>
+        <p>Socket identifier returned from
+          <seemfa marker="#open/0"><c>open/*</c></seemfa>.</p>
+        <marker id="exports"></marker>
+      </desc>
+    </datatype>
+  </datatypes>
+
+  <datatypes>
+    <datatype_title>Internal data types</datatype_title>
     <datatype>
       <name name="elementary_option"/>
     </datatype>
@@ -110,14 +132,6 @@
     </datatype>
     <datatype>
       <name name="ro_option"/>
-    </datatype>
-    <datatype>
-      <name>sctp_socket()</name>
-      <desc>
-        <p>Socket identifier returned from
-          <seemfa marker="#open/0"><c>open/*</c></seemfa>.</p>
-        <marker id="exports"></marker>
-      </desc>
     </datatype>
   </datatypes>
 
--- a/lib/kernel/doc/src/gen_sctp.xml
+++ b/lib/kernel/doc/src/gen_sctp.xml
@@ -79,14 +79,37 @@
       </desc>
     </datatype>
     <datatype>
-      <name name="option"/>
+      <name name="setoption"/>
+      <desc>
+        <p>One of the
+          <seeerl marker="#options">SCTP Socket Options</seeerl> used to set an option.</p>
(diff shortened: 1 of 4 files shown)


dotsimon:
On master, the following additional changes would need to be made to accommodate the new connectx API:
```udiff
***************
*** 396,402 ****
        SockAddrs:: [{inet:ip_address(), inet:port_number()} |
                     inet:family_address() |
                     socket:sockaddr_in() | socket:sockaddr_in6()],
-       Opts     :: [option()].
  %%
  connectx_init(S, SockAddrs, Opts) ->
      case do_connectx(S, SockAddrs, Opts) of
--- 405,411 ----
        SockAddrs:: [{inet:ip_address(), inet:port_number()} |
                     inet:family_address() |
                     socket:sockaddr_in() | socket:sockaddr_in6()],
+       Opts     :: [setoption()].
  %%
  connectx_init(S, SockAddrs, Opts) ->
      case do_connectx(S, SockAddrs, Opts) of
***************
*** 411,417 ****
        Socket :: sctp_socket(),
        Addrs :: [inet:ip_address() | inet:hostname()],
        Port :: inet:port_number() | atom(),
-       Opts :: [option()].
  %%
  connectx_init(S, Addrs, Port, Opts) ->
      connectx_init(S, Addrs, Port, Opts, infinity).
--- 420,426 ----
        Socket :: sctp_socket(),
        Addrs :: [inet:ip_address() | inet:hostname()],
        Port :: inet:port_number() | atom(),
+       Opts :: [setoption()].
  %%
  connectx_init(S, Addrs, Port, Opts) ->
      connectx_init(S, Addrs, Port, Opts, infinity).
***************
*** 421,427 ****
        Socket :: sctp_socket(),
        Addrs :: [inet:ip_address() | inet:hostname()],
        Port :: inet:port_number() | atom(),
-       Opts :: [option()],
        Timeout :: timeout().
  %%
  connectx_init(S, Addrs, Port, Opts, Timeout) ->
--- 430,436 ----
        Socket :: sctp_socket(),
        Addrs :: [inet:ip_address() | inet:hostname()],
        Port :: inet:port_number() | atom(),
+       Opts :: [setoption()],
        Timeout :: timeout().
  %%
  connectx_init(S, Addrs, Port, Opts, Timeout) ->

```

dotsimon:
In the documentation I removed the references to the option type because it was only partially expanded
```
setoption() = elementary_option() | record_option()
```
Which did not seem useful, especially since all the options are explicitly defined in the _SCTP Socket Options_ section

dotsimon:
Regarding the automated check:
I don't understand why the "Build Erlang/OTP (documentation)" action fails.
The doc builds ok locally and also in the "Build Erlang/OTP (32-bit) (pull_request)" action.

garazdawi:
It is the xml lint that fails, if you run `make xmllint` on the top you should be able to reproduce it locally. The reason why it fails is because you have removed the option type, but somewhere in the docs it is still being used.

garazdawi:
https://github.com/erlang/otp/blob/master/lib/kernel/doc/src/gen_sctp.xml#L529 this seems to be the place where it is used. It should probably be replaced by `<seeerl marker="#options"`.

lib/kernel/doc/src/gen_sctp.xml
```diff
@@ -79,15 +79,12 @@
       </desc>
     </datatype>
     <datatype>
-      <name name="option"/>
+      <name>setoption()</name>
       <desc>
         <p>One of the
           <seeerl marker="#options">SCTP Socket Options</seeerl>.</p>
       </desc>
     </datatype>
```
garazdawi, line 87: You need to add `getoption` in the list of documented types as the inet docs now refer to it.

lib/kernel/src/inet.erl
```diff
@@ -144,10 +144,13 @@
 -type inet_backend() :: {'inet_backend', 'inet' | 'socket'}.
 
 -type socket_setopt() ::
-        gen_sctp:option() | gen_tcp:option() | gen_udp:option().
+        gen_sctp:setoption() | gen_tcp:option() | gen_udp:option().
+
+-type socket_optval() ::
```
garazdawi, line 149: This type needs to be inlined into the `getopts/2` documentation or documented at the top like `socket_setopt`.

garazdawi (dismissed):

dotsimon:
Is there some way to recreate what the check is doing?
It's kind of frustrating since a local build of the documentation succeeds albeit with some dangling links.

lib/kernel/doc/src/gen_sctp.xml
dotsimon: Ok yes I see that now, and also the optionval (return from inet:getopts) case needs to be fixed.
Although, I'm not sure what the best way to handle this is.
The type specs in gen_sctp now match what the code does.
But they are not so great for automatically generating the doc.
As I mentioned earlier, just to include `setoption` in the xml results in an expansion like this:
```
setoption() = elementary_option() | record_option()
```
Of course I could include all those types but it feels like that would just bloat the doc without really adding anything. Especially since the same would also apply for `getoption()` and `optionval()`

I'd be interested in what approach you suggest. I assume there is not a way to get the docgen to expand certain types?

garazdawi:
I created #5946 to expand on how to do the documentation validation locally.

lib/kernel/doc/src/gen_sctp.xml
garazdawi: This is one of the bad parts of having the typespecs as the source for the documentation. Sometimes you need to make compromises either in the type annotations, or the documentation. I'll leave the recommendations about what to do to the people responsible for inet/sctp as I don't really know my way around these parts.

> I assume there is not a way to get the docgen to expand certain types?

Not that I know of.

lib/kernel/src/inet.erl
dotsimon: I expanded everything out.
It is now consistent with other documentation

dotsimon:
> I created #5946 to expand on how to do the documentation validation locally.

I had already found the github action scripts to help myself. But this is great that it will now be documented for future reference.
Thanks for your time with this PR.

dotsimon:
@garazdawi Hi! Do you need something else from me on this PR?

garazdawi:
I don't know much about the inet/sctp API so you will have to wait for the people that do know for the final review. I only commented here to help out with how to write the correct xml for what you wanted to achieve.

dotsimon:
Your change looks fine to me. I hadn't considered existing users of the gen_sctp types.

No worries about the delay, this is obviously a non-priority improvement.

RaimoNiskanen (approved):

PR #5940, opened 2022-04-28
Fix multi-init record initialization edge case

bjorng:
The compiler would crash when attempting to compile the
following code:

    -record(r, {a}).

    foo() ->
        R = #r{a = [], _ = V = 42},
        {R,V}.

The reason is that when all fields in the record had been explicitly
initialized, the `erl_expands_records` pass would rewrite the code
to:

    foo() ->
        R = {[]},
        {R,V}.

This pull request eliminates that bug by updating `erl_expand_records`
to ensure that the initializing expression is still executed even
when all records fields are explicitly initialized:

    foo() ->
        begin
            V = 42,
            R = {[]}
        end,
        {R,V}.

### DIFF
--- a/lib/compiler/test/record_SUITE.erl
+++ b/lib/compiler/test/record_SUITE.erl
@@ -28,7 +28,8 @@
 	 init_per_testcase/2,end_per_testcase/2,
 	 errors/1,record_test_2/1,record_test_3/1,record_access_in_guards/1,
 	 guard_opt/1,eval_once/1,foobar/1,missing_test_heap/1,
-	 nested_access/1,coverage/1,grab_bag/1,slow_compilation/1]).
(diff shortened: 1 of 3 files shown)


jhogberg (approved):

bootstrap/lib/kernel/ebin/kernel.appup
```diff
@@ -1,7 +1,7 @@
 %% -*- erlang -*-
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 1999-2022. All Rights Reserved.
+%% Copyright Ericsson AB 1999-2021. All Rights Reserved.
```
jhogberg, line 4: Nitpick: maybe skip this file?

bjorng: This commit is here for testing only. I intend to remove it before merging.

PR #5941, opened 2022-04-28
Strengthen tests of relational operators

bjorng:
Strengthen tests of relational operators to better test current
and potential future optimizations done by the JIT.

### DIFF
--- a/erts/emulator/test/op_SUITE.erl
+++ b/erts/emulator/test/op_SUITE.erl
@@ -24,18 +24,18 @@
 
 -export([all/0, suite/0,
          bsl_bsr/1,logical/1,t_not/1,relop_simple/1,relop/1,
-         complex_relop/1,unsafe_fusing/1]).
+         complex_relop/1,unsafe_fusing/1,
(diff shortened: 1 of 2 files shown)


erts/emulator/test/op_SUITE.erl
```diff
@@ -450,6 +457,225 @@ usec_to_seconds(Usec) when is_integer(Usec) ->
     Sec = Usec div 1000000,
     Sec rem 1000000.
 
+range_tests(_Config) ->
+    {MinSmall,MaxSmall} = small_limits(),
+
+    small = range(-1 bsl 64),
```
jhogberg, line 463: I was a bit confused by `small_limits/0` and then seeing that `-1 bsl 64` is somehow `small`. Maybe use `lesser` / `greater` instead?

jhogberg (approved):

erts/emulator/test/op_SUITE.erl
bjorng: Good idea. Changed.

PR #5944, opened 2022-04-28
global fixes

rickard-green:
Fixed inconsistency bugs in `global` due to `nodeup`/`nodedown` messages not being delivered before/after traffic over connections. Also fixed various inconsistency bugs and deadlocks in `global_group`.

As building blocks for these fixes, a new BIF `erlang:nodes/2` has been introduced and `net_kernel:monitor_nodes/2` has been extended.

The `-hidden` and `-connect_all` command line arguments did not work if multiple instances was present on the command line which has been fixed. The new kernel parameter `connect_all` has also been introduced in order to replace the `-connect_all` command line argument.


erts/emulator/beam/dist.c
```diff
@@ -6205,15 +6362,62 @@ erts_monitor_nodes(Process *c_p, Eterm on, Eterm olist)
     if (on != am_true && on != am_false)
 	return THE_NON_VALUE;
 
-    if (is_not_nil(opts_list)) {
+    if (is_map(options)) {
```
garazdawi, line 6365: Could we move some of this code to net_kernel:monitor_nodes? That is something like this:

```erlang
monitor_nodes(Flag, Opts) when is_list(Opts) ->
    case catch process_flag({monitor_nodes, maps:from_list(Opts)#{ info_type => list }}, Flag) of
	N when is_integer(N) -> ok;
	_ -> mk_monitor_nodes_error(Flag, Opts)
    end;
monitor_nodes(Flag, Opts) when is_map(Opts) ->
    case catch process_flag({monitor_nodes, Opts#{ info_type => map }}, Flag) of
	N when is_integer(N) -> ok;
	_ -> mk_monitor_nodes_error(Flag, Opts)
    end;
```

erts/emulator/test/exception_SUITE.erl
```diff
@@ -1029,6 +1029,7 @@ error_info(_Config) ->
          {nif_error, 2},
          {node, [abc]},
          {nodes, [abc]},
+         {nodes, [abc, 17]},
```
garazdawi, line 1032: Some more tests should probably be added here, for example

```suggestion
         {nodes, [abc, 17], [{1, ".*not a valid node type.*"},{2, ".*not a map.*"}]},
         {nodes, [hidden, 17], [{2, ".*not a map.*"}]},
         {nodes, [hidden, #{ 17 => true }], [{2, ".*invalid options in map.*"}]}},
         {nodes, [hidden, #{ connection_id => 17 }], [{2, ".*invalid options in map.*"}]},
         {nodes, [hidden, #{ node_type => 17 }], [{2, ".*invalid options in map.*"}]},
```

erts/emulator/beam/dist.c
```diff
@@ -5887,50 +5946,146 @@ BIF_RETTYPE nodes_1(BIF_ALIST_1)
...
+                if (eni->cid ==  ~((Uint32) 0))
+                    cid = am_undefined;
+                else if (IS_USMALL(0, (Uint) eni->cid))
+                    cid = make_small((Uint) eni->cid);
+                else {
+                    cid = uint_to_big((Uint) eni->cid, hp);
+                    hp += BIG_UINT_HEAP_SIZE;
+                }
+            }
+            map = (flatmap_t *) hp;
```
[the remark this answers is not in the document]
rickard-green: Don't like to use `erts_map_from_ks_and_vs()` since it unconditionally sorts the keys even though I already have provided them sorted, and (even worse) wont share the key tuple if multiple instances of maps with the same keys are created (which is frequent in the nodes/2 case). However, I introduced a `erts_map_from_sorted_ks_and_vs()` which does not have these issues.

erts/emulator/beam/dist.c
```diff
@@ -6426,26 +6646,92 @@ send_nodes_mon_msgs(Process *c_p, Eterm what, Eterm node, Eterm type, Eterm reas
...
+                    nodedown_reason_opt = !0;
+                    map_size++;
+                }
+
+                /* key tuple... */
+                keys = (Eterm *) hp;
+                *keys = make_arityval(map_size);
+                hp += 1 + map_size;
+
+                /* flat map structure... */
```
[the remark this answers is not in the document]
rickard-green: Same as above

erts/emulator/beam/dist.c
rickard-green: Implemented

rickard-green:
Replaced by #6007

PR #5945, opened 2022-04-29
Update distribution protocol doc

sile:
First of all, thank you for providing the detailed doc (https://www.erlang.org/doc/apps/erts/erl_dist_protocol.html) about the Erlang distribution protocol.
I wrote a Rust implementation of this protocol ([erl_dist](https://github.com/sile/erl_dist)) years ago, and recently updated the library to support the latest protocol version. 
During the work, I noticed some tiny mistakes and a lack of explanation in the doc. This PR fixes them.


erts/doc/src/erl_dist_protocol.xml
```diff
@@ -741,7 +741,9 @@ io:format("old/unused name ~ts at port ~p, fd = ~p ~n",
 	      node name by setting flag <seeguide marker="#DFLAG_NAME_ME">
 	      <c>DFLAG_NAME_ME</c></seeguide>. The dynamic node name of
 	      <c>A</c> is supplied at the end of the status message from
-	      <c>B</c>.</p>
+	      <c>B</c>. The host name part of <c>Name</c> sent in the
+              <c>send_name</c> message will be used by node <c>B</c> to
+              generate the dynamic node name.</p>
```
sverker, line 746: I think this should be worded something like "The host name of A which was sent as Name in send_name will be used by node B to generate the full dynamic node name."
And it should be added in send_name that Name must be just the host name (without @) when DFLAG_NAME_ME is set.
That is what dist_util:is_name_ok demands: https://github.com/erlang/otp/blob/96872ae52c4461be1791104a511f52ba40afce48/lib/kernel/src/dist_util.erl#L822

sverker (commented):
Thanks for "testing" the documentation and the feedback.

erts/doc/src/erl_dist_protocol.xml
sile: Thank you for your comment! (those are very important points)
I updated the doc to reflect your review in https://github.com/erlang/otp/pull/5945/commits/c77b25a3ce301043c28e0dd54ddfa0f2c6c61675 and https://github.com/erlang/otp/pull/5945/commits/2b203ad498217502e4ad00c0ba71599dd621a3e1.

sverker:
I pushed an extra commit that rearranged the xml under send_name->named:

sverker:
I'm satisfied, if you also are I suggest prepare for merge by squashing all commits except my last one which contains no textual changes but mostly white space indentation changes.

sile:
Thank you for your review 😄 
Squashed!

PR #5947, opened 2022-04-29
Backport: epmd: don't fail if non-critical interfaces fail to bind

big-r81:
Backporting e8c9e12 (#5762) to v23.

When specifying bind address(es) via ERL_EPMD_ADDRESS or the -address
option, epmd will bind to the IPv4/IPv6 loopback intefaces in addition
to the user-supplied address(es).

However, it is possible that one of the loopback interfaces does not
exist.  For example, if the host system has ipv6 disabled via the
disable_ipv6 sysctl, there will be no loopback interface "::1", and
attmepting to bind to "::1" will fail.

Under these circumstances, log the failure but do not consider the
failure to bind to the loopback interface as fatal.

If any of the user-supplied addresses fail to bind, however, that is
still considered fatal and epmd will terminate.

### DIFF
--- a/erts/epmd/src/epmd_srv.c
+++ b/erts/epmd/src/epmd_srv.c
@@ -239,6 +239,7 @@ void run(EpmdVars *g)
   char socknamebuf[INET_ADDRSTRLEN];
 #endif
   int num_sockets = 0;
+  int nonfatal_sockets = 0;
   int i;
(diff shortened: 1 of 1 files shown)


sverker:
Released in OTP-23.3.4.14 and OTP-24.3.4.

PR #5953, opened 2022-05-02
Replace inferior 35-bit PRNGs, OTP-18011 (again)

RaimoNiskanen:
This PR replaces the small fast classical and bad 35-bit PRNGs `mcg35` and `lcg35` that were added in PR #5836 with a new custom made 59-bit `mwc59`.

In cooperation with Prof. Sebastiano Vigna, I have been trying out a Multiply With Carry PRNG, which is a special form of a Multiplicative Congruential Generator with a safe prime modulus and thanks to a power of 2 multiplier has got a very effective implementation.  Due to the power of 2 multiplier it has a known statistical problem, that can be compensated for by using a scrambler function on the generator state.

Despite the 59 bit state this generator is just as fast as the fastest (`lcg35`) generator for a state update, and when generating for a range 16 bits or less no scrambler is needed.  For range 32 bits or less there is a light weight scrambler that adds 60% to the execution time.  For full range (58 bits) there is a thorough scrambler that makes the total execution time about 2 times the base generator, but this is still twice as fast as the `exsp` generator without the `rand` plug-in framework.

To put some numbers on that from my laptop:
* State update: 4 ns
* Add light scramble: 6.5 ns
* Add instead thorough scramble: 8 ns
* `exsp` direct call: 16 ns


max-au:
Nice one! Yet saving the state is... hard. I wonder if there is something that can be done for such case

RaimoNiskanen:
@max-au: Well, I still cannot think of any other than saving in the process dictionary (adds about 10 ns or 250% (2..3 times)), or to use `erlang:phash2(erlang:unique_integer())`.  The process dictionary solution is actually faster per generated number.

But IIRC you pulled just a few numbers per seed, so seeding was the greater problem, which leaves only phash2 over unique integer, that I can think of.

A BIF solution that handles seeding and stores the state in thread specific data would also solve your problem, I guess, but we (the OTP team) would really like to not have to maintain such a beast...

max-au:
I found an interesting function here: https://github.com/erlang/otp/blob/master/erts/emulator/beam/erl_process.h#L2874=

I wonder if we could use that for seed-less RNG...

PR #5954, opened 2022-05-02
peer: Add detach option to allow peers to remain attached

garazdawi:
Add the option `detach` that can be set to false in order
for the peer process to remain attached to the peer node
and forward any output to stdout or stderr from that node.

This mode is mostly useful when testing the startup and
shutdown of Erlang.

### DIFF
--- a/erts/doc/src/erl_cmd.xml
+++ b/erts/doc/src/erl_cmd.xml
@@ -294,7 +294,7 @@ $ <input>erl \
           <c><![CDATA[-setcookie]]></c>. Use <c><![CDATA[-setcookie]]></c>
           instead.</p>
       </item>
-      <tag><c><![CDATA[-detached]]></c></tag>
+      <tag><marker id="detach"/><c><![CDATA[-detached]]></c></tag>
       <item>
         <p>Starts the Erlang runtime system detached from the system
           console. Useful for running daemons and backgrounds processes. Implies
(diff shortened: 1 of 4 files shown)


max-au:
I had a similar code originally, but looking several more times I figured out that `standard_io` alternative connection covers this use-case as well.

I wonder what isn't covered with `standard_io` connection? Is it because internally `peer` uses `-user` command line argument to start up before any console i/o driver?

garazdawi:
I want to be able to capture things printed by erlang:display at startup/shutdown. From what I’ve seen doing erlang:display will break standard_io? I would be very happy to be wrong about that.

I also want to be able to start non-detached nodes inside tmux/screen I order to test the interactive shell better.

max-au:
> From what I’ve seen doing erlang:display will break standard_io?

The `peer` is designed to specifically support `erlang:display` with standard_io. I use it extensively to debug startup problems, with, say, adding `init_debug` option (see the test case: https://github.com/erlang/otp/blob/master/lib/stdlib/test/peer_SUITE.erl#L410= ). If I haven't added a test case that ensures `erlang:display` works, then it's my bad, I should've, and I also should've documented it.

There are some (hopefully not too strict) limitations on what you cannot output in `standard_io` mode, see here: https://github.com/erlang/otp/blob/master/lib/stdlib/src/peer.erl#L682=
(it's intentionally not documented).

garazdawi:
I did try to use standard_io, but failed so I assumed that it did not work. Turns out that it was only the stderr_to_stdout part that did not work. No matter, was a good exercise to get to know the inner workings of peer.

> I also want to be able to start non-detached nodes inside tmux/screen I order to test the interactive shell better.

I may re-open this PR with some changes when I get closer to being done with this work. When testing the shell, peer cannot use `-user`, so I will have to come up with some alternative there.

PR #5955, opened 2022-05-02
logger: Fix simple logger to work when init process is busy

garazdawi:
When wanting to log during system shutdown, the init process
is not always available to load new modules. So we need to
make sure that the simple logger can handle the fact that it
cannot load the correct formatting modules.

This PR needs #5954 in order to test that the fix works.


erts/emulator/beam/bif.c
```diff
@@ -4240,17 +4240,11 @@ BIF_RETTYPE display_string_1(BIF_ALIST_1)
     if (res != 0 || written != len)
 	erts_exit(ERTS_ERROR_EXIT, "%s:%d: Internal error (%d)\n", __FILE__, __LINE__, res);
     str[len] = '\0';
-    erts_fprintf(stderr, "%s", str);
+    erts_fprintf(stdout, "%s", str);
```
max-au, line 4243: If `stderr_to_stdout` is defined in `peer`, then maybe we can keep the `stderr` here?

garazdawi: I still want to change this so that calls to `erlang:display_string` and `erlang:display` end up in the same place. For instance [this code](https://github.com/erlang/otp/blob/master/lib/kernel/src/logger_simple_h.erl#L183-L198) needs it.

The stderr_to_stdout in peer is because I want to catch the "Crash dump is being written..." text.

I'm a bit torn about whether we want to control stderr via an option or not... do you have an opinion?

Issue #5870, opened 2022-04-07
Storing Client Certificate in TLS 1.3 stateless tickets and configurable Shard/IV


sindrip:
**Feature Proposal and Use Case**
I have a use case where I need to verify Client Certificates and pull some data out of the certificate when a client connects. 

Now I am implementing session resumption with TLS 1.3 and stateless tickets. During session resumption the client certificate is not requested and so the `peer_certificate` field is not populated. My proposal is to store the `peer_certificate` in the `opaque ticket` field in NewSessionTicket message [RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.1)

With the stateless tickets, I am running multiple instances of the same service and I want any of the instances to be able to accept tickets generated by any of them. This is not possible as the `Shard` and `IV` used to encrypt and decrypt the tickets is generated randomly on starting listener in `ssl`. My proposal is to allow the user to configure these values. Note that doing this weakens replay protection of the tickets which can be documented.

**Proof of Concept implementation**
I have made a proof of concept of these two features based on `OTP-24.1.3` implementing the feature, the code changes can be seen here: https://github.com/sindrip/otp/pull/4/files

What would be required to get this feature merged in? (Other than basing off master, adding tests, cleaning up the commits etc.)

**Other implementations**
Note that encoding the certificate in the ticket seems to be in line with the implementation in the standard library of Go: https://go.dev/src/crypto/tls/handshake_server_tls13.go (see line 278)

![image](https://user-images.githubusercontent.com/15222674/162232308-c01807aa-f041-42c0-9cec-0fb0d6be1577.png)

IngelaAndin:
We have been very busy with the OTP-25 deadline. We will start taking a look at this hopefully next week.

u3s:
closing, as #6362 was merged

Issue #5951, opened 2022-05-02
tls_server_session_ticket_SUITE is being skipped


ahovgaard:
**Describe the bug**

The `tls_server_session_ticket_SUITE` test cases are all getting skipped, since the call to `tls_server_session_ticket:start_link/6` is missing the listener socket argument, so it gives "init_per_testcase failed" and the CT report shows something like the following.

```
=== ERROR! init_per_testcase crashed!
	Location: [{tls_server_session_ticket,start_link},
 {tls_server_session_ticket_SUITE,init_per_testcase,74},
 {test_server,do_init_per_testcase,1553},
 {test_server,run_test_case_eval1,1254},
 {test_server,run_test_case_eval,1224}]
	Reason: {undef,
    [{tls_server_session_ticket,start_link,[stateful,1,1,100,undefined],[]},
     {tls_server_session_ticket_SUITE,init_per_testcase,2,
         [{file,"tls_server_session_ticket_SUITE.erl"},{line,74}]},
     {test_server,do_init_per_testcase,2,
         [{file,"test_server.erl"},{line,1553}]},
     {test_server,run_test_case_eval1,6,
         [{file,"test_server.erl"},{line,1254}]},
     {test_server,run_test_case_eval,9,
         [{file,"test_server.erl"},{line,1224}]}]}
```

**To Reproduce**

Simply run the suite and inspect the test report.

```bash
cd lib/ssl
make
make test ARGS="-suite tls_server_session_ticket_SUITE"
```

**Expected behavior**

It is expected that all test cases in the suite are run and pass.

PR #5957, opened 2022-05-03
ssl: Store ClientCert in Stateless SessionTicket

sindrip:
During session resumption with stateless tickets a previously verified
client certificate will not be available in the resumed session.
Storing the certificate in the session ticket allows us to retrieve it
during session resumption handshake.

This PR is related to #5870 but only covers the case of storing the client certificate.

### DIFF
--- a/lib/ssl/src/ssl_cipher.erl
+++ b/lib/ssl/src/ssl_cipher.erl
@@ -1306,12 +1306,20 @@ encrypt_ticket(#stateless_ticket{
                   pre_shared_key = PSK,
                   ticket_age_add = TicketAgeAdd,
                   lifetime = Lifetime,
-                  timestamp = Timestamp
+                  timestamp = Timestamp,
+                  certificate = Certificate
                  }, Shard, IV) ->
-    Plaintext = <<(ssl_cipher:hash_algorithm(Hash)):8,PSK/binary,
+    Plaintext1 = <<(ssl_cipher:hash_algorithm(Hash)):8,PSK/binary,
                    ?UINT64(TicketAgeAdd),?UINT32(Lifetime),?UINT32(Timestamp)>>,
+    CertificateLength = case Certificate of
+        undefined -> 0;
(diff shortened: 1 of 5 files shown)


sindrip:
tls_server_session_ticket_SUITE seems to be skipped, this is now dependent on #5951 (along with work on the tests)

u3s:
please resolve conflicts and push an update to enable test execution on our side

sindrip:
Hi @u3s, I have resolved the conflicts

IngelaAndin:
We will get to this we just want to resolve the other PR regarding stateless tickets first.

sindrip:
Perfectly reasonable, I appreciate the update

sindrip:
Hi @IngelaAndin and @u3s, have you had any time to look at this? 

I just quickly rebased to clear up the conflict

u3s:
we will have a discussion and let you know. should happen this or next week.

u3s (commented):
Finally, we've discussed the PR. See notes below:

1. we would like this behavior to be optional, not default on server side.
   so only interested users are affected
   1. preferably we would like to avoid new option, maybe we could extend
      server_session_tickets() = disabled | stateful | stateless
      with values: stateful_with_cert, stateless_with_cert ?
2. it would be preferred to have equivalent behavior for stateful tickets,
   although having different implementation (cert stored on server side).
3. ssl_session_ticket_SUITE - we would like to have blackbox tests using ssl API
   1. if we lack some test coverage, we might add some whitebox
   2. with new code ssl:peercert/1 should return {ok, Cert} instead of {error,
      no_peercert} when stateless_with_cert or stateful_with_cert
   3. maybe also add instructions in existing tests, for checking behavior with existing code
4. for new option values we would need necessary doc update
5. see also minor comments in code

sindrip:
Thanks for the review, I will try and find the time to implement the requested changes as soon as possible.

sindrip:
This PR can be closed in favor of #6362

PR #5959, opened 2022-05-04
ssl: Use milliseconds for TLS 1.3 session ticket age

### LINKED ISSUE #5943: TLS 1.3 session resumption, comparison of client ticket age in milliseconds and ticket lifetime in seconds
**Describe the bug**
In TLS 1.3, when using stateless session tickets, the session ticket contains a random `ticket_age_add` value and a `ticket_lifetime` (in seconds), see [RFC 8446 Section 4.6.1](https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.1). The client uses the `ticket_age_add` to add to the client-side ticket age, to obscure the age of the ticket, and the client includes this in the `obfuscated_ticket_age` in the `pre_shared_key` extension in the `ClientHello`, when attempting a session resumption.

The `obfuscated_ticket_age` is obtained by taking the client-side ticket age **in milliseconds**  and adding the `ticket_age_add`. From [RFC 8446 Section 4.2.11.1](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.11.1):

> The "obfuscated_ticket_age" field of each PskIdentity contains an obfuscated version of the ticket age formed by taking the age in milliseconds and adding the "ticket_age_add" value that was included with the ticket (see [Section 4.6.1](https://datatracker.ietf.org/doc/html/rfc8446#section-4.6.1)), modulo 2^32. (..) Note that the "ticket_lifetime" field in the NewSessionTicket message is in seconds but the "obfuscated_ticket_age" is in milliseconds.

The problem is that the OTP `ssl` implementation compares the (de-)obfuscated ticket age **in milliseconds** with the ticket lifetime **in seconds**. In `tls_server_session_ticket.erl`:
```erlang
stateless_living_ticket(0, _, _, _, _) ->
    true;
stateless_living_ticket(ObfAge, TicketAgeAdd, Lifetime, Timestamp, Window) ->
    ReportedAge = ObfAge - TicketAgeAdd,
    RealAge = erlang:system_time(second) - Timestamp,
    (ReportedAge =< Lifetime)
        andalso (RealAge =< Lifetime)
        andalso (in_window(RealAge, Window)).
```
With the default session ticket lifetime of 7200 seconds (2 hours), the actual ticket lifetime becomes 7.2 seconds.

**To Reproduce**

Generate some test certificates:
```bash
openssl req -x509 -newkey rsa:4096 \
    -keyout rootCA.key -out rootCA.pem \
    -sha256 -days 365 -nodes \
    -subj '/CN=Test root CA'

openssl genrsa -out server.key 2048
openssl req -new -key server.key -out server.csr \
    -subj "/CN=Test server"

openssl x509 -req -in server.csr \
    -CA rootCA.pem -CAkey rootCA.key \
    -CAcreateserial -out server.pem -days 1000 -sha256
```

`server.erl`:
```erlang
-module(server).

-export([run/0,
         start_trace/0,
         stop_trace/0]).

run() ->
    application:ensure_all_started(ssl),
    LOpts = [{cacertfile, "rootCA.pem"},
             {certfile, "server.pem"},
             {keyfile, "server.key"},
             {versions, ['tlsv1.3']},
             {session_tickets, stateless}],
    {ok, LSock} = ssl:listen(8001, LOpts),
    handle(LSock).

handle(LSock) ->
    {ok, CSock} = ssl:transport_accept(LSock),
    {ok, Sock} = ssl:handshake(CSock),
    {ok, Opts} = ssl:connection_information(Sock, [session_resumption]),
    io:format('Connected with session resumption ~p~n', [Opts]),
    handle(LSock).

start_trace() ->
    dbg:tracer(),
    dbg:p(all, [c]),
    dbg:tpl({tls_server_session_ticket, stateless_living_ticket, '_'}, [{'_',[],[{return_trace}]}]).

stop_trace() ->
    dbg:stop_clear().
```

Start tracing and run the server:
```
1> c(server).
{ok,server}
2> server:start_trace().
{ok,[{matched,nonode@nohost,1},{saved,1}]}
3> server:run().
```

Connect with OpenSSL `s_client`, waiting 10 seconds before trying to reconnect and resuming the session:
```bash
sleep 3s | openssl s_client -tls1_3 -connect localhost:8001 -sess_out sess1.pem -debug
sleep 10s
sleep 3s | openssl s_client -tls1_3 -connect localhost:8001 -sess_in sess1.pem -debug
```

The server shows that the connection failed to use session resumption:
```
Connected with session resumption [{session_resumption,false}]
(<0.123.0>) call tls_server_session_ticket:stateless_living_ticket(1066054369,1066042369,7200,1651161160,undefined)
(<0.123.0>) returned from tls_server_session_ticket:stateless_living_ticket/5 -> false
Connected with session resumption [{session_resumption,false}]
```

If we change to only sleep 2 seconds between the connections, resumption works:
```
Connected with session resumption [{session_resumption,false}]
(<0.123.0>) call tls_server_session_ticket:stateless_living_ticket(394807085,394803085,7200,1651161373,undefined)
(<0.123.0>) returned from tls_server_session_ticket:stateless_living_ticket/5 -> true
Connected with session resumption [{session_resumption,true}]
```

**Expected behavior**
It is expected that the OpenSSL `s_client` can resume the session after 10 seconds, when the default session ticket lifetime is 2 hours. I.e., we should get `{ok, [{session_resumption, true}]} = ssl:connection_information(Sock, [session_resumption])` and the internal `tls_server_session_ticket:stateless_living_ticket/5` should return `true`.

**Affected versions**
Tested on OTP 24.1.3 and OTP 24.3.2. Both versions are affected by this issue.

**Additional context**
I would be happy to submit a PR for this.

### LINKED ISSUE #5951: tls_server_session_ticket_SUITE is being skipped

ahovgaard:
TLS 1.3 session tickets contain a random `ticket_age_add` value and a
`ticket_lifetime` (in seconds). The client adds the `ticket_age_add` to
the ticket age (in milliseconds) and sends this value in the
`obfuscated_ticket_age` field in the `pre_shared_key` extension in the
`ClientHello`, when attempting a session resumption. The
`ticket_lifetime` is in seconds while the `obfuscated_ticket_age` is in
milliseconds. For reference, see RFC 8446 Section 4.2.11.1.

The implementation incorrectly used seconds for the ticket age in both
the client and server code. This means that when using a client which
correctly reports the age in milliseconds, the OTP server would reject
the ticket after only `server_session_ticket_lifetime/1000` seconds.

Fixes #5943 and fixes #5951.


ahovgaard:
Note that this PR also contains the commit from #5952, which was needed to run the unit tests. If that PR gets merged first, I'll rebase and remove the commit from this branch. Otherwise, I can cancel the other PR.

IngelaAndin:
For some reason, the other PR was commented out from our testing. It can happen if you make a new push after the testing label was added or if it causes problems to the build (maybe clashing with some internal testing branch). So we might as well test it here instead and you can cancel the other PR.

ahovgaard:
@IngelaAndin Okay, I will cancel #5952. Then this PR fixes #5951 as well.

IngelaAndin:
Ah, I see now that it was because your branch was making this PR for maint. But this close to OTP-25 we are only concentrating our efforts on master. I changed the target for the PR. But we do not need to rebase it, as it is based further back makes it easier to perhaps let it be patched on OTP-24 if an opportunity arises to piggyback it.

IngelaAndin:
I now see the following failure on several of our test machines. It might be that the test is too sensitive to
the passing of time!? But I have not had time to investigate it properly yet as there are a lot of things to
finish up for the release.
 
*** CT Error Notification 2022-05-07 10:20:55.189 ***🔗
tls_client_ticket_store_SUITE:ticket_obfuscated_age failed on line 83
Reason: {badmatch,false}

ahovgaard:
@IngelaAndin I made the test case a bit less timing sensitive and also added a CT log, which would have made the error easier to identify. I hope this fixes it.

I also added another commit to change the client ticket store time measurement to use monotonic time, which I think makes more sense since this time stamp is internal to the store and is used to compute the ticket age.

dgud:
Should not tls_connection_1_3 and tls_handshake_1_3 be changed to use monotonic time as well?

Some tests also failed with:
tls_client_ticket_store_SUITE:ticket_obfuscated_age failed on line 80
Reason: {badmatch,undefined}

Which I don't understand, since both store_ticket and find_ticket are synchronous calls.

ahovgaard:
I'm not sure. As far as I understand,  in `tls_connection_1_3`, the session ticket is sent to the client when using the `{session_tickets, manual}` option. When the client uses the ticket to resume a session, it is handled in `tls_handshake_1_3` and used to compute the (obfuscated) ticket age. So if we can assume that the client doesn't inspect the `timestamp` field and that the ticket isn't received on one machine and used on another machine, then I think it's correct to use monotonic time for those as well. Do you think it is to correct to make those assumptions?

Regarding the test failure, I don't really understand how this can happen since, as you mention, the calls are synchronous and there should be 800 ms (now 900 ms) of time slack before the ticket expires.

dgud:
Ok, keep it as it in tls_*_1_3.

ahovgaard:
@dgud I updated the `LIFETIME` to 2 seconds and changed to use `timer:sleep/1` instead, as suggested.

I squashed the change into the previous [Use milliseconds for TLS 1.3 session ticket age](https://github.com/erlang/otp/pull/5959/commits/3f881856abc22830878ac0965cba6cabeb44f0a1) commit, in order to keep the list of commits clean. You can see the diff below.

```diff
diff --git a/lib/ssl/test/tls_client_ticket_store_SUITE.erl b/lib/ssl/test/tls_client_ticket_store_SUITE.erl
index a4b72f0026..ebc602c4e8 100644
--- a/lib/ssl/test/tls_client_ticket_store_SUITE.erl
+++ b/lib/ssl/test/tls_client_ticket_store_SUITE.erl
@@ -35,7 +35,7 @@
          ticket_expired/1]).
 
 -define(TICKET_STORE_SIZE, 2).
--define(LIFETIME, 1). % tickets expire after 1 second
+-define(LIFETIME, 2). % tickets expire after 2 second
 
 %%--------------------------------------------------------------------
 %% Common Test interface functions -----------------------------------
@@ -74,7 +74,7 @@ ticket_obfuscated_age(_Config) ->
 
     ok = tls_client_ticket_store:store_ticket(Ticket, CipherSuite, SNI, PSK),
 
-    ct:sleep(100),
+    timer:sleep(100),
 
     {Key, _} = tls_client_ticket_store:find_ticket(self(), [Cipher], [HashAlgo], SNI, undefined),
     [#ticket_data{identity = Identity}] = tls_client_ticket_store:get_tickets(self(), [Key]),
@@ -102,7 +102,7 @@ ticket_expired(_Config) ->
 
     ok = tls_client_ticket_store:store_ticket(Ticket, CipherSuite, SNI, PSK),
 
-    ct:sleep(1500),
+    timer:sleep(?LIFETIME * 1000 + 500),
 
     {undefined, undefined} = tls_client_ticket_store:find_ticket(
                                self(), [Cipher], [HashAlgo], SNI, undefined).
```

lib/ssl/src/tls_server_session_ticket.erl
```diff
@@ -385,10 +385,10 @@ stateless_living_ticket(0, _, _, _, _) ->
 stateless_living_ticket(ObfAge, TicketAgeAdd, Lifetime, Timestamp, Window) ->
     ReportedAge = ObfAge - TicketAgeAdd,
     RealAge = erlang:system_time(second) - Timestamp,
```
u3s, line 387: why you replaced `system_time` with `monotonic_time` in tls_client_ticket_store but left it in `tls_server_session_ticket`?

ahovgaard: I kept `system_time` in `tls_server_session_ticket` because this time stamp is stored in the encrypted stateless ticket and it could potentially be generated in one server instance and used in a different server instance. This doesn't work currently, since the encryption seed would be different between instances, but I have another issue that would make it possible (#5962). It doesn't make sense to compare monotonic times between difference BEAM instances. Hence, I prefer to keep `system_time` with those things in mind.

lib/ssl/src/tls_server_session_ticket.erl
```diff
@@ -385,10 +385,10 @@ stateless_living_ticket(0, _, _, _, _) ->
...
-    (ReportedAge =< Lifetime)
+    (ReportedAge =< Lifetime * 1000)
```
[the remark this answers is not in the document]
ahovgaard: I think it doesn't matter, since this is stored in the encrypted ticket and only visible to the server, so the unit of the time stamp generated in line 329 just needs to match the time stamp in line 387, when computing the real ticket age. These are both in seconds and the elapsed time is compared to the `Lifetime`, which is also in seconds.

dgud:
Thanks the test run cleanly now.

Could you rebase this onto OTP-24.3.4? 
If I do it we loose the github pull requests reference.

ahovgaard:
@dgud Thanks, that's great to hear.

Sure, I'm trying to do `git rebase OTP-24.3.4` on my branch, but I'm getting a number of conflicts in various files that I didn't touch in my commits (`make/otp_version_ticket`, various xml files in `lib/kernel/doc`, etc.). I suspect this is because the branch was originally created from `maint` and @IngelaAndin changed the target branch to `master`. Could you offer some advice on what would be the easiest way to go about this rebase?

IngelaAndin:
The target branch does not matter what matters is from where it is based and it is probably based on a maint that includes things that mismatch with  OTP-24.3.4. 
`git rebase --onto=OTP-24.3.4 HEAD~3  ` should do it.

dgud:
Does `git rebase --onto OTP-24.3.4 399d2ea589a4492b ssl/fix-ticket-age` work?

ahovgaard:
Thanks, it worked with `git rebase --onto=OTP-24.3.4 HEAD~3`.

PR #5966, opened 2022-05-05
Cons/features/documentation/otp 18049

lisztspace:
Documentation for features support,

### DIFF
--- a/lib/compiler/doc/src/compile.xml
+++ b/lib/compiler/doc/src/compile.xml
@@ -4,7 +4,7 @@
 <erlref>
   <header>
     <copyright>
-      <year>1996</year><year>2021</year>
+      <year>1996</year><year>2022</year>
       <holder>Ericsson AB. All Rights Reserved.</holder>
     </copyright>
     <legalnotice>
@@ -225,6 +225,26 @@
             reduce the paths in stack traces to the module name alone.
 	    This option will make it easier to achieve reproducible builds.
 	    </p>
+          </item>
+
+          <tag><c>{feature, Feature, enable | disable}</c></tag>
+          <item>
+            <p><marker id="feature-option"/>
+            Enable (disable) the <seeguide
+            marker="system/reference_manual:features#features">feature</seeguide>
+            <c>Feature</c> during compilation.  The special feature
+            <c>all</c> can be used to enable (disable) all features.
+	    </p>
+            <note>
+              <p><marker id="feature-directive"/>
+              A feature can also be enabled (disabled) using the
+              <c>-feature(Feature, enable | disable).</c> module attribute.
+              Note that this directive can only be present in a prefix
+              of the file, before exports and function definitions.
+              This is the preferred method of enabling and disabling
+              features, since it is a local property of a module.
+              </p>
+            </note>
           </item>
 
           <tag><c>makedep</c></tag>
@@ -653,6 +673,18 @@ module.beam: module.erl \
 	      in functional objects or list comprehensions with the same 
 	      name as some already defined variable. Default is to 
 	      emit warnings for such variables.</p>   
+          </item>
+
+	  <tag><c>warn_keywords</c></tag>
+          <item>
+	    <p><marker id="warn-keywords"/>
+              Emits warnings when the code contains atoms that are
+            used as keywords in some <seeguide
+            marker="system/reference_manual:features#features">feature</seeguide>.
+            When the feature is enabled, any occurrences will lead to
+            a syntax error.  To prevent this, the atom has to
+            be renamed or quoted.
+            </p>
           </item>
 
 	  <tag><c>nowarn_unused_function</c></tag>
--- a/system/doc/reference_manual/features.xml
+++ b/system/doc/reference_manual/features.xml
@@ -1,0 +1,218 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!DOCTYPE chapter SYSTEM "chapter.dtd">
+
+<chapter>
+  <header>
+    <copyright>
+      <year>2022</year><year>2022</year>
+      <holder>Ericsson AB. All Rights Reserved.</holder>
+    </copyright>
+    <legalnotice>
+      Licensed under the Apache License, Version 2.0 (the "License");
+      you may not use this file except in compliance with the License.
+      You may obtain a copy of the License at
+
+          http://www.apache.org/licenses/LICENSE-2.0
+
+      Unless required by applicable law or agreed to in writing, software
+      distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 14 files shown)


lib/compiler/doc/src/compile.xml
```diff
@@ -655,6 +675,18 @@ module.beam: module.erl \
 	      emit warnings for such variables.</p>   
           </item>
 
+	  <tag><c>warn_keywords</c></tag>
+          <item>
+	    <p><marker id="warn-keywords"/>
+              Emits warnings when the code contains atoms that are
+            use as keywords in some <seeguide
```
garazdawi, line 682:
```suggestion
            used as keywords in some <seeguide
```

system/doc/reference_manual/features.xml
```diff
@@ -0,0 +1,204 @@
...
+    <title>Features</title>
+    <prepared></prepared>
+    <docno></docno>
+    <date></date>
+    <rev></rev>
+    <file>features.xml</file>
+  </header>
+  <p>
+    <marker id="features"/>
+    Introduced in OTP25, Erlang has the concept of selectable features.
```
garazdawi, line 34:
```suggestion
    Introduced in OTP 25, Erlang has the concept of selectable features.
```

system/doc/reference_manual/features.xml
```diff
@@ -0,0 +1,204 @@
...
+    selectable.
+  </p>
+
+  <section>
+    <title>Life cycle of features</title>
+    <p>A feature is in one of four possible states:</p>
+    <taglist>
+      <tag>Experimental</tag>
+      <item>Initial state, meant for try out and collecting
+      feedback.  The feature can be enabled, but is disabled by
```
garazdawi, line 65:
```suggestion
      <item>The initial state, is meant for trying out and collecting
      feedback.  The feature can be enabled but is disabled by
```

system/doc/reference_manual/features.xml
```diff
@@ -0,0 +1,204 @@
...
+      can change state only in connection with a release.
+    </p>
+    <p>
+      A feature may be in the approved state for several releases.
+    </p>
+    <table>
+      <row>
+        <cell>State</cell>
+        <cell>Default</cell>
+        <cell>Controllable</cell>
```
garazdawi, line 90:
```suggestion
        <cell>Configurable</cell>
```
?

system/doc/reference_manual/features.xml
```diff
@@ -0,0 +1,204 @@
...
+      marker="erts:erlc#disable-feature"><c>-disable-feature</c></seecom>
+      can be used to enable or disable individal features.</item>
+      <tag>Compiler options</tag>
+      <item>The compiler option <seeerl
+      marker="compiler:compile#feature-option"><c>{feature,
+      &lt;feature&gt;, enable|disable}</c></seeerl> can be used either
+      as a <c>+&lt;term&gt;</c> option to <c>erlc</c>, in the options
+      argument to function in the <c>compile</c> module, or in the
+      <c>-compile(..)</c> attribute in a file.</item>
+      <tag>The feature directive</tag>
```
garazdawi, line 148:
```suggestion
      <tag>The feature attribute</tag>
```

system/doc/reference_manual/features.xml
```diff
@@ -0,0 +1,204 @@
...
+      To allow for conditional compilation during transitioning of a
+      code base and/or trying out experimental features <seeguide
+      marker="system/reference_manual:macros#predefined-macros">feature</seeguide>
+      <c>predefined macros</c> <c>?FEATURE_AVAILABLE(Feature)</c> and
+      <c>?FEATURE_ENABLED(Feature)</c> are available.
+    </p>
+  </section>
+
+  <section>
+    <title>Information about Existing Features</title>
```
garazdawi, line 181: I'm wondering if there should be a place I'm the documentation where all the experimental features are listed? Maybe we don't need to keep the permanent or rejected features there, but the experimental and accepted might be nice to have? If nothing else just links to the actual documentation of the features, i.e. to the maybe feature now.

system/doc/reference_manual/macros.xml
```diff
@@ -134,7 +134,8 @@ bar(X) ->
 
   <section>
     <title>Predefined Macros</title>
-    <p>The following macros are predefined:</p>
+    <p><marker id="predefined-macros"/>
```
garazdawi, line 137: The title should auto-generate a marker with that name, so it should not be needed.

system/doc/reference_manual/features.xml
lisztspace: Attributes have only one argument, while the `-feature` directive has two, although the documentation mentions other constructs, e.g., `-record(..)` sharing the same syntax as attributes.  The documentation on the preprocessor mentions directives.  I see this more akin to a directive since it has a direct affect will/can affect the parsing of the file.

system/doc/reference_manual/features.xml
lisztspace: I agree.  Where?  In the documentation for `maybe` it just says that it is a experimental, but it might be a good idea to have a more "formal" marking of it as well.  Just having _since_ is not really enough.  I added a link to this general section from it, though.

system/doc/reference_manual/macros.xml
lisztspace: Ah, good to know!

system/doc/reference_manual/features.xml
garazdawi: Yes, now I see it. Use directive.

system/doc/reference_manual/features.xml
garazdawi: The best place I can think of is at the end of this page in the reference manual.

lisztspace: Added now, with a reference to the sole existing feature.  Ideally, this should be generated automagically for each release.  It could be done using the functions in `erl_features`, but not enough time for that now.

system/doc/reference_manual/features.xml
```diff
@@ -0,0 +1,218 @@
...
+    <title>Existing Features</title>
+    <p>
+      The following configurable features exist:
+    </p>
+    <taglist>
+      <tag><c>maybe_expr</c> (experimental)</tag>
+      <item>
+        Implementation of the <seeguide
+        marker="expressions#maybe"><c>maybe</c></seeguide> expression
+        proposed in EEP49.</item>
```
garazdawi, line 215:
```suggestion
        proposed in <url href="https://www.erlang.org/eeps/eep-0049">EEP 49</url>.</item>
```

system/doc/reference_manual/features.xml
garazdawi: Great! and yes, it should indeed be automated.

PR #5969, opened 2022-05-06
Efficiency caching callbacks

essen:
Followup to https://github.com/erlang/otp/pull/5831#issuecomment-1118371747

Make it more explicit that caching would be benefitial for callback modules.

Also fixes a small error at the same time (Module:function is changed into apply, not apply into Module:function, I believe).

### DIFF
--- a/system/doc/efficiency_guide/functions.xml
+++ b/system/doc/efficiency_guide/functions.xml
@@ -158,22 +158,23 @@ explicit_map_pairs(Map, Xs0, Ys0) ->
   <section>
     <title>Function Calls</title>
 
-    <p>This is an intentionally rough guide to the relative costs of
-    different calls. It is based on benchmark figures run on
-    Solaris/Sparc:</p>
+    <p>This is a rough hierarchy of the performance of the
+    different types of function calls:</p>
 
     <list type="bulleted">
     <item>Calls to local or external functions (<c>foo()</c>, <c>m:foo()</c>)
     are the fastest calls.</item>
 
     <item>Calling or applying a fun (<c>Fun()</c>, <c>apply(Fun, [])</c>)
-    is about <em>three times</em> as expensive as calling a local
-    function.</item>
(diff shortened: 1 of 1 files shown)


jhogberg:
Thanks for the PR!

> Also fixes a small error at the same time (Module:function is changed into apply, not apply into Module:function, I believe).

Both are `apply/3`, yes, but some prehistoric version used to compile the former into the fixed-arity apply instruction and the latter into the variable one, so it's not entirely wrong to say it "translates the latter into the former" now.

It's quite confusing with modern eyes though, so I think the section should be reworded to just say that apply with a fixed number of arguments is faster than apply with a variable number.

Come to think of it, the section on the relative costs of calls should be updated too. Fun calls are now only a few percent slower than external calls (the difference comes from the type test, which is skipped if the compiler has figured it out it's a fun with matching arity), and `apply/3` is also a lot quicker though I don't know the ballpark off the top of my head. Feel free to update it with modern benchmarks if you want. :)

essen:
I will reword the apply stuff and remove the exact benchmark results only keeping the precedence with some of the details you have listed. I do not think new benchmark info will stand the test of time with all the work that's being done on the compiler and everything.

essen:
Updated!

essen:
Test failure is because GitHub is having trouble right now. Will retry later.

essen:
Pushed an XML error fix.

jhogberg:

vans163:
I know this is merged but randomly reading and a big difference between local and `my_func()` vs `Module:my_func()` call is the `my_func()` call if the module hotloads will keep calling the stale version (until its purged and proc will die), vs latter will call into the newly loaded module. I feel this is important pitfall that often happens.

essen:
Yes of course but that has nothing to do with performance or efficiency. It is documented in https://www.erlang.org/doc/reference_manual/code_loading.html#code-replacement

PR #5970, opened 2022-05-06
Fix peer options peer_down=>crash with connection=>0

sverker:
When calling `peer:start_link` with option `peer_down=>crash`, the controlling process should exit abnormally if peer node disconnects. This did not work with alternative `connection=>0` option.

### DIFF
--- a/lib/stdlib/test/peer_SUITE.erl
+++ b/lib/stdlib/test/peer_SUITE.erl
@@ -21,6 +21,7 @@
 -export([
     dist/0, dist/1,
     peer_down_crash/0, peer_down_crash/1,
+    peer_down_crash_tcp/1,
     peer_down_continue/0, peer_down_continue/1,
     peer_down_boot/0, peer_down_boot/1,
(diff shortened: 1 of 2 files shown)


lib/stdlib/test/peer_SUITE.erl
```diff
@@ -154,6 +156,30 @@ peer_down_crash(Config) when is_list(Config) ->
...
+    unlink(Peer),
+    MRef = monitor(process, Peer),
+    %% Make the node go down
+    ok = erpc:cast(Node, erlang, halt, [0]),
+
+    %% since two-way link is requested, it triggers peer to stop
+    receive
+        {'DOWN', MRef, process, Peer, tcp_closed} ->
+            ok
+    after 2000 ->
```
max-au, line 178: You may want to bump this to 5 sec, debug build may sometimes be a bit too slow on that path.

max-au (approved):
Thanks! Likely it worked with standard_io and distribution, and I missed the TCP part.

garazdawi (approved):

PR #5976, opened 2022-05-09
Fix spec for peer:start_link/0

sverker:
`peer:start_link/0` returns 3-tuple `{ok,Pid,Node}`.

Other peer issues:
- A lot of functions accept type `Dest :: gen_statem:server_ref()` as identifier of the peer node to communicate with. But nowhere is a  `gen_statem:server_ref()` returned. Why? Whether the controlling process is a gen_statem server or not, seems to me as an implementation issue. 
- I carelessly wrote `peer:start().` in the shell and got strange ERROR REPORT followed by a _frozen shell prompt_. It turns out that `start/0` is not part of the API but must be exported to satisfy the user.erl I/O interface. This is unfortunate, as it's an easy lets-try-this mistake to do.

### DIFF
--- a/lib/stdlib/src/peer.erl
+++ b/lib/stdlib/src/peer.erl
@@ -174,7 +174,7 @@ random_name(Prefix) ->
 %% @doc Starts a distributed node with random name, on this host,
 %%      and waits for that node to boot. Returns full node name,
 %%      registers local process with the same name as peer node.
--spec start_link() -> {ok, node()} | {error, Reason :: term()}.
+-spec start_link() -> {ok, pid(), node()} | {error, Reason :: term()}.
 start_link() ->
     start_link(#{name => random_name()}).
 


sverker:
@max-au @rickard-green I had some concerns above.

max-au (approved):
The only reason to return `gen_statem:server_ref()` was lack of a common spec somewhere in `gen` or `gen_server`. I vaguely remember some PR that was exposing it, but it surely wasn't a part of OTP codebase by the time I worked on the `peer`. If this PR has been merged, I can change `peer` to accept that instead of `server_ref` from `gen_statem`.

And for `peer:start`, yes, it might be better to completely remove reliance on the undocumented `-user` command line argument. The reason it's there is simple, this code needs to be executed as early in boot sequence as possible, to be able to pass the stdio output to the origin node. Using `-s` or `-eval` command line argument won't work, it's too late in the sequence.

garazdawi:
> And for peer:start, yes, it might be better to completely remove reliance on the undocumented -user command line argument. The reason it's there is simple, this code needs to be executed as early in boot sequence as possible, to be able to pass the stdio output to the origin node. Using -s or -eval command line argument won't work, it's too late in the sequence.

The nice thing about using `user` is that it exists in previous versions of Erlang. But having a command-line argument that can be used to run some code before kernel is started could be generally useful I think.

Regarding peer:start/0, could we add a check that the caller of that function is the `user_sup` supervisor? For instance, if we check that `proc_lib:initial_call(self()) == {supervisor_bridge,user_sup,1}`. It is a little brittle, but it should work.

max-au:
> The nice thing about using user is that it exists in previous versions of Erlang
On the other hand it introduces non-trivial interaction with `-noinput` argument (and it also breaks if user specifies extra `-user` arguments). Existing logic in user_sup takes the last one, IIRC.

I'll come up with a PR that fixes spec (I found that gen_server now also exports server_ref) and provides some mechanism to detect whether it's called when starting user, or later. I think less brittle would be to check whether there is already registered `user` process (`whereis(user) =/= undefined`).

sverker:
> I'll come up with a PR that fixes spec (I found that gen_server now also exports server_ref)

Why not specify the `Dest` arguments as `pid()`. The start functions returns the `pid()` of the "controlling process". Isn't that enough? Why bring the type of server interface, that the controlling process happens to implement, into the peer API. To me that seems like implementation details that may even change in the future without having to affect the users of peer.

PR #5977, opened 2022-05-10
[features] Don't warn for quoted atoms being keywords

lisztspace:


lib/stdlib/src/epp.erl
```diff
@@ -627,7 +627,8 @@ init_server(Pid, FileName, Options, St0) ->
 			 path=Path, location=AtLocation, macs=Ms1,
 			 default_encoding=DefEncoding,
                          erl_scan_opts =
-                             [{reserved_word_fun, ResWordFun}],
+                             [text,
```
michalmuskala, line 630: Always turning this on might significantly increase memory use of the parser (especially that EEP is copying stuff a lot between processes), which in turn might have an impact on performance of the compiler. Was this impact measured?

lisztspace: Yes, we are aware that this might be a problem.  This PR is just an initial test run to see how much of the tests break.  It turns out that there were fewer than expected, but this rather simple change might have an affect on other areas as well, e.g., parse transforms not being prepared to see full `erl_anno:anno()` in the tokens.  I'm working on another approach that will be more selective in including the original text for atoms.  This check for atoms colliding with keywords for future features is done in `erl_lint`, so we need that information there.

lisztspace: Back to this after ~~EUC~~ `Code BEAM Europe`, I have now pushed a version with selective enabling of when to keep the original string (so it can be used in `erl_lint`).  Now keeping only for quoted atoms that are keywords in some feature.  Comparing time and memory has the latest implementation slightly faster and similar in memory than what is present in OTP25.0, while being more precise.  The previous implementation on this branch was slightly slower and much worse in terms of memory (but that was expected).

PR #5978, opened 2022-05-10
Add type peer:server_ref()

sverker:
To replace gen_statem:server_ref().

I gave peer its own `server_ref()` type, currently only specified as `pid()` to
- not expose current gen_server implementation unnecessarily.
- make it expandable in future independent of used implementation.

@max-au Thoughts?

### DIFF
--- a/lib/stdlib/doc/src/peer.xml
+++ b/lib/stdlib/doc/src/peer.xml
@@ -247,6 +247,14 @@
 
   <datatypes>
     <datatype>
+      <name name="server_ref"/>
+      <desc>
+        <p>
+	  Identifies the controlling process of a peer node.
+	</p>
+      </desc>
+    </datatype>
+    <datatype>
       <name name="start_options"/>
       <desc>
         <p>
--- a/lib/stdlib/src/peer.erl
+++ b/lib/stdlib/src/peer.erl
@@ -59,6 +59,12 @@
          send/3
         ]).
 
+-export_type([server_ref/0]).
+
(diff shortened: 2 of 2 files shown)


max-au:
I was going to simply replace `gen_statem:server_ref` with `gen_server:server_ref` (and have a PR baking with the other change too, will submit soon).
My reasoning is:
- `peer` works with everything `gen:call` works with (and I actually use it in many of my tests, this way: `{ok, Pid, Node} = peer:start_link(), register(Node, Pid), <...>, peer:call(Node, mod, fun, [Arg]).` to avoid passing both Node and Pid to all inner functions (originally I even had that as a part of `peer` API, but my belief is, the module is the best not when you can't add anything to it, but when you can't remove anything)
- many if not most other `gen_server` implementations would need to follow this very same scheme, because they also return `{ok, Pid}` from start_link, but also accept calls by registered or a global name

On the other side, I understand that reusing `gen_server` type leaks a bit about internal implementation. I personally don't think it's a big problem.

max-au:
Here is my take at reported problems: #5980

sverker:
Sorry, I should have said that the code freeze for OTP 25.0 has passed and we only accept the must urgent fixes at this point. In this case the incorrect server_ref type. The peer:start/0 problem can wait for 25.1.

I'll make the call to take this PR for 25.0. It _removes_ the API dependency to gen_server and is expandable.

max-au (approved):

PR #5980, opened 2022-05-10
[peer] implement peer:start/0 and improve specs

max-au:
Follow-up for #5976:
 - change the spec from gen_statem:server_ref to gen_server:server_ref (original implementation was based on gen_statem)
 - allow start/0 to be used similar to start_link/1
 - update documentation


max-au:
- rebased to master
- removed duplicate `server_ref` type

max-au:
@sverker shall I just close this one? It would be great to remove the extra type (`server_ref()`), but now that it's been part of the official API, it could be a bit harder than just cleanup/removal.

PR #5982, opened 2022-05-11
Add option to configure encryption seed for TLS 1.3 stateless tickets

### LINKED ISSUE #5962: ssl: Configurable seed for TLS 1.3 stateless session tickets encryption

The TLS 1.3 stateless session tickets in OTP are not truly stateless in the sense that the listener's session tickets tracker process has state (the IV and shard seed) needed to encrypt/decrypt the session tickets. This seed is generated on calling `ssl:listen/2`, with option `{session_tickets, stateless}`, when starting the session tickets tracker (`tls_server_session_ticket`) for the listener.

`tls_server_session_ticket.erl`:
```erlang
inital_state([stateless, Lifetime, _, MaxEarlyDataSize, undefined]) ->
    #state{nonce = 0,
           stateless = #{seed => {crypto:strong_rand_bytes(16), 
                                  crypto:strong_rand_bytes(32)},
                         window => undefined},
           lifetime = Lifetime,
           max_early_data_size = MaxEarlyDataSize
          };
```

This prevents using stateless session resumption across multiple server instances.

The server session ticket encryption seed could be made configurable, e.g., by providing the seed shard and IV in the `ssl:listen/2` options:
```erlang
IV = crypto:strong_rand_bytes(16),
Shard = crypto:strong_rand_bytes(32),
ssl:listen(Port, [{session_tickets, stateless}, {stateless_session_tickets_seed, {IV, Shard}}, ...]).
```

It was considered making the seed configurable for the `ssl` application, however this would mean that all `ssl` listeners within an application would have the same seed, which may not be what you would want.

**Additional context**

Suggestions for alternative solutions to achieving the same goal would be appreciated.

I will begin working on a draft PR implementing the solution described above.

ahovgaard:
This enables TLS 1.3 stateless session tickets to work across multiple
server instances by allowing the user to configure the same encryption
seed in each instance, through a newly added `stateless_tickets_seed`
ssl server option.

Closes #5962.


u3s:
current status update
in general the code looks fine, but we're wondering/checking 2 more areas:
1. adding some warning note about consequences of re-using random values with multiple server instance on different machines (such as anti_replay will not work if bloom filter structure is not shared for example, any others? relevance of time synchronization between machines/zones?)
2. RFC 8446 contains such description which relates to startup/restart of server (considering single server instance):
> When implementations are freshly started, they SHOULD reject 0-RTT as
> long as any portion of their recording window overlaps the startup
> time. Otherwise, they run the risk of accepting replays which were
> originally sent during that period.
Today our bloom filter is accepting tickets before 1st rotation of data structure. Even though, when generating random values (as today) for a listen socket, we will reject a ticket from previous incarnation of ssl server.
However I wonder if with your option used we should not improve bloom filter implementation as well. So that when same randoms are used before and after restart, we have some "warm-up" period lasting for 1 Window (?) when we treat all tickets as used.
Such a "warm-up" would assure ticket received was create by current ssl server incarnation.

lib/ssl/test/ssl_session_ticket_SUITE.erl
```diff
@@ -1118,6 +1121,69 @@ early_data_basic_auth(Config) when is_list(Config) ->
...
+    [{doc, "Test session resumption with session tickets, resuming on different server"}].
+stateless_multiple_servers(Config) when is_list(Config) ->
+    ClientOpts0 = ssl_test_lib:ssl_options(client_rsa_verify_opts, Config),
+    ServerOpts0 = ssl_test_lib:ssl_options(server_rsa_verify_opts, Config),
+    {ClientNode, ServerNode, Hostname} = ssl_test_lib:run_where(Config),
+
+    Seed = crypto:strong_rand_bytes(64),
+
+    %% Configure session tickets
+    ClientOpts = [{session_tickets, auto}, {log_level, debug},
```
u3s, line 1134: please remove  `{log_level, debug},` from proposed test suite.
I've cleaned it on maint/master already.

u3s:
#6055 is merged

lib/ssl/test/ssl_session_ticket_SUITE.erl
ahovgaard: Done.

ahovgaard:
> current status update
> in general the code looks fine, but we're wondering/checking 2 more areas:
> 
> 1. adding some warning note about consequences of re-using random values with multiple server instance on different machines (such as anti_replay will not work if bloom filter structure is not shared for example, any others? relevance of time synchronization between machines/zones?)

Are you thinking about adding a warning to the `ssl` documentation (`ssl.xml`) together with the proposed `stateless_tickets_seed` option? I think that would make sense. Those are the two main issues that I can think of, as well.

> 2. RFC 8446 contains such description which relates to startup/restart of server (considering single server instance):
> 
> > When implementations are freshly started, they SHOULD reject 0-RTT as
> > long as any portion of their recording window overlaps the startup
> > time. Otherwise, they run the risk of accepting replays which were
> > originally sent during that period.
> 
> Today our bloom filter is accepting tickets before 1st rotation of data structure. Even though, when generating random values (as today) for a listen socket, we will reject a ticket from previous incarnation of ssl server. However I wonder if with your option used we should not improve bloom filter implementation as well. So that when same randoms are used before and after restart, we have some "warm-up" period lasting for 1 Window (?) when we treat all tickets as used. Such a "warm-up" would assure ticket received was create by current ssl server incarnation.

Yes, I think it makes sense to add this "warm-up" period. Do you want me to do that in this PR or can it be deferred to another PR?

> 
> I've added some tests in #6055 `ticket_reuse_anti_replay_server_restart` test will probably start failing if you combine it with option proposed ... maybe you can confirm/check that ?

The tests pass as they are right now, but `ticket_reuse_anti_replay_server_restart` does indeed fail when extended to configure the same ticket seed before and after the server restart. With the warm-up period added, we should be able to make it pass using the proposed option as well.

u3s:
> > 1. adding some warning note about consequences of re-using random values with multiple server instance on different machines (such as anti_replay will not work if bloom filter structure is not shared for example, any others? relevance of time synchronization between machines/zones?)
> 
> Are you thinking about adding a warning to the `ssl` documentation (`ssl.xml`) together with the proposed `stateless_tickets_seed` option? I think that would make sense. Those are the two main issues that I can think of, as well.

Yes, please - to make option users aware/remind of risks associated with it. Maybe use `<warning>` tag.

> > Today our bloom filter is accepting tickets before 1st rotation of data structure. Even though, when generating random values (as today) for a listen socket, we will reject a ticket from previous incarnation of ssl server. However I wonder if with your option used we should not improve bloom filter implementation as well. So that when same randoms are used before and after restart, we have some "warm-up" period lasting for 1 Window (?) when we treat all tickets as used. Such a "warm-up" would assure ticket received was create by current ssl server incarnation.
> 
> Yes, I think it makes sense to add this "warm-up" period. Do you want me to do that in this PR or can it be deferred to another PR?
> 
> > I've added some tests in #6055 `ticket_reuse_anti_replay_server_restart` test will probably start failing if you combine it with option proposed ... maybe you can confirm/check that ?
> 
> The tests pass as they are right now, but `ticket_reuse_anti_replay_server_restart` does indeed fail when extended to configure the same ticket seed before and after the server restart. With the warm-up period added, we should be able to make it pass using the proposed option as well.

I think it would be preferred if this PR would include bloom filter adjustment, because after adding encryption seed option bloom filter missing implementation will be potentially exposed. With random being generated upon socket creation (before this PR), missing warm-up is not a problem as tickets from previous instance of server will not be usable.

This would be nice to have (pretty easy to implement I guess):
1. adjustment of `tls_bloom_filter` module - so it rejects all ClientHellos before it rotates twice since startup (2*WindowSize must pass)
2. new test case based on `ticket_reuse_anti_replay_server_restart` using encryption seed option

What will be more tricky is what happens with legacy tests after adding warm-up to `tls_bloom_filter`. I haven't got an idea on preferred solution for that one yet. But maybe you could add parts above first - and then we could see how big the problem is and discuss how we could solve it. Preferably we would like to avoid too many, unnecessary sleep instructions in tests ... maybe warm-up should work only when encryption option is specified ... I'm not sure yet. I need to think about. @dgud, @IngelaAndin - maybe you have some ideas?

ahovgaard:
> What will be more tricky is what happens with legacy tests after adding warm-up to `tls_bloom_filter`. I haven't got an idea on preferred solution for that one yet. But maybe you could add parts above first - and then we could see how big the problem is and discuss how we could solve it. (..)

@u3s Please review the pushed changes. As expected, the added warm-up breaks a a few of the test cases in `ssl_session_ticket_SUITE` at least.

lib/ssl/src/tls_server_session_ticket.erl
```diff
@@ -148,20 +155,23 @@ format_status(_Opt, Status) ->
...
     erlang:send_after(Window * 1000, self(), rotate_bloom_filters),
     #state{nonce = 0,
            stateless = #{bloom_filter => tls_bloom_filter:new(K, M),
-                         seed => {crypto:strong_rand_bytes(16),
-                                  crypto:strong_rand_bytes(32)},
+                         %% Warm up for 2*WindowSize to ensure tickets from a
+                         %% previous instance of the server (before a restart)
+                         %% cannot be reused, if the ticket encryption seed is
+                         %% reused
+                         warm_up_windows_remaining => 2,
```
u3s, line 173: can you make initial value 0 for legacy runs (when new option is not specified)?
and 2 for flows with hardcoded random - i.e. new option is specified?

we discussed it, and think this is preferred. not only it solves test case issues but also make PR not changing legacy flows.

ahovgaard: Sure, makes sense. Done.

ahovgaard:
@u3s  I'm not sure what went wrong with the PR pipeline. Some build tasks failed with `write /dev/stdout: no space left on device`, while the `ssl` CT tests failed the test case `openssl_session_ticket_SUITE:openssl_server_basic/1`. However, this test case shouldn't be affected by the changes in this PR since the test uses the OpenSSL `s_server` and not the OTP `ssl` server. Besides, with the recent change, existing functionality shouldn't change unless the newly added option is specified.

ahovgaard:
@u3s Would you mind taking another look at this PR?

u3s:
sorry for holiday period related delay. we will have a look on this soon.

u3s:
thanks for contribution!

PR #5983, opened 2022-05-11
Extend timer tests

juhlig:
This PR introduces some new tests to cover more of the `timer` module. Test coverage has risen from 84 to 93%. The remaining 7% are IMO not coverable.

A side issue that I found with the new tests is related to starting the timer server. If the child spec is present in `kernel_sup` but not started, the current implementation fails with `{error, already_present}`. This was fixed so it would restart the child.

A more general change contained in this PR is turning messages to be sent around by the test cases from enumerated atoms into references, which makes them automatically unique. The same approach is used when registering a process (mostly, the test process itself), which was previously done by turning the pid into an atom.

Finally, I took the opportunity to change all tabs to spaces in `timer.erl` and `timer_simple_SUITE.erl`. I understand that this is the preferred way.

### DIFF
--- a/lib/stdlib/src/timer.erl
+++ b/lib/stdlib/src/timer.erl
@@ -20,16 +20,16 @@
 -module(timer).
 
 -export([apply_after/4,
-	 send_after/3, send_after/2,
-	 exit_after/3, exit_after/2, kill_after/2, kill_after/1,
-	 apply_interval/4, send_interval/3, send_interval/2,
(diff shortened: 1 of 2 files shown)


rickard-green:
Thanks for the PR! We do not want whitespace changes on other lines than the ones changed for other reasons. I'll keep these as is, but it can be good to know for the future.

juhlig:
> Thanks for the PR! We do not want whitespace changes on other lines than the ones changed for other reasons. I'll keep these as is, but it can be good to know for the future.

Thanks, I'll remember that :)

PR #5996, opened 2022-05-17
fix(crl): insert the CRLs individually when an URI is provided

thalesmg:
Currently, if one uses `ssl_crl_cache:insert/2` providing the URI of
the distribution point of a CRL, when a connection is attempted, it
fails with a `{unexpected_error,function_clause}`.

This traces to [`ssl_handshake:dps_and_crls/3`](https://github.com/erlang/otp/blob/8c182956e1bde761ea1541fd1d061b208356cb48/lib/ssl/src/ssl_handshake.erl#L2244), which eventually ends
up calling `public_key:der_decode/2` with a list-wrapped CRL DER
binary instead of simply the DER binary.

### DIFF
--- a/lib/ssl/src/ssl_crl_cache.erl
+++ b/lib/ssl/src/ssl_crl_cache.erl
@@ -175,7 +175,7 @@ cache_lookup(URL, {{Cache, _}, _}) ->
     case ssl_pkix_db:lookup(string:trim(Path, leading, "/"), Cache) of
 	undefined ->
 	    [];
-	CRLs ->
+	[CRLs] ->
 	    CRLs
     end.
 
--- a/lib/ssl/src/ssl_crl_cache.erl
+++ b/lib/ssl/src/ssl_crl_cache.erl
@@ -64,7 +64,7 @@ fresh_crl(#'DistributionPoint'{distributionPoint = {fullName, Names}}, CRL) ->
     case get_crls(Names, undefined) of
 	not_available ->
 	    CRL;
-	[NewCRL] ->
+	NewCRL ->
 	    NewCRL
     end.
 
--- a/lib/ssl/src/ssl_pkix_db.erl
+++ b/lib/ssl/src/ssl_pkix_db.erl
@@ -365,7 +365,7 @@ remove_crls([_,_,_, {Cache, Mapping} | _], Path) ->
     case lookup(Path, Cache) of
 	undefined ->
 	    ok;
-	CRLs ->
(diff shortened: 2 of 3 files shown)


IngelaAndin:
I think flattening the list is symptom treatment.  Would you make the patch do the following instead?

```
diff --git a/lib/ssl/src/ssl_crl_cache.erl b/lib/ssl/src/ssl_crl_cache.erl
index 095e3e8b44..7a8bcf5758 100644
--- a/lib/ssl/src/ssl_crl_cache.erl
+++ b/lib/ssl/src/ssl_crl_cache.erl
@@ -175,7 +175,7 @@ cache_lookup(URL, {{Cache, _}, _}) ->
     case ssl_pkix_db:lookup(string:trim(Path, leading, "/"), Cache) of
        undefined ->
            [];
-       CRLs ->
+       [CRLs] ->
            CRLs
     end.
```
That is we know that the internal table used in the internal implementation is an ets table of type set (pkix_db handles several ets tables ) and we can remove the outer list as our value which happens to be a list will be the one term inside. And is what ssl_crl_cache:lookup is supposed to return according to the documented callback API.

thalesmg:
@IngelaAndin Thanks for the feedback.

I've updated the patch to use that: https://github.com/erlang/otp/pull/5996/commits/f3948cd90e84c59439c00bff473b5cd40adefbb3

dgud:
You will need to remove the enclosing []  in fresh_crl since get_crls have already removed the list.

```
fresh_crl(#'DistributionPoint'{distributionPoint = {fullName, Names}}, CRL) ->
    case get_crls(Names, undefined) of
	not_available ->
	    CRL;
->	[NewCRL] ->
	    NewCRL
    end.
```

thalesmg:
Hi!  Do I need to do some other changes here?

IngelaAndin:
Well, the test case is a bit sparse!  It is only testing that the delete function does not crash. 
You could use the following code to find the reference of the CRL-cache and do lookups before and after the deletion to see that it works.

 ```
   {status, _, _, StatusInfo} = sys:get_status(whereis(ssl_manager)),
    [_, _,_, _, Prop] = StatusInfo,
    State = ssl_test_lib:state(Prop),
    case element(5, State) =  [_, _, _, {CRLCache,_}],
```

IngelaAndin:
@thalesmg Thank you for the PR

thalesmg:
Thanks for the feedback!  :tada: :confetti_ball: :beers:

PR #5997, opened 2022-05-18
dialyzer: Add incremental analysis mode

TD5:
Implements a new `dialyzer --incremental` mode.

Incremental mode primarily differs from
the previous, "classic", ways of running Dialyzer, in that its model is
optimised around the common usecase of regularly analysing a single
codebase, tweaking the code, analysing it again, and so on, without
explicit reference to the building and checking of a PLT.

In this mode the PLT file acts much more like a true cache, where users provide
a codebase and a set of files they care about, and Dialyzer does the
legwork in terms of deciding how to most efficiently report all of the
relevant warnings given the cached results it may already have in the
PLT (and if a PLT doesn't exist, incremental mode will make one).

This change also includes additional supporting options to aid with
debugging, for example, for dumping the dependency graph as
incrementality sees it (which modules depend on which other modules via
function calls, type definitions, etc.), and which modules Dialyzer
thinks changed, and hence which modules it decided to (re-)analyse.

Example usage scenarios:
- Analyse two apps and report warnings on both:
    `dialyzer --incremental --apps dialyzer erts`
- Analyse two apps, and report warnings on only one:
    `dialyzer --incremental --apps dialyzer erts --warning_apps dialyzer`
- Analyse the apps given in config*:
    `dialyzer --incremental`
- Debug an analysis by dumping the dependency graph Dialyzer computed:
    `dialyzer --incremental --apps dialyzer erts --dump_full_dependencies_graph ~/deps.dot`

*Config defaults to standard location such as
`~/.config/erlang/dialyzer.config`, but can be overridden via the
`DIALYZER_CONFIG` environment variable. An example config:
```
{incremental, {default_apps, [stdlib, erts, kernel, compiler, dialyzer]}}.
```

Further design details
======================

Key considerations:
- "Pay for what you use"
- Correctness
- Portability
- Migratability

"Pay for what you use"
----------------------
One key principle of the new incremental mode was that users should be
able to invoke Dialyzer and have the tool itself work out precisely
which parts of the analysis can be read straight back from the cache,
and which need to be (re-)analysed, due to any changes in the code
since the PLT was last updated.

Relative to the typical cost of the Dialyzer analysis itself, the
administrative overhead of incrementality is trivial in practice. There
are some extra things kept in the PLT (such as a record of the warnings
previously computed) which need to be read/written, and some changes to
how modules are hashed when checking for changes, but these costs aren't
big, and some other optimisations, such as indexing jobs more
efficiently, mean that I've not seen an incremental Dialyzer analysis that is
any slower than the corresponding building/checking of a PLT in the
classical way.

The gains from using incremental mode depend significantly on the changes
made to the code between two incremental analyses, both in terms of the
number of modules directly modified by the change, but also the size of the
transitive closure of the modules which depend on the modified code. For
usage scenarios with large changes, or changes to modules upon which
many other modules transitively depend, the benefits of incremental mode
will be more modest, because fewer of the cached results will still be
unaffected by the change. For usage scenarios where the changes are at
the "leaf" of the module dependency graph, the benefits of incremental
mode will be more significant. Re-running an incremental analysis
without making code changes should be very fast indeed, since the result
of the analysis can be read entirely from the cached results in the PLT.

Correctness
-----------

The earlier PR to erlang/OTP #5498 improved Dialyzer's tracking of
dependencies in preparation for this change. The result should be that
running Dialyzer in incremental mode for a set of modules should result
in it "doing the right thing", giving the same results as an analysis
from scratch, but leveraging an existing PLT to potentially avoid doing
some unnecessary work.

I've tested incremental mode in production for a large codebase for many
months, comparing the issues it reports with those reported by
building/checking a PLT in the classical way, and I have seen no
discrepancy between the two in practice.

I've noticed a practical improvement in correctness for some users who
would attempt to build, merge and check PLTs for subsets of
their code manually, in order to approximate something like
incrementality. Unfortunately, doing this manually risked giving incorrect
results, because it was easy to end up not analysing a module that was
transitively affected by a change to the codebase.

For simplicity and correctness reasons, incremental mode does not
support merging PLTs as it stands.

Portability
-----------

Incremental mode can provide significant performance wins if most of the
analysis can be cached. This benefit can be leveraged by sharing a PLT
file between checkouts or even machines, allowing, for example, a
previous run on a CI/CD server to generate a PLT file that a user can
then base their local analyses on.

In order to make this sort of PLT sharing portable, Dialyzer needed to
stop using absolute paths for indexing modules in the PLT and
internally, and switch to indexing on module name (which must already be
unique within a single codebase). A corollary of this is that warning
messages from incremental mode refer to only the module name, rather
than the full path. To offer some mitigation for this, incremental mode
can be configured to write out the mapping from module name to absolute
path separately to the PLT file, for use cases which may need this
information (for example, for reporting in IDE or CI/CD).

When building code from a different checkout or on a different machine,
build determinism becomes quite important. Previously, building code
from a different absolute path would result in different build artefacts
(see, for example, erlang/OTP issue #4482), which can affect Dialyzer's
perception of whether the code has changed. To mitigate this, Dialyzer's
module hashing logic now ignores some non-semantic features, such as
-file attribute paths, but running on top of a deterministic build will
still give the best results. erlang/OTP PR #5965 makes this much easier.

Migratability
-------------

Switching to the incremental mode from building/checking a PLT means
skipping any `--build_plt` step, and passing `--incremental` rather
than `--check_plt`, and providing those files, directories or apps
to analyse, either as command line arguments or via config, rather
than than relying on a list of files that were already stored in the PLT.

This change introduces a distinction between classic PLT (CPLT) and
the new incremental PLT (IPLT) files. The two formats contain slightly
different information, and aren't compatible. This change adds new
warnings which detect and report if an IPLT is used when a CPLT file
was expected, or vice versa. The hope is that, by having a clear
distinction between the two, mistakes during migration to incremental
mode will be picked up quickly.

The separation of the `dialyzer_plt` module into `dialyzer_cplt` and
`dialyzer_iplt` for the serialisation-specific logic with some
intentional duplication between the two should allow the two to
naturally evolve independently (there's no technical reason why they
must operate similarly), and give a route to deprecating the
classic mechanism in time.

Incremental mode offers the ability to analyse a set of modules, but
only report the warnings on a subset of them. One usecase for this is
where the codebase is made up of first-party code, for which you want
warnings, and some third-party dependencies, which you don't want to see
warnings for, but are needed as "context" for the analysis of the
first-party code. This was implicitly possible in classic Dialyzer modes
by virtue of the arguments used when building and checking a PLT file.

> Special thanks to @pergu and @mheiber for their research into Dialyzer's
existing logic and early proof of concept which significantly accelerated the
development of this feature.


TD5:
This is a draft for now so anyone who is interested can take a look, but I plan to break this up into a few commits  to aid readability, fix whitespace, etc.

bjorng:
> does that mean, for example, that this PR will only be reviewed when we get nearer to the release of the next major version?

No, we add new stuff to `master` continuously during development of the next upcoming release.

@jhogberg and I have been busy. We hope to get to this pull request within a few weeks.

TD5:
@bjorng, brilliant, thanks for clarifying. I look forward to your feedback 👍

bjorng (commented):
I have taken a first look at this pull request.

I've pushed a fixup commit to your branch that fixes some indentation, comment, and long line problems. Please take look and squash the commits. At the same time I suggest that you rebase to the latest `master`.

The test case `incremental_SUITE,default_apps_config_xdg/1` fails when I run it. I have not created any `dialyzer.config` file but the test case seems to expect that there is one. Here is the log from the failed test case:

```
=== Location: {incremental_SUITE,default_apps_config_xdg}
=== === Reason: {thrown,
                     {dialyzer_error,
                         "no such file or directory: \"/Users/bjorng/git/otp/release/tests/test_server/ct_run.test_server@zaphod.2022-06-23_06.56.00/tests.dialyzer_test.logs/run.2022-06-23_06.56.01/log_private/default_apps_config_xdg/Library/Application Support/erlang/dialyzer.config\""}}
```

TD5:
> The test case `incremental_SUITE,default_apps_config_xdg/1` fails when I run it. I have not created any `dialyzer.config` file but the test case seems to expect that there is one. Here is the log from the failed test case:
> 
> ```
> === Location: {incremental_SUITE,default_apps_config_xdg}
> === === Reason: {thrown,
>                      {dialyzer_error,
>                          "no such file or directory: \"/Users/bjorng/git/otp/release/tests/test_server/ct_run.test_server@zaphod.2022-06-23_06.56.00/tests.dialyzer_test.logs/run.2022-06-23_06.56.01/log_private/default_apps_config_xdg/Library/Application Support/erlang/dialyzer.config\""}}
> ```

The test is meant to check the ability of Dialyzer to read from your home directory, but it does it by changing your `$HOME` to be under the test's directory during the test case run, and the test case itself writes out a config file to that location, so there shouldn't be a need for any existing config in your real home directory. However, I am surprised by the path in the error above:

Seen from your error:
```
/log_private/default_apps_config_xdg/Library/Application Support/erlang/dialyzer.config
```

However, I expected:
```
/log_private/default_apps_config_xdg/.config/erlang/dialyzer.config
```

I suspect this is due to OS differences: I run the tests on Linux, and I presume CI does too, but it looks like you're using MacOS, which uses a different `user_config` location (see https://www.erlang.org/doc/man/filename.html#basedir-3), and the test I wrote presumed you'd be using the `.config` default for Linux. I've updated the test to detect MacOS (`darwin`), and generate what I think is the correct path in that case.

lib/dialyzer/src/dialyzer_incremental.erl
```diff
@@ -0,0 +1,811 @@
...
+   }.
+
+assert_metrics_file_valid(#options{metrics_file = none}) ->
+    ok;
+
+assert_metrics_file_valid(#options{metrics_file = MetricsFile}) ->
+    case check_if_writable(MetricsFile) of
+        true -> ok;
+        false ->
+            Msg = io_lib:format("    The metrics file ~ts is not writable", [MetricsFile]),
```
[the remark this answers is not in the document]
TD5: Yep, there's a bit of a pattern where these sorts of messages written to the console are indented.

lib/dialyzer/test/incremental_SUITE.erl
```diff
@@ -0,0 +1,881 @@
...
+    HomeEnv = case os:type() of
+                  {win32, _} ->
+                      [Drive | Path] = filename:split(TestHome),
+                      [{"APPDATA", filename:join(TestHome,"AppData")},
+                       {"HOMEDRIVE", Drive}, {"HOMEPATH", Path}];
+                  _ ->
+                      [{"HOME", TestHome}]
+              end,
+
+    ConfigSubDirs =
```
garazdawi, line 737: How about starting the peer node earlier and then doing `erpc:call(Node, filename, basedir, ...` to figure out where to put the config file?

TD5: I am not entirely sure what you mean. Is your proposal to query the `Node` for the appropriate path?

garazdawi: I would say that this test case tests "if I put a config file in the XDG directory it should be used", i.e. it is not important to test where the XDG directory is, but rather that the config file is used if places there.

In any case, this test case will fail on windows as they have yet another way to write the XDG path.

```erlang
5> filename:basedir(user_config, "erlang", #{ os => windows }). 
"/home/lukas/Roaming/erlang"
```

This raises another question, what happens if I set `DIALYZER_CONFIG` and have an XDG config? Does that behavior need a test?

TD5: Ah, yes, good spot!

Regarding `DIALYZER_CONFIG` and XDG, `default_apps_config_env_var` covers this implicitly, but the semantics are that if `DIALYZER_CONFIG` exists, it is used, and if not, we fallback to the XDG location.

garazdawi: >  `default_apps_config_env_var` covers this implicitly

I must be missing something as it is not obvious to me that this is the case.

TD5: I agree this should be clearer. I’ve got another test that demonstrates this explicitly. I’m away for a few days at the moment, but I’ll push it when I get back.

TD5:
@bjorng / @garazdawi I think everything you mentioned has been resolved now. Is there anything else you'd want me to modify?

bjorng:
Thanks! I will add the PR to our daily builds and if no problems turn up I will have a final look and then merge it.

bjorng:
There is a failed test case on Windows:

```
=== Ended at 2022-07-27 06:33:22
=== Location: {incremental_SUITE,default_apps_config_xdg}
=== === Reason: {thrown,
                     {{dialyzer_error,
                          "no such file or directory: \"c:/ldisk/daily_build/26_master-opu_32_4384.2022-07-26_21/test/test_server/ct_run.test_server@Fobi.2022-07-27_06.28.59/test.dialyzer_test.logs/run.2022-07-27_06.29.03/log_private/default_apps_config_xdg/AppData/erlang/dialyzer.config\""},
                      [{erpc,call,5,[{file,"erpc.erl"},{line,698}]},
                       {incremental_SUITE,default_apps_config_xdg,1,
                           [{file,"incremental_SUITE.erl"},{line,760}]},
                       {test_server,ts_tc,3,
                           [{file,"test_server.erl"},{line,1782}]},
                       {test_server,run_test_case_eval1,6,
                           [{file,"test_server.erl"},{line,1291}]},
                       {test_server,run_test_case_eval,9,
                           [{file,"test_server.erl"},{line,1223}]}]}}
```

Everything else looks good.

(I will not work tomorrow and on Friday; I will be back on Monday next week.)

TD5:
Thanks @bjorng, I'll take a look and get that sorted out

TD5:
I don't have access to a Windows machine to directly check this myself, but I've updated the test to reflect the details from the log you pasted.

bjorng:
The test case still fails, but I have decided to merge the branch anyway. I will take care of the test case later.

PR #6001, opened 2022-05-19
Fix encoding fun containing export fun or bitstring

sverker:
toward a remote node with a not yet established connection.

PR #5987 is removing the symptom on the receiving node for this problem.

This PR fixes the encoding problem as erl_interface still relies on a correct encoding of the NEW_FUN_EXT size field.


garazdawi (approved):

sverker:
Released in
OTP-23.3.4.16
OTP-24.3.4.3

PR #6002, opened 2022-05-19
Add new Linux paths to load OS cacerts

codeadict:
Adds new paths to load OS CA certificates for the following
Linux distros:

* CentOS
* RHEL 7
* Alpine Linux

Tested on a CentOS server and an Alpine linux Docker container.


dgud:
Can rebase this onto maint branch instead, so we can patch this.

codeadict:
@dgud - Do you need me to change base to the `maint` branch os is that something the OTP team takes care of?

dgud:
I can do it but the github integration works better if you do it, i.e. PR closes automagicly with a merge tag and so on.

codeadict:
This is rebased and opened against maint now

PR #6008, opened 2022-05-23
gen_event: store callbacks in handler record

mikpe:
This was inspired by the PR to cache callback funs in gen_server (https://github.com/erlang/otp/pull/5831) but applies to gen_event instead. Same idea: store callback funs and invoke those instead of going though apply-like Mod:Func(...) calls. In my environments I'm generally seeing a 2x to 3x speedup for the fun calls using Lukas' micro-benchmark in that other PR.

There's also some refactoring to reduce the amount of duplicated code for initializing handlers. The server_update functions are refactored to avoid fetching data we don't need in the fast paths. The format_status/2 change is there to prevent leaking the internal data structure change which would regress gen_event_SUITE.

### DIFF
--- a/lib/stdlib/src/gen_event.erl
+++ b/lib/stdlib/src/gen_event.erl
@@ -65,6 +65,27 @@
               del_handler_ret/0, request_id/0, request_id_collection/0]).
 
 -record(handler, {module             :: atom(),
+		  handle_event = error(uninitialized_handle_event)
+			       :: fun((Event :: term(), State :: term()) ->
+				      {ok, NewState :: term()} |
+				      {ok, NewState :: term(), hibernate} |
+				      {swap_handler, Args1 :: term(), NewState :: term(),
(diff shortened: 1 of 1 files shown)


mikpe:
I just realized a drawback of this: gen_event also stores a handler's state in the handler record, and this is updated every time the handler is invoked. By enlarging the handler record by three fields we also force gen_event to allocate and initialize three more fields on every handler invocation. I can shrink the handler record to two words by moving all invariant fields to a sub-tuple, but it costs another indirection when retrieving any of those invariant fields (Mod, callback funs, ...). :disappointed:

mikpe:
I'm declining this PR. `gen_event` has design issues that work against attempts to reduce its overheads while forwarding events, making the export fun caching in this PR ineffective. I've replaced our most heavily used `gen_event` process with a custom `gen_server` because of that.

PR #6018, opened 2022-05-24
dialyzer: Correct documentation for missing/extra_return

### LINKED ISSUE #6017: Dialyzer: flags missing_return and extra_return go against documentation (they are swapped)
**Describe the bug**

`-Wextra_return` command line option does what is described for `-Wmissing_return` option in documentation and vice versa.

**To Reproduce**

```erlang
-module(xxx).
-export([foo/1, bar/1]).

-spec foo(integer()) -> integer() | binary().
foo(1) -> 2;
foo(_) -> 4.

-spec bar(any()) -> integer().
bar(1) -> 2;
bar(_) -> ok.
```

```
dialyzer xxx.erl
.. - no warnings
```
----

```
dialyzer -Wmissing_return xxx.erl
xxx.erl:8:2: The success typing for xxx:bar/1 implies that the function might also return 
          'ok' but the specification return is 
          integer()
```

Documentation:

> -Wmissing_return
> Warn about functions whose specification includes types that the function cannot return.

-----

```
dialyzer -Wextra_return xxx.erl
xxx.erl:4:2: The specification for xxx:foo/1 states that the function might also return 
          binary() but the inferred return is 
          2 | 4

```

Documentation:

> -Wextra_return
> Warn about functions that return values that are not part of the specification.

----

The flags are "swapped" in command line

**Expected behavior**
According to documentation - https://www.erlang.org/doc/man/dialyzer.html

**Affected versions**
25.0

jhogberg:
Fixes #6017

### DIFF
--- a/lib/dialyzer/doc/src/dialyzer.xml
+++ b/lib/dialyzer/doc/src/dialyzer.xml
@@ -318,13 +318,13 @@ dialyzer --plts plt_1 ... plt_n -- files_to_analyze</code>
       </item>
       <tag><c>-Wextra_return</c> (***)</tag>
       <item>
-        <p>Warn about functions that return values that are not part of the
-          specification.</p>
+        <p>Warn about functions whose specification includes types that the
+          function cannot return.</p>
       </item>
       <tag><c>-Wmissing_return</c> (***)</tag>
       <item>
-        <p>Warn about functions whose specification includes types that the
-          function cannot return.</p>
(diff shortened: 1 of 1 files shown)


ilya-klyuchnikov:
There are also pieces about `-Wno_extra_return` and `-Wno_missing_return`

jhogberg:
Ah, I was too quick. Thanks :)

PR #6019, opened 2022-05-24
ssl: Fix stateless ticket in_window comparison

### LINKED ISSUE #6014: TLS 1.3 stateless session resumption with anti_replay ignores ticket lifetime
**Describe the bug**
When using TLS 1.3 session resumption with `anti_replay` enabled, the ticket is only valid for `WindowSize` seconds (see: https://www.erlang.org/doc/man/ssl.html#type-anti_replay).

Note that at the bottom I have a proposed resolution to the problem.

**To Reproduce**
We can use the test setup from #5943 with slight modifications.
 
Generate some test certificates:

```shell
openssl req -x509 -newkey rsa:4096 \
    -keyout rootCA.key -out rootCA.pem \
    -sha256 -days 365 -nodes \
    -subj '/CN=Test root CA'

openssl genrsa -out server.key 2048
openssl req -new -key server.key -out server.csr \
    -subj "/CN=Test server"

openssl x509 -req -in server.csr \
    -CA rootCA.pem -CAkey rootCA.key \
    -CAcreateserial -out server.pem -days 1000 -sha256
```

`server.erl`:

```erlang
-module(server).

-export([run/0,
         run_anti_replay/0,
         start_trace/0,
         stop_trace/0]).

run() ->
    application:ensure_all_started(ssl),
    LOpts = [{cacertfile, "rootCA.pem"},
             {certfile, "server.pem"},
             {keyfile, "server.key"},
             {versions, ['tlsv1.3']},
             {session_tickets, stateless}],
    {ok, LSock} = ssl:listen(8001, LOpts),
    handle(LSock).

run_anti_replay() ->
    application:ensure_all_started(ssl),
    LOpts = [{cacertfile, "rootCA.pem"},
             {certfile, "server.pem"},
             {keyfile, "server.key"},
             {versions, ['tlsv1.3']},
             {session_tickets, stateless},
             {anti_replay, {10, 5, 72985}}], % {10, 5, 72985} is equivalent to the 10k default
    {ok, LSock} = ssl:listen(8001, LOpts),
    handle(LSock).

handle(LSock) ->
    {ok, CSock} = ssl:transport_accept(LSock),
    {ok, Sock} = ssl:handshake(CSock),
    {ok, Opts} = ssl:connection_information(Sock, [session_resumption]),
    io:format('Connected with session resumption ~p~n', [Opts]),
    handle(LSock).

start_trace() ->
    dbg:tracer(),
    dbg:p(all, [c]),
    dbg:tpl({tls_server_session_ticket, stateless_living_ticket, '_'}, [{'_',[],[{return_trace}]}]).

stop_trace() ->
    dbg:stop_clear().
```

Start tracing and run the server:

```
1> c(server).
{ok,server}
2> server:start_trace().
{ok,[{matched,nonode@nohost,1},{saved,1}]}
3> server:run().
```

Connect with OpenSSL `s_client`, waiting 15 seconds before trying to reconnect and resuming the session:

```shell
sleep 3s | openssl s_client -tls1_3 -connect localhost:8001 -sess_out sess1.pem -debug
sleep 15s
sleep 3s | openssl s_client -tls1_3 -connect localhost:8001 -sess_in sess1.pem -debug
```

The server shows that the connection succeeded to use session resumption:

```
Connected with session resumption [{session_resumption,false}]
(<0.120.0>) call tls_server_session_ticket:stateless_living_ticket(1956700463,1956683463,7200,1653382359,undefined)
(<0.120.0>) returned from tls_server_session_ticket:stateless_living_ticket/5 -> true
Connected with session resumption [{session_resumption,true}]

```

If we now start a new server with `anti_replay` enabled with `server:run_anti_replay()`

```
1> c(server).
{ok,server}
2> server:start_trace().
{ok,[{matched,nonode@nohost,1},{saved,1}]}
3> server:run_anti_replay().
```

The server shows that the connection failed to use session resumption:

```
Connected with session resumption [{session_resumption,false}]
(<0.120.0>) call tls_server_session_ticket:stateless_living_ticket(1176305235,1176288235,7200,1653382568,10)
(<0.120.0>) returned from tls_server_session_ticket:stateless_living_ticket/5 -> false
Connected with session resumption [{session_resumption,false}]
```

**Expected behavior**
Enabling `anti_replay` does not change the behavior of normal client flows.

**Affected versions**
Tested on master (`38f6106`) and a fork of `24.1.3`

**Additional context**
This is the current implementation

`tls_server_session_ticket.erl`
```erlang
stateless_living_ticket(0, _, _, _, _) ->
    true;
stateless_living_ticket(ObfAge, TicketAgeAdd, Lifetime, Timestamp, Window) ->
    ReportedAge = ObfAge - TicketAgeAdd,
    RealAge = erlang:system_time(second) - Timestamp,
    (ReportedAge =< Lifetime * 1000)
        andalso (RealAge =< Lifetime)
        andalso (in_window(RealAge, Window)).
```

A proposed solution would be to subtract the ReportedAge in seconds from RealAge and take the absolute value.

```erlang
stateless_living_ticket(ObfAge, TicketAgeAdd, Lifetime, Timestamp, Window) ->
    ...
    DriftTime = abs(RealAge - (ReportedAge / 1000)),
    (ReportedAge =< Lifetime * 1000)
        andalso (RealAge =< Lifetime)
        andalso (in_window(DriftTime, Window)).

```

The above change fixes the simple example that is presented above, but might not be the ideal solution due to client-side clock drift.

sindrip:
If a stateless ticket is older than the WindowSize of the bloom filter
then it will be rejected. By subtracting the ReportedAge from the
RealAge we have the drift time, which should become too large during
a replay. Previously the absolute age of the ticket was being used.

Fixes #6014

### DIFF
--- a/lib/ssl/doc/src/using_ssl.xml
+++ b/lib/ssl/doc/src/using_ssl.xml
@@ -776,7 +776,7 @@ ssl:connect("localhost", 9999, [{verify, verify_peer},
       less than ticket lifetime.</p></item>
       <item><p>Actual ticket age shall be less than the ticket lifetime (stateless session
       tickets contain the servers timestamp when the ticket was issued).</p></item>
-      <item><p>Ticket shall be used within specified time window (freshness checks).</p></item>
+      <item><p>ClientHello created with the ticket shall be sent relatively recently (freshness checks).</p></item>
       <item><p>If all above checks passed both <em>current</em> and <em>old</em> Bloom filters
       are checked to detect if binder was already seen. Being a probabilistic data structure,
       false positives can occur and they trigger a full handshake.</p></item>
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -1456,7 +1456,7 @@ fun(srp, Username :: binary(), UserState :: term()) ->
 	  <p>Allowed values are the pre-defined <c>'10k'</c>, <c>'100k'</c> or a custom 3-tuple that
 	  defines the properties of the bloom filters: <c>{WindowSize, HashFunctions, Bits}</c>.
 	  <c>WindowSize</c> is the number of seconds after the current Bloom filter is rotated
-	  and also the window size used for freshness checks. <c>HashFunctions</c> is the number
+	  and also the window size used for freshness checks of ClientHello. <c>HashFunctions</c> is the number
 	  hash functions and <c>Bits</c> is the number of bits in the bit vector.
 	  <c>'10k'</c> and <c>'100k'</c> are simple defaults with the following properties:</p>
 	  <list type="bulleted">
--- a/lib/ssl/src/tls_server_session_ticket.erl
+++ b/lib/ssl/src/tls_server_session_ticket.erl
@@ -382,15 +382,50 @@ stateless_usable_ticket(#stateless_ticket{hash = Prf,
 
 stateless_living_ticket(0, _, _, _, _) ->
     true;
+%% If `anti_replay` is not enabled, then a ticket is considered to be living
+%% if it has not exceeded its lifetime.
+%%
+%% If `anti_replay` is enabled, we must additionally perform a freshness check
+%% as is outlined in section 8.3 Freshness Checks - RFC 8446
 stateless_living_ticket(ObfAge, TicketAgeAdd, Lifetime, Timestamp, Window) ->
-    ReportedAge = ObfAge - TicketAgeAdd,
+    %% RealAge is the server's view of the age of the ticket in seconds.
     RealAge = erlang:system_time(second) - Timestamp,
+
+    %% ReportedAge is the client's view of the age of the ticket in milliseconds.
+    ReportedAge = ObfAge - TicketAgeAdd,
+
+    %% DeltaAge is the difference of the client's view of the age of the ticket
+    %% and the server's view of the age of the ticket in seconds.
+    DeltaAge = abs(RealAge - (ReportedAge / 1000)),
+
+    %% We ensure that both the client's view of the age of the ticket and the
+    %% server's view of the age of the ticket do not exceed the lifetime specified.
     (ReportedAge =< Lifetime * 1000)
         andalso (RealAge =< Lifetime)
-        andalso (in_window(RealAge, Window)).
+        andalso (in_window(DeltaAge, Window)).
 
 in_window(_, undefined) ->
     true;
+%% RFC 8446 - section 8.2 Client Hello Recording
+%% describes an anti-replay implementation that can use bounded memory
+%% by storing a unique value from a ClientHello (in our case the PSK binder)
+%% withing a given time window.
+%%
+%% In order implement this, when a ClientHello is received, the server
+%% must ensure that a ClientHello has been sent relatively recently.
+%% We do this by ensuring that the client and server view of the age
+%% of the ticket is not larger than our recording window.
+%%
+%% In the case of an attempted replay attack, there are 2 possible
+%% outcomes:
+%%   - A ClientHello is replayed within the recording window
+%%      * The ticket looks valid, `in_window` returns true
+%%        so we proceed to check the unique value
+%%      * The unique value (PSK Binder) is stored in the bloom filter
+%%        and we reject the ticket.
+%%
+%%   - A ClientHello is replayed outside the recording window
+%%      * We reject the ticket as `in_window` returns false.
 in_window(Age, Window) when is_integer(Window) ->
     Age =< Window.
 
--- a/lib/ssl/test/ssl_session_ticket_SUITE.erl
+++ b/lib/ssl/test/ssl_session_ticket_SUITE.erl
@@ -243,7 +243,7 @@ ticketage_smaller_than_windowsize_anti_replay(Config) when is_list(Config) ->
 ticketage_bigger_than_windowsize_anti_replay() ->
     [{doc, "Session resumption with stateless tickets and anti_replay enabled."
       "Fresh ClientHellos."
-      "Ticket age bigger than windowsize. 0-RTT is expected to fail."
+      "Ticket age bigger than windowsize. 0-RTT is expected to succeed."
       "(Erlang client - Erlang server)"}].
 ticketage_bigger_than_windowsize_anti_replay(Config) when is_list(Config) ->
     WindowSize = 3,
@@ -252,10 +252,10 @@ ticketage_bigger_than_windowsize_anti_replay(Config) when is_list(Config) ->
     ssl_test_lib:check_result(Server0, ok, Client0, ok),
     Client1 = anti_replay_helper_connect(Server0, Client0, Port0, ClientNode,
                                          Hostname, ClientOpts,
-                                         {seconds, WindowSize + 2}, false),
+                                         {seconds, WindowSize + 2}, true),
     Client2 = anti_replay_helper_connect(Server0, Client0, Port0, ClientNode,
                                          Hostname, ClientOpts,
-                                         {seconds, 2*WindowSize + 2}, false),
+                                         {seconds, 2*WindowSize + 2}, true),
     process_flag(trap_exit, false),
     [ssl_test_lib:close(A) || A <- [Server0, Client0, Client1, Client2]].
 
--- a/lib/ssl/test/tls_server_session_ticket_SUITE.erl
+++ b/lib/ssl/test/tls_server_session_ticket_SUITE.erl
@@ -37,14 +37,17 @@
          main_test/0,
          main_test/1,
          misc_test/0,
-         misc_test/1]).
+         misc_test/1,
+         valid_ticket_older_than_windowsize_test/0,
+         valid_ticket_older_than_windowsize_test/1]).
 
--define(LIFETIME, 1). % tickets expire after 1s
+-define(LIFETIME, 3). % tickets expire after 3s
 -define(TICKET_STORE_SIZE, 1).
 -define(MASTER_SECRET, "master_secret").
 -define(PRF, sha).
 -define(VERSION, {3,4}).
 -define(PSK, <<15,168,18,43,216,33,227,142,114,190,70,183,137,57,64,64,66,152,115,94>>).
+-define(WINDOW_SIZE, 1).
 
 %%--------------------------------------------------------------------
 %% Common Test interface functions -----------------------------------
@@ -55,12 +58,12 @@ all() ->
 groups() ->
     [{stateful, [], [main_test, expired_ticket_test, invalid_ticket_test]},
      {stateless, [], [expired_ticket_test, invalid_ticket_test, main_test]},
-     {stateless_antireplay, [], [main_test, misc_test]}
+     {stateless_antireplay, [], [main_test, misc_test, valid_ticket_older_than_windowsize_test]}
     ].
 
 init_per_group(stateless_antireplay, Config) ->
     check_environment([{server_session_tickets, stateless},
-                       {anti_replay, {10, 20, 30}}]
+                       {anti_replay, {?WINDOW_SIZE, 20, 30}}]
                       ++ Config);
 init_per_group(Group = stateless, Config) ->
     check_environment([{server_session_tickets, Group} | Config]);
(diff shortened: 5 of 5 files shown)


IngelaAndin:
Your test case actually fails on some of our test machines. It might be due to making the window size smaller the possibility for false negatives become higher. There actually is no guarantee with stateless tickets that all valid tickets will be considered valid and used. I think this section of the RFC is what we should go on is:

```
8.3.  Freshness Checks

   Because the ClientHello indicates the time at which the client sent
   it, it is possible to efficiently determine whether a ClientHello was
   likely sent reasonably recently and only accept 0-RTT for such a
   ClientHello, otherwise falling back to a 1-RTT handshake.  This is
   necessary for the ClientHello storage mechanism described in
   Section 8.2 because otherwise the server needs to store an unlimited
   number of ClientHellos, and is a useful optimization for self-
   contained single-use tickets because it allows efficient rejection of
   ClientHellos which cannot be used for 0-RTT.

   In order to implement this mechanism, a server needs to store the
   time that the server generated the session ticket, offset by an
   estimate of the round-trip time between client and server.  I.e.,

       adjusted_creation_time = creation_time + estimated_RTT

   This value can be encoded in the ticket, thus avoiding the need to
   keep state for each outstanding ticket.  The server can determine the
   client's view of the age of the ticket by subtracting the ticket's
   "ticket_age_add" value from the "obfuscated_ticket_age" parameter in
   the client's "pre_shared_key" extension.  The server can determine
   the expected_arrival_time of the ClientHello as:

     expected_arrival_time = adjusted_creation_time + clients_ticket_age

   When a new ClientHello is received, the expected_arrival_time is then
   compared against the current server wall clock time and if they
   differ by more than a certain amount, 0-RTT is rejected, though the
   1-RTT handshake can be allowed to complete.

 There are several potential sources of error that might cause
   mismatches between the expected_arrival_time and the measured time.
   Variations in client and server clock rates are likely to be minimal,
   though potentially the absolute times may be off by large values.
   Network propagation delays are the most likely causes of a mismatch
   in legitimate values for elapsed time.  Both the NewSessionTicket and
   ClientHello messages might be retransmitted and therefore delayed,
   which might be hidden by TCP.  For clients on the Internet, this
   implies windows on the order of ten seconds to account for errors in
   clocks and variations in measurements; other deployment scenarios may
   have different needs.  Clock skew distributions are not symmetric, so
   the optimal tradeoff may involve an asymmetric range of permissible
   mismatch values.

   Note that freshness checking alone is not sufficient to prevent
   replays because it does not detect them during the error window,
   which -- depending on bandwidth and system capacity -- could include
   billions of replays in real-world settings.  In addition, this
   freshness checking is only done at the time the ClientHello is
   received and not when subsequent early Application Data records are
   received.  After early data is accepted, records may continue to be
   streamed to the server over a longer time period.
```
I do not think our implementation takes estimated round trip time into account, and that is probably what we want. But  I do not have an emiditat good idea of how to best do that.

IngelaAndin:
Your test case actually fails on some of our test machines. It might be due to making the window size smaller the possibility for false negatives become higher. There actually is no guarantee with stateless tickets that all valid tickets will be considered valid and used. I think this section of the RFC is what we should go on is:

```
8.3.  Freshness Checks

   Because the ClientHello indicates the time at which the client sent
   it, it is possible to efficiently determine whether a ClientHello was
   likely sent reasonably recently and only accept 0-RTT for such a
   ClientHello, otherwise falling back to a 1-RTT handshake.  This is
   necessary for the ClientHello storage mechanism described in
   Section 8.2 because otherwise the server needs to store an unlimited
   number of ClientHellos, and is a useful optimization for self-
   contained single-use tickets because it allows efficient rejection of
   ClientHellos which cannot be used for 0-RTT.

   In order to implement this mechanism, a server needs to store the
   time that the server generated the session ticket, offset by an
   estimate of the round-trip time between client and server.  I.e.,

       adjusted_creation_time = creation_time + estimated_RTT

   This value can be encoded in the ticket, thus avoiding the need to
   keep state for each outstanding ticket.  The server can determine the
   client's view of the age of the ticket by subtracting the ticket's
   "ticket_age_add" value from the "obfuscated_ticket_age" parameter in
   the client's "pre_shared_key" extension.  The server can determine
   the expected_arrival_time of the ClientHello as:

     expected_arrival_time = adjusted_creation_time + clients_ticket_age

   When a new ClientHello is received, the expected_arrival_time is then
   compared against the current server wall clock time and if they
   differ by more than a certain amount, 0-RTT is rejected, though the
   1-RTT handshake can be allowed to complete.

 There are several potential sources of error that might cause
   mismatches between the expected_arrival_time and the measured time.
   Variations in client and server clock rates are likely to be minimal,
   though potentially the absolute times may be off by large values.
   Network propagation delays are the most likely causes of a mismatch
   in legitimate values for elapsed time.  Both the NewSessionTicket and
   ClientHello messages might be retransmitted and therefore delayed,
   which might be hidden by TCP.  For clients on the Internet, this
   implies windows on the order of ten seconds to account for errors in
   clocks and variations in measurements; other deployment scenarios may
   have different needs.  Clock skew distributions are not symmetric, so
   the optimal tradeoff may involve an asymmetric range of permissible
   mismatch values.

   Note that freshness checking alone is not sufficient to prevent
   replays because it does not detect them during the error window,
   which -- depending on bandwidth and system capacity -- could include
   billions of replays in real-world settings.  In addition, this
   freshness checking is only done at the time the ClientHello is
   received and not when subsequent early Application Data records are
   received.  After early data is accepted, records may continue to be
   streamed to the server over a longer time period.
```
I do not think our implementation takes estimated round trip time into account, and that is probably what we want. But  I do not have an emiditat good idea of how to best do that.

sindrip:
Wouldn't adding the estimated round-trip time be considered an enhancement? As it stands, this fix is to implement the original intention of the code, because the anti-replay mechanism is unusable in its current state. Currently the deviation from the RFC is that we calculate `expected_arrival_time` without adjusting for the estimated connection latency `estimated_RTT`. It is also not clear to me how one is supposed to derive `estimated_RTT` when receiving a ClientHello, if it is a constant or per connection.

Regarding the point that not all valid tickets will be considered valid and used. I'm assuming this is induced by the bloom filter that is used to validate tickets. Doesn't that imply that each test-case for valid tickets should be ran multiple times and expected to pass x% of times (97% with the default bloom filter values)? Alternatively the tests could be ran with a deterministic data structure for the ticket validation storage, but that is I believe a big rework and requires explicit testing of the bloom filter (I think it's only indirectly tested right now).

> [8.2](https://datatracker.ietf.org/doc/html/rfc8446#section-8.2).  Client Hello Recording
   ...
    Servers MAY
   also implement data stores with false positives, such as Bloom
   filters, in which case they MUST respond to apparent replay by
   rejecting 0-RTT but MUST NOT abort the handshake.

Regarding my original concern about the clock drift time:
> [8.2](https://datatracker.ietf.org/doc/html/rfc8446#section-8.2).  Client Hello Recording
   ...
   Note: If the client's clock is running much faster than the server's,
   then a ClientHello may be received that is outside the window in the
   future, in which case it might be accepted for 1-RTT, causing a
   client retry, and then acceptable later for 0-RTT.  This is another
   variant of the second form of attack described in [Section 8](https://datatracker.ietf.org/doc/html/rfc8446#section-8).

Regarding this specific PR, I originally reduced the window_size to keep the test execution time down. I could try tweaking the `Lifetime` and `WindowSize` and we could see how that works. Let me know if you want to try this.

u3s:
I would like clarify discussion in related issue #6014 , before proceeding with this PR.

u3s:
@peterdmv wrote:
> Regarding the solution I would prefer using the naming scheme of the RFC and keeping variables even if they are not used for future reference and maintainability e.g. EstimatedRTT.

I'm not sure how to read that. Examples provided in issue discussion are
different than existing code and also PR content. At the moment, I'm not sure if it is worth to rewrite more code just to match the RFC content - assuming the logic is good and code execution result as expected.

### Other comments
1. changing that might be potentially incompatible - because tickets will start
be usable(or useful as is suggested) after WindowSize seconds.
   - I understand it that with stateless tickets, user can potentially use them as many times
     as they want (although it is recommended for the clients to use it only once -
     RFC C.4)
   - Today with anti_replay: this is limited to Bloom filter WindowSize seconds
     (which is suggestd to not be correct) - tickets with RealAge bigger than WindowSize will be rejected
   - also Today without anti_replay: this is limited to ticket lifetime
   - With the change proposed: re-use possibility will be limited only with ticket lifetime
     (no matter if with or without anti_replay enabled)
2. I wonder if readability of code could be improved a little (so conclusion from our
   discussion is reflected):
   - I was considering adding some explanatory comment and/or renaming some functions
     (I'm not sure which is better - what do you think)?
     - function names tls_server_session_ticket:stateless*ticket are kind of ticket oriented
     - they don't cover ClientHello replay protection (anti_replay) flow and might
       be misguiding - I mean it is not only ticket that is checked with anti_replay
3. docs - maybe anti_replay option manual could be improved so our discussion is
   somehow reflected there as well.

u3s:
I've added more tests for anti_replay in https://github.com/erlang/otp/pull/6055/.
`ticket_reuse_anti_replay` will be failing with PR6019, because I wrote it to verify existing behavior which we agreed is not expected.

I hope you will adjust it in this PR, once I merge PR-6055 to master.

This will be needed:
```
@@ -243,7 +243,7 @@ ticketage_smaller_than_windowsize_anti_replay(Config) when is_list(Config) ->
 ticketage_bigger_than_windowsize_anti_replay() ->
     [{doc, "Session resumption with stateless tickets and anti_replay enabled."
       "Fresh ClientHellos."
-      "Ticket age bigger than windowsize. 0-RTT is expected to fail."
+      "Ticket age bigger than windowsize. 0-RTT is expected to succeed."
       "(Erlang client - Erlang server)"}].
 ticketage_bigger_than_windowsize_anti_replay(Config) when is_list(Config) ->
     WindowSize = 3,
@@ -252,10 +252,10 @@ ticketage_bigger_than_windowsize_anti_replay(Config) when is_list(Config) ->
     ssl_test_lib:check_result(Server0, ok, Client0, ok),
     Client1 = anti_replay_helper_connect(Server0, Client0, Port0, ClientNode,
                                          Hostname, ClientOpts,
-                                         {seconds, WindowSize + 2}, false),
+                                         {seconds, WindowSize + 2}, true),
     Client2 = anti_replay_helper_connect(Server0, Client0, Port0, ClientNode,
                                          Hostname, ClientOpts,
-                                         {seconds, 2*WindowSize + 2}, false),
+                                         {seconds, 2*WindowSize + 2}, true),
     process_flag(trap_exit, false),
     [ssl_test_lib:close(A) || A <- [Server0, Client0, Client1, Client2]].

```

sindrip:
Let me know once #6055 is merged and I can make the suggested changes

u3s:
#6055 is merged. please rebase and fix the `ticketage_bigger_than_windowsize_anti_replay`.

sindrip:
I rebased on master and updated the testcase from #6055, what do you think about the testcases that I wrote originally in this PR, the new testcases you added test the same thing @u3s?

> * I wonder if readability of code could be improved a little (so conclusion from our
>   discussion is reflected):
>   
>   * I was considering adding some explanatory comment and/or renaming some functions
>     (I'm not sure which is better - what do you think)?
>     
>     * function names tls_server_session_ticket:stateless*ticket are kind of ticket oriented
>     * they don't cover ClientHello replay protection (anti_replay) flow and might
>       be misguiding - I mean it is not only ticket that is checked with anti_replay

I think comments surrounding the checks with references to the RFC would be fine by me. Do you want me to cover this?

> * docs - maybe anti_replay option manual could be improved so our discussion is
>   somehow reflected there as well.

Do you want me to add it to this PR?

u3s:
>I rebased on master and updated the testcase from https://github.com/erlang/otp/pull/6055, what do you think about the testcases that I wrote originally in this PR, the new testcases you added test the same thing @u3s?

I'm fine with leaving them as your tests are more of a white box tests. My recent addition is more end2end.

> think comments surrounding the checks with references to the RFC would be fine by me. Do you want me to cover this?

yes please.

>>        docs - maybe anti_replay option manual could be improved so our discussion is
>>        somehow reflected there as well.
> Do you want me to add it to this PR?

yes, if not a problem for you to make a small adjustment. Maybe its a matter of adding ClientHello around "freshness checks" - to indicate what we're recording and checking with anti_replay - I'm looking [here](https://www.erlang.org/doc/man/ssl.html#type-anti_replay).

sindrip:
I did some slight documentation updates, but it is kind of adressed here: https://www.erlang.org/doc/apps/ssl/using_ssl.html#anti-replay-protection-in-tls-1.3

lib/ssl/doc/src/using_ssl.xml
```diff
@@ -776,7 +776,7 @@ ssl:connect("localhost", 9999, [{verify, verify_peer},
       less than ticket lifetime.</p></item>
       <item><p>Actual ticket age shall be less than the ticket lifetime (stateless session
       tickets contain the servers timestamp when the ticket was issued).</p></item>
-      <item><p>Ticket shall be used within specified time window (freshness checks).</p></item>
+      <item><p>Ticket shall be sent relatively recently (freshness checks).</p></item>
```
u3s, line 779: Ticket -> ClientHello?

Issue #6022, opened 2022-05-26
Building OTP 25.0 on Ubuntu 22.04 LTS with wxWidgets 3.1.6 causes compilation error


jj1bdx:
**Describe the bug**
Building OTP 25.0 on Ubuntu 22.04 LTS with wxWidgets 3.1.6 wxGTK causes compilation error in `lib/wx/c_src/gen/wxe_wrapper_7.cpp` due to type difference of a class wxBitmap (on wx-3.0) and wxBitmapBundle (on wx-3.1). 

**To Reproduce**

Execution environment:

* Ubuntu 22.04 LTS with uname: `Linux gluonite 5.15.0-33-generic #34-Ubuntu SMP Wed May 18 13:34:26 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux` (gluonite is my machine name)
* gcc (Ubuntu 11.2.0-19ubuntu1) 11.2.0
* wxWidgets 3.1.6.0 wxGTK built with the following configure options: `--prefix="/opt/wx-3.1" --enable-compat28 --enable-compat30 --enable-unicode --with-gtk` (GTK version: 3)
* Following GTK packages installed on Ubuntu: libgth-4-dev libgtk-3-dev
* kerl version 2.3.0
* Locally built OpenSSL 1.1.1o under /opt/openssl/1.1.1o

Build shell command using kerl:

```shell
OTP_VERSION=25.0
OPENSSL_LOCAL="/opt/openssl/1.1.1o"
INSTALL_DIR_PREFIX="/home/kenji/otp/"
INSTALL_NAME=${OTP_VERSION}-wxgtk-buildtest
#
env \
    CC=/usr/bin/cc \
    CXX=/usr/bin/c++ \
    CFLAGS="-O3 -fstack-protector-strong" LDFLAGS="-fstack-protector-strong" \
    MAKEFLAGS="-j12" \
    KERL_BUILD_DOCS="yes" \
    KERL_BUILD_PLT="yes" \
[...]

PR #6023, opened 2022-05-26
Fix AddTool bug of wx module for wxWidgets 3.1.6

jj1bdx:
* Add additional type casting to wxToolBar_AddTool_4() as a workaround
  for compilation error on gcc 11.2.0 of Ubuntu
  - wxBitmapBundle is introduced since wxWidgets 3.1.6
  - wxBitmap is assignble to wxBitmapBundle object
    - See https://docs.wxwidgets.org/3.1.6/classwx_bitmap_bundle.html,
      quote:

> Also note that the existing code using wxBitmap is compatible with the functions taking wxBitmapBundle in wxWidgets 3.1.6 and later because bitmaps are implicitly convertible to the objects of this class, so just passing wxBitmap to the functions taking wxBitmapBundle continues to work and if high resolution versions of bitmap are not (yet) available for the other toolbar tools, single bitmaps can continue to be used instead.

* See also https://github.com/erlang/otp/issues/6022

Build tested on Ubuntu 22.04 amd64 and macOS 12.4 x86_64.

### DIFF
--- a/lib/wx/c_src/gen/wxe_wrapper_7.cpp
+++ b/lib/wx/c_src/gen/wxe_wrapper_7.cpp
@@ -2339,7 +2339,13 @@ void wxToolBar_AddTool_4(WxeApp *app, wxeMemEnv *memenv, wxeCommand& Ecmd)
     } else        Badarg("Options");
   };
   if(!This) throw wxe_badarg("This");
+// AddTool() type check workaround
+// See https://docs.wxwidgets.org/3.1.6/classwx_bitmap_bundle.html
+#if wxCHECK_VERSION(3,1,6)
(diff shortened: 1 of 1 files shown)


dgud:
I can't reproduce the building problem.
Does it work if you build without --enable-compat28 (which is not needed anymore)

Also please don't change the generated code, you will need to change the generator or its input.

jj1bdx:
@dgud Sorry for having not understood the whole API generation process described in `lib/wx/api_gen/README`.

Also, rebuilding the wxWidgets 3.1.6 by removing `--enable-compat28` as you suggested stopped causing the compilation error with the OTP-25.0 source code.

PR #6028, opened 2022-05-29
Fix case clause errors in erl_erts_errors

josevalim:
`Cause` defaults to `none`, not `true`.
A catch-all was used for consistency with
the remaining of the file.


bjorng:
Can you rebase the branch to `maint` so that we can include it in OTP 25.1?

josevalim:
@bjorng done!

bjorng:
Thanks for your pull request.

PR #6030, opened 2022-05-30
Improve observer nodes list menu

gomoripeti:
- Also include connected, hidden nodes ([Observer User Guide](https://www.erlang.org/doc/apps/observer/observer_ug.html#getting-started) uses a hidden node as example)
- Always include current node. If dist_listen is false it won't show up
in epdm names
- If the default epmd module `erl_epmd` is used and `erl_epmd_port` is set
to `DistPort` then it is only possible to connect to the node listening
on `DistPort` (if any), so exclude other nodes registered in EPMD

The last change is controversial, maybe it should be implemented elsewhere (in net_adm or erl_epmd?) and it's not the responsibility of observer to know about this detail. Or the exclusion should not happen at all (Observer shows a nice error dialogue when connection fails and maybe that's good enough)

### DIFF
--- a/lib/observer/src/observer_wx.erl
+++ b/lib/observer/src/observer_wx.erl
@@ -747,17 +747,20 @@ get_nodes() ->
     Nodes0 = case erlang:is_alive() of
 		false -> [];
 		true  ->
-		    case net_adm:names() of
-			{error, _} -> nodes();
-			{ok, Names} ->
-			    epmd_nodes(Names) ++ nodes()
-		    end
+                    case net_adm:names() of
+                        {error, _} -> [];
+                        {ok, Names} -> epmd_nodes(Names)
+                    end
+                    ++
+                    nodes(connected)
 	     end,
     Nodes = lists:usort(Nodes0),
+    WarningText = "WARNING: connecting to non-erlang nodes may crash them",
     {_, Menues} =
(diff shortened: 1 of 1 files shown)


dgud:
> Could there be some check only for connected nodes if they support observer functionality and only list those that do?
I don't know how to test that.

> Do c/java nodes show up in epmd names? (If yes, they would also be listed without being connected)
They can do that but then they expose themselves to maybe get rpc:call's so I think that is ok.

gomoripeti:
A connected node is listed among `nodes(hidden)` if either the local node or the remote node is hidden. Would it be an acceptable heuristic that if the current node is hidden then entries from `nodes(hidden)` are listed? (There is an edge case when a c/java node is connected to a hidden node - but I dont know how likely that is)

dgud:
After (not) sleeping on this for a couple of nights, I I think I like your last suggestion :-)

gomoripeti:
happy to hear that, I will update the PR accordingly
wish you some sleep-full nights :)

gomoripeti:
there are some recent changes in net_kernel around hidden type which I made use of. Naming is bit inconsistent with `connection_type` and `publish_type` having the same values (but bit different use)

gomoripeti:
ok, I will add the warning text
(had to try to know it was a bad idea)

PR #6031, opened 2022-05-30
JIT: Optimize creation of binaries

bjorng:
In the JIT, improve the code generation for binary construction for segments of fixed widths of no more than 64 bits and for variable width segments that are a set to 0.
    
I have run the benchmark linked to from https://github.com/erlang/otp/issues/5639.
    
On my Intel iMac from 2017, the result without this pull request was:
    
        == Testing with 1 MB ==
        fun base64:encode/1: 1000 iterations in 16264 ms: 61 it/sec
        fun base64:decode/1: 1000 iterations in 18597 ms: 53 it/sec
    
With this pull request:
    
        == Testing with 1 MB ==
        fun base64:encode/1: 1000 iterations in 10955 ms: 91 it/sec
        fun base64:decode/1: 1000 iterations in 10629 ms: 94 it/sec
    
On my M1 MacBook Pro from 2020 the result before was:
    
        == Testing with 1 MB ==
        fun base64:encode/1: 1000 iterations in 12299 ms: 81 it/sec
        fun base64:decode/1: 1000 iterations in 15147 ms: 66 it/sec
    
After:
    
        == Testing with 1 MB ==
        fun base64:encode/1: 1000 iterations in 8550 ms: 116 it/sec
        fun base64:decode/1: 1000 iterations in 10066 ms: 99 it/sec

### DIFF
--- a/erts/emulator/beam/jit/arm/instr_bs.cpp
+++ b/erts/emulator/beam/jit/arm/instr_bs.cpp
@@ -1429,10 +1429,49 @@ void BeamGlobalAssembler::emit_bs_bit_size_shared() {
     a.ret(a64::x30);
 }
 
+/*
+ * ARG1 = tagged bignum term
+ */
+void BeamGlobalAssembler::emit_get_sint64_shared() {
+    Label success = a.newLabel();
+    Label fail = a.newLabel();
+
+    emit_is_boxed(fail, ARG1);
+    arm::Gp boxed_ptr = emit_ptr_val(TMP3, ARG1);
+    a.ldr(TMP1, emit_boxed_val(boxed_ptr));
+    a.ldr(TMP2, emit_boxed_val(boxed_ptr, sizeof(Eterm)));
+    a.and_(TMP1, TMP1, imm(_TAG_HEADER_MASK));
+    a.cmp(TMP1, imm(POS_BIG_SUBTAG));
+    a.b_eq(success);
+
+    a.cmp(TMP1, imm(NEG_BIG_SUBTAG));
+    a.b_ne(fail);
+
+    a.neg(TMP2, TMP2);
+
+    a.bind(success);
+    {
+        a.mov(ARG1, TMP2);
+        /* Clear Z flag.
+         *
+         * TMP1 is known to be POS_BIG_SUBTAG or NEG_BIG_SUBTAG at this point.
+         */
+        ERTS_CT_ASSERT(POS_BIG_SUBTAG != 0 && NEG_BIG_SUBTAG != 0);
+        a.tst(TMP1, TMP1);
+        a.ret(a64::x30);
+    }
+
+    a.bind(fail);
+    {
+        a.tst(ZERO, ZERO);
+        a.ret(a64::x30);
+    }
+}
+
 struct BscSegment {
     BscSegment()
             : type(am_false), unit(1), flags(0), src(ArgNil()), size(ArgNil()),
-              error_info(0), effectiveSize(-1) {
+              error_info(0), effectiveSize(-1), action(action::DIRECT) {
     }
 
     Eterm type;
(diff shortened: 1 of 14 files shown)


erts/emulator/beam/jit/x86/instr_bs.cpp
```diff
@@ -1582,10 +1582,44 @@ void BeamGlobalAssembler::emit_bs_create_bin_error_shared() {
...
+
+    a.bind(success);
+    a.mov(ARG1, ARG3);
+    mov_imm(ARG2, 1);
+    a.test(ARG2, ARG2); /* Clear Z flag */
+    a.ret();
+
+    a.bind(fail);
+    a.xor_(ARG2, ARG2); /* Set Z flag */
+    a.ret();
```
jhogberg, line 1616:
```suggestion
    a.bind(success);
    {
        a.mov(ARG1, ARG3);
        /* Clear Z flag.
         *
         * ARG2 is known to be POS_BIG_SUBTAG or NEG_BIG_SUBTAG at this point. */
        ERTS_CT_ASSERT(POS_BIG_SUBTAG && NEG_BIG_SUBTAG);
        a.test(ARG2, ARG2);
        a.ret();
    }

    a.bind(fail);
    {
        a.xor_(ARG2, ARG2); /* Set Z flag */
        a.ret();
    }
```

erts/emulator/beam/jit/x86/instr_bs.cpp
```diff
@@ -1582,10 +1582,44 @@ void BeamGlobalAssembler::emit_bs_create_bin_error_shared() {
     a.jmp(labels[raise_exception_shared]);
 }
 
+/*
+ * ARG1 = tagged bignum term
+ *
+ * On return, Z is set if ARG1 is not a bignum. Otherwise, Z is clear and
+ * ARG1 is ththe 64 least significant bits of the bignum.
```
jhogberg, line 1589:
```suggestion
 * ARG1 is the 64 least significant bits of the bignum.
```

erts/emulator/beam/jit/x86/instr_bs.cpp
```diff
@@ -1596,8 +1630,187 @@ struct BscSegment {
...
+}
+
+void BeamModuleAssembler::set_zero(Sint effectiveSize) {
+    Label words = a.newLabel();
+    Label less_than_a_word = a.newLabel();
+    Sint store_unit = 1;
+
+    update_bin_state(RET, ARG2, -1, ARG3);
+
+    if (effectiveSize >= 128) {
```
jhogberg, line 1736: When the size is unknown, `REP STOSB` is quicker and more compact than manual zeroing on modern processors (after "Enhanced REP MOVSB support" was added in Ivy Bridge circa 2013). Specialized AVX implementations start to be slightly quicker after about two cache lines, but that feels rare enough that I don't think we need to care about it.

erts/emulator/beam/jit/x86/instr_bs.cpp
```diff
@@ -1596,8 +1630,187 @@ struct BscSegment {
...
+
+        a.bind(done);
+    } else if (effectiveSize % (store_unit * 8 * sizeof(Eterm)) != 0) {
+        /* The size is known, and we know that there are less than
+         * 128 bits to initialize. */
+        ASSERT(store_unit = 1);
+
+        if ((effectiveSize & 127) >= 64) {
+            a.mov(x86::Mem(ARG1, 0, 8), RET);
+            a.add(ARG1, imm(8));
```
jhogberg, line 1790: It would be faster and more compact to use static offsets here and below, doing a single update of `ARG1` at the end if the caller expects it to be updated.

erts/emulator/beam/jit/arm/instr_bs.cpp
```diff
@@ -1429,10 +1429,41 @@ void BeamGlobalAssembler::emit_bs_bit_size_shared() {
...
+
+    a.bind(success);
+    a.mov(ARG1, TMP2);
+    mov_imm(TMP1, 1);
+    a.tst(TMP1, TMP1);
+    a.ret(a64::x30);
+
+    a.bind(fail);
+    a.tst(ZERO, ZERO);
+    a.ret(a64::x30);
```
jhogberg, line 1460:
```suggestion
    a.bind(success);
    {
        a.mov(ARG1, TMP2);
        /* Clear Z flag.
         *
         * TMP1 is known to be POS_BIG_SUBTAG or NEG_BIG_SUBTAG at this point. */
        ERTS_CT_ASSERT(POS_BIG_SUBTAG && NEG_BIG_SUBTAG);
        a.tst(TMP1, TMP1);
        a.ret(a64::x30);
    }

    a.bind(fail);
    {
        a.tst(ZERO, ZERO);
        a.ret(a64::x30);
    }
```

erts/emulator/beam/jit/x86/instr_bs.cpp
bjorng: Changed to use `REP STOSB` if the size is unknown or greater than 16 bytes.

erts/emulator/beam/jit/x86/instr_bs.cpp
bjorng: Done.

erts/emulator/beam/jit/arm/instr_bs.cpp
```diff
@@ -1443,8 +1482,174 @@ struct BscSegment {
 
     Uint error_info;
     Sint effectiveSize;
+    enum action { NONE, ACCUMULATE_FIRST, ACCUMULATE, STORE } action;
```
jhogberg, line 1485: Using `enum class` will prevent setting this to invalid values by accident, e.g. `seg.action = 0;` will no longer work.
```suggestion
    enum class action { NONE, ACCUMULATE_FIRST, ACCUMULATE, STORE } action;
```

erts/emulator/beam/jit/arm/instr_bs.cpp
```diff
@@ -1443,8 +1482,174 @@ struct BscSegment {
...
+}
+
+void BeamModuleAssembler::update_bin_state(arm::Gp bin_base,
+                                           arm::Gp bin_offset,
+                                           Sint bit_offset,
+                                           Sint size,
+                                           arm::Gp size_reg) {
+    int cur_bin_offset =
+            offsetof(ErtsSchedulerRegisters, aux_regs.d.erl_bits_state) +
+            offsetof(struct erl_bits_state, erts_current_bin_);
```
jhogberg, line 1547: Is there any particular reason for splitting this?
```suggestion
    int cur_bin_offset =
            offsetof(ErtsSchedulerRegisters, aux_regs.d.erl_bits_state.erts_current_bin_);
```

erts/emulator/beam/jit/arm/instr_bs.cpp
```diff
@@ -1443,8 +1482,174 @@ struct BscSegment {
...
+    Label less_than_a_word = a.newLabel();
+    Sint store_unit = 1;
+
+    update_bin_state(ARG1, ARG2, -1, -1, ARG3);
+
+    if (effectiveSize >= 128) {
+        /* Store two 64-bit words machine words when the size is
+         * known and at least 128 bits. */
+        store_unit = 2;
+    }
```
jhogberg, line 1592: As an aside, we can use a `store_unit` of 4 by using `STP` with vector registers.

bjorng: OK. Done that.

erts/emulator/beam/jit/x86/instr_bs.cpp
```diff
@@ -1596,8 +1638,200 @@ struct BscSegment {
 
     Uint error_info;
     Sint effectiveSize;
+    enum class action { NONE, ACCUMULATE_FIRST, ACCUMULATE, STORE } action;
```
garazdawi, line 1641: Would you mind adding a small description what the different actions do and when they trigger?

garazdawi (approved):

erts/emulator/beam/jit/x86/instr_bs.cpp
```diff
@@ -2044,42 +2365,261 @@ void BeamModuleAssembler::emit_i_bs_create_bin(const ArgLabel &Fail,
...
+                a.bind(done);
+                break;
+            }
+            case BscSegment::action::NONE:
+                comment("construct integer segment");
+                if (seg.effectiveSize >= 0) {
+                    mov_imm(ARG3, seg.effectiveSize);
+                } else if (seg.unit == 0) {
+                    /* Silly but legal. */
+                    mov_imm(ARG3, 0);
```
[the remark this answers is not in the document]
bjorng: Contrary to the comment, a unit size of zero is not legal (at least for `integer` segments). I have removed the code.

erts/emulator/beam/jit/x86/instr_bs.cpp
bjorng: I have renamed `NONE` to `DIRECT` and added more comments.

PR #6035, opened 2022-05-30
Add application:get_supervisor/1

josevalim:
It returns the Pid of the supervisor running at
the root of the given application.

This allows tools for introspection and debugging
to print supervision trees without relying on
private APIs.

### DIFF
--- a/lib/kernel/src/application_master.erl
+++ b/lib/kernel/src/application_master.erl
@@ -75,7 +75,7 @@ call(AppMaster, Req) ->
     AppMaster ! {Req, Tag, self()},
     receive 
 	{'DOWN', Ref, process, _, _Info} ->
-	    ok;
+	    error;
 	{Tag, Res} ->
 	    erlang:demonitor(Ref, [flush]),
 	    Res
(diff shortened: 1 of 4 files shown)


lib/kernel/src/application_master.erl
```diff
@@ -75,7 +75,7 @@ call(AppMaster, Req) ->
     AppMaster ! {Req, Tag, self()},
     receive 
 	{'DOWN', Ref, process, _, _Info} ->
-	    ok;
+	    error;
```
josevalim, line 78: The master process already returns `error` is the child
process is dead, so I choose to align it.

It should not affect anything, because all callers seem
to simply assume a tuple is returned. We only care about
this in case there is a race condition and then master
process terminates after we retrieve it.

okeuday:
@josevalim If you used the name `application:get_root_process/1` the concept would be more general because it looks like there isn't a requirement for the root process to be a supervisor.  It would also be helpful if an `application:get_group_leader/1` function existed too (just a call to `application_controller:get_master/1`).

josevalim:
I am almost sure it has to be a supervisor and part of the tooling assumes do. To quote the docs for the start/2 callback:

> The function is to return {ok,Pid} or {ok,Pid,State}, where Pid is the pid of the top supervisor and State is any term.

in any case, I will be glad to change the name anytime. :)

essen:
Maybe `get_top_supervisor`.

PR #6036, opened 2022-05-31
Fix `logger_formatter` about `msg` by making templates linear before split

gfngfn:
I have encountered a somewhat strange behavior of [`logger_formatter`](https://www.erlang.org/doc/man/logger_formatter.html) about `msg` in templates. This pull request attempts to remedy it.

The behavior is exemplified by the following application:

- https://github.com/gfngfn/a_strange_behavior_of_logger_formatter

In summary, `msg`s occurring at conditional branching `{metakey(), template(), template()}` in templates will not be formatted as intended.

Looking into the implementation of `logger_formatter:format`, I have found that the implementation presumably assumes that given templates are of the form `[Elem_1, …, Elem_{i - 1}, msg, Elem_{i + 1}, …, Elem_n]` or do not contain `msg`. According to [the definition of `logger_formatter:template()`](https://www.erlang.org/doc/man/logger_formatter.html#type-template), this assumption is not documented. Therefore, one of the following amendments seems to be needed:

1. State in the manual that `msg` must occur as a top level element of templates and at most once;
2. Fix the implementation of `logger_formatter:format` so that `msg`s can occur under conditional branching but at most once for every “path”; or
3. Fix the implementation of `logger_formatter:format` so that `msg`s can occur everywhere in templates possibly more than once.

The present pull request suggests an amendment based on the second approach. This is because using `msg` more than once in a “path” does not seem so usual and also its support will require a bit complicated implementation due to [the `chars_limit` option](https://www.erlang.org/doc/man/logger_formatter.html#type-config).

I would appreciate it if you could give any comments or suggestions (especially since this is my first pull request to this repository).

### DIFF
--- a/lib/kernel/src/logger_formatter.erl
+++ b/lib/kernel/src/logger_formatter.erl
@@ -48,7 +48,8 @@ format(#{level:=Level,msg:=Msg0,meta:=Meta},Config0)
     Config = add_default_config(Config0),
     Meta1 = maybe_add_legacy_header(Level,Meta,Config),
     Template = maps:get(template,Config),
-    {BT,AT0} = lists:splitwith(fun(msg) -> false; (_) -> true end, Template),
+    LinearTemplate = linearize_template(Meta1,Template),
+    {BT,AT0} = lists:splitwith(fun(msg) -> false; (_) -> true end, LinearTemplate),
     {DoMsg,AT} =
         case AT0 of
             [msg|Rest] -> {true,Rest};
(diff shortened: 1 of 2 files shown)


gfngfn:
Rebased the branch onto `maint` and pushed it with `--force` (because the CI workflow had failed probably due to some other cause than this pull request). The workflow seems passing in this time on the forked repository: https://github.com/gfngfn/otp/actions/runs/2429845241

garazdawi:
Thanks! I think the change looks good and makes sense. Putting it into testing to make sure it works on all OSs and configurations.

garazdawi:
thanks!

gfngfn:
Thank you for the review!

PR #6040, opened 2022-05-31
Make specs for hd/1 and tl/1 more specific 

eksperimental:
This also stream lines with the development of the analogous functions in Elixir:

- https://github.com/elixir-lang/elixir/blob/01361af9fe582c71abdb9746c854f6fe59fbf511/lib/elixir/lib/kernel.ex#L550
- https://github.com/elixir-lang/elixir/blob/01361af9fe582c71abdb9746c854f6fe59fbf511/lib/elixir/lib/kernel.ex#L1249-L1250

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -2994,12 +2994,17 @@ os_prompt%</pre>
       <fsummary>Head of a list.</fsummary>
       <desc>
         <p>Returns the head of <c><anno>List</anno></c>, that is,
-          the first element, for example:</p>
+          the first element.</p>
+        <p>It works with improper lists.</p>
+        <p>Examples:</p>
         <pre>
 > <input>hd([1,2,3,4,5]).</input>
 1</pre>
+        <pre>
+> <input>hd([first, second, third, so_on | improper_end]).</input>
+first</pre>
         <p>Allowed in guard tests.</p>
-        <p>Failure: <c>badarg</c> if <c><anno>List</anno></c> is the empty
+        <p>Failure: <c>badarg</c> if <c><anno>List</anno></c> is an empty
           list <c>[]</c>.</p>
       </desc>
     </func>
@@ -11853,7 +11858,9 @@ timestamp() ->
       <fsummary>Tail of a list.</fsummary>
       <desc>
         <p>Returns the tail of <c><anno>List</anno></c>, that is,
-          the list minus the first element, for example:</p>
+          the list minus the first element</p>
+        <p>It works with improper lists.</p>
+        <p>Examples:</p>
         <pre>
 > <input>tl([geesties, guilies, beasties]).</input>
 [guilies, beasties]</pre>
@@ -11868,7 +11875,7 @@ timestamp() ->
 improper_end</pre>
         <p>Allowed in guard tests.</p>
         <p>Failure: <c>badarg</c> if <c><anno>List</anno></c>
-          is the empty list <c>[]</c>.</p>
+          is an empty list <c>[]</c>.</p>
       </desc>
     </func>
 
--- a/erts/preloaded/src/erlang.erl
+++ b/erts/preloaded/src/erlang.erl
@@ -2197,8 +2197,9 @@ get_module_info(_Module, _Item) ->
     erlang:nif_error(undefined).
 
 %% Shadowed by erl_bif_types: erlang:hd/1
--spec hd(List) -> term() when
-      List :: [term(), ...].
+-spec hd(List) -> Head when
+      List :: nonempty_maybe_improper_list(),
+      Head :: term().
 hd(_List) ->
     erlang:nif_error(undefined).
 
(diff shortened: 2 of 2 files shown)


eksperimental:
It may require updating 
https://github.com/erlang/otp/blob/master/lib/dialyzer/src/erl_bif_types.erl
https://github.com/erlang/otp/blob/master/lib/dialyzer/src/erl_types.erl

essen:
The input is one thing, but for the output, if you have a `term()` as a possible type, then adding more types like `[] | term()` provides no value, it'll still be just a `term()`. So I'm not sure what that part accomplishes.

erts/preloaded/src/erlang.erl
```diff
@@ -2781,8 +2783,13 @@ term_to_iovec(_Term, _Options) ->
...
 %% Shadowed by erl_bif_types: erlang:tl/1
--spec tl(List) -> term() when
-      List :: nonempty_maybe_improper_list().
+-spec tl(List) -> Tail when
+      List :: nonempty_maybe_improper_list(T, End),
+      Tail :: maybe_improper_list(T, End) | End,
+      T :: term(),
+      End :: term().
+  
+when elem: term(), last: term()
```
essen, line 2792: I don't think this line is Erlang.

eksperimental: Oops. Removed. Thank you @essen

eksperimental:
> The input is one thing, but for the output, if you have a `term()` as a possible type, then adding more types like `[] | term()` provides no value, it'll still be just a `term()`. So I'm not sure what that part accomplishes.

Yes. It's more for documentation purposes.  I don't know if this is idiomatic in Erlang.

erts/doc/src/erlang.xml
```diff
@@ -2998,9 +2998,20 @@ os_prompt%</pre>
...
+       <p>Allowed in guard tests.</p>
         <p>Failure: <c>badarg</c> if <c><anno>List</anno></c> is the empty
           list <c>[]</c>.</p>
+        <pre>
+> <input>hd([]).</input>
+** exception error: bad argument
+     in function  hd/1
+        called as hd([])
+        *** argument 1: not a nonempty list
+</pre>
```
bjorng, line 3014: I don't see the point of this example. I suggest you remove it.

erts/doc/src/erlang.xml
```diff
@@ -11779,6 +11792,13 @@ improper_end</pre>
         <p>Allowed in guard tests.</p>
         <p>Failure: <c>badarg</c> if <c><anno>List</anno></c>
           is the empty list <c>[]</c>.</p>
+        <pre>
+> <input>tl([]).</input>
+** exception error: bad argument
+     in function  tl/1
+        called as tl([])
+        *** argument 1: not a nonempty list
+</pre>
```
bjorng, line 11801: I don't see the point of this example. I suggest you remove it.

erts/preloaded/src/erlang.erl
```diff
@@ -2197,8 +2197,11 @@ get_module_info(_Module, _Item) ->
     erlang:nif_error(undefined).
 
 %% Shadowed by erl_bif_types: erlang:hd/1
--spec hd(List) -> term() when
-      List :: [term(), ...].
+-spec hd(List) -> Head when
+      List :: nonempty_maybe_improper_list(T, Last),
+      Head :: term(),
+      T :: term(),
+      Last :: term().
```
bjorng, line 2204: I don't find the extra variables helpful.

```suggestion
-spec hd(List) -> Head when
      List :: nonempty_maybe_improper_list(),
      Head :: term().
```

erts/preloaded/src/erlang.erl
```diff
@@ -2781,8 +2784,11 @@ term_to_iovec(_Term, _Options) ->
     erlang:nif_error(undefined).
 
 %% Shadowed by erl_bif_types: erlang:tl/1
--spec tl(List) -> term() when
-      List :: nonempty_maybe_improper_list().
+-spec tl(List) -> Tail when
+      List :: nonempty_maybe_improper_list(T, End),
+      Tail :: maybe_improper_list(T, End) | End,
+      T :: term(),
+      End :: term().
```
bjorng, line 2791: I find the extra variables and the union in the `Tail` type confusing.

```suggestion
-spec tl(List) -> Tail when
      List :: nonempty_maybe_improper_list(),
      Tail :: term().
```

bjorng (changes requested):
Thanks for your pull request.

I have suggested some improvements. If you agree, please squash the commits down to one commit when you have done the changes.

erts/doc/src/erlang.xml
eksperimental: done

erts/preloaded/src/erlang.erl
eksperimental: done

erts/doc/src/erlang.xml
eksperimental: done

eksperimental:
Please review again @bjorng .
Additionally I streamline both examples, and change "is the empty list" to "is an empty list".

bjorng (approved):

bjorng:
Thanks! Add to our daily builds (along with updated preloaded modules).

bjorng:
Thanks for your pull request.

PR #6041, opened 2022-06-01
ssl: Reject unsupported previous version with protocol alert

### LINKED ISSUE #5950: TLS1.3 handshake fails for some SSL clients
**Describe the bug**
When a client attempts to use BearSSL (and possibly other TLS implementations) to access my service, my service crashes; the service is based on OTP's `ssl` library, that is, the server is running on OTP and using it to provide a TLS-wrapped listener that it wants to accept connections on. The bug only happens with a subset of the TLS client implemenations that are available, and I can only easily reproduce it with BearSSL, but I believe there exist several other client implementations which expose the same problem. Many client implementations work *fine* and consistently fail to expose the bug.

As I understand it, a client ought not to be able to crash OTP's TLS server *at all*, no matter what it transmits.

The crashes all look like this:

```
12:59:04.909 [error] gen_statem <0.347.0> in state internal terminated with reason: {error,[tls_connection_1_3]}
12:59:04.912 [error] CRASH REPORT Process <0.347.0> with 1 neighbours crashed with reason: no function clause matching tls_handshake_1_3:get_supported_groups({elliptic_curves,[{1,2,840,10045,3,1,7},{1,3,132,0,34},{1,3,132,0,35},{1,3,101,110}]}) line 2448
12:59:04.916 [error] Supervisor tls_connection_sup had child undefined started with {ssl_gen_statem,start_link,undefined} at <0.347.0> exit with reason no function clause matching tls_handshake_1_3:get_supported_groups({elliptic_curves,[{1,2,840,10045,3,1,7},{1,3,132,0,34},{1,3,132,0,35},{1,3,101,110}]}) line 2448 in context child_terminated
12:59:04.917 [error] CRASH REPORT Process <0.348.0> with 0 neighbours exited with reason: {{function_clause,[{tls_handshake_1_3,get_supported_groups,[{elliptic_curves,[{1,2,840,10045,3,1,7},{1,3,132,0,34},{1,3,132,0,35},{1,3,101,110}]}],[{file,"tls_handshake_1_3.erl"},{line,2448}]},{tls_handshake_1_3,do_start,2,[{file,"tls_handshake_1_3.erl"},{line,619}]},{tls_connection_1_3,start,3,[{file,"tls_connection_1_3.erl"},{line,268}]},{gen_statem,loop_state_callback,11,[{file,"gen_statem.erl"},{line,1194}]},{tls_connection_1_3,init,1,[{file,"tls_connection_1_3.erl"},{line,194}]},{proc_lib,...}]},...} in gen:do_call/4 line 220
```

**To Reproduce**
The error seems to occur during the response to the initial SSL handshake.

1) start a TLS-wrapped TCP listener
2) attempt to connect via BearSSL

Specifically,

1) start my `blizanci` server, which is an Erlang/OTP process that listens on a TCP port in the usual way
2) try to connect to it with `gmni`, which uses BearSSL under the hood.

**Expected behavior**
A normal SSL handshake would occur, without crashing the Erlang process and terminating the connection.

**Affected versions**

```
Erlang/OTP 24 [erts-12.0.3] [source] [64-bit] [smp:1:1] [ds:1:1:10] [async-threads:1] [jit]
```
and
```
[{sasl,"SASL  CXC 138 11","4.1"},
 {blizanci,"A Gemini protocol server","4.1.0"}, 
 {ppool,"Worker pool library from LearnYouSomeErlang",
        "1.1.1"},
 {mime_lookup,"MIME types lookup service","0.1.1"},
 {realpath,"Filename canonicaliser","1.3.6"},
 {erlexec,"OS Process Manager","1.18.8"},
 {ranch,"Socket acceptor pool for TCP protocols.","2.0.0"},
 {lager,"Erlang logging framework","3.9.2"},
 {goldrush,"Erlang event stream processor","0.1.9"},
 {compiler,"ERTS  CXC 138 10","8.0.2"},
 {syntax_tools,"Syntax tools","2.6"},
 {runtime_tools,"RUNTIME_TOOLS","1.16.2"},
 {inets,"INETS  CXC 138 49","7.4.1"},
 {ssl,"Erlang/OTP SSL application","10.4.2"},
 {public_key,"Public key infrastructure","1.11.1"},
 {asn1,"The Erlang ASN1 compiler version 5.0.16","5.0.16"},
 {crypto,"CRYPTO","5.0.2"},
 {stdlib,"ERTS  CXC 138 10","3.15.2"},
 {kernel,"ERTS  CXC 138 10","8.0.2"}]
```

**Additional context**
Not a lot here. The thing reliably occurs whenever accepting a TLS/1.3 connection (but not other versions), and appears to implicate the handling of elliptic curve -based ciphers.

IngelaAndin:
Closes #5950

### DIFF
--- a/lib/ssl/src/dtls_connection.erl
+++ b/lib/ssl/src/dtls_connection.erl
@@ -324,14 +324,12 @@ hello(internal, #hello_verify_request{cookie = Cookie},
                         },
     dtls_gen_connection:next_event(?FUNCTION_NAME, no_record, State, Actions);
 hello(internal, #client_hello{extensions = Extensions} = Hello,
-      #state{ssl_options = #{handshake := hello},
-             handshake_env = HsEnv,
+      #state{handshake_env = #handshake_env{continue_status = pause},
              start_or_recv_from = From} = State0) ->
     try tls_dtls_connection:handle_sni_extension(State0, Hello) of
         #state{} = State ->
-            {next_state, user_hello, State#state{start_or_recv_from = undefined,
-                                                 handshake_env = HsEnv#handshake_env{hello = Hello}},
-             [{reply, From, {ok, Extensions}}]}
+            {next_state, user_hello, State#state{start_or_recv_from = undefined},
+             [{postpone, true}, {reply, From, {ok, Extensions}}]}
     catch throw:#alert{} = Alert ->
             alert_or_reset_connection(Alert, ?FUNCTION_NAME, State0)
     end;
@@ -354,16 +352,11 @@ hello(internal, #client_hello{cookie = Cookie} = Hello, #state{static_env = #sta
                     hello(internal, Hello#client_hello{cookie = <<>>}, State)
             end
     end;
-hello(internal, #server_hello{extensions = Extensions} = Hello, 
-      #state{ssl_options = #{
-                 handshake := hello},
-             handshake_env = HsEnv,
+hello(internal, #server_hello{extensions = Extensions}, 
+      #state{handshake_env = #handshake_env{continue_status = pause},
              start_or_recv_from = From} = State) ->
-    {next_state, user_hello, State#state{start_or_recv_from = undefined,
-                                         handshake_env = HsEnv#handshake_env{
-                                                           hello = Hello}},
-     [{reply, From, {ok, Extensions}}]};       
-
+    {next_state, user_hello, State#state{start_or_recv_from = undefined},
+     [{postpone, true},{reply, From, {ok, Extensions}}]};
 hello(internal, #server_hello{} = Hello,
       #state{
          static_env = #static_env{role = client},
--- a/lib/ssl/src/ssl_connection.hrl
+++ b/lib/ssl/src/ssl_connection.hrl
@@ -55,7 +55,7 @@
 
 
 -record(handshake_env, {
-                        client_hello_version  :: ssl_record:ssl_version() | 'undefined',
+                        client_hello_version  :: ssl_record:ssl_version() | 'undefined', %% Legacy client hello
                         unprocessed_handshake_events = 0    :: integer(),
                         tls_handshake_history :: ssl_handshake:ssl_handshake_history() | secret_printout()
                                                | 'undefined',
@@ -67,7 +67,9 @@
                         early_data_accepted = false :: boolean(), %% TLS 1.3
                         allow_renegotiate = true                    ::boolean(),
                         %% Ext handling
-                        hello,                %%:: #client_hello{} | #server_hello{}            
+                        %% continue_status reflects handling of the option handshake that is either full or
+                        %% hello (will pause at hello message to allow user to act on hello extensions)
+                        continue_status = full,  %% full | pause | {pause, ClientVersionsExt} | continue
                         sni_hostname = undefined,
                         max_frag_enum :: undefined | {max_frag_enum, integer()},
                         expecting_next_protocol_negotiation = false ::boolean(),
--- a/lib/ssl/src/tls_connection.erl
+++ b/lib/ssl/src/tls_connection.erl
@@ -111,7 +111,8 @@
 %% Setup
 -export([init/1]).
 
--export([renegotiate/2]).
+-export([renegotiate/2,
+        choose_tls_fsm/2]).
 
 %% gen_statem state functions
 -export([initial_hello/3,
@@ -207,23 +208,18 @@ config_error(Type, Event, State) ->
 	    #state{}) ->
 		   gen_statem:state_function_result().
 %%--------------------------------------------------------------------
-hello(internal, #client_hello{extensions = Extensions} = Hello, 
-      #state{ssl_options = #{handshake := hello},
-             static_env = #static_env{role = server},
-             handshake_env = HsEnv,
+hello(internal, #client_hello{extensions = Extensions},
+      #state{static_env = #static_env{role = server},
+             handshake_env = #handshake_env{continue_status = pause},
              start_or_recv_from = From} = State) ->
-    {next_state, user_hello, State#state{start_or_recv_from = undefined,
-                                         handshake_env = HsEnv#handshake_env{hello = Hello}},
-     [{reply, From, {ok, Extensions}}]};
-hello(internal, #server_hello{extensions = Extensions} = Hello,
-      #state{ssl_options = #{handshake := hello},
+    {next_state, user_hello, State#state{start_or_recv_from = undefined},
+     [{postpone, true}, {reply, From, {ok, Extensions}}]};
+hello(internal, #server_hello{extensions = Extensions},
+      #state{handshake_env = #handshake_env{continue_status = pause},
              static_env = #static_env{role = client},
-             handshake_env = HsEnv,
-             start_or_recv_from = From} = State) ->   
+             start_or_recv_from = From} = State) ->
     {next_state, user_hello,
-     State#state{start_or_recv_from = undefined,
-                 handshake_env = HsEnv#handshake_env{
-                                   hello = Hello}}, [{reply, From, {ok, Extensions}}]};
+     State#state{start_or_recv_from = undefined}, [{postpone, true}, {reply, From, {ok, Extensions}}]};
 hello(internal, #client_hello{client_version = ClientVersion} = Hello,
       #state{static_env = #static_env{role = server}, connection_env = CEnv} = State0) ->
     try
--- a/lib/ssl/src/tls_connection_1_3.erl
+++ b/lib/ssl/src/tls_connection_1_3.erl
@@ -237,19 +237,33 @@ user_hello({call, From}, cancel, State) ->
     ssl_gen_statem:handle_own_alert(?ALERT_REC(?FATAL, ?USER_CANCELED, user_canceled),
                                     ?FUNCTION_NAME, State);
 user_hello({call, From}, {handshake_continue, NewOptions, Timeout},
-           #state{static_env = #static_env{role = Role},
-                  handshake_env = #handshake_env{hello = Hello},
+           #state{static_env = #static_env{role = client = Role},
+                  handshake_env = HSEnv,
                   ssl_options = Options0} = State0) ->
-    Options = ssl:handle_options(NewOptions, Role, Options0#{handshake => full}),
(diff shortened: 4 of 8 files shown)


u3s (approved):

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -257,41 +257,72 @@ user_hello(_, _, _) ->
...
+                {State, negotiated} ->
+                    {next_state, negotiated, State, [{next_event, internal, {start_handshake, undefined}}]};
+                {State, negotiated, PSK} ->  %% Session Resumption with PSK
+                    {next_state, negotiated, State, [{next_event, internal, {start_handshake, PSK}}]}
+            end;
+        false ->
+            ssl_gen_statem:handle_own_alert(?ALERT_REC(?FATAL, ?PROTOCOL_VERSION), start, State0)
+        end;
+start(internal, #client_hello{}, State0) ->
+    ssl_gen_statem:handle_own_alert(?ALERT_REC(?FATAL, ?PROTOCOL_VERSION), ?FUNCTION_NAME, State0);
```
KennethL, line 293: I would like a comment for each clause matching `#client_hello`explaining what it means and when it can occur.
As I understand it the clauses handle the following cases:
1. The user has requested to handle part of the handshake indicated by existence of the `{handshake, hello}` option. We  return to the user which can continue the handshake
2. The client has sent which versions it wants to negotiate and we will check it that is acceptable
3. ??? When can this happen (a `#client_hello{}`without the extension `client_hello_versions`

I suggest that we merge the first 2 clauses together and that the test regarding the `handshake` option is done only if `tls_record:is_acceptable_version({3,4}, ClientVersions)` returns `true`.

KennethL (changes requested):

lib/ssl/src/tls_connection_1_3.erl
IngelaAndin: Sure I will add some comments. I actually think there are some more things that we might need to think about for this issue so I will work on a new version.

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -237,61 +237,94 @@ user_hello({call, From}, cancel, State) ->
...
+        false ->
+            ssl_gen_statem:handle_own_alert(?ALERT_REC(?FATAL, ?PROTOCOL_VERSION), ?FUNCTION_NAME, State)
+    end;
+start(internal, #client_hello{extensions = #{client_hello_versions :=
+                                                 #client_hello_versions{versions = ClientVersions}
+                                            }= Extensions},
+      #state{ssl_options = #{handshake := hello},  %% User requested pause to be able to react to hello extensions.
              start_or_recv_from = From,
-             handshake_env = HsEnv} = State) ->
+             handshake_env = #handshake_env{handshake_continue = Continue} = HSEnv} = State) when Continue =/= continue ->
```
u3s, line 289: (low priority comment about variable naming style)

that part: `Continue =/= continue`

I understand it, it is probably fully functionally correct.
But from ease of code reading perspective ... maybe variable name could be adjusted ... `ContinuePhase`, `ContinueStage`, maybe some other word used in RFC ?

It might sound ridiculous, but I think that as long as it come at no cost and is doing the job - code should be readable for someone not knowing Erlang.

lib/ssl/src/tls_connection_1_3.erl
IngelaAndin: I made a new version now, that will avoid double-checking the version when {handshake, hello} is used. I took the opportunity to get rid of a legacy mechanism and use gen_statem to postpone the handling of the client/server-hello until we receive the continuation with possible new options. I think this can also warrant some work on version handling in TLS-1.2 with considerations of adoption of some TLS-1.3 mechanisms mentioned in the RFC that preferably should affect also TLS-1.2 which I am not certain that we have adopted. This I think should be handled as a separate issue I will make a ticket.

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -237,61 +237,94 @@ user_hello({call, From}, cancel, State) ->
     ssl_gen_statem:handle_own_alert(?ALERT_REC(?FATAL, ?USER_CANCELED, user_canceled),
                                     ?FUNCTION_NAME, State);
 user_hello({call, From}, {handshake_continue, NewOptions, Timeout},
-           #state{static_env = #static_env{role = Role},
-                  handshake_env = #handshake_env{hello = Hello},
+           #state{static_env = #static_env{role = client = Role},
+                  handshake_env = HSEnv,
                   ssl_options = Options0} = State0) ->
-    Options = ssl:handle_options(NewOptions, Role, Options0#{handshake => full}),
+    Options= ssl:handle_options(NewOptions, Role, Options0),
```
u3s, line 243: if you would be making a next version, consider replacing all `Options=` with `Options =`.
Ignore otherwise.

u3s (approved):

lib/ssl/src/tls_connection_1_3.erl
IngelaAndin: I will fix this, I agree it is more readable with the space ( I usually do that, I consider it a typo).

lib/ssl/src/tls_connection_1_3.erl
IngelaAndin: It does not sound ridiculous :) I changed in an other way so that we get a positive test and do not need the variable!

lib/ssl/src/dtls_connection.erl
```diff
@@ -325,13 +325,12 @@ hello(internal, #hello_verify_request{cookie = Cookie},
     dtls_gen_connection:next_event(?FUNCTION_NAME, no_record, State, Actions);
 hello(internal, #client_hello{extensions = Extensions} = Hello,
       #state{ssl_options = #{handshake := hello},
-             handshake_env = HsEnv,
-             start_or_recv_from = From} = State0) ->
+             handshake_env = #handshake_env{handshake_continue = Continue},
+             start_or_recv_from = From} = State0) when Continue =/= continue ->
```
KennethL, line 329:
```suggestion
             handshake_env = #handshake_env{handshake_continue = continue},
             start_or_recv_from = From} = State0)  ->
```
If the variable `Continue` only is used in the guard testing for `continue` this can be done directly in the pattern and no guard is needed.

lib/ssl/src/ssl_connection.hrl
```diff
@@ -67,7 +67,7 @@
                         early_data_accepted = false :: boolean(), %% TLS 1.3
                         allow_renegotiate = true                    ::boolean(),
                         %% Ext handling
-                        hello,                %%:: #client_hello{} | #server_hello{}            
+                        handshake_continue = hello,  %% hello | client_hello_versions (when entering user_hello in TLS-1.3) | continue (after user_hello)
```
KennethL, line 70: I think the name if this new field `handshake_continue` is confusing. Maybe it would be better if it was named `handshake_mode` or `handshake_progress` and since it is a field in the record `handshake_env` the prefix `handshake` can be omitted and we land in `mode` or `progress`. What do you think?

lib/ssl/src/tls_connection.erl
```diff
@@ -207,23 +208,20 @@ config_error(Type, Event, State) ->
...
       #state{ssl_options = #{handshake := hello},
              static_env = #static_env{role = server},
-             handshake_env = HsEnv,
-             start_or_recv_from = From} = State) ->
-    {next_state, user_hello, State#state{start_or_recv_from = undefined,
-                                         handshake_env = HsEnv#handshake_env{hello = Hello}},
-     [{reply, From, {ok, Extensions}}]};
-hello(internal, #server_hello{extensions = Extensions} = Hello,
+             handshake_env = #handshake_env{handshake_continue = Continue},
+             start_or_recv_from = From} = State) when Continue =/= continue ->
```
KennethL, line 215:
```suggestion
             handshake_env = #handshake_env{handshake_continue = continue},
             start_or_recv_from = From} = State)  ->
```
The guard is not necessary

lib/ssl/src/tls_connection.erl
```diff
@@ -207,23 +208,20 @@ config_error(Type, Event, State) ->
...
+    {next_state, user_hello, State#state{start_or_recv_from = undefined},
+     [{postpone, true}, {reply, From, {ok, Extensions}}]};
+hello(internal, #server_hello{extensions = Extensions},
       #state{ssl_options = #{handshake := hello},
+             handshake_env = #handshake_env{handshake_continue = Continue},
              static_env = #static_env{role = client},
-             handshake_env = HsEnv,
-             start_or_recv_from = From} = State) ->   
+             start_or_recv_from = From} = State) when Continue =/= continue ->
```
KennethL, line 222: The guard is not necessary

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -237,61 +237,94 @@ user_hello({call, From}, cancel, State) ->
...
+                                            handshake_env = HSEnv#handshake_env{handshake_continue = continue}},
+             [{{timeout, handshake}, Timeout, close}]};
+        undefined ->
+            ssl_gen_statem:handle_own_alert(?ALERT_REC(?FATAL, ?PROTOCOL_VERSION), ?FUNCTION_NAME, State);
+        _Else ->
+            {next_state, hello, State#state{start_or_recv_from = From,
+                                            handshake_env = HSEnv#handshake_env{handshake_continue = continue}},
+             [{change_callback_module, tls_connection},
+              {{timeout, handshake}, Timeout, close}]}
+    end;
```
KennethL, line 266: When does this happen?

KennethL (changes requested):

lib/ssl/src/ssl_connection.hrl
KennethL: Another possible name could be `continue_status`

IngelaAndin: Sure

lib/ssl/src/tls_connection_1_3.erl
IngelaAndin: This is a corner case that could happen for the server if the user changes the version option and client happens to work a certain way.

IngelaAndin:
The content of this PR will soon be included in a patch, but as the solution now has a merge conflict I need to handle this locally.

PR #6044, opened 2022-06-01
erts: Improve internal hashing

jhogberg:
x86-64 has come a long way since the first processors were released in 2003, but most of us still build Erlang/OTP so that it could in theory run on a vintage Athlon 64. This PR introduces a configure flag to make it easier to use new instruction set extensions, and emits a warning when it hasn't been used to remind people that they're leaving performance on the table.

For example, map lookups become a lot quicker when the compiler is free to emit instructions for "bit count" and "leading zero count" instead of having to twiddle the bits manually, and internal hashing (used for maps, ETS, etc) can likewise become quicker if we use hash instructions instead of the current Jenkins96 hash.

To demonstrate this, I've included a quick-and-dirty implementation around the `crc32` / `crc32cw` instructions in this PR. The results are pretty sweet:

```erlang
-module(t).
-export([t/0, id/1]).

t() ->
    Map = maps:from_keys(lists:seq(1, 1000), id(1)),
    erlang:garbage_collect(),
    element(1, timer:tc(fun() -> bench(100000000, id(1), Map) end)).

bench(0, _, _) ->
    ok;
bench(N, Key, Map) ->
    #{ Key := I } = Map,
    bench(N - I, Key, Map).

id(I) -> I.
```

Built with `./otp_build setup`:
```
1> t:t().
1627577
2> t:t().
1648546
3> t:t().
1627494
```
Built with `./otp_build setup --with-march=skylake`:

```
1> t:t().
1155951
2> t:t().
1151126
3> t:t().
1129490
```


garazdawi:
I wonder if it is not better for us to add some reasonable default value that the user will have to switch off if they do not want it?

I don't really like that running configure without any options always emits a big warning.

garazdawi:
Also, would it be possible to do the refactoring and improvements of the hash function in separate commits so that it is possible to see what changes have been made?

KennethL:
I also think that we should have reasonable defaults assuming reasonable modern CPU architecture. And that the user has to turn off support for features instead of turning them on.

By the way what happens if an alternative not supported by the current CPU is given to configure? Will it test it and fail in an understandable way?

Also if a binary of `erts` is built with use of certain instructions will that crash in a bad way on a cpu that does not support them or can this be tested at startup and then we can do a graceful and for the user understandable termination of `erts`?

Not that I am very positive to that we try to use new instructions if they can give significant performance improvements.

jhogberg:
> I wonder if it is not better for us to add some reasonable default value that the user will have to switch off if they do not want it?

> I also think that we should have reasonable defaults assuming reasonable modern CPU architecture. And that the user has to turn off support for features instead of turning them on.

My first attempt did that, with a check of all extensions at startup so that it would fail immediately with an error message telling the user to rebuild with a certain flag, but after discussing it with @rickard-green we decided to leave it entirely to the user.

The gist of it is that if someone distributes binary builds for their Linux distribution, using modern instructions might break for some end-user that uses old hardware, so we can't be as aggressive as we'd like in our defaults. Even something as simple as requiring SSE4.2 will break for people who run on reasonably modern Intel Atom processors.

I'm not a huge fan of the warning either, and there's always the risk that a user sets this and forgets about it for 10 years, losing out on progress they'd otherwise have if we had a default that got regularly bumped. We need to decide which is the lesser evil, and I'm fine either way.

> Also, would it be possible to do the refactoring and improvements of the hash function in separate commits so that it is possible to see what changes have been made?

Sure thing, I've pushed an update with that change now.

erts/emulator/beam/erl_term_hashing.c
```diff
@@ -0,0 +1,1724 @@
...
+                (((((hash)*(Prime1) + (x & 0xFF)) * (Prime1) +                \
+                ((x >> 8) & 0xFF)) * (Prime1) +                                \
+                ((x >> 16) & 0xFF)) * (Prime1) +                        \
+                 (x >> 24));                                                \
+        } while(0)
+
+#define UINT32_HASH_RET(Expr, Prime1, Prime2)           \
+        UINT32_HASH_STEP(Expr, Prime1);                        \
+        hash = hash * (Prime2);                                \
+        break
```
bjorng, line 191: While at I, I suggest that you clean up the arrangement of the backslashes.

KennethL: I am not sure what you mean with this:
`My first attempt did that, with a check of all extensions at startup so that it would fail immediately with an error message telling the user to rebuild with a certain flag, but after discussing it with @rickard-green we decided to leave it entirely to the user.`

At startup of the Erlang VM we test if the instructions we intend to use (chosen at configure and defaults) actually is supported on the CPU where we are running. We then terminate in an understandable way.

With "leave entirely to the user" I don't understand what you mean. Should it just crash later at some random time depending on what the Erlang system is doing? Should it crash with the user reporting a suspected bug to us?

I would also think that we should clearly state that we don't guarantee that our binaries built on or for one system works on other systems without the user taking care of that by selecting the appropriate options at configure time. I also assume that cross builds has a role to play here. A cross build it intended for a specific OS and architecture, it is up to the user to get this right.

KennethL: I just talked to @rickard-green and know more now. You don't need to answer my previous comment

jhogberg:
I just saw your message about not needing to reply, but I figure might reply anyway in case someone's interested:

> At startup of the Erlang VM we test if the instructions we intend to use (chosen at configure and defaults) actually is supported on the CPU where we are running. We then terminate in an understandable way.

Yes, that's roughly what it did, sorry for being unclear.

When we select which extensions to use it's easy add checks for them. For example if we say that we're compiling with SSE4.2 then we can try to execute an SSE4.2 instruction and crash with a proper error message.

> With "leave entirely to the user" I don't understand what you mean. Should it just crash later at some random time depending on what the Erlang system is doing? Should it crash with the user reporting a suspected bug to us?

This PR leaves the problem to the user: they select which extensions they want to use, and then it might crash somewhere with `SIGILL` if they run it on an unsupported processor. It's entirely on them to do things right.

The reason we cannot add a feature test when the user selects the extensions is that there's no way to check it in a forward-compatible manner.

If Intel releases a cool new processor tomorrow and someone compiles with `--with-march=coolnewprocessor`, the compiler will happily set the `__COOL_NEW_INSTRUCTIONS__` feature macro, but if we haven't updated the check to account for that, then the check will erroneously pass and we're back to square one.

garazdawi:
I'm wondering a bit why we need a configure flag for `-march`? It has always been possible to do `./configure CFLAGS="-march=native -O2 -g"` to do the same thing.

jhogberg:
The main reason is that this started as `--enable/disable-modern-instructions` where we controlled the options :smile: 

I kept it a configure option as I wanted to warn users when they hadn't made a choice, because it's basically "free performance" for most people. While we can scan `CFLAGS` for the `-march` option, there's no way to know whether the user omitted it on purpose or not, so someone who really wants to compile for the baseline would still get a warning.

mikpe:
Could support for newer instructions be detected at configure-time? I would prefer if this happens automatically as the result of either CC/CXX defaulting to some newer arch level, or CFLAGS selecting that.

Also, "modern instructions" is a poor choice of option, since its meaning varies over time and architecture. IMO it would be better to have explicit --enable-<specific-instruction-or-ISA-level> options.

I would hate if OTP by default builds an executable that doesn't run on the build system.

mikpe:
> > Could support for newer instructions be detected at configure-time? I would prefer if this happens automatically as the result of either CC/CXX defaulting to some newer arch level, or CFLAGS selecting that.
> 
> That was the plan back when it was `--enable-modern-instructions`: one check at configure-time and then another at runtime in case someone moved it to a different machine. I removed those after changing it to `--with-march=arch` which left all responsibility to the user.

Ok, passing --with-march=arch to OTP's ./configure is entirely reasonable.

PR #6045, opened 2022-06-02
Introduce receive optimisation to dets calls

ssmyczynski:
Current implementation of `dets:req/2` function in `dets` module is not using receive optimisation.

Module `dets` is used as a backend for `mnesia`'s `disc_only_copies` tables. All dirty operations, or symmetric transactions (`sym_trans`) are executed in context of a single `mnesia_tm` process which means that in case of using `disc_only_copies` tables we are actually vulnerable to non-optimised `dets` calls in case when `mnesia_tm` message queue is busy.

The proposed solution introduces receive optimisation to avoid that potential problem.

### DIFF
--- a/lib/stdlib/src/dets.erl
+++ b/lib/stdlib/src/dets.erl
@@ -93,7 +93,8 @@
               tab_name/0]).
 
 -compile({inline, [{einval,2},{badarg,2},{undefined,1},
-                   {badarg_exit,2},{lookup_reply,2}]}).
+                   {badarg_exit,2},{lookup_reply,2},
+                   {pidof,1},{resp,2}]}).
 
 -include_lib("kernel/include/file.hrl").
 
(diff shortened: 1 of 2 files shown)


lib/stdlib/src/dets.erl
```diff
@@ -1237,15 +1238,27 @@ treq(Tab, R) ->
...
+    Pid;
+pidof(Pid) ->
+    Pid.
+
+%% Inlined.
+resp({Pid, Ref} = _From, Message) ->
+    Pid ! {self(), Ref, Message}, ok;
+resp(Pid, Message) ->
+    Pid ! {self(), Message}, ok.
+
```
mikpe, line 1261: Do you really need to keep support for replying to old Ref-less calls?

ssmyczynski: You are right, we do not need to keep it anymore. Fixed.

ssmyczynski: `dets` server sends some messages to itself without using references (`auto_save` | `may_grow` | `{delayed_write, _}`) but in that case it does not send nor expect any response.

mikpe:
LGTM

garazdawi:
Thanks!

PR #6048, opened 2022-06-03
erts: Apply -fno-omit-frame-pointer to erts/lib_src when JIT is enabled

d1jang:
Currently OMIT_OMIT_FP is set to true in erts/emulator/Makefile.in
when JIT is enabled, but not necessarily in erts/lib_src/Makefile.in.

This change factors out OMIT_OMIT_FP computation to be shared between
those two builds to apply this flag consistently to them.

Note that this bug was confirmed by @garazdawi in an Erlang forum post: https://erlangforums.com/t/unify-builds-for-otp-erts-targets/1419/3

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -270,6 +270,45 @@ jobs:
           name: otp_win32_installer
           path: otp/release/win32/otp*.exe
 
+  build-flavors:
+    name: Build Erlang/OTP (Types and Flavors)
+    runs-on: ubuntu-latest
+    needs: [pack, changed-apps]
+    if: contains(needs.changed-apps.outputs.changes, 'emulator')
+
+    steps:
+      - uses: actions/checkout@v2
+      ## Download docker images
+      - name: Cache BASE image
+        id: cache-base-image
+        uses: actions/cache@v3
+        with:
+            path: otp_docker_base.tar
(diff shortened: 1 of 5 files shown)


garazdawi:
Hello! Thanks for the PR! Would you mind signing the CLA?

garazdawi (changes requested):
You need to also change [erts/configure.ac](https://github.com/erlang/otp/blob/master/erts/configure.ac#L3598) to include flavor.mk.in as [erts/configure](https://github.com/erlang/otp/blob/master/erts/configure) is a generated file.

d1jang:
> You need to also change [erts/configure.ac](https://github.com/erlang/otp/blob/master/erts/configure.ac#L3598) to include flavor.mk.in as [erts/configure](https://github.com/erlang/otp/blob/master/erts/configure) is a generated file.

Thanks for the comment! I included the requested change in [erts/configure.ac].

d1jang:
@garazdawi I made the requested change. Can I ask for a quick review? Thanks!

garazdawi:
Hello again! The changes in this PR break the building of a non-jit emulator. To reproduce run:

```
./otp_build setup -a
make FLAVOR=emu
```

garazdawi:
I've pushed changes to your branch to fix the issues and add CI in order to check that the different types and flavors works.

garazdawi (approved):

PR #6051, opened 2022-06-05
ets.xml: fix grammar in update_counter/N doc

mikpe:
The sentence talks about a singular function so it needs "updates" not "update".


bjorng:
Thanks for your pull request.

PR #6060, opened 2022-06-08
Fix public_key removed version

garazdawi:
The `public_key:ssh_*` functions were removed in Erlang/OTP 25, but were not labeled with the correct version in the DEPRECATIONS file.


HansN (approved):
OK.

Issue #5789, opened 2022-03-14
IPv6 multicast options are broken


binaryseed:
Using `gen_udp` multicast options with an IPv6 address fail with a `badarg` error. Beyond the argument parsing, the `inet_driver` code doesn't have support for handling IPv6 addresses when doing multicast operations (see https://github.com/erlang/otp/pull/5726#discussion_r814790826). The underlying multicast functionality does work however, once the kernel gets the correct socket options.

For example, opening a socket with a multicast interface and an IPv4 address:
```erlang
gen_udp:open(5000, [inet, binary, {multicast_if, {10, 11, 12, 1}}]).
```

This fails with an IPv6 address:
```erlang
gen_udp:open(5000, [inet6, binary, {multicast_if, {65152, 0, 0, 0, 0, 0, 0, 1}}]).
```

There is a workaround, which involves giving the kernel `raw` socket options, for example, joining a multicast group with this:

```erlang
Mad = <<65298:16, 0:16, 0:16, 0:16, 0:16, 0:16, 0:16, 1:16>>,
Ifindex = <<10:32/native>>,
Ip6 = 41,
Ip6am = 12,
Raw = {raw, Ip6, Ip6am, <<Mad/binary, Ifindex/binary>>},
gen_udp:open(5000, [inet6, binary, Raw]).
```

The issue has been around for a while, and was described here: https://stackoverflow.com/questions/38376640/cant-add-multicast-group . The workaround is complicated because it involves passing OS-specific socket flags:

* Linux: `IPV6_ADD_MEMBERSHIP 20` https://github.com/torvalds/linux/blob/master/include/uapi/linux/in6.h#L172
* Mac: `IPV6_JOIN_GROUP 12` https://github.com/apple/darwin-xnu/blob/main/bsd/netinet6/in6.h#L568

bmk:
If you are looking for an IPv6 version of multicast_if, it is currently not supported.
Even though the documentation claims it is.
Also, the IPv6 version of multicast_if takes an interface index (integer), not an (IPv6) address.
Also, the add/drop membership option has the same problem. It is documented to
work with IPv6, but does not.

bmk:
A solution for this problem has now been merged (into maint and master)

njouanin:
Hi,
I'm running Erlang/OTP 25.2 and it seems that `badarg` error still appears with IPv6. 
Here's my test:

```
[...]

PR #6061, opened 2022-06-08
bmk/kernel/20220518/fix multicast/otp 18091

bmk:
Attempt to fix IPv6 handling for multicast options.


bmk:
I forgot to mention that the branch is based on another topic branch, which deals with inet-driver debug.

erts/preloaded/src/prim_inet.erl
```diff
@@ -2078,21 +2097,58 @@ enc_value_tuple(_, _, _, _) -> [].
...
+       is_integer(Idx) ->
+    [?int32(?INET_AF_INET), ?int32(Idx), ip4_to_bytes(IP1), ip4_any()];
+enc_value_2(membership, {IP, Idx})
+  when (tuple_size(IP) =:= 8) andalso
+       is_integer(Idx) ->
+    %% The reason for turning this thing around (the interface
+    %% before the address) is because of the inet-driver (it reads out a
+    %% 32-bit value for *all* options, so we might as well put a 32-nit
+    %% value "first".
+    [?int32(?INET_AF_INET6), ?int32(Idx), ip6_to_bytes(IP)];
```
binaryseed, line 2150: does this still get encoded into a socket option binary in the same order as using the `raw` option in the workaround described here: https://github.com/erlang/otp/issues/5789#issue-1168721524 ?

bmk: Sorry for the delay!

Are you referring to: '<<Mad/binary, Ifindex/binary>>'?
Since we use the "proper" type, 'struct ipv6_mreq', I am
not sure why you think the order matter.
Or am I missing something?

binaryseed: Oh, didn't realize that. Disregard!

bmk:
The "internal" version of this branch has now been merged.

PR #6063, opened 2022-06-08
observer: correct the units shown for memory data

gonzalobf:
The current code does the conversion from bytes to kibibyte (KiB), mebibyte(MiB) and gibibyte(GiB), that is dividing by 1024.

However, when the units are printed, the units are in kilobytes (KB), megabytes (MB) and gigabytes (GB), so the bytes should be divided by 1000 instead of 1024.

We realised about the error when we were discussing in [erlang forum](https://erlangforums.com/t/ets-memory-occupy-the-erl-shell-query-occupies-more-than-2m-memory-but-the-observer-query-occupies-more-than-18m-memory/1525/5), and I couldn't explain why it was a difference between calculating the ets memory footprint in the shell and looking in `observer` (I must say that I wasn't very awake so I didn't think it could be in KiB).

The change is not a big deal but I think it will show more precise values.

### DIFF
--- a/lib/observer/src/observer_lib.erl
+++ b/lib/observer/src/observer_lib.erl
@@ -299,13 +299,13 @@ to_str({Unit, X}) when (Unit==bytes orelse Unit==time_ms) andalso is_list(X) ->
     catch error:badarg -> X
     end;
 to_str({bytes, B}) ->
-    KB = B div 1024,
-    MB = KB div 1024,
-    GB = MB div 1024,
+    KiB = B div 1024,
+    MiB = KiB div 1024,
+    GiB = MiB div 1024,
     if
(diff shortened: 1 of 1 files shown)


gonzalobf:
@dgud Sound good, it makes sense to have the same value as `top`. I changed my MR so it only changes the units. 

Thank you

dgud:
Can you rebase onto maint branch for inclusion in OTP-25.1?

gonzalobf:
@dgud I rebased my branch but the Windows test is failing with `Error: The process 'C:\hostedtoolcache\windows\Ubuntu\18.4.0\x64\ubuntu1804.exe' failed with exit code 1`, I suspect that is not related to my change.

PR #6080, opened 2022-06-13
Fix SEGV while crash dumping ETS ordered_set

### LINKED ISSUE #5981: Segmentation fault while writing crash dump
**Describe the bug**
BEAM dumps the core in this line: https://github.com/erlang/otp/blob/master/erts/emulator/beam/erl_process.h#L2897=

This line assumes that `esdp->current_process` is not NULL, which isn't true all the time. 

The trivial fix (check that it's not NULL) would be on a hot path. Given that the only way to reproduce this is via writing crash dump, I'd be interested in a more efficient fix.

** To reproduce **

It happens when BEAM is writing crash dump and dumping an `ordered_set` CATree table.

```
#0  0x00000000007878d0 in erts_sched_local_random (additional_seed=140670306335632) at beam/erl_process.h:2899
#1  do_random_join_with_low_probability (seed=140670306335632, tb=0x7ff05ba3b858) at beam/erl_db_catree.c:757
#2  runlock_base_node (tb=0x7ff05ba3b858, base_node=<optimized out>) at beam/erl_db_catree.c:841
#3  unlock_iter_base_node (iter=0x7ff098b8d390) at beam/erl_db_catree.c:928
#4  catree_find_nextprev_root (iter=iter@entry=0x7ff098b8d390, search_keyp=search_keyp@entry=0x0, forward=1) at beam/erl_db_catree.c:1755
#5  0x000000000078834e in catree_find_next_root (keyp=0x0, iter=0x7ff098b8d390) at beam/erl_db_catree.c:2205
#6  db_print_catree (to=0x832660 <erts_write_fp>, to_arg=0x7ff0901ab600, show=0, tbl=0x7ff05ba3b858) at beam/erl_db_catree.c:2205
#7  0x00000000006276d6 in print_table (to=to@entry=0x832660 <erts_write_fp>, to_arg=to_arg@entry=0x7ff0901ab600, show=show@entry=0, tb=tb@entry=0x7ff05ba3b858) at beam/erl_db.c:5291
#8  0x00000000006396e7 in db_info_print (vpdbip=<synthetic pointer>, tb=0x7ff05ba3b858) at beam/erl_db.c:5319
#9  erts_db_foreach_table (func=<optimized out>, alive_only=1, arg=<synthetic pointer>) at beam/erl_db.c:5356
#10 db_info (to=to@entry=0x832660 <erts_write_fp>, to_arg=to_arg@entry=0x7ff0901ab600, show=show@entry=0) at beam/erl_db.c:5330
#11 0x000000000067d36a in erl_crash_dump_v (file=file@entry=0x0, line=line@entry=0, fmt=fmt@entry=0x8b2dde "erl_child_setup closed\n", args=args@entry=0x7ff098b8e6c0) at beam/break.c:1026
#12 0x00000000005372e1 in erts_exit_vv (n=-3, flush_async=flush_async@entry=0, fmt=fmt@entry=0x8b2dde "erl_child_setup closed\n", args1=args1@entry=0x7ff098b8e6c0, args2=args2@entry=0x7ff098b8e6d8) at beam/erl_init.c:2594
#13 0x0000000000537410 in erts_exit (n=n@entry=-3, fmt=fmt@entry=0x8b2dde "erl_child_setup closed\n") at beam/erl_init.c:2621
#14 0x00000000007a00d3 in forker_ready_input (e=<optimized out>, fd=<optimized out>) at sys/unix/sys_drivers.c:1768
#15 0x00000000005e8875 in erts_port_task_execute (runq=runq@entry=0x7ff09b03db40, curr_port_pp=curr_port_pp@entry=0x7ff09b03eee0) at beam/erl_port_task.c:1764
#16 0x0000000000467bc4 in erts_schedule (esdp=<optimized out>, p=<optimized out>, calls=<optimized out>) at beam/erl_process.c:9788
```

sverker:
Fixes #5981.

Crash dump of ETS table with options `ordered_set` and `{write_concurrency,true}` could lead the segmentation violation crash if the crash dumping happens to not be done on behalf of an Erlang process running thread (current_process==NULL).

### DIFF
--- a/erts/emulator/beam/erl_db_catree.c
+++ b/erts/emulator/beam/erl_db_catree.c
@@ -1196,7 +1196,7 @@ static void join_catree(DbTableCATree *tb,
     DbTableCATreeNode *neighbor_parent;
 
     ASSERT(thiz->is_base_node);
-    if (parent == NULL) {
+    if (parent == NULL || ERTS_IS_CRASH_DUMPING) {
         BASE_NODE_STAT_SET(thiz, 0);
         wunlock_base_node(thiz);
         return;
@@ -1342,7 +1342,7 @@ static void split_catree(DbTableCATree *tb,
     DbTableCATreeNode* ERTS_RESTRICT new_right;
     DbTableCATreeNode* ERTS_RESTRICT new_route;
 
-    if (less_than_two_elements(base->u.base.root)) {
+    if (less_than_two_elements(base->u.base.root) || ERTS_IS_CRASH_DUMPING) {
         if (!(tb->common.status & DB_CATREE_FORCE_SPLIT))
             BASE_NODE_STAT_SET(base, 0);
         wunlock_base_node(base);
--- a/erts/emulator/beam/erl_process.h
+++ b/erts/emulator/beam/erl_process.h
@@ -673,6 +673,7 @@ struct ErtsSchedulerData_ {
     } pending_signal;
 
     Uint64 reductions;
+    Uint64 rand_state;
     ErtsSchedWallTime sched_wall_time;
     ErtsGCInfo gc_info;
     ErtsPortTaskHandle nosuspend_port_task_handle;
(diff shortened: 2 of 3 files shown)


erts/emulator/beam/erl_db_catree.c
```diff
@@ -1196,7 +1196,7 @@ static void join_catree(DbTableCATree *tb,
     DbTableCATreeNode *neighbor_parent;
 
     ASSERT(thiz->is_base_node);
-    if (parent == NULL) {
+    if (parent == NULL || ERTS_IS_CRASH_DUMPING) {
```
max-au, line 1199: As I originally stated in #5981, it's possible to add such a condition. But it appears to be on a hot path, and it may affect performance to a degree.

erts/emulator/beam/erl_process.h
```diff
@@ -673,6 +673,7 @@ struct ErtsSchedulerData_ {
     } pending_signal;
 
     Uint64 reductions;
+    Uint64 rand_state;
```
max-au, line 676: This would be super-amazing to have for @RaimoNiskanen RNG improvements. It would be possible to reuse this per-scheduler state for mwc59.
Right now the problem with mwc59 is lack of initial state (seed needs to be generated and kept on the process heap). But if we have this per-scheduler random state already in, we can do `mwc59` with no extra state needed.

erts/emulator/beam/erl_db_catree.c
sverker: The random number generation is hot patch. But the split/join operations are not particularly hot as they only happen when a threshold of lock contention has been reached (split) or when it decides to randomly let "gravity" slowly shrink the tree (join). The split/join ops are quite heavy themself; seize locks, allocate/free memory, tree rotation.

PR #6081, opened 2022-06-13
fix +hmax command line argument to accept > 4 Gb

max-au:
It is possible to set the maximum heap size for the process
using process_flag(max_heap_size, #{size => (1 bsl 59) -1}).
However command line parameter +hmax would not work with anything
beyond 4 Gb (536870911 words) due to signed int32 used to parse
the command line.


garazdawi:
Thanks! It fails to build on 32-bit, I suggest using MAX_SMALL to compare with. That way we get the same limit for process_flag/2 and this option.

max-au:
Thanks! I thought that 32-bit build are history, but apparently I'm wrong.

erts/emulator/test/process_SUITE.erl
```diff
@@ -2294,6 +2296,15 @@ process_flag_heap_size(Config) when is_list(Config) ->
     VHSize = erlang:process_flag(min_bin_vheap_size, OldVHmin),
     ok.
 
+%% test that max_heap_size is correctly handled when passed via command line
+command_line_max_heap_size(Config) when is_list(Config) ->
+    %% test maximum heap size
+    HMax = 1 bsl 59 - 1,
```
garazdawi, line 2302:
```suggestion
    HMax = case erlang:system_info(wordsize) of 8 -> 1 bsl 59 - 1; 4 -> 1 bsl 27 - 1 end,
```

erts/emulator/beam/erl_init.c
```diff
@@ -162,7 +162,7 @@ int erts_initialized = 0;
 
 int H_MIN_SIZE;			/* The minimum heap grain */
 int BIN_VH_MIN_SIZE;		/* The minimum binary virtual*/
-int H_MAX_SIZE;			/* The maximum heap size */
+Uint H_MAX_SIZE;		/* The maximum heap size */
```
garazdawi, line 165:
```suggestion
Sint H_MAX_SIZE;		/* The maximum heap size */
```

erts/emulator/beam/erl_vm.h
```diff
@@ -254,7 +254,7 @@ extern Uint erts_instr_count[];
 
 extern int H_MIN_SIZE;		/* minimum (heap + stack) */
 extern int BIN_VH_MIN_SIZE;	/* minimum virtual (bin) heap */
-extern int H_MAX_SIZE;          /* maximum (heap + stack) */
+extern Uint H_MAX_SIZE;          /* maximum (heap + stack) */
```
garazdawi, line 257:
```suggestion
extern Sint H_MAX_SIZE;          /* maximum (heap + stack) */
```

erts/emulator/beam/erl_init.c
max-au: This makes H_MAX_SIZE signed, and max_heap_size is `Uint` everywhere else. I'll keep the Uint, and use Sint only for initial conversion.

garazdawi:
Thanks!

PR #6084, opened 2022-06-14
Add transaction_log_dir option to Mnesia

gustawlippa:
Hi!
We've noticed `mnesia overloaded` warnings in one system, and wanted to address this somehow. Since the warnings are happening when the transaction log files are to be dumped before the previous dumping has finished, the issue seems to be with disc performance. Changing `dump_log_write_treshold​` or `dc_dump_limit​` did not help, from what I understand. So the proposed solution was to put the transaction log files on a different filesystem, possibly on a different, faster drive then the Mnesia directory. This would require adding a configuration parameter with a path specifying where to put the log files, something like proposed in this PR - `transaction_log_dir`.

I've tried to write testcases for this option, and from what I checked locally, the proposal seems to work. I did not write the documentation yet, because it maybe better to get some feedback before that.

### DIFF
--- a/lib/mnesia/src/mnesia.erl
+++ b/lib/mnesia/src/mnesia.erl
@@ -2551,6 +2551,7 @@ system_info2(transaction_commits) -> mnesia_lib:read_counter(trans_commits);
 system_info2(transaction_restarts) -> mnesia_lib:read_counter(trans_restarts);
 system_info2(transaction_log_writes) -> mnesia_dumper:get_log_writes();
 system_info2(core_dir) ->  mnesia_monitor:get_env(core_dir);
+system_info2(transaction_log_dir) -> mnesia_monitor:get_env(transaction_log_dir);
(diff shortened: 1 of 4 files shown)


dgud:
Did using this patch remove the warnings?

Normally if you have peak of work you will get warning but as long as it doesn't happen several times in a row
mnesia will catch up and it is not a problem, the problem is when you are consistently writing more data then
what the disk/mnesia can handle so the logfile will increase the size.

Otherwise the code looks ok.

gustawlippa:
Being honest, we didn't yet test this solution. I will keep you updated and tell you about the results when we will. The reason I've opened this PR before that, was to see if adding such an option to Mnesia would be at all acceptable.

As for the load, it's rather consistent in our case, and we see a lot of these warnings for longer periods.

dgud:
Closing this, please re-open if/when it have been proved to increase the performance

PR #6087, opened 2022-06-15
Property-based tests for the lists module

Maria-12648430:
This PR provides property-based tests for the `lists` module. As the test suite is comparatively huge (92 properties, plus some custom generators, plus helpers), we expect a review to take a long time.

As the `lists` module is built around a native data type, tests that do not repeat the implementation or are at least close to it are often hard to come up with. The methods we employ where appropriate are the following:

* Build the expected outcome iteratively along with the input list, then compare the output of the tested function against the expected outcome. This is generally the approach used in properties that use the `gen_list_fold` and `gen_list_deepfold` generators, for example `prop_enumerate_1`, `prop_flatlength`, `prop_sum` etc.
* Run the tested function, then check the result to fulfill certain expectations. Properties employing this approach make use of special helper functions, named `check_*`, for example `prop_sort_1`, `prop_umerge_1` etc.
* The input list(s) are constructed in such a way that the output to be expected can be predicted. By and large, this is done by generating two lists and a special element and fusing the three together, and let the generator return the fused construct as well as the special item. For example, `prop_delete`, `prop_member`, `prop_nth` etc.

Exceptions are `prop_append/2` and `prop_subtract`, which indeed do repeat the current implementation. However, this is on purpose, to reflect the statements made in the documentation that say "lists:append(A, B) is equivalent to A ++ B" and "lists:subtract(A, B) is equivalent to A -- B", respectively.


lib/stdlib/test/property_test/lists_prop.erl
```diff
@@ -0,0 +1,1764 @@
...
+        false ->
+            find_remove_in_heads(Found, Fn, E, Ls, [L|Seen])
+    end.
+
+%% --------------------------------------------------------------------
+check_uniqed(L, UL) ->
+    check_uniqed(fun(X) -> X end, L, UL).
+
+check_uniqed(Fn, L, UL) ->
+    check_uniqed1(Fn, L, UL, sets:new()).
```
bjorng, line 1745:
```suggestion
    check_uniqed1(Fn, L, UL, sets:new([{version, 2}])).
```

juhlig: We were thinking of that, but our impression was that `sets:new/1` was supposed to be a temporary addition to the `sets` module for the time while there are several versions to choose from, which would disappear again afterwards?

bjorng: We have not decided yet when we will make version 2 default. There are potential backward compatibility problems if some application has saved a set to a disk file. Not sure if that is a problem in practice.

I suggest that you use `sets:new/1` for now. When we'll change the default, we will change all `sets:new/1` calls in the Erlang/OTP source code to `sets:new/0`.

juhlig: Ok, done :)

Btw, the docs for `sets:new/0` should probably say what version you will get from it. Right now, that can only be guessed from the paragraph at the top of the module description.

Maria-12648430:
> @Maria-12648430 forgot to mention...

@juhlig Yeah, sue me already :stuck_out_tongue_closed_eyes:

> As the default `atom` generator produces atoms randomly and without bounds, it quickly exhausts the limits of the atom table.

If we are adding more property tests to OTP (and I expect we will), I think it will become necessary to move the custom generators into a separate utility module that they all can (should/must) use, and revisit the existing property tests also. Otherwise they will eat up the atom supply rather sooner than later.

That's material for another PR I guess.

garazdawi:
If you want you can rebase to latest master and get the property tests running in github (working since I merged #6093).

Maria-12648430:
> If you want you can rebase to latest master and get the property tests running in github (working since I merged #6093).

Neat :heart_eyes:

bjorng (commented):
Overall looks good to me. I only have a few minor comments.

Maria-12648430:
@bjorng we did most of the requested changes already and will probably finish and push them tomorrow. Too tired now :yawning_face:

lib/stdlib/test/property_test/lists_prop.erl
```diff
@@ -0,0 +1,1819 @@
...
+    ExpElems =:= FoundElems;
+check_usorted(SortFun, [E], ExpElems, FoundElems) ->
+    maps:is_key(E, ExpElems) andalso
+    check_usorted(SortFun, [], ExpElems, maps:update_with(E, fun(Cnt) -> Cnt + 1 end, 1, FoundElems));
+check_usorted(SortFun, [E1|[E2|_]=L], ExpElems, FoundElems) ->
+    SortFun(E1, E2) andalso
+    maps:is_key(E1, ExpElems) andalso
+    check_usorted(SortFun, L, ExpElems, maps:update_with(E1, fun(Cnt) -> Cnt + 1 end, 1, FoundElems));
+check_usorted(_SortFun, _L, _ExpElems, _FoundElems) ->
+    false.
```
bjorng, line 1800:
```suggestion
    check_sorted(SortFun, Sorted, ExpElems, #{}).
```

lib/stdlib/test/property_test/lists_prop.erl
```diff
@@ -0,0 +1,1819 @@
...
+
+ucount_elems(SortFun, L) ->
+    ucount_elems(SortFun, L, #{}).
+
+ucount_elems(SortFun, [E|Es], Acc) ->
+    K = ufind_key(SortFun, E, maps:keys(Acc)),
+    ucount_elems(SortFun, Es, maps:put(K, 1, Acc));
+ucount_elems(SortFun, [], Acc) ->
```
bjorng, line 1808:
```suggestion
ucount_elems(_SortFun, [], Acc) ->
```

bjorng (commented):
Looks good to me! I only have two more minor suggestions. When you have fixed them, please squash your commits. I hope to merge this pull request tomorrow.

Maria-12648430:
@bjorng there, suggestions applied, squashed, ready :smiley:

bjorng:
Thanks!

PR #6094, opened 2022-06-17
Added `zip:zip_get_crc32/2` to read the CRC32 sums from an open zip handle

dominicletz:


dominicletz:
I've added to the docs xml `since="OTP 25.x"` not sure what the right value here is for master commits

lib/stdlib/doc/src/zip.xml
```diff
@@ -506,6 +506,14 @@
       </desc>
     </func>
 
+    <func>
+      <name name="zip_get_crc32" arity="2" since="OTP 25.x"/>
```
bjorng, line 510: For this PR to appear in OTP 25.1, you'll need to rebase it on the `maint` branch and edit the PR to change its target to `maint`.

To get the correct version number, do the following change and our scripts will ensure that the correct version is inserted at the time of the release:

```suggestion
      <name name="zip_get_crc32" arity="2" since="OTP @OTP-18159"/>
```

dominicletz: Cool, I've merged that. There is no need to be in 25.1, can be a later version as well. Is @OTP-18159 a magic number that will always be replaced?

bjorng: `@OTP-18159@` is a magic token that will only work for this pull request.

I now see that I forgot the `@` at the end. That is, the line should be:

```
      <name name="zip_get_crc32" arity="2" since="OTP @OTP-18159@"/>
```

> There is no need to be in 25.1, can be a later version as well.

The next later version for the `master` branch will be OTP 26.0 in about  a years time.

dominicletz:
I've added the @ at the end. That's it for me. The timing for the merge is for me secondary as we're working in the meantime on our own branch here. Let me know your preference I can rebase on `maint` as well or keep it here. 

Cheers!

bjorng:
`master` is fine for me.

Please squash the fixup commit into the first commit (or squash all three into one; your choice).

dominicletz:
done. thanks!

bjorng:
Thanks for your pull request.

PR #6095, opened 2022-06-17
Cons/features/tweaks/otp 18199

lisztspace:
Fix a number of smaller issues, add more precision, improve documentation and move tests to common test.


erts/test/erlc_SUITE.erl
```diff
@@ -435,6 +487,613 @@ make_dep_options(Config) ->
...
+features_macros(Config) when is_list(Config) ->
+    {Run, _, OutDir} = run_fun_dirs(Config),
+
+    Run("f_macros.erl", "", ?OK([])),
+
+    {ok, Peer0, Node0} =
+        ?CT_PEER(#{args => ["-pa", OutDir],
+                   connection => 0}),
+    Call = fun(Node, Fun) ->
+                   rpc:call(Node, f_macros, Fun, [])
```
bjorng, line 795: In new code, we should always use `erpc` instead of `rpc`.

lib/stdlib/doc/src/erl_features.xml
```diff
@@ -63,7 +63,20 @@
...
+        <c>rejected</c>) and features that are no longer configurable
+        (status <c>permanent</c>).</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="configurable" arity="0" since="OTP 25.0"/>
+      <fsummary>Return all configurable features.</fsummary>
+      <desc>
+        <p>Return a list of all configurable features, i.e., features
```
bjorng, line 77:
```suggestion
        <p>Return a list of all configurable features, that is, features
```

(Our documentation guidelines says that we should not use abbreviations such a "i.e." or "e.g.".)

lib/stdlib/src/erl_features.erl
```diff
@@ -187,20 +200,15 @@ keyword_fun(Opts, KeywordFun) ->
                (_) -> false
             end,
     FeatureOps = lists:filter(IsFtr, Opts),
-    {AddFeatures, DelFeatures} = collect_features(FeatureOps),
+    {AddFeatures, DelFeatures, RawFtrs} = collect_features(FeatureOps),
     %% FIXME check that all features are known at this stage so we
     %% don't miss out on reporting any unknown features.
```
bjorng, line 205: Is this FIXME still relevant? If so, could you fix it in this pull request?

bjorng (changes requested):
Overall looks good to me.

lib/stdlib/src/erl_features.erl
lisztspace: No, this has been fixed in this pull request, so I'll just remove the `FIXME`

erts/test/erlc_SUITE.erl
lisztspace: Ok, will fix.

bjorng (approved):

PR #6100, opened 2022-06-20
compiler: Add type info for bs_init_writable

frej:
The SSA instruction bs_init_writable is not known to the SSA type
analysis pass, nor is its result type known to the BEAM
validator. This patch adds type information for bs_init_writable and
as it allows the SSA type pass to elide is_bitstring tests, it also
extends the BEAM validator to consider the result of bs_init_writable
to be a bitstring.

### DIFF
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -2148,6 +2148,8 @@ type(update_tuple, [Src | Updates], _Anno, Ts, _Ds) ->
 type(wait_timeout, [#b_literal{val=infinity}], _Anno, _Ts, _Ds) ->
     %% Waits forever, never reaching the 'after' block.
     beam_types:make_atom(false);
+type(bs_init_writable, [_Size], _, _, _) ->
+    beam_types:make_type_from_value(<<>>);
 type(_, _, _, _, _) ->
     any.
 
--- a/lib/compiler/src/beam_validator.erl
+++ b/lib/compiler/src/beam_validator.erl
@@ -3279,6 +3279,9 @@ join_tuple_elements(I, Tuple, Type0) ->
 call_types({extfunc,M,F,A}, A, Vst) ->
     Args = get_call_args(A, Vst),
     beam_call_types:types(M, F, Args);
+call_types(bs_init_writable, A, Vst) ->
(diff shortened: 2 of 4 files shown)


lib/compiler/src/beam_ssa_type.erl
```diff
@@ -2148,6 +2148,8 @@ type(update_tuple, [Src | Updates], _Anno, Ts, _Ds) ->
 type(wait_timeout, [#b_literal{val=infinity}], _Anno, _Ts, _Ds) ->
     %% Waits forever, never reaching the 'after' block.
     beam_types:make_atom(false);
+type(bs_init_writable, [_Size], _, _, _) ->
+    #t_bitstring{};
```
bjorng, line 2152: The type will be better if you set the type in the same way as for an empty binary:

```suggestion
    beam_types:make_type_from_value(<<>>);
```

With this change, the compiler will eliminate the `is_binary/1` test in the following example:

```
foo(L) ->
    Bs = << <<C:8>> || C <- L >>,
    true = is_binary(Bs),                       %Assertion.
    Bs.
```

lib/compiler/src/beam_validator.erl
```diff
@@ -3279,6 +3279,8 @@ join_tuple_elements(I, Tuple, Type0) ->
 call_types({extfunc,M,F,A}, A, Vst) ->
     Args = get_call_args(A, Vst),
     beam_call_types:types(M, F, Args);
+call_types(bs_init_writable, A, Vst) ->
+    {#t_bitstring{}, get_call_args(A, Vst), false};
```
bjorng, line 3283:
```suggestion
    T = beam_types:make_type_from_value(<<>>),
    {T, get_call_args(A, Vst), false};
```

bjorng (changes requested):
Looks good, but the type for `bs_init_writable` can be made even better as suggested in my comments.

lib/compiler/src/beam_validator.erl
frej: done

frej:
Suggestions implemented and the regression test is updated to check for both `is_binary` and `is_bitstring`.

bjorng:

bjorng (approved):

PR #6101, opened 2022-06-20
Update windows configure cache

garazdawi:


dgud:
win32.config.cache?

garazdawi:
> win32.config.cache?

Working on it... takes a while to run configure without the cache...

PR #6102, opened 2022-06-20
Handle AST nodes without location on *_opt_info

### LINKED ISSUE #6097: beam_ssm_bsm and beam_ssm_recv fail on generated code
Both `beam_ssm_bsm` and `beam_ssm_recv` assume that there is a file+line information attached to the AST node, which is not true for Erlang nodes with annotation equal to `{location, 0}`. Therefore they error while producing a report on such code.

For example, they both fallback to `[]` as the location:

https://github.com/erlang/otp/blob/a37a8ea81ceeee452daa933b2bbeca6c138a20b8/lib/compiler/src/beam_ssa_bsm.erl#L935

https://github.com/erlang/otp/blob/a37a8ea81ceeee452daa933b2bbeca6c138a20b8/lib/compiler/src/beam_ssa_recv.erl#L837

But then match on `{File, Line}`:

https://github.com/erlang/otp/blob/a37a8ea81ceeee452daa933b2bbeca6c138a20b8/lib/compiler/src/beam_ssa_bsm.erl#L1013

https://github.com/erlang/otp/blob/a37a8ea81ceeee452daa933b2bbeca6c138a20b8/lib/compiler/src/beam_ssa_recv.erl#L908

Which leads to:

```
exception error: no match of right hand side value []
  in function  beam_ssa_bsm:make_warning/3 (src/beam_ssa_bsm.erl, line 1015)
  in call from beam_ssa_bsm:add_unopt_binary_info/5 (src/beam_ssa_bsm.erl, line 1006)
  in call from lists:foldl/3 (lists.erl, line 1350)
  in call from lists:foldl_1/3 (lists.erl, line 1355)
  in call from beam_ssa:fold_instrs_1/4 (beam_ssa.erl, line 833)
  in call from lists:foldl_1/3 (lists.erl, line 1355)
  in call from beam_ssa_bsm:module/2 (src/beam_ssa_bsm.erl, line 98)
  in call from compile:'-select_passes/2-anonymous-0-'/3 (compile.erl, line 678)
```

I can submit a pull request but it is unclear if we want to use a fake file/line OR ignore such nodes. Suggestions are welcome!

## Other info

This happens on OTP 24/25.

Originally reported here: https://github.com/elixir-lang/elixir/issues/11914

josevalim:
Closes #6097.

### DIFF
--- a/lib/compiler/src/beam_ssa_bsm.erl
+++ b/lib/compiler/src/beam_ssa_bsm.erl
@@ -1012,7 +1012,11 @@ make_promotion_warning(I, Nested, Anno, Where) ->
     make_warning({binary_created, I, Nested}, Anno, Where).
 
 make_warning(Term, Anno, Where) ->
-    {File, Line} = maps:get(location, Anno, Where),
+    {File, Line} =
+        case maps:get(location, Anno, Where) of
+            {_, _} = Location -> Location;
+            _ -> {"no_file", none}
+        end,
     {File,[{Line,?MODULE,Term}]}.
 
 format_opt_info(context_reused) ->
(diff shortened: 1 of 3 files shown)


bjorng:

PR #6103, opened 2022-06-21
Call `mach_clock_getres` only on initialized clocks

evanmiller:
Mac OS X 10.6 and earlier produce a segmentation fault due to a condition mismatch between calls to `host_get_clock_service` and `mach_clock_getres`. This change brings consistency to the conditionals and fixes the segfault.


rickard-green:
Thanks for the PR! Please sign the [CLA](https://github.com/erlang/otp/pull/6103#issuecomment-1161746214).

Issue #6106, opened 2022-06-24
Allow disabling certificate_authorities extension with peer-to-peer setup with TLS 1.3


jjcarstens:
**Is your feature request related to a problem? Please describe.**

Since OTP >=24.3, `certificate_authorities` extension is configurable for the client and defaults to `false`. However, in a peer-to-peer setup (i.e. with `verify_peer`), the server may still try to use the `certificate_authorities` extension and fails to handshake in some situations, such as when you might manually manage a separate store of client signer CA certificates and are unable to include all CA's with all requests.

**Describe the solution you'd like**

The ability to set `{certificate_authorities, false}` with using `verify_peer` for peer-to-peer connections to disable the extension for both client and server

**Additional context**

This was originally investigated in https://github.com/nerves-project/nerves_system_rpi4/issues/160#issuecomment-1165615614 and more context is available there.
You can also see @tonnenpinguin [bare minimum project](https://github.com/tonnenpinguin/otp25_ssl_error) to reproduce the error with this setup

/cc @fhunleth

IngelaAndin:
@jjcarstens Would the solution in #6204 be sufficient to solve your use case? Or do you have uses cases where some but not all CA (except for ROOT Cas) would be excluded?

jjcarstens:
Thanks @IngelaAndin ! I think this issue is technically different than #6105. I tested your #6204 solution and I technically can get it to work from the client (with the minor change from my PR comment), but it seems like it is still forcing the `certificate_authorities` check on the server when its not necessarily wanted (since the majority of the case we we looking up pinned certificates and dont need to validate the signer CA coming in and can just ignore it)

That said, I'm not too incredibly verse with SSL practices and there may be reasoning I'm missing and this new implementation could be totally acceptable

IngelaAndin:
I agree that this is different from #6105, I will now merge that solution and I think rather an option to not use the certificate_authorities extension is what would be appropriate for your use case. It would be fairly easy to add such an option, however, we are currently reworking the option handling so we would prefer to wait for that before introducing new options, but if you are in a hurry you are always welcome to make a suggestion in a PR.

PR #6107, opened 2022-06-24
Fix certificate_authorities misspelling

jjcarstens:
Total nit found while investigating https://github.com/erlang/otp/issues/6106 and figured I'd be at least a little helpful 😄


IngelaAndin:
Thank you for the PR, merge was delayed due to the vacation period.

PR #6108, opened 2022-06-25
Allow the timezone part of the string to contain only an hour

dischoen:
RFC3339 declares the minutes part of the timezone diff as optional,
which is now enabled with this commit.
E.g.: 2022-06-25 11:04:00+02 is now possible.

### DIFF
--- a/lib/stdlib/src/calendar.erl
+++ b/lib/stdlib/src/calendar.erl
@@ -718,9 +718,14 @@ offset_string_adjustment(_Time, _Unit, "Z") ->
 offset_string_adjustment(_Time, _Unit, "z") ->
     0;
 offset_string_adjustment(_Time, _Unit, Tz) ->
-    [Sign, H1, H2, $:, M1, M2] = Tz,
+    [Sign, H1, H2 | MinutesDiff] = Tz,
     Hour = list_to_integer([H1, H2]),
-    Min = list_to_integer([M1, M2]),
+    Min = case MinutesDiff of
+              [$:, M1, M2] ->
+                  list_to_integer([M1, M2]);
+              [] ->
+                  0
+          end,
     Adjustment = 3600 * Hour + 60 * Min,
     case Sign of
         $- -> -Adjustment;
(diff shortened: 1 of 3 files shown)


lib/stdlib/src/calendar.erl
```diff
@@ -717,7 +717,15 @@ offset_string_adjustment(_Time, _Unit, "Z") ->
...
-offset_string_adjustment(_Time, _Unit, Tz) ->
+offset_string_adjustment(_Time, _Unit, Tz) when length(Tz) == 3 ->
+    [Sign, H1, H2] = Tz,
+    Hour = list_to_integer([H1, H2]),
+    Adjustment = 3600 * Hour,
+    case Sign of
+        $- -> -Adjustment;
+        $+ -> Adjustment
+    end;
+offset_string_adjustment(_Time, _Unit, Tz) when length(Tz) == 6 ->
```
bjorng, line 728:
```suggestion
offset_string_adjustment(_Time, _Unit, Tz) ->
    [Sign, H1, H2 | MinutesDiff] = Tz,
    Hour = list_to_integer([H1, H2]),
    Min = case MinutesDiff of
              [$:, M1, M2] ->
                  list_to_integer([M1, M2]);
              [] ->
                  0
          end,
    Adjustment = 3600 * Hour + 60 * Min,
    case Sign of
        $- -> -Adjustment;
        $+ -> Adjustment
    end.
```

(and also remove the rest of the function after your changes)

bjorng (changes requested):
Thanks for your pull request.

I've change the target of the pull request to maint so that it can be included in OTP 25.1.

I have suggested one simplification of the code in a comment. Also, I would want you to add a note the documentation that starting from OTP 25.1, the minutes part of the timezone diff is now optional.

dischoen:
@bjorng I have accepted your suggestion (hope I did it right, I don't have much experience with pull requests on github).
I will do the documentation change tonight.

bjorng (approved):

bjorng:
Thanks for your pull request.

PR #6110, opened 2022-06-27
inets: Add a helper function for client ssl connections

dgud:
httpc:ssl_verify_host_options/1 returns ssl client options
to verify the host.

Added as a helper function to 25.1 to be added as the default option
in 26.

### DIFF
--- a/lib/inets/doc/src/httpc.xml
+++ b/lib/inets/doc/src/httpc.xml
@@ -647,7 +647,23 @@
         <marker id="get_options"></marker>
       </desc>
     </func>
-     
+
+    <func>
+      <name since="OTP @OTP-18118@">ssl_verify_host_options(WildcardHostName) -> list() </name>
+      <fsummary>Returns ssl options for host verification.</fsummary>
+      <type>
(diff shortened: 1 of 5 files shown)


lib/inets/doc/src/httpc.xml
```diff
@@ -647,7 +647,20 @@
...
+
+    <func>
+      <name since="OTP @OTP-18118@">ssl_verify_host_options(WildcardHostName) -> list() </name>
+      <fsummary>Returns ssl options for host verification.</fsummary>
+      <type>
+        <v>WildcardHostName = boolean()</v>
+      </type>
+      <desc>
+        <p>Returns ssl options which can be used to verify the host, with or without
+        wildcard matching of the hostname.
```
u3s, line 659: > `wildcard matching of the hostname`

While reading, I'm not sure about purpose of this wildcard. I'm guessing its a pattern matching with something containing hostname ... but maybe could be explained little more so guessing need could be reduced.
Maybe reference to `public_key:pkix_verify_hostname_match_fun(https)` manual would be good enough?

Issue #6070, opened 2022-06-09
Incorrect .app bundle location on macOS


wojtekmach:
**Describe the bug**
When packaging up a Erlang in an .app bundle on macOS, the system runtime incorrectly reports the bundle location and cannot retrieve any of the bundle keys.

Here is at least one place in OTP that could benefit from improvements in this area: https://github.com/erlang/otp/blob/OTP-25.0.1/lib/wx/c_src/wxe_ps_init.c#L38:L41

It might be a problem in OTP but it is also very possible it's just a matter of proper packaging.

**To Reproduce**

1. Apply this patch to get additional debugging information

```diff
diff --git a/lib/wx/c_src/wxe_ps_init.c b/lib/wx/c_src/wxe_ps_init.c
index 473354a671..56a0d13ec9 100644
--- a/lib/wx/c_src/wxe_ps_init.c
+++ b/lib/wx/c_src/wxe_ps_init.c
@@ -58,6 +58,8 @@ void * wxe_ps_init2() {

    // Setup and enable gui
    pool = [[NSAutoreleasePool alloc] init];
+   NSLog(@"bundleURL=%@", [[NSBundle mainBundle] bundleURL]);
+   NSLog(@"bundleIdentifier=%@", [[NSBundle mainBundle] bundleIdentifier]);

    if( !is_packaged_app() ) {
       // Undocumented function (but no documented way of doing this exists)
```

2. Run this script 

```bash
#!/bin/sh
set -e pipefail
export MAKEFLAGS=-j8
export ERL_TOP=$PWD

./otp_build setup -a
./otp_build boot -a
./otp_build release -a

rm -rf MyApp.app
mkdir -p MyApp.app/Contents/Resources
cp -R release/`$ERL_TOP/make/autoconf/config.guess` MyApp.app/Contents/Resources/erlang
(cd MyApp.app/Contents/Resources/erlang && ./Install -sasl $PWD)

cat << EOF > MyApp.app/Contents/Info.plist
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
  <key>CFBundleName</key>
  <string>MyApp</string>
  <key>CFBundleVersion</key>
  <string>1.0.0</string>
  <key>CFBundleIdentifier</key>
  <string>myapp</string>
</dict>
</plist>
EOF

[...]

PR #6112, opened 2022-06-28
erts: Remove `erts/etc/darwin/Info.plist`

wojtekmach:
Ref: https://github.com/erlang/otp/issues/6070#issuecomment-1168676066


wojtekmach:
Also, please see the related PR #5393 that was already merged.

If accepted, I'm not sure if this should be against master or it is ok to be against maint.

cc @dominicletz

dominicletz:
@wojtekmach if this is accepted it will not be possible with the default Erlang build will to use the WebView against localhost addresses. Which means electron like use cases, but even development can only be done with custom builds.

For that reason I recommend against emerging this.

josevalim:
@dominicletz this is tricky because, by having this file included, then it becomes impossible for us to customize our own builds. So I would prefer to make custom builds required in order to access some features rather than not allowing configuring custom builds at all.

dominicletz:
How does the default file impact a custom build? When you do a custom build you can choose any other plist file to be used instead - unless I'm missing something.

josevalim:
@dominicletz Mac actually looks inside the running program first. If the running program has a plist, then it doesn't fallback to the one in the app bundle.

dominicletz:
For a custom build you can control which plist file will be embedded and just use yours instead of the default. I'm not sure if you can change the embedded plist after compilation again? - Would be nice if there is an easier way than a full recompile.

josevalim:
That's exactly the issue though. AFAIK it requires full compilation without the .plist. :(

wojtekmach:
@dominicletz I put together a sample app that does two things:

1. starts a web server at port 8000, rendering some html and loading an image at `/logo.png`
2. starts a `wxWebView` pointed to `http://localhost:8000`

Here's the code: https://github.com/wojtekmach/demo

I built the app using the OTP compiled on this branch, without the Info.plist. As far as I can tell it works well, no problems with loading things from the localhost.

To simulate distributing the app, I copied the app to a blank slate Mac VM and it worked too.

If you'd like to try it out, here's the zipped app: [demo.zip](https://github.com/erlang/otp/files/9111505/demo.zip)

If I'm not triggering the use case behind the `NSAppTransportSecurity` setting, please let me know and I'll happily do more tests.

Btw, this is how I've confirmed that the Info.plist is gone:

```
# the app
~/demo[main]% otool -P _build/dev/demo.app/Contents/Resources/rel/erts-13.0.2/bin/beam.smp
_build/dev/demo.app/Contents/Resources/rel/erts-13.0.2/bin/beam.smp:

# my OTP checkout
~/src/otp[wm-remove-Info.plist]% otool -P release/x86_64-apple-darwin20.6.0/erts-13.0.2/bin/beam.smp
release/x86_64-apple-darwin20.6.0/erts-13.0.2/bin/beam.smp:

# OTP checkout at master, at another machine:
~/src/otp[master]% otool -P release/aarch64-apple-darwin21.5.0/erts-13.0.2/bin/beam.smp
release/aarch64-apple-darwin21.5.0/erts-13.0.2/bin/beam.smp:
(__TEXT,__info_plist) section
<plist version="1.0">
<dict>
	<key>NSAppTransportSecurity</key>
	<dict>
		<key>NSExceptionDomains</key>
		<dict>
			<key>localhost</key>
			<dict>
				<key>NSExceptionAllowsInsecureHTTPLoads</key>
				<true/>
				<key>NSIncludesSubdomains</key>
				<true/>
			</dict>
		</dict>
	</dict>
</dict>
</plist>
```

wojtekmach:
@dominicletz sorry for a ping. If you could take a look at the comment above and help us out we’d really appreciate. Put it another way, if there is a sample app that requires this plist and that I could run myself, I’d be happy to dig into that. Thanks!

dominicletz:
Hey @wojtekmach, no worries on the ping - This is an interesting experiment with bundling a plist file after compilation. Is there any way to run the development environment this way so that the WebView works in the development environment (e.g. while editing code) as well?

wojtekmach:
@dominicletz sorry but just so we're on the same page what I'm saying is this setting isn't needed. I'm running OTP without the Info.plist and it works well. I have tested that demo project above, `erl> wx:demo().`, and https://github.com/elixir-desktop/desktop-example-app just now. Everything works well.

If you can help me reproduce a situation where that setting _is_ needed, I'm happy to keep digging.

wojtekmach:
I forgot to mention, fwiw my setup is the following:

macOS Monterey 12.4
OTP with my patch off maint (just updated; see https://github.com/wojtekmach/otp/commits/wm-remove-Info.plist)
wxWidgets 3.1.7 (installed with brew)

Regarding running development environment and some sort of custom `MyApp.app/Contents/Info.plist` I think it's tricky to make it ergonomic but not impossible to do. I think what we'd need is some sort of `$ mix app` task that would build the project, build the `.app`, and `$ open /path/to/my.app`. So we'd have potentially two VMs running, one that started the Mix task (or equivalent) and one from the .app, and probably communication between the two so they exit together, etc. On one hand it is definitely adding decent amount of complexity but on the other hand something like this is required to test app events like "open url", "open file", etc, anyway. I'm keen on exploring this down the road, I think it will be very useful.

wojtekmach:
I changed base branch to maint.

dominicletz:
In terms of reproducing @wojtekmach the effect of the plist there are no doubts it's working as documented by Apple. I'm attaching a screen recording with two runs. One with the current build including the plist and one with the plist entry invalidated. You can see that without the `localhost` plist entry the WebView only renders a white page in the example. I'm running the default brew-based build of OTP 24.3.2 which comes with the built-in plist. For the second run, I invalidate the plist entry by binary replacing the word `localhost` with `localpost` so that it doesn't match anymore. 

https://user-images.githubusercontent.com/2987674/181038052-f6af9a27-3d53-45b3-b477-52da5d2c044b.mov

 This is precisely the reason why I'm opposed to merging this change, it will cause users to only see a white page -- there is not even an error message.

wojtekmach:
Excellent, thank you, I will try to reproduce this behaviour on my machine. Again, in my OTP build per this branch everything worked correctly as in the beginning of the video, the app window was filled in with web content, hence I am pushing for this change. But I clearly need to understand more pieces so thank you again for providing a way to hopefully reproduce the results.

dominicletz:
After some offline discussion with @wojtekmach we found that this behaviour changed in macos - When I contributed the plist file I was still using a pre macOS 10.12 - In addition, it seems the macOS versions only using the new default if no plist instructions `NSAppTransportSecurity` are present. This is explaining the behavior in the video above. From the docs:
 
https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsallowslocalnetworking 

> In iOS 10 and macOS 10.12 and later, ATS allows all three of these connections by default, so you no longer need an exception for any of them. However, if you need to maintain compatibility with older versions of the OS, set both of the [NSAllowsArbitraryLoads](https://developer.apple.com/documentation/bundleresources/information_property_list/nsapptransportsecurity/nsallowsarbitraryloads) and NSAllowsLocalNetworking keys to YES.

I've tested new builds now without the plist (as per this PR) and those seem to work fine. All this said I'm in favor of merging this PR. Thanks, @wojtekmach for being this persistent and for helping get to the root of this.

wojtekmach:
Thank you @dominicletz!!!

frazze-jobb:
Sorry for waiting so long, but this seems fine I will merge it

PR #6113, opened 2022-06-28
wx: Add `WX_MACOS_NON_GUI_APP`

wojtekmach:
We cannot depend on `is_packaged_app()` check because it is not a perfect way of checking if app is bundled.

It would return true when the app is bundled with a launcher like this:

```swift
import AppKit

let task = Process()
task.launchPath = "/Users/wojtek/src/otp/bin/erl"
task.arguments = ["-noshell", "-eval", "wx:new(), timer:sleep(5000), halt()"]
try task.run()
task.waitUntilExit()
```

It would return false for this launcher, however:

```swift
import AppKit

class AppDelegate: NSObject, NSApplicationDelegate {
    func applicationDidFinishLaunching(_ aNotification: Notification) {
        let task = Process()
        task.launchPath = "/Users/wojtek/src/otp/bin/erl"
        task.arguments = ["-noshell", "-eval", "wx:new(), timer:sleep(5000), halt()"]
        try! task.run()
    }
}

let app = NSApplication.shared
let delegate = AppDelegate()
app.delegate = delegate
app.run()
```

Btw, there has recently been a fix in wx [1] and we'd only be able to take advantage of it when `OSXIsGUIApplication` returns true (and when wx recognizes our app as properly bundled, see [2]). And so I believe to support variety of scenarios, we need precise control.

[1] https://github.com/wxWidgets/wxWidgets/pull/22508
[2] https://github.com/erlang/otp/issues/6070#issuecomment-1168676066

### DIFF
--- a/lib/wx/c_src/wxe_impl.cpp
+++ b/lib/wx/c_src/wxe_impl.cpp
@@ -225,14 +225,16 @@ void WxeApp::MacReopenApp() {
   send_msg("reopen_app", &empty);
 }
 
+// See: https://github.com/wxWidgets/wxWidgets/blob/v3.1.5/src/osx/cocoa/utils.mm#L76:L93
 bool WxeApp::OSXIsGUIApplication() {
-  // wx manually activates the application if it's not in the bundle [1]. In particular, it calls
-  // `[NSApp setActivationPolicy: NSApplicationActivationPolicyRegular]` which prevents the app
-  // from running in the background and we cannot control it with `LSUIElement` [2]. Their check
-  // if it's a bundle always returns false for wxErlang. Thus we use our own way of detecting it.
-  // [1] https://github.com/wxWidgets/wxWidgets/blob/v3.1.5/src/osx/cocoa/utils.mm#L76:L93
-  // [2] https://developer.apple.com/documentation/bundleresources/information_property_list/lsuielement
-  return !is_packaged_app();
+   char val_buf[8];
+   size_t val_len = 7;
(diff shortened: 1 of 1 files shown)


wojtekmach:
This is PR is really only useful for the 2nd approach I mentioned in the PR description, the AppDelegate one, but there are some caveats with that which I have learned since so let's hold off on this change for a moment.

wojtekmach:
Yeah, as noted in the PR description there is a variety of scenarios for using wx so I believe more precise control is really useful here. We'd appreciate if we could land this PR so we ([Livebook Desktop](https://news.livebook.dev/introducing-the-livebook-desktop-app-4C8dpu)) could stop using our fork. :)

dgud:
Can this be done without an env variable, i.e. in some cleaner way?

Or is this such a special case that I should just accept it as it is?

wojtekmach:
I unfortunately couldn't think of a better way of solving this issue, I think we need to override `OSXIsGUIApplication` but at the same time control what exactly it returns (so we can opt-out of certain behaviour in wx).

Would controlling this with `application:put_env` instead be an improvement?

wojtekmach:
@dgud fyi, for Livebook we are also exploring an altogether different way of building Mac apps with OTP described in https://github.com/erlang/otp/issues/6070#issuecomment-1261512487 and for that we wouldn't need this patch. I still think the patch is useful though as there are so many different ways of running OTP and wx and so I think it is very helpful to have some flexibility.

PR #6118, opened 2022-06-30
peer: Add detached and post_process_args

garazdawi:
In order to test interactive shells, peer nodes sometimes need to be started without detaching the shell. This PR adds two mechanism that help when testing such systems:

1. The `detached` option can be set to false which causes the peer when running using distributed or tcp connection to not close its terminal.
2. The `post_process_args` option can be set to a fun which allows the peer to process all arguments before the peer node is started, this notably includes any arguments that `peer` itself has added.

These two changes in combination allows a peer node to be started in a screen/tmux session, or using run_erl/to_erl.

This PR also contains some changes to how cover is started when peer is running in CT so that cover is started earlier in the boot sequence allowing for better code coverage reports.

### DIFF
--- a/lib/stdlib/test/peer_SUITE.erl
+++ b/lib/stdlib/test/peer_SUITE.erl
@@ -45,7 +45,8 @@
     duplicate_name/0, duplicate_name/1,
     old_release/0, old_release/1,
     ssh/0, ssh/1,
-    docker/0, docker/1
+    docker/0, docker/1,
+    post_process_args/0, post_process_args/1
 ]).
 
 suite() ->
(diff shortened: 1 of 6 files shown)


lib/stdlib/test/peer_SUITE.erl
```diff
@@ -563,3 +581,26 @@ docker(Config) when is_list(Config) ->
...
+    RunErl = os:find_executable("run_erl"),
+    [throw({skip, "Could not find run_erl"}) || RunErl =:= false],
+    Erl = string:split(ct:get_progname()," ",all),
+    RunErlDir = filename:join(proplists:get_value(priv_dir, Config),?FUNCTION_NAME),
+    filelib:ensure_path(RunErlDir),
+    {ok, Peer, _Node} =
+        peer:start(
+          #{ name => ?CT_PEER_NAME(),
+             exec => {RunErl, Erl},
+             detached => false,
```
max-au, line 598: Could you add a test to veify that `detached => false` works with both TCP and `standard_io` alternative connections? Technically I do that by adding test case name to `alternative()` function so it gets added to two test groups.

If it is not intended to, it should be disallowed in `verify_args/1` function with relevant user-friendly error.

garazdawi: Added tests (and functionality) for tcp connection type. Also expanded verify_args to disallow standard_io.

max-au:
Looks good in general, but I think test failures are related.

PR #6121, opened 2022-07-01
Documentation about debugging NIFs & drivers

sverker:


system/doc/tutorial/debugging.xml
```diff
@@ -0,0 +1,345 @@
...
+	  and data structures.
+	</p>
+      </item>
+      <item>
+	<p>
+	  <em>Generate a core dump that is easier to analyze</em>. Compiler
+	  optimizations are turned off, which can otherwise make local variables
+	  to be "optmized away" and therefor much harder or even impossible to
+	  see.
+	</p>
```
frej, line 73: I understand what you mean, but the word order is strange (and with some typos), what about:

Compiler optimizations are turned off, which stops the compiler from "optimizing away" variables, thus making it easier to inspect the state of the NIF.

sverker: Thanks.

garazdawi (approved):

PR #6127, opened 2022-07-04
inets: Make https requests verify host by default

dgud:
Make ssl connections more secure by default.

### DIFF
--- a/lib/inets/src/http_client/httpc.erl
+++ b/lib/inets/src/http_client/httpc.erl
@@ -116,11 +116,11 @@ request(Url, Profile) ->
 %%                   {ssl,  ssl_options()} | 
 %%                   {essl, ssl_options()}
 %%      ssl_options() = [ssl_option()]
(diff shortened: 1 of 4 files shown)


max-au:
@voltone - it's finally happening, one of the first items we had under "what EEF Security WG should attempt". Thanks @dgud !

dgud:
Any opinions of we should use wildcard matching on the hostnames or not?
Currently it used, so slightly more insecure but used almost on all hosts.

lib/inets/src/http_client/httpc.erl
```diff
@@ -765,12 +765,14 @@ http_options_default() ->
 		error
 	end,
 
+    Ssl = httpc:ssl_verify_host_options(true),
+
```
IngelaAndin, line 769: Why is this an external call? It could be` Ssl = ssl_verify_host_options(true),` ?!

IngelaAndin (approved):

PR #6130, opened 2022-07-05
compiler: Add missing bs_create_bin prim_op() to type definition

frej:
The missing bs_create_bin operation has probably remained undiscovered
as the Dialyzer limit mentioned in the comment preceding the prim_op()
type definition collapses the effective type to atom().


bjorng:

PR #6138, opened 2022-07-07
Docs: Move maps:groups_from_list docs to correct place

gustawlippa:
When reading the documentation for the new functions, I've noticed that they are rendered at the bottom of the page, not following the alphabetical order: https://www.erlang.org/doc/man/maps.html#groups_from_list-2.
Notice that the new functions are below the description of `maps:without/2`, not following the correct order from the left pane: 
image

This change simply moves the whole description to a correct place


garazdawi:
Thanks! I rebased your branch to maint so that this will be fixed in the next minor release.

PR #6143, opened 2022-07-08
Add dbg:tracer(file, Filename)

garazdawi:
This tracer allows the user to easily trace to a file in cleartext.


gomoripeti:
will this handy addition be included in 25.x as well?

garazdawi:
I was not planning on it, but if you want it then you are welcome to open a PR towards maint :)

PR #6157, opened 2022-07-15
eunit: exact execution option

u3s:
- option for avoiding automatic execution of a module with "_tests" suffix

### DIFF
--- a/lib/eunit/src/eunit_data.erl
+++ b/lib/eunit/src/eunit_data.erl
@@ -31,10 +31,8 @@
 
 -include_lib("kernel/include/file.hrl").
 
--export([iter_init/2, iter_next/1, iter_prev/1, iter_id/1,
-	 enter_context/3, get_module_tests/1]).
-
--import(lists, [foldr/3]).
+-export([iter_init/3, iter_next/1, iter_prev/1, iter_id/1,
+	 enter_context/3, get_module_tests/2]).
 
 -define(TICKS_PER_SECOND, 1000).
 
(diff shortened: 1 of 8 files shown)


u3s:
- @vkatsuba alternative approach compared to #5461
- intention is to leave legacy execution sequence unchanged
- but automated execution of "_tests" module can be disabled if not expected (such as when m and m_tests modules are placed in src folder and execution of all modules from folder is started, I observed this with rebar3 today :-) so I decided it is high time to come back to this one :-)) )

lib/eunit/src/eunit_data.erl
```diff
@@ -131,8 +132,8 @@ iter_id(#iter{pos = N, parent = Ns}) ->
 
 %% @spec (testIterator()) -> none | {testItem(), testIterator()}
 
-iter_next(I = #iter{next = []}) ->
-    case next(I#iter.tests) of
+iter_next(I = #iter{next = [], options = Options}) ->
```
KennethL, line 135: How will this work? `iter_next/1` is exported in the beginning of the file but after this change there is no `iter_next/1` anymore

IngelaAndin: It looks to me that it still only has one argument, the map!

u3s: It's a new field inside map, arity of function is unchanged I think.

lib/eunit/src/eunit_data.erl
```diff
@@ -573,22 +574,24 @@ get_module_tests(M) ->
 	    throw({module_not_found, M})
     end.
 
-get_module_tests_1(M, Es) ->
+get_module_tests_1(M, Es, Options) ->
```
IngelaAndin, line 577: So I know this is not part of your change, just an observation that the very short variable names make the code hard to read I think. A few more characters that would make it easier to get the correct association without analyzing the context could be beneficial. It could perhaps be an enhancement if you think it is worth the time.

u3s: I agree with observation. Unfortunately this is a style eunit code is written :-/ There is limited number of tests of application as well.

I will see what can be improved for eunit_data module - assuming some time box for activity.

u3s: I've pushed a [commit](https://github.com/erlang/otp/pull/6157/commits/dd366cb65accf1b257f73fcbb1b1cfd572db68de) please have a look and let me know what you think
IMHO, result is a better clarity of the code but it is costly to implement and then review.

IngelaAndin (approved):

PR #6167, opened 2022-07-24
otp: fix compilation - deprecated OpenSSL APIs

neheb:
Signed-off-by: Rosen Penev <rosenp@gmail.com>

Compile with -DOPENSSL_API_COMPAT=0x10100000L to simulate.

### DIFF
--- a/lib/crypto/c_src/engine.c
+++ b/lib/crypto/c_src/engine.c
@@ -513,7 +513,11 @@ ERL_NIF_TERM engine_load_dynamic_nif(ErlNifEnv* env, int argc, const ERL_NIF_TER
 #ifdef HAS_ENGINE_SUPPORT
     ASSERT(argc == 0);
 
+# if OPENSSL_VERSION_NUMBER < PACKED_OPENSSL_VERSION_PLAIN(1,1,0)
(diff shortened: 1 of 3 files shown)


lib/crypto/c_src/engine.c
```diff
@@ -513,7 +513,7 @@ ERL_NIF_TERM engine_load_dynamic_nif(ErlNifEnv* env, int argc, const ERL_NIF_TER
 #ifdef HAS_ENGINE_SUPPORT
     ASSERT(argc == 0);
 
-    ENGINE_load_dynamic();
+    ENGINE_load_builtin_engines();
```
HansN, line 516: Although the ENGINE_load_builtin_engines() enables the dynamic engine, it does more.
The substitution
```suggestion
# if OPENSSL_VERSION_NUMBER < PACKED_OPENSSL_VERSION_PLAIN(1,1,0)
    ENGINE_load_dynamic();
# else
    OPENSSL_init_crypto(OPENSSL_INIT_ENGINE_DYNAMIC, NULL);
# endif
```
seems better.

neheb: Fixed.

PR #6170, opened 2022-07-26
compiler: Add missing operations to beam_ssa:prim_op() type

frej:
This patch adds BEAM SSA operations that are produced by the compiler,
but are not in the beam_ssa:prim_op() type.


bjorng:

PR #6171, opened 2022-07-26
Eliminate compiler crash in beam_clean

### LINKED ISSUE #6139: Compiler is not leveraging literal map tuples as frequently as possible
**Describe the bug**

Take the following code:

```erlang
-module(foo).
-compile(export_all).

decimal(Int) ->
  #{type => decimal, int => Int, exp => 0}.
```

I would expect the following property to be true:

```erlang
A = foo:decimal(0),
B = foo:decimal(1),
erts_debug:same(erts_internal:map_to_tuple_keys(A), erts_internal:map_to_tuple_keys(B)).
```

However, that's not the case! This means that every invocation of the function is allocating a new tuple keys, not fully leveraging #1498.

I found the optimization only happens when we are creating a new map where all values are not literals (or all values are literals)! So this will optimize:

```erlang
-module(foo).
-compile(export_all).

decimal(Int) ->
  #{type => Int, int => Int, exp => Int}.
```

Having maps mixed with literal and non-literal keys is likely the most common scenario and therefore it would make more sense if the optimization applied for all maps with _literal keys_. This has the benefit of considerably reducing the memory allocation of programs that use maps.

**Affected versions**
Probably all versions with maps.

bjorng:
Fixes #6139

### DIFF
--- a/lib/compiler/src/beam_call_types.erl
+++ b/lib/compiler/src/beam_call_types.erl
@@ -29,9 +29,14 @@
 %%
 %% Returns whether a call will succeed or not.
 %%
-%% Note that it only answers 'yes' for functions in the 'erlang' module as
+%% Notes:
+%%
+%% This function only answers 'yes' for functions in the 'erlang' module as
 %% calls to other modules may fail due to not being loaded, even if we consider
 %% the module to be known.
+%%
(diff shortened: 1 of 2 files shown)


bjorng:
Wrong issue number in branch name.

PR #6177, opened 2022-07-27
Optimize comparison of equal terms

bjorng:
When doing comparisons, be sure to avoid calling the quite expensive
generic comparison routine for equal tagged values.  In particular,
the generic comparison function for atoms, `erts_cmp_atoms()`, will
always call `memcmp()` for equal atoms longer than 4 characters, so we
want to avoid that.


michalmuskala:
Should this be always done by the `CMP_TERM` macro? I believe the `EQ` macro already does something similar

bjorng:
Yes, I did consider modifying `CMP_TERM()` to always do the equal test, but then I looked at how it was used. It is only used for maps. In most of the places it is used, the keys being equal is impossible or highly unlikely. So in the end I only added it to the places there were were a reasonable probability that the keys would be equal.

PR #1498, opened 2017-07-03
Introduce new_map_lit operation in the loader

michalmuskala:
Optimise construction of maps when all keys are literals but building a literal tuple in the loader. This benefits especially small maps since they basically consist of two tuples - one for the keys and one for the values.

This change should benefit especially Elixir, because of the record replacement - structs are always built with all literal keys. 

The change effectively means significantly reducing memory use (almost by a half) of programs using maps to replace records.

PR #6178, opened 2022-07-27
Enhance creation of maps with literal keys

### LINKED ISSUE #6139: Compiler is not leveraging literal map tuples as frequently as possible

bjorng:
An optimization in the loader was added in: https://github.com/erlang/otp/pull/1498

When a small map is created where all keys are literals, the key tuple
is created as a literal. That means that the key tuple can be shared
for all instances of the map created by the same code.

However, that does not always work, for example for code such
as:

    decimal(Int) ->
        #{type => decimal, int => Int, exp => 0}.

The compiler would try to be smart and rewrite the code to:

    decimal(Int) ->
        (#{type => decimal, exp => 0})#{int => Int}.

Expressed in that way, the loader optimization cannot be applied and
a new key tuple will be created every time `decimal/1` is called.

This pull request removes that problematic transformation. The implementation
is based on José Valim's suggested implementation.

Fixes #6139

### DIFF
--- a/lib/compiler/src/beam_ssa_opt.erl
+++ b/lib/compiler/src/beam_ssa_opt.erl
@@ -1025,6 +1025,30 @@ cse_is([#b_set{op={succeeded,_},dst=Bool,args=[Src]}=I0|Is], Es, Sub0, Acc) ->
             Sub = Sub0#{Bool=>#b_literal{val=true}},
             cse_is(Is, Es, Sub, Acc)
     end;
+cse_is([#b_set{op=put_map,dst=Dst,args=[_Kind,Map|_]}=I0|Is],
+       Es0, Sub0, Acc) ->
+    I1 = sub(I0, Sub0),
+    {ok,ExprKey} = cse_expr(I1),
+    case Es0 of
+        #{ExprKey:=PrevPutMap} ->
+            Sub = Sub0#{Dst=>PrevPutMap},
+            cse_is(Is, Es0, Sub, Acc);
+        #{Map:=PutMap} ->
+            case combine_put_maps(PutMap, I1) of
+                none ->
+                    Es1 = Es0#{ExprKey=>Dst},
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you for the PR @bjorng!

Do you think it is worth adding the code in the issue description as a test?

```erlang
decimal(Int) ->
  #{type => decimal, int => Int, exp => 0}.

key_space_sharing_test(_Config) ->
  A = ?MODULE:decimal(0),
  B = ?MODULE:decimal(1),
  true = erts_debug:same(erts_internal:map_to_tuple_keys(A), erts_internal:map_to_tuple_keys(B)).
```

bjorng:
> Do you think it is worth adding the code in the issue description as a test?

Good idea. Will do.

bjorng:
Added a test case.

josevalim (approved):

PR #6179, opened 2022-07-28
[filelib] Fix wildcard function to not follow symlinks

jendis:
Avoid possible DOS when more than 1 symlinks create loop inside
the (current) working directory

### DIFF
--- a/lib/stdlib/src/filelib.erl
+++ b/lib/stdlib/src/filelib.erl
@@ -354,15 +354,25 @@ prepare_base(Base0) ->
     lists:reverse(Base2).
 
 do_double_star(Base, [H|T], Patterns, Result0, Mod, Root) ->
-    Full = case Root of
-               false -> filename:join(Base, H);
-               true -> H
+    {Full, Loop} = case Root of
+               false ->
+                    F = filename:join(Base, H),
+                    L = case file:read_link(F) of
+                            {ok, LinkTarget} ->
+                                lists:prefix(filename:split(LinkTarget), filename:split(Base));
+                            _ -> false
+                        end,
+                    {F, L};
+               true -> {H, false}
            end,
-    Result1 = case do_list_dir(Full, Mod) of
-                  {ok, Files} ->
-                      do_double_star(Full, Files, Patterns, Result0, Mod, false);
-                  _ -> Result0
-              end,
+    Result1 = if Loop -> Result0;
+                true ->
+                    case do_list_dir(Full, Mod) of
+                        {ok, Files} ->
+                            do_double_star(Full, Files, Patterns, Result0, Mod, false);
+                        _ -> Result0
+                    end
+                end,
     Result2 = case Patterns of
                   %% The root is never included in the result.
                   _ when Root -> Result1;
--- a/lib/stdlib/test/filelib_SUITE.erl
+++ b/lib/stdlib/test/filelib_SUITE.erl
@@ -31,7 +31,7 @@
          ensure_path_invalid_path/1,
 	 wildcard_symlink/1, is_file_symlink/1, file_props_symlink/1,
          find_source/1, find_source_subdir/1, find_source_otp/1,
-         safe_relative_path/1, safe_relative_path_links/1]).
+         safe_relative_path/1, safe_relative_path_links/1, wildcard_symlink_loop/1]).
 
 -import(lists, [foreach/2]).
 
@@ -59,7 +59,7 @@ all() ->
      ensure_path_invalid_path,
      wildcard_symlink, is_file_symlink, file_props_symlink,
      find_source, find_source_subdir, find_source_otp,
-     safe_relative_path, safe_relative_path_links].
+     safe_relative_path, safe_relative_path_links, wildcard_symlink_loop].
 
 groups() -> 
     [].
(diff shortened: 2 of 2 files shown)


josevalim:
Unfortunately this is a backwards incompatible change and it also does not match the behaviour of wildcards on Unix (which do traverse symlinks). Perhaps we could store the expanded symlink in an accumulator and stop traversal if we visit a previously visited node?

jendis:
Honestly my first intention to fix it was in a way you are proposing. But as I have seen more troubles than benefits in following symlinks, I came with this easy approach.
One of the problems that can arise is the computational time in case of a big tree of files.
The question is if we can't just make a note in the doc that symlinks are not followed and close eyes regarding to the backward compatibility - I'm really not sure if this can cause a harm somewhere but I faced a big troubles with the current state in production system :).
In the meantime I'll try to bring the solution that detects loops and you can finally decide.

jendis:
I should probably better explain the whole point of this PR.
I'll explain the problem with currently available implementation of wildcard funtion on the example.
Let's have the following directory structure with symbolic links pointing to:

- "/here/is/target/symlink/a -> /here/is/target

- "/here/is/target/symlink/b -> /here/is/target

Then after the function call `filelib:wildcard("/here/is/**")` the function `do_double_star` accumulates combinations of paths "`/here/is/target/symlink/a, /here/is/target/symlink/a/symlink, /here/is/target/symlink/a/symblink/b ...` " where the final number of path combinations is limited by the OS or FS limits. The number of combinations is definitely huge enough (`2^n` where `n > 100` on my Linux where `PATH_MAX=4096`) to exhaust available memory on standard Linux distributions or Windows OS on powerful machines with tens of gigabytes but even unrealistically more.
The testcase `wildcard_symlink_loop/1` https://github.com/erlang/otp/pull/6179/files#diff-82e5eb1f83e28b1235e24a05ae2c2f4d115a19695b67415b8f18eadf50eb7601 can be easily applied on the current codebase so you can quickly raise the issue.

I generally consider to follow symlinks by default as unsafe (and maybe even unexpected) but I realize that your point about backward compatibility is crucial, so the [4c97f9a](https://github.com/erlang/otp/pull/6179/commits/4c97f9ac87f39720bc39bbcc57c3b0e1f271f4d7) is really bad idea.

The change [6c2110e](https://github.com/erlang/otp/pull/6179/commits/6c2110e2ad90dedfea6d4ac41d2f222acb34a955) fulfills your idea about detection of loop. In fact that was my first local fix before I argued to myself that it: 

1. makes the function slower 
2. follow symlink is bad idea

There is still another solution for this problem we can discuss, which is to extend each `filelib:wildcard/1` and `filelib:wildcard/2` for 
a `FollowSymlinks = boolean()` argument. That will cause build time error so users of this functions will be quickly aware of the incompatibility and can fix it according to their needs. I would expect such change in the next major version of course.
The main benefit would be the better performance because there will be no need to check the loops.

But please consider seriously the [6c2110e](https://github.com/erlang/otp/pull/6179/commits/6c2110e2ad90dedfea6d4ac41d2f222acb34a955) for the really next release because there is a real risk of exploitation.

jendis:
As no activity here for a significant amount of time, I'd like to ask if I can do anything to help to solve this issue?

bjorng (changes requested):
Please squash the commit and edit the commit message to better explain why the change is needed and what has been changed.

bjorng:
Could you give some more details how the current implementation of `filelib` could be exploited in a DOS attack?

It seems to me, that for an attack to be possible, you are allowing untrusted users to specify their own wildcards and/or to create symlinks in the file system. That seems to be dangerous in other ways.

jendis:
I've pushed the required changes.

As a chance for possible DoS, imagine e.g. an email service that accepts emails from the online wild and one of its task is filtration of content of attachments (archived data etc.). Using this function to get all files from e.g. already extracted archive can give a chance to attacker if he just prepare such archive with symlink loops and then periodically sends this content. The problem is that process heap is exhausted which means the crash of the whole application.

frej:
> As a chance for possible DoS, imagine e.g. an email service that accepts emails from the online wild and one of its task is filtration of content of attachments (archived data etc.). Using this function to get all files from e.g. already extracted archive ...

Extracting unsanitized filenames from an untrusted archive sounds like a recipe for disaster. In your example, a safer way would be to assigning the filenames yourself and just keep track of the mapping between your filenames and the originals for reporting the results.

lib/stdlib/src/filelib.erl
```diff
@@ -354,15 +354,25 @@ prepare_base(Base0) ->
...
 do_double_star(Base, [H|T], Patterns, Result0, Mod, Root) ->
-    Full = case Root of
-               false -> filename:join(Base, H);
-               true -> H
+    {Full, Loop} = case Root of
+               false ->
+                    F = filename:join(Base, H),
+                    L = case file:read_link(F) of
+                            {ok, LinkTarget} ->
+                                lists:prefix(filename:split(LinkTarget), filename:split(Base));
```
frej, line 362: `file:readlink/1` does not canonicalize the returned path. If the target for example is  `../foo` the prefix check will not trigger. I think the traditional UNIX way of detecting symlink loops (or for that matter hard links) is to do a `stat` and then check if you have seen the device and inode before. That will only require _O(n)_ memory, where _n_ is the number of existing files.

On UNIX-like platforms you could probably use `file:read_file_info/1` but I don't know if you will get meaningful inode+device values on Windows.

jendis: Yes, you are absolutely right about the relative path symlinks. It reminds me another reason why the first proposal here was to not follow symlinks at all.
I don't think hard links can cause any problem.
I'll definitely need more time to come with right solution.

bjorng:
Feel free to reopen this pull request if you come up with a solution.

PR #6180, opened 2022-07-28
add xmerl_xml_indent-module

fnchooft:
The indent module was added to provide out of the box
indented output instead of the standard xmerl_xml-module.

The code is based on the Elixir xmerl_xml_indent-package (https://hex.pm/packages/xmerl_xml_indent)

To me it seems a perfect addition even though it has the same caveats as the Elixir-package or the xmerl_xml-module.
It seems overkill to have to include calls to xslt, or other packages to do some formatting.

I have not found a way in the callback to make it more configurable, if such an option exists please let me know.

PS: I tried to add many more tests from the tgz-files. However even the standard xmerl_xml-module breaks on these since there are many xml-constellations which is does not support, such as unicode, or element-names which are not latin1.

Kind regards,

Fabian

### DIFF
--- a/lib/xmerl/src/xmerl_xml_indent.erl
+++ b/lib/xmerl/src/xmerl_xml_indent.erl
@@ -1,0 +1,86 @@
+%%
+%% %CopyrightBegin%
+%% 
+%% Copyright Ericsson AB 2003-2016. All Rights Reserved.
+%% 
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%% 
+%% %CopyrightEnd%
+%%
+
+%% Description  : Callback module for exporting complete or simple forms to indented XML.
+%%
+%% This module indents the xml with 2 spaces and a newline \n.
+%% Currently the implementation does not allow it to be configured.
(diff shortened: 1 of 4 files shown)


fnchooft:
Please let me know if I can add or aid in any other matter - or if additional information is needed.
Kind regards,
Fabian

lthor:
Hi,
I had a first look and think it's ok and a good feature to add (will test it myself this week, just have to finish up another thing).

But I would like you to put the test case in xmerl_SUITE instead (as a separate formatting test group), I just want the standard tests in xmerl_std_SUITE.

It would also be nice if you have the time/possibility to just add a section about the formatting last in the xmerl_ug.xmlsrc so the feature is visible in the documentation.

BR Lars

lthor:
Hi Fabian
just to clarify the indent feature will be added after I done the above changes but I can't promise when it will be done due to other things. If you help us fix it we can get it in quicker.
BR Lars

fnchooft:
Thanks for the info. I have been swamped but will try to advance with the
cleanup.
Kind regards,

Fabian

lib/xmerl/src/xmerl_xml_indent.erl
```diff
@@ -0,0 +1,86 @@
...
+                ] ++ ident(LengthParents);
+            false ->
+                Data
+        end,
+    markup(Tag, Attrs, NewData).
+
+is_char(Data) when is_list(Data) ->
+    is_list(hd(Data)) and not is_integer(lists:nth(1, hd(Data))).
+
+ident(Level) ->
```
leonardb, line 85: I know this is an old PR and has not seen activity for a while, but surely this is meant to be `indent`, not `ident`?

IngelaAndin:
@fnchooft  ping

fnchooft:
Hi, my apologies for the long delays in responding and fixing the issues.
I will try to make the time available to fix and update the patch as I
believe that it is a simple feature but none the less - very useful.

Kind regards,

Fabian

IngelaAndin:
You need to add the email address that the commits are using as one of the registered emails on github, I believe.

lthor:
This pull request is replaced by #7942.

PR #6188, opened 2022-07-31
[kernel|pg] Add tagged processes feature

paulozulato:
Added possibility to associate a process with a tag when joining pg
groups. This tag can be useful to identify particular processes within a
given group.

For example, suppose some resource is controlled by three processes -
each one controls one aspect of this resource. Let's name these aspects
top, middle and bottom.
These processes are added to a pg group. In this way, it's easy to get
the processes associated to a resource, but it's hard to identify the
process related to a specific aspect among the returned ones.
With tagged processes, these processes could be associated with some tag
in order to be easily identified within that group.

This use case may be executed as follow:
  %%%%
  %% The resource instance 42 is controlled by three processes (local or
  %% remote).
  %% Each process manages one aspect (top, middle and bottom) of that
  %% resource.
  %%%%

  %% Set up group
  Scope = some_scope,
  GroupId = 42,
  Top = new_pid(top, GroupId), % returns pid()
  Middle = new_pid(middle, GroupId), % returns pid()
  Bottom = new_pid(bottom, GroupId), % returns pid()
  Processes = [{top, Top}, {middle, Middle}, {bottom, Bottom}].
  pg:join(Scope, GroupId, Processes).

  %% Getting members
  Members = pg:get_members(Scope, GroupId).
  % something like [{top, Pid1}, {middle, Pid2}, {bottom, Pid3}]

  %% it's easy to get the process which controls the aspect "middle" in
  %% this group
  {middle, MiddlePid} = lists:keyfind(middle, 1, Members).

### DIFF
--- a/lib/kernel/doc/src/pg.xml
+++ b/lib/kernel/doc/src/pg.xml
@@ -55,6 +55,12 @@
       <seemfa marker="#get_local_members/1"><c>get_local_members/1</c></seemfa>
       to determine which processes are members of the group.
       Then the message can be sent to one or more group members.</p>
+    <p>To help identifying particular processes within a group, processes can be
+       associated with a tag. These tagged processes are returned along with
+       untagged processes by <c>pg</c> functions. There are no special functions
+       for handling tagged or untagged processes only.
+    </p>
+
     <p>If a member terminates, it is automatically removed from the group.</p>
 
     <p>A process may join multiple groups. It may join the same group multiple times.
@@ -98,6 +104,14 @@
     <datatype>
       <name name="group"/>
       <desc><p>The identifier of a process group.</p></desc>
+    </datatype>
+    <datatype>
+      <name name="tag"/>
+      <desc><p>The tag associated with a process.</p></desc>
+    </datatype>
+    <datatype>
+      <name name="pid_or_tagged_pid"/>
+      <desc><p>An untagged or tagged process.</p></desc>
     </datatype>
   </datatypes>
 
--- a/lib/kernel/src/pg.erl
+++ b/lib/kernel/src/pg.erl
@@ -53,6 +53,7 @@
 %%  * all leave/join operations are serialised through pg server process
 %%
 -module(pg).
+-behaviour(gen_server).
 
 %% API: default scope
 -export([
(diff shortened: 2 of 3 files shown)


max-au:
Thank you, it is indeed a good alternative to the original flow!
I have been considering something in this vain to add "region-aware process groups" (https://github.com/max-au/spg/issues/7), but realised there are a couple of issues with this:
1. Protocol change is a long process. Just making this change will crash all `pg` running prior OTP versions. Any change in `pg` protocol needs to go through several years long process (with negotiation protocol added to `pg` discovery). It is my fault to miss that in the original implementation. I need to come up with a forward-compatible fix for `discover` message that is currently missing "version" field.
2. In a large deployment with tens of thousands groups (and million published processes) `pg` performance is critical to a degree that we look for any optimisation opportunity possible (e.g. https://github.com/erlang/otp/pull/4615).

I ended up keeping existing approach. Instead of turning pid into more complex structure (by adding metadata/tag to every pid), I "tag" processes by adding them to multiple groups. That is, instead of calling
`pg:join(group, {tag, self()})` I do `pg:join({group, tag}, self())`. It is better from performance standpoint - in most cases `pg:get_members({group, tag})` succeeds returning smaller list of pids (and generating less garbage, especially for large groups). 

One more way to add this feature without changing `pg` protocol could be via scope monitoring. You can, for example, run your own gen_server that subscribes to all updates of the selected scope, and maintains a different view (with extra metadata attached, e.g. tags). I have a similar thing, gen_server subscribed to `pg` updates, to filter our "unwanted" pids (in a gist - when some process joins a group, it's "raw service", in order to use it, my gen_server negotiates which protocol that "raw service" provides, and accepts only those understood by the client).

rickard-green:
The incompatible protocol change is a major hurdle. An OTP node of release X needs to be able to talk to OTP nodes from release X-2 to release X+2. That is, this PR cannot be introduced as is.

Looking into how to change the protocol under the above restrictions, so that this feature in the long run could be introduced could be an option. Another option is @max-au's scope monitoring.

rickard-green:
Closing this since it cannot be introduced without causing compatibility issues with older releases. The PR #6433 introduce a protocol version in order to make it easier to do changes like this, so this PR might be revisited in the future.

PR #6192, opened 2022-08-02
dialyzer: Add warning_options `missing_return` and `extra_return`

vtm9:
The options are documented in [docs](https://www.erlang.org/doc/man/dialyzer.html#using-dialyzer-from-the-command-line) 
But they are not in the `warn_option() type` and not in the `dialyzer -Whelp`.
This commit add the options in dialyzer docs and cli help.

### DIFF
--- a/lib/dialyzer/src/dialyzer.hrl
+++ b/lib/dialyzer/src/dialyzer.hrl
@@ -130,7 +130,9 @@
                        | 'unmatched_returns'
                        | 'overspecs'
                        | 'specdiffs'
+                       | 'extra_return'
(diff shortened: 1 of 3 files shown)


bjorng:
Thanks for your pull request.

The contents looks good, but because it is based on `master` it would be included in OTP 26, which will be released next year.

I was going to ask you to rebase it to `maint` so that we could include it in OTP 25.1, but then I saw that your branch was named `master` and rebasing it to `maint` would be confusing.

Therefore, I cherry-picked your commit to `maint`.

PR #6193, opened 2022-08-02
compiler: Start all potential nif functions with a nif_start

frej:
Since OTP 25 the `nifs` attribute can be used to declare which
functions are nifs. When the `nifs` attribute is present, the runtime
system will ensure that only functions listed in the attribute are
replaced by a native implementation when `erlang:load_nif/2` is
called. Functions listed in the `nifs` attribute are flagged by
emitting a `nif_start` operation as the first instruction in the
function.

Due to backwards compatibility, the `nifs` attribute is not required,
so when a module containing a call to `erlang:load_nif/2` is
encountered, the compiler is forced to conservatively assume that in
such a module, all functions are potential nifs. Currently most of the
more powerful compiler optimizations (beam_ssa_opt.erl) are completely
disabled when a module contains a call to `erlang:load_nif/2`. It has
been suggested by Björn Gustavsson and Sverker Eriksson that switching
to a more fine-grained approach where only functions that are
potential nifs are excluded from, or treated as external functions,
during optimization would be an improvement.

This patch extends the v3_core-pass to make all functions start with a
`nif_start` instruction when the `nifs` attribute isn't present in the
module but a call to `erlang:load_nif/2` is. With this change, later
compiler passes can safely assume that a function not starting with a
`nif_start` instruction is not a nif regardless of the presence of a
call to `erlang:load_nif/2` in the module.

### DIFF
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -160,7 +160,8 @@
 	       opts=[]     :: [compile:option()], %Options.
                dialyzer=false :: boolean(),     %Help dialyzer or not.
 	       ws=[]    :: [warning()],		%Warnings.
-               file=[{file,""}]			%File.
+               file=[{file,""}],                %File.
+               load_nif=false :: boolean()      %true if calls erlang:load_nif/2
 	      }).
 
 %% XXX: The following type declarations do not belong in this module
@@ -171,12 +172,16 @@
 
 -record(imodule, {name = [],
 		  exports = ordsets:new(),
-                  nifs = sets:new([{version, 2}]),
+                  nifs = none ::
+                    'none' | sets:set(), % Is a set if the attribute is
+                                         % present in the module.
 		  attrs = [],
 		  defs = [],
 		  file = [],
 		  opts = [],
-		  ws = []}).
+		  ws = [],
+                  load_nif=false :: boolean() %true if calls erlang:load_nif/2
+                 }).
 
 -spec module([form()], [compile:option()]) ->
         {'ok',cerl:c_module(),[warning()]}.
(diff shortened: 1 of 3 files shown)


lib/compiler/src/v3_core.erl
```diff
@@ -171,12 +172,16 @@
 
 -record(imodule, {name = [],
 		  exports = ordsets:new(),
-                  nifs = sets:new([{version, 2}]),
+                  nifs = false ::
+                    'false' | sets:set(), % a set if the attribute is
+                                          % present in the module.
```
bjorng, line 177: I suggest adding `is` for clarity:

```suggestion
                    'none' | sets:set(), %is a set if the attribute is
                                          %present in the module.
```

Also, in the compiler we prefer to use `none` instead of `false` in this situation.

lib/compiler/src/v3_core.erl
```diff
@@ -849,6 +869,7 @@ expr({named_fun,L,'_',Cs}, St) ->
     fun_tq(Cs, L, St, unnamed);
 expr({named_fun,L,Name,Cs}, St) ->
     fun_tq(Cs, L, St, {named,Name});
+
```
bjorng, line 872:
```suggestion
```

bjorng (commented):
Looks good. I have only a couple of nitpicks.

lib/compiler/src/v3_core.erl
frej: Fixed.

lib/compiler/src/v3_core.erl
frej: Fixed, although it prevents the cute use of field's value as a truth value at line 202 :)

frej:
Unless someone tells me otherwise, I won't consider the failed tests to require any action by me, `write /dev/stdout: no space left on device` during Docker wrangling is probably not related to a change in the compiler.

bjorng:

bjorng:
Thanks for your pull request.

PR #6196, opened 2022-08-03
crypto: Fix AES and Blowfish names for OpenSSL 3.0

### LINKED ISSUE #6146: aes_*_cfb128 wrongly reported as unsupported with OpenSSL 3.0
**Describe the bug**

I ran into this problem after building OTP 24.3.4.2 with OpenSSL 3.0.2.  `aes_128_cfb128` is not supported

    > lists:member(aes_128_cfb128, crypto:supports(ciphers)).
    false

At the same time I can see `aes-128-cfb` in the output of the following command

    $ openssl enc -ciphers

**To Reproduce**

Build OTP with OpenSSL 3.0.

**Expected behavior**

The following expression evaluates to `true`

    lists:member(aes_128_cfb128, crypto:supports(ciphers))

The same for `aes_192_cfb128` and `aes_256_cfb128`.

**Affected versions**

OTP 24, OTP 25.

**Additional context**

According to _openssl-enc_ man page `aes-128-cfb` is 128 bit AES in 128 bit CFB mode.  It seems like crypto uses wrong name in `cipher_type_t.str_v3`.  I can confirm that it works as expected if I replace `-cfb128` with `-cfb` here

<https://github.com/erlang/otp/blob/OTP-24.3.4.2/lib/crypto/c_src/cipher.c#L82-L84>

dmitrivereshchagin:
This should fix #6146. It turned out that Blowfish in CFB/OFB modes is also not detected as supported.

Tested with OpenSSL 3.0.5. Corresponding groups for AES and Blowfish in crypto_SUITE are not skipped and pass.

### DIFF
--- a/lib/crypto/c_src/cipher.c
+++ b/lib/crypto/c_src/cipher.c
@@ -60,15 +60,15 @@ static struct cipher_type_t cipher_types[] =
 #endif
 
 #ifdef HAVE_BF
-    {{"blowfish_cbc"},   "BF-CBC",   {&EVP_bf_cbc},   0, NO_FIPS_CIPHER, NOT_AEAD},
-    {{"blowfish_cfb64"}, "BF-CFB64", {&EVP_bf_cfb64}, 0, NO_FIPS_CIPHER, NOT_AEAD},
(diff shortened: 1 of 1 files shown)


HansN:
Thanks for the PR!

PR #6197, opened 2022-08-04
Regular expression replacement with a function

juhlig:
I rarely use regular expressions with Erlang. But when I do, I often would need a function to process matches and create replacements.

This PR extends `re:replace/3,4` to accept a 2-ary function for the replacement. This function will be called with the entire match as the first and a list of subexpression matches as second argument. The value returned from the function will be inserted into the result.

A problem exists in case of regular expressions like this:
```
".(x)?(y)?"
```
When applied to a string like `"axy"`, the function will be called with arguments `<<"axy">>` and `[<<"x">>, <<"y">>]`.
When applied to `"ax"`, the arguments will be `<<"ax">>` and `[<<"x">>]`, ie the subexpression match list _does not_ contain the optional-ish match related to `(y)`.
When applied to `"ay"`, the arguments will be `<<"ay">>` and `[<<>>, <<"y">>]`, ie the subexpression match list _does_ contain the optional-ish match related to `(x)` as an empty binary.

I'm not sure how to handle this (and that is why it is not documented yet). The implementation for replacements using strings with back references inserts empty binaries if the match index does not exist in the actual match. For creating the subexpression match list for replacement via a function however, something like this is not possible. The easiest way out may be to recommend a catch-all clause that will return an empty binary, in the documentation.

### DIFF
--- a/lib/stdlib/doc/src/re.xml
+++ b/lib/stdlib/doc/src/re.xml
@@ -75,6 +75,9 @@
     <datatype>
       <name name="compile_option"/>
     </datatype>
+    <datatype>
+      <name name="replace_fun"/>
+    </datatype>
   </datatypes>
 
   <funcs>
@@ -363,7 +366,7 @@
         elements with Replacement.</fsummary>
       <desc>
         <p>Replaces the matched part of the <c><anno>Subject</anno></c> string
-          with the contents of <c><anno>Replacement</anno></c>.</p>
+          with <c><anno>Replacement</anno></c>.</p>
         <p>The permissible options are the same as for
           <seemfa marker="#run/3"><c>run/3</c></seemfa>, except that option<c>
           capture</c> is not allowed. Instead a <c>{return,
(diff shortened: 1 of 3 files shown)


bjorng (changes requested):
Looks good to me. I think that we should handle optional capturing patterns as you suggested, namely documenting the behavior and leave it up to the user's fun to handle it. In practice, I wouldn't expect such patterns to be very common, so most replacement funs wouldn't have to handle it.

Please add more details to the commit message, so that if anyone wonders what happened 10 years from now don't have to locate the pull request, but can find the details in the commit message.

juhlig:
Ok, great :) I applied your suggestions and will finish up the docs, squash the commits and write a more elaborate commit message tomorrow.

juhlig:
@bjorng I documented the short subexpression list behavior as best as I could, what do you think? I also augmented the commit message as requested.

bjorng (approved):
Looks good. Added to our daily builds.

juhlig:
There is some strange failure in the doc build, that seems to be unrelated to my changes :man_shrugging:

bjorng:
> There is some strange failure in the doc build, that seems to be unrelated to my changes.

Yes, it is unrelated to your changes.

bjorng:
Thanks for your pull request.

PR #6199, opened 2022-08-04
gen_event: eliminate selective-receive in fetch_msg/6

mikpe:
Perform a non-selective receive with timeout, and pass the
message to factored-out decode_msg/7, similar to gen_server.

One of our more heavily used `gen_event` servers got into a state of extreme memory consumption and slow processing a couple of times this summer. When I looked into it recently I discovered that `gen_event`'s fetch-and-dispatch loop performs a _selective_ receive looking for `system` and `EXIT` messages (without using the ref trick), which could explain the issues. `gen_server` always picks the first message, and I strongly believe `gen_event` should do the same.

This bug seems to have been there "for ever" -- even R6B0 has this difference between `gen_event` and `gen_server`.

### DIFF
--- a/lib/stdlib/src/gen_event.erl
+++ b/lib/stdlib/src/gen_event.erl
@@ -499,19 +499,25 @@ wake_hib(Parent, ServerName, MSL, HibernateAfterTimeout, Debug) ->
 
 fetch_msg(Parent, ServerName, MSL, HibernateAfterTimeout, Debug, Hib) ->
     receive
+	Msg ->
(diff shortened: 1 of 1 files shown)


RaimoNiskanen:
Since the `receive` statement has got a catchall `Msg` your proposed change should be absolutely safe.

This was once probably just a matter of code style and nobody has happened to detect the performance bug...

RaimoNiskanen:
Thank you for your pull request.  Great finding!

PR #6203, opened 2022-08-08
[crypto/4871] Add support for SHAKE128 and SHAKE256.

### LINKED ISSUE #4871: Add support for SHAKE128 and SHAKE256
### Rationale

SHAKE128 and SHAKE256 are part of the SHA-3 hash family (are also based on Keccak-f[1600] permutations), and while they are not as commonly used as SHA3-*, there are notorious projects, such as Tor, which [use them](https://github.com/torproject/torspec/blob/baf7bc9cf7d474e77daa9922471c2b090d90053d/rend-spec-v3.txt#L214) as key derivation functions. OpenSSL 1.1.1 [supports](https://www.openssl.org/docs/man1.1.1/man3/EVP_shake256.html) both.

### Purposed changes

As those functions are XOFs (Extendable Output Functions), supporting them might require some tweaks to existing APIs for providing variable output length. I think the cleanest way would be to pass tuples with algorithm name and size of output needed in bits:

```erlang
crypto:hash({shake256, 32}, Msg)

State = crypto:hash_init({shake128, 256})
State2 = crypto:hash_update(State, Msg)
crypto:hash_final(State2)
```

marcellanz:
This should fix #4871.
Passes tests with /lib/crypto for `openssl@3/3.0.5` and `openssl@1.1/1.1.1q`.


lib/crypto/doc/src/crypto.xml
```diff
@@ -1107,6 +1114,18 @@ end
       </desc>
     </func>
 
+    <func>
+      <name name="hash_xof" arity="3" since="OTP 26.0"/>
```
marcellanz, line 1118: I'm not sure what version to set here for this PR.

lib/crypto/c_src/digest.c
```diff
@@ -114,6 +114,23 @@ static struct digest_type_t digest_types[] =
...
+#else
+    {NULL, NULL}
+#endif
+    },
+
+    {"shake256", "SHAKE-256", 0, 0,
+#ifdef HAVE_SHAKE256
+    {&EVP_shake256, NULL}
+#else
+    {NULL}
```
marcellanz, line 129: this line is superfluous – left from a change while adapting to changes in digest.c
```suggestion
```

lib/crypto/c_src/digest.c
```diff
@@ -114,6 +114,23 @@ static struct digest_type_t digest_types[] =
 #endif
     },
 
+    {"shake128", "SHAKE-128", 0, 0,
+#ifdef HAVE_SHAKE128
+    {&EVP_shake128, NULL}
+#else
+    {NULL, NULL}
```
marcellanz, line 121: align with format used in this file.
```suggestion
    {NULL,NULL}
```

lib/crypto/c_src/digest.c
```diff
@@ -114,6 +114,23 @@ static struct digest_type_t digest_types[] =
...
+#endif
+    },
+
+    {"shake256", "SHAKE-256", 0, 0,
+#ifdef HAVE_SHAKE256
+    {&EVP_shake256, NULL}
+#else
+    {NULL}
+    {NULL, NULL}
```
marcellanz, line 130: align with format used in this file.
```suggestion
    {NULL,NULL}
```

HansN:
Great work! And complete with tests and documentation - very good!

But, for certain reasons it must be compileable also with OpenSSL versions less than 1.1.1.  Less that 0.9.8c is not needed.  It is not necessary to return any other result than that is not supported when running given that the algorithms are not present in crypto:supported.

If the PR is compiled with 1.1.0 or less the function hash_final_xof_nif is not declared. We usually solve this by another declaration  in the \#else part which just only returns an EXCP_NOTSUP

marcellanz:
Thanks. I missed that case and will adapt to have returned `EXCP_NOTSUP` for OPENSSL_VERSION_NUMBER < 1.0.
With returning `EXCP_NOTSUP`, consequently, I would not remove `shake128` and `shake256` from `crypto:supports(hashs)` at all when compiled with OpenSSL < 1.0, right?

HansN:
If an algorithm is present in crypto:supports, it should return a value and not NOTSUP.   That exception means that the algorithm is supported by Erlang/OTP, but not by the current cryptolib.  The algorithm is therefore not in crypto:supports for the current cryptolib.

So yes, keep the `#ifdef HAVE_SHAKExxx` in algorithms.c.

You could then `#undef HAVE_SHAKExxx` in openssl_config.h for versions you do not want to support with old function calls.

HansN:
Thanks for the PR!

marcellanz:
>Thanks for the PR!

You're welcome. Thanks for guiding me through the details.

lorenzosinisi:
Hi and great work on this particular algo! It will be very useful :) May I ask what is needed to include it in the next release and why wasn't it added in the latest one already? Is there any way to help?

lorenzosinisi:
> Thanks @lorenzosinisi. As it introduces addition to library API I assume it would be released with the next major version of OTP?
> 
> I think it would be listed here https://github.com/erlang/otp/milestone/162?closed=1 but its not so far.

Thank you! I'll take a closer look there 👍👍 yes I can imagine how the APIs would change given that shake256 would take also additional args

IngelaAndin:
This PR was merged to master and will hence be part of OTP-26. We are still adopting our ways of working with GitHub features so I guess it never got a milestone attached to it.

PR #6204, opened 2022-08-09
ssl: Enhance client certificate selection

### LINKED ISSUE #6105: TLS handshake issue
**Describe the bug**
Fail to establish TLS connection when version is not specified, but works when version is set to 1.2 at the client side. This issue is found when server is started with `verify_peer` and the request for certificate returns empty list. 
Error on client side:
```
1> ssl:start().
ok
2> ssl:connect("localhost", 9999, [
2>  {certfile, "tls-gen/basic/result/client_certificate.pem"},
2>  {keyfile, "tls-gen/basic/result/client_key.pem"}
2>  ], infinity).
=WARNING REPORT==== 24-Jun-2022::10:24:37.745907 ===

{ok,{sslsocket,{gen_tcp,#Port<0.5>,tls_connection,undefined},
               [<0.115.0>,<0.114.0>]}}
3> =NOTICE REPORT==== 24-Jun-2022::10:24:37.919848 ===
TLS client: In state connection received SERVER ALERT: Fatal - Certificate required
```
Client side error with log_level debug.
```
>>> Handshake, Certificate
[{certificate_request_context,<<>>},{certificate_list,[]}]
writing (30 bytes) TLS 1.2 Record Protocol, application_data
0000 - 17 03 03 00 19 57 1e 75  42 83 a8 79 40 b8 1a bf    .....W.uB..y@...
0010 - 63 af 5e f0 10 c4 cd 2f  6d e2 51 19 3f f7          c.^..../m.Q.?.
>>> Handshake, Finished
[{verify_data,<<133,128,28,173,99,120,66,202,147,101,220,174,217,147,30,69,
                44,180,236,191,254,162,174,41,16,136,150,72,113,139,73,47,
                224,94,213,71,59,81,69,24,45,119,68,15,31,3,195,92>>}]
writing (74 bytes) TLS 1.2 Record Protocol, application_data
0000 - 17 03 03 00 45 81 54 11  7b 5b af 36 c8 e2 85 3a    ....E.T.{[.6...:
0010 - f6 7b b4 5f cf ea 87 cc  79 c6 bb 71 74 51 ea 99    .{._....y..qtQ..
0020 - a1 d0 7b 60 c3 26 fd 0d  ee d7 91 60 5c d6 6f c6    ..{`.&.....`\.o.
0030 - 1b ea 8c 5d 5d 38 df 7f  02 dc 2c 49 84 aa b1 15    ...]]8....,I....
0040 - 49 7f 41 49 62 3c ba 63  72 fd                      I.AIb<.cr.
{ok,{sslsocket,{gen_tcp,#Port<0.5>,tls_connection,undefined},
               [<0.119.0>,<0.118.0>]}}
reading (24 bytes) TLS 1.2 Record Protocol, application_data
0000 - 17 03 03 00 13 f0 00 e4  0a 92 fd b3 5d bf cb e9    ............]...
0010 - cd 75 83 89 8d 83 87 42                             .u.....B
3> =NOTICE REPORT==== 24-Jun-2022::11:44:00.781779 ===
TLS client: In state connection received SERVER ALERT: Fatal - Certificate required

writing (24 bytes) TLS 1.2 Record Protocol, application_data
0000 - 17 03 03 00 13 7b c1 45  25 0b c5 92 5e 06 fb 6a    .....{.E%...^..j
0010 - 3b 81 43 9e e4 c2 d3 04                             ;.C.....
```
Error on server side:
```
Error: {'EXIT',
           {{badmatch,
                {error,
                    {tls_alert,
                        {certificate_required,
                            "TLS server: In state wait_cert at tls_handshake_1_3.erl:1469 generated SERVER ALERT: Fatal - Certificate required\n certificate_required"}}}},
            [{erl_eval,expr,6,[{file,"erl_eval.erl"},{line,496}]},
             {erl_eval,exprs,6,[{file,"erl_eval.erl"},{line,136}]},
             {erl_eval,expr,6,[{file,"erl_eval.erl"},{line,480}]},
             {erl_eval,expr,6,[{file,"erl_eval.erl"},{line,492}]},
             {erl_eval,exprs,6,[{file,"erl_eval.erl"},{line,136}]},
             {shell,exprs,7,[{file,"shell.erl"},{line,691}]},
             {shell,eval_exprs,7,[{file,"shell.erl"},{line,647}]},
             {shell,eval_loop,3,[{file,"shell.erl"},{line,632}]}]}}
Restarting socket
=NOTICE REPORT==== 24-Jun-2022::10:24:37.909543 ===
TLS server: In state wait_cert at tls_handshake_1_3.erl:1469 generated SERVER ALERT: Fatal - Certificate required
 - certificate_required

```

**To Reproduce**
```
# generate certs and setup server.
ssl:start().
F = fun L() ->
    {ok, ListenSocket} = ssl:listen(9999,
        [
            {verify, verify_peer},
            {fail_if_no_peer_cert, true},
       	    {cacertfile, "tls-gen/basic/result/ca_certificate.pem"},
            {certs_keys, [#{certfile => "tls-gen/basic/result/server_certificate.pem", keyfile => "tls-gen/basic/result/server_key.pem"}]},
            {reuseaddr, true}
        ]
    ),
    R = fun L2() ->
        {ok, TLSTransportSocket} = ssl:transport_accept(ListenSocket),
        {ok, Socket} = ssl:handshake(TLSTransportSocket),
        io:format("~p", [{ssl:peercert(Socket), Socket}]),
        L2()
    end,
    Error = catch R(),
    io:format("Error: ~p \nRestarting socket\n", [Error]),
    ssl:close(ListenSocket),
    L()
end.
F().

Or if we use openssl to start the server.

openssl s_server -port 9999 -key tls-gen/basic/result/server_key.pem -cert tls-gen/basic/result/server_certificate.pem -Verify 1 -CAfile tls-gen/basic/result/ca_certificate.pem -debug
```
```
# start client
ssl:start().
ssl:connect("localhost", 9999, [
 {certfile, "tls-gen/basic/result/client_certificate.pem"},
 {keyfile, "tls-gen/basic/result/client_key.pem"}
 ], infinity).
```

**Expected behavior**
I would expect connection setup to be successful.

**Affected versions**
I tried to reproduce the issue on 25.0.2 on a mac, but the issue exists even on linux.
```
% erl
Erlang/OTP 25 [erts-13.0.2] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns] [dtrace]

Eshell V13.0.2  (abort with ^G)
1> crypto:info().
#{compile_type => normal,
  cryptolib_version_compiled => "OpenSSL 1.1.1o  3 May 2022",
  cryptolib_version_linked => "OpenSSL 1.1.1p  21 Jun 2022",
  link_type => dynamic,otp_crypto_version => "5.1.1"}
2> ssl:info().
** exception error: undefined function ssl:info/0
3> ssl:version().
** exception error: undefined function ssl:version/0
4> ssl:versions().
[{ssl_app,"10.8.2"},
 {supported,['tlsv1.3','tlsv1.2']},
 {supported_dtls,['dtlsv1.2']},
 {available,['tlsv1.3','tlsv1.2','tlsv1.1',tlsv1]},
 {available_dtls,['dtlsv1.2',dtlsv1]},
 {implemented,['tlsv1.3','tlsv1.2','tlsv1.1',tlsv1]},
 {implemented_dtls,['dtlsv1.2',dtlsv1]}]
5>
```
```
 >brew list erlang
/usr/local/Cellar/erlang/25.0.2/bin/ct_run
/usr/local/Cellar/erlang/25.0.2/bin/dialyzer
/usr/local/Cellar/erlang/25.0.2/bin/epmd
/usr/local/Cellar/erlang/25.0.2/bin/erl
/usr/local/Cellar/erlang/25.0.2/bin/erlc
/usr/local/Cellar/erlang/25.0.2/bin/escript
/usr/local/Cellar/erlang/25.0.2/bin/run_erl
/usr/local/Cellar/erlang/25.0.2/bin/to_erl
/usr/local/Cellar/erlang/25.0.2/bin/typer
```

**Additional context**
It looks like this is a problem when server is configured to validate client certificate and client is sending an empty response

IngelaAndin:
In TLS-1.3 if chain certs are missing (so server auth domain adherence
can not be determined) send peer cert and hope the server is able to
recreate a chain in its auth domain.

Closes #6105

### DIFF
--- a/lib/ssl/src/tls_handshake_1_3.erl
+++ b/lib/ssl/src/tls_handshake_1_3.erl
@@ -1446,7 +1446,7 @@ process_certificate_request(#certificate_request_1_3{
     CertKeyPairs = ssl_certificate:available_cert_key_pairs(CertKeyAlts, Version),
     Session = select_client_cert_key_pair(Session0, CertKeyPairs,
                                           ServerSignAlgs, ServerSignAlgsCert, ClientSignAlgs,
-                                          CertDbHandle, CertDbRef, CertAuths),
+                                          CertDbHandle, CertDbRef, CertAuths, undefined),
     {ok, {State#state{client_certificate_status = requested, session = Session}, wait_cert}}.
 
 process_certificate(#certificate_1_3{
@@ -3029,16 +3029,24 @@ default_or_fallback(Default, _) ->
 
 select_client_cert_key_pair(Session0,
                             [#{private_key := NoKey, certs := [[]] = NoCerts}],
-                            _,_,_,_,_,_) ->
+                            _,_,_,_,_,_, _) ->
     %% No certificate supplied : send empty certificate
     Session0#session{own_certificates = NoCerts,
                      private_key = NoKey};
-select_client_cert_key_pair(Session, [],_,_,_,_,_,_) ->
-    %% No certificate compliant with supported algorithms and extensison : send empty certificate in state 'wait_finished'
+select_client_cert_key_pair(Session, [],_,_,_,_,_,_, undefined) ->
+    %% No certificate compliant with supported algorithms and
+    %% extensison : send empty certificate in state 'wait_finished'
     Session#session{own_certificates = [[]],
                     private_key = #{}};
+select_client_cert_key_pair(_,[],_,_,_,_,_,_, #session{} = Plausible) ->
+    %% If we do not find an alternative chain with a cert signed in auth_domain,
+    %% but have a single cert without chain certs it might be verifiable by
+    %% a server that has the means to recreate the chain 
+    Plausible;
 select_client_cert_key_pair(Session0, [#{private_key := Key, certs := [Cert| _] = Certs} | Rest],
-                            ServerSignAlgs, ServerSignAlgsCert, ClientSignAlgs, CertDbHandle, CertDbRef, CertAuths) ->
+                            ServerSignAlgs, ServerSignAlgsCert, 
+                            ClientSignAlgs, CertDbHandle, CertDbRef, 
+                            CertAuths, Plausible0) ->
     {PublicKeyAlgo, SignAlgo, SignHash, MaybeRSAKeySize, Curve} = get_certificate_params(Cert),
     case select_sign_algo(PublicKeyAlgo, MaybeRSAKeySize, ServerSignAlgs, ClientSignAlgs, Curve) of
         {ok, SelectedSignAlg} ->
@@ -3051,15 +3059,27 @@ select_client_cert_key_pair(Session0, [#{private_key := Key, certs := [Cert| _] 
                                              own_certificates = EncodedChain,
                                              private_key = Key
                                             };
-                        {error, _, not_in_auth_domain} ->
+                        {error, EncodedChain, not_in_auth_domain} ->
+                            Plausible = plausible_missing_chain(EncodedChain, Plausible0,
+                                                                SelectedSignAlg, Key, Session0),
                             select_client_cert_key_pair(Session0, Rest, ServerSignAlgs, ServerSignAlgsCert,
-                                                        ClientSignAlgs, CertDbHandle, CertDbRef, CertAuths)
+                                                        ClientSignAlgs, CertDbHandle, CertDbRef, CertAuths,
+                                                        Plausible)
                     end;
                 _ ->
                     select_client_cert_key_pair(Session0, Rest, ServerSignAlgs, ServerSignAlgsCert, ClientSignAlgs,
-                                                CertDbHandle, CertDbRef, CertAuths)
+                                                CertDbHandle, CertDbRef, CertAuths, Plausible0)
             end;
         {error, _} ->
             select_client_cert_key_pair(Session0, Rest, ServerSignAlgsCert, ServerSignAlgsCert, ClientSignAlgs,
-                                        CertDbHandle, CertDbRef, CertAuths)
+                                        CertDbHandle, CertDbRef, CertAuths, Plausible0)
     end.
+
+plausible_missing_chain([_] = EncodedChain, undefined, SignAlg, Key, Session0) ->
+    Session0#session{sign_alg = SignAlg,
+                     own_certificates = EncodedChain,
+                     private_key = Key
+                    };
+plausible_missing_chain(_,Plausible,_,_,_) ->
+    Plausible.
+
--- a/lib/ssl/test/ssl_cert_SUITE.erl
+++ b/lib/ssl/test/ssl_cert_SUITE.erl
@@ -72,6 +72,8 @@
          verify_fun_always_run_server/1,
          incomplete_chain_auth/0,
          incomplete_chain_auth/1,
+         no_chain_client_auth/0,
+         no_chain_client_auth/1,
          invalid_signature_client/0,
          invalid_signature_client/1,
          invalid_signature_server/0,
@@ -233,6 +235,7 @@ all_version_tests() ->
      verify_fun_always_run_client,
      verify_fun_always_run_server,
      incomplete_chain_auth,
+     no_chain_client_auth,
      invalid_signature_client,
      invalid_signature_server,
      critical_extension_auth,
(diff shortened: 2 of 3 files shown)


lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -3051,15 +3059,27 @@ select_client_cert_key_pair(Session0, [#{private_key := Key, certs := [Cert| _]
...
-                                                CertDbHandle, CertDbRef, CertAuths)
+                                                CertDbHandle, CertDbRef, CertAuths, Plausible0)
             end;
         {error, _} ->
             select_client_cert_key_pair(Session0, Rest, ServerSignAlgsCert, ServerSignAlgsCert, ClientSignAlgs,
-                                        CertDbHandle, CertDbRef, CertAuths)
+                                        CertDbHandle, CertDbRef, CertAuths, Plausible0)
     end.
+
+plausible_missing_chain([_] = EncodedChain, undefined, SignAlg, Key, Session0) ->
```
jjcarstens, line 3078:
```suggestion
plausible_missing_chain([_ | _] = EncodedChain, undefined, SignAlg, Key, Session0) ->
```
Does this need to be restricted to only one single cert in the EncodedChain?

IngelaAndin: Well, the assumption I made was that the case where the domain adherence check becomes a problem is when the client is "misconfigured" to only supply its own cert and not the other certs in the chain.  As in this case, we cannot perform the check,  but the server may be able to verify it anyway (TLS-1.3 spec mandates to try and do that). So in this case we don't really know and it is worth sending what we got. I could send a default always if the check fails, but that kind of makes the check pointless, and the TLS-1.3 spec also says the client SHOULD adhere to the server's auth domain.

IngelaAndin: @jjcarstens do you have an actual use case when only part of the chain is supplied? (Root cert can safely be left out).

jjcarstens: Maybe. We roughly follow the same setup for [AWS IOT Client authentication using X.509 Certificates](https://docs.aws.amazon.com/iot/latest/developerguide/x509-client-certs.html). The server authentication to complete handshake is the same - The device has the certificate chain and includes it in `cacerts` of the request. But the device/client also includes the signer CA used to sign it's device certificate as well which is typically a self-signed and managed CA and also put into the `cacerts` request.

This signer CA is not required by our server and is only used when the device certificate being presented is unknown (i.e. not pinned in our DB) and is required to be registered with our service first (also like [AWS IOT requires registering CA's first](https://docs.aws.amazon.com/iot/latest/developerguide/register-CA-cert.html)). In that situation, we'll run PKIX validation with this signer CA and presented client cert before allowing the connection on the server. If we already know about the client certificate (after looking up its calculated pinned value in the DB), then the signer CA is simply ignored.

It's this extra, unchained signer CA that is getting us with the TLS 1.3 changes. I also don't claim to be an SSL/TLS expert, so it could be that I'm doing things completely wrong and don't fully understand the standards.

IngelaAndin: Well,  TLS-1.3  RFC says  this about the extension:

  ```
 The client MAY send the "certificate_authorities" extension in the
   ClientHello message.  The server MAY send it in the
   CertificateRequest message.
```

```
4.4.2.3.  Client Certificate Selection

   The following rules apply to certificates sent by the client:

   -  The certificate type MUST be X.509v3 [RFC5280], unless explicitly
      negotiated otherwise (e.g., [RFC7250]).

   -  If the "certificate_authorities" extension in the
      CertificateRequest message was present, at least one of the
      certificates in the certificate chain SHOULD be issued by one of
      the listed CAs.

   -  The certificates MUST be signed using an acceptable signature
      algorithm, as described in Section 4.3.2.  Note that this relaxes
      the constraints on certificate-signing algorithms found in prior
      versions of TLS.

   -  If the CertificateRequest message contained a non-empty
      "oid_filters" extension, the end-entity certificate MUST match the
      extension OIDs that are recognized by the client, as described in
      Section 4.2.5.

```

So I think that your use case fits better with having an option to disable the sending of the certificat_authorties extension.

In TLS 1.2 this particular extension is not present but an equivalent value exists sent in the certificate_reques, although the RFC is less clear on how to act, and before it was just ignored, so we let it be backward compatible.

PR #6206, opened 2022-08-09
dialyzer: Parallelise hashing of source files

TD5:
Unlike with classic Dialyzer, incremental Dialyzer doesn't just hash a
`.beam` as a binary blob - it first filters out some non-deterministic
data, such as absolute file paths. For historical reasons, we did this
serially, but running it in parallel seems to make a fully cached
analysis about 2x faster in the cases that I have tried.

For longer analyses, the module hashing is proportionally a much smaller
amount of the elapsed time, so the saving in time spent hashing isn't
material.


bjorng:
Thanks! Added our daily builds.

PR #6207, opened 2022-08-09
dialyzer: Fix missing incremental analysis type option

TD5:
Adds the `incremental` analysis mode into the appropriate option types.

This was causing Dialyzer warnings for users that consumed the
incremental API to Dialyzer itself (e.g. erlang/rebar3#2736), since the `incremental`
option was available at runtime, but not declared in `dial_option()`.

### DIFF
--- a/lib/dialyzer/src/dialyzer.hrl
+++ b/lib/dialyzer/src/dialyzer.hrl
@@ -150,7 +150,8 @@
                                            'plt_add' |
                                            'plt_build' |
                                            'plt_check' |
-                                           'plt_remove'}
+                                           'plt_remove' |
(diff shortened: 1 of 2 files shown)


bjorng:

oliver-kriska:
Hello, I know it's closed issue but I would like to ask when this fix will be released. Because it's not in current 25.2 release. Thanks https://github.com/erlang/otp/blob/OTP-25.2/lib/dialyzer/test/opaque_SUITE_data/src/recrec/dialyzer.hrl#L102

bjorng:
It will be included in OTP 26, planned to be released in May next year.

PR #6212, opened 2022-08-10
kernel: Fix Logger filter removal if message format is atom or binary

dmitrivereshchagin:
Log event check performed in `logger_backend:do_apply_filters/4` doesn't tolerate atoms and binaries as message format strings.  This may cause a filter to be removed

    > logger:notice(<<"hello">>, []).
    Logger - error: {removed_failing_filter,no_domain}

For atom format strings it was partially fixed in OTP 25 by forcing atoms to strings before filters are applied (17c8fcf).


garazdawi:
Thanks!

PR #6223, opened 2022-08-12
[inets/3392] Fix for CVE-2016-1000107

### LINKED ISSUE #3392: ERL-198: httpoxy
Original reporter: `dominics`
Component: `inets`
Migrated from: https://bugs.erlang.org/browse/ERL-198

---

```
I'm the original researcher on httpoxy.org. I just wanted to reach out to make sure you're aware that this would affect any HTTP client that reads HTTP_PROXY without checking for a CGI environment.

And also that common CGI implementations are probably going to start refusing to carry a Proxy header into HTTP_PROXY just to make a common library-level mistake less of a security problem. I wonder if you might consider a similar mitigation in mod_cgi (I think this is what Go does at the language level.)

I should probably stress that inets HTTP Server's CGI handler is likely implementing RFC 3875 perfectly. It's just that when the spec was finalized, in 2003, the issue was already known about and fixed in libwww-perl, and I guess they just didn't notice and make it a special case in the spec (the informal use of HTTP_PROXY predates the spec, but then CGI as a whole goes way back into the early days of the web - I'm leaning toward blaming nobody; just an unfortunate coincidence that has been exploitable for over a decade.)

Through the DWF, CVE-2016-1000107 has been assigned to the issue of CGI applications running under the http://erlang.org/doc/apps/inets/http_server.html mod_cgi receiving the injected HTTP_PROXY environment variable. (Obviously, you also need to trust it to be vulnerable, but this does not seem uncommon.)

Please let me know if you have any questions. Sorry if this is the wrong place to report this.

```

marcellanz:
Fixes #3392

Open tasks
- [ ] add test(s)
- [ ] probably document the behaviour for `httpd_script_env:create_env` that it skips `PROXY` HTTP headers because of this CVE.
- [ ] review my _beginners erlang code_

### DIFF
--- a/lib/inets/src/http_server/httpd_script_env.erl
+++ b/lib/inets/src/http_server/httpd_script_env.erl
@@ -132,6 +132,8 @@ create_http_header_elements(ScriptType, [{Name, [Value | _] = Values } |
 create_http_header_elements(ScriptType, [{Name, Value} | Headers], Acc, OtherAcc) 
   when is_list(Value) ->
     try http_env_element(ScriptType, Name, Value) of
+        skipped ->
+            create_http_header_elements(ScriptType, Headers, Acc, [OtherAcc]);
         Element ->
             create_http_header_elements(ScriptType, Headers, [Element | Acc],
                                        OtherAcc)
@@ -141,6 +143,11 @@ create_http_header_elements(ScriptType, [{Name, Value} | Headers], Acc, OtherAcc
                                        [{Name, Value} | OtherAcc])
     end.
 
+http_env_element(cgi, "proxy", _Value)  ->
+  %% CVE-2016-1000107 – https://github.com/erlang/otp/issues/3392
+  skipped;
+http_env_element(cgi, "PROXY", _Value)  ->
+  skipped;
 http_env_element(cgi, VarName0, Value)  ->
     VarName = re:replace(VarName0,"-","_", [{return,list}, global]),
     {"HTTP_"++ http_util:to_upper(VarName), Value};


artman41:
@marcellanz is `proxy` a case-insentive varname or is it either `proxy` or `PROXY`?

If it's case-insensitive, I'd be looking to add a case statement doing `case string:lowercase(VarName0) of` to the existing cgi  case for `http_env_element/3`

RaimoNiskanen:
Should we handle mixed case variants too, such as pRoXy?

KennethL:
Since we are creating an upper case HTTP_PROXY if the request contains a HTTP header named `proxy` in whatever mixed case the comparision ought to be done on a header name which is normalized to lower or upper case. 
Something like this:
```erlang
http_env_element(cgi, VarName0, Value)  ->
    case http_util:to_upper(VarName0) of
           "PROXY" -> 
                   skipped;
          VarName1 ->
                   VarNameUpper = re:replace(VarName1,"-","_", [{return,list}, global]),
                  {"HTTP_"++ VarNameUpper, Value}
     end.
```

marcellanz:
I agree. According to the defining RFC 7230, http header field names are case-insensitive, normalizing to an uppercase value seems to make perfect sense. Similar the CGI [spec](https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.18) defines the protocol specific meta variables are converted to upper case names; so it would be safe to use that normalization here too.

RaimoNiskanen:
According to [RFC 3875](https://www.rfc-editor.org/rfc/rfc3875#section-4.1), in particular [section 4.1.18](https://www.rfc-editor.org/rfc/rfc3875#section-4.1.18) variable names are case insensitive, so a case canonicalization before comparison is required.

Why not speculatively do both `re:replace/4` and `http_util:to_upper/1` and then compare with "PROXY"?

marcellanz:
This makes sense for sure. I was not sure if the current implementation actually does this somewhere before with how CGI is implemented and a variable would arrive already upper case. Implementing this in a defensive way still is preferable I think.

marcellanz:
I've updated the branch to reflect the discussion above by canonicalizing the HTTP variable name to upper case before comparing. I'm not sure with tests. I've tested the behaviour like I have described it [here](https://github.com/erlang/otp/issues/3392#issuecomment-1213289354); for a permanent test with this change, would one do a integration like test or would a scoped test towards `http_env_element` be enough? I'm not used to work with the libraries so far, so my question.

IngelaAndin:
I think I could be ok with a white box test in http_format_SUITE. You would have to create some internal data to be able to call
httpd_script_env:create_env/3

marcellanz:
I'll provide such a test.

marcellanz:
I was absorbed with a lot of "work work" lately; I will find some time over xmas to provide this test.

IngelaAndin:
@marcellanz  ping

marcellanz:
true – I'd do that soon enough. Actually "work work" reduced a bit so I shall be able to do it. Thanks for the patience.

RaimoNiskanen:
Ping?

PR #6227, opened 2022-08-17
kernel: invoke init:run_on_load_handlers/0 after code_server start

### LINKED ISSUE #6085: kernel_config fails to sync nodes due to crypto nifs not being loaded (TLS)
**Describe the bug**
Given an **embedded** Erlang node is configured to use **TLS** and wait for node syncing through the **sync_nodes_optional**/**sync_nodes_mandatory** and **sync_nodes_timeout** parameters, when starting up the node an error is thrown:
```
=ERROR REPORT==== 14-Jun-2022::17:56:28.705000 ===
Error in process <0.380.0> on node test2@myhost with exit value:
{undef,[{crypto,supports,[],[]},
        {tls_record,sufficient_crypto_support,1,
                    [{file,"tls_record.erl"},{line,378}]},
        {ssl,validate_versions,2,[{file,"ssl.erl"},{line,2478}]},
        {ssl,handle_option,4,[{file,"ssl.erl"},{line,1928}]},
        {ssl,process_options,3,[{file,"ssl.erl"},{line,1648}]},
        {ssl,handle_options,3,[{file,"ssl.erl"},{line,1602}]},
        {ssl,connect,4,[{file,"ssl.erl"},{line,604}]},
        {inet_tls_dist,do_setup_connect,10,
                       [{file,"c:/Program Files/erl-23.2/lib/ssl-10.2/src/inet_tls_dist.erl"},
                        {line,540}]}]}
```

This leads to the kernel_config to wait for the full duration configured through the sync_nodes_timeout and proceeds without syncing the specified nodes configured through either sync_nodes_mandatory or  sync_nodes_optional despite the other nodes being up.

**To Reproduce**
Common steps:
1. Create a boot script with the necessary applications so that the node would use TLS for distribution, e.g.:
```
{release, {"Test", "0"}, {erts, "11.1.4"},
      [{kernel, "7.2"},
      {stdlib, "3.14"},
      {sasl, "4.0.1"},
      {asn1, "5.0.14"},
      {crypto, "4.8.1"},
      {public_key, "1.9.2"},
      {ssl, "10.2"}]}.
```

Node A:
1. Start node with the following arguments: 
`erl -sname test1 -mode embedded -boot path_to_boot_script -proto_dist inet_tls`

Node B:
1. Create a node config file with the following parameters:
```
[{kernel, 
	[{sync_nodes_optional, ['test1@myhost']},
	 {sync_nodes_timeout, 10000}]}].
```
2. Start node with the following arguments:
`erl -sname test2 -mode embedded -boot path_to_boot_script -config path_to_config  -proto_dist inet_tls`

**Expected behavior**
Node B has to sync with node A without any errors being thrown and without the sync_node_timeout being breached.

**Affected versions**
All.

**Additional context**
The root of this issue seems to be that the crypto application nifs are not loaded by the time when the kernel_config supervisor attempts to establish a connection with the other nodes, which leads to the undefined function error. The crypto nifs most likely are expected to be loaded during the [kernel_safe_sup supervisor](https://github.com/erlang/otp/blob/master/lib/kernel/src/kernel.erl#L189) initialization which, according to [this](https://github.com/erlang/otp/blob/master/lib/kernel/src/kernel.erl#L173) list, is done after the kerner_config supervisor is started, thus leading to the above issue.

vngthb:
Moving the invocation of init:on_load_handlers/0 from kernel:init(safe) to code:do_start/0.

init:on_load_handlers/0 has been added as part of https://github.com/erlang/otp/commit/7f82edd and it was implemented in such a way that it will be invoked, quoting, "after having started most servers in the kernel application", problem is that some of those servers, under specific configurations (see #6085), need the on load handlers in order for them to properly start.

As mentioned, this should fix #6085.

### DIFF
--- a/lib/ssl/Makefile
+++ b/lib/ssl/Makefile
@@ -40,4 +40,6 @@ include $(ERL_TOP)/make/otp_subdir.mk
 
 DIA_PLT_APPS=crypto runtime_tools inets public_key
 
+TEST_NEEDS_RELEASE=true
+
 include $(ERL_TOP)/make/app_targets.mk
--- a/lib/ssl/src/inet_tls_dist.erl
+++ b/lib/ssl/src/inet_tls_dist.erl
@@ -421,6 +421,7 @@ wait_for_code_server() ->
 	    timer:sleep(10),
 	    wait_for_code_server();
 	Pid when is_pid(Pid) ->
+            init:run_on_load_handlers([crypto,asn1rt_nif]),
 	    ok
     end.
 
@@ -552,6 +553,8 @@ setup_fun(Driver, Kernel, Node, Type, MyNode, LongOrShortNames, SetupTime) ->
 
 -spec do_setup(_,_,_,_,_,_,_) -> no_return().
 do_setup(Driver, Kernel, Node, Type, MyNode, LongOrShortNames, SetupTime) ->
+    wait_for_code_server(),
+
     {Name, Address} = split_node(Driver, Node, LongOrShortNames),
     ErlEpmd = net_kernel:epmd_module(),
     {ARMod, ARFun} = get_address_resolver(ErlEpmd, Driver),
--- a/lib/ssl/test/ssl_dist_SUITE.erl
+++ b/lib/ssl/test/ssl_dist_SUITE.erl
@@ -37,6 +37,8 @@
 %% Test cases
 -export([basic/0,
          basic/1,
+         embedded/0,
+         embedded/1,
          monitor_nodes/1,
          payload/0,
          payload/1,
@@ -105,6 +107,7 @@ start_ssl_node_name(Name, Args) ->
 %%--------------------------------------------------------------------
 all() ->
     [basic,
+     embedded,
      monitor_nodes,
      payload,
      dist_port_overload,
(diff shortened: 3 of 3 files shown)


bjorng (changes requested):
Thanks for your pull request.

Please sign the [Contributor License Agreement](https://cla-assistant.io/erlang/otp?pullRequest=6227).

With this pull request, the test case [code_SUITE:bit_boot_embedded/1](https://github.com/erlang/otp/blob/d079ee1e20cb289fa858849cd4cfdf2f689dee91/lib/kernel/test/code_SUITE.erl#L1198-L1206) fails. The reason is that the on_load handler for the `dyntrace` module expects the file server process to be started.

I think that a better place to run the on_load handlers would be in `kernel_config:start_link/0`. Can you do that change and verify that it solves #6085?

When you have done that, please update the commit message to contain the same information as the pull request description. We want anyone that want to know about the change in 10 years time to find the information directly in the commit message instead of having to locate the pull request.

bjorng:
@rickard-green Could there be a problem for on_load handlers if the `kernel_refc` process is not started? If so, could the `kernel_refc` server be started before the `kernel_config process`?

@garazdawi The logger supervisor is started before the `kernel_config` process in `minimal` mode, but after the `safe` supervisor in other modes. I suppose there is a good reason for that? Can there be any consequences if an on_load handler is called before the logger supervisor is started?

vngthb:
> Please sign the [Contributor License Agreement](https://cla-assistant.io/erlang/otp?pullRequest=6227).

Done!

> With this pull request, the test case [code_SUITE:bit_boot_embedded/1](https://github.com/erlang/otp/blob/d079ee1e20cb289fa858849cd4cfdf2f689dee91/lib/kernel/test/code_SUITE.erl#L1198-L1206) fails. The reason is that the on_load handler for the `dyntrace` module expects the file server process to be started.
> 
> I think that a better place to run the on_load handlers would be in `kernel_config:start_link/0`. Can you do that change and verify that it solves #6085?

Made the suggested change and it does solve it, updated the commit.

> When you have done that, please update the commit message to contain the same information as the pull request description. We want anyone that want to know about the change in 10 years time to find the information directly in the commit message instead of having to locate the pull request.

Updated the commit message as well.

bjorng:

bjorng (approved):

garazdawi:
> @garazdawi The logger supervisor is started before the `kernel_config` process in `minimal` mode, but after the `safe` supervisor in other modes. I suppose there is a good reason for that? Can there be any consequences if an on_load handler is called before the logger supervisor is started?

I think it should be fine. Any log messages sent during on_load will be handled by the simple logger and get logged correctly when the logger supervisor is started.

garazdawi:
Hmm, one of the problems with this change is that any on_load callback will no longer have access to standard_out or standard_error. So any call to `io:format()` will crash. Logging will still work as it falls back to erlang:display if needed...

I'm unsure how common it is print to stdout/stderr when on_load is called. All OTP code uses logger to report any errors.

garazdawi:
I've also found that #6144 needs this PR as the tty nif has to be loaded before user is started in the kernel supervision tree.

garazdawi:
I looked a bit closer at this, and at least for the new tty nif added in #6144 I've decided that the best way is to extend init to allow calling of on_load for a specific module earlier in the boot sequence.

You can view the changes I did [here](https://github.com/erlang/otp/pull/6144/commits/eb0165e169a628f19e92a7ab9fc060db6e18db5d). If we are unsure if moving the place where on_load are called then we could make the tls dist module call the same init function for crypto.

garazdawi:
Me and @bjorng have talked a bit and agreed that the solution is to use the [init function](https://github.com/erlang/otp/pull/6144/commits/cf981d3849a8b737fc2a7be44bc86835015ced5d) I've introduced in #6144 for loading crypto if needed for the distribution.

I will push new commits to this PR when I've merged that PR so that you can test that they work in your system before merging.

garazdawi:
I've updated your branch with the changes needed and added a test case to make sure that it works. Please try it and make sure that it also solves your problem.

@IngelaAndin @RaimoNiskanen I would appreciate any comments you have on the contents of this PR.

vngthb:
@garazdawi, sorry for the late response, it does indeed solve the issue described in #6085, thank you!

garazdawi:
Thanks for the trouble report and the PR!

PR #6228, opened 2022-08-17
Allow disabling `certificate_authorities` extension on server with `verify_peer`

### LINKED ISSUE #6106: Allow disabling certificate_authorities extension with peer-to-peer setup with TLS 1.3

jjcarstens:
Fixes #6106

/cc @IngelaAndin 

Based on the discussion from the issue (and #6204), it was decided that adding the
ability to disable the `certificate_authorities` extension on the server would be
the fitting resolution for situations where you do not want that extension forced
in TLS 1.3.

This adds that ability to specify as a server option and defaults to `true` to keep
with existing expected functionality.

### DIFF
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -1033,7 +1033,7 @@ fun(srp, Username :: binary(), UserState :: term()) ->
     </datatype>
 
     <datatype>
-      <name name="certificate_authorities"/>
+      <name name="client_certificate_authorities"/>
       <desc>
 	<p>If set to true, sends the certificate authorities extension in TLS-1.3 client hello.
 	The default is false. Note that setting it to true may result in a big overhead if you
@@ -1248,6 +1248,15 @@ fun(srp, Username :: binary(), UserState :: term()) ->
       is supplied it overrides option <c>cacertfile</c>.</p>
 	</desc>
       </datatype>
+
+  <datatype>
+    <name since="OTP 25.2" name="server_certificate_authorities"/>
+      <desc>
+      <p>When used with <c>{verify, verify_peer}</c> on the server with TLS-1.3, the certificate
+      authorities extension will be included or excluded from the certificate request. The Default
+      is true.</p>
+      </desc>
+  </datatype>
             
       <datatype>
 	<name name="server_cafile"/>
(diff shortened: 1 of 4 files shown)


jjcarstens:
Sorry @IngelaAndin ! Life got crazy for a bit and this got neglected, but I'm back on track. I've added a test for my use case and _think_ I updated the docs as expected. Let me know if I need to change anything

lib/ssl/test/ssl_cert_SUITE.erl
```diff
@@ -1287,6 +1290,41 @@ basic_rsa_1024(Config) ->
                   {fail_if_no_peer_cert, true} | ServerOpts1],
     ssl_test_lib:basic_test(ClientOpts, ServerOpts, Config).
 
+%%--------------------------------------------------------------------
+server_certificate_authorities_disabled() ->
+     [{doc,"TLS 1.3: Disabling certificate_authorities extension on the server when verify_peer is set to true" 
+       " allows client auth with empty chain"}].
```
IngelaAndin, line 1296: I think we need to change the comment to say something like  "allows the client to send a chain that could be verifiable by the server but that would not adhere to the certificate_authorities extension as it is not part of the regular trusted certificate set"

jjcarstens:
Thanks for the feedback! Implemented the suggestions and ready for another look

lib/ssl/doc/src/ssl.xml
```diff
@@ -1248,6 +1248,15 @@ fun(srp, Username :: binary(), UserState :: term()) ->
       is supplied it overrides option <c>cacertfile</c>.</p>
 	</desc>
       </datatype>
+
+  <datatype>
+    <name name="server_certificate_authorities"/>
```
IngelaAndin, line 1253:
```suggestion
    <name since ="OTP 25.2" name="server_certificate_authorities"/>
```

IngelaAndin:
I think it looks good now :) If you would add the since tag I suggested in the documentation this is good for merging to maint for OTP-25.2 inclusion.

jjcarstens:
🎉 Updated with `since` tag. Thanks!

IngelaAndin:
Thanks, will merge this tomorrow, as after your change I need to reenable it in the test to see that the documentation builds correctly.

PR #6231, opened 2022-08-19
Add timeout for an incoming ssh connection that does not open a channel after completed authentication

HansN:
Adds the daemon timeout `max_initial_idle_time` for an incoming ssh connection that does not open a channel after completed authentication.  It defaults to `infinity` to maintain compatibility.

The existing `idle_time` timeout is for terminating a connection that opens channels, closes them but does not close the connection.


HansN:
Will be released in OTP-25.1

PR #6232, opened 2022-08-19
ets:lookup_element badarg for Pos

leonardb:
Document badarg when Pos is larger than the tuple the element it to be extracted from.

This came up when testing cases in the discussion on adding support for ets:lookup_element/4

https://erlangforums.com/t/proposal-adding-ets-lookup-element-as-list-3/1743/11


garazdawi:
Thanks!

PR #6234, opened 2022-08-22
Add lookup_element/4, which takes a default value

RobinMorisset:
Originally proposed in https://erlangforums.com/t/proposal-adding-ets-lookup-element-as-list-3/1743/4.

lookup_element/3 tends to be faster than lookup/2 when the key is always found, and is also a bit more ergonomic.
Unfortunately, when the key is absent, it throws an exception which is both somewhat slow and
less ergonomic to catch than matching the [] returned by lookup.
LeonardB suggested that a simple solution would be to offer an alternate version of lookup_element that takes an
extra argument, and returns that argument if the key is not found. This matches several other functions in the
standard library such as maps:get and proplists:get_value.

I confirmed the theory with some very simple microbenchmark: https://gist.github.com/RobinMorisset/62030c8fcb7376e16291904a9098e342

Code                                                         ||        QPS       Time     Rel
bench_lookup:bench_lookup_element_default_present().          1     133 Ki    7538 ns    100%
bench_lookup:bench_lookup_element_present().                  1     131 Ki    7604 ns     99%
bench_lookup:bench_lookup_present().                          1      89464   11177 ns     67%

bench_lookup:bench_lookup_element_default_absent().          1     140 Ki    7156 ns    100%
bench_lookup:bench_lookup_absent().                          1     134 Ki    7477 ns     95%
bench_lookup:bench_lookup_element_absent().                  1      88187   11339 ns     63%

As expected, lookup_element/4 is as fast as lookup_element/3 when the element is present, and as fast as lookup/2 when it is absent.

I've updated ets_SUITE.erl, and also done some manual testing. Finally there is the microbenchmark above.

### DIFF
--- a/erts/emulator/beam/erl_db.c
+++ b/erts/emulator/beam/erl_db.c
@@ -2670,8 +2670,44 @@ BIF_RETTYPE ets_lookup_element_3(BIF_ALIST_3)
     }
 }
 
-/* 
- * BIF to erase a whole table and release all memory it holds 
+/*
+** Get an element from a term
+** get_element_4(Tab, Key, Index, Default)
+** return the element or a list of elements if bag or Default if the element is not present
+*/
+BIF_RETTYPE ets_lookup_element_4(BIF_ALIST_4)
+{
+    DbTable* tb;
+    Sint index;
+    int cret;
+    Eterm ret;
+
+    CHECK_TABLES();
+
+    DB_BIF_GET_TABLE(tb, DB_READ, LCK_READ, BIF_ets_lookup_element_4);
+
+    if (is_not_small(BIF_ARG_3) || ((index = signed_val(BIF_ARG_3)) < 1)) {
+	    db_unlock(tb, LCK_READ);
+	    BIF_ERROR(BIF_P, BADARG);
+    }
+
+    cret = tb->common.meth->db_get_element(BIF_P, tb,
+					   BIF_ARG_2, index, &ret);
+    db_unlock(tb, LCK_READ);
+    switch (cret) {
+        case DB_ERROR_NONE:
+            BIF_RET(ret);
+        case DB_ERROR_BADKEY:
+            BIF_RET(BIF_ARG_4);
+        case DB_ERROR_SYSRES:
+            BIF_ERROR(BIF_P, SYSTEM_LIMIT);
+        default:
+            BIF_ERROR(BIF_P, BADARG);
+    }
+}
+
+/*
+ * BIF to erase a whole table and release all memory it holds
  */
 BIF_RETTYPE ets_delete_1(BIF_ALIST_1)
 {
--- a/lib/stdlib/test/ets_SUITE.erl
+++ b/lib/stdlib/test/ets_SUITE.erl
@@ -39,7 +39,7 @@
 -export([tab2file/1, tab2file2/1, tabfile_ext1/1,
 	 tabfile_ext2/1, tabfile_ext3/1, tabfile_ext4/1, badfile/1]).
 -export([heavy_lookup/1, heavy_lookup_element/1, heavy_concurrent/1]).
--export([lookup_element_mult/1]).
+-export([lookup_element_mult/1, lookup_element_default/1]).
 -export([foldl_ordered/1, foldr_ordered/1, foldl/1, foldr/1, fold_empty/1]).
 -export([t_delete_object/1, t_init_table/1, t_whitebox/1,
          select_bound_chunk/1, t_delete_all_objects/1, t_test_ms/1,
@@ -191,7 +191,7 @@ groups() ->
        privacy]},
      {insert, [], [empty, badinsert]},
      {lookup, [], [badlookup, lookup_order]},
-     {lookup_element, [], [lookup_element_mult]},
+     {lookup_element, [], [lookup_element_mult, lookup_element_default]},
      {delete, [],
       [delete_elem, delete_tab, delete_large_tab,
        delete_large_named_table, evil_delete, table_leak,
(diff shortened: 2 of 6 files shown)


RobinMorisset:
I've been unable to reproduce the failure locally, and I don't see how adding a bif to ets could cause a bug in persistent terms, so I suspect the test is just flaky.
Is there a way to convince the CI to just retry it?

juhlig:
> Is there a way to convince the CI to just retry it?

AFAIK, this can be done only with another push.

RobinMorisset:
I've done a new push (after rebasing on top of master), can one of the maintainers approve running the test workflow?

RobinMorisset:
The failure is different from before, and seems shared with another recent PR: #6240, so I believe it has to be an issue with master rather than my PR.

Maria-12648430:
I like that this is being addressed, but I'm torn on whether `lookup_element` with a default value is a good way to do it. Actually, I liked the original suggestion (**Proposal** in the [opening post of the thread on Erlang Forums](https://erlangforums.com/t/proposal-adding-ets-lookup-element-as-list-3/1743?u=maria-12648430)) better. Well, except for the name, that is :sweat_smile: But returning a (possibly empty) list of lookup results seems to me preferable.

There exists an ambiguity in `lookup_element/3`, given that the key actually exists, which is related to the different table types. For the `set` types it returns the raw value that was found, for the `bag` types it returns a list of the values that were found. So, without always paying attention the table type, it is unclear if the value returned by the call was a list from a single object in a `set` table or a list of elements from multiple objects in a `bag`table.

***Example:***
```erlang
1> ets:new(t1, [named_table, set]).
t1
2> ets:insert(t1, {a, [z, y, x]}).
true
3> ets:lookup_element(t1, a, 2).
[z,y,x]
```
... and ...
```erlang
4> ets:new(t2, [named_table, bag]).
t2
5> ets:insert(t2, [{a, x}, {a, y}, {a, z}]).
true
6> ets:lookup_element(t2, a, 2).
[z,y,x]
```
... will return equal results (with the elements in undefined order in the latter case, yes).

With the original approach, the first snippet would return `[[x,y,z]]` instead, making it clear that the element returned stems from a single object, not from multiple ones.

`lookup_element/4` may even add to the confusion:
```erlang
7> ets:new(t3, [named_table, bag]).
t3
8> ets:lookup_element(t3, a, 2, [z, y, x]).
[z,y,x]
9> ets:insert(t2, [{a, x}, {a, y}, {a, z}]).
true
10> ets:lookup_element(t3, a, 2, [x,y,z]).
[z,y,x]
```

To be sure, I _like_ the proposal itself. But alluding to what @juhlig [said in the discussion](https://erlangforums.com/t/proposal-adding-ets-lookup-element-as-list-3/1743/13?u=maria-12648430), we should try to make it _good_, not another function with slightly different peculiarities.

mikpe:
IMO, the whole point of `lookup_element` is that it just returns the requested value without any wrapping, so to me adding a default (if the key isn't there you get the default, error on wrong tuple arity) is perfect. And it fits with other `ets` update ops that allow a default to be specified.

As for bags, I think the most rational solution is to return a list of those elements for the records that exist (and error on wrong tuple arity). Users should know if they're accessing a set or a bag, so the return shape shouldn't be a surprise to them.

max-au:
I'd love to have this function. Quite often I find myself writing either `case ets:lookup(Tab, PrimaryKey)` or `try ets:lookup_element(Tab, PrimaryKey, 2) catch`.

Maria-12648430:
> And it fits with other `ets` update ops that allow a default to be specified.

Hm, I disagree. This is not an update but a retrieval op, and the other retrieval ops (like `lookup`) return lists, regardless of the table type.

<hr />

Be that as it may, I think that we don't have to pick either the one or the other.

There could be `lookup_element/4` as in the current state of this PR, _and_ `lookup_elements/3` to return a list, and everyone could choose what he likes/needs and be happy.

From the discussion on Erlang Forums I understand that @RobinMorisset already had this done before switching to the default value approach. For completeness' sake (ie, not sure about useful), there could even be a `lookup_elements/4` accompanying it, taking a default value that would be used for object(s) that are too short.

RobinMorisset:
What is the next step on this @sverker ? Is there anything I should do?

sverker:
Next thing to happen if for me to read through and form an opinion. Then we will probably decide in an OTP Technical Board meeting.

sverker:
I think I prefer the PR basically as it is.

`ets:lookup_element(T, Key, Pos, Default) -> SetElement | [BagElements] | Default.`

But OTP Technical Board will have the final say.

RobinMorisset:
> I think I prefer the PR basically as it is.
> 
> `ets:lookup_element(T, Key, Pos, Default) -> SetElement | [BagElements] | Default.`
> 
> But OTP Technical Board will have the final say.

Thanks! How often does the OTP technical board meet/review proposals?

sverker:
The OTP Technical Board approved the API as it is.

I have some minor issues with the PR...

erts/emulator/beam/erl_db.c
```diff
@@ -2663,13 +2663,51 @@ BIF_RETTYPE ets_lookup_element_3(BIF_ALIST_3)
     case DB_ERROR_BADKEY:
         BIF_P->fvalue = EXI_BAD_KEY;
         BIF_ERROR(BIF_P, BADARG | EXF_HAS_EXT_INFO);
+    // TODO: error message for the DB_ERROR_BADITEM case
```
sverker, line 2666: Either fix the "TODO" or remove it. The error formatting seems to work without it.

erts/emulator/beam/erl_db.c
```diff
@@ -2663,13 +2663,51 @@ BIF_RETTYPE ets_lookup_element_3(BIF_ALIST_3)
...
+					   BIF_ARG_2, index, &ret);
+    db_unlock(tb, LCK_READ);
+    switch (cret) {
+        case DB_ERROR_NONE:
+            BIF_RET(ret);
+        case DB_ERROR_BADKEY:
+            BIF_RET(BIF_ARG_4);
+        case DB_ERROR_SYSRES:
+            BIF_ERROR(BIF_P, SYSTEM_LIMIT);
+        // TODO: error message for the DB_ERROR_BADITEM case
```
sverker, line 2703: Either fix TODO or remote it.

lib/stdlib/test/ets_SUITE.erl
```diff
@@ -4033,6 +4033,25 @@ fill_tab(Tab,Val) ->
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+lookup_element_default(Config) when is_list(Config) ->
```
sverker, line 4036: Complete the test with `ordered_set` and `duplicate_bag`.

RobinMorisset:
Thanks a lot for the review! I've removed the comments and added the tests requested.

RobinMorisset:
All failing tests affect either peer_SUITE or edlin_expand_SUITE. Since I did not touch the peer module in any way and I see that there have been several recent patches to it merged into master (https://github.com/erlang/otp/commit/cf6668e83463e4f5b9f9b9157c1f74c0b15d08ed), I suspect I am not to blame. I'll try to relaunch the tests in a few days to confirm it.

sverker:
@RobinMorisset 
I pushed small administrative fixup to the documentation.
I will squash and merge once the tests have gone through again, assuming you are satisfied.

RobinMorisset:
Thank you for the help and reviews, I'd be very happy to see the PR merged.

sverker:
Merged to master for OTP 26.
Thanks for the contribution, @RobinMorisset .

PR #6243, opened 2022-08-25
dialyzer: Fix unsound behaviour subtype checks

### LINKED ISSUE #6221: Dialyzer erroneously processes callback and spec types differently
### Describe the bug

> **tl;dr** Callback and spec types are "simplified" differently, which can break behaviour subtype checking since it is not always safe to grow a type, unlike normal overlap checking of specs and implementations. This results in Dialyzer raising an error about code which I consider to be correct.

Dialyzer sometimes simplifies types (e.g. types such as `1 | 2 | 3 | 4 | 5 | ...` can be simplified to `integer()`, or `[a] | [b]` to `[a | b]`). Dialyzer also limits how far it'll recurse into types, and has other mechanisms to keep it's analysis time quicker than it would otherwise be by making approximations.

Usually, these approximations are fine because Dialyzer checks types for compatibility by essentially taking the intersection of two types, and seeing whether it's empty (i.e. do the types "overlap"), and this sort of simplification taking place doesn't change whether a later intersection is empty or not (since "simplifications" always a superset of the original), and nor does it matter whether one of the two types is simplified more than the other. 

**However**, checking of specifications against a behaviour's callbacks are done via a different mechanism which Dialyzer considers "subtyping", which works a bit more like classical subtyping in object oriented systems and is sensitive to the growth in a type, and also isn't commutative (e.g. what was a subtype can be "simplified" into a bigger type which is no longer subtype). Typically, this works fine since its usually the case that the two types being compared will be simplified in the same way, but _types with the same structure can end up being "simplified" differently in specs vs. callbacks_, resulting in a callback and spec with the same type not being considered subtypes of one another. This results in Dialyzer rejecting valid code (i.e. code where the callback, spec and implementation all agree).

### To Reproduce

Given a behaviour in `i.erl`:
```
-module(i).

-callback foo() -> #{ {{{f,f}, f}, f} => x }.
```

And an implementation of it in `j.erl`:

```
-module(j).
-behaviour(i).

-export([foo/0]).

-type pair(A,B) :: {A,B}.

-type nested() :: pair(pair(pair(f,f),f),f).

-spec foo() -> #{ nested() => x }.
foo() ->
  Ret = #{ {{{f,f}, f}, f} => x },
  Ret.
```

We get a `callback_spec_type_mismatch` error like:
```
The return type #{{{_,'f'},'f'}=>'x'} in the specification of foo/0 is not a subtype of #{{{{'f','f'},'f'},'f'}=>'x'}, which is the expected return type for the callback of i behaviour
```

Note the `_` in the spec's return type - and not the callback's return type(!) - in the error message, despite both the spec and the callback both, once expanded, declaring to return a `#{ {{{f,f}, f}, f} => x }`.

Also note that if the declarations of the spec and callback were swapped, the `_` would be in the callback's type, and so the spec _would_ be considered a subtype, despite them not being simplified in the same way.

### Expected behavior
I would expect this case to be accepted, because both spec and callback return structurally the same type (and both should be simplified in the same way).

### Affected versions
Recent HEAD of OTP, but also older official releases.

### Additional context
In the above example, I showed this issue occurring with nested tuples where one element became `_`, breaking subtyping. However, with much bigger examples (as you might see in a production codebase, but much harder to put into a bug report), you can end up with breaking it more egregiously, making two identical types in spec and callback become an error like:

```
The return type {ok, map()} in the specification of foo/0 is not a subtype of {ok, #{ ... some huge map here ...}}, which is the expected return type for the callback of i behaviour
```

In fact, in sufficiently big examples with types defined across multiple files, and with header files including records and other types, this can happen even when the spec and callbacks _are defined in precisely the same way_, i.e. they don't need to just expand to the same structure, but they're literally syntactically the same type in the source files of both.

### Notes

I've not validated my explanation of what's going on here by stepping through Dialyzer during a debugging session, so I could be wrong about the precise mechanism at play, but whatever the explanation, I think this must be a bug since Dialyzer seems to be rejecting a callback, spec and implementation which express the same structure, which goes against the philosophy that "if Dialyzer complains, there's _something_ wrong in your specs or implementation".

TD5:
An invariant inside Dialyzer is that it is always safe to "simplify" a
type, (e.g. `1..1000` -> `pos_integer()`, or `a | b | c | .. | z` -> `atom()`).
This is safe when checking for overlap between specs and inferred types,
since the overlap never gets smaller when one of the types is grown, so
we don't generate false positives.

For the case of behaviours, we use subtyping when checking the specs
and inferred types with respect to the behaviour definition. In this
case, the spec in source code may correctly be a subtype of the behaviour,
but due to internal simplifications which "should" always be safe, we can
make what was a subtype of the behaviour into something which is not,
and hence generate a false positive error about valid code.

The fix here is to switch behaviour checking to use overlap checking as
is used elsewhere in Dialyzer: it's a well tested codepath and is
consistent with the simplification logic. The drawback is that if it
weren't for the simplification logic (which is required to keep
performance reasonable), subtyping would be the correct thing to enforce,
so those cases where a spec does have an overlap with the behaviour, but
is not a subtype, will no longer be reported, since Dialyzer can't
safely tell that situation apart from valid code which has been
simplified.

We also add special casing to match the original intent, which allows
callbacks to crash (i.e. have a `none()` return type).

Fixes #6221.


TD5:
There's a lot of subtlety hidden in Dialyzer's type checking logic, so I may well have overlooked something: @aronisstav, you might be well-placed to check this new logic.

lib/dialyzer/test/behaviour_SUITE_data/src/callbacks_and_specs/my_callbacks_wrong.erl
```diff
@@ -28,13 +28,13 @@
 callback_init(Parent) -> #state{parent = Parent}. %% Wrong return
 
 -spec callback_cast(state(), pid() | atom(), cast_message()) ->
-			   {'noreply' | 'reply', state()}. %% More generic spec
+			   {'reply', state()}. %% Non-overlapping spec
```
aronisstav, line 31: Only taking a quick look, and missing context, so wondering, just to double check: is the case of a "more-generic" spec (and therefore no warnings should be emitted) covered by other cases?

aronisstav (approved):
LGTM

lib/dialyzer/test/behaviour_SUITE_data/src/callbacks_and_specs/my_callbacks_wrong.erl
TD5: No, it's not covered elsewhere. I didn't really want to add that because in a sense catching a more generic spec would be a good thing - it's just that we can't because of the false-positives it would create, so I just created a test for that situation.

lib/dialyzer/test/behaviour_SUITE_data/src/otp_6221/my_callbacks_correct.erl
```diff
@@ -0,0 +1,17 @@
...
+
+-export([foo/0]).
+
+-type pair(A,B) :: {A,B}.
+
+-type nested() :: pair(pair(pair(f,f),f),f).
+
+% This is correctly implemented, but a combination of Dialyzer
+% "simplification" logic and subtyping rules for behaviours means
+% this implementation has historically been erroneously rejected
```
bjorng, line 13:
```suggestion
%% This is correctly implemented, but a combination of Dialyzer
%% "simplification" logic and subtyping rules for behaviours means
%% this implementation has historically been erroneously rejected
```

bjorng (commented):
Looks good to me, except for a nit pick.

TD5:
Formatting nitpick resolved 👍 Thanks for your suggestion

bjorng:

bjorng:
Thanks for your pull request.

Issue #6242, opened 2022-08-25
Erlang cannot read nor write files with to 2^31 bytes or more on macOS


josevalim:
**Describe the bug**

```erlang
1> file:write_file("ohno", binary:copy(<<$a>>, 2147483648)).
{error,einval}
2> file:write_file("ohno", binary:copy(<<$a>>, 2147483648-1)).
ok
```

where 2147483648 is 2^31.

I have been told this error does not happen on Windows. It has been reproduced on macOS.

Upon further inspection, it also happens with the file descriptor read/write APIs, even when reading small files. In this case, however, the limit is slightly lower:

```erlang
2> {ok, Fd0} = file:open("README.md", [read]).
{ok,<0.89.0>}
3> file:read(Fd0, 2147483648).
{error,einval}
5> {ok, Fd1} = file:open("README.md", [read]).
{ok,<0.93.0>}
6> file:read(Fd1, 2147483648-1).
{error,einval}
7> {ok, Fd2} = file:open("README.md", [read]).
{ok,<0.97.0>}
8> file:read(Fd1, 2147479552-1).
{ok,<<...>>}
```

[Note the Linux manual says](https://man7.org/linux/man-pages/man2/read.2.html):

```
       On Linux, read() (and similar system calls) will transfer at most
       0x7ffff000 (2,147,479,552) bytes, returning the number of bytes
       actually transferred.  (This is true on both 32-bit and 64-bit
       systems.)
```

Although I could only read `2147479552-1`. So perhaps it is generally safer to not try to read more than `2147479552-1` bytes in said systems?

**Expected behavior**

That it can read files with 2^31 bytes or more. :)

**Affected versions**

Checked it on 25 and master as of 23b7b7e29ed003f7a9419432e78111e015822874.

garazdawi:
How odd. We have [test cases](https://github.com/erlang/otp/blob/master/lib/kernel/test/file_SUITE.erl#L4053-L4123) that should cover those scenarios.

mikpe:
I can't reproduce:
```
Erlang/OTP 25 [erts-13.0.4] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Eshell V13.0.4  (abort with ^G)
1> file:write_file("ohno", binary:copy(<<$a>>, 2147483648)).
ok
2> file:write_file("README.md", <<"foo\n">>).
ok
3> {ok, Fd0} = file:open("README.md", [read]).
[...]

PR #6248, opened 2022-08-26
erts: Fix large read/writes on macOS 

garazdawi:
On macOS there is an (as far as I can tell) undocumented limitation
for read and write that the size is not allowed to be greater than
2GB. So if we get an EINVAL from read/write and the size is > 2GB
we try again with a smaller size.

Closes https://github.com/erlang/otp/issues/6242

### DIFF
--- a/erts/emulator/nifs/unix/unix_prim_file.c
+++ b/erts/emulator/nifs/unix/unix_prim_file.c
@@ -42,6 +42,8 @@
 
 #include <utime.h>
 
+#define FALLBACK_RW_LENGTH ((1ull << 31) - 1)
+
 /* Macros for testing file types. */
 #ifdef NO_UMASK
 #define FILE_MODE 0644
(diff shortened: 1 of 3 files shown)


erts/emulator/nifs/unix/unix_prim_file.c
```diff
@@ -284,6 +286,11 @@ Sint64 efile_readv(efile_data_t *d, SysIOVec *iov, int iovlen) {
 
         if(use_fallback) {
             result = read(u->fd, iov->iov_base, iov->iov_len);
+
+            /* Some OSs (e.g. macOS) does not allow reads greater than 4 GB,
```
mikpe, line 290: Here and elsewhere: I think you meant "2 GB", as that is what was observed in the issue and also fits with overflowing a signed 32-bit int.

garazdawi: thanks, fixed!

PR #6249, opened 2022-08-26
ssl: Correct middle_box negotiation

### LINKED ISSUE #6241: OTP-24.3.4.3 `{unexpected_msg,{internal,{encrypted_extensions,#{}}}}`
**Describe the bug**

On OTP-24.3.4.3 I noticed the error below. OTP-24.3.4.2, or other versions, didn't have such behaviour.

**To Reproduce**

```
$ erl
1> ssl:start(), ssl:connect('elixir-lang.org', 443, []).
=WARNING REPORT==== 25-Aug-2022::11:28:04.717049 ===

=NOTICE REPORT==== 25-Aug-2022::11:28:04.811253 ===
TLS client: In state hello_middlebox_assert at ssl_gen_statem.erl:736 generated CLIENT ALERT: Fatal - Unexpected Message
 - {unexpected_msg,{internal,{encrypted_extensions,#{}}}}
{error,{tls_alert,{unexpected_message,"TLS client: In state hello_middlebox_assert at ssl_gen_statem.erl:736 generated CLIENT ALERT: Fatal - Unexpected Message\n {unexpected_msg,{internal,{encrypted_extensions,#{}}}}"}}}
```

**Expected behavior**
No error.

**Affected versions**
24.3.4.3

IngelaAndin:
Closes #6241

### DIFF
--- a/lib/ssl/src/ssl_session.erl
+++ b/lib/ssl/src/ssl_session.erl
@@ -1,8 +1,8 @@
 %%
 %% %CopyrightBegin%
-%% 
+%%
 %% Copyright Ericsson AB 2007-2022. All Rights Reserved.
-%% 
+%%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
 %% You may obtain a copy of the License at
@@ -14,7 +14,7 @@
 %% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 %% See the License for the specific language governing permissions and
 %% limitations under the License.
-%% 
+%%
 %% %CopyrightEnd%
 %%
 
@@ -30,9 +30,13 @@
 -include("ssl_api.hrl").
 
 %% Internal application API
--export([is_new/2, client_select_session/5, server_select_session/5, valid_session/2, legacy_session_id/1]).
+-export([is_new/2,
+         client_select_session/5,
+         server_select_session/5,
+         valid_session/2,
+         legacy_session_id/1]).
 
--type seconds()   :: integer(). 
+-type seconds()   :: integer().
 
 %%--------------------------------------------------------------------
 -spec legacy_session_id() -> ssl:session_id().
@@ -68,8 +72,8 @@ is_new(_ClientSuggestion, _ServerDecision) ->
     true.
 
 %%--------------------------------------------------------------------
--spec client_select_session({ssl:host(), inet:port_number(), map()}, db_handle(), atom(),
-                            #session{}, list()) -> #session{}.
+-spec client_select_session({ssl:host(), inet:port_number(), map()},
+                            db_handle(), atom(), #session{}, list()) -> #session{}.
 %%
 %% Description: Should be called by the client side to get an id
 %%              for the client hello message.
@@ -77,24 +81,27 @@ is_new(_ClientSuggestion, _ServerDecision) ->
 client_select_session({_, _, #{versions := Versions,
                                protocol := Protocol} = Opts} = ClientInfo,
                       Cache, CacheCb, NewSession, CertKeyPairs) ->
-    
+
     RecordCb = record_cb(Protocol),
     LVersion = RecordCb:lowest_protocol_version(Versions),
     HVersion = RecordCb:highest_protocol_version(Versions),
-    
+
     case LVersion of
         {3, 4} ->
-            %% Session reuse is not supported, do pure legacy middlebox comp mode
-            %% negotiation, by providing either empty session id (no middle box)
-            %% or random id (middle box mode).
+            %% Session reuse is not supported, do pure legacy
+            %% middlebox comp mode negotiation, by providing either
+            %% empty session id (no middle box) or random id (middle
+            %% box mode).
             NewSession#session{session_id = legacy_session_id(Opts)};
         _ ->
-            Session =  do_client_select_session(ClientInfo, Cache, CacheCb, NewSession, CertKeyPairs),
-            %% If TLS-1.3 is highest version and there was no previous session
-            %% id that could be reused, if TLS-1.3 is not negotiated,
-            %% possible use random id for middle box mode negotiation.
+            Session =  do_client_select_session(ClientInfo, Cache, CacheCb,
+                                                NewSession, CertKeyPairs),
+            %% If TLS-1.3 is highest version and there was no previous
+            %% session id that could be reused, if TLS-1.3 is not
+            %% negotiated, possibly use random id for middle box mode
+            %% negotiation.
             maybe_handle_middlebox(HVersion, Session, Opts)
-    end.  
+    end.
 
 %%--------------------------------------------------------------------
 -spec server_select_session(ssl_record:ssl_version(), pid(), ssl:session_id(), map(),
@@ -128,8 +135,9 @@ valid_session(#session{time_stamp = TimeStamp}, LifeTime) ->
 %%--------------------------------------------------------------------
 %%% Internal functions
 %%--------------------------------------------------------------------
-do_client_select_session({_, _, #{reuse_session := {SessionId, SessionData}}}, _, _, NewSession, _) when is_binary(SessionId) andalso
-                                                                                                         is_binary(SessionData) ->
+do_client_select_session({_, _, #{reuse_session := {SessionId, SessionData}}},
+                         _, _, NewSession, _) when is_binary(SessionId) andalso
+                                                   is_binary(SessionData) ->
     try binary_to_term(SessionData, [safe]) of
         Session ->
             Session#session{is_resumable = true}
@@ -137,7 +145,8 @@ do_client_select_session({_, _, #{reuse_session := {SessionId, SessionData}}}, _
         _:_ ->
             NewSession#session{session_id = ?EMPTY_ID}
     end;
-do_client_select_session({Host, Port, #{reuse_session := SessionId}}, Cache, CacheCb, NewSession, _) when is_binary(SessionId)->
+do_client_select_session({Host, Port, #{reuse_session := SessionId}},
+                         Cache, CacheCb, NewSession, _) when is_binary(SessionId)->
     case CacheCb:lookup(Cache, {{Host, Port}, SessionId}) of
         undefined ->
 	    NewSession#session{session_id = ?EMPTY_ID};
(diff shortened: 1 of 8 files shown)


lib/ssl/src/ssl_session.erl
```diff
@@ -207,3 +210,11 @@ record_cb(tls) ->
     tls_record;
 record_cb(dtls) ->
     dtls_record.
+
+legacy_session_id() ->
```
KennethL, line 214: Why is this function deleted from its previous place in the beginning of the file and then moved here?
I find it more natural that `legacy_session/0` and `legacy_session/1` are placed together.

lib/ssl/src/ssl_session.erl
```diff
@@ -66,17 +67,19 @@ is_new(_ClientSuggestion, _ServerDecision) ->
...
+                               protocol := Protocol} = Opts} = ClientInfo,
                       Cache, CacheCb, NewSession, CertKeyPairs) ->
     
     RecordCb = record_cb(Protocol),
-    Version = RecordCb:lowest_protocol_version(Versions),
+    LVersion = RecordCb:lowest_protocol_version(Versions),
+    HVersion = RecordCb:highest_protocol_version(Versions),
     
-    case Version of
+    case LVersion of
```
KennethL, line 77: I think we need comment in the code here which explain what the case statement and the logic is supposed to do.
Maybe something like:
If the lowest acceptable protocol version is TLS 1.3 ({3,N} when N>=4 then .....
else if the lowest acceptable protocol version is TLS 1.2 or even lower then ... and if highest acceptable protocol version is TLS 1.3 then we do what?

KennethL (changes requested):

lib/ssl/src/ssl_session.erl
IngelaAndin: Because we keep exported function at the beginning of the file and local functions at the bottom.

lib/ssl/src/ssl_session.erl
IngelaAndin: Added comment!

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -446,13 +445,15 @@ wait_sh(internal, #server_hello{} = Hello,
...
     case tls_handshake_1_3:do_wait_sh(Hello, State0) of
         #alert{} = Alert ->
             ssl_gen_statem:handle_own_alert(Alert, wait_sh, State0);
-         {State1, start, ServerHello} ->
-            %% hello_retry_request: go to start
+        {State1 = #state{session = #session{session_id = Id}}, start, ServerHello} when Id =/= ?EMPTY_ID ->
+            %% hello_retry_request
+            {next_state, hello_retry_middlebox_assert, State1, [{next_event, internal, ServerHello}]};
+        {State1 = #state{session = #session{session_id = Id}}, start, ServerHello} when Id == ?EMPTY_ID ->
             {next_state, hello_retry_middlebox_assert, State1, [{next_event, internal, ServerHello}]};
```
u3s, line 457: just checking: return values on lines 455, 457 are the same - is this intended?

currently 2 patterns from line 453, 456 could be collapsed into single one ... you have a code comment anyway... maybe it could be extended to decribe that no matter what Id value is the action is the same ...

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -488,7 +489,7 @@ wait_ee(enter, _, State0) ->
     State = handle_middlebox(State0),
     {next_state, ?FUNCTION_NAME, State,[]};
 wait_ee(internal = Type, #change_cipher_spec{} = Msg,
-        #state{ssl_options = #{middlebox_comp_mode := true}} = State) ->
+                #state{session = #session{session_id = Id}} = State) when Id =/= ?EMPTY_ID ->
```
u3s, line 492: indentation fault?

lib/ssl/test/tls_1_3_version_SUITE.erl
```diff
@@ -263,3 +277,93 @@ tls12_client_tls_server(Config) when is_list(Config) ->
...
+        <<>> ->
+            ok;
+        _ ->
+            {nok, {{expected, <<>>}, {got, SessionId}}}
+    end;
+check_session_id(Socket, not_empty) ->
+    {ok, [{session_id, SessionId}]} = ssl:connection_information(Socket, [session_id]),
+    case SessionId of
+        <<>> ->
+            {nok, {{expected, no_empty}, {got, SessionId}}};
```
u3s, line 364: typo: no_empty -> not_empty?

lib/ssl/test/tls_1_3_version_SUITE.erl
```diff
@@ -263,3 +277,93 @@ tls12_client_tls_server(Config) when is_list(Config) ->
...
+                   ['tlsv1.3']} | ssl_test_lib:ssl_options(client_cert_opts, Config)],
+    ServerOpts =  [{versions, ['tlsv1.3']} |
+                   ssl_test_lib:ssl_options(server_cert_opts, Config)],
+    middlebox_test(true, not_empty, ClientOpts, ServerOpts, Config),
+    middlebox_test(false, empty, ClientOpts, ServerOpts, Config).
+
+middle_box_tls12_enabled_client() ->
+       [{doc,"Test that a TLS 1.2 enabled client can connect to a TLS 1.3 server with and without middle box compatible mode."}].
+middle_box_tls12_enabled_client(Config) when is_list(Config) ->
+   ClientOpts = [{versions, ['tlsv1.2', 'tlsv1.3']} | ssl_test_lib:ssl_options(client_cert_opts, Config)],
```
u3s, line 295: indentation fault

lib/ssl/test/tls_1_3_version_SUITE.erl
```diff
@@ -263,3 +277,93 @@ tls12_client_tls_server(Config) when is_list(Config) ->
...
+middle_box_tls13_client(Config) when is_list(Config) ->
+    ClientOpts = [{versions,
+                   ['tlsv1.3']} | ssl_test_lib:ssl_options(client_cert_opts, Config)],
+    ServerOpts =  [{versions, ['tlsv1.3']} |
+                   ssl_test_lib:ssl_options(server_cert_opts, Config)],
+    middlebox_test(true, not_empty, ClientOpts, ServerOpts, Config),
+    middlebox_test(false, empty, ClientOpts, ServerOpts, Config).
+
+middle_box_tls12_enabled_client() ->
+       [{doc,"Test that a TLS 1.2 enabled client can connect to a TLS 1.3 server with and without middle box compatible mode."}].
```
u3s, line 293: indentation fault?

lib/ssl/test/tls_1_3_version_SUITE.erl
```diff
@@ -263,3 +277,93 @@ tls12_client_tls_server(Config) when is_list(Config) ->
                    ['tlsv1','tlsv1.1', 'tlsv1.2', 'tlsv1.3']} |
                    ssl_test_lib:ssl_options(server_cert_opts, Config)],
     ssl_test_lib:basic_test(ClientOpts, ServerOpts, Config).
+
+
+middle_box_tls13_client() ->
+       [{doc,"Test that a TLS 1.3 client can connect to a 1.3 server with and without middle box compatible mode."}].
```
u3s, line 283: indentation fault

lib/ssl/src/ssl_session.erl
```diff
@@ -156,7 +182,9 @@ select_session(Sessions, #{ciphers := Ciphers}, CertKeyPairs) ->
                     end,
 		not (resumable(Session#session.is_resumable) andalso
 		     lists:member(Session#session.cipher_suite, Ciphers)
-		     andalso (is_owncert(SessionOwnCert, CertKeyPairs) orelse (SessionOwnCert == undefined)))
+		     andalso (is_owncert(SessionOwnCert, CertKeyPairs)
```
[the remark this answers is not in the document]
IngelaAndin: Humm ... it is now! Fixed.

lib/ssl/src/tls_connection_1_3.erl
IngelaAndin: Good catch, one of those clauses where dead code. I made some clarifications to the code.

lib/ssl/test/tls_1_3_version_SUITE.erl
```diff
@@ -263,3 +277,92 @@ tls12_client_tls_server(Config) when is_list(Config) ->
...
+            {nok, {{expected, <<>>}, {got, SessionId}}}
+    end;
+check_session_id(Socket, not_empty) ->
+    {ok, [{session_id, SessionId}]} = ssl:connection_information(Socket, [session_id]),
+    case SessionId of
+        <<>> ->
+            {nok, {{expected, not_empty}, {got, SessionId}}};
+        _ ->
+            ok
+    end.
```
KennethL, line 366: The two clauses are almost similar, may I propose something like this:
```suggestion
check_session_id(Socket, Expected) -> % Expected is empty | not_empty
   {ok, [{session_id, SessionId}]} = ssl:connection_information(Socket, [session_id]),
    case {Expected, SessionId} of
        {empty, <<>>} ->
            ok;
        {not_empty, Bin} when binary_size(Bin) > 0 -> 
            ok;
        {Exp, _} when Exp == empty; Exp == not_empty ->    
           {nok, {{expected, Exp}, {got, SessionId}}}
    end.
```

KennethL (approved):

lib/ssl/test/tls_1_3_version_SUITE.erl
KennethL: Sorry got it wrong with use of `binary_size` which should be `byte_size`.

IngelaAndin: No worries, I used SessionId =/= <<>>  instead. We do not need to have a type check here we just need to verify that we did not get an empty id!

GitJadhav:
@IngelaAndin We are using Rabbit 3.11.27 and Erlang OTP 25.3.2.7 and still seeing this errors in our RMQ logs 
```
no_suitable_ciphers
no_suitable_ciphers

"TLS server: In state hello at tls_handshake.erl:348 generated SERVER
ALERT: Fatal - Insufficient Security"
 
unexpected_msg,{internal,{change_cipher_spec,<<1>>}}}
```

IngelaAndin:
@GitJadhav I have no reason to believe that OTP-25.3.2.7 is not handling middlebox compatibility mode correctly, we also
will ignore one such message per state  if middlebox mode is not used just to increase interoperability. 

You are not providing enough context for me to help you. I guess you are only asking about the last line in the log. What might help is if you know what client that is connecting when this happens.  If you could connect with this client and set the server log_level to debug that could help.

PR #6251, opened 2022-08-27
Introduce file:flock/2

lukaszsamson:
This PR extends `file` module with new function `flock/2`. It allows locking files for shared or exclusive access on Unix and Windows.

Rationale: Currently the OTP lacks suitable APIs allowing fine grained control over access to filesystem. Those can be especially useful for applications like language tools, compilers (e.g. elixir compiler), IDEs (like language servers) that share access to the same filesystem directories.

Please note that this is a first draft that received only a little testing on macOS platform. I would like to get some guidance if the direction I took is acceptable and if the OTP Team would be interested in merging this. The documentation will also need further improvements.

Related manpages and documentation:
1. https://linux.die.net/man/2/flock
2. https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/flock.2.html
3. http://www.polarhome.com/service/man/?qf=flock&tf=2&of=Solaris&sf=3UCB
4. https://www.freebsd.org/cgi/man.cgi?query=flock&sektion=2
5. https://man.openbsd.org/flock.2
6. https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex
7. https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-unlockfileex
8. https://docs.microsoft.com/en-us/windows/win32/fileio/locking-and-unlocking-byte-ranges-in-files

CC @josevalim

### DIFF
--- a/lib/kernel/doc/src/file.xml
+++ b/lib/kernel/doc/src/file.xml
@@ -267,6 +267,25 @@
       </desc>
     </func>
     <func>
+      <name name="flock" arity="2" since="OTP 26"/>
+      <fsummary>Locks or unlocks a file for shared or exclusive access.</fsummary>
+      <desc>
+        <p>Locks the file referenced by <c><anno>IoDevice</anno></c>. Returns <c>ok</c>
+          if the operation succedes or <c>{error, atom()}</c> otherwise.</p>
+        <p>If the flag <c>unlock</c> is set then the function unlocks the file.</p>
+        <p>On Unix shared locks can be upgraded to exclusive and vice versa.
+          Note however, that it is not atomic.</p>
+        <p>On Unix this function uses <c>flock(2)</c> syscall</p>
+        <p>On Windows one process can hold both shared and exclusive locks
+          over the same file range. This situation needs two unlock operations.</p>
+        <p>On Windows this function uses <c>LockFileEx</c> and <c>UnlokcFileEx</c>
+          win32 API calls</p>
+        <p>Notice that unless option <c>non_blocking</c> was
+          set, <c>flock/2</c> may block if the file is already locked by
+          some other operating system process with an incompatible lock type.</p>
+      </desc>
+    </func>
+    <func>
       <name name="consult" arity="1" since=""/>
       <fsummary>Read Erlang terms from a file.</fsummary>
       <desc>
--- a/lib/kernel/src/file.erl
+++ b/lib/kernel/src/file.erl
@@ -39,7 +39,7 @@
 %% Specialized
 -export([ipread_s32bu_p32bu/3]).
 %% Generic file contents.
--export([open/2, close/1, advise/4, allocate/3,
+-export([open/2, close/1, flock/2, advise/4, allocate/3,
 	 read/2, write/2, 
 	 pread/2, pread/3, pwrite/2, pwrite/3,
 	 read_line/1,
@@ -560,6 +560,24 @@ close(File) when is_pid(File) ->
 close(#file_descriptor{module = Module} = Handle) ->
     Module:close(Handle);
 close(_) ->
+    {error, badarg}.
+
+-spec flock(IoDevice, Flags) -> ok | {error, Reason} when
+      IoDevice :: io_device(),
+      Flags :: list(Flag),
+      Flag :: shared | exclusive | non_blocking | unlock,
+      Reason :: posix() | badarg | terminated.
+
+flock(File, Flags) when is_pid(File) ->
+    case file_request(File, {flock, Flags}) of
+	{error, terminated} ->
+	    ok;
+	Other ->
+	    Other
+    end;
+flock(#file_descriptor{module = Module} = Handle, Flags) ->
+    Module:flock(Handle, Flags);
+flock(_, _) ->
     {error, badarg}.
 
 -spec advise(IoDevice, Offset, Length, Advise) -> ok | {error, Reason} when
--- a/erts/emulator/nifs/common/prim_file_nif.c
+++ b/erts/emulator/nifs/common/prim_file_nif.c
@@ -59,6 +59,12 @@ static ERL_NIF_TERM am_skip_type_check;
 static ERL_NIF_TERM am_read_write;
 static ERL_NIF_TERM am_none;
 
+/* Lock modes */
+static ERL_NIF_TERM am_shared;
+// static ERL_NIF_TERM am_exclusive;
+static ERL_NIF_TERM am_non_blocking;
+static ERL_NIF_TERM am_unlock;
+
 /* enum efile_advise_t */
 static ERL_NIF_TERM am_normal;
 static ERL_NIF_TERM am_random;
@@ -101,6 +107,8 @@ static ERL_NIF_TERM read_file_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM a
 
 static ERL_NIF_TERM open_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[]);
 static ERL_NIF_TERM close_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[]);
+
+static ERL_NIF_TERM flock_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[]);
 
 static ERL_NIF_TERM file_desc_to_ref_nif(ErlNifEnv *env, int argc, const ERL_NIF_TERM argv[]);
 
@@ -158,6 +166,7 @@ static ERL_NIF_TERM file_handle_wrapper(file_op_impl_t operation, ErlNifEnv *env
         return file_handle_wrapper( name ## _impl , env, argc, argv); \
     }
 
+WRAP_FILE_HANDLE_EXPORT(flock_nif)
 WRAP_FILE_HANDLE_EXPORT(read_nif)
 WRAP_FILE_HANDLE_EXPORT(write_nif)
 WRAP_FILE_HANDLE_EXPORT(pread_nif)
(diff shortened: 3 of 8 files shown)


lib/kernel/doc/src/file.xml
```diff
@@ -266,6 +266,25 @@
           <seemfa marker="#open/2"><c>open/2</c></seemfa>.</p>
       </desc>
     </func>
+    <func>
+      <name name="flock" arity="2" since="OTP 26"/>
+      <fsummary>Locks or unlocks a file for shared or exclusive access.</fsummary>
+      <desc>
+        <p>Locks the file referenced by <c><anno>IoDevice</anno></c>. Returns <c>ok</c>
+          if the operation succedes or <c>{error, atom()}</c> otherwise.</p>
```
josevalim, line 274:
```suggestion
          if the operation succeeds or <c>{error, atom()}</c> otherwise.</p>
```

lib/kernel/doc/src/file.xml
```diff
@@ -266,6 +266,25 @@
...
+        <p>If the flag <c>unlock</c> is set then the function unlocks the file.</p>
+        <p>On Unix shared locks can be upgraded to exclusive and vice versa.
+          Note however, that it is not atomic.</p>
+        <p>On Unix this function uses <c>flock(2)</c> syscall</p>
```
josevalim, line 278:
```suggestion
        <p>On Unix this function uses <c>flock(2)</c> syscall.</p>
```

lib/kernel/doc/src/file.xml
```diff
@@ -266,6 +266,25 @@
...
+        <p>On Windows one process can hold both shared and exclusive locks
+          over the same file range. This situation needs two unlock operations.</p>
+        <p>On Windows this function uses <c>LockFileEx</c> and <c>UnlokcFileEx</c>
```
josevalim, line 281:
```suggestion
        <p>On Windows this function uses <c>LockFileEx</c> and <c>UnlockFileEx</c>
```

lib/kernel/src/file.erl
```diff
@@ -562,6 +562,24 @@ close(#file_descriptor{module = Module} = Handle) ->
...
+      Flag :: shared | exclusive | non_blocking | unlock,
+      Reason :: posix() | badarg | terminated.
+
+flock(File, Flags) when is_pid(File) ->
+    case file_request(File, {flock, Flags}) of
+	{error, terminated} ->
+	    ok;
+	Other ->
+	    Other
+    end;
```
josevalim, line 577: IIRC the emacs indentation is only for old code. :)

```suggestion
    case file_request(File, {flock, Flags}) of
        {error, terminated} ->
            ok;
        Other ->
            Other
    end;
```

josevalim:
Awesome @lukaszsamson! This would indeed be useful for Elixir!

Some additional comments:

* I believe you are missing adding a `flock(_Device, _Mode) -> {error, enotsup}.` in `ram_file.erl`
* Are we missing the `flock_nif/2` stub and `flock/2` itself in `prim_file.erl`?

If the OTP team accepts this feature, there are also the following questions:

* Should we call it `file:lock/2` instead of `file:flock/2`?
* Should the unlock flag be its own function instead (`file:unlock/2`)?
* Should we default to "exclusive" mode and then only provide shared as an option?

I would answer yes to all of the above but it is up to the OTP team, so I would wait for feedback before further changes.

garazdawi:
Hello!

We will need to have an OTB (OTP Technical Board) meeting to talk about if and how we want to have this, but since it can be implemented on all supported platforms I don't see any reason why we would not want it.

> * Should we call it `file:lock/2` instead of `file:flock/2`?
> * Should the unlock flag be its own function instead (`file:unlock/2`)?
> * Should we default to "exclusive" mode and then only provide shared as an option?
>
> I would answer yes to all of the above but it is up to the OTP team, so I would wait for feedback before further changes.

I would also say yes to all of the above.

lukaszsamson:
Here's a build from a branch with updated preloaded (a check in kernel is failing as my machine differs from the one running GHA) https://github.com/lukaszsamson/otp/runs/8089001687?check_suite_focus=true

josevalim:
I believe `flock` is more than enough for Mix/rebar3 use cases, so I am happy to go with a smaller surface area.

lukaszsamson:
> Should the unlock flag be its own function instead (`file:unlock/2`)?

Then maybe `file:unlock/1`. On unix it would be trivial to implement with `flock` syscall. The windows implementation would need to unlock both shared and exclusive locks but it would make the behaviour on this platform closer to unix one.
And regarding windows, maybe we shouldn't let shared and exclusive locks over the same file and just unlock one when `file:lock/2` is called again with other type.

garazdawi:
I started looking at this PR as we will soon have a meeting to discuss it. I'm putting some notes here in case you have any opinions.

It would seem like Windows has decided to make the default `shared`. From what I can tell python+ruby+java follow the unix default (i.e. you have to specify what you want), while rust has created different functions (i.e. `lock_exclusive` and `lock_shared`). So it would seem like most/all other languages have chosen to make it mandatory to specify which mode you want. Maybe that is a reason for us to do the same...

I don't think we should support locking byte ranges. If we need that we can add it in another PR.

It is too bad that there is no asynchronous API that we can use. Blocking a `dirty_io_scheduler` while waiting for a lock to be released is not ideal, and neither is polling for the lock to be released. Though there is not much we can do about it.

garazdawi:
Hello!

We have (finally) made a decision and we would like this functionality. We propose that the API should be:

```erlang
 -spec file:lock(file:io_device(), [shared | exclusive | non_blocking]) ->
  ok | {error, already_locked | file:posix()}.

-spec file:unlock(file:io_device()) -> ok | {error, terminated}.
```

I think that is the same as what has been discussed in this PR?

lukaszsamson:
Nice. We need to look into some details to make the API consistent
1. On unix subsequent calls to flock succeed even if file is already locked. I'm not sure about windows
2. On windows it's possible to lock the same file twice (shared and exclusive). Each lock needs an unlock with matching type.
3. I'm not sure what's the effect of unlocking an unlocked file on windows. On unix it succeedes without error (if the file descriptor is valid)

garazdawi:
I did some testing on Windows.

> 1. On unix subsequent calls to flock succeed even if file is already locked. I'm not sure about windows.

Windows will fail with `ERROR_LOCK_VIOLATION` if file is already locked using the same HANDLE. I would suggest that we keep some metadata about whether the file is locked or not and with which lock type so that we can just return success if something is already locked.

> 2. On windows it's possible to lock the same file twice (shared and exclusive). Each lock needs an unlock with matching type.

On Windows you cannot upgrade a lock, but you can downgrade. That is you can do:

```c
FileLockEx(handle, EXCLUSIVE);
FileLockEx(handle, SHARED);
FileUnlockEx(handle, EXCLUSIVE);
```

but you cannot do:

```c
FileLockEx(handle, SHARED);
FileLockEx(handle, EXCLUSIVE); // this will block or return ERROR_LOCK_VIOLATION
FileUnlockEx(handle, SHARED);
```

Also, you can do as many shared locks are you like, but you need to do the same amount of unlocks for all the locks to be released. That is:

```c
FileLockEx(handle, SHARED);
FileLockEx(handle, SHARED);
FileLockEx(handle, SHARED);
FileLockEx(handle, SHARED);
FileUnlockEx(handle, SHARED);
FileLockEx(handle, EXCLUSIVE|NB); // ERROR_LOCK_VIOLATION
FileUnlockEx(handle, SHARED);
FileLockEx(handle, EXCLUSIVE|NB); // ERROR_LOCK_VIOLATION
FileUnlockEx(handle, SHARED);
FileLockEx(handle, EXCLUSIVE|NB); // ERROR_LOCK_VIOLATION
FileUnlockEx(handle, SHARED);
FileLockEx(handle, EXCLUSIVE|NB); // Success
```

Here again, I think it would be good to keep a flag in the file structure so that any extra locks are just ignored so that we behave as on Unix.

> 3. I'm not sure what's the effect of unlocking an unlocked file on windows. On unix it succeedes without error (if the file descriptor is valid)

You get `ERROR_NOT_LOCKED` on Windows.

garazdawi:
Also locks on Windows are not only advisory, so if you do `FileLockEx(handle, EXCLUSIVE)` you cannot do `file:read_file(File)` on that file from the same or another process.

lukaszsamson:
@garazdawi I applied your suggestions and split `flock/2` into `lock/2` and `unlock/1`. I was able to test POSIX implementation on macos, will try to set up windows machine later.

I'm not sure about `already_locked` error. Do we need it? If we'd like to ignore extra locks on windows than probably not.

> On Windows you cannot upgrade a lock, but you can downgrade

Do you think we should make it closer to POSIX, i.e. add support for nonatomic lock upgrade and make downgrade nonatomic?

lukaszsamson:
I made changes to Windows implementation to match unix. If a lock of a different type is already held then unlock is called before attempting to lock.

jhogberg:
We had another technical board meeting today to try to reconcile the differences between Unix and Windows, and have realized that the functionality differs too much between operating systems to pretend that it's the same function. There's no good way to compensate for this in Erlang. :-(

We think it's time for us to refactor the file system support in Erlang by dividing it into several modules, one for generic and actually platform independent operations, and other modules for platform specific ones such as Windows, POSIX, BSD, Linux (which I suppose would cover those not in POSIX).

This is something we've been thinking about for a long time but haven't had enough motivation for until now. We plan to get this done early in OTP 27 and we'll bring this PR in with it.

lukaszsamson:
Any updates @jhogberg @garazdawi? Is OTP27 still the target?

PR #6256, opened 2022-08-29
Improve interval timer handling

Maria-12648430:
This PR was inspired and is competing (to a degree) with #6145.

With the current (old) implementation, the apply messages resultingh from interval timers were handled by the timer server process itself, which may lead to its congestion. Specifically, it has been noticed that, with many short-interval timers running, starting as cancelling of timers takes longer and longer, stalling the calling process.

The changes in this PR run each interval timer in a separate process, so their apply messages don't compete with other (start/cancel/etc) messages. This reduces the load of the timer server, and at the same time reduces the tendency of interval timers to lag.

### DIFF
--- a/lib/stdlib/doc/src/timer.xml
+++ b/lib/stdlib/doc/src/timer.xml
@@ -122,7 +122,15 @@
           finish before starting the next.</p>
         <p>If the execution time of the spawned process is greater than the
           given <c><anno>Time</anno></c>, the next process is spawned immediately
-          after the one currently running has finished.</p>
+          after the one currently running has finished. Assuming that execution
+          times of the spawned processes performing the applies on average
+          are smaller than <c><anno>Time</anno></c>, the amount of applies
+          made over a large amount of time will be the same even if some
+          individual execution times are larger than
+          <c><anno>Time</anno></c>. The system will try to catch up as soon
+          as possible. For example, if one apply takes
+          <c>2.5*<anno>Time</anno></c>, the following two applies will be
+          made immediately one after the other in sequence.</p>
         <p>Returns <c>{ok, <anno>TRef</anno>}</c> or
           <c>{error, <anno>Reason</anno>}</c>.</p>
       </desc>
--- a/lib/stdlib/doc/src/timer.xml
+++ b/lib/stdlib/doc/src/timer.xml
@@ -71,10 +71,10 @@
   <funcs>
     <func>
       <name name="apply_after" arity="4" since=""/>
-      <fsummary>Apply <c>Module:Function(Arguments)</c> after a specified
-        <c>Time</c>.</fsummary>
+      <fsummary>Spawn a process evaluating <c>Module:Function(Arguments)</c>
+        after a specified <c>Time</c>.</fsummary>
       <desc>
-        <p>Evaluates <c>apply(<anno>Module</anno>, <anno>Function</anno>,
+        <p>Evaluates <c>spawn(<anno>Module</anno>, <anno>Function</anno>,
           <anno>Arguments</anno>)</c> after <c><anno>Time</anno></c>
           milliseconds.</p>
         <p>Returns <c>{ok, <anno>TRef</anno>}</c> or
@@ -84,12 +84,45 @@
 
     <func>
       <name name="apply_interval" arity="4" since=""/>
-      <fsummary>Evaluate <c>Module:Function(Arguments)</c> repeatedly at
-        intervals of <c>Time</c>.</fsummary>
+      <fsummary>Spawn a process evaluating <c>Module:Function(Arguments)</c>
+        repeatedly at intervals of <c>Time</c>.</fsummary>
       <desc>
-        <p>Evaluates <c>apply(<anno>Module</anno>, <anno>Function</anno>,
+        <p>Evaluates <c>spawn(<anno>Module</anno>, <anno>Function</anno>,
           <anno>Arguments</anno>)</c> repeatedly at intervals of
-          <c><anno>Time</anno></c>.</p>
+          <c><anno>Time</anno></c>, irrespective of whether a previously
+          spawned process has finished or not.</p>
+        <warning>
+          <p>If the execution time of the spawned process is, on average,
+            greater than the given <c><anno>Time</anno></c>, multiple such
+            processes will run at the same time. With long execution times,
+            short intervals, and many interval timers running, this may even
+            lead to exceeding the number of allowed processes. As an extreme
+            example, consider
+            <c>[timer:apply_interval(1, timer, sleep, [1000]) || _ &lt;- lists:seq(1, 1000)]</c>,
+            that is, 1,000 interval timers executing a process that takes 1s
+            to complete, started in intervals of 1ms, which would result in
+            1,000,000 processes running at the same time, far more than a node
+            started with default settings allows (see the
+            <seeguide marker="system/efficiency_guide:advanced#system-limits">System
+            Limits section in the Effiency Guide</seeguide>).</p>
+        </warning>
+        <p>Returns <c>{ok, <anno>TRef</anno>}</c> or
+          <c>{error, <anno>Reason</anno>}</c>.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="apply_repeatedly" arity="4" since="OTP @OTP-18236@"/>
+      <fsummary>Spawn a process evaluating <c>Module:Function(Arguments)</c>
+        repeatedly at intervals of <c>Time</c>.</fsummary>
+      <desc>
+        <p>Evaluates <c>spawn(<anno>Module</anno>, <anno>Function</anno>,
+          <anno>Arguments</anno>)</c> repeatedly at intervals of
+          <c><anno>Time</anno></c>, waiting for the spawned process to
+          finish before starting the next.</p>
+        <p>If the execution time of the spawned process is greater than the
+          given <c><anno>Time</anno></c>, the next process is spawned immediately
+          after the one currently running has finished.</p>
         <p>Returns <c>{ok, <anno>TRef</anno>}</c> or
           <c>{error, <anno>Reason</anno>}</c>.</p>
       </desc>
--- a/lib/stdlib/src/timer.erl
+++ b/lib/stdlib/src/timer.erl
@@ -22,7 +22,8 @@
 -export([apply_after/4,
          send_after/3, send_after/2,
          exit_after/3, exit_after/2, kill_after/2, kill_after/1,
-         apply_interval/4, send_interval/3, send_interval/2,
+         apply_interval/4, apply_repeatedly/4,
+         send_interval/3, send_interval/2,
          cancel/1, sleep/1, tc/1, tc/2, tc/3, now_diff/2,
          seconds/1, minutes/1, hours/1, hms/3]).
 
@@ -61,7 +62,7 @@
                    Reason :: term().
 apply_after(0, M, F, A)
   when ?valid_mfa(M, F, A) ->
-    do_apply({M, F, A}),
+    _ = do_apply({M, F, A}, false),
     {ok, {instant, make_ref()}};
 apply_after(Time, M, F, A)
   when ?valid_time(Time),
@@ -158,6 +159,21 @@ apply_interval(Time, M, F, A)
        ?valid_mfa(M, F, A) ->
     req(apply_interval, {system_time(), Time, self(), {M, F, A}});
 apply_interval(_Time, _M, _F, _A) ->
+    {error, badarg}.
+
+-spec apply_repeatedly(Time, Module, Function, Arguments) ->
+          {'ok', TRef} | {'error', Reason}
+              when Time :: time(),
+                   Module :: module(),
+                   Function :: atom(),
+                   Arguments :: [term()],
+                   TRef :: tref(),
+                   Reason :: term().
+apply_repeatedly(Time, M, F, A)
+  when ?valid_time(Time),
+       ?valid_mfa(M, F, A) ->
+    req(apply_repeatedly, {system_time(), Time, self(), {M, F, A}});
+apply_repeatedly(_Time, _M, _F, _A) ->
     {error, badarg}.
 
 -spec send_interval(Time, Destination, Message) -> {'ok', TRef} | {'error', Reason}
@@ -399,15 +415,11 @@ handle_call({apply_once, {Started, Time, MFA}}, _From, Tab) ->
     {reply, Reply, Tab};
 %% Start an interval timer.
 handle_call({apply_interval, {Started, Time, Pid, MFA}}, _From, Tab) ->
-    Tag = make_ref(),
-    TimeServerPid = self(),
-    {TPid, TRef} = spawn_monitor(fun() ->
-        TimeServerRef = monitor(process, TimeServerPid),
-        TargetRef = monitor(process, Pid),
-        NextTimeout = Started + Time,
-        TimerRef = erlang:start_timer(NextTimeout, self(), {apply_interval, NextTimeout, Time, MFA}, [{abs, true}]),
-        _ = interval_loop(TimeServerRef, TargetRef, Tag, TimerRef)
-    end),
+    {TRef, TPid, Tag} = start_interval_loop(Started, Time, Pid, MFA, false),
+    ets:insert(Tab, {TRef, TPid, Tag}),
+    {reply, {ok, {interval, TRef}}, Tab};
+handle_call({apply_repeatedly, {Started, Time, Pid, MFA}}, _From, Tab) ->
+    {TRef, TPid, Tag} = start_interval_loop(Started, Time, Pid, MFA, true),
     ets:insert(Tab, {TRef, TPid, Tag}),
     {reply, {ok, {interval, TRef}}, Tab};
 %% Cancel a one-shot timer.
@@ -432,9 +444,9 @@ handle_call(_Req, _From, Tab) ->
               when Tab :: ets:tid().
 %% One-shot timer timeout.
 handle_info({timeout, TRef, {apply_once, MFA}}, Tab) ->
-    case ets:take(Tab, TRef) of
+    _ = case ets:take(Tab, TRef) of
         [{TRef}] ->
-            do_apply(MFA);
+            do_apply(MFA, false);
         [] ->
             ok
     end,
@@ -466,8 +478,8 @@ terminate(Reason, Tab) ->
                 TPid ! {cancel, Tag},
                 Acc
         end,
-	undefined,
-	Tab),
+        undefined,
+        Tab),
     true = ets:delete(Tab),
     terminate(Reason, undefined).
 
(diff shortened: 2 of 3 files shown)


rickard-green:
The only potential concern that I can see with this approach is that you get a lot of processes if you set up a lot of interval timers. This should however be much less of a concern today, than when the original timer server was implemented.

Maria-12648430:
Yes, that is true. But I don't think this will become a practical problem, at least not anytime soon. If a system uses so many interval timers that the number of extra processes becomes a problem, I believe that system will also experience other problems, probably earlier.

juhlig:
Hm, I just ran into a problem :sweat_smile: Looks like the interval timers run too fast now. That is related to a problem with interval timers that @Maria-12648430 hinted at in other places already.

```
[timer:apply_interval(1, timer, sleep, [1000]) || _ <- lists:seq(1, 10000)].
```
This leads to a `system_limit` exception with reason "too many processes". Not because of the extra processes the interval timers use, but because each action is itself being spawned in a fire-and-forget manner, that is, the timer does not wait for the action to complete but just spawns again as soon as the timer triggers again. In the old implementation, there was a sort of natural slowdown here in that the actions would arrive in the time servers message queue and the spawn would happen slower, but with the interval timers running in separate threads, the spawns happen much faster.

So, I would suggest a different approach: Leave interval timers as they are (for backwards compatibility), but make up a new kind of timer and associated functions, like `apply_repeatedly`. This kind of timer would be implemented like the interval timers in this PR, but wait for the action to complete before spawning a new action. When such a timer triggers, it would calculate the time for the next event from the scheduled time of the current event and the interval, then spawn and wait for the action to complete, then if the next scheduled time has already passed fire again immediately, otherwise set a timer to fire at the calculated next event.

Maria-12648430:
@rickard-green what do you think?

rickard-green:
My initial thought is that if anyone do that, they intentionally want to shoot them selves in the foot 😄 I'll think some more about it, and ask around what people think here at OTP, though.

rickard-green:
BTW if I do the above using OTP 25.0 on my machine I get tons of `Too many processes` error reports as well.

Maria-12648430:
Yes, it is a pretty extreme example ;) But it illustrates the point... With actions that take 1s to run, startet naively at every millisecond, that amounts ~1,000 actions (ie, processes) run at the same time (apprx. `$action_runtime / $timer_interval`). With 10,000 such timers, that would make it ~10,000,000 processes running at the same time, way more than the system can handle in the default setup.

Unlikely as something as the given scenario is in itself... Systems tend to grow (both in features and number of users) and change. What I'm thinking of is, like, somebody puts some new code in an action that was quick enough to complete, but with the new code drags it out; that the action itself is slower may not be a problem to the implementor/users, but it may suddenly and unexpectedly become a problem for the system; or maybe the action talks to an external system which may be temporarily slow, so the behavior of the external system may suddenly (and non-obviosuly) cause problems and crash the Erlang system. And what may work well with 10 users in a dev system may cause all hell to break loose when released to production with 10,000 or 10,000,000 users. Things like that happen, I hear :grin:

So, what I'm trying to say is that @juhlig's approach of offering a safer alternative to `apply_interval` is IMO not a bad idea. I would, however, keep `apply_interval` as it currently is in this PR, that is, in a separate process and spawning without second thoughts, but add a warning and link to `apply_repeatedly` to the docs.
The current (before this PR) problem with `apply_interval` delivering events to the timer server itself is that it may clog the message queue when many short-interval timers are running. `apply_*`/`send_*` and `cancel` messages have to wait until all events that happened before them are processed. As those messages are delivered synchronously to the timer server via `gen_call`, the process that wants to start/cancel a timer gets blocked. In an extreme scenario, it may be that the time for which an event is scheduled has already passed when the timer start message is received and processed by the timer server.

Maria-12648430:
Ok, the last commit adds `apply_repeatedly/4`, which is like `apply_interval/4` but waits for the action to complete. The example [given above](https://github.com/erlang/otp/pull/6256#issuecomment-1231222720) by @juhlig works with `apply_repeatedly` without touching the process limit. `apply_interval` has been left as it is now in this PR and causes `Too many processes` errors. Both (in the case of `apply_interval`, with _sane_ timers of course) do not clog the timer server regarding starting and cancelling of timers.

No tests and docs yet, let's hear what @rickard-green thinks first :wink:

rickard-green:
Yes, keeping `apply_interval` as in this PR and add an `apply_repeatedly` which won't `apply` until the previous `apply` has completed sounds good.

I don't see it as a problem that `apply_interval` is better at timely delivering timeouts than before and you cannot rely on the old behavior as some kind of "flow control". If you create enough `apply_interval` timers, set a small enough interval, and your computer is fast enough *it will* exceed maximum amount of concurrent processes also with the old implementation. I also do not see it as a problem with one extra process for each interval timer.

When it comes to `apply_repeatedly` one could (at least) have two different approaches for when to trigger next timeout:
1) Trigger X milliseconds after the previous timeout was triggered
2) Trigger X milliseconds after the last `apply` completed

I guess 1. probably is what one wants, but in case the `apply` takes longer time than the interval the `apply` calls will be performed back to back all the time. I do however not consider that a problem as long as it is obvious in the documentation how it works.

lib/stdlib/src/timer.erl
```diff
@@ -480,34 +467,121 @@ handle_cast(_Req, Tab) ->
...
+    of _ -> {ok, exit}
+    catch _:_ -> error
+    end;
+do_apply({M,F,A}, false) ->
+    try spawn(M, F, A)
+    of _ -> {ok, spawn}
+    catch _:_ -> error
+    end;
+do_apply({M, F, A}, true) ->
+    try spawn_monitor(M, F, A)
```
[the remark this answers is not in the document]
rickard-green: Yes, good point about unintentionally stealing messages from the timer process. That is a problem that `rpc` doesn't have.

Maria-12648430:
The last commits adds tests and documentation for `apply_repeatedly/4`.

juhlig:
Btw @rickard-green, before (and if) this gets merged, we need something to put in the `since` attribute in the docs for `apply_repeatedly`.

lpgauth:
Why not have a pool of timer servers instead? Having a server per timer seems like a big overkill...

Maria-12648430:
@lpgauth it is a (small, leightweight) _process_ per _interval_ timer, not full-fledged timer _server_ for _every kind of_ timer. That would be overkill, yes, but having just some small processes is not a big deal (IMO).

That said, if you were doing it with a pool, you would have to answer the question what would be a good size for it. What is overkill for some applications migt be too few for others. And if you went as far as auto-scaling, that would blow complexity way through the roof, and there might also be scenarios where more work is going on scaling back and forth than actual timing.

On the other hand, using small dedicated processes like suggested in this PR is a rather simple and solid approach which puts no or only very little extra load on a system.

rickard-green:
@juhlig 
> Btw @rickard-green, before (and if) this gets merged, we need something to put in the since attribute in the docs for apply_repeatedly.

You can put "OTP @OTP-18236@" in the since tag

Maria-12648430:
@rickard-green

> You can put "OTP @OTP-18236@" in the since tag

Is there a way to find that out without asking? Because I'm _bound_ to forget, already did in the past, and also this time :sweat_smile:

Anyway. I added the given info to the tag, and also augmented the commit message with more details. I kept the interval timer improvement and the addition of `apply_repeatedly/4` in separate commits (vs squashing them into one) because they _are_ seperate things ;)

lib/stdlib/src/timer.erl
```diff
@@ -480,34 +467,121 @@ handle_cast(_Req, Tab) ->
...
     case ets:take(Tab, TRef) of
-        [{TRef, SRef}] ->
-            ok = erlang:cancel_timer(SRef, [{async, true}, {info, false}]),
+        [{TRef}] -> % One-shot timer.
+            ok = cancel_timer(TRef),
+            true;
+        [{TRef, TPid, Tag}] -> % Interval timer.
+            Mon = monitor(process, TPid),
+            TPid ! {cancel, Tag},
+            receive {'DOWN', Mon, process, _, _} -> ok end,
```
juhlig, line 551: The monitoring and waiting for the `'DOWN'` message is unnecessary IMO. It accomplishes nothing but slowing down the timer server for a little bit.

Maria-12648430: You do have a point... :sweat_smile:

Maria-12648430: Done

Maria-12648430:
Ok... @rickard-green, I guess that at least from our side, this is finished :slightly_smiling_face:

lib/stdlib/doc/src/timer.xml
```diff
@@ -84,12 +84,45 @@
...
+      <fsummary>Spawn a process evaluating <c>Module:Function(Arguments)</c>
+        repeatedly at intervals of <c>Time</c>.</fsummary>
+      <desc>
+        <p>Evaluates <c>spawn(<anno>Module</anno>, <anno>Function</anno>,
+          <anno>Arguments</anno>)</c> repeatedly at intervals of
+          <c><anno>Time</anno></c>, irrespective of whether a previosuly
+          spawned process has finished or not.</p>
+        <warning>
+          <p>If the execution time of the spawned process is, on average,
+            greater than the given <c><anno>Time</anno></c>, muliple such
```
bjorng, line 96:
```suggestion
            greater than the given <c><anno>Time</anno></c>, multiple such
```

lib/stdlib/src/timer.erl
```diff
@@ -399,37 +415,25 @@ handle_call({apply_once, {Started, Time, MFA}}, _From, Tab) ->
...
-            true ->
-                demonitor(TRef, [flush]);
-            false ->
-                ok
-        end,
+        true ->
+            demonitor(TRef, [flush]);
+        false ->
+            ok
+    end,
```
u3s, line 436: indentation seems spoiled?

lib/stdlib/src/timer.erl
```diff
@@ -440,31 +444,14 @@ handle_call(_Req, _From, Tab) ->
...
 handle_info({timeout, TRef, {apply_once, MFA}}, Tab) ->
-    case ets:take(Tab, TRef) of
-        [{TRef, _SRef}] ->
-            do_apply(MFA);
+    _ = case ets:take(Tab, TRef) of
+        [{TRef}] ->
+            do_apply(MFA, false);
         [] ->
             ok
     end,
```
u3s, line 452: indentation

lib/stdlib/src/timer.erl
```diff
@@ -480,34 +467,119 @@ handle_cast(_Req, Tab) ->
...
+    _ = ets:foldl(fun
+            ({TRef}, Acc) ->
+                _ = cancel_timer(TRef),
+                Acc;
+            ({_TRef, TPid, Tag}, Acc) ->
+                TPid ! {cancel, Tag},
+                Acc
+        end,
+        undefined,
+        Tab),
```
u3s, line 482: indentation

u3s (commented):
I've marked some places where I thought indentation is not as normal in Emacs - fix if you think it would add some value for readers.

lib/stdlib/doc/src/timer.xml
Maria-12648430: Keen eyes ;) Corrected.

lib/stdlib/doc/src/timer.xml
```diff
@@ -84,12 +84,45 @@
...
     <func>
       <name name="apply_interval" arity="4" since=""/>
-      <fsummary>Evaluate <c>Module:Function(Arguments)</c> repeatedly at
-        intervals of <c>Time</c>.</fsummary>
+      <fsummary>Spawn a process evaluating <c>Module:Function(Arguments)</c>
+        repeatedly at intervals of <c>Time</c>.</fsummary>
+      <desc>
+        <p>Evaluates <c>spawn(<anno>Module</anno>, <anno>Function</anno>,
+          <anno>Arguments</anno>)</c> repeatedly at intervals of
+          <c><anno>Time</anno></c>, irrespective of whether a previosuly
```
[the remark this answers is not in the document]
Maria-12648430: Thanks, corrected :)

lib/stdlib/src/timer.erl
Maria-12648430: Hm... seems ok to me, what's wrong with it? :thinking::sweat_smile: (I'm using vim, not Emacs)

lib/stdlib/src/timer.erl
Maria-12648430: Dto?

lib/stdlib/src/timer.erl
Maria-12648430: Dto?

lib/stdlib/src/timer.erl
```diff
@@ -480,34 +467,119 @@ handle_cast(_Req, Tab) ->
...
 
+start_interval_loop(Started, Time, TargetPid, MFA, WaitComplete) ->
+    Tag = make_ref(),
+    TimeServerPid = self(),
+    {TPid, TRef} = spawn_monitor(fun() ->
+        TimeServerRef = monitor(process, TimeServerPid),
+        TargetRef = monitor(process, TargetPid),
+        TimerRef = schedule_interval_timer(Started, Time, MFA),
+        _ = interval_loop(TimeServerRef, TargetRef, Tag, WaitComplete, TimerRef)
+    end),
```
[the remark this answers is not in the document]
Maria-12648430: Dto?

lib/stdlib/src/timer.erl
u3s: In code you've changed you got patterns (`true ->`, `false ->`) intended in relation to `case`. it is different in you variant.
I don't think its related to Editor (although Emacs seems to do it the way it was by default).

proposal use if you like it
```
handle_call({cancel, {interval, TRef}}, _From, Tab) ->
    _ = case remove_timer(TRef, Tab) of
            true ->
                demonitor(TRef, [flush]);
            false ->
                ok
        end,
    {reply, {ok, cancel}, Tab};
```

lib/stdlib/src/timer.erl
u3s: what do you mean with DTO? (I did google query and still not sure) :-)

a proposal use if you like it

```
handle_info({timeout, TRef, {apply_once, MFA}}, Tab) ->
    _ = case ets:take(Tab, TRef) of
            [{TRef}] ->
                do_apply(MFA, false);                                                                                                                                                                                                                                                                       [] ->
                ok
        end,                                                                                                                                                                                                                                                                                        {noreply, Tab};

```

lib/stdlib/src/timer.erl
u3s: a proposal use if you like it

```
terminate(Reason, Tab) ->
    _ = ets:foldl(fun
                      ({TRef}, Acc) ->
                         _ = cancel_timer(TRef),
                         Acc;
                      ({_TRef, TPid, Tag}, Acc) ->
                         TPid ! {cancel, Tag},
                         Acc
                  end,
                  undefined,
                  Tab),
    true = ets:delete(Tab),
    terminate(Reason, undefined).
```

lib/stdlib/src/timer.erl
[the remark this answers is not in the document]
u3s: a proposal use if you like it

```
start_interval_loop(Started, Time, TargetPid, MFA, WaitComplete) ->
    Tag = make_ref(),
    TimeServerPid = self(),
    {TPid, TRef} =
        spawn_monitor(fun() ->
                              TimeServerRef = monitor(process, TimeServerPid),
                              TargetRef = monitor(process, TargetPid),
                              TimerRef =
                                  schedule_interval_timer(Started, Time, MFA),
                              _ = interval_loop(TimeServerRef, TargetRef,
                                                Tag, WaitComplete, TimerRef)
                      end),
    {TRef, TPid, Tag}.
```

lib/stdlib/src/timer.erl
Maria-12648430: Ah, you mean to align on `case`, got it. I grepped through the stdlib, and in the modules there the alignment of the clauses and bodies is as you suggest. The alignment of the `end` differs sometimes, but is mostly as you suggest, too. So, yeah, I'll fix it here to fit in :wink:

lib/stdlib/src/timer.erl
Maria-12648430: Dto --> dito, meaning something like "likewise" or "the same (as above)". I didn't want to repeat "Seems ok to me, what's wrong with it?" in every answer :wink:

Maria-12648430: Fixed.

lib/stdlib/src/timer.erl
Maria-12648430: Dto. :wink:

lib/stdlib/src/timer.erl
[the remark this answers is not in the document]
Maria-12648430: Dto.

lib/stdlib/src/timer.erl
juhlig: > or @juhlig

Not me :innocent::grin:

lib/stdlib/doc/src/timer.xml
```diff
@@ -84,12 +84,45 @@
...
       <desc>
-        <p>Evaluates <c>apply(<anno>Module</anno>, <anno>Function</anno>,
+        <p>Evaluates <c>spawn(<anno>Module</anno>, <anno>Function</anno>,
           <anno>Arguments</anno>)</c> repeatedly at intervals of
-          <c><anno>Time</anno></c>.</p>
+          <c><anno>Time</anno></c>, waiting for the spawned process to
+          finish before starting the next.</p>
+        <p>If the execution time of the spawned process is greater than the
+          given <c><anno>Time</anno></c>, the next process is spawned immediately
+          after the one currently running has finished.</p>
```
rickard-green, line 125:
```suggestion
          after the one currently running has finished. Assuming that execution
          times of the spawned processes performing the applies on average
          are smaller than <c><anno>Time</anno></c>, the amount of applies
          made over a large amount of time will be the same even if some
          individual execution times are larger than
          <c><anno>Time</anno></c>. The system will try to catch up as soon
          as possible. For example, if one apply takes
          <c>2.5*<anno>Time</anno></c>, the following two applies will be
          made immediately one after the other in sequence.</p>
```

We discussed the `apply_repeatedly` functionality on the technical board and some comments came up about the documentation not being clear enough on what will happen when the apply greatly exceeds the interval. I've made a suggestion about an addition to the documentation above that makes this a little clearer.

rickard-green:
> @rickard-green
> 
> > You can put "OTP @OTP-18236@" in the since tag
> 
> Is there a way to find that out without asking? Because I'm _bound_ to forget, already did in the past, and also this time sweat_smile
> 

Sorry, no. We need to create an internal ticket for this. It is however one of the things we at OTP should check for when new functions are introduced, so I should have checked it :-)

lib/stdlib/doc/src/timer.xml
Maria-12648430: Looks fine to me, so I committed your suggestion :slightly_smiling_face:

Maria-12648430:
> so I should have checked it :-)

No worries, and I'll try to remember as well :wink:

lib/stdlib/doc/src/timer.xml
Maria-12648430: ... and by doing so, I popped it out of the test run again, did I? :laughing:

Maria-12648430:
@rickard-green this PR seems to have stalled? :thinking:

rickard-green:
> @rickard-green this PR seems to have stalled? thinking

I'll soon get to this. There has been other stuff taking precedence, e.g. the soon to be released OTP 25.1

Issue #6247, opened 2022-08-25
erts_no_of_hidden_dist_entries=-1 when racing happens between erts_do_net_exits and abort_pending_connection


zzydxm:
**Describe the bug**
We occasionally see coredumps with backtraces like:
```
#0  0x00000000006139bf in erts_do_net_exits (dep=<optimized out>, reason=<optimized out>) at beam/dist.c:947
#1  0x000000000046c6fb in erts_continue_exit_process (p=0x17798b8) at beam/erl_process.c:13975
#2  0x00007f7c6a663fe0 in erts_beamasm:continue_exit/0 ()
#3  0x00007f7c61b347a8 in proc_lib:exit_p/3 () at proc_lib.erl:261
#4  0x00007f7c6a663f38 in erts_beamasm:normal_exit/0-CodeInfoPrologue ()
```
```
#0  0x0000000000523107 in unlink_carrier (crr=0x7fc1411a8ec0, cl=0x7fc0f385bee0) at beam/erl_alloc_util.c:1258
#1  destroy_carrier (allctr=0x7fc0f385bd80, blk=0x7fc1411a8ee0, busy_pcrr_pp=<optimized out>) at beam/erl_alloc_util.c:4305
#2  0x0000000000530e8e in do_erts_alcu_free (busy_pcrr_pp=0x0, p=0x7fc1411a8ee8, allctr=0x7fc0f385bd80, type=1250)
    at beam/erl_alloc_util.c:5953
#3  do_erts_alcu_free (busy_pcrr_pp=0x0, p=<optimized out>, allctr=0x7fc0f385bd80, type=1250) at beam/erl_alloc_util.c:5929
#4  erts_alcu_free_thr_spec (type=1250, extra=<optimized out>, p=<optimized out>) at beam/erl_alloc_util.c:5993
#5  0x0000000000613b7e in erts_do_net_exits (dep=<optimized out>, reason=<optimized out>) at beam/dist.c:950
#6  0x000000000046c6fb in erts_continue_exit_process (p=0x17b6750) at beam/erl_process.c:13975
```
The coredump happens on node shutdown when there is pending connections. In both case, we see that erts_no_of_hidden_dist_entries=-1. This means there are duplicated calls of erts_set_dist_entry_not_connected, causing wrong value of erts_no_of_hidden_dist_entries. (Even though there is rwlock for dist table, it is still possible that one operation is blocked and the other operation pass through. Then blocked operation decreased the value again.)

[...]

PR #6258, opened 2022-08-29
Fix DistEntry corruption at node shutdown

sverker:
Proposed fix for #6247.

Two commits.

1. Assert the invariant that seems to fail. Must be CONNECTED to enter EXITING.
2. Fix. Don't hook a PENDING DistEntry into controlling port.


sverker:
Merged to maint for release in OTP 25.1

PR #6262, opened 2022-08-31
Make Ctrl+L clear the screen in the terminal

josevalim:
This also mirrors both Mac and Linux terminals.


lib/stdlib/src/edlin.erl
```diff
@@ -192,7 +195,8 @@ key_map($\^F, none) -> forward_char;
 key_map($\^H, none) -> backward_delete_char;
 key_map($\t, none) -> tab_expand;
 key_map($\^K, none) -> kill_line;
-key_map($\^L, none) -> redraw_line;
+key_map($\^L, none) -> clear;
+key_map($\^O, none) -> redraw_line;
```
garazdawi, line 199: Why did you use `^O` for line redraw?

lib/kernel/src/user_drv.erl
```diff
@@ -640,6 +642,8 @@ io_request({insert_chars, unicode, Chars}, TTY) ->
     write(prim_tty:handle_request(TTY, {insert, unicode:characters_to_binary(Chars)}));
 io_request({delete_chars, N}, TTY) ->
     write(prim_tty:handle_request(TTY, {delete, N}));
+io_request(clear, TTY) ->
```
garazdawi, line 645: `clear` also needs to be handled in [`ssh_cli.erl`](https://github.com/erlang/otp/blob/master/lib/ssh/src/ssh_cli.erl#L412-L436).

garazdawi (changes requested):
I think this is a good change to do.

^L is used by [to_erl](https://github.com/erlang/otp/blob/master/erts/etc/unix/to_erl.c#L353-L360) to make things work. I'm unsure if changing this would break any functionality there, but I'm pretty sure there are no tests for it so we need to make sure that it works as it should.

I would also like to see a test case in lib/kernel/test/interactive_shell_SUITE.erl using the new tmux testing interface as I did in #6260.

lib/kernel/src/prim_tty.erl
```diff
@@ -535,6 +536,8 @@ handle_request(State = #state{ xn = OrigXn, unicode = U }, {insert, Chars}) ->
      NewState};
 handle_request(State, beep) ->
     {<<7>>, State};
+handle_request(State, clear) ->
+    {<<"\e[H\e[2J">>, State};
```
garazdawi, line 540: Use `tgetstr("clear")` to fetch this value so that if we happen across a shell that prefers to use another ANSI code for clear we will do so.

lib/stdlib/src/edlin.erl
josevalim: I picked one that was not in use and has [no general meaning in the terminal](https://www.makeuseof.com/linux-bash-terminal-shortcuts/) (unless under history search).

lib/kernel/src/user_drv.erl
josevalim: Should we expose `prim_tty:tgetstr/1` so we can use it from `ssh_cli`? Or should we expose a specific `prim_tty:clear/1` instead?

garazdawi: I don't know how ssh is supposed to work in this regard. It seems like it is hard coded in the get_tty_command function right now, so I think we should be safe to do the same for `clear` in ssh.

ssh does some type of handshake procedure to get some of the information about the terminal that is connected, but how that is supposed to work I have no idea.

lib/stdlib/src/edlin.erl
garazdawi: The reason I ask is that we used ^O in #5924 to trigger opening of an editor to edit the current line.

josevalim: ^Oh! I think ^T is also fine, so I will roll with that.

josevalim: Scratch that, ^T is transpose. I think ^N is fine. I will give it a try.

josevalim: ^N is the down arrow. :'(

We are pretty much running out of characters here. Given ^L is used mostly for #6260, I wonder if it is worth removing the old feature altogether so we get one character back?

Anyway, I think ^Q is actually fine.

lib/kernel/src/user_drv.erl
josevalim: I see. SSH actually assumes VT100 and "clear" is part of VT100:

> Cursor to Home
> ESC H 	 
>
> Move the cursor to the home position.

> Erase to End of Screen
> ESC J 	 
>
> Erase all characters from the active position to the end of the screen. The active position is not changed.

So I think we are good with hard-coding alongside the other characters?

garazdawi: > So I think we are good with hard-coding alongside the other characters?

Sounds good

josevalim:
I should have addressed all feedback.

garazdawi:
Tests? Also, I just realized that https://github.com/erlang/otp/blob/master/erts/doc/src/tty.xml will also have to be updated.

josevalim:
@garazdawi any suggestions to where the tests should go? I haven't found suites for `user_drv` or `edlin`.

I have done manual tests in relation to run_erl/to_erl and it seems to work as expected (it connects and ^L still clears).

garazdawi:
Quoting myself:

> I would also like to see a test case in lib/kernel/test/interactive_shell_SUITE.erl using the new tmux testing interface as I did in https://github.com/erlang/otp/pull/6260.

josevalim:
> Also, I just realized that https://github.com/erlang/otp/blob/master/erts/doc/src/tty.xml will also have to be updated.

Done! Btw, I decided to not document Ctrl+Q because I don't know how to use it directly from the terminal. It feels it is for internal use only? But I will be glad to document it too if desired.

> Quoting myself:

Thanks, I missed it the first time around. :)

josevalim:
Tests are in.

garazdawi:
As you noticed, `C-Q` is intercepted by the terminal, so we cannot use that. I redid the PR to use `M-l` instead which seems to work just as well. I've rebased the commit and pushed some suggested changes, if you are happy with the changes please squash and I'll put it into test as soon as #6260 has been merged.

josevalim:
Looks great! Squashed and pushed!

garazdawi (approved):

garazdawi:
Rebased on latest master now that #6260 is merged and put into testing!

garazdawi:
Thanks!

PR #6266, opened 2022-08-31
kernel: Fix standard_error config race

### LINKED ISSUE #6263: Garbled standard error output while shell is initializing
**Describe the bug**

See those shell snippets:

```
$ erl -noshell -eval 'shell:start_interactive({io, put_chars, [standard_error, "oops\noops"]}).'
Erlang/OTP 26 [DEVELOPMENT] [erts-13.0.4] [source-098208c93f] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

oops
    oops*** ERROR: Shell process terminated! (^G to start new job) ***
```

```
$ erl -noshell -eval 'shell:start_interactive(), io:put_chars(standard_error, "oops\noops").'
Erlang/OTP 26 [DEVELOPMENT] [erts-13.0.4] [source-098208c93f] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

oops
    oops*** ERROR: Shell process terminated! (^G to start new job) ***
```

**Affected versions**

While this error happens in previous Erlang/OTP versions, it is easier to reproduce as of the master branch with the new shell. Historically, Elixir has chosen to pass its `erl -user` process to work around this bug. However, if this is addressed, then we can rely exclusively on the new `shell:start_interactive/1` API.

Maybe this could be addressed by invoking the shell callback AND returning from `shell:start_interactive/1` only after the `standard_error` has been replaced?

garazdawi:
When starting a shell using start:interactive_shell, standard_error
would not be correctly configured to use the correct newline sequence
when the shell was started causing staggering printouts.

    erl -noshell -eval 'shell:start_interactive(), io:put_chars(standard_error, "oops\noops").'
    oops
        oopsErlang/OTP 26

Closes #6263

### DIFF
--- a/lib/kernel/src/prim_tty.erl
+++ b/lib/kernel/src/prim_tty.erl
@@ -216,8 +216,9 @@ init_term(State = #state{ tty = TTY, options = Options }) ->
         case maps:get(tty, Options) of
             true ->
                 ok = tty_init(TTY, stdout, Options),
+                NewTTYState = init(State, os:type()),
                 ok = tty_set(TTY),
-                init(State, os:type());
+                NewTTYState;
             false ->
                 State
         end,
--- a/lib/kernel/src/user_drv.erl
+++ b/lib/kernel/src/user_drv.erl
@@ -258,8 +258,6 @@ init_local_shell(State, InitialShell) ->
 
 init_shell(State, Slogan) ->
 
-    init_standard_error(State#state.tty, State#state.shell_started),
-
     {next_state, server, State#state{ current_group = gr_cur_pid(State#state.groups) },
      {next_event, info,
       {gr_cur_pid(State#state.groups),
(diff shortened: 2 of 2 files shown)


josevalim:
@garazdawi thanks! This fixes the regular case but it leaves the dumb shell case (with enotsup) borked:

```
$ TERM=dumb erl -noshell -eval 'erlang:display(shell:start_interactive()), io:put_chars(standard_error, "oops\noops").'
{error,enotsup}
oops
    oops
```

garazdawi:
Pushed a new commit that should solve that problem as well.

josevalim:
@garazdawi perfect. It is working flawlessly on all of my test scenarios!

garazdawi:
Solved in #6279

PR #6269, opened 2022-09-01
Fix typo of min/max for +Q option

gomoripeti:
(Unless I am mistaken) the `ERTS_(MIN|MAX)_` constants for `PROCESSES` and `PORTS` have the same value, so this change is purely cosmetic.


rickard-green:
Thanks for the PR.

> (Unless I am mistaken) the ERTS_(MIN|MAX)_ constants for PROCESSES and PORTS have the same value, so this change is purely cosmetic.

Yes, but that might change in the future, so this fix might prevent a bug from appearing in the future.

PR #6270, opened 2022-09-01
ssl: Avoid partial connection trees

### LINKED ISSUE #6244: Potential process leakage of tls_dyn_connection_sup processes
**Describe the bug**

I believe there is a race condition in the `ssl` application that can lead to a leak of `tls_dyn_connection_sup` processes.

We are assuming this because, in a high-performance system, we are retrieving all instances of `tls_dyn_connection_sup` and some of them have either no children or 1 child. One hour later, we still see the same number of leaked processes with no or 1 child.

If someone calls `ssl:connect`, the following call trace might happen:

    ssl:connect -> tls_socket:connect
    tls_socket:connect -> ssl_gen_statem:connect
    ssl_gen_statem:connect -> tls_gen_connection:start_fsm

[`tls_gen_connection:start_fsm` has the following code](https://github.com/erlang/otp/blob/master/lib/ssl/src/tls_gen_connection.erl#L85-L86):

```erlang
start_fsm(Role, Host, Port, Socket,
          {#{erl_dist := false, sender_spawn_opts := SenderOpts}, _, Trackers} = Opts,
    User, {CbModule, _, _, _, _} = CbInfo, 
    Timeout) -> 
    try
        {ok, DynSup} = tls_connection_sup:start_child([]),
        {ok, Sender} = tls_dyn_connection_sup:start_child(DynSup, sender, [[{spawn_opt, SenderOpts}]]),
        {ok, Pid} = tls_dyn_connection_sup:start_child(DynSup, receiver, [Role, Sender, Host, Port, Socket,
                                                             Opts, User, CbInfo]),
        {ok, SslSocket} = ssl_gen_statem:socket_control(?MODULE, Socket, [Pid, Sender], CbModule, Trackers),
        ssl_gen_statem:handshake(SslSocket, Timeout)
    catch
  error:{badmatch, {error, _} = Error} ->
      Error
    end;
```

I assume the issue is with the code above: if the current process crashes for any reason (such as a broken process link), it can fail before `tls_dyn_connection_sup:start_child(DynSup, receiver, ...)` is invoked. My undestanding is that the receiver child (the second one) is the one monitoring the user process and therefore is the one responsible for triggering the significant cleanup in the supervisor. However, if the child process is never started due to a broken link, both `tls_dyn_connection_sup` and `tls_sender` will leak (and we see both leaking).

**To Reproduce**

Unfortunately I do not have an easy way to reproduce this issue, due to the race, but I believe the call trace above is reasonable?

**Potential solutions**

Feel free to ignore this section. :)

One of the possible solutions is to change `tls_dyn_connection_sup` to start both children as part of its initialization and then call `supervisor:which_children` to retrieve their PIDs. A potential downside of this approach is that `tls_dyn_connection_sup` takes longer to start, increasing the odds of `tls_connection_sup` itself becoming a bottleneck? It may be necessary to make `tls_connection_sup` a pool of processes that are randomly routed to.

**Affected versions**

OTP 25.

IngelaAndin:
If the "User" process, the process starting the TLS connection, gets
killed in the middle of spawning the dynamic connection tree make sure
we do not leave any processes behind.

Close #6244

### DIFF
--- a/lib/ssl/src/tls_gen_connection.erl
+++ b/lib/ssl/src/tls_gen_connection.erl
@@ -72,41 +72,68 @@
 
 %%====================================================================
 %% Internal application API
-%%====================================================================	     
+%%====================================================================
 %%====================================================================
 %% Setup
 %%====================================================================
 start_fsm(Role, Host, Port, Socket,
-          {#{erl_dist := false, sender_spawn_opts := SenderOpts}, _, Trackers} = Opts,
-	  User, {CbModule, _, _, _, _} = CbInfo, 
-	  Timeout) -> 
-    try
-        {ok, DynSup} = tls_connection_sup:start_child([]),
-        {ok, Sender} = tls_dyn_connection_sup:start_child(DynSup, sender, [[{spawn_opt, SenderOpts}]]),
-	{ok, Pid} = tls_dyn_connection_sup:start_child(DynSup, receiver, [Role, Sender, Host, Port, Socket,
-                                                             Opts, User, CbInfo]),
-	{ok, SslSocket} = ssl_gen_statem:socket_control(?MODULE, Socket, [Pid, Sender], CbModule, Trackers),
-        ssl_gen_statem:handshake(SslSocket, Timeout)
-    catch
-	error:{badmatch, {error, _} = Error} ->
-	    Error
-    end;
+          {#{erl_dist := ErlDist, sender_spawn_opts := SenderSpawnOpts}, _, Trackers} = Opts,
+	  User, {CbModule, _, _, _, _} = CbInfo,
+	  Timeout) ->
+    SenderOptions = handle_sender_options(ErlDist, SenderSpawnOpts),
+    Starter = start_connection_tree(User, ErlDist, SenderOptions,
+                                    Role, [Host, Port, Socket, Opts, User, CbInfo]),
+    receive
+        {Starter, SockReceiver, SockSender} ->
+            socket_control(Socket, SockReceiver, SockSender, CbModule, Trackers, Timeout);
+        {Starter, Error} ->
+            Error
+    end.
 
-start_fsm(Role, Host, Port, Socket,
-          {#{erl_dist := true, sender_spawn_opts := SenderOpts}, _, Trackers} = Opts,
-	  User, {CbModule, _, _, _, _} = CbInfo, 
-	  Timeout) -> 
-    try
-        SenderOpts1 = [{priority, max} | proplists:delete(priority, SenderOpts)],
-        {ok, DynSup} = tls_connection_sup:start_child_dist([]),
-        {ok, Sender} = tls_dyn_connection_sup:start_child(DynSup, sender, [[{spawn_opt, SenderOpts1}]]),
-	{ok, Pid} = tls_dyn_connection_sup:start_child(DynSup, receiver, [Role, Sender, Host, Port, Socket,
-                                                                 Opts, User, CbInfo]),
-	{ok, SslSocket} = ssl_gen_statem:socket_control(?MODULE, Socket, [Pid, Sender], CbModule, Trackers),
-        ssl_gen_statem:handshake(SslSocket, Timeout)
-    catch
-	error:{badmatch, {error, _} = Error} ->
-	    Error
+handle_sender_options(ErlDist, SpawnOpts) ->
+    case ErlDist of
+        true ->
+            [[{spawn_opt, [{priority, max} | proplists:delete(priority, SpawnOpts)]}]];
+        false ->
+            [[{spawn_opt, SpawnOpts}]]
+    end.
+
+start_connection_tree(User, IsErlDist, SenderOpts, Role, ReceiverOpts) ->
+    StartConnectionTree =
+        fun() ->
+                case start_dyn_connection_sup(IsErlDist) of
+                    {ok, DynSup} ->
+                        case tls_dyn_connection_sup:start_child(DynSup, sender, SenderOpts) of
+                            {ok, Sender} ->
+                                case tls_dyn_connection_sup:start_child(DynSup, receiver,
+                                                                        [Role, Sender | ReceiverOpts]) of
+                                    {ok, Receiver} ->
+                                        User ! {self(), Receiver, Sender};
+                                    {error, Error} ->
+                                        User ! {self(), Error},
+                                        exit(DynSup, shutdown)
+                                end;
+                            {error, Error} ->
+                                User ! {self(), Error},
+                                exit(DynSup, shutdown)
+                        end;
+                    {error, Error} ->
+                        User ! {self(), Error}
+                end
+        end,
+    spawn(StartConnectionTree).
+
+start_dyn_connection_sup(true) ->
+    tls_connection_sup:start_child_dist([]);
+start_dyn_connection_sup(false) ->
+    tls_connection_sup:start_child([]).
+
+socket_control(Socket, SockReceiver, SockSender, CbModule, Trackers, Timeout) ->
+    case ssl_gen_statem:socket_control(?MODULE, Socket, [SockReceiver, SockSender], CbModule, Trackers) of
+        {ok, SslSocket} ->
+            ssl_gen_statem:handshake(SslSocket, Timeout);
+        Error ->
+            Error
     end.
 
 pids(#state{protocol_specific = #{sender := Sender}}) ->
--- a/lib/ssl/test/ssl_basic_SUITE.erl
+++ b/lib/ssl/test/ssl_basic_SUITE.erl
@@ -76,8 +76,8 @@
          fake_root_no_intermediate_legacy/1,
          fake_intermediate_cert/0,
          fake_intermediate_cert/1,
-         incompleat_chain_length/0,
-         incompleat_chain_length/1,
+         incomplete_chain_length/0,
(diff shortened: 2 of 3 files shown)


sneako:
Thanks @IngelaAndin ! I can test this out on Monday and let you know if it resolves the issue we were seeing. I am also curious if this new process that starts the connection tree will add any noticeable overhead. I will report back soon.

lib/ssl/src/tls_gen_connection.erl
```diff
@@ -77,36 +77,68 @@
...
+                                end;
+                            {error, Error} ->
+                                User ! {self(), Error},
+                                exit(DynSup, shutdown)
+                        end;
+                    {error, Error} ->
+                        User ! {self(), Error}
+                end
+        end,
+    spawn(StartConnectionTree).
```
KennethL, line 134: I suggest something like below in order to avoid a dynamic call  with the function name in a variable
```suggestion
dyn_connection_sup_child(true,ArgList) ->
    tls_connection_sup:start_child_dist(ArgList);
dyn_connection_sup_child(false,ArgList) ->
    tls_connection_sup:start_child(ArgList);
    
start_connection_tree(User, ErlDist, SenderOpts, Role, ReceiverOpts) ->
    StartConnectionTree =
        fun() ->
                case dyn_connection_sup_child(ErlDist, []) of
                    {ok, DynSup} ->
                        case tls_dyn_connection_sup:start_child(DynSup, sender, SenderOpts) of
                            {ok, Sender} ->
                                case tls_dyn_connection_sup:start_child(DynSup, receiver,
                                                                        [Role, Sender | ReceiverOpts]) of
                                    {ok, Receiver} ->
                                        User ! {self(), Receiver, Sender};
                                    {error, Error} ->
                                        User ! {self(), Error},
                                        exit(DynSup, shutdown)
                                end;
                            {error, Error} ->
                                User ! {self(), Error},
                                exit(DynSup, shutdown)
                        end;
                    {error, Error} ->
                        User ! {self(), Error}
                end
        end,
    spawn(StartConnectionTree).
```

KennethL (changes requested):

lib/ssl/src/tls_gen_connection.erl
IngelaAndin: Dynamic function calls can be really useful, but in this case, it is not really that valuable so we can make it along the suggested lines. I thought I made the choice really obvious though!

lib/ssl/src/tls_gen_connection.erl
```diff
@@ -77,36 +77,67 @@
...
     end;
-
 start_fsm(Role, Host, Port, Socket,
-          {#{erl_dist := true, sender_spawn_opts := SenderOpts}, _, Trackers} = Opts,
+          {#{erl_dist := true = ErlDist, sender_spawn_opts := SenderOpts}, _, Trackers} = Opts,
 	  User, {CbModule, _, _, _, _} = CbInfo, 
-	  Timeout) -> 
-    try
+	  Timeout) ->
         SenderOpts1 = [{priority, max} | proplists:delete(priority, SenderOpts)],
```
KennethL, line 95: Since the 2 start_fsm clauses are identical except for one line, I propose that they are put together like this:
```erlang
start_fsm(Role, Host, Port, Socket,
          {#{erl_dist := true = ErlDist, sender_spawn_opts := SenderOpts}, _, Trackers} = Opts,
	  User, {CbModule, _, _, _, _} = CbInfo, 
	  Timeout) ->
        SenderOpts1 = set_erl_dist_priority(ErlDist,SenderOpts),
    Starter = start_connection_tree(User, ErlDist, [[{spawn_opt, SenderOpts1}]],
                                    Role, [Host, Port, Socket,Opts, User, CbInfo]),
    receive
        {Starter, SockReceiver, SockSender} ->
            socket_control(Socket, SockReceiver, SockSender, CbModule, Trackers, Timeout);
        {Starter, Error} ->
            Error
    end.

%% put this helper function in a suitable place in the module
set_erldist_priority(true,SenderOpts) ->
        [{priority, max} | proplists:delete(priority, SenderOpts)];
set_erldist_priority(false,SenderOpts) ->
       SenderOpts.

```

KennethL (changes requested):

IngelaAndin:
@sneako How is the testing going?

sneako:
I was waiting for the requested changes to be addressed so that I could be sure to test the final version. If you think this is ready to go I can start testing this afternoon!

lib/ssl/src/tls_gen_connection.erl
IngelaAndin: Sure we can remove some code duplication here. I prefer to generalize the helper function to handle the sender options.

IngelaAndin:
@sneako well the changes were mainly cosmetic, but I made a new version now.

IngelaAndin:
@sneako I added a test case cuddle, but I would not worry about it for the purpose of your testing, the test case is sensitive to timing and testing timeout functionality is tricky, I do not think this is actualy related to my change.

sneako:
Sounds good! I am going to deploy the test shortly. Will have some results for you by the end of the day.

KennethL (approved):

sneako:
This patch has been running for almost 4 hours now on a subset of our prod nodes and we have not seen a single instance of a `tls_dyn_connection_sup` having less than 2 children during this period. We have observed the other nodes running OTP 25.0.2 slowly accumulate `tls_dyn_connection_sup` with 0, or 1 children. Performance is very similar across the two OTP versions as well.

As far as I can tell, this PR seems to have completely resolved the reported issue, thanks @IngelaAndin !

lib/ssl/test/ssl_basic_SUITE.erl
```diff
@@ -986,3 +1012,42 @@ chipher_suite_checks(Version) ->
...
+
+loop(L) ->
+    {ok, A} = ssl:transport_accept(L),
+    spawn(fun() ->
+                  ssl:handshake(A)
+          end),
+    loop(L).
+                 
+check_process_count() ->
+    check_process_count(5).
```
u3s, line 1041: maybe minor detail but: I don't see value added for having check_process_count/0 function wrapper. I would probably skip adding it in that situation.

lib/ssl/test/ssl_basic_SUITE.erl
```diff
@@ -77,7 +77,9 @@
          fake_intermediate_cert/0,
          fake_intermediate_cert/1,
          incompleat_chain_length/0,
-         incompleat_chain_length/1
+         incompleat_chain_length/1,
```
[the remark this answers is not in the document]
IngelaAndin: This function is not really part of this PR (that is old typo). But I can make a clean-up commit.

lib/ssl/test/ssl_basic_SUITE.erl
IngelaAndin: It hides the implementation detail of waiting long enough for the processes that should have been terminated to terminate.
However, I add an argument so now it is a wrapper with one argument.

lib/ssl/test/ssl_basic_SUITE.erl
```diff
@@ -782,6 +786,28 @@ incompleat_chain_length(Config) when is_list(Config)->
...
+    ClientOpts = ssl_test_lib:ssl_options(client_rsa_verify_opts, Config),
+    {_, ServerNode, Hostname} = ssl_test_lib:run_where(Config),
+
+    Port = ssl_test_lib:inet_port(ServerNode),
+    Server = spawn_link(fun() ->
+                                {ok, L} = ssl:listen(Port, ServerOpts),
+                                loop(L)
+                        end),
+    {ok, _} = ssl:connect(Hostname, Port, ClientOpts),
+    2 = proplists:get_value(supervisors, supervisor:count_children(tls_connection_sup)),
```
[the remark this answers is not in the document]
IngelaAndin: It could have been replaced by check_process_count/0, but I added an argument to the function to keep the knowledge of the expected count in the main test case function.

lib/ssl/src/tls_gen_connection.erl
```diff
@@ -72,41 +72,68 @@
...
+                    {ok, DynSup} ->
+                        case tls_dyn_connection_sup:start_child(DynSup, sender, SenderOpts) of
+                            {ok, Sender} ->
+                                case tls_dyn_connection_sup:start_child(DynSup, receiver,
+                                                                        [Role, Sender | ReceiverOpts]) of
+                                    {ok, Receiver} ->
+                                        User ! {self(), Receiver, Sender};
+                                    {error, Error} ->
+                                        User ! {self(), Error},
+                                        exit(DynSup, shutdown)
```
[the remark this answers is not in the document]
IngelaAndin: If you shut down the supervisor it will shut down its children, it is its job!

lib/ssl/test/ssl_basic_SUITE.erl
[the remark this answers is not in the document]
u3s: yes. I meant you could replace line 804 with check_process_count(0) call and reduce code repetition ...

sneako:
I ended up leaving this test running overnight, almost 24 hours now, and things still look great :grinning:

IngelaAndin:
@sneako  Super :) I will be merging once the test enhancements (suggested by @u3s)  have passed the test step.

PR #6275, opened 2022-09-02
[pg] fix pg:monitor_scope/0,1 return value

max-au:
These functions were expected to return the entire scope content,
but due to a bug only local pids were returned.


max-au:
I verified that test failure in global_SUITE is unrelated to my changes.

PR #6268, opened 2022-09-01
Store current_group and shell in pdict

josevalim:

PR #6279, opened 2022-09-05
kernel: Incorporate "oldshell" into user_drv

garazdawi:
This commit cleans up #6144 by:

1. Removing the old `oldshell` implementation and merging it into `newshell`
2. Made `shell:start_interactive/1` start an `oldshell` if it cannot start a `newshell`
3. Improved `-remsh` to work for both `oldshell` and `newshell`
4. Added shell:whereis_shell() to find the pid of the current shell process.

This PR also implements the fixes in #6266 and #6268.

This PR needs to have more tests before it is merged.

@josevalim Could you take a look at this and see if this works for Elixir and if you still need a way to detect if you are running an oldshell or not? My goal is that, after you stop supporting Erlang/OTP 25, IEx should be able to stop using `-user` and instead start like this: `erl -noinput -eval 'shell:start_interactive({IEx, start, []})'`, or for `iex --remsh` `erl -noinput -eval 'shell:start_interactive({remote, Node, {IEx, start, []}})'`. You could even put the `shell:start_interactive/1` call in the application start callback of the `IEx` application if you wanted to.


josevalim:
@garazdawi thanks, I will check tomorrow the latest!

But for documentation purposes, I needed to detect if we are in an old shell for two reasons:

1. The oldshell does not support remsh. If oldshell now supports remsh, then we are good!

2. The equivalent to `shell:whereis_shell()` in the past could not find oldshells, so I had to detect those cases myself. If it can now find the old shell, then I don't need it anymore :)

garazdawi:
> * The oldshell does not support remsh. If oldshell now supports remsh, then we are good!
> * The equivalent to `shell:whereis_shell()` in the past could not find oldshells, so I had to detect those cases myself. If it can now find the old shell, then I don't need it anymore :)

The implementation in this PR should fix both of these problems, but please try and make sure that it does work for you.

josevalim:
@garazdawi this is amazing! I have done extensive tests and everything works, both using `-user_drv` and the eval approach you mentioned in the issue description.

Not only that, you are actually removing several LOCs here and I will be able to do the same in Elixir too. :)

> My goal is that, after you stop supporting Erlang/OTP 25, IEx should be able to stop using -user 

Btw, if you want to remove `-user` support by any chance (AFAIK it is undocumented), I should be able to deal with it accordingly. :)

garazdawi:
> Not only that, you are actually removing several LOCs here and I will be able to do the same in Elixir too. :)

Always a good feeling when you make something better and remove code at the same time.

> Btw, if you want to remove -user support by any chance (AFAIK it is undocumented), I should be able to deal with it accordingly. :)

Yes, I think that I would like to do that eventually, though [peer](https://www.erlang.org/doc/man/peer.html#) uses it right now to get into nodes very early in the boot sequence, so I'll need to figure out how to solve its needs also.

PR #6280, opened 2022-09-05
Add selectable alphabets (standard vs urlsafe) to base64

juhlig:
In #5639 an addition to the `base64` module to support an alternative, URL-safe encoding alphabet (`base64url`) was suggested. In the alternative encoding, the characters `+` and `/` ([RFC 4648 Section 4](https://www.rfc-editor.org/rfc/rfc4648#section-4)), which may be problematic in URLs and file names, are replaced with `-` and `_` ([RFC 4648 Section 5](https://www.rfc-editor.org/rfc/rfc4648#section-5)), respectively.

This PR adds the new functions `encode/2`, `encode_to_string/2`, `decode/2`, `decode_to_string/2`, `mime_decode/2` and `mime_decode_to_string/2` as supplements of the respective existing 1-ary functions of the same names. The second parameters may be one of the atoms `standard` (meaning the section 4 alphabet; default for the 1-ary functions) and `urlsafe` (the section 5 alphabet), which denote the alphabet to be used for encoding and decoding. The decoding functions also accept `undefined`, via which they accept characters from both alphabets.

I'm still not clear on how to best spec this, and documentation is another matter, so this is a WIP.
As for tests, I added two new unit tests for now. If this PR is to be accepted, I'll extend the property test suite accordingly.

### DIFF
--- a/lib/stdlib/src/base64.erl
+++ b/lib/stdlib/src/base64.erl
@@ -21,12 +21,19 @@
 
 -module(base64).
 
--export([encode/1, decode/1, mime_decode/1,
-	 encode_to_string/1, decode_to_string/1, mime_decode_to_string/1]).
-
+-export([encode/1, encode/2,
+	 decode/1, decode/2,
+	 mime_decode/1, mime_decode/2,
+	 encode_to_string/1, encode_to_string/2,
+	 decode_to_string/1, decode_to_string/2,
+	 mime_decode_to_string/1, mime_decode_to_string/2]).
 
 %% RFC 4648: Base 64 Encoding alphabet
--type base64_alphabet() :: $A..$Z | $a..$z | $0..$9 | $+ | $/ | $=.
+-type base64_alphabet() :: $A..$Z | $a..$z | $0..$9 | $+ | $/ | $- | $_ | $=.
+
+%% Selector for the Base 64 alphabet, `standard'  for RFC 4648
+%% Section 4, `urlsafe'  for RFC 4648 Section 5.
+-type base64_mode() :: 'standard' | 'urlsafe'.
 
 %% The following type is a subtype of string() for return values
 %% of encoding functions.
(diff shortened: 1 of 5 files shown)


bjorng:
Having worked on improving the performance of `base64` (mainly in #6031, #6259, and #6274), I would prefer that additions to the `base64` module are done in a way that doesn't lower performance more than necessary.

Running [base64_bench](https://gist.github.com/bjorng/522470f313ac01d16156a597657f97bb) on my M1 MacBook Pro on the `master` branch, the results are the following:

```
== Testing with 1 MB ==
fun base64:encode/1: 1000 iterations in 3386 ms: 295 it/sec
fun base64:decode/1: 1000 iterations in 5167 ms: 193 it/sec
```

The results with this pull request are:

```
== Testing with 1 MB ==
fun base64:encode/1: 1000 iterations in 4142 ms: 241 it/sec
fun base64:decode/1: 1000 iterations in 6232 ms: 160 it/sec
```

One way to improve the performance for encoding without having completely separate code for the `standard` and `urlsafe` modes is to rewrite `b64e/2` like this:

```
b64e(X, Offset) ->
    element(X+Offset,
	    {$A, $B, $C, $D, $E, $F, $G, $H, $I, $J, $K, $L, $M, $N,
	     $O, $P, $Q, $R, $S, $T, $U, $V, $W, $X, $Y, $Z,
	     $a, $b, $c, $d, $e, $f, $g, $h, $i, $j, $k, $l, $m, $n,
	     $o, $p, $q, $r, $s, $t, $u, $v, $w, $x, $y, $z,
	     $0, $1, $2, $3, $4, $5, $6, $7, $8, $9, $+, $/,
             $A, $B, $C, $D, $E, $F, $G, $H, $I, $J, $K, $L, $M, $N,
	     $O, $P, $Q, $R, $S, $T, $U, $V, $W, $X, $Y, $Z,
	     $a, $b, $c, $d, $e, $f, $g, $h, $i, $j, $k, $l, $m, $n,
	     $o, $p, $q, $r, $s, $t, $u, $v, $w, $x, $y, $z,
	     $0, $1, $2, $3, $4, $5, $6, $7, $8, $9, $-, $_}).
```

where `Offset` is to be calculated in the following way:

```
    Offset = case Mode of
                 standard -> 1;
                 urlsafe -> 65
             end
```

The benchmark results are:

```
== Testing with 1 MB ==
fun base64:encode/1: 1000 iterations in 3523 ms: 283 it/sec
fun base64:decode/1: 1000 iterations in 6240 ms: 160 it/sec
```

which is closer to the original performance and probably acceptable.

It should be possible to optimize decoding in a similar way.

juhlig:
@bjorng I refactored as you suggested. Running the `base64_bench` on my machine, it looks like it is roughly on par with current `master`, but it would be good if you could double-check ;)

lib/stdlib/src/base64.erl
```diff
@@ -512,13 +542,60 @@ b64d(X) ->
              bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,
              bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,
              bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,
+             bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,bad,
+
+	     % alternative base64url alphabet (RFC 4648 Section 5)
```
bjorng, line 547:
```suggestion
	     %% alternative base64url alphabet (RFC 4648 Section 5)
```

and similarly for other `%` comments in similar positions (i.e. not comments at the end of a line).

lib/stdlib/src/base64.erl
```diff
@@ -140,367 +158,379 @@ mime_decode(List) when is_list(List) ->
...
     <<A/bits,B1:6,B2:6,(B3 bsr 2):4>>.
 
-mime_decode_binary(<<0:8, Cs/bits>>, A) ->
-    mime_decode_binary(Cs, A);
-mime_decode_binary(<<C1:8, Cs/bits>>, A) ->
-    case b64d(C1) of
-        B1 when is_integer(B1) -> mime_decode_binary(Cs, A, B1);
-        _ -> mime_decode_binary(Cs, A)  % eq is padding
+mime_decode_binary(ModeOffset, <<0:8, Cs/bits>>, A) ->
+    mime_decode_binary(ModeOffset, Cs, A);
```
bjorng, line 243: If you change the decode offsets to be 1, 256, and 512 (with the corresponding adjustments to the tuple), you can omit this clause and all similar clauses.

```suggestion
```

bjorng (changes requested):

lib/stdlib/src/base64.erl
juhlig: Done (but it is 257 and 513 ;))

bjorng (approved):
Looks good to me.

Benchmark results on my Intel Mac are almost exactly the same as for the master branch.

Regarding the specs, I think that the easiest solution is to extend `base64_alphabet()` to include the characters from both variants of Base64.

It seems that you are not testing the new `mime_decode` functions. I suggest that you add tests for them in the funs in the `mime_decode/1` and `mime_decode_to_string/1` test cases.

Since this is an extension of an API, the OTP Technical Board will have to approve it.

juhlig:
> Looks good to me.
> 
> Benchmark results on my Intel Mac are almost exactly the same as for the master branch.

Great =)

> Regarding the specs, I think that the easiest solution is to extend `base64_alphabet()` to include the characters from both variants of Base64.

Ok, then I'll do it that way.

> It seems that you are not testing the new `mime_decode` functions. I suggest that you add tests for them in the funs in the `mime_decode/1` and `mime_decode_to_string/1` test cases.

Yeah, and there is also the property test suite that needs to be extended to cover the new functions. I'm currently working on that.

> Since this is an extension of an API, the OTP Technical Board will have to approve it.

Sure.

Maria-12648430:
Something that just occurred to me when reading this PR... the `undefined` decoding mode (accepting both `+` / `-` and `/` / `_` to refer to `62` and `63` respectively) may be a bit dangerous, at least material for surprises, especially when used with `mime_decode/2` (vs `decode/2`). You may want to bring that up with the OTB when discussing this extension.

<hr />

Given the encoded string `<<"AB+C+D/EF/G-H-I_J_KL">>`, this mime-decodes differently in the different modes:
```text
Mode      | ignored chars | cleaned string equivalent  | decodes to
=================================================================================================================
standard  | "-" and "_"   | <<"AB+C+D/E/FGHIJKL">>     | <<0,31,130,248,63,196,252,81,135,32,146,139>>
----------+---------------+----------------------------+---------------------------------------------------------
urlsafe   | "+" and "/"   | <<"ABCDEFG-H-I_J_KL">>     | <<0,16,131,16,81,190,31,226,63,39,242,139>>
----------+---------------+----------------------------+---------------------------------------------------------
undefined | none          | <<"AB+C+D/E/FG-H-I_J_KL">> | <<0,31,130,248,63,196,252,81,190,31,226,63,39,242,139>>
----------+---------------+----------------------------+---------------------------------------------------------
```
(It is similar with `decode/2`: the first two cases would result in an exception, the last case would decode.)

IMO, it is ok and expected that the same string may decode differently in `standard` and `urlsafe` modes, given that there the user has explicitly specified exactly how he wants the data to be interpreted. However, that `undefined` decodes yet differently than any one of the specific modes looks like a problem to me.

<hr />

The possible ways I can think of to address the problem:
* remove `undefined`, so the user _has_ to specify how he wants the data to be interpreted and decoded
* keep `undefined` but put a warning in the docs ([but who ever reads docs before it is too late?](https://github.com/ninenines/ranch/pull/313#issuecomment-793192238))
* rename `undefined` to something like `mixed`, to remove the impression of auto-detection (`undefined` --> "dunno, figure it out")

juhlig:
Good point @Maria-12648430, you do have a knack for things like that :grimacing:

I'll leave things as they are for now, any of your suggestions is easy to do, whatever the decision may be.

bjorng:
@Maria-12648430 Yes, I will bring that up with the OTB.

Maria-12648430:
> you do have a knack for things like that

More like a dark spot in my soul that urges me to poke holes in shiny things :innocent:

juhlig:
> More like a dark spot in my soul that urges me to poke holes in shiny things :innocent:

That is why no company that values its pride should ever hire you :grin: You would find three different ways to capsize their Unsinkable Flagship<sup>:registered:</sup> on your first day.

Maria-12648430:
> You would find three different ways to capsize their Unsinkable Flagship<sup>:registered:</sup> on your first day.

_And_ enjoy it :smiley_cat: But make that 5 ways :grin:

juhlig:
The last push adds specs, tests for `mime_decode/2` (in separate cases, I think they don't fit well into the existing ones), and documentation. (Don't worry about the sloppy commit messages, I'll squash and tidy them up in the end :wink:)

(On a side note, some of the links to RFC 4648 were broken in the docs, and I fixed them while I was at it.)

The property test suite will be extended later by @Maria-12648430 (thanks :hugs:), since she implemented the current test suite and has more experience with property testing than I do.

Last but not least, I have been discussing the `undefined` decoding mode some more with @Maria-12648430, and we came to the conclusion that it would be best to remove/not include it. It is brittle, prone to be mistaken for auto-detection, and even if understood correctly (as `mixed`) there seems little real use for it. I have spec'ed and documented it all the same, in case the OTB sees any value in it after all, but by and large we would opt for removal and force users to be explicit.

RaimoNiskanen:
`undefined` could make a decision at the first `+`, `/`, `-`, `_` character and then require all to be from the same alphabet.
I do not know if that would improve usefulness (enough)...

Maria-12648430:
Hi @RaimoNiskanen :slightly_smiling_face:

> `undefined` could make a decision at the first `+`, `/`, `-`, `_` character and then require all to be from the same alphabet.

We have been discussing this also, but, well...

First, it would make the code more complex. There would either have to be extra clauses and/or guards on practically all functions involved in the decoding (and quite many of them, like `undefined` and `+`, `undefined` and `-`, etc; and as many functions work on several encoded characters at once, even `undefined` and `+` in the first character, `undefined` and `+` in the second character, etcetc), or on the helper function `b64d` with repercussions into the decoding functions, primarily meaning more `case` clauses or entire new `case` constructs where there were none before.
Both would degrade performance, something that @bjorng will understandably _not_ like to see, given that he has worked hard in the past to improve base64 performance.

Second, it would still be brittle, just in a different way. It would be guesswork. I think it is better to not give raise to the illusion that the right alphabet will always be detected automatically.

Third and last, I don't think it is very useful, possibly dangerous even. Users should be able to know where the data to decode comes from, and what kind of alphabet (not) to expect.
If we had `undefined`, users would just use it for convenience and good measure (if you can have both, why bother deciding on one?), under the illusion that the right alphabet will be detected reliably, as I said.
Moreover, that may even _work_ in development and pass test cases, and even work in production for a long time, but every now and then strange things (incorrectly decoded and therefore corrupt or even mailcious data) may happen, for no _obvious_ reason.

Maria-12648430:
The last commit adds the property tests. I didn't bother with `undefined` mode, though. That can be added later if necessary :wink:

bjorng:
The OTB meeting agreed with the conclusion reached in the previous comments on this PR that the `undefined` option is harmful and that it should be removed. When you have removed it and cleaned up the commits, I will add this PR to our daily builds.

Maria-12648430:
> The last commit adds the property tests.

Btw, I forgot to mention that I also fixed something that was pointed out by @seriyps after I did the initial property test suite, [here](https://github.com/erlang/otp/pull/5967#discussion_r887815311).

juhlig:
@bjorng all done, I removed `undefined` mode from code, tests and docs, squashed the commits and provided a nicer commit message.

bjorng:

bjorng:
I forgot about the `since` tags. Please add `"OTP @OTP-18247@"` to the `since` tags for the new functions.

juhlig:
> I forgot about the `since` tags.

Dto. :laughing:

> Please add `"OTP @OTP-18247@"` to the `since` tags for the new functions.

Done

bjorng:
Thanks!

PR #6286, opened 2022-09-08
inets: add missing typespecs

kikofernandez:
This PR adds missing typespecs and the XML documentation into the source file.


KennethL:
The xml documentation should still be in a separate file but it should be different than before when it references -specs and -types from the source code. I think there is some misunderstanding there.

kikofernandez:
@KennethL you are absolutely right. sorry about that

PR #6287, opened 2022-09-08
ssl: adjust cacertfile option

u3s:
- perform conversions before loading cert data from file
- convert relative path to absolute for avoiding collisions
- convert symlink to actual file path

### DIFF
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -2102,10 +2102,10 @@ validate_option(cacertfile, undefined, _) ->
    <<>>;
 validate_option(cacertfile, Value, _)
   when is_binary(Value) ->
-    Value;
+    unambiguous_path(Value);
 validate_option(cacertfile, Value, _)
   when is_list(Value), Value =/= ""->
-    binary_filename(Value);
+    binary_filename(unambiguous_path(Value));
 validate_option(cacerts, Value, _)
   when Value == undefined;
        is_list(Value) ->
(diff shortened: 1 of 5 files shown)


lib/ssl/src/ssl.erl
```diff
@@ -2102,10 +2102,10 @@ validate_option(cacertfile, undefined, _) ->
    <<>>;
 validate_option(cacertfile, Value, _)
   when is_binary(Value) ->
-    Value;
+    adjust_cacertfile(Value);
```
IngelaAndin, line 2105: I think something like unambiguous_path would be more explanatory.

lib/ssl/src/ssl_manager.erl
```diff
@@ -465,7 +465,7 @@ invalidate_session(Cache, CacheCb, Key, _Session,
 
 clean_cert_db(Ref, CertDb, RefDb, FileMapDb, File) ->
     case ssl_pkix_db:ref_count(Ref, RefDb, 0) of
-	0 ->
```
IngelaAndin, line 468: I think that it is nice to make white space changes as a separate commit for cosmetic enhancments so that it is easy to see what code is important to the code logic. This is however such a small change that it perhaps matters less.

lib/ssl/test/ssl_pem_cache_SUITE.erl
```diff
@@ -390,12 +390,12 @@ alternative_path_hardlink(Config) when is_list(Config) ->
 alternative_path_symlink() ->
     [{doc,"Test that internal reference table contains expected data for"
       " absolute and symbolic link. "
-      "This test verifies handling of same file with an alternative reference."}].
-%% see alternative_path_hardlink for specification
+      "This test verifies handling of same file with an alternative reference."
```
IngelaAndin, line 393: Maybe add some explanation of why counters change in the way they do!? See for instance the manual clean-up case.

lib/ssl/src/ssl_manager.erl
u3s: Yes I agree in general. Will avoid that in future.

lib/ssl/test/ssl_pem_cache_SUITE.erl
u3s: I tried to improve the comment.

PR #6291, opened 2022-09-08
inets: Implement HTTP 308 Permanent Redirect

### LINKED ISSUE #6290: [inets] Add support of HTTP status code 308 (RFC 7538)
**Describe the bug**
* httpd_util:reason_phrase(308). returns "Internal Server Error" instead of "Permanent Redirect"
* httpc doesn't follow redirect upon reception of HTTP 308 even if autodirect option is set to true

**To Reproduce**
* Call httpd_util:reason_phrase(308).
* Call httpc:request with autoredirect=true towards an HTTP server that returns 308 and a location header

**Expected behavior**
* Return "Permanent Redirect"
* Follow redirect if autoredirect option is set to true

**Affected versions**
OTP <= 25

**Additional context**
HTTP 308 status code has been introduced by RFC 7538 https://www.rfc-editor.org/rfc/rfc7538

DataTriny:
Fixes #6290 

Adds support for the 308 Permanent Redirect HTTP status code to `inets`, as described in [RFC 9110](https://datatracker.ietf.org/doc/rfc9110/).

### DIFF
--- a/lib/inets/doc/src/httpd_util.xml
+++ b/lib/inets/doc/src/httpd_util.xml
@@ -264,7 +264,10 @@
       <name since="">reason_phrase(StatusCode) -> Description</name>
       <fsummary>Returns the description of an HTTP 1.1 status code.</fsummary>
       <type>
-        <v>StatusCode = 100| 200 | 201 | 202 | 204 | 205 | 206 | 300 | 301 | 302 | 303 | 304 | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 410 411 | 412 | 413 | 414 415 | 416 | 417 | 500 | 501 | 502 | 503 | 504 | 505</v>
+        <v>StatusCode = 100 | 200 | 201 | 202 | 204 | 205 | 206 | 300 | 301 | 
+          302 | 303 | 304 | 308 | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 
+          410 | 411 | 412 | 413 | 414 | 415 | 416 | 417 | 500 | 501 | 502 | 
+          503 | 504 | 505</v>
         <v>Description = string()</v>
       </type>
       <desc>
--- a/lib/inets/src/http_client/httpc_response.erl
+++ b/lib/inets/src/http_client/httpc_response.erl
@@ -122,17 +122,19 @@ result(Response = {{_, Code, _}, _, _},
                            (Code =:= 303) ->
     redirect(Response, Request#request{method = get});
 result(Response = {{_, Code, _}, _, _}, 
-       Request = #request{settings =
-              #http_options{autoredirect = true},
-              method = post}) when (Code =:= 307) ->
+        Request = #request{settings =
+            #http_options{autoredirect = true},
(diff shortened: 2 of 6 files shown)


lib/inets/doc/src/httpd_util.xml
```diff
@@ -264,7 +264,7 @@
       <name since="">reason_phrase(StatusCode) -> Description</name>
       <fsummary>Returns the description of an HTTP 1.1 status code.</fsummary>
       <type>
-        <v>StatusCode = 100| 200 | 201 | 202 | 204 | 205 | 206 | 300 | 301 | 302 | 303 | 304 | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 410 411 | 412 | 413 | 414 415 | 416 | 417 | 500 | 501 | 502 | 503 | 504 | 505</v>
+        <v>StatusCode = 100| 200 | 201 | 202 | 204 | 205 | 206 | 300 | 301 | 302 | 303 | 304 | 308 | 400 | 401 | 402 | 403 | 404 | 405 | 406 | 410 411 | 412 | 413 | 414 415 | 416 | 417 | 500 | 501 | 502 | 503 | 504 | 505</v>
```
u3s, line 267: While editing this line. Could you maybe fix formatting details ?
1. add space after `100`
2. add | after `410`
3. add | after `414`

lib/inets/src/http_client/httpc_response.erl
```diff
@@ -124,15 +124,17 @@ result(Response = {{_, Code, _}, _, _},
...
     redirect(Response, Request);
 result(Response = {{_, Code, _}, _, _}, 
        Request = #request{settings = 
 			  #http_options{autoredirect = true},
 			  method = Method}) when (Code =:= 301) orelse
 					       (Code =:= 302) orelse
 					       (Code =:= 303) orelse
-					       (Code =:= 307) ->
+					       (Code =:= 307) orelse
+                           (Code =:= 308) ->
```
u3s, line 137: white at it, could you reformat that code block and make nesting little more consistent?

u3s (commented):
it would be great is you could improve formatting a little in sections you've updated.

lib/inets/doc/src/httpd_util.xml
DataTriny: I even splitted the line. Is it OK?

u3s: thanks

IngelaAndin:
We also think that the files in otp/lib/inets/doc/archive are legacy, which we plan to remove, so no need to add files to that directory.

DataTriny:
Hello @IngelaAndin , file removed.

IngelaAndin:
@DataTriny you need to also list the test case in the list of test cases belonging to a test group in the test suite for it to be executed. I would add it along line 183. 

Also, do you aim for this to be fixed in OTP-26? If you want it to be fixed on the OTP-25 track you need to base it on maint.

DataTriny:
Hi @IngelaAndin ,

Sorry for the delay. Yes I forgot to actually run my test... Rebased as well!

PR #6294, opened 2022-09-11
Adjust language around the usage of receive in a gen_statem.

starbelly:
- Change can not to should not and give a brief explanation of why one should not use a receive in a gen_* process.

I suspect this will go through a few revisions. We may want to provide examples or simply expand on all this in another chapter (the dos and donts of gen_*) and link to that. However, I didn't think letting perfect be the enemy of good here was ideal.

### DIFF
--- a/system/doc/design_principles/statem.xml
+++ b/system/doc/design_principles/statem.xml
@@ -1918,7 +1918,7 @@ do_unlock() ->
       Using a receive with a catch all match may result in system messages
       being discarded which in turn may lead to unexpected behaviour.
       If a selective receive must be used then great care should be taken to
-      ensure messages only messages pertitnent to the operation are caught.
+      ensure messages only messages pertinent to the operation are caught.
       Likewise, long running operations may result in timeouts and the
       deferment of system messages, which in turn can lead to undesirable behaviour;
       thus a reasonable timeout for any blocking operation should be used.
--- a/system/doc/design_principles/statem.xml
+++ b/system/doc/design_principles/statem.xml
@@ -1918,7 +1918,7 @@ do_unlock() ->
       Using such a receive may result in system messages
       being discarded which in turn may lead to unexpected behaviour.
       If a selective receive must be used then great care should be taken to
-      ensure messages only messages pertinent to the operation are caught.
+      ensure only messages pertinent to the operation are caught.
       Likewise, long running operations may result in timeouts and the
       deferment of system messages, which in turn can lead to undesirable behaviour;
       thus a reasonable timeout for any blocking operation should be used.
--- a/system/doc/design_principles/statem.xml
+++ b/system/doc/design_principles/statem.xml
@@ -1915,7 +1915,7 @@ do_unlock() ->
       A catch-all receive should never be used from a <c>gen_statem</c>
       behaviour (or from any <c>gen_*</c> behaviour),
       as the receive statement is within the <c>gen_*</c> engine itself.
-      Using a receive with a catch all match may result in system messages
+      Using such a receive may result in system messages
       being discarded which in turn may lead to unexpected behaviour.
       If a selective receive must be used then great care should be taken to
       ensure messages only messages pertinent to the operation are caught.
--- a/system/doc/design_principles/statem.xml
+++ b/system/doc/design_principles/statem.xml
@@ -1912,15 +1912,16 @@ do_unlock() ->
       to implicitly postpone any events to the <c>locked</c> state.
     </p>
     <p>
-      A receive should not be used from a <c>gen_statem</c>
+      A receive all should never be used from a <c>gen_statem</c>
       behaviour (or from any <c>gen_*</c> behaviour),
       as the receive statement is within the <c>gen_*</c> engine itself.
-      Doing so without taking all cases into consideration could result
-      in system messages being discarded leading to unexpected behaviour.
-      Even if all cases are taken into consideration the recvieve would
-      still block system messages from arriving in a timely manor or
-      indefinitely if a timeout is not provided.
-      It must be there because all
+      Using a receive with a catch all match may result in system messages
+      being discarded which in turn may lead to unexpected behaviour.
+      If a selective receive must be used then great care should be taken to
+      ensure messages only messages pertitnent to the operation are caught.
+      Likewise, long running operations may result in timeouts and the
+      deferment of system messages, which in turn can lead to undesirable behaviour;
+      thus a reasonable timeout for any blocking operation should be used.
       <seeerl marker="stdlib:sys"><c>sys</c></seeerl>
       compatible behaviours must respond to system messages and therefore
       do that in their engine receive loop,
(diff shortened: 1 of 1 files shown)


system/doc/design_principles/statem.xml
```diff
@@ -1912,9 +1912,14 @@ do_unlock() ->
       to implicitly postpone any events to the <c>locked</c> state.
     </p>
     <p>
-      A selective receive cannot be used from a <c>gen_statem</c>
+      A receive should not be used from a <c>gen_statem</c>
       behaviour (or from any <c>gen_*</c> behaviour),
       as the receive statement is within the <c>gen_*</c> engine itself.
+      Doing so without taking all cases into consideration could result
+      in system messages being discarded leading to unexpected behaviour.
+      Even if all cases are taken into consideration the recvieve would
```
starbelly, line 1920: The only problem here as I see it, is one may say, well I just use gen_server:call/* and then you still get into a blocking problem. So, it may be best to leave the blocking part out or only comment on the indefinite blocking part.

paulo-ferraz-oliveira, line 1920:
```suggestion
      Even if all cases are taken into consideration the recieve would
```

system/doc/design_principles/statem.xml
```diff
@@ -1912,9 +1912,14 @@ do_unlock() ->
...
+      still block system messages from arriving in a timely manor or
```
paulo-ferraz-oliveira, line 1921:
```suggestion
      still block system messages from arriving in a timely manner or
```

system/doc/design_principles/statem.xml
```diff
@@ -1912,9 +1912,14 @@ do_unlock() ->
...
+      indefinitely if a timeout is not provided.
```
paulo-ferraz-oliveira, line 1922:
```suggestion
      indefinitely if a timeout was not provided.
```

paulo-ferraz-oliveira: For consistency with "would", in the same sentence (?)

lhoguin:
I think both old and new are wrong.

A catch-all receive should not be used, we can all agree on that.

A selective receive may be used but great care should be taken so that it only catches messages that it has to deal with, here I also agree. But blocking is fine and it's in fact impossible or undesirable not to block in a lot of scenarios. For example if you use a `gen_statem` for TCP connections or if you have to perform a long operation in general (not necessarily involving another process). So for that part I would just mention that long-running operations may result in timeouts being triggered and that system messages will not be processed while that is going on.

starbelly:
> I think both old and new are wrong.
> 
> A catch-all receive should not be used, we can all agree on that.
> 
> A selective receive may be used but great care should be taken so that it only catches messages that it has to deal with, here I also agree. But blocking is fine and it's in fact impossible or undesirable not to block in a lot of scenarios. For example if you use a `gen_statem` for TCP connections or if you have to perform a long operation in general (not necessarily involving another process). So for that part I would just mention that long-running operations may result in timeouts being triggered and that system messages will not be processed while that is going on.

Yeah, agreed.

system/doc/design_principles/statem.xml
starbelly: Sorry, @paulo-ferraz-oliveira  you'll have to another pass as I put up a second pass ;)

system/doc/design_principles/statem.xml
```diff
@@ -1912,10 +1912,16 @@ do_unlock() ->
       to implicitly postpone any events to the <c>locked</c> state.
     </p>
     <p>
-      A selective receive cannot be used from a <c>gen_statem</c>
+      A receive all should never be used from a <c>gen_statem</c>
```
paulo-ferraz-oliveira, line 1915:
```suggestion
      A catch-all receive should never be used from a <c>gen_statem</c>
```

system/doc/design_principles/statem.xml
```diff
@@ -1912,10 +1912,16 @@ do_unlock() ->
...
       behaviour (or from any <c>gen_*</c> behaviour),
       as the receive statement is within the <c>gen_*</c> engine itself.
-      It must be there because all
+      Using a receive with a catch all match may result in system messages
```
paulo-ferraz-oliveira, line 1918:
```suggestion
      Using such a receive may result in system messages
```

system/doc/design_principles/statem.xml
```diff
@@ -1912,10 +1912,16 @@ do_unlock() ->
...
+      being discarded which in turn may lead to unexpected behaviour.
+      If a selective receive must be used then great care should be taken to
+      ensure messages only messages pertitnent to the operation are caught.
```
paulo-ferraz-oliveira, line 1921:
```suggestion
      ensure messages only messages pertinent to the operation are caught.
```

system/doc/design_principles/statem.xml
```diff
@@ -1912,10 +1912,16 @@ do_unlock() ->
...
     <p>
-      A selective receive cannot be used from a <c>gen_statem</c>
+      A catch-all receive should never be used from a <c>gen_statem</c>
       behaviour (or from any <c>gen_*</c> behaviour),
       as the receive statement is within the <c>gen_*</c> engine itself.
-      It must be there because all
+      Using such a receive may result in system messages
+      being discarded which in turn may lead to unexpected behaviour.
+      If a selective receive must be used then great care should be taken to
+      ensure messages only messages pertinent to the operation are caught.
```
RaimoNiskanen, line 1921: First "messages" is redundant

Maria-12648430: Btw, this phrasing still leaves a backdoor open, ie messages of the form `{system, ..., ...}` or `{'$gen_call', ..., ...}` _could_ be pertinent to the operation but would also be prone to stealing non-pertinent messages. However, I think that this is unlikely, and that you can't point out every single way in which someone could do something wrong. Just nitpicking then :sweat_smile:

starbelly: Yeah, but I agree. We're jamming a whole lot into a few sentences here That's why I figure a long-term plan might be to add a lot more to a section where longer explanations with lots of examples make sense. 

Either way, I've adjusted a few things, see if it read better now.

system/doc/design_principles/statem.xml
```diff
@@ -1912,14 +1912,20 @@ do_unlock() ->
...
       compatible behaviours must respond to system messages and therefore
       do that in their engine receive loop,
       passing non-system messages to the <em>callback module</em>.
+      Using a catch-all receive may result in system messages
+      being discarded which in turn may lead to unexpected behaviour.
+      If a selective receive must be used then great care should be taken to
+      ensure that only non-system messages which are pertinent to the operation
+      are caught. Likewise, long-running operations may result in timeouts and the
+      deferment of system messages, which in turn can lead to undesirable behaviour;
+      thus, a reasonable timeout for any blocking operation should be used.
```
RaimoNiskanen, line 1928: Suggestion; maybe a simplification:

ensure that only messages pertinent to the operation are received.
Likewise, a callback <em>must</em> return in due time to let the engine
receive loop handle system messages, or they might time out also
leading to unexpected behaviour.

starbelly: Much better :) I'll apply this.

RaimoNiskanen:
Sorry about the delay.  Thank you for your pull request!

PR #6296, opened 2022-09-13
dialyzer: Fix literal value being used instead of type

TD5:
A missing translation of a literal atom from a record name to the corresponding atom type meant Dialyzer/TypEr would crash due to trying to pattern match on a `c_literal` value, where a type was expected.

I suspect this hasn't been noticed yet because it requires specs to be set explicitly to be ignored (which results in the table of records being empty, activating this codepath where the first literal atom in a tuple corresponding to the record name is not converted into a type, and is instead erroneously left as a value-level literal).

### DIFF
--- a/lib/dialyzer/test/typer_SUITE.erl
+++ b/lib/dialyzer/test/typer_SUITE.erl
@@ -20,14 +20,18 @@
 -module(typer_SUITE).
 
 -export([all/0,suite/0,
-         smoke/1, smoke_incremental_plt/1]).
+         smoke/1,
+         smoke_incremental_plt/1,
+         gh_6296_no_spec_flag_does_not_break_records/1]).
 
 -include_lib("common_test/include/ct.hrl").
 
 suite() -> [{ct_hooks,[ts_install_cth]}].
 
 all() ->
-    [smoke,smoke_incremental_plt].
+    [smoke,
+     smoke_incremental_plt,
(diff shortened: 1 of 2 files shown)


bjorng:
Looks good to me. If you'll change the pull request from a draft to a normal pull request, I'll add it to our daily builds.

TD5:
Thanks. It should be fine to add to the daily tests, but I was holding off to make sure my unit testing properly covered everything.

TD5:
> Thanks. It should be fine to add to the daily tests, but I was holding off to make sure my unit testing properly covered everything.

It just so happens I finished this now, so should be all set 👍

bjorng:

lib/dialyzer/test/typer_SUITE.erl
```diff
@@ -51,6 +55,39 @@ smoke(Config) ->
           ok
     end.
 
+otp_6296_no_spec_flag_does_not_break_records(Config) ->
+    Code = <<"-module(otp_6296).
```
bjorng, line 59: One more thing. `otp_xxxx` is for our internally created tickets, not for github issues. To avoid confusion:

```suggestion
gh_6296_no_spec_flag_does_not_break_records(Config) ->
    Code = <<"-module(gh_6296).
```

bjorng:
Thanks!

PR #6297, opened 2022-09-14
stdlib: Add `binary:encode_hex/2`

gilbertwong96:
The `binary` module only exposes `encode_hex/1`, which only supports uppercase output. 

This change adds `binary:encode_hex/2` to support uppercase and lowercase hex-encoding.  After this change, `encode_hex/1` supports uppercase by default.


Maria-12648430:
Something similar is currently being done on the `base64` module, see #6280.

The solution initially proposed there was much like this one here, and it slowed down the encoding and decoding process considerably. So, you might want to benchmark the performance before and after your change here :wink:

After changes suggested by @bjorng were applied, the revised solution was about on par with the current implementation. The same approach should also be applicable for this PR.

lib/stdlib/src/binary.erl
```diff
@@ -365,31 +365,39 @@ get_opts_replace(_,_) ->
...
-    << <<?HEX(N)>> || <<N>> <= Data >>;
 encode_hex(Bin) ->
+    encode_hex(Bin, upper).
+
+-define(HEX(X, Case), (hex(X, Case)):16).
+-compile({inline,[hex/2]}).
+-spec encode_hex(Bin, Case) -> Bin2 when
+      Bin :: binary(),
+      Case :: lower | upper,
+      Bin2 :: <<_:_*16>>.
```
Maria-12648430, line 380: Would `lowercase` / `uppercase` be better namings than `upper` / `lower`? Not sure what the precedents are.

gilbertwong96: Yes, `lowercase`/`uppercase` is better.

gilbertwong96:
> Something similar has recently been done for the `base64` module, see #6280.
> 
> The solution initially proposed there was much like this one here, and it slowed down the encoding and decoding process considerably. So, you might want to benchmark the performance before and after your change here 😉
> 
> After changes suggested by @bjorng were applied, the revised solution was about on par with the then current implementation. The same approach should also be applicable for this PR.

The performance must be the same as the original version. The algorithm hasn't been changed. It just provide new APIs.

bjorng:
> The performance must be the same as the original version. 

Running a modified version of [base64_bench](https://gist.github.com/bjorng/522470f313ac01d16156a597657f97bb) on the `maint` branch, I get the following numbers on my computer:

```
== Testing with 1 MB ==
fun binary:encode_hex/1: 1000 iterations in 9622 ms: 103 it/sec
fun binary:decode_hex/1: 1000 iterations in 41326 ms: 24 it/sec
```

With this PR, there is a slight loss of performance:

```
== Testing with 1 MB ==
fun binary:encode_hex/1: 1000 iterations in 10541 ms: 94 it/sec
fun binary:decode_hex/1: 1000 iterations in 41008 ms: 24 it/sec
```

On the `master` branch, the results are:

```
== Testing with 1 MB ==
fun binary:encode_hex/1: 1000 iterations in 2607 ms: 383 it/sec
fun binary:decode_hex/1: 1000 iterations in 37013 ms: 27 it/sec
```

With this PR rebased on top of the `master` branch, the results are:

```
== Testing with 1 MB ==
fun binary:encode_hex/1: 1000 iterations in 3574 ms: 279 it/sec
fun binary:decode_hex/1: 1000 iterations in 36800 ms: 27 it/sec
```

Maria-12648430:
@bjorng out of interest, if you run the benchmarks with `lowercase` (with these changes applied), does that make it worse? I mean, is the difference in performance more pronounced? Going by gut feeling, it should, as there is now an extra check? (And sorry, can't try myself right now, I'm only on my flimsy mobile :sweat_smile:)

bjorng:
@Maria-12648430 There were no noticeable differences in benchmark results between the two modes.

> Going by gut feeling, it should, as there is now an extra check? 

In general, matching is not done one clause at a time in order. In this case, the selection between the two clauses is done by the following BEAM instruction:

```
    {select_val,{x,2},
                {f,490},
                {list,[{atom,lower},{f,488},{atom,upper},{f,487}]}}.
```

If there are many values in a `select_val` a binary search will be used. When there are only a few, as in this case, the JIT will emit code that does a linear search:

```
# i_select_val_lins_sfI
    mov rsi, qword ptr [rbx+16]
    cmp rsi, 441611
    je label_487
    cmp rsi, 441931
    je label_488
    jmp label_490
```

I expect the most expensive part of this sequence to be the first instruction, which loads a BEAM register into CPU register `rsi`. Comparing `rsi` to a value and not branching is very cheap, relatively speaking. Therefore, it doesn't really matter which of the atoms that is compared first.

Maria-12648430:
@bjorng thanks for the detailed explanation, very much appreciated :smiling_face_with_three_hearts:

juhlig (changes requested):

gilbertwong96:
I improve the performance in some way ugly.

Here is the bench on my computer:

On the master branch, the results are:

```
== Testing with 1 MB ==
fun binary:encode_hex/1: 1000 iterations in 9091 ms: 109 it/sec
fun binary:decode_hex/1: 1000 iterations in 32389 ms: 30 it/sec
```

With this PR rebased on top of the master branch, the results are:

```
== Testing with 1 MB ==
fun binary:encode_hex/1: 1000 iterations in 9060 ms: 110 it/sec
fun binary:decode_hex/1: 1000 iterations in 31815 ms: 31 it/sec
```

juhlig:
I'll take a look at it tomorrow 🙂

juhlig:
Ok. I want to say that my review is inspired (or maybe biased) by similar work recently done on the `base64` module. @bjorng may have a different opinion on this. He is waaay more of an expert than I am. Anyway, here goes :)

<hr />

I would suggest to conflate the functions `hex_uppercase/1` and `hex_lowercase/1` into one function `hex/2`, with the second parameter being an offset. Put both the uppercase and lowercase integers into one big tuple, such that the uppercase integers are in positions 1-256, lowercase in 257-512 (don't mind the duplicates):
```erlang
hex(X, Off) ->
    element(
        X+Off,
        {
            % integers for uppercase
            16#3030, 16#3031, 16#3032, 16#3033, 16#3034, ...
            ..., 16#4642, 16#4643, 16#4644, 16#4645, 16#4646,
            % integers for lowercase
            16#3030, 16#3031, 16#3032, 16#3033, 16#3034, ...
            ..., , 16#6662, 16#6663, 16#6664, 16#6665, 16#6666
        }
    ).
```

So if you want to query for an uppercase integer, you pass in the offset 1, for a lowercase integer you pass in 257:
```erlang
1> hex(16#4a, 257).
13371
2> hex(16#4a, 1).
13409
```

Having only one function for the translation, you can go with only one macro instead of two:
```erlang
-define(HEX(X, Off), (hex(X, Off)):16).
```

And having only one function and macro, you don't need two separate functions `encode_hex_upper/1` and `encode_hex_lower/1` but only one, `encode_hex1/2` (I suffixed the name with a `1` since a function `encode_hex/2` already exists) which takes the data to encode and an offset:
```erlang
encode_hex1(Data, Off) when byte_size(Data) rem 8 =:= 0 ->
    << <<?HEX(A, Off), ?HEX(B, Off), ?HEX(C, Off), ?HEX(D, Off),
         ?HEX(E, Off), ?HEX(F, Off), ?HEX(G, Off), ?HEX(H, Off)>>
       || <<A, B, C, D, E, F, G, H>> <= Data >>;
encode_hex1(Data, Off) when byte_size(Data) rem 7 =:= 0 ->
    << <<?HEX(A, Off), ?HEX(B, Off), ?HEX(C, Off), ?HEX(D, Off),
         ?HEX(E, Off), ?HEX(F, Off), ?HEX(G, Off)>>
       || <<A, B, C, D, E, F, G>> <= Data >>;
...
encode_hex1(Data, Off) when is_binary(Data) ->
    << <<?HEX(N, Off)>> || <<N>> <= Data >>;
encode_hex1(Bin, Off) ->
    badarg_with_info([Bin, Off]).
```

*(Actually, if you define your macro like `-define(HEX(X), (hex(X, Off)):16).`, you can even go without the `Off` in all the `?HEX(...)` usages. However, I'm pretty sure this is bad style :sweat_smile:)*

The proper value for the offset can be determined and passed in from `encode_hex/2`:
```erlang
encode_hex(Bin, uppercase) ->
    encode_hex1(Bin, 1);
encode_hex(Bin, lowercase) ->
    encode_hex1(Bin, 257).
```

<hr />

I think by going like this, the code would be smaller and cleaner (but make sure to add comments explaining what that Offset is about ;)), without any performance penalties. Anyway, I would advise to wait for @bjorng's comment before doing anything :wink:

bjorng (changes requested):
I agree with @juhlig's suggestions.

We don't want to have only property tests; there should always be some tests for if `proper` is not installed. Please update `binary_module_SUITE`.

Make sure that the documentation have the correct name for the options.

gilbertwong96:
Thank you for your suggestions. @bjorng @juhlig . I have fixed the points you mentioned. Please review again.😄

lib/stdlib/doc/src/binary.xml
```diff
@@ -256,6 +256,25 @@
       </desc>
     </func>
 
+    <func>
+      <name name="encode_hex" arity="2" since="OTP 26.0"/>
```
juhlig, line 260: You could conflate the docs for the two functions `encode_hex/1` and `encode_hex/2` into one, and mention in the description that `uppercase` is the default.

lib/stdlib/doc/src/binary.xml
```diff
@@ -256,6 +256,25 @@
...
+      <fsummary>Encode a binary into a hex encoded binary with specified case</fsummary>
+      <desc>
+          <p>Encode a binary into a hex encoded binary with specified case.
+          If the case is <c>lowercase</c>, the <c>a</c> through <c>f</c> in hex encoded binary
+          will be in lower case. If the case is <c>uppercase</c>, the <c>A</c> through <c>F</c>
+          in hex encoded binary will be in upper case.</p>
```
juhlig, line 266:
```suggestion
      <fsummary>Encodes a binary into a hex encoded binary with specified case</fsummary>
      <desc>
          <p>Encodes a binary into a hex encoded binary using the specified case for the
          Hex digits "a" to "f".</p>
```
I think this is clear enough, and the examples below illustrate what will happen.

juhlig (changes requested):
The code is fine now as far as I'm concerned. More comments and suggestions inline ;)

lib/stdlib/test/property_test/binary_prop.erl
```diff
@@ -0,0 +1,19 @@
+-module(binary_prop).
+
+-export([prop_hex/0]).
+
+-include_lib("common_test/include/ct_property_test.hrl").
+
+prop_hex() ->
```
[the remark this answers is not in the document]
gilbertwong96: Yes, `prop_hex_encode_2` is a better name.

lib/stdlib/test/binary_property_test_SUITE.erl
```diff
@@ -0,0 +1,35 @@
+%%
+%% %CopyrightBegin%
+%%
+%% Copyright Gilbert Wong 2021. All Rights Reserved.
```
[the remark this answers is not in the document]
gilbertwong96: Sorry. This is my first contribution to the OTP, and I have no idea how to write a correct copyright for a new file.

lib/stdlib/src/binary.erl
```diff
@@ -365,48 +365,77 @@ get_opts_replace(_,_) ->
...
-encode_hex(Data) when byte_size(Data) rem 3 =:= 0 ->
-    << <<?HEX(A),?HEX(B),?HEX(C)>> || <<A,B,C>> <= Data >>;
-encode_hex(Data) when byte_size(Data) rem 2 =:= 0 ->
-    << <<?HEX(A),?HEX(B)>> || <<A,B>> <= Data >>;
-encode_hex(Data) when is_binary(Data) ->
-    << <<?HEX(N)>> || <<N>> <= Data >>;
 encode_hex(Bin) ->
+    encode_hex(Bin, uppercase).
+
+-spec encode_hex(binary(), uppercase | lowercase) -> <<_:_*16>>.
```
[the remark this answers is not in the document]
gilbertwong96: Done.

lib/stdlib/doc/src/binary.xml
gilbertwong96: Thank you for your advice.

lib/stdlib/doc/src/binary.xml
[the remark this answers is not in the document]
gilbertwong96: I have updated this line to  `<name name="encode_hex" arity="2" since=""/>`. I don't know how to give a placeholder here. I don't find anywhere mention of the `placeholder` in [CONTRIBUTING.md](https://github.com/erlang/otp/blob/master/CONTRIBUTING.md)

lib/stdlib/doc/src/binary.xml
```diff
@@ -256,6 +256,22 @@
       </desc>
     </func>
 
+    <func>
+      <name name="encode_hex" arity="2" since=""/>
```
gilbertwong96, line 260: Could you please tell me how to give a placeholder here? @bjorng

lib/stdlib/doc/src/binary.xml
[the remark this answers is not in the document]
Maria-12648430: You can't put it there at this point because you don't know yet. It is a ticket number that is internal to the OTP team, AFAIK, and will eventually be given to you by the person a PR is assigned to.

lib/stdlib/test/binary_property_test_SUITE.erl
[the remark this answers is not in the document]
Maria-12648430: I usually copy the header from an existing file, like
```erlang
%% Copyright Ericsson AB 2022. All Rights Reserved.
```

lib/stdlib/doc/src/binary.xml
```diff
@@ -256,6 +256,22 @@
       </desc>
     </func>
 
+    <func>
+      <name name="encode_hex" arity="2"/>
+      <fsummary>Encodes a binary into a hex encoded binary with specified case</fsummary>
+      <desc>
+          <p>Encodes a binary into a hex encoded binary using the specified case for the Hex digits "a" to "f".</p>
```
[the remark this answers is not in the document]
gilbertwong96: Done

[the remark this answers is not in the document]
gilbertwong96: I have seen this style first in OTP Documentation. Could you please give the reference for the new style?  @Maria-12648430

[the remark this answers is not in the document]
Maria-12648430: > Could you please give the reference for the new style?

Not sure if you can call it a "new" style, the possibility for it probably existed for quite a while. It is not used everywhere where it could be used, so I guess most of the documentation that was done before this became possible just didn't get reworked.

Anyway, [here in the Reference Manual DTDs](https://www.erlang.org/doc/apps/erl_docgen/refman_dtds.html#%3Cfunc%3E) it says that `<func>` "Contains one **or more** `<name>`...". A place where it is used is for example [the docs for `string:equal/2,3,4`](https://github.com/erlang/otp/blob/7ad716dfdbd674f96d4301cf6dab4f8791fbcdf8/lib/stdlib/doc/src/string.xml#L167-L169), which thereby looks [like this](https://github.com/erlang/otp/blob/7ad716dfdbd674f96d4301cf6dab4f8791fbcdf8/lib/stdlib/doc/src/string.xml#L167-L169) in the compiled documentation.

lib/stdlib/test/property_test/binary_prop.erl
```diff
@@ -0,0 +1,18 @@
...
+-compile([export_all, nowarn_export_all]).
+
+-include_lib("common_test/include/ct_property_test.hrl").
+
+prop_hex_encode_2() ->
+    ?FORALL(Data, data(),
+            begin
+                UpperHex = binary:encode_hex(Data, uppercase),
+                LowerHex = binary:encode_hex(Data, lowercase),
+                binary:decode_hex(LowerHex) =:= Data andalso binary:decode_hex(UpperHex) =:= Data
```
Maria-12648430, line 14: This is nice as far as it goes, and you should keep it, but this only tests that what was hex-encoded can be hex-decoded.  In other words, it is not tested if `encode_hex` produces the correct output, only that `decode_hex` can decode whatever `encode_hex` produced back into the initial binary.

A way you could test correct output (without repeating the actual implementation) could be with a helper function like this:
```erlang
check_hex_encoded(<<I1:4, I2:4, Ins/binary>>, <<U1:8, U2:8, UCs/binary>>, <<L1:8, L2:8, LCs/binary>>) ->
    check_hex_chars_match(I1, U1, L1) andalso
    check_hex_chars_match(I2, U2, L2) andalso
    check_hex_encoded(Ins, UCs, LCs);
check_hex_encoded(<<>>, <<>>, <<>>) ->
    true;
check_hex_encoded(_, _, _) ->
    false.

check_hex_chars_match(X, U, L) when X < 10 ->
    (U =:= $0 + X) andalso (L =:= $0 + X);
check_hex_chars_match(X, U, L) ->
    (U =:= $A + X -10) andalso (L =:= $a + X -10).
```
... and call it from the property like `check_hex_encoded(Data, UpperHex, LowerHex)`.
*(Note that I just typed this off the top of my head, didn't test it, so there may be bugs and/or typos :sweat_smile:)*

I admit this won't be hideously fast (for which reason I would just use the normally-growing `binary()` generator instead of resizing it), but with tests it is more important to be thorough than to be fast :wink:

gilbertwong96: Done

juhlig (changes requested):
IMO, this is fine now. Only some minor indendation issues ;)

Maria-12648430:
Nice, I think this is almost good to go now, only the indendation :wink:

lib/stdlib/doc/src/binary.xml
```diff
@@ -244,15 +244,23 @@
...
 1> binary:encode_hex(&lt;&lt;"f"&gt;&gt;).
-&lt;&lt;"66"&gt;&gt;</code>
+&lt;&lt;"66"&gt;&gt;
+2> binary:encode_hex(&lt;&lt;"/"&gt;&gt;).
+&lt;&lt;"2F"&gt;&gt;
+3> binary:encode_hex(&lt;&lt;"/"&gt;&gt;, lowercase).
+&lt;&lt;"2f"&gt;&gt;
+4> binary:encode_hex(&lt;&lt;"/"&gt;&gt;, uppercase).
+&lt;&lt;"2F"&gt;&gt;
+      </code>
```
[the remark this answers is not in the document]
gilbertwong96: Done

gilbertwong96:
> Nice, I think this is almost good to go now, only the indendation 😉

I have already fixed the indentations you mentioned. What about introducing format tools like [erlfmt](https://github.com/WhatsApp/erlfmt) in [otp](https://github.com/erlang/otp)?

bjorng:
I've been busy the last week or so and I haven't had time to fully review the latest changes. Thanks @juhlig and @Maria-12648430 for the reviewing you've done.

However, when I took a quick look and ran the test suite and attempted to build the documentation, I noticed that the test case `binary_module_SUITE:error_info/1` fails, and that the documentation fails to build.

bjorng:
By the way, I've created a ticket number to use in the `since` attribute: `OTP-18354`

(@juhlig No, I don't think it's mentioned in our contribution guidelines.)

lib/stdlib/src/erl_stdlib_errors.erl
```diff
@@ -1069,6 +1075,8 @@ expand_error(not_atom) ->
     <<"not an atom">>;
 expand_error(not_binary) ->
     <<"not a binary">>;
+expand_error(bad_hex_case) ->
+    <<"not uppercase or lowercase">>;
 expand_error(not_compiled_regexp) ->
```
gilbertwong96, line 1080: > @gilbertwong96
> 
> > However, when I took a quick look yesterday and ran the test suite and attempted to build the documentation, I noticed that the test case `binary_module_SUITE:error_info/1` fails, and that the documentation fails to build.
> 
> The documentation build failure is probably because you removed the `since` tag from your new function altogether. I believe it can be left empty, but its presence is required. Since you now have a ticket number, you should put it in the `since` tag like this `since="OTP @OTP-18354@"`, and that should also solve the build issue.
> 
> About the `error_info` test failure, I have no idea though =^^=

I have fixed the `error_info` test failure now.  @juhlig

gilbertwong96:
image

The stdlib tests ran successfully on my machine, but it failed in [CI].(https://github.com/gilbertwong96/otp/actions/runs/3655307834/jobs/6176652422)

image

I don't know why.

Do you have any ideas?

bjorng:
We know that there are some test cases that tend to fail in github actions. It is not your fault.

lib/stdlib/doc/src/binary.xml
```diff
@@ -244,15 +244,23 @@
 
     <func>
       <name name="encode_hex" arity="1" since="OTP 24.0"/>
-      <fsummary>Encodes a binary into a hex encoded binary.</fsummary>
+      <name name="encode_hex" arity="2" since="OTP @OTP-18354@"/>
+      <fsummary>Encodes a binary into a hex encoded binary with specified case</fsummary>
       <desc>
-      <p>Encodes a binary into a hex encoded binary.</p>
+        <p>Encodes a binary into a hex encoded binary using the specified case for the Hex digits "a" to "f".</p>
```
bjorng, line 250:
```suggestion
        <p>Encodes a binary into a hex encoded binary using the specified case for the hexadecimal digits "a" to "f".</p>
```

bjorng (approved):
I have only one more nit pick in the documentation. Otherwise looks good. When you have fixed that, please squash your commits, rebase on the master branch, and retarget the pull request to master.

lib/stdlib/doc/src/binary.xml
gilbertwong96: Done

bjorng:

bjorng:
Thanks for your pull request!

PR #6298, opened 2022-09-14
JIT: Minor optimizations

bjorng:
This PR improves the code generation of the JIT in several ways.

The most important is eliminating reading from a BEAM register written by the previous BEAM instructions if the contents is already available in a CPU register.

For example, consider:

```
foo(<<X:16>>) ->
    (X band 7) bor 1.
```

The BEAM code for the expression body looks like this:

```
    {gc_bif,'band',
            {f,0},
            3,
            [{tr,{x,2},{t_integer,{0,65535}}},{integer,7}],
            {x,0}}.
    {gc_bif,'bor',{f,0},1,[{tr,{x,0},{t_integer,{0,7}}},{integer,1}],{x,0}}.
```

The result of the `band` operand is written X-register 0 (`{x,0}`). The `bor` operation then reads X-register 0.

The JIT in Erlang/OTP 25 generates the following code:

```
# i_band_ssjd
    mov rsi, qword ptr [rbx+16]
    mov eax, 127
    and rax, rsi
    mov qword ptr [rbx], rax
# i_bor_jssd
    mov rsi, qword ptr [rbx]
    mov eax, 31
    or rax, rsi
    mov qword ptr [rbx], rax
```

Here we can see that the last instruction of `band` stores the result to memory (`mov qword ptr [rbx], rax`). The very first thing `bor` does is to read back the value (albeit to another CPU register).

The new optimization will rewrite the first instruction of `bor` to get the value from the `rax` register:

```
# i_bor_jssd
    mov rsi, rax
    mov eax, 31
    or rax, rsi
    mov qword ptr [rbx], rax
```

But that is not actually the code that will be emitted because there is another optimization in this pull request that avoids loading a small integer operand into a register for logical operations. Here is the actual code for the two BEAM instructions:

```
# i_band_ssjd
    and eax, 127
    mov qword ptr [rbx], rax
# i_bor_jssd
    or rax, 31
    mov qword ptr [rbx], rax
```

As it happened, the result of the instruction before `band` (not shown) was available in the `rax` register. Since one operand for `band` is the integer 127, it can be used as an immediate operand for the `and` instruction. Similarly, the `bor` can also be reduced to two instructions because the result from `band` happened to be available in exactly the register it needed to be.

### DIFF
--- a/erts/emulator/beam/jit/arm/beam_asm.hpp
+++ b/erts/emulator/beam/jit/arm/beam_asm.hpp
@@ -1353,6 +1353,8 @@ protected:
 
     void emit_validate_unicode(Label next, Label fail, arm::Gp value);
 
+    void ubif_comment(const ArgWord &Bif);
+
     void emit_bif_is_eq_ne_exact_immed(const ArgSource &LHS,
                                        const ArgSource &RHS,
                                        const ArgRegister &Dst,
--- a/erts/emulator/beam/jit/arm/beam_asm.hpp
+++ b/erts/emulator/beam/jit/arm/beam_asm.hpp
@@ -1003,6 +1003,12 @@ class BeamModuleAssembler : public BeamAssembler {
      * fragments as if they were local. */
     std::unordered_map<void (*)(), Label> _dispatchTable;
 
+    /* Skip unnecessary moves in load_source(), load_sources(), and
+     * mov_arg(). */
+    size_t last_destination_offset = 0;
+    arm::Gp last_destination_from;
+    arm::Mem last_destination_to;
+
 public:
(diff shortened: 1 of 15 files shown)


garazdawi (approved):
lgtm.

I wonder what the overhead of the `comment` function and its arguments are for when we actually don't emit any log... maybe the compiler is smart enough to optimize it away.

bjorng:
The compiler will not optimize away the call to `ubif2mfa()` because it is in another compilation unit and it can't know whether the call has any side effects.

I have updated the annotation code to test whether logging is enabled before doing anything potentially expensive. I sneaked in a few other improvements and simplifications.

PR #6300, opened 2022-09-14
Add missing type specs in http client (httpc.erl)

### LINKED ISSUE #6245: httpc module is unspecced
**Describe the bug**
I would expect all public functions from OTP have specs - which would make them better for tools like dialyzer and eqwalizer (https://github.com/whatsapp/eqwalizer). The httpc module doesn't have any proper specs.

https://www.erlang.org/doc/man/httpc.html has "pseudo-specs" of no use for practical tooling.

kikofernandez:
This PR adds type specs to the API of the `http_client` app, file `httpc.erl`.
I have updated the XML documentation to use the type specs from `httpc.erl`and this fixes #6245

### DIFF
--- a/lib/inets/doc/src/httpc.xml
+++ b/lib/inets/doc/src/httpc.xml
@@ -78,7 +78,7 @@
       <tag>{data_dir, path()}</tag>
       <item>Directory where the profile
 	can save persistent data. If omitted, all cookies are treated
-	as session cookies.</item>
+	as session cookies. <c>path() = string()</c> represents a file path or directory path.</item>
     </taglist>
 
     <p>The client can be stopped using <seemfa marker="inets#stop/2"><c>inets:stop(httpc, Pid)</c></seemfa> or
@@ -330,15 +330,13 @@
             structure:</p>
 
             <pre>
-              {RequestId, saved_to_file}
-              {RequestId, {error, Reason}}
-              {RequestId, Result}
-              {RequestId, stream_start, Headers}
-              {RequestId, stream_start, Headers, HandlerPid}
-              {RequestId, stream, BinBodyPart}
-            {RequestId, stream_end, Headers}
-            </pre>
-
+  {RequestId, saved_to_file}
+  {RequestId, {error, Reason}}
+  {RequestId, Result}
+  {RequestId, stream_start, Headers}
+  {RequestId, stream_start, Headers, HandlerPid}
+  {RequestId, stream, BinBodyPart}
+  {RequestId, stream_end, Headers}</pre>
 	        </item>
           <tag><c>ipv6_host_with_brackets</c></tag>
           <item>
@@ -413,6 +411,9 @@
       <fsummary>Returns the cookie header that would have been sent when making
       a request to <c><anno>URL</anno></c> using the profile
       <c><anno>Profile</anno></c>.</fsummary>
+      <type_desc variable="Profile">
+		    When started <c>stand_alone</c> only the pid can be used.
+      </type_desc>
       <desc>
         <p>Returns the cookie header that would have been sent when making a
         request to <c><anno>Url</anno></c> using profile
@@ -428,6 +429,9 @@
       <name since="OTP R15B01" name="get_options" arity="1"/>
       <name since="OTP R15B01" name="get_options" arity="2"/>
       <fsummary>Gets the currently used options.</fsummary>
+      <type_desc variable="Profile">
+		    When started <c>stand_alone</c> only the pid can be used.
+      </type_desc>
       <desc>
         <p>Retrieves the options currently used by the client.</p>
       </desc>
@@ -437,6 +441,9 @@
       <name since="OTP R15B02" name="info" arity="0"/>
       <name since="OTP R15B02" name="info" arity="1"/>
       <fsummary>Produces a list of miscellaneous information.</fsummary>
+      <type_desc variable="Profile">
+		    When started <c>stand_alone</c> only the pid can be used.
+      </type_desc>
       <desc>
         <p>Produces a list of miscellaneous information. Intended for debugging.
         If no profile is specified, the default profile is used.</p>
@@ -448,6 +455,9 @@
       <name since="OTP R13B04" name="reset_cookies" arity="0"/>
       <name since="OTP R13B04" name="reset_cookies" arity="1"/>
       <fsummary>Resets the cookie database.</fsummary>
+      <type_desc variable="Profile">
+		    When started <c>stand_alone</c> only the pid can be used.
+      </type_desc>
       <desc>
         <p>Resets (clears) the cookie database for the specified
         <c><anno>Profile</anno></c>. If no profile is specified the default
@@ -459,6 +469,9 @@
       <name since="OTP R13B04" name="request" arity="1"/>
       <name since="OTP R13B04" name="request" arity="2"/>
       <fsummary>Sends a get HTTP request.</fsummary>
+      <type_desc variable="Profile">
+		    When started <c>stand_alone</c> only the pid can be used.
+      </type_desc>
       <desc>
         <p>Equivalent to <seemfa marker="httpc#request/4"><c>httpc:request(get, {Url, []}, [], [])</c></seemfa>.</p>
       </desc>
@@ -536,7 +549,7 @@
 	        </item>
         </taglist>
 
-        <p>Request options (<seetype marker="#option_request">option_request()</seetype>) details:</p>
+        <p>Options (type <seetype marker="#option_request">option_request()</seetype>) details:</p>
         <taglist>
           <tag><c>sync</c></tag>
           <item>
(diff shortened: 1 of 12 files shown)


lib/inets/doc/src/httpc.xml
```diff
@@ -143,517 +72,620 @@
...
-      <fsummary>Sends an HTTP request.</fsummary>
-      <type>
-        <v>Method = method()</v>
-        <v>Request = request()</v>
-        <v>HTTPOptions = http_options()</v>
-        <v>http_options() = [http_option()]</v>
-        <v>http_option() = {timeout,         timeout()} | 
-                           {connect_timeout, timeout()} | 
-                           {ssl,             ssloptions()} | 
-                           {essl,            ssloptions()} |
```
KennethL, line 295: I think this option should be removed from the doc and from the code. It is not documented and it seems like `ssl` and `essl` will have the same effect. I don't know the history of `essl` but now it seems redundant.

lib/inets/src/http_client/httpc.erl
```diff
@@ -58,6 +83,98 @@
 
 -define(DEFAULT_PROFILE, default).
 
+%%%=========================================================================
+%%%  Type specifications
+%%%=========================================================================
+
+%% These are used more than once in this module
```
KennethL, line 90: There are still no exported types from `httpc` or from any other modules in the `inets` app  I think. But if we should export types we must be very careful not to export to many types and we also have to be very careful with the names of the types. I think we have to go through `httpd`as well before we decide about exported types because there might be some types which should be common for the client and the server.

lib/inets/src/http_client/httpc.erl
```diff
@@ -75,85 +193,190 @@ profile_name(Profile) ->
...
+%%       {RequestId, stream_start, Headers}
+%%       {RequestId, stream_start, Headers, HandlerPid}
+%%       {RequestId, stream, BinBodyPart}
+%%       {RequestId, stream_end, Headers}
+%%       '''
+%%
+%%       Default is the pid of the process calling the request function (self()).</dd>
+%%     </dl>
+%%   </dd>
+%% </dl>
```
KennethL, line 349: This chunk of edoc style documentation is new isn't it? But it is redundant as long as we have the same text in the `httpc.xml` file. So if there is no additions here compared with in the xml file I suggest that we skip documentation in the .erl file.

KennethL (changes requested):

lib/inets/src/http_client/httpc.erl
kikofernandez: Got it. I agree with you for this wall of text. However, I am not sure that we should remove minimal (one-liners) documentation from source code. My reasoning is that wall-of-text belongs to XML files, but short and concise info may stay in *.erl files, even if creating a minor duplication because it helps us, developers, understand what the function should do without navigating to another file.

- Do you think we should remove all documentation from source code if it is already present in XML?

lib/inets/doc/src/httpc.xml
kikofernandez: I understand. Changing the code to remove `essl` may have further implications if some of our users were already using this option, since it would be a breaking change. I am not sure that removing `essl` from the code belongs to this PR.
I can, if it is a middle ground, remove the type spec with the `essl` option and hope that dialyzer does not complain about it.
Is this (removal of `essl` type spec) what we want for now?

lib/inets/src/http_client/httpc.erl
KennethL: I agree that we should have short few line descriptions of each function and also other comments explaining what's going on in the code. But not the complete documentation intended for the users of the API.

lib/inets/doc/src/httpc.xml
KennethL: I think this PR should remove {essl,ssloptions() } from the documentation, it is mentioned as a valid option but it is not described.

lib/inets/doc/src/httpc.xml
```diff
@@ -143,517 +72,769 @@
...
+    </datatype>
+    <datatype>
+      <name name="socket_opts"/>
+    </datatype>
+    <datatype>
+      <name name="socket_opt"/>
+      <desc>
+        <p>See the options used by
+        <seeerl marker="kernel:gen_tcp">gen_tcp(3)</seeerl> and
+        <seeerl marker="ssl:ssl">gen_tcp(3)</seeerl>
```
IngelaAndin, line 104: The link test ought to be ssl(3) here!

kikofernandez: Good catch!

lib/inets/src/http_server/httpd.erl
```diff
@@ -83,17 +151,33 @@ info(Pid, Properties) when is_pid(Pid) andalso is_list(Properties) ->
 	    info(Address, Port, Profile, Properties)
     end; 
 
-info(Address, Port) when is_integer(Port) ->    
+info(Address, Port) when is_integer(Port) ->
     info(Address, Port, default).
 
-info(Address, Port, Profile) when is_integer(Port), is_atom(Profile) ->    
+-spec info(Address, Port, Profile) -> HttpInformation when
+      Address :: io:format(),
```
IngelaAndin, line 158: io:format?  did you intend inet:ip_address()?

lib/inets/src/http_server/httpd.erl
```diff
@@ -83,17 +151,33 @@ info(Pid, Properties) when is_pid(Pid) andalso is_list(Properties) ->
...
+info(Address, Port, Profile) when is_integer(Port), is_atom(Profile) ->
     httpd_conf:get_config(Address, Port, Profile);
 
 info(Address, Port, Properties) when is_integer(Port) andalso 
 				     is_list(Properties) ->    
     httpd_conf:get_config(Address, Port, default, Properties).
 
-info(Address, Port, Profile, Properties) when is_integer(Port) andalso 
+-spec info(Address, Port, Profile, Properties) -> HttpInformation when
+      Address :: io:format(),
```
IngelaAndin, line 175: Same comment as above!

lib/inets/src/http_server/httpd.erl
```diff
@@ -45,16 +45,75 @@
...
+                          | {server_name, string()}
+                          | {server_root, path()}
+                          | {document_root, path()}.
+
+-type path() :: file:name_all().
+
+-type httpd_comm_properties() ::
+                            {bind_address, inet:ip_address() | inet:hostname() | any}
+                          | {profile, atom()}
+                          | { socket_type, ip_comm | {ip_comm, proplists:proplist()} | {essl, proplists:proplist()}}
```
IngelaAndin, line 79: I think this is the place where @KennethL  would like the documentation to say ssl instead of essl (that will have the same effect ). It also should say [ssl:ssl_option]  and I think [gen_tcp:option()]  instead of proplists:proplist. To make this change of essl -> ssl I think we ought to change the code in http_transport also to make essl the wrapper and ssl the default as not it is actually the other way around even if that is not what we think is best.

kikofernandez: Got it. I'll see if I can add some more types around. Thanks

lib/inets/doc/src/httpc.xml
```diff
@@ -143,593 +72,809 @@
...
+    <datatype>
+      <name name="socket_opt"/>
+      <desc>
+        <p>See the options used by
+        <seeerl marker="kernel:gen_tcp">gen_tcp(3)</seeerl> and
+        <seeerl marker="ssl:ssl">ssl(3)</seeerl>
+        </p>
+      </desc>
+    </datatype>
+  </datatypes>
```
KennethL, line 115: For each time the `<datatypes> ...</datatypes>` block occurs a subtitle "Data Types" appears in the html doc. This is not what we want. It is better to have only one `<datatypes> </datatypes>` block and inside that using `<datatype_title>` several times like this:
```xml
<datatypes>
  <datatype_title>My first group of datatypes</datatype_title>
  <datatype>... </datatype>
  <datatype>... </datatype>
  ...
  <datatype_title>My second group of datatypes</datatype_title>
  <datatype>... </datatype>
  <datatype>... </datatype>
  ...
</datatypes>
```

KennethL (changes requested):

lib/inets/test/httpc_SUITE.erl
```diff
@@ -2076,7 +2076,7 @@ server_config(http_internal, Config) ->
     ];
 server_config(https, Config) ->
     Ssl = proplists:get_value(test_ssl, Config, ssl),
-    [{socket_type, {Ssl, ssl_config(Config)}} | server_config(http, Config)];
+    [{socket_type, {Ssl, ssl_config(Config)}} | server_config(http, [{test_ssl, essl} | Config] )];
```
kikofernandez, line 2079: if `Ssl =:= ssl` we get an error. This happens in `maint` as well.
If `Ssl =:= essl`, then all tests pass

kikofernandez: @IngelaAndin Do you know if this behaviour is expected in `maint`?

IngelaAndin: You probably ran the modified test case with maint!? I do not see it failing in the nightly builds on maint.

kikofernandez:
I have removed the `essl` option from `socket_type` option creating a new wrapper that takes any `ssl` and replaces it by `essl` and vice versa. This wrapper must be in place until #6380 gets fixed.

kikofernandez:
Closing PR to create one with the correct naming convention for OTP team

PR #6301, opened 2022-09-15
[stdlib, kernel] Ensure that peer node halts

rickard-green:
`peer` nodes failed to halt when the process supervising the control connection crashed. When an alternative control connection was used, this supervision process also quite frequently crashed when the `peer` node was stopped by the node that started it which caused the `peer` node to linger without ever halting.

### DIFF
--- a/lib/stdlib/test/peer_SUITE.erl
+++ b/lib/stdlib/test/peer_SUITE.erl
@@ -45,7 +45,9 @@
     duplicate_name/0, duplicate_name/1,
     old_release/0, old_release/1,
     ssh/0, ssh/1,
-    docker/0, docker/1
+    docker/0, docker/1,
+    cntrl_channel_handler_crash/0, cntrl_channel_handler_crash/1,
+    cntrl_channel_handler_crash_old_release/0, cntrl_channel_handler_crash_old_release/1
 ]).
 
 suite() ->
@@ -56,12 +58,14 @@ shutdown_alternatives() ->
 
 alternative() ->
     [basic, peer_states, cast, detached, dyn_peer, stop_peer,
-     io_redirect, multi_node, duplicate_name] ++ shutdown_alternatives().
+     io_redirect, multi_node, duplicate_name, cntrl_channel_handler_crash,
+     cntrl_channel_handler_crash_old_release | shutdown_alternatives()].
 
 groups() ->
     [
         {dist, [parallel], [errors, dist, peer_down_crash, peer_down_continue, peer_down_boot,
-            dist_up_down, dist_localhost] ++ shutdown_alternatives()},
+                            dist_up_down, dist_localhost, cntrl_channel_handler_crash,
+                            cntrl_channel_handler_crash_old_release | shutdown_alternatives()]},
         {dist_seq, [], [dist_io_redirect,      %% Cannot be run in parallel in dist group
                         peer_down_crash_tcp]},
         {tcp, [parallel], alternative()},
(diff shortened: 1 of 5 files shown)


max-au:
In the original design, `peer` should be halting the node as a consequence of closing the control connection. It is expected that process supervising the control connection is linked to `kernel` processes. Hence crashing that process must crash the entire node.

I think the bug I have in the implementation (`user` process starts unlinked) could be corrected with a smaller change that does not touch `kernel.erl` (and does not introduce formal `peer_supervision`). I believe that just linking `RelayPid` (which is registered as the `user` process) to the one spawned in `start/0` should be enough.

The test case you provided, is it the only observed failure mode? I can try coming up with a smaller fix that will make the test case pass.

rickard-green:
> In the original design, `peer` should be halting the node as a consequence of closing the control connection. It is expected that process supervising the control connection is linked to `kernel` processes. Hence crashing that process must crash the entire node.
> 

This currently does not work. If the process supervising the control connection crash in the alternative control connection cases, it will bring down `user_sup`, but the `user_sup` supervisor is configured as `restart => temporary` (which I think it needs to be), so it will just terminate and no more actions will be taken. That is, the node will linger without ever being taken down. In the dist control connection case, the process supervising the control connection isn't linked to anything, so if that process crash, the same will happen.

> I think the bug I have in the implementation (`user` process starts unlinked) could be corrected with a smaller change that does not touch `kernel.erl` (and does not introduce formal `peer_supervision`). I believe that just linking `RelayPid` (which is registered as the `user` process) to the one spawned in `start/0` should be enough.
> 

No that won't be enough, since it will just bring down `user_sup` as described above. I guess your worry about `kernel.erl` is that it won't be modified and recompiled when starting old nodes? Otherwise I'm not sure why you are worried about modifying kernel. I've modified the pull request so that it starts an "orphan" peer supervision in the case when `peer` is started on an old node. I don't think it is possible to do any better regarding old nodes. The problem will also go away when we release OTP 27 and won't support communication with OTP nodes older than OTP 25 which will have the supervision.

> The test case you provided, is it the only observed failure mode?

Well yes in the sense that `peer` failed to stop the node if the process supervising the control connection crashed. In the tcp case there were a top catch all which was supposed to catch all failures and then halt, but it crashed when doing [`io:format()`](https://github.com/erlang/otp/blob/OTP-25.0.4/lib/stdlib/src/peer.erl#L953) before the halt since it had itself as group leader. Does that (being its [own group leader](https://github.com/erlang/otp/blob/OTP-25.0.4/lib/stdlib/src/peer.erl#L948)) actually have any purpose? In the port case there is no top catch so in that case the process just dies without attempting to halt the node. The same in the dist case even though that code is a bit simpler, so it is not that exposed to unexpected crashes.

By the way, in the dist case the io relaying and the control connection supervision are two different processes. In the other cases one and the same. Is there any specific point to this, or did that just happen? I merged, those two processes into one to make it simpler (only one process to keep track of).

I noted that the supervision of the control channel didn't work when I "peer-ified" global_SUITE (#6303) and it left 5 to 10 lingering nodes after each run. All of these nodes have tcp control connections. What I suspect happened is that the `peer` side tried to write at the control connection at the same time as the mother node stopped it. The write failed, then was caught in the top catch which then failed in the `io:format()`.

> I can try coming up with a smaller fix that will make the test case pass.

I think `peer` should be properly supervised by the kernel supervisor tree, and I don't see how that can be accomplished much simpler than in this PR, but I don't claim that this is the cleanest/nicest way to do it. I'm open to suggestions.

max-au:
I have two thoughts.
1. Instead of adding proper supervision to `kernel`, just link the control connection to one of the critical system processes (e.g. init). In that case crashing control connection would result in a node halting, but without explicit calls to `halt/1` in the `peer` code (I am a bit afraid of making these calls)
2. If there is no such way to link, can we add this child dynamically, from `peer` code? That is, by calling `supervisor:start_child` from the peer (asynchronously, after returning control from the undocumented `user` callback).

You guessed right - old nodes (25 and before) won't have that supervision if we bake it into kernel.erl, but they would if we start it dynamically, or they won't need it if we ensure that lifecycle of the controlling process is tied to some critical kernel process. Maybe 1-liner `link(whereis(init))` would do the trick (although it needs to be something that does not trap exit, or some smarter way to add a "critical" child to init, - I'll take a look when I return to a computer).

rickard-green:
> I have two thoughts.
> 
>     1. Instead of adding proper supervision to `kernel`, just link the control connection to one of the critical system processes (e.g. init). In that case crashing control connection would result in a node halting, but without explicit calls to `halt/1` in the `peer` code (I am a bit afraid of making these calls)
> 

Linking to `init` doesn't work. It ignores such exit signals. The only processes that I've found that would work is `global`'s registrar process and `prim_file`'s helper process. However, if someone "fixes" those processes, `peer`'s supervision will be lost which by no means will be obvious, so I don't think linking to one of those is a good idea (except perhaps in the orphan case; see below).

>     2. If there is no such way to link, can we add this child dynamically, from `peer` code? That is, by calling `supervisor:start_child` from the peer (asynchronously, after returning control from the undocumented `user` callback).
> 

My first attempt was trying to do that from the `user` callback, but that just cause a deadlock. Doing it asynchronously leaves a blind spot where the supervision could fail to start.

> 
> You guessed right - old nodes (25 and before) won't have that supervision if we bake it into kernel.erl, but they would if we start it dynamically, or they won't need it if we ensure that lifecycle of the controlling process is tied to some critical kernel process. Maybe 1-liner `link(whereis(init))` would do the trick (although it needs to be something that does not trap exit, or some smarter way to add a "critical" child to init, - I'll take a look when I return to a computer).

By setting up an explicit supervision for `peer`, it wont be lost by mistake, which may be the case with the other solutions, so I think that is preferable. Also, as we get to OTP 27, the problem with supervision of old nodes will have solved itself since all supported nodes will have `peer` supervision.

It could perhaps be a good idea to link the orphan `peer` supervision to `global`'s registrar. One more safety-belt just in case. A "hack" but it won't be forever, only for two releases.

lib/stdlib/test/peer_SUITE.erl
```diff
@@ -563,3 +564,24 @@ docker(Config) when is_list(Config) ->
...
             peer:stop(Peer)
     end.
+
+user_crash() ->
+    [{doc, "Test that peer node is halted if peer user process crashes"}].
+
+user_crash(Config) ->
+    Opts = case proplists:get_value(connection, Config) of
+               undefined -> #{name => ?FUNCTION_NAME};
+               Conn -> #{connection => Conn, name => ?FUNCTION_NAME}
```
max-au, line 574: Is there a reason to use `?FUNCTION_NAME` instead of `?CT_PEER_NAME`?

The problem it creates, this test is in 3 different groups, but all 3 use the same peer node name, hence 2 tests can never succeed if the first one fails to shut down the node (they will fail peer node boot).

max-au:
After banging a few times against the wall I came down to a solution that feels embarrassingly simple... my attempt is in https://github.com/erlang/otp/pull/6317

I still have a feeling that I must be missing something, but trying your tests for `global` also appears to work. It does feel a tiny bit hacky, but that solution avoids making `kernel_sup` aware of `peer` (which is in `stdlib`).

lib/stdlib/test/peer_SUITE.erl
rickard-green: No reason, I just copied that from another testcase. I'll change that.

rickard-green:
> After banging a few times against the wall I came down to a solution that feels embarrassingly simple... my attempt is in #6317
> 

It will still have the same properties as an orphan supervision, even though it happens to be linked to a process in the supervision tree. If the introduced process terminates, the supervision is lost. Sure, it is unlikely that that process will crash due to a bug, but it can still be terminated.

> I still have a feeling that I must be missing something, but trying your tests for `global` also appears to work. It does feel a tiny bit hacky, but that solution avoids making `kernel_sup` aware of `peer` (which is in `stdlib`).

I don't see it as a good thing not making `kernel_sup` aware of `peer`, but a bad thing. By properly introducing it into the `kernel` supervision tree we avoid the risk of the supervision going away without proper action being taken regardless of how small or big that risk is.

Everything in `stdlib` that needs supervision is supervised via the `kernel` supervison tree, like for example the `timer` server.

max-au:
I think that "orphan supervision" is perfectly good for this specific case. `peer` is expected to be used for testing only, therefore lack of kernel-backed supervision should not cause frequent issues. I also don't see any way for this code to crash due to a bug:
```
receive
    {'EXIT', Self, _Reason} ->
        erlang:halt(0)
end
```
The only way for that to happen would be to send `exit(Pid, kill)` to that process. It can be navigated by trapping exists in control connection and calling `halt` as well, but I think even that isn't strictly necessary.

My main concern with added supervision, it contains a fair amount of code that requires test coverage for various failure cases (e.g. timer code in `init_supervision`, high-pri process behaviour, various potential races). Latest PR version also misses testing for TCP/standard_io connections. 

What do you think of adding just the basic "orphan supervision" from #6317 for "main" branch (it is anyway needed for compatibility reasons) and postponing supervision redesign for "master" branch and OTP 26? I am a bit scared of major `peer` supervision changes in a "maint" release.

rickard-green:
> I think that "orphan supervision" is perfectly good for this specific case. `peer` is expected to be used for testing only, therefore lack of kernel-backed supervision should not cause frequent issues.

This can also be used as an argument for it not being sensitive for larger changes.

> I also don't see any way for this code to crash due to a bug:
> 
> ```
> receive
>     {'EXIT', Self, _Reason} ->
>         erlang:halt(0)
> end
> ```
> 
> The only way for that to happen would be to send `exit(Pid, kill)` to that process.

Yes, all supervision could be handled like this with small risk of things going wrong, but the point of the supervision tree is to prevent the small risk of islands of orphan supervision dying without properly propagating errors. It feels very wrong to introduce an orphan supervision when it is not needed.

> It can be navigated by trapping exists in control connection and calling `halt` as well, but I think even that isn't strictly necessary.
> 
> My main concern with added supervision, it contains a fair amount of code that requires test coverage for various failure cases (e.g. timer code in `init_supervision`, high-pri process behaviour, various potential races).

The timer code was a leftover from a rewrite which I've replaced with an `after` which is very straight forward. The `high` priority is hardly a concern here. This process will do nothing during normal operation. The only time it executes is when it receives an exit signal, then immediately either terminating itself or halting the system, so there is no risk of it starving anything else. The reason why I chose to set the priority to `high` priority was so that it wouldn't be starved if something else on `high` priority was running amok. If you see various potential races somewhere, please point them out. Testing of this code is now in my opinion good (compared to before; very good...). 

> Latest PR version also misses testing for TCP/standard_io connections.
> 

No. I only moved the test-case name atom from the group declaration into the `alternative()` function. Testing of `peer` nodes running old releases were however not tested. I've added tests for that now.

> What do you think of adding just the basic "orphan supervision" from #6317 for "main" branch (it is anyway needed for compatibility reasons) and postponing supervision redesign for "master" branch and OTP 26? I am a bit scared of major `peer` supervision changes in a "maint" release.

I removed the changes I made in the `dist` case (merging the `relay` and `origin_link`), so now only obvious bugs have been fixed in the processes managing the control connection. That is, now more or less only the supervision has been added which in my opinion is quite straight forward. I don't see how this would be a too big of a change for a patch.

PR #6302, opened 2022-09-15
Allow arbitrary binary Logger formatters

eproxus:
Data from formatters that is already in binary format is passed through as-is to the handler, allowing for binary logging. Lists are still assumed to possible contain unicode data that will be converted with unicode:characters_to_binary/1 as before.

This opens up for re-using the Logger framework as is with a simple formatter that e.g. calls term_to_binary/1 for binary logging:

```erlang
-module(my_binlog).
-export([start_binlog/0, format/2]).

start_binlog() ->
    logger:add_handler(default, logger_std_h, #{formatter => {my_binlog, #{}}}).

format(Event, _Config) -> term_to_binary(Event).
```

This will write reports in the Erlang term format to disk.


eproxus:
Took a look at the test logs and couldn't find any obvious relation to my changes. Are the Kernel tests flaky normally? Any hints as to why these particular failures?

garazdawi:
The test failure in `logger_std_h_SUITE` is most likely due to changes in this PR. The others are due to #6314

eproxus:
@garazdawi Thanks, I did spot the Logger test case on Friday but forgot to push. I can rebase on `master` as soon as the fix for the other tests is in.

eproxus:
@garazdawi I tried rebasing on the latest master after the changes in #6314 were merged, but I still see the same errors. Is there anything else I need to do, or are the failures actually related to this change?

garazdawi:
I did a mistake in #6314, should be fixed in maint/master now.

Issue #6255, opened 2022-08-29
Unicode filepaths and ERLC_USE_SERVER


garazdawi:
**Describe the bug**
If I use a unicode filepath and have set ERLC_USE_SERVER=true then any compilation using the server will fail with "no such file or directory.

**To Reproduce**

```sh
export ERLC_USE_SERVER=true
echo "-module(t). -export([t/0]). t() -> ok." > "∑/t.erl"
erlc "∑/t.erl"
erlc "∑/t.erl"
∑/t.erl: no such file or directory
```

**Expected behavior**
The second call to erlc should not fail.

**Affected versions**
I've only tested 25, but probably exists since introduction of the compile server.

frazze-jobb:
The first call, will fail to connect to the compilation server and proceed to starting it but compile without the server. The second call will connect successfully to the compilation server and doing so it will encode the command line arguments into an Erlang term, among the functions ei_string_len is used which does not support unicode

bjorng:
Yes, changing the protocol to use binaries instead of strings should be easy. It **might** be enough to do the change only in `erlc`. Glancing at `erl_compile_server` code, it seems that it can already handle binaries as well as strings.

garazdawi:
Fixed in #6306

PR #6306, opened 2022-09-16
erts: support unicode cmdline in compilation server

frazze-jobb:
Updates compilation server so that it supports input arguments in binary format, and converting them to unicode lists.

More information in https://github.com/erlang/otp/issues/6255

### DIFF
--- a/erts/test/erlc_SUITE.erl
+++ b/erts/test/erlc_SUITE.erl
@@ -28,6 +28,7 @@
          compile_yecc/1, compile_script/1,
          compile_mib/1, good_citizen/1, deep_cwd/1, arg_overflow/1,
          make_dep_options/1,
+         unicode_paths/1,
          features_erlc_describe/1,
          features_erlc_unknown/1,
          features_directives/1,
@@ -56,7 +57,8 @@ groups() ->
 
 tests() ->
     [compile_erl, compile_yecc, compile_script, compile_mib,
-     good_citizen, deep_cwd, arg_overflow, make_dep_options].
+     good_citizen, deep_cwd, arg_overflow, make_dep_options,
+     unicode_paths].
 
 feature_tests() ->
     [features_erlc_describe,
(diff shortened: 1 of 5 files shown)


erts/test/erlc_SUITE.erl
```diff
@@ -487,6 +489,25 @@ make_dep_options(Config) ->
...
+        _ ->
+            DepRE = ["_OK_"],
+            {SrcDir,OutDir,Cmd0} = get_cmd(Config),
+            OutDir2 = OutDir++"/😀",
+            os:cmd("mkdir "++OutDir2),
+            Cmd = Cmd0 ++ " +brief -o "++OutDir2,
+            FileName = filename:join(SrcDir, "😀/erl_test_unicode.erl"),
+            BeamFileName = filename:join(OutDir2, "erl_test_unicode.beam"),
+            run(Config, Cmd, FileName, "", DepRE),
+            %run(Config, Cmd, FileName, 0, 0), %% Do we need to run this twice to acctually use the compile server?
```
frazze-jobb, line 505: will remove the comment

lib/kernel/src/erl_compile_server.erl
```diff
@@ -145,8 +145,9 @@ do_compile(ErlcArgs, Cwd, Enc) ->
             {error, StdOutput, StdErrorOutput}
     end.
 
-parse_command_line(#{command_line := CmdLine, cwd := Cwd}) ->
-    parse_command_line_1(CmdLine, Cwd, [], []).
+parse_command_line(#{command_line := CmdLine, cwd := Cwd, encoding := Enc}) ->
+    CmdLine2 = lists:map(fun(A) -> unicode:characters_to_list(A, Enc) end, CmdLine),
+    parse_command_line_1(CmdLine2, unicode:characters_to_list(Cwd, Enc), [], []).
```
bjorng, line 150:
```suggestion
parse_command_line(#{command_line := CmdLine0, cwd := Cwd, encoding := Enc}) ->
    CmdLine = lists:map(fun(A) -> unicode:characters_to_list(A, Enc) end, CmdLine0),
    parse_command_line_1(CmdLine, unicode:characters_to_list(Cwd, Enc), [], []).
```

erts/test/erlc_SUITE.erl
```diff
@@ -487,6 +489,25 @@ make_dep_options(Config) ->
...
+    case os:type() of
+        {win32, _} -> {skip, "Unicode on windows is tricky"};
+        {win64, _} -> {skip, "Unicode on windows is tricky"};
+        _ ->
+            DepRE = ["_OK_"],
+            {SrcDir,OutDir,Cmd0} = get_cmd(Config),
+            OutDir2 = OutDir++"/😀",
+            os:cmd("mkdir "++OutDir2),
+            Cmd = Cmd0 ++ " +brief -o "++OutDir2,
+            FileName = filename:join(SrcDir, "😀/erl_test_unicode.erl"),
```
bjorng, line 502:
```suggestion
            FileName = filename:join([SrcDir, "😀", "erl_test_unicode.erl"]),
```

bjorng (commented):
Looks good to me, except for a few nitpicks.

garazdawi:
I had hoped that the 7 test cases that fail in kernel would be fixed by this change… I wonder why they are not…

frazze-jobb:
Were they being skipped before? Did my commit break them?

garazdawi:
> Were they being skipped before? Did my commit break them?

I filed #6255 because I was investigating failures in those test cases and found that they failed because of using unicode paths. Looking closer now I can see that those test cases use older Erlang versions, so this fix does not effect them yet...

I opened #6314 to fix the problem with older versions so you can ignore those failures in this PR.

PR #6308, opened 2022-09-16
ssl: clarify docs for PEM cache

u3s:


lib/ssl/doc/src/ssl.xml
```diff
@@ -1054,8 +1054,15 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
-      certificates are used during server authentication and when building the
-      client certificate chain.</p>
+        certificates are used during server authentication and when building the
+        client certificate chain.</p>
+        <note><p>When PEM caching is enabled, files provided with
+        this option will be checked for updates at fixed time interval specified with
+        <seeapp marker="ssl:ssl_app#configuration">ssl_pem_cache_clean</seeapp> environment parameter.</p>
+        </note>
+        <note><p>Built-in PEM caching mechanism can be replaced with the use of
+        <seetype marker="#client_cacerts">client_cacerts</seetype> option.</p>
```
Maria-12648430, line 1064:
```suggestion
        this option will be checked for updates at fixed time intervals specified by the
        <seeapp marker="ssl:ssl_app#configuration">ssl_pem_cache_clean</seeapp> environment parameter.</p>
        </note>
        <note><p>The built-in PEM caching mechanism can be replaced with the
        <seetype marker="#client_cacerts">client_cacerts</seetype> option.</p>
```

lib/ssl/doc/src/ssl.xml
```diff
@@ -1258,6 +1265,13 @@ fun(srp, Username :: binary(), UserState :: term()) ->
 	client when a certificate is requested. Can be omitted if
 	there is no need to verify the client and if there are no
 	intermediate CAs for the server certificate.</p>
+        <note><p>When PEM caching is enabled, files provided with
+        this option will be checked for updates at fixed time interval specified with
+        <seeapp marker="ssl:ssl_app#configuration">ssl_pem_cache_clean</seeapp> environment parameter.</p>
+        </note>
+        <note><p>Built-in PEM caching mechanism can be replaced with the use of
+        <seetype marker="#server_cacerts">server_cacerts</seetype> option.</p>
```
Maria-12648430, line 1273:
```suggestion
        this option will be checked for updates at fixed time intervals specified by the
        <seeapp marker="ssl:ssl_app#configuration">ssl_pem_cache_clean</seeapp> environment parameter.</p>
        </note>
        <note><p>The built-in PEM caching mechanism can be replaced with the
        <seetype marker="#server_cacerts">server_cacerts</seetype> option.</p>
```

lib/ssl/doc/src/ssl.xml
```diff
@@ -1610,12 +1624,14 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
       <name name="clear_pem_cache" arity="0" since="OTP 17.5"/>
-      <fsummary> Clears the pem cache</fsummary>
+      <fsummary>Clears the pem cache</fsummary>
       <desc><p>PEM files, used by ssl API-functions, are cached. The
       cache is regularly checked to see if any cache entries should be
       invalidated, however this function provides a way to
       unconditionally clear the whole cache.
       </p>
+      <warning><p>Calling this function ensures that cache becomes empty and
+      previously cached data will not be used for following connections.</p></warning>
```
Maria-12648430, line 1634: I don't think this is necessary, at least not as a warning. It sort of says what the previous paragraph already says, so boils down to a warning that it will do what you want :wink:

lib/ssl/doc/src/ssl_app.xml
```diff
@@ -139,9 +139,9 @@
 	<p>
 	  Number of milliseconds between PEM cache validations. Defaults to 2 minutes.
 	</p>
+        <p>Note: cache can be cleared by calling
```
Maria-12648430, line 142:
```suggestion
        <p>Note: The cache can be cleared by calling
```

Maria-12648430 (changes requested):

u3s:
@Maria-12648430 thanks for reviewing PR!

lib/ssl/doc/src/ssl.xml
u3s: You're right. I tried reformulating it. Is it better now?

lib/ssl/doc/src/ssl.xml
```diff
@@ -1610,12 +1624,14 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
-      <fsummary> Clears the pem cache</fsummary>
+      <fsummary>Clears the pem cache</fsummary>
       <desc><p>PEM files, used by ssl API-functions, are cached. The
       cache is regularly checked to see if any cache entries should be
       invalidated, however this function provides a way to
       unconditionally clear the whole cache.
       </p>
+      <note><p>When PEM cache is enabled, calling this function ensures that current file content specified
+      with <c>cacertfile</c> option is used for following connection.</p></note>
       </desc>
```
Maria-12648430, line 1635: Sorry, it still sounds somewhat awkward (to me) :sweat_smile:

I don't think it should be separated out into a note (which is like "Oh, by the way, $this is what will also happen"), either. It describes what the function does, from a slightly different point of view, not a (possibly unexpected) side effect.

Consider this suggestion?
```suggestion
      <desc>
        <p>PEM files, used by ssl API-functions, are cached for
          performance reasons. The cache is automatically checked at
          regular intervals to see if any cache entries should be invalidated.</p>
        <p>This function provides a way to unconditionally clear the entire
          cache, thereby forcing a reload of previously cached PEM files.</p>
      </desc>
```

lib/ssl/doc/src/ssl.xml
```diff
@@ -1054,8 +1054,15 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
 	<p>Path to a file containing PEM-encoded CA certificates. The CA
-      certificates are used during server authentication and when building the
-      client certificate chain.</p>
+        certificates are used during server authentication and when building the
+        client certificate chain.</p>
+        <note><p>When PEM caching is enabled, files provided with
+        this option will be checked for updates at fixed time intervals specified by the
+        <seeapp marker="ssl:ssl_app#configuration">ssl_pem_cache_clean</seeapp> environment parameter.</p>
+        </note>
+        <note><p>The built-in PEM caching mechanism can be replaced with the
```
IngelaAndin, line 1063: This reads really strange to me. If we directly provide the file content as a binary instead of embedding it in a PEM file there is nothing to be cached. So the note I think should reflect this.

u3s:
I needed to rename the branch name. #6360 replaces this one

PR #6310, opened 2022-09-17
wx: fix build issue when GL/glu.h is not present

tcrossland:
Currently the build of wx fails if `GL/gl.h` is present but `GL/glu.h` is not. This may happen, for example, in macOS if the homebrew `mesa` package is installed.

This change ensures that both `GL/gl.h` and `GL/glu.h` are present before including them.

### DIFF
--- a/lib/wx/c_src/egl_impl.h
+++ b/lib/wx/c_src/egl_impl.h
@@ -29,10 +29,10 @@
 #include <windows.h>  
 #include <gl/gl.h>
 #include <gl/glu.h>
-#elif defined(HAVE_GL_GL_H)
+#elif defined(HAVE_GL_GL_H) && defined(HAVE_GL_GLU_H)
 #include <GL/gl.h>
 # include <GL/glu.h>	
-#elif defined(HAVE_OPENGL_GL_H)
+#elif defined(HAVE_OPENGL_GL_H) && defined(HAVE_OPENGL_GLU_H)
 #include <OpenGL/gl.h> 
 #include <OpenGL/glu.h> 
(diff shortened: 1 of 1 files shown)


dgud:
Can you make it so that only **glu** is skipped if the header is not there, 
since glu is kind of unusable if a modern opengl is used.

Also please rebase on top of maint branch so that it can be included in the next maintenance release

tcrossland:
> Can you make it so that only **glu** is skipped if the header is not there, since glu is kind of unusable if a modern opengl is used.

I'm not clear on the specifics of different GL / OpenGL headers on different platforms, so the idea behind this PR was simply to prevent build failures by ensuring that the `glu.h` file was detected by `configure` before it is included, without breaking anything else or changing the order of the `#ifdef`.

In my particular scenario, the build was failing on macOS if the Homebrew `mesa` formula was installed (but not `mesa-glu`), so `GL/gl.h` (from `mesa`), `OpenGL/gl.h` and `OpenGL/glu.h` (from Xcode Command Line Tools macOS 12.3 SDK OpenGL.framework) were present but not `GL/glu.h`. In this scenario, with this patch, the `OpenGL` headers can be used, the build succeeds, wx and observer work.

Having said that, it's worth nothing that OpenGL was deprecated in macOS 10.14, so I some guidance on building otp on macOS will probably be needed in the near future. Maybe Hombrew `mesa` and `mesa-glu` is the way to go?

> Also please rebase on top of maint branch so that it can be included in the next maintenance release

Done.

dgud:
Thanks

PR #6317, opened 2022-09-21
[peer] Link peer node lifecycle to control connection

max-au:
Crashing controller process should always lead to halting the entire node. Before this change it was possible for controlling process to crash itself and user_sup, but not the node.


rickard-green:
Closing this since #6301 was released in [OTP 25.1.1](https://www.erlang.org/patches/otp-25.1.1).

Issue #6307, opened 2022-09-16
verify_peer warning message is unclear/confusing


danj3:
The warning message produced at this line (in maint-25):

https://github.com/erlang/otp/blob/c028b855ee3f12c835ff3538a90ac5dbc155631b/lib/ssl/src/ssl.erl#L2877

"Option {verify, verify_peer} and cacertfile/cacerts is missing"

Is less clear than is beneficial as it seems to be triggered by missing {verify, verify_peer} rather than cacerts being missing. This is partly a defaults expectation (of this developer), but I believe can be simply addressed if this message is triggered specifically by verify missing (as I understand the code). 

Suggesting the text be changed to (something like):

"Option {verify, verify_peer} not provided, (cacertfile/cacerts would be unused if provided)"

This message appeared in otp-24 and exists in maint-24 as well.

Some additional rationale: as SSL options typically come through multiple layers, and there are many ssl options, it is easy to misunderstand (and get wrong) what options are in effect and have what effect. The above message is greatly appreciated as a point to clarify the function of verify, but pointing the developer more precisely at the problem is the point here.

IngelaAndin:
Well, actually we are thinking about changing the default, in OTP-26,  of verify-option to verify_peer and then failing if not also "cacerts" or "cacertfiles" also is specified. In the spirit of always making the user actively choose the less secure options. This warning is a backward compatible compromise to make sure that users do not think that they have verified something they have not as currently the verify option defaults to verify_none. The intent of the warning is to specify that both verify_peer and one of the options cacertfile or cacerts are needed to get the verification.  We will consider your suggestion.

KennethL:
Could an alternative solution in OTP 26 be to let `{verify,verify_peer}` be implicitly set if one of the options `cacerts` or `cacertfiles`are given?

IngelaAndin:
@KennethL hum .. setting options implicitly might be surprising to the user! Also, there can still be uses for the certs in these options without verifying the peer. It could of course have been made differently could we start over, but now we have considered how it happens to work.

KennethL:
If the cacert options makes sense also in other situations without verify_peer then of course this does not make sense. But if the 2 options mutually dependent on each others existence then in might make sense. (Of course not by introducing incompatibilities).

PR #6319, opened 2022-09-21
ssl: Clarify warning

IngelaAndin:
Closes #GH-6307

### DIFF
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -2873,8 +2873,9 @@ add_filter(Filter, Filters) ->
 maybe_client_warn_no_verify(#{verify := verify_none,
                              warn_verify_none := true,
                              log_level := LogLevel}, client) ->
-            ssl_logger:log(warning, LogLevel, #{description => "Authenticity is not established by certificate path validation",
-                                                reason => "Option {verify, verify_peer} and cacertfile/cacerts is missing"}, ?LOCATION);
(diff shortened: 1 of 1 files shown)


lib/ssl/src/ssl.erl
```diff
@@ -2873,8 +2873,9 @@ add_filter(Filter, Filters) ->
 maybe_client_warn_no_verify(#{verify := verify_none,
                              warn_verify_none := true,
                              log_level := LogLevel}, client) ->
-            ssl_logger:log(warning, LogLevel, #{description => "Authenticity is not established by certificate path validation",
-                                                reason => "Option {verify, verify_peer} and cacertfile/cacerts is missing"}, ?LOCATION);
+    ssl_logger:log(warning, LogLevel, #{description => "Authenticity is not established by certificate path validation",
+                                        reason => "Option {verify, verify_peer} together with one of the options cacertfile or "
+                                        "cacerts is needed to accomplish this"}, ?LOCATION);
```
RaimoNiskanen, line 2878: Proposal, to clarify that authenticity isn't verified, since the text now may suggest that it isn't verified by certificate path validation, but maybe some other way, who knows...

Description: Server authenticity is unverified since certificate path validation is not enabled
Reason: Option {verify, verify_peer}, and one of 'cacertfile' or 'cacerts', are required to enable this

IngelaAndin: Thank you for the suggestion, pushed a new version.

PR #6321, opened 2022-09-22
Fix when ASN.1 module name does not match file

andreasanta:
When a module identifier does not match the file name, the `.hrl` and the `.erl` files are generated with the same name as the input file. However, the include directive in the `.erl` file uses the module identifier, resulting in a compilation error. This fixes the issue.

See example error, the module identifier in the ASN file is `Modkeystoretypes`:

```
erlc mod_keystore_types.asn
/src/mod_keystore_types.erl:8:10: can't find include file "Modkeystoretypes.hrl"
%    8| -include("Modkeystoretypes.hrl").
%     |          
```

### DIFF
--- a/lib/asn1/src/asn1ct_gen.erl
+++ b/lib/asn1/src/asn1ct_gen.erl
@@ -1325,7 +1325,7 @@ gen_head(#gen{options=Options}=Gen, Mod, Hrl) ->
          ]),
     case Hrl of
 	0 -> ok;
-	_ -> emit(["-include(\"",Mod,".hrl\").",nl])
+	_ -> emit(["-include(\"",get(outfile),".hrl\").",nl])
     end,
     Deterministic = proplists:get_bool(deterministic, Options),
     Options1 =


KennethL:
If the ASN.1 module name and the filename it is defined in does not match you will get other troubles as well so I don't think this is the right way to fix the problem (if it is a problem).
Take this example:
We have the file Suck.asn1 with the contents:
```asn1
Seq DEFINITIONS IMPLICIT TAGS ::= 
BEGIN
...
```
Compiling this with `asn1ct:compile("Suck.asn1") will result in:
```
./Suck.beam: Module name 'Seq' does not match file name 'Suck'
{error,none}
```
This error printout comes from the Erlang compiler which requires the Erlang module and the file it is defined in to have the same name.

So I think there are things that can be improved in this area but this could for example be that the ASN.1 compiler refuses to accept that the ASN.1 module and the file it is defined in have different names.

KennethL:
Another comment is that if the generated include files have different names from the ASN.1 modules which is their source then how would the user know which files to include in his module creating records and calling the encode and decode functions?

andreasanta:
Thanks @KennethL I did not account for BEAM modules being generated with the module name also.

Concerning the include statement, in our case it's pretty simple: we do not care about the ASN.1 module identifier, we include the generated hrl/erl files, and we would like the source file to have the same name.

So for instance, we can start with:

this_file_sux.asn

And we can obtain

```
this_file_sux.hrl
this_file_sux.erl
this_file_sux.asn1db
```

So we always know which module to include, given it **always** matches the filename.

Maybe allowing some special characters such as underscores and lowercase names in the ASN.1 module identifier can help solve this?

KennethL:
I don't really understand your problem with naming the file exactly the same as the ASN.1 module that is defined inside the file.
With that approach you will get a generated `.erl` and `.hrl` file which corresponds to the ASN.1 module and the Erlang module inside the `.erl` file.

Note that the `.asn1db` file is only used by the asn1 compiler if one ASN.1 module import types from another ASN.1 module. It is not needed in runtime or when compiling the generated `.erl` files.

The format of the ASN.1 module identifier is specified in the standard (X.680) so we don't want to change that.

KennethL:
You can accomplish the thing you want with the multi file compilation feature https://www.erlang.org/doc/apps/asn1/asn1_getting_started.html#multi-file-compilation

Create a file `this_file_sux.set.asn`. Inside that file you list the files with ASN.1 modules (one or several) like this:
```
Seq.asn1
```
Then you compile that file like this `erlc this_file_sux.set.asn` and the following files will be generated:
```
this_file_sux.asn1db % You don't need this
this_file_sux.erl
this_file_sux.hrl
this_file_sux.beam
```

andreasanta:
Sorry for the late reply. I have tried your suggestion with the multiple files compilation, and it does not accept the asn1 modules, for instance

`key_attestation_types.asn1`

My objective is to have everything consistently named as I explained, like this, since I have other files with the same naming convention:

```
this_file_sux.hrl
this_file_sux.erl
```

I would like to avoid having a file that is called `ModuleTest.hrl` but `module_test.hrl` instead. Not sure if I managed to explain correctly :(

KennethL:
I think the method using so called multi-file compilation like I described above will work for you.
You create one `this_file_sux.set.asn` file and inside that one you have a list of filenames to your ASN.1 files. 
When you compile you will get the generated files with the name `this_file_sux.set.erl`.

But you cant have a file named `key_attestation_types.asn1` since that filename must still correspond to the ASN.1 module declared inside the file.

KennethL:
And underscore `_` is not an allowed character in an ASN.1 module name

andreasanta:
@KennethL I managed to solve this with an awkward sequence of Bash commands and the multi-file compilation you suggested, so I am closing the PR for now. It would be interesting to find a way to make this simple and have the module name *not* correspond to the ASN1 module name in the future. Thanks.

PR #6336, opened 2022-09-29
rand: Slightly optimize the mwc59 family of functions

bjorng:
Recent improvements in the compiler (92611e5d2d75) and the JIT (f8a9ccda6c4a) allows us to write more natural code using guard tests instead of `band` operations to establish ranges for integers. This also happens to be slightly faster.


RaimoNiskanen (approved):
Looks fine to me

PR #6338, opened 2022-09-29
ets: improve ets:new/2 already_exists named_table error

### LINKED ISSUE #6252: ets:new/2 with named_table option - confusing behavour
**Describe the bug**
A call to create a named ets table with an existing name produces very confusing error message: `argument 2: invalid options` instead of providing more user-friendly and understandable message that a table with the given name already exists.

~~Moreover: it destroys already existing table and data.~~

**To Reproduce**

```
% erl
Erlang/OTP 25 [erts-13.0.4] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns] [dtrace]

Eshell V13.0.4  (abort with ^G)
1> ets:whereis(a).
undefined
2> ets:new(a, [named_table]).
a
3> ets:whereis(a).
#Ref<0.2920346062.238682117.68831>
4> ets:new(a, [named_table]).
     in function  ets:new/2
        called as ets:new(a,[named_table])
        *** argument 2: invalid options
5> ets:whereis(a).
undefined
```

**Expected behavior**

1. Error message that the table already exists.
2. ~~Not dropping the existing table.~~

**Affected versions**
25.0.4

gonzalobf:
Adds extra information in `erl_db.c` before returning the error for already used table name. The error is only returned if the option `named_table` is give and the name is already in use.

The extra information is used later to format the error message.

Closes #6252

### DIFF
--- a/erts/emulator/beam/atom.names
+++ b/erts/emulator/beam/atom.names
@@ -90,6 +90,7 @@ atom allocator
 atom allocator_sizes
 atom alloc_util_allocators
 atom allow_passive_connect
+atom already_exists
 atom already_loaded
 atom amd64
 atom anchored
--- a/erts/emulator/beam/erl_db.c
+++ b/erts/emulator/beam/erl_db.c
@@ -57,6 +57,7 @@
 #define EXI_POSITION am_position /* The position is out of range. */
 #define EXI_OWNER    am_owner	 /* The receiving process is already the owner. */
 #define EXI_NOT_OWNER am_not_owner /* The current process is not the owner. */
+#define EXI_ALREADY_EXISTS am_already_exists /* The table identifier already exists. */
 
 #define DB_WRITE_CONCURRENCY_MIN_LOCKS 1
 #define DB_WRITE_CONCURRENCY_MAX_LOCKS 32768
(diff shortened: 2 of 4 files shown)


gonzalobf:
I tried only changing the function in `erl_stdlib_errors`  by adding something like `ets:whereis(Name)` so I could check if the name was already in use, but the option only works if the `ets` is created in another process. If both tables are created under the same process, when the `format_ets_error` function is called the first table has been already destroyed and doesn't exist anymore.

Since I'm touching `erl_db.c` and adding an atom, I suspect it shouldn't be based on `maint` but let me know if I'm wrong and I'll change it.

sverker:
Thanks for the contribution.

I have pushed three additional commits here:
https://github.com/sverker/otp/tree/gonzalobf/gonzalo/ets/already-exists/GH-6252

I propose you
1. Add them to your branch for this PR (I did not have access rights to do so).
2. Remove your `ets_SUITE:already_exists`, I think it was made redundant by my extension of `ets_SUITE:error_info`.
3. If you are satisfied, squash all of it into one commit.
4. Rebase the resulting branch on maint: `git rebase --onto maint master`
5. Edit the PR to be based on maint.

Bug fixes are normally done on maint, especially if they are low risk and without backward compatibility problems.

gonzalobf:
@sverker Thank you for your help. Your changes make sense so I added to my branch and squash them. I also deleted the ets_SUITE test and changed the PR be based in `maint` as you suggested.

I'm happy to do more modifications if there is something missing or anything that can be improved

sverker:
@bjorng I will merge this unless you see any problems.

sverker:
Merged to maint for OTP 25.2

Thanks for the contribution.

PR #6340, opened 2022-10-01
erts: Decouple use of single-mapped memory from `perf` support

frej:
Running the JIT with single-mapped RWX memory for JIT:ed machine code is needed for `perf` support. But as it is also useful for running under QEMU user mode emulation [1] this patch adds a new flag `+JMsingle bool` which controls the use of single-mapped RWX memory without triggering the output of perf-related metadata.

The naming of the flag is intended to follow the (perceived) pattern introduced by the `+JPperf`-flag, that is, `J` is a JIT-related option and `M` stands for memory related things.

Personally I use the additional patch below in order to use the same file system image for development on an amd64 build host and for deployment to an aarch64-target board. The patch detects when the VM is running using QEMU and if so, enables single-mapped memory. As the patch only works on Linux and doesn't try to identify the version of QEMU (if the bug gets fixed), I don't think it belongs to OTP master. But as it is nevertheless useful, I include it here:

```
diff --git a/erts/emulator/beam/erl_init.c b/erts/emulator/beam/erl_init.c
index 42216da0b2..6f11c7586b 100644
--- a/erts/emulator/beam/erl_init.c
+++ b/erts/emulator/beam/erl_init.c
@@ -2408,6 +2408,40 @@ erl_start(int argc, char **argv)
 	erts_usage();
     }

+
+#if defined(HAVE_LINUX_PERF_SUPPORT)
+    /* We want to detect if we are running under QEMU user mode
+       emulation in order to activate single-mapped RWX memory for the
+       JIT:ed code. This only works on Linux, so we use
+       HAVE_LINUX_PERF_SUPPORT to conditionally compile this piece of
+       detection code.
+
+       To detect QEMU user mode we make use of the fact that our
+       parent process will always be QEMU, we detect that by looking
+       up our parent process and then checking if the final path
+       component of the symlink /proc/<parent>/exe starts with
+       "qemu-".
+    */
+    {
+        char symlink_buf[21 + 11]; /* space for any 64 bit integer
+                                      + /proc/%d/exe and terminator */
+        char target_buf[MAXPATHLEN];
+        ssize_t l;
+        erts_snprintf(symlink_buf, sizeof(symlink_buf),
+                      "/proc/%d/exe", getppid());
+        l = readlink(symlink_buf, target_buf, sizeof(target_buf));
+        if (l > 0 && l != sizeof(target_buf)) {
+            char *last_path_separator;
+
+            target_buf[l] = 0;
+            last_path_separator = memrchr(target_buf, '/', l);
+            if (last_path_separator && strncmp(last_path_separator + 1,
+                                               "qemu-", 5) == 0)
+                erts_jit_single_map = 1;
+        }
+    }
+#endif
+
 /* Output format on windows for sprintf defaults to three exponents.
  * We use two-exponent to mimic normal sprintf behaviour.
  */
```

[1] There is a bug in QEMU,
    see https://gitlab.com/qemu-project/qemu/-/issues/1034

### DIFF
--- a/erts/emulator/beam/jit/beam_jit_main.cpp
+++ b/erts/emulator/beam/jit/beam_jit_main.cpp
@@ -42,6 +42,8 @@ ErtsFrameLayout ERTS_WRITE_UNLIKELY(erts_frame_layout);
 #ifdef HAVE_LINUX_PERF_SUPPORT
 int erts_jit_perf_support;
 #endif
+/* Force use of single-mapped RWX memory for JIT code */
+int erts_jit_single_map = 0;
 
 /*
  * Special Beam instructions.
@@ -120,6 +122,13 @@ static JitAllocator *create_allocator(JitAllocator::CreateParams *params) {
     void *test_ro, *test_rw;
     Error err;
 
+#if defined(__APPLE__) && defined(__aarch64__)
+    /* Using a single map will not work on Apple Silicon. */
+    if (params->options == JitAllocatorOptions::kNone) {
+        return nullptr;
+    }
+#endif
+
     auto *allocator = new JitAllocator(params);
 
     err = allocator->alloc(&test_ro, &test_rw, 1);
@@ -140,16 +149,19 @@ static JitAllocator *pick_allocator() {
 #if defined(HAVE_LINUX_PERF_SUPPORT)
     /* `perf` has a hard time showing symbols for dual-mapped memory, so we'll
      * use single-mapped memory when enabled. */
-    if (erts_jit_perf_support & (BEAMASM_PERF_DUMP | BEAMASM_PERF_MAP)) {
+    if (erts_jit_perf_support & (BEAMASM_PERF_DUMP | BEAMASM_PERF_MAP))
+        erts_jit_single_map = 1;
+
+#endif
+    if (erts_jit_single_map) {
         if (auto *alloc = create_allocator(&single_params)) {
             return alloc;
         }
 
         ERTS_INTERNAL_ERROR("jit: Failed to allocate executable+writable "
-                            "memory. Either allow this or disable the "
-                            "'+JPperf' option.");
+                            "memory. Either allow this or disable both the "
+                            "'+JPperf' and '+JMsingle' options.");
     }
-#endif
 
 #if !defined(VALGRIND)
     /* Default to dual-mapped memory with separate executable and writable
--- a/erts/emulator/test/jit_SUITE.erl
+++ b/erts/emulator/test/jit_SUITE.erl
@@ -24,7 +24,7 @@
          init_per_suite/1, end_per_suite/1,
          init_per_group/2, end_per_group/2,
          init_per_testcase/2, end_per_testcase/2]).
--export([annotate/1, named_labels/1, symbols/1]).
+-export([annotate/1, jmsingle/1, named_labels/1, symbols/1]).
 
 suite() ->
     [{timetrap, {minutes, 4}}].
@@ -33,7 +33,7 @@ groups() ->
     [{perf, [symbols, annotate]}].
 
 all() ->
-    [{group, perf}, named_labels].
+    [{group, perf}, jmsingle, named_labels].
 
 init_per_suite(Config) ->
     case erlang:system_info(emu_flavor) of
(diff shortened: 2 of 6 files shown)


okeuday:
@frej The current QEMU bug discussion indicated it wasn't fixable at that level (i.e., it is related to the Linux kernel instead of QEMU).  So, the `+JMsingle bool` flag is likely necessary for other virtualization using Erlang/OTP + JIT too, right? (not only a QEMU problem)

dmorneau:
What is the downside / tradeoff of enabling the flag? It could be useful to mention in the doc. This looks really useful!

frej:
@okeuday My take away from the bug discussion was that the _current_ translation strategy in QEMU for user mode emulation doesn't consider the case when a physical page is mapped into more than one place in the emulated process' address space. When the JIT updates a page using the RW-mapping, QEMU fails to update the executable mapping and when we then branch into the new code, we crash. 

As the memory is (at least on Linux) allocated by acquiring an fd to an anonymous file (or POSIX SHM) and then `mmap`:ing it with different permissions, you could track that in QEMU, it would probably be costly and only be needed for JITs. The information is also available in `/proc/` if your process has the `CAP_SYS_ADMIN` capability, so that's why I don't consider it unfixable even if it realistically never will be fixed.

> So, the +JMsingle bool flag is likely necessary for other virtualization using Erlang/OTP + JIT too, right? (not only a QEMU problem)

That depends, if the virtualization implementation does dynamic translation of the JIT:ed code (and doesn't consider multiple maps), it will be needed. If it does not and the virtualization implementation doesn't screw up multiple `mmap`:ings of a single file, it won't be needed.

frej:
> What is the downside / tradeoff of enabling the flag? It could be useful to mention in the doc. This looks really useful!

The downside of enabling the flag is that you are using RWX pages, but the pro is that BEAM runs without segfaulting. The case for not using RWX pages is to complicated to cover in the OTP documentation, look to what [SpiderMonkey says about it](https://jandemooij.nl/blog/wx-jit-code-enabled-in-firefox/).

erts/emulator/beam/jit/beam_jit_main.cpp
```diff
@@ -149,7 +155,6 @@ static JitAllocator *pick_allocator() {
                             "memory. Either allow this or disable the "
                             "'+JPperf' option.");
```
bjorng, line 156: Shouldn't the `+JMsingle` option be mentioned as well?

bjorng (commented):
Looks good to me, but I suggest you add a test case with a smoke test of the new option (that is, that the emulator can be started when the option is given).

bjorng:

bjorng:
The test case fails on an Apple Silicon Mac because the emulator crashes with `JMsingle true`.

Apparently, on Apple Macs it is not allowed to have memory that is both writable and executable, but that is not detected when initializing the allocator. The crash occurs later when attempting to write into the allocated memory that does not appear to be writable.

I have pushed a suggestion for a fix as a fixup commit.

erts/emulator/test/jit_SUITE.erl
```diff
@@ -180,6 +180,48 @@ annotate(Config) ->
                     [Symbol, Anno])
     end.
 
+run_jmsingle_test(Param, ExpectSuccess, ErrorMsg) ->
+    Cmd = "erl +JMsingle " ++ Param ++ " -noshell " ++
+        "-eval 'io:format(\"All is well~n\"),erlang:halt(0).'",
```
bjorng, line 185: This will most likely not work on Windows. As far as I know, single quotes are not supported on Windows. The following is more likely to work:

```suggestion
        "-eval erlang:display(all_is_well),erlang:halt(0).",
```

erts/emulator/test/jit_SUITE.erl
```diff
@@ -180,6 +180,48 @@ annotate(Config) ->
...
+    end.
+
+jmsingle(_Config) ->
+    %% Smoke test to check that the emulator starts with the +JMsingle
+    %% true/false option and fails with a non-boolean, that is, we
+    %% parse the command line correctly.
+    case erlang:system_info(system_architecture) of
+        "aarch64-apple" ++ _ ->
+            %% +JMsingle true does not work on macOS running
+            %% on Apple Silicon computers.
```
garazdawi, line 212: `os:type() == {unix,netbsd}` has the same issue.

frej: Does the allocator error out at creation time or do we have to check for the os in `JitAllocator::create_allocator()` just like on the mac?

garazdawi: It errors out at creation time. So the only thing we need to fix is the testcase.

erts/emulator/test/jit_SUITE.erl
frej: I have pushed a fixup for this, I will wait for a clarification from @garazdawi before dealing with the netbsd issue.

frej:
@garazdawi , @bjorng: I have pushed a fixup for the NetBSD issue, should I wait for results from more platforms, or is it time to squash the fixups (if the CI passes)?

bjorng:
You can squash the commits now. Tests have been run successfully on all platforms.

bjorng:
Thanks!

PR #6343, opened 2022-10-03
Fix erl_tar compatibility with compressed archives

dotsimon:
In OTP 24, `erl_tar` cannot read compressed tar files that have zero padding after the compressed data.
Examples of such tar files are those from tapes (yes, even in 2022) or created by `bsdtar`
Prior to commit [e95f648](https://github.com/erlang/otp/commit/e95f648dd49d8699225ba46cea81ab6b831525bc) this data was correctly handled at least for binaries.

This commit introduces a new option to `file:open` Modes to handle this case: `compressed_one`


lib/stdlib/src/erl_tar.erl
```diff
@@ -354,7 +361,13 @@ open1({file, Fd}=Handle, read, [raw], Opts) ->
     end;
 open1({file, _Fd}=Handle, read, [], _Opts) ->
     {error, {Handle, {incompatible_option, cooked}}};
-open1(Name, Access, Raw, Opts) when is_list(Name) or is_binary(Name) ->
+open1(Name, Access, Raw, Opts0) when is_list(Name) or is_binary(Name) ->
+    Opts = case lists:member(compressed, Opts0) and lists:member(read, Opts0) of
```
bjorng, line 365:
```suggestion
open1(Name, Access, Raw, Opts0) when is_list(Name); is_binary(Name) ->
    Opts = case lists:member(compressed, Opts0) andalso lists:member(read, Opts0) of
```

bjorng (changes requested):
Looks good, except for a nitpick.

lib/stdlib/src/erl_tar.erl
dotsimon: No worries, looks like I forgot which language I was writing for!

bjorng:
> the test failures (peer and global) are problems with the tests and not this PR.

Yes.

> Pretty sure dialyzer is wrong about that.

Running the test with cover, I notice that `true` clause in lines 366-367 are never executed.

Looking further, I see that the `Opts` variable originates from a call to `open_mode/2`. According to `typer`, the spec is:

```
-spec open_mode(maybe_improper_list()) ->
    {'error','einval'} | {'ok','false' | 'read' | 'write',['raw'],['compressed' | 'read_ahead']}.
```

`Opts` comes from the last element in the `ok` tuple, which is:

```
['compressed' | 'read_ahead']
```

Please add a test case to verify that your fix fixes the bug.

dotsimon:
I added a test case and fixed the conditional.

bjorng:

bjorng (approved):

bjorng:
Thanks!

PR #6345, opened 2022-10-04
Support including off-heap binaries in max_heap_size

### LINKED ISSUE #5889: Support vheap size limit for off-heap binaries
It's possible that a single process consumes all memory on the host, even when `process_flag(max_heap_size, #{size => 1000, kill => true})` is set for that process.

It happens when the process creates huge off-heap binaries that are accounted separately (vheap_size), and there is no way to be notified about a process that grows vheap beyond reasonable size (e.g. there is no `system_monitor(large_vheap)`

Either account vheap_size as a part of a heap_size (breaks existing behaviour though), or implement `process_flag(max_vheap_size)` and `system_monitor(large_vheap)`.

Hope to find some, but so far unsuccessful.

sverker:
Fixes #5889.

Instead of adding a new limit "max_vheap_size", this PR proposes an option to the existing max_heap_size map:

`#{include_shared_bins => true | false}
`

If set to `true` off-heap binaries ("vheap") is included in the total heap size compared against the `max_heap_size` limit.
Default is false to keep existing semantics.

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -6468,6 +6468,26 @@ receive_replies(ReqId, N, Acc) ->
               or <seeerl marker="#system_flag_max_heap_size">
               <c>erlang:system_flag(max_heap_size, MaxHeapSize)</c></seeerl>.</p>
           </item>
+          <tag><c>include_shared_binaries</c></tag>
+          <item>
+            <p>
+	      When set to <c>true</c>, off-heap binaries are included in the
+	      total sum compared against the <c>size</c> limit. Off-heap binaries
+	      are typically larger binaries that may be shared between
+	      processes. The size of a shared binary is included by all
+	      processes that are referring it. Also, the entire size of a large
+	      binary may be included even if only a smaller part of it is
+	      referred by the process.
+            </p>
+	    <p>
+	      If <c>include_shared_binaries</c> is not defined in the map, the
+	      system default is used. The default system default is <c>false</c>.
+              It can be changed by either the option
+              <seecom marker="erl#+hmaxib">+hmaxib</seecom> in <c>erl(1)</c>,
+              or <seeerl marker="#system_flag_max_heap_size">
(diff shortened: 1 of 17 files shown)


erts/doc/src/erlang.xml
```diff
@@ -6285,6 +6285,22 @@ receive_replies(ReqId, N, Acc) ->
...
+	      are typically larger binaries that may be shared between
+	      processes. The size of a shared binary is included by all
+	      processes that are referring it. Also, the entire size of a large
+	      binary may be included even if only a smaller part of it is
+	      referred by the process.
+            </p>
+	    <p>
+	      Default is <c>false</c>, in which case off-heap binaries are not
+	      included in the total sum.
+            </p>
```
garazdawi, line 6302:
```suggestion
	    <p>If <c>include_shared_binaries</c> is not defined in the map, the system
              default is used. The default system default is <c>false</c>.
              It can be changed by either the option
              <seecom marker="erl#+hmaxib">+hmaxib</seecom> in <c>erl(1)</c>,
              or <seeerl marker="#system_flag_max_heap_size">
              <c>erlang:system_flag(max_heap_size, MaxHeapSize)</c></seeerl>.
            </p>
```

erts/emulator/beam/erl_map.h
```diff
@@ -74,7 +74,7 @@ typedef struct flatmap_s {
 #define flatmap_get_size(x)          (((flatmap_t*)(x))->size)
 
 #ifdef DEBUG
-#define MAP_SMALL_MAP_LIMIT    (3)
+#define MAP_SMALL_MAP_LIMIT    (4)
```
garazdawi, line 77: :) isn't this cheating? Maybe rewrite the info function to use erts_map_from_ks_and_vs?

garazdawi (commented):
[erlang:max_heap_size()](https://www.erlang.org/doc/man/erlang.html#type-max_heap_size) type module needs to be updated and some other small fixes, but I think the API and the code looks good.

PR #6346, opened 2022-10-04
stdlib: fix some autocompletion bugs

frazze-jobb:
* strip newlines after format_matches, we only want one
* find_type should recognize {type, any, []} as the any type
* functions with arity 0 and multiple other arities but
without typespecs should output the function/0


garazdawi:
If I do:

<pre>
erl -stdlib shell_expand_location above
1> esc&lt;TAB>
<b>modules</b>
escript:
1> escript:
</pre>

the only completion is printed when it should just complete on the same line. Does this PR fix that? (The problem is also there for when expand is below, but just results in a blank expand area).

frazze-jobb:
nope, but I can add a fix for that

PR #6352, opened 2022-10-07
doc: add missing syntax case to Test Case Groups

### LINKED ISSUE #6342: Documentation for Test case groups has a missing case in its def. of the syntax
The documentation in [Section 5.8 Test Case Group](https://www.erlang.org/docs/23/apps/common_test/write_test_chapter.html#test-case-groups) declares the syntax to be the following:

```
groups() -> GroupDefs

 Types:

 GroupDefs = [GroupDef]
 GroupDef = {GroupName,Properties,GroupsAndTestCases}
 GroupName = atom()
 GroupsAndTestCases = [GroupDef | {group,GroupName} | TestCase |
                      {testcase,TestCase,TCRepeatProps}]
 TestCase = atom()
 TCRepeatProps = [{repeat,N} | {repeat_until_ok,N} | {repeat_until_fail,N}]
```

This misses the case where `GroupsAndTestCases` allows also calls to functions, as in `{https, [], [def_ssl_opt | real_requests()]},` in [httpc_SUITE.erl](https://github.com/erlang/otp/blob/master/lib/inets/test/httpc_SUITE.erl#L77). 

**To Reproduce**
It is documentation

**Expected behavior**
I expected the syntax of `GroupsAndTestCases` to include function calls

**Affected versions**
OTP 23, 24, 25 (AFAIK)

kikofernandez:
This` PR adds a missing case on how to write test case groups.
The missing case is that one can perform function calls at any point of the test case group as long as the result of the function call returns a `GroupsAndTestCases` ([Syntax for the Erlang TestCaseGroup docs](https://www.erlang.org/docs/23/apps/common_test/write_test_chapter.html#test-case-groups)).
The following example follows, which is also used in the OTP code:

```erlang
groups() ->
    [
     {http, [], real_requests()}
   ]
```
Closes #6342

### DIFF
--- a/lib/common_test/doc/src/write_test_chapter.xml
+++ b/lib/common_test/doc/src/write_test_chapter.xml
@@ -551,6 +551,21 @@
       [{tests2, default,
         [{tests3, [parallel,{repeat,100}]}]}]}].</pre>
 
+    <p>For ease of readibility, all syntax definitions can be replaced by a function
+    call whose return value should match the expected syntax case.</p>
+    <p>Example</p>
+    <pre>
+ all() ->
+    [{group, tests1, default, test_cases()},
+     {group, tests1, default, [shuffle_test(),
+                               {tests3, default}]}].
+ test_cases() ->
+    [{tests2, [parallel]}, {tests3, default}].
+
+ shuffle_test() ->
+    {tests2, [shuffle,{repeat,10}]}.
+    </pre>
+
     <p>The described syntax can also be used in test specifications
       to change group properties at the time of execution,
       without having to edit the test suite. For more information, see
--- a/lib/common_test/doc/src/write_test_chapter.xml
+++ b/lib/common_test/doc/src/write_test_chapter.xml
@@ -449,7 +449,7 @@
     functions and execution properties. Test case groups are defined by
     function
     <seemfa marker="ct_suite#Module:groups/0"><c>groups/0</c></seemfa>
-    according to the following syntax:</p>
+    that should return a term having the following syntax:</p>
     <pre>
  groups() -> GroupDefs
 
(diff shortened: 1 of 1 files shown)


lib/common_test/doc/src/write_test_chapter.xml
```diff
@@ -551,6 +551,21 @@
       [{tests2, default,
         [{tests3, [parallel,{repeat,100}]}]}]}].</pre>
 
+    <p>For ease of readibility, all syntax definitions can be replaced by a function
```
bjorng, line 554:
```suggestion
    <p>For ease of readability, all syntax definitions can be replaced by a function
```

lib/common_test/doc/src/write_test_chapter.xml
```diff
@@ -551,6 +551,21 @@
...
+    call whose return value should match the expected syntax case.</p>
+    <p>Example</p>
```
bjorng, line 556: For consistency with the rest of the chapter:

```suggestion
    <p><em>Example:</em></p>
```

lib/common_test/doc/src/write_test_chapter.xml
```diff
@@ -551,6 +551,21 @@
...
+ all() ->
+    [{group, tests1, default, test_cases()},
+     {group, tests1, default, [shuffle_test(),
+                               {tests3, default}]}].
+ test_cases() ->
+    [{tests2, [parallel]}, {tests3, default}].
+
+ shuffle_test() ->
+    {tests2, [shuffle,{repeat,10}]}.
+    </pre>
```
bjorng, line 567: To avoid an empty line and consistency with the rest of the chapter:

```suggestion
    {tests2, [shuffle,{repeat,10}]}.</pre>
```

bjorng (commented):
Perhaps you should also rewrite the beginning of the section, too. It is currently:

> Test case groups are defined by function [groups/0](https://erlang.github.io/prs/6352/lib/common_test-1.23.1/doc/html/ct_suite.html#Module:groups-0) according to the following syntax:

To avoid falsely implying that this is syntax that must be exactly followed, perhaps it could be reformulated as:

> Test case groups are defined by function [groups/0](https://erlang.github.io/prs/6352/lib/common_test-1.23.1/doc/html/ct_suite.html#Module:groups-0) that should return a term having the following structure:

or perhaps like this:

> ... return a term having the following type:

That would be the followed by one or more `-type` specs.

kikofernandez:
thanks for the feedback @bjorng , I made the appropriate changes :)

Issue #6289, opened 2022-09-08
`term()` vs. `any()` vs. `_`


TD5:
**Is your feature request related to a problem? Please describe.**
Periodically, I see users confused by `term()` vs. `any()` vs. `_` - they're not sure if there is a difference in meaning between them, or which they should use once they discover they all amount to the same thing.

Each seems to be preferred in at least one sort of scenario, so there doesn't seem to be a consensus around a particular convention:
- `any()` seems to be the default printed by `erl_pp`
- `_` is used by `erl_pp` for map keys/values and tuple elements, presumably for compactness?
- `term()` seems to be preferred in OTP function specs

**Describe the solution you'd like**
- Perhaps some official word on when to use `term()` vs. `any()` vs. `_`, plus maybe some context on why all three exist.
- Long term, unify on a single syntax, or at least a long and a short name, e.g., say, `term()` and `_`

josevalim:
If I am not mistaken, `_` is not valid in typespecs (except in bitstrings?), only in patterns. It exists in a different context than `term()` and `any()`. 

`term()` and `any()` are considered equivalent and such is mentioned here: https://www.erlang.org/doc/reference_manual/typespec.html - the table says `term()` is defined as `any()`. AFAIK which one to use is totally a matter of personal preference but note tools like Gradualizer consider them to be distinct.

TD5:
> If I am not mistaken, _ is not valid in typespecs (except in bitstrings?), only in patterns. It exists in a different context than term() and any().

That's just it: despite it not being documented as a valid way to _input_ an `any()` type, `erl_pp` (the module powers the rendering of types in Dialyzer, etc) will _output_ it in place of `any()` in several circumstances, which can cause some confusion for users.

> the table says `term()` is defined as `any()`

I think the documentation does a good job on making this clear, but it is a common question I see from Erlang beginners which causes a small amount of friction (they typically presume there must be _some_ difference, else why do they both exist?)

ilya-klyuchnikov:
> If I am not mistaken, _ is not valid in typespecs

https://github.com/erlang/otp/blob/54df3e34d288a18ff5846196fb09e1cdceb07fea/lib/stdlib/src/gb_sets.erl#L201

josevalim:
@TD5 @ilya-klyuchnikov TIL, thanks.

Then I agree that `_` should most likely be phased out in favor of `any()` or `term()`.

kostis:
[...]

PR #6355, opened 2022-10-10
docs: add note about anonymous type variables

kikofernandez:
Add note and examples about anonymous type variables (#6289)

### DIFF
--- a/system/doc/reference_manual/typespec.xml
+++ b/system/doc/reference_manual/typespec.xml
@@ -607,5 +607,14 @@
       of the following form:
     </p>
     <pre>  -spec my_error(term()) -> no_return().</pre>
+
+  <note>
+    <p>Erlang uses the shorthand version <c>_</c> as an anonymous type variable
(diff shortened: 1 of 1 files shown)


system/doc/reference_manual/typespec.xml
```diff
@@ -607,5 +607,29 @@
...
+    <pre>
+  -spec id(_) -> _.
+    </pre>
+    is different from the identity function defined with type variables
+    <pre>
+  -spec id(X) -> X.
+    </pre>
+    where the difference is that the former specifies that the argument and return
+    types can be any type (<c>any()</c>), while the later specified that
+    the argument and return types are of the same type.
```
bjorng, line 631: I think this part is unnecessary because there already is a discussion about that kind spec above.

bjorng (approved):
Looks good! I suggest that you squash the commits.

kikofernandez:
I have squashed the commits :)

PR #6357, opened 2022-10-11
pg: Note for local-only use cases

essen:
A tiny note to point how it can be used without the distributed features.

### DIFF
--- a/lib/kernel/doc/src/pg.xml
+++ b/lib/kernel/doc/src/pg.xml
@@ -89,8 +89,15 @@
       If there is another process registered under this name, or another ETS table
       exists, scope fails to start.</p>
       <p>Local membership is not preserved if scope process exits and
-        restarts.
-    </p></note>
+        restarts.</p>
+      <p>A scope can be kept local-only by using a scope name that is unique
+        cluster-wide, e.g. the node name:</p>
+      <taglist>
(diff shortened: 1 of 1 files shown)


lib/kernel/doc/src/pg.xml
```diff
@@ -89,7 +89,9 @@
       If there is another process registered under this name, or another ETS table
       exists, scope fails to start.</p>
       <p>Local membership is not preserved if scope process exits and
-        restarts.
+        restarts.</p>
+      <p>A scope can be kept local-only by using a scope name that is unique
+        cluster-wide.
```
max-au, line 94: It might be useful to give an example here:

"<p>A scope can be kept local-only by using a scope name that is unique
        cluster-wide, e.g. node name: `pg:start_link(node())`."

essen: Done, thanks!

essen:
I have no idea how to make xmlint happy. I tried to do like the `socket` manual page.

garazdawi:
Your changes are fine. There was merge done to master that broke the xmllint which was then fixed in #6361.

If you force a CI rebuild by pushing a `git commit --amend` with no changes the checks should now pass.

max-au (approved):

garazdawi:
thanks!

PR #6359, opened 2022-10-11
Replace some uses of ets:lookup by ets:lookup_element in the stdlib

RobinMorisset:
When only one field of an ETS record is needed, using ets:lookup_element is usually faster than ets:lookup. This patch makes sure that we do so when possible in various erlang files in OTP, taking advantage of the recently landed ets:lookup_element/4 (https://github.com/erlang/otp/pull/6234) in places.

### DIFF
--- a/lib/asn1/src/asn1_db.erl
+++ b/lib/asn1/src/asn1_db.erl
@@ -82,11 +82,11 @@ loop(#state{parent = Parent, monitor = MRef, table = Table,
             includes = Includes} = State) ->
     receive
         {set, Mod, K2, V} ->
-            [{_, Modtab}] = ets:lookup(Table, Mod),
+            Modtab = ets:lookup_element(Table, Mod, 2),
             ets:insert(Modtab, {K2, V}),
             loop(State);
         {set, Mod, Kvs} ->
-            [{_, Modtab}] = ets:lookup(Table, Mod),
+            Modtab = ets:lookup_element(Table, Mod, 2),
             ets:insert(Modtab, Kvs),
             loop(State);
         {From, {get, Mod, K2}} ->
@@ -105,7 +105,7 @@ loop(#state{parent = Parent, monitor = MRef, table = Table,
             end,
             loop(State);
         {save, OutFile, Mod} ->
-            [{_,Mtab}] = ets:lookup(Table, Mod),
+            Mtab = ets:lookup_element(Table, Mod, 2),
 	    TempFile = OutFile ++ ".#temp",
             ok = ets:tab2file(Mtab, TempFile),
 	    ok = file:rename(TempFile, OutFile),
@@ -146,10 +146,7 @@ get_table(Table, Mod, Includes) ->
     end.
 
 lookup(Tab, K) ->
-    case ets:lookup(Tab, K) of
-        []      -> undefined;
-        [{K,V}] -> V
-    end.
+	ets:lookup_element(Tab, K, 2, undefined).
 
 info(EruleMaps) ->
     {asn1ct:vsn(),EruleMaps}.
--- a/lib/kernel/src/application_controller.erl
+++ b/lib/kernel/src/application_controller.erl
@@ -27,7 +27,7 @@
 	 change_application_data/2, prep_config_change/0, config_change/1,
 	 which_applications/0, which_applications/1,
 	 loaded_applications/0, info/0, set_env/2,
-	 get_pid_env/2, get_env/2, get_pid_all_env/1, get_all_env/1,
+	 get_pid_env/2, get_env/2, get_env/3, get_pid_all_env/1, get_all_env/1,
 	 get_pid_key/2, get_key/2, get_pid_all_key/1, get_all_key/1,
 	 get_master/1, get_application/1, get_application_module/1,
 	 start_type/1, permit_application/2, do_config_diff/2,
@@ -343,10 +343,14 @@ get_pid_env(Master, Key) ->
     end.
 
 get_env(AppName, Key) ->
-    case ets:lookup(ac_tab, {env, AppName, Key}) of
-	[{_, Val}] -> {ok, Val};
-	_ -> undefined
+    NotFound = make_ref(),
+    case ets:lookup_element(ac_tab, {env, AppName, Key}, 2, NotFound) of
+        NotFound -> undefined;
+        Val -> {ok, Val}
     end.
+
+get_env(AppName, Key, Default) ->
+    ets:lookup_element(ac_tab, {env, AppName, Key}, 2, Default).
 
 get_pid_all_env(Master) ->
     case ets:match(ac_tab, {{application_master, '$1'}, Master}) of
@@ -442,10 +446,7 @@ start_type(Master) ->
 
 
 get_master(AppName) ->
-    case ets:lookup(ac_tab, {application_master, AppName}) of
-	[{_, Pid}] -> Pid;
-	_ -> undefined
-    end.
+    ets:lookup_element(ac_tab, {application_master, AppName}, 2, undefined).
 
 get_application(Master) ->
     case ets:match(ac_tab, {{application_master, '$1'}, Master}) of
--- a/lib/stdlib/src/beam_lib.erl
+++ b/lib/stdlib/src/beam_lib.erl
@@ -832,8 +832,7 @@ symbol(_, AT, I1, I2, _I3, _Cnt) ->
     {atm(AT, I1), I2}.
 
 atm(AT, N) ->
-    [{_N, S}] = ets:lookup(AT, N),
-    S.
+    ets:lookup_element(AT, N, 2).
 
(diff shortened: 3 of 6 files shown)


lib/asn1/src/asn1_db.erl
```diff
@@ -82,11 +82,11 @@ loop(#state{parent = Parent, monitor = MRef, table = Table,
             includes = Includes} = State) ->
     receive
         {set, Mod, K2, V} ->
-            [{_, Modtab}] = ets:lookup(Table, Mod),
+            Modtab = ets:lookup_element(Table, Mod, 2),
```
Maria-12648430, line 85: This is not fully equivalent. The old code will badmatch if the found element is a tuple of size >2, but the new code will pass.

lib/kernel/src/application_controller.erl
```diff
@@ -343,9 +343,9 @@ get_pid_env(Master, Key) ->
     end.
 
 get_env(AppName, Key) ->
-    case ets:lookup(ac_tab, {env, AppName, Key}) of
-	[{_, Val}] -> {ok, Val};
-	_ -> undefined
+    case ets:lookup_element(ac_tab, {env, AppName, Key}, 2, not_found) of
+        not_found -> undefined,
+        Val -> {ok, Val}
```
Maria-12648430, line 348: This (apart from the already mentioned issue with tuples of size >2) is pretty questionable. If `not_found` is stored in the table, it is indistinguishable from there being no record at all and `undefined` will be returned. I suggest to create a reference beforehand, use it as the default, and match on that.

lib/asn1/src/asn1_db.erl
```diff
@@ -82,11 +82,11 @@ loop(#state{parent = Parent, monitor = MRef, table = Table,
...
             ets:insert(Modtab, {K2, V}),
             loop(State);
         {set, Mod, Kvs} ->
-            [{_, Modtab}] = ets:lookup(Table, Mod),
+            Modtab = ets:lookup_element(Table, Mod, 2),
```
Maria-12648430, line 89: Not fully equivalent.

lib/asn1/src/asn1_db.erl
```diff
@@ -105,7 +105,7 @@ loop(#state{parent = Parent, monitor = MRef, table = Table,
             end,
             loop(State);
         {save, OutFile, Mod} ->
-            [{_,Mtab}] = ets:lookup(Table, Mod),
+            Mtab = ets:lookup_element(Table, Mod, 2),
```
Maria-12648430, line 108: Not fully equivalent.

lib/asn1/src/asn1_db.erl
```diff
@@ -146,10 +146,7 @@ get_table(Table, Mod, Includes) ->
     end.
 
 lookup(Tab, K) ->
-    case ets:lookup(Tab, K) of
-        []      -> undefined;
-        [{K,V}] -> V
-    end.
+	ets:lookup_element(Tab, K, 2, undefined).
```
Maria-12648430, line 149: Not fully equivalent.

lib/kernel/src/application_controller.erl
```diff
@@ -442,10 +442,7 @@ start_type(Master) ->
 
 
 get_master(AppName) ->
-    case ets:lookup(ac_tab, {application_master, AppName}) of
-	[{_, Pid}] -> Pid;
-	_ -> undefined
-    end.
+    ets:lookup_element(ac_tab, {application_master, AppName}, 2, undefined).
```
Maria-12648430, line 445: Not fully equivalent.

lib/stdlib/src/beam_lib.erl
```diff
@@ -832,8 +832,7 @@ symbol(_, AT, I1, I2, _I3, _Cnt) ->
     {atm(AT, I1), I2}.
 
 atm(AT, N) ->
-    [{_N, S}] = ets:lookup(AT, N),
-    S.
+    ets:lookup_element(AT, N, 2).
```
Maria-12648430, line 835: Not fully equivalent.

Maria-12648430 (changes requested):

lib/kernel/src/application_controller.erl
```diff
@@ -343,9 +343,9 @@ get_pid_env(Master, Key) ->
     end.
 
 get_env(AppName, Key) ->
-    case ets:lookup(ac_tab, {env, AppName, Key}) of
-	[{_, Val}] -> {ok, Val};
-	_ -> undefined
+    case ets:lookup_element(ac_tab, {env, AppName, Key}, 2, not_found) of
+        not_found -> undefined,
```
juhlig, line 347: There should be a semicolon at the end of this line, not a comma.

juhlig (changes requested):

lib/asn1/src/asn1_db.erl
bjorng: @Maria-12648430 The lookup is intended to always succeed. Testing the tuple size is only done because it was the most convenient way to get the second element. It is also an (accidental) assertion to crash if there is some strange bugs that stores garbage in the table. In practice, that assertion is very unlikely to ever fire because the table is only writeable by the process owning it.

@RobinMorisset On the other hand, this module is part of the ASN.1 compiler, so the optimization will give very little.

lib/kernel/src/application_controller.erl
bjorng: I agree that a reference should be used, but since creating a reference does not come for free, I think you should do some benchmarking whether it would be an improvement compared to the original code.

This table is writable by all processes in the system, so there is a (very) slight possibility that testing the tuple size could detect a garbage value written by another process. However, since this is a system process always running, I think that (very) slightly improving performance is worth more than keeping the tuple test that is very unlikely to go wrong.

lib/stdlib/src/beam_lib.erl
bjorng: This ETS table is short-lived and only writeable by the owner. I find it very unlikely that the tuple size assertion would ever find a bug.

bjorng (changes requested):

lib/asn1/src/asn1_db.erl
RobinMorisset: I found this code by a naive grep for ets:lookup, I am quite willing to drop this change.
The change in this PR that I consider important is the application_controller:get_env one, as that function shows surprisingly heavily in some profiling I did.

lib/kernel/src/application_controller.erl
RobinMorisset: I've just done some benchmarking.
Running `erlperf -l 10K --init 'application:set_env(foobar, key, 42).' 'application:get_env(foobar, key).'` 10 times for each configuration I get the following results:
Baseline: 15314 Ki to 15649 Ki QPS
With this patch: 18939 Ki to 21598 Ki QPS (most values around 21000)
Using a reference instead of an atom: 17391 Ki to 19531 Ki (most values around 18000)

I then checked whether using `{key, 0}` as the key instead of just the atom `key` changes anything, and I got the following results:
Baseline: 12225 Ki to 12903 Ki
With this patch: 18018 Ki to 18727 Ki
Using a reference instead of an atom: 15106 Ki to 17361 Ki (most values around 16000)

So lookup_element + make_ref seems to always be faster than lookup, but the make_ref is far from free. Would sticking to using an atom and just making it more unique (something like `not_found_application_controller_get_env`) be considered safe enough?

lib/kernel/src/application_controller.erl
```diff
@@ -344,7 +344,7 @@ get_pid_env(Master, Key) ->
 
 get_env(AppName, Key) ->
```
michalmuskala, line 345: This is used in two ways from `application`. In particular `application:get_env/3` has a default value, which could use `ets:lookup_element` directly with no wrapping.

It might make sense to export 2 functions from `application_controller` - one that returns the `{ok, _} | undefined` pattern, one that returns `_ | Default`. This would have the additional benefit of avoiding allocating the `{ok, _}` tuple on lookup successes.

bjorng: @RobinMorisset I suggest that you implement this suggestion and that you use a reference in the version that has to return an `ok` tuple.

lib/asn1/src/asn1_db.erl
bjorng: > I found this code by a naive grep for ets:lookup, I am quite willing to drop this change.

In that case, I suggest that you drop the changes in `asn1db`.

lib/kernel/src/application_controller.erl
RobinMorisset: Done, thanks for the suggestions and comments.

bjorng (approved):

bjorng:

bjorng:
Thanks!

PR #6362, opened 2022-10-12
ssl: Store PeerCertificate for session resumption

sindrip:
This PR subsumes PR: #5957 and closes issue: #5870 

Calling `ssl:peercert/1` after a client connects with session resumption returns `{error, no_peercert}`. This PR adds the new options `stateful_with_cert` and `stateless_with_cert` to the `server_session_tickets` option that allow `ssl:peercert/1` to return the client certificate even after session resumption.

### DIFF
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -1461,15 +1461,34 @@ fun(srp, Username :: binary(), UserState :: term()) ->
       <datatype>
 	<name name="server_session_tickets"/>
 	<desc>
-	  <p>Configures the session ticket functionality. Allowed values are <c>disabled</c>,
-	  <c>stateful</c> and <c>stateless</c>.</p>
-	  <p>If it is set to <c>stateful</c> or
-	  <c>stateless</c>, session resumption with pre-shared keys is enabled and the server will
-	  send stateful or stateless session tickets to the client after successful connections.</p>
+          <p>Configures the session ticket functionality. Allowed values are <c>disabled</c>,
+          <c>stateful</c>, <c>stateless</c>, <c>stateful_with_cert</c>, <c>stateless_with_cert</c>.</p>
+          <p>If it is not set to <c>disabled</c>,
+          session resumption with pre-shared keys is enabled and the server will
+          send stateful or stateless session tickets to the client after successful connections.</p>
+          
+          <note><p>
+          Pre-shared key session ticket resumption does not include any certificate exchange, 
+          hence the function <seemfa marker="ssl:ssl#peercert/1">ssl:peercert/1</seemfa> will not 
+          be able to return the peer certificate as it is only communicated in the initial handshake.
+          The server options <c>stateful_with_cert</c> or <c>stateless_with_cert</c> may be used 
+          to make a server associate the client certificate from the original handshake 
+          with the tickets it issues.
+          </p></note>
+
 	  <p>A stateful session ticket is a database reference to internal state information.
 	  A stateless session ticket is a self-encrypted binary that contains both cryptographic keying
 	  material and state data.
 	  </p>
+          
+          <warning><p>
+          If it is set to <c>stateful_with_cert</c> the client certificate
+          is stored with the internal state information, increasing memory consumption.
+          If it is set to <c>stateless_with_cert</c> the client certificate is
+          encoded in the self-encrypted binary that is sent to the client,
+          increasing the payload size.
+          </p></warning>
+
 	  <note><p>This option is supported by TLS 1.3 and above. See also
 	  <seeguide marker="ssl:using_ssl#session-tickets-and-session-resumption-in-tls-1.3">
 	  SSL's Users Guide, Session Tickets and Session Resumption in TLS 1.3</seeguide>
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -386,7 +386,7 @@
 -type log_alert()                :: boolean().
 -type logging_level()            :: logger:level() | none | all.
 -type client_session_tickets()   :: disabled | manual | auto.
--type server_session_tickets()   :: disabled | stateful | stateless.
+-type server_session_tickets()   :: disabled | stateful | stateless | stateful_with_cert | stateless_with_cert.
 -type session_tickets()          :: client_session_tickets() | server_session_tickets().
 -type key_update_at()            :: pos_integer().
 -type bloom_filter_window_size()    :: integer().
@@ -1899,17 +1899,24 @@ opt_tickets(UserOpts, #{versions := Versions} = Opts, #{role := client}) ->
     assert_server_only(stateless_tickets_seed, UserOpts),
     Opts#{session_tickets => SessionTickets, use_ticket => UseTicket, early_data => EarlyData};
 opt_tickets(UserOpts, #{versions := Versions} = Opts, #{role := server}) ->
-    {_, SessionTickets} = get_opt_of(session_tickets, [disabled, stateful, stateless], disabled, UserOpts, Opts),
+    {_, SessionTickets} =
+        get_opt_of(session_tickets,
+                   [disabled, stateful, stateless, stateful_with_cert, stateless_with_cert],
+                   disabled,
+                   UserOpts,
+                   Opts),
     assert_version_dep(SessionTickets =/= disabled, session_tickets, Versions, ['tlsv1.3']),
 
     {_, EarlyData} = get_opt_of(early_data, [enabled, disabled], disabled, UserOpts, Opts),
     option_incompatible(SessionTickets =:= disabled andalso EarlyData =:= enabled,
                         [early_data, {session_tickets, disabled}]),
 
+    Stateless = lists:member(SessionTickets, [stateless, stateless_with_cert]),
+
     AntiReplay =
         case get_opt(anti_replay, undefined, UserOpts, Opts) of
             {_, undefined} -> undefined;
-            {_,AR} when SessionTickets =/= stateless ->
+            {_,AR} when not Stateless ->
                 option_incompatible([{anti_replay, AR}, {session_tickets, SessionTickets}]);
             {_,'10k'}  -> {10, 5, 72985};  %% n = 10000 p = 0.030003564 (1 in 33) m = 72985 (8.91KiB) k = 5
             {_,'100k'} -> {10, 5, 729845}; %% n = 10000 p = 0.03000428 (1 in 33) m = 729845 (89.09KiB) k = 5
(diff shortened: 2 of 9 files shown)


sindrip:
Addressing review from PR: #5957 @u3s 

- [x]  1. we would like this behavior to be optional, not default on server side.  
so only interested users are affected. 
preferably we would like to avoid new option, maybe we could extend
`server_session_tickets() = disabled | stateful | stateless`
with values: `stateful_with_cert | stateless_with_cert` ?

 

- [x] 2. it would be preferred to have equivalent behavior for stateful tickets,
although having different implementation (cert stored on server side).
 
- [ ] 3. `ssl_session_ticket_SUITE` - we would like to have blackbox tests using ssl API
        1. if we lack some test coverage, we might add some whitebox
        2. with new code `ssl:peercert/1` should return `{ok, Cert}` instead of
           `{error, no_peercert}` when `stateless_with_cert` or `stateful_with_cert`
        3. maybe also add instructions in existing tests, for checking behavior with existing code

This is done, but please take a look at the new test groups that I have set up in ssl_session_ticket_SUITE.
Let me know if you agree with what I have there.

Can you clarify what you mean by point 3, adding instructions to existing tests?
 
- [x] 4. for new option values we would need necessary doc update

- [x] 5. see also minor comments in code

I think the indentation is done correctly now, and the tls_server_session_ticket_SUITE is passing now. (The suite was being skipped on the commit that this branch is based on and I also fixed that).

u3s:
>     * [ ]  3. `ssl_session_ticket_SUITE` - we would like to have blackbox tests using ssl API
>       1. if we lack some test coverage, we might add some whitebox
>       2. with new code `ssl:peercert/1` should return `{ok, Cert}` instead of
>       `{error, no_peercert}` when `stateless_with_cert` or `stateful_with_cert`
>       3. maybe also add instructions in existing tests, for checking behavior with existing code
> 
> 
> This is done, but please take a look at the new test groups that I have set up in ssl_session_ticket_SUITE. Let me know if you agree with what I have there.
> Can you clarify what you mean by point 3, adding instructions to existing tests?

I initially thought that adding `peercert_stateless_stateful_without_cert` test case might not be required as one of existing test cases already covers that flow - if so such _instruction_ `{error, no_peercert} = ssl:peercert(SSocket1)` could be added to improve verification done there.
Do you think we need `peercert_stateless_stateful_without_cert` ?

PS. we will do code review a bit late and come back to you.

sindrip:
> I initially thought that adding `peercert_stateless_stateful_without_cert` test case might not be required as one of existing test cases already covers that flow - then such _instruction_ `{error, no_peercert} = ssl:peercert(SSocket1)` could be added to improve verification done there. Do you think we need `peercert_stateless_stateful_without_cert` ?
> 
> PS. we will do code review a bit later and come back to you.

Now I see, yes that makes sense. I'm looking at it, and I might even be able to put it into `basic_test` with a conditional on what the response should be. Then we can get rid of the 2 new testcases, and run the same testcases for both implementations.

sindrip:
Made the tests simpler, I think this looks much better. I also found one bug with option handling between `anti_replay` and the new options and fixed that.

sindrip:
Rebased due to conflicts, fixed the option handling in `ssl.erl` that had changed.

u3s (commented):
thanks for update. see comments inline.

lib/ssl/src/ssl.erl
```diff
@@ -1899,17 +1899,19 @@ opt_tickets(UserOpts, #{versions := Versions} = Opts, #{role := client}) ->
     assert_server_only(stateless_tickets_seed, UserOpts),
     Opts#{session_tickets => SessionTickets, use_ticket => UseTicket, early_data => EarlyData};
 opt_tickets(UserOpts, #{versions := Versions} = Opts, #{role := server}) ->
-    {_, SessionTickets} = get_opt_of(session_tickets, [disabled, stateful, stateless], disabled, UserOpts, Opts),
+    {_, SessionTickets} = get_opt_of(session_tickets, [disabled, stateful, stateless, stateful_with_cert, stateless_with_cert], disabled, UserOpts, Opts),
```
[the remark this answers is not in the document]
sindrip: Definitely a good point to not make it worse, I'm just used to auto-formatters taking care of my long lines. I made it better and far from the longest line!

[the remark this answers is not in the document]
sindrip: Definitely a good point to not make it worse, I'm just used to auto-formatters taking care of my long lines. I made it better and far from the longest line!

lib/ssl/test/tls_server_session_ticket_SUITE.erl
```diff
@@ -63,16 +66,18 @@ all() ->
```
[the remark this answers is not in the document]
u3s: > Also as an aside, if you drag across the line-numbers on the left you can make a comment on a code-block (if you weren't aware of it)

Yes, I know. I'm trying to use it - but in this case I commented line which was unchanged. This was not possible for line with `all()` (github limitation).  So i tried to add comment in the closest line.

lib/ssl/test/tls_server_session_ticket_SUITE.erl
```diff
@@ -114,19 +126,25 @@ end_per_testcase(_TestCase, Config) ->
 main_test() ->
     [{doc, "Ticket main scenario"}].
 main_test(Config) when is_list(Config) ->
+    case ?config(anti_replay, Config) of
+        undefined -> undefined;
+        _ ->
+            % Sleep longer than warmup period
+            ct:sleep({seconds, 2 * ?LIFETIME})
+    end,
```
u3s, line 134: 3 occurences of ct:sleep case statements for anti_replay tests - could you move them to init_per_testcase.
the less sleeping and copy/code code the better.

lib/ssl/test/ssl_session_ticket_SUITE.erl
```diff
@@ -155,6 +161,10 @@ init_per_group(stateful, Config) ->
     [{server_ticket_mode, stateful} | proplists:delete(server_ticket_mode, Config)];
 init_per_group(stateless, Config) ->
     [{server_ticket_mode, stateless} | proplists:delete(server_ticket_mode, Config)];
+init_per_group(stateful_with_cert, Config) ->
+    [{server_ticket_mode, stateful_with_cert} | proplists:delete(server_ticket_mode, Config)];
+init_per_group(stateless_with_cert, Config) ->
+    [{server_ticket_mode, stateless_with_cert} | proplists:delete(server_ticket_mode, Config)];
```
u3s, line 167: If possible, reduce those 4 init_per_group into 1 with guard expression checking for `stateful`, `stateless`, `stateful_with_cert`, `stateless_with_cert`.

u3s:
Thanks for work so far. Tests in our environment look good.
I would @IngelaAndin to also have a look before we merge.

lib/ssl/test/ssl_session_ticket_SUITE.erl
sindrip: Fixed

lib/ssl/test/tls_server_session_ticket_SUITE.erl
sindrip: I moved it into `init_per_testcase`, the difference is that now we also sleep for `misc_test` (4 extra seconds). I don't see a reason to make an auxiliary function taking into account which test to sleep in, unless more anti-replay tests are added that need to not sleep.

u3s: Ah. I missed misc_test that one. I agree it shouldn't be a problem. Actually I think this should be a default behavior for anti_replay tests since PR for customizing seed.

lib/ssl/doc/src/ssl.xml
```diff
@@ -1462,14 +1462,31 @@ fun(srp, Username :: binary(), UserState :: term()) ->
 	<name name="server_session_tickets"/>
 	<desc>
 	  <p>Configures the session ticket functionality. Allowed values are <c>disabled</c>,
-	  <c>stateful</c> and <c>stateless</c>.</p>
+	  <c>stateful</c>, <c>stateless</c>, <c>stateful_with_cert</c>, <c>stateless_with_cert</c>.</p>
 	  <p>If it is set to <c>stateful</c> or
```
IngelaAndin, line 1466: Maybe here it would make more sense to write. If it is not disabled, the session resumption with pre-shared keys.

sindrip: I agree, fixed

lib/ssl/doc/src/ssl.xml
```diff
@@ -1461,15 +1461,32 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
-	  <p>If it is set to <c>stateful</c> or
-	  <c>stateless</c>, session resumption with pre-shared keys is enabled and the server will
-	  send stateful or stateless session tickets to the client after successful connections.</p>
+          <p>Configures the session ticket functionality. Allowed values are <c>disabled</c>,
+          <c>stateful</c>, <c>stateless</c>, <c>stateful_with_cert</c>, <c>stateless_with_cert</c>.</p>
+          <p>If it is not set to <c>disabled</c>,
+          session resumption with pre-shared keys is enabled and the server will
+          send stateful or stateless session tickets to the client after successful connections.</p>
+
+          <warning><p>If it is set to <c>stateful</c> or <c>stateless</c>
```
IngelaAndin, line 1470: I think we probably should make this a note instead of a warning. A warning is usually used to alert you that something is less secure, very memory-consuming, or might be a bad idea under some circumstances.  I was thinking something along the lines: Pre-shard key session ticket resumption does not include any certificate exchange, hence the function ssl:peercert /1 will not be able to return the peer certificate as it is only communicated in the initial handshake. The 
server options <c>stateful_with_cert</c> or <c>stateless_with_cert</c> may be used to make a server associate the client certificate from the original handshake with the tickets it issues. 

Here we might need a warning that using these options will increase memory consumption for stateful tickets and
for stateless tickets it will increase the handshake payload.

sindrip: Makes sense, I changed it into a note with your text, and then I changed the last paragraph into a warning specifying the increased memory consumption and increased payload size. What do you think?

u3s (commented):
1. please extend `ssl_api_SUITE:options_whitebox` to cover added options
2. I've pushed a commit adjusting indentation for 3 case blocks
3. finally please squash the commits

sindrip:
I added the tests and squashed

sindrip:
I just checked the failure, and it's not in a testcase that I changed and the tests are passing on my end at least. Can you re-run?

sindrip:
I see that the tests passed on the re-run, but the results don't seem to have been updated in the ct results check. I don't think there's anything that needs to be done from my side at this moment, just want to confirm.

u3s:
thanks for contribution!

PR #6365, opened 2022-10-13
Further optimize binary matching

bjorng:
This pull request builds on the improvements in #6259 to further optimize binary matching. See the individual commit messages for more details.

### DIFF
--- a/erts/emulator/beam/jit/arm/instr_bs.cpp
+++ b/erts/emulator/beam/jit/arm/instr_bs.cpp
@@ -682,7 +682,7 @@ void BeamModuleAssembler::emit_bs_get_integer2(const ArgLabel &Fail,
         if (emit_bs_get_field_size(Sz, unit, fail, ARG5) >= 0) {
             /* This operation can be expensive if a bignum can be
              * created because there can be a garbage collection. */
-            auto [_, max] = getClampedRange(Sz);
+            auto max = std::get<1>(getClampedRange(Sz));
             bool potentially_expensive =
                     max >= SMALL_BITS || (max * Unit.get()) >= SMALL_BITS;
 
@@ -1915,7 +1915,7 @@ void BeamModuleAssembler::emit_i_bs_create_bin(const ArgLabel &Fail,
             comment("size binary/integer/float/string");
 
             if (always_small(seg.size)) {
-                auto [min, _] = getClampedRange(seg.size);
+                auto min = std::get<0>(getClampedRange(seg.size));
                 if (min >= 0) {
                     can_fail = false;
                 }
@@ -2687,7 +2687,8 @@ struct BsmSegment {
         GET_BINARY,
         SKIP,
         DROP,
-        GET_TAIL
+        GET_TAIL,
+        EQ
     } action;
     ArgVal live;
     Uint size;
@@ -3111,6 +3112,7 @@ static std::vector<BsmSegment> opt_bsm_segments(
         if (heap_need != 0 && seg.live.isWord()) {
             BsmSegment s = seg;
 
+            read_action_pos = -1;
             s.action = BsmSegment::action::TEST_HEAP;
             s.size = heap_need;
             segs.push_back(s);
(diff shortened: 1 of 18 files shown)


erts/emulator/beam/jit/arm/instr_bs.cpp
```diff
@@ -658,19 +679,51 @@ void BeamModuleAssembler::emit_bs_get_integer2(const ArgLabel &Fail,
         int unit = Unit.get();
 
         if (emit_bs_get_field_size(Sz, unit, fail, ARG5) >= 0) {
+            /* This operation can be expensive if a bignum can be
+             * be created because there can be a garbage collection. */
```
garazdawi, line 683:
```suggestion
             * created because there can be a garbage collection. */
```

erts/emulator/beam/jit/arm/instr_bs.cpp
```diff
@@ -56,18 +54,39 @@ int BeamModuleAssembler::emit_bs_get_field_size(const ArgSource &Size,
         return -1;
     } else {
         auto size_reg = load_source(Size, TMP2);
+        bool can_fail = true;
+
+        if (always_small(Size)) {
+            auto [min, max] = getClampedRange(Size);
+            can_fail = !(0 <= min && (max >> 52) == 0);
```
garazdawi, line 61: Is there a macro or constant we can use instead of 52 here?

garazdawi (approved):
I've checked the emu and x86 code and it looks good to me.

erts/emulator/beam/jit/arm/instr_bs.cpp
bjorng: I have replaced `52` with `(SMALL_BITS - ERL_UNIT_BITS)`, where `ERL_UNIT_BITS` is a new define I've added to `erl_bits.h`.

michalmuskala:
Would this optimise something like the following (which is how I'd decode varints or Unsigned LEB128 - common in many compact networking protocols):
```erlang
varint(<<0:1, Len:7, Rest/bits>>) ->
    continue(Rest, Len);
varint(<<1:1, Len1:7, 0:1, Len2:7, Rest/bits>>) ->
    Len = (Len2 bsl 7) bor Len1,
    continue(Rest, Len);
...
```

bjorng:
@michalmuskala Yes, this code will be nicely optimized by this pull request and #6259.

Internally in the compiler, your example would be rewritten to something similar like this:

```
varint(Bin) ->
    case Bin of
        <<X:1, Tail/bits>> ->
            case X of
                0 ->
                    case Tail of
                        <<Len:7, Rest/bits>> ->
                            continue(Rest, Len)
                    end;
                1 ->
                    case Tail of
                        <<Len1:7, 0:1, Len2:7, Rest/bits>> ->
                            Len = (Len2 bsl 7) bor Len1,
                            continue(Rest, Len)
                    end
            end
    end.
```

#6259 makes matching of segments having "odd" sizes such as `1` and `7` more efficient than in OTP 25.

This pull request will make the matching of `<Len1:7, 0:1, Len2:7, Rest/bits>>` more efficient. The entire match will be contained in one `bs_match` instruction:

```
    {bs_match,{f,4},
              {x,1},
              {commands,[{ensure_at_least,15,1},
                         {integer,2,{literal,[]},7,1,{x,2}},
                         {'=:=',3,1,0},
                         {integer,3,{literal,[]},7,1,{x,0}},
                         {get_tail,3,1,{x,1}}]}}.
```

Before this pull request, there would be two `bs_match` instructions with a `bs_match_string` instruction in between:

```
    {bs_match,{f,4},
              {x,1},
              {commands,[{ensure_at_least,7,1},
                         {integer,2,{literal,[]},7,1,{x,2}}]}}.
    {test,bs_match_string,{f,4},[{x,1},1,{string,<<0>>}]}.
    {bs_match,{f,4},
              {x,1},
              {commands,[{ensure_at_least,7,1},
                         {integer,3,{literal,[]},7,1,{x,0}},
                         {get_tail,3,1,{x,1}}]}}.
```

bjorng:
> It's a shame the two sub-byte reads are separated - I assume always reading full bytes is more efficient. In this case in particular it would be better to read `<<Tag:1, Len:7, ...>>` since we'll need the `Len` in both branches either way.

Well, doing one read instead of two read is faster. (Only one read instead of two would be needed for extracting a "short" length.)  Since #6259, there is less difference between reading whole bytes and parts of bytes or doing unaligned reads.

It is in fact possible to restructure the source code so we avoid separating the first two sub-byte reads:

```
varint(Bin) ->
    case Bin of
        <<X:1, Len1:7, Tail/bits>> ->
            case X of
                0 ->
                    continue(Tail, Len1);
                1 ->
                    case Tail of
                        <<0:1, Len2:7, Rest/bits>> ->
                            Len = (Len2 bsl 7) bor Len1,
                            continue(Rest, Len)
                    end
            end
    end.
```

I suppose that it would be possible for the compiler to do that kind of transformation by itself. I don't know how tricky that would be or how much it would give.

PR #6372, opened 2022-10-15
Improve boot and code loading error messages

josevalim:
Today if there is a crash during boot or eval,
because a more recent Erlang/OTP version is
required, the user is prompted with this message:

    $ erl -noshell -s oops
    {"init terminating in do_boot",{undef,
    [{oops,start,[],[]},{init,start_em,1,[]}
    ,{init,do_boot,3,[]}]}}
    init terminating in do_boot ({undef,[{oops,
    start,[],[]},{init,start_em,1,[]},{init,
    do_boot,3,[]}]})

    Crash dump is being written to:
    erl_crash.dump...done

Unfortunately, this message gives little indication
of what is wrong besides the "undef" atom. This
particularly problematic when users are running
a system software (Rebar, Elixir, etc) against an
old system version of Erlang/OTP.

This commit improves the boot error messages
such that:

  1. It removes the double reports from erlang:display/1 and crash dump

  2. Validates the module given to -s and attempts to provide a human error for common cases

  3. Makes sure the -eval failure is printed and uses a consistent format with -s

  4. Improve error messages to not include internal details such as "do_boot"

  5. It suggests that, besides using an old version of Erlang/OTP, they may also use a new one

Now we get:

    $ erl -noshell -s oops
    Error! Failed to load module oops because:
    the module cannot be found. Make sure the
    module name is correct and its .beam is in
    the code path

    Runtime terminating during boot ({undef,
    [{init,start_it,1,[]},{init,start_em,1,[]},
    {init,do_boot,3,[]}]})

    Crash dump is being written to:
    erl_crash.dump...done

If the .beam file exists but cannot be loaded:

    $ erl -noshell -s oops
    Error! Failed to load module oops because:
    the module requires a more recent Erlang/OTP
    version or its .beam was corrupted (you are
    running Erlang/OTP 26)

    Runtime terminating during boot ({undef,
    [{init,start_it,1,[]},{init,start_em,1,[]},
    {init,do_boot,3,[]}]})

    Crash dump is being written to:
    erl_crash.dump...done

### DIFF
--- a/erts/emulator/beam/beam_load.c
+++ b/erts/emulator/beam/beam_load.c
@@ -47,7 +47,7 @@ Uint erts_total_code_size;
 
 static int load_code(LoaderState *stp);
 
-#define PLEASE_RECOMPILE "please re-compile this module with an Erlang/OTP " ERLANG_OTP_RELEASE " compiler"
+#define PLEASE_RECOMPILE "please re-compile this module with an Erlang/OTP " ERLANG_OTP_RELEASE " compiler or update your Erlang/OTP version"
 
 /**********************************************************************/
 
--- a/erts/preloaded/src/init.erl
+++ b/erts/preloaded/src/init.erl
@@ -396,8 +396,8 @@ boot_loop(BootPid, State) ->
 	    loop(State#state{status = {started,PS},
 			     subscribed = []});
 	{'EXIT',BootPid,Reason} ->
-	    erlang:display({"init terminating in do_boot",Reason}),
-	    crash("init terminating in do_boot", [Reason]);
+	    % erlang:display({"init terminating in do_boot",Reason}),
+	    crash("Runtime terminating during boot", [Reason]);
 	{'EXIT',Pid,Reason} ->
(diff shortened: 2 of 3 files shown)


josevalim:
The test is failing because I have not updated the preloaded files.

bjorng:
I like the idea, but I have some nitpicks about the formatting. I have pushed a fixup commit with my modifications. If you agree, please squash the commits. I've done the following changes:

* Added CRs before LFs (`\r` before each `\n`). This is necessary when running in a terminal to get the following message produced by the runtime to start at the beginning of a line.
* I have rephrased the error messages to read more naturally and avoid the "because" followed by a colon.
* Since the messages are quite long, I using binaries instead of string literals to reduce the size of the code.

josevalim:
@bjorng thank you! Squashed and pushed!

bjorng:
Thanks! I've added the PR (along with an updated `init.beam`) to our daily builds.

bjorng:
Thanks!

Issue #6376, opened 2022-10-18
[MnesiaTM] Lightweight method for obtaining an active mnesia transactions counters


phiotr:
**Is your feature request related to a problem? Please describe.**

I am building a monitoring/visualization tool for the erlang system that heavily uses Mnesia. And instead of the basic counters available in the `mnesia_stats` table (accessible via mnesia:system_info/1) I wish to also include active transactions statistics from all of the cluster nodes. The naive version of `erlang:length(mnesia_tm:get_transactions())` poller is extremely dangerous during high load periods when the Mnesia can glitch itself and accumulate up to a few thousand waiting transactions to be processed. In such cases, calling the `mnesia_tm:get_transactions()` every few seconds becomes a significant part of the problem. Because the transaction manager is too busy with the preparation of the response for the monitoring tool, collecting a list of transactions, sending it (copying huge structures to the caller's message queue!), and pre-formatting, to handle these transactions. Even when a high-priority flag was set for `mnesia_tm`. So much work to be done and completely unnecessary when all I need at the end is just a number.

Seeing mnesia transactions glitches on the Prometheus/Grafana dashboards is very important for us for debugging and discovering any potential bottlenecks in the system. I could not find any lightweight way of achieving it without modifying mnesia_tm's codebase. 

**Describe the solution you'd like**

The perfect solution for this problem would be a very light method that counts in real-time the transactions stored inside of `mnesia_tm` process' state and returns just a number of active ones. Without copying long lists of detailed `#tid{}` records and formatting them.

I have prepared an example of `mnesia_tm:get_transactions_count/0` method that will be attached to the following pull request.

**Describe alternatives you've considered**

I tried to use the sys:install/2-3 and write down a custom one-shot debug function to obtain active transactions statistics, but hence mnesia_tm supports many of the system messages, it is not a `gen_server` and `{debug, {install, _}}` is not on that list.

```erlang
CallerPid = self(),
ReturnRef = erlang:make_ref(),
sys:install(
	erlang:whereis(mnesia_tm), 
	{fun(_, _, MnesiaTmState) -> 
		ParticipantsCount = gb_trees:size(erlang:element(3, MnesiaTmState)),
		CoordinatorsCount = gb_trees:size(erlang:element(2, MnesiaTmState)),
		CallerPid ! {ReturnRef, ParticipantsCount, CoordinatorsCount},
		done 
	 end, ok}
),

receive ...
```

[...]

PR #6377, opened 2022-10-18
[MnesiaTM] Lightweight method for obtaining an active mnesia transactions counters

phiotr:
This is a pull request for idea #6376. A simple and light way for obtaining the active mnesia transactions counters for monitoring purposes


dgud:
Can you add a testcase for regression on this.

phiotr:
Ok, I pushed one (very simple) test which starts a transaction and checks whether `mnesia_tm:get_transactions()` and `mnesia_tm:get_transactions_count()` returns the same statistics. Hmm... Do you think it is enough?

PR #6378, opened 2022-10-18
erts: Add acful, abandon carried free utilization limit

garazdawi:
This limit can be used to let erts_alloc know that it should
use MADV_FREE on any pages within a carrier that are currently
unused.

Before this change we have tried to have this as default, but that
caused performance issues as carriers oscillated in and out of the
migration pool. So in 25 we decided to remove it. However, this caused
MemAvailable from /proc/meminfo to become significantly lower and thus
machines that have a tight memory budget need this option.

So the solution is to add a new config option where the user can set
at which utilization limit that free pages in a carrier should be placed
in MemAvailable.

The option is by default set to 0, to mimic the behaviour of OTP 25.

### DIFF
--- a/erts/doc/src/erts_alloc.xml
+++ b/erts/doc/src/erts_alloc.xml
@@ -514,6 +514,29 @@
 	  <p>See also <seecref marker="#M_acul"><c>acul</c></seecref>.</p>
         </item>
 
+        <tag><marker id="M_acful"/><c><![CDATA[+M<S>acful <utilization>|de]]></c>
+        </tag>
+        <item>
+            <p>Abandon carrier free utilization limit. When the utilization of
+              a carrier falls belows this limit erts_alloc instructs the OS that
+              unused memory in the carrier can be re-used for allocation by other OS
+              procesesses. On Unix this is done by calling <c>madvise(..., ..., MADV_FREE)</c>
+              on the unused memory region, on Windows it is done by calling
+              <c>VirtualAlloc(..., ..., MEM_RESET, PAGE_READWRITE)</c>. Defaults to 0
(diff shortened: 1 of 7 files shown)


erts/emulator/beam/erl_alloc_util.h
```diff
@@ -694,9 +698,11 @@ struct Allctr_t_ {
         ErtsAlcCPoolData_t  *sentinel;
 
 	UWord		abandon_limit;
+        UWord           discard_limit;
```
sverker, line 701: `Allctr.cpool.discard_limit` do not seem to be used other than in ASSERT.

The `abandon_limit` check is done both on allocator level and on carrier, but the `discard_limit` check is only done on carrier.

erts/doc/src/erts_alloc.xml
```diff
@@ -514,6 +514,29 @@
...
+              a carrier falls belows this limit erts_alloc instructs the OS that
+              unused memory in the carrier can be re-used for allocation by other OS
+              procesesses. On Unix this is done by calling <c>madvise(..., ..., MADV_FREE)</c>
+              on the unused memory region, on Windows it is done by calling
+              <c>VirtualAlloc(..., ..., MEM_RESET, PAGE_READWRITE)</c>. Defaults to 0
+              which means that no memory will be marked as re-usable by the OS.</p>
+            <p>A valid
+              <c><![CDATA[<utilization>]]></c> is an integer in the range
+              <c>[0, 100]</c> representing utilization in percent. If this value is
+              larger than the <c>acul</c> limit it will be lowered the current
```
sverker, line 530: it will be lowered **_to_** the current acul limit

sverker (approved):

PR #6382, opened 2022-10-19
kernel: Fix os:cmd error report for open_port errors

garazdawi:
The error_info code for os:cmd inadvertently hid errors when open_port failed to start due to some reason, for example out of file descriptors.

This PR fixes the bug and adds tests to make sure it is correct.

Found while lurking here: https://elixirforum.com/t/error-when-using-os-cmd-not-a-list-of-characters-panic-kernel-8-4-2-os-erl-434/51123


lib/kernel/src/erl_kernel_errors.erl
```diff
@@ -71,6 +73,14 @@ format_os_error(unsetenv, [Name], _) ->
 format_os_error(_, _, _) ->
     [].
 
+maybe_posix_message(Reason) ->
+    case erl_posix_msg:message(Reason) of
+        "unknown POSIX error" ->
+            io_lib:format("open_port failed with readon: ~tp",[Reason]);
```
KennethL, line 79:
```suggestion
            io_lib:format("open_port failed with reason: ~tp",[Reason]);
```

KennethL (changes requested):

PR #6388, opened 2022-10-24
Further improve the code emitted by the JIT

bjorng:
This pull request improves the code emitted by the JIT for miscellaneous BEAM instructions, generally by taking advantage of types determined by the Erlang compiler.

Here are the major enhancements introduced in this pull request:

* Inlining of the following guard BIFs when the types of their arguments are known: `bit_size/1`, `byte_size/1`, `element/2`, `map_size/1`.
* Generating better code for `is_map_key/2` and `map_get/2` when the second argument is a known map.
* Generating better code for the relational operators `<`, `=<`, `>=`, and `>` when used as guard BIFs returning a boolean.
* Some improvements of the guard tests `is_binary/1` and `is_bitstring/1`.
* Enhancement of the optimization introduced in #6298 that avoids loading a BEAM register whose value is already present in a CPU register.

There are more details in the individual commit messages.


erts/emulator/beam/jit/x86/instr_guard_bifs.cpp
```diff
@@ -91,6 +93,121 @@ void BeamModuleAssembler::emit_bif_is_ne_exact(const ArgRegister &LHS,
     emit_bif_is_eq_ne_exact(LHS, RHS, Dst, am_true, am_false);
 }
 
+void BeamModuleAssembler::emit_cond_to_bool(uint32_t instId,
+                                            const ArgRegister &Dst) {
+    mov_imm(RET, am_true);
+    mov_imm(ARG1, am_false);
+    a.emit(instId, RET, ARG1);
+    mov_arg(Dst, RET);
```
michalmuskala, line 103: I wonder, if it would make sense to reserve some special integer values for the true and false atoms. In particular, it could be possible to make the representation differ by only one bit. 
This might allow using some combination of `CMP` + `SETcc` + `ADD` or `AND` which might be shorter than this.

bjorng: We already have assigned special values for `true` and `false` atoms. We take advantage of that in the `is_boolean/1` guard test:

https://github.com/erlang/otp/blob/cdc7c4c6b46ad9558bf4eec20e7a75cc8e5ff196/erts/emulator/beam/jit/x86/instr_common.cpp#L895-L907

We currently don't have type information about booleans in the BEAM files (we know when something is an atom, but not which possible atoms), so we can't do the trick of comparing only a single bit.

For the boolean guards BIFs, the current sequence is:

```
mov eax, 75
mov edi, 11
cmove rax, rdi
```
which is 14 bytes.

I think this can be rewritten like this (not tested):

```
setne al
movzx eax, al
shl al, 6
or al, 11
```

which is 4 instructions instead of 3, but only 11 bytes instead of 14. I am not sure if there is any difference in speed.

Since my current version is slightly simpler, I think I'll keep it.

michalmuskala: Makes sense. Thanks for the explanation.

Issue #6368, opened 2022-10-14
systools:make_script error reporting is misleading


max-au:
**Describe the bug**
When an application contains duplicate dependencies, systools:make_script fails with "Undefined applications: [dep]". 

**To Reproduce**
Example of myapp.app:
```
{application, myapp, [
    {vsn, "1.0.0"},
    {applications, [kernel, stdlib,
        dep, dep]},
]}.
```
Note that "dep" dependency is mentioned twice. This is not an error anywhere, until I try to build a release with myrel.rel file:

```
{release,{"myrel","1.0.0"},
         {erts,"13.0.2"},
         [{myapp,"1.0.0"},
          {kernel,"8.4.1"},
          {stdlib,"4.0.1"},
          {dep, "1.0.0"}]}.
```

Both `dep` and `myapp` are in the myrel.rel file, but since `dep` is mentioned twice in myapp.app, `systools_make:check_applications/1` throws "Undefined applications: [dep]". 

**Expected behavior**
I would expect to see an error like "myapp has duplicate dependency on dep".

**Affected versions**
All of them.

frazze-jobb:
Thanks for the bug report, I fixed it in #6389

PR #6389, opened 2022-10-24
sasl: add error for duplicate atoms in app parameters

frazze-jobb:
Fixes bug reported in GH-6368

This PR adds functionality to **systools:make_script**, to check .app parameters for duplicates.
The parameters that will be checked are **modules**, **applications** and **registered**.

### DIFF
--- a/lib/sasl/src/sasl.app.src
+++ b/lib/sasl/src/sasl.app.src
@@ -42,6 +42,6 @@
     {applications, [kernel, stdlib]},
     {env, []},
     {mod, {sasl, []}},
-    {runtime_dependencies, ["tools-2.6.14","stdlib-3.4","kernel-6.0",
+    {runtime_dependencies, ["tools-2.6.14","stdlib-4.0","kernel-6.0",
 			    "erts-10.2"]}]}.
 
--- a/lib/sasl/src/systools_make.erl
+++ b/lib/sasl/src/systools_make.erl
@@ -696,7 +696,15 @@ specified([], _) ->
     [].
 
 get_items([H|T], Dict) ->
-    Item = check_item(keysearch(H, 1, Dict),H),
+    Item = case check_item(keysearch(H, 1, Dict),H) of
+        [Atom|_]=Atoms when is_atom(Atom), is_list(Atoms) ->
+            %% Check for duplicate entries in lists
+            case Atoms =/= lists:uniq(Atoms) of
+                true -> throw({dupl_entry, H, lists:subtract(Atoms, lists:uniq(Atoms))});
+                false -> Atoms
+            end;
+        X -> X
+    end,
     [Item|get_items(T, Dict)];
 get_items([], _Dict) ->
     [].
@@ -2436,6 +2444,8 @@ form_reading({read,File}) ->
     io_lib:format("Cannot read ~tp~n",[File]);
 form_reading({{bad_param, P},_}) ->
     io_lib:format("Bad parameter in .app file: ~tp~n",[P]);
+form_reading({{dupl_entry, P, DE},_}) ->
+    io_lib:format("~tp parameter contains duplicates of: ~tp~n", [P, DE]);
 form_reading({{missing_param,P},_}) ->
     io_lib:format("Missing parameter in .app file: ~p~n",[P]);
 form_reading({badly_formatted_application,_}) ->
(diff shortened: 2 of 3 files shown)


lib/sasl/src/systools_make.erl
```diff
@@ -696,7 +696,15 @@ specified([], _) ->
     [].
 
 get_items([H|T], Dict) ->
-    Item = check_item(keysearch(H, 1, Dict),H),
+    Item = case check_item(keysearch(H, 1, Dict),H) of
+        [Atom|_]=Atoms when is_atom(Atom), is_list(Atoms) ->
+            %% Check for duplicate entries in lists
+            case Atoms =/= lists:uniq(Atoms) of
+                true -> throw({dupl_entry, H, lists:subtract(Atoms, lists:uniq(Atoms))});
+                _ -> Atoms
```
bjorng, line 704: Clearer, may lead to better types in Dialyzer, inexpensive or free:

```suggestion
                false -> Atoms
```

lib/sasl/test/systools_SUITE.erl
```diff
@@ -467,6 +467,19 @@ variable_script(Config) when is_list(Config) ->
     ok = file:set_cwd(OldDir),
     ok.
 
+%% make_script: Duplicate entries in app file
+duplicate_entries_script(Config) when is_list(Config) ->
+    DataDir = ?datadir,
+    %Name = fname(PrivDir, duplicate_entries),
```
bjorng, line 473: Don't comment out code:

```suggestion
```

lib/sasl/src/systools_make.erl
```diff
@@ -696,7 +696,15 @@ specified([], _) ->
     [].
 
 get_items([H|T], Dict) ->
-    Item = check_item(keysearch(H, 1, Dict),H),
+    Item = case check_item(keysearch(H, 1, Dict),H) of
+        [Atom|_]=Atoms when is_atom(Atom), is_list(Atoms) ->
+            %% Check for duplicate entries in lists
+            case Atoms =/= lists:uniq(Atoms) of
```
bjorng, line 702: You will need to update the dependencies for `sasl` application since you use a function was introduced in OTP 25.

lib/sasl/src/sasl.app.src
```diff
@@ -42,6 +42,6 @@
     {applications, [kernel, stdlib]},
     {env, []},
     {mod, {sasl, []}},
-    {runtime_dependencies, ["tools-2.6.14","stdlib-3.4","kernel-6.0",
+    {runtime_dependencies, ["tools-2.6.14","stdlib-4.1.1","kernel-6.0",
```
rickard-green, line 45: The `stdlib` dependency should only be bumped to `stdlib-4.0` if the bump is only due to using `lists:uniq()`

lib/sasl/src/sasl.appup.src
```diff
@@ -35,12 +35,16 @@
...
   {<<"^4\\.0\\.0(?:\\.[0-9]+)+$">>,[restart_new_emulator]},
   {<<"^4\\.0\\.1(?:\\.[0-9]+)*$">>,[restart_new_emulator]},
   {<<"^4\\.0\\.2(?:\\.[0-9]+)*$">>,[restart_new_emulator]},
   {<<"^4\\.1$">>,[restart_new_emulator]},
   {<<"^4\\.1\\.0(?:\\.[0-9]+)+$">>,[restart_new_emulator]},
   {<<"^4\\.1\\.1(?:\\.[0-9]+)*$">>,[restart_new_emulator]},
-  {<<"^4\\.1\\.2(?:\\.[0-9]+)*$">>,[restart_new_emulator]}]}.
+  {<<"^4\\.1\\.2(?:\\.[0-9]+)*$">>,[restart_new_emulator]},
+  {<<"^4\\.2$">>,[restart_new_emulator]}
+]}.
```
rickard-green, line 50: The `sasl.appup.src` file is automatically updated based on automatically selected versions, so you should not update it.

lib/sasl/vsn.mk
```diff
@@ -1 +1 @@
-SASL_VSN = 4.2
+SASL_VSN = 4.2.1
```
rickard-green, line 1: Versions are automatically updated based on ticket states, so you should not update the version manually.

frazze-jobb: Right thanks, done.

rickard-green (approved):

PR #6390, opened 2022-10-24
Boost thread progress later op execution

sverker:
To keep up and not lag behind for too long with memory deallocation jobs.
    
Symptom seen:
`ets_SUITE:table_leak`, which repeatedly creates and deletes ETS tables, caused memory consumption running amok as scheduled later ops to free large lock arrays could not keep up.


sverker:
**Merged** to maint for 25.2.

PR #6391, opened 2022-10-24
ftp: add missing case nlist

### LINKED ISSUE #5823: ftp:nlist returns `ok` which is unexpected
**Describe the bug**
Frequently, when calling `ftp:nlist/2` a bare `ok` atom is returned, as opposed to a list result or error.

I reproduce it at `24.2` and   I was also able to reproduce with OTP `24.3.2`

Basically the same issue as described here https://github.com/erlang/otp/issues/4177

**To Reproduce**
Call multiple times :ftp.nlist(pid, path) it will return `:ok` atom 

**Expected behavior**
Expected `{:ok, result}` or `{:error, reason}`

**Affected versions**
24.2
24.3.2 (latest)

kikofernandez:
This PR adds a missing pattern match case to the function `handle_ctrl_result` where the `caller` record attribute has 2 arguments instead of 3 arguments. When `caller` has 2 arguments and the data necessary to return it, we will return it. It seems like due to the timing of events, sometimes the `state` record may have the `caller` set with a 2-tuple or 3-tuple value.
Without the case added in here, one gets spurious results as `ok` , when one expects `{ok, "foobar\r\nbaz"}`

Closes #5823

### DIFF
--- a/lib/ftp/src/ftp.erl
+++ b/lib/ftp/src/ftp.erl
@@ -1773,23 +1773,25 @@ handle_ctrl_result({pos_compl, _}, #state{caller = {handle_dir_result, Dir,
             %% %% overhead to be able to give a good return value. Alas not
             %% %% all ftp implementations behave the same and returning
             %% %% an error string is allowed by the FTP RFC.
-            %% case lists:dropwhile(fun(?CR) -> false;(_) -> true end,
-            %%                          binary_to_list(Data)) of
-            %%         L when (L =:= [?CR, ?LF]) orelse (L =:= []) ->
-            %%             send_ctrl_message(State, mk_cmd("PWD", [])),
-            %%             activate_ctrl_connection(State),
-            %%             {noreply,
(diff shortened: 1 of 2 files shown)


lib/ftp/src/ftp.erl
```diff
@@ -1791,6 +1822,27 @@ handle_ctrl_result({pos_compl, _}, #state{caller = {handle_dir_result, Dir,
...
+            {noreply, State#state{client = undefined,
+                                  caller = undefined}};
+        [] ->
+            %% end of stream
+            gen_server:reply(From, {ok, Data}),
+            {noreply, State#state{client = undefined,
+                                  caller = undefined}};
+        _ ->
+            %% recurse the state one more time until we reach the end of the stream
+            {noreply, State#state{caller = {handle_dir_result, Dir, Data}}}
```
kikofernandez, line 1843: This is not right and may leaves the state machine in a stuck state

IngelaAndin (approved):

Issue #3771, opened 2019-02-12
ERL-853: use_srtp extension (DTLS/SRTP)


OTP-Maintainer:
Original reporter: `benbro`
Affected version: `OTP-21.2.3`
Component: `ssl`
Migrated from: https://bugs.erlang.org/browse/ERL-853

---

```
The use_srtp extension is needed for DTLS/SRTP (WebRTC) and TLS-1.3.

There is currently no way to manually add the extension with a transport module without modifying the ssl app source.

The extension is defined in the code but not used:
https://github.com/erlang/otp/blob/master/lib/ssl/src/dtls_handshake.hrl#L65
https://github.com/erlang/otp/blob/master/lib/ssl/src/ssl_handshake.erl#L3007
```

PR #6392, opened 2022-10-25
ssl: implement use_srtp hello extension with options

stolen:
This PR implements encoding and decoding of use_srtp hello extension (RFC 5764) which is necessary for setting up an SRTP channel. SRTP is used in WebRTC.

This feature was requested in #3771

To add use_srtp to hello extensions, you need to pass 'use_srtp_protection_profiles' option, e.g.
  `ssl:connect("localhost", 17423, [{protocol, dtls},
    {use_srtp_protection_profiles, [<<0,1>>, <<0,5>>]}, {handshake, hello}]).`
Proper value for use_srtp_protection_profiles option is a list of 2-byte binary profile ids (see section 4.1.2 of RFC 5764)

Optionally you may specify use_srtp_mki option to set srtp_mki field (I don't need it now but someone may need it later).

### DIFF
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -1231,6 +1231,36 @@ fun(srp, Username :: binary(), UserState :: term()) ->
 	</desc>
       </datatype>
 
+      <datatype>
+        <name name="use_srtp_protection_profiles"/>
+        <desc>
+          <p>Configures the list of the client's acceptable SRTP Protection Profiles.</p>
+          <p>In order to negotiate the use of SRTP data protection, clients
+          include an extension of type "use_srtp" in the DTLS extended client
+          hello.  This extension MUST only be used when the data being
+          transported is RTP or RTCP.</p>
+          <p>This extension MUST only be used with DTLS, and not with TLS</p>
+          <p>OTP does not handle SRTP, so an external implementation of SRTP
+          encoder/decoder and a packet demultiplexer are needed to make use
+          of the use_srtp extension. See also <c>cb_info</c> option</p>
+        </desc>
+      </datatype>
+
+      <datatype>
+        <name name="use_srtp_mki"/>
+        <desc>
+          <p>Configures the SRTP Master Key Identifier chosen by the client.</p>
+          <p>This options only has effect only when <c>use_srtp_protection_profiles</c>
+          option is specified.</p>
+          <p>The srtp_mki field contains the value of the SRTP MKI which is associated
+          with the SRTP master keys derived from this handshake.  Each SRTP
+          session MUST have exactly one master key that is used to protect
+          packets at any given time.  The client MUST choose the MKI value so
+          that it is distinct from the last MKI value that was used, and it
+          SHOULD make these values unique for the duration of the TLS session.</p>
+        </desc>
+      </datatype>
+
     <!--   <datatype> -->
   <!-- 	<name name="ocsp_stapling"/> -->
   <!-- 	<desc><p>If true, OCSP stapling will be enabled, an extension of type "status_request" will be -->
@@ -1579,6 +1609,31 @@ fun(srp, Username :: binary(), UserState :: term()) ->
           </p>
 	  <warning><p>This option is a placeholder, early data is not yet implemented on the server side.
           </p></warning>
+        </desc>
+      </datatype>
+
+      <datatype>
+        <name name="use_srtp_protection_profiles"/>
+        <desc>
+          <p>Configures the server's chosen SRTP Protection Profile (as a list of a single profile).</p>
+          <p>Servers that receive an extended hello containing a "use_srtp"
+          extension can agree to use SRTP by including an extension of type
+          "use_srtp", with the chosen protection profile in the extended server
+          hello.</p>
+        </desc>
+      </datatype>
+
+      <datatype>
+        <name name="use_srtp_mki"/>
+        <desc>
+          <p>Configures the server's SRTP Master Key Identifier.</p>
+          <p>Upon receipt of a "use_srtp" extension containing a "srtp_mki" field,
+          the server MUST either (assuming it accepts the extension at all):</p>
+	  <list type="bulleted">
+	    <item><p>include a matching "srtp_mki" value in its "use_srtp" extension
+            to indicate that it will make use of the MKI, or</p></item>
+	    <item><p>return an empty "srtp_mki" value to indicate that it cannot make use of the MKI.</p></item>
+	  </list>
         </desc>
       </datatype>
 
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -1232,32 +1232,30 @@ fun(srp, Username :: binary(), UserState :: term()) ->
       </datatype>
 
       <datatype>
-        <name name="use_srtp_protection_profiles"/>
+        <name name="use_srtp"/>
         <desc>
-          <p>Configures the list of the client's acceptable SRTP Protection Profiles.</p>
+          <p>Configures the <c>use_srtp</c> DTLS hello extension.</p>
           <p>In order to negotiate the use of SRTP data protection, clients
           include an extension of type "use_srtp" in the DTLS extended client
           hello.  This extension MUST only be used when the data being
           transported is RTP or RTCP.</p>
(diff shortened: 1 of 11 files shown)


stolen:
I've noticed ssl handle_options was refactored in master. I can port my commit to the new way of handling options, if needed.

lib/ssl/src/dtls_handshake.hrl
```diff
@@ -51,7 +51,6 @@
 %% RFC 7764  Datagram Transport Layer Security (DTLS) Extension to Establish Keys
```
u3s, line 51: RFC7764 seems not correct any longer. Replace it with RFC5764.

stolen: Yep, seems like an old typo. Fixed.

stolen:
> am I correct thinking this option is only for DTLS1.2 version?

I never wondered about protocol compatibility. I didn't find any strict rules, only this mentions:

https://www.rfc-editor.org/rfc/rfc5764#section-4.1.2
```
Note that this specification is compatible with DTLS 1.0
      or DTLS 1.2
```

Other extenson is meant to be used along with `use_srtp` in DTLS 1.3:
https://www.rfc-editor.org/rfc/rfc8870#section-5.2

So, seems like any DTLS version is compatible.  

Also https://www.rfc-editor.org/rfc/rfc5764#section-9
```
This extension MUST only be used with DTLS, and not with TLS
```
So, it seems proper to throw an error if use_srtp is specified for non-DTLS protocol, not just inform user. I'll do that.

u3s:
> So, it seems proper to throw an error if use_srtp is specified for non-DTLS protocol, not just inform user. I'll do that.

Yeah, this is what I meant maybe not clearly enough :-).
`ssl:assert_version_dep/4` might be useful.

stolen:
added a protocol dependency

u3s (commented):
`srtp_profiles` test case fails with latest commit. please see other comments inline.

stolen:
I had a problem with stateful config parsing. Solved by storing default MKI value `<<>>` instead of undefined.  
The long lines are shortened, test client/server start calls are simplified.  
Please take a look.

IngelaAndin:
@stolen Sorry for the delay there has been many TLS-related work packages so we have been quite busy. I and some co-workers have been looking into the RFC:s for this a little further. So the first question-mark is, that as far as we can tell only handling the extension enoding/decoding and picking a profile will not be sufficient?!  It looks like there also needs to be some package recognition/mappings and callbacks to handlers that will do SRTP encoding.
 
Also, we have been doing other changes that will require you to rebase.

Latest test also fails:

*** CT Error Notification 2023-01-23 08:00:47.953 ***🔗
ssl_handshake_SUITE:encode_decode_srp failed on line 222
Reason: {badmatch,<<0,20,0,0,0,8,0,6,0,...>>}

=== Ended at 2023-01-23 08:00:47
=== Location: [{ssl_handshake_SUITE,encode_decode_srp,222},
              {test_server,ts_tc,1782},
              {test_server,run_test_case_eval1,1291},
              {test_server,run_test_case_eval,1223}]
=== === Reason: no match of right hand side value 
                 <<0,20,0,0,0,8,0,6,0,0,3,98,97,114,0,12,0,4,3,102,111,111>>
  in function  ssl_handshake_SUITE:encode_decode_srp/1 (ssl_handshake_SUITE.erl, line 222)
  in call from test_server:ts_tc/3 (test_server.erl, line 1782)
  in call from test_server:run_test_case_eval1/6 (test_server.erl, line 1291)
  in call from test_server:run_test_case_eval/9 (test_server.erl, line 1223)

stolen:
> also needs to be some package recognition/mappings and callbacks to handlers that will do SRTP encoding

Yes, I consider that orthogonal task to extension handling.  
Our code uses a separate worker to do ICE+SRTP stuff, forwarding DTLS packets to the dtls_connection. We indeed use custom callback module (specified via `cb_info`) for outgoing packets.  

> require you to rebase

OK, will check that.

> ssl_handshake_SUITE:encode_decode_srp

Seems unrelated. I hope this test will be OK after proper rebase.

stolen:
Rebased.  
CI has failures in `ssl_dist_SUITE` — passes locally. Maybe unstable test.  
Re-checked `ssl_handshake_SUITE` locally — no failures.

@IngelaAndin do you expect anything more from me on this PR?

lib/ssl/src/ssl.erl
```diff
@@ -2233,6 +2242,27 @@ opt_handshake(UserOpts, Opts, _Env) ->
 
     Opts#{handshake => HS, max_handshake_size => MHSS}.
 
+opt_use_srtp(UserOpts, #{protocol := Protocol} = Opts, _Env) ->
```
dgud, line 2245: Should we not make the user option a map here to reduce the number of options?
{use_strp, #{mki := Mki, protection_profiles := PPs} 

When included in Opts it shall be a map,  and it shall not be included in Opts if it is not enabled.
We are trying to reduce the size of Opts map.

lib/ssl/doc/src/ssl.xml
```diff
@@ -1231,6 +1231,33 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
+          option is specified.</p>
+          <p>The srtp_mki field contains the value of the SRTP MKI which is associated
+          with the SRTP master keys derived from this handshake.  Each SRTP
+          session MUST have exactly one master key that is used to protect
+          packets at any given time.  The client MUST choose the MKI value so
+          that it is distinct from the last MKI value that was used, and it
+          SHOULD make these values unique for the duration of the TLS session.</p>
+        </desc>
+      </datatype>
+
```
dgud, line 1260: It should be mentioned that the user is required to handle SRTP via the via a custom backend so that
nobody expects this to work of shelf.

lib/ssl/test/ssl_api_SUITE.erl
```diff
@@ -2145,7 +2145,9 @@ invalid_options(Config) when is_list(Config) ->
 
     TestOpts2 =
         [{[{supported_groups, []}, {versions, [tlsv1]}],
-          {options,incompatible,[supported_groups,{versions,['tlsv1']}]}}],
+          {options,incompatible,[supported_groups,{versions,['tlsv1']}]}},
+         {[{use_srtp_protection_profiles, [<<0,2>>]}],
+          {options,incompatible,[use_srtp_protection_profiles,{protocol,tls}]}}],
```
dgud, line 2151: Move these tests to the whitebox testcase instead of here.

lib/ssl/src/ssl.erl
stolen: Reducing the size of Opts is good.  
Are there any options implemented this way?  
At first I was going to implement options exactly as you propose, but then decided to do it more consistently with majority of other options (or all of them?).

dgud: With a map as input?
  Only certs_keys comes to mind currently (but we want to change ocsp to be a map).
  
 Search for set_opt_new for example of optionally setting Opts map.

IngelaAndin:
Could you also update the table in https://www.erlang.org/doc/apps/ssl/standards_compliance.html ?

stolen:
Updated `standards_compliance`. Didn't know which version to write in Since, guessed 26.0.

IngelaAndin:
> Updated `standards_compliance`. Didn't know which version to write in Since, guessed 26.0.

Yes, it should be 26.0 as it is master. We have an internal way to get it filled in but that is mainly useful if you do not know which version the code will be included in when you write it. As this is not the case now I think just putting 26.0 will be easier.

lib/ssl/src/ssl.erl
dgud: Looks good to me, tough a question should we translate use_srtp record in the internal options directly?
Then we don't need the translation in "runtime".

dgud (approved):

stolen:
@IngelaAndin This PR still has `waiting` label while it is approved by Dan Gudmundsson.  
Do you expect anything else from me here?

IngelaAndin:
@stolen not at the moment, we forgot to remove the label. We are now waiting for some test cuddling so that the tests can pass.  The previous failing test case was not your fault, hopefully, they should pass now.

IngelaAndin:
@stolen I do need you to rebase on the latest master. I removed a test case that was redundant and problematic as that specific case relied on the map order which you may not rely on!

PR #6398, opened 2022-10-28
ssl test cuddle

dgud:
Cache openssl versions, speed up tests on windows where starting openssl via wsl takes ~0.5 seconds and have become slower on windows 11 which causes tests to time out.


dgud:
Merged

PR #6401, opened 2022-10-29
dec_term: must use unsigned for BINARY_EXT size

### LINKED ISSUE #6393: Cannot allocate 18446744071562067968 bytes on erlang:binary_to_term/1
**Describe the bug**

Take this code:

```erlang
Bin = binary:copy(<<0>>, 2 * 1024 * 1024 * 1024),
erlang:binary_to_term(erlang:term_to_binary(Bin)).
```

It fails with reason:

```
binary_alloc: Cannot allocate 18446744071562067968 bytes of memory (of type "binary").

```

**Expected behavior**

I would expect it to succeed or a clearer error message. :)

**Affected versions**

I have seen the failure on Erlang/OTP 24 and 25, on both Linux and macOS.

mikpe:
Fixes #6393

The fix needs to be backported to OTP-25 and OTP-24 (OTP-23 doesn't crash but throws an IMO bogus error instead).

There may be similar bugs elsewhere in this code, but I haven't investigated that in detail (yet).


sverker:
@mikpe I rebased your commit on OTP-23.3.4, added some more fixes and tests, and then merged it forward to 24.3.4 to solve some conflicts. And then force push to your repo to keep this PR going. Hope that was ok.

mikpe:
> @mikpe I rebased your commit on OTP-23.3.4, added some more fixes and tests, and then merged it forward to 24.3.4 to solve some conflicts. And then force push to your repo to keep this PR going. Hope that was ok.

@sverker no problem

PR #6408, opened 2022-10-31
Add new XML tag <change>

sverker:
Just like `<note>` boxes but with a different color and label. To make it easier to find (or ignore) talk about semantic differences between OTP versions.

For demonstration, the `erlang` module has been updated with `<change>` tags.
Example: https://erlang.github.io/prs/6408/erts-13.1.2/doc/html/erlang.html#binary_to_atom-2


kikofernandez (approved):
The change looks good.

PR #6413, opened 2022-11-01
kernel: support paste on terminals that does not filter paste brackets

frazze-jobb:


frazze-jobb:
This was reported by Dan, but neither him or I was able to replicate his issue, so I'm not 100% sure yet that this solves it. Therefore I prefer to wait with this one until someone reports it again and I can replicate it.

PR #6415, opened 2022-11-02
Lift restrictions for matching of binaries and maps

### LINKED ISSUE #6348: "binary patterns cannot be matched in parallel using =" on an assignment
```
foo(X) ->
  <<>> = (<<>> = 42).
```
is rejected by erlc with test.erl:5:3: binary patterns cannot be matched in parallel using '='.

This surprised me because I expected it to be parsed as follows:
```
foo(X) ->
  X = (<<>> = 42),
  <<>> = X.
```
Which is accepted.

Even weirder, the following is accepted:
```
foo(X) ->
  <<>> = begin <<>> = 42 end.
```
Showing that `(Expr)` and `begin Expr end` are not treated equivalently by the compiler.

Looking at the documentation in
- https://www.erlang.org/doc/reference_manual/expressions.html#parenthesized-expressions
- https://www.erlang.org/doc/reference_manual/expressions.html#bit-syntax-expressions
- https://www.erlang.org/doc/programming_examples/bit_syntax.html#segments

I could not find anything warning of this rather surprising behavior.

So I don't know whether this is a bug in erlc or just a missing piece of documentation (or maybe it is documented somewhere else?).

### LINKED ISSUE #6444: [erlc] internal error in pass beam_kernel_to_ssa
On master, both with and without cherry-picking https://github.com/erlang/otp/pull/6415/commits/2983a2585a494a99a921ae5446a59bc594a8571c:
```erlang
-module(kernel_to_ssa).
-compile([export_all]).

f(X) ->
    [<<_:Y>> | _] = [_ | Y] = X.
```
causes erlc to crash with the following error message:
```
Function: f/1
minimized/kernel_to_ssa.erl: internal error in pass beam_kernel_to_ssa:
exception error: bad key: 'Y'
  in function  map_get/2
     called as map_get('Y',
 #{0 => {b_var,0},
   6 => {b_var,6},
   7 => {b_var,7},
   8 => {b_var,8},
   '@ssa_ret' => {b_var,'@ssa_ret'}})
     *** argument 1: not present in map
  in call from beam_kernel_to_ssa:ssa_arg/2 (beam_kernel_to_ssa.erl, line 1127)
  in call from beam_kernel_to_ssa:select_extract_bin/9 (beam_kernel_to_ssa.erl, line 445)
  in call from beam_kernel_to_ssa:select_bin_seg/4 (beam_kernel_to_ssa.erl, line 394)
  in call from beam_kernel_to_ssa:select_binary/5 (beam_kernel_to_ssa.erl, line 314)
  in call from beam_kernel_to_ssa:select_cons/5 (beam_kernel_to_ssa.erl, line 300)
  in call from beam_kernel_to_ssa:match_cg/3 (beam_kernel_to_ssa.erl, line 177)
  in call from beam_kernel_to_ssa:do_match_cg/3 (beam_kernel_to_ssa.erl, line 167)
```

### LINKED ISSUE #6467: [erlc] internal crash in pass beam_kernel_to_ssa
```erlang
-module(kernel_to_ssa4).

f(X, _, []) ->
    [
        0
     || _ <- (<<_:Y>> = (Y = ((_ = X) = X)))
    ];
f(_, <<Y>>, _) ->
    a.
```
crashes erlc with the following error message:
```
Function: f/3
minimized/kernel_to_ssa4.erl: internal error in pass beam_kernel_to_ssa:
exception error: bad key: 'Y'
  in function  map_get/2
     called as map_get('Y',
 #{0 => {b_var,0},
   1 => {b_var,1},
   2 => {b_var,2},
   30 => {b_var,30},
   '@ssa_ret' => {b_var,'@ssa_ret'}})
     *** argument 1: not present in map
  in call from beam_kernel_to_ssa:ssa_arg/2 (beam_kernel_to_ssa.erl, line 1127)
  in call from beam_kernel_to_ssa:select_extract_bin/9 (beam_kernel_to_ssa.erl, line 445)
  in call from beam_kernel_to_ssa:select_bin_seg/4 (beam_kernel_to_ssa.erl, line 394)
  in call from beam_kernel_to_ssa:select_binary/5 (beam_kernel_to_ssa.erl, line 314)
  in call from beam_kernel_to_ssa:match_cg/3 (beam_kernel_to_ssa.erl, line 177)
  in call from beam_kernel_to_ssa:do_match_cg/3 (beam_kernel_to_ssa.erl, line 167)
  in call from beam_kernel_to_ssa:cg/2 (beam_kernel_to_ssa.erl, line 142)
```

This affects master. Cherry-picking 55eae53f9911b5eeb615081bc691c23df4e3e92e and 28fce1f8016a8bf52ca6a109aa471700f17e772b does not fix the issue.

This affects the same pass as https://github.com/erlang/otp/issues/6465, but the error message and testcase are sufficiently different that I believe it to probably be a different bug.

bjorng:
Lift restrictions for matching of binaries and maps

There has always been an implementation limitation for matching
of binaries (for technical reasons). For example:

    foo(Bin) ->
        <<A:8>> = <<X:4,Y:4>> = Bin,
        {A,X,Y}.

This would fail to compile with the following message:

    t.erl:5:5: binary patterns cannot be matched in parallel using '='
    %    5|     <<A:8>> = <<X:4,Y:4>> = Bin,
    %     |     ^

This commit lifts this restriction, making the example legal.

A restriction for map matching is also lifted, but before we can
describe that, we'll need a digression to talk about the `=` operator.

The `=` operator can be used for two similar but slightly differently
purposes.

When used in a pattern in a clause, for example in a function head,
both the left-hand and right-hand side operands must be patterns:

    Pattern1 = Pattern2

For example:

    bar(#{a := A} = #{b := B}) -> {A, B}.

The following example will not compile because the right-hand side
is not a pattern but an expression:

    wrong(#{a := A} = #{b => B}) -> {A, B}.

    t.erl:4:23: illegal pattern
    %    4| wrong(#{a := A} = #{b => B}) -> {A, B}.
    %     |                       ^

Used in this context, the `=` operator does not imply that the two
patterns are matched in any particular order. Attempting to use a
variable matched out on the left-hand side on the right-hand side, or
vice versa, will fail:

    also_wrong1(#{B := A} = #{b := B}) -> {A,B}.
    also_wrong2(#{a := A} = #{A := B}) -> {A,B}.

    t.erl:6:15: variable 'B' is unbound
    %    6| also_wrong1(#{B := A} = #{b := B}) -> {A,B}.
    %     |               ^

    t.erl:7:27: variable 'A' is unbound
    %    7| also_wrong2(#{a := A} = #{A := B}) -> {A,B}.
    %     |                           ^

The other way to use `=` is in a function body. Used in this way,
the right-hand side must be an expression:

    Pattern = Expression

For example:

    foobar(Value) ->
        #{a := A} = #{a => Value},
        A.

Used in this context, the right-hand side of `=` must **not** be a pattern:

    illegal_foobar(Value) ->
        #{a := A} = #{a := Value},
        A.

    t.erl:18:21: only association operators '=>' are allowed in map construction
    %   18|     #{a := A} = #{a := Value},
    %     |                     ^

When used in a body context, the value of the `=` operator is the
value of its right-hand side operand. When multiple `=` operators are
combined, they are evaluted from right to left. That means that any
number of patterns can be matched at once:

    Pattern1 = Pattern2 = ... = PatternN = Expr

which is equivalent to:

    Var = Expr
    PatternN = Var
       .
       .
       .
    Pattern2 = Var
    Pattern1 = Var

Given that there is a well-defined evaluation order from right to
left, one would expect that the following example would be legal:

    baz(M) ->
        #{K := V} = #{k := K} = M,
        V.

It is not. In Erlang/OTP 25 or earlier, the compilation fails with the
following message:

    t.erl:28:7: variable 'K' is unbound
    %   28|     #{K := V} = #{k := K} = M,
    %     |       ^

That restriction is now lifted, making the example legal.

Closes #6348
Closes #6444
Closes #6467

### DIFF
--- a/lib/compiler/src/v3_core.erl
+++ b/lib/compiler/src/v3_core.erl
@@ -81,8 +81,8 @@
 
 -export([module/2,format_error/1]).
 
--import(lists, [reverse/1,reverse/2,map/2,member/2,foldl/3,foldr/3,mapfoldl/3,
-                splitwith/2,keyfind/3,sort/1,droplast/1,last/1,
+-import(lists, [any/2,reverse/1,reverse/2,map/2,member/2,foldl/3,foldr/3,mapfoldl/3,
+                splitwith/2,keydelete/3,keyfind/3,keymember/3,sort/1,droplast/1,last/1,
                 duplicate/2]).
 -import(ordsets, [add_element/2,del_element/2,is_element/2,
 		  union/1,union/2,intersection/2,subtract/2]).
@@ -896,85 +896,47 @@ expr({call,L,FunExp,As0}, St0) ->
     Lanno = lineno_anno(L, St2),
     {#iapply{anno=#a{anno=Lanno},op=Fun,args=As1},Fps ++ Aps,St2};
 expr({match,L,P0,E0}, St0) ->
-    %% First fold matches together to create aliases.
-    {P1,E1} = fold_match(E0, P0),
-    St1 = set_wanted(P1, St0),
-    {E2,Eps1,St2} = novars(E1, St1),
-    St3 = St2#core{wanted=St0#core.wanted},
-    {P2,St4} = try
-                   pattern(P1, St3)
-               catch
-                   throw:Thrown ->
-                       {Thrown,St3}
-               end,
-    {Fpat,St5} = new_var(St4),
-    Lanno = lineno_anno(L, St5),
-    Fc = fail_clause([Fpat], Lanno, c_tuple([#c_literal{val=badmatch},Fpat])),
-    case P2 of
-	nomatch ->
-	    %% The pattern will not match. We must take care here to
-	    %% bind all variables that the pattern would have bound
-	    %% so that subsequent expressions do not refer to unbound
-	    %% variables.
-	    %%
-	    %% As an example, this code:
-	    %%
-	    %%   [X] = {Y} = E,
-	    %%   X + Y.
-	    %%
-	    %% will be rewritten to:
-	    %%
-	    %%   error({badmatch,E}),
-	    %%   case E of
-	    %%      {[X],{Y}} ->
-	    %%        X + Y;
-	    %%      Other ->
-	    %%        error({badmatch,Other})
-	    %%   end.
-	    %%
-	    St6 = add_warning(L, {nomatch,pattern}, St5),
-	    {Expr,Eps3,St7} = safe(E1, St6),
-	    SanPat0 = sanitize(P1),
-	    {SanPat,St} = pattern(SanPat0, St7),
-	    Badmatch = c_tuple([#c_literal{val=badmatch},Expr]),
-	    Fail = #iprimop{anno=#a{anno=Lanno},
-			    name=#c_literal{val=match_fail},
-			    args=[Badmatch]},
-	    Eps = Eps3 ++ [Fail],
-	    {#imatch{anno=#a{anno=Lanno},pat=SanPat,arg=Expr,fc=Fc},Eps,St};
-	Other when not is_atom(Other) ->
-            %% We must rewrite top-level aliases to lets to avoid unbound
-            %% variables in code such as:
+    St1 = set_wanted(P0, St0),
+    case fold_match(E0, P0) of
+        {{sequential_match,_,_,_}=P1,E1} ->
+            %% Matching of an expression to more than one pattern. Example:
             %%
-            %%     <<42:Sz>> = Sz = B
+            %%    #{Key := Value} = #{key := Key} = Expr
+            {E2,Eps1,St2} = safe(E1, St1),
+            St3 = St2#core{wanted=St0#core.wanted},
+
+            %% If necessary, bind the expression to a variable to ensure it is
+            %% only evaluted once.
+            {Var,Eps2,St4} =
+                case E2 of
+                    #c_var{} ->
+                        {E2,[],St3};
+                    _ ->
+                        {Var0,StInt} = new_var(St3),
+                        {Var0,[#iset{var=Var0,arg=E2}],StInt}
+                end,
+
+            %% Rewrite to a begin/end block matching one pattern at the time
+            %% (using the `single_match` operator). Example:
             %%
-            %% If we would keep the top-level aliases the example would
-            %% be translated like this:
-            %%
-            %% 	   case B of
-            %%         <Sz = #{#<42>(Sz,1,'integer',['unsigned'|['big']])}#>
-            %%            when 'true' ->
-            %%            .
-            %%            .
-            %%            .
-            %%
-            %% Here the variable Sz would be unbound in the binary pattern.
-            %%
-            %% Instead we bind Sz in a let to ensure it is bound when
-            %% used in the binary pattern:
-            %%
-            %%     let <Sz> = B
-            %% 	   in case Sz of
-            %%         <#{#<42>(Sz,1,'integer',['unsigned'|['big']])}#>
-            %%            when 'true' ->
-            %%            .
-            %%            .
-            %%            .
-            %%
-            {P3,E3,Eps2} = letify_aliases(P2, E2),
-            Eps = Eps1 ++ Eps2,
-            {#imatch{anno=#a{anno=Lanno},pat=P3,arg=E3,fc=Fc},Eps,St5}
+            %% begin
+            %%   V = Expr,
+            %%   #{key := Key} = V,
+            %%   #{Key := Value} = V
+            %% end
+            Block = blockify(L, P1, Var),
+            {E3,Eps3,St5} = expr({block,L,Block}, St4),
+            {E3,Eps1 ++ Eps2 ++ Eps3,St5};
+        {P0,E1} ->
+            %% Matching of an expression to a single pattern. Example:
+            %%    {A,B} = Expr
+            {E2,Eps1,St2} = novars(E1, St1),
+            St3 = St2#core{wanted=St0#core.wanted},
+            {E3,Eps2,St4} = single_match(L, P0, E2, St3),
+            {E3,Eps1 ++ Eps2,St4}
     end;
+expr({single_match,L,P,#c_var{}=E}, St0) ->
+    single_match(L, P, E, St0);
 expr({op,_,'++',{lc,Llc,E,Qs0},More}, St0) ->
     %% Optimise '++' here because of the list comprehension algorithm.
     %%
@@ -1015,6 +977,56 @@ expr({op,L,Op,L0,R0}, St0) ->
 	    module=#c_literal{anno=LineAnno,val=erlang},
 	    name=#c_literal{anno=LineAnno,val=Op},args=As},Aps,St1}.
 
+blockify(L0, {sequential_match,_L1,First,Then}, E) ->
+    [{single_match,L0,First,E}|blockify(L0, Then, E)];
+blockify(L, P, E) ->
+    [{single_match,L,P,E}].
+
+%% single_match(Line, AbstractPattern, CoreExpr, State0) -> {Expr,Pre,State}.
+%%  Generate the code for matching an expression against a single pattern.
+single_match(L, P0, E, St0) ->
+    {Fpat,St1} = new_var(St0),
+    Lanno = lineno_anno(L, St1),
+    Fc = fail_clause([Fpat], Lanno, c_tuple([#c_literal{val=badmatch},Fpat])),
+    try pattern(P0, St1) of
+        {P1,St2} ->
+            St3 = set_wanted(P0, St2),
+            St4 = St3#core{wanted=St0#core.wanted},
+            {#imatch{anno=#a{anno=Lanno},pat=P1,arg=E,fc=Fc},[],St4}
+    catch
+        throw:nomatch ->
+            %% The pattern will not match. We must take care here to
+            %% bind all variables that the pattern would have bound
+            %% so that subsequent expressions do not refer to unbound
+            %% variables.
+            %%
+            %% As an example, this code:
+            %%
+            %%   ([X] = {Y}) = E,
+            %%   X + Y.
+            %%
+            %% will be rewritten to:
+            %%
+            %%   error({badmatch,E}),
+            %%   case E of
+            %%      {[X],{Y}} ->
+            %%        X + Y;
+            %%      Other ->
+            %%        error({badmatch,Other})
+            %%   end.
+            %%
+            St2 = add_warning(L, {nomatch,pattern}, St1),
+            {Expr,Eps0,St3} = force_safe(E, St2),
+            SanPat0 = sanitize(P0),
+            {SanPat,St} = pattern(SanPat0, St3),
+            Badmatch = c_tuple([#c_literal{val=badmatch},Expr]),
+            Fail = #iprimop{anno=#a{anno=Lanno},
+                            name=#c_literal{val=match_fail},
+                            args=[Badmatch]},
+            Eps = Eps0 ++ [Fail],
+            {#imatch{anno=#a{anno=Lanno},pat=SanPat,arg=Expr,fc=Fc},Eps,St}
+    end.
+
 %% set_wanted(Pattern, St) -> St'.
 %%  Suppress warnings for expressions that are bound to the '_'
 %%  variable and variables that begin with '_'.
@@ -1028,12 +1040,6 @@ set_wanted({var,_,Var}, St) ->
             St
     end;
 set_wanted(_, St) -> St.
-
-letify_aliases(#c_alias{var=V,pat=P0}, E0) ->
-    {P1,E1,Eps0} = letify_aliases(P0, V),
-    {P1,E1,[#iset{var=V,arg=E0}|Eps0]};
-letify_aliases(P, E) ->
-    {P,E,[]}.
 
 %% sanitize(Pat) -> SanitizedPattern
 %%  Rewrite Pat so that it will be accepted by pattern/2 and will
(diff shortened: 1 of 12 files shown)


RobinMorisset:
Thank you for this change! It makes the language a bit simpler, and I like that it clarifies the semantics of assignment.

RobinMorisset:
This PR seems to have introduced https://github.com/erlang/otp/issues/6445

bjorng:
@RobinMorisset Thanks for catching this problem. Fixed.

kikofernandez (approved):
Great work! I only have a minor suggestion to simplify a function implementation.

lib/compiler/src/v3_core.erl
```diff
@@ -2025,10 +2025,18 @@ is_safe(_) -> false.
 %% fold_match(MatchExpr, Pat) -> {MatchPat,Expr}.
 %%  Fold nested matches into one match with aliased patterns.
 
-fold_match({match,L,P0,E0}, P) ->
-    {P1,E1} = fold_match(E0, P),
-    {{match,L,P0,P1},E1};
-fold_match(E, P) -> {P,E}.
+fold_match(P, E) ->
```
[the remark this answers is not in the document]
bjorng: Thanks! That **is** simpler and works if you add a clause to handle the case that the first argument is not a `match` tuple:

```
fold_match({match, L0, P0, E1}, E0) ->
    fold_match_helper(E1, {match, L0, P0, E0});
fold_match(P, E) ->
    {E, P}.
```

Staring some more at the code, I saw that it could be further simplified by removing `fold_match/2` and renaming the helper function to `fold_match/2`. I have pushed a fixup commit that does that.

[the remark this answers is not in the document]
kikofernandez: Perfect!

RobinMorisset:
This PR introduces an erlc crash on the following code:
```erlang
-module(validator).
-compile([export_all]).

f(#{[] := <<_>>, [] := <<_>>}) -> 
    ok.
``` 
It results in the following error message:
```
Function: f/1
minimized/validator.erl: internal error in pass beam_validator_strong:
exception error: no case clause matching {test_tail,8}
  in function  beam_validator:validate_bs_match/4 (beam_validator.erl, line 1743)
  in call from beam_validator:branch/4 (beam_validator.erl, line 2889)
  in call from beam_validator:validate_instrs/4 (beam_validator.erl, line 333)
  in call from beam_validator:validate_1/5 (beam_validator.erl, line 276)
  in call from beam_validator:validate_0/4 (beam_validator.erl, line 114)
  in call from beam_validator:validate/2 (beam_validator.erl, line 56)
  in call from compile:beam_validator_1/3 (compile.erl, line 1663)
  in call from compile:fold_comp/4 (compile.erl, line 408)
```
I could not reproduce this bug on master, only with this PR.

@bjorng I'm not sure whether such PR-specific bug reports should go here or in their own issue, please tell me if you have any preference (or more generally if I should report the bugs I find in a different way).

bjorng:
As long as a PR is not merged, I prefer bug reports to go into that PR as comments.

bjorng:
I have pushed a bug fix. Thanks for noticing the problem.

bjorng:
Thanks! Fixed.

RobinMorisset:
Here is a testcase that breaks with the latest version of this PR:
```erlang
f(X, <<Y>>) when ((Y > X) xor X); (not X); X ->
    ok.
```
Error message:
```
Sub pass sanitize
Function: f/2
minimized/precodegen.erl: internal error in pass beam_ssa_pre_codegen:
exception error: no match of right hand side value [{b_literal,integer},
 {b_var,13},
 {b_literal,[unsigned,big]},
 {b_literal,8},
 {b_literal,1}]
  in function  beam_ssa_pre_codegen:sanitize_instr/3 (beam_ssa_pre_codegen.erl, line 900)
  in call from beam_ssa_pre_codegen:do_sanitize_is/9 (beam_ssa_pre_codegen.erl, line 805)
  in call from beam_ssa_pre_codegen:sanitize/5 (beam_ssa_pre_codegen.erl, line 667)
  in call from beam_ssa_pre_codegen:sanitize/1 (beam_ssa_pre_codegen.erl, line 657)
  in call from compile:run_sub_passes_1/3 (compile.erl, line 422)
  in call from beam_ssa_pre_codegen:function/2 (beam_ssa_pre_codegen.erl, line 160)
  in call from beam_ssa_pre_codegen:functions/2 (beam_ssa_pre_codegen.erl, line 89)
  in call from beam_ssa_pre_codegen:module/2 (beam_ssa_pre_codegen.erl, line 84)
```

RobinMorisset:
And here is another with sensibly the same error message, I'm putting it here in case it helps with debugging:
```erlang
f(<<X>>, Y) when {(X / 1), (Y andalso true)}; Y ->
    ok.
```

RobinMorisset:
A third one with the same error message:
```erlang
f(<<X>>, Y) when ((<<X>> == false) orelse (Y andalso true)); Y ->
    Y.
```

bjorng:
Thanks! Fixed. I included all three examples in the test suite.

RobinMorisset:
Thank you for your quick fixes.
Here is a fresh testcase that crashes erlc with this PR but not without:
```erlang
f(X) ->
    case 0 of
        <<_:(X bor X)>> ->
            <<_:Y>> = (Y = X)
    end,
    Y.
```
The error message is:
```
Function: f/1
minimized/kernel_to_ssa6.erl: internal error in pass beam_kernel_to_ssa:
exception error: bad key: 'Y'
  in function  map_get/2
     called as map_get('Y',
 #{0 => {b_var,0},
   '@exception' => {b_var,'@exception'},
   '@ssa_bool' => {b_var,'@ssa_bool'},
   '@ssa_ret' => {b_var,'@ssa_ret'}})
     *** argument 1: not present in map
  in call from beam_kernel_to_ssa:ssa_arg/2 (beam_kernel_to_ssa.erl, line 1127)
  in call from beam_kernel_to_ssa:cg/2 (beam_kernel_to_ssa.erl, line 131)
  in call from beam_kernel_to_ssa:cg/2 (beam_kernel_to_ssa.erl, line 114)
  in call from beam_kernel_to_ssa:do_match_cg/3 (beam_kernel_to_ssa.erl, line 167)
  in call from beam_kernel_to_ssa:cg_fun/2 (beam_kernel_to_ssa.erl, line 91)
  in call from beam_kernel_to_ssa:function/3 (beam_kernel_to_ssa.erl, line 75)
  in call from beam_kernel_to_ssa:'-functions/3-lc$^0/1-0-'/3 (beam_kernel_to_ssa.erl, line 65)
```

RobinMorisset:
Another one that fails with this PR (both with and without 28fce1f8016a8bf52ca6a109aa471700f17e772b), and does not on master (both with and without 28fce1f8016a8bf52ca6a109aa471700f17e772b):
```erlang
f(#{0 := <<_:0>>, 0 := <<_>>}) -> 
    ok.
```
results in
```
Function: f/1
minimized/bug1.erl: internal error in pass beam_validator_strong:
exception error: no case clause matching {test_tail,0}
  in function  beam_validator:validate_bs_match/4 (beam_validator.erl, line 1716)
  in call from beam_validator:branch/4 (beam_validator.erl, line 2879)
  in call from beam_validator:validate_instrs/4 (beam_validator.erl, line 333)
  in call from beam_validator:validate_1/5 (beam_validator.erl, line 276)
  in call from beam_validator:validate_0/4 (beam_validator.erl, line 114)
  in call from beam_validator:validate/2 (beam_validator.erl, line 56)
  in call from compile:beam_validator_1/3 (compile.erl, line 1663)
  in call from compile:fold_comp/4 (compile.erl, line 408)
```

bjorng:
Thanks! Fixed.

RobinMorisset:
It seems to still be possible to trigger the same bug (or at least one with the same error message) with different testcases, such as:
```erlang
f(#{0 := <<_>>, 0 := <<X>>})  -> 
	X.
```
or
```erlang
f(#{0 := <<_>>, 0 := <<_, 0>>}) ->
    ok.
```
or
```erlang
f(#{0 := <<_>>, 0 := <<_, _:(0 div 0)>>}) ->
    0.
```
or
```erlang
f(#{0 := <<X:0>>, 0 := <<X>>}) ->
    ok.
```
or
```erlang
f() ->
    case (catch <<0 || true>>) of
        #{[] := <<_>>, [] := <<X>>} ->
            X
    end.
```

bjorng:
Yes, it is the same bug. I have now pushed a fix that cares of it in a different way than my previous attempts.

RobinMorisset:
Thank you for your reactivity. The fuzzer can no longer trigger that particular bug, but it now found some very similar testcases that trigger a different error message:
```erlang
f(#{0 := <<_, _:(true andalso 0)>>, 0 := <<_>>}) ->
    ok.
```
and
```erlang
f() ->
    case catch ({}+{}) of
        #{[] := <<X:0, _:X>>, [] := <<_>>} ->
            ok
    end.
```
Both cause a crash with the following error message (that is for the second case, the first one is the same except for some numbers):
```
Sub pass ssa_opt_bsm_shortcut
Function: f/0
minimized/bug2.erl: internal error in pass beam_ssa_opt:
exception error: bad key: {b_var,14}
  in function  map_get/2
     called as map_get({b_var,14},#{39 => {0,8},{b_var,13} => 0})
     *** argument 1: not present in map
  in call from beam_ssa_opt:bsm_positions/2 (beam_ssa_opt.erl, line 1959)
  in call from beam_ssa_opt:ssa_opt_bsm_shortcut/1 (beam_ssa_opt.erl, line 1944)
  in call from compile:run_sub_passes_1/3 (compile.erl, line 422)
  in call from beam_ssa_opt:phase/4 (beam_ssa_opt.erl, line 114)
  in call from beam_ssa_opt:run_phases/3 (beam_ssa_opt.erl, line 77)
  in call from beam_ssa_opt:module/2 (beam_ssa_opt.erl, line 69)
  in call from compile:'-select_passes/2-anonymous-0-'/3 (compile.erl, line 681)
```

bjorng:
Your latest example triggered a bug in the `ssa_opt_bsm` sub pass. I fixed it... I thought. The new test cases based on your examples fail when some optimizations are disabled. It seems to be an unrelated bug. I can't push my fix to this PR until I have fixed that bug, too, because there would be failed test cases in our daily builds. Meanwhile, if your want to find more bugs (with all compiler optimizations enabled), here is another branch which includes my latest fix:

https://github.com/bjorng/otp/tree/bjorn/compiler/parallel-match-WIP

bjorng:
I have now fixed the other bugs and updated this pull request.

jhogberg (approved):

bjorng:
@rvirding Do you have any opinions about this pull request and the `=` operator?

In particular, what do you think about that in bodies, multiple `=` operators will be evaluated from right to left? As an example, if the matching is defined to be right to left, that will make the following example legal:

```
baz(M) ->
    #{K := V} = #{k := K} = M,
    V.
```

I am asking because we have had some discussions in the OTP team as a preparation for an OTP Technical Board meeting we will have later this week.

bjorng:
The OTP Technical Board has now approved this pull request.

 I have updated the documentation to clarify the semantics of the **match operator** and the **compound pattern operator** (a new name invented by @RaimoNiskanen in order to more easily distinguish the two flavors of the `=` operator).

After the updates of the documentation have been reviewed, I hope to merge this pull request next week.

system/doc/reference_manual/expressions.xml
```diff
@@ -192,22 +201,112 @@ case {Value, Result} of
...
+    sub patterns of the same compound pattern.</p>
+    <p><em>Examples:</em></p>
+
+    <pre>
+1> <input>fun(#{Key := Value} = #{key := Key}) -> Value end.</input>
+* 1:7: variable 'Key' is unbound
+2> <input>F = fun({A, B} = E) -> {E, A + B} end, F({1,2}).</input>
+{{1,2},3}
+3> <input>G = fun(&lt;&lt;A:8,B:8>> = &lt;&lt;C:16>>) -> {A, B, C} end, G(&lt;&lt;42,43>>).</input>
+{42,42,10795}</pre>
```
RobinMorisset, line 270: Shouldn't the second element be 43 here?

system/doc/reference_manual/expressions.xml
```diff
@@ -192,22 +201,112 @@ case {Value, Result} of
...
+    <title>The Match Operator</title>
+    <p>The following matches <c>Pattern</c> against
+      <c>Expr</c>:</p>
     <pre>
-Expr1 = Expr2</pre>
+Pattern = Expr</pre>
     <p>If the matching succeeds, any unbound variable in the pattern
-      becomes bound and the value of <c>Expr2</c> is returned.</p>
+    becomes bound and the value of <c>Expr</c> is returned.</p>
+    <p>If multiple match operators are applied in sequence, they will be
```
RobinMorisset, line 212: I'd maybe add that even if there is a single match operator, its right hand side will be executed before the pattern on its left hand side is checked. So for example the following is valid:
```erlang
<<X:Y>> = begin Y = 8, <<42:8>> end
```
And it correctly binds X to 42, as Y is bound to 8 before the validity of the pattern on the left hand side is checked.

RobinMorisset (commented):
Thank you again for your work on this!

system/doc/reference_manual/expressions.xml
bjorng: Yes, it should. I changed example to make it clearer and changed it manually instead of copy-pasting the whole thing from the output of the shell.

system/doc/reference_manual/expressions.xml
bjorng: Yes, I will add that clarification and an example.

RaimoNiskanen (approved):
The documentation changes looks very good to me

rickard-green (approved):
The doc changes looks good to me too :+1:

PR #6420, opened 2022-11-03
ct: prevent documentation errors and ease maintenance

paulo-ferraz-oliveira:
## Description

We:

1. expose a new type (`ct:handle/0`) as per the current doc.
   - alternatively we can just _spec_ this as `pid()` and remove the newly added type
2. export (existing, though private) type `ct:conn_log_option/0` as per the current doc
   - same alternative approach (see 1.) applies
3. typespec all the functions in a very lazy approach (basically copy-paste from the doc. and then fix any Dialyzer issues/missing bits)
4. refer (in the .xml) to the newly added specs for ease of maintenance

## Some generic notes

- `make dialyzer` shows no issues
- `make xmllint` shows no issues
- local result for `make release_docs` seems to yield good result
- I've targeted the `erlang:master` branch, but maybe this can be released sooner (?)
- I followed https://github.com/erlang/otp/wiki/Documentation and got a note stating that `./otp_build autoconf` is no longer required, so maybe [it can be removed](https://github.com/erlang/otp/issues/6421) from that Wiki page

## Some function-specific notes

- `get_testspec_terms/0` mentioned `Tag` but had no type for it (is now `atom()`, as per `get_testspec_terms/1`)
- `listenv/1` returns an `{ok, Ok}` or `{error, Error}` though it was only listed as returning `Ok`; I'm not sure the `Error` is also appropriate, in this case
- `require/2` mentioned `Reason` but no type for it: it's now `term()`
  - the same is applicable for other functions (I didn't keep track of all of them)
- `run/3` mentioned `TestResult` but no type for it: it's now `term()`
- `run_test/1` mentioned type `string(HHMMSS)`: I've now created spec.type `HHMMSS` as `string()` and `string(HHMMSS)` is now simply `HHMMSS` (I tried to give it some semantics) 
  - the same goes for `string(YYMoMoDDHHMMSS)`
- `fail/1,2` was said to return `ok`: it's now spec.ed as `no_return()`
- `Category`, from `log/2,3,4,5`, was tweaked to what seems more appropriate as per the function's internal pattern matching
  - the same goes for `print/2,3,4,5`
  - the same goes for `pal/2,3,4,5`

### DIFF
--- a/lib/common_test/doc/src/ct.xml
+++ b/lib/common_test/doc/src/ct.xml
@@ -67,20 +67,20 @@
 
   <datatypes>
     <datatype>
-      <name>handle() = pid()</name>
+      <name name="handle" />
       <desc>
         <p>The identity (handle) of a connection.</p>
       </desc>
     </datatype>
 
     <datatype>
-      <name>config_key() = atom()</name>
+      <name name="config_key" />
       <desc>
 	<p>A configuration key which exists in a configuration file</p>
       </desc>
     </datatype>
     <datatype>
-      <name>target_name() = atom()</name>
+      <name name="target_name" />
       <desc>
         <p>A name and association to configuration data introduced
         through a require statement, or a call to
@@ -90,8 +90,11 @@
       </desc>
     </datatype>
     <datatype>
-      <name>key_or_name() = config_key() | target_name()</name>
-      <name>conn_log_options() = [conn_log_option()]</name>
+      <name name="key_or_name" />
+      <desc></desc>
+    </datatype>
+    <datatype>
+      <name name="conn_log_options" />
       <desc>
 	<p>Options that can be given to the <c>cth_conn_log</c> hook,
 	which is used for logging of NETCONF and Telnet
@@ -103,21 +106,24 @@
     </datatype>
 
     <datatype>
-      <name>conn_log_option() = {log_type,conn_log_type()} | {hosts,[key_or_name()]}</name>
-      <name>conn_log_type() = raw | pretty | html | silent</name>
-      <name>conn_log_mod() = ct_netconfc | ct_telnet</name>
+      <name name="conn_log_option" />
+      <desc></desc>
+    </datatype>
+    <datatype>
+      <name name="conn_log_type" />
+      <desc></desc>
+    </datatype>
+    <datatype>
+      <name name="conn_log_mod" />
+      <desc></desc>
     </datatype>
 
   </datatypes>
 
   <funcs>
     <func>
-      <name since="">abort_current_testcase(Reason) -&gt; ok | {error, ErrorReason}</name>
+      <name since="" name="abort_current_testcase" arity="1" />
       <fsummary>Aborts the currently executing test case.</fsummary>
-      <type>
-        <v>Reason = term()</v>
-        <v>ErrorReason = no_testcase_running | parallel_group</v>
-      </type>
       <desc><marker id="abort_current_testcase-1"/>
         <p>Aborts the currently executing test case. The user must know with
           certainty which test case is currently executing. The function is
@@ -130,14 +136,9 @@
     </func>
 
     <func>
-      <name since="OTP R14B">add_config(Callback, Config) -&gt; ok | {error, Reason}</name>
+      <name since="OTP R14B" name="add_config" arity="2" />
       <fsummary>Loads configuration variables using the specified callback
         module and configuration string.</fsummary>
-      <type>
-        <v>Callback = atom()</v>
-        <v>Config = string()</v>
-        <v>Reason = term()</v>
-      </type>
       <desc><marker id="add_config-2"/>
         <p>Loads configuration variables using the specified callback module and
           configuration string. The callback module is to be either loaded or
@@ -149,14 +150,9 @@
     </func>
 
     <func>
-      <name since="OTP R15B02">break(Comment) -&gt; ok | {error, Reason}</name>
+      <name since="OTP R15B02" name="break" arity="1" />
       <fsummary>Cancels any active timetrap and pause the execution of the
         current test case until the user calls function continue/0.</fsummary>
-      <type>
-        <v>Comment = string()</v>
-        <v>Reason = {multiple_cases_running, TestCases} | 'enable break with release_shell option'</v>
-        <v>TestCases = [atom()]</v>
-      </type>
       <desc><marker id="break-1"/>
         <p>Cancels any active timetrap and pauses the execution of the
           current test case until the user calls function <c>continue/0</c>.
@@ -179,14 +175,9 @@
     </func>
 
     <func>
-      <name since="OTP R15B02">break(TestCase, Comment) -&gt; ok | {error, Reason}</name>
+      <name since="OTP R15B02" name="break" arity="2" />
       <fsummary>Works the same way as break/1, only argument TestCase makes it
         possible to pause a test case executing in a parallel group.</fsummary>
-      <type>
-        <v>TestCase = atom()</v>
-        <v>Comment = string()</v>
-        <v>Reason = 'test case not running' | 'enable break with release_shell option'</v>
-      </type>
       <desc><marker id="break-2"/>
         <p>Works the same way as
           <seemfa marker="#break/1"><c>ct:break/1</c></seemfa>, only
@@ -201,11 +192,8 @@
     </func>
 
     <func>
-      <name since="OTP R15B">capture_get() -&gt; ListOfStrings</name>
+      <name since="OTP R15B" name="capture_get" arity="0" />
       <fsummary>Equivalent to capture_get([default]).</fsummary>
-      <type>
-        <v>ListOfStrings = [string()]</v>
-      </type>
       <desc><marker id="capture_get-0"/>
         <p>Equivalent to
           <seemfa marker="#capture_get/1">ct:capture_get([default])</seemfa>.</p>
@@ -213,13 +201,9 @@
     </func>
 
     <func>
-      <name since="OTP R15B">capture_get(ExclCategories) -&gt; ListOfStrings</name>
+      <name since="OTP R15B" name="capture_get" arity="1" />
       <fsummary>Returns and purges the list of text strings buffered during
         the latest session of capturing printouts to stdout.</fsummary>
-      <type>
-        <v>ExclCategories = [atom()]</v>
-        <v>ListOfStrings = [string()]</v>
-      </type>
       <desc><marker id="capture_get-1"/>
         <p>Returns and purges the list of text strings buffered during the
           latest session of capturing printouts to <c>stdout</c>. Log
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
@rickard-green, is there an expectation on when this will be reviewed? Thanks.

rickard-green:
> @rickard-green, is there an expectation on when this will be reviewed? Thanks.

Sorry, this issue is not handled by my team. @IngelaAndin might be able to give you an answer.

IngelaAndin:
Someone from the PS team will look at it when we can find some time. Although not unimportant common_test is not a top-priority application.

paulo-ferraz-oliveira:
Thank you @IngelaAndin, I'll wait for your analysis.

lib/common_test/src/ct.erl
```diff
@@ -89,34 +92,135 @@
...
+      when TestDir :: string(),
+           Suite :: atom(),
+           Result :: [TestResult] | {error, Reason},
+           TestResult :: term(),
+           Reason :: term().
 run(TestDir,Suite) ->
     ct_run:run(TestDir,Suite).
 
+-spec run(TestDirs) -> Result
+      when TestDirs :: TestDir | [TestDir],
```
kikofernandez, line 128: I think the type variable `TestDir` should be `TestDir :: string()`

lib/common_test/src/ct.erl
```diff
@@ -89,34 +92,135 @@
...
+           Label :: string() | atom(),
+           CfgFiles :: [string()] | string(),
+           UserConfig :: [{CallbackMod, CfgStrings}] | {CallbackMod, CfgStrings},
+           CallbackMod :: atom(),
+           CfgStrings :: [string()] | string(),
+           LogDir :: string(),
+           Conns :: all | [atom()],
+           CSSFile :: string(),
+           CoverSpecFile :: string(),
+           StepOpts :: [StepOpt] | [],
```
kikofernandez, line 153: This definition `StepOpts` is redundant and should be `StepOpts :: [StepOpt]`, since a list type already includes an empty list.

lib/common_test/src/ct.erl
```diff
@@ -89,34 +92,135 @@
...
+      when TestDirs :: TestDir | [TestDir],
+           Result :: [TestResult] | {error, Reason},
+           TestResult :: term(),
+           Reason :: term().
 run(TestDirs) ->
     ct_run:run(TestDirs).
 
+-spec run_test(Opts) -> Result
+      when Opts :: [OptTuples],
+           OptTuples :: {dir, TestDirs} | {suite, Suites} | {group, Groups} | {testcase, Cases} | {spec, TestSpecs} | {join_specs, Bool} | {label, Label} | {config, CfgFiles} | {userconfig, UserConfig} | {allow_user_terms, Bool} | {logdir, LogDir} | {silent_connections, Conns} | {stylesheet, CSSFile} | {cover, CoverSpecFile} | {cover_stop, Bool} | {step, StepOpts} | {event_handler, EventHandlers} | {include, InclDirs} | {auto_compile, Bool} | {abort_if_missing_suites, Bool} | {create_priv_dir, CreatePrivDir} | {multiply_timetraps, M} | {scale_timetraps, Bool} | {repeat, N} | {duration, DurTime} | {until, StopTime} | {force_stop, ForceStop} | {decrypt, DecryptKeyOrFile} | {refresh_logs, LogDir} | {logopts, LogOpts} | {verbosity, VLevels} | {basic_html, Bool} | {esc_chars, Bool} | {keep_logs,KeepSpec} | {ct_hooks, CTHs} | {enable_builtin_hooks, Bool} | {release_shell, Bool},
```
kikofernandez, line 137: The definition of the type `OptTuples` is way too long, maybe it can be break in multiple lines to make it easier to read

kikofernandez: The type variable `Bool` is not defined and probably should be `boolean()`? Please check :)

lib/common_test/src/ct.erl
```diff
@@ -89,34 +92,135 @@
...
+           Case :: atom(),
+           Result :: term().
 step(TestDir,Suite,Case) ->
     ct_run:step(TestDir,Suite,Case).
 
+-spec step(TestDir, Suite, Case, Opts) -> Result
+      when TestDir :: string(),
+           Suite :: atom(),
+           Case :: atom(),
+           Opts :: [Opt] | [],
```
kikofernandez, line 212: the empty list is already contained in a `[Opt]`, so `Opts :: [Opt],`

lib/common_test/src/ct.erl
```diff
@@ -79,6 +81,7 @@
 -type config_key() :: atom(). % Config key which exists in a config file
 -type target_name() :: atom().% Name associated to a config_key() though 'require'
 -type key_or_name() :: config_key() | target_name().
+-type handle() :: pid().
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: It is, as per [my changes](https://github.com/erlang/otp/pull/6420/files/42be66e1968d0880c1119e481a7a826082ba9afe#diff-d48f5f891bb937c2d1e2a83abd3f8be520652aea7f8cda5470ed9ef01b86659bR70). Is this just a confirmation, regarding what I wrote in the initial description?

Feel free to Resolve this conversation if there's no follow-up. 👍

paulo-ferraz-oliveira:
@kikofernandez, I'll be doing individual commits, with replies to your comments pointing to them, for a smoother review, but can/will squash everything at the end. You can (to help me) Resolve conversation as you revalidate stuff. Thank you.

lib/common_test/src/ct.erl
paulo-ferraz-oliveira: 2d6d21f01d

lib/common_test/src/ct.erl
paulo-ferraz-oliveira: 24a446a766, for the second comment.

paulo-ferraz-oliveira: c32d21fb46, for the first comment.

lib/common_test/src/ct.erl
paulo-ferraz-oliveira: I've found this elsewhere, in the past, and do agree it's redundant. I'm wondering if the producer just wants to signal (sometimes) that this may be empty so it's easier to eyeball?

paulo-ferraz-oliveira: a4c7e5a7b8

lib/common_test/src/ct.erl
paulo-ferraz-oliveira: d02030ecd2

lib/common_test/src/ct.erl
```diff
@@ -125,24 +229,65 @@ stop_interactive() ->
...
+           Value :: term().
 get_config(Required,Default) ->
     ct_config:get_config(Required,Default,[]).
 
+-spec get_config(Required, Default, Opts) -> ValueOrElement
+      when Required :: KeyOrName | {KeyOrName, SubKey} | {KeyOrName, SubKey, SubKey},
+           KeyOrName :: atom(),
+           SubKey :: atom(),
+           Default :: term(),
+           Opts :: [Opt] | [],
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: 24488490e2

lib/common_test/src/ct.erl
```diff
@@ -171,9 +322,17 @@ escape_chars(Format, Args) ->
 	    {error,Reason}
     end.
 
+-spec log(Format) -> ok
+      when Format :: string().
 log(Format) ->
     log(default,?STD_IMPORTANCE,Format,[],[]).
 
+-spec log(X1, X2) -> ok
+      when X1 :: Category | Importance | Format,
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: I'll check it but from my initial description we read "typespec all the functions in a very lazy approach", so I basically just copy-pasted everything, no real verification work after that (except for the static analysis which showed some issues that were solved).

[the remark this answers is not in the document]
paulo-ferraz-oliveira: 2d2a1454bc

lib/common_test/src/ct.erl
```diff
@@ -489,37 +814,87 @@ get_testdata(Key) ->
...
+           M :: atom(),
+           F :: atom(),
+           A :: list().
 timetrap(Time) ->
     test_server:timetrap_cancel(),
     test_server:timetrap(Time).
 
+-spec get_timetrap_info() -> {Time, {Scaling,ScaleVal}}
+      when Time :: integer() | infinity,
+           Scaling :: true | false,
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: 3791db70d4

lib/common_test/doc/src/ct.xml
```diff
@@ -103,21 +103,17 @@
     </datatype>
 
     <datatype>
-      <name>conn_log_option() = {log_type,conn_log_type()} | {hosts,[key_or_name()]}</name>
-      <name>conn_log_type() = raw | pretty | html | silent</name>
-      <name>conn_log_mod() = ct_netconfc | ct_telnet</name>
+      <name name="conn_log_option" />
```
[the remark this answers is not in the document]
paulo-ferraz-oliveira: fb6822d639

(though now some are without description) - if you tell me their description I can add it.

lib/common_test/src/ct.erl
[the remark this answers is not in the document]
kikofernandez: my mistake, I did not see it.
Thanks!

lib/common_test/doc/src/ct.xml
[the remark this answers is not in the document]
kikofernandez: Thanks a lot. I think it is ok that some do not have description. 
See for example [here](https://www.erlang.org/doc/man/ct.html#type-conn_log_option): `conn_log_option` does not have a description and that's ok.

lib/common_test/src/ct.erl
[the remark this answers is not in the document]
kikofernandez: Yes, and I think that this is something that Dialyzer should have caught (see #6508 ) :)

lib/common_test/src/ct.erl
kikofernandez: Maybe, no clue...
In any case, to me, it is confusing that we put it explicitly as if indicating that the other list is non-empty... which cannot be the case because it would be `[StepOpt,...]`. Thanks for the fix!

kikofernandez:
This looks good so far, we will run some more internal tests and hopefully we can merge it tomorrow :) 

If I find some other issue, I will let you know. 

The only thing that we should do is to squash your first two commits into a single one, since we know that there were some types missing (and probably Dialyzer complained?). Today's changes can be placed into the single commit you just created by squashing the commits or into a new commit, up to you :) 

Thanks a lot for your contribution

paulo-ferraz-oliveira:
> and probably Dialyzer complained?

It did 😓 

I'll squash everything now that your review is done. Thank you.

kikofernandez (approved):
Looks good, I proceed with the merge.
Thanks a lot for your contribution!

PR #6425, opened 2022-11-03
ssl: Add debug logging on error cases

dgud:
To ease the debugging add info/debug logs whenever and error caught so that is easy to find what happened.

### DIFF
--- a/lib/ssl/src/ssl_internal.hrl
+++ b/lib/ssl/src/ssl_internal.hrl
@@ -276,8 +276,11 @@
         fun() ->
                 case get(log_level) of
                     undefined ->
-                        %% exit({log_level_not_set, self(), process_info(self(), current_stacktrace)}),
-                        ignore;
+                        %% Use debug here, i.e. log everything and let loggers
+                        %% log_level decide if it should be logged
+                        ssl_logger:log(Level, debug,
+                                       #{description => Descr, reason => Reason},
+                                       ?LOCATION);
                     __LogLevel__ ->
                         ssl_logger:log(Level, __LogLevel__,
                                        #{description => Descr, reason => Reason},
--- a/lib/ssl/src/ssl_internal.hrl
+++ b/lib/ssl/src/ssl_internal.hrl
@@ -23,7 +23,8 @@
 -ifndef(ssl_internal).
 -define(ssl_internal, true).
 
--include_lib("public_key/include/public_key.hrl"). 
+-include_lib("kernel/include/logger.hrl").
+-include_lib("public_key/include/public_key.hrl").
 
 -define(SECRET_PRINTOUT, "***").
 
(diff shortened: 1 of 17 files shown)


lib/ssl/src/ssl_internal.hrl
```diff
@@ -270,6 +271,21 @@
 				{stop, any(), any()}.
 -type ssl_options()          :: map().
 
+
+-define(SSL_LOG(Level, Descr, Reason),
+        fun() ->
+                case get(log_level) of
+                    undefined ->
```
IngelaAndin, line 278: Will not log_level always be defined ? We default to notice! And if you want no logs it should be none!

lib/ssl/test/ssl_cipher_SUITE.erl
```diff
@@ -74,9 +74,14 @@ end_per_group(_GroupName, Config) ->
 
 init_per_testcase(_TestCase, Config) ->
     ct:timetrap({seconds, 5}),
-    Config.
+    put(log_level, debug),  %% SSL log level
```
IngelaAndin, line 77: Maybe I am missing something but is this verifying that the debug-level macro works by manual inspection of the logs?

lib/ssl/src/ssl_internal.hrl
dgud: Because a function with the macro can be called from a process that have not set the log_level.
Such as in the testcases below :-)

It is also done to make sure we don't crash if we later wrongly call the macro from such a process.

lib/ssl/src/tls_connection_1_3.erl
```diff
@@ -604,6 +604,7 @@ do_client_start(ServerHello, State0) ->
 
 initial_state(Role, Sender, Host, Port, Socket, {SSLOptions, SocketOptions, Trackers}, User,
 	      {CbModule, DataTag, CloseTag, ErrorTag, PassiveTag}) ->
+    put(log_level, maps:get(log_level, SSLOptions)),
```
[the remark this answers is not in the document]
dgud: Because then we don't have to carry ssl_options as as argument to a lot of functions just to be able log
something at deep down in the call stack.

I asked about this in our meeting.

lib/ssl/test/ssl_cipher_SUITE.erl
u3s: shouldn't we check log content explicitly in one test case?

u3s (approved):

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -591,21 +591,21 @@ certificate_entry(DER) ->
...
-    try ssl_handshake:digitally_signed({3,4}, Content, HashAlgo, PrivateKey, SignAlgo) of
-        Signature ->
-            {ok, Signature}
-    catch
-        error:badarg ->
-            {error, ?ALERT_REC(?FATAL, ?INTERNAL_ERROR, badarg)}
+    try
+        {ok, ssl_handshake:digitally_signed({3,4}, Content, HashAlgo, PrivateKey, SignAlgo)}
+    catch throw:Alert ->
+            {error, Alert}
```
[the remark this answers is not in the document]
dgud: Check ssl_handshake:digitally_signed it doesn't exit(badarg) it throws an Alert

IngelaAndin (approved):

dgud:
Merged

PR #6432, opened 2022-11-04
Improve pg_SUITE

sverker:
Fix sometimes failing test cases.

@max-au

### DIFF
--- a/lib/kernel/test/pg_SUITE.erl
+++ b/lib/kernel/test/pg_SUITE.erl
@@ -266,14 +266,14 @@ empty_group_by_remote_leave(Config) when is_list(Config) ->
     RemotePid = erlang:spawn(Node, forever()),
     % remote join
     ?assertEqual(ok, rpc:call(Node, pg, join, [?FUNCTION_NAME, ?FUNCTION_NAME, RemotePid])),
-    sync({?FUNCTION_NAME, Node}),
+    sync_via({?FUNCTION_NAME, Node}, ?FUNCTION_NAME),
     ?assertEqual([RemotePid], pg:get_members(?FUNCTION_NAME, ?FUNCTION_NAME)),
     % inspecting internal state is not best practice, but there's no other way to check if the state is correct.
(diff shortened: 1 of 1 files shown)


lib/kernel/test/pg_SUITE.erl
```diff
@@ -715,7 +726,18 @@ sync(GS) ->
...
 sync_via({RegName, Node}, GS) ->
-    rpc:call(Node, sys, replace_state, [RegName, fun (S) -> (catch sys:get_state(GS)), S end]).
+    MyNode = node(),
+    rpc:call(Node, sys, replace_state,
+             [RegName, fun (S) -> (catch sys:get_state({GS,MyNode})), S end]);
+
+%% flush remote GS queue from local process RegName
+sync_via(RegName, {GS, Node}) ->
+    sys:replace_state(RegName,
+                      fun (S) -> _R = (catch sys:get_state({GS,Node})),
```
max-au, line 736: Same formatting nit, `{GS, Node}`

max-au (approved):
Thank you for fixing these issues!

lib/kernel/test/pg_SUITE.erl
```diff
@@ -67,13 +67,46 @@ suite() ->
...
+            Mon = erlang:monitor(process, Tracer),
+            Tracer ! flush,
+            normal = receive
+                         {'DOWN', Mon, process, Tracer, R} -> R
+                     end
+    end.
+
+tracer() ->
+    receive flush -> ok end,
+    io:format("Flush trace messages:\n"),
```
max-au, line 98: Do you mind removing these debug messages in the final version?

sverker: Yes, that was the plan. Just want to make sure pg_SUITE runs clean for a while before I merge.
Thanks for the reminder.

sverker: I might keep the trace code deactivated, if that's ok with you. I have already rewritten it twice.

max-au: That'd be fine, but then could you move it next to `sync`/`sync_via` functions to preserve the overall structure of the suite?

PR #6433, opened 2022-11-04
Prepare pg protocol for upgrade

sverker:
Add a protocol version to the discover message, but only to accepted and ignored for now.

`{discover, Peer, ProtocolVersion}`

Future pg version (OTP 28) may use it for protocol enhancements.

@max-au What do you think?


lib/kernel/src/pg.erl
```diff
@@ -395,17 +395,13 @@ handle_info({leave, Peer, PidOrPids, Groups}, #state{scope = Scope, remote = Rem
...
-            MRef = erlang:monitor(process, Peer),
-            erlang:send(Peer, {discover, self()}, [noconnect]),
-            {noreply, State#state{remote = Remote#{Peer => {MRef, #{}}}}}
-    end;
+handle_info({discover, Peer}, State) ->
+    handle_discover(Peer, State);
+
+%% New discover msg sent by a future pg version.
+%% Accepted first in OTP 26, to be used by OTP 28 or later.
+handle_info({discover, Peer, _ProtocolVersion}, State) ->
```
max-au, line 403: `handle_info` spec is missing this clause (probably spec should be rewritten completely, using overloaded function specs).
With the spec, it'd also be clearer what is the type of _ProtocolVersion (I don't know whether it should be an integer, a bitmask, or a map of "capabilities" for easier expansion).

max-au (commented):
Looks good.

Do you mind adding a test case that serves as a contract (literally to confirm "yes we want to pass 'version' and version is integer/map/...").

lib/kernel/src/pg.erl
sverker: I don't see much reason to specify the ProtocolVersion type now. I think it's better to leave that up to the author of a future protocol update.

max-au (approved):
Thanks!
I'd still think of adding at least a test case clause, mainly for a "contract" purpose.

sverker:
> Thanks! I'd still think of adding at least a test case clause, mainly for a "contract" purpose.

Yes, I'm thinking of how to do that. I guess it would entail faking a future pg process. Do you have any smart ideas?

max-au:
I would take a shortcut and simply start a process that sends both version-less and version-enabled discover message, without crashing the pg scope process. White-box, similar to `errors` case (that actually tests that `pg` does crash on unexpected messages).

sverker:
Test added and I cleaned up the branch.
@max-au Ok to merge?

PR #6437, opened 2022-11-06
Fix example code in ct_hooks_chapter.xml

fabjan:
Problem
======
When I copied the example hook from the Common Test User's Guide to use as a base for making a new hook
my editor complained about the code.

The example code has a syntax error (missing a period), and uses the deprecated erlang:now/0 function.

Fix
==
I followed the suggestions in ["How to Work with the New API"] to replace the `erlang:now/0` calls.

["How to Work with the New API"]: https://www.erlang.org/doc/apps/erts/time_correction.html#how-to-work-with-the-new-api

### DIFF
--- a/lib/common_test/doc/src/ct_hooks_chapter.xml
+++ b/lib/common_test/doc/src/ct_hooks_chapter.xml
@@ -99,6 +99,10 @@
     <item><c>{ct_hooks,[{my_cth_module,[{debug,true}],500}]}</c></item>
     </list>
 
+    <p>Note that regardless of how you install a CTH, its BEAM file
+      must be available in the code path when Common Test runs.
+      <c>ct_run</c> accepts the <c>-pa</c> command line option.</p>
+
     <section>
       <title>Overriding CTHs</title>
       <p>By default, each installation of a CTH causes a new instance of it
@@ -373,118 +377,93 @@
 
   <section>
     <marker id="example"/>
-     <title>Example CTH</title>
-     <p>The following CTH logs information about a test run into a format 
-       parseable by <seemfa marker="kernel:file#consult/1">file:consult/1</seemfa> 
-       (in Kernel):
-     </p>
-     <code>
- %%% Common Test Example Common Test Hook module.
- -module(example_cth).
+      <title>Example CTH</title>
+      <p>The following CTH logs information about a test run into a format 
+        parseable by <seemfa marker="kernel:file#consult/1">file:consult/1</seemfa> 
+        (in Kernel):
+      </p>
+      <code>
+%%% Common Test Example Common Test Hook module.
+%%%
+%%% To use this hook, on the command line:
+%%%     ct_run -suite example_SUITE -pa . -ct_hooks example_cth
+%%%
+%%% Note `-pa .`: the hook beam file must be in the code path when installing.
+-module(example_cth).
 
- %% Callbacks
- -export([id/1]).
- -export([init/2]).
+%% Mandatory Callbacks
+-export([init/2]).
 
- -export([pre_init_per_suite/3]).
- -export([post_init_per_suite/4]).
- -export([pre_end_per_suite/3]).
- -export([post_end_per_suite/4]).
+%% Optional Callbacks
+-export([id/1]).
 
- -export([pre_init_per_group/4]).
- -export([post_init_per_group/5]).
- -export([pre_end_per_group/4]).
- -export([post_end_per_group/5]).
+-export([pre_init_per_suite/3]).
+-export([post_end_per_suite/4]).
 
- -export([pre_init_per_testcase/4]).
- -export([post_init_per_testcase/5]).
- -export([pre_end_per_testcase/4]).
- -export([post_end_per_testcase/5]).
+-export([pre_init_per_testcase/4]).
+-export([post_end_per_testcase/5]).
 
- -export([on_tc_fail/4]).
- -export([on_tc_skip/4]).
+-export([on_tc_skip/4]).
 
- -export([terminate/1]).
+-export([terminate/1]).
 
- -record(state, { file_handle, total, suite_total, ts, tcs, data }).
+%% This hook state is threaded through all the callbacks.
+-record(state, {filename, total, suite_total, ts, tcs, data, skipped}).
+%% This example hook prints its results to a file, see terminate/1.
+-record(test_run, {total, skipped, suites}).
 
- %% Return a unique id for this CTH.
- id(Opts) ->
-   proplists:get_value(filename, Opts, "/tmp/file.log").
+%% Return a unique id for this CTH.
+%% Using the filename means the hook can be used with different
+%% log files to separate timing data within the same test run.
+%% See Installing a CTH for more information.
+id(Opts) ->
+    %% the path is relative to the test run directory
+    proplists:get_value(filename, Opts, "example_cth.log").
 
- %% Always called before any other callback function. Use this to initiate
- %% any common state. 
- init(Id, Opts) ->
-     {ok,D} = file:open(Id,[write]),
-     {ok, #state{ file_handle = D, total = 0, data = [] }}.
+%% Always called before any other callback function. Use this to initiate
+%% any common state.
+init(Id, _Opts) ->
+    {ok, #state{filename = Id, total = 0, data = []}}.
 
- %% Called before init_per_suite is called.
- pre_init_per_suite(Suite,Config,State) ->
-     {Config, State#state{ suite_total = 0, tcs = [] }}.
+%% Called before init_per_suite is called.
+pre_init_per_suite(_Suite,Config,State) ->
+    {Config, State#state{suite_total = 0, tcs = []}}.
 
- %% Called after init_per_suite.
- post_init_per_suite(Suite,Config,Return,State) ->
-     {Return, State}.
+%% Called after end_per_suite.
+post_end_per_suite(Suite,_Config,Return,State) ->
+    Data = {suites, Suite, State#state.suite_total,
+            lists:reverse(State#state.tcs)},
+    {Return, State#state{data = [Data | State#state.data],
+                         total = State#state.total + State#state.suite_total}}.
 
- %% Called before end_per_suite.
- pre_end_per_suite(Suite,Config,State) ->
-     {Config, State}.
+%% Called before each init_per_testcase.
+pre_init_per_testcase(_Suite,_TC,Config,State) ->
+    Now = erlang:monotonic_time(microsecond),
+    {Config, State#state{ts = Now, suite_total = State#state.suite_total + 1}}.
 
- %% Called after end_per_suite.
- post_end_per_suite(Suite,Config,Return,State) ->
-     Data = {suites, Suite, State#state.suite_total, lists:reverse(State#state.tcs)},
-     {Return, State#state{ data = [Data | State#state.data] ,
-                           total = State#state.total + State#state.suite_total } }.
+%% Called after each end_per_testcase.
+post_end_per_testcase(Suite,TC,_Config,Return,State) ->
+    Now = erlang:monotonic_time(microsecond),
+    TCInfo = {testcase, Suite, TC, Return, Now - State#state.ts},
+    {Return, State#state{ts = undefined, tcs = [TCInfo | State#state.tcs]}}.
 
- %% Called before each init_per_group.
- pre_init_per_group(Suite,Group,Config,State) ->
-     {Config, State}.
+%% Called when a test case is skipped by either user action
+%% or due to an init function failing.
+on_tc_skip(_Suite, _TC, _Reason, State) ->
+    State#state{skipped = State#state.skipped + 1}.
 
- %% Called after each init_per_group.
- post_init_per_group(Suite,Group,Config,Return,State) ->
-     {Return, State}.
+%% Called when the scope of the CTH is done.
+terminate(State) ->
+    %% use append to avoid data loss if the path is reused
+    {ok, File} = file:open(State#state.filename, [write, append]),
+    io:format(File, "~p.~n", [results(State)]),
+    file:close(File),
+    ok.
 
- %% Called before each end_per_group.
- pre_end_per_group(Suite,Group,Config,State) ->
-     {Config, State}.
-
- %% Called after each end_per_group.
- post_end_per_group(Suite,Group,Config,Return,State) ->
-     {Return, State}.
-
- %% Called before each init_per_testcase.
- pre_init_per_testcase(Suite,TC,Config,State) ->
-     {Config, State#state{ ts = now(), total = State#state.suite_total + 1 } }.
-
- %% Called after each init_per_testcase (immediately before the test case).
- post_init_per_testcase(Suite,TC,Config,Return,State) ->
-     {Return, State}
-
-%% Called before each end_per_testcase (immediately after the test case).
- pre_end_per_testcase(Suite,TC,Config,State) ->
-     {Config, State}.
-
- %% Called after each end_per_testcase.
- post_end_per_testcase(Suite,TC,Config,Return,State) ->
-     TCInfo = {testcase, Suite, TC, Return, timer:now_diff(now(), State#state.ts)},
-     {Return, State#state{ ts = undefined, tcs = [TCInfo | State#state.tcs] } }.
-
- %% Called after post_init_per_suite, post_end_per_suite, post_init_per_group,
- %% post_end_per_group and post_end_per_testcase if the suite, group or test case failed.
- on_tc_fail(Suite, TC, Reason, State) ->
-     State.
-
- %% Called when a test case is skipped by either user action
- %% or due to an init function failing.  
- on_tc_skip(Suite, TC, Reason, State) ->
-     State.
-
- %% Called when the scope of the CTH is done
- terminate(State) ->
-     io:format(State#state.file_handle, "~p.~n",
-                [{test_run, State#state.total, State#state.data}]),
-     file:close(State#state.file_handle),
-     ok.</code>
+results(State) ->
+    #state{skipped = Skipped, data = Data, total = Total} = State,
+    #test_run{total = Total, skipped = Skipped, suites = lists:reverse(Data)}.
+    </code>
   </section>
 
   <section>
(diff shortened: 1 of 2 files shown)


lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -454,19 +454,21 @@
 
  %% Called before each init_per_testcase.
  pre_init_per_testcase(Suite,TC,Config,State) ->
-     {Config, State#state{ ts = now(), total = State#state.suite_total + 1 } }.
+     Now = erlang:monotonic_time(microsecond),
```
paulo-ferraz-oliveira, line 457: Mind you where this once returned a [`timestamp()`](https://www.erlang.org/doc/man/erlang.html#type-timestamp), it's now returning an integer. Maybe a better equivalent to `now()` would be `erlang:timestamp()`?

fabjan: Yeah, perhaps.

`erlang:timestamp/0` is probably just fine if it makes the example simpler. And the monotonic part is probably not very interesting when analyzing test suite run times. It would make the diff in this PR smaller since it can be used with the same `now_diff`.

I just followed the "Measure Elapsed Time" advice in the docs from the deprecation and arrived at the solution in the PR:

> ##### Measure Elapsed Time
>
> Take time stamps with [erlang:monotonic_time/0](https://www.erlang.org/doc/man/erlang.html#monotonic_time-0) and calculate the time difference using ordinary subtraction. The result is in native [time unit](https://www.erlang.org/doc/man/erlang.html#type_time_unit). If you want to convert the result to another time unit, you can use [erlang:convert_time_unit/3](https://www.erlang.org/doc/man/erlang.html#convert_time_unit-3).
>
> An easier way to do this is to use [erlang:monotonic_time/1](https://www.erlang.org/doc/man/erlang.html#monotonic_time-1) with the desired time unit. However, you can then lose accuracy and precision.

paulo-ferraz-oliveira: Sure... In the same documentation, though, you'll find

> **Don't**
Use erlang:now/0 to retrieve the current Erlang system time.

> **Do**
Use [erlang:system_time/1](https://www.erlang.org/doc/man/erlang.html#system_time-1) to retrieve the current Erlang system time on the [time unit](https://www.erlang.org/doc/man/erlang.html#type_time_unit) of your choice.
If you want the same format as returned by erlang:now/0, use [erlang:timestamp/0](https://www.erlang.org/doc/man/erlang.html#timestamp-0).

I'm pretty sure backward compatibility isn't important here, but I was suggesting the change to have a simpler example. 😄 

And upon closer inspection I missed `timer:now_diff(now(), State#state.ts)` from your changes, so it's probably Ok to keep as-is.

fabjan: Oh, I see what you mean then. Yeah I changed both sides of the subtraction.

I agree, it would look like a simpler change if we use `erlang:timestamp()` instead. I don't have any strong opinion, I just followed the deprecation link and scanned the headings to find "Measure Elapsed Time" and thought that this example code was about measuring elapsed time more than "Retrieve Erlang System Time" so I didn't see the timestamp suggestion.

lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -454,19 +454,21 @@
 
  %% Called before each init_per_testcase.
  pre_init_per_testcase(Suite,TC,Config,State) ->
```
u3s, line 456: Example still generates warnings for multiple unused variables (e.g. `Suite`), which might be confusing for beginners.
Should unused variable names be prefixed with '_' to silence the unwanted output? What do you think?

fabjan: Hmm, OK I guess we had different linters configured. I can see what it looks like if we fix that as well.

But I think using a deprecated function is a bigger, and "new" issue.

The unused variables were there before and are the same issue now as before, whereas the problem with the deprecated function was introduced (well, kind of) recently by OTP evolving.

u3s: I agree with what you wrote :-)

lib/common_test/doc/src/ct_hooks_chapter.xml
fabjan: Hmm, the link from the reference manual for CT to this example section in the user guide says:

> For a minimal example of a CTH, see section [Example CTH](https://www.erlang.org/doc/apps/common_test/ct_hooks_chapter.html#example) in the User's Guide.

And I don't think this example is minimal.

Here is a version where only the warnings have been removed:
https://gist.github.com/fabjan/cc178d21f42a773cd45e43fe1b1416d7

Here is a minimal version keeping all the features of the current example: https://gist.github.com/fabjan/754840a6daf30058c00880b465448974

fabjan: But I also tried the code now for real and it's not really working at all any way I do it on my local machine, regardless of any changes.

The hooks crash when trying to use the opened file:
```
*** CT 2022-11-26 16:28:33.664 *** Suite Hook
Call to CTH failed: error:{terminated,
                           [{io,format,
                             [<0.656.0>,"~p.~n",
                              [{test_run,1,
                                [{suites,foo_SUITE,0,
                                  [{testcase,foo_SUITE,test,ok,
                                    409}]}]}]],
                             [{file,"io.erl"},
                              {line,94},
                              {error_info,
                               #{cause => {io,terminated},
                                 module => erl_stdlib_errors}}]},
                            {example_cth,terminate,1,
                             [{file,
                               "/Users/fabian/code/example_cth/src/example_cth.erl"},
                              {line,58}]},
                            {ct_hooks,catch_apply,3,
                             [{file,"ct_hooks.erl"},{line,491}]},
                            {ct_hooks,call_terminate,3,
                             [{file,"ct_hooks.erl"},{line,218}]},
                            {ct_hooks,call,4,
                             [{file,"ct_hooks.erl"},{line,301}]},
                            {ct_hooks,terminate,1,
                             [{file,"ct_hooks.erl"},{line,100}]},
                            {ct_hooks,terminate_if_scope_ends,3,
                             [{file,"ct_hooks.erl"},{line,374}]},
                            {ct_hooks,call,4,
                             [{file,"ct_hooks.erl"},{line,306}]}]}
```

It works If I move the file open to the terminate function together with the writing.
But if I use the hook in multiple suites (started via `Module:suite/0`) they stomp over each other's content in the log file.

Is this a problem with my local machine perhaps? Or is this related to the information at https://www.erlang.org/doc/apps/common_test/ct_hooks_chapter.html#cth-processes-and-tables

> CTHs are run with the same process scoping as normal test suites, that is, a different process executes the init_per_suite hooks then the init_per_group or per_testcase hooks. So if you want to spawn a process in the CTH, you cannot link with the CTH process, as it exits after the post hook ends. Also, if you for some reason need an ETS table with your CTH, you must spawn a process that handles it.

Is the IoDevice returned by file:open a process that is linked?

The fact that the suite data is overwritten by the next suite might be due to how I installed the hook I guess.

I'm creating a separate little repository where I can run some real code and tests, instead of just manipulating this example code inside a documentation article and running it in my head: https://github.com/fabjan/example_cth

(I don't expect anyone to treat this PR any differently, at this point I'm just curious as to how one could make this CT hook actually work)

u3s:
Thanks for taking a deeper look into that area!

I like the reduced [minimal example] (https://gist.github.com/fabjan/754840a6daf30058c00880b465448974). I think current example is more of a code skeleton than a minimal example.

I would suggest having it(minimal_example) in ct_hooks_chapter with following adjustments:
1. including underscored variable name - e.g. `_Suite` says more than `_`
2. leaving one of `on_tc_*` callbacks (even if empty) to indicate the presence of that callback category
3. I agree using a file in common hook is not a robust solution and has it disadvantages, but maybe it could be left alone with a small explanation of the limitations in the code comment. my thinking is that a file is one of easiest things to have with some observable side effects at the same time - at that is the main purpose of `minimal example`.
4. I'm not sure about exact problems which you have. I tried following and it worked:
   - take your minimal example
   - compile it with `erlc example_cth.erl`
   - create example_SUITE.erl from "Small Common Test suite" skeleton under Emacs
   - run suite with a hook with
     `ct_run -suite example_SUITE -pa . -ct_hooks example_cth`
   - Above worked even when /tmp/file.log was opened in Emacs.
5. add example command for trying hook module out (as a comment in minimal_example code)
   `ct_run -suite example_SUITE -pa . -ct_hooks example_cth`

Additionally (does not have to be included in this PR), I thought about following:
1. while at it, I think it would be worth adding a note in this [section](https://www.erlang.org/doc/apps/common_test/ct_hooks_chapter.html#installing-a-cth) explaining that a hook beam file needs to be added to code path to install it, therefore using `-pa .` might be needed

_I brought points 1 and 5 because regularly when I'm explaining CT hooks to students I got a question/suggestion that feature does not work when actually the forget to adjust the code path or compile the hook module before hand._

3. The example in current shape (with callbacks doing nothing) could be considered more of a template/skeleton and maybe could fit nicely in https://www.erlang.org/doc/apps/common_test/example_chapter.html ?
4. If you would like to be nice to Emacs users and know how to test it (a template code skeleton) - maybe you could also add a skeleton code `erlang-skels.el`

fabjan:
Cheers for the feedback! I'll take a stab at pulling the minimal example in here (and look at the other files you mentioned, I always want to be nice to Emacs users).

> I'm not sure about exact problems which you have. I tried following [...]

Yes it works when installing the hook via e.g. the CT command line, but when the hook is installed by a `spec/1` callback in the SUITE, the version of the hook that opens the file on init and writes at the end gets io errors for me. If the open is moved to the end just before the write, the errors go away but if the mode `append` is included no data is lost (but there is one summary per hook install so it seems the init happens multiple times, even if the same hook ID was returned at the start.

u3s:
1. for issues with running hook from spec - can you check the pids executing each hook callback and if the one executing `init` is alive in `terminate` ?- AFAIK, they might differ. if process opening file is not alive anymore when `terminate` is executed you might face issues (hard to say why this is not visible when hook is installed via command line, I see process differ as well for each callback).
 - if this improves robustness and is aligned with _minimal_ concept of the example - maybe file should be opened and closed in `terminate` callback (example hook is collecting execution data in the memory anyway)
 - I've noticed in your repo that you're installing 2 hooks which is of course fine - but please check if the hooks code doesn't interact in unwanted manner
 - see shell interaction below (I think it illustrates your error)
2. BTW, please check indentation (so 4 spaces are used)

```
1> {ok, Device} = file:open("/tmp/file.log", [write]).
{ok,<0.86.0>}
2> self()
2> .
<0.84.0>
3> io:format(Device, "Hello", []).
ok
4> 1/0.
** exception error: an error occurred when evaluating an arithmetic expression
     in operator  '/'/2
        called as 1 / 0
5> self().
<0.90.0>
6> Device.
<0.86.0>
7> io:format(Device, "Hello", []).
** exception error: terminated
     in function  io:format/3
        called as io:format(<0.86.0>,"Hello",[])
        *** argument 1: the device has terminated
```

fabjan:
I added the minimal example hook now, with the adjustments discussed above. I think I have addressed all the suggestions above, either in the code or in my comments here (except for the emacs skeleton, will look at that next).

For the sake of the review, may I suggest that any further issues with the code changes here are discussed under the ["Files changed"](https://github.com/erlang/otp/pull/6437/files) tab? I think that makes it easier to manage the discussions as separate threads when the issues are separate.

The indentation question I think is not quite resolved, see [my side note comment above](https://github.com/erlang/otp/pull/6437#issuecomment-1339685026).

fabjan:
I put the suggested Emacs skeleton in this PR as well, we can remove it if you prefer.

Tested by putting this in my ~/.emacs.d/init.el:

```elisp
;; Set up Erlang mode
(setq load-path (cons "~/code/otp/lib/tools/emacs" load-path))
(require 'erlang-start)
```

The skeleton showed up in the "Erlang" menu, and produces this:

Skärmavbild 2022-12-06 kl  21 16 02

I copied that module and tested in my example_cth repository (linked in the thread above) and it logged this:

```
=NOTICE REPORT==== 6-Dec-2022::21:15:26.870679 ===
test_hook is done: {state,4,2,0,1,0}
```

lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -99,6 +99,10 @@
     <item><c>{ct_hooks,[{my_cth_module,[{debug,true}],500}]}</c></item>
     </list>
 
+    <p>Note that regardless of how you install a CTH, its beam file
```
paulo-ferraz-oliveira, line 102:
```suggestion
    <p>Note that regardless of how you install a CTH, its BEAM file
```

lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -380,111 +384,85 @@
      </p>
      <code>
  %%% Common Test Example Common Test Hook module.
+ %%%
+ %%% To use this hook, on the command line:
+ %%%     ct_run -suite example_SUITE -pa . -ct_hooks example_cth
+ %%%
+ %%% N.B. `-pa .`, the hook beam file must be in the code path when installing.
```
paulo-ferraz-oliveira, line 391:
```suggestion
 %%% N.B. `-pa .`, the hook BEAM file must be in the code path when installing.
```

lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -380,111 +384,85 @@
...
-                [{test_run, State#state.total, State#state.data}]),
-     file:close(State#state.file_handle),
-     ok.</code>
+     %% use append to avoid data loss if the path is reused
+     {ok, File} = file:open(State#state.filename, [write, append]),
+     io:format(File, "~p.~n", [results(State)]),
+     file:close(File),
+     ok.
+
+ results(State) ->
```
u3s, line 462: > I left the extra space at the beginning to try and make sure the diff for my changes is a as legible as possible, showing only the meaningful changes. Would you like me to in this PR also get rid of that first extra column of space characters? (Maybe the diff preservation is not as interesting now that we are discussing bigger changes anyway)

Please remove the extra space in the example code.  We will see how it behaves ... it looks artificial, maybe there is some reason I'm missing ... let us check and hope it will be fine:-)

u3s (commented):
Sorry for delay. This PR has lower priority and must wait longer.

1. Please fix minor comments I've added in code.
2. can you set target of PR to master branch, please?
3. I've played around with adding hook modules via ct_run parameter and from suite/0 function in test suite. My observations are that, `Module:init/2` behaves as expected (i.e. if the hook module returns the same id - init will be called only once).  There are more calls to `Module:id/1`, but I presume this is needed for checking hook ID by the framework and making the decision.

So from my perspective, there is nothing suspicious at this stage. However if you still observe some anomaly worth checking, please create separate issue with reproduction steps and we will try to clarify that.

lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -99,6 +99,10 @@
     <item><c>{ct_hooks,[{my_cth_module,[{debug,true}],500}]}</c></item>
     </list>
 
+    <p>Note that regardless of how you install a CTH, its BEAM file
+      must be available in the code path when Common Test runs.
```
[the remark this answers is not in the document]
fabjan: I'm using "its" as a possessive of the pronoun “it” in this paragraph. I'm not a native speaker of English but I think there is no problem here. "It" refers to the hook (CTH). Its BEAM file must be present in the path.

lib/common_test/doc/src/ct_hooks_chapter.xml
```diff
@@ -380,111 +384,85 @@
      </p>
      <code>
  %%% Common Test Example Common Test Hook module.
+ %%%
+ %%% To use this hook, on the command line:
+ %%%     ct_run -suite example_SUITE -pa . -ct_hooks example_cth
+ %%%
+ %%% N.B. `-pa .`, the hook BEAM file must be in the code path when installing.
```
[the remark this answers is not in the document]
fabjan: Ah, no problem! Will fix. How about I just write "Note: " instead?

lib/common_test/doc/src/ct_hooks_chapter.xml
fabjan: 👍

lib/common_test/doc/src/ct_hooks_chapter.xml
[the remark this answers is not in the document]
u3s: OK. leave it. just wanted to clarify.

lib/common_test/doc/src/ct_hooks_chapter.xml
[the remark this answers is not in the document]
u3s: Good enough!

fabjan:
> Sorry for delay. This PR has lower priority and must wait longer.

No worries, that makes sense.

1 and 2: will fix and update PR.

3: yeah I agree it should be treated separately, I think that's something I experienced when I was trying the file handle issue out in different ways. If I can reproduce something meaningful I can open a separate issue.

fabjan:
The suggestions should all be adressed now.

fabjan:
The build failed after curling the GitHub API into `jq`: https://github.com/erlang/otp/actions/runs/3741792360/jobs/6353967342#step:8:6181

Since it's inside a shell pipeline we can't see what the cause was, but it's possible `curl -sL "https://api.github.com/repos/$1/tags"`was rate limited. If you authenticate that call its rate limit will increase from 60 per hour to 1000 per hour (as per current GitHub API policy).

u3s:
I've pushed small commit removing some spaces and wrapping code line. I wanted to check if we can get rid of scrollbar under code example and improve readability.

If it it works and you are fine with this suggestion - please squash the commits on the branch and we will merge it.

fabjan:
I'm fine with that formatting 👍, but not sure how to check if it renders better. I'll squash the commits though.

u3s:
> I'll squash the commits though.

Thanks

> ... but not sure how to check if it renders better. 

easiest way I'm aware of is to navigate to HTML doc generated for PR. once processed link should appear in area visualized on screenshot below (it usually works).

image

u3s:
no horizontal scroll bar now :-). at least in my browsers.

https://erlang.github.io/prs/6437/lib/common_test-1.23.2/doc/html/ct_hooks_chapter.html#example-cth

fabjan:
Nice, yeah for me too 👍

PR #6443, opened 2022-11-07
emulator: statistics_SUITE timetraps on nenya, increase timetrap

frazze-jobb:
Make timetrap scale based on number of schedulers, but set a lower limit to 4 minutes


frazze-jobb:
This commit doesn't really improve anything in this testcase so I will just close it.

PR #6449, opened 2022-11-08
 Add missing type specs in http library and replace `essl` option by `ssl`

### LINKED ISSUE #6245: httpc module is unspecced

kikofernandez:
This PR adds type specs to the API of the http_client app, file httpc.erl.
I have updated the XML documentation to use the type specs from httpc.erland this fixes https://github.com/erlang/otp/issues/6245

For some addressed comments, see #6300
Closes #6245

### DIFF
--- a/lib/inets/src/http_client/httpc.erl
+++ b/lib/inets/src/http_client/httpc.erl
@@ -53,11 +53,11 @@
 	 stop_service/1, 
 	 services/0, service_info/1]).
 
+
 -include_lib("inets/src/http_lib/http_internal.hrl").
 -include("httpc_internal.hrl").
 
 -define(DEFAULT_PROFILE, default).
-
 
 %%%=========================================================================
 %%%  API
@@ -66,7 +66,8 @@
 default_profile() ->
     ?DEFAULT_PROFILE.
 
-
+-spec profile_name(pid()) -> pid();
+                  (atom()) -> pid() | atom().
 profile_name(?DEFAULT_PROFILE) ->
     httpc_manager;
 profile_name(Profile) when is_pid(Profile) -> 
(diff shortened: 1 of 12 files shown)


paulo-ferraz-oliveira:
@kikofernandez, good job on this `-spec(_)`ing. Do you think this PR could be linked to #6245 so that that one gets closed when this gets merged? Maybe add `Closes #6245` to the description on the comment (?).

lib/inets/src/http_client/httpc.erl
```diff
@@ -53,11 +53,93 @@
...
 -include_lib("inets/src/http_lib/http_internal.hrl").
 -include("httpc_internal.hrl").
 
 -define(DEFAULT_PROFILE, default).
 
+%%%=========================================================================
+%%%  Type specifications
+%%%=========================================================================
+
+-opaque request_id() :: reference().
```
KennethL, line 82: `request_id` looks strange in the documentation since it is a type without definition. 
I don't think it can be an `opaque` type since the purpose is that the user is matching `request_id`s for equality.
Compare with ssl_socket in the ssl module.
Probably the type should be `request_id() :: any()` or Dialyzers support for `opaque` or the definition of `opaque` must be revisited.

KennethL (changes requested):

lib/inets/src/http_client/httpc.erl
kikofernandez: Yes, good catch!

lib/inets/src/http_client/httpc.erl
```diff
@@ -53,11 +53,93 @@
...
+             , body/0, http_body_result/0, request_result/0]).
+
+% SSL options
+-export_type([ ssloption/0, http_option/0]).
+
+% HTTP Option types
+-export_type([ option_request/0, stream_to/0, body_format/0
+             , receiver/0, receiver_module/0, receiver_function/0, receiver_args/0
+             , cookie_header_opt/0]).
+
```
KennethL, line 72: I think we have to discuss which types we should export. For example `receiver_module`, `receiver_function` and `receiver_args` are quite unnecessary to have types for at all. See next comment.

lib/inets/src/http_client/httpc.erl
```diff
@@ -53,11 +53,93 @@
...
+                        | {headers_as_is, boolean()}
+                        | {socket_opts, [socket_opt()]}
+                        | {receiver, receiver()}
+                        | {ipv6_host_with_brackets, boolean()}.
+-type stream_to() :: none | self | {self, once} | file:name_all().
+-type body_format() :: string() | binary() | atom().
+-type receiver() :: pid() | fun((term()) -> term()) | {receiver_module(), receiver_function(), receiver_args()}.
+-type receiver_module() :: atom().
+-type receiver_function() :: atom().
+-type receiver_args() :: list().
```
KennethL, line 141:
```suggestion
-type receiver() :: pid() 
    |  fun((term()) -> term())
    | {Receiver_module::atom(), Receiver_function::atom(), Receiver_args::list()}.
```

KennethL (changes requested):
I think we should discuss what types we should export, I am not sure what is the best, but I don't think it is good to export too many types since they become part of the public interface and you have a hard time changing them later.

I think simple types which are only atom() or list() are especially questionable as it can be suspected that it is just to get a name in the documentation, which I showed is possible anyway.

KennethL (approved):

PR #6450, opened 2022-11-08
beam_lib: Fix spec for all_chunks/1

the-mikedavis:
`beam_lib:all_chunks/1` returns the module contained in the given beam file/binary rather than the `beam_lib` atom if successful. Also see [`beam_lib:scan_beam/2`](https://github.com/erlang/otp/blob/54e4353c70fc4eb6ddc8fefd1475ca60366ffeeb/lib/stdlib/src/beam_lib.erl#L579-L581) which originally returns the module.


bjorng:

PR #6447, opened 2022-11-07
beam_disasm: Fix bs_create_bin/6 resolution

the-mikedavis:
The `bs_create_bin/6` instruction was previously not fully translated by `beam_disasm`. This change resolves some of the args in the instruction (unwrapping `u` tags) and resolves the operation list to satisfy the checks in `beam_validator:verify_create_bin_list/2`.

PR #6451, opened 2022-11-08
beam_disasm: Resolve commands for bs_match/3 instruction

the-mikedavis:
This change is similar to #6447: the command list was not yet resolved when translating the `bs_match/3` instruction. This change decodes the command list into the shape expected by `beam_validator:validate_bs_match/4`.


bjorng:
Thanks!

PR #6452, opened 2022-11-08
beam_disasm: Decode field flags

the-mikedavis:
`beam_validator:bs_integer_type/2` expects field flags to be a list of atoms (little, signed, native, etc.). This change reverses the bitflag encoding done in `beam_asm:flag_to_bit/1` and `beam_asm:encode_arg/2` to decode the field flags in `beam_disasm` as `beam_validator` expects.


bjorng:
Thanks!

PR #6457, opened 2022-11-10
empty containers: more precise specs

ilya-klyuchnikov:
with #6441 merged, - it's now possible to provide the most precise specs for empty containers.

(the rationale is to make OTP specs more precise and accurate for type-checking)


bjorng:

Maria-12648430:
@ilya-klyuchnikov @bjorng

I have only seen this now because I went to investigate why dialyzer suddenly started complaining to me. I understand that now, all fine. But it made me think that there is something definitely wrong with the specs in the modules touched by this PR. I'll use `queue` as illustration, that is probably the most simple.

```erlang
-spec in(Item, Q1 :: queue(Item)) -> Q2 :: queue(Item).
```
This is _not_ correct, is it? If we assume that `Q1` is a fresh queue as returned from `new/0`, and `new/0` now returning `queueu(none())`, obviously we can't insert anything into such a queue without violating this spec. It was ok before because the return type of `new/0` was specced as `queue(any())`; but even back then, queues returned from `from_list/1` were a different story, prone to violate the spec.

Long story short: shouldn't the specs of `in/2` and friends (and their respective equivalents in the other modules) be like this?
```erlang
-spec in(InItem, Q1 :: queue(QueueItem)) -> Q2 :: queue(QueueItem | InItem).
```

jhogberg:
> This is _not_ correct, is it? If we assume that `Q1` is a fresh queue as returned from `new/0`, and `new/0` now returning `queue(none())`, obviously we can't insert anything into such a queue without violating this spec. It was ok before because the return type of `new/0` was specced as `queue(any())`; but even back then, queues returned from `from_list/1` were a different story, prone to violate the spec.

Nope, and it doesn't even work for non-`none()` either: you can't add anything other than `1` to a `queue(1)`!

Repeated type variables in a spec are nearly always wrong.

> Long story short: shouldn't the specs of `in/2` and friends (and their respective equivalents in the other modules) be like this?
> 
> ```erlang
> -spec in(InItem, Q1 :: queue(QueueItem)) -> Q2 :: queue(QueueItem | InItem).
> ```

Yes, that's closer to the truth, but the type language is a mess at the moment so we've decided not to change any specs before we've cleaned it up.

(For example, `queue(1)` is not a subset, that is `::`, of `queue(integer())` :vomiting_face:)

Maria-12648430:
Okaaay... :thinking::raised_eyebrow::flushed::exploding_head:

ilya-klyuchnikov:
> (For example, queue(1) is not a subset, that is ::, of queue(integer()) 🤮)

`queue:queue` is an opaque type and subtyping for parameterised opaque types would involve explicit declaration of variance - see https://github.com/erlang/otp/issues/5240 for detailed discussion.

Logically, `queue(A)` should be covariant: that is if `A <: B` then `queue(A)` <: `queue(B)`.

With covariance `queue(1)` is a subtype of `queue(integer())`. Unfortunately, the official Erlang documentation doesn't have clarity on this.

PR #6459, opened 2022-11-11
Make encoding errors of known extensions handable by the verify_fun

### LINKED ISSUE #6402: `pubkey_cert_records:decode_extensions` throws on malformed extensions
**Describe the bug**
When a certificate contains a known extension which is malformed, `pubkey_cert_records:decode_extensions` throws instead of returning a validation error, or simply skipping the extension. This breaks valid certificate validation paths and does not allow graceful handling of the error.

**To Reproduce**
```erlang
public_key:pkix_decode_cert(
   base64:decode(
<<"MIICXDCCAgKgAwIBAgIBATAKBggqhkjOPQQDAjApMRkwFwYDVQQFExBjOTY4NDI4OTMyNzUwOGRiMQwwCgYDVQQMDANURUUwHhcNMjIxMDI5MTczMTA3WhcNMjkwNDE2MjAzNDUzWjAfMR0wGwYDVQQDExRBbmRyb2lkIEtleXN0b3JlIEtleTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABFmIQDus/jIZ0cPnRCITCzUUuCjQBw8MetO6154mmTL8O/fFlGgYkZ6C8jSSntKC/lMwaZHxAgW1AGgoCrPuX5ejggEjMIIBHzALBgNVHQ8EBAMCB4AwCAYDVR0fBAEAMIIBBAYKKwYBBAHWeQIBEQSB9TCB8gIBAgoBAQIBAwoBAQQgyvsSa116xqleaXs6xA84wqpAPWFgaaTjCWBnZpHslmoEADBEv4VFQAQ+MDwxFjAUBAxjb20ud2hhdHNhcHACBA0+oAQxIgQgOYfQQ9EK769ahxCzZxQY/lfg4ZtlPJ34JVj+tf/OXUQweqEFMQMCAQKiAwIBA6MEAgIBAKUIMQYCAQYCAQSqAwIBAb+DdwIFAL+FPQgCBgGEJMweob+FPgMCAQC/hUAqMCgEIFNB5rJkaXmnDldlMAeh8xAWlCHsm92fGlZI91reAFrxAQH/CgEAv4VBBQIDAV+Qv4VCBQIDAxUYMAoGCCqGSM49BAMCA0gAMEUCIF0BwvRQipVoaz5SIhsYbIeK+FHbAjWPgOxWgQ6Juq64AiEA83ZLsK37DjZ/tZNRi271VHQqIU8mdqUIMboVUiy3DaM=">>
   ),
otp).
```

**Expected behavior**
Ideally, the extension should not be included in the final parsed result, and optionally an error should be returned.

**Affected versions**
Erlang/OTP 25 [erts-13.0.2]

**Additional context**
The certificate above does not adhere to [RFC5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.13) and has an empty `cRLDistributionPoints` extension. However, the certificate is generated by Android in hardware and it is valid for all other purposes. It is part of a valid path and the remaining extensions are valid. Unfortunately, OTP's behaviour in this case is "fail all" and does not allow us to recover gracefully.

IngelaAndin:
close #6402

### DIFF
--- a/lib/public_key/src/pubkey_cert_records.erl
+++ b/lib/public_key/src/pubkey_cert_records.erl
@@ -292,12 +292,17 @@ decode_extensions(asn1_NOVALUE) ->
 decode_extensions(Exts) ->
     lists:map(fun(Ext = #'Extension'{extnID=Id, extnValue=Value0}) ->
 		      case extension_id(Id) of
-			  undefined -> Ext;
+			  undefined ->
+                              Ext;
 			  Type ->
-			      {ok, Value} = 'OTP-PUB-KEY':decode(Type, iolist_to_binary(Value0)),
-			      Ext#'Extension'{extnValue=transform(Value,decode)}
-		      end
-	      end, Exts).
+                              case 'OTP-PUB-KEY':decode(Type, iolist_to_binary(Value0)) of
+                                  {ok, Value} ->
+                                      Ext#'Extension'{extnValue=transform(Value,decode)};
+                                  {error, _} ->
+                                      Ext
+                              end
+                      end
+              end, Exts).
 
 encode_extensions(asn1_NOVALUE) ->
     asn1_NOVALUE;
--- a/lib/public_key/test/public_key_SUITE.erl
+++ b/lib/public_key/test/public_key_SUITE.erl
@@ -89,6 +89,8 @@
          pkix_countryname/1,
          pkix_emailaddress/0,
          pkix_emailaddress/1,
+         pkix_decode_cert/0,
+         pkix_decode_cert/1,
          pkix_path_validation/0,
          pkix_path_validation/1,
          pkix_path_validation_root_expired/0,
@@ -147,6 +149,7 @@ all() ->
      pkix, 
      pkix_countryname, 
      pkix_emailaddress, 
+     pkix_decode_cert,
      pkix_path_validation,
      pkix_path_validation_root_expired,
      pkix_iso_rsa_oid, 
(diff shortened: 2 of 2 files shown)


andreasanta:
Thank you! I think that almost solves it! I added a minor comment because I think there is an error being thrown when parsing fails. I agree that we should delegate control to `verify_fun` and let the caller decide.

IngelaAndin:
@andreasanta I adjusted the implementation as you are correct that an error will be raised. Do you have a certificate that you could contribute as a test case with an extension that will fail the decoding?  As for your other question, the module `OTP-PUBLIC-KEY` is generated by the asn1 compiler so enhancements in that area would be enhancments to the ASN1 compiler.

IngelaAndin:
@andreasanta  just realized you provided one in the issue, thanks I will use that!

IngelaAndin:
@andreasanta  also realized my first solution was correct an error tuple will actually be returned, so I changed the code back.

andreasanta:
> @andreasanta also realized my first solution was correct an error tuple will actually be returned, so I changed the code back.

Fantastic, thanks.

Please use this for unit tests:

```erlang
public_key:pkix_decode_cert(
   base64:decode(
<<"MIICXDCCAgKgAwIBAgIBATAKBggqhkjOPQQDAjApMRkwFwYDVQQFExBjOTY4NDI4OTMyNzUwOGRiMQwwCgYDVQQMDANURUUwHhcNMjIxMDI5MTczMTA3WhcNMjkwNDE2MjAzNDUzWjAfMR0wGwYDVQQDExRBbmRyb2lkIEtleXN0b3JlIEtleTBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABFmIQDus/jIZ0cPnRCITCzUUuCjQBw8MetO6154mmTL8O/fFlGgYkZ6C8jSSntKC/lMwaZHxAgW1AGgoCrPuX5ejggEjMIIBHzALBgNVHQ8EBAMCB4AwCAYDVR0fBAEAMIIBBAYKKwYBBAHWeQIBEQSB9TCB8gIBAgoBAQIBAwoBAQQgyvsSa116xqleaXs6xA84wqpAPWFgaaTjCWBnZpHslmoEADBEv4VFQAQ+MDwxFjAUBAxjb20ud2hhdHNhcHACBA0+oAQxIgQgOYfQQ9EK769ahxCzZxQY/lfg4ZtlPJ34JVj+tf/OXUQweqEFMQMCAQKiAwIBA6MEAgIBAKUIMQYCAQYCAQSqAwIBAb+DdwIFAL+FPQgCBgGEJMweob+FPgMCAQC/hUAqMCgEIFNB5rJkaXmnDldlMAeh8xAWlCHsm92fGlZI91reAFrxAQH/CgEAv4VBBQIDAV+Qv4VCBQIDAxUYMAoGCCqGSM49BAMCA0gAMEUCIF0BwvRQipVoaz5SIhsYbIeK+FHbAjWPgOxWgQ6Juq64AiEA83ZLsK37DjZ/tZNRi271VHQqIU8mdqUIMboVUiy3DaM=">>
   ),
otp).
```

IngelaAndin:
Unfortunately, this sort of general discarding of decoding errors could cause problems in other places, such as CRL validation. And probably in general is better not to allow arbitrary decoding errors. We are thinking of making the solution more of a specific workaround for this specific error that will also not mess with the CRL validation.  I will update the PR once we have had time to work something out.

andreasanta:
I see. But in theory, if the default `verify_fun` rethrows an error in this case, unless the callee override the fun, it should still behave as expected by other libraries, no?

andreasanta:
Hi @IngelaAndin may I ask how is the change coming along, do you have any blockers you'd like to discuss? Thanks.

potatosalad:
@IngelaAndin Is there anything we could do to help out with to move forward with a solution?

I assume the hold-up is partly because the resulting decoded `#'OTPCertificate'{}` is invalid because it cannot be re-encoded as-is?

Would it make sense to have (1) a function or module callback that could be optionally provided during decode so that these errors could be handled and (2) some sort of extension to the `SupportedExtensions` in `#'Extension'{}` which allows `extnValue` to be a "raw value" or "decoding error value"?

Context: we are running into a significant number of production decoding errors related to [Android-based hardware-backed Key Attestation](https://developer.android.com/training/articles/security-key-attestation).

We appreciate your help, please let me know if there's anything we could do to arrive at an agreeable solution.

andreasanta:
@potatosalad @IngelaAndin Potentially we can also keep the current behaviour by rethrowing in case a malformed exception is found. And we could control that behaviour with a flag passed as options to `pkix_decode_cert` and  `pkix_path_validation`.

For instance:
`public_key:pkix_decode_cert(..., otp, {do_not_throw, true}).`

In any case, currently the implementation throws a `badmatch`, I would at least recommend to throw a specific error with the extension name that failed validation, so it can be trapped by the caller and floated to the user. If we fixed this and passed the error to `verify_fun` passed to `pkix_path_validation`, by default it could rethrow and a customised one could instead trap a specific extension name and handle gracefully.

Thank you both for your efforts 🙏

KennethL:
@IngelaAndin and I have discussed this and we think we can find a solution. We have now primarily thought about the problem when setting up a TLS connection and the certificate contains extensions of which one or some of them have an erroneous data.
The problem as we see it is that we have a number of known extensions (hard coded) and if the decoding of one of them fails the whole connection setup fails or the whole handling of the certificate fails. And this failure happens even if the faulty extension is not needed further on in the connection setup (or other handling of the certificate). In this particular case it is the extension `CRLDistributionPoints` which is faulty but it is not really needed unless the user has ordered a CRL validation with `{crl_check true}`. What we can do is to remove the `CRLDistributionPoints` from the known extensions or redefine the known extensions to mean the mandatory decoded extensions and let the other extensions be decoded when they are needed. This would solve this particular problem. It would also be good to return the failing extension whenever there is a failure during decode. We we look further into this after we have released OTP 26 RC1 next week.

IngelaAndin:
Replaced by #6883

PR #6462, opened 2022-11-14
beam_validator: Eliminate type mismatch

### LINKED ISSUE #6458: [erlc] internal consistency check failed
On master, both with and without 705a255dffae09d07f9ba4087b4e90267b21d5e2:
```erlang
-module(call_only).
-compile([export_all]).

f1({X}) when X; (X orelse false) ->
    (X orelse {}),
    [
        {X}#{
            f2() orelse X => []
        }
     || _ <- []
    ].

f2() ->
    0.
```
results in an erlc crash with the following message:
```
call_only:1: function f1/1+27:
  Instruction: {call_only,3,{f,16}}
  Error:       {bad_arg_type,
                   {x,1},
                   {t_tuple,1,true,#{1 => {t_atom,[false,true]}}},
                   {t_union,none,none,none,
                       [{{1,{t_atom,[false]}},
                         {t_tuple,1,true,#{1 => {t_atom,[false]}}}},
                        {{1,{t_atom,[true]}},
                         {t_tuple,1,true,#{1 => {t_atom,[true]}}}}],
                       none}}:
```

Somewhat surprisingly, I've not been able to reduce this testcase any more; every single part seems required for the crash to reproduce.

bjorng:
When compiling the following code:

    f1({X}) when X; (X orelse false) ->
        (X orelse {}),
        [
            {X}#{
                f2() orelse X => []
            }
         || _ <- []
        ].

    f2() ->
        0.

the `beam_validator` pass rejects it:

    call_only:1: function f1/1+27:
      Instruction: {call_only,3,{f,16}}
      Error:       {bad_arg_type,
                       {x,1},
                       {t_tuple,1,true,#{1 => {t_atom,[false,true]}}},
                       {t_union,none,none,none,
                           [{{1,{t_atom,[false]}},
                             {t_tuple,1,true,#{1 => {t_atom,[false]}}}},
                            {{1,{t_atom,[true]}},
                             {t_tuple,1,true,#{1 => {t_atom,[true]}}}}],
                           none}}:

What happens here is that when types are updated in different ways, the result will be different. That is not supposed to happen; there should be one canonical type representation regardless how it's reached.

This pull request resolves the bug for this particular example, but a more general solution might be needed.

Closes #6458

### DIFF
--- a/lib/compiler/src/beam_validator.erl
+++ b/lib/compiler/src/beam_validator.erl
@@ -2547,12 +2547,29 @@ update_container_type(Type, Ref, #vst{current=#st{vs=Vs}}=Vst) ->
     case Vs of
         #{ Ref := #value{op={bif,element},
                          args=[{integer,Index},Tuple]} } when Index >= 1 ->
-            Es = beam_types:set_tuple_element(Index, Type, #{}),
-            TupleType = #t_tuple{size=Index,elements=Es},
-            update_type(fun meet/2, TupleType, Tuple, Vst);
+            case {Index,Type} of
+                {1,#t_atom{elements=[_,_|_]}} ->
+                    %% The first element is one atom out of a set of
+                    %% at least two atoms. We must take care to
+                    %% construct an atom set.
+                    update_type(fun meet_tuple_set/2, Type, Tuple, Vst);
+                {_,_} ->
+                    Es = beam_types:set_tuple_element(Index, Type, #{}),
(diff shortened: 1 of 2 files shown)


bjorng:
After some discussions and thinking, @jhogberg and I have decided to not pursue a general solution at this point. We will reconsider if similar bugs turn up. I have turned this PR into a real PR and added it our daily builds.

PR #6469, opened 2022-11-14
Fix mistranslations in beam_disasm

the-mikedavis:
This PR builds on some prior work in #6447, #6451, #6452 for resolving instructions in `beam_disasm` so that they mirror `beam_asm` and are accepted by `beam_validator`. Each commit is a separate fix and the commit messages explains each change. (I can break these into separate PRs if that is desired.)

Thank you!

### DIFF
--- a/lib/compiler/src/beam_disasm.erl
+++ b/lib/compiler/src/beam_disasm.erl
@@ -761,9 +761,13 @@ resolve_names(Fun, Imports, Str, Lbls, Lambdas, Literals, M) ->
     [resolve_inst(Instr, Imports, Str, Lbls, Lambdas, Literals, M) || Instr <- Fun].
 
 %%
-%% New make_fun2/4 instruction added in August 2001 (R8).
-%% We handle it specially here to avoid adding an argument to
+%% Instructions that need to look up an entry in the Lambda table.
+%% We handle these specially here to avoid adding an argument to
 %% the clause for every instruction.
+%%
+%% - make_fun2/4 (R8, added in August 2001)
+%% - make_fun3/3 (OTP 24)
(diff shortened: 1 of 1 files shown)


bjorng:
The reason that we have the `beam_disasm` module at all is because it was used as a part of the [HiPE](https://www.it.uu.se/research/group/hipe/) native compiler when compiling from BEAM files. (HiPE was removed in OTP 24.) It was never a priority to make the output of `beam_diasam` identical to that of `.S` files. For example, `arithbif` was more convenient for the HiPE compiler because arithmetic instructions were translated differently compared to other guard BIF calls. You might want to adjust some of your commit messages (especially those about `arithbif` and `raise`).

the-mikedavis:
Oh, I see now, thank you for the pointer! I have reworded some of the commit messages (mostly the arithfbif and raise ones).

> It was never a priority to make the output of `beam_diasam` identical to that of `.S` files.

Now that HiPE is removed, is it ok to patch `beam_disasm` to emit `.S` output? Or are you looking to remove `beam_disasm` in the future?

We use this in [rabbitmq/khepri](https://github.com/rabbitmq/khepri) to compile transaction functions into self-contained modules (through `code:get_object_code/1`, `beam_disasm:file/1`, and `compile:forms/2` with the `from_asm` option). It's very close to working out-of-the-box: we just needed these translations and to add some `var_info` type and match-context annotations to satisfy `beam_validator`.

bjorng:
> Now that HiPE is removed, is it ok to patch `beam_disasm` to emit `.S` output?

Yes.

> Or are you looking to remove `beam_disasm` in the future?

We have no plans to remove it.

bjorng:
Thanks for your pull request!

PR #6487, opened 2022-11-18
pg: Do not send redundant sync 'messages'

sverker:
When working on fixing `pg_SUITE:forced_sync` sporadic failures, I found out there are quite common for redundant `discover` and `sync` messages to be sent. So this PR is me wondering...

@max-au: Why would a second `sync` message toward the same `pg` process be necessary at all?

When we receive a `discover` we monitor the sender and send a `sync` with all our local groups as reply. That's the only way a `sync` is sent. If we then get a second `discover` from the same process, why would that need a second `sync` reply? Any updates to our local groups are sent anyway as `join` and `leave` messages as I understand it.

All `pg_SUITE` tests passed unchanged on my 8-core machine, except white-box test `pg_SUITE:forced_sync` that needed to be tweaked to not expect sending of redundant `sync`. With that change one would question its purpose, or am I missing something?

### DIFF
--- a/lib/kernel/src/pg.erl
+++ b/lib/kernel/src/pg.erl
@@ -438,12 +438,12 @@ terminate(_Reason, #state{scope = Scope}) ->
 %% Internal implementation
 
 handle_discover(Peer, #state{remote = Remote, local = Local} = State) ->
-    gen_server:cast(Peer, {sync, self(), all_local_pids(Local)}),
     %% do we know who is looking for us?
     case maps:is_key(Peer, Remote) of
         true ->
             {noreply, State};
         false ->
+            gen_server:cast(Peer, {sync, self(), all_local_pids(Local)}),
             MRef = erlang:monitor(process, Peer),
             erlang:send(Peer, {discover, self()}, [noconnect]),
             {noreply, State#state{remote = Remote#{Peer => {MRef, #{}}}}}
--- a/lib/kernel/test/pg_SUITE.erl
+++ b/lib/kernel/test/pg_SUITE.erl
@@ -551,12 +551,12 @@ forced_sync(Config) when is_list(Config) ->
     {?FUNCTION_NAME, Node} ! {discover, whereis(?FUNCTION_NAME)},
     sync({?FUNCTION_NAME, Node}),
     sync(?FUNCTION_NAME),
-    ?assertEqual(Expected, lists:sort(pg:get_members(?FUNCTION_NAME, one))),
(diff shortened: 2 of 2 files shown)


max-au:
The change makes sense to me, but I am also interested in the source of redundant `discover` messages. I can think of only one case, a race condition between `net_kernel:monitor_nodes` and `nodes()` call in the `init/1` call. It is possible that `nodeup` message was in the queue during `pg` startup. This appears to be a very rare event. Did I miss any other cases?

In the early stages I used redundant `sync` message to force synchronisation (mainly to work around some heisenbug). I left the capability there because I did not expect `discover` duplicates.

I attempted to run this patch with the property-based test that I have, https://github.com/max-au/spg/blob/master/test/spg_cluster_SUITE.erl and it fails consistently with `badkey`

```
{{{badkey,three}, [{erlang,map_get, [three,#{}], [{error_info, #{module => erl_erts_errors}}]},
  {spg,'-leave_remote/3-fun-0-',3, [{file, "/home/max-au/git/max-au/spg/src/spg.erl"}, {line,630}]},
```

The very same test passes before the change. That test is extra slow, taking 25 min on AMD 5950X, and it's quite sensitive to hardware performance (hence I never attempted to add it to the OTP).

I guess this change triggers another race condition when `leave_group` message races with `sync` and 'DOWN' from the monitor.

sverker:
> The change makes sense to me, but I am also interested in the source of redundant `discover` messages.

Well, consider you got an already established Erlang connection between nodes A and B. Node A has pg started but its `discover` message to B was not received by anyone. Then a pg process is started on B and it will broadcast `discover` to A. In this case A must reply with a potentially redundant `discover` message to B, as that is the only way to get a `sync` back from B.

That extra discover-after-sync could be avoided if the `sync` also acted as a `discover` when received from an undiscovered process. But that would need a protocol upgrade I think, as `sync` is only sent as reply from `discover` today.

To be continued on Monday...

max-au:
Now I see what you mean! Optimising a case when scopes are starting on nodes that are already running and connected. I did not think of it originally, because in our case `pg` starts before everything else. It is indeed possible to add `sync` response here: https://github.com/erlang/otp/blob/master/lib/kernel/src/pg.erl#L478 but as you already mentioned it requires a protocol upgrade (and the future-proof design of protocol upgrade).

max-au:
I updated `spg` to fully match `pg` code (except for OTP 21 compatibility that `pg` does not need to provide), so if you want to try debugging why this patch causes `pg` crashes, you're welcome to do so by running `rebar3 ct --suite spg_cluster_SUITE`.

PR #6488, opened 2022-11-19
Quote the atom `maybe` in the `ssl` application

belltoy:
Quote the `maybe` atoms and functions in the `ssl` application.

### DIFF
--- a/lib/ssl/src/ssl_gen_statem.erl
+++ b/lib/ssl/src/ssl_gen_statem.erl
@@ -518,7 +518,7 @@ initial_hello({call, From}, {start, Timeout},
     %% ServerHello is processed.
     RequestedVersion = tls_record:hello_version(Versions),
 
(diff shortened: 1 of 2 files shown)


u3s:
Can you agree on CLA please ? see automated comment above.

belltoy:
> Can you agree on CLA please ? see automated comment above.

👌🏼

u3s:
upon review of code it was proposed to rename `maybe` instead of quoting which looks ugly in source.
_Maybe_ `maybe` could become `maybe_do` instead of `'maybe'` :-) ?

belltoy:
> upon review of code it was proposed to rename `maybe` instead of quoting which looks ugly in source.
> _Maybe_ `maybe` could become `maybe_do` instead of `'maybe'` :-) ?

Sounds good. Updated and amended commit.

PR #6493, opened 2022-11-22
Add lists:combine (WIP, request for comments)

juhlig:
I had this idea in mind for a while, and decided to publish it here to see if there is any interest in it. It allows to traverse and process two lists at once (a bit like how `merge` does it) and combine them into a new term (a bit like `foldl` does).

The constituent parts are:
* a comparison function of arity 2, `Cmp`, which as usual returns `true` if the first argument is smaller than or equal to the second, otherwise `false` (optional, defaults to `=<`)
* a fold function of arity 3, `Fun` (more on that later)
* a starting accumulator (more on that later)
* two lists which must be sorted via `Cmp` before passing them to `combine`
* a 2-tuple of defaults used for padding, `{PadA, PadB}` (more on that later)

The lists are traversed in parallel. With each iteration, the two head elements are compared via `Cmp`
* if both elements compare equal, they are both passed as arguments to `Fun`, along with the accumulator. Both lists are then advanced.
* if the elements do not compare equal, the smaller one is passed to `Fun` while the respective other is filled with the respective element from the padding tuple, plus the current accumulator. Only the list from which the (smaller) element was taken is advanced, the other stays.

The result of `Fn` becomes the new accumulator.

`combine` itself returns the value of the final accumulator.

As an example, a call like this...
```erlang
lists:combine(
    fun(X, Y, Acc) -> [{X, Y}|Acc] end, % the fold function
    [], % the starting accumulator
    [a, b, c, e], % list 1, already sorted
    [b, d, e], % list 2, already sorted
    {x, y} % defaults for padding
).
```
... will traverse the lists like this...
```text
a      --> Fun(a, y, [])                                  --> [{a, y}]
b b    --> Fun(b, b, [{a, y}])                            --> [{b, b}, {a, y}]
c      --> Fun(c, y, [{b, b}, {a, y}])                    --> [{c, y}, {b, b}, {a, y}]
  d    --> Fun(x, d, [{c, y}, {b, b}, {a, y}])            --> [{x, d}, {c, y}, {b, b}, {a, y}]
e e    --> Fun(e, e, [{x, d}, {c, y}, {b, b}, {a, y}])    --> [{e, e}, {x, d}, {c, y}, {b, b}, {a, y}]
```
and finally return `[{e, e}, {x, d}, {c, y}, {b, b}, {a, y}]`.

<hr />

The padding thing is debatable I guess. In a nutshell, it makes it possible to have a nicer fold function by pushing it to the user how he wants missing elements in one of the lists represented, or filled up with defaults:
```erlang
Absent=make_ref(),
lists:combine(
    fun(A, Absent, Acc) -> ...; (Absent, B, Acc) -> ...; (A, B, Acc) -> ... end,
    [],
    [a, b, c, e],
    [b, d, e],
    {Absent, Absent}
).
```

Without it, missing elements would have to be something like `undefined`, and the present element wrapped in a tuple like `{value, ...}`:

```erlang
lists:combine(
    fun({value, A}, undefined, Acc) -> ...; (undefined, {value, B}, Acc) -> ...; ({value, A}, {value, B}, Acc) -> ... end,
    [],
    [a, b, c, e],
    [b, d, e]
).
```

<hr />

For a use case, I remember some work I did for `ranch` in the past. For context, a `ranch` listener can be configured with alarms, to call a given function when the number of active connections exceeds a given treshold, and call it again if that treshold is still exceeded after a given cooldown period; it it possible to have multiple alarms, for which reason they are named. It is possible to change the alarms setup by providing a new one, on a running listener. In that case...
* alarms in the old setup which are not present in the new one need their cooldown timers cancelled
* alarms in the new setup with are not present in the old one will just be added to the setup
* alarms existing in both the old and new setups may need to be updated, ie existing cooldown timers may have to be cancelled and restarted (taking into account the already elapsed time)

Back then, this was done manually by essentially doing what `lists:combine` is proposed for.

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -27,7 +27,8 @@
 
 %% Miscellaneous list functions that don't take funs as
 %% arguments. Please keep in alphabetical order.
--export([append/1, append/2, concat/1,
+-export([append/1, append/2, combine/5,
+         combine/6, concat/1,
          delete/2, droplast/1, duplicate/2,
          enumerate/1, enumerate/2,
          flatlength/1, flatten/1, flatten/2,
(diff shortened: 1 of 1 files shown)


josevalim:
Not sure if you are looking for general interest, but my initial impression is that this is too specific. If I run into a similar problem, it is most likely that I would implement the algorithm myself than remembering that `lists:combine` exists or trying to make it fit within the constraints of the proposed function. :)

juhlig:
> Not sure if you are looking for general interest

Depends on what you mean by "general" ;)

I have been implementing this thing several times in projects in my old job, with only slight differences (the way the lists are ordered and of course what is done if one or the other or both are found the lists), but the overall approach of stepping through both lists and advancing either the one or the other or both. So I wondered if that may be a thing that other people are implementing over and over again and how it could be abstracted.

So I guess what I'm asking is "is this useful to someone", or rather useful to enough people that would warrant putting it in the `lists` module. It is a draft/WIP, too, meant to illustrate the general idea, so open for changes and suggestions of any kind.

Maria-12648430:
This would probably be clearer (and arguably make more sense) in the context of maps. With maps, there would be no need for a comparison/ordering function. Also, the use case you gave for `ranch` could be even better with that, because the alarms are really stored in a map to begin with. To use the proposed `lists:combine` there, you still have to convert the old and new maps to lists, sort them, then `combine`, and in the end convert the result back to a map. If you had `maps:combine`, all the conversion and sorting would not be needed.

juhlig:
Hey @Maria-12648430, welcome back :hugs:

> This would probably be clearer (and arguably make more sense) in the context of maps. With maps, there would be no need for a comparison/ordering function. Also, the use case you gave for `ranch` could be even better with that, because the alarms are really stored in a map to begin with. To use the proposed `lists:combine` there, you still have to convert the old and new maps to lists, sort them, then `combine`, and in the end convert the result back to a map. If you had `maps:combine`, all the conversion and sorting would not be needed.

Hm, you do have a point, duly taken :sweat_smile:

Maria-12648430:
If this PR gets accepted, I would want to try an implementation for maps :smiley_cat:

bjorng:
I agree with @josevalim that the `combine` function is too specific. If I encountered a real-world problem where it could applied, I would probably fail to recognize that it could be used and I would still implement my own solution. So I don't think it is useful to enough people to warrant including it in the `lists` module.

For your use case, another way to handle it is to use [sofs:symmetric_partition/2](https://www.erlang.org/doc/man/sofs.html#symmetric_partition-2). It will give you three sets of values, one with timers to be cancelled, one with the timers to be added, and one with timers that need to be updated.

Here is an example to show how to use it:

```
1> A = sofs:set([a,b,c,e]),
   B = sofs:set([b,d,e]),
   {X,Y,Z} = sofs:symmetric_partition(A, B),
   {sofs:to_external(X),sofs:to_external(Y),sofs:to_external(Z)}.
{[a,c],[b,e],[d]}
```

juhlig:
> I agree with @josevalim that the `combine` function is too specific. If I encountered a real-world problem where it could applied, I would probably fail to recognize that it could be used and I would still implement my own solution. So I don't think it is useful to enough people to warrant including it in the `lists` module.

Ok. Nobody else expressed any interest, so that kinda proves it further =^^= So I'm closing this PR.

> For your use case, another way to handle it is to use [sofs:symmetric_partition/2](https://www.erlang.org/doc/man/sofs.html#symmetric_partition-2). It will give you three sets of values, one with timers to be cancelled, one with the timers to be added, and one with timers that need to be updated.

Thanks :) I have never used `sofs` before, but sounds interesting, I will have a look.

Maria-12648430:
> Nobody else expressed any interest

Hey, _I_ was interested! 😧

juhlig:
> Hey, _I_ was interested! :anguished:

Only because it would have given you an opportunity to pull off the same with `maps` :grin:

Maria-12648430:
> Only because it would have given you an opportunity to pull off the same with `maps` :grin:

You know me too well :roll_eyes:

Issue #6448, opened 2022-11-07
lookup_tracer_nif crashes when tracer = erts_tracer_nil


zzydxm:
**Describe the bug**
We occasionally see this back trace in our coredumps:
```
#0  0x0000000000728c07 in lookup_tracer_nif (tracer=59) at beam/erl_trace.c:2636
#1  0x0000000000734699 in call_enabled_tracer (t_p_id=19756849570803, tag=43083, topt=TRACE_FUN_ENABLED, tnif_ret=0x0, tracer=59) at beam/erl_trace.c:2845
#2  is_tracer_enabled (tag=43083, topt=TRACE_FUN_ENABLED, tnif_ret=0x0, t_p=0x7fc406851da8, c_p_locks=1, c_p=0x7fc406851da8) at beam/erl_trace.c:2889
#3  erts_is_tracer_proc_enabled (c_p=c_p@entry=0x7fc406851da8, c_p_locks=c_p_locks@entry=1, t_p=t_p@entry=0x7fc406851da8) at beam/erl_trace.c:2950
#4  0x00000000006035d8 in trace_schedule_in (p=p@entry=0x7fc406851da8, state=687874090) at beam/erl_process.c:10246
#5  0x0000000000620579 in erts_schedule (esdp=<optimized out>, esdp@entry=0x7fc4d62346c0, p=0x7fc406851da8, p@entry=0x7fc405dd5a78, calls=0, calls@entry=1) at beam/erl_process.c:10051
#6  0x00000000006a866e in erts_dirty_process_main (esdp=esdp@entry=0x7fc4d62346c0) at beam/beam_common.c:174
#7  0x00000000006066f6 in sched_dirty_io_thread_func (vesdp=0x7fc4d62346c0) at beam/erl_process.c:8755
#8  0x00000000009e1518 in ethr_init (id=0x7fff792b3230) at pthread/ethread.c:280
```

It happens that ```tracer = 59 = erts_tracer_nil = []``` in this case, but ```lookup_tracer_nif``` is trying to visit ```((Eterm*)tracer - TAG_PRIMARY_LIST)[0]``` which will surely crash.

I checked the code that ```erts_tracer_nil``` should be a valid value on ```tracer```, so I think a ```if(tracer == erts_tracer_nil)``` should be added in ```lookup_tracer_nif``` and return null if it does equal.

**To Reproduce**
Although this happens not rarely in our code base, I don't know how to reproduce this reliably due to lack of knowledge in erl_trace

[...]

PR #6495, opened 2022-11-22
avoid crash when tracer = erts_tracer_nil

zzydxm:
This closes https://github.com/erlang/otp/issues/6448


rickard-green:
Thanks! I rebased it to `maint`, so it will be part of OTP 25.3

Issue #5585, opened 2022-01-05
erlang:display/1 errored output for large maps


pehuen-rodriguez:
**Describe the bug**
When given a map of size > 32, `erlang:display/1` will print out the map incorrectly formatted and including some garbage. `erlang:display/1` implementation for [large maps](https://www.erlang.org/doc/efficiency_guide/maps.html#how-large-maps-are-implemented) seems to be missing or incorrect.

**To Reproduce**
`35>` shows a **working** example of 32 positions. `36>` is how to reproduce it: a Map with size 33
```
35> erlang:display(maps:from_list([{K, v} || K <- lists:seq(1, 32)])).
#{1=>v,2=>v,3=>v,4=>v,5=>v,6=>v,7=>v,8=>v,9=>v,10=>v,11=>v,12=>v,13=>v,14=>v,15=>v,16=>v,17=>v,18=>v,19=>v,20=>v,21=>v,22=>v,23=>v,24=>v,25=>v,26=>v,27=>v,28=>v,29=>v,30=>v,31=>v,32=>v}
true
36> erlang:display(maps:from_list([{K, v} || K <- lists:seq(1, 33)])).
#<fffb>{<1201>{[16|v],[18|v],[5|v]},<4400>{[19|v],[27|v]},<80>{<4080>{[10|v],[24|v]}},<14>{[4|v],[21|v]},<240>{[22|v],[17|v]},<2a>{[3|v],[8|v],[1|v]},<8512>{[7|v],[2|v],[14|v],[15|v],[20|v]},[13|v],[6|v],<1400>{[28|v],[25|v]},<212>{[32|v],[9|v],[11|v]},[31|v],[26|v],<2040>{[30|v],<8020>{[29|v],[23|v]}},<8200>{[12|v],[33|v]}}
true
```

**Expected behavior**
Output should be a correctly shaped map.

**Affected versions**
OTP 23~24. Elder versions: not sure

garazdawi:
That output is the internal representation of large maps, so it is correct. I believe it was done this way for large maps because when debugging the implementation of large maps it was very useful to know the internal details of the data structure.

I'm unsure how easy it is to fix the printing logic to print a normal map for large maps as the entire implementation for `erlang:display` has to be in C.

pehuen-rodriguez:
Thank you @garazdawi for the clarification! So it's not a bug but an inconsistency and a counter-intuitive result, you'd say?

[...]

PR #6497, opened 2022-11-22
[erts] Display large maps similarly to small maps

lmth:
The formatting of large maps done by erlang:display/1 used to reveal interesting information about the map internals, but was difficult to read, compared to the formatting of small maps. This commit aims at a similar (but still not sorted) display of large maps as for small maps.

This is important to those of us who regularly try to understand the state of our system from erlang:process_info(...,backtrace) or erlang:system_info(procs).

Pull request submitted based on issue #5585 discussion.

### DIFF
--- a/erts/emulator/beam/erl_printf_term.c
+++ b/erts/emulator/beam/erl_printf_term.c
@@ -726,40 +726,66 @@ print_term(fmtfn_t fn, void* arg, Eterm obj, long *dcount) {
                 switch (MAP_HEADER_TYPE(*head)) {
                 case MAP_HEADER_TAG_HAMT_HEAD_ARRAY:
                 case MAP_HEADER_TAG_HAMT_HEAD_BITMAP:
-                    PRINT_STRING(res, fn, arg, "#<");
-                    PRINT_UWORD(res, fn, arg, 'x', 0, 1, mapval);
-                    PRINT_STRING(res, fn, arg, ">{");
-                    WSTACK_PUSH(s,PRT_CLOSE_TUPLE);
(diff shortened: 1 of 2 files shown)


bjorng (changes requested):
Thanks for you pull request.

Your implementation of map printing looks good.

However, I think that the test case should be more challenging, testing nested maps, and that the maps are printed correctly. Therefore, I've updated the test case and pushed it to your branch. I've also made sure that indentation in the C code is done using spaces instead of tabs.

If you agree to those changes, please squash all four commits down to a single commit.

Also, since this is not a bug fix, I think we should not release it in a maintenance release. Please rebase the branch on `master` and retarget the pull request to `master`.

lmth:
Hi, @bjorng !

I obviously like the improvements you made to the branch, and have squashed and rebased.

/Magnus T

bjorng:

bjorng (approved):

bjorng:
My suggested test case failed on runtime system using the old non-JIT BEAM emulator because of subtle differences in how the stack is managed.

I turns out that there is much easier way to get back what `erlang:display/1` would print. It's an undocumented function called `erts_debug:display/1`. I have updated test case and force-pushed.

lmth:
Hi @bjorng,

I see - I was looking for some way of capturing the output of display but failed and thought I had to resort to a process stack dump of some sort. This is more to-the-point.

Do I have to do anything?

/Magnus T

bjorng:
@lmth You don't have to do anything.

bjorng:
Thanks!

PR #6500, opened 2022-11-23
Kuba/ssl/add trace profiles/otp 18312

u3s:


KennethL:
I think it is ok with init:get_argument to fetch the name of the trace file

Issue #3264, opened 2018-04-08
ERL-608: Better time handling (grabbing native time unit/ timer.tc variant/alternative)


OTP-Maintainer:
Original reporter: `pragtob`
Affected version: `Not Specified`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-608

---

```
(first thanks for Erlang et. al. <3)

 I have a suggestion for making time measurements better/more accurate. The normal built-in way to measure function execution time is [timer.tc/n|http://erlang.org/doc/man/timer.html#tc-1] but as described in its documentation and implemented it [always returns|https://github.com/erlang/otp/blob/master/lib/stdlib/src/timer.erl#L164-L169].

However, for instance on my Linux system the time measuring according to `erts_internal:time_unit/0` is nano seconds so precision is given up.

One could just reimplement timer.tc but without knowing what time unit monotonic_time returns it's hard as usage of erts_internal functions is clearly discouraged. Making the "native" time unit part of the public API would probably be the easiest fix. 

(There are workarounds like converting the monotonic time returned to another unit until you figure out the unit where it doesn't change).

Another idea would to offer an alternative to timer.tc that doesn't only return the time but the (native) unit it was measured in as well. So instead of {1000, result} the return value might be {{1000, microscond}, result}.

That'd be great and I think would help people who want to casually measure execution times (for logging purposes) or do more advanced things with them.

(I maintain a benchmarking library on the BEAM and stumbled across this as the native unit on Windows seems to be milliseconds which was a surprise to us)

Sorry if I didn't fill out some fields I should have.
```

rickard-green:
Implemented by #6507 which has been merged into master (to be released in OTP 26)

PR #6507, opened 2022-11-24
stdlib/timer: better resolution control

jdamanalo:
- added a final `erlang:time_unit()` argument to `tc/[1,2,3]`

Tackles #3264. Primarily for systems with worse resolution than a microsecond.


rickard-green (approved):

rickard-green:
Thanks!

PR #6511, opened 2022-11-25
handle DB2 32Bit null values on data encoding

esterlus:
The 64Bit DB2 ODBC driver reports null values in 32Bit fashion instead of 64 Bit.

This is a minimal change to always check for 32Bit null value as well as the configured null value in `SQL_NULL_DATA` when receiving data in ODBC.

Tested with DB2 `v10.5` and DB2 ODBC driver `v10.5fp11_linuxx64_odbc_cli` and OTP `v24.3.4.6`.

If you consider merging this, I would also be happy to get it into master.

### DIFF
--- a/lib/odbc/c_src/odbcserver.c
+++ b/lib/odbc/c_src/odbcserver.c
@@ -1215,7 +1215,9 @@ static db_result_msg encode_out_params(db_state *state,
             }
             column = params[i];
             if (column.type.len == 0 ||
-                column.type.strlen_or_indptr == SQL_NULL_DATA) {
+                column.type.strlen_or_indptr == SQL_NULL_DATA ||
+                // DB2 64Bit Driver uses 32Bit null
+                column.type.strlen_or_indptr == 0xffffffff) {
                 ei_x_encode_atom(&dynamic_buffer(state), "null");
             } else {
                 void* values = retrive_param_values(&column);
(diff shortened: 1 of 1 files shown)


IngelaAndin:
Do you mean that you want this to be fixed for OTP-24?  The normal case is to target PR for maint or master, although if it is a serious bug that we want to patch as soon as possible it would require the branch to be based on a release tag of the maint-track, and not on maint.  Everything fixed on maint should be merged also into master unless the master branch has diverged so that the fix is not relevant. This would not be a candidate for making an OTP-24 patch. We could possibly piggyback it for another OTP-24 patch if the branch would be based on an OTP-24 release tag (of green version https://erlang.org/download/otp_versions_tree.html)  but targeted for maint in the PR. That would probably be easy for odbc which has been in maintenance mode and this should not cause any merge conflicts.

esterlus:
Changed target branch to maint as recommended.
Currently using this with OTP-24. Would be happy if this is also applied to versions upwards of that.

IngelaAndin:
Alas, it seems our test environment for ODBC has some issues and I have not been able to verify that the change does not break anything. While waiting for it to be fixed it would be great if you could give us some more background information and maybe some reference to DB2 documentation to help us verify that your fix is the logical way of solving the problem you are facing. ODBC application is very much in maintenance mode and the code has not been actively worked on in a long time, so what seems obvious to you is not necessarily obvious to us.

esterlus:
Just double checked the driver documentation and noticed that there is different libraries available.
The problem here is that with the default library `libdb2.so`, `SQL_NULL_DATA` does not hold the correct value to determine nulls. This leads to weird overflow/conversion issues on numbers.

However the driver also comes with a library specific for odbc: `libdb2o.so`. (Notice the additional **o**)
When configuring odbc to use this library everything works as expected.

Thank you for your patience and helping me to figure this out.

PR #6513, opened 2022-11-25
Reference to split/3 function corrected

jurmc:
There is minor mistake in documentation of the string module. After docs generation it is visible as follows (split/2 is repeated twice and links to incorrect function):

'''
Grapheme clusters for codepoints of class prepend and non-modern (or decomposed) Hangul is not handled for performance reasons in find/3, replace/3, split/2, split/2 and trim/3.
'''


u3s:
can you sign the CLA please?

jurmc:
> can you sign the CLA please?

Done.

PR #6522, opened 2022-11-29
[erts, kernel] reuseaddr/reuseport/exclusiveaddruse support/fixes

rickard-green:
* Introduce socket option `reuseport` which *may or may not* have load balancing.

* Introduce socket option `reuseport_lb` *with* load balancing.

* Introduce socket option `exclusiveaddruse`. This socket option is Windows specific and will silently be ignored on other systems.

* Change behavior on Windows so that `SO_REUSEADDR` is only set if both `reusaddr` and `reuseport` have been set. This since `SO_REUSEADDR` on Windows behaves as BSD behaves if both `SO_REUSEADDR` and `SO_REUSEPORT` have been set.


okeuday:
@rickard-green There is a problem with using `SO_REUSEPORT` regarding its meaning on Linux and FreeBSD.  FreeBSD had `SO_REUSEPORT` in the past, before it was present on Linux (mainly for UDP multicast usage probably) and Linux added the `SO_REUSEPORT` with extra load-balancing functionality.  After that, FreeBSD added `SO_REUSEPORT_LB` to be equivalent to the `SO_REUSEPORT` option on Linux.  Other BSDs also have the older meaning of `SO_REUSEPORT`, so all BSDs would need to have support for the `SO_REUSEPORT_LB` option to have the equivalent functionality to `SO_REUSEPORT` on Linux.  So, you likely need to support `reuseport` and `reuseportlb` (or some other name) as separate options to avoid confusion.

rickard-green:
@okeuday Yes, your right. We probably need to expose the load balancing option as well.

rickard-green:
@okeuday 

Added an `reuseport_lb` option as well (using FreeBSD's name for it).

`reuseport` is documented as *may or may not* be load balancing while `reuseport_lb` is documented as *with* load balancing. On Linux they will both set `SO_REUSEPORT`.

okeuday:
@rickard-green You may want to consider `reuseport` on Linux causing an error, due to lack of support, to ensure it represents the BSD-meaning of `SO_REUSEPORT` because the use-case of the BSD `SO_REUSEPORT` should be very different.  The potential for the use of the BSD `SO_REUSEPORT` in Erlang is interesting too, even though load-balancing is not provided, because it could be used to have a backup TCP connection with a backup Erlang process to own the backup connection (it would be only used if the other Erlang process dies).  However, doing something like that, would be clearly different from needing the BSD `SO_REUSEPORT_LB` functionality which is what people would normally use for ensuring TCP traffic is evenly distributed among Erlang processes.  So, the `reuseport_lb` option use is likely more common to ensure efficient load-balancing can occur.

lib/kernel/doc/src/inet.xml
```diff
@@ -1566,18 +1582,88 @@ setcap cap_sys_admin,cap_sys_ptrace,cap_dac_read_search+epi beam.smp</code>
...
+          <item>
+            <p>
+              Allows or disallows reuse of local address. By default, reuse
+              is disallowed.
+            </p>
+            <note>
+              <p>
+                On windows <c>{reuseaddr, true}</c> will have no effect unless
+                also <seeerl marker="#option-reuseport"><c>{reuseport,
+                true}</c></seeerl> is set. If both are set, the
```
bmk, line 1595: ...is also set...

lib/kernel/test/inet_sockopt_SUITE.erl
```diff
@@ -914,6 +924,46 @@ all_connect_options() ->
     ].
 
 
+%% Mandatory an a lot of system other than those listed below. Please add more...
```
bmk, line 927: Mandatory *on* a lot of...?

bmk (approved):

RaimoNiskanen (approved):

bjorng (approved):

lib/kernel/test/inet_sockopt_SUITE.erl
rickard-green: yes fixed

lib/kernel/doc/src/inet.xml
rickard-green: there is in also before the option which I guess is correct

rickard-green:
@okeuday We've decided not to error out when reuseport without load balancing is not available in this API. The socket API is expected to get a more fine-grained option API.

Issue #4733, opened 2021-04-13
erl_syntax_lib:annotate_bindings for named funs


garazdawi:
**Describe the bug**

When calling `erl_syntax_lib:annotate_bindings` on a named fun the arguments and 'F' are not annotated as bound.

**To Reproduce**

```erlang
test() ->
  Fn = {named_fun,{6,5},
           'F',
           [{clause,{6,9},
                    [{var,{6,11},'Test'}],
                    [],
                    [{var,{7,13},'Test'}]}]},
  erl_syntax_lib:annotate_bindings(Fn, []).
```

returns

```erlang
{tree,named_fun_expr,
    %% Neither 'F' nor 'Test' should be free here, they are both bound, or possibly should not even be part of the attribute.
    {attr,{6,5},[{env,[]},{bound,[]},{free,['F','Test']}],none},
    {named_fun_expr,
        {tree,variable,
            %% 'F' should be bound here
            {attr,{6,5},[{env,[]},{bound,[]},{free,['F']}],none},
            'F'},
        [{tree,clause,
             %% 'Test' should be bound here
             {attr,{6,9},[{env,[]},{bound,[]},{free,['Test']}],none},
             {clause,
                 [{wrapper,variable,
                      {attr,{6,11},[{env,[]},{bound,[]},{free,['Test']}],none},
                      {var,{6,11},'Test'}}],
                 none,
                 [{wrapper,variable,
                      {attr,{7,13},[{env,[]},{bound,[]},{free,['Test']}],none},
                      {var,{7,13},'Test'}}]}}]}}
```

**Affected versions**
Any Erlang/OTP version after named funs were introduced.

**Additional context**
This was discovered when using erlang-ls. See https://github.com/erlang-ls/erlang_ls/issues/982

MarkoMin:
@garazdawi This is closed via https://github.com/erlang/otp/pull/6523

garazdawi:
Thanks!

PR #6523, opened 2022-11-30
named_fun_expr annotate bindings fix

MarkoMin:
Attempt to fix the behavior of `erl_syntax_lib:annotate_bindings/1-2` described in #4733 (labeled with "help wanted").
Issue is also mentioned in `erlang_ls` repo, so potential fix may also allow better UX while using it.

Description:
My plan was to include a case in `vann/2` function to handle `named_fun_expr`nodes correctly. Annotation process is almost identical to the one of `fun_expr`, obviously with the difference of handling `named_fun_expr`'s name. Name of the function is set into the environment for each of the clause patterns, because patterns can shadow the function name.

I included one concrete example for the testing purpose.
GitHub actions passed successfully on the fork repo.


jhogberg:

PR #6528, opened 2022-12-02
kernel: fix inet_backend option in connect with sockaddr

RoadRunnr:


RoadRunnr:
The problem exists since at least OTP-24.3 (but might older).

ebengt (approved):
Very good find.

bmk (approved):

bmk:
I would have liked this to be based on 24.3.4, to make it possible to make both 24 and 25 patches.
But I am currently unsure exactly how we (Erlang/OTP) handle this. I will check.

/BMK

bmk:
Never mind. I will take care of the 24 and 25 patches.

bmk:
Prepared for OTP 24 and 25 patches.

IngelaAndin:
Solved in OTP-25.2.1  and OTP 24.3.4.7

PR #6530, opened 2022-12-03
kernel: os_SUITE: fix close_stdin test

MatthewGentoo:
* os_SUITE_data/my_fds.c: `read(stdin, buff, 1)` is invalid because stdin is a FILE* and not an fd, will become an error by default in clang 16.
* os_SUITE.erl: my_fds program now returns 0 bytes read instead of -1 for error, update assertion.

***

```
my_fds.c:11:20: error: incompatible pointer to integer conversion passing 'FILE *' (aka 'struct _IO_FILE *') to parameter of type 'int' [-Wint-conversion]
    int res = read(stdin, buff, 1);
                   ^~~~~
```


sverker:
Merged to maint and master.
@MatthewGentoo Thanks for the fix.

PR #6536, opened 2022-12-06
Add Tau constant to math module

hauleth:
https://tauday.com/tau-manifesto

Tau constant is already present and available in some prominent languages and libraries

- [.NET](https://learn.microsoft.com/en-us/dotnet/api/system.math.tau?view=net-7.0)
- [Rust](https://doc.rust-lang.org/std/f64/consts/constant.TAU.html)
- [Java](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Math.html#TAU)
- [Python](https://docs.python.org/3/library/math.html#math.tau)
- [Boost](https://www.boost.org/doc/libs/1_38_0/boost/units/systems/si/codata/tau_constants.hpp)

### DIFF
--- a/lib/stdlib/doc/src/math.xml
+++ b/lib/stdlib/doc/src/math.xml
@@ -102,9 +102,10 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
 
     <func>
       <name name="pi" arity="0" since=""/>
-      <fsummary>A useful number.</fsummary>
+      <fsummary>Ratio of the circumference of a circle to its diameter.</fsummary>
       <desc>
-        <p>A useful number.</p>
+          <p>Ratio of the circumference of a circle to its diameter.</p>
+          <p>Floating point approximation of mathematical constant pi.</p>
       </desc>
     </func>
 
--- a/lib/stdlib/doc/src/math.xml
+++ b/lib/stdlib/doc/src/math.xml
@@ -109,6 +109,16 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
       </desc>
     </func>
 
+    <func>
+      <name name="tau" arity="0" since="OTP @OTP-18361@"/>
+      <fsummary>Ratio of the circumference of a circle to its radius.</fsummary>
+      <desc>
+        <p>Ratio of the circumference of a circle to its radius.</p>
+        <p>This constant is equivalent to a full turn when described in radians.</p>
+        <p>The same as <c>2 * pi()</c>.</p>
+      </desc>
+    </func>
+
   </funcs>
 
   <section>
--- a/lib/stdlib/src/math.erl
+++ b/lib/stdlib/src/math.erl
@@ -19,7 +19,7 @@
 %%
 -module(math).
 
--export([pi/0]).
+-export([pi/0,tau/0]).
 
 %%% BIFs
 
@@ -154,5 +154,7 @@ tanh(_) ->
 %%% End of BIFs
 
 -spec pi() -> float().
-
 pi() -> 3.1415926535897932.
+
+-spec tau() -> float().
+tau() -> 6.2831853071795864.
--- a/lib/stdlib/test/math_SUITE.erl
+++ b/lib/stdlib/test/math_SUITE.erl
@@ -28,7 +28,7 @@
 	 init_per_testcase/2, end_per_testcase/2]).
 
 %% Test cases
--export([floor_ceil/1, error_info/1]).
+-export([floor_ceil/1, error_info/1, constants/1]).
 
 
 suite() ->
@@ -36,7 +36,7 @@ suite() ->
      {timetrap,{minutes,1}}].
 
 all() ->
-    [floor_ceil, error_info].
+    [floor_ceil, error_info, constants].
 
 groups() ->
     [].
@@ -58,6 +58,11 @@ init_per_testcase(_Case, Config) ->
     Config.
 
 end_per_testcase(_Case, _Config) ->
+    ok.
+
+constants(_Config) ->
+    3.1415926535897932 = math:pi(),
+    6.2831853071795864 = math:tau(),
     ok.
 
 floor_ceil(_Config) ->


lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
 
     <func>
       <name name="pi" arity="0" since=""/>
-      <fsummary>A useful number.</fsummary>
+      <fsummary>Ratio between circumfence and diameter of the circle.</fsummary>
```
Maria-12648430, line 105:
```suggestion
      <fsummary>Ratio of circumfence to diameter of a circle.</fsummary>
```
(actually, to be _very_ pedantic, the ratio of circumference to diameter of a circle ***on a plane*** :grin: (just joking))

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
...
       <desc>
-        <p>A useful number.</p>
+          <p>Ratio between circumfence and diameter of the circle.</p>
+          <p>Floating point approximation of mathematical constant pi.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="tau" arity="0" since=""/>
+      <fsummary>Ratio between circumfence and radius of the circle aka full turn.</fsummary>
```
Maria-12648430, line 114:
```suggestion
      <fsummary>Ratio of circumfence to radius of a circle.</fsummary>
```
Do not use abbreviations like "aka" in the documentation. Actually, I would leave that part out entirely here, it is not necessary in the summary.

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
...
+      <desc>
+        <p>Ratio between circumfence and radius of the circle.</p>
+        <p>This constant is equivalent to full turn when described in radians.</p>
+        <p>The same as <c>2</c> * <c>pi</c>.</p>
```
Maria-12648430, line 118:
```suggestion
        <p>Ratio of circumfence to radius of a circle.</p>
        <p>This constant is equivalent to full turn when described in radians.</p>
        <p>The same as <c>2 * pi()</c>.</p>
```

lib/stdlib/src/math.erl
```diff
@@ -154,5 +154,7 @@ tanh(_) ->
 %%% End of BIFs
 
 -spec pi() -> float().
-
 pi() -> 3.1415926535897932.
+
+-spec tau() -> float().
+tau() -> 6.2831853071795864.
```
Maria-12648430, line 160: Just a suggestion. Instead of writing out the full (yeah, I know) sequence of digits, I would put the value for pi in a constant and define both `pi/0` and `tau/0` in terms of it:
```erlang
-define(PI, 3.1415926535897932).

pi() -> ?PI.

tau() -> 2 * ?PI.
```
You know, in case some day god decides to [make math things simpler](https://www.smbc-comics.com/comic/irrational-3) :smile_cat:

Maria-12648430 (changes requested):
Not sure how useful it _really_ is to have this, but well... :wink:

lib/stdlib/src/math.erl
hauleth: I didn't wanted to do it to avoid loosing precision when compiling this module on different machines. We can make math simpler, but IEEE 754 will always stand in our way.

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
 
     <func>
       <name name="pi" arity="0" since=""/>
-      <fsummary>A useful number.</fsummary>
+      <fsummary>Ratio of circumfence to diameter of a circle.</fsummary>
```
paulo-ferraz-oliveira, line 105:
```suggestion
      <fsummary>Ratio of the circumference of a circle to its diameter.</fsummary>
```

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
...
       <desc>
-        <p>A useful number.</p>
+          <p>Ratio between circumfence and diameter of the circle.</p>
```
paulo-ferraz-oliveira, line 107:
```suggestion
          <p>Ratio of the circumference of a circle to its diameter.</p>
```

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
...
+          <p>Floating point approximation of mathematical constant pi.</p>
+      </desc>
+    </func>
+
+    <func>
+      <name name="tau" arity="0" since=""/>
+      <fsummary>Ratio of circumfence to radius of a circle.</fsummary>
```
paulo-ferraz-oliveira, line 114:
```suggestion
      <fsummary>Ratio of the circumference of a circle to its radius.</fsummary>
```

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
...
+      <desc>
+        <p>Ratio of circumfence to radius of a circle.</p>
```
paulo-ferraz-oliveira, line 116:
```suggestion
        <p>Ratio of the circumference of a circle to its radius.</p>
```

lib/stdlib/doc/src/math.xml
```diff
@@ -102,9 +102,20 @@ erf(X) = 2/sqrt(pi)*integral from 0 to X of exp(-t*t) dt.</pre>
...
+        <p>This constant is equivalent to full turn when described in radians.</p>
```
paulo-ferraz-oliveira, line 117:
```suggestion
        <p>This constant is equivalent to a full turn when described in radians.</p>
```

lib/stdlib/doc/src/math.xml
paulo-ferraz-oliveira: At the same time, if you don't want to accept the suggestion, I point out that "circumfence" is a typo.

sverker:
Rebased into two commits. One for `tau` and one for `pi`.

Added test and removed all extra decimals which had no effect. Argue for that in separate PR if there is something I don't get.

sverker (approved):

PR #6545, opened 2022-12-09
ftp: fix types in ftp module

kikofernandez:
Fixes some type specs from the ftp module. These have been validated by eqWAlizer.

### DIFF
--- a/lib/ftp/src/ftp.erl
+++ b/lib/ftp/src/ftp.erl
@@ -21,8 +21,6 @@
 
 -module(ftp).
 
--behaviour(gen_server).
-
 -deprecated([{start_service, 1, "use ftp:open/2 instead"},
              {stop_service, 1,  "use ftp:close/1 instead"}]).
 
@@ -31,8 +29,6 @@
          stop/0,
          stop_service/1
         ]).
-
--export([start_link/1, start_link/2]).
 
 %%  API - Client interface
 -export([cd/2, close/1, delete/2, formaterror/1,
@@ -50,82 +46,7 @@
          append_chunk/2, append_chunk_end/1, append_chunk_start/2,
          info/1, latest_ctrl_response/1]).
 
-%% gen_server callbacks
--export([init/1, handle_call/3, handle_cast/2,
-         handle_info/2, terminate/2, code_change/3]).
-
 -include("ftp_internal.hrl").
-
-%% Constants used in internal state definition
--define(CONNECTION_TIMEOUT,  60*1000).
--define(DATA_ACCEPT_TIMEOUT, infinity).
--define(DEFAULT_MODE,        passive).
--define(PROGRESS_DEFAULT,    ignore).
--define(FTP_EXT_DEFAULT,     false).
-
-%% Internal Constants
--define(FTP_PORT, 21).
--define(FTPS_PORT, 990).
--define(FILE_BUFSIZE, 4096).
-
-
-%%%=========================================================================
-%%%  Data Types
-%%%=========================================================================
-
-%% Internal state
--record(state, {
-          csock   = undefined, % socket() - Control connection socket
-          dsock   = undefined, % socket() - Data connection socket
-          tls_options = undefined, % list()
-          verbose = false, % boolean()
-          ldir    = undefined, % string() - Current local directory
-          type    = ftp_server_default, % atom() - binary | ascii
-          chunk   = false, % boolean() - Receiving data chunks
-          mode    = ?DEFAULT_MODE, % passive | active
-          timeout = ?CONNECTION_TIMEOUT, % integer()
-          %% Data received so far on the data connection
-          data    = <<>>, % binary()
-          %% Data received so far on the control connection
-          %% {BinStream, AccLines}. If a binary sequence
-          %% ends with ?CR then keep it in the binary to
-          %% be able to detect if the next received byte is ?LF
-          %% and hence the end of the response is reached!
-          ctrl_data = {<<>>, [], start}, % {binary(), [bytes()], LineStatus}
-          %% pid() - Client pid (note not the same as "From")
-          latest_ctrl_response = "",
-          owner = undefined,
-          client = undefined, % "From" to be used in gen_server:reply/2
-          %% Function that activated a connection and maybe some
-          %% data needed further on.
-          caller = undefined, % term()
-          ipfamily, % inet | inet6 | inet6fb4
-          sockopts_ctrl = [],
-          sockopts_data_passive = [],
-          sockopts_data_active = [],
-          progress = ignore, % ignore | pid()
-          dtimeout = ?DATA_ACCEPT_TIMEOUT, % non_neg_integer() | infinity
-          tls_ctrl_session_reuse = false, % boolean()
-          tls_upgrading_data_connection = false,
-          ftp_extension = ?FTP_EXT_DEFAULT
-         }).
-
--record(recv_chunk_closing, {
-          dconn_closed = false,
-          pos_compl_received = false,
-          client_called_us = false
-         }).
-
-
--type shortage_reason()  :: 'etnospc' | 'epnospc'.
--type restriction_reason() :: 'epath' | 'efnamena' | 'elogin' | 'enotbinary'.
--type common_reason() ::  'econn' | 'eclosed' | term().
--type file_write_error_reason() :: term(). % See file:write for more info
-
--define(DBG(F,A), 'n/a').
-%%-define(DBG(F,A), io:format(F,A)).
-%%-define(DBG(F,A), ct:pal("~p:~p " ++ if is_list(F) -> F; is_atom(F) -> atom_to_list(F) end, [?MODULE,?LINE|A])).
-
 
 %%%=========================================================================
 %%%  API
@@ -138,23 +59,19 @@ start() ->
 %% ftp client processes should always be part of ftp supervisor tree.
 %% We consider it a bug that the "standalone" concept of inets was 
 %% not removed when ftp was broken out, and it is now fixed.
+-spec start_service(ServiceConfig) -> {ok, Pid} | {error, Reason} when
+      ServiceConfig :: [{Property, Value}],
+      Property :: proplists:property(),
+      Value :: term(),
+      Pid :: pid(),
+      Reason :: term().
 start_service(Options) ->
-    try
-        {ok, StartOptions} = start_options(Options),
-        case ftp_sup:start_child([[[{client, self()} | StartOptions], []]]) of
-            {ok, Pid} ->
-                call(Pid, {open, ip_comm, Options}, plain);
-            Error1 ->
-                Error1
-        end
-    catch
-        throw:Error2 ->
-            Error2
-    end.
+    ftp_internal:start_service(Options).
 
 stop() ->
     application:stop(ftp).
 
+-spec stop_service(Pid :: pid() | term()) -> ok.
 stop_service(Pid) ->
     close(Pid).
 
@@ -163,17 +80,11 @@ stop_service(Pid) ->
 %%%=========================================================================
 
 %%--------------------------------------------------------------------------
-%% open(HostOrOtpList, <Port>, <Flags>) -> {ok, Pid} | {error, ehost}
-%%        HostOrOtpList = string() | [{option_list, Options}]
-%%      Port = integer(),
-%%      Flags = [Flag],
-%%      Flag = verbose | debug | trace
-%%
 %% Description:  Start an ftp client and connect to a host.
 %%--------------------------------------------------------------------------
 
 -spec open(Host :: string() | inet:ip_address()) ->
-    {'ok', Pid :: pid()} | {'error', Reason :: 'ehost' | term()}.
+    {'ok', Pid :: pid()} | {'error', Reason :: term()}.
 
 %% <BACKWARD-COMPATIBILLITY>
 open({option_list, Options}) when is_list(Options) ->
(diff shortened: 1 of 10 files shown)


lib/ftp/src/ftp.erl
```diff
@@ -2087,9 +2090,9 @@ handle_caller(#state{caller = {transfer_data, {Cmd, Bin, RemoteFile}}} =
 %% Connect to FTP server at Host (default is TCP port 21)
 %% in order to establish a control connection.
 -spec setup_ctrl_connection(Host, Port, Timeout, State) -> Result when
-      Host    :: inet:socket_address() | inet:hostname(),
+      Host    :: inet:ip_address() | inet:hostname(),
       Port    :: inet:port_number(),
-      Timeout :: timeout(),
+      Timeout :: integer(),
```
michalmuskala, line 2095: Is this right?

`connect/5` supports `infinity` timeout, however this does not because of the case below. Was the intention to support `infinity` and implementation is wrong, or there was no expectation of supporting this.

kikofernandez: AFAIK, it is supposed to be an `non_neg_integer()`. 

From the usage of the record, the documentation, and the [code documentation](https://github.com/erlang/otp/pull/6545/files/9407e3d3402c28aa2d1ad3065a5f65384ffabae8#diff-bbc31375ed1562b0f69299e0b126ac627bfd216c1164f0c36a38d46cafb1c0f8L86), it is a `non_neg_integer()`.

It is probably not too hard to support `infinity`, which AFAIK, means that there is no timeout to consider :)
If someone using the FTP app really needs this, we can look into it :)

michalmuskala: Makes sense. I this case I think it does make sense to keep it just as an integer

kikofernandez: why `integer()` instead of `non_neg_integer()`? we know that a timeout must be a positive number, correct?
Just wondering :)

michalmuskala: Yeah. The more precise type makes sense here. As far as eqwalizer is concerned though, it doesn't distinguish various numeric types, see https://github.com/WhatsApp/eqwalizer/blob/main/docs/reference/types.md#numeric-types

kikofernandez: For traceability, I created a ticket [eqWA-11](https://github.com/WhatsApp/eqwalizer/issues/11) :)
I hope this is ok

IngelaAndin:
Branch name and comments are misleading as you also change inets files!

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -44,7 +44,7 @@
 
 -record(timers, 
         {
-          request_timers = [] :: [reference()],
+          request_timers = [] :: [reference() | {reference() | 'undefined', term()}],
```
IngelaAndin, line 47: Why do you quote the atom undefined?

kikofernandez:
Thanks for the feedback.

I am updating two modules and will close this PR and create the appropriate one with a better branch name.

I would like to get feedback regarding the changes, if we think that those guards make sense.
- If we add them, eqWAlizer is happy about the types because it can type check those functions. 
- If we do not have them, it would reject those functions. 

From my point of view:
- it makes sense to add them, to say that the function expects lists / values different from `undefined` or similar
- there will be a small overhead in those checks, but AFAIK they are almost for free (?)
- if there is a catch all clause in those functions, adding these guards changes the semantics of the program, as instead of crashing at runtime, the code will go through the catch-all function

kikofernandez:
I have remove the parts that dealt with the inets module to be in its own PR (#6661 )
A new review is welcome :)

lib/ftp/src/ftp.erl
```diff
@@ -75,39 +75,41 @@
...
+                               | { tcp | ssl, inet:socket() | ssl:socket() | ssl:sslsocket()},
+          dsock   = undefined :: undefined % Data connection socket
+                               | { tcp | ssl | lsock, inet:socket() | ssl:socket() | ssl:sslsocket() | gen_tcp:socket()},
+          tls_options = undefined :: undefined | [tuple()],
+          verbose = false  :: boolean(),
+          ldir    = undefined :: undefined | file:filename_all(), % Current local directory
+          type    = ftp_server_default :: atom(), % binary | ascii
+          chunk   = false :: boolean(), % Receiving data chunks
+          mode    = ?DEFAULT_MODE :: active | passive,
+          timeout = ?CONNECTION_TIMEOUT :: integer(),
```
IngelaAndin, line 88: I think it should be timeout as it is passed to gen_tcp:connect that expects a timeout()

lib/ftp/src/ftp.erl
```diff
@@ -1285,7 +1287,7 @@ handle_info({Trpt, Socket, Data},
 
 handle_info({Trpt, Socket, Data}, #state{dsock = {Trpt,Socket}, client = From,
                                         caller = recv_chunk}
-            = State) when Trpt==tcp;Trpt==ssl ->
+            = State) when (Trpt==tcp orelse Trpt==ssl) andalso (From =/= undefined) ->
```
IngelaAndin, line 1290: What value does this defensive approach add?  It is a bug if From is undefined here and now you will get a function clause on handle_info instead of gen_server:reply which is less informative for finding the bug!

IngelaAndin: The comment is relevant for several of the changes below!

kikofernandez: I completely agree but from the point of view of a static analysis tool, if `From` can be `undefined` because it is not excluded, so eqWAlizer and any other type checker will report this as an error, because it knows that `gen_server:reply` cannot accept an `undefined`. 

I think it depends on the approach that we want to take to make type checkers happy :)

kikofernandez: I am torn on this one. 

I agree that you do get the error on `gen_server:reply` before, and now you get it in the function clause.

At the same time, if there is an error with an input argument, maybe you want to crash early. 

In this case, it is obvious that the call to `gen_server:reply` crashes. However, if the `From` value were to travel to a bunch of other functions until reaching the `gen_server:reply` call, I do wonder: isn't it more useful to know the first place that should not have received the unexpected value? It seems like the closer you get to the root cause, the better?

lib/ftp/src/ftp.erl
```diff
@@ -54,6 +54,11 @@
 -export([init/1, handle_call/3, handle_cast/2,
          handle_info/2, terminate/2, code_change/3]).
 
+-export_type([common_reason/0
```
kikofernandez, line 57: This is important to notice in the review.
I was exporting these types. I think it makes sense for users to be able to use these error types.
If I cannot export them, I need to copy-paste the same definition of the type in all functions that use them (which is quite a few).

IngelaAndin: I think it is necessary to export all types that are input or output types for the API functions. @KennethL maybe you have thoughts about the name of the type?

kikofernandez: I think it would be ok to call them instead as follows, so that the format is the same for all of them :) 
```erlang
export_type([common_error_reason/0
             ,restriction_error_reason/0
             ,shortage_error_reason/0
             ,file_write_error_reason/0]).
```

lib/ftp/src/ftp.erl
IngelaAndin: I do not feel I like to make my code defensive to make a tool happy. I think it is better to let it crash whenever the parameter is used because then you immediately know which parameter has an unexpected value. Typically state machine/gen server function clause will match several values and you will first have to determine which value was it unhappy with and then you can start analyzing when did it become wrong. As it is the state that has an unexpected value the fault is somewhere else than in the function clause that failed in the defensive version. It will be in another previously executed function clause where the returned state was not properly populated.

lib/ftp/src/ftp.erl
IngelaAndin: Maybe there is a type in file that  we should reference?!

lib/ftp/src/ftp.erl
kikofernandez: I got rid of the `From =/= undefined` :+1:

lib/ftp/src/ftp.erl
IngelaAndin: Yes, I meant that the file module should have a type that we should use instead of term(), I did not remember which one it was without checking, so I suggested you check. I suppose you also deduced that badarg or terminated could be returned in this case.

kikofernandez: Yes, those can be returned from `file:write/2` and the comment about `file_write_error_reason/0` says that the type is that of `file:write/2`

KennethL: Why is it important for users to declare the exact type of the `Reason` in {error, `Reason`}? 
The `Reason` is not intended to be matched is it? It is just an informative code that can be translated to a string with `format_error(Reason)`.
By the way I think `format_error(atom())` takes an `atom()` as input.

If there still is need for an exported `Reason` type I could propose that it is only one with all the possible error atoms. I don't think it is important if a function only returns a subset of the atoms.
But please give me arguments to why I am wrong. Or maybe it is enough to specify `Reason` as `atom()`.

lib/ftp/src/ftp.erl
```diff
@@ -120,7 +127,7 @@
 -type shortage_reason()  :: 'etnospc' | 'epnospc'.
 -type restriction_reason() :: 'epath' | 'efnamena' | 'elogin' | 'enotbinary'.
 -type common_reason() ::  'econn' | 'eclosed' | term().
--type file_write_error_reason() :: term(). % See file:write for more info
+-type file_write_error_reason() :: file:posix() | badarg | terminated. % See file:write for more info
```
KennethL, line 130: Isn't `badarg` something we associate with an exception? 
A bad type of argument can result in an exception while a correct type of argument can result in `ok | {error, Reason}`?

lib/ftp/src/ftp.erl
```diff
@@ -54,6 +54,11 @@
 -export([init/1, handle_call/3, handle_cast/2,
          handle_info/2, terminate/2, code_change/3]).
 
+-export_type([common_reason/0
+             ,restriction_reason/0
+             ,shortage_reason/0
+             ,file_write_error_reason/0]).
+
```
KennethL, line 61: I do not like these groups of error atoms, I think it is overkill. `I vote for `atom()` or possibly export one type `ftp:error_codes()`

lib/ftp/src/ftp.erl
kikofernandez: I don't think there is right or wrong, it is more about managing drawbacks :) 
Here is my analysis and understanding:

### Why we should export types
- If we provide the `Reason`, it is easier for us developers of OTP to know what possible error reason we can return in each function. This helps in understanding that some functions cannot return certain errors, so when fixing bugs one can easily see which reasonable error types can appear.  
- The existing documentation already mentions these types, they are simply not exported.
- If we use a superset of atoms (all possible error reason atoms) or simply say `atom()`, we lose some error information that may or may not be useful, because we are way too general.

### Why we should not use those types
- The `Reason` is not to be pattern match'ed and we do not want users to statically know the reason for an error
- If all other APIs treat `Reason` as `term()` or `atom()`, maybe for consistency `ftp` should too.

### Not viable options
- We cannot declare the `Reason` types as opaques because we break backwards compatibility (users would not be able to pattern match on the error reason when they were able to do it before).

My reasoning is that being concrete about the return type may help us (OTP maintainers) in understanding what can go wrong with the functions and this could be useful information to keep around.

lib/ftp/src/ftp.erl
kikofernandez: in this case, since the previous code was referring to `file:write`, I simply placed those types in here

lib/ftp/src/ftp.erl
kikofernandez: For our users, I agree that maybe `atom()` is enough.
For ourselves, OTP maintainers, I think that being explicit about the possible error type could be useful for Dialyzer to rule out certain possible errors, like mistakes where we could return `{error, foo}`.

I am just not sure yet on how to handle them... One "solution" is that the public interface calls an internal API that has the precise type of error. In this way, the public interface says `{error, atom()}` while the private interface says `{error, common_reason()}`. An example of what I mean follows:

```erlang
-spec user(Pid  :: pid(),
           User :: string(),
           Pass :: string()) ->
    'ok' | {'error', Reason :: atom()}.

user(Pid, User, Pass) -> user_(Pid, User, Pass).
    

-spec user_(Pid  :: pid(),
           User :: string(),
           Pass :: string()) ->
    'ok' | {'error', Reason :: 'euser' | common_reason()}.

user_(_Pid, _User, _Pass) ->
    case {is_name_sane(User), is_name_sane(Pass)} of
        {true, true} ->
            call(Pid, {user, User, Pass}, atom);
        _ ->
            {error, euser}
    end.
```

This allows us to hide the internal return type at the expense of some more lines of information.

IngelaAndin: According to what is said above I think the internal types are useful for the maintainers but the API type I think should be something like possix_atom() | term(). Even if we possibly only return atoms at the moment. I know we have an error section on the man page listing the possix_atoms (or at least the expected ones).

kikofernandez:
- I decided that to keep the types that OTP maintainers could benefit, and leave a more general interface to our users; I used an internal FTP module that contains the expected types for maintainers (`ftp_internal.erl`), while I removed the expected reason of the error from the public API (`ftp.erl`). In this way, `ftp.erl` is simply a wrapper around our internal API.

- Type-wise, we are not breaking anything in Dialyzer or other typing tools, since the change means that `(Reason :: {error, common_reason() | ...}) <: (Reason :: {error, term()})` because `(common_reason() | ...) <: term()` 

Please let me know if this is a viable solution or if we need to find another way. Alternatives:
- Remove the reason types (but I think that we maintainers do benefit from having the expected type error...)
- Collpase all Reason types into a single one (but I think this gives a superset of types to some functions where we do not expect certain kinds of errors)

IngelaAndin:
I think that this solution feels good. I do not think that for example then commit
[ftp: removal of defensive programming checks] feels like an interesting one to keep for the final solution. I think you should try to squash some of the commits to logical steps of what we think the final result should look like.

PR #6560, opened 2022-12-15
Correct some details in the operator table

bjorng:
* Change "priority" to the standard terminology "precedence".

* Document the changed precedence of the `catch` operator.

* Mention the restrictions of `?=` in a note after the operand table.

* Mark the non-associative binary operators as such in the operand table.

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1599,10 +1599,27 @@ catch Expr</code>
       returns the value <c>Any</c>.</p>
       <p><em>Example:</em></p>
     <pre>
-5> <input>catch throw(hello).</input>
+3> <input>catch throw(hello).</input>
 hello</pre>
     <p>If <c>throw/1</c> is not evaluated within a catch, a
       <c>nocatch</c> run-time error occurs.</p>
+
+      <change><p>Before Erlang/OTP 24, the <c>catch</c> operator had
+      the lowest precedence, making it necessary to add parentheses when
+      combining it with the <c>match</c> operator:</p>
+      <pre>
+1> <input>A = (catch 42).</input>
+42
+2> <input>A.</input>
+42</pre>
+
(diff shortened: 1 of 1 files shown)


system/doc/reference_manual/expressions.xml
```diff
@@ -2076,58 +2093,69 @@ end</pre>
...
+    which can only be used in patterns.</p>
+    <p><c>?=</c> is restricted in that it can only be used at
+    the top-level inside a <c>maybe</c> block.</p>
+    </note>
     <p>When evaluating an expression, the operator with the highest
-      priority is evaluated first. Operators with the same priority
-      are evaluated according to their associativity.</p>
-      <p><em>Example:</em></p>
-      <p>The left associative arithmetic operators are evaluated left to
+      precedence is evaluated first. Operators with the same priority
```
jhogberg, line 2143: Did you mean...?
```suggestion
      precedence is evaluated first. Operators with the same precedence
```

bjorng: Yes.

system/doc/reference_manual/expressions.xml
```diff
@@ -2076,58 +2093,69 @@ end</pre>
...
+    the top-level inside a <c>maybe</c> block.</p>
+    </note>
     <p>When evaluating an expression, the operator with the highest
-      priority is evaluated first. Operators with the same priority
-      are evaluated according to their associativity.</p>
-      <p><em>Example:</em></p>
-      <p>The left associative arithmetic operators are evaluated left to
+      precedence is evaluated first. Operators with the same precedence
+      are evaluated according to their associativity. Non-associative
+      operators cannot be combined.</p>
```
RaimoNiskanen, line 2145: It would be nice to have a clarification of "combined" such as
...` be combined (with operator of the same precedence).`

bjorng: OK. I've added that clarification.

PR #6565, opened 2022-12-15
ssl: Refactor to avoid confusion of what is the API

### LINKED ISSUE #6506: Error in type spec in `ssl:format_error/1`
**Describe the bug**
The current type spec for ssl:format_error is:
```erlang
-spec format_error({error, Reason}) -> string() when
      Reason :: any().
```
But its implementation pattern matches on the type being a tuple or a list, and matches on only in the tuple `{error, Reason}` as seen below:

```erlang
-spec format_error({error, Reason}) -> string() when
      Reason :: any().

format_error({error, Reason}) ->
    format_error(Reason);
format_error(Reason) when is_list(Reason) ->
    Reason;
format_error(closed) ->
    "TLS connection is closed";
format_error({tls_alert, {_, Description}}) ->
    Description;
```

**Expected behavior**
I would expect a type specification that matches its implementation.

**Affected versions**
OTP-24+ probably earlier versions as well

**Additional context**
At the moment, Dialyzer does not seem to complain but the type is not correct. This is more a semantic error than an existing bug and it can lead to Dialyzer throwing errors in the future.

IngelaAndin:
Closes #6506

### DIFF
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -1408,39 +1408,15 @@ clear_pem_cache() ->
     ssl_pem_cache:clear().
 
 %%---------------------------------------------------------------
--spec format_error({error, Reason}) -> string() when
+-spec format_error(Reason | {error, Reason}) -> string() when
       Reason :: any().
 %%
 %% Description: Creates error string.
 %%--------------------------------------------------------------------
 format_error({error, Reason}) ->
-    format_error(Reason);
-format_error(Reason) when is_list(Reason) ->
-    Reason;
-format_error(closed) ->
-    "TLS connection is closed";
-format_error({tls_alert, {_, Description}}) ->
-    Description;
-format_error({options,{FileType, File, Reason}}) when FileType == cacertfile;
-						      FileType == certfile;
-						      FileType == keyfile;
-						      FileType == dhfile ->
-    Error = file_error_format(Reason),
-    file_desc(FileType) ++ File ++ ": " ++ Error;
-format_error({options, {socket_options, Option, Error}}) ->
-    lists:flatten(io_lib:format("Invalid transport socket option ~p: ~s", [Option, format_error(Error)]));
-format_error({options, {socket_options, Option}}) ->
-    lists:flatten(io_lib:format("Invalid socket option: ~p", [Option]));
-format_error({options, Options}) ->
-    lists:flatten(io_lib:format("Invalid TLS option: ~p", [Options]));
-
-format_error(Error) ->
-    case inet:format_error(Error) of
-        "unknown POSIX" ++ _ ->
-            unexpected_format(Error);
-        Other ->
-            Other
-    end.
+    do_format_error(Reason);
+format_error(Reason) ->
+    do_format_error(Reason).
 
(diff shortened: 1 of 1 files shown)


IngelaAndin:
I rewrote it to be behavior compatible even if it was not documented. Although I still do not want to specify all possible error reasons as they should never be matched and can not be reliably specified.  Some errors are partially documented with special atoms that may be matched, but they are documented by the functions returning them and not here.

lib/ssl/src/ssl.erl
```diff
@@ -1408,39 +1408,15 @@ clear_pem_cache() ->
     ssl_pem_cache:clear().
 
 %%---------------------------------------------------------------
--spec format_error({error, Reason}) -> string() when
+-spec format_error({error, Reason} | Reason) -> string() when
```
kikofernandez, line 1411: I think the type is semantically correct but maybe too general?
Essentially, as a user who has never used this function, I would need to read the source code to see what I can put as input argument. I am not sure if this is the intended case. If not, I think the following type specs could "help", although they are a bit verbose:

```erlang
-spec format_error({error, ErrorReason} | Reason) -> string() when
      Reason :: string()
              | closed
              | {tls_alert, {_, Description :: string()}}
              | {options, Options :: term()}
              | {options, {socket_options, Option :: term()}}
              | {options, {socket_options, Option :: term(), Error}}
              | {options, {FileType, File :: string(), Error}}
              | InetError,
      FileType    :: cacertfile | certfile | keyfile | dhfile,
      Error       :: {error, ErrorReason} | ErrorReason,
      InetError   :: inet:posix()| system_limit,
      ErrorReason :: term().
```
and also:
```erlang
-spec do_format_error( string()
                     | closed
                     | {tls_alert, {_, Description :: string()}}
                     | {options, Options :: term()}
                     | {options, {socket_options, Option :: term()}}
                     | {options, {socket_options, Option :: term(), Error}}
                     | {options, {FileType, File :: string(), Error}}
                     | InetError) -> string()
              when
      FileType   :: cacertfile | certfile | keyfile | dhfile,
      Error      :: {error, ErrorReason} | ErrorReason,
      ErrorReason     :: term(),
      InetError :: inet:posix() | system_limit.
```

kikofernandez (approved):
I have one comment to consider about the typing.
I think we need to start thinking if we want to have more explicit and verbose types or types that work because they are possibly quite general, so we over-specify. 

Verbosity comes at the price of longer documentation, and over-specification comes at the price of leaving users to have to dig into the source code to understand what can be expected.

Choose your poison :)

lib/ssl/src/ssl.erl
IngelaAndin: Well, the purpose of this function is for the user to be able to provide a logging string from any error reason returned by any of the ssl API functions that may return {error, Reason}. For a legacy reason, the input can even be the tagged error tuple.  The whole point is that the user should not need to know what the error looks like but they want something that they can print in an error log. The reason may for instance include stack trace information for debugging purposes. The user will never need to create the input, it should be a matched Reason. We might want to spec the internal function for our own benefit, but I think that the documented value should be general.

PR #6569, opened 2022-12-16
Call erlang:error/2 with the actual args in erl_anno:anno_info/2,3

erszcz:
I've been playing around with Gradualizer on random community projects and spotted this. It's present as early as `maint-21` and then further up until the current master. I haven't checked branches older than `maint-21`. Most likely it only manifests when not-an-annotation is passed to this function, hence it's gone unnoticed for a while.


bjorng:
Thanks for your pull request. We only accept pull request on the `maint` and `master` branches. I have merged your branch to `maint` and `master`.

PR #6574, opened 2022-12-18
[erts] report writable binaries as off-heap binaries

### LINKED ISSUE #6573: process_info/2 doesn't report some large appended binaries.
**Describe the bug**
process_info/2 doesn't report some large appended binaries. This behavior seems to be dependent on the size of the first binary in the append. 
There is also a difference sometimes between calling process_info for self process vs another process. Where calling process_info for self usually has more chance of getting results.

**To Reproduce**
Compile this module
```erlang
-module(bin_test).

-behaviour(gen_server).

-export([init/1, handle_call/3, handle_cast/2, test/1]).

init(_) ->
  {ok, #{bin => ""}}.

handle_call({add_binary, Size}, _, State) ->
  RandomBytes = crypto:strong_rand_bytes(Size),
  Bin = <<RandomBytes/binary, "a">>,
  NewState = maps:put(bin, Bin, State),
  timer:sleep(200 + rand:uniform(1000)), % To rule out internal racing
  check_binaries(self, Size, self()),
  {reply, ok, NewState}.

handle_cast(_, S) ->
  {noreply, S}.

test(From) ->
  {ok, Pid} = gen_server:start_link(?MODULE, nil, []),
  To = From + 16,
  lists:foreach(fun(Size) ->
    gen_server:call(Pid, {add_binary, Size}),
    check_binaries(other, Size, Pid)
  end, lists:seq(From, To)).

check_binaries(From, Size, Pid) ->
  {binary, Bins} = process_info(Pid, binary),
  % remove small bins, these are used to call io:fwrite/2
  Bins = lists:filter(fun({_, BinSize, _}) -> BinSize > 100 end, Bins),
  case Bins of
    [] -> io:fwrite("~p~n", [[From, Size, "emtpy!"]]);
    Bins -> io:fwrite("~p~n", [[From, Size, Bins]])
  end.
````

then run 
```erlang
bin_test:test(1024 * 1024 * 1024).
```

On my machine I get these results (consistently)
```
[self,1073741824,"emtpy!"]
[other,1073741824,"emtpy!"]
[self,1073741825,[{6730649640,1073741825,1}]]
[other,1073741825,[{6730649640,1073741825,1}]]
[self,1073741826,[{10381389864,1073741826,1}]]
[other,1073741826,[{10381389864,1073741826,1}]]
[self,1073741827,"emtpy!"]
[other,1073741827,"emtpy!"]
[self,1073741828,[{10381389864,1073741828,1}]]
[other,1073741828,"emtpy!"]
[self,1073741829,"emtpy!"]
[other,1073741829,"emtpy!"]
[self,1073741830,"emtpy!"]
[other,1073741830,"emtpy!"]
[self,1073741831,"emtpy!"]
[other,1073741831,"emtpy!"]
[self,1073741832,[{15105876008,1073741832,1}]]
[other,1073741832,"emtpy!"]
[self,1073741833,[{15105876008,1073741833,1}]]
[other,1073741833,"emtpy!"]
[self,1073741834,[{15105876008,1073741834,1}]]
[other,1073741834,[{15105876008,1073741834,1}]]
[self,1073741835,[{15105876008,1073741835,1}]]
[other,1073741835,[{15105876008,1073741835,1}]]
[self,1073741836,[{6730649640,1073741836,1}]]
[other,1073741836,[{6730649640,1073741836,1}]]
[self,1073741837,[{10381389864,1073741837,1}]]
[other,1073741837,[{10381389864,1073741837,1}]]
[self,1073741838,[{6730649640,1073741838,1}]]
[other,1073741838,[{6730649640,1073741838,1}]]
[self,1073741839,[{10381389864,1073741839,1}]]
[other,1073741839,[{10381389864,1073741839,1}]]
[self,1073741840,[{6730649640,1073741840,1}]]
[other,1073741840,[{6730649640,1073741840,1}]]
```

This line triggers the issue:
`Bin = <<RandomBytes/binary, "a">>,`
Replacing it with just
`Bin = RandomBytes/binary,`
does work as expected.

**Expected behavior**
To see at least the currently referenced large binaries reported. 

**Affected versions**
Erlang/OTP 25 [erts-13.1.1] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns] [dtrace]
I have seen this issue on older versions too.

**Additional context**
Running on an Intel Mac with macOS 12.6.

max-au:
When writable binaries were moved to a separate list `proc->wrt_bins`, they weren't added to the output of `process_info(Pid, binary)`. It led to interesting effects of binaries suddenly disappearing after running implicit GC caused by BIF calls, including `process_info` itself. Effectively, running `process_info(self(), binary)` twice in a row resulted in different lists, because first call triggers GC clearing out original binary from the off-heap list:

    start() ->
        OffHeapBin = crypto:strong_rand_bytes(1024 * 1024),
        Bin = <<OffHeapBin/binary, "a">>,
        Before = erlang:process_info(self(), [binary]),
        After = erlang:process_info(self(), [binary]),
        Before =/= After andalso erlang:display({wtf, Before, After}),
        Bin.

Fixes #6573

### DIFF
--- a/erts/emulator/beam/erl_bif_info.c
+++ b/erts/emulator/beam/erl_bif_info.c
@@ -1998,12 +1998,15 @@ process_info_aux(Process *c_p,
 
     case ERTS_PI_IX_BINARY: {
         ErlHeapFragment *hfrag;
+        ErlOffHeap wrt_bins;
         Uint sz;
 
         res = NIL;
         sz = 0;
+        wrt_bins.first = rp->wrt_bins;
 
         (void)erts_bld_bin_list(NULL, &sz, &MSO(rp), NIL);
+        (void)erts_bld_bin_list(NULL, &sz, &wrt_bins, NIL);
         for (hfrag = rp->mbuf; hfrag != NULL; hfrag = hfrag->next) {
             (void)erts_bld_bin_list(NULL, &sz, &hfrag->off_heap, NIL);
         }
@@ -2011,6 +2014,7 @@ process_info_aux(Process *c_p,
         hp = erts_produce_heap(hfact, sz, reserve_size);
 
         res = erts_bld_bin_list(&hp, NULL, &MSO(rp), NIL);
+        res = erts_bld_bin_list(&hp, NULL, &wrt_bins, res);
         for (hfrag = rp->mbuf; hfrag != NULL; hfrag = hfrag->next) {
             res = erts_bld_bin_list(&hp, NULL, &hfrag->off_heap, res);
         }
--- a/erts/emulator/test/bs_construct_SUITE.erl
+++ b/erts/emulator/test/bs_construct_SUITE.erl
@@ -28,6 +28,7 @@
 	 huge_float_field/1, system_limit/1, badarg/1,
 	 copy_writable_binary/1, kostis/1, dynamic/1, bs_add/1,
 	 otp_7422/1, zero_width/1, bad_append/1, bs_append_overflow/1,
+         bs_append_offheap/1,
          reductions/1, fp16/1, error_info/1]).
 
 -include_lib("common_test/include/ct.hrl").
@@ -41,7 +42,8 @@ all() ->
      in_guard, mem_leak, coerce_to_float, bjorn, append_empty_is_same,
      huge_float_field, system_limit, badarg,
      copy_writable_binary, kostis, dynamic, bs_add, otp_7422, zero_width,
-     bad_append, bs_append_overflow, reductions, fp16, error_info].
+     bad_append, bs_append_overflow, bs_append_offheap,
+     reductions, fp16, error_info].
 
 init_per_suite(Config) ->
     Config.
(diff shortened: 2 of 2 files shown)


sverker:
@max-au Thanks for the PR. Looks good, test case and all.

One little nitpick. Writable binaries were not introduced by #5195, it's a quite old performance feature.
However, #5195 moved them into a separate list `proc->wrt_bins` as an optimization of the GC. Can you update the commit message, not to confuse people.

sverker (approved):

max-au:
Updated, thanks for the correction!

While at it, I wonder if there is any way to take a peek into what's stored in these binaries. Something like `process_info(Pid, binaries)` that returns actual binaries (with their content) rather than a reference (VM internal pointer). Sometimes I'm getting curious why a process keeps off-heap binaries way longer than I expect. Knowing what those binaries contain can help.

sverker:
> While at it, I wonder if there is any way to take a peek into what's stored in these binaries. Something like `process_info(Pid, binaries)`

There is nothing like that, but it would be easy to implement. However, then you probably also want to know which parts of large shared binaries that are actually referred by the process. That is a bit trickier, as you have to scan the heap to find all ErlSubBin terms.

markmeeus:
Thanks @max-au !

Just one thing I would like to add:

While it is totally correct to report orig_size for writable ProcBins, it is a bit unexpected to see the total size there. If there was more info in the tuple, (maybe a used_size and an is_writable) it would be more transparent.

Just mentioning this because I think it relates to your 'peeking into binaries' question.

PR #6579, opened 2022-12-19
Improve specs and docs for maps:groups_from_list/2,3

Maria-12648430:
I found the documentation for the recently introduced functions `maps:groups_from_list/2,3` particularly lacking:

* The purpose of the functions was present only in the `fsummary`, but not in the actual description.
* The word "split", commonly used in the sense of "chop into pieces", was IMO misleading. I replaced it with "partition" therefore.
* Some of the example lines were too long and overflowed the code display box of my browser.
* I improved the examples grouping on the 2-remainder of numbers to group on `even` and `odd` instead. I think this is nicer to the eye.
* I improved the wording of how the result map is constructed.
* I also changed the specs of the functions a bit to be more expressive.

Something I would like to put up for discussion is whether the statement "The order of elements within each group list is preserved from the original list" should be kept or removed. While it is true and makes sense, I'm not sure whether it should be explicitly guaranteed. For instance, `lists:partition/2`, which does something similar, does _not_ say anything about element order of the result lists.

### DIFF
--- a/lib/stdlib/doc/src/maps.xml
+++ b/lib/stdlib/doc/src/maps.xml
@@ -232,15 +232,19 @@ val1
 
     <func>
       <name name="groups_from_list" arity="2" since="OTP 25.0"/>
-      <fsummary>Splits the list into groups using a function as discriminator.</fsummary>
+      <fsummary>Partitions a list into groups using a function as discriminator.</fsummary>
       <desc>
-        <p>The result is a map where each key is given by <anno>Fun</anno>
-        and each value is a list of elements. The order of elements within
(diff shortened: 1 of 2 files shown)


bjorng (approved):
Looks good to me!

I think that the guarantee to retain the order is useful. There already is [sofs:relation_to_family/1](https://www.erlang.org/doc/man/sofs.html#relation_to_family-1) that can be used if the order is not important (and especially if one is only interested in the groups but not the map). Also, removing the sentence now would be an incompatibility.

bjorng:
Thanks for your pull request.

PR #6581, opened 2022-12-19
Prevent the compiler from hanging

### LINKED ISSUE #6578: [erlc] hang on small testcase
On master (whether cherry-picking https://github.com/erlang/otp/pull/6559 and https://github.com/erlang/otp/pull/6575 or not), running erlc on the following testcase:
```erlang
f(X) ->
    <<
        0
     || ((+(is_alive())) <
            [
                f(f(0))
             || X
            ]) andalso ok
    >>.
```
does not terminate even after 3 minutes (I've not waited longer, so it is possible that erlc is merely extremely slow and not completely stuck here). Usually, even programs several order of magnitude larger are compiled in seconds at the very most. And reducing this testcase even a tiny bit makes erlc compile it near-instantly.

bjorng:
Closes #6578


jhogberg:
Closed in favor of #6582

PR #6584, opened 2022-12-20
Fix race receiving fragmented messages on terminating connection

sverker:
Could potentially cause memory leaks as well as double free crashes.

Have not been seen to cause problem other then `ASSERT(... (dep->dflags & DFLAG_SEND_SENDER) ...)` failing on debug built beam.


sverker:
CLOSED.
Wrong branch pushed.

PR #6590, opened 2022-12-21
BEAM SSA checker

frej:
[@bjorng and @jhogberg this is the SSA Checker i have been talking about]

This is a series of patches which adds a mechanism to check the emitted
BEAM SSA against patterns embedded in comments in the source code.
This greatly simplifies writing tests for compiler optimizations as it becomes
possible to directly check properties of the emitted code instead of having to create a test which exhibit different behavior depending on the triggering of the optimization.

For examples on how the patterns look like, check [sanity_checks.erl](https://github.com/erlang/otp/blob/f08007ac250712ee3463607bd2f8fd76b9f2ff70/lib/compiler/test/beam_ssa_check_SUITE_data/sanity_checks.erl). As a non-synthetic example, this PR includes [an improvement to ](https://github.com/erlang/otp/commit/7cca58fe31c9f7fd046c41ac097e0341ce7f79a9) `beam_ssa_type:bs_size_unit/2` where it is tested using the new pattern language.

What this PR is missing is internal documentation of the pattern language. As I think that the examples given above are enough to form an opinion about the syntax and provide feedback, I will wait for review comments and finalization of the syntax before creating the documentation.

A work-In-Progress branch using the SSA checking framework for an additional optimization is also [available](https://github.com/frej/otp/tree/frej/private_append).

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -291,6 +291,8 @@ expand_opt(no_type_opt=O, Os) ->
      no_ssa_opt_type_finish | Os];
 expand_opt(no_module_opt=O, Os) ->
     [O,no_recv_opt | Os];
+expand_opt({check_ssa,Tag}, Os) ->
+    [check_ssa, Tag | Os];
 expand_opt(O, Os) -> [O|Os].
 
 -spec format_error(error_description()) -> iolist().
@@ -773,6 +775,16 @@ select_list_passes_1([P|Ps], Opts, Acc) ->
 select_list_passes_1([], _, Acc) ->
     {not_done,reverse(Acc)}.
 
+make_ssa_check_pass(PassFlag) ->
+    F = fun (Code, St) ->
+                case beam_ssa_check:module(Code, PassFlag) of
+                    ok -> {ok, Code, St};
+                    {error, Errors} ->
+                        {error, St#compile{errors=St#compile.errors++Errors}}
+                end
+        end,
+    {iff, PassFlag, {PassFlag, F}}.
+
 %% The standard passes (almost) always run.
 
 standard_passes() ->
@@ -877,6 +889,7 @@ kernel_passes() ->
        {unless,no_bsm_opt,{iff,ssalint,{pass,beam_ssa_lint}}},
 
        {unless,no_ssa_opt,{pass,beam_ssa_opt}},
+       make_ssa_check_pass(post_ssa_opt),
        {iff,dssaopt,{listing,"ssaopt"}},
        {unless,no_ssa_opt,{iff,ssalint,{pass,beam_ssa_lint}}},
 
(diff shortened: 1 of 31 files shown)


lib/stdlib/src/erl_scan.erl
```diff
@@ -93,7 +93,7 @@
 -type text_fun() :: fun((atom(), string()) -> boolean()).
 -type option() :: 'return' | 'return_white_spaces' | 'return_comments'
                 | 'text' | {'reserved_word_fun', resword_fun()}
-                | {'text_fun', text_fun()}.
+                | {'text_fun', text_fun()} | 'checks'.
```
bjorng, line 96: This new `checks` option will unfortunately be visible in the documentation, and as far as I know, there is no easy way to avoid that.

Therefore, I think that the option should have a more specific name, such as `ssa_checks` or `internal_compiler_checks`, and that it should be documented but with a note that is experimental and that it can be changed or removed at any time without prior warning. The same goes for the other options introduced in the other modules.

lib/compiler/src/compile.erl
```diff
@@ -773,6 +775,16 @@ select_list_passes_1([P|Ps], Opts, Acc) ->
 select_list_passes_1([], _, Acc) ->
     {not_done,reverse(Acc)}.
 
+make_ssa_check_pass(PassFlag) ->
+    F = fun (Code, St) ->
+                case beam_ssa_check:module(PassFlag, Code) of
```
bjorng, line 780: All other `module/2` functions take the code as the first argument and the option list as the second argument. I suggest that you adhere to that convention.

```suggestion
                Options = [{pass_flag, PassFlag} | St#compile.options], 
                case beam_ssa_check:module(Code, Options) of
```

bjorng (changes requested):
Overall looks good to me. It seems this will help us ensure that future optimizations are effective in much more direct way.

The `bs_size_unit` module seems obviously generated. How is it generated? Would it be possible to do the generation with a generator written in Erlang? (Nitpick: the indentation is off for the `-export()` attribute.)

Please rebase the branch on the latest `master`.

lib/stdlib/src/erl_scan.erl
frej: Should we have an option that is `{compiler_internal, term()}` and to clearly signal when something is experimental  (with `{compiler_internal, ssa_checks}` used for this particular case)?

frej:
>The bs_size_unit module seems obviously generated. How is it generated? Would it be possible to do the generation with a generator written in Erlang? (Nitpick: the indentation is off for the -export() attribute.)

It's produced by an ugly Erlang generator. I thought about trying to create something quine-like, but it's probably better to just create some helpers in `beam_ssa_check_SUITE` to run a generator and then run the checks on the output. The generator executes faster than the SSA checks, so it won't affect the test run time much.

Will handle the other comments, no objections.

lib/stdlib/src/erl_scan.erl
bjorng: Yes, that's a good idea.

frej:
@bjorng, I have addressed the review comments and also added a short guide describing the SSA checker language in the compiler's internal documentation.

bjorng:

bjorng (approved):

bjorng:
Thanks for your pull request.

PR #6608, opened 2022-12-30
Ssl/warn verify none

riccardomanfrin:
This patch somehow completes a work that I think has never been finished about the option "warn_verify_none" which exists but is not publicly available, nor apparently actually works, when provided in client tls config.

The option must be ignored when provided in server side config, and used on client side to let the user inform the ssl subsystem that he is aware that it's not verifying the server peer, and therefore doesn't want any warning emitted.

### DIFF
--- a/lib/ssl/src/ssl.erl
+++ b/lib/ssl/src/ssl.erl
@@ -405,6 +405,7 @@
 %% -------------------------------------------------------------------------------------------------------
 
 -type client_option()        :: {verify, client_verify_type()} |
+                                {warn_verify_none, boolean()} |
                                 {reuse_session, client_reuse_session()} |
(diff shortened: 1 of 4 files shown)


dgud:
warn_verify_none is an temporary internal setting only,
you are supposed to set `{verify, verify_none}` in options to not get any warnings.

riccardomanfrin:
I see thanks (hadn't seen the note in the doc)

IngelaAndin:
As Dan said the API that suppresses the warning is to set verify to verify_none explicitly. The warning is for the cases when it is set implicitly. If we later change the default for choosing security over ease of use, which is the tradeoff, we can remove the warning.

PR #6615, opened 2023-01-02
system: Clarify documentation for enabling 'maybe'

jhogberg:


system/doc/reference_manual/expressions.xml
```diff
@@ -400,10 +400,11 @@ is_valid_signal(Signal) ->
     <note><p><c>maybe</c> is an experimental new <seeguide
     marker="system/reference_manual:features#features">feature</seeguide>
     introduced in OTP 25. By default, it is disabled. To enable
-    <c>maybe</c>, use compiler option
-    <c>{feature,maybe_expr,enable}</c>.  The feature must also be
-    enabled in the runtime using the <c>-enable-feature</c> option to
-    <c>erl</c>.</p></note>
+    <c>maybe</c>, either use the <c>-feature(maybe_expr,enable)</c>
+    directive (from within source code), or the compiler option
+    <c>{feature,maybe_expr,enable}</c>. The feature must also be enabled
```
KennethL, line 405: In OTP 26 the maybe_expr is enabled in runtime by default and it cannot be disabled. It cannot be disabled in runtime because that could make code (for example in OTP applications ) which relies on the feature to stop working.

KennethL (changes requested):

system/doc/reference_manual/expressions.xml
KennethL: I think @bjorng already implemented it that way on the master branch

jhogberg: Yes, it'll be on by default on `master`, though on a second look this part of the manual doesn't seem to have been updated. I'll fix that in a different PR once this is merged.

I made this PR (for `maint`) to avoid confusion on how to enable the feature from source code, as this note had led people to believe that they could use the `-compile()` directive to enable features.

PR #6618, opened 2023-01-02
Add identity function to the `erlang` module

Maria-12648430:
This PR provides the `identity` function for the `erlang` module. I wonder why this hasn't been implemented earlier, this being one of the most trivial functions out there which practically every other (functional) language sports, including Elixir. Maybe it is because it is so trivial to implement ad hoc, so nobody ever bothered...?

That being said, the `erlang` module is a little different from most of the others, and it is the first time that both @juhlig and me have done anything in there. As documentation for this is sparse, we went pretty much by example and trial and error, and don't know if we did everything needed and everything correct.


KennethL:
I know what an identity function can be used for, but I still think it would be good with an explanation of when to use it and or an example use case in the documentation.

Maria-12648430:
Hi @KennethL :wave: I have been thinking about use cases and found it surprisingly hard. Not that there aren't any, rather it appears that there are no _typical_ use cases that can be sketched out in just one or two sentences or one-liner examples. Put differently, the function can be used in many places, but those places are all narrow and specific. Thus, to outline one use case, I'd have to give comparatively much context which does not translate easily to other contexts.

So, I guess I have to admit that I took the easy way out by not-saying "you'll know you need it when you do" :sweat_smile:

Seriously though, I have been looking at the docs for Elixirs' and Clojures' `identity` functions for inspiration, but they don't provide much in terms of use cases, either. Clojure docs just says "Returns its argument", same as what I put in. Elixir adds "This function can be passed as an anonymous function to transformation functions", IMO not very useful.

The only simple, context-free, somewhat-interesting usage examples I can think of are `lists:all/2` and `lists:any/2` used on lists of booleans...
```erlang
1> lists:all(fun erlang:identity/1, [true, false, true]).
false
2> lists:any(fun erlang:identity/1, [false, true, false]).
true
```
... and with the recently added `maps:groups_from_lists/2,3`...
```erlang
3> maps:groups_from_lists(fun erlang:identity/1, "abcaba").
#{97 => "aaa", 98 => "bb", 99 => "c"}
```

<hr />

However, I think the _real_ usefulness of having this function actually lies elsewhere.

For one, it is a standardized, idiomatic, universally available "handle" for an algorithm, instead of an otherwise needed ad-hoc _spelling out_ of an algorithm (which may be written in a multitude of ways) which will have to be re-implemented in projects, modules, or in-place whenever needed. Using it by name also lets it carry a notion of _intention_, which makes it easier to read and grasp.

For another, IDEs can highlight it, follow it around, look it up in possibly large code bases etc.

On a downside, it is more wordy to write `fun erlang:identity/1` (21 characters) than `fun(X) -> X end` (15 characters), but I don't think that is too bad. We considered shortening it to just `id` (like in Haskell; it would make the above also 15 characters), but decided against as `id` is associated more likely with "identiFIER" that with "identiTY".

michalmuskala:
FWIW Elixir does have this function as `Function.indentity/1`. There has been quite a lengthy discussion on the Elixir mailing list before it was added: https://groups.google.com/g/elixir-lang-core/c/tB61BHYIH1s/m/V99m6dVTBAAJ

bjorng:
I like identity functions, **but**...

We use identity functions quite a lot in test suites for the compiler and runtime system to suppress compiler optimizations. It is defined like this:

```
id(I) -> I.
```

Here is an example how we use it:

```
5 = id(3) + id(2)
```

The call to the `id/1` function ensures that the compiler does not evaluate the entire expression at compile-time, and it suppresses type hints that the compiler would add to the BEAM file. That ensures that we test the actual `+/2` operator in the runtime system without any type hints.

However, I don't think that we'll need an identify function in the `erlang` module. We don't want to write:

```
5 = identity(3) + identity(2)
```

(Assuming that `identifiy/1` is auto-imported.)

Changing the definition of `id/1` is also not much help:

```
id(I) -> identity(I).
```

So I don't see any real improvement for test cases.

While it could still be nice to have an `erlang:identity/1` function, I don't think that it useful enough to earn a place in the already crowded `erlang` module. In the discussion in the Elixir mailing list that @michalmuskala mentioned, the suggestion to add the `identity` function to Elixir's [Kernel](https://hexdocs.pm/elixir/Kernel.html) module was rejected for similar reasons (too limited utility, already too crowded). At the end, it was decided to add it to the [Function](https://hexdocs.pm/elixir/Function.html) module, because that module only contained a few functions and `identify/1` fitted in well there.

We don't have any equivalent to the `Function` module in Erlang, and I can't think of any other module where it would fit in, so I don't think we should add an identify function.

Maria-12648430:
> We don't have any equivalent to the `Function` module in Erlang, and I can't think of any other module where it would fit in, so I don't think we should add an identify function.

Fair enough I think :smile:

Maria-12648430:
@bjorng In the bigger picture, this gives raise to the question how we are going to tackle the overcrowding of some modules. Not adding functions of little use like `identity/1` is one thing, but other functions may be of more utility, or a necessity even for new features (`alias/0,1` and `unalias/1` for example, for the Process Aliases feature that appeared in OTP 24 (?)). The `lists` module is another notoriously overcrowded module. In the long term, I think it is safe to say that modules will become more crowded, some more than others.

So, I'm interested, what is your and the OTP teams opinion/long term plan on this topic, if any? :smile:

juhlig:
@Maria-12648430 the only way to de-crowd modules is to create focused supplementary modules. With a heavy emphasis on _focused_, otherwise we end up with a set of hodgepodge odds-and-ends modules.

For example, the `lists` module could be de-crowded to some degree by moving the `key*` family of functions into an own module, say `tuplelists`. IMO, they were always misplaced in the `lists` module to begin with, as they are concerned with lists of a specific type of elements (tuples of given sizes; and what about lists of maps, for that matter?), while the real purpose of the `lists` module is to work on lists of anything.

So, what I would suggest is to move certain functions targeting a specific use case or type into own modules (`lists:key*` --> `tuplelists`/`maplists`) and make the old functions in the old module delegate to them. Then put notes in their documentations stating that the functions of the new module should be used. Later, remove them from the respective documentation altogether, but only from the documentation, so as not to break any existing code.

Maria-12648430:
> and what about lists of maps, for that matter?

Wasn't there a PR about that, a while ago? \*rummage\* Ah: https://github.com/erlang/otp/pull/4831, followed by https://github.com/erlang/eep/pull/29, which kinda dried up. And I remember drafting up a complimentary EEP for `tuplelists` even, which I subsequently forgot about :sweat_smile:

PR #6619, opened 2023-01-02
beam_ssa_type: Track substitutions on a per-block basis

### LINKED ISSUE #6599: [erlc] internal consistency check failed: "illegal_branch, try_handler"
On master,
```erlang
f(X, #{0 := X, 0 := Y}) ->
    try #{ok => ok} of
        Y ->
            bnot (Y = X)
    after
        ok
    end.
```
crashes erlc with the message:
```
consistency_try:1: function f/2+13:
  Instruction: {jump,{f,8}}
  Error:       {illegal_branch,try_handler,8}:
```

jhogberg:
Fixes #6599

### DIFF
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -210,14 +210,14 @@ sig_function_1(Id, StMap, State0, FuncDb) ->
     Ds = maps:from_list([{Var, FakeCall#b_set{dst=Var}} ||
                             #b_var{}=Var <- Args]),
 
-    Ls = #{ ?EXCEPTION_BLOCK => {incoming, Ts},
-            0 => {incoming, Ts} },
+    Ls = #{ ?EXCEPTION_BLOCK => {incoming, Ts, #{}},
+            0 => {incoming, Ts, #{}} },
 
     {Meta, State2} = sig_init_metadata(Id, Linear, Args, State1),
 
     Wl0 = State1#sig_st.wl,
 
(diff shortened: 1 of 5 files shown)


RobinMorisset:
Thanks for the quick fix!
Unfortunately it seems to have introduced an infinite loop in the compiler (or exponential blow-up); the following testcase compiles in less than 1s on master, but does not compile (at least with a 2mn timeout) with your patch:
```erlang
f() ->
    (X =
        not (X =
            (ok <
                fun() ->
                    ok
                end))) orelse ok.
```

jhogberg:
Oops. I'll have a look tomorrow :)

RobinMorisset:
Another very similar testcase for the infinite loop issue:
```erlang
f() ->
    case _V4 = (not (_V4 = (catch (0 = ok)))) of
        _ when _V4 ->
            ok
    end.
```

Another, likely related issue, is that the following testcase:
```erlang
f(X) when not ((X =:= false) == X) ->
    ok;
f(false) ->
    ok.
```
crashes erlc with this PR (but not on master), with the following error message:
```
type_conflict:1: function f/1+7:
  Instruction: {test,is_eq_exact,{f,6},[{x,1},{atom,true}]}
  Error:       {type_conflict,{t_atom,[true]},{t_atom,[false]}}:
```

jhogberg:
Thanks, it looks like we've dragged the current type pass about as far as we can take it. I'll have to redesign it to fix this issue completely, which will take some time. I'll close this PR and get back to the drawing board.

Thankfully the validator will find all limitations of this kind so there's very little risk of bogus code leaking through.

PR #6628, opened 2023-01-04
erts: Add current stack trace to tracing

### LINKED ISSUE #5333: Tracing: Allow traced function to return current stack trace
Recently in #5305 the `caller_line` option was added (PR is not yet merged to master but this PR is complete). This feature allows match spec bodies to return the calling function name _with_ the call site position using `message`.

The next logical request would be to allow a new match spec body tracing option `stack_trace` that will allow traced functions to return their whole stack trace.

**Justification for feature**
Is this feature useful? I think so. But better than that -- this feature is already available today in the Redbug Erlang tracer. The Redbug tracer has the the ability to return the stack trace of a function via `stack`

e.g.
```erlang
redbug:start('test_mod:some_func->stack', [{msgs, 400}, {time, 10000}])
```

When you think about it, how is the Redbug tracer implementing this feature without any tracing primitives that give you a stack trace??? The answer is simple, Redbug uses tracing option `process_dump` to first obtain a full process dump.

This process dump is later parsed for the all `Return addr` lines to construct the stack trace (see [here](https://github.com/massemanet/redbug/blob/14f6cf50d740d8e7a9226b98907bd66ea68390a0/src/redbug.erl#L495))

As this is currently implemented: 
* It is inefficient because a lot of other data is collected during the `process_dump`
* It is fragile because tomorrow if the process dump text output changes a bit, the functionality will break
* Currently the stack trace provided is a bit crude because it only has the function names. We can provide source file and location just like we did for `caller_line`. This will make the stack trace more useful

This is a useful functionality and  in my view Erlang/OTP should provide a primitive for traced functions to return stack traces.

**Implementation**
The implementation looks relatively straightforward. I have looked at the stack dump related code within the `process_dump` routines in `otp` source. With the experience I got on #5305 I think I can implement it. _But before I spend any time on this problem I'd like to ask if this is something you'd like to see implemented_.

cc: @garazdawi @jhogberg

P.S. There are the usual issues with the stack trace looking funny when `return_trace` is used. However, this problem is present with Redbug even today because it consults the process dump `Return Addr` lines to construct the stack trace. You will see a lot of non-useful "undefined" frames (non-useful to humans anyways) when `return_trace` is in effect.

jdamanalo:
Closes #5333

### DIFF
--- a/erts/emulator/beam/erl_db_util.c
+++ b/erts/emulator/beam/erl_db_util.c
@@ -295,6 +295,7 @@ typedef enum {
     matchTrace2,
     matchTrace3,
     matchCallerLine,
+    matchCurrentStacktrace
 } MatchOps;
 
 /*
@@ -2804,6 +2805,64 @@ restart:
                }
             }
             break;
+        case matchCurrentStacktrace: {
+            Uint sz;
+            Uint heap_size;
+            Eterm mfa;
+            Eterm res;
+            struct StackTrace *s;
+            int depth;
+            FunctionInfo* stk;
+            FunctionInfo* stkp;
+
+            ASSERT(c_p == self);
+
+            depth = unsigned_val(esp[-1]);
+            esp--;
+
+            sz = offsetof(struct StackTrace, trace) + sizeof(ErtsCodePtr) * depth;
+            s = (struct StackTrace *) erts_alloc(ERTS_ALC_T_TMP, sz);
+            s->depth = 0;
+            s->pc = NULL;
+
+            erts_save_stacktrace(c_p, s, depth);
+
+            depth = s->depth;
+            stk = stkp = (FunctionInfo *) erts_alloc(ERTS_ALC_T_TMP,
+                                                     depth*sizeof(FunctionInfo));
+
+            heap_size = 0;
+            for (i = 0; i < depth; i++) {
+                erts_lookup_function_info(stkp, s->trace[i], 1);
+                if (stkp->mfa) {
+                    heap_size += stkp->needed + 2;
+                    stkp++;
+                }
+            }
+
+            res = NIL;
+
+            if (heap_size > 0) {
+                int count = stkp - stk;
+
+                ASSERT(count > 0 && count <= MAX_BACKTRACE_SIZE);
+
+                ehp = HAllocX(build_proc, heap_size, HEAP_XTRA);
+
+                for (i = count - 1; i >= 0; i--) {
+                    ehp = erts_build_mfa_item(&stk[i], ehp, am_true, &mfa, NIL);
+                    res = CONS(ehp, mfa, res);
+                    ehp += 2;
+                }
+            }
+
+            *esp++ = res;
+
+            erts_free(ERTS_ALC_T_TMP, stk);
+            erts_free(ERTS_ALC_T_TMP, s);
+
+            break;
+        }
         case matchSilent:
             ASSERT(c_p == self);
 	    --esp;
@@ -4970,6 +5029,57 @@ static DMCRet dmc_caller_line(DMCContext *context,
     return retOk;
 }
 
+static DMCRet dmc_current_stacktrace(DMCContext *context,
+                                    DMCHeap *heap,
+                                    DMC_STACK_TYPE(UWord) *text,
+                                    Eterm t,
+                                    int *constant)
+{
+    Eterm *p = tuple_val(t);
+    Uint a = arityval(*p);
+    DMCRet ret;
+    int depth;
+
+    if (!check_trace("current_stacktrace", context, constant,
+                    (DCOMP_CALL_TRACE|DCOMP_ALLOW_TRACE_OPS), 0, &ret))
+        return ret;
+
+    switch (a) {
+    case 1:
+        *constant = 0;
+        do_emit_constant(context, text, make_small(erts_backtrace_depth));
+        DMC_PUSH(*text, matchCurrentStacktrace);
+        break;
+    case 2:
+        *constant = 0;
+
+        if (!is_small(p[2])) {
+            RETURN_ERROR("Special form 'current_stacktrace' called with non "
+                         "small argument.", context, *constant);
+        }
+
+        depth = signed_val(p[2]);
+
+        if (depth < 0) {
+            RETURN_ERROR("Special form 'current_stacktrace' called with "
+                         "negative integer argument.", context, *constant);
+        }
+
+        if (depth > erts_backtrace_depth) {
+            p[2] = make_small(erts_backtrace_depth);
+        }
+
+        do_emit_constant(context, text, p[2]);
+        DMC_PUSH(*text, matchCurrentStacktrace);
+        break;
+    default:
+        RETURN_TERM_ERROR("Special form 'current_stacktrace' called with wrong "
+                          "number of arguments in %T.", t, context,
+                          *constant);
+    }
+    return retOk;
+}
+
 static DMCRet dmc_silent(DMCContext *context,
  			 DMCHeap *heap,
 			 DMC_STACK_TYPE(UWord) *text,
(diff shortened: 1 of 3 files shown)


erts/emulator/beam/erl_db_util.c
```diff
@@ -2804,6 +2805,72 @@ Eterm db_prog_match(Process *c_p,
...
+
+            while (stkp > stk) {
+                stkp--;
+                hp = erts_build_mfa_item(stkp, hp, am_true, &mfa, NIL);
+                *ehp++ = mfa;
+                *ehp = make_list(ehp + 1);
+                ehp++;
+            }
+
+            ehp[-1] = NIL;
```
jhogberg, line 2867: This section is a bit odd: we've already calculated the amount of heap space we need in the loop above this, so we don't need to allocate an extra `2 * depth` elements. It should also be a bit easier to build the list using our macros, e.g.

```suggestion
            Eterm res = NIL; /* declare where appropriate. feel free to
                              * clean up the declarations at the top
                              * while you're at it. */

            if (heap_size > 0) {
                ehp = HAllocX(build_proc, heap_size, HEAP_XTRA);

                ASSERT((stkp - stk) > 0 && (stkp - stk) < MAX_BACKTRACE_SIZE);
                for (i = stkp - stk; i >= 0; i--) {
                    Eterm mfa;

                    ehp = erts_build_mfa_item(stk[i], ehp, am_true, &mfa, NIL);
                    res = CONS(ehp, mfa, res);
                    ehp += 2;
                }
            }

            *esp++ = res;
```

(I haven't tested the code above)

erts/emulator/beam/erl_db_util.c
```diff
@@ -2804,6 +2805,72 @@ Eterm db_prog_match(Process *c_p,
...
+            s->pc = NULL;
+
+            erts_save_stacktrace(c_p, s, depth);
+
+            depth = s->depth;
+            stk = stkp = (FunctionInfo *) erts_alloc(ERTS_ALC_T_TMP,
+					     depth*sizeof(FunctionInfo));
+
+            heap_size = 0;
+            for (i = depth - 1; i >= 0; i--) {
```
jhogberg, line 2838: Why are we iterating backwards here?

erts/emulator/beam/erl_db_util.c
```diff
@@ -2804,6 +2805,72 @@ Eterm db_prog_match(Process *c_p,
...
+            sz = offsetof(struct StackTrace, trace) + sizeof(ErtsCodePtr) * depth;
+            s = (struct StackTrace *) erts_alloc(ERTS_ALC_T_TMP, sz);
+            s->depth = 0;
+            s->pc = NULL;
+
+            erts_save_stacktrace(c_p, s, depth);
+
+            depth = s->depth;
+            stk = stkp = (FunctionInfo *) erts_alloc(ERTS_ALC_T_TMP,
+					     depth*sizeof(FunctionInfo));
```
jhogberg, line 2835: Nitpick: please don't mix tabs and spaces, new code should be formatted using only spaces.

erts/emulator/beam/erl_db_util.c
```diff
@@ -2804,6 +2805,72 @@ Eterm db_prog_match(Process *c_p,
...
+            FunctionInfo* stk;
+            FunctionInfo* stkp;
+
+            ASSERT(c_p == self);
+
+            depth = unsigned_val(esp[-1]);
+
+            if (depth > erts_backtrace_depth) {
+                depth = erts_backtrace_depth;
+            }
```
jhogberg, line 2824: Why not clamp the depth during compilation?

jhogberg (changes requested):
Thanks for the PR! Could you look into a few things before I add it to our nightly tests?

erts/emulator/beam/erl_db_util.c
jdamanalo: Thanks for your review, @jhogberg! :) It is better, I have applied it. However, I am unsure what cleaning up the declarations mean. Please advise.

jhogberg: There were a bunch of variable declarations near `case matchCurrentStacktrace: {` that are no longer needed, but I see you've already done that :)

erts/emulator/test/match_spec_SUITE.erl
```diff
@@ -239,6 +240,85 @@ test_5b(Config) when is_list(Config) ->
...
+                   [
+                    {?MODULE, f3_test6, 0, [{file, "test6.erl"}, {line, 21}]},
+                    {?MODULE, f5_test6, 0, [{file, "test6.erl"}, {line, 14}]}
+                   ]},
+                {trace, P3, call, {?MODULE, f1_test6, []},
+                   [
+                    {?MODULE, f2_test6, 1, [{file, "test6.erl"}, {line, 25}]},
+                    {?MODULE, f3_test6, 0, [{file, "test6.erl"}, {line, 21}]}
+                   ]}
+               ]),
```
jhogberg, line 318: We had a bunch of failed testcases on account of `backtrace_depth` not being restored to what it was before. Try restoring it when you're done, e.g.

```suggestion
    %% Test when erts_backtrace_depth is less than depth
    OldDepth = erlang:system_flag(backtrace_depth, 2),
    try
        P3 = spawn(?MODULE, fixed_runner, [self(), Fun]),
        erlang:trace(P3, true, [call]),
        %% Notice `local` function tracing
        erlang:trace_pattern({?MODULE, f2_test6, 1}, Pat2, [local]),
        erlang:trace_pattern({?MODULE, f1_test6, 0}, Pat2, [local]),
        collect(P3, [{trace, P3, call, {?MODULE, f2_test6, [f1]},
                      [
                       {?MODULE, f3_test6, 0, [{file, "test6.erl"}, {line, 21}]},
                       {?MODULE, f5_test6, 0, [{file, "test6.erl"}, {line, 14}]}
                      ]},
                     {trace, P3, call, {?MODULE, f1_test6, []},
                      [
                       {?MODULE, f2_test6, 1, [{file, "test6.erl"}, {line, 25}]},
                       {?MODULE, f3_test6, 0, [{file, "test6.erl"}, {line, 21}]}
                      ]}
                    ])
    after
        erlang:system_flag(backtrace_depth, OldDepth)
    end,
```

jdamanalo: Oh! I apologize for the trouble. I will apply this.

But first I want to consult you about something. I remembered the difference of clamping the depth during execution instead of compilation. When setting the backtrace_depth flag, the stack tracing will immediately update instead of changing only when compiled again. Which is the correct behavior?

jhogberg: Setting it on compilation is the correct behavior IMO.

erts/emulator/beam/erl_db_util.c
```diff
@@ -2804,6 +2805,63 @@ Eterm db_prog_match(Process *c_p,
...
+            Eterm mfa;
+            Eterm res;
+            struct StackTrace *s;
+            int depth;
+            FunctionInfo* stk;
+            FunctionInfo* stkp;
+
+            ASSERT(c_p == self);
+
+            depth = unsigned_val(esp[-1]);
```
jhogberg, line 2820: We forgot to consume the depth constant, so the stack was overrun when writing the result.
```suggestion
            depth = unsigned_val(esp[-1]);
            esp--;
```

jhogberg:
Thanks, the tests look good after the last fix. I'm going on leave tomorrow and will be back on Wednesday next week, so I'll let it percolate in our test runs until then and will merge it after I get back :)

jhogberg:
Sorry for not getting around to this earlier, I fell ill and this PR sort of went on the back burner. I've merged it now though, thanks again for the PR! :-)

jdamanalo:
No worries, it is good that you are well now. Thank you too! I learned a lot.

PR #6632, opened 2023-01-05
Support for fully asynchronous distributed signaling

rickard-green:
Support for *fully asynchronous distributed signaling* where asynchronous send operations **never** block. This functionality is by default disabled and can be enabled per process. For more information see the documentation of [`process_flag(async_dist, Bool)`](https://erlang.github.io/prs/6632/erts-13.1.3/doc/html/erlang.html#process_flag_async_dist).


erts/doc/src/erlang.xml
```diff
@@ -6154,6 +6154,66 @@ receive_replies(ReqId, N, Acc) ->
...
+        <p>
+          Enable or disable <i>truly asynchronous distributed signaling</i>
+          for the calling process. When disabled, the process sending a
+          distributed signal will block in the send operation if the buffer
+          for the distribution channel reach the
+          <seecom marker="erts:erl#+zdbbl">distribution buffer busy
+          limit</seecom>. The process will remain blocked until the buffer
+          shrinks enough. This might in some cases take a substantial amount
+          of time. When <i>truly asynchronous distributed signaling</i> is
+          enabled, operations sending asynchronous distributed signals will
```
KennethL, line 6172: No need to repeat "truly asynchronous distributed signaling" that many times. Is enough to mention it once and then refer to the name of the flag `async_dist` Or maybe call the flag `async_dist_signaling`?

How much is buffered when async_dist = true? Is there a system_limit? configurable? What happens when the limit is reached? Or when no more memory can be allocated? 
Is it ok if one process sending to another problematic node will potentially exhaust all memory and crash the node?

If there was a configurable system_limit the sending process could for example get an exception from the send operation and crash with system_limit as reason.

I am just thinking here because I don't know how it is supposed to work. True asynchronous signaling will always be limited by the memory at some point.

okeuday: @KennethL Are you suggesting a command-line argument like `+zdbbl` but for a higher limit that may prevent an out of memory crash when the `async_dist` is used?  I was also concerned about this, but was happy the warning about use of the feature potentially causing a crash due to running out of memory (in absence of flow control or other limitations) was larger and more explicit in the documentation.

rickard-green:
> I am interested in the same answers, would it be possible to document it in greater details?

When `async_dist` is enabled there is no hard limit on how much you can buffer which is the intention. Signals will be buffered without restrictions just as in the node local case. It is up to the implementer that use this functionality to ensure that the system does not consume an unlimited amount of memory, for example by implementing flow control.

I'm not sure exactly what you want documented in greater detail. I thought it was quite clear in the provided documentation.

okeuday:
@rickard-green If each process had its own memory limit as an offset in kilobytes on the distribution buffer busy limit the option could become more similar to `max_heap_size` (for distributed send memory consumption specific to the Erlang process).  The default could be `{async_dist, 0}` (equivalent to what is currently `{async_dist, false}`) with `{async_dist, true}` becoming `{async_dist, infinity}`.  If `{async_dist, pos_integer()}` was used an exit kill signal could kill the process if it was seen exceeding that distribution buffer busy limit.  Using `infinity` for memory consumption can help make the decision more explicit because the developer is clearly agreeing to the potential for unbounded memory growth.

I can understand something similar may have already been discussed, but wanted to mention the idea in case it hasn't.

rickard-green:
@okeuday Values of `{async, Limit}` where `Limit /= infinity` would just be another flavor of the same problematic semi asynchronous strategy as we already have. The point of this PR was to make it possible to utilize **truly** asynchronous signaling also in the distributed case.

PR #6652, opened 2023-01-10
Fix more buggy assertions

bjorng:
Fix a few more buggy assertions similar to the one in #6646.

To make it easier to do the right thing, introduce new members to enum `Update`. If we are about to call a function that allocates heap space without using the stack pointer, we can now write:

```
emit_enter_runtime<Update::eHeapOnlyAlloc>();
```

which will make sure that both the heap pointer and stack pointer are saved in a debug build, but only the heap pointer in a release build.

For convenience, also introduce a shorthand for saving both the heap and stack pointer (as when doing a normal alloc using `HAlloc()`). We can now write:

```
emit_enter_runtime<Update::eHeapAlloc>();
```

instead of:

```
emit_enter_runtime<Update::eStack | Update::eHeap>();
```

### DIFF
--- a/erts/emulator/beam/jit/arm/beam_asm.hpp
+++ b/erts/emulator/beam/jit/arm/beam_asm.hpp
@@ -417,12 +417,28 @@ protected:
         return arm::Mem(ARG1, CodeIndex, arm::lsl(3));
     }
 
+    /* Prefer `eHeapAlloc` over `eStack | eHeap` when calling
+     * functions in the runtime system that allocate heap
+     * memory (`HAlloc`, heap factories, etc).
+     *
+     * Prefer `eHeapOnlyAlloc` over `eHeapAlloc` for functions
+     * that assume there's already a certain amount of free
+     * space on the heap, such as those using `HeapOnlyAlloc`
+     * or similar. It's slightly cheaper in release builds,
+     * and in debug builds it updates `eStack` to ensure that
+     * we can make heap size assertions. */
     enum Update : int {
         eStack = (1 << 0),
         eHeap = (1 << 1),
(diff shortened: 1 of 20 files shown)


erts/emulator/beam/jit/arm/beam_asm.hpp
```diff
@@ -417,12 +417,23 @@ class BeamAssembler : public ErrorHandler {
         return arm::Mem(ARG1, CodeIndex, arm::lsl(3));
     }
 
+    /* Prefer `eHeapAlloc` over `eStack | eHeap` when calling a
+     * function in the runtime system that calls HAlloc().
+     * Always use `eHeapOnlyAlloc` when calling a function that
+     * calls HeapOnlyAlloc().
+     */
```
jhogberg, line 424:
```suggestion
    /* Prefer `eHeapAlloc` over `eStack | eHeap` when calling
     * functions in the runtime system that allocate heap
     * memory (`HAlloc`, heap factories, et cetera).
     *
     * Prefer `eHeapOnlyAlloc` over `eHeapAlloc` for functions
     * that assume there's already a certain amount of free
     * space on the heap, such as those using `HeapOnlyAlloc`
     * or similar. It's slightly cheaper in release builds,
     * and in debug builds it updates `eStack` to ensure that
     * we can make heap size assertions. */
```

jhogberg (approved):
LGTM

erts/emulator/beam/jit/arm/beam_asm.hpp
bjorng: Yes, that's better. Updated.

PR #1889, opened 2018-07-25
Do not allocate a new map when the value is the same

josevalim:
This patch optimizes map operations to not allocate new maps
when the key is being replaced by the exact same value in memory.

Imagine this very common idiom:

    Map#{key := compute_new_value(Value, Condition)}

where:

    compute_new_x(X, true) -> X + 1;
    compute_new_x(X, false) -> X;

In many cases, we are not changing the value in `Key`, however
the code prior to this patch would still allocate a new array
for the map values. This optimization changes this.

The cost of optimization is minimum, as in the worst case scenario
it only adds a pointer comparison and boolean check. The major benefit
is reducing the GC pressure by avoiding allocating data.

Next we list the operations we have changed alongside the benchmark
results. The benchmarks basically create a map and perform the same
operations, roughly 20000 times, once replacing the key with the same
value, and another with a different value.

## `Map#{Key := Value}`

For a map with 4 keys, replacing the fourth key 20000 times went from
718us to 539us.

For a map with 8 keys, replacing the fourth key 20000 times went from
976us to 555us.

## `maps:update/3`

For a map with 4 keys, replacing the fourth key 20000 times went from
673us to 575us.

For a map with 8 keys, replacing the fourth key 20000 times went from
827us to 585us.

## `maps:put/3`

For a map with 4 keys, replacing the fourth key 20000 times went from
763us to 553us.

For a map with 8 keys, replacing the fourth key 20000 times went from
788us to 561us.

Note that we have ported some optimizations found in maps:update/3
to maps:put/3 while creating this patch.

PR #6657, opened 2023-01-11
Do not allocate a new map when the value is the same encore

michalmuskala:
This is a followup to PR #1889 almost 5 years in the making. The optimisation there was implemented for many cases, but somehow `M#{existing_key => same_value}` and `M#{existing_key => new_value}` cases were missed.

This is a bit more complicated since there are two cases to handle:
* the keys and values didn't change, and we can just return the original map
* the keys didn't change, but values did - we return a new map but keeping original keys tuple

This seems to be hit fairly frequently - while I was working on it it wasn't even possible to start the VM successfully.

I'll provide benchmarks at a later time

### DIFF
--- a/erts/emulator/beam/beam_common.c
+++ b/erts/emulator/beam/beam_common.c
@@ -2076,6 +2076,8 @@ erts_gc_update_map_assoc(Process* p, Eterm* reg, Uint live,
     Eterm new_key;
     Eterm* kp;
     Eterm map;
+    int changed_values = 0;
+    int changed_keys = 0;
 
     num_updates = n / 2;
     map = reg[live];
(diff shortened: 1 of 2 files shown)


bjorng:
I have pushed fixup commit with some improvements. The main improvement is that the copying of the map is avoided by creating the map and keys skeletons in the opposite order, which makes it trivial to discard the new key tuple and reuse the old one.

If you agree with my changes, please squash the commits.

michalmuskala:
This is indeed much better. I considered this, but I wasn't actually sure if this will work well with GC since it makes the heap not monotonic - the reference to keys tuple comes before the keys tuple itself.

bjorng:
Added to our daily builds.

bjorng:
Thanks for your pull request.

PR #6661, opened 2023-01-12
inets: add type specifications

kikofernandez:
This PR adds some type specifications to the `inets` module.

These have been checked by Dialyzer and eqWAlizer. 

Note: Not all types have been added because that would require bigger changes in `stdlib`, but the ones that were possible have been added and are type checked by eqWalizer. The ones that are not possible usually refer to prop lists, because the return type in the spec is always `term()`, which makes eqWAlizer to think that `term()` is the input type to another function. When this other function expects (e.g.,) `integer()`, it reports a mismatch between `term()` and `integer()`.

### DIFF
--- a/lib/inets/src/http_client/httpc_handler.erl
+++ b/lib/inets/src/http_client/httpc_handler.erl
@@ -44,8 +44,8 @@
 
 -record(timers, 
         {
-          request_timers = [] :: [reference()],
-          queue_timer         :: reference() | 'undefined'
+          request_timers = [] :: [reference() | {reference(), term()}],
+          queue_timer         :: reference() | undefined
          }).
 
 -record(state, 
@@ -622,13 +622,11 @@ do_handle_info({ssl_closed, _}, State = #state{request = undefined}) ->
     {stop, normal, State};
 
 %%% Error cases
-do_handle_info({tcp_closed, _}, #state{session = Session0} = State) ->
-    Socket  = Session0#session.socket,
+do_handle_info({tcp_closed, _}, #state{session = #session{socket = Socket}=Session0} = State) ->
     Session = Session0#session{socket = {remote_close, Socket}},
     %% {stop, session_remotly_closed, State};
     {stop, normal, State#state{session = Session}};
-do_handle_info({ssl_closed, _}, #state{session = Session0} = State) ->
-    Socket  = Session0#session.socket,
+do_handle_info({ssl_closed, _}, #state{session = #session{socket = Socket}=Session0} = State) ->
     Session = Session0#session{socket = {remote_close, Socket}},
     %% {stop, session_remotly_closed, State};
     {stop, normal, State#state{session = Session}};
@@ -878,7 +876,7 @@ connect_and_send_upgrade_request(Address, Request, #state{options = Options0} = 
     end.
 
 handler_info(#state{request     = Request, 
-		    session     = Session, 
+		    session     = #session{socket = Socket}=Session,
 		    status_line = _StatusLine, 
 		    pipeline    = Pipeline, 
 		    keep_alive  = KeepAlive, 
@@ -906,8 +904,7 @@ handler_info(#state{request     = Request,
 			  queue:len(KeepAlive)
 		  end,
     Scheme     = Session#session.scheme, 
-    Socket     = Session#session.socket, 
-    SocketType = Session#session.socket_type, 
+    SocketType = Session#session.socket_type,
 
     SocketOpts  = http_transport:getopts(SocketType, Socket), 
     SocketStats = http_transport:getstat(SocketType, Socket), 
@@ -983,11 +980,10 @@ handle_http_body(Body, #state{headers = Headers,
 handle_http_body(_Body, #state{request = #request{method = head}} = State) ->
     handle_response(State#state{body = <<>>});
 
-handle_http_body(Body, #state{headers       = Headers, 
+handle_http_body(Body, #state{headers       = #http_response_h{'transfer-encoding' = TransferEnc}=Headers,
 			      max_body_size = MaxBodySize,
 			      status_line   = {_,Code, _},
 			      request       = Request} = State) ->
-    TransferEnc = Headers#http_response_h.'transfer-encoding',
     case case_insensitive_header(TransferEnc) of
         "chunked" ->
 	    try http_chunk:decode(Body, State#state.max_body_size, 
@@ -1064,7 +1060,7 @@ handle_response(#state{status = Status0} = State0) when Status0 =/= new ->
     handle_cookies(Headers, Request, Options, ProfileName), 
     case httpc_response:result({StatusLine, Headers, Body}, Request) of
 	%% 100-continue
-	continue -> 
+	continue ->
 	    %% Send request body
 	    {_, RequestBody} = Request#request.content,
 	    send_raw(Session, RequestBody),
@@ -1080,7 +1076,7 @@ handle_response(#state{status = Status0} = State0) when Status0 =/= new ->
 
 	%% Ignore unexpected 100-continue response and receive the
 	%% actual response that the server will send right away. 
-	{ignore, Data} -> 
+	{ignore, Data} ->
 	    Relaxed = (Request#request.settings)#http_options.relaxed,
 	    MFA     = {httpc_response, parse,
 		       [State#state.max_header_size, Relaxed]}, 
@@ -1116,8 +1112,7 @@ handle_cookies(_,_, #options{cookies = disabled}, _) ->
 %% so the user will have to call a store command.
 handle_cookies(_,_, #options{cookies = verify}, _) ->
     ok;
-handle_cookies(Headers, Request, #options{cookies = enabled}, ProfileName) ->
-    {Host, _ } = Request#request.address,
+handle_cookies(Headers, #request{address = {Host, _}}=Request, #options{cookies = enabled}, ProfileName) ->
     Cookies = httpc_cookie:cookies(Headers#http_response_h.other, 
 				  Request#request.path, Host),
     httpc_manager:store_cookies(Cookies, Request#request.address,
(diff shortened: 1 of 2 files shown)


lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -622,12 +622,12 @@ do_handle_info({ssl_closed, _}, State = #state{request = undefined}) ->
     {stop, normal, State};
 
 %%% Error cases
-do_handle_info({tcp_closed, _}, #state{session = Session0} = State) ->
+do_handle_info({tcp_closed, _}, #state{session = Session0} = State) when Session0 =/= undefined->
```
IngelaAndin, line 625: It is a bug if  Session0 is undefined at this point.  I think it would be less confusing to assert it by matching session0 = #session{socket = Socket} in the head than checking it is not undefined (that happens to be the record default value).

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -622,12 +622,12 @@ do_handle_info({ssl_closed, _}, State = #state{request = undefined}) ->
...
     Socket  = Session0#session.socket,
     Session = Session0#session{socket = {remote_close, Socket}},
     %% {stop, session_remotly_closed, State};
     {stop, normal, State#state{session = Session}};
-do_handle_info({ssl_closed, _}, #state{session = Session0} = State) ->
+do_handle_info({ssl_closed, _}, #state{session = Session0} = State) when Session0 =/= undefined->
```
IngelaAndin, line 630: Same as above!

IngelaAndin: Well I guess this goes for the whole file. I think it is better to assert that you get the expected type by matching the expected record type `#session{}, #request{} ,  #http_response_h{},` than checking that it is not undefined.

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -1300,15 +1301,15 @@ is_keep_alive_enabled_server("HTTP/1.0",
 is_keep_alive_enabled_server(_,_) ->
     false.
 
-is_keep_alive_connection(Headers, #session{client_close = ClientClose}) ->
+is_keep_alive_connection(Headers, #session{client_close = ClientClose}) when is_boolean(ClientClose) ->
```
IngelaAndin, line 1304: I think this would be better by setting the default in the record to false. There is no use for this field to be undefined!

lib/inets/src/http_client/httpc_response.erl
```diff
@@ -194,7 +194,7 @@ parse_status_code(<<?LF>>, StatusCodeStr,
 		      MaxHeaderSize, Result, true);
 
 parse_status_code(<<?CR, ?LF, Rest/binary>>, StatusCodeStr, 
-		  MaxHeaderSize, Result, true) ->
+		  MaxHeaderSize, Result, true) when is_list(StatusCodeStr) ->
```
IngelaAndin, line 197: Do we really need this guard? We usually trust internal code and do not guard it!

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -44,8 +44,8 @@
 
 -record(timers, 
         {
-          request_timers = [] :: [reference()],
-          queue_timer         :: reference() | 'undefined'
+          request_timers = [] :: [reference() | {reference() | undefined, term()}],
+          queue_timer         :: reference() | undefined
```
IngelaAndin, line 48: I do not understand this one, it is already set to the empty list that should be included in the type [reference()] !?

kikofernandez: not sure I follow. the addition was to add `[referece()] | [{reference() | undefined, term()}]`

KennethL: I don't understand either. Do you mean that request_timers can be a list containing?:
reference() |
{reference(), term()} |
{undefined | term()}

kikofernandez: image
Yes, as shown in the image

IngelaAndin: It looks like the code was updated later than the spec!  The code seems to think that  a request timer is a
`{Requestid::reference(), TimerRef::referencer() | undefined},`  so that would imply` request_timers = [] ::[{reference(), referencer() | undefined}]`

kikofernandez: the reason for why eqWAlizer deduces that it could be `[{undefined, term()}]` is the following code:

image

Line 1399 shows that we are removing a tuple from the `RequestTimers` variable, and because `ReqId` value comes from `Request#request.id`, it can see that there is no default value for `id` field of the record, thus `undefined` is allowed.

A further investigation (but eqWAlizer cannot see this because it analyses function body by function body) is that the `activate_request_timers` actually adds to ReqTimers a tuple with `{Request#request.id, Ref}` to a list of request timers. So, technically, `request_timers` does indeed contain `[reference() | {undefined | reference(), term()}]`

image

If I have not explained good enough why this happens, we can discuss this offline (I got a good grasp of it now )

kikofernandez:
This has been stalled a bit, but I plan on fixing all related issues this week :)
Thanks for the feedback

lib/inets/src/http_client/httpc_handler.erl
kikofernandez: @IngelaAndin I am not sure about this:
> The code seems to think that a request timer is a
{Requestid::reference(), TimerRef::referencer() | undefined}, so that would imply request_timers = [] ::[{reference(), referencer() | undefined}]

I do see that eqWAlizer sees that `request_timer = [ ] :: [reference()] | [{undefined | reference(), term()}]`, where the reason for the `{undefined | reference(), term()}` comes from line 1277 where #request.id could be `undefined` and from line 1399, where no one prevents `RequestTimers` from having multiple `{request.id, term()}` in its list.

IngelaAndin: > Line 1399 shows that we are removing a tuple from the `RequestTimers` variable, and because `ReqId` value comes from `Request#request.id`, it can see that there is no default value for `id` field of the record, thus `undefined` is allowed.

Although the record field has no default value it will always be set when functions in this module are called.

IngelaAndin: > @IngelaAndin I am not sure about this:
> 
> > The code seems to think that a request timer is a
> > {Requestid::reference(), TimerRef::referencer() | undefined}, so that would imply request_timers = [] ::[{reference(), referencer() | undefined}]
> 
> I do see that eqWAlizer sees that `request_timer = [ ] :: [reference()] | [{undefined | reference(), term()}]`, where the reason for the `{undefined | reference(), term()}` comes from line 1277 where #request.id could be `undefined` and from line 1399, where no one prevents `RequestTimers` from having multiple `{request.id, term()}` in its list.

Of course, the list can have multiple  `{request.id, term()}` in its list, that is the point of it being a list, the spec only says it should be a list of such terms, and also the empty list will be a valid one. But the spec saying it could be only a reference seems to no longer be expected by the code!  But let's discuss it together and see what the conclusion will be.

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -622,13 +622,11 @@ do_handle_info({ssl_closed, _}, State = #state{request = undefined}) ->
...
 %%% Error cases
-do_handle_info({tcp_closed, _}, #state{session = Session0} = State) ->
-    Socket  = Session0#session.socket,
+do_handle_info({tcp_closed, _}, #state{session = #session{socket = Socket}=Session0} = State) ->
     Session = Session0#session{socket = {remote_close, Socket}},
     %% {stop, session_remotly_closed, State};
     {stop, normal, State#state{session = Session}};
-do_handle_info({ssl_closed, _}, #state{session = Session0} = State) ->
-    Socket  = Session0#session.socket,
+do_handle_info({ssl_closed, _}, #state{session = #session{socket = Socket}=Session0} = State) when Session0 =/= undefined->
```
IngelaAndin, line 629: Should we not remove the guard here too?

kikofernandez: Yes, thanks!

kikofernandez: Done!

IngelaAndin (approved):

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -622,13 +622,11 @@ do_handle_info({ssl_closed, _}, State = #state{request = undefined}) ->
...
 %%% Error cases
-do_handle_info({tcp_closed, _}, #state{session = Session0} = State) ->
-    Socket  = Session0#session.socket,
+do_handle_info({tcp_closed, _}, #state{session = #session{socket = Socket}=Session0} = State) ->
     Session = Session0#session{socket = {remote_close, Socket}},
     %% {stop, session_remotly_closed, State};
     {stop, normal, State#state{session = Session}};
-do_handle_info({ssl_closed, _}, #state{session = Session0} = State) ->
-    Socket  = Session0#session.socket,
+do_handle_info({ssl_closed, _}, #state{session = #session{socket = Socket}=Session0} = State) ->
```
KennethL, line 629: Would be nice if we could handle this with the same code in only one function body like this:
```erlang
do_handle_info({Closed, _}, #state{session = #session{socket = Socket}=Session0} = State) when Closed == tcp_closed orelse Closed == ssl_closed ->
      Session = Session0#session{socket = {remote_close, Socket}},
￼    %% {stop, session_remotly_closed, State};
￼    {stop, normal, State#state{session = Session}};
```

kikofernandez: I merged the PR before I saw these comments.
Please review #6904 which contains the requested changes :)

PR #6669, opened 2023-01-16
ct: Replace size/1 with either tuple_size/1 or byte_size/1

bjorng:
The `size/1` BIF is not optimized by the JIT, and its use can result in worse types for Dialyzer.

When one knows that the value being tested must be a tuple, `tuple_size/1` should always be preferred.

When one knows that the value being tested must be a binary, `byte_size/1` should be preferred. However, `byte_size/1` also accepts a bitstring (rounding up size to a whole number of bytes), so one must make sure that the call to `byte_size/1` is preceded by a call to `is_binary/1` to ensure that bitstrings are rejected. Note that the compiler removes redundant calls to `is_binary/1`, so if one is not sure whether previous code had made sure that the argument is a binary, it does not harm to add an `is_binary/1` test immediately before the call to `byte_size/1`.

### DIFF
--- a/lib/common_test/src/ct_framework.erl
+++ b/lib/common_test/src/ct_framework.erl
@@ -443,7 +443,7 @@ get_suite_name(Mod, _) ->
 %% Check that alias names are not already in use
 check_for_clashes(TCInfo, [CurrGrInfo|Path], SuiteInfo) ->
     ReqNames = fun(Info) -> [element(2,R) || R <- Info,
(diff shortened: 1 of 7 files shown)


kikofernandez (approved):

bjorng:
I have now also replaced a call to `size/1` in `test_server_SUITE` with a call to `byte_size/1`.

kikofernandez (approved):

PR #6683, opened 2023-01-16
Default -code_path_choice to strict 

josevalim:
This PR may be a bit controversial but I would say it is worth it.

The `init` process has a feature where it tries to preemptively
fix code paths. To do so, it has to traverse all paths and perform
(most commonly) two additional file system lookups.

https://github.com/erlang/otp/blob/bdf564f8ee2ef847a33dded0f30bb9073779ace7/erts/preloaded/src/init.erl#L1125-L1161

While trying to optimize the boot process, I have noticed this
does show up when profiling. 1-2ms on my (fast) local machine
for `erl` and 3-4ms with Elixir (it uses -pa and -pz with increases
the cost for every loaded application).

When trying this on an actual application, such as Livebook which
has 50 dependencies, I got up to 10ms spent on patching paths
on a Linux-like machine typically used for deployment (emulated
with VirtualBox).

The times may be higher for embedded devices and those where
filesystem lookups are more expensive.

Luckily, this can be disabled by setting `-code_path_choice strict`,
which I argue should be the default. We could also try to optimize
this code but I would argue that, unless there are very good reasons,
it should not be the job of the `init` system to fix invalid paths.
The faster booter times can be seen as an additional positive side-effect
from removing this behaviour.

Therefore my suggestion is to change the default for this release and
completely remove the relaxed behaviour in future releases.

PS: This PR will most likely break the build. I am waiting for the desired
outcome to fix tests.

### DIFF
--- a/erts/preloaded/src/init.erl
+++ b/erts/preloaded/src/init.erl
@@ -292,7 +292,7 @@ code_path_choice() ->
 	{ok,[["relaxed"]]} ->
 	    relaxed;
 	_Else ->
-	    relaxed
+	    strict
     end.
 
 boot(Start,Flags,Args) ->


josevalim:
> What will break if we merge this pull request? Do tools that package BEAM files into archive files need to be updated? Do you users of those tools need to do something?

While Mix uses archives, the archives are unpacked on installation, so this change does not affect Elixir. However, even before then, we would always point to the correct path. I will request feedback from the rebar3 team and also from the community on ErlangForums.

In any case, there is a chance this pull request will break existing users. If someone is building an archive and they are not passing the correct path inside the archive to the code path, then it will fail unless the appropriate path is given. In other words, the current mode allows you to pass:

    -pa app.ez/app/ebin

and

    -pa app.ez/ebin

And it will be able to load either app.ez/app/ebin or app.ez/ebin in both cases. This pull request will require the exact path to be given, so I would say it is more of a concern of how users are using archives rather than how they are being built. 

Finally, it is important to note that tools that process archives can keep the existing behaviour by doing the traversal themselves once they detect an archive. So they can keep the behaviour if it is extremely important to do so, but outside of the init process.

josevalim:
Thanks @dgud for the additional context. The code responsible for loading `ERL_LIBS` is in the `code_server` module and the slow down I measured was in relation to `init`. Furthermore, the relaxed code path feature does not impact `code_server` as much because it is only invoked when the given path does not exist.

So one alternative suggestion is to disable the relaxed code path feature from `-pa`, `-pz`, and bootscripts, but still keep it enabled in the code server. That would effectively address the same concerns.

josevalim:
@dgud further updates from studying the code: the code responsible for loading `ERL_LIBS` and Erlang/OTP lib dir already attempts to load the correct path, regardless if the code path choice is relaxed or strict.

The code starts here:

https://github.com/erlang/otp/blob/master/lib/kernel/src/code_server.erl#L79-L89

And then leads here:

https://github.com/erlang/otp/blob/master/lib/kernel/src/code_server.erl#L464-L483

So the behavior you describe will continue working regardless of the  `code_path_choice` option. So I propose to continue with the original proposal in that whoever is loading the archives must pass the correct path (which is what ERL_LIBS/lib_dir handling is doing).

josevalim:
Thank you @frazze-jobb! Should I add a note to myself to send a PR to remove the option altogether on Erlang/OTP 27?

frazze-jobb:
Yes do that!

bjorng:
We have a page in the documentation for things that are to be removed:

https://www.erlang.org/doc/general_info/scheduled_for_removal.html

Here is an example how we added the text for things to be removed in OTP 26:

[system/doc/general_info/scheduled_for_removal_26.inc](https://github.com/erlang/otp/blob/0f5d7d729f072ce18e73009301965bcf4eb68a2a/system/doc/general_info/scheduled_for_removal_26.inc)

josevalim:
I have added the page and changed the docs in #6777.

PR #6688, opened 2023-01-16
Also normalize expand_fun when set via io:setopts

josevalim:
This addresses a regression on Erlang/OTP 26 master.


frazze-jobb:
Thanks for this fix, looks good!

PR #6689, opened 2023-01-16
os_mon: Close cpu_sup and memsup port on terminate

### LINKED ISSUE #4221: ERL-1374: application:stop(os_mon) causes unexpected error
Original reporter: `paulo.ferraz.oliveira@gmail.com`
Component: `Not Specified`
Migrated from: https://bugs.erlang.org/browse/ERL-1374

---

```
The _error_ in question seems to be coming from {{cpu_sup.c}}:
{code:java}
if(rc == 0)
  error("Erlang has closed");
{code}
Is it possible to shut {{os_mon}} down in such a way that it doesn't provoke this?
```

jdamanalo:
Closes #4221

There seems to be a similar issue with memsup. Should I include it here?

### DIFF
--- a/lib/os_mon/c_src/memsup.c
+++ b/lib/os_mon/c_src/memsup.c
@@ -609,6 +609,8 @@ message_loop(int erlin_fd)
 		    case SHOW_SYSTEM_MEM:
 			extended_show_mem();
 			break;
+		    case EXIT:
+			return;
 		    default:	/* ignore all other messages */
 			break;
 		    }
--- a/lib/os_mon/c_src/memsup.h
+++ b/lib/os_mon/c_src/memsup.h
@@ -30,6 +30,8 @@
 /* Extended memory statistics */
 /*IG*/ #define SHOW_SYSTEM_MEM 2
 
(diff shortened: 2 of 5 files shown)


jdamanalo:
Added closing of `memsup` too.

I considered closing `cpu_sup` using `stop()` as mentioned in the issue but I am unsure if it is actually working as intended (It does not seem to bubble up until the internal port). Also, it is unused by `os_mon`. Hence, I decided to just close it on the exit signal. Maybe it can be removed?

I also added more context in the open issue.

rickard-green:
Thanks for the PR. I've pushed a commit that
* only send the quit message to the `cpu_sup` port server when we actually have a port server which prevents an error being logged in the case we do not have a port server
* removed the receive of the never sent `quit` message in the `cpu_sup` measurement server
* takes care of all places where the `close` message can be received in the `memsup` port handler process

Please squash this commit with your commit and force push it.

jdamanalo:
Squashed and rebased. Thanks for getting what I missed! 😅

Issue #6692, opened 2023-01-17
Paths added with -pa and -pz cannot be removed from the module lookup


josevalim:
**Describe the bug**

See the snippet below. I am using Elixir's `ebin` but any `ebin` will work:

```erlang
$ erl -pa lib/elixir/ebin
Erlang/OTP 26 [DEVELOPMENT] [erts-13.1.3] [source-bdf564f8ee] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1] [jit]

Eshell V13.1.3 (press Ctrl+G to abort, type help(). for help)
1> elixir:module_info(module).
elixir
2> code:del_path("lib/elixir/ebin").
true
3> elixir:module_info(module).
elixir
4> code:purge(elixir), code:delete(elixir).
true
5> code:purge(elixir), code:delete(elixir).
true
6> elixir:module_info(module).
elixir
```

The reason why this happens is because when trying to load a module, the `code_server` will fallback to the `erl_prim_loader`:

https://github.com/erlang/otp/blob/master/lib/kernel/src/code_server.erl#L1191

And the `erl_prim_loader` always keep it in its code paths all the -pa/-pz entries plus the last call to `path` from `init`.

I would say this is problematic for two reasons:

1. Looking up a module that does not exist can be twice slower because it has to traverse the same paths twice

2. Because some applications use `erl_prim_loader:get_file` to load files, you can end-up in a scenario where we are looking at code in different code paths and getting different results

~I think the solution to this problem may be relatively simple: move the code loading functionality from the `code_server` to `erl_prim_loader`. All of the path functionality in `code_server` (`_path` functions + `dir` + `get_object_code`) feel like they could be extracted rather cleanly. The functionality is also abstracted away via the `code` module.~ (see comments below why this is bad)

**OTP Versions**

Reproduced in Erlang/OTP 25 and master.

josevalim:
I have found more information that perhaps should be considered when working on this issue. It is related to the `-loader inet` mode:

1. Paths added via `code:add_path*` will not use `inet` loading when using `code:ensure_loaded`/`code:get_object_file` and friends. In this case, only the paths given via -pa/-pz or on boot use `inet` lookups

[...]

Issue #5811, opened 2022-03-24
Add support for caching load paths


josevalim:
**Is your feature request related to a problem? Please describe.**

Currently, code loading has two options: embedded or interactive. During a release, embedded mode restricts the code path to a single application and loads all modules upfront. This avoids code loading after the application boots but uses more memory, as you may load unused modules. On the other hand, interactive mode loads the modules as necessary.

For applications with many dependencies, we have started to noticed that interactive mode is actually slower to boot than embedded mode. That's because the cost of traversing the filesystem to find a module becomes too expensive. This can be particularly true on embedded devices. Therefore, this puts us in an awkward position here we have to choose a faster boot with more memory usage (embedded) or a slower boot with less memory usage (interactive). The slow loading in interactive mode can be noticed when running tests too, where there is an increased time until the first test runs.

**Describe the solution you'd like**

I would like to propose the ability of caching load paths. First, we will change all `code:all_*/1` functions to receive an optional second argument, called `cache`. For example:

    code:add_path("some_dep/ebin", cache)

If the cache option is given, we will perform a `file:list_dir/1` operation whenever we first attempt to lookup the cached directory and then store all files in a map to speed up future lookups. This is particularly useful in interactive mode, as we will no longer need to traverse the filesystem multiple times. In embedded mode, it has no effect.

I also propose additional changes:

  * [ ] Add `-pac` and `-pzc` to `erl`
  * [ ] Make it so all paths for Erlang/OTP apps are cached by default (can this be a breaking change? How does it play with ERL_LIBS?)
  * [ ] Add an option or a mechanism to cache all paths in a release

I will be glad to work on this, but I thought I would submit an initial description of the problem and the possible solution before I jump into it.

KennethL:
We had more or less exactly the functionality you are suggested and removed it in OTP 19.0 (kernel-5.0).
The release note was like this:
```
The code path cache feature turned out not to be very useful in practice and has been removed. If an attempt is made to enable the code path cache, there will be a warning report informing the user that the feature has been removed.
```
In the documentation for OTP 18.0 is should be described how it worked.

KennethL:
In OTP 18 we had this option to the `erl` command. 

`-code_path_cache`

Enables the code path cache of the code server.

That said I am positive to optimizing the speed of code loading during startup, if it is the traversing of directories which is the bottleneck. 
We have to remind ourselves of why we thought the code path cache was not very useful and what the problems was.

KennethL:
[...]

PR #6693, opened 2023-01-17
Remove archives fallback in code path

josevalim:
Archives have a feature where both the archive directory and the application directory can co-exist and the contents of the archive will be read as a fallback. This pull request removes this feature as it is implicit behaviour and has a direct impact on each code path added to the code server with two additional file system lookups per code path.

Furthermore, I struggle to see the benefit of this feature:

1. This could be used to provide a fallback `ebin` directory for a given application, but this requires an application without `ebin` to be added to the code path, which is slightly contradictory as the `ebin` directory is the one typically added to the code path in the first place

2. You most likely cannot  add `priv` from an archive because the functions in the `file` module cannot read contents from archives. Therefore, if you do `file:read_info(code:priv_dir(megaco))`, it will return `error` if the path points to an archive (and we most likely don't want to make that support archives as it will make all error paths from the `file` module more expensive)

Finally, reading any content from an archive is actually considerably slower than reading from an unpacked archive. Therefore we should promote users to unpack archives and use them directly whenever possible. Still, Erlang/OTP should still support archives, especially for cases like escripts, and this pull request removes only one specific implicit feature from them.

---

Apologies for sending another pull request which is removing existing behaviour. But I believe there is merit on this proposal, similar to the one in #6683. Both are implicit behaviours that have an impact in all users while they are likely rarely used. And I keep running into those scenarios while looking into #5811 and #6692.

Given archives are still tagged as experimental, perhaps this is a good opportunity to clean up those implicit behaviours and streamline code loading, with the goal of eventually marking archives as stable.

### DIFF
--- a/lib/kernel/doc/src/code.xml
+++ b/lib/kernel/doc/src/code.xml
@@ -162,34 +162,6 @@ zip:create("mnesia-4.4.7.ez",
      would list the contents of a directory inside an archive.
      See <seeerl marker="erts:erl_prim_loader"><c>erl_prim_loader(3)</c></seeerl>.</p>
 
-    <p>An application archive file and a regular application directory
-     can coexist. This can be useful when it is needed to have
-     parts of the application as regular files. A typical case is the
-     <c>priv</c> directory, which must reside as a regular directory
-     to link in drivers dynamically and start port programs.
-     For other applications that do not need this, directory
-     <c>priv</c> can reside in the archive and the files
-     under the directory <c>priv</c> can be read through
-     <c>erl_prim_loader</c>.</p>
-
-    <p>When a directory is added to the code path and
-     when the entire code path is (re)set, the code server
-     decides which subdirectories in an application that are to be
-     read from the archive and which that are to be read as regular
-     files. If directories are added or removed afterwards, the file
-     access can fail if the code path is not updated (possibly to the
-     same path as before, to trigger the directory resolution
-     update).</p>
-
-     <p>For each directory on the second level in the application archive
-     (<c>ebin</c>, <c>priv</c>, <c>src</c>, and so on), the code server first
-     chooses the regular directory if it exists and second from the
-     archive. Function <c>code:lib_dir/2</c> returns the path to the
-     subdirectory. For example, <c>code:lib_dir(megaco,ebin)</c> can return
-     <c>/otp/root/lib/megaco-3.9.1.1.ez/megaco-3.9.1.1/ebin</c> while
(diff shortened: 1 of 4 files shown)


michalmuskala:
You can read priv directories from archives with `erl_prim_loader(code:priv_dir(megaco))` - and this is where the "fallback" feature works. You can have an application in an archive (probably shared as a build system artifact) and priv dir on disk to provide some host-specific configuration - at least this seems like the intention of how the feature was designed - weather it's used like that in practice is a different question

josevalim:
Correct, the issue is that all code upfront has to be written with erl_prim_loader but most projects that have to read do so through the file API. So for example, if you have a CSV file, you most likely cannot use the built function from a CSV library to read files and rather roll your custom one. At this point, you might as well read the configuration from anywhere. Is there any benefit from reading it as a priv fallback? Also consider that the priv directory cannot be partially replaced, it is all or nothing, so this sort of configuration fully blocks you from using priv for your own purposes in the future.

Other than fallbacks, archives do have their uses, so +1 in making them faster. At the same time, the implicit behavior does have a measurable effect on the system:

* `2*pa*pz+Paths` lookups for every `{path, Paths}` instruction in the boot scripts

* 2 lookups for each path, -pa, -pz, apps in OTP root, apps in erl libs in the code server

michalmuskala:
Yes, I agree the double traversal is problematic. 

However, I think that in general we might need to rethink a bit the whole code loading process and how it's commonly integrated into various tooling to make it really fast. I see leveraging archives as great opportunity for that.

frazze-jobb:
I researched this a little bit to get an understanding on how we should deal with host specific parts, like NIFs that you could bundle in a Zip file (with application structure) together with a .ez archive.

I feel like it would be more convenient if it was put inside the .ez archive (without compression). To streamline them some more. We are not the only ones having this problem. Google put .so files in their .apk and extracted them out of the .apk on installation. So then you would have them both in the .apk and on disk. They solved that by forking glibc and added a dlopen_with_offset. But since we can't do that, then probably the currently most efficient way to distribute an archive with a .so is like it has currently been designed.
I was thinking if you could take out the .so file of the .ez archive on startup, either to disk or a ramdisk and then load it with dlopen().

We want to have archives, and we want them to be performant. I don't know why they aren't that currently, I need to dig some more, @michalmuskala feel free to enlighten me!

frazze-jobb:
I asked around and we accept this change, and we will try to get the dynamic libraries inside the .ez/escript file in the future. If we get a lot of complaints for removing this functionality without providing an alternative we may have to revert this change.

PR #6698, opened 2023-01-17
[stdlib] Optimize gen_server:multi_call()

rickard-green:
Utilize the "new" selective receive optimizations (introduced in OTP 24) and process alias (also introduced in OTP 24) to implement `gen_server:multi_call()` without the use of a middleman process. Also removed support for communication with R6B nodes.

### DIFF
--- a/lib/stdlib/src/gen_server.erl
+++ b/lib/stdlib/src/gen_server.erl
@@ -665,7 +665,7 @@ do_abcast([], _,_) -> abcast.
 %%
 multi_call(Name, Request)
   when is_atom(Name) ->
-    do_multi_call([node() | nodes()], Name, Request, infinity).
+    multi_call([node() | nodes()], Name, Request, infinity).
 
 -spec multi_call(
         Nodes   :: [node()],
@@ -679,7 +679,7 @@ multi_call(Name, Request)
 %%
 multi_call(Nodes, Name, Request)
   when is_list(Nodes), is_atom(Name) ->
-    do_multi_call(Nodes, Name, Request, infinity).
+    multi_call(Nodes, Name, Request, infinity).
 
 -spec multi_call(
         Nodes   :: [node()],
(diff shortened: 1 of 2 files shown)


lib/stdlib/src/gen_server.erl
```diff
@@ -694,8 +694,92 @@ multi_call(Nodes, Name, Request)
...
+                        [noconnect]),
+        ok
+    catch
+        _:_ ->
+            ok
+    end,
+    mc_send(Nodes, Name, Alias, Request, Timer, [[Node|Mon]|Reqs]);
+mc_send(_BadNodes, _Name, Alias, _Request, Timer, Reqs) ->
+    %% Cleanup then fail...
+    unalias(Alias),
```
michalmuskala, line 736: Isn't `unalias` already guaranteed from the `try/after` on line 705? What happens if `unalias` is called twice?

rickard-green: The `try/after` guarantees that we will remove the alias, but we need to remove the alias before calling `mc_recv_tmo()` which clean up the message queue of any unwanted replies. Without the `unalias()` prior to the `mc_recv_tmo()` call, unwanted replies may be received inbetween return from `mc_recv_tmo()` and call to `unalias()` which would be left in the queue after we have left this code.

Multiple calls to `unalias()` will just change the return value of `unalias()`. It returns `true` if an alias was removed; otherwise, `false`.

michalmuskala: Makes sense. Thank you for explaining.

max-au:
I meant "reusable receive markers". I recall they are allocated in "blocks", but I don't remember whether there is a test case that causes a very large amount of these markers to enter the queue (e.g. a multi_call in 40,000 nodes cluster).

rickard-green:
@max-au Yes there can at most be 8 receive markers used at the same time in a signal queue. This code will, however, only use one receive marker regardless of how many nodes `multi_call()` communicates with. This receive marker is inserted into the signal queue just before the alias is created and is bound to the alias which also is matched on in all receive clauses in `multi_call`.

Issue #6672, opened 2023-01-16
Remove calls to `size/1` and use instead `tuple_size/1` or `byte_size/1`


kikofernandez:
**Describe the bug**
The BIF `size/1` works on tuples and binaries. 
To get the JIT to optimise code, OTP code should call in its stead `tuple_size/1` or `byte_size/1`.
Affected modules:
- [x] lib/sasl/src/systools_rc.erl (#6699 )
- [x] lib/eunit/src/eunit_data.erl (#6671 )
- [x] lib/mnesia/test/mnesia_consistency_test.erl (#6695 )
- [x] lib/kernel/test/gen_udp_SUITE.erl (#6700 )
- [x] lib/kernel/src/erts_debug.erl
- [x] lib/kernel/doc/src/socket.xml
- [x] lib/inets/test/httpd_time_test.erl (#6670 )
- [x] lib/inets/test/httpd_poll.erl (#6670 )
- [x] lib/snmp/test/snmp_test_lib.erl (#6694 )
- [x] lib/snmp/test/snmp_agent_SUITE.erl (#6694 )
- [x] lib/snmp/test/snmp_test_mgr_misc.erl (#6694 )
- [x] lib/snmp/test/snmp_test_mgr.erl (#6694 )
- [x] lib/snmp/src/agent/snmpa_net_if.erl (#6694 )
- [x] lib/public_key/src/public_key.erl (#6673 )
- [x] lib/stdlib/test/erl_eval_SUITE.erl (these seem to test `size/1`, so I let them be)
- [x] lib/stdlib/test/erl_lint_SUITE.erl..(same as above)
- [x] lib/stdlib/test/shell_SUITE.erl ....(same as above)
- [x] lib/stdlib/test/qlc_SUITE.erl .......(same as above)
- [x] lib/stdlib/src/gb_sets.erl
- [x] lib/stdlib/src/gb_trees.erl
- [x] lib/ssh/test/ssh_basic_SUITE.erl (#6696 )
- [x] lib/ssh/src/ssh_dbg.erl (#6696 )
- [x] lib/ssh/src/ssh_options.erl (#6696 )
- [x] lib/ssh/src/ssh.erl (#6696 )
- [x] lib/eldap/ebin/ELDAPv3.erl
- [x] lib/diameter/test/diameter_gen_tcp_SUITE.erl (#6702 )
- [x] lib/diameter/test/diameter_codec_test.erl (#6702 )
[...]

PR #6703, opened 2023-01-18
debugger: replace size/1 by xxx_size/1

kikofernandez:
Relates to #6672


bjorng:
You branch is based on `master`, but you are trying to merge it into `maint`.

bjorng:
I've changed the base to `master`. I don't see any reason to include to fix it in `maint`.

bjorng (approved):

kikofernandez:
I do not think that changing the `size/1` fn is correct here, since we are actually also testing that `size/1` works as expected in guards with complex `and/andalso/or/orelse/not`, so I will close the PR

PR #6715, opened 2023-01-19
Remove the file_io_servers ETS table

michalmuskala:
~~Accessing a table by reference is [around 10% faster][1] on a quiet system with a lot of named tables since looking up the table name is relatively expensive. It's even worse in a system with a lot of ETS operations, since looking up a table involves locking, which can have high contention.~~

~~This doesn't seem to impact performance of file:open/2 in a measurable way on a quiet system, but it doesn't look detrimental either and logically should be saving some CPU.~~

This table was used to store Pids of files opened through the file server, however this data was never used by OTP itself. We can skip the overhead of storing this information.

[1]: https://gist.github.com/michalmuskala/5e7fc0031048f89f64b6cc4468d124a3

### DIFF
--- a/lib/kernel/src/file.erl
+++ b/lib/kernel/src/file.erl
@@ -173,16 +173,9 @@ format_error(ErrorId) ->
       Pid :: pid().
 
 pid2name(Pid) when is_pid(Pid) ->
-    case whereis(?FILE_SERVER) of
-	undefined ->
-	    undefined;
-	_ ->
-	    case ets:lookup(?FILE_IO_SERVER_TABLE, Pid) of
-		[{_, Name} | _] ->
-		    {ok, Name};
-		_ ->
-		    undefined
-	    end
+    case file_request(Pid, pid2name) of
+        {ok, _} = Ok -> Ok;
+        _ -> undefined
     end.
 
 %%%-----------------------------------------------------------------
--- a/lib/kernel/src/file_io_server.erl
+++ b/lib/kernel/src/file_io_server.erl
@@ -26,7 +26,7 @@
 
 -export([count_and_find/3]).
 
--record(state, {handle,owner,mref,buf,read_mode,unic}).
+-record(state, {handle,owner,mref,name,buf,read_mode,unic}).
 
 -include("file_int.hrl").
 
@@ -80,8 +80,9 @@ do_start(Spawn, Owner, FileName, ModeList) ->
 				  Self ! {Ref, ok},
 				  server_loop(
 				    #state{handle    = Handle,
-					   owner     = Owner, 
-					   mref      = M, 
+					   owner     = Owner,
+					   mref      = M,
+					   name      = FileName,
 					   buf       = <<>>,
 					   read_mode = ReadMode,
 					   unic = UnicodeMode})
@@ -322,7 +323,10 @@ file_request({read_handle_info, Opts},
         Reply ->
             {reply,Reply,State}
     end;
-file_request(Unknown, 
+file_request(pid2name,
+             #state{name=Name}=State) ->
+    {reply,{ok,Name},State};
+file_request(Unknown,
 	     #state{}=State) ->
     Reason = {request, Unknown},
     {error,{error,Reason},State}.
--- a/lib/kernel/src/file_server.erl
+++ b/lib/kernel/src/file_server.erl
@@ -37,8 +37,6 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2, 
 	 terminate/2, code_change/3]).
 
--define(FILE_IO_SERVER_TABLE, file_io_servers).
-
 -define(FILE_SERVER, file_server_2).      % Registered name
 -define(FILE_IO_SERVER, file_io_server).  % Module
 -define(PRIM_FILE, prim_file).            % Module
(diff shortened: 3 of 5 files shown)


bjorng:
The name of the ETS table is not documented. Therefore, there should not be any code that tries to use the table to find out which files are opened, and this change should be safe. I have added the PR to our daily builds.

However, when looking at what the ETS table is used for, it is not used for anything what I can see. That is, when a file is opened, its name and pid are stored into the ETS table, and it is removed from the table when the table is closed. The file server process itself has no actual use for the table.

It is possible to look at the table using the table viewer in the Observer application. The table will be included among the system tables. This is not documented. Therefore, it seems that it would be safe to remove the use of the table entirely. I will discuss this with the OTP team.

bjorng:
We have discussed the matter in the OTP team, and see no good reason to keep the ETS table since it is unused and undocumented. If it turns out to be a good reason, there is ample time to revert the commit before the release of OTP 26. Please amend this pull request to eliminate the ETS table (or create a new pull request, if you prefer).

michalmuskala:
I pushed a change removing the table entirely.

TBH with this change, I wonder do we actually need the file server at all. Most operations seem to be atomic (with exception of copy - an it looks like on many OSes it's possible to implement copy atomically as well), which means there's not much point in serialising them through the single server, since the OS will serialise them one way or another already.
It could be possible to bypass the server for the atomic operations, after a check that the server process is on the local node.

max-au:
I think this breaks `file:pid2name/1`: https://www.erlang.org/doc/man/file.html#pid2name-1
The warning in the documentation says "This function is intended for debugging only." Personally I'm fine with removing it without any deprecation period at all, but it may not be expected by some other users.

Most frequently used `file` module functions have `[raw]` option that cause `prim_file` NIFs to be used for filesystem access. So technically developer has control whether to use local (raw) or potentially distributed way of accessing files. Checking that `file_server_2` process is alive and non-local may be a good way to automatically apply 'raw' access.

bjorng:
> I think this breaks `file:pid2name/1`: https://www.erlang.org/doc/man/file.html#pid2name-1

Yes, it does. Thanks for noticing this. I missed it when grepping for uses of the table. The original version of the pull request will also break `file:pid2name/1` because it depends on the table having a name.

Even if it is a debug function, I don't think we should break it without prior warning. I think we should close this pull request.

bjorng:
Raw files are not supported by the `io` module:

```
1> {ok,Fd} = file:open("some_file", [write,raw]).
{ok,{file_descriptor,prim_file,
                     #{handle => #Ref<0.243821546.3403808786.124240>,
                       owner => <0.85.0>,r_ahead_size => 0,
                       r_buffer => #Ref<0.243821546.3403808771.124454>}}}
2> io:format(Fd, "~p\n", [{hello,world}]).
** exception error: no function clause matching io:request({file_descriptor,prim_file,
                                                               #{handle => #Ref<0.243821546.3403808786.124240>,
                                                                 owner => <0.85.0>,r_ahead_size => 0,
                                                                 r_buffer => #Ref<0.243821546.3403808771.124454>}},
                                                           {format,"~p\n",[{hello,world}]},
                                                           #Ref<0.243821546.3403677699.124463>) (io.erl, line 588)
     in function  io:format/3 (io.erl, line 91)
```

Therefore, we cannot default to `raw`.

michalmuskala:
> Raw files are not supported by the `io` module:

What I had in mind was not necessarily defaulting to raw files - rather, we can still start the process, but skip the file server. I think this gets us best of both worlds - flexible processing with a process, rather than `raw` file, but we can easily do stuff in parallel without the bottleneck of the file server process.

> I think this breaks `file:pid2name/1`

Thanks for noticing this! I missed it and I agree we shouldn't break it. What if we stored the filename in the file process itself and called it to get the name (this could be done as an extension to the io protocol)? In terms of memory use this should be actually better since we'd only add one extra word to the size of the file process, while we remove a 3 word tuple from the ETS table.

bjorng:
We have decided to deprecate `file:pid2name/1`. See #6743.

bjorng:
> What I had in mind was not necessarily defaulting to raw files - rather, we can still start the process, but skip the file server.

Yes, that could work.

max-au:
>I think we should close this pull request.

I think the original one (where the table was referenced by `tid` instead of public table name) is still useful. Saves a lookup in the names table.
I'm not sure about storing the pid <-> filename association in the process state (large states are notoriously bad at GC, hence the ETS table used).

michalmuskala:
I pushed a commit that restores the `pid2name` functionality, without the ETS table. It should still be beneficial in terms of memory. The function `file:pid2name` is more expensive in this case and somewhat more unpredictable if called on a pid that is not a file, but as a debug utility and [about to be deprecated](#6743) I think this is an acceptable trade-off. This also makes the function work for remote files.

With this change, it means `file:open` could bypass the file server, if it's local since there's no centralised storage.

bjorng:
Added to our daily builds.

bjorng:
The test case `error_logger_h_SUITE:logfile/1` fails.

michalmuskala:
I've pushed a fix to the test. Sorry for the delay, I've been a bit under the weather last couple days

bjorng:
Thanks!

bjorng:
We are ready to merge; please squash your commits.

michalmuskala:
I've rebased, squashed, and updated the commit message

PR #6716, opened 2023-01-19
prim_file:read_file_info/2 avoid copying the record for posix time read

michalmuskala:
When requested time format is already posix, we don't need to adjust anything, but the old code would still copy the `#file_info` record, we can skip the adjustments a function earlier and avoid this copy.

Before `prim_file:read_file_info("mix.exs", [raw, {time, posix}])` would allocate 432 bytes, now it only allocates 312 bytes.

This will require bootstrap update to propagate, but I skipped this as usually requested for external contributors.


jhogberg:
Merged, thanks for the PR! :-)

PR #6718, opened 2023-01-20
Add maps:iterator/2 with ~k and ~K format options for printing ordered maps.

potatosalad:
This PR add support for both term-ordered and custom order of keys for map iterators.  It also includes support for specifying the ordering of maps when printing or formatting.

Summary of changes:

1. New function: `maps:iterator/2`
    > Accepts a `map()` as the first argument and `undefined` (the default for `maps:iterator/1`), `ordered`, or a 2-arity comparison function `fun((A :: Key, B :: Key) -> boolean())`.
    > 
    > For example:
    > ```erlang
    > %% Ordered
    > Iter = maps:iterator(#{a => b, c => d}, ordered),
    > {a, b, NextIter} = maps:next(Iter).
    > 
    > %% Reversed
    > Iter = maps:iterator(#{a => b, c => d}, fun(A, B) -> B =< A end),
    > {c, d, NextIter} = maps:next(NextIter).
    > ```
2. <del>New function: `maps:next_key/1`
    > Accepts a `maps:iterator()` and returns `{Key, NextIter}`.  This is the same as `maps:next/1`, but does not return the `Value`.
    </del>
3. New type: `maps:iterator_order/0` and `maps:iterator_order/1`
    > ```erlang
    > -type iterator_order(Key) :: undefined | ordered | fun((A :: Key, B :: Key) -> boolean()).
    > -type iterator_order() :: iterator_order(term()).
    > ```
4. Changed type: `maps:iterator/2`
    > Now accepts type `nonempty_improper_list(list(Key), #{Key => Value})` which is used for ordering of keys during iteration.
5. Changed function: `maps:to_list/1`
    > Now accepts `map() | maps:iterator()`.
    > 
    > For example:
    > ```erlang
    > %% Ordered
    > [{a, b}, {c, d}] = maps:to_list(maps:iterator(#{a => b, c => d}, ordered)).
    > 
    > %% Reversed
    > [{c, d}, {a, b}] = maps:to_list(maps:iterator(#{a => b, c => d}, fun(A, B) -> B =< A end)).
    > ```
6. New `io` modifier: `~k` for ordered map formatting
    > ```erlang
    > Map = #{a => b, c => d},
    > FormatList = io_lib:scan_format("~kp", [Map]),
    > {"~kp", [Map]} = io_lib:unscan_format(FormatList).
    > 
    > "#{a => b,c => d}" = lists:flatten(io_lib:format("~kp", [Map])).
    > ```
7. New `io` modifier: `~K` for custom order map formatting
    > ```erlang
    > Map = #{a => b, c => d},
    > RevCmpFun = fun(A, B) -> B =< A end,
    > FormatList = io_lib:scan_format("~Kp", [RevCmpFun, Map]),
    > {"~Kp", [RevCmpFun, Map]} = io_lib:unscan_format(FormatList).
    > 
    > "#{c => d,a => b}" = lists:flatten(io_lib:format("~Kp", [RevCmpFun, Map])).
    > ```
8. Tests for all of the above in `io_SUITE` and `maps_SUITE`
    > ```bash
    > make stdlib_test ARGS="-suite io_SUITE"
    > make stdlib_test ARGS="-suite maps_SUITE"
    > ```
9. Updated `maps.xml`, `io.xml`, and `io_lib.xml` documentation.

### DIFF
--- a/erts/emulator/beam/erl_map.c
+++ b/erts/emulator/beam/erl_map.c
@@ -3488,6 +3488,60 @@ BIF_RETTYPE erts_internal_map_next_3(BIF_ALIST_3) {
         BIF_ERROR(BIF_P, BADARG);
     }
 
+    /* Handle an ordered iterator. */
+    if (type == iterator && (is_list(path) || is_nil(path))) {
+#ifdef DEBUG
+#define ORDERED_ITER_FACTOR 200
+#else
+#define ORDERED_ITER_FACTOR 32
+#endif
+        int orig_elems = MAX(1, ERTS_BIF_REDS_LEFT(BIF_P) / ORDERED_ITER_FACTOR);
+        int elems = orig_elems;
+        Uint needed = 4 * elems + 2;
+        Eterm *hp = HAlloc(BIF_P, needed);
+        Eterm *hp_end = hp + needed;
+        Eterm result = am_none;
+        Eterm *patch_ptr = &result;
+
+        while (is_list(path) && elems > 0) {
+            Eterm *lst = list_val(path);
+            Eterm key = CAR(lst);
+            Eterm res = make_tuple(hp);
+            const Eterm *value = erts_maps_get(key, map);
+            if (!value) {
+            ordered_badarg:
+                HRelease(BIF_P, hp_end, hp);
+                BIF_ERROR(BIF_P, BADARG);
+            }
+            hp[0] = make_arityval(3);
+            hp[1] = key;
+            hp[2] = *value;
+            *patch_ptr = res;
+            patch_ptr = &hp[3];
+            hp += 4;
+            path = CDR(lst);
+            elems--;
+        }
+
+        if (is_list(path)) {
+            Eterm next = CONS(hp, path, map);
+            hp += 2;
+            ASSERT(hp == hp_end);
+            *patch_ptr = next;
+            BUMP_ALL_REDS(BIF_P);
+            ASSERT(is_tuple(result));
+            BIF_RET(result);
+        } else if (is_nil(path)) {
+            HRelease(BIF_P, hp_end, hp);
+            *patch_ptr = am_none;
+            BUMP_REDS(BIF_P, ORDERED_ITER_FACTOR * (orig_elems - elems));
+            ASSERT(result == am_none || is_tuple(result));
+            BIF_RET(result);
+        } else {
+            goto ordered_badarg;
+        }
+    }
+
     if (is_flatmap(map)) {
         Uint n;
 	Eterm *ks,*vs, res, *hp;
--- a/lib/stdlib/doc/src/maps.xml
+++ b/lib/stdlib/doc/src/maps.xml
@@ -582,10 +582,12 @@ error</code>
       <fsummary></fsummary>
       <desc>
         <p>Returns a list of pairs representing the key-value associations of
-          <c><anno>Map</anno></c>, where the pairs
+          <c><anno>MapOrIterator</anno></c>, where the pairs
           <c>[{K1,V1}, ..., {Kn,Vn}]</c> are returned in arbitrary order.</p>
         <p>The call fails with a <c>{badmap,Map}</c> exception if
-          <c><anno>Map</anno></c> is not a map.</p>
+        <c><anno>MapOrIterator</anno></c> is not a map or an iterator obtained
+        by a call to <seemfa marker="#iterator/1">iterator/1</seemfa> or
+        <seemfa marker="#iterator/2">iterator/2</seemfa>.</p>
         <p><em>Example:</em></p>
         <code type="none">
 > Map = #{42 => value_three,1337 => "value two","a" => 1},
--- a/lib/stdlib/doc/src/maps.xml
+++ b/lib/stdlib/doc/src/maps.xml
@@ -42,16 +42,36 @@
       <desc>
         <p>An iterator representing the associations in a map with keys of type
           <c><anno>Key</anno></c> and values of type <c><anno>Value</anno></c>.</p>
-        <p>Created using <seemfa marker="#iterator/1"><c>maps:iterator/1</c></seemfa>.</p>
+        <p>Created using <seemfa marker="#iterator/1"><c>maps:iterator/1</c></seemfa> or
+          <seemfa marker="#iterator/2"><c>maps:iterator/2</c></seemfa>.</p>
         <p>Consumed by <seemfa marker="#next/1"><c>maps:next/1</c></seemfa>,
           <seemfa marker="#filter/2"><c>maps:filter/2</c></seemfa>,
-          <seemfa marker="#fold/3"><c>maps:fold/3</c></seemfa> and
-          <seemfa marker="#map/2"><c>maps:map/2</c></seemfa>.</p>
+          <seemfa marker="#fold/3"><c>maps:fold/3</c></seemfa>,
+          <seemfa marker="#map/2"><c>maps:map/2</c></seemfa>, and
+          <seemfa marker="#to_list/1"><c>maps:to_list/1</c></seemfa>.</p>
       </desc>
     </datatype>
 
     <datatype>
       <name name="iterator" n_vars="0"/>
+    </datatype>
+
+    <datatype>
+      <name name="iterator_order" n_vars="1"/>
+      <desc>
+        <p>Key-based iterator order option that can be one of <c>undefined</c>
+          (default for <seemfa marker="#iterator/1"><c>maps:iterator/1</c></seemfa>),
+          <c>ordered</c> (sorted in map-key order), or a custom sorting function.</p>
+        <p>Used by <seemfa marker="#iterator/2"><c>maps:iterator/2</c></seemfa>.</p>
+        <p>The
+        <seeguide
+            marker="system/reference_manual:expressions#term-comparisons">
+          Expressions section</seeguide> contains descriptions of how terms are ordered.</p>
+      </desc>
+    </datatype>
+
+    <datatype>
+      <name name="iterator_order" n_vars="0"/>
     </datatype>
   </datatypes>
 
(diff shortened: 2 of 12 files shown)


bjorng:
I have a suggestion for a simplification of the ordered iterators. In `maps:iterator/2`, we can expand the sorted list fully, eliminating the need for a more complicated `maps:next/1` function:

```
iterator(M, undefined) when is_map(M) ->
    [0 | M];
iterator(M, ordered) when is_map(M) ->
    make_iterator(lists:sort(maps:to_list(M)));
iterator(M, CmpFun) when is_map(M), is_function(CmpFun, 2) ->
    make_iterator(lists:sort(CmpFun, maps:to_list(M)));
iterator(M, Order) ->
    badarg_with_info([M, Order]).

make_iterator([{K, V} | T]) ->
    {K, V, make_iterator(T)};
make_iterator([]) ->
    none.
```

The iterator will look like this:

```
1> Map = sets:from_list(lists:seq(1, 5), [{version,2}]).
#{1 => [],2 => [],3 => [],4 => [],5 => []}
2> maps:iterator(Map, ordered).
{1,[],{2,[],{3,[],{4,[],{5,[],none}}}}}
```

Each call to `maps:next/1` on this iterator is very cheap with no need to build anything:

```
next({K, V, I} = KVI) ->
    KVI;
```

The iterator suggested in this pull request would require doing a lookup in the original map and building a list and tuple for each call:

```
next([[Key | Keys] | Map]) when is_map(Map) ->
    {Key, maps:get(Key, Map), [Keys | Map]};
```

If the map will no longer be used after the call to `maps:iterator/2`, my suggested iterator requires less memory than the original iterator, which would keep a list of the keys in the map **and** the original map itself. (The nested tuples takes twice as much memory as the list of keys.)

bjorng:
> New function: `maps:next_key/1`

Can you give a motivation for adding this function?

As I see it, from the user's point of view, using `maps:next_key/1`:

```
case maps:next_key(Iter0) ->
    {K, Iter} ->
        . . .
```

is slightly more convenient than:

```
case maps:next(Iter0) ->
    {K, _, Iter} ->
        . . .
```

However, that slight extra convenience comes with an extra cost of building a tuple in the call to `maps:next_key/1`:

```
next_key({K, _V, I}) ->
    {K, I};
```

which is a cost doesn't exist in `maps:next/1`. As a user, not knowing anything about the implementation, I could reasonably expect that just retrieving the key would be more efficient than retrieving both the key and the value.

bjorng:
After a long discussion with @sverker and @rickard-green, we have decided that we want to avoid the worst case memory utilization that my proposed iterator would cause (that is, when the map would be kept alongside the iterator). Therefore, we will use the iterator suggested in this pull request. However, we want to implement it in slightly different way by moving part of the implementation into the `erts_internal:map_next/3` BIF in the runtime system. That way, the performance of using the iterator can be improved. (I am interested in that because I am working on implementing map comprehensions.)

We also think that `maps:next_key/1` should be removed because it cannot be implemented efficiently with the implementation

bjorng:
I've now force-pushed an update with the following changes:

* `maps:next_key/1` has been removed.
* Part of the implementation of `maps:next/1` has been moved to the BIF `erts_internal:map_next/1`. (@sverker, please take a look.)
* I have updated the documentation for `maps:to_list/1` because the change of the spec for it broke the documentation build.
* I've silenced Dialyzer warnings for creation of improper lists.

erts/emulator/beam/erl_map.c
```diff
@@ -3501,6 +3501,58 @@ BIF_RETTYPE erts_internal_map_next_3(BIF_ALIST_3) {
...
+            path = CDR(lst);
+            elems--;
+        }
+
+        if (is_list(path)) {
+            Eterm next = CONS(hp, path, map);
+            hp += 2;
+            *patch_ptr = next;
+            BUMP_ALL_REDS(BIF_P);
+            ERTS_ASSERT(result == am_none || is_tuple(result));
```
sverker, line 3544: I don't see how `result` could be `am_none` here.

and it should probably be an ASSERT macro for debug build only.

sverker: and I think you could also ASSERT(hp == hp_end).

lib/stdlib/src/maps.erl
```diff
@@ -455,16 +483,38 @@ size(Map) ->
...
 
+-spec iterator(Map, Order) -> Iterator when
+      Map :: #{Key => Value},
+      Order :: iterator_order(Key),
+      Iterator :: iterator(Key, Value).
+
+iterator(M, undefined) when is_map(M) ->
+    [0 | M];
+iterator(M, ordered) when is_map(M) ->
+    Keys = lists:sort(maps:keys(M)),
```
sverker, line 497: Should we let the sort order be arithmetic or the map-key order which will
- give a strict order where integer and float keys may not get arbitrary ordered
- mimic the legacy order for small maps

potatosalad: I'm not sure I understand, what would the arithmetic order look like here versus the term order?

potatosalad:
Updated the `maps.xml` docs, still working on the `io.xml`, `io_lib.xml`, etc doc updates.

erts/emulator/beam/erl_map.c
bjorng: Yes, it should be `ASSERT()`, and `result` can only be `am_none` if the list is empty. Fixed.

bjorng:
@potatosalad, I have again force-pushed the branch. While at it, I shortened the commit message for your document update commit because it was too long and got broken in an ugly way when displayed by github. Generally, the first line in a commit message should be no longer than 72 characters (and it should not end in a period).

lib/stdlib/src/maps.erl
bjorng: I think that we should use the map-key order to ensure a consistent order for numeric keys. I have done that change and added a test.

lib/stdlib/doc/src/maps.xml
```diff
@@ -582,15 +641,24 @@ error</code>
...
-          <c><anno>Map</anno></c> is not a map.</p>
+        <c><anno>MapOrIterator</anno></c> is not a map or an iterator obtained
+        by a call to <seemfa marker="#iterator/1">iterator/1</seemfa> or
+        <seemfa marker="#iterator/2">iterator/2</seemfa>.</p>
         <p><em>Example:</em></p>
         <code type="none">
 > Map = #{42 => value_three,1337 => "value two","a" => 1},
   maps:to_list(Map).
 [{42,value_three},{1337,"value two"},{"a",1}]</code>
+        <p><em>Example (using <seemfa marker="#iterator/2">iterator/2</seemfa>):</em></p>
```
bjorng, line 655: This line breaks the documentation build.

```suggestion
        <p><em>Example (using </em><seemfa marker="#iterator/2">iterator/2</seemfa><em>):</em></p>
```

potatosalad:
@bjorng Fixed the `xmllint` issues and updated `io.xml` and `io_lib.xml` documentation and force-pushed the most recent 2 commits.  Also kept my commit message less than 72 characters this time 🙂 

One additional thought: would it be worth having a `reversed` option for `maps:iterator_order()`?  This would allow `maps:fold/3` and to emulate both [`lists:foldl/3`](https://www.erlang.org/doc/man/lists.html#foldl-3) and [`lists:foldr/3`](https://www.erlang.org/doc/man/lists.html#foldr-3) by specifying `ordered` or `reversed`.

For example:

```erl
> M = #{a => 1, b => 2, c => 3}.
#{a => 1,b => 2,c => 3}
> maps:to_list(maps:iterator(M, ordered)).
[{a,1},{b,2},{c,3}]
> maps:to_list(maps:iterator(M, reversed)).
[{c,3},{b,2},{a,1}]
```

Reason I ask:

With the switch for `ordered` from `fun(A, B) -> A =< B end` to `fun(A, B) -> erts_internal:cmp_term(A, B) =< 0 end`, it's no longer as easy for someone to iterate through the map in "reversed" order without having to use `erts_internal:cmp_term/2`.  I think this could lead to unexpected behavior for those trying to iterate in reverse using "term-order" instead of the internal map-key order.

For example:

```erl
> M = #{-1.0 => a, 0.0 => b, -1 => c, 0 => d}.
#{-1 => c,0 => d,-1.0 => a,0.0 => b}
> maps:to_list(maps:iterator(M, ordered)).
[{-1,c},{0,d},{-1.0,a},{0.0,b}]
> % Potentially unexpected: this is *not* the reverse of `ordered` above.
> maps:to_list(maps:iterator(M, fun(A, B) -> B =< A end)).
[{0,d},{0.0,b},{-1,c},{-1.0,a}]
> % This is what users would need to use instead to achieve "reversed" order.
> maps:to_list(maps:iterator(M, fun(A, B) -> erts_internal:cmp_term(B, A) =< 0 end)).
[{0.0,b},{-1.0,a},{0,d},{-1,c}]
```

potatosalad:
Force-pushed one more time in order to resolve conflicts with `master` branch.

sverker:
There is a failing test case [map_SUITE:t_bif_map_to_list](https://erlang.github.io/prs/6718/ct_logs/index.html). Looks like it gets a `badarg` instead of expected `badmap`.

lib/stdlib/src/maps.erl
```diff
@@ -220,14 +225,37 @@ remove(_,_) -> erlang:nif_error(undef).
...
+    MapOrIterator :: #{Key => Value} | iterator(Key, Value).
 
 to_list(Map) when is_map(Map) ->
     to_list_internal(erts_internal:map_next(0, Map, []));
+to_list({_K, _V, _NextIterator} = Iterator) ->
+    try to_list_from_iterator(Iterator) of
+        Result -> Result
+    catch
+        error:badarg ->
+            badarg_with_info([Iterator])
```
bjorng, line 238: Regarding the failed test case, the tuple was confused with an iterator. @sverker  and I discussed it and we'll think that the most logical is to raise a `badmap` exception for bad iterators too. I think that the following (not tested) will work:

```suggestion
            error_with_info(badmap, [Iterator])
```

sverker:
@potatosalad About `maps:iterator(M, reversed)`.
It seems reasonable. However, we would like to get this PR tested and merged soon in order to use the ~k for other things (like some failing dialyzer tests).
Either add the `reversed` rather quick or we do that later in a separate PR.

potatosalad:
Force-pushed again, hopefully everything should be green now.

@bjorng Fixed the `badarg` -> `badmap` issue with the emulator's `map_SUITE`.

@sverker Sounds good, I'll submit a separate PR for `reversed` support.

lib/stdlib/src/maps.erl
```diff
@@ -220,13 +225,36 @@ remove(_,_) -> erlang:nif_error(undef).
...
+            error_with_info({badmap, Iterator}, [Iterator])
+    end;
+to_list([PathOrKeys | Map] = Iterator)
+  when (is_integer(PathOrKeys) orelse is_list(PathOrKeys)), is_map(Map) ->
+    try to_list_from_iterator(Iterator) of
+        Result -> Result
+    catch
+        error:badarg ->
+            error_with_info({badmap, Iterator}, [Iterator])
+    end;
```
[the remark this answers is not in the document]
Maria-12648430: Sure, will do :smiley_cat:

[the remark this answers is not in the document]
Maria-12648430: See #6762 :wink:

PR #6720, opened 2023-01-23
Add `erlang:info_timer/0`

manuel-rubio:
The problem with the timers is that we have no information about how many are running and we need to know the reference for retrieving data and canceling these timers. If we want to provide monitoring information for Observer, we need a listing function that gives us all of the timers.

This function is returning a map where the key is the `id` of the timer and it's containing another map with the following keys:

- `time_left` is the remaining time for the timer to be triggered.
- `message` is the message that will be sent to the `receiver`.
- `receiver` is the process (PID or name) that will receive the `message`.

An example:

```
Erlang/OTP 26 [DEVELOPMENT] [erts-13.1.4] [source-c1dfdb4031] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Eshell V13.1.4 (press Ctrl+G to abort, type help(). for help)
1> erlang:info_timer().
#{#Ref<0.3315389924.154664961.229914> =>
      #{receiver => <0.51.0>,message => refresh_timeout,
        time_left => 3598516},
  #Ref<0.3315389924.154664961.230108> =>
      #{receiver => <0.85.0>,
        message => {<0.85.0>,write_cache},
        time_left => 638}}
```


michalmuskala:
Could this return maps rather than lists of tuple? I think it would be much more ergonomic to use. Additionally, with the key part pre-allocated just once, it should also use much less memory

manuel-rubio:
> Could this return maps rather than lists of tuple? I think it would be much more ergonomic to use. Additionally, with the key part pre-allocated just once, it should also use much less memory

good idea! I'll modify it to use maps... I'm thinking about doing something like this:

```erlang
#{#Ref<0.3987089801.1463287815.22163> =>
      #{message => refresh_timeout,receiver => <0.51.0>,
        time_left => 3574886},
  #Ref<0.3987089801.1463287815.22166> =>
      #{message => {<0.85.0>,write_cache},
        receiver => <0.85.0>,time_left => 1992}}
```

About the memory, I'm getting SIGVABRT when I'm debugging and SIGSEGV when I'm in console running this code:

```erlang
lists:foreach(fun(_) -> erlang:info_timer() end, lists:seq(1, 1_000)).
```

I was checking with `lldb` but it's not helping about where could be the issue:

```
* thread #6, name = '2_scheduler', stop reason = signal SIGABRT
  * frame #0: 0x00007ff8180b000e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007ff8180e61ff libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007ff818031d24 libsystem_c.dylib`abort + 123
    frame #3: 0x00000001001c2d24 beam.debug.smp`check_memory(start=0x0000000146127420, end=0x00000001461285f8) at erl_debug.c:336:6
    frame #4: 0x00000001001c2bd4 beam.debug.smp`erts_check_for_holes(p=0x0000000142d74bf0) at erl_debug.c:311:5
    frame #5: 0x0000000100045f85 beam.debug.smp`beam_jit_call_bif(c_p=0x0000000142d74bf0, reg=0x0000700008b55dc0, I=0x0000000143d91010, vbf=<unavailable>, arity=0)(process*, unsigned long*, void const*), unsigned long) at beam_jit_common.cpp:523:9 [opt]
    frame #6: 0x0000000143d004b0
```

Any ideas?

manuel-rubio:
This PR is related to #6722

manuel-rubio:
@michalmuskala I've implemented the code using maps:
```
Erlang/OTP 26 [DEVELOPMENT] [erts-13.1.4] [source-c1dfdb4031] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Eshell V13.1.4 (press Ctrl+G to abort, type help(). for help)
1> erlang:info_timer().
#{#Ref<0.3315389924.154664961.229914> =>
      #{receiver => <0.51.0>,message => refresh_timeout,
        time_left => 3598516},
  #Ref<0.3315389924.154664961.230108> =>
      #{receiver => <0.85.0>,
        message => {<0.85.0>,write_cache},
        time_left => 638}}
```

max-au:
On the naming. There is already a function named `read_timer` (which I'd arguably want to be called `timer_info`).

And there is function named `processes()` that returns list of all processes in the system.

Should timers behave the same way? E.g. `erlang:timers()` BIF returning only the timer references, then read with `timer_info`?

In heavily loaded systems it's likely to have millions of timers. Reading all of them and returning all of them from a single BIF may take forever (and must be yielding, although just iterating should also be).

manuel-rubio:
> On the naming. There is already a function named `read_timer` (which I'd arguably want to be called `timer_info`).
> 
> And there is function named `processes()` that returns list of all processes in the system.
> 
> Should timers behave the same way? E.g. `erlang:timers()` BIF returning only the timer references, then read with `timer_info`?
> 
> In heavily loaded systems it's likely to have millions of timers. Reading all of them and returning all of them from a single BIF may take forever (and must be yielding, although just iterating should also be).

Hi @max-au thanks for your comment, as far as I could read, you prefer to get two functions instead of only one in the same way as you can get ports or processes, right? 

About yielding, not sure if that's something it should be done from the C code or if modifying the code to put the iteration (recursion) in Erlang code is good enough.

About the examples, I think I was explicit about what I wanted to get, as you can see in the description. I guess that with the new way you are proposing this could be something like:

```
1> [Timer1 | _] = erlang:timers().
[#Ref<0.3315389924.154664961.229914>,#Ref<0.3315389924.154664961.230108>]

2> erlang:timer_info(Timer1).
#{receiver => <0.51.0>,message => refresh_timeout, time_left => 3598516}
```

Right?

rickard-green (changes requested):
As mentioned by @max-au functionality like this **must** be yielding since there might be a huge amount of timers in the system.

Current solution also access the timer structures in a thread unsafe way which is a no go (I've made a comment where this is done). The solution is to send requests to schedulers and let each scheduler collect information and send it back to the process making the `timer_info()` call as ordinary messages and let that process put together the result.

> The problem with the timers is that we have no information about how many are running and we need to know the reference for retrieving data and canceling these timers.

In general the program using the timers are expected to distribute these references to process that are to retrieve information about timers. This especially when it comes to cancelling timers. It sounds like a recipe for disaster to lookup timers in some global timer storage and cancel timer them. This should be handled by the process setting the timer or another process that it has communicated the timer reference with.

I've only taken a quick peek at the code and we have not looked into what we think of the usefulness of this as such yet, so in order for you not to do work that later will be rejected please wait for us to have a closer look at the idea as such first before putting a lot of work into this.

rickard-green:
We do not want to include this functionality. Cancelling timers that you've retrieved references for in an API like will just introduce potential for new errors and is not something that we want to introduce. It could *perhaps* be interesting the get overall information about amount of BIF timers used or similar, though, but we see no need to expose all details of all timers for that. Also timers can be implemented in a lot of ways, e.g. a timer service of your own like the `timer` module do, so you will never be able to get information about all timers in a system.

PR #6721, opened 2023-01-23
Fixes `ftp:nlist/2` returning `{error, epath}` for existing file

kikofernandez:
`ftp:nlist(Pid, "file")` was returning `{error, epath}` for an existing file.
This PR fixes this.


KennethL:
What made it go wrong before and what is the logic of the correction?

kikofernandez:
- before it was going wrong because there was a pattern match case from the `ftp` protocol that was not handled .
That missing case made the `ftp:nlist/1/2` to return sometimes `ok`, when we expect `{ok | error, term()}`. (missing code from a previous PR (#6391 ) here)

```erlang
handle_ctrl_result({pos_compl, _}=Operation, #state{caller = {handle_dir_result, Dir},
                                                    data   = Data}= State) ->
    handle_ctrl_result(Operation, State#state{caller = {handle_dir_result, Dir, Data}});
```

- At the time, I thought that we had to make a difference between knowing whether something is a folder or not. This is not needed at all and simplifies the logic.

- We were missing test cases for non-existing folder / files, so the output was not consistent, i.e., sometimes we may be returning `{error, epath}`, and some other times `{ok, []}` to indicate in both cases that a file does not exist and the documentation that we have does not specify what we return in case of an error. To make the behaviour of `ftp:nlist/ls` consistent, simply removing the code mentioned in this PR makes the behaviour consistent for `ftp:ls/nlist` when a file/folder does not exist. If that ever happens, we know have tests that will indicate if this contract is broken.

IngelaAndin (approved):

PR #6724, opened 2023-01-23
Add options to inet:i functions

manuel-rubio:
When I go into the Erlang shell and I run `inet:i/0` I can see something like this:

```
2> inet:i().
Port Module   Recv Sent Owner    Local Address Foreign Address State Type   
24   inet_tcp 0    0    <0.88.0> *:pda-gate    *:*             LN:O  STREAM 
ok
```

And I ask myself, what was `pda-gate`?!? ... I have to go to `/etc/services` and search to find that is the port 4012:

```
1> gen_tcp:listen(4012, []).                                   
{ok,#Port<0.3>}
```

When I run `netstat` in the operating system it provides the `-n` option that's very helpful because most of the time I'm placing different web services in different ports and those are not related to the name stored in the services file. I was thinking and I've provided a way for solving this:

```
3> inet:i([show_numbers]).
Port Module   Recv Sent Owner    Local Address Foreign Address State Type   
24   inet_tcp 0    0    <0.88.0> 0.0.0.0:4012  0.0.0.0:*       LN:O  STREAM 
ok
```

If we are using `inet:i/1` and provide a list of elements, those are considered options. There's also a new `inet:i/3` function where the last parameter is the options. The options are:

- `show_numbers` is the capability to avoid changing IP or port numbers to names.
- `proto` we can specify the proto to be displayed (all of them by default: `tcp`, `udp`, and `sctp`).
- `tcp` only show TCP connections.
- `udp` only show UDP connections.
- `sctp` only show SCTP connections.
- `fields` is a list of fields to show in the output.

### DIFF
--- a/lib/kernel/src/inet.erl
+++ b/lib/kernel/src/inet.erl
@@ -43,7 +43,7 @@
 -export([udp_module/1, tcp_module/1, tcp_module/2, sctp_module/1]).
 -export([gen_tcp_module/1, gen_udp_module/1]).
 
--export([i/0, i/1, i/2]).
+-export([i/0, i/1, i/2, i/3]).
 
 -export([getll/1, getfd/1, open/8, open_bind/8, fdopen/6]).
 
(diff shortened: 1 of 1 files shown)


bmk:
There is already an 'options' argument (the second argument) when calling inet:i/2,
which makes this a bit confusing.
And they seem to be doing similar things (for instance "tcp show only TCP connections").

manuel-rubio:
> There is already an 'options' argument (the second argument) when calling inet:i/2, which makes this a bit confusing. And they seem to be doing similar things (for instance "tcp show only TCP connections").

The options help with filtering but the main point is avoiding translating ports to the service names.

IngelaAndin:
FYI: This needs some investigation and attention from the OTP technical board. It has been stalled to after OTP-26

bmk:
A solution for this has been merged to master and will be released with 27.0

PR #6731, opened 2023-01-25
JIT: Miscellaneous optimizations

bjorng:
Optimize `min/2`, `max/2`, and relational operators for both x86_64 and AArch64. Also do a few other optimizations for AArch64.

### DIFF
--- a/erts/emulator/beam/jit/arm/instr_guard_bifs.cpp
+++ b/erts/emulator/beam/jit/arm/instr_guard_bifs.cpp
@@ -192,27 +192,10 @@ void BeamModuleAssembler::emit_bif_is_ne_exact(const ArgRegister &LHS,
     }
 }
 
-void BeamModuleAssembler::emit_bif_is_ge(const ArgSource &LHS,
-                                         const ArgSource &RHS,
-                                         const ArgRegister &Dst) {
-    if (always_small(LHS) && RHS.isSmall() && RHS.isImmed()) {
-        auto lhs = load_source(LHS, ARG1);
-
-        comment("simplified compare because one operand is an immediate small");
-        cmp(lhs.reg, RHS.as<ArgImmed>().get());
-        emit_cond_to_bool(arm::CondCode::kGE, Dst);
-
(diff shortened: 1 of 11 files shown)


erts/emulator/beam/jit/arm/instr_guard_bifs.cpp
```diff
@@ -1001,6 +987,88 @@ void BeamModuleAssembler::emit_bif_map_size(const ArgLabel &Fail,
...
+    auto [lhs, rhs] = load_sources(LHS, ARG1, RHS, ARG2);
+    auto dst = init_destination(Dst, ARG1);
+    bool both_small = always_small(LHS) && always_small(RHS);
+    bool need_generic = !both_small;
+    Label generic = a.newLabel(), next = a.newLabel();
+
+    if (both_small) {
+        comment("skipped test for small operands since they are always small");
+    } else if (always_small(RHS) &&
+               always_one_of(LHS, BEAM_TYPE_INTEGER | BEAM_TYPE_MASK_BOXED)) {
```
jhogberg, line 1009: `BEAM_TYPE_MASK_BOXED` includes all types that may be boxed, including some that may be immediates like pids and ports. You probably want `BEAM_TYPE_MASK_ALWAYS_BOXED` here.
```suggestion
               always_one_of(LHS, BEAM_TYPE_INTEGER | BEAM_TYPE_MASK_ALWAYS_BOXED)) {
```

Looking further into it, I see we have the same problem in `emit_is_lt` et al on `maint`. I'll make a PR that fixes that and uses `enum class` to prevent us from running into this problem again.

jhogberg (changes requested):

erts/emulator/beam/jit/arm/instr_guard_bifs.cpp
bjorng: Fixed.

bjorng:
Rebased on latest `master`.

jhogberg (approved):

PR #6733, opened 2023-01-25
jit: Fix type hint error in emit_is_lt

jhogberg:
Fixes a bug we noticed while reviewing #6731

### DIFF
--- a/erts/emulator/beam/jit/arm/instr_common.cpp
+++ b/erts/emulator/beam/jit/arm/instr_common.cpp
@@ -1362,8 +1362,10 @@ void BeamModuleAssembler::emit_is_lt(const ArgLabel &Fail,
         comment("skipped test for small operands since they are always small");
         a.cmp(ARG1, ARG2);
         a.b_ge(resolve_beam_label(Fail, disp1MB));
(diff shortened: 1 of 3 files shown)


erts/emulator/test/type_hint_SUITE.erl
```diff
@@ -0,0 +1,59 @@
...
+
+init_per_group(_GroupName, Config) ->
+    Config.
+
+end_per_group(_GroupName, Config) ->
+    Config.
+
+
+compare(Config) when is_list(Config) ->
+    [compare_integer_pid(1 bsl N) || N <- lists:seq(1, 64)],
```
bjorng, line 50:
```suggestion
    _ = [compare_integer_pid(1 bsl N) || N <- lists:seq(1, 64)],
```

bjorng (approved):
The correction looks good.

Maybe with don't need a new test module for this test? We already have `op_SUITE` with test of operators, and there already are additional test cases in that module to test the relational operators (including the use of type hints). I think it is better to put the new test in `op_SUITE`.

jhogberg:
Sure thing, thanks :)

PR #6736, opened 2023-01-25
Reduce contention on the code_server

josevalim:
load_file/1, load_abs/2, and load_binary/3 now
perform most of the work on the client.

is_sticky/1 and is_loaded/1 now read directly
from the module database.

PS: I could not move ensure_loaded/1 to the
client for reasons I could not understand yet.

### DIFF
--- a/lib/kernel/src/code.erl
+++ b/lib/kernel/src/code.erl
@@ -186,7 +186,7 @@ objfile_extension() ->
 load_file(Mod) when is_atom(Mod) ->
     case get_object_code(Mod) of
         error -> {error,nofile};
-        {Mod,Binary,File} -> load_module(Mod, File, Binary, false)
+        {Mod,Binary,File} -> load_module(Mod, File, Binary, false, false)
     end.
 
 -spec ensure_loaded(Module) -> {module, Module} | {error, What} when
@@ -195,7 +195,12 @@ load_file(Mod) when is_atom(Mod) ->
 ensure_loaded(Mod) when is_atom(Mod) ->
     case erlang:module_loaded(Mod) of
         true -> {module, Mod};
-        false -> call({ensure_loaded,Mod})
+        false ->
+            case get_object_code(Mod) of
+                error -> call({sync_ensure_on_load, Mod});
+                {Mod,Binary,File} ->
+                    load_module(Mod, File, Binary, false, true)
+            end
     end.
 
 %% XXX File as an atom is allowed only for backwards compatibility.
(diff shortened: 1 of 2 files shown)


bjorng:
A thought about `ensure_loaded`: If two processes calls `code:ensure_loaded/1` for the same module, the module must only be loaded once. That is, when the code server has done `finish_loading` for the first request to load the module, it must discard the second request request to load the module. One way to implement that would be to compare the MD5 for the module. Not sure if that was the actual issue you encountered.

bjorng:
Have you observed actual contention caused by the functions you have optimized? (That is, will this optimization make a meaningful difference?)

When running in interactive mode, I would expect most code loading being caused by a call to an unloaded module and that the `error_handler` module would then call `code:ensure_loaded/1` to load the module in question. I could be wrong.

josevalim:
@bjorng no, I didn't measure anything. My thoughts are that the approach is generally preferable in terms of design. In particular:

* `is_sticky` and `is_loaded` are positive changes with no drawbacks (perhaps we should even enable read_concurrency in the table?)

* Moving `load_abs` and `load_binary` to the client have no harm - but may not provide any benefit

* `load_file` though needs to talk to the code server twice, passing the .beam binary around, so I understand if we would prefer to not change it

Do you think those are fair?

I will share what I did for `ensure_loaded` later.

josevalim:
@bjorng here is the commit for `ensure_loaded`:

https://github.com/josevalim/otp/commit/8f7b6365b042070b521de122868fccd8d2b836c4

and here is the test failure I get:

```
code_SUITE:on_load_binary failed on line 1304
Reason: {badmatch,{undef,[{on_load_binary,ok,[],[]},
        {code_S...}
```

bjorng:
I am only worried about potential unintended consequences because of lack serialization of the request.

Moving relatively cheap calls such as `is_sticky` and `is_loaded` to the client might not be worth it.

Loading potentially very expensive calls such as the call to `prepare_loading` is probably worth moving to the client (possibly even if there will be two calls to the server).

I will try to take a look at why `ensure_loaded` does not work tomorrow. I think that there is a much greater potential for more concurrency if that would work.

josevalim:
> Moving relatively cheap calls such as is_sticky and is_loaded to the client might not be worth it.

Can you think of a scenario where it would introduce bugs? Given the writes are serialized, we will always read our own writes. The only race would be if two different processes are writing and reading at the same time, but then the result is not guaranteed in any way.

In any case, you are right in that they are cheap and the benefit in moving them is minimal. I will be glad to amend it in any way desired.

frazze-jobb:
I have now taken a look at this new ensure loaded function and fixed the issue. 
The issue was that the second process needed to wait for the on_load function to finish before it can use the module.
I added a call in ensure loaded that asks the code_server if the module has an on_load function pending. If so wait for that to finish.

frazze-jobb:
Merged, thanks @josevalim  for your contribution!

PR #6737, opened 2023-01-26
Start children applications concurrently

josevalim:
We build a DAG with all application dependencies
and start leaves recursively. As each application
starts, we further traverse the graph looking for
more leaves.

This PR always traverse children applications
concurrently by default. Let me know if there is
an interest in making this setting configurable.

### DIFF
--- a/lib/kernel/doc/src/application.xml
+++ b/lib/kernel/doc/src/application.xml
@@ -69,12 +69,21 @@
     <func>
       <name name="ensure_all_started" arity="1" since="OTP R16B02"/>
       <name name="ensure_all_started" arity="2" since="OTP R16B02"/>
-      <fsummary>Load and start an application and its dependencies, recursively.</fsummary>
+      <name name="ensure_all_started" arity="3" since="OTP 26.0"/>
+      <fsummary>Loads and starts all applications and their dependencies, recursively.</fsummary>
       <desc>
-        <p>Equivalent to calling
+        <p><c><anno>Applications</anno></c> is either an an <c>atom()</c> or a list
+            of <c>atom()</c> representing multiple applications.</p>
+        <p>This function is equivalent to calling
         <seemfa marker="#start/1"><c>start/1,2</c></seemfa>
-            repeatedly on all dependencies that are not yet started for an application.
+            repeatedly on all dependencies that are not yet started of each application.
             Optional dependencies will also be loaded and started if they are available.</p>
+        <p>The <c><anno>Mode</anno></c> argument controls if the applications should
+            be started in <c>serial</c> mode (one at a time) or <c>concurrent</c> mode.
+            In concurrent mode, a dependency graph is built and the leaves of the graph
+            are started concurrently and recursively. In both modes, no assertion can be
+            made about the order the applications are started. If not supplied, it defaults
+            to <c>serial</c>.</p>
         <p>Returns <c>{ok, AppNames}</c> for a successful start or for an already started
             application (which is, however, omitted from the <c>AppNames</c> list).</p>
         <p>The function reports <c>{error, {AppName,Reason}}</c> for errors, where
--- a/lib/kernel/src/application.erl
+++ b/lib/kernel/src/application.erl
@@ -19,7 +19,8 @@
 %%
 -module(application).
 
--export([ensure_all_started/1, ensure_all_started/2, start/1, start/2,
+-export([ensure_all_started/1, ensure_all_started/2, ensure_all_started/3,
+	 start/1, start/2,
 	 start_boot/1, start_boot/2, stop/1, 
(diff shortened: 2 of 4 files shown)


josevalim:
Also note that this has no impact on releases at the moment. I am not well-versed on reltool/systool to make a contribution there. Perhaps `ensure_all_started` can concurrently start apps by default and the actual configuration goes to reltool/systool.

Suggestions on how to test this feature is welcome, if any tests at all are necessary.

josevalim:
Apologies, one last comment: if this direction is accepted, I would also propose for `ensure_all_started` to also accept a list of applications. The return type signature would be the same, so I think it would be an acceptable change.

josevalim:
After some thought, I think this functionality should be configurable (either opt-in or opt-out), because I can see disabling it being convenient for debugging. Therefore, I think the following questions apply:

1. What is the API? Do we want to make it `ensure_all_loaded(Nodes, Types)` where Types is `[permanent | temporary | concurrent | serial]` or two separate functions?

2. If it is a single function, what is the default?

3. Implementation wise, we can keep two completely separate implementations, or is it ok to build the DAG on both serial and concurrent cases and then just change how we traverse the DAG? In theory doing so for the serial mode would be a bit slower than today's implementation but I don't think it would be measurable. The benefit of sharing is that they will be less variation in the order dependencies are started between both modes (apart from concurrency).

frazze-jobb:
Since concurrent / serial is orthogonal to the restart types I think they should be a separate argument, where the serial would be the default. And you can run it concurrently with ensure_all_loaded(Apps, RestartType, Mode = concurrent).
I think its okey that they share the implementation between the two modes unless the additional cost is too great, then we will pull in the old implementation back in.
Maybe that is safe enough, and we can change it to concurrent by default in the future, if it is requested.

josevalim:
@frazze-jobb cool, I will address this today or tomorrow the latest.

josevalim:
I have amended the PR to define `application:ensure_all_started/3` that receives the `Mode` as third argument. It defaults to `serial`.

josevalim:
Thanks @stolen! Do you use releases? If so, how do you handle said cases? Because a release does not guarantee an order for application starts either.

@frazze-jobb let me know if you want me to partially revert this and keep the old code for serial starts.

stolen:
No, we don't use releases for start up.  
Rebar3 and relx are used to build a bundle of libs, but the startup is done via special module which performs some preparation and thel calls `application:ensure_all_started(toplevel_app, permanent)`.

frazze-jobb:
@stolen could you elaborate what breaks in OTP 25? I think the best way forward is to add the dependencies.

stolen:
@frazze-jobb Checked it again. TL;DR: **OTP 25 is OK**, it is our CI script failure.  

At first I saw the failed CI job with some application error and just dropped the commit.  

Turned out the problem was not in the main app but in our CI script.  
In short:
  * `toplevel_app` depends on `[communicate_app, storage_app]` (in this order)
  * `communicate_app` requires extra preparation, which is done by custom boot script
  * fully working `storage_app` relies on `communicate_app` — easily fixed by dependency
  * In CI, `check_storage.escript` starts `storage_app` to use its utility functions (`communicate_app` not needed)
  * `communicate_app` fails to start now because no preparation was done
  * the whole CI job fails

So, OTP 26 forces to write dependencies properly, and specified dependencies raise problems with other dependencies.  

Sorry for bothering you.

stolen:
@frazze-jobb Sorry for the late reply.  
`optional_applications` do not change anything.

I created a demo of start failures. See it here: https://github.com/stolen/otp-app-deps  

As for real application, I 'fixed' the issue by explicitly starting the communicate_app before starting the toplevel_app. It works for now and does not need attention for the nearest couple of months.

frazze-jobb:
I miss interpreted the documentation, I thought optional applications would help, but it says that the dependency must be in applications to ensure that its started before the app. It doesn't matter if a "parent" application has the dependency in its applications list. So according to documentation its working as stated.
I don't know why it has been designed that way.
Its good that you found a workaround in your case.

josevalim:
>  I don't know why it has been designed that way.

For two reasons:

1. One application may be optional but not listed in `applications` (which means it is an optional dependency that does not need to be started before)

2. Some tools preserve app order, so we need to have a single ordering

More info back on #2675. :)

josevalim:
@frazze-jobb definitely.

FWIW, I pinged @tsloughter about the app ordering and he confirmed that relx does not rely on ordering today. In any case, we probably want to avoid changing the behavior if we can, and I have a patch that keeps the behavior and is actually fewer LOC!

josevalim:
Done in https://github.com/erlang/otp/pull/6958.

PR #6740, opened 2023-01-26
Update io_protocol.xml

ityonemo:
The documentation appears to be incorrect, judging from the following LOC:

https://github.com/erlang/otp/blob/a213a9a9541731f1f68a85d9cc14af9535d9be14/lib/stdlib/src/io.erl#L148-L154


bjorng:
Please squash your commits.

bjorng:
Thanks for noticing the incorrect documentation. I have fixed it in the `master` branch, using a slightly different wording.

PR #5048, opened 2021-07-12
Update the index.html with some up to date info regarding editors, books etc.

KennethL:

PR #6745, opened 2023-01-27
Update index.html.src

KennethL:
Update the entry page for the documentation. Will come more before OTP 26

### DIFF
--- a/system/doc/top/templates/index.html.src
+++ b/system/doc/top/templates/index.html.src
@@ -178,9 +178,9 @@ tags support and more. See the <a href="#tools#/index.html">
 Tools</a> application for details.  
 <p>
 There are also Erlang plugins for other code editors
-<a href="http://github.com/vim-erlang">Vim (vim-erlang)</a> ,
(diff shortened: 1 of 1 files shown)


system/doc/top/templates/index.html.src
```diff
@@ -200,14 +199,12 @@ module and function name completion (tab) if the module is loaded.
 <p>
 
 <li>OpenSource users can ask questions
-and share experiences on the <a href="http://www.erlang.org/static/doc/mailinglist.html">
-Erlang questions mailing list</a>.
+and share experiences on <a href="http://erlangforums.com">
```
max-au, line 202: Could you change it to `https` ?

Here, and pretty much everywhere else.

KennethL: will do that and more changes before OTP 26, thanks for the suggestion

KennethL: Closes #5048 as well

KennethL:
continue work in a new PR

Issue #6748, opened 2023-01-27
Unknown posix errors hard to debug since error is thrown away


onno-vos-dev:
**Describe the bug**
Errors that boil down into [erl_posix_msg.erl ](https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_posix_msg.erl#L175) are hard to debug since the error reason is thrown away. 

**Expected behavior**
I'd expect the error reason to be included in the `unknown POSIX error` error to aid in debugging. 

**Affected versions**
OTP26-rc0 and everything below. 

**Additional context**
This issue was discovered as part of publish docs for [aws-beam/aws-erlang](https://github.com/aws-beam/aws-erlang) described in https://github.com/hexpm/hex_core/issues/136

PR #6749, opened 2023-01-27
[#6748] Unknown posix errors should include the error reason

onno-vos-dev:
Fixes https://github.com/erlang/otp/issues/6748 

I updated the PR in `hex_core` to not leak the `too_big` error down into `erl_posix_msg.erl` but I believe the PR still stands. Someone else may end up in a similar scenario and end up requiring to patch OTP in order to figure out where the error is originating from.


bjorng (approved):
Seems reasonable. Added to our daily builds.

bjorng:
The test case `io_SUITE:error_info/1` failed in our daily build.

I have force-pushed a correction to your branch to fix that problem, as well as fixes to the same problem in other places not caught by any test case.

bjorng:
Thanks for your pull request.

PR #6751, opened 2023-01-27
erl_nif.xml: fix typo in documentation

max-au:


max-au:
Err, documentation was correct. But parsing it is, uhm, challenging.

Issue #6730, opened 2023-01-24
ets:insert/2 for bag table stores values for same key in reversed order


madlep:
**Describe the bug**
When a list of values are passed to `ets:insert/2` for a `bag` table type, and there are multiple values for the same key, they are incorrectly stored in reversed order. In OTP versions prior to R23, they were stored in the order passed.

We are upgrading from OTP 22 to OTP 25, and have logic that relies on insert order being preserved.

**To Reproduce**
Short code to reproduce:
```erlang
-module(ets_insert_reversed_regression).

-export([example/0, example_multiple_insert/0]).

example() ->
  Tab = ets:new(my_table, [bag]),

  OriginalValues = [{a, 1}, {a, 2}, {a, 3}],
  ets:insert(Tab, OriginalValues),
  ReturnedValues = ets:lookup(Tab, a),

  io:fwrite("Expected =~w~n", [OriginalValues]),
  io:fwrite("Actual   =~w~n", [ReturnedValues]).

example_multiple_insert() ->
  Tab = ets:new(my_table, [bag]),

  OriginalValues1 = [{a, 1}, {a, 2}, {a, 3}],
  ets:insert(Tab, OriginalValues1),

  OriginalValues2 = [{a, 4}, {a, 5}, {a, 6}],
  ets:insert(Tab, OriginalValues2),

  ReturnedValues = ets:lookup(Tab, a),

  io:fwrite("Expected =~w~n", [OriginalValues1 ++ OriginalValues2]),
  io:fwrite("Actual   =~w~n", [ReturnedValues]).
```

**Correct** original behaviour running on OTP 22 - note ordering of values for key `a` is preserved:
```
Erlang/OTP 22 [erts-10.7.2.18] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1]

Eshell V10.7.2.18  (abort with ^G)
1> c(ets_insert_reversed_regression).
{ok,ets_insert_reversed_regression}
2> ets_insert_reversed_regression:example().
Expected =[{a,1},{a,2},{a,3}]
Actual   =[{a,1},{a,2},{a,3}]
ok
3> ets_insert_reversed_regression:example_multiple_insert().
Expected =[{a,1},{a,2},{a,3},{a,4},{a,5},{a,6}]
Actual   =[{a,1},{a,2},{a,3},{a,4},{a,5},{a,6}]
ok
```

[...]

PR #6752, opened 2023-01-28
ets:insert/2 insert order preserved when list with same keys for bag table

madlep:
Since OTP 23, when inserting a list of values into an ets `bag` table, values for the same key are inserted in reversed order (prior to OTP 22, they were inserted in the order in the list).

Implementing fix suggested by @sverker to build linked list of `DbTerm` values stored against a key by _appending_ (rather than _prepending_ as it had been doing).

Fix for https://github.com/erlang/otp/issues/6730 (see description there for code to reproduce + original correct behaviour on OTP 22, and regression behaviour on OTP >= 23)

### DIFF
--- a/erts/emulator/beam/erl_db_tree.c
+++ b/erts/emulator/beam/erl_db_tree.c
@@ -519,7 +519,7 @@ DbTableMethod db_tree =
     db_lookup_dbterm_tree,
     db_finalize_dbterm_tree,
     db_eterm_to_dbterm_tree_common,
-    db_dbterm_list_prepend_tree_common,
+    db_dbterm_list_append_tree_common,
     db_dbterm_list_remove_first_tree_common,
     db_put_dbterm_tree,
     db_free_dbterm_tree_common,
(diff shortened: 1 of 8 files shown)


erts/emulator/beam/erl_db_tree.c
```diff
@@ -3533,11 +3533,11 @@ void* db_eterm_to_dbterm_tree_common(int compress, int keypos, Eterm obj)
     return term;
 }
 
-void* db_dbterm_list_prepend_tree_common(void *list, void *db_term)
+void* db_dbterm_list_append_tree_common(void *last_term, void *db_term)
```
madlep, line 3536: Not sure if/where this code is even used, but have modified it to be consistent with `erl_db_hash` for consistency.

sverker: It's for `ordered_set` tables.

sverker:
I pushed a commit with some documentation improvements.
@madlep @gomoripeti What do you think?

madlep:
> I pushed a commit with some documentation improvements. @madlep @gomoripeti What do you think?

makes sense to me 👍

sverker:
Expanded the test case to include provoking bug for `duplicate_bag`.
Squashed into two commits and merged.

@madlep Thanks for good report and fix.

PR #6757, opened 2023-01-30
kernel: `gen_tcp:connect/3` should not accept empty strings

### LINKED ISSUE #6353: gen_tcp:connect/3 raises an exception for empty string addresses
**Describe the bug**
I was running some dynamic tests with HTTP requests using hackney and realized after some debugging that gen_tcp.connect with String address might respond in 3 different ways:
1. Success tuple in case you set a valid URL like: `'something.com'`
2. Error tuple in case you set a string but invalid URL like: `'something'`
3. Raises an exception with badarg in case you set an empty string like: `''` 

the third response doesn't feel right, sometimes you might get a misconfiguration issue just like I had, and it will cost you some debugging time to understand what happened, while it could simply say `{error, nxdomain}` instead of raising an Exception badarg with no details.

I was digging into the gen_tcp:connect/3 and I found out that it happens because the `inet_parse:visible_string/1` [here](https://github.com/erlang/otp/blob/master/lib/kernel/src/inet.erl#L1514)

**Steps to reproduce the behavior.**
```
> gen_tcp:connect('something.com', 80, []).
{ok, #Port<...>}

> gen_tcp:connect('something', 80, []).
{error, nxdomain}

> gen_tcp:connect('', 80, [])
** exception exit: badarg
     in function  gen_tcp:connect/4 (gen_tcp.erl, line 171)
```

**Expected behavior**
empty string also classifies as an invalid URL, I guess it should respond with an error tuple the same as below.
```
> gen_tcp:connect('something', 80, []).
{error, nxdomain}

> gen_tcp:connect('', 80, [])
{error, nxdomain}
```

**Affected versions**
I'm using Erlang/OTP 24, but I believe its happening not only on this version.

**Additional context**
I'd like to understand better if this is the best approach of dealing with an empty string, but I believe that returning an error tuple might be more friendly. 

If so, can I work on a PR for that?
I guess that we could simply accept an empty list/string at `inet:getaddrs_tm/3` and dispatch `inet:gethostbyname_tm/3` to return `{error, nxdomain}`

kikofernandez:
- This PR changes the type spec of the `gen_tcp:connect/3` to specify that its first argument cannot contain an empty string. In this case, that means replacing `inet:hostname()` which is equivalent to `atom() | string()` by `atom() | nonempty_string()`.
- The PR also adds a test case that checks that calling the `gen_tcp:connect/3` with an empty string as first argument throws a `badarg` error and this is the expected behaviour.
- Closes #6353

### DIFF
--- a/lib/kernel/src/gen_tcp.erl
+++ b/lib/kernel/src/gen_tcp.erl
@@ -161,7 +161,7 @@ connect(SockAddr, Opts) ->
     connect(SockAddr, Opts, infinity).
 
 -spec connect(Address, Port, Opts) -> {ok, Socket} | {error, Reason} when
-      Address  :: inet:socket_address() | inet:hostname(),
+      Address  :: inet:socket_address() | nonempty_string() | atom(),
       Port     :: inet:port_number(),
       Opts     :: [inet:inet_backend() | connect_option()],
       Socket   :: socket(),
@@ -202,7 +202,7 @@ connect(#{family := Fam} = SockAddr, Opts, Timeout)
 
 -spec connect(Address, Port, Opts, Timeout) ->
                      {ok, Socket} | {error, Reason} when
-      Address :: inet:socket_address() | inet:hostname(),
+      Address :: inet:socket_address() | nonempty_string() | atom(),
       Port    :: inet:port_number(),
       Opts    :: [inet:inet_backend() | connect_option()],
       Timeout :: timeout(),
--- a/lib/kernel/test/gen_tcp_api_SUITE.erl
+++ b/lib/kernel/test/gen_tcp_api_SUITE.erl
@@ -362,6 +362,9 @@ t_connect_bad(Config) when is_list(Config) ->
                                        ?INET_BACKEND_OPTS(Config)),
     io:format("Error for connection attempt to non-existing host: ~p",
 	      [Reason2]),
+
+    {'EXIT', badarg} = catch gen_tcp:connect("", 80,
+                                             ?INET_BACKEND_OPTS(Config)),
     ok.
 
 


lib/kernel/src/gen_tcp.erl
```diff
@@ -161,7 +161,7 @@ connect(SockAddr, Opts) ->
     connect(SockAddr, Opts, infinity).
 
 -spec connect(Address, Port, Opts) -> {ok, Socket} | {error, Reason} when
-      Address  :: inet:socket_address() | inet:hostname(),
+      Address  :: inet:socket_address() | nonempty_string() | atom(),
```
IngelaAndin, line 164: Should we not change `inet:hostname() `to -`type hostname() :: atom() | nonempty_string().` rather that changing this spec? @RaimoNiskanen ?

kikofernandez: We should, if that does not break anything :)

kikofernandez: Since changing the `inet:hostname()` may involve further changes to handle other cases, do we accept this simple change or not?

kikofernandez: AFAIK, empty string was not allowed but an empty atom is. 
There were some edge cases here...

RaimoNiskanen: I did some studying, acquired a new opinion, and wrote it in issue #6353, which was shortly that empty atom should work just like empty string, and since we cannot use the type language to exclude both we should allow both, hence we should not do this type change.

Furthermore I think that we should consistently change so that the empty host name is treated just like the host name "." i.e the root domain.  This would be consistent in the way that all other (non-empty) host names are always treated as absolute and have an implied dot at the end (in DNS's terminology - it is the way to denote an absolute host name in a DNS domain file and when quering to a DNS resolver).

For example `gen_tcp:connect/*` looks up an address for a host name, that is, an A or AAAA record.  And the root domain today has no address records (however e.g a SOA record).  So such a lookup should result in `{error, nxdomain}`, now, and probably in the future.  The right thing (tm) would be to do the lookup...

PR #6759, opened 2023-01-31
dialyzer: Do some minor optimizations and clean ups

bjorng:


lib/dialyzer/src/dialyzer_callgraph.erl
```diff
@@ -337,15 +337,15 @@ module_postorder_from_funs(Funs, #callgraph{digraph = DG,
...
-  catch
-    _:_ -> error
+  case ets:lookup_element(Table, Key, 2, error) of
+    error -> error;
+    Val -> {ok, Val}
   end.
 
 ets_lookup_set(Key, Table) ->
-  ets:lookup(Table, Key) =/= [].
+  ets:member(Table, Key).
```
kikofernandez, line 348: Why is this (`=/= []`) no longer needed?

bjorng: [ets:lookup/2](https://www.erlang.org/doc/man/ets.html#lookup-2) returns an empty list if the key is not present in the table, and otherwise a list with one or more elements. The original code tests whether the there is an object with that key in the table, which is exactly what `ets:member/2` does.

kikofernandez (approved):

lib/dialyzer/src/dialyzer_codeserver.erl
```diff
@@ -320,11 +314,8 @@ finalize_records(#codeserver{temp_records = TmpRecords,
 
 lookup_mod_contracts(Mod, #codeserver{contracts = ContDict})
   when is_atom(Mod) ->
-  case ets_dict_find(Mod, ContDict) of
-    error -> maps:new();
-    {ok, Keys} ->
-      maps:from_list([get_file_contract(Key, ContDict)|| Key <- Keys])
-  end.
+  Keys = ets:lookup_element(ContDict, Mod, 2, []),
+  maps:from_list([get_file_contract(Key, ContDict)|| Key <- Keys]).
```
jhogberg, line 318: Nitpick:
```suggestion
  maps:from_list([get_file_contract(Key, ContDict) || Key <- Keys]).
```

lib/dialyzer/src/dialyzer_coordinator.erl
```diff
@@ -135,22 +136,20 @@ parallel_job(Mode, Jobs, InitData, Timing) ->
...
               ok
           end
-      catch
-        _:_ ->
-          %% Already finished.
-          ok
-      end
-    end,
+      end,
   lists:foreach(F, Labels).
```
jhogberg, line 153: Why not get rid of the fun while we're at it? It seems a bit redundant.
```erlang
wait_for_success_typings(Labels, {_Collector, _Regulator, JobLabelsToPid}) ->
    wait_for_success_typings_1(Labels, JobLabelsToPid).

wait_for_success_typings_1([JobLabel | Labels], JobLabelsToPid) ->
   %% The jobs that job depends on have always been started.
    case ets:lookup_element(JobLabelsToPid, JobLabel, 2, ok) of
        Pid when is_pid(Pid) ->
            Ref = erlang:monitor(process, Pid),
            receive
                {'DOWN', Ref, process, Pid, _Info} -> ok
            end;
        ok ->
            %% Already finished.
            ok
    end,
    wait_for_success_typings_1(Labels, JobLabelsToPid);
wait_for_success_typings_1([], _JobLabelsToPid) ->
    ok.
```

jhogberg (approved):
LGTM

lib/dialyzer/src/dialyzer_coordinator.erl
bjorng: I don't see any particular problem with the fun. The host function is not very large and the fun is not too deeply indented (which are things that annoy me).

If I had written the function myself from scratch, I would probably have written something like this (using a list comprehension only for its side effects, which I know that not everybody approves of):

```
wait_for_success_typings(Labels, {_Collector, _Regulator, JobLabelsToPid}) ->
    _ = [wait_for_success_typing(Label, JobLabelsToPid) || Label <- Labels],
    ok.
```

PR #6760, opened 2023-01-31
Simplify some calls to application:get_env/2

bjorng:
Replace some calls to `application:get_env/2` with `application:get_env/3`. This simplifies the code at the call site, and as an additional benefit, `application:get_env/3` is slightly more efficient than `application:get_env/2`.


frazze-jobb:
Looks good

PR #6762, opened 2023-01-31
Simplify implementation of maps:to_list/1 for iterators

Maria-12648430:
`maps:to_list` for iterators can be written in a much simpler fashion by simply _trying_ the given iterator with `next` (see https://github.com/erlang/otp/pull/6718#discussion_r1091660046).

### DIFF
--- a/lib/stdlib/src/maps.erl
+++ b/lib/stdlib/src/maps.erl
@@ -231,31 +231,17 @@ take(_,_) -> erlang:nif_error(undef).
 
 to_list(Map) when is_map(Map) ->
     to_list_internal(erts_internal:map_next(0, Map, []));
-to_list({_K, _V, _NextIterator} = Iterator) ->
-    try to_list_from_iterator(Iterator) of
-        Result -> Result
+to_list(Iter) ->
+    try to_list_from_iterator(next(Iter))
     catch
-        error:badarg ->
-            error_with_info({badmap, Iterator}, [Iterator])
-    end;
-to_list([PathOrKeys | Map] = Iterator)
-  when (is_integer(PathOrKeys) orelse is_list(PathOrKeys)), is_map(Map) ->
-    try to_list_from_iterator(Iterator) of
-        Result -> Result
-    catch
-        error:badarg ->
-            error_with_info({badmap, Iterator}, [Iterator])
-    end;
-to_list(Map) ->
-    error_with_info({badmap, Map}, [Map]).
-
-to_list_from_iterator(Iterator) ->
-    case maps:next(Iterator) of
-        {Key, Value, NextIterator} ->
-            [{Key, Value} | to_list_from_iterator(NextIterator)];
-        none ->
-            []
+        error:_ ->
+            error_with_info({badmap, Iter}, [Iter])
     end.
+
+to_list_from_iterator({Key, Value, NextIter}) ->
+    [{Key, Value} | to_list_from_iterator(next(NextIter))];
+to_list_from_iterator(none) ->
+    [].
 
 to_list_internal([Iter, Map | Acc]) when is_integer(Iter) ->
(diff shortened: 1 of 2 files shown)


Maria-12648430:
Btw...

```shell
This application needs to be tested using a released Erlang/OTP.
Erlang/OTP will now be released to /home/maria/contrib/otp/lib/stdlib/make_test_dir/Erlang ∅⊤℞.
```

Whose bright idea was the [∅](https://www.compart.com/en/unicode/U+2205)[⊤](https://www.compart.com/en/unicode/U+22A4)[℞](https://www.compart.com/en/unicode/U+211E), pray tell? :sweat_smile:

bjorng:
Looking a little bit closer at the code, I see that there will be an ugly crash for an iterator that will not crash the first call to `maps:next/1`:

```
1> maps:to_list({a,b,x}).
     in function  maps:next/1
        called as maps:next(x)
        *** argument 1: not a valid iterator
     in call from maps:to_list_from_iterator/1 (maps.erl, line 243)
```

Maria-12648430:
Ok, just a minute ;)

Maria-12648430:
There, better? :hugs:

bjorng:
Thanks, that works better. Since this was not caught by an existing test, do you think you could add a test for bad iterators?

Maria-12648430:
> do you think you could add a test for bad iterators?

You mean for non-iterators and almost-but-not-exactly-itorator-ish-things-that-fail-later-on? :grin:

Well, depends...
* for only `maps:to_list/1` --> sure :smile:
* for all `maps` functions that accept iterators --> yes, but outside of the scope of this PR?
* for iterators and their uses in general --> yes, but probably stuff for property testing?

Maria-12648430:
Ok... that was even easier than expected, just add stuff to the `error_info` case... _but_, in doing so, I noticed that most (all?) other functions that can use iterators (`map`, `filter`, ...) also only try the _first_ call to `next` and fail ugly on later-failing almost-iterators :cry:

Maria-12648430:
Then again... for functionslike `map` and `filter`, an distinction must be made between _errors caused by faulty iterators_ and _errors caused by the given function_, so other than with `to_list`, every subsequent call to `next` should be in its own `try` for these functions.

bjorng:
Thanks. Added to our daily builds.

Your other code is indeed material for another PR. Performance can be improved if the `try` / `catch` and the call to `next/1` is avoided when not needed in `try_next/2`:

```
try_next({_, _, _} = KVI, _ErrorRef) ->
    KVI;
try_next(none) ->
    none;
.
.
.
```

Another slight optimization would be to avoid creating the reference if the input is a map. (Move the `make_ref/0` call into the `if` and return `[]` for `ErrorRef` if the input is a map.)

(I'm taking the day off; back on Thursday.)

Maria-12648430:
> Thanks. Added to our daily builds.

:+1:

> Your other code is indeed material for another PR.

Ok, I'll make a draft just for `map` so we have something to talk about and play with, then when it is sound adapt it to the other functions.

> Another slight optimization would be to avoid creating the reference if the input is a map. (Move the `make_ref/0` call into the `if` and return `[]` for `ErrorRef` if the input is a map.)

Why an empty list? :thinking: Instinctively and out of habit, I would have reached for `undefined`.

Actually, if the input is a map, we can skip the `try`ing altogether, assuming that `iterator/1` (and subsequent calls to `next/1`) will always return a good iterator, so any errors must come from the mapping function, which should bubble up.

> (I'm taking the day off; back on Thursday.)

Enjoy, see you tomorrow :slightly_smiling_face:

Maria-12648430:
Follow-up in #6766 :smile:

bjorng:
> Why an empty list? 🤔 Instinctively and out of habit, I would have reached for `undefined`.

Yes, `undefined` is better.

PR #6763, opened 2023-01-31
Silent init error for gen_server and gen_statem / otp-18423

bmk:
The point of this extension is to provide a way for a user of gen_server | gen_statem to "silently"
(*no* automatic error reports) return an error during *init* (new return value from Module:init/1).
In essence, behave as 'ignore'  but provide a way to return an '{error, Reason}' (from the 
[gen_server|gen_statem]:start_link functions).

### DIFF
--- a/lib/stdlib/doc/src/gen_server.xml
+++ b/lib/stdlib/doc/src/gen_server.xml
@@ -1967,7 +1967,7 @@ format_status(Status) ->
           <item>
             <p>
               Initialization failed.
-	      An exit signal with</p>
+	      An exit signal with exit reason</p>
 	      <taglist>
 		<tag>stop:</tag>
 		<item><c>Reason</c></item>
--- a/lib/stdlib/doc/src/gen_server.xml
+++ b/lib/stdlib/doc/src/gen_server.xml
@@ -1916,6 +1916,7 @@ format_status(Status) ->
         <v>&nbsp;&nbsp;| {ok,State,hibernate}</v>
         <v>&nbsp;&nbsp;| {ok,State,{continue,Continue}}</v>
         <v>&nbsp;&nbsp;| {stop,Reason}</v>
+        <v>&nbsp;&nbsp;| {shutdown,Reason}</v>
         <v>&nbsp;&nbsp;| ignore</v>
         <v>&nbsp;State = term()</v>
         <v>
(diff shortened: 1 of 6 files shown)


lib/stdlib/src/gen_statem.erl
```diff
@@ -225,8 +228,9 @@
     {ok, State :: StateType, Data :: DataType} |
     {ok, State :: StateType, Data :: DataType,
      Actions :: [action()] | action()} |
```
RaimoNiskanen, line 230: Fix the indentation of these since you change the lines below (?)

RaimoNiskanen:
For the callback returns with `stop` it is `Reason = shutdown | {shutdown,term()}` that avoids the `error_logger`.

This implies (or?) that it should be the same for the `init` callback, i.e returning `{stop, shutdown} | {stop, {shutdown, term()}}` should avoid the `error_logger`.  This would not introduce a new return term but could potentially change the behaviour of existing code (which could be regarded as a bug fix).

The documentation for `gen_statem:start_link/4` talks about the return value from `Module:init/1` and should probably also be updated concerning the special treatment of a `shutdown` tuple.  There may be more places.  The behaviours user's guide in system documentation, for example...

bmk:
I suppose that we could view the current situation as a documentation-bug.
That is, the special case of {stop, {shutdown, Reason}} is not documented.

bmk:
Another issue with using stop (with {shutdown, Reason}) is that the return from start|start_link
is that the return will be {error, {shutdown, Reason}}, so the shutdown part leaks through.
The original problem was "excessive" error reporting.

bmk:
Use '{error, Reason}' instead of '{shutdown, Reason}' and behave like 'ignore'!

lib/stdlib/doc/src/gen_statem.xml
```diff
@@ -2510,6 +2512,9 @@ handle_event(_, _, State, Data) ->
           <c>Module:init/1</c> returns <c>ignore</c>) is set to linked processes
           and ports, including the process calling <c>start_link/3,4</c>.
 	</p>
+	<p>The difference between returning <c>{stop, Reason}</c> and
+	<c>{shutdown, Reason}</c> (from <c>Module:init/1</c>) is that
+	'shutdown' results in a "silent" termination. </p>
```
rickard-green, line 2517:
```suggestion
	<c>shutdown</c> results in a "silent" termination. </p>
```

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -1960,14 +1961,22 @@ format_status(Status) ->
...
+            <c>{shutdown,Reason}</c><br/>
             <c>ignore</c>
           </tag>
           <item>
             <p>
               Initialization failed.
-              An exit signal with this <c>Reason</c>
-              (or with reason <c>normal</c> if <c>ignore</c> is returned)
-              is sent to linked processes and ports,
+	      An exit signal with</p>
```
rickard-green, line 1970:
```suggestion
	      An exit signal with exit reason</p>
```

PR #6766, opened 2023-02-01
maps: always raise a `badmap` error if an iterator fails

Maria-12648430:
Follow-up of #6762 (Thanks @bjorng for suggestions made there :smiley:)

The current implementations of `maps` functions that accept iterators (namely, `filter/2`, `filtermap/2`, `fold/3`, `foreach/2` and `map/2`) test if the input is really an iterator by calling `next/1` on it once in a `try`. If this call fails by `next` raising a **`badarg`** error, the calling function catches it and raises a **`badmap`** error instead. However, if the iterator fails later on (eg giving a tuple like `{a, b, c}` as input passes the first call to `next`, but calling it again on the follow-up "iterator" `c` fails), the **`badarg`** error is bubbled up instead of the **`badmap`** error.

Running the entire algorithm (eg. the actual mapping) in a `try` (vs only the first call to `next`) is not feasible, as this may steal errors caused by the given function (vs the iterator), and cloak them as a `badmap` error.

The proposed implementation (for `map`) tries calls to `next` (if they _could_ fail) and raises an unique error that can be identified by the calling code to reliably distinguish it from other errors caused by the given mapping function, which must be bubbled up.

Unfortunately, for basically the same reason, the function `must_be_map_or_iter` in `erl_stdlib_errors`, which decides if a given $thing is (a map or) an iterator, had to be augmented to traverse the entire iterator to produce accurate error reports.

~This PR is currently a proof of concept and only implements the proposed changes for `map`, for illustration purposes and discussion. The concept can be easily adapted to the other functions named above once this has been accepted.~

### DIFF
--- a/lib/stdlib/src/erl_stdlib_errors.erl
+++ b/lib/stdlib/src/erl_stdlib_errors.erl
@@ -982,11 +982,9 @@ must_be_map_iterator_order(_) ->
 must_be_map_or_iter(Map) when is_map(Map) ->
     [];
 must_be_map_or_iter(Iter) ->
-    try maps:next(Iter) of
-        _ -> []
-    catch
-        error:_ ->
-            not_map_or_iterator
+    case maps:is_iterator_valid(Iter) of
+        true -> [];
+        false -> not_map_or_iterator
     end.
 
 must_be_number(N) ->
--- a/lib/stdlib/src/maps.erl
+++ b/lib/stdlib/src/maps.erl
@@ -30,6 +30,9 @@
          merge_with/3,
          groups_from_list/2, groups_from_list/3]).
 
+%% Internal
+-export([is_iterator_valid/1]).
+
 %% BIFs
 -export([get/2, find/2, from_list/1, from_keys/2,
          is_key/2, keys/1, merge/2,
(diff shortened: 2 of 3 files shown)


Maria-12648430:
... I have no idea what those errors in the CI are about XD

bjorng:
This looks good me. I suggest you go ahead and implement it for the other functions.

> Unfortunately, for basically the same reason, the function `must_be_map_or_iter` in `erl_stdlib_errors`, which decides if a given $thing is (a map or) an iterator, had to be augmented to traverse the entire iterator to produce accurate error reports.

The performance of error reporting is not that important. However, the testing for an invalid iterator could be optimized. If you want to do that optimization, I suggest that you add a new function called `maps:is_iterator_valid/1` (to avoid spreading the knowledge of the internal details of iterators to more modules). The trick is that if the iterator is not a tuple and the call to `next/1` succeeds, the iterator is valid; there is no need to continue, because the iterator returned from `erts_internal:map_next/3` is guaranteed to be valid.

Maria-12648430:
> This looks good me. I suggest you go ahead and implement it for the other functions.

Ok, I'm on it :smiley:

> The performance of error reporting is not that important. However, the testing for an invalid iterator could be optimized. If you want to do that optimization, I suggest that you add a new function called `maps:is_iterator_valid/1` (to avoid spreading the knowledge of the internal details of iterators to more modules). The trick is that if the iterator is not a tuple and the call to `next/1` succeeds, the iterator is valid; there is no need to continue, because the iterator returned from `erts_internal:map_next/3` is guaranteed to be valid.

See my updated commit, like this?

bjorng:
Yes, looks good.

PR #6767, opened 2023-02-01
Update list of map functions consuming iterators

Maria-12648430:
Obvious :wink:


bjorng:
Thanks for noticing the out-of-date list of functions.

We try to avoid putting lists of things in sentences. The original list of four items should really have been a bulleted list instead of a sentence. I have updated the commit to put the items into a bulleted list.

PR #6768, opened 2023-02-01
crypto: replace size/1 by byte_size/1

kikofernandez:

When one knows that the value being tested must be a binary, <c>byte_size/1</c> should be preferred.

Links to #6672

### DIFF
--- a/lib/crypto/src/crypto.erl
+++ b/lib/crypto/src/crypto.erl
@@ -1282,7 +1282,7 @@ rand_plugin_aes_next(Key, GenWords, F, Count) ->
     {V,Cache}.
 
 block_encrypt(Key, Data) ->
-    Cipher = case size(Key) of
+    Cipher = case byte_size(Key) of
                  16 -> aes_128_ecb;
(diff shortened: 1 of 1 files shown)


lib/crypto/src/crypto.erl
```diff
@@ -2276,7 +2276,7 @@ srp_pad_length(Width, Length) ->
     (Width - Length rem Width) rem Width.
 
 srp_pad_to(Width, Binary) ->
-    case srp_pad_length(Width, size(Binary)) of
+    case srp_pad_length(Width, byte_size(Binary)) of
         0 -> Binary;
         N -> << 0:(N*8), Binary/binary>>
```
bjorng, line 2281:
```suggestion
    case srp_pad_length(Width, bit_size(Binary)) of
        0 -> Binary;
        N -> << 0:N, Binary/binary>>
```

bjorng (changes requested):

bjorng (approved):

kikofernandez:
All tests pass locally, so I merge

PR #6770, opened 2023-02-01
ftp: replace size/1 by XXX_size/1

kikofernandez:

When one knows that the value being tested must be a binary, <c>byte_size/1</c> should be preferred. However, <c>byte_size/1</c> also accepts a bitstring (rounding up size to a whole number of bytes), so one must make sure that the call to <c>byte_size/</c> is preceded by a call to <c>is_binary/1</c> to ensure that bitstrings are rejected. Note that the compiler removes redundant calls to <c>is_binary/1</c>, so if one is not sure whether previous code had made sure that the argument is a binary, it does not harm to add an <c>is_binary/1</c> test immediately before the call to <c>byte_size/1</c>.

### DIFF
--- a/lib/ftp/src/ftp.erl
+++ b/lib/ftp/src/ftp.erl
@@ -2338,8 +2338,8 @@ file_close(Fd) ->
 
 file_read(Fd) ->
     case file:read(Fd, ?FILE_BUFSIZE) of
-        {ok, Bytes} ->
-            {ok, size(Bytes), Bytes};
+        {ok, Bytes} when is_binary(Bytes) ->
+            {ok, byte_size(Bytes), Bytes};
         eof ->
             {ok, 0, []};
         Other ->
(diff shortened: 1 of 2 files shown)


lib/ftp/src/ftp.erl
```diff
@@ -2338,8 +2338,8 @@ file_close(Fd) ->
 
 file_read(Fd) ->
     case file:read(Fd, ?FILE_BUFSIZE) of
-        {ok, Bytes} ->
-            {ok, size(Bytes), Bytes};
+        {ok, Bytes} when is_binary(Bytes) ->
```
bjorng, line 2341: The `is_binary/1` test is not strictly necessary here, because `file:read/2` is guaranteed to never return a bitstring.

lib/ftp/src/ftp.erl
```diff
@@ -2519,7 +2519,7 @@ open_options(Options) ->
         fun(Host) when is_list(Host) ->
                 true;
            (Host) when is_tuple(Host) andalso
-                       ((size(Host) =:= 4) orelse (size(Host) =:= 8)) ->
+                       ((tuple_size(Host) =:= 4) orelse (tuple_size(Host) =:= 8)) ->
```
bjorng, line 2522: You don't need the `is_tuple/1` test here. The clause can be simplified to:

```
           (Host) when tuple_size(Host) =:= 4; tuple_size(Host) =:= 8 ->
```

bjorng (changes requested):

kikofernandez:
Thanks for the feedback. I applied your fixes

kikofernandez:
This PR is closed due to the branch naming convention.
Follow #6793 for its continuation.

lib/ftp/src/ftp.erl
kikofernandez: Thinking about this, `file:read` can return `string() | binary()`, so I think that the when should be re-added?

bjorng: Yes.

PR #6791, opened 2023-02-04
add ets:first/next/last/prev_lookup

mittalanshul:
A common use case for ets:next is to read the next object stored in key order. This currently requires an ets:next followed by an ets:lookup for the object . Combining them in one call makes the overall operation much faster by getting rid of the second call to fetch the value. I also saw similar pattern being used in ets:foldl, so we can get speed up there as well by using this in ets:foldl .

Here is a very simple benchmark using erlperf: https://gist.github.com/mittalanshul/819e4d8d95cf993b3d31a36c3d72ac9c

There are a few problems with the PR as of now:
1. The way I passed the function pointers in erl_db code makes the code much cleaner, but will also negate any benefits that were being obtained by inlining (which I am not sure how much they were)
2. The ets_suite test uses a somewhat convoluted way to test (added a comment, looking for suggestions to do better)
3. We should probably add analogous functions for first/prev/last as well to maintain full parity.
4. I have not made changes to ets:foldl yet to use the new API

I can address the last 2 (as well as the first 2 if there are any suggestions), as long as there is agreement on the overall value of the new APIs

### DIFF
--- a/erts/emulator/beam/bif.tab
+++ b/erts/emulator/beam/bif.tab
@@ -357,6 +357,7 @@ bif ets:delete/1
 bif ets:delete/2
 bif ets:delete_object/2
 bif ets:first/1
+bif ets:first_lookup/1
 bif ets:is_compiled_ms/1
 bif ets:lookup/2
 bif ets:lookup_element/3
@@ -364,6 +365,7 @@ bif ets:lookup_element/4
 bif ets:info/1
 bif ets:info/2
 bif ets:last/1
+bif ets:last_lookup/1
 bif ets:match/1
 bif ets:match/2
 bif ets:match/3
@@ -372,7 +374,9 @@ bif ets:match_object/2
 bif ets:match_object/3
 bif ets:member/2
 bif ets:next/2
+bif ets:next_lookup/2
 bif ets:prev/2
+bif ets:prev_lookup/2
 bif ets:insert/2
 bif ets:insert_new/2
 bif ets:rename/2
--- a/erts/emulator/beam/erl_db.c
+++ b/erts/emulator/beam/erl_db.c
@@ -1115,6 +1115,29 @@ BIF_RETTYPE ets_first_1(BIF_ALIST_1)
     BIF_RET(ret);
 }
 
+/*
+** Returns the first {key, object(s)} in a table
+*/
+BIF_RETTYPE ets_first_lookup_1(BIF_ALIST_1)
+{
+    DbTable* tb;
+    int cret;
+    Eterm ret;
+
+    CHECK_TABLES();
+
+    DB_BIF_GET_TABLE(tb, DB_READ, LCK_READ, BIF_ets_first_lookup_1);
+
+    cret = tb->common.meth->db_first_lookup(BIF_P, tb, &ret);
+
+    db_unlock(tb, LCK_READ);
+
+    if (cret != DB_ERROR_NONE) {
+	BIF_ERROR(BIF_P, BADARG);
+    }
+    BIF_RET(ret);
+}
+
 /* 
 ** The next BIF, given a key, return the "next" key 
 */
@@ -1129,6 +1152,30 @@ BIF_RETTYPE ets_next_2(BIF_ALIST_2)
     DB_BIF_GET_TABLE(tb, DB_READ, LCK_READ, BIF_ets_next_2);
 
     cret = tb->common.meth->db_next(BIF_P, tb, BIF_ARG_2, &ret);
+
+    db_unlock(tb, LCK_READ);
+
+    if (cret != DB_ERROR_NONE) {
+	BIF_ERROR(BIF_P, BADARG);
+    }
+    BIF_RET(ret);
+}
+
+
+/*
+** The next_lookup BIF, given a key, return the "next" {key, object(s)}
+*/
+BIF_RETTYPE ets_next_lookup_2(BIF_ALIST_2)
+{
+    DbTable* tb;
+    int cret;
+    Eterm ret;
+
+    CHECK_TABLES();
+
+    DB_BIF_GET_TABLE(tb, DB_READ, LCK_READ, BIF_ets_next_lookup_2);
+
+    cret = tb->common.meth->db_next_lookup(BIF_P, tb, BIF_ARG_2, &ret);
 
     db_unlock(tb, LCK_READ);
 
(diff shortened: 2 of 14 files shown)


Maria-12648430:
I like the idea, this would be pretty handy. Here's my 2 cents :wink: _(I didn't look at the C stuff btw, I'm no good with that)_

IMO, extending this functionality to `first`, `last` and `prev` is more of a requirement than an option. I mean, it would be pretty awkward if I had to go through `first`+`lookup` for the first element while I could go through the rest by `next_object`. And in that sense, it would be as awkward if you could iterate through the table comfortably only _forwards_ but not _backwards_.

And with those functions at hand, using them in `foldl` (and `foldr`) would just be a natural choice, especially if as you say it is faster (which I didn't check, but sounds reasonable to me).

A problem I see with the implementation is that it returns an _object_, but to continue the iteration you need the _key_. And to get at the key, you need to know the key position in the object. This is all well if everything is in the same module etc, but with `protected` or `public` tables it may be that other processes in other modules want to access the table, too (not that I encourage that :wink:) And yes, you can get that via `ets:info/1,2`, but this again makes in unwieldy, and more likely than not will force you to write additional code to extract the key from the object.

Another thing that is not clear to me is how `bag`/ `duplicate_bag` tables should be traversed via `next_object`. If I understand correctly, `next_object` returns only a _single_ object (at least that is what I gather from your test), but in a `bag` there may be several with the same key. With `next`+`lookup`, you will get them all, but with `next_object`, you will only get one of them.

As to your tests. _ (Yes, I would probably do them differently :sweat_smile:)_
I see that your test implicitly tests that `next` and `next_object` traverse the table in the same order. This property, however, is something that is IMO somewhat essential, as otherwise it would not be possible to mix `next` and `next_object` calls in a traversal (not that I encourage this, but if it _didn't_ work this way the docs would have to be sprinkled with respective warnings all over). What I'm trying to say, this should be in a dedicated test, and preferably a property-based one, as this is more likely to find a problematic combination of objects, key positions, table types, (un-)fixed tables etc than what you can come up with manually.

mittalanshul:
> IMO, extending this functionality to first, last and prev is more of a requirement than an option.
+1, that is my proposal too (I should've not used the word "probably" in my original comment)

> A problem I see with the implementation is that it returns an object, but to continue the iteration you need the key. And to get at the key, you need to know the key position in the object.
The end result for a next + lookup pattern is to get the value stored in the object (why else will you do a lookup if you just need key?). Any module that is, 1) creating a table using a non default keypos value, 2) making the table public, is making even the next+lookup pattern unusable at worst, and error prone at best. I don't think replacing next+lookup with next_object there makes it any worse. 

> Another thing that is not clear to me is how bag/ duplicate_bag tables should be traversed via next_object.
+1 on this, the usability with those table types is definitely questionable. I expect an actual application will more likely use this for set/ordered_set (I know that's where my use cases are coming from). I don't know how to address this though, any suggestions? Is it advisable to restrict some APIs to certain table types?

> This property, however, is something that is IMO somewhat essential, as otherwise it would not be possible to mix next and next_object calls in a traversal
I cannot think of an actual use case where you would want to do this, but once the API is out there, we'll have no control over the usage. So I agree with having compatibility between the two APIs. Happy to add a test for it too.

> Therefore I believe there is a value of its own in this PR - even without fold/reverse. Although I agree it would be useful to at least design APIs for such fold/reverse operations (and have them implemented later).
Agreed. For foldl, this should be a very straightforward modification of https://fburl.com/zfcrw7d7 . Will follow it up in a subsequent PR

Maria-12648430:
> > A problem I see with the implementation is that it returns an object, but to continue the iteration you need the key. And to get at the key, you need to know the key position in the object.
> 
> The end result for a next + lookup pattern is to get the value stored in the object (why else will you do a lookup if you just need key?). Any module that is, 1) creating a table using a non default keypos value, 2) making the table public, is making even the next+lookup pattern unusable at worst, and error prone at best. I don't think replacing next+lookup with next_object there makes it any worse.

I guess I expressed myself badly, that wasn't meant as harsh criticism. Specifically, I didn't say anything about bad/worse :wink: I see potential for improvement, that's all :smiley_cat:

To illustrate what I mean, consider this typical use of first/next+lookup for traversal:
```erlang
trav(Tab) ->
    trav(Tab, ets:first(Tab)).

trav(_Tab, '$end_of_table') ->
    done;
trav(Tab, Key) ->
    Objects=ets:lookup(Tab, Key),
    do_stuff(Objects),
    trav(Tab, next(Tab, Key)).
```

While I will need to know something about the _shape_ of the objects to make use of them in the hypothetical `do_stuff` function, I don't need to know anything about which of the elements is the key. `first` will give me the first key, no matter where in the object it is; `next` will give me the key following the one I give it, which I got from a previous call to `next` or `first`. `lookup` will give me the objects under the key I give it, no matter where that key is in the object. Even if I change the key position, I don't need to touch any of the code above.

This is different when traversing with `next_object` as it is implemented now, since that requires knowledge about the key position, which is cumbersome. And it doesn't _have_ to be cumbersome ;) My suggestion would be to return a tuple of `{NextKey, NextObject}` from `next_object` instead (or `'$end_of_table'` for that matter). `NextKey` can then be used for the following `next_object` call, so _that_ comfortable property of `next` is preserved.

> > Another thing that is not clear to me is how bag/ duplicate_bag tables should be traversed via next_object.
> 
> +1 on this, the usability with those table types is definitely questionable. I expect an actual application will more likely use this for set/ordered_set (I know that's where my use cases are coming from). I don't know how to address this though, any suggestions? Is it advisable to restrict some APIs to certain table types?

Well, I'm not the one calling the shots here, but I'd say no. That would be a first. Why not return a list of next objects, those that share the same key? For `set`/`ordered_set`, this would always be a single element list, for `bag`/`duplicate_bag` there could be more.

In a nutshell, and in conjunction with what I said above: if I were to implement something like `next_object` in current Erlang, I would do it somewhat like this (and call it `next_objects` accordingly :stuck_out_tongue_winking_eye:):
```erlang
next_objects(Tab, Key) ->
    case ets:next(Tab, Key) of
        '$end_of_table' ->
            '$end_of_table';
        NextKey ->
            {NextKey, ets:lookup(Tab, NextKey)}
    end.
```

mittalanshul:
Sorry, nothing yet. I haven't been able to find time to complete it (also, from what I'm told OTP team is busy with OTP 26 release, so this will not be merged anytime soon anyways even if I do). Will try to address the comments and update as soon as I can.

sverker:
Sorry for forgetting about this PR.

About the semantics of `next_object`, I like the suggestion from @Maria-12648430:
```erlang
next_objects(Tab, Key) ->
    case ets:next(Tab, Key) of
        '$end_of_table' ->
            '$end_of_table';
        NextKey ->
            {NextKey, ets:lookup(Tab, NextKey)}
    end.
```
Which means
- Return both key and object(s).
- Return list of all objects with shared key for `bag` and `duplicate_bag`. The same that `ets:lookup` does.

But implemented atomically of course, only locking the table once.

I think we could document that it's ok to mix `next` and `next_object` during a traversal. That is, they are equivalent wrt which key they will return.
And a complete implementation with `first_object`, `next_object`, `last_object` and `prev_object`.

Hmm...maybe call it `next_lookup`.
- Says exactly what it does; next+lookup.
- Avoids decision whether call it "object" or "object**s**".

mittalanshul:
@sverker all the recommended changes are in.

Both make xmllint and make docs were failing locally were for me, but they were also failing without my changes, so most likely a setup problem. I did confirm the docs looked ok on the html page though.

Please let me know if there is something else to follow up on

sverker:
@mittalanshul Thanks for the contribution!

PR #6795, opened 2023-02-06
ssl: Maximize compatibility

### LINKED ISSUE #6772: TLS 1.3 middlebox incompatibilities
There are two related bugs - one server side, one client side.

**Describe the server bug**

OTP 25.? changes the SSL server behaviour in deciding if/when to send `change_cipher_spec`
The simplistic explanation of that being a _client_ problem given in #6374 ignores the real world.

Although [RFC 8446](https://www.rfc-editor.org/rfc/rfc8446) D.4 describes a method for "partially negotiated" middlebox compatibility, it also states:

> Either side can send `change_cipher_spec` at any time during the handshake

Furthermore, section 5 states an implementation receiving `change_cipher_spec`

> MUST simply drop it without further processing

For this reason many servers send `change_cipher_spec` regardless - OTP 24 is an example, as is www.google.com

The bug is being too pedantic in interpreting D.4, introducing a breaking change for cases where the clients can't (easily) be changed.

**Describe the client bug**

OTP 25.2 and also 24.3.4.8 have the change introduced for #6241 
The problem is that this change violates the handling of `change_cipher_spec` stated in section 5.

A TLS 1.3  client **MUST** drop `change_cipher_spec`

**To Reproduce**
Connect to an OTP 24 server or, you know, the internet in general
```erlang
1> ssl:start(), ssl:connect("www.google.com", 443, [{verify,verify_none},{middlebox_comp_mode,false}], 1000).
=NOTICE REPORT==== 1-Feb-2023::17:29:17.067186 ===
TLS client: In state wait_ee at ssl_gen_statem.erl:738 generated CLIENT ALERT: Fatal - Unexpected Message
 - {unexpected_msg,{internal,{change_cipher_spec,<<1>>}}}
{error,{tls_alert,{unexpected_message,"TLS client: In state wait_ee at ssl_gen_statem.erl:738 generated CLIENT ALERT: Fatal - Unexpected Message\n {unexpected_msg,{internal,{change_cipher_spec,<<1>>}}}"}}}
```

**Expected behavior**
- Regardless of `middlebox_comp_mode` the client **MUST** ignore  `change_cipher_spec`
- The server should send `change_cipher_spec` when `middlebox_comp_mode := true` for greater compatibility. 
In any case the behavioural change should have been clearly documented.

**Affected versions**
OTP 25.2 and OTP 24.4.3.8

IngelaAndin:
Always disregard one change_cipher_spec during handshake, even though middlebox mode is not used, to maximize compatibility.

Closes #6772

### DIFF
--- a/lib/ssl/src/tls_connection_1_3.erl
+++ b/lib/ssl/src/tls_connection_1_3.erl
@@ -288,8 +288,7 @@ start(internal = Type, #change_cipher_spec{} = Msg,
         _ ->
             handle_change_cipher_spec(Type, Msg, ?FUNCTION_NAME, State)
         end;
-start(internal = Type, #change_cipher_spec{} = Msg,
-      #state{session = #session{session_id = Id}} = State) when Id =/= ?EMPTY_ID ->
+start(internal = Type, #change_cipher_spec{} = Msg, State) ->
     handle_change_cipher_spec(Type, Msg, ?FUNCTION_NAME, State);
 start(internal, #client_hello{extensions = #{client_hello_versions :=
                                                  #client_hello_versions{versions = ClientVersions}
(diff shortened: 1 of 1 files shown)


KennethL:
Is the `change_cipher_spec` ignored in every state of the handshake? I mean that `change_cipher_spec` seems to be ignored in every handshake state so in practice we will allow many `change_cipher_spec` during the handshake. Is that the intention?

IngelaAndin:
Yes this is expected:

```
"An implementation may receive an unencrypted record of type
   change_cipher_spec consisting of the single byte value 0x01 at any
   time after the first ClientHello message has been sent or received
   and before the peer's Finished message has been received and MUST
   simply drop it without further processing."
```
I think however that a server should not send a change_chiper_spec  unless middle box  mode is negotiated or really mandated by the client, but here the spec is a little to vague and  others have interpreted it differently.  So we can make this change to have maximum interop with out loosing any thing important.

PR #6804, opened 2023-02-06
`private_append` optimization for binaries

frej:
When a binary is grown by appending data to it using `bs_create_bin`,
a considerable performance improvement can be achieved if the append
can be done using the destructive `private_append` instead of
`append`. Using `private_append` flavor of `bs_create_bin` is only
possible when the binary being extended has been created by
`bs_writable_binary` or by another `bs_create_bin` using
`private_append`. As `private_append` is destructive, an additional
requirement is that there is only one live reference to the binary
being being extended.

This set of patches implements the compiler optimization described above together with its dependencies.

### DIFF
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -38,6 +38,16 @@
 -define(DP(FMT), skip).
 -endif.
 
+%% A code location refering to either the #b_set{} defining a variable
+%% or the terminator of a block.
+-type kill_loc() :: #b_var{} | {terminator, beam_ssa:label()}.
+
+%% Map a code location to the set of variables which die at that
+%% location.
+-type kill_set() :: #{ kill_loc() => sets:set(#b_var{}) }.
+
+-type kills_map() :: #{ func_id() => kill_set() }.
+
 %% Record holding the liveness information for a code location.
 -record(liveness_st, {
                       in = sets:new([{version,2}]) :: sets:set(#b_var{}),
@@ -55,7 +65,8 @@ opt(StMap0, FuncDb0) ->
     %% called) or are stubs for nifs.
     Funs = [ F || F <- maps:keys(StMap0),
                   is_map_key(F, FuncDb0), not is_nif(F, StMap0)],
-    _Liveness = liveness(Funs, StMap0),
+    Liveness = liveness(Funs, StMap0),
+    _KillsMap = killsets(Liveness, StMap0),
     {StMap0, FuncDb0}.
 
 %%%
@@ -147,3 +158,45 @@ is_nif(F, StMap) ->
         _ -> false
     end.
 
+%%%
+%%% Calculate the killset for all functions in the liveness
+%%% information.
+%%%
+-spec killsets([{func_id(),
+                 #{func_id() => {beam_ssa:label(), #liveness_st{}}}}],
+               st_map()) -> kills_map().
+
+killsets(Liveness, StMap) ->
+    maps:from_list([{F, kills_fun(F, StMap, Live)} || {F, Live} <- Liveness]).
+
+%%%
+%%% Calculate the killset for a function. The killset allows us to
+%%% look up the variables that die at a code location.
+%%%
+kills_fun(Fun, StMap, Liveness) ->
+    #opt_st{ssa=SSA} = map_get(Fun, StMap),
+    kills_fun1(SSA, #{}, Liveness).
+
+kills_fun1([{Lbl,Blk}|Blocks], KillsMap0, Liveness) ->
+    KillsMap = kills_block(Lbl, Blk, map_get(Lbl, Liveness), KillsMap0),
+    kills_fun1(Blocks, KillsMap, Liveness);
+kills_fun1([], KillsMap, _) ->
+    KillsMap.
+
+kills_block(Lbl, #b_blk{is=Is,last=Last}, #liveness_st{out=Out}, KillsMap0) ->
+    kills_is([Last|reverse(Is)], Out, KillsMap0, Lbl).
+
+kills_is([I|Is], Live0, KillsMap0, Blk) ->
+    {Live, Key} = case I of
+                      #b_set{dst=Dst} ->
+                          {sets:del_element(Dst, Live0), Dst};
+                      _ ->
+                          {Live0, {terminator, Blk}}
+                  end,
+    Uses = sets:from_list(beam_ssa:used(I), [{version,2}]),
+    RemainingUses = sets:union(Live0, Uses),
+    Killed = sets:subtract(RemainingUses, Live0),
+    KillsMap = KillsMap0#{Key => Killed},
+    kills_is(Is, sets:union(Live, Killed), KillsMap, Blk);
+kills_is([], _, KillsMap, _) ->
+    KillsMap.
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -23,7 +23,11 @@
 
 -export([opt/2]).
 
--import(lists, [foldl/3, reverse/1]).
+-import(lists, [foldl/3, reverse/1, zip/2]).
+
+%% The maximum number of iterations when calculating alias
+%% information.
+-define(MAX_REPETITIONS, 16).
 
 -include("beam_ssa_opt.hrl").
 -include("beam_types.hrl").
@@ -37,6 +41,19 @@
 -define(DP(FMT, ARGS), skip).
 -define(DP(FMT), skip).
 -endif.
+
+-type call_args_status_map() :: #{ #b_local{} => ['aliased' | 'unique'] }.
+
+%% Alias analysis state
+-record(aas, {
+              caller :: func_id() | 'undefined',
+              call_args = #{} :: call_args_status_map(),
+              alias_map = #{},
+              func_db :: func_info_db(),
+              kills :: kills_map(),
+              st_map :: st_map(),
+              orig_st_map :: st_map()
+             }).
 
 %% A code location refering to either the #b_set{} defining a variable
 %% or the terminator of a block.
@@ -66,8 +83,9 @@ opt(StMap0, FuncDb0) ->
     Funs = [ F || F <- maps:keys(StMap0),
                   is_map_key(F, FuncDb0), not is_nif(F, StMap0)],
     Liveness = liveness(Funs, StMap0),
-    _KillsMap = killsets(Liveness, StMap0),
-    {StMap0, FuncDb0}.
+    KillsMap = killsets(Liveness, StMap0),
+
+    aa(Funs, KillsMap, StMap0, FuncDb0).
 
 %%%
 %%% Calculate liveness for each function using the standard iterative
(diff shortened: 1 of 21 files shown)


lib/compiler/test/beam_types_SUITE.erl
```diff
@@ -101,6 +101,16 @@ binary_absorption(Config) when is_list(Config) ->
     A = beam_types:meet(A, beam_types:join(A, B)),
     A = beam_types:join(A, beam_types:meet(A, B)),
 
+    %% Tests that the appendable flag behaves as intended.
```
jhogberg, line 104: Feel free to update `beam_types_prop:gen_bs_matchable/0` too, the following ought to do it:

```erlang
gen_bs_matchable() ->
    oneof([?LET(Unit, range(1, 16), #t_bs_matchable{tail_unit=Unit}),
           ?LET(Unit, range(1, 16), #t_bs_context{tail_unit=Unit}),
           ?LET({Unit, Appendable}, {range(1, 16), boolean()},
                #t_bitstring{size_unit=Unit})]).
```

jhogberg, line 104: Feel free to update `beam_types_prop:gen_bs_matchable/0` too, the following ought to do it:

```erlang
gen_bs_matchable() ->
    oneof([?LET(Unit, range(1, 16), #t_bs_matchable{tail_unit=Unit}),
           ?LET(Unit, range(1, 16), #t_bs_context{tail_unit=Unit}),
           ?LET({Unit, Appendable}, {range(1, 16), boolean()},
                #t_bitstring{size_unit=Unit})]).
```

lib/compiler/test/property_test/beam_types_prop.erl
```diff
@@ -217,7 +217,9 @@ gen_atom() ->
 gen_bs_matchable() ->
     oneof([?LET(Unit, range(1, 16), #t_bs_matchable{tail_unit=Unit}),
            ?LET(Unit, range(1, 16), #t_bs_context{tail_unit=Unit}),
-           ?LET(Unit, range(1, 16), #t_bitstring{size_unit=Unit})]).
+           ?LET(Unit, range(1, 16), #t_bitstring{size_unit=Unit}),
+           ?LET(Unit, range(1, 16), #t_bitstring{size_unit=Unit,
+                                                 appendable=true})]).
```
jhogberg, line 222:
```suggestion
           ?LET({Unit, Appendable}, {range(1, 16), boolean()},
                #t_bitstring{size_unit=Unit,appendable=Appendable})]).
```

bjorng (approved):

lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -0,0 +1,997 @@
...
+
+aa_is([I=#b_set{dst=Dst,op=Op,args=Args,anno=Anno0}|Is], SS0, Acc, AAS) ->
+    SS1 = aa_new_ssa_var(Dst, unique, SS0),
+    SS = case Op of
+             {bif,Bif} ->
+                 aa_bif(Dst, Bif, Args, SS1, AAS);
+             {float,_} ->
+                 SS1;
+             {succeeded,_} ->
+                 SS1;
```
jhogberg, line 351: Nitpick: I think this `case` is a bit noisy, do you think it makes sense to split the ones needing special attention and these `SS1` ones into separate groups?

lib/compiler/src/beam_types.erl
```diff
@@ -628,7 +631,7 @@ mtfv_1(A) when is_atom(A) ->
 mtfv_1(B) when is_bitstring(B) ->
     case bit_size(B) of
         0 ->
-            #t_bitstring{size_unit=256};
+            #t_bitstring{size_unit=256,appendable=true};
```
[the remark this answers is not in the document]
frej: I have added a comment, but it feels somewhat tautological, so perhaps your suggestion is a hint that the information from the commit message should be in the code. To try to remedy that, I also added a comment in `beam_types.hrl` where `t_bitstring` is defined.

lib/compiler/test/property_test/beam_types_prop.erl
frej: My Proper-fu failed me, that's nicer. Fixed.

lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -0,0 +1,997 @@
...
+             bs_match ->
+                 aa_set_aliased([Dst|Args], SS1);
+             bs_start_match ->
+                 [_,Bin] = Args,
+                 aa_set_aliased([Dst,Bin], SS1);
+             bs_test_tail ->
+                 SS1;
+             build_stacktrace ->
+                 SS1; % XXXX
```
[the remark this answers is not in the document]
frej: It means i should have done a `git grep XXXX` before pushing :) I use at as a marker for things I should take a second look at. 

At this particular location my thinking is that `build_stacktrace` is special in that it is only executed when we're exiting and additionally its only argument isn't a term which can contain a binary. I could do a `aa_set_aliased([Dst|Args], SS1)` but I think that is redundant. Correct? I added a comment about that.

lib/compiler/src/beam_ssa_private_append.erl
```diff
@@ -0,0 +1,613 @@
...
+
+find_appends_is([#b_set{dst=Dst, op=bs_create_bin,
+                        args=[#b_literal{val=append},SegmentInfo,Var|_],
+                        anno=Anno}|Is], Fun, Found0) ->
+    case is_unique(Var, Anno) andalso is_appendable(Anno, SegmentInfo) of
+        true ->
+            AlreadyFound = maps:get(Fun, Found0, []),
+            Found = Found0#{Fun => [{append,Dst,Var}|AlreadyFound]},
+            find_appends_is(Is, Fun, Found);
+        _ ->
```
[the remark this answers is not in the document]
frej: fixed.

lib/compiler/src/beam_types.erl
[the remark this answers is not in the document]
jhogberg: Yes, I wanted a comment to explain why `appendable=true` here. `%% See definition in beam_types.hrl for why empty binaries are appendable` or something similar would work. :-)

bjorng:
Some [benchmark](https://gist.github.com/bjorng/522470f313ac01d16156a597657f97bb) results on my Intel iMac:

Erlang/OTP 25.2:
```
== Testing with 1 MB ==
fun base64:encode/1: 1000 iterations in 16307 ms: 61 it/sec
fun base64:decode/1: 1000 iterations in 18402 ms: 54 it/sec
```

Current `master` branch:
```
== Testing with 1 MB ==
fun base64:encode/1: 1000 iterations in 7832 ms: 127 it/sec
fun base64:decode/1: 1000 iterations in 10878 ms: 91 it/sec
```

This PR:
```
== Testing with 1 MB ==
fun base64:encode/1: 1000 iterations in 5555 ms: 180 it/sec
fun base64:decode/1: 1000 iterations in 6433 ms: 155 it/sec
```

That makes both encode and decode of Base64 almost three times as fast compared to OTP 25.

lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -0,0 +1,1000 @@
...
+                  ({#t_bs_context{},_},{#t_bs_matchable{},_}) ->
+                       true;
+                  ({any,_},{none,_}) ->
+                       true;
+                  ({none,_},{any,_}) ->
+                       false;
+                  ({_,_},{any,_}) ->
+                       true;
+                  ({_,_},{none,_}) ->
+                       true;
```
jhogberg, line 673: Are we trying to sort the list from "least specific" to "most specific" here? If yes, doesn't this break `if A < B and B < C, then A < C`?

lib/compiler/src/beam_ssa_alias.erl
frej: Do you mean something like:

```Erlang
`case` Op of
  %% Instructions changing the alias status
  {bif,Bif}  ->
     aa_bif(Dst, Bif, Args, SS1, AAS);
  ...
  % Instructions which don't change the alias status
  {succeeded,_} ->
    SS1;
  ...
end
```
Or something more radical? The current order comes from when I expanded `beam_ssa:op()`, but I see the beenfit of grouping them differently. What I don't want to do is to just do:

```Erlang
 ...
  % Instructions which don't change the alias status
  _ -> SS1
end
```
as I prefer a hard crash in the compiler when a new instruction is added and it's not added here.

jhogberg: I was thinking the former, grouping the instructions that don't change the status together at the bottom of the list.

lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -0,0 +1,1000 @@
...
+        #{} -> false
+    end.
+
+aa_set_status(V=#b_var{}, Status, State) ->
+    Repr = aa_get_representative(V, State),
+    State#{ Repr => {status, Status} };
+aa_set_status(#b_local{}, _Status, State) ->
+    State;
+aa_set_status(#b_remote{}, _Status, State) ->
+    State;
```
bjorng, line 528: Not covered by our tests. Can this happen? Is there a test case missing?

lib/compiler/src/beam_ssa_private_append.erl
```diff
@@ -0,0 +1,613 @@
...
+
+%% The way find_defs and find_appends work, we can end up with
+%% multiple patches pacting different parts of a tuple or pair. We
+%% merge them here.
+merge_arg_patches([{Idx,Lit,P0},{Idx,Lit,P1}|Patches]) ->
+    P = case {P0, P1} of
+            {{hd,H},{tl,T}} ->
+                {pair,H,T};
+            {{tl,T},{hd,H}} ->
+                {pair,H,T};
```
bjorng, line 489: Not covered.

michalmuskala:
This is an awesome improvement! The results for `base64` are really impressive.

Is it possible to "pre-allocate" the binary? In many cases, like `base64` you actually know exactly how big the result is going to be - it would be great if there was a way to communicate this information to the runtime system. This could probably improve the results even further.

lib/compiler/src/beam_ssa_alias.erl
frej: The opposite, we want the list to start with the most specific type. And we never want to try a candidate for `any` as long as we have something more specific. So the "end game" should be:

```Erlang
                  %% For handling any and none, we want none at the
                  %% end of the list immediately preceeded by the
                  %% any-typed candidates.
                  ({any,_},{none,_}) ->
                       true;
                  ({none,_},{any,_}) ->
                       false;
                  ({_,_},{_,_}) ->
                       false
```

I see that I also fail to handle cases where the type is `identifier`. will fix that

lib/compiler/src/beam_ssa_alias.erl
frej: done

lib/compiler/src/beam_ssa_alias.erl
frej: I completely rewrote that part, instead of having a combinatorial mess of different pairwise comparisons I just assigned a rank to each type and sorted on that. Fixup pushed.

bjorng:
Added to our daily builds along with an updated primary bootstrap.

bjorng:
> Is it possible to "pre-allocate" the binary?

Yes, it would be possible, and it is already done for binary comprehensions for which the exact size can be calculated beforehand. This PR does not attempt to do any pre-allocating (the existing analysis for pre-allocation used for binary comprehensions is not general enough), but is something we will consider doing in OTP 27.

lib/compiler/src/beam_types.erl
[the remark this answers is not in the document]
frej: Good suggestion.

michalmuskala:
I don't think it even needs to be fully automated. I think giving control to the user would also work. Additionally, I don't think getting the size "wrong" would lead to any unsafely - the binary will either be re-allocated, or just remain with some unused bytes at the end wasting some memory. In particular, an API like `binary:empty_appendable(100)` could be useful. 
There are probably cases where automated analysis wouldn't be useful, but optimising for the most common case could be extremely beneficial. For example, for JSON when encoding a string we know the result will be at least as long as the input string + some padding (some implementations have heuristics for that, e.g. over-allocate 10%), so we can start the allocations much higher and avoid couple rounds of re-allocations, even if the final result is a little off.

max-au:
Binary preallocation is really useful for protocol serialisation. For example, protobuf and Thrift could be sped up significantly, and avoid cumbersome double transformation (term -> iolist -> binary).

bjorng:
> I don't think it even needs to be fully automated. I think giving control to the user would also work. Additionally, I don't think getting the size "wrong" would lead to any unsafely - the binary will either be re-allocated, or just remain with some unused bytes at the end wasting some memory.

Yes, getting the size wrong is safe. The compiler's size calculation is extremely conservative to avoid allocating a huge binary that is not actually needed, not because it would be unsafe.

> In particular, an API like `binary:empty_appendable(100)` could be useful.

Yes, that would work.

lib/compiler/src/beam_ssa_alias.erl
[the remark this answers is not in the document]
frej: Any verdict on `build_stacktrace` @bjorng ? The updated comment is in 0af6c67fc3dc4c5072bc77575e4c2cac7fea48b8.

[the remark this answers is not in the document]
bjorng: `build_stacktrace` is not only used when exiting. The code that follows `build_stacktrace` can do anything, for example print the stack trace to a log file and continue executing.

The argument can potentially contain binaries:

https://github.com/erlang/otp/blob/a2311845930de6dccba93c58b733dc4a75b1d754/erts/emulator/beam/beam_common.c#L1181-L1189

garazdawi:
> In particular, an API like binary:empty_appendable(100) could be useful.

```erlang
list_to_binary(List) ->
  list_to_binary(List, <<length(List)/size-hint>>). %% Creates an empty binary with a size hint.
list_to_binary([H|T], Acc) ->
  list_to_binary(T, <<H,Acc/binary>>);
list_to_binary([], Acc) ->
  Acc.
```

lib/compiler/src/beam_ssa_alias.erl
frej: As @jhogberg  pointed out offline, `aa_get_status_by_type` is not correct. The fix is in 1f35d0df1d717484398a04c9a2c1532f177d0816.

lib/compiler/src/beam_ssa_alias.erl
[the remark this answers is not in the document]
frej: So it's even worse, anything live at this point can be captured and aliased.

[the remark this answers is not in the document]
jhogberg: Sort of: a stack trace (currently) only contains the arguments of the operation that crashed, so we only need to consider the arguments of all operations that can crash into the current exception handler.

I'm not sure that tracking it would be worth it though, so maybe the optimization should just give up when encountering these. I don't think that doing so would cost too much since it's pretty rare to use `try/catch` with captured stack traces in the middle of a hot loop that appends binaries.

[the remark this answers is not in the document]
frej: That's not too bad. At least it can't alias terms which are not known when the analysis encounters `build_stacktrace`. Analyzing exactly which operations can crash into the handler sounds complicated, but just aliasing everything is quite simple.Hopefully users will wrap the loop with the try catch and not have the in the body. For example (with the fix in 63fd18d429824ed11e53451a525ef02e6d7ed289):

```Erlang
stacktrace() ->
    X = ..., % something unique
    try
	ex:code_that_fails(),
	X % still unique
    catch
	_:_:Stacktrace ->
	    io:format("Stacktrace: ~p~n", [Stacktrace]),
	    X % now conservatively aliased
    end.
```
Implementing this correction does not remove any private-append transforms in the set of files compiled by diffable.

Are there any plans to include more in the stack trace, i.e. arguments to functions higher on the stack? That would require the stack-crawler (or what it is called) to deep copy terms.

lib/compiler/src/beam_ssa_alias.erl
frej: I can't create an example which covers it, so away it goes! (fix in beb1977950b934fbd926bc223208c3ed2df3c895)

lib/compiler/src/beam_ssa_private_append.erl
frej: Currently the type analyzer does not handle the case when you "misuse" cons cells. For example:
```Erlang
b(L) ->
    b(L, [<<>>|<<>>]).

b([H|T], [A|B]) ->
    b([H|T], [<<A/binary,H:8>>|<<B/binary,H:8>>]);
b([], Acc) ->
    Acc.
```
Here the type analyzer will conclude that the type of the accumulator is `nonempty_improper_list(bitstring(8,appendable), bitstring(8))` and not `nonempty_improper_list(bitstring(256,appendable), bitstring(256,appendable))` which would be expected. If the type analysis pass is fixed so both fields of the cons are transformed by private append, we will hit this code with the example above.

Apart from a couple of lines dealing with nifs (will fix that), all other non-covered lines in  [beam_ssa_private_append.erl](https://github.com/erlang/otp/pull/6804/files/0af6c67fc3dc4c5072bc77575e4c2cac7fea48b8#diff-4cdc13ddd828821dd63acadce3756539e7a875e7ee4a8f62d9d9f28fda35c5ae) are due to this type analysis fault as we never try to transform pairs of binaries.

When I last looked at this, during the autumn, I seem to recall that the type was less specific, i.e. `nonempty_improper_list(bitstring(8,appendable), any)` so maybe this can now be fixed. I'll have a look.

frej:
I have pushed a squashed version of all the fixups on fixups. If you want to pull this branch for nightly testing @bjorng, now is good time.

bjorng:
Added the updated branch along with an updated bootstrap to our daily builds.

frej:
While working to increase the code-coverage in `beam_ssa_alias.erl` I (re-)discovered that for modules containing NIFs  `beam_ssa_opt:module/2` completely gives up on creating a function-db for all functions, not just the ones calling NIFs.
As `beam_ssa_alias.erl` is written to gracefully handle NIFs, but needs the caller/callee information provided by the function-db, this of course leads to some dead code (the predicates ) as we won't hit code paths where we deal with partial information.

Modifying `beam_ssa_opt` to build a partial function-db seems dangerous less than two days before code freeze, so (@bjorng, @jhogberg) should I excise the code dealing with NIFs in this PR or should I leave it in? As far as I can tell, this is 2 lines of code and I would be replacing them with `TODO` comments.

bjorng:
Don't worry about the dead code for this release candidate.

Can you squash the commits again? I will add it to our daily builds and we plan to merge tomorrow.

lib/compiler/src/beam_ssa_private_append.erl
frej: This is now fixed in d1211f5b498615c999d147ed7a6c241469db19e0, `beam_ssa_type:type/5` had problems handling `bs_create_bin` when the
first fragment was typed as a #t_union{} containing an appendable bitstring. The same error was in beam_ssa_private_append:is_appendable/2` (fixed in 6952cb6310a06ea79e96f80ebdf0c1f68fda55a7).

Additionally 6952cb6310a06ea79e96f80ebdf0c1f68fda55a7 adds a tests which exercises `ht`/`tl` lines above, this getting the coverage to 99% (the remaining non-covered lines are due to the NIF problem mentioned below).

frej:
> Can you squash the commits again? I will add it to our daily builds and we plan to merge tomorrow.

Will do within an hour or so, right now I'm in a meeting (where I'm supposed to be presenting).

frej:
Took a bit longer, squashed and pushed @bjorng.

lib/compiler/src/beam_ssa_private_append.erl
frej: ` {{tl,T},{hd,H}}` never happens, patches are stored in an orddict, so only ` {{hd,H},{tl,T}}` can occur. So the non-reachable code is removed in b33d99e2dbb6a40a0941c02db6fc29e31dac3310.

bjorng:

frej:
Looking at other recently pushed PRs it looks like I'm not the only one who gets tripped on the "Test Erlang/OTP (system)" step of the CI, so unless someone tells me differently, I will consider the CI failure not-my-fault™.

(@bjorng and @jhogberg, you have my phone number in an email if you need my urgent attention this afternoon)

PR #6810, opened 2023-02-07
Do not emit "unexpected" message on shutdown

max-au:
During shutdown, net_kernel process that is linked to an acceptor, gracefully terminates, triggering
`{EXIT, net_kernel, shutdown}` message sent to the acceptor.

However acceptor does not expect it, and instead of a graceful shutdown logs a confusing warning.

### DIFF
--- a/lib/ssl/src/inet_tls_dist.erl
+++ b/lib/ssl/src/inet_tls_dist.erl
@@ -232,7 +232,7 @@ gen_accept(Driver, Listen) ->
 %%  smaller than MaxPending
 accept_loop(DLK, undefined, MaxPending, Pending) when map_size(Pending) < MaxPending ->
     accept_loop(DLK, spawn_accept(DLK), MaxPending, Pending);
-accept_loop(DLK, HandshakePid, MaxPending, Pending) ->
+accept_loop({_, _, NetKernelPid} = DLK, HandshakePid, MaxPending, Pending) ->
     receive
         {continue, HandshakePid} when is_pid(HandshakePid) ->
             accept_loop(DLK, undefined, MaxPending, Pending#{HandshakePid => true});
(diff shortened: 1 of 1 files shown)


max-au:
See the discussion: https://erlangforums.com/t/tls-distribution-unexpected-message-exit-0-178-0-shutdown and original error https://github.com/rabbitmq/rabbitmq-server/discussions/7182

Note: this does not fix another issue with the supervision tree of the TLS dist. That second problem manifests like this:
```
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>     supervisor: {<0.2756.0>,tls_dyn_connection_sup}
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>     errorContext: child_terminated
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>     reason: killed
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>     offender: [{pid,<0.2757.0>},
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>                {id,sender},
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>                {mfargs,{tls_sender,start_link,undefined}},
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>                {restart_type,temporary},
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>                {significant,false},
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>                {shutdown,5000},
2023-02-03 15:07:44.565913+00:00 [error] <0.2756.0>                {child_type,worker}]
```
This appears to be caused by `net_kernel` shutdown (e.g `net:stop()` is called for dynamic distribution). `net_kernel` is linked to a TLS connection receiver (gen_statem process), and not to the `tls_dyn_connection_sup`. I tracked it down to 0078ebf6c531 (when TLS was split into sender/receiver), and I believe the fix would be to link the supervisor process instead of the receiver. 

Currently `#sslsocket` structure does not contain the supervisor pid. I can add a second commit that would use `process_info(DistCtrl, parent)` to fetch that in an obscure way, but before doing so, I'd be interested in what OTP team thinks (@RaimoNiskanen and @IngelaAndin ).

RaimoNiskanen:
Regarding df0ef03fee5b797c48f009774bf42747f817baf7; I think it is spot on.  It will collide with changes I have in 'master', but that is what it is.

Regarding the SSL distribution supervision tree vs. `net_kernel` and Kernel's `net_sup` tree, and how they should cooperate to avoid such an ugly and incorrect report during shutdown; we (I and @IngelaAndin) would have to take an internal discussion with the VM team e.g @rickard-green and maybe more to sort this out.

`net_kernel` does not care much for distribution implementation that has other supervisors.  It kills the distribution process it is linked to.  And SSL's supervision tree, I suppose, should to be taken down from the top in an orderly manner...

We will see

PR #6817, opened 2023-02-07
Update on_load handling during boot

RaimoNiskanen:
Introduce a transient worker that is started by kernel:init and that only calls init:run_on_load_handlers(), which will run all so far postponed (after code load) on_load handlers. The worker exits immediately after that.

This on_load worker is run after code_server and standard_error, which should provide enough fundamentals for the on_load handlers themselves.

The need for explicitly calling on_load handlers from e.g user_drv and the SSL distribution inet_tls_dist should vanish.

Therefore init:run_on_load_handlers/1 should be removed, but to avoid bootstrap problems that can wait to a later commit.


RaimoNiskanen:
I added commits to remove `init:run_on_load_handlers/1`

Issue #5695, opened 2022-02-08
Why is Dialyzer's `-Wunknown` off by default?


TD5:
I've found myself explicitly switching on `-Wunknown` in various places, and being confused when I've forgotten to set it and therefore not had Dialyzer warn when I expected it to. I started to wonder why it's not on by default, because I find it quite useful. My best guess is that it's off by default to allow people to run Dialyzer over a subset of their codebase for performance reasons, and doing this would potentially then result in `unknown` warnings purely due to "missing" functions/modules being outside the set of things under analysis.

From the docs:
> **-Wunknown**
> Let warnings about unknown functions and types affect the exit status of the command-line version. The default is to ignore warnings about unknown functions and types when setting the exit status. When using Dialyzer from Erlang, warnings about unknown functions and types are returned; the default is not to return these warnings.

What's the official view on why `unknown` warnings aren't always returned, and does it still make sense for this to be the case?

max-au:
Could it be done in order to support hot code upgrades (where a module/function may appear)?

paulo-ferraz-oliveira:
@max-au, I believe so. I don't have an "official view" but that's what makes sense to me, too.

bjorng:
> What's the official view on why `unknown` warnings aren't always returned, and does it still make sense for this to be the case?

I don't know reasoning behind the decision, but I think it's partly because you might want to run Dialyzer on part of a codebase for speed and also it fits with Dialyzer's philosophy to only warn for things that Dialyzer is sure is wrong.

We will not change the default in OTP 25. We will postpone to OTP 26 the decision on whether the current default makes sense or whether it should be changed.

jesperes:
I think it would be a good idea to make `unknown` on by default. If does make it difficult to run dialyzer on a subset of your code, but disabling it hides a whole category of problems, such as misspelled types.

TD5:
I perhaps should have been clearer in the original post: I really did intend this as a question of history and original intent to aid my own understanding, rather than a proposal to change things. That said, I am happy if the question sparks a debate and results in some sort of improvement for users of OTP.

The point made above about hot code reloading is an interesting one, and makes sense to me. My question very much came from my personal experience, where I have all my code and dependencies available at build and analysis time, and I run Dialyzer over everything on every change to stay on top of the warnings. I appreciate this isn't necessarily reflective of how some (or even most?) people use Erlang and OTP, so it makes sense that the option default to something that supports the majority of users.

[...]

PR #6822, opened 2023-02-08
dialyzer: make -Wunknown a default; add -Wno_unknown flag

kikofernandez:
Prior to this PR, Dialyzer used to have disabled (by default) warnings about unknown types and functions. 

This default value has been overwritten; Dialyzer now warns about unknown types and functions (as requested by the community in GH-5695). Thus, the following two examples are equivalent, i.e., passing the `-Wunknown` function is enabled by default:

```
dialyzer moduler.erl -Wunknown -Wmissing_return
dialyzer moduler.erl -Wmissing_return
```

Dialyzer has a new flag, `-Wno_unknown`. Its purpose is to supress warnings about unknown functions and types.

Users who wish to suppress these warnings can invoke Dialyzer using this flag. Example:

```
dialyzer module.erl -Wno_unknown
```

### DIFF
--- a/lib/dialyzer/doc/src/dialyzer.xml
+++ b/lib/dialyzer/doc/src/dialyzer.xml
@@ -386,19 +386,17 @@ dialyzer --plts plt_1 ... plt_n -- files_to_analyze</code>
       <item>
         <p>Suppress warnings for unused functions.</p>
       </item>
+      <tag><c>-Wno_unknown</c></tag>
+      <item>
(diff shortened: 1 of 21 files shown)


bjorng (approved):

TD5:
I'm a bit late to the party, but I just wanted to say thank you for looking into this and making the change. It'll certain make my life easier when working with Dialyzer.

Issue #4326, opened 2016-04-16
ERL-128: Close stdin of a port


OTP-Maintainer:
Original reporter: `fnanninga`
Affected version: `Not Specified`
Component: `erts`
Migrated from: https://bugs.erlang.org/browse/ERL-128

---

```
There is no way to close the stdin of a port (send EOF). This is necessary for piping data to programs which usually read from a file but can read from stdin.

It's possible to work around this by creating a wrapper executable and communicating with that wrapper through some frame format; but the effort seems out of proportion with the effort needed to add this feature to Erlang.

Using named pipes instead seems not to be an option as as far as I know Erlang does not support those.

*EDIT:* I still want to be able to receive future data from the program, so erlang:close_port/1 is too much.
```

bezborodow:
I have also implemented a patch on https://github.com/erlang/otp/compare/master...bezborodow:otp:port_eof and am currently working on a test case.

I have just noticed https://github.com/rinpatch/otp/commit/ce8d0750b7e441a66be065fb272727825f3bb4a9 and will look to it for further inspiration.

bezborodow:
I have also implemented a test case, but would need feedback and review.

adkron:
I need this. I must send something on STDIN, and the output returns on STDOUT. The issue is that STDOUT closes when I try to get an EOF to STDIN.

ivanov:
I am not sure if this issue was closed as "won't fix" or just accidentally closed at the same time that it was marked as a low priority feature. 

I also came across this issue today, and found some of the resources linked from @bezborodow's #6824 summary description useful to peruse for possible workarounds, along with [the comparisons in `Rambo` README](https://github.com/jayjun/rambo#comparisons). Hope that helps you figure out some paths forward.

adamwight:
I ran into the same need, when interfacing with the `bzip2` binary, which won't send the last packet of data until it knows its input is fully consumed.

It would be nice if the port call looked something like this, and the implementation between port and the spawn driver can be opaque.
```erlang
erlang:port_command(Port, eof)
```

Nicd:
This has been closed as "completed", though it seems it has not been completed. Can we reopen this or mark it as "won't fix" if that is the resolution?

adamwight:
[...]

PR #6824, opened 2023-02-08
Port: spawn_control: Close the Write File Descriptor (Implying EOF)

bezborodow:
This implements a means to close the stdin of a port (implying EOF). This is necessary for piping data to programs which read from standard input and wait for the input stream to be closed, such as `wc`.

It works by closing the write descriptor when the magic code *0x0112c000* (`SPAWN_CTRL_OP_EOF`) is passed to `port_control/3` for the `spawn_driver` [driver entry](https://www.erlang.org/doc/man/driver_entry.html). The write descriptor is replaced by `/dev/null`, with further writes being discarded to the bit bucket. Replacing the write file descriptor with `/dev/null` avoids complexities when dealing with "half-open" ports.

Only implemented for Unix systems.  See `erts/emulator/sys/unix/sys_drivers.c`.

Includes a test case. See `erts/emulator/test/port_bif_SUITE.erl`.

Resolves:

  - https://github.com/erlang/otp/issues/4326

Users having this issue:

  * [Issues with stdin on ports](https://erlang.org/pipermail/erlang-questions/2013-July/074916.html)
  * [EOF to external program](http://erlang.org/pipermail/erlang-questions/2010-November/054330.html)
  * [struggling with port_command()](http://erlang.org/pipermail/erlang-questions/2010-October/053944.html)
  * [open_pipe({spawn, "cat"}) --- how to close stdin of cat?](http://erlang.org/pipermail/erlang-questions/2009-March/042123.html)
  * [Erlang Ports: Interfacing with a "wc"-like program?](https://stackoverflow.com/questions/8792376/erlang-ports-interfacing-with-a-wc-like-program)
  * [erlang ports bewlider me](https://gist.github.com/timruffles/77e9b69cdecdd7b3ef08)
  * [Use an OS process like a bash pipe: Send it STDIN and get its STDOUT](https://stackoverflow.com/questions/74833431/use-an-os-process-like-a-bash-pipe-send-it-stdin-and-get-its-stdout)
  * [Rambo - Run your command. Send EOF. Get output](https://elixirforum.com/t/rambo-run-your-command-send-eof-get-output/25052)

Users solving this issue with middleware:

  * [mattsta/erlang-stdinout-pool](https://github.com/mattsta/erlang-stdinout-pool#why-is-this-special)
  * [saleyn/erlexec](https://github.com/saleyn/erlexec#communicating-with-an-os-process-via-stdin-and-sending-end-of-file)
  * [petrkozorezov/estdinout](https://github.com/petrkozorezov/estdinout#estdinout)
  * [jayjun/rambo](https://github.com/jayjun/rambo#why)
  * [alco/porcelain](https://github.com/alco/porcelain#overview)

Example usage of the *0x0112c000* control code with `wc`:

```
-module(test).
-export([wc/1]).

wc(InputText) ->
    P = open_port({spawn, "wc"}, [stream, use_stdio]),
    P ! {self(), {command, InputText}},
    port_control(P, 16#0112c000, []),
    receive {P, X} -> X end.
```

Running the example:

```
Erlang/OTP 26 [DEVELOPMENT] [erts-13.1.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit:ns]

Eshell V13.1.4 (press Ctrl+G to abort, type help(). for help)
1> test:wc("Hello, world!\n").
{data,"      1       2      14\n"}
2>
```

### DIFF
--- a/erts/emulator/sys/unix/sys_drivers.c
+++ b/erts/emulator/sys/unix/sys_drivers.c
@@ -570,7 +570,7 @@ static ErlDrvData spawn_start(ErlDrvPort port_num, char* name,
 	memcpy((void *) (cmd_line + CMD_LINE_PREFIX_STR_SZ), (void *) name, len);
 	cmd_line[CMD_LINE_PREFIX_STR_SZ + len] = '\0';
 	len = CMD_LINE_PREFIX_STR_SZ + len + 1;
-}
+    }
 
     if ((cwd = getcwd(wd_buff, MAXPATHLEN+1)) == NULL) {
         /* on some OSs this call opens a fd in the
--- a/erts/emulator/sys/unix/sys_drivers.c
+++ b/erts/emulator/sys/unix/sys_drivers.c
@@ -757,11 +757,32 @@ static ErlDrvData spawn_start(ErlDrvPort port_num, char* name,
 #undef CMD_LINE_PREFIX_STR_SZ
 }
 
+/* Control op */
+#define SPAWN_CTRL_OP_EOF 0x0112c000U
+
 static ErlDrvSSizeT spawn_control(ErlDrvData e, unsigned int cmd, char *buf,
                                   ErlDrvSizeT len, char **rbuf, ErlDrvSizeT rlen)
 {
     ErtsSysDriverData *dd = (ErtsSysDriverData*)e;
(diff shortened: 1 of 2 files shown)


bezborodow:
I have not convinced myself that this is a good solution, but I hope at least that it will provide some inspiration.

PR #6777, opened 2023-02-03
Document -code_path_choice is scheduled for removal in OTP 27

josevalim:

PR #6832, opened 2023-02-09
Revert #6683 and #6777

jhogberg:
#6683 and #6777 broke standalone systems with code archives created by `reltool`, but we didn't see any test failures until we updated preloaded modules yesterday.


josevalim:
I tried to take a look at the tests but unfortunately `create_standalone_app` fails on my machine even without those changes.

I took a quick look around though and the source of the issue may be on `reltool_target:cr_path/3`. It doesn't consider if an application is an escript. If it is, it may be a matter of changing the escript path from "someapp-1.0/ebin" to "someapp-1.0/someapp-1.0/ebin". A similar change may be necessary for archives. But alas, I cannot test it. :(

josevalim:
Ok, I tested it and I can verify that is also broken: #6693. `reltool` uses archives by default and, unless there is a use case, we should stop doing that? I will send a PR and you folks can choose to keep it or not. :)

PR #6836, opened 2023-02-09
Remove archives support in reltool

josevalim:
reltool would package applications in archives
by default but there is little to no gain in
doing so. Furthermore, this ability pushed
complexity into other parts of Erlang/OTP,
such as the code server, erl_prim_loader, and
the init module.

This release makes it so archives are no longer
supported but the underlying options still warn
to avoid breaking changes. No functionality should
be removed from the assembled released (except
that they will have 100% fewer archives).


josevalim:
This should enable safe merging of #6683, #6777, and #6693.

Perhaps the best route forward is to merge this PR for Erlang/OTP 26 and schedule the three PRs above for OTP 27.

jhogberg:
We had a technical board meeting today, and decided to merge this for OTP 26 RC2. We'll also add release notes in OTP 26 to warn people that the default code path choice will be changed to `strict` in OTP 27.

jhogberg:
Merged, thanks for the PR!

PR #6837, opened 2023-02-09
Update maybe_expr documentation note

sabiwara:
Hello,

I noticed that the documentation for `maybe` was still mentioning a need to enable the feature, but if I understand correctly this is now enabled by default?

Feel free to close my PR if mistaken.


bjorng:
It is no longer necessary to enable the `maybe_expr` feature in the runtime system, but it is still necessary to enable it when compiling a module using `maybe`.

Issue #6789, opened 2023-02-04
[diameter] Unsigned64 treated like Unsigned32


sebastiw:
**Describe the bug**

When defining an Unsigned64 with values greater than 2^32-1 diameterc faulty outputs
```
** @define MIP6-Feature-Vector specifies value 1099511627776 at line number that is out of range for a value of Diameter type Unsigned32
```

**To Reproduce**
An example from 3GPP TS 29.273 which declares and define the MIP6-Feature-Vector AVP.

```
$ cat dia_3gpp_ts_29273.dia

@name dia_3gpp_ts_29273
@vendor   10415 3gpp

@avp_types
MIP6-Feature-Vector                                         124  Unsigned64        M 

@define MIP6-Feature-Vector
PMIP6_SUPPORTED                                             0x0000010000000000

$ lib/diameter/bin/diameterc dia/dia_3gpp_ts_29273.dia
** @define MIP6-Feature-Vector specifies value 1099511627776 at line number that is out of range for a value of Diameter type Unsigned32
```

**Expected behavior**

I expect this dictionary to compile with Unsigned64 type checking.

**Affected versions**
25.0.4

otp-anders:
Closing following the discussion in PR 6849.

PR #6849, opened 2023-02-11
diameter: Fix Unsigned64 defines

sebastiw:
Fix for #6789 

Unsigned64 diameter types are currently treated like Unsigned32. Any dictionary which defines values in range 2^32 to 2^64 will therefor fail with uint32_out_of_range error.

Solution is to check which type is carried for the define and assert the correct range.


otp-anders:
Doesn't look quite right since you're checking an Integer64 with a function made for unsigned integers, but the same problem looks to have existed for 32-bit. I'll have to take a closer look.

sebastiw:
yes, I couldn't find any tests or checks for int32/int64s, so I assume the bug is there as well.

otp-anders:
Sorry this took me a while, but the reason the check is only for Unsigned32 is hinted at by commit c03ac3db: @define is a historical remnant that's just used to generate macros for Result-Code, which is Unsigned32. It's not documented in the file format since the usability is fairly questionable. It could be made to support other types, but then it should be all Diameter types, not just Unsigned64 (hence my initial confusion about Integer32).

The example you give in issue #6789 would give you a ?'DIA_3GPP_TS_29273_MIP6-FEATURE-VECTOR_PMIP6_SUPPORTED' (56 characters!), but is that really a macro you want to use? I'd just define my own macro with a shorter name if I needed it, as I wrote in the commit comment 11+ years ago.

sebastiw:
I see. ok, then no probably not.
I thought it was documented, but now when I look I don't see it :)

Thank you

PR #6852, opened 2023-02-13
[argparse] Command line parser for Erlang

max-au:
Inspired by Python argparse library.


max-au:
Submitting a draft PR for the review. The documentation is lacking, and it will take some time for me to come up with all descriptions plus a separate User Guide with code examples (given the complexity, it would be too much to have just one document describing module reference and user scenarios).

In general, it keeps most of the original `argparse` logic, so I'll be adjusting and improving documentation I have on (https://hexdocs.pm/argparse/argparse_ref.html). There are however a few differences: exceptions are no longer raised for parser errors (only for invalid command specifications), and a helper function `run/3` is added to argparse.erl itself (as opposed to more complicated cli.erl in the original library).

Looking forward for a discussion (and working on the docs in the background).

michalmuskala:
What is the reason for shipping this with OTP? Is it going to be used by some OTP CLI tools? Otherwise I'm not sure I see huge advantage over including stuff in the standard library vs importing from hex, given how easy it is these days.

lib/stdlib/src/argparse.erl
```diff
@@ -0,0 +1,1305 @@
...
+    parse(Args, Command, #{}).
+
+%% @doc Parses supplied arguments according to expected command specification.
+%% @param Args command line arguments (e.g. `init:get_plain_arguments()')
+%% @returns argument map, or argument map with deepest matched command
+%%  definition.
+-spec parse(Args :: [string()], command(), Options :: parser_options()) -> parse_result().
+parse(Args, Command, Options) ->
+    Prog = validate(Command, Options),
+    %% use maps as set_v2, FIXME: replace before firing a PR!
```
TD5, line 228: > FIXME: replace before firing a PR!

Probably worth taking a look at this, @max-au

lib/stdlib/src/argparse.erl
```diff
@@ -0,0 +1,1305 @@
...
+default(_) ->
+    undefined.
+
+%% command path is now in direct order
+format_path(Commands) ->
+    lists:join(" ", Commands).
+
+%%--------------------------------------------------------------------
+%% Validation and preprocessing
+%% Theoretically, Dialyzer should do that too.
```
TD5, line 855: Am I right in thinking the below "just" generates clear error messages, rather than actually performs some static analysis? If so, perhaps this comment is more confusing than anything else.

Also, it might be nice to actually export the validation functions, so they can be run at build time or in-IDE by users?

TD5:
> What is the reason for shipping this with OTP? Is it going to be used by some OTP CLI tools? Otherwise I'm not sure I see huge advantage over including stuff in the standard library vs importing from hex, given how easy it is these days.

I can't comment on the appropriateness of this library in particular, because I don't have any experience as a user of it, but:
- Loads of bits of OTP use hand-rolled arg parsing, e.g. in Dialyzer. It might be nice to have a more standardised way to do these things so they don't continue to grow into even more complex beasts
- It'd be nice to have a standard/official argument parsing option, since it removes the potential bike-shedding argument around something which typically doesn't need it

arpunk:
> What is the reason for shipping this with OTP? Is it going to be used by some OTP CLI tools? Otherwise I'm not sure I see huge advantage over including stuff in the standard library vs importing from hex, given how easy it is these days.

One advantage that I see (and would love to have) is to have an uniform way to handle command line arguments within OTP for enhancing CLI/escripts tools. It is shipped by default in Elixir as `OptionParser` too. Importing from hex is easy with `mix` and `Mix.install/1` but not *that* easy on stock OTP.

paulo-ferraz-oliveira:
It'll also be important to have this once `rebar3` is integrated into Erlang/OTP as the default build tool, to not have to rely on `getopt` (I believe @max-au even considered replacing it there, in the past, so this move seems to make sense). But I understand where @michalmuskala is coming from: if we keep putting everything into Erlang/OTP the platform not only becomes more rigid, but release schedules for updates are harder to manage (especially for consumers).

max-au:
I see several reasons to have this in the OTP:
* standalone interpreted escripts with a standard way to parse command line. It isn't possible with a separate library, because `argparse` won't be in a code path, and there is no simple way to use hex.pm from a textual escript
* command line tools included in the OTP (dialyzer, ct_run) and future integrations (rebar3)
* uniform way to understand command line for all BEAM tooling

I agree that moving any code into the lower layer should be considered with care. It may also make sense to mark the module as experimental, and a subject to change APIs in the future.

michalmuskala:
If the argument is that this can be used in OTP (and rebar) itself, ideally we'd have some examples that this is indeed possible - especially taking into account backwards compatibility and current interfaces. Without concrete migration plans, I'm afraid we'd end up in [the xkcd classic](https://xkcd.com/927/).

As to `escripts` - while I agree argument parsing is useful for them, the fact that you can't install hex dependencies for escripts is entirely a limitation of the current Erlang tooling. [`Mix.install`](https://hexdocs.pm/mix/Mix.html#install/2) has been an awesome recent addition to Elixir that indeed makes experimentation and building small, standalone scripts, so much easier.
I think this case or argument parsing is just an example of a much larger gap in the Erlang tooling.

okeuday:
@michalmuskala From a security perspective, it is better to have escripts that don't need to install remote dependencies.  Security is required for fault-tolerance to be possible, i.e., it is part of having reliable systems.

max-au:
Updated documentation, it should be reasonably comprehensive now. I will come up with a separate "cookbook" containing more examples, hints, tips & tricks if the OTP team decides to move forward with this PR, and convert the draft into an actual PR.

I agree that a demonstration of an actual conversion is necessary, and will do it to the `typer` tool - it appears that almost the entire typer.erl is a hand-rolled parser, a perfect candidate for a demo.

KennethL:
I think there are good reasons to include an argument parser in OTP because:
- Having a uniform way to handle arguments to Escripts is nice
- Implementing OTP tools as Escripts using the argument parser would also be nice (converting existing tools)
- Getting rid of non OTP dependencies from rebar3 would make it possible to include it in OTP
- Even if it is easy to handle dependencies with rebar3 and hex every dependency has an extra administrative cost if it should be included in a product. 

I think there are 2 questions to answer here:
1. Is it good to include an argument parser in OTP? (the answer is most probably yes)
2. What argument parser should that be? It could be `argparse`, `getopt` or something else.
    What are the pros and cons for argparse and getopt?

I think we need a real world non trivial example implemented in both of them, for example `rebar3`, `erlc` or `dialyzer`.

Just a thought: An argument parser is working with a string as input , converts it to individual options and checks them according to declared rules. Would it be possible to use the same approach for functions with a complicated option lists (or maps)? Can this be achieved by separating the text parsing from the actual checking of options. Checking if they are allowed and if they have the right type of value. Handling default values etc.

ferd:
Rebar3 has various well-formed arguments lines based on some commands, such as `release`:

```
→ rebar3 help release
Build release of project.
Usage: rebar3 release [--all <all>] [-n <relname>] [-v <relvsn>]
                      [-u <upfrom>] [-o <output_dir>] [-h] [-l <lib_dir>]
                      [-d <dev_mode>] [-i <include_erts>] [-a <override>]
                      [-c [<config>]] [--overlay_vars <overlay_vars>]
                      [--vm_args <vm_args>] [--sys_config <sys_config>]
                      [--system_libs <system_libs>] [--version]
                      [-r <root_dir>] [-m <relnames>]

  --all               If true runs the command against all configured
                      releases
  -n, --relname       Specify the name for the release that will be
                      generated
  -v, --relvsn        Specify the version for the release
  -u, --upfrom        Only valid with relup target, specify the release to
                      upgrade from
  -o, --output-dir    The output directory for the release. This is `./`
                      by default.
  -h, --help          Print usage
  -l, --lib-dir       Additional dir that should be searched for OTP Apps
  -d, --dev-mode      Symlink the applications and configuration into the
                      release instead of copying
  -i, --include-erts  If true include a copy of erts used to build with,
                      if a path include erts at that path. If false, do
                      not include erts
  -a, --override      Provide an app name and a directory to override in
                      the form <appname>:<app directory>
  -c, --config        The path to a config file [default: ]
  --overlay_vars      Path to a file of overlay variables
  --vm_args           Path to a file to use for vm.args
  --sys_config        Path to a file to use for sys.config
  --system_libs       Boolean or path to dir of Erlang system libs
  --version           Print relx version
  -r, --root          The project root directory
  -m, --relnames      Like --all, but only build the releases in the list,
                      e.g. --relnames rel1,rel2
```

And then others that are extremely flexible and almost random and require custom parsing such as `new` (the templates):

```
$ rebar3 new
<lists templates>
$ rebar3 new help 
<lists templates>
$ rebar3 new help lib
<describes the lib template>
$ rebar3 new lib name=appname desc="a fun app for everyone" ...
<applies the template>
```

Chances are that whatever we do we'd need to scope the argument specification to individual tasks.

My expectation however is that bringing Rebar3 into OTP would likely break the plugin ecosystem by shifting dependencies and require us to re-think some features we no longer can support, and also allow us to clean up a bunch of cruft as well. (It's also a bit unclear how we'd exactly go about generating protobuffs, but we can likely generate them offline and vendor the generated source artifacts like we do today)

Either way, since moving to OTP is likely to break things, I would in theory be fine with depending on an OTP argparsing library and I assume most people would as well.

max-au:
I was looking at a Typer, but given Dialyzer is a more complex tool to cover, I'll do the Dialyzer instead.

KennethL:
We have discussed this in the OTP technical board and came to the following conclusion:

- We need a side by side comparision of `argparse` and `getopt` before we can judge. Would be good with some bigger example of existing CLI command. For example Dialyzer. We think the use of maps as input and output is a pro for `argparse`.
- It is today quite hard to understand all features of `argparse`.
- We think it would be possible to add `argparse` (or `getopt`) as experimental in OTP 26.0 if we can take the decision in April

max-au:
This PR now contains `dialyzer` CLI conversion. The conversion itself was surprisingly easy. Probably, CLI test coverage is not complete, so some features are not working as expected. I will do a few more manual tests later.

While working in the conversion, I found a number of pre-existing issues that I also fixed in this PR:
* `dial_option()` type in dialyzer.hrl had many options missing (but actively used). I made a decision to add those options there
* dialyzer would imply `--src` option if there is a source file with "*.erl" extension, even when there are also *.beam files. I made a decision to not imply this, but I think a better behaviour would be to produce an error when there is a mix of *.beam and *.erl files
* there are several undocumented options (`--metrics_file`, `--warning_apps`, `--solver` and `--resources`). I left them undocumented (`help => hidden`), but I suspect that all of them (except the `solver`) should be documented. 
* help/usage printout was inconsistent (some lines are <80 columns, but some exceed 100). By default, `argparse` wraps help/usage lines at 80 characters (it is configurable).

I do not think there is a way to adopt `getopt` to support `dialyzer`, as there are too many features and validations missing from `getopt`.

One more note. Help/usage produced by `argparse` is generated automatically (from the actual options/arguments list), hence formatting is slightly different. I believe it is just as readable (if not better) compared to the original one.


Appreciate a more thorough review! I'm sure there are some things missing, but so far it looks like retrofitting `argparse` into existing applications is not that hard.

josevalim:
Question: should it be part of "stdlib" or of a separate application such as "tools"?

eproxus:
> I'm afraid we'd end up in [the xkcd classic](https://xkcd.com/927/).

This is more like going from 0 → 1 standard?

> As to `escripts` - while I agree argument parsing is useful for them, the fact that you can't install hex dependencies for escripts is entirely a limitation of the current Erlang tooling. [`Mix.install`](https://hexdocs.pm/mix/Mix.html#install/2) has been an awesome recent addition to Elixir that indeed makes experimentation and building small, standalone scripts, so much easier.

While I agree that is an awesome addition—and I'd love to see that in Erlang too—the benefit of being able to run an escript out-of-the-box without hitting the network is _huge_.

eproxus:
@max-au 
> By default, `argparse` wraps help/usage lines at 80 characters (it is configurable).

Is there a reason it doesn't wrap at `io:columns()` by default?

eproxus:
@josevalim 
> Question: should it be part of "stdlib" or of a separate application such as "tools"?

Or why not `argparse`? In the interest of being able to create minimal Erlang releases, ideally one should be able to depend on just the applications needed without bringing in too many extras.

KennethL:
@josevalim I think argparse fits in `stdlib`. I don't think it should be in tools since what we have there is normally not used or needed to run a system or application. If we are thinking of using `argparse` for all or many of the command line utilities then we don't want a dependencey to `tools`. I also think it would be bad to add a dependency to `tools` from all non OTP applications implmenting a CLI with help of `argparse`.

max-au:
> Is there a reason it doesn't wrap at `io:columns()` by default?

Initially I did exactly that, to fin that:
- dumb terminals (which are at majority in out classic environment) return an error
- when it works, it's actually less readable to have 153 (or any other random number) column word wrap compared to 80

Hence I decided that it should be a developers decision to explicitly specify column width (as it should be really simple).

josevalim:
@max-au I believe the usual approach I saw on the codebase is to get `io:columns` and do a `min(io:columns(), 100)` (or some other random number which can also be used as default).

max-au:
@josevalim Honestly I  have no opinion on that one, and would rather follow OTP team direction. Also, there has to be a minimal setting (I think there is a bug in the wordwrap algorithm that'd make it fail if some words are longer than the columns limit).

michalmuskala:
I looked through the proposed CLI implementation for dialyzer and it's definitely a huge improvement. The fact that `argparse` can even handle a "hand-grown" CLI like that is also pretty impressive.

TD5:
> I looked through the proposed CLI implementation for dialyzer and it's definitely a huge improvement. The fact that `argparse` can even handle a "hand-grown" CLI like that is also pretty impressive.

I have to agree. I also found it interesting that @max-au found that some docs/help was missing during the translation (I think I am the one that added some of these, such as `--metrics_file` and `--warning_apps`), and certainly I feel like these would have been much harder to forget to do if we'd been using `argparse`, since there's a single place where you set all aspects of an argument.

max-au:
Rebased to resolve merge conflict (command line options documentation)

max-au:
I erroneously assumed that `\n` would not work on Windows (because of `\r\n` newline on that platform). Now that I checked that `io:format(<<"\n">>)` works just fine on Windows, I removed the unnecessary complexity in the newline code.

Binaries are accepted as help text, so I changed Dialyzer command line to pass binaries as help. While at it, rebased to the current master branch.

jhogberg:
Merged, thanks for the PR! :)

PR #6854, opened 2023-02-13
dialyzer: Fix crash when adding multiple PLTs

### LINKED ISSUE #6850: Does the --plts Dialyzer option work correctly?
Hi,

(using Erlang/OTP 25 [erts-13.1.3])

Maybe there is a bug in the management of the Dialyzer command-line --plts option (or the doc is maybe unclear), as a call like:

`dialyzer --add_to_plt --output_plt X.plt -r ./src ./test --plts Y.plt` is accepted, whereas `dialyzer --add_to_plt --output_plt X.plt -r ./src ./test --plts Y.plt Z.plt` will fail (for all Z) with:

`{"init terminating in do_boot",{function_clause,[{dialyzer_cl,plt_common,[{options,[],["./src","./test"],plt_add,false,[],byte_code,false,["Y.plt","Z.plt"],[],"X.plt",[warn_behaviour,warn_bin_construction,[...]`

The man page has the following example: `dialyzer files_to_analyze --plts plt_1 ... plt_n`.

Thanks for any feedback,

Olivier.

TD5:
The code previously assumed that multiple PLTs would only be used in combination with an analysis, but adding multiple PLTs to another without analysis was not considered.

The existing logic for merging PLTs was used in the context of an analysis, with different data to hand, so here we just validate the merge, then sequentially add each PLT.

Resolves #6850

### DIFF
--- a/lib/dialyzer/src/dialyzer_cl.erl
+++ b/lib/dialyzer/src/dialyzer_cl.erl
@@ -105,7 +105,16 @@ init_opts_for_build(Opts) ->
 add_to_plt(Opts) ->
   Opts1 = init_opts_for_add(Opts),
   AddFiles = get_files_from_opts(Opts1),
-  plt_common(Opts1, [], AddFiles).
+  case Opts1#options.init_plts of
+    [] -> plt_common(Opts1, [], AddFiles);
+    [_] -> plt_common(Opts1, [], AddFiles);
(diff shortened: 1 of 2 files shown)


jhogberg:
Thanks for the PR, I've added it to our nightly test runs.

jhogberg:

PR #6855, opened 2023-02-13
compiler: Don't generate unsafe code for binary append

frej:
[This patch is flawed, will revise it in the morning]

The private-append optimization worked on the assumption that the `private_append` version of `bs_create_bin` checked that the type of the first argument was a bitstring. As it doesn't, the compiler shouldn't do the optimization when it cannot guarantee that the type is and only is a bitstring.


frej:
This patch is stupid, there is no need for an new `is_typesafe`-function, I should just modify `is_appendable/2`. I'll revise it in the morning.

Note to self, don't try to rush something out before you leave for the day...

bjorng:
(If you remove the colons from your `Fixes` lines in the commit message, Github should automatically add links to the issues.)

bjorng (approved):
Looks good to me. I will add it to our daily builds as soon as the first release candidate has been released (probably tomorrow).

PR #6856, opened 2023-02-13
inets: fix typespec httpc:cookie_header/{1,2,3}

kikofernandez:
The typespec was incorrectly setting the returned type of `httpc:cookie_header/{1,2,3}` to be `[HttpHeader]` where `HttpHeader :: {Field :: [byte()], Value :: binary() | iolist()}`.

The returned type should have been simply `HttpHeader` (i.e., not a list just a singleton item).

This can be seen also in OTP 24:
image
image

And the wrong modification in OTP-25, where we clearly can see that `[HttpHeader]` should have been just `HttpHeader`

image


IngelaAndin (commented):
Maybe also change the variable name in httpc_manager  to exclude the s!? There can be several cookies but only one header.

```
handle_call({which_cookies, Url, Options}, _, 
	    #state{cookie_db = CookieDb} = State) ->
    ?hcrv("which cookies", [{url, Url}, {options, Options}]),
    case uri_parse(Url) of
	{ok, {Scheme, Host, Port, Path}} ->
	    CookieHeaders = 
		httpc_cookie:header(CookieDb, erlang:list_to_existing_atom(Scheme), {Host, Port}, Path),
	    {reply, CookieHeaders, State};
	{error, _} = ERROR ->
	    {reply, ERROR, State}
    end;

```

kikofernandez:
@IngelaAndin Done! I have renamed the variable named as suggested :)

IngelaAndin (approved):

PR #6857, opened 2023-02-13
Add take_all_objects/1 for atomic take and delete

sorentwo:
This is a concurrency-safe alternative to taking all objects in a loop or calling `tab2list` and `delete_all_objects` sequentially. It fixes the table and iterates through all keys, taking all objects to return a single list.

After writing a similar internal function twice in the same week (for statsd/prometheus style metrics collection), I realized that others might benefit from a standard function!

To start, I've only implemented the function and typespec. If this is accepted as a sound approach, I'll add tests and documentation.

### DIFF
--- a/lib/stdlib/src/ets.erl
+++ b/lib/stdlib/src/ets.erl
@@ -36,6 +36,7 @@
 	 tab2list/1,
          table/1,
          table/2,
+	 take_all_objects/1,
 	 fun2ms/1,
 	 match_spec_run/2,
 	 repair_continuation/2]).
@@ -477,6 +478,21 @@ slot(_, _) ->
 
 take(_, _) ->
     erlang:nif_error(undef).
+
+-spec take_all_objects(Table) -> [Object] when
+      Table :: table(),
+      Object :: tuple().
+
+take_all_objects(T) ->
+    ets:safe_fixtable(T, true),
+    Objects = internal_take_all(T, ets:first(T), []),
+    ets:safe_fixtable(T, false),
+    Objects.
+
+internal_take_all(_, '$end_of_table', Acc) ->
+    Acc;
+internal_take_all(T, Key, Acc) ->
+    ets:internal_take_all(T, ets:next(T, Key), ets:take(T, Key) ++ Acc).
 
 -spec update_counter(Table, Key, UpdateOp) -> Result when
       Table :: table(),


lib/stdlib/src/ets.erl
```diff
@@ -478,6 +479,21 @@ slot(_, _) ->
...
+take_all_objects(T) ->
+    ets:safe_fixtable(T, true),
+    Objects = internal_take_all(T, ets:first(T), []),
+    ets:safe_fixtable(T, false),
+    Objects.
+
+internal_take_all(_, '$end_of_table', Acc) ->
+    Acc;
+internal_take_all(T, Key, Acc) ->
+    ets:internal_take_all(T, ets:next(T, Key), ets:take(T, Key) ++ Acc).
```
sorentwo, line 495: According to my benchmarks, appending with `++` is the same as flattening afterward. Optimizing the standard case with a single element, like in `set`, makes this about three times faster. Here's the benchmark results:

```
Operating System: macOS
CPU Information: Apple M1 Pro
Number of Available Cores: 10
Available memory: 32 GB
Elixir 1.14.2
Erlang 25.1

warmup: 2 s
time: 5 s
memory time: 0 ns
reduction time: 0 ns
parallel: 1
inputs: none specified
Estimated total run time: 21 s

Benchmarking Append ...
Benchmarking Flatten ...
Benchmarking Optimized ...

Name                ips        average  deviation         median         99th %
Optimized       32.73 K       30.55 μs   ±166.00%       32.67 μs       46.29 μs
Append          10.72 K       93.25 μs     ±3.64%       92.33 μs      106.25 μs
Flatten         10.70 K       93.47 μs     ±4.88%       92.33 μs      106.97 μs

Comparison:
Optimized       32.73 K
Append          10.72 K - 3.05x slower +62.70 μs
Flatten         10.70 K - 3.06x slower +62.92 μs
```

Then again, there may be a faster way to accomplish all of this 🤷

michalmuskala: The problematic case with `++` in terms of performance is usually appending to an accumulator, since that ends up being quadratic in the number of elements - here you're prepending to an accumulator, so that's why it's not very problematic.

michalmuskala:
A more powerful version of this would be some sort of `select_take` - I wished several times something like that existed.

That does not preclude this function from existing as well, though.

max-au:
What are the atomicity guarantees provided by this function? `safe_fixtable` works for enumeration, but then the function also modifies the table that was fixed, and does it without taking into account any other concurrent operations happening.

How would it work, for example, when there are concurrent processes also removing entries from the database?

sorentwo:
> What are the atomicity guarantees provided by this function?

I genuinely don't know! That's partly why I created this PR—to find out what's the correct way to "take all objects" in a concurrency-safe way.

> How would it work, for example, when there are concurrent processes also removing entries from the database?

Concurrent access seems to be prevented if both processes use `safe_fixtable`; otherwise, there may be argument errors due to missing keys. Essentially, this doesn't work atomically.

I'd still like to know _how_ to do this or some type of `take_delete` safely.

sverker:
This implementation is not atomic, in the sense as one single isolated operation serializable with other concurrent operations on the table. Concurrent read ops may see intermediate results and concurrent write ops may be "randomly" seen or not  by `take_all_objects`.

The only thing that `ets:safe_fixtable` guarantees is that
- `ets:next(Key)` will succeed even if `Key` has been deleted (by keeping deleted objects in table marked as deleted).
- keys existing during the entire traversal will be hit once.
- keys inserted or deleted during the traversal will (unpredictably) be hit one or zero times.

A truly atomic implementation cannot be done in Erlang, at least not entirely. It would probably be done something similar to `ets:delete_all_objects`, if done at all.

Maria-12648430:
I'm not sure if this does what you think it does, especially regarding atomicity. AFAIK, `safe_fixtable` does not create a transaction-like snapshot of or lock on a table. It only ensures that a call to `next` and a few others will succeed, whereas without `safe_fixtable` it may fail if something gets deleted during a traversal. In other words, other processes may still insert, update and/or delete objects in the table.

Maria-12648430:
@sverker you beat me to it :sweat_smile:

sorentwo:
Thank you for the feedback and clarifications. After the last round of feedback, I wrote some concurrent tests and discovered that this certainly doesn't accomplish what I had hoped.

> A truly atomic implementation cannot be done in Erlang, at least not entirely. It would probably be done something similar to `ets:delete_all_objects`, if done at all.

I'll use an alternate design for my use cases—but I still think some variants of `take`, like `select_take` or `take_all_objects`, would be truly helpful.

PR #6858, opened 2023-02-14
Exit process immediately when the max heap size is exceeded

bjorng:
Executing receive-related instructions would cause the process to enter a suspended state forever. Avoid that by scheduling out immediately after any GC if the process is exiting.

### DIFF
--- a/erts/emulator/beam/emu/macros.tab
+++ b/erts/emulator/beam/emu/macros.tab
@@ -74,6 +74,12 @@ GC_SWAPOUT() {
     c_p->i = I;
 }
 
+MAYBE_EXIT_AFTER_GC() {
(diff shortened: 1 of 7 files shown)


erts/emulator/beam/emu/macros.tab
```diff
@@ -74,6 +74,12 @@ GC_SWAPOUT() {
     c_p->i = I;
 }
 
+MAYBE_EXIT_AFTER_GC() {
+    if (ERTS_PROC_IS_EXITING(c_p)) {
```
rickard-green, line 78:
```suggestion
    if (ERTS_PSFLG_EXITING & erts_atomic32_read_nob(&c_p->state)) {
```
Avoids an unnecessary memory barrier. It might not be that big of a deal though compared to the gc if it is only used after a gc.

Issue #6455, opened 2022-11-09
gen_udp:send execution time increases with caller's queue length


gonzalobf:
**Describe the bug**
`gen_udp:send` gets slowed down by the message queue that the caller has.

I suspect the problem is inside the `gen_udp:send` implementation which calls to `erlang:port_command` and waits in a `receive` for the completion [in this line](https://github.com/erlang/otp/blob/master/erts/preloaded/src/prim_inet.erl#L651). I don't think it puts any mark in the queue so the `receive` needs to do a full queue scan looking for the `inet_reply`.

We have been hit by this issue a few times in production recently. We have a process with the sole responsibility of receiving messages and forwarding them to a udp socket that built up a queue of 2M messages. At some point, the forwarder was getting messages at a regular rate but the queue wasn't going down. We found that the clients weren't receiving messages at the usual rate. In the server side, we found the process was spending all its time in `prim_inet:do_sendto` so we thought it was stuck in `gen_udp:send`, but it was actually processing them very slowly.  Looking at my benchmark, it would take approximately 50ms to process each message with a queue of 2M as we had. 

**To Reproduce**
I wrote a [benchmark](https://gist.github.com/gonzalobf/9186724c72c47fc80e7d2265a6b8741f) that reproduces the issue. 

**Expected behavior**
I would expect `gen_udp:send` to take constant time or not be influenced so much by the caller's queue size.

**Affected versions**
I checked in OTP25.1 and master.

**Additional context**
I opened a related topic in [erlang forums](https://erlangforums.com/t/gen-udp-server-stuck-in-send/2003) before opening this one where I posted the code and the debug I did in more details.

Polexy:
I think we got same issue for our app.
We use `syslog_logger` which uses `gen_udp:send`. we make experiments with cpu, and found that after increased it to ~85% it's start stocking on `prim_inet:do_sendto` with an endlessly growing queue
```
[{registered_name,syslog_logger},
 {current_function,{prim_inet,do_sendto,4}},
 {initial_call,{proc_lib,init_p,5}},
 {status,running},
 {message_queue_len,121553},
 {links,[#Port<0.8>,<0.1031.0>]},
 {dictionary,[{'$initial_call',{syslog_logger,init,1}},
              {'$ancestors',[<0.1031.0>,<0.1030.0>]}]},
 {trap_exit,false},
 {error_handler,error_handler},
 {priority,normal},
 {group_leader,<0.1029.0>},
 {total_heap_size,7932638},
[...]

PR #6859, opened 2023-02-14
Introduce erl_drv_command_error()

rickard-green:
* [`int erl_drv_command_error(ErlDrvPort port, char *error_reason)`](https://erlang.github.io/prs/6859/erts-13.1.4/doc/html/erl_driver.html#erl_drv_command_error)

    Generate an `error` exception, with the reason `error_reason` as an atom, in a call to the `erlang:port_command/2` BIF or the `erlang:port_command/3` BIF. The error can only be generated from the drivers `outputv()` or `output()` callbacks, which implements port command, when the port command has been initiated via one of the synchronous `port_command()` BIFs. When the port command has been initiated by the asynchronous `Port ! {Owner, {command, Data}}` construct, this call will fail and **no** exception will be generated.

    `error_reason` needs to be a null terminated latin1 string. If the string is longer than 255 characters (excluding the null termination character), it will be truncated to 255 characters.

    Returns zero on success and a non-zero integer on failure.


rickard-green:
@RaimoNiskanen for use when solving #6455

sverker:
What about changing argument `char* error_reason` to `ErlDrvTermData error_reason_atom` to be created by existing `driver_mk_atom()`. Two advantages:

- Avoid introducing Latin1 string in new API.
- Can be used more efficient by calling `driver_mk_atom()` once at driver load time.

rickard-green:
@RaimoNiskanen Closing this since it cannot be used in solving #6455. Looking into other alternatives...

PR #6862, opened 2023-02-15
compiler: Avoid crash in private-append optimization

### LINKED ISSUE #6847: Compiler crash in pass beam_ssa_opt, sub pass ssa_opt_private_append
**Describe the bug**
The compiler crashes like this when compiling the module shown below:

```
Sub pass ssa_opt_private_append
beam_ssa_private_append_crash.erl: internal error in pass beam_ssa_opt:
exception error: no function clause matching beam_ssa_private_append:patch_literal_term([],[self],5) 
  in function  beam_ssa_private_append:patch_appends_ret/3 (beam_ssa_private_append.erl, line 462)
  in call from beam_ssa_private_append:patch_appends_f/5 (beam_ssa_private_append.erl, line 416)
  in call from beam_ssa_private_append:'-patch_appends/3-anonymous-2-'/3 (beam_ssa_private_append.erl, line 390)
  in call from maps:fold_1/4 (maps.erl, line 416)
  in call from beam_ssa_private_append:private_append/3 (beam_ssa_private_append.erl, line 72)
  in call from beam_ssa_opt:run_phases/3 (beam_ssa_opt.erl, line 75)
``` 

**To Reproduce**
Compile the following module:

```
-module(beam_ssa_private_append_crash).
-compile([export_all]).

foo() ->
    <<(bar(ok))/binary>>.

bar(#{}) ->
    [];
bar(_) ->
    <<>>.
```

**Expected behavior**
The compilation should succeed.

**Affected versions**
The current `master` branch.

frej:
Avoid a crash in the private-append optimization due to attempting to rewrite a literal which isn't a bitstring.

As value tracking is done without type information, it can happen that when following def chains, we encounter literals which are not bitstrings. We should never schedule a literal which isn't a bitstring for rewriting when the complete literal is what we are tracking, i.e. when the element is `self`.

Closes #6847

### DIFF
--- a/lib/compiler/src/beam_ssa_private_append.erl
+++ b/lib/compiler/src/beam_ssa_private_append.erl
@@ -168,6 +168,13 @@ get_results(SSA, Element, Fun, DefSt) ->
 get_results([{_,#b_blk{last=#b_ret{arg=#b_var{}=V}}}|Rest],
             Acc, Element, Fun, DefSt) ->
     get_results(Rest, [{V,Element}|Acc], Element, Fun, DefSt);
+get_results([{_,#b_blk{last=#b_ret{arg=#b_literal{val=Lit}}}}|Rest],
+            Acc, Element=self, Fun, DefSt) when not is_bitstring(Lit) ->
+    %% As value tracking is done without type information, we can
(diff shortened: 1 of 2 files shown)


bjorng:

Issue #6508, opened 2022-11-24
Inconsistent error report in Dialyzer and the use of type variables


kikofernandez:
**Describe the bug**
Dialyzer reports an error for this type spec:

```erlang
-module(example).
-export([run_test_error/1]).

-spec run_test_error(Opts) -> term()
      when Opts :: {join_specs, Bool}.
run_test_error(Opts) ->
    {error, fail}.
```
with error: `type variable 'Bool' is only used once (is unbound)`.

If the unbound variable is used more than once, there is no error:

```erlang
-spec run_test(Opts) -> term()
      when Opts :: {join_specs, Bool} | {test, Bool}.
run_test(Opts) ->    {error, fail}.
```
It is not clear to me why there is not error if the type variable `Bool` is used twice.

**To Reproduce**
`dialyzer example.erl`

**Expected behavior**
I expected that either both of them are an error or none of them are an error. Since the input type is not used
as part of the output type, I do not see why it matters that the type variable is used multiple times in the input type.

**Affected versions**
OTP-23+ I think

kikofernandez:
Confirmed: this is a bug and we should fix it

kikofernandez:
This is specific to tuples and unions. The type variable is not bound between arguments or argument and return types (e.g., identity function), therefore the variable is not bound and it should be an error. (maybe maps too).

PR #6864, opened 2023-02-15
dialyzer: fix singleton type var in union types

kikofernandez:
Dialyzer (or `erl_lint.erl`) was accepting code of the following form:

```erlang
-spec run_test(Opts) -> term()
      when Opts :: {join_specs, Bool} | {test, Bool}.
run_test(Opts) ->    {error, fail}.
```

But this cannot be ever checked by Dialyzer as the type variable `Bool` is in an union, which means that `Bool` happened only once and thus, `Bool` is an unbound type variable. Essentially, the example above should be rejected by Dialyzer as is the one below:

```erlang
-spec run_test_error(Opts) -> term()
      when Opts :: {join_specs, Bool}.
run_test_error(Opts) ->
    {error, fail}.
```

This commit fixes GH-6508 and adds a bunch of tests to `erl_lint.erl` so that we can keep track of which singleton type variables should throw an error and which ones are ok.

### DIFF
--- a/lib/kernel/src/wrap_log_reader.erl
+++ b/lib/kernel/src/wrap_log_reader.erl
@@ -116,11 +116,11 @@ open(File, FileNo) when is_list(File), is_integer(FileNo) ->
 close(#wrap_reader{fd = FD}) ->
     file:close(FD).
 
--type chunk_ret() :: {Continuation2, Terms :: [term()]}
-                   | {Continuation2,
+-type chunk_ret() :: {Continuation2 :: term(), Terms :: [term()]}
+                   | {Continuation2 :: term(),
                       Terms :: [term()],
                       Badbytes :: non_neg_integer()}
-                   | {Continuation2, 'eof'}
+                   | {Continuation2 :: term(), 'eof'}
                    | {'error', Reason :: term()}.
 
 -spec chunk(Continuation) -> chunk_ret() when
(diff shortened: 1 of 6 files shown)


lib/kernel/src/wrap_log_reader.erl
```diff
@@ -116,11 +116,11 @@ open(File, FileNo) when is_list(File), is_integer(FileNo) ->
 close(#wrap_reader{fd = FD}) ->
     file:close(FD).
 
--type chunk_ret() :: {Continuation2, Terms :: [term()]}
-                   | {Continuation2,
+-type chunk_ret() :: {Continuation2 :: term(), Terms :: [term()]}
```
kikofernandez, line 119: Errors detected by the new linter / dialyzer

jhogberg (approved):

essen:
Please note this a potential incompatibility in the release notes because this breaks compilation of code that was compiling for many years.

kikofernandez:
It will be marked as a potential incompatibility in the release notes

josevalim:
@kikofernandez, would it be possible to ship this as a warning, at least for Erlang/OTP 26?

This is restricting the ability of projects to try out Erlang/OTP 26, because working code no longer compiles (and there were no past warnings), and I assume it may also restrict the ability of said projects to migrate to Erlang/OTP 26 once it is out. If this could be a warning, it would at least give folks a year to update their project and dependencies.

Thank you.

kikofernandez:
We will try to make it a warning now, and an error in OTP-27

PR #6865, opened 2023-02-15
Improve the `lists:merge` family of functions

Maria-12648430:
This PR addresses a few shortcomings and inconcistencies that I discovered in the `merge` family of functions in the `lists` module.

* if the second (and third, in the variants that take 3 lists) list is empty, the call is fast and returns the first list; however, if the first list is not empty but the second (and third) list is, the call is slow as it does not simply return the other list but runs through the algorithm
* if the second (and third) list is empty, the first argument can really be anything, it does not have to be a list at all; for example, `lists:merge(foo, [])` will happily return `foo`. For `merge/1`/`umerge/1` this will also happen if the list of lists contains only empty lists and a single non-list; for example, `lists:merge([[], foo, []])` will return `foo`

The changes I made here allow the calls to be equally fast if all but maybe one list are empty. Also, giving the functions non-list arguments is detected and causes an error. The `keymerge` functions are somewhat of an exception here: if any list contains non-tuples or too short tuples while the others are empty, this is not detected. Previously, this was detected if the second list was not empty; if the second list was empty, however, the first list could again be anything.

For the record, I'm not sure about the filtering I'm doing as a preparatory step in `merge/1` and `umerge/1`. It could suffice to just ensure that the first thing that is not an empty list is a list. Comments welcome :smiley:

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -701,24 +701,44 @@ merge(L) ->
       Y :: term(),
       Z :: term().
 
-merge3(L1, [], L3) ->
-   merge(L1, L3);
-merge3(L1, L2, []) ->
-   merge(L1, L2);
-merge3(L1, [H2 | T2], [H3 | T3]) ->
-   lists:reverse(merge3_1(L1, [], H2, T2, H3, T3), []).
+merge3([_|_]=L1, [H2 | T2], [H3 | T3]) ->
+   lists:reverse(merge3_1(L1, [], H2, T2, H3, T3), []);
+merge3([_|_]=L1, [_|_]=L2, []) ->
+    merge(L1, L2);
+merge3([_|_]=L1, [], [_|_]=L3) ->
+    merge(L1, L3);
+merge3([_|_]=L1, [], []) ->
+    L1;
+merge3([], [_|_]=L2, [_|_]=L3) ->
+    merge(L2, L3);
+merge3([], [_|_]=L2, []) ->
+    L2;
+merge3([], [], [_|_]=L3) ->
+    L3;
+merge3([], [], []) ->
+    [].
 
 %% rmerge3(X, Y, Z) -> L
 %%  merges three reversed sorted lists X, Y and Z
 
 -spec rmerge3([X], [Y], [Z]) -> [(X | Y | Z)].
 
-rmerge3(L1, [], L3) ->
-   rmerge(L1, L3);
-rmerge3(L1, L2, []) ->
-   rmerge(L1, L2);
-rmerge3(L1, [H2 | T2], [H3 | T3]) ->
-   lists:reverse(rmerge3_1(L1, [], H2, T2, H3, T3), []).
+rmerge3([_|_]=L1, [H2 | T2], [H3 | T3]) ->
+   lists:reverse(rmerge3_1(L1, [], H2, T2, H3, T3), []);
+rmerge3([_|_]=L1, [_|_]=L2, []) ->
+    rmerge(L1, L2);
+rmerge3([_|_]=L1, [], [_|_]=L3) ->
+    rmerge(L1, L3);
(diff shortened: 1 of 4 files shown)


juhlig:
Nice 🤗

> It could suffice to just ensure that the first thing that is not an empty list is a list.

Yes, that is probably better.

bjorng:
I think this is a nice improvement of the `merge` functions.

Regarding the implementation of `merge/1` and `umerge/1`, perhaps it would better to not change those functions themselves, but instead add the extra error checking to `mergel/2` and `umergel/2`?

Maria-12648430:
> Regarding the implementation of `merge/1` and `umerge/1`, perhaps it would better to not change those functions themselves, but instead add the extra error checking to `mergel/2` and `umergel/2`?

I can try :slightly_smiling_face:

Maria-12648430:
@bjorng done, what do you think?

lib/stdlib/src/lists.erl
```diff
@@ -1789,25 +1889,25 @@ split_2_1(X, Y, [], R, Rs, S) ->
...
-mergel([A, B, [] | L], Acc) ->
-    mergel([A, B | L], Acc).
-
+mergel([[_|_]=A, [_|_]=B, [] | L], Acc) ->
+    mergel([A, B | L], Acc);
+mergel([[_|_]=T1, [H2 | T2], [H3 | T3] | L], Acc) ->
+    mergel(L, [merge3_1(T1, [], H2, T2, H3, T3) | Acc]);
+mergel([[_|_]=T1, [H2 | T2]], Acc) ->
+    rmergel([merge2_1(T1, H2, T2, []) | Acc], []).
+
```
bjorng, line 1910: Please remove the spaces on this line.

bjorng (commented):
**Looks** good to me, but I don't trust this update without a thoroughly updated test suite that feeds non-list values to the merge functions.

It might also be worth testing that if the input is a single non-empty list among empty lists, the non-empty list is indeed returned unchanged. That can be done by calling `erts_debug:same/2`. It returns `true` if and only if both its arguments are the same term.

lib/stdlib/src/lists.erl
Maria-12648430: Done

Maria-12648430:
I'll tackle the test suite tomorrow.

juhlig:
I'm helping with the tests...

> It might also be worth testing that if the input is a single non-empty list among empty lists, the non-empty list is indeed returned unchanged. That can be done by calling `erts_debug:same/2`. It returns `true` if and only if both its arguments are the same term.

@bjorng there is something strange going on which I don't understand.

In the shell:
```erlang
1> L=[a, b, c].
[a,b,c]
2> erts_debug:same(L, lists.merge([L])).
true
```
But if I do this in a test suite:
```erlang
L=[a, b, c],
true=erts_debug:same(L, lists:merge([L]),
```
... I get a `{badmatch, false}` error.
It doesn't matter if I insert more empty lists, in the shell I get `true` and in the test suite I get `false`.

However, it works as expected with the functions that accept two or three lists ( `merge/2` and `merge3/3`), ie
```erlang
L=[a, b, c],
true = erts_debug:same(L, lists:merge([], L)),
```
passes.

juhlig:
The latest commit extends common test cases and adds more property tests for invalid input.

bjorng:

bjorng:
This looks good in our daily builds. Please squash the commits.

Maria-12648430:
@bjorng
> This looks good in our daily builds.

Great :partying_face:

> Please squash the commits.

Done. Big commit message though, hope it captures all of what this was about.

<hr />

@juhlig
> The latest commit extends common test cases and adds more property tests for invalid input.

Thanks for your help :hugs:

juhlig:
> Thanks for your help :hugs:

Don't mention it :hugs:

bjorng:
Thanks for your pull request.

PR #6875, opened 2023-02-16
compiler: Remove dead code from private-append optimization

frej:
With the decision to not introduce a private append variant of `bs_create_bin` which gracefully tolerates non-binary string arguments, we can remove many code paths in `beam_ssa_private_append.erl` which are never reached as we will never try to track and patch the tail field of a cons.


bjorng (approved):

frej:
It turns out that this patch depends on #6855 and maybe #6862, will rebase when it has been merged.

bjorng:
I have merged your other three pull requests. You can rebase now.

frej:
Rebased.

bjorng:
Thanks!

PR #6882, opened 2023-02-17
leex column number support

MarkoMin:
Commit message:

    Added:
        Support for including column number in token, end and error locations.
        Variables in Rules: TokenCol and TokenLoc
        Leex options: error_location and tab_size
    
    Fixed:
        Bug in testcase OTP14285

Extension of my Spawnfest 2022 project. Feature is backward compatible, because column number is optional. Column numbers are tracked always. If `error_location` is set to `column`, column number just get filtered out in the API functions (`string`, `token`, `tokens`). That's why I set `{error_location, column}` in majority of tests cases.

Also, while editing test cases, I found what I think is a bug in testcase `opt14285_2`. If the binary string is not labeled with `utf8`, then `\x{400}` is printed as NULL character to the file. However test succeeded because NULL was on the both sides of the test, but column line was not checked. However, `\x{400}` takes 2 bytes and `NULL` takes 1 byte, so testcase indicated that something is wrong. Fix was to label binary string with `utf8`.

I'm not sure if `error_location` is THE name, because it covers both error location and end location of the parser. However, it shouldn't cover token location because it is determined by ".xrl" file, not leex options. If somebody got better name I'd like to hear it. 

I've added documentation of the feature, but if some things are unclear, hit me up.

Thanks to @peerst for reaching me out so I knew that somebody else is interested in this feature. Thanks to Invariant for giving me hours to open-source and ofc. thanks to all people around Spawnfest.


lib/parsetools/doc/src/leex.xml
```diff
@@ -115,6 +115,22 @@
               <p>Causes generated -file() attributes to only include
                 the basename of the file path.</p>
             </item>
+            <tag><c>{error_location, line | column}</c></tag>
+            <item>
+              <p>If set to <c>column</c>, error location will be 
+              <c>{Line,Column}</c> tuple instead of just <c>Line</c>. Also, <c>StartLoc</c>
+              and <c>EndLoc</c> in <c>string/2</c>, <c>token/3</c> and <c>tokes/3</c>
```
bjorng, line 122:
```suggestion
              and <c>EndLoc</c> in <c>string/2</c>, <c>token/3</c>, and <c>tokens/3</c> 
```

bjorng (changes requested):
This looks good to me, except for a typo in the documentation, and that you have added trailing whitespaces on many lines.

When fixing the documentation typo, please amend the commit so that there is still only one commit. Then please use use the following command to get rid of the trailing whitespace in your added or modified code:

    git rebase --whitespace=fix HEAD~1

Then force-push the branch.

bjorng:
Thanks for your pull request.

PR #6883, opened 2023-02-17
public_key: Move decode of CRLDistributionPoints extension

### LINKED ISSUE #6402: `pubkey_cert_records:decode_extensions` throws on malformed extensions

IngelaAndin:
As different solutions of verifying certificate revocation exists move the decode of 'CRLDistributionPoints' so that it will only be decode when it is actually used in the verification process. This would enable interoperability with systems that use certificates with an invalid empty CRLDistributionPoints extension that they want to ignore and make verification by other means.

Closes #6402


lib/public_key/src/pubkey_cert.erl
```diff
@@ -370,6 +370,9 @@ select_extension(_, asn1_NOVALUE) ->
     undefined;
 select_extension(_, []) ->
     undefined;
+select_extension(Id, [#'Extension'{extnID = ?'id-ce-cRLDistributionPoints' = Id,
+                                   extnValue = Value} = Extension | _]) when is_binary(Value) ->
+    Extension#'Extension'{extnValue = public_key:der_decode('CRLDistributionPoints', Value)};
 select_extension(Id, [#'Extension'{extnID = Id} = Extension | _]) ->
     Extension;
```
KennethL, line 377: Why is it only the extension ?id-ce-cRLDistributionPoints that is decoded here? Could there not be other extensions as well that is not decoded until used?
Would a test of `is_binary(Value)` imply that the extension needs to be decoded here?

IngelaAndin: This CRL extension is not decode initially  but then decode later when used by public_key functions.  Other  possibly unencoded  extensions are not know by public key and handled totally by some other code that will have to do the decoding. We could possible find a reason to add more known public_key extensions here later, but I do not think there is an obvious one at the moment.  Yes is_binary implies that  the value is DER encoded.

PR #6884, opened 2023-02-17
Removed unnecessary <em> tags

gertab:
The docs ([example](https://www.erlang.org/doc/man/erl_syntax_lib.html#analyze_function-1)) contain some emphasis tags (within code snippets) which end up visible in the documentation. I fixed these by removing the unwanted tags.


bjorng:
Thanks for your pull request.

PR #6888, opened 2023-02-17
Add best practice for atoms in NIFs

NAR:
Mention that atoms created in load are available in all environments. The best practice is to pre-create all of them before usage, make sure this best practice is included in the documentation.

Also fixed a "has" that should have been a "have".

### DIFF
--- a/erts/doc/src/erl_nif.xml
+++ b/erts/doc/src/erl_nif.xml
@@ -171,11 +171,37 @@ $> erl
           to query terms, like <c>enif_is_atom</c>, <c>enif_is_identical</c>,
           and <c>enif_compare</c>.</p>
         <p>All terms of type <c>ERL_NIF_TERM</c> belong to an environment of
-          type <seecref marker="#ErlNifEnv"><c>ErlNifEnv</c></seecref>. The
-          lifetime of a term is controlled by the lifetime of its environment
-          object. All API functions that read or write terms has the
-          environment that the term belongs to as the first function
-          argument.</p>
+          type <seecref marker="#ErlNifEnv"><c>ErlNifEnv</c></seecref>,
+          except atoms created during loading (by callbacks
+          <seecref marker="#load"><c>load</c></seecref> or
+          <seecref marker="#upgrade"><c>upgrade</c></seecref>). The lifetime of
+          a term is controlled by the lifetime of its environment object. All
+          API functions that read or write terms have the environment that the
+          term belongs to as the first function argument. However, the atoms
+          created during loading can be referred as a term in any <c>ErlNifEnv</c>.
+          That is, the best practice it to create all your atoms during
+          loading and store them in static/global variables, for example:</p>
+          <code type="none"><![CDATA[
+#include <erl_nif.h>
+
+ERL_NIF_TERM world_atom;
+
+static int load(ErlNifEnv* env, void** priv_data, ERL_NIF_TERM load_info)
+{
+    world_atom = enif_make_atom(env, "world");
+    return 0;
+}
+
+static ERL_NIF_TERM hello(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
+{
+    ERL_NIF_TERM hello_string = enif_make_string(env, "Hello", ERL_NIF_LATIN1);
+    return enif_make_tuple2(env, hello_string, world_atom);
+}
+
+static ErlNifFunc nif_funcs[] = { { "hello", 0, hello } };
+
+ERL_NIF_INIT(niftest, nif_funcs, load, NULL, NULL, NULL)
+]]></code>
       </item>
       <tag>Binaries</tag>
       <item>


erts/doc/src/erl_nif.xml
```diff
@@ -171,11 +171,26 @@ $> erl
...
+          API functions that read or write terms have the environment that the
+          term belongs to as the first function argument. However, the atoms
+          created during loading can be referred as a term in any ErlNifEnv.
+          That is, the best practice it to create all your atoms during
+          loading and store them in static/global variables, for example:</p>
+          <code type="none"><![CDATA[
+static ERl_NIF_TERM atom_ok;
+static ERl_NIF_TERM atom_error;
+
+static int load(ErlNifEnv* env, ...)
```
starbelly, line 188: Would a full example work better here? It would make the document longer is the downside, but we could expand with : 

```
#include <erl_nif.h>

ERL_NIF_TERM atom_world;

static int load(ErlNifEnv* env, void** priv_data, ERL_NIF_TERM load_info)
{
    atom_world = enif_make_atom(env, "world");
    return 0;
}

static ERL_NIF_TERM hello(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
{
    return enif_make_tuple2(env, enif_make_string(env, "Hello", ERL_NIF_LATIN1), atom_world);
}

static ErlNifFunc nif_funcs[] = { { "hello", 0, hello } };

ERL_NIF_INIT(nif_test, nif_funcs, load, NULL, NULL, NULL)
```

Something like that.

jhogberg:
Not to sound too picky, but documenting it this way makes it difficult to support EEP-20 or an “atom GC” in the future.

Pre-creating the atoms in an `ErlNifEnv` created at load-time and copying them out of that environment when needed is only slightly more expensive than storing them in plain global variables (the copy routine sees that there’s nothing to copy and just assigns them, so the difference in performance is just function call overhead), and avoids painting us into a corner.

sverker:
@jhogberg 
My reasoning was that such atoms created during nif loading should be considered as "literals" (or "global" according to [EEP-20](https://www.erlang.org/eeps/eep-0020)) and could be added to the set of atoms introduced by the module. That is, for EEP-20 they would never be garbage collected and for a global atom GC (if ever implemented) they would be kept alive by the module until purged.

erts/doc/src/erl_nif.xml
```diff
@@ -171,11 +171,26 @@ $> erl
           to query terms, like <c>enif_is_atom</c>, <c>enif_is_identical</c>,
           and <c>enif_compare</c>.</p>
         <p>All terms of type <c>ERL_NIF_TERM</c> belong to an environment of
-          type <seecref marker="#ErlNifEnv"><c>ErlNifEnv</c></seecref>. The
-          lifetime of a term is controlled by the lifetime of its environment
-          object. All API functions that read or write terms has the
-          environment that the term belongs to as the first function
-          argument.</p>
+          type <seecref marker="#ErlNifEnv"><c>ErlNifEnv</c></seecref>,
+          expect atoms created during loading (by callbacks
```
sverker, line 175:
```suggestion
          except atoms created during loading (by callbacks
```

erts/doc/src/erl_nif.xml
```diff
@@ -171,11 +171,26 @@ $> erl
...
+          <seecref marker="#load"><c>load</c></seecref> or
+          <seecref marker="#upgrade"><c>upgrade</c></seecref>). The lifetime of
+          a term is controlled by the lifetime of its environment object. All
+          API functions that read or write terms have the environment that the
+          term belongs to as the first function argument. However, the atoms
+          created during loading can be referred as a term in any ErlNifEnv.
```
sverker, line 181:
```suggestion
          created during loading can be referred as a term in any <c>ErlNifEnv</c>.
```

jhogberg:
> @jhogberg
> My reasoning was that such atoms created during nif loading should be considered as "literals" (or "global" according to [EEP-20](https://www.erlang.org/eeps/eep-0020)) and could be added to the set of atoms introduced by the module. That is, for EEP-20 they would never be garbage collected and for a global atom GC (if ever implemented) they would be kept alive by the module until purged.

Sure. If we're going down that road, wouldn't it make sense to turn all terms created in the `load`/`upgrade` environment into literals? I think it would be useful and we wouldn't have to distinguish between atoms and other terms in the documentation.

max-au:
> wouldn't it make sense to turn all terms created in the load/upgrade environment into literals

Sounds interesting. Right now, it's not exactly possible to create literals in the NIF itself, so I have to do some "weird stuff": create literals (textual error messages, etc.) in the NIF on_load function and then pass it into `load` (may sounds like another "best practice" if there is no way to create literals in the NIF itself).

It made me think however, maybe, in fact we _should be_ creating atoms (and all other literals) in the Erlang code, then passing it to NIFs. Just for simplicity.

sverker:
> It made me think however, maybe, in fact we should be creating atoms (and all other literals) in the Erlang code, then passing it to NIFs. Just for simplicity.

I think the solution with creating literals in load&upgrade is cleaner. It does not make Erlang code dependent on what is and is not a literal.
It will still be possible to pass down terms (literals or not) that load&upgrade can turn into literals with `enif_make_copy`.

sverker:
@NAR I propose you squash this into one commit for merge.

Then we'll do the more general solution for literals in a separate PR.

NAR:
@sverker I squashed (and rebased) the commits. If that rebase was not a good idea, I could try to revert that.

Issue #6074, opened 2022-06-10
mod_alias failed to handle re_write config


xning:
**Describe the bug**
mod_alias module of inets http server failed to hadnle re_write config.

**To Reproduce**
Start a http server to view erlang help doc as follows, then try open http://localhost:8080/doc/index.html

```erlang
inets:start().
inets:start(httpd, [{server_name, "localhost"},
                             {port, 8080}, {server_root, Tmpdir},
                             {document_root, "/usr/share/doc/erlang-24.3.4"},
                             {error_log, "error.log"},
                             {security_log, "secure.log"},
                             {transfer_log, "access.log"},
                             {re_write, {"^/(^[^/]+)/([+]+)/doc/(.*)$", "/\\1/\\2\\3"}},
                             {bind_address, "localhost"}]).
```

**Expected behavior**

http server return 505

**Affected versions**
erlang-24.3.4

**Additional context**

erlang-24.3.4-2.fc36.x86_64

I test the regular expression in re module, it works.

u3s:
- can you give some examples showing user URIs and rewriting result?
- why do you expect server returning HTTP 505 ?

PR #6892, opened 2023-02-19
fix 6074 (mod_alias failed to handle re_write config)

MarkoMin:
Fixed functionallity of re_write property (Issue #6074).
Previous code coudn't work because there was "^"++FakeName line,
   but FakeName could be MatchPattern so this would fail.
I changed the behaviour to append "^" while storeing the re_write, alias or script_alias proplist.

Also, I included one simple re_write property into test, to make sure everything works correcty.

### DIFF
--- a/lib/inets/src/http_server/mod_alias.erl
+++ b/lib/inets/src/http_server/mod_alias.erl
@@ -123,8 +123,7 @@ real_name(ConfigDB, RequestURI, [{MP,Replacement}| _] = Aliases)
 real_name(ConfigDB, RequestURI,  [{_,_}|_] = Aliases) ->
     case longest_match(Aliases, RequestURI) of
 	{match, {FakeName, RealName}} ->
-	    ActualName = re:replace(RequestURI,
-				    "^" ++ FakeName, RealName, [{return,list}]),
+	    ActualName = re:replace(RequestURI, FakeName, RealName, [{return,list}]),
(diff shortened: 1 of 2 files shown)


u3s:
Thanks for contribution and fix proposal.
Unfortunately, inets work has lower priority so it might take a while before we process it.

u3s:
please rebase the PR to latest maint and solve the conflict

u3s:
I think the PR looks good.
- Please let me know if it would be possible to cover also `script_re_write` option in `alias` test case?
- inets has low test coverage this would be a small step in wanted direction and inline with functionality touched in this PR

Such calls could be added in alias test case together with setting script_re_write option earlier:
GET /cgi-bin/erl/httpd_example:get
GET /cgi-UNWANTED-bin/erl/httpd_example:get

u3s:
> Is this what you meant? I'm not sure I understood "UNWANTED" part. Should it return 404?

:-D. I just thought about something that is pattern matched successfully. So we have a URI towards script that is expected to pattern match and rewritten by httpd. Success HTTP code is expected as well.

Sorry for confusion.

MarkoMin:
I misunderstood it... "/cgi-bin/erl/..." is actually for ESI scripts, while "script_re_write" is only for CGI scripts. I've amended changes to the latest commit, take a look :D

IngelaAndin (approved):

PR #6893, opened 2023-02-19
supervisor: export sup_ref type

tsloughter:
dialyzer in OTP-26 caught that I was using this unexported type `supervisor:sup_ref/0`, https://github.com/open-telemetry/opentelemetry-erlang/blob/6a29857f39ebf027bb77bc9a59d752d4a52eef11/apps/opentelemetry_experimental/src/otel_meter_server_sup.erl#L32

### DIFF
--- a/lib/stdlib/src/supervisor.erl
+++ b/lib/stdlib/src/supervisor.erl
@@ -56,9 +56,10 @@
 
 %%--------------------------------------------------------------------------
 
--export_type([sup_flags/0, child_spec/0, strategy/0,
-              startchild_ret/0, startchild_err/0,
-              startlink_ret/0, startlink_err/0]).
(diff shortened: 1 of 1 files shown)


IngelaAndin:
We will look into this, we of course want API types to be available but in the same time want to be causes in exporting old types so that we do not lock ourselves into corners by making type that is sort of opaque but still can be matched too specified so that the user start having the wrong expectation.  So that said it might take some time for us to have time to make the good decision.

tsloughter:
OK, thanks :+1:

kikofernandez:
Sorry for the delay in the answer. I am trying to see if instead of simply exporting this type, I can also unify the types of `supervisor`, `gen_server`, `gen_event`, and `gen_statem`.

kikofernandez:
Super-seeded by #7205

PR #6899, opened 2023-02-20
Optimize code:which/1 to use code path cache

josevalim:
We only optimize code:which/1 when a single
module is looked up. We still perform the
traversal on the client when looking up several
modules.

---

More context: I am helping improve the compilation
time of applications with more than a thousand modules
in a single `ebin` directory. The way code:which/1 works
today is very expensive in such cases, because it calls
`list_dir/1` on every directory along the way, and `list_dir/1`
will literally return thousands of entries. Perhaps the
`list_dir/1` could be removed altogether _but_ using
the new code cache is a better solution for when looking
up a single module anyway.


lib/kernel/src/code_server.erl
```diff
@@ -1178,12 +1182,19 @@ get_object_code(#state{path=Path} = St, Mod) when is_atom(Mod) ->
     ModStr = atom_to_list(Mod),
     case erl_prim_loader:is_basename(ModStr) of
         true ->
-            case mod_to_bin(Path, ModStr ++ objfile_extension(), []) of
-                {Binary, File, NewPath} ->
+            ModFile = ModStr ++ objfile_extension(),
+            case find_module(Path, [], ModFile, get_file) of
```
josevalim, line 1186:
```suggestion
            case find_module(Path, [], ModFile, fun erl_prim_loader:get_file/1) of
```

lib/kernel/src/code_server.erl
```diff
@@ -1246,34 +1257,45 @@ loader_down(#state{loading = Loading0} = St, {Mod, Bin, FName}) ->
             St
     end.
 
-mod_to_bin([{Dir, Cache0}|Tail], ModFile, Acc) ->
+which(#state{path=Path} = St, Mod) when is_atom(Mod) ->
+    ModStr = atom_to_list(Mod),
+    case erl_prim_loader:is_basename(ModStr) of
+        true ->
+            ModFile = ModStr ++ objfile_extension(),
+            case find_module(Path, [], ModFile, read_file_info) of
```
josevalim, line 1265:
```suggestion
            case find_module(Path, [], ModFile, fun erl_prim_loader:read_file_info/1) of
```

lib/kernel/src/code_server.erl
```diff
@@ -1246,34 +1257,45 @@ loader_down(#state{loading = Loading0} = St, {Mod, Bin, FName}) ->
...
+            {non_existing, St}
+    end.
+
+find_module([{Dir, Cache0}|Tail], Acc, ModFile, Fun) ->
     case with_cache(Cache0, Dir, ModFile) of
         {true, Cache1} ->
             File = filename:append(Dir, ModFile),
 
-            case erl_prim_loader:get_file(File) of
+            case erl_prim_loader:Fun(File) of
```
josevalim, line 1282:
```suggestion
            case Fun(File) of
```

josevalim:
@frazze-jobb I have rebased this PR. :)

frazze-jobb:
This breaks cover_SUITE:misc, cover_SUITE:compile, and code_SUITE:ext_mod_dep

josevalim:
Thank you, I will go ahead and close this then. My assumption was that this would not break anything, but it at least seem to make some return paths absolute instead of relative. I will explore other approaches later too. I'll be back. :D

josevalim:
Nevermind. A full recompile fixed all tests. Pushed another approach here, which should be simpler and faster: #7711

PR #6900, opened 2023-02-20
Enable decoding URIs with ipv6 address in host.

### LINKED ISSUE #6898: erlang:decode_packet/3 doesn't support IPv6 hosts
**Describe the bug**
`erlang:decode_packet/3` is used in the `bandit` library ([see here](https://github.com/mtrudel/bandit/blob/main/lib/bandit/http1/adapter.ex#L59)) to decode HTTP requests and pull the hostname out.

This doesn't work with IPv6 addresses in the host.

**To Reproduce**
Call `erlang:decode_packet/3` with the following arguments:

`erlang:decode_packet(http_bin, <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]/echo_components HTTP/1.1\r\nhost: orange\r\n\r\n">>, [])`

Result:

`{ok, {http_request, GET,  {absoluteURI, http, <<"[FEDC">>, undefined, <<"/echo_components">>}, {1, 1}}, <<"host: orange\r\n\r\n">>}`

**Expected behavior**
I expect the return value to extract the host in its entirety, `<<"[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]">>`

**Affected versions**
I tested using OTP 25.

**Additional context**

I believe the bug in question lives somewhere close to this part of the code: https://github.com/erlang/otp/blob/master/erts/emulator/beam/packet_parser.c#L556-L586 (in a BIF, unfortunately).

See for context https://github.com/mtrudel/bandit/pull/97

derekkraan:
Fixes #6898

### DIFF
--- a/erts/emulator/beam/packet_parser.c
+++ b/erts/emulator/beam/packet_parser.c
@@ -518,6 +518,7 @@ static void
 http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
 {
     const char* p;
+    const char* v;
     
     if ((p = memchr(uri_ptr, '/', uri_len)) == NULL) {
         /* host [":" port] */
@@ -533,15 +534,39 @@ http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
 
     uri->s1_ptr = uri_ptr;
     uri->port = 0; /* undefined */
-    /* host[:port]  */
     if ((p = memchr(uri_ptr, ':', uri_len)) == NULL) {
         uri->s1_len = uri_len;
     }
+    /* ipv6
+     * eg [::1]:4000
+     * eg [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:80
+     */
+    else if (memchr(uri_ptr, '[', uri_len) == uri_ptr &&
+        (v = memchr(uri_ptr, ']', uri_len)) != NULL) {
+      int n = (v - uri_ptr) + 1;
+      int port = 0;
+      uri->s1_len = n;
+      n = uri_len - (n+1);
+      // parse port if the next char is `:`
+      if (sys_memrchr(uri_ptr, ':', uri_len) == v + 1) {
+        // Skip over `]:`
+        v = v + 2;
+        while(n && isdigit((int) *v)) {
+            port = port*10 + (*v - '0');
+            n--;
+            v++;
+        }
+        if (n==0 && port!=0)
+            uri->port = port;
+      }
+    }
+    /* host[:port]  */
     else {
         int n = (p - uri_ptr);
         int port = 0;        
         uri->s1_len = n;
         n = uri_len - (n+1);
+        // Skip over port delimiter `:`
         p++;
         while(n && isdigit((int) *p)) {
             port = port*10 + (*p - '0');
--- a/erts/emulator/beam/packet_parser.c
+++ b/erts/emulator/beam/packet_parser.c
@@ -293,7 +293,7 @@ int packet_get_length(enum PacketParseType htype,
     case TCP_PB_LINE_LF: {
         /* TCP_PB_LINE_LF:  [Data ... Delimiter]  */
         const char* ptr2;
-        if ((ptr2 = memchr(ptr, delimiter, n)) == NULL) {
+        if ((ptr2 = sys_memchr(ptr, delimiter, n)) == NULL) {
             if (n > max_plen && max_plen != 0) { /* packet full */
                 DEBUGF((" => packet full (no NL)=%d\r\n", n));
                 goto error;
@@ -407,7 +407,7 @@ int packet_get_length(enum PacketParseType htype,
 	    }
 
             while (1) {
-                const char* ptr2 = memchr(ptr1, '\n', len);
+                const char* ptr2 = sys_memchr(ptr1, '\n', len);
                 
                 if (ptr2 == NULL) {
                     if (max_plen != 0) {
@@ -520,7 +520,7 @@ http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
     const char* p;
     const char* v;
     
-    if ((p = memchr(uri_ptr, '/', uri_len)) == NULL) {
+    if ((p = sys_memchr(uri_ptr, '/', uri_len)) == NULL) {
         /* host [":" port] */
         uri->s2_ptr = "/";
         uri->s2_len = 1;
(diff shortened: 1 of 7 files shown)


erts/emulator/beam/packet_parser.c
```diff
@@ -533,15 +533,37 @@ http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
...
-    /* host[:port]  */
     if ((p = memchr(uri_ptr, ':', uri_len)) == NULL) {
         uri->s1_len = uri_len;
     }
+    /* ipv6
+     * eg [::1]:4000
+     * eg [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:80
+     */
+    else if (memchr(uri_ptr, '[', uri_len) == uri_ptr) {
+      if ((p = memchr(uri_ptr, ']', uri_len)) != NULL) {
```
sverker, line 544: I don't think this is correct.
If there is a `[` without terminating `]` then `uri->s1_len` will not be set leading to potential disaster.

It should probably be combined into one if-statement.

erts/emulator/beam/packet_parser.c
```diff
@@ -533,15 +533,37 @@ http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
...
+        int n = (p - uri_ptr) + 1;
+        int port = 0;        
+        uri->s1_len = n;
+        n = uri_len - (n+1);
+        // Need to skip over `]:`
+        p = p + 2;
+        while(n && isdigit((int) *p)) {
```
sverker, line 551: Here we blindly skip over an assumed `:` after `]`.

There are no error cases. Maybe it's better to let port number be `undefined` instead of potentially wrong in case of missing `:`.

derekkraan:
Thanks for the feedback, I'll take a look this week hopefully and see if I can address your comments.

erts/emulator/beam/packet_parser.c
derekkraan: I combined both if statements into a single one in the latest commit.

erts/emulator/beam/packet_parser.c
derekkraan: I added a check here to skip over port parsing if the next character after the `]` isn't a `:`

derekkraan:
@sverker I pushed another commit, this should cover the cases. Now it checks for both `[` and `]` in a single if statement and only parses the port out of integers coming after the `]:` (now also checking for the presence of the `:`).

sverker:
Sorry, a missing `]` still fails hard:
```
3> erlang:decode_packet(http_bin, <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210:4000/echo_components HTTP/1.1\r\nhost: orange\r\n\r\n">>, []).
Segmentation fault (core dumped)
```
You will get here with `p` as NULL:
```
/* host[:port]  */
    else {
        int n = (p - uri_ptr);
```

Add the "missing ]" into the test case.

derekkraan:
You're absolutely correct. I have forgotten the intricacies of C after so many years of not using it.

We can choose between running `memchr` a second time (because we are overwriting it with the location of `]`, and need it to represent the location of `:`), or adding a new variable. I chose to add a new variable. Let me know what you think.

erts/emulator/beam/packet_parser.c
```diff
@@ -533,15 +534,38 @@ http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
...
     uri->port = 0; /* undefined */
-    /* host[:port]  */
     if ((p = memchr(uri_ptr, ':', uri_len)) == NULL) {
         uri->s1_len = uri_len;
     }
+    /* ipv6
+     * eg [::1]:4000
+     * eg [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:80
+     */
+    else if (memchr(uri_ptr, '[', uri_len) == uri_ptr && (v = memchr(uri_ptr, ']', uri_len)) != NULL) {
```
sverker, line 544: We usually try keep lines under 80 characters.

erts/emulator/test/decode_packet_SUITE.erl
```diff
@@ -206,6 +206,17 @@ pack_ssl(Content, Major, Minor, Body) ->
...
 
+ipv6(Config) when is_list(Config) ->
+    Packet = <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:4000/echo_components HTTP/1.1\r\nhost: orange\r\n\r\n">>,
+    {ok, {http_request, 'GET',  {absoluteURI, http, <<"[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]">>, 4000, <<"/echo_components">>}, {1, 1}}, <<"host: orange\r\n\r\n">>} =
+      erlang:decode_packet(http_bin, Packet, []),
+    Packet2 = <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]/1234 HTTP/1.1\r\nhost: orange\r\n\r\n">>,
+    {ok, {http_request, 'GET',  {absoluteURI, http, <<"[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]">>, undefined, <<"/1234">>}, {1, 1}}, <<"host: orange\r\n\r\n">>} =
+      erlang:decode_packet(http_bin, Packet2, []),
+    Packet3 = <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210:4000/echo_components HTTP/1.1\r\nhost: orange\r\n\r\n">>,
+    logger:info(erlang:decode_packet(http_bin, Packet3, [])),
```
sverker, line 217: Remove redundant logger:info line.

erts/emulator/test/decode_packet_SUITE.erl
```diff
@@ -206,6 +206,17 @@ pack_ssl(Content, Major, Minor, Body) ->
...
     {Res, {ssl_tls,[],C,{Major,Minor}, Data}}.
 
+ipv6(Config) when is_list(Config) ->
+    Packet = <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:4000/echo_components HTTP/1.1\r\nhost: orange\r\n\r\n">>,
+    {ok, {http_request, 'GET',  {absoluteURI, http, <<"[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]">>, 4000, <<"/echo_components">>}, {1, 1}}, <<"host: orange\r\n\r\n">>} =
+      erlang:decode_packet(http_bin, Packet, []),
+    Packet2 = <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]/1234 HTTP/1.1\r\nhost: orange\r\n\r\n">>,
+    {ok, {http_request, 'GET',  {absoluteURI, http, <<"[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]">>, undefined, <<"/1234">>}, {1, 1}}, <<"host: orange\r\n\r\n">>} =
+      erlang:decode_packet(http_bin, Packet2, []),
+    Packet3 = <<"GET http://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210:4000/echo_components HTTP/1.1\r\nhost: orange\r\n\r\n">>,
```
sverker, line 216: Add comment of testing missing `]`.

sverker (commented):
Could you write something in the documentation of `erlang:decode_packet` about how the `Host` string can be an ipv6 address within `[]` from OTP 26.0.

derekkraan:
Yes, I can do that.

derekkraan:
I added that requested documentation @sverker, hope it's sufficient. I also added a link to the RFC there.

erts/emulator/beam/packet_parser.c
```diff
@@ -533,15 +534,39 @@ http_parse_absoluteURI(PacketHttpURI* uri, const char* uri_ptr, int uri_len)
...
+     * eg [FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:80
+     */
+    else if (memchr(uri_ptr, '[', uri_len) == uri_ptr &&
+        (v = memchr(uri_ptr, ']', uri_len)) != NULL) {
+      int n = (v - uri_ptr) + 1;
+      int port = 0;
+      uri->s1_len = n;
+      n = uri_len - (n+1);
+      // parse port if the next char is `:`
+      if (memrchr(uri_ptr, ':', uri_len) == v + 1) {
```
sverker, line 551: `memrchr` is a GNU extension not available on Windows:
`
packet_parser.o : error LNK2019: unresolved external symbol memrchr referenced in function http_parse_absoluteURI
`

sverker: I will fix this

PR #6902, opened 2023-02-20
Ensure that the bs_match instruction can be extended

bjorng:
If we were to introduce new sub commands for the `bs_match` instruction in the future, we want to make sure that loading a BEAM file produced by a future Erlang/OTP version will not crash the runtime when loaded into an earlier version. Therefore, fail loading gracefully if any unknown sub commands are found. For example, if we were to introduce support for matching floats in a future release, loading would fail like so if we try to load that module in Erlang/OTP 26:

    =ERROR REPORT==== 20-Feb-2023::13:03:21.664064 ===
    beam/beam_load.c(583): Error loading function bad_bs_match:f/1:
      bs_match: invalid sub instruction 'float' (is this BEAM file generated by a future version of Erlang/OTP?)


erts/emulator/beam/jit/arm/generators.tab
```diff
@@ -571,6 +571,50 @@ gen.bs_match(Fail, Ctx, N, List) {
     int fixed_args;
     int i;
 
+    /*
+     * If a BEAM file produced by a later version of Erlang/OTP
+     * is accidentally loaded into an earlier version, ensure
+     * that the loading fails (as opposed to crashing the runtime)
+     * if there are any unknown sub commands.
+     */
+    i = 0;
```
jhogberg, line 580: Can we break this part out into a common macro in the top-level `generators.tab`? It seems a bit repetitive.

jhogberg (approved):

erts/emulator/beam/jit/arm/generators.tab
bjorng: Yes, it is repetitive, but it is only used by the two JITs, not the emulator.

PR #6905, opened 2023-02-20
compiler: Fix mishandling of make_fun in alias analysis

### LINKED ISSUE #6890: segmentation fault when appending binary in recursive call
**Describe the bug**

Getting a `segmentation fault` when appending a binary twice in a recursive function:
- once to build the accumulator parameter
- once to build the return value

**To Reproduce**

Here is a minimal reproduction based on the original error found in [Elixir](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/lib/io/ansi/docs.ex#L259-L267):

```erlang
-module(erl_repro).
-export([segfault/0]).

segfault() ->
  format([<<"hello">>, <<"world">>], <<"">>).

format(List, Indent) when is_list(List) ->
  lists:map(fun (X) -> [<<Indent/binary,"  ">>, format(X, <<Indent/binary,"    ">>) ] end, List);
format(Binary, Indent) when is_binary(Binary) ->
  [Binary, Indent].
```

**Expected behavior**
No segfault.

**Affected versions**
```
Erlang/OTP 26 [RELEASE CANDIDATE 1] [erts-14.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]
```

frej:
When a value is copied into the environment of a fun, alias analysis should assume that it has been aliased as there is no obvious way to track and ensure that the value is only used once, even if the argument dies at the location of the make_fun.

Closes #6890

### DIFF
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -324,150 +324,150 @@ aa_fun(F, #opt_st{ssa=Linear0,anno=_Anno,args=Args}=St,
                         aa_new_ssa_var(Var, Status, Acc)
                 end, #{}, ArgsStatus),
     ?DP("@@ Args: ~p~n", [ArgsStatus]),
-    {Linear1,SS} = aa_blocks(Linear0, SS0, AAS0),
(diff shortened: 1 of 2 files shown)


bjorng:

bjorng:
Thanks for your pull request.

Issue #6903, opened 2023-02-20
deprecate `dbg:stop_clear` since it is a wrapper around `dbg:stop`


kikofernandez:
**Describe the bug**
The function `dbg:stop_clear` should be deprecated, as per the source code documentation comments.

**Expected behavior**
Removal of `dbg:stop_clear` in future releases, so as to remove non-needed wrappers.

**Affected versions**
OTP-24+

PR #6906, opened 2023-02-20
dbg: deprecates function dbg:stop_clear/1

kikofernandez:
function `dbg:stop_clear/1` is not documented in the API but was kept for compatibility reasons. in this commit we make a step forward to deprecating its use such that it will be eventually removed.

This link shows the lines with the reason why I think this function should be deprecated
(https://github.com/erlang/otp/pull/6906/files#diff-7c55527e108fff46bd5f57925d681195f2b4a127a6f3e247c61332e1e8fa743bR618-R624)

image

closes GH-6903.


sverker:
Should there be a
`-deprecated([{stop_clear,0, "use dbg:stop/0 instead"}]).`
in dbg.erl?

kikofernandez:
Absolutely, I have missed that one! Thanks!
I have pushed it as a new commit but I will squash it in the end to have everything in a single commit
Thanks!

sverker (approved):

kikofernandez:
Closing and reopening a new PR to see if the Github actions run correctly

PR #6907, opened 2023-02-20
New log type 'rotate' for disk_log

anupamasingh10:
Uses a sequence of rotate log files of limited size. As a log file is filled up, it's rotated and then compressed.


KennethL:
I think the naming scheme of rotate_logs should be described in the documentation, for example here in the disk_log:open function.:
`{file, FileName}`
Specifies the name of the file to be used for logged terms. If this value is omitted and the log name is an atom or a string, the filename defaults to `lists:concat([Log, ".LOG"])` for halt logs.

For wrap logs, this is the base name of the files. Each file in a wrap log is called <base_name>.N, where N is an integer. Each wrap log also has two files called <base_name>.idx and <base_name>.siz.

PR #6908, opened 2023-02-21
compiler: Improve code coverage in alias analysis

frej:
Add a test case which exercises the portions of the alias analysis which deals with a non-converging fixpoint iteration.


bjorng:

bjorng:
Thanks for your pull request.

PR #6685, opened 2023-01-16
Include timing on -init_debug to further aid debugging

josevalim:
This is a snippet of how it looks like:

```erlang
{progress,preloaded}
{progress,kernel_load_completed}
{progress,modules_loaded}
{start,heart}
{'done_in_µs',17}
{start,logger}
{'done_in_µs',1473}
{start,application_controller}
{'done_in_µs',19451}
{progress,init_kernel_started}
{apply,{application,load,[{application,stdlib,[{description,"ERTS  CXC 138 10"},{vsn,"4.2"},{id,[]},{modules,[array,base64,beam_lib,binary,c,calendar,dets,dets_server,dets_sup,dets_utils,dets_v9,dict,digraph,digraph_utils,edlin,edlin_context,edlin_expand,edlin_type_suggestion,epp,eval_bits,erl_abstract_code,erl_anno,erl_bits,erl_compile,erl_error,erl_eval,erl_expand_records,erl_features,erl_internal,erl_lint,erl_parse,erl_posix_msg,erl_pp,erl_scan,erl_stdlib_errors,erl_tar,error_logger_file_h,error_logger_tty_h,escript,ets,file_sorter,filelib,filename,gb_trees,gb_sets,gen,gen_event,gen_fsm,gen_server,gen_statem,io,io_lib,io_lib_format,io_lib_fread,io_lib_pretty,lists,log_mf_h,maps,math,ms_transform,orddict,ordsets,otp_internal,peer,pool,proc_lib,proplists,qlc,qlc_pt,queue,rand,random,re,sets,shell,shell_default,shell_docs,slave,sofs,string,supervisor,supervisor_bridge,sys,timer,unicode,unicode_util,uri_string,win32reg,zip]},{registered,[timer_server,rsh_starter,take_over_monitor,pool_master,dets]},{applications,[kernel]},{optional_applications,[]},{included_applications,[]},{env,[]},{maxT,infinity},{maxP,infinity}]}]}}
{'done_in_µs',13}
[...]

PR #6911, opened 2023-02-22
init: Avoid outputting non-ASCII characters from -init_debug

bjorng:
ba18068dc28693 (#6685) added printing of timing for the `-init_debug` option:

    {start,heart}
    {'done_in_µs',10}

With that change, the test case `peer_SUITE:init_debug/1` started to fail sporadically. That test case tries to start a peer node with option `-init_debug` using the connection type `standard_io`:

    {ok, Peer, _Node} = peer:start_link(#{name => peer:random_name(?FUNCTION_NAME),
        shutdown => 1000,
        connection => standard_io, args => ["-init_debug"]}),

The test case would fail when the two bytes that comprise the `µ` character were split up into two separate IO requests.

We could update the `peer` module so that it would do buffering and re-combine split UTF-8 sequences. That would fix the failing test case. However, introducing re-combining is not sufficient to make it safe to print arbitrary non-ASCII characters, because of an undocumented restriction mentioned in a comment for the `standard_io` connection type:

    %% Characters in range of 192-255 are reserved for control sequences,
    %%  see encode_port_data for details. If peer node attempts to print
    %%  characters in this range, an controlling process on the origin
    %%  node may terminate with an error (because CRC check will fail).

(The actual restriction is actually slightly less severe than the comment implies, which is why `µ` would work after re-combining but others characters such as `ö` would not. Note that the other connection modes don't have this restriction, but they are not able to catch the output from `-init_debug`.)

Since printing of arbitrary non-ASCII characters cannot be made safe for the `standard_io` connection mode, change the printout of timing information to only use ASCII characters:

    {start,heart}
    {done_in_microseconds,10}


josevalim:
That's what happens when I try to be fancy!

PR #6913, opened 2023-02-22
ssl: fix keylog mechanism

u3s:
- fetch application traffic secret from TLS sender process

### DIFF
--- a/lib/ssl/src/ssl_gen_statem.erl
+++ b/lib/ssl/src/ssl_gen_statem.erl
@@ -1897,7 +1897,8 @@ connection_info(#state{handshake_env = #handshake_env{sni_hostname = SNIHostname
 
 security_info(#state{connection_states = ConnectionStates,
                      static_env = #static_env{role = Role},
-                     ssl_options = #{keep_secrets := KeepSecrets}}) ->
+                     ssl_options = #{keep_secrets := KeepSecrets},
+                     protocol_specific = ProtocolSpecific}) ->
     ReadState = ssl_record:current_connection_state(ConnectionStates, read),
     #{security_parameters :=
 	  #security_parameters{client_random = ClientRand,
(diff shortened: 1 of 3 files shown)


lib/ssl/src/ssl_gen_statem.erl
```diff
@@ -1911,10 +1912,18 @@ security_info(#state{connection_states = ConnectionStates,
...
-                                       client_early_data_secret = ClientEarlyData
-                                      }} =
+                  #security_parameters{
+                     application_traffic_secret = AppTrafSecretWrite0,
+                     client_early_data_secret = ClientEarlyData}} =
                 ssl_record:current_connection_state(ConnectionStates, write),
+            Sender = maps:get(sender, ProtocolSpecific, undefined),
+            AppTrafSecretWrite =
+                case call(Sender, get_application_traffic_secret) of
+                    {ok, SenderAppTrafSecretWrite} ->
```
IngelaAndin, line 1922: This is a blocking call for something that I do note believe needs synchronization between the sender and receiver process! I believe we want to avoid that. You could solve it in several ways. Probably the best way is to let the API function ask both processes and combine the answer. It would also be possible to have a "cast and bang"  between the sender and receiver process although requiring the receiver to save the "From-reference" to reply the user later. It is also possible to pass the From-reference to the sender with some data and make it do the reply (but that makes the code harder to understand).

dgud: Hmm, I think I would prefer forwarding the partial answer and From in that case, otherwise we need another
field in the state, and do the gen_server:reply from sender process.

u3s: I pushed updated proposal. Please have a look. Senders PID is delivered to API user process, which tries to fetch/update current key before returning value to user.

PR #6914, opened 2023-02-23
Compiler: alias analysis speedup

frej:
This is a set of patches which speeds up the alias analysis pass by reducing the number of iterations required to find a fix point and reduces the amount of work in each iteration.

### DIFF
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -318,7 +318,7 @@ aa_fixpoint([], Order, _OldAliasMap,
     aa_fixpoint(NewOrder, Order, AliasMap,
                 AAS#aas{repeats=sets:new([{version,2}])}, Limit - 1).
 
-aa_fun(F, #opt_st{ssa=Linear0,anno=_Anno,args=Args}=St,
+aa_fun(F, #opt_st{ssa=Linear0,args=Args}=St,
        AAS0=#aas{alias_map=AliasMap0,func_db=FuncDb,repeats=Repeats0}) ->
     %% Initially assume all formal parameters are unique for a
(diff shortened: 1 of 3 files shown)


RobinMorisset:
This looks interesting! Have you done any benchmarking confirming that it is a performance win?

lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -325,11 +330,22 @@ aa_fun(F, #opt_st{ssa=Linear0,anno=_Anno,args=Args}=St,
...
+    ?DP("SS:~n~s~n~n", [SS]),
     AAS = aa_merge_call_args_status(SS, AAS1),
 
     AliasMap = AliasMap0#{ F => SS },
-    {St#opt_st{ssa=Linear1}, AAS#aas{alias_map=AliasMap}}.
+    PrevSS = maps:get(F, AliasMap0, #{}),
+    Repeats = case PrevSS =/= SS of
+                  true ->
+                      %% Alias status has changed, so schedule both
+                      %% our callers and callees for renews analysis.
```
bjorng, line 341:
```suggestion
                      %% our callers and callees for renewed analysis.
```

bjorng (approved):
Looks good, except for a typo in a comment.

lib/compiler/src/beam_ssa_alias.erl
frej: Fixed and force-pushed.

bjorng:

frej:
> This looks interesting! Have you done any benchmarking confirming that it is a performance win?

For the set of modules compiled by the `diffable`-script these patches reduces the time spent in the alias pass with around 20%.

bjorng:
Thanks for your pull request.

PR #6918, opened 2023-02-24
Fix factual errors and omissions for the bit syntax

### LINKED ISSUE #6706: bitstring-unit:X - discrepancy between documentation and implementation
**Describe the bug**

According to https://www.erlang.org/doc/reference_manual/expressions.html#bit-syntax-expressions

> Unit= unit:IntegerLiteral
> The allowed range is 1..256. Defaults to 1 for integer, float, and bitstring, and to 8 for binary. No unit specifier must be given for the types utf8, utf16, and utf32.

However, for `bitstring-unit:X` compiler rejects anything but `1`.

**To Reproduce**

```erlang
-module(bitsting_unit).

make(B) -> <<B:42/bitstring>>.
make_unit(B) -> <<B:42/bitstring-unit:3>>.

match(<<B:42/bitstring>>) -> B.
match_unit(<<B:42/bitstring-unit:3>>) -> B.
```

```
erlc bitstring.erl
bitsting_unit.erl:5:19: conflict in unit specification for bit field: '3' and '1'
%    5| make_unit(B) -> <<B:42/bitstring-unit:3>>.
%     |                   ^

bitsting_unit.erl:8:14: conflict in unit specification for bit field: '3' and '1'
%    8| match_unit(<<B:42/bitstring-unit:3>>) -> B.
%     |              ^

```

**Expected behavior**
No discrepancy between the documentation and the actual behavior.

**Affected versions**
23, 24, 25

bjorng:
While at it, reorganize the documentation for the bit syntax into separate sections for each type of segment.

Closes #6706

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1390,14 +1390,27 @@ Ei = Value |
       <item>For <c>binary</c> and <c>bitstring</c> it is
       the whole binary or bit string.</item>
     </list>
-    <p>In matching, this default value is only
-    valid for the last element. All other bit string or binary
-    elements in the matching must have a size specification.</p>
+    <p>In matching, the default value for a binary or bit string
+    segment is only valid for the last element. All other bit string
+    or binary elements in the matching must have a size
+    specification.</p>
+
+    <p><strong>Example:</strong></p>
+
+    <pre>
+1> <input>&lt;&lt;A/binary, B/binary>> = &lt;&lt;"abcde">>.</input>
+* 1:3: a binary field without size is only allowed at the end of a binary pattern
+2> <input>&lt;&lt;A:3/binary, B/binary>> = &lt;&lt;"abcde">>.</input>
+&lt;&lt;"abcde">>
+3> <input>A.</input>
+&lt;&lt;"abc">>
+4> <input>B.</input>
+&lt;&lt;"de">></pre>
 
     <p>For the <c>utf8</c>, <c>utf16</c>, and <c>utf32</c> types,
     <c>Size</c> must not be given. The size of the segment is implicitly
     determined by the type and value itself.</p>
-    
+
     <p><c>TypeSpecifierList</c> is a list of type specifiers, in any
     order, separated by hyphens (-). Default values are used for any
     omitted type specifiers.</p>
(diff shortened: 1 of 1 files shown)


system/doc/reference_manual/expressions.xml
```diff
@@ -1390,14 +1390,15 @@ Ei = Value |
       <item>For <c>binary</c> and <c>bitstring</c> it is
       the whole binary or bit string.</item>
     </list>
-    <p>In matching, this default value is only
-    valid for the last element. All other bit string or binary
-    elements in the matching must have a size specification.</p>
+    <p>In matching, the default value for a a binary or bit string
```
kikofernandez, line 1393: double a's `the default value for a a binary`.

system/doc/reference_manual/expressions.xml
```diff
@@ -1390,14 +1390,15 @@ Ei = Value |
...
+    segment is only valid for the last element. All other bit string
+    or binary elements in the matching must have a size
+    specification.</p>
```
kikofernandez, line 1396: (This is a suggestion, take it or leave it)
I think this is understandable but, for newcomers, I would add here an example that shows what you mean. That is, an example where you show other bit string / binary elements in the matching with their size specification and also the last element with the specification.

system/doc/reference_manual/expressions.xml
```diff
@@ -1423,60 +1424,142 @@ Ei = Value |
...
-    <c><![CDATA[<<$a/utf8,$b/utf8,$c/utf8>>]]></c>.</p>
+    <section>
+      <title>Integer segments</title>
+      <p>The value of <c>Size</c> multiplied with the unit gives the
+      size of the segment in bits.</p>
+
+      <p>When constructing binaries, if the size <c>N</c> of an integer
+      segment is too small to contain the given integer, the most significant
+      bits of the integer are silently discarded and only the <c>N</c> least
+      significant bits are put into the binary. For example, <c>&lt;&lt;16#ff:4&gt;&gt;</c>
```
kikofernandez, line 1444: Perfect! A simple example that guides to reader to unpack the sentence from above.
Love it!

system/doc/reference_manual/expressions.xml
```diff
@@ -1423,60 +1424,142 @@ Ei = Value |
...
+    </section>
+
+    <section>
+      <title>Float segments</title>
+      <p>The value of <c>Size</c> multiplied with the unit gives
+      the size of the segment in bits. The size of a float segment in bits must be
+      one of 16, 32, or 64.</p>
+
+      <p>When constructing binaries, if the size of a float segment is too small
+      to contain the representation of the given float value, an exception is raised.</p>
```
kikofernandez, line 1455: Not related to the PR but why are we not consistent between integers and floats?
In the Integer section:
> if the size <c>N</c> of an integer segment is too small to contain the given integer, the most significant bits of the integer are silently discarded and only the <c>N</c> least significant bits are put into the binary.

Should we make it consistent to raise an error if the segment is too small in integers? Apart from consistency, if the user writes an integer segment that is too small, getting an error is probably better than not knowing why the most significant bits are not part of the binary? Essentially, I question whether we should open up a ticket to fix this discrepancy in the current behaviour between integers and floats.

system/doc/reference_manual/expressions.xml
```diff
@@ -1423,60 +1424,142 @@ Ei = Value |
...
+      one of the segment types <c>binary</c>, <c>bitstring</c>,
+      <c>bytes</c>, and <c>bits</c>.</p>
+
+      <p>When constructing binaries and no size is specified for a
+      binary segment, the entire binary value is interpolated into the
+      binary being constructed. However, the size in bits of the
+      binary being interpolated must be evenly divisible by the unit
+      value for the segment; otherwise an exception is raised.</p>
+
+      <p>For example, attempting to interpolate a bit string of size 1 into
```
kikofernandez, line 1473: Again, I think this example is great and helps the reader in understanding.
I wonder if we should add also an example that succeeds, so that the difference is clear.
AFAIK and from the docs, that should be something of (e.g.,) size 8 and unit 1.

kikofernandez (changes requested):
I think this PR takes the documentation for bit syntax to another level. Well done!
The comments I wrote are mostly optional and I simply request the changes for a misspelling (first comment). The other comments are suggestions.

system/doc/reference_manual/expressions.xml
bjorng: Changing that would be incompatible. There is code that depend on the current behavior. What we could do is to come up with a new flag for a segment to cause an error if an integer value does not fit.

kikofernandez: right, so we cannot remove it but we can prevent it so that users get consistency (if so they choose).
I think this is a good trade-off. We cannot remove it because we rely on it, but users can get a consistent behaviour.
To me, this sounds good :)

system/doc/reference_manual/expressions.xml
bjorng: I will add an example.

system/doc/reference_manual/expressions.xml
bjorng: Will add.

system/doc/reference_manual/expressions.xml
bjorng: Will fix.

system/doc/reference_manual/expressions.xml
bjorng: The hardest part of doing that is finding a good name for the flag. :wink:

kikofernandez (approved):

PR #6920, opened 2023-02-24
gh: Update GitHub actions setup

TD5:
- Switch to new Docker container namespace
- Update actions dependencies, and keep them updated with a new action updater action

The new container namespace (`ghcr.io`) now supports `GITHUB_TOKEN` (see https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry), allowing us to migrate. When debugging locally, I found the new namespace was more reliable.

To see more about the migration process in general, visit: https://docs.github.com/en/packages/working-with-a-github-packages-registry/migrating-to-the-container-registry-from-the-docker-registry


TD5:
Looks like this PR is a duplicate of #6808. Might be worth merging the contents of both.

PR #6924, opened 2023-02-26
ms_transform: allow caller_line and current_stacktrace action functions

gomoripeti:
caller_line/0 was added in https://github.com/erlang/otp/pull/5305, while current_stacktrace/0,/1 were added in
https://github.com/erlang/otp/pull/6628


bjorng:
Thanks for your pull request.

PR #6929, opened 2023-02-27
Address a few omissions in documentation of comprehensions

### LINKED ISSUE #6454: Undocumented scoping rules in list comprehensions
I don't know whether this is a bug, or (more likely imho) something missing in the documentation or something that I missed.
On master, the following code:
```erlang
f() ->
  [Y || _ <- (Y = 42)].
```
is rejected with
```
test.erl:7:4: variable 'Y' is unbound
%    7|   [Y || _ <- (Y = 42)].
%     |    ^
```

whereas `[Y || Y <- 42]` naturally is fine.

I first thought that this means that variables bound in an expression in a generator are only bound for the rest of that comprehension's generators. But the following: `[Y || _ <- (Y = 42), Y <- Y]` is also rejected, showing that it is even more restrictive.

My current guess is that they are bound only for the duration of that expression itself. So I have 3 questions:
- Is my guess correct?
- If so, is it intended?
- If so, did I miss it in the documentation? (I checked https://www.erlang.org/doc/reference_manual/expressions.html#list-comprehensions and only found the following snippet: "The variables in the generator patterns shadow previously bound variables, including variables bound in a previous generator pattern.")

bjorng:
* Document that variables bound in a generator expression is not visible outside that expression.

* Document the failure modes for filters.

Closes #6454

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1985,29 +1985,31 @@ KeyPattern := ValuePattern &lt;- MapExpression</pre>
     marker="stdlib:maps#iterator/2">maps:iterator/2</seemfa>.</p>
 
     <p>A <em>filter</em> is an expression that evaluates to
-    <c>true</c> or <c>false</c>, or a
-    <seeguide marker="#guard_expressions">guard expression</seeguide>.
-    If the filter is not a guard expression and evaluates
-    to a non-Boolean value <c>Val</c>, an exception
-    <c>{bad_filter, Val}</c> is triggered at runtime.</p>
+    <c>true</c> or <c>false</c>.</p>
 
     <p>The variables in the generator patterns shadow previously bound variables,
     including variables bound in a previous generator pattern.</p>
 
-    <p>A list comprehension returns a list, where the lists elements are the
+    <p>Variables bound in a generator expression are not visible outside the expression:</p>
+
+    <pre>
+1> <input>[{E,L} || E &lt;- L=[1,2,3]].</input>
+* 1:5: variable 'L' is unbound</pre>
+
+    <p>A <strong>list comprehension</strong> returns a list, where the list elements are the
     result of evaluating <c>Expr</c> for each combination of generator
     elements for which all filters are true.</p>
 
-    <p>A bit string comprehension returns a bit string, which is
+    <p>A <strong>bit string comprehension</strong> returns a bit string, which is
     created by concatenating the results of evaluating <c>BitStringExpr</c> for
     each combination of bit string generator elements for which all
     filters are true.</p>
 
-    <p>A map comprehension returns a map, where the map elements are
-    the result of evaluating <c>KeyExpr</c> and <c>ValueExpr</c> for
-    each combination of generator elements for which all filters are
-    true. If the key expressions are not unique, the last occurrence
-    is stored in the map.</p>
+    <p>A <strong>map comprehension</strong> returns a map, where the
+    map elements are the result of evaluating <c>KeyExpr</c> and
+    <c>ValueExpr</c> for each combination of generator elements for
+    which all filters are true. If the key expressions are not unique,
+    the last occurrence is stored in the map.</p>
 
     <p><strong>Examples:</strong></p>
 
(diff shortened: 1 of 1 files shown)


system/doc/reference_manual/expressions.xml
```diff
@@ -1985,29 +1985,31 @@ KeyPattern := ValuePattern &lt;- MapExpression</pre>
...
-    If the filter is not a guard expression and evaluates
-    to a non-Boolean value <c>Val</c>, an exception
-    <c>{bad_filter, Val}</c> is triggered at runtime.</p>
+    <c>true</c> or <c>false</c>.</p>
 
     <p>The variables in the generator patterns shadow previously bound variables,
     including variables bound in a previous generator pattern.</p>
 
-    <p>A list comprehension returns a list, where the lists elements are the
+    <p>Variables bound in a generator expression is not visible outside the expression:</p>
```
kikofernandez, line 1993: I think we need to use the plural `are` instead of `is`, i.e., replace
```text
expression is not
```
by
```text
expression are not
```

system/doc/reference_manual/expressions.xml
```diff
@@ -1985,29 +1985,31 @@ KeyPattern := ValuePattern &lt;- MapExpression</pre>
...
+
+    <pre>
+1> <input>[{E,L} || E &lt;- L=[1,2,3]].</input>
+* 1:5: variable 'L' is unbound</pre>
+
+    <p>A <strong>list comprehension</strong> returns a list, where the lists elements are the
```
kikofernandez, line 1999: I think there is a minor misspelling here, but I could be wrong.

I think you either want to say `where the list's elements` making clear that the list is in possession of the elements, or you want to say `where the list elements`. The current `where the lists elements` may not be grammatically correct, but I am not 100% sure...

system/doc/reference_manual/expressions.xml
```diff
@@ -1985,29 +1985,31 @@ KeyPattern := ValuePattern &lt;- MapExpression</pre>
...
     each combination of bit string generator elements for which all
     filters are true.</p>
 
-    <p>A map comprehension returns a map, where the map elements are
-    the result of evaluating <c>KeyExpr</c> and <c>ValueExpr</c> for
-    each combination of generator elements for which all filters are
-    true. If the key expressions are not unique, the last occurrence
-    is stored in the map.</p>
+    <p>A <strong>map comprehension</strong> returns a map, where the
+    map elements are the result of evaluating <c>KeyExpr</c> and
```
kikofernandez, line 2009: for consistency, if you previously write `where the list's elements`, then I would say here `where the map's elements`.
Essentially, pick one writing style and keep it consistent 😄

system/doc/reference_manual/expressions.xml
```diff
@@ -2072,9 +2078,50 @@ KeyPattern := ValuePattern &lt;- MapExpression</pre>
 2> <input>[x || is_integer(x)].</input>
 []</pre>
 
-    <p>More examples are provided in
-    <seeguide marker="system/programming_examples:list_comprehensions">
-    Programming Examples.</seeguide></p>
+    <p>The behavior when a filter expression fails to evaluate to a
+    boolean value depends on whether the expression is a <seeguide
+    marker="#guard_expressions">guard expression</seeguide> or
+    not:</p>
```
kikofernandez, line 2084: The sentence is correct, but it seems like it could be improved. 

I propose the suggestion below, but feel free to either leave as is or change to something that you like better 👍🏽 :
```html
<p>Filter expressions may change the behaviour of their evaluation logic 
when their result is a non-boolean value as follows:</p>
```

system/doc/reference_manual/expressions.xml
```diff
@@ -2072,9 +2078,50 @@ KeyPattern := ValuePattern &lt;- MapExpression</pre>
...
+      to evaluating to <c>false</c>.</p></item>
+
+      <item><p>If the expression is not a guard expression and
+      evaluates to a non-Boolean value <c>Val</c>, an exception
+      <c>{bad_filter, Val}</c> is triggered at runtime. If the
+      evaluation of the expression raises an exception, it is not
+      caught by the comprehension.</p></item>
+    </list>
+
+    <p><strong>Examples</strong> (using a guard expression as filter):</p>
```
kikofernandez, line 2098: Great again with the example!

kikofernandez (changes requested):
The changes improve the documentation, IMHO.
I mark the PR not ready simply because I believe there are misspellings. If I am mistaken/when fixed, please let me know and I can approve 👍🏽

system/doc/reference_manual/expressions.xml
bjorng: Yes, the sentence felt awkward to me. Changed to:

> What happens when the filter expression does not evaluate to a boolean value depends on the expression:

PR #6932, opened 2023-02-27
Allow node_type as option on pg

josevalim:
This allows pg overlays to work on
visible nodes, hidden nodes, or both.

### DIFF
--- a/lib/kernel/doc/src/pg.xml
+++ b/lib/kernel/doc/src/pg.xml
@@ -131,6 +131,41 @@
     </func>
 
     <func>
+      <name name="start" arity="2" since="OTP 26.0"/>
+      <name name="start_link" arity="2" since="OTP 26.0"/>
+      <fsummary>Start additional scope with options.</fsummary>
+      <desc>
+        <p>Starts additional scope.</p>
+        <p><c>Options</c> is a map where the following associations
+        are allowed:</p>
+        <taglist>
+          <tag><c>node_type => <anno>NodeType</anno></c></tag>
+          <item>
+            <p>Valid values for <c>NodeType</c>:</p>
+            <taglist>
+              <tag><c>visible</c></tag>
+              <item><p>The groups in this scope are exchanged between visible
+               nodes only (the nodes returned by <c>nodes()</c>).</p></item>
+              <tag><c>hidden</c></tag>
+              <item><p>The groups in this scope are exchanged between hidden
+               nodes only.</p></item>
+              <tag><c>all</c></tag>
+              <item><p>The groups in this scope are exchanged between all
+               nodes.</p></item>
+            </taglist>
+            <p>
+              It defaults to <c>visible</c>. If the <c>node_type</c> is set
+              for a given scope, then the same scope on all nodes must have
+              a matching <c>node_type</c>. If the values are different,
+              different nodes may have different views of which processes
+              belong to the scope and the overall the behaviour is unspecified.
+            </p>
+          </item>
+        </taglist>
+      </desc>
+    </func>
+
+    <func>
       <name name="join" arity="2" since="OTP 23.0"/>
       <name name="join" arity="3" since="OTP 23.0"/>
       <fsummary>Join a process or a list of processes to a group.</fsummary>
--- a/lib/kernel/test/pg_SUITE.erl
+++ b/lib/kernel/test/pg_SUITE.erl
@@ -56,7 +56,7 @@
     missing_scope_join/1,
     disconnected_start/1,
     forced_sync/0, forced_sync/1,
-    group_leave/1,
+    group_leave/1, node_type/1,
     monitor_nonempty_scope/0, monitor_nonempty_scope/1,
     monitor_scope/0, monitor_scope/1,
     monitor/1,
@@ -93,7 +93,7 @@ groups() ->
         {performance, [], [thundering_herd]},
         {cluster, [parallel], [process_owner_check, two, initial, netsplit, trisplit, foursplit,
             exchange, nolocal, double, scope_restart, missing_scope_join, empty_group_by_remote_leave,
-            disconnected_start, forced_sync, group_leave]},
+            disconnected_start, forced_sync, group_leave, node_type]},
         {monitor, [parallel], [monitor_nonempty_scope, monitor_scope, monitor]},
         {old_release, [parallel], [process_owner_check, two, overlay_missing,
                                    empty_group_by_remote_leave, initial, netsplit,
@@ -194,7 +194,7 @@ dyn_distribution(Config) when is_list(Config) ->
     %%  distributed, and calling net_kernel:start/1, however
     %%  the effect is still the same as simply sending nodeup,
     %%  which is also documented.
-    ?FUNCTION_NAME ! {nodeup, node()},
+    ?FUNCTION_NAME ! {nodeup, node(), #{}},
     %%
     ?assertEqual(ok, pg:join(?FUNCTION_NAME, ?FUNCTION_NAME, self())),
     ?assertEqual([self()], pg:get_members(?FUNCTION_NAME, ?FUNCTION_NAME)),
@@ -387,7 +387,7 @@ netsplit(Config) when is_list(Config) ->
     1 = erlang:trace(PgPid, true, ['receive']),
     pong = net_adm:ping(Node),
     receive
-        {trace, PgPid, 'receive', {nodeup, Node}} -> ok
+        {trace, PgPid, 'receive', {nodeup, Node, _}} -> ok
     end,
     1 = erlang:trace(PgPid, false, ['receive']),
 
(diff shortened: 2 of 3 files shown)


lib/kernel/test/pg_SUITE.erl
```diff
@@ -620,13 +620,55 @@ group_leave(Config) when is_list(Config) ->
...
+node_type(Config) when is_list(Config) ->
+    NodeTypeDefault = ?FUNCTION_NAME,
+    NodeTypeVisible = node_type_visible,
+    NodeTypeHidden = node_type_hidden,
+    NodeTypeAll = node_type_all,
+    Scopes = [NodeTypeDefault, NodeTypeVisible, NodeTypeHidden, NodeTypeAll],
+
+    {ok, _} = pg:start(NodeTypeVisible, #{node_type => visible}),
+    {ok, _} = pg:start(NodeTypeHidden, #{node_type => hidden}),
+    {ok, _} = pg:start(NodeTypeAll, #{node_type => all}),
```
max-au, line 639: Could you please change `start` to `start_link` here?
The reason is, if the test case fails somewhere in the middle, unlinked processes will continue to run, and may cause side effects (e.g. running this test in the same shell twice fails with an unrelated reason).

lib/kernel/test/pg_SUITE.erl
```diff
@@ -620,13 +620,55 @@ group_leave(Config) when is_list(Config) ->
...
+    Pid = erlang:spawn(forever()),
+    [?assertEqual(ok, pg:join(Scope, key, Pid)) || Scope <- Scopes],
+
+    {Peer, Node} = spawn_disconnected_node(NodeTypeDefault, node_type_visible, Config, []),
+    {ok, _} = peer:call(Peer, pg, start, [NodeTypeVisible, #{node_type => visible}]),
+    {ok, _} = peer:call(Peer, pg, start, [NodeTypeHidden, #{node_type => hidden}]),
+    {ok, _} = peer:call(Peer, pg, start, [NodeTypeAll, #{node_type => all}]),
+    true = net_kernel:connect_node(Node),
+
+    ?assertEqual([Pid], rpc:call(Node, pg, get_members, [NodeTypeDefault, key])),
```
max-au, line 650: The test flakes here:

```
=== Location: [{pg_SUITE,node_type,650},
              {test_server,ts_tc,1782},
              {test_server,run_test_case_eval1,1291},
              {test_server,run_test_case_eval,1223}]
=== === Reason: {assertEqual,
                     [{module,pg_SUITE},
                      {line,650},
                      {expression,
                          "rpc : call ( Node , pg , get_members , [ NodeTypeDefault , key ] )"},
                      {expected,[<0.119765.0>]},
                      {value,[]}]}
  in function  pg_SUITE:node_type/1 (pg_SUITE.erl, line 650)
```

Because all operations are asynchronous, and by the time test runner reaches (and proceeds through) `rpc:call`, scopes haven't finished scope discovery.

If you look into other test cases, there are instructions like this:
```
    sync({?FUNCTION_NAME, Node}),
    sync(?FUNCTION_NAME),
```
Which ensure that a scope named `?FUNCTION_NAME` on the `Node` has processed the message queue.

Something like this should help:
`[begin sync(Proc, Node), sync(Proc) end || Proc <- [NodeTypeDefault, NodeTypeVisible, NodeTypeHidden, NodeTypeAll]],

max-au (commented):
Looks good in general.
I recall that we also discussed `this` node_type for node-local pg scope, that does not monitor remote nodes. I can add that later on.

I also ran the property-based test (this one: https://github.com/max-au/spg/blob/master/test/pg_cluster_SUITE.erl ) and found no regressions, but that test does not have knowledge of all/visible/hidden nodes (using default).

josevalim:
@max-au thank you. Feedback addressed and force pushed.

okeuday:
@josevalim If hidden node connections are becoming more common and accepted for explicit use, you will want to document the `net_kernel:hidden_connect_node/1` function because that currently remains undocumented.  `net_kernel:hidden_connect_node/1` is the way to explicitly create a hidden node connection in Erlang/OTP.

lib/kernel/doc/src/pg.xml
```diff
@@ -130,6 +130,18 @@
...
 
+    <func>
+      <name name="start" arity="2" since="OTP 26.0"/>
+      <name name="start_link" arity="2" since="OTP 26.0"/>
+      <fsummary>Start additional scope with options.</fsummary>
+      <desc>
+        <p>Starts additional scope.</p>
+        <p><c>Options</c> are a map with optional keys. Currently
+          only the <c>node_type</c> key is supported, and it must be
+          one of <c>visible</c> (default), <c>hidden</c>, or <c>all</c>.</p>
```
max-au, line 141: I think it's also necessary to add a `<datatype>` documentation that explains the difference between all/hidden/visible. Otherwise there is no way to know what this parameter means.

max-au (approved):
Apart from the documentation nit, I don't have any other comments.

lib/kernel/doc/src/pg.xml
josevalim: Pushed better docs using taglist as in net_kernel:connect_node.

rickard-green:
Hidden connections between nodes exist in order to provide means to connect from something acting as a node without the services provided by full fledged Erlang/OTP nodes. This something would typically be an `erl_interface` "node" or a `jinterface` "node". The ability for an Erlang/OTP node to pose as a hidden node was added later since we thought it would be useful, typically for inspection, to connect from an Erlang/OTP node into a cluster without interfering with those services.

When a hidden/all scope has been enabled, non full fledged Erlang/OTP nodes not prepared for this `pg` communication might crash since they are not prepared for such communication. This will typically be the case for `erl_interface` nodes.

Today there are no services trying to talk to a hidden node that connects. This since those services are not required to exist on such a node. I think this is nice and clean, so I'm not that fond of this PR. Do you have actual use cases for hidden/all scopes? Why aren't all the nodes visible?

@okeuday 
> @josevalim If hidden node connections are becoming more common and accepted for explicit use, you will want to document the `net_kernel:hidden_connect_node/1` function because that currently remains undocumented. `net_kernel:hidden_connect_node/1` is the way to explicitly create a hidden node connection in Erlang/OTP.

No `net_kernel:hidden_connect_node()` will not be documented, will eventually be removed, and is not to be used. If a node is supposed to act as a hidden node, it should be started as a hidden node.

josevalim:
@rickard-green thanks! I will try solving my problem with `-connect_all false` instead and, if it works, I will close this.

okeuday:
@rickard-green The problem is that hidden nodes are useful for avoiding a fully connected network topology.  I enjoy being able to make some node connections hidden while others are visible for [cpg](https://github.com/okeuday/cpg/), so it is disappointing that Ericsson's Erlang/OTP team is unable to support `net_kernel:hidden_connect_node()` use in the future.

josevalim:
FWIW, I think `hidden_connect_node` could work on my use cases, although I would not rely on private APIs. Basically I have two clusters, which are fully meshed inside themselves, but may have a hidden connection when connecting to a node in the other cluster.

@rickard-green what are the concerns and difficulties with making `hidden` a property of the connection instead of the nodes? Where the node property becomes the default for the connection?

rickard-green:
@okeuday wrote:
> I enjoy being able to make some node connections hidden while others are visible

This is one reason for this not being supported. When you setup connections like that, services cannot rely on that all nodes in a cluster will get notified in the same manner. For example, `global`'s algorithm will break down when a connection to a specific node triggers `nodeup` for some nodes in the cluster, but not for other nodes in the cluster.

@josevalim wrote:
> 1. Is it expected that if node A connects to node B, where node B has `-connect_all false`, the global names in B won't be visible to A? Or is this an accident? And is the fact `global:sync()` never returns expected?

I'm quite sure you are expected to use the same configuration for `connect_all` on all visible nodes for `global` to work as expected. The documentation could however be a bit clearer on this. When a node with `connect_all=true` connects to a node with `connect_all=false`, `global`'s connection setup will hang forever which in turn will cause `global:sync()` to hang. This has always been the case as far as I know.

> 2. Unfortunately I cannot use global groups because my understanding is that all nodes need to be declared upfront. It would be nice if the groups could be declared based on the name itself. For example, if I say I belong to the global group `foobar`, every node that declares such group is automatically part of the group. We already ping the connecting node to see if our interpretations of groups are the same, perhaps we could support `{GroupName, dynamic}` and if they are both are dynamic, it means they both agree?

Yes it is unfortunate that everything has to be statically declared beforehand when it comes to `global_group`. I agree that it would be nice with a dynamic version.

> FWIW, I think hidden_connect_node could work on my use cases, although I would not rely on private APIs. Basically I have two clusters, which are fully meshed inside themselves, but may have a hidden connection when connecting to a node in the other cluster.
>
> @rickard-green what are the concerns and difficulties with making hidden a property of the connection instead of the nodes? Where the node property becomes the default for the connection?

One problem is what I wrote in response to @okeuday (above in this comment). Another problem is that if you have auto connect enabled, the connection might become visible if the connection is lost and then automatically set up again. In your scenario above, this will cause the two clusters to join together into one big cluster.

I wrote:
> When a hidden/all scope has been enabled, non full fledged Erlang/OTP nodes not prepared for this pg communication might crash since they are not prepared for such communication. This will typically be the case for `erl_interface` nodes.
> 
> Today there are no services trying to talk to a hidden node that connects. This since those services are not required to exist on such a node. I think this is nice and clean, so I'm not that fond of this PR. Do you have actual use cases for hidden/all scopes? Why aren't all the nodes visible?

Note that this does not mean that we have rejected this PR, but it would be nice to know if this can be solved some other way. If not, at a minimum there need to be a warning in the documentation about the issues hidden/all scopes can cause.

rickard-green:
@josevalim Sorry for the long silence!

We've discussed this and have come to the conclusion that we don't want services to communicate with hidden nodes like this.

We instead would like to solve situations like this by an improved `global_group` that supports dynamically adding nodes to groups. I cannot say when we will be able to have a look at this, though. A PR is of course welcome.

josevalim:
@rickard-green thanks! I will try a PR but definitely do not wait for me. Last time I checked it was way above my pay grade. :)

josevalim:
Closing this one as it is not accepted right now.

PR #6938, opened 2023-02-27
ssl: use macros instead of hardcoded version numbers

kikofernandez:
Replaces TLS and SSL version tuple codes (e.g., `{3,4}` for means TLS-1.3) by a macro descriptive name, and some refactorings as follows:
- addition of macros `?'TLS-1.3`, `?'TLS-1.2`, `?'TLS-1.1`, `?'TLS-1.0`, `?'TLS-1.0`, `?'TLS-1.X`,
- addition of macros `?'SSL-3.0`, `?'SSL-2.0`,
- addition of macros  `?'DTLS-1.2`, `?'DTLS-1.1`, `?'DTLS-1.0`
- refactor of functions that receive arguments which are not used

The addition of the macros is self-contained in its own commit. The refactorings are placed in different commits for ease of scanning and reversal (if necessary).

### DIFF
--- a/lib/ssl/src/dtls_gen_connection.erl
+++ b/lib/ssl/src/dtls_gen_connection.erl
@@ -545,6 +545,8 @@ handle_info({CloseTag, Socket}, StateName,
     %% with widespread implementation practice.
     case (Active == false) andalso (CTs =/= []) of
         false ->
+            %% the =< is a leaking implementation detail.
+            %% this means `Version={254, N} when N =< 253`
             if (Version =< ?'DTLS-1.2') ->
                     ok;
                true ->
--- a/lib/ssl/src/dtls_gen_connection.erl
+++ b/lib/ssl/src/dtls_gen_connection.erl
@@ -545,10 +545,9 @@ handle_info({CloseTag, Socket}, StateName,
     %% with widespread implementation practice.
     case (Active == false) andalso (CTs =/= []) of
         false ->
-            case Version of
-                {254, N} when N =< 253 ->
+            if (Version =< ?'DTLS-1.2') ->
                     ok;
-                _ ->
+               true ->
                     %% As invalidate_sessions here causes performance issues,
                     %% we will conform to the widespread implementation
                     %% practice and go against the spec
--- a/lib/ssl/src/ssl_cipher.erl
+++ b/lib/ssl/src/ssl_cipher.erl
@@ -328,7 +328,7 @@ suites(?'DTLS-1.X'=Version) ->
 all_suites(?'TLS-1.3' = Version) ->
     suites(Version) ++ tls_testing_suites(?'TLS-1.2');
 all_suites(?'TLS-1.X' = Version) ->
-    suites(Version) ++ tls_testing_suites(?'TLS-1.2');
+    suites(Version) ++ tls_testing_suites(Version);
 all_suites(Version) ->
     dtls_v1:all_suites(Version).
 
--- a/lib/ssl/src/ssl_cipher.erl
+++ b/lib/ssl/src/ssl_cipher.erl
@@ -214,8 +214,7 @@ build_cipher_block(BlockSz, Mac, Fragment) ->
     [Fragment, Mac, padding_with_len(TotSz, BlockSz)].
 
 block_cipher(Fun, BlockSz, #cipher_state{key=Key, iv=IV} = CS0,
-	     Mac, Fragment, Version)
-  when Version == ?'SSL-3.0'; Version == ?'TLS-1.0' ->
+	     Mac, Fragment, ?'TLS-1.0') ->
     L = build_cipher_block(BlockSz, Mac, Fragment),
     T = Fun(Key, IV, L),
     NextIV = next_iv(T, IV),
@@ -326,13 +325,13 @@ suites(?'TLS-1.X'=Version) ->
 suites(?'DTLS-1.X'=Version) ->
     dtls_v1:suites(Version).
 all_suites(?'TLS-1.3' = Version) ->
-    suites(Version) ++ tls_testing_suites(?'TLS-1.2');
+    suites(Version) ++ tls_legacy_suites(?'TLS-1.2');
 all_suites(?'TLS-1.X' = Version) ->
-    suites(Version) ++ tls_testing_suites(Version);
+    suites(Version) ++ tls_legacy_suites(Version);
 all_suites(Version) ->
     dtls_v1:all_suites(Version).
 
-tls_testing_suites(Version) ->
+tls_legacy_suites(Version) ->
     Tests = [ fun tls_v1:psk_suites/1
             , fun tls_v1:srp_suites/1
             , fun tls_v1:rsa_suites/1
@@ -685,7 +684,7 @@ mac_hash({_,_}, ?NULL, _MacSecret, _SeqNo, _Type,
 	 _Length, _Fragment) ->
     <<>>;
 mac_hash(?'TLS-1.X' = Version, MacAlg, MacSecret, SeqNo, Type, Length, Fragment)
-  when Version =/= ?'SSL-3.0' ->
+  when Version =< ?'TLS-1.2', Version =/= ?'SSL-3.0' ->
     tls_v1:mac_hash(MacAlg, MacSecret, SeqNo, Type, Version,
 		      Length, Fragment).
 
@@ -825,7 +824,7 @@ block_size(Cipher) when Cipher == aes_128_cbc;
 			Cipher == chacha20_poly1305 ->
     16.
 
-prf_algorithm(default_prf, ?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
+prf_algorithm(default_prf, ?'TLS-1.2') ->
     ?SHA256;
 prf_algorithm(default_prf, ?'TLS-1.X') ->
     ?MD5SHA;
@@ -930,8 +929,7 @@ signature_algorithm_to_scheme(#'SignatureAlgorithm'{algorithm = ?'id-RSASSA-PSS'
 %%   We return the original (possibly invalid) PadLength in any case.
 %%   An invalid PadLength will be caught by is_correct_padding/2
 %%
-generic_block_cipher_from_bin(Version, T, IV, HashSize)
-  when Version == ?'SSL-3.0'; Version == ?'TLS-1.0' ->
+generic_block_cipher_from_bin(?'TLS-1.0', T, IV, HashSize)->
     Sz1 = byte_size(T) - 1,
     <<_:Sz1/binary, ?BYTE(PadLength0)>> = T,
     PadLength = if
@@ -946,7 +944,7 @@ generic_block_cipher_from_bin(Version, T, IV, HashSize)
 			  next_iv = IV};
 
 generic_block_cipher_from_bin(Version, T, IV, HashSize)
-  when Version == ?'TLS-1.1'; Version == ?'TLS-1.2'; Version == ?'TLS-1.3' ->
+  when Version == ?'TLS-1.1'; Version == ?'TLS-1.2' ->
     Sz1 = byte_size(T) - 1,
     <<_:Sz1/binary, ?BYTE(PadLength)>> = T,
     IVLength = byte_size(IV),
@@ -968,7 +966,7 @@ is_correct_padding(#generic_block_cipher{padding_length = Len,
 					 padding = Padding}, ?'SSL-3.0', _) ->
     Len == byte_size(Padding); %% Only length check is done in SSL 3.0 spec
 %% For interoperability reasons it is possible to disable
-%% the padding check when using TLS 1.0, as it is not strictly required 
+%% the padding check when using TLS 1.0 (mimicking SSL-3.0), as it is not strictly required
 %% in the spec (only recommended), however this makes TLS 1.0 vunrable to the Poodle attack 
 %% so by default this clause will not match
 is_correct_padding(GenBlockCipher, ?'TLS-1.0', false) ->
@@ -1052,12 +1050,12 @@ filter_suites_pubkey(ecdsa, Ciphers, _, OtpCert) ->
 
 filter_suites_signature(_, Ciphers, ?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
      Ciphers;
-filter_suites_signature(rsa, Ciphers, Version) ->
-    (Ciphers -- ecdsa_signed_suites(Ciphers, Version)) -- dsa_signed_suites(Ciphers, Version);
-filter_suites_signature(dsa, Ciphers, Version) ->
-    (Ciphers -- ecdsa_signed_suites(Ciphers, Version)) -- rsa_signed_suites(Ciphers, Version);
-filter_suites_signature(ecdsa, Ciphers, Version) ->
-    (Ciphers -- rsa_signed_suites(Ciphers, Version)) -- dsa_signed_suites(Ciphers, Version).
+filter_suites_signature(rsa, Ciphers, _Version) ->
+    (Ciphers -- ecdsa_signed_suites(Ciphers)) -- dsa_signed_suites(Ciphers);
+filter_suites_signature(dsa, Ciphers, _Version) ->
+    (Ciphers -- ecdsa_signed_suites(Ciphers)) -- rsa_signed_suites(Ciphers);
+filter_suites_signature(ecdsa, Ciphers, _Version) ->
+    (Ciphers -- rsa_signed_suites(Ciphers)) -- dsa_signed_suites(Ciphers).
 
 
 %% From RFC 5246 - Section  7.4.2.  Server Certificate
@@ -1076,15 +1074,7 @@ filter_suites_signature(ecdsa, Ciphers, Version) ->
 %% extension.  The names DH_DSS, DH_RSA, ECDH_ECDSA, and ECDH_RSA are
 %% historical.
 %% Note: DH_DSS and DH_RSA is not supported
-rsa_signed(?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
-    fun(rsa) -> true;
-       (dhe_rsa) -> true;
-       (ecdhe_rsa) -> true;
-       (rsa_psk) -> true;
-       (srp_rsa) -> true;
-       (_) -> false
-    end;
-rsa_signed(_) ->
+rsa_signed() ->
     fun(rsa) -> true;
        (dhe_rsa) -> true;
        (ecdhe_rsa) -> true;
@@ -1094,24 +1084,20 @@ rsa_signed(_) ->
        (_) -> false
     end.
 %% Cert should be signed by RSA
-rsa_signed_suites(Ciphers, Version) ->
-    filter_suites(Ciphers, #{key_exchange_filters => [rsa_signed(Version)],
+rsa_signed_suites(Ciphers) ->
+    filter_suites(Ciphers, #{key_exchange_filters => [rsa_signed()],
                              cipher_filters => [],
                              mac_filters => [],
                              prf_filters => []}).
-ecdsa_signed(?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
-    fun(ecdhe_ecdsa) -> true;
-       (_) -> false
-    end;
-ecdsa_signed(_) ->
+ecdsa_signed() ->
     fun(ecdhe_ecdsa) -> true;
        (ecdh_ecdsa) -> true;
        (_) -> false
     end. 
 
 %% Cert should be signed by ECDSA
-ecdsa_signed_suites(Ciphers, Version) ->
-    filter_suites(Ciphers, #{key_exchange_filters => [ecdsa_signed(Version)],
+ecdsa_signed_suites(Ciphers) ->
+    filter_suites(Ciphers, #{key_exchange_filters => [ecdsa_signed()],
                              cipher_filters => [],
                              mac_filters => [],
                              prf_filters => []}).
@@ -1161,12 +1147,12 @@ dss_keyed_suites(Ciphers) ->
                              prf_filters => []}).
 
 %% Cert should be signed by DSS (DSA)
-dsa_signed_suites(Ciphers, Version) ->
-    filter_suites(Ciphers, #{key_exchange_filters => [dsa_signed(Version)],
+dsa_signed_suites(Ciphers) ->
+    filter_suites(Ciphers, #{key_exchange_filters => [dsa_signed()],
                              cipher_filters => [],
                              mac_filters => [],
                              prf_filters => []}).
-dsa_signed(_) ->
+dsa_signed() ->
     fun(dhe_dss) -> true;
        (_) -> false
     end.
--- a/lib/ssl/src/ssl_cipher.erl
+++ b/lib/ssl/src/ssl_cipher.erl
@@ -214,16 +214,16 @@ build_cipher_block(BlockSz, Mac, Fragment) ->
     [Fragment, Mac, padding_with_len(TotSz, BlockSz)].
 
 block_cipher(Fun, BlockSz, #cipher_state{key=Key, iv=IV} = CS0,
-	     Mac, Fragment, {3, N})
-  when N == 0; N == 1 ->
+	     Mac, Fragment, Version)
+  when Version == ?'SSL-3.0'; Version == ?'TLS-1.0' ->
     L = build_cipher_block(BlockSz, Mac, Fragment),
     T = Fun(Key, IV, L),
     NextIV = next_iv(T, IV),
     {T, CS0#cipher_state{iv=NextIV}};
 
 block_cipher(Fun, BlockSz, #cipher_state{key=Key, iv=IV, state = IV_Cache0} = CS0,
-	     Mac, Fragment, {3, N})
-  when N == 2; N == 3; N == 4 ->
+	     Mac, Fragment, ?'TLS-1.X'=Version)
+  when Version == ?'TLS-1.1'; Version == ?'TLS-1.2'; Version == ?'TLS-1.3' ->
     IV_Size = byte_size(IV),
     <<NextIV:IV_Size/binary, IV_Cache/binary>> =
         case IV_Cache0 of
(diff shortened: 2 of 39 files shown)


lib/ssl/src/ssl_cipher.erl
```diff
@@ -214,16 +214,16 @@ build_cipher_block(BlockSz, Mac, Fragment) ->
     [Fragment, Mac, padding_with_len(TotSz, BlockSz)].
 
 block_cipher(Fun, BlockSz, #cipher_state{key=Key, iv=IV} = CS0,
-	     Mac, Fragment, {3, N})
-  when N == 0; N == 1 ->
+	     Mac, Fragment, Version)
+  when Version == ?'SSL-3.0'; Version == ?'TLS-1.0' ->
```
IngelaAndin, line 218: We have dropped support for SSL-3.0, we only keep it around to allow legacy hello messages that will negotiate higher versions so it should no longer be needed here.

lib/ssl/src/ssl_cipher.erl
```diff
@@ -828,9 +825,9 @@ block_size(Cipher) when Cipher == aes_128_cbc;
 			Cipher == chacha20_poly1305 ->
     16.
 
-prf_algorithm(default_prf, {3, N}) when N >= 3 ->
+prf_algorithm(default_prf, ?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
```
IngelaAndin, line 828: This check can be changed to match TLS-1-2

lib/ssl/src/ssl_cipher.erl
```diff
@@ -933,8 +930,8 @@ signature_algorithm_to_scheme(#'SignatureAlgorithm'{algorithm = ?'id-RSASSA-PSS'
 %%   We return the original (possibly invalid) PadLength in any case.
 %%   An invalid PadLength will be caught by is_correct_padding/2
 %%
-generic_block_cipher_from_bin({3, N}, T, IV, HashSize)
-  when N == 0; N == 1 ->
+generic_block_cipher_from_bin(Version, T, IV, HashSize)
+  when Version == ?'SSL-3.0'; Version == ?'TLS-1.0' ->
```
IngelaAndin, line 934: Here we should also be able to remove SSL-3.0

lib/ssl/src/ssl_cipher.erl
```diff
@@ -948,8 +945,8 @@ generic_block_cipher_from_bin({3, N}, T, IV, HashSize)
 			  padding=Padding, padding_length=PadLength0,
 			  next_iv = IV};
 
-generic_block_cipher_from_bin({3, N}, T, IV, HashSize)
-  when N == 2; N == 3; N == 4 ->
+generic_block_cipher_from_bin(Version, T, IV, HashSize)
+  when Version == ?'TLS-1.1'; Version == ?'TLS-1.2'; Version == ?'TLS-1.3' ->
```
IngelaAndin, line 949: Version can not be TLS-1.3!

lib/ssl/src/ssl_cipher.erl
```diff
@@ -1079,7 +1076,7 @@ filter_suites_signature(ecdsa, Ciphers, Version) ->
 %% extension.  The names DH_DSS, DH_RSA, ECDH_ECDSA, and ECDH_RSA are
 %% historical.
 %% Note: DH_DSS and DH_RSA is not supported
-rsa_signed({3,N}) when N >= 3 ->
+rsa_signed(?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
```
IngelaAndin, line 1079: This looks wrong! The functions will only be called when version is lower then TLS-1.2!

lib/ssl/src/ssl_cipher.erl
```diff
@@ -1102,7 +1099,7 @@ rsa_signed_suites(Ciphers, Version) ->
                              cipher_filters => [],
                              mac_filters => [],
                              prf_filters => []}).
-ecdsa_signed({3,N}) when N >= 3 ->
+ecdsa_signed(?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
```
IngelaAndin, line 1102: Same comment as above!

lib/ssl/src/ssl_handshake.erl
```diff
@@ -596,7 +599,7 @@ encode_handshake(#certificate_verify{signature = BinSig, hashsign_algorithm = Ha
 encode_handshake(#finished{verify_data = VerifyData}, _Version) ->
     {?FINISHED, VerifyData}.
 
-encode_hello_extensions(_, {3, 0}) ->
+encode_hello_extensions(_, ?'SSL-3.0') ->
```
IngelaAndin, line 602: This clause should no longer be needed! I would remove only this clause and then run the test cases.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -894,9 +897,9 @@ decode_handshake(_Version, ?CERTIFICATE_REQUEST,
 			 certificate_authorities = decode_cert_auths(EncCertAuths, [])};
 decode_handshake(_Version, ?SERVER_HELLO_DONE, <<>>) ->
     #server_hello_done{};
-decode_handshake({Major, Minor}, ?CERTIFICATE_VERIFY,<<HashSign:2/binary, ?UINT16(SignLen),
-						       Signature:SignLen/binary>>)
-  when Major == 3, Minor >= 3 ->
+decode_handshake(?'TLS-1.X'=Version, ?CERTIFICATE_VERIFY,<<HashSign:2/binary, ?UINT16(SignLen),
+                                                           Signature:SignLen/binary>>)
+  when Version >= ?'TLS-1.2' ->
```
IngelaAndin, line 902: Will not be TLS-1.3 so change TLS-1.2

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1121,17 +1124,21 @@ server_select_cert_key_pair_and_params(CipherSuites, [#{private_key := Key, cert
...
     {SignAlgo0, Param, _, _, _} = get_cert_params(Cert),
     SignAlgo = sign_algo(SignAlgo0, Param),
     is_acceptable_hash_sign(SignAlgo, HashSigns);
 is_acceptable_cert(_,_,_) ->
     %% Not negotiable pre TLS-1.2. So if cert is available for version it is acceptable
     true.
 
-supported_ecc({Major, Minor}) when ((Major == 3) and (Minor >= 1)) orelse (Major > 3) ->
-    Curves = tls_v1:ecc_curves(Minor),
+%% TODO: this function is not used. Can it be removed or does it need
```
IngelaAndin, line 1135: I think we can and should remove it. I think it the use of it probably got refactored out and was left here by accident!

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1973,7 +1979,7 @@ int_to_bin(I) ->
 
 %% The end-entity certificate provided by the client MUST contain a
 %% key that is compatible with certificate_types.
-certificate_types(Version) when Version =< {3,3} ->
+certificate_types(Version) when Version =< ?'TLS-1.2' ->
```
IngelaAndin, line 1982: Can not be TLS-1.3

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2336,11 +2344,11 @@ master_secret(Version, MasterSecret,
...
      ssl_record:set_pending_cipher_state(ConnStates2, ClientCipherState,
 					 ServerCipherState, Role)}.
-setup_keys({3,N}, PrfAlgo, MasterSecret,
+setup_keys({3,_}=Version, PrfAlgo, MasterSecret,
 	   ServerRandom, ClientRandom, HashSize, KML, _EKML, IVS) ->
-    tls_v1:setup_keys(N, PrfAlgo, MasterSecret, ServerRandom, ClientRandom, HashSize,
+    tls_v1:setup_keys(Version, PrfAlgo, MasterSecret, ServerRandom, ClientRandom, HashSize,
 			KML, IVS).
-calc_master_secret({3,_}, PrfAlgo, PremasterSecret, ClientRandom, ServerRandom) ->
+calc_master_secret(?'TLS-1.X', PrfAlgo, PremasterSecret, ClientRandom, ServerRandom) ->
```
IngelaAndin, line 2351: Can not be TLS-1.3

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2699,7 +2707,7 @@ dec_server_key(<<?UINT16(NLen), N:NLen/binary,
 dec_server_key(_, KeyExchange, _) ->
     throw(?ALERT_REC(?FATAL, ?HANDSHAKE_FAILURE, {unknown_or_malformed_key_exchange, KeyExchange})).
 
-dec_client_key(PKEPMS, ?KEY_EXCHANGE_RSA, {3, 0}) ->
+dec_client_key(PKEPMS, ?KEY_EXCHANGE_RSA, ?'SSL-3.0') ->
```
IngelaAndin, line 2710: Also a SSL-3.0 clause that we should be able to remove!

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2725,7 +2733,7 @@ dec_client_key(<<?UINT16(Len), Id:Len/binary,
 	       ?KEY_EXCHANGE_EC_DIFFIE_HELLMAN_PSK, _) ->
     #client_ecdhe_psk_identity{identity = Id, dh_public = DH_Y};
 dec_client_key(<<?UINT16(Len), Id:Len/binary, PKEPMS/binary>>,
-	       ?KEY_EXCHANGE_RSA_PSK, {3, 0}) ->
+	       ?KEY_EXCHANGE_RSA_PSK, ?'SSL-3.0') ->
```
IngelaAndin, line 2736: And here too!

lib/ssl/src/ssl_logger.erl
```diff
@@ -290,19 +290,20 @@ get_server_version(Version, Extensions) ->
...
+version(?'TLS-1.2') ->
     "TLS 1.2";
-version({3,2}) ->
+version(?'TLS-1.1') ->
     "TLS 1.1";
-version({3,1}) ->
+version(?'TLS-1.0') ->
     "TLS 1.0";
-version({3,0}) ->
+version(?'SSL-3.0') ->
```
IngelaAndin, line 302: We should keep this one as we still allow this version for legacy client hellos!

lib/ssl/src/ssl_record.hrl
```diff
@@ -180,4 +180,17 @@
...
 
+
+-define('TLS-1.X', {3, _}).
+-define('TLS-1.3', {3, 4}).
+-define('TLS-1.2', {3, 3}).
+-define('TLS-1.1', {3, 2}).
+-define('TLS-1.0', {3, 1}).
+-define('DTLS-1.X', {254, _}).
+-define('DTLS-1.0', {254, 255}).
+-define('DTLS-1.1', {254, 254}).
```
IngelaAndin, line 191: This version does not exist! There is only DTLS-1.0 corresponds to TLS-1.1 and DTLS-1.2 corresponds to TLS-1.2

lib/ssl/src/tls_record.erl
```diff
@@ -718,15 +718,15 @@ encode_fragments(Type, Version, [Text|Data],
 %% 1/n-1 splitting countermeasure Rizzo/Duong-Beast, RC4 ciphers are
 %% not vulnerable to this attack.
 split_iovec(Data, Version, BCA, one_n_minus_one, MaxLength)
-  when (BCA =/= ?RC4) andalso ({3, 1} == Version orelse
-                               {3, 0} == Version) ->
+  when (BCA =/= ?RC4) andalso (?'TLS-1.0' == Version orelse
```
IngelaAndin, line 721: We can remove check for SSL-3.0

lib/ssl/src/tls_record.erl
```diff
@@ -718,15 +718,15 @@ encode_fragments(Type, Version, [Text|Data],
...
+                               ?'SSL-3.0' == Version) ->
     {Part, RestData} = split_iovec(Data, 1, []),
     [Part|split_iovec(RestData, MaxLength)];
 %% 0/n splitting countermeasure for clients that are incompatible with 1/n-1
 %% splitting.
 split_iovec(Data, Version, BCA, zero_n, MaxLength)
-  when (BCA =/= ?RC4) andalso ({3, 1} == Version orelse
-                               {3, 0} == Version) ->
+  when (BCA =/= ?RC4) andalso (?'TLS-1.0' == Version orelse
```
IngelaAndin, line 728: We can remove check for SSL-3.0

lib/ssl/src/tls_gen_connection.erl
```diff
@@ -822,15 +822,15 @@ next_record_done(#state{protocol_buffers = Buffers} = State, CipherTexts, Connec
 %% field in TLS-1.3 client hello). The versions are instead negotiated with an hello extension. When
 %% decoding the server_hello messages we want to go through TLS-1.3 decode functions to be able
 %% to handle TLS-1.3 extensions if TLS-1.3 will be the negotiated version.
-handle_unnegotiated_version({3,3} , #{versions := [{3,4} = Version |_]} = Options, Data, Buffer, client, hello) ->
+handle_unnegotiated_version(?'TLS-1.2' , #{versions := [?'TLS-1.3' = Version |_]} = Options, Data, Buffer, client, hello) ->
```
IngelaAndin, line 825: Maybe we should use the legacy macro here for TLS-1.2?!

IngelaAndin: @dgud @u3s

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -79,7 +79,7 @@ server_hello(MsgType, SessionId, KeyShare, PSK, ConnectionStates) ->
     #{security_parameters := SecParams} =
 	ssl_record:pending_connection_state(ConnectionStates, read),
     Extensions = server_hello_extensions(MsgType, KeyShare, PSK),
-    #server_hello{server_version = {3,3}, %% legacy_version
+    #server_hello{server_version = ?'TLS-1.2', %% legacy_version
```
IngelaAndin, line 82: Here we probably should use the legacy-macro!

lib/ssl/src/tls_record_1_3.erl
```diff
@@ -288,7 +288,7 @@ encode_plain_text(#inner_plaintext{
     Encoded = cipher_aead(PlainText, BulkCipherAlgo, Key, Seq, IV, TagLen),
     %% 23 (application_data) for outward compatibility
     #tls_cipher_text{opaque_type = ?OPAQUE_TYPE,
-                     legacy_version = {3,3},
+                     legacy_version = ?'TLS-1.2',
```
IngelaAndin, line 291: Legacy macro!

lib/ssl/src/tls_sender.erl
```diff
@@ -582,7 +582,7 @@ update_bytes_sent(_, #data{static = #static{bytes_sent = Sent} = Static} = State
 key_update_at(Version, #{security_parameters :=
                              #security_parameters{
                                 bulk_cipher_algorithm = CipherAlgo}}, KeyUpdateAt)
-  when Version >= {3,4} ->
+  when Version >= ?'TLS-1.3' ->
```
IngelaAndin, line 585: Maybe we should test for TLS-1.3 here, we do not know when and if TLS-1.4 will be equivalent! Same comment for the lines below.

lib/ssl/src/tls_v1.erl
```diff
@@ -170,10 +169,15 @@ master_secret(PrfAlgo, PreMasterSecret, ClientRandom, ServerRandom) ->
 	[ClientRandom, ServerRandom], 48).
 %% TLS 1.0 -1.2  ---------------------------------------------------
 
--spec finished(client | server, integer(), integer(), binary(), [binary()]) -> binary().
+-spec finished(Role, Version, PrfAlgo, MasterSecret, Handshake) -> binary() when
+      Role :: client | server,
+      Version :: {non_neg_integer(), non_neg_integer()},
```
IngelaAndin, line 174: we should use  ssl:tls_version() type

lib/ssl/src/tls_v1.erl
```diff
@@ -203,30 +207,29 @@ finished(Role, Version, PrfAlgo, MasterSecret, Handshake)
 
 %% TODO 1.3 finished
 
--spec certificate_verify(md5sha | sha, integer(), [binary()]) -> binary().
-
+-spec certificate_verify(HashAlgo, Handshake) -> binary() when
+      HashAlgo :: md5sha | sha,
```
IngelaAndin, line 211: should have type ssl:hash() even though ignored by the second clause for optimization reasons.

lib/ssl/src/tls_v1.erl
```diff
@@ -279,9 +281,9 @@ setup_keys(Version, _PrfAlgo, MasterSecret, ServerRandom, ClientRandom, HashSize
 %% TLS v1.1 ---------------------------------------------------
 
 %% TLS v1.2  ---------------------------------------------------
-setup_keys(Version, PrfAlgo, MasterSecret, ServerRandom, ClientRandom, HashSize,
+setup_keys(?'TLS-1.X'=Version, PrfAlgo, MasterSecret, ServerRandom, ClientRandom, HashSize,
 	   KeyMatLen, IVSize)
-  when Version == 3; Version == 4 ->
+  when Version == ?'TLS-1.2'; Version == ?'TLS-1.3' ->
```
IngelaAndin, line 286: Can not be TLS-1.3

IngelaAndin: Also remove comment on line 510!

lib/ssl/test/ssl_api_SUITE.erl
```diff
@@ -3224,55 +3228,63 @@ log(#{msg:={report,_Report}},#{config:=Pid}) ->
...
+exclusive_default_up_to_version(?'TLS-1.3', Acc) ->
+    exclusive_default_up_to_version(?'TLS-1.2', [?'TLS-1.3' | Acc]).
+
+dtls_exclusive_default_up_to_version(?'DTLS-1.0', Acc) ->
+    lists:flatmap( fun (Vsn) -> ssl:cipher_suites(exclusive, Vsn) end
+                 , [?'DTLS-1.0' | Acc]);
+dtls_exclusive_default_up_to_version(?'DTLS-1.2', Acc) ->
+    dtls_exclusive_default_up_to_version(?'DTLS-1.0', [?'DTLS-1.2' | Acc]).
+
+%% TODO: wip
```
IngelaAndin, line 3257: ?

lib/ssl/test/ssl_cert_SUITE.erl
```diff
@@ -1355,7 +1356,7 @@ rsa_alg(rsa_pss_pss_1_3) ->
 rsa_alg(Atom) ->
     Atom.
 
-no_reuse({3, N}) when N >= 4 ->
+no_reuse(?'TLS-1.X'=Version) when Version >= ?'TLS-1.3' ->
```
IngelaAndin, line 1359: Not be possible future compatible, that is use when equals!

lib/ssl/test/ssl_cert_tests.erl
```diff
@@ -462,8 +462,8 @@ test_ciphers(_, 'tlsv1.3' = Version) ->
                  end, Ciphers);
 test_ciphers(_, Version) when Version == 'dtlsv1';
                                 Version == 'dtlsv1.2' ->
-    {_, Minor} = dtls_record:protocol_version(Version),
-    Ciphers = [ssl_cipher_format:suite_bin_to_map(Bin) ||  Bin <- dtls_v1:suites(Minor)],
+    Versioning = dtls_record:protocol_version(Version),
```
IngelaAndin, line 465: Not very important but called NVersion in other places

lib/ssl/test/ssl_cipher_SUITE.erl
```diff
@@ -108,19 +109,19 @@ aes_decipher_fail() ->
 aes_decipher_fail(Config) when is_list(Config) ->
     HashSz = 32,
     CipherState = incorrect_cipher_state(),
-    decipher_check_fail(HashSz, CipherState, {3,0}),
-    decipher_check_fail(HashSz, CipherState, {3,1}),
-    decipher_check_fail(HashSz, CipherState, {3,2}),
-    decipher_check_fail(HashSz, CipherState, {3,3}).
+    decipher_check_fail(HashSz, CipherState, ?'SSL-3.0'),
```
IngelaAndin, line 112: SSL-3.0 can be removed

lib/ssl/test/ssl_cipher_SUITE.erl
```diff
@@ -108,19 +109,19 @@ aes_decipher_fail() ->
...
 
 %%--------------------------------------------------------------------
 padding_test(Config) when is_list(Config)  ->
     HashSz = 16,
     CipherState = correct_cipher_state(),
-    pad_test(HashSz, CipherState, {3,0}),
-    pad_test(HashSz, CipherState, {3,1}),
-    pad_test(HashSz, CipherState, {3,2}),
-    pad_test(HashSz, CipherState, {3,3}).
+    pad_test(HashSz, CipherState, ?'SSL-3.0'),
```
IngelaAndin, line 121: SSL-3.0 can be removed

lib/ssl/test/ssl_handshake_SUITE.erl
```diff
@@ -128,7 +128,7 @@ decode_hello_handshake(_Config) ->
 		    16#00, 16#00, 16#33, 16#74, 16#00, 16#07, 16#06, 16#73,
 		    16#70, 16#64, 16#79, 16#2f, 16#32>>,
 	
-    Version = {3, 0},
+    Version = ?'SSL-3.0',
```
IngelaAndin, line 131: This is also testing SSL-3.0, I think we should aim to remove this suite all together. Thought so for a while, just have not have time to check that all that it test is actually covered by the property tests. It should be, but we  should check!

lib/ssl/test/ssl_npn_hello_SUITE.erl
```diff
@@ -123,12 +123,12 @@ encode_and_decode_npn_server_hello_test(Config) ->
 
 %%--------------------------------------------------------------------
 create_server_hello_with_no_advertised_protocols_test(_Config) ->
-    Hello = ssl_handshake:server_hello(<<>>, {3, 0}, create_connection_states(), #{}),
+    Hello = ssl_handshake:server_hello(<<>>, ?'SSL-3.0', create_connection_states(), #{}),
```
IngelaAndin, line 126: This suite I also think is a candidate for removal!

IngelaAndin (changes requested):
Most of the changes I requested are due enhancements  I realized can be done when looking at the code due to the improvements you did.

lib/ssl/src/ssl_record.hrl
KennethL, line 191: When I was thinking about this several years ago (when we optimized the distribution over SSL) I was thinking if it could be a good idea to represent the version as an integer. And when we now are declaring the macros there is actually not much point in having the version as a tuple.
```suggestion
-define('is_TLS-1.X'(V), (V bsr 8) == 3). % to be used in guard or case or if statement
-define('TLS-1.3', 16#304).
-define('TLS-1.2', 16#303).
-define('TLS-1.1', 16#302).
-define('TLS-1.0', 16#301).
-define('is_DTLS-1.X'(V), (V bsr 8) == 254. % to be used in guard or case or if statement
-define('DTLS-1.0', (254 bsl 8) + 255.
-define('DTLS-1.1', (254 bsl 8) + 254).
```
There will be a need for minor changes at a few other places as well, for example when matching with TLS-1.X and when encoding /decoding the SSL_record from and to bytes.

KennethL: I don't know if you like this change, but I do.

IngelaAndin: I can see both benefits and drawbacks. I think we should not make this change for this PR but we can evaluate if we think it is beneficial enough later,

KennethL: What are the drawbacks?

lib/ssl/src/ssl_cipher.erl
kikofernandez: ok, I am not sure I fully understand what is implied :)
Do you mean that I should `Version =< ?'TLS-1.2'? If I were to do this, then the next function clause for `rsa_signed(_) ->` would never be executed, correct?

IngelaAndin: I am saying that the firs clause is dead code and we could remove the Version argument to the function.

lib/ssl/src/ssl_cipher.erl
```diff
@@ -1053,7 +1050,7 @@ filter_suites_pubkey(ecdsa, Ciphers, _, OtpCert) ->
                                    ec_ecdhe_suites(Ciphers)),
     filter_keyuse_suites(keyAgreement, Uses, CiphersSuites, ec_ecdh_suites(Ciphers)).
 
-filter_suites_signature(_, Ciphers, {3, N}) when N >= 3 ->
+filter_suites_signature(_, Ciphers, ?'TLS-1.X'=Version) when Version >= ?'TLS-1.2' ->
```
kikofernandez, line 1053: @IngelaAndin I am guessing I can remove this first clause as well, as pointed out by your comments from below, correct?

lib/ssl/src/ssl_cipher.erl
```diff
@@ -321,45 +321,42 @@ block_decipher(Fun, #cipher_state{key=Key, iv=IV} = CipherState0,
...
-        ++ tls_v1:des_suites(Version)
-        ++ tls_v1:rc4_suites(Version);
+suites(?'TLS-1.X'=Version) ->
+    tls_v1:suites(Version);
+suites(?'DTLS-1.X'=Version) ->
+    dtls_v1:suites(Version).
+all_suites(?'TLS-1.3' = Version) ->
+    suites(Version) ++ tls_testing_suites(?'TLS-1.2');
+all_suites(?'TLS-1.X' = Version) ->
+    suites(Version) ++ tls_testing_suites(?'TLS-1.2');
```
kikofernandez, line 331: this is a bug! This should be `suites(Version) ++ tls_testing_suites(Version);`
phew!

lib/ssl/src/ssl_cipher.erl
IngelaAndin: No here you should match on TLS-1.2. That is how I know that the function rsa_signed_suites will only be called for lower than TLS-1.2. Also I know that this code will never be called when the version is TLS-1.3.

lib/ssl/src/ssl_cipher.erl
IngelaAndin: I think they should be called leagcy_suites rather than testing_suites

u3s (changes requested):
1. I remember our "ssl mission statement" as follows:

make ssl application:
- less buggy (preferably make it harder to introduce new bugs)
- easier to maintain
- and understand

The only improvement I see with current code is that version constants are not hardcoded everywhere (but without version handling API, maybe macro names should rather be named TUPLE_3_4 ,TUPLE_3_3 ... TUPLE_254_253 ...).
Unfortunately I see some drawbacks (see comments inline).

2. PR is much bigger than I expected it to be. I think we should:
- reduce the scope of it, by making a separate PR for additional 'code cleanup' suggestions
- since it is not going to be cheap, make sure we agree what we want to achieve with it and do it right

3. Should have a small workshop and agree on what we want to achieve and how. Maybe agree implementation steps?

4. Maybe we should iterate on smaller code chunks (rather whole ssl updates) for finding the wanted solution.

lib/ssl/src/ssl_logger.erl
kikofernandez: got it, so nothing to do :)

lib/ssl/src/tls_gen_connection.erl
kikofernandez: what legacy macro are you referring to? :)

lib/ssl/src/tls_handshake_1_3.erl
kikofernandez: what legacy macro?

lib/ssl/test/ssl_api_SUITE.erl
kikofernandez: please ignore, I marked myself where I was doing work before the end of some day. simply I reminder that I have already finished.

lib/ssl/src/tls_handshake_1_3.erl
IngelaAndin: ?LEGACY_VERSION from tls_record_1_3.hrl

lib/ssl/src/dtls_gen_connection.erl
```diff
@@ -545,10 +545,9 @@ handle_info({CloseTag, Socket}, StateName,
     %% with widespread implementation practice.
```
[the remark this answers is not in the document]
kikofernandez: https://github.com/erlang/backlog/issues/130
but this seems to be a private repo

lib/ssl/src/ssl_handshake.erl
kikofernandez: so this change is correct because it states that the version must be less or equal than TLS-1.2?

lib/ssl/src/dtls_gen_connection.erl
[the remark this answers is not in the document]
kikofernandez: got it, so the branch name should be changed then.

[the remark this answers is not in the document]
IngelaAndin: Well yes, to avoid any confusion.  I think you could change the name of your branch and later (when next step feels ready) make a new PR replacing this one.

lib/ssl/src/ssl_cipher.erl
kikofernandez: We have a test that breaks if I remove it. 
So either the test is wrong or we should keep it...

lib/ssl/src/ssl_cipher.erl
```diff
@@ -1097,24 +1084,20 @@ rsa_signed(_) ->
...
     end.
 %% Cert should be signed by RSA
-rsa_signed_suites(Ciphers, Version) ->
-    filter_suites(Ciphers, #{key_exchange_filters => [rsa_signed(Version)],
+rsa_signed_suites(Ciphers) ->
+    filter_suites(Ciphers, #{key_exchange_filters => [rsa_signed()],
                              cipher_filters => [],
                              mac_filters => [],
                              prf_filters => []}).
-ecdsa_signed({3,N}) when N >= 3 ->
```
kikofernandez, line 1105: @IngelaAndin the code here is not semantically the same, I guess this is my error and not correct?

lib/ssl/src/ssl_cipher.erl
```diff
@@ -1079,15 +1074,7 @@ filter_suites_signature(ecdsa, Ciphers, Version) ->
 %% extension.  The names DH_DSS, DH_RSA, ECDH_ECDSA, and ECDH_RSA are
 %% historical.
 %% Note: DH_DSS and DH_RSA is not supported
-rsa_signed({3,N}) when N >= 3 ->
```
kikofernandez, line 1082: @IngelaAndin  here is the same, is my code change correct? I am guessing not...

kikofernandez:
This PR is closed in favour of #7065

PR #6940, opened 2023-02-28
Improve `lists:nth/2` and `nthtail/2`

Maria-12648430:
This PR provides performance improvements for `nth/2`, `nthtail/2` ~and `append/1,2`~ in the `lists` module.

* `nth/2` and `nthtail/2`: There was a guard in one of the clauses that would be checked in each step of the recursion. Another drawback was that both functions would accept floats as values for the index, and only crash after having gone through the recursion to the point where it dropped below 1.
I moved this check, augmented to also check for the indexes being integers, plus some shortcuts for common edge cases, in front functions and put the actual work in helper functions that run unchecked. This implementation is (AFAICS) fully equivalent to the old implementation, but performs better, with back-off-a-napkin benchmarks showing the runtime being around 75% of that of the old implementation.

* ~`append/1`: The old implementation appended everything by means of `++`. In the case of appending an empty list to non-empty lists, this is wasteful, as the non-empty list still needs to be traversed.
In my implementation, empty lists are ignored. Also, I added a shortcut that, when a list consists only of a single element, this element is consed instead of the entire singleton list being prepended via `++` (TBH, I didn't expect much from this, but benchmarking with long lists of singletons showed the runtime to be at about 50% of the old implementation).
This implementation is _not_ fully equivalent to the old one, but still in accordance with the documentation.~
  * ~The old implementation allowed the last element of the argument list to be any term, with the result being an improper list or the term itself if it was the only term in the list or all other elements before it were empty lists.
In my implementation, non-list arguments are not allowed anywhere in the argument list.~
  * ~The old implementation would accept an improper list only if it was the last or only element in the list of lists.
In my implementation, an improper list is also allowed if it is followed by only empty lists.~

* ~`append/2`: The old implementation simply mapped to `++`. If the second list was empty, this was wasteful for the same reasons as outlined in `append/1`.
In my implementation, if either of the input lists is empty, the respective other is simply returned, thereby avoiding the needless appending of an empty list.
This implementation is _not_ fully equivalent to the old one.~
  * ~The old implementation allowed the second argument to be any term, with the result being an improper list or the term itself if the first argument was an empty list.
In my implementation, non-list arguments are not allowed. And while this is in accordance with the documented specs, the documentation unfortunately contains the sentence "`lists:append(A, B)` is equivalent to `A ++ B`", and `A ++ B` _allows_ `B` to be any term (which actually contradicts the specs). This behavior can be added back if you think this is required, but given the choice, I'd rather not :sweat_smile:~
  * ~The old implementation would accept improper lists only in the second argument.
In my implementation, an improper list is allowed as first argument if the second argument is an empty list. This also contradicts the stated equivalence to `A ++ B`, which does _not_ allow `A` to be an improper list even if `B` is an empty list.~

### DIFF
--- a/lib/stdlib/src/lists.erl
+++ b/lib/stdlib/src/lists.erl
@@ -196,8 +196,12 @@ reverse([A, B | L]) ->
       T :: term().
 
 nth(1, [H|_]) -> H;
-nth(N, [_|T]) when N > 1 ->
-    nth(N - 1, T).
+nth(N, [_|_]=L) when is_integer(N), N > 1 ->
+    nth_1(N, L).
+
+nth_1(1, [H|_]) -> H;
+nth_1(N, [_|T]) ->
+    nth_1(N - 1, T).
 
 -spec nthtail(N, List) -> Tail when
       N :: non_neg_integer(),
@@ -205,10 +209,15 @@ nth(N, [_|T]) when N > 1 ->
       Tail :: [T],
       T :: term().
 
+nthtail(0, []) -> [];
+nthtail(0, [_|_]=L) -> L;
 nthtail(1, [_|T]) -> T;
-nthtail(N, [_|T]) when N > 1 ->
-    nthtail(N - 1, T);
-nthtail(0, L) when is_list(L) -> L.
+nthtail(N, [_|_]=L) when is_integer(N), N > 1 ->
+    nthtail_1(N, L).
+
+nthtail_1(1, [_|T]) -> T;
+nthtail_1(N, [_|T]) ->
+    nthtail_1(N - 1, T).
 
 %% prefix(Prefix, List) -> (true | false)
 


michalmuskala:
I'm pretty sure that `List ++ []` is already just returning `List` without traversal. I don't think the optimisation of `append/2` actually changes any behaviour
```erlang
Eshell V13.1.4  (abort with ^G)
1> List = [1, 2, 3].
[1,2,3]
2> New = List ++ [].
[1,2,3]
3> erts_debug:same(List, New).
true
4> New2 = [] ++ List.
[1,2,3]
5> erts_debug:same(List, New2).
true
```

Maria-12648430:
> I'm pretty sure that List ++ [] is already just returning List without traversal.

```erlang
Erlang/OTP 26 [RELEASE CANDIDATE 1] [erts-13.1.5] [source-955178d235] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns]

Eshell V13.1.5 (press Ctrl+G to abort, type help(). for help)
1> L=lists:seq(1, 10000000).
[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22, 23,24,25,26,27,28,29|...]
2> element(1, timer:tc(erlang, '++', [L, []])).
297071
3> element(1, timer:tc(erlang, '++', [[], L])).
1
```

michalmuskala:
Perhaps the change in behaviour should be propagated to `++` then? The equivalence of `++` and `lists:append/2` is a fairly long standing documentation item.

The biggest behaviour change will in cases where this delays discovery of improper lists. However, give the behaviour where the second argument is an empty/improper list, perhaps this is justifiable.

bjorng:
I approve the nice improvements of `nth/2` and `nthtail/2`.

I don't like the change of behavior for `append/1` and `append/2`. As currently implemented, it is very easy to understand in what circumstances they will fail. Also, I think that `'++'/2` and `lists:append/2` should continue to be equivalent.

In any case, changing the behavior of two commonly used functions that have existed basically since the birth of Erlang should not be done this late in the release cycle (after the release of the first release candidate). If it should be done at all, it should be done in Erlang/OTP 27 after approval of the OTP Technical Board.

I suggest that you take out of the behavioral changes of `append/1` and `append/2` from this pull request. If you still think that those changes should be done, submit a new pull request after the release of Erlang/OTP 26.

Regarding the optimizations in `append/1` for empty lists and singletons, have you reason to think that is a common enough use case worth optimizing for? Something that you have seen in your own programs? When using the JIT, that extra test is very cheap, but it is probably a little bit more noticeable for the traditional BEAM emulator (which is still for used 32-bit systems).

bjorng:
> I'm pretty sure that `List ++ []` is already just returning `List` without traversal.

> The value varies quite a bit on repeated calls, but never anywhere near to what I'd expect if it was just returned 😜

`++/2` always copies its left-hand side operand to ensure that a improper list will be found. If, after copying, it turns out that the right-hand side operand is `[]`, the copy will be discarded and the original left-hand side operand will be returned:

https://github.com/erlang/otp/blob/dab35a8a65080c50fbb73f9189e536e666ccb11d/erts/emulator/beam/erl_bif_lists.c#L198-L200

Maria-12648430:
> Perhaps the change in behaviour should be propagated to `++` then?

Now _that_ would be a change in semantics of one of the building blocks of the very language, pretty dangerous if you ask me :sweat_smile:

Maria-12648430:
@bjorng I took out `append` with the last commit.

bjorng:

bjorng:
Thanks for your pull request.

PR #6941, opened 2023-02-28
dialyzer: Add the --no_spec option for ignoring specs

bjorng:
This option is useful for debugging when one suspects that some specs are incorrect.

(Technically, this option is not new, but it was broken and undocumented.)

### DIFF
--- a/lib/dialyzer/doc/src/dialyzer.xml
+++ b/lib/dialyzer/doc/src/dialyzer.xml
@@ -86,8 +86,8 @@ dialyzer [--add_to_plt] [--apps applications] [--build_plt]
          [--check_plt] [-Ddefine]* [-Dname]* [--dump_callgraph file]
          [--error_location flag] [files_or_dirs] [--fullpath]
          [--get_warnings] [--gui] [--help] [-I include_dir]*
-         [--incremental] [--no_check_plt] [--no_indentation] [-o outfile]
-         [--output_plt file] [-pa dir]* [--plt plt] [--plt_info]
+         [--incremental] [--no_check_plt] [--no_indentation] [--no_spec]
+         [-o outfile] [--output_plt file] [-pa dir]* [--plt plt] [--plt_info]
          [--plts plt*] [--quiet] [-r dirs] [--raw] [--remove_from_plt]
          [--shell] [--src] [--statistics] [--verbose] [--version]
          [-Wwarn]*</code>
@@ -208,6 +208,11 @@ dialyzer --apps inets ssl ./ebin ../other_lib/ebin/my_module.beam</code>
       <item>
 	<p>Do not insert line breaks in types, contracts, and Erlang
 	  Code when formatting warnings.</p>
+      </item>
+      <tag><c>--no_spec</c></tag>
+      <item>
+	<p>Ignore functions specs. This is useful for debugging when
+	one suspects that some specs are incorrect.</p>
       </item>
       <tag><c>-o outfile</c> (or
         <c>--output outfile</c>)</tag>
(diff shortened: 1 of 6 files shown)


lib/dialyzer/doc/src/dialyzer.xml
```diff
@@ -86,8 +86,8 @@ dialyzer [--add_to_plt] [--apps applications] [--build_plt]
          [--check_plt] [-Ddefine]* [-Dname]* [--dump_callgraph file]
          [--error_location flag] [files_or_dirs] [--fullpath]
          [--get_warnings] [--gui] [--help] [-I include_dir]*
-         [--incremental] [--no_check_plt] [--no_indentation] [-o outfile]
-         [--output_plt file] [-pa dir]* [--plt plt] [--plt_info]
+         [--incremental] [--no_check_plt] [--no_indentation] [-no_spec]
```
jhogberg, line 89:
```suggestion
         [--incremental] [--no_check_plt] [--no_indentation] [--no_spec]
```

jhogberg (approved):
LGTM

ilya-klyuchnikov:
Would it ignore field types for records? or would dialyzer still consider and take into account fields types of records?

lib/dialyzer/test/options3_SUITE_data/dialyzer_options
```diff
@@ -0,0 +1 @@
+{dialyzer_options, [{use_spec, false}]}.
```
kikofernandez, line 1: not sure if we can replace the `use_spec` by `no_spec`. 
The main reason is to make the use of these flags as close as possible to the CLI flags.
It is an idea to think about :)

kikofernandez (approved):
I think this is good. I simply added a comment to think whether we want the CLI flags to be similar to the internal flags, that is, if `use_spec` should be replaced by `no_spec`, so that it is easy to see which flag triggers what behaviour in the `dialyzer_options` files

lib/dialyzer/doc/src/dialyzer.xml
bjorng: Fixed.

lib/dialyzer/test/options3_SUITE_data/dialyzer_options
bjorng: `use_spec` does not seem to be documented, so we should be able to change it.

bjorng:
> Would it ignore field types for records?

Yes, it will.

lib/dialyzer/test/options3_SUITE_data/dialyzer_options
bjorng: The Erlang API is under-documented with only the most important options included in the spec.

I will not address that in this PR. I will keep `use_spec`, but add `no_spec` as a synonym.

TD5:
> This option is useful for debugging when one suspects that some specs are incorrect.
> 
> (Technically, this option is not new, but it was broken and undocumented.)

I use this with TypEr: https://github.com/erlang/otp/blob/master/lib/dialyzer/src/typer.erl#L86, and I find it very handy there to compare the types with and without specs, to see whether the specs are introducing unexpected constraints on the resulting type Dialyzer uses (i.e. after intersecting the set of values for each argument and the return type for both spec and inferred success type).

lib/dialyzer/src/dialyzer_typegraph.erl
```diff
@@ -29,7 +29,7 @@ module_type_deps(UseContracts, CodeServer, Modules) ->
   Contracts =
     case UseContracts of
       true -> maps:from_list(dict:to_list(dialyzer_codeserver:get_contracts(CodeServer)));
-      false -> []
+      false -> #{}
```
TD5, line 32: What impact did this bug have in practice?

bjorng: If the `--no_spec` option was given, this bug caused Dialyzer to crash in a subsequent call to `maps:to_list/1`.

PR #6948, opened 2023-03-01
Use map comprehensions

bjorng:


frej (approved):
LGTM for the changes in `lib/compiler/src/beam_ssa_alias.erl`.

PR #6955, opened 2023-03-01
Add missing message send operator ! to erl_parse:binary_op/0 type

erszcz:
The message send operator `!` is listed in terminals of the grammar and an expression node for it is built using `?mkop2()` macro, but it's not listed as a valid operator in the definition of the `binary_op/0` type. This commit adds it there.

This PR is the same as #6954 but it targets `maint` instead of `master`.


bjorng:
I am closing this PR because I've included its correction in #7013.

PR #6959, opened 2023-03-02
Speedup valgrind_beamasm_update script

hizel:
Hi,

valgrind_beamasm_update script to slow, with my data (5Mb logs and 2Mb perf map) it's works hours. Patched script now works less than minute.


jhogberg:
Merged, thanks for the PR! :-)

PR #6961, opened 2023-03-02
Update unicode build script to take 'update_tests' argument

dgud:
Make it possible to update the binary testfile when updating version.


dgud:
Replace

PR #6968, opened 2023-03-03
compiler: Cache use/def sets in alias analysis pass

frej:
Reduce the cost of the alias analysis pass by hoisting the calculation of use and def sets out of the liveness fixpoint iteration using a cache. On the modules compiled by `scripts/diffable` the time spent in the alias analysis pass is reduced by 16%.

### DIFF
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -105,22 +105,23 @@ liveness([], _StMap) ->
 liveness_fun(F, StMap0) ->
     #opt_st{ssa=SSA} = map_get(F, StMap0),
     State0 = #{Lbl => #liveness_st{} || {Lbl,_} <- SSA},
-    liveness_blks_fixp(reverse(SSA), State0, false).
+    UseDefCache = liveness_make_cache(SSA),
+    liveness_blks_fixp(reverse(SSA), State0, false, UseDefCache).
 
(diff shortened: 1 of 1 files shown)


lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -153,17 +154,28 @@ blk_effective_uses(#b_blk{is=Is,last=Last}) ->
...
-                   sets:del_element(Dst, Uses0);
-               _ -> % A terminator, no defs
-                   Uses0
-           end,
+                #b_set{dst=Dst} ->
+                    %% The uses after the def do not count
+                    sets:del_element(Dst, Uses0);
+                _ -> % A terminator, no defs
+                    Uses0
+            end,
```
bjorng, line 162: I see only whitespace changes on these lines. (And the indentation was correct in my Emacs before the extra whitespace was added.)

bjorng (commented):
Looks good to me except for lines 157-161.

bjorng:
I've removed the whitespace changes and force-pushed. Added to our daily builds.

lib/compiler/src/beam_ssa_alias.erl
frej: Sloppy last minute revert of a change which did not have an impact on the performance, bad Frej! (I see that you already fixed it).

PR #6971, opened 2023-03-03
Optimise handling of missing data with `ets:lookup_element`

TD5:
A common pattern is to write code of the form
`try ets:lookup_element(A,B,C) of ... catch ... end` to handle missing data in an ETS table, since `ets:lookup_element/3` would throw an error in that case. More recently `ets:lookup_element/4` has been added, which introduces a 4th parameter to use as a default parameter in the case that there is no data in the ETS table corresponding to the given table and key. The new way is faster (see GitHub #6234), so we migrate to it here.


lib/dialyzer/src/dialyzer_codeserver.erl
```diff
@@ -98,9 +98,9 @@
 
 %% We KNOW that `error` is not a valid value in the table.
 ets_dict_find(Key, Table) ->
-  case ets:lookup_element(Table, Key, 2, error) of
-    error -> error;
-    Val -> {ok, Val}
+  case ets:lookup_element(Table, Key, 2, '$missing') of
+      '$missing' -> error;
+      Val -> {ok, Val}
```
bjorng, line 103: Fixing indentation:

```suggestion
    '$missing' -> error;
    Val -> {ok, Val}
```

lib/dialyzer/src/dialyzer_codeserver.erl
```diff
@@ -98,9 +98,9 @@
 
 %% We KNOW that `error` is not a valid value in the table.
```
Maria-12648430, line 99: This comment needs to be changed I think. But why change `error` to `'$missing'` at all?

lib/dialyzer/src/dialyzer_codeserver.erl
```diff
@@ -98,9 +98,9 @@
...
 ets_dict_find(Key, Table) ->
-  case ets:lookup_element(Table, Key, 2, error) of
-    error -> error;
-    Val -> {ok, Val}
+  case ets:lookup_element(Table, Key, 2, '$missing') of
```
Maria-12648430, line 101: I'm always suspicious when actually possible values are used on the grounds that it should not be possible to have them present because of circumstances (circumstances may change), or that it is unlikely to have such a value because it is obscure.

To be safe, the default should be a reference created with `make_ref()`, which is guaranteed to be unique.

Otherwise, admitting that creating references does not come for free as pointed out when `lookup_element/4` was introduced, it may be allowed to do it this way but should be noted somewhere that the atom `'$missing'` (and `'$missing_key'` for that matter) _must not_ be inserted into the table.

Maria-12648430 (changes requested):
See inline comments.

Maria-12648430 (approved):

sverker:
All removed try-catch would not only catch a missing key but also a missing table. Now they don't. Have you verified that to not be a potential problem.
There's a lot of failed mnesia tests. Are a missing table the cause?

TD5:
> So, @TD5... my suggestion is to keep the use of `lookup_element/4` that you introduced (I think I have heard somewhere that it is supposed to be faster if the key is absent, dunno if it is true... @sverker?), but also still wrap it all in a `try` as before to be prepared for a missing table.

Yep, that makes sense and I agree. I'll update the PR as soon as I can

Maria-12648430:
As for the other modules you changed, I believe it is not supposed to happen that the table is absent, and that catching this actually conceals that something went wrong somewhere else. With `lookup_element/3`, there is no way to distinguish "missing key" from "missing table", with `lookup_element/4` there is, so not catching the missing table may actually be an improvement.

However, I gave this only a cursory look, closer inspection is required.

max-au:
> Yep, that makes sense and I agree. I'll update the PR as soon as I can

I wonder if that actually leads to performance regression (rather than improvement). I would at least recommend benchmarking the happy case.

TD5:
Seems like this might not have been the quick improvement that I thought, so I'll leave this for now. Happy for someone else to resurrect it.

PR #6973, opened 2023-03-03
Avoid truncating thread names for better runtime observability

lexprfuncall:
The Erlang runtime gives many of its threads descriptive names.  When those threads are part of a logical group, a unique ID is added to the name for disambiguation.  After construction, many thread names in Erlang have a string length greater than 16 characters.  To fit within operating system limits, Erlang then truncates them from right to left.  To minimize confusion after truncation, the unique ID is always placed at the left of a thread name so its information is not likely to be lost.

The convention used by Erlang presents challenges to the use of thread names as keys when reporting on thread activity at the operating system level.  The more common convention, used by other runtimes, is to have the description followed by a unique ID.  When followed, sorting threads by name places like workers next to each other and the unique ID can be dropped to create a grouping key. Placing the unique ID first, as Erlang does, means that a different strategy needs to be used for sorting the threads of an Erlang process.  Furthermore, the truncation necessitates a complicated strategy for analyzing the description to identify a possible common substring to be used as a grouping key.

This change switches the Erlang runtime to use the more common convention in order to make reporting on the thread usage in an Erlang process easier for tooling.  To do so, it shortens the content of the initial printf(3) format strings to ensure their output is always 16 or fewer characters so the name is never truncated.  It also moves the unique ID in the format string to the right of the description, so the names of worker threads appear next to each other after sorting alphabetically from left to right.

To prevent the accidental creation of long thread names in the future, the silent truncation has been eliminated from the lowest-layer of thread functionality.  It now returns an EINVAL when given a long name which will be caught when the runtime is started.  This would break the NIF and driver libraries so the silent truncation has moved up to a higher layer in order to preserve compatibility.  New unit tests have been added to test setting and getting thread names.

### DIFF
--- a/erts/emulator/beam/erl_async.c
+++ b/erts/emulator/beam/erl_async.c
@@ -199,7 +199,7 @@ erts_init_async(void)
 	for (i = 0; i < erts_async_max_threads; i++) {
 	    ErtsAsyncQ *aq = async_q(i);
 
-            erts_snprintf(thr_opts.name, sizeof(thr_name), "async_%d", i+1);
+            erts_snprintf(thr_opts.name, sizeof(thr_name), "erts_async_%d", i+1);
 
 	    erts_thr_create(&aq->thr_id, async_main, (void*) aq, &thr_opts);
 	}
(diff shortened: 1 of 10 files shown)


rickard-green:
Please sign the CLA (above)

rickard-green:
Thanks for the PR!

I made some changes and force pushed those to you repository:
* Changed `snprintf()` to `erts_snprintf()` in `erl_drv_thr_create()`. There at least used to exist platforms without `snprintf()` support. I don't know if there will be other problems building on such platforms, but just to be safe...
* Changed the names so that all thread names for threads created by the runtime system itself have the prefix `erts_` making it easier to distinguish ERTS threads in output from for example `top -H` from threads of other processes.
* Added `erts_thr_setname()` and used it for setting the name on the main thread which was the only unnamed thread.

I'll run some more tests, if nothing unforeseen happens I'll merge this later this evening.

Issue #6098, opened 2022-06-20
[DNS] Handling of the LOC resource record type


bortzmeyer:
The current DNS code in `lib/kernel/src/inet_dns.hrl` does not handle the `LOC` resource record type ([RFC 1876](https://www.rfc-editor.org/info/rfc1876)). True, it is not used a lot but I have one or two projects that may use it. 

Therefore, I would be happy if  `LOC` and its encoding/decoding would be added to `lib/kernel/src/inet_dns.hrl` .

bortzmeyer:
Note there is Erlang code to deal with LOC types in [this library](https://github.com/dnsimple/dns_erlang).

RaimoNiskanen:
I am working on it...

RaimoNiskanen:
I have implemented a suggestion in PR #6982.

I selected a different data content format than in the `dns_erlang` library because `inet_dns` does not expose any records for RR data content.

Comments are welcome, preferably over there...

RaimoNiskanen:
Fixed in PR #6982.  Thank you for reporting this issue.

PR #6982, opened 2023-03-06
Implement DNS LOC record (RFC 1876)

RaimoNiskanen:
The record data Erlang representation is chosen to be similar to the Master File Format described in RFC section 3, i.e latitude and longitude in degrees, and altitude, size and precision in meters; all as floating point values.

Encoding also supports integers with latitude and longitude in thounsands of an arc second, and alitude, size and precision in centimeters, as the RDATA format in RFC section 2.

Encoding of size and precision encodes the smallest encoding value not less than the input value; e.g 100 encodes as 100, 4711 encodes as 5000, and 90001 encodes as 100000.

### DIFF
--- a/lib/kernel/src/inet_dns.erl
+++ b/lib/kernel/src/inet_dns.erl
@@ -1,7 +1,7 @@
 %%
 %% %CopyrightBegin%
 %%
-%% Copyright Ericsson AB 1997-2021. All Rights Reserved.
+%% Copyright Ericsson AB 1997-2023. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
@@ -48,6 +48,7 @@
 	 make_dns_query/0, make_dns_query/1,
 	 make_dns_query/2, make_dns_query/3]).
 -include("inet_dns_record_adts.hrl").
+
 
 %% Function merge of #dns_rr{} and #dns_rr_opt{}
 %%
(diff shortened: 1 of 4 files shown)


bortzmeyer:
> Awaiting comments/approval from issue #6098 reporter @bortzmeyer.

I'm not an expert in Erlang internal code (I use it via Elixir) but this code seems OK.

RaimoNiskanen:
@bortzmeyer: What I would like to have feedback on is the API choices e.g having the main representation as floating point degrees for latitude and longitude, and floating point meters for the other parameters.  That representation is intended to be similar to the RFC's Master File Format description.

The secondary value representation (possible to set but not what you get) is integers as defined in the RFC but offset to signed, just like in the DNS library you linked to, except for rounding up size and precision.

Also, I have chosen to allow several alternative tuple formats (when setting) with default values as in the RFC's Master File Format description: `{{Lat,Long}, Alt, Size, {HorizPrec,VertPre}}` | `{{Lat,Long}, Alt, Size}` | `{{Lat,Long}, Alt}`.  Is that useful or just code bloat?

bortzmeyer:
For [code I wrote in Elixir](https://framagit.org/bortzmeyer/drink/-/blob/master/lib/drink/utils.ex#L144), I also used the very same representation (because it is more "natural" than the integers used in the wire format), so I vote yes.
I used default values for things like precision but I did not really think about the best way to represent it, so I abstain on this one.

RaimoNiskanen:
Great!  Luck on my part that your feedback was positive, since i merged 7 minutes before...

PR #6984, opened 2023-03-06
Unicode 15

dgud:
Update to use Unicode 15 tables.

Contains the original PR see #6312 and some missing fixes.


dgud:
Does this look ok @eksperimental?

Issue #5087, opened 2021-08-02
add support for truncated SHA-512/t: SHA-512/256 and SHA-512/224


faried:
**Is your feature request related to a problem? Please describe.**

I was trying to port some code over from Javascript when I found out that `crypto` did not implement SHA-512.  Truncated SHA-512 is SHA-512 with a different initial value, and with the output truncated at the right place (32 bytes for SHA-512/256, for example).  The `crypto` module does not appear to export anything that will let me fiddle with this initial value.

OpenSSL has supported these functions 1.1.1, released almost [three years ago](https://www.openssl.org/news/changelog.txt).

**Describe the solution you'd like**

Add support for SHA512/256 to `crypto`.

**Describe alternatives you've considered**

I [forked](https://github.com/faried/erlsha2/commits/sha512-truncated) a pure Erlang implementation of SHA2 from 2015, disabled the code that loaded the NIF, and added my functions.  It's not ideal, but it'll work for my current project.

HansN:
This is currently not prioritized by Erlang/OTP, but Pull Requests are welcome.

KayEss:
I'm wondering about taking a stab at adding the two truncated hash functions to the crypto module. I guess my strategy will be to go through occurrences of `sha512` in the crypto module and add `sha512_224` and `sha512_256` where appropriate.

The open_ssl documentation https://www.openssl.org/docs/manmaster/man3/EVP_sha512_224.html lumps the truncated functions in with the ones that are already being used, so it should be a very simple change (I hope).

Is there anything I should keep in mind? So far I've cloned and built this project and I can run the tests using `make ARGS="-suite crypto_SUITE"` and I've read the contributing guide.

sverker:
@KayEss 
If this is just about adding a new digest version then it should be quite straight forward. You will end up down in `lib/crypto/c_src/digest.c` and the array `digest_types` containing all the different supported digest types. Note that OpenSSL >= 3.0 uses a slightly different initialization with EVP_MD_fetch.
Then also add tests and documentation.

KayEss:
@sverker I have a commit that does the hash part of this: https://github.com/BiggestLab/otp/commit/d595652a47c763b0b491550ee94d77083e2a6ad7

Should I do a PR with this, or is there something obvious I ought to fix first?

sverker:
Looks good. Make it a PR.

PR #6985, opened 2023-03-06
inet_dns support for UPDATE, NOTIFY and TSIG

jimdigriz:
Some proposed enhancements to `inet_dns`:
 * guard to catch when there is more than one `EDNS(0)` option present as [discussed](https://github.com/erlang/otp/pull/2959#issuecomment-759405956)
 * support for `NOTIFY` and `UPDATE`
     * `UPDATE` is a little more complicated as zero length data is now allowed and a new `NONE` class
     * removed the `UPDATE{A,D,DA,M,MA}` to kick up a conversation if these should still exist
     * includes the defines for `IXFR` and a few other rcodes
 * Initial support for `TSIG` [suggested some time back that there would be interest to have this baked into `inet_dns.erl`](https://github.com/erlang/otp/pull/2959#issuecomment-759405956)
     * ~~Not yet added support for `TSIG` over TCP as [described by RFC8945, section 5.3.1](https://datatracker.ietf.org/doc/html/rfc8945#section-5.3.1); this is on my list of things to resolve though~~ - this has now been added though mostly untested

I suspect none of this is particularly controversial but the `TSIG` API does need to be thrashed out.

I am submitting what I see as version zero (0) and finally worth sharing around. I really would prefer the focus be on the API as the implementation will need burning and redoing a few more times from scratch as we carve out the interface.

In the spirit of `inet_dns`, this is an undocumented API in that it expects the user to handle policy whilst it only sets out to handles the encoding/decoding/verification.

I tried a few interface/API iterations and found having settled on passing in `#dns_rec{}` ended up being the least offensive and made the most sense from a possible user perspective but this is my view only. Similarly I tried putting this all in `inet_dns.erl` but after a while the sign/verify functionality just felt more correct in its own module, again this is my view, but it did mean I had to export the name encode/decode routines from `inet_dns` so maybe this should be folded back into `inet_dns`.

Broken out of the commit so to hack out the conversation here, this is what it smells like (for a simple client expecting a single UDP response packet to its request):
```
Zone = "example.com".
Name = "cheese".
Request = inet_dns:make_msg([
        {header,inet_dns:make_header([
                {id,rand:uniform(65536) - 1},
                {opcode,update}
        ])},
        % RFC2136, section 2
        {qdlist,[       % Zone
                inet_dns:make_dns_query([
                        {domain,Zone},
                        {class,in},
                        {type,soa}
                ])
        ]},
        {nslist,[       % Update
                inet_dns:make_rr([ 
                        {domain,Name ++ "." ++ Zone},
                        {ttl,300},
                        {class,in},
                        {type,a},
                        {data,{192,0,2,1}}
                ])
        ]}
]).

% my thinking here is to be like crypto:mac_init as we need rolling state especially for TSIG over TCP
TSigState0 = inet_dns_tsig:init([{keys,[{"mykey","moocowmoocow"}]}]).

{ok,RequestSigned,TSigState1} = inet_dns_tsig:sign(Request, TSigState0).
RequestPkt = inet_dns:encode(RequestSigned).

{ok,Sock} = gen_udp:open(0, [binary,{active,false}]).
gen_udp:send(Sock, {{127,0,0,1},5354}, RequestPkt).
{ok,{_,_,ResponsePkt}} = gen_udp:recv(Sock, 0).

{ok,ResponseSigned} = inet_dns:decode(ResponsePkt).

% will return {error,...} if something is wrong
% like inet_dns:decode, will throw formerr if something is wrong with the packet...personally not a fan of this behaviour
{ok,TSigState2} = inet_dns_tsig:verify(ResponseSigned, TSigState1).

Response = inet_dns:msg(ResponseSigned).
```

Usage for a client (shown above and mostly tested):
1. `tsig:init/1`
    * configuration `algs` really is only useful as a single item as we use the first algorithm (default: `sha256`) provided
    * configuration `keys` really is only useful as a single item as we use the first name/secret entry
3. `tsig:sign/2` the request and send it
4. for each response packet (there can be one or typically for TCP more than one) call `tsig:verify/2`
5. ...remember to consider the gotchas/warnings in the comments at the top of `inet_dns_tsig.erl`

Usage for a server (not tested as I am still in the middle of building my DNS server...will get back to you...):
1. `tsig:init/1`
    * configuration `algs` lists all the algorithms we are willing to support (default: only `sha256`, maybe the default for 'server' should be changed to all the algorithms)
    * configuration `keys` lists all the name/secret combinations that are accepted
3. call `tsig:verify/2` on the request packet
4. construct the response and call `tsig:sign/2`, sending it to the client; repeat this until the response is completely sent


jimdigriz:
Okay, the name compression tripped me up[1] so the API now requires the user to pass in the original binary packet. TSIG over TCP now works.

An example of a TSIG signed DNS update is as shown:
```
Zone = "example.com".
Name = "cheese".
RequestPkt = inet_dns:encode(inet_dns:make_msg([
        {header,inet_dns:make_header([
                {id,rand:uniform(65536) - 1}, 
                {opcode,update}
        ])},
        % RFC2136, section 2
        {qdlist,[       % Zone
                inet_dns:make_dns_query([
                        {domain,Zone},
                        {class,in},
                        {type,soa}
                ])  
        ]}, 
        {nslist,[       % Update
                inet_dns:make_rr([
                        {domain,Name ++ "." ++ Zone},
                        {ttl,300},
                        {class,in},
                        {type,a},
                        {data,{192,0,2,1}}
                ])  
        ]}  
])).
TSigState0 = inet_dns_tsig:init([{keys,[{"mykey","moocowmoocow"}]}]).
{ok,RequestPktSigned,TSigState1} = inet_dns_tsig:sign(RequestPkt, TSigState0).
{ok,Sock} = gen_udp:open(0, [binary,{active,false}]).
ok = gen_udp:send(Sock, {{127,0,0,1},5354}, RequestPktSigned).
{ok,{_,_,ResponsePktSigned}} = gen_udp:recv(Sock, 0). 
ok = gen_udp:close(Sock),
{ok,TSigState2} = inet_dns_tsig:verify(ResponsePktSigned, TSigState1).
{ok,ResponseSigned} = inet_dns:decode(ResponsePktSigned).
Response = inet_dns:msg(ResponseSigned).
```
    
An example of a TSIG signed DNX AXFR query is as shown:
```
ZoneAXFR = "com".
RequestAXFRPkt = inet_dns:encode(inet_dns:make_msg([
        {header,inet_dns:make_header([
                {id,rand:uniform(65536) - 1}, 
                {opcode,query}
        ])},
        {qdlist,[
                inet_dns:make_dns_query([
                        {domain,ZoneAXFR},
                        {class,in},
                        {type,axfr}
                ])
        ]}  
])).
TSigStateAXFR0 = inet_dns_tsig:init([{keys,[{"mykey","moocowmoocow"}]}]).
{ok,RequestAXFRPktSigned,TSigStateAXFR1} = inet_dns_tsig:sign(RequestAXFRPkt, TSigStateAXFR0).
{ok,SockAXFR} = gen_tcp:connect({127,0,0,1}, 5354, [binary,{active,false},{nodelay,true},{packet,2}]).
ok = gen_tcp:send(SockAXFR, RequestAXFRPktSigned).
Recv = fun(Recv, A) ->
        case gen_tcp:recv(SockAXFR, 0) of
                {ok,Pkt} ->
                        Recv(Recv, [Pkt|A]);
                {error,closed} ->
                        lists:reverse(A)
        end 
end.
ResponseAXFRPktsSigned = Recv(Recv, []),
ok = gen_tcp:close(SockAXFR).
TSigStateAXFR2 = lists:foldl(fun(P, TS0) ->
        {ok,TS} = inet_dns_tsig:verify(P, TS0), TS
end, TSigStateAXFR1, ResponseAXFRPktsSigned).
[ case inet_dns:decode(P) of {ok,M} -> inet_dns:msg(M) end || P <- ResponseAXFRPktsSigned ].
```

[1] we need to do an HMAC on the original DNS packet and every DNS name compression algorithm is different so by decoding and re-encoding the packet we end up with a different output

jimdigriz:
Would be helpful to receive a summary of either "this is a pile of garbage, please haul it away" or "we like it in principle, we are currently too busy to chew through the sharp edges".

Just depressing for a contributor to watch other PRs being worked through whilst receiving radio silence and being left in limbo.

Really it is absolutely zero problem if this is considered a hot steaming pile of garbage, more than happy to just move it into my own private library.

For my purposes it does not need to be in OTP, I only submitted as it was mentioned there was some interest in getting something like this into OTP.

jimdigriz:
I think the biggest bit here is that the decision needs to be made to either:
 1. have one (the existing) complete DNS decoder and a second one (for TSIG) that breaks each RR section into a list of binaries
 2. you rework the existing DNS decoder to be a two pass affair, the first one breaks the everything up into a list of binaries where then TSIG can do its thing, whilst the main decoder then immediately goes into the binary to record parser

Most of my PR is just a reimplementation of the DNS decoder; I did it this way so you could see what is needed rather than just jumping in an butchering the existing decoder.

lib/kernel/src/inet_dns.erl
```diff
@@ -681,17 +715,6 @@ decode_name_label(Label, Name, N) ->
...
-encode_data(Comp, _, ?S_A, in, Addr) ->
-    {A,B,C,D} = Addr,
-    {<<A,B,C,D>>,Comp};
-encode_data(Comp, _, ?S_AAAA, in, Addr) ->
-    {A,B,C,D,E,F,G,H} = Addr,
-    {<<A:16,B:16,C:16,D:16,E:16,F:16,G:16,H:16>>,Comp};
-encode_data(Comp, _, ?S_WKS, in, Data) ->
-    {{A,B,C,D},Proto,BitMap} = Data,
-    BitMapBin = iolist_to_binary(BitMap),
-    {<<A,B,C,D,Proto,BitMapBin/binary>>,Comp};
```
RaimoNiskanen, line 694: You made these no longer class `in` specific.  How come?
In RFC 1035 they are (not AAAA) listed under section 3.4. ARPA Internet specific RRs.

RaimoNiskanen:
I am happy you have made the effort to understand the RFC:s, and to write code that works.  From here we can look at details and structure.

I have read RFC 8945 (could have been more thorough) and now try to understand how your code implements it...

jimdigriz:
Let me know if I can help in any way, even if it is "hey, we think we want the DNS decoder to work like this, can you refactor things for us and we'll review"

Thanks!

RaimoNiskanen:
I'll write down things in the order I stumble on them...

I belong to the camp that limits lines to 80 characters, but we have people from the other camp in my team...  If you have a strong opinion about not limiting code to 80 characters I will accept that, but I would really like the line limit to be 80 characters...

I do not see the point of converting to Gregorian time for the user interface format.  Why not simply have Posix time there.  Furthermore I think using `os:system_time(seconds)` would be less of a surprise to system administrators.  `erlang:system_time/*` might be quite a bit off from the OS Posix time after changing the system clock. I believe `os:system_time/*` is the best value we have for the Posix time that the sysadmin has set.

I have a hard time understanding the code in `inet_dns_tsig:do_verify/2` regarding `#tsig_state.qr` as in when it is the request, the first response, an unsigned intermediate response, the peculiar mac update for `QR < 3` - when and why does that happen.  The interval `1 < QR, QR < 100` should be 98 messages, not 99, etc.  More comments would be nice, and perhaps it is possible to have a flatter or in some other way clearer decision tree?

There was something in the RFC about at least 10 bytes for a truncated MAC, but I cannot find code for that.

Should `#tsig_state.key` be `{string(), iodata()} | #{string() := iodata()}`, and `.alg` be `tsig_alg() | #{tsig_alg() := []}` to facilitate lookup for the server.  The client can take  a`Config` that specifies one key and alg, but the server knows all supported algs and gets all keys it should know of. `.alg` can have a default of all supported.

I think the abstraction border between `inet_dns` and `inet_dns_tsig` is basically right, but it would be nice if we could get the separation more like what is indicated in the RFC by adding specialized helper functions in `inet_dns`:
* For encoding: Introduce `inet_dns:encode_comp(Q) -> {Binary, Comp}`, sign that `Binary` with a function in `inet_dns_tsig` to get a `TsigRR`, and use a new `inet_dns:encode_tsig(Binary, Comp, TsigRR) -> NewBinary`.  With this it would be possible to append a TSIG record to `Binary`, and patch ARCOUNT without unpacking Binary, right?
* Would it be possible to decode first, extract the TSIG record, then verify?  Use the normal `inet_dns:decode(Buffer)`, work on the RR level to find the TSIG record, then implement `inet_dns:decode_tsig(Buffer, TsigRR) -> NewBuffer` where `NewBuffer` is the original not signed binary message that can be verified with a function in `inet_dns_tsig`.

Is this a feasible API, or is there something in the RFC:s that I have missed?  Is this any better?

I think the duplicated functions in `inet_dns_tsig` such as `encode_name/1` and `decode_name/1` miss some tests, for example for maximum length, for compression pointer, etc, so as you say; it would be good to use the ones in `inet_dns` to have the encode/decode details in one module.

RaimoNiskanen:
> Sure, I don't mind either way, but if you asked me to s/\t/    /g we might have a fight on our hands!

Now that you mention it ;-) ... no, but we prefer to use space only indent for new modules. However, if this is already written with tabs I will not insist on changing that.  But I would be happy if you can rewrite to keep the line length under 80 (76) characters.

And use Posix seconds in the user data, also use `os:system_time(seconds)` for that - you had no comment on `os:system_time` vs. `erlang:system_time`?

> To preserve the sanity of others, should I remove the overloading of `.qr` and have counter to run alongside it? Alternatively better documentation?

I think better comments and clearer code will be good enough.  I still think that `2..99` would be 98 intermediate non-signed messages, not 99.  For integer ranges I find it often clearer to consistently use `=<` which demonstrates the interval end.

Would using `request | response | 1..99` for `#tsig_state.qr` cause too clumsy code, or could it be a clarification?

I missed `?MAC_SIZE_MIN`.  Now it is clear.

> ... but maybe that undocumented'ness is a good reason why this is not a problem?

We can somewhat treat it as an internal module and modify it as we like, but we know we have external users, so we want to only do backwards compatible changes, if possible.  Adding specialized functions for TSIG would be fine.

> Only way I see it is we have to decode the whole thing.

Then it would be preferable to not duplicate code getting a parallel decoder implementation.

> One idea I have is we could update #dns_rr{} to include the byte offset of where the RR starts in the binary ...

I think it would be enough to add a byte offset field to the TSIG RR  `#dns_rr.data` term.
Or maybe like this...:
* Change `inet_dns:decode_rr_section/5` to have special treatment for `?S_TSIG` so it verifies that `Rest` is `<<>>` and returns `{{TsigRR, Offset}, reverse(RRs), boolean()}`.
* Change `inet_dns:do_decode/1` to return `{#dns_rec, {TsigRR,Offset} | undefined}`.
* Change `inet_dns:decode/1` to append the `TsigRR` to `#dns_rec.arlist` unless `undefined`.
* Add a function `inet_dns:decode_tsig/1` that returns `{ok, {#dns_rec{}, TsigRR, BinaryPatchedToOriginal}`.

`Offset` can be part of `TsigRR`, or as suggested above, be in a wrapping tuple.

jimdigriz:
> > Sure, I don't mind either way, but if you asked me to s/\t/    /g we might have a fight on our hands!
> 
> Now that you mention it ;-) ... no, but we prefer to use space only indent for new modules. However, if this is already written with tabs I will not insist on changing that. But I would be happy if you can rewrite to keep the line length under 80 (76) characters.

Only a suggestion, it is useful at times when committing code to *purposely* merge incorrectly whitespaced segments.

The theory is that at some later point that someone with OCD will come along later, fix it up...and then `git blame` not longer points to you :)

> ...also use `os:system_time(seconds)` for that - you had no comment on `os:system_time` vs. `erlang:system_time`?

Did not see any need to discuss it, took this as a signal on what the OTP team want to see in a v1 rework.

> > One idea I have is we could update #dns_rr{} to include the byte offset of where the RR starts in the binary ...
> 
> I think it would be enough to add a byte offset field to the TSIG RR `#dns_rr.data` term. Or maybe like this...:
> 
> * Change `inet_dns:decode_rr_section/5` to have special treatment for `?S_TSIG` so it verifies that `Rest` is `<<>>` and returns `{{TsigRR, Offset}, reverse(RRs), boolean()}`.

I suspect it will never occur, mostly due to the contract around TSIG, but treating things 'special' feels like something that will burn someone later on; also to date 100% of OTP only Erlang processed DNS packets are TSIG less :)
 
> `Offset` can be part of `TsigRR`, or as suggested above, be in a wrapping tuple.

I think I would prefer to put it in the `#dns_rr{}` record.

It is easier to reason about (no RR is special, I hate them all equally!) plus it will help others working on decoding problems where they can now tie back each RRs to where they appeared in the originating binary; my thinking here is when you are inspecting your PCAPs in Wireshark.

I think I have everything I need from a first review to rework the TSIG component.

Did you want to cherry pick out the EDNS(0) guard commit from this PR? Also what do you want me to do (if anything) with the NOTIFY/UPDATE commit...cherry pick or does it need some massaging in places?

RaimoNiskanen:
> I suspect it will never occur, mostly due to the contract around TSIG, but treating things 'special' feels like something that will burn someone later on; also to date 100% of OTP only Erlang processed DNS packets are TSIG less :)

Well, since `inet_dns` parses DNS messages, and we are adding TSIG parsing, then I do not see why that parsing should not handle TSIG like the special record it is, just like the OPT record has special treatment.  (We could even add a `#dns_rr_tsig{}` record, but no real point in that, I guess)  A TSIG record has to be last otherwise it is a parse error, and part of the parsing result is how large the original message was.  This behaviour is separate from signature verification.

> I think I would prefer to put it in the #dns_rr{} record.

I am afraid we much prefer to not change the `#dns_rr{}` record.  Some customers use the record definitions form the .hrl file, instead of the existing (also undocumented) functional abstraction, so adding fields would force them to recompile, which is not popular (and these are internal corporate customers that pay our salaries).  Such a change would have to be for a major release, and have warnings attached to it (even though this module is "unsupported").

I understand your reasoning about decoding, and debugging the wire format, but unfortunately ... the above.

> Did you want to cherry pick out the EDNS(0) guard commit from this PR? Also what do you want me to do (if anything) with the NOTIFY/UPDATE commit...cherry pick or does it need some massaging in places?

If I should take the first two commits, then you would have to wait until I have merged to master for you to rebase on that, to avoid merge conflicts, furthermore, it seems UPDATE/NOTIFY are the major use cases you have for TSIG.  So I do not mind keeping these commits in this PR.

But, there seem to be two diff hunks in the 3:rd commit (inet_dns: support for TSIG requests), moving A, AAAA and WSIG records from class IN RRs to any class RRs, that maybe should be in the 2:nd commit (inet_dns: support for UPDATE and NOTIFY DNS packets).  If so those diff hunks should be moved.

jimdigriz:
> > I think I would prefer to put it in the #dns_rr{} record.
> 
> I am afraid we much prefer to not change the `#dns_rr{}` record. Some customers use the record definitions form the .hrl file, instead of the existing (also undocumented) functional abstraction, so adding fields would force them to recompile, which is not popular (and these are internal corporate customers that pay our salaries).
>
> Such a change would have to be for a major release, and have warnings attached to it (even though this module is "unsupported").
>
> I understand your reasoning about decoding, and debugging the wire format, but unfortunately ... the above.

I am 100% okay with waiting till OTP 28 for this; not the [first time I have waited on a PR to be merged](https://github.com/erlang/otp/pull/1968).

For now I am happy to slum the solution at my end for my use case with double parsing and passing a [binary blob into the DNS encoder as I used to for CAA RRs](https://github.com/erlang/otp/pull/2827#issuecomment-721837629).

If though the preference is to get this done sooner rather than later, then I'll start the work to drag that byte offset kicking'n'screaming through the decoder. I assume this is what you want unless I hear otherwise.

> > Did you want to cherry pick out the EDNS(0) guard commit from this PR? Also what do you want me to do (if anything) with the NOTIFY/UPDATE commit...cherry pick or does it need some massaging in places?
> 
> If I should take the first two commits, then you would have to wait until I have merged to master for you to rebase on that, to avoid merge conflicts, furthermore, it seems UPDATE/NOTIFY are the major use cases you have for TSIG. So I do not mind keeping these commits in this PR.

I suspect if the TSIG stuff was ready *today* for merging, it would not make it into OTP 26.x.

Rebasing and tracking master is really little to no effort, and the TSIG work is not tangibly impacted by it.

Only asking as whatever is easier for you? If you prefer to have the PR retain all three features, that works for me...the cost at my side is no different carrying one or three features out of tree.
 
> But, there seem to be two diff hunks in the 3:rd commit (inet_dns: support for TSIG requests), moving A, AAAA and WSIG records from class IN RRs to any class RRs, that maybe should be in the 2:nd commit (inet_dns: support for UPDATE and NOTIFY DNS packets). If so those diff hunks should be moved.

Good catch, looks like I crossed the streams, I'll get that fixed.

RaimoNiskanen:
> If though the preference is to get this done sooner rather than later, ...

It would be nice to get this finished while I have not gotten too deep into to much other stuff and forgotten about the details.  Other than that there is no hurry.

> ... then I'll start the work to drag that byte offset kicking'n'screaming through the decoder. I assume this is what you want unless I hear otherwise.

To just get the byte offset out there would not be much difference to have it as a new field in the `#dns_rr{}` record compared to in the `#dns_rr.data` field for a TSIG record.  But I think the DNS decoder should be aware  of the TSIG record semantics and verify that it is last in the AR section.

Adding an API function that also removes the TSIG record and returns it separated from the DNS message would be an optimization to avoid completing a DNS message you do not want and then having to parse the DNS whole message again to do exactly that.

> I suspect if the TSIG stuff was ready today for merging, it would not make it into OTP 26.x.

Maybe in theory, but, ... no.

> If you prefer to have the PR retain all three features, that works for me...

I think, then, I would prefer to keep all 3 in the same branch, with the 2 first as separate commits, as it is now.

jimdigriz:
@RaimoNiskanen I have made the TSIG related updates we discussed, here are the highlights:
 * I actually went and created `#dns_rr_tsig{}` (similar to `#dns_rr_opt{}`) inspired by the [passing remark you made](https://github.com/erlang/otp/pull/6985#issuecomment-1491993176)
     * this actually has made it easier (I think) for the user to interacting with `inet_dns_tsig.erl` and the implementation a lot easier to understand
     * also we avoid the two pass decoding/encoding
 * I have created an example client and server in the form of unit tests
     * client talks to 'nsd' so no changes needed there
     * server implementation expects `dig` to be installed (part of `bind9-dnsutils`)

Other possible points of interest:
 * not able to do `UPDATE`/`NOTIFY` end-to-end tests as `nsd` does not support Dynamic DNS updates
     * maybe you want to consider moving to Knot DNS?
 * I have split up the previously bundled/aggregated commits into standalone parts
 * included some fixes for some of your DNS unit tests
 * I have rebased against 'master'

jimdigriz:
One outstanding problem I just discovered is what to do with the truncated hash algorithms described in https://github.com/erlang/otp/issues/5087

Though here we would just disable their use though maybe I should just do a crypto PR just to solve this would be easier...really do not want this turning into a can of worms :(

jimdigriz:
@RaimoNiskanen one thought is as you are using `bash`, I could amend https://github.com/erlang/otp/pull/6985/commits/53742e6052d813b5ac174fc6da81b8a1b9ba088e to wait until `nsd` has opened the TCP port instead and remove the use of `nsd-control`?

lib/kernel/src/inet_dns_tsig.erl
```diff
@@ -0,0 +1,340 @@
...
+%%  * truncated messages must include the TSIG RR as per RFC8945, section 5.3
+%%  * if running as a server, generate suitable error responses
+%%    as per RFC8945, section 5.2
+
+-export([init/1, sign/2, sign/3, verify/3]).
+
+-include("inet_dns.hrl").
+
+-define(ALGS_SUPPORTED, [md5, sha, sha224, sha256, sha384, sha512, sha512]).
+-define(ALG_DEFAULT, sha256). % RFC8945, section 6
```
RaimoNiskanen, line 40: How about
```erlang
-define(ALG_DEFAULT, sha256). % RFC8945, section 6
-define(ALGS_SUPPORTED, [md5, sha, sha224, ?ALG_DEFAULT, sha384, sha512]).
```
(by the way, `sha512` seems to be accidentally doubly enumerated)

lib/kernel/src/inet_dns_tsig.erl
```diff
@@ -0,0 +1,340 @@
...
+-define(FUDGE_DEFAULT, 300). % RFC8945, section 10
+
+-define(MAC_SIZE_MIN, 10). % RFC8945, section 5.2.2.1
+-define(TTL, 0).
+
+-define(NOW, os:system_time(seconds)). %  RFC8945, section 5.2.3
+
+-type tsig_key() :: {string(),iodata()}.
+
+-type tsig_algs() :: crypto:hmac_hash_algorithm() | gss_tsig.
```
RaimoNiskanen, line 51: The `crypto:` type is not exported so this causes a Dialyzer error.
Although maybe the `crypto:` type should be supported, I do not think it should be used here, instead enumerate the supported algorithms.  I think 
``` erlang
-type tsig_algs() :: ?ALGS_SUPPORTED.
```
might be what we want.

Btw, `gss_tsig`; how on earth could that even be supported?  Seems complicated.

lib/kernel/src/inet_dns.erl
```diff
@@ -246,9 +254,38 @@ decode_rr_section(Bin, N, Buffer, RRs) ->
                           z                = Z,
                           data             = D,
                           do               = DnssecOk};
+                   ?S_TSIG ->
+                       % RFC 8945: 5.2. FORMERR if not last
+                       Rest =/= <<>> andalso throw(?DECODE_ERROR),
+                       % RFC 8945: 5.2. FORMERR if more than one dns_rr_tsig
+                       lists:keymember(dns_rr_tsig, 1, RRs) andalso
+                        throw(?DECODE_ERROR),
```
RaimoNiskanen, line 262: Isn't this redundant, since the only possibility for `?S_TSIG` to occur more than once is to not be last?

lib/kernel/src/inet_dns_tsig.erl
```diff
@@ -0,0 +1,340 @@
...
+    fudge     :: 0..65535,
+
+    qr    = 0 :: 0..100, % 0 (false): request, 1+ (true): response
+                         % this value is overloaded to handle TSIG
+                         % over TCP where we need to accept up to
+                         % 99 messages without TSIG so each time a
+                         % response with a TSIG RR is received this
+                         % value is reset to 2
+    id        :: undefined | 0..65535,
+    mac       :: undefined | binary() | crypto:mac_state()
```
RaimoNiskanen, line 68: Since `crypto:mac_state()` is an opaque type it cannot be assumed to not be e.g `undefined`, so when comparing if this field is `undefined` you break the opaque type specification.  It cannot be compared.

Therefore a field with an opaque type has to have a different field (or other way of knowing) that flags if it contains a `crypto:mac_state()` or not.

RaimoNiskanen:
I am not completely done reviewing, and tomorrow we have an all day workshow, so I will not get further until Friday...

lib/kernel/src/inet_dns_tsig.erl
jimdigriz: > The `crypto:` type is not exported so this causes a Dialyzer error. Although maybe the `crypto:` type should be supported, I do not think it should be used here, instead enumerate the supported algorithms. I think
> 
> ```erlang
> -type tsig_algs() :: ?ALGS_SUPPORTED.
> ```
> 
> might be what we want.

Sure.

Resolved in https://github.com/erlang/otp/pull/6985/commits/af2995b2f3c6db178c34a076856c83497569546d
 
> Btw, `gss_tsig`; how on earth could that even be supported? Seems complicated.

I did look into it, if you want to bleed from the eyes, head on over to https://stackoverflow.com/a/55936492

From what I can tell the dnspython people avoided the GSSAPI component by having the end user create it OOB and reference it.

Overall, I suspect no one wants OTP to depend on KRB5 to support DNS packets :)

Cheers

RaimoNiskanen:
> @RaimoNiskanen one thought is as you are using bash, I could amend 53742e6 to wait until nsd has opened the TCP port instead and remove the use of nsd-control?

You lost me.  Is there some exciting bash feature that can be used in an esoteric way?
Nevertheless, I do not like to rely on bash, prefer Posix shell.  Where is this bash usage?

I will try to take a look at Knot DNS...

lib/kernel/src/inet_dns.erl
jimdigriz: Yep.

Resolved by https://github.com/erlang/otp/pull/6985/commits/d3e5974876f4a1f3fb9a356731b10ff806076846

lib/kernel/src/inet_dns_tsig.erl
jimdigriz: What about instead something like?
```
mac :: {?MODULE,undefined | binary()} | {crypto,crypto:mac_state()}
```

Partitions off "this is mine" and "this is yours".

If you prefer a separate element in the record, happy to do that instead?

jimdigriz:
> > @RaimoNiskanen one thought is as you are using bash, I could amend [53742e6](https://github.com/erlang/otp/commit/53742e6052d813b5ac174fc6da81b8a1b9ba088e) to wait until nsd has opened the TCP port instead and remove the use of nsd-control?
> 
> You lost me. Is there some exciting bash feature that can be used in an esoteric way?

> Nevertheless, I do not like to rely on bash, prefer Posix shell.

Agreed.

> Where is this bash usage?

My mistake, I swear I saw it in https://github.com/erlang/otp/blob/master/lib/kernel/test/inet_res_SUITE_data/run-ns

Probably was late at night...sorry.

> I will try to take a look at Knot DNS...

`nsd` is nice, but it would be great to have the end-to-end `UPDATE` tests against something other than my ropey testing code. I could not find something (other than `bind9`) that would do this.

lib/kernel/src/inet_dns_tsig.erl
RaimoNiskanen: That would be just fine, it does not have to be a separate element

lib/kernel/src/inet_dns_tsig.erl
jimdigriz: ...spent too long on this staring at the obvious, Dialyzer does not understand Erlang so fixed the fix.

https://github.com/jimdigriz/otp/commit/c1b83588632c499ba6cc251b68cbbbe7e155371e

lib/kernel/src/inet_dns_tsig.erl
jimdigriz: Resolved by 5c1728c8591a398683d64918ff9127e4c44bada6

RaimoNiskanen:
I just pushed a suggestion that you can use as you like.  It is incomplete and we should not use it as it is.  See the commit comment.

> ```
> exec 3<>/dev/tcp/<host>/<port>
> ```
I'll be darned.  That was almost too much...

Could maybe `telnet` or `netcat` be used for the same purpose?

> ... but it would be great to have the end-to-end UPDATE tests against something other than my ropey testing code. I could not find something (other than bind9) that would do this.

Do you say that KnotDNS cannot be used for UPDATE tests?

I found that Bind9 is almost impossible to use in our test cases because it has system default Apparmour and such definitions, so that it is not possible / very awkward to write and use a Bind9 configuration file from a test case since it does not get placed in in /etc/

I am on vacation the next week, so I will not work on this for a while...

jimdigriz:
> I just pushed a suggestion that you can use as you like. It is incomplete and we should not use it as it is. See the commit comment.
>
> I am on vacation the next week, so I will not work on this for a while...

Gives me time to pick through the commit and update things.

> > ```
> > exec 3<>/dev/tcp/<host>/<port>
> > ```
> 
> I'll be darned. That was almost too much...
>
> Could maybe `telnet` or `netcat` be used for the same purpose?

I suspect at that point you may as well rewrite the `start-ns` script in Erlang and use `gen_tcp`.

As you use `bash` already in OTP for a few things, such as `run-dialyzer` and the SSH test suite, I would probably consider biting the bullet and just use that; or as we are installing `dig` with this PR just poll with `dig` rather than `nsd-control`?

> > ... but it would be great to have the end-to-end UPDATE tests against something other than my ropey testing code. I could not find something (other than bind9) that would do this.
> 
> Do you say that KnotDNS cannot be used for UPDATE tests?

Typo, I meant "other than KnotDNS".

KnotDNS *can* be used for both `UPDATE` and `AXFR` with `TSIG` tests. I used it for my local testing and whilst building the faux unit tests to exercise it all.

Documented over at https://www.knot-dns.cz/docs/3.2/html/configuration.html#dynamic-updates

Have to find something else if you [ever want `SIG(0)` though](https://datatracker.ietf.org/doc/html/rfc2931)...

> I found that Bind9 is almost impossible to use in our test cases because it has system default Apparmour and such definitions, so that it is not possible / very awkward to write and use a Bind9 configuration file from a test case since it does not get placed in in /etc/

...that and BIND9 is completely ghetto.

On a more serious note, if you are fighting AppArmour with BIND9 you will stumble on that elsewhere so you may need to find at some stage a long term solution anyway.

jimdigriz:
> The idea for the start-ns script in the test suite is that it should start the name server and figure out when it should be ready to serve requests, without doing a request. If we poll for successful start by using requests then we cannot tell the difference between an unsuccessful start vs. a failed first request.
>
> Maybe use some name server debug flag that produces a stdout printout indicating server start?

We can, you poll for a DNS lookup you know the answer for, anything that does not match what you expect, you keep trying until some deadline timer is exceeded.

This is no different to polling using some CLI tool such as `nsd-control status` where instead you are just waiting for the exit code to reflect what you want or groking the output of stdout, right?

I do not need the advantage of pinning our hopes here on stdout :)
 
> Regarding installing Bind vs. Nsd for testing; I have had problems with system installed Nsd that installs a start script that starts Nsd and which fails with the default configuration - causing the packet manager to indicate a failure for every future packet update since the system Nsd fails to restart and regards this as a packet dependency problem for packets that Nsd depends on...
>
> So I have no good idea of what a long term solution should be :-(

I do not think you should not be starting any services on install, I would recommend:
```
# systemctl mask nsd
# apt install nsd
```

Just tested it:
```
$ lxc launch images:debian/bullseye nsd-test

$ lxc exec nsd-test -- /bin/bash

root@nsd-test:~# apt update
root@nsd-test:~# systemctl mask nsd
Unit nsd.service does not exist, proceeding anyway.
Created symlink /etc/systemd/system/nsd.service → /dev/null.
root@nsd-test:~# apt install nsd
[snipped]
Selecting previously unselected package nsd.
Preparing to unpack .../archives/nsd_4.3.5-1_amd64.deb ...
Unpacking nsd (4.3.5-1) ...
Setting up libfstrm0:amd64 (0.6.0-1+b1) ...
Setting up libprotobuf-c1:amd64 (1.3.3-1+b2) ...
Setting up libevent-2.1-7:amd64 (2.1.12-stable-1) ...
Setting up openssl (1.1.1n-0+deb11u4) ...
Setting up nsd (4.3.5-1) ...
Failed to preset unit: Unit file /etc/systemd/system/nsd.service is masked.
/usr/bin/deb-systemd-helper: error: systemctl preset failed on nsd.service: No such file or directory
nsd.service is a disabled or a static unit, not starting it.
Processing triggers for libc-bin (2.31-13+deb11u6) ...
root@nsd-test:~# echo $?
0
root@nsd-test:~# 
```

RaimoNiskanen:
> I do not see the advantages of pinning our hopes here on `stdout` :)

The test philosophy here is that we should check the outcome of the start through "the same channel" as the start itself.  If polling with legit requests we can mistake network issues for server configuration problems.

For example I have seen Windows issues where the network stack per default does not route between loopback and a real interface, so the name server may start successfully but the test request does not get routed correctly.

I am working on a solution...


> ```
> # systemctl mask nsd
> ```

Thanks for the pointer.  I try to learn as little as possible about systemd. ;-)

jimdigriz:
> The test philosophy here is that we should check the outcome of the start through "the same channel" as the start itself

`stdout` is not the same channel as `fork()`/`exec()` though.

> I try to learn as little as possible about systemd. ;-)

...says thee who proclaims spaces over tabs?!

Vg jnf gur ynfg terng jne hf terlornqf sbhtug naq ybfg, naq abj gur jbeyq oheaf fheebhaqvat hf jvgu x8f, flfgrzq naq Ryvkve...

RaimoNiskanen:
Would this commit (03807638c73445b853f6b465c0d129c206158801) work for you?

jimdigriz:
> Would this commit ([0380763](https://github.com/erlang/otp/commit/03807638c73445b853f6b465c0d129c206158801)) work for you?

What I do not like is I did not think of using `SIGPIPE` originally!

Looks good to me though and tested. Thanks

RaimoNiskanen:
> What I do not like is I did not think of using SIGPIPE originally!

Sry.  What about `SIGPIPE`?

jimdigriz:
> > What I do not like is I did not think of using SIGPIPE originally!
> 
> Sry. What about `SIGPIPE`?

`sed '/.../ q'` closes `STDOUT` for `tail` which receives a `SIGPIPE` on `write()` causing it to die...

Wish I had thought of this strategy before a burnt an hour or two getting `nsd-control` to work.

RaimoNiskanen:
Ah.  Now I see.  I was under the misconception that that the pipeline finishes when the last command `sed ... q`, exits, but now when I read the documentation all commands on the pipeline must finish, which as you say will happen due to `SIGPIPE` for `tail`...  Thank you for the enlightenment.

I will try to rewrite for KnotDNS, see how that works out.

RaimoNiskanen:
I tried a rewrite for KnotDNS.  See if it fits.  If so you can add tests for UPDATE.

Then I think I am waiting for you completing something I mentioned in 90e5b4f3f076bf60c5ba07097d7276b1daad8e82 ...

jimdigriz:
Yep, all on me now!

jimdigriz:
> Then I think I am waiting for you completing something I mentioned in https://github.com/erlang/otp/commit/90e5b4f3f076bf60c5ba07097d7276b1daad8e82 ...
>
> * Single % should be used at the end of a line
> * Double %% should be used at the code's indentation level

Think I picked up a few you missed, but stayed "non-compliant" for `inet_dns.hrl` to fit in with the existing style used there.

> * The type tsig_state() that is #tsig_state{} very often needs to have
>   all fields default values to follow the specified type for the field,
>   or everywhere the record is constructed Dialyzer may figure out that
>   the constructed record breaks the type spec.

[Yes I know](https://erlang.org/pipermail/erlang-questions/2018-December/096819.html), but as Dialyzer (for me, see below) does not run on `lib/kernel` this is going to have to be something the OTP team are going to have to fix for me.

> * I messed with the sign/2,3 functions to parse the binary
>   in steps through helper function sign/5 to avoid later using
>   binary:part/2 to dissect the message again.  Use the code style
>   if you like it.  I have not touched verify/3.

Why the hate on `binary:part`?

I thinks it made things easier to follow rather than dumping on the reader scaffolding that exists only for the sake of allowing the binary syntax to drive it.

Performance makes no sense either as we are about to dispatch into calculating a SHA256 HMAC.

That said, I really am not that fussed.

Not sure what you want me to do with `verify/3`? Annihilate `binary:part` here too?

> * There is still a problem on line 209 where the call to macN/2
>   has a first argument that is {Module, MacState} instead of the
>   expected MacState.  I have not figured out how to solve this.

I'm confused, it *is* meant to be `{?MODULE,binary()}` as it carries the resulting calculated HMAC from the last packet.

If this was wrong, signed AXFR transfers would not work.

Maybe I am missing something, do you have an example that triggers unexpected behaviour?

> * I think encode_algname/1 and decode_algname/1 should be in inet_dns.erl
>   where encoding and decoding belongs.

Okay, moved.

> * Moving those functions will probably also solve a bunch of Dialyzer
>   complaints about them handling unused values.

I have no idea why Dialyzer does not work for me on the OTP tree, maybe I'm just too stupid and skipped something important in `HOWTO/DEVELOPMENT.md`:
```
alex@sarasti:~/src/otp$ make distclean
alex@sarasti:~/src/otp$ git clean -xf
alex@sarasti:~/src/otp$ export ERL_TOP=`pwd`
alex@sarasti:~/src/otp$ ./configure
alex@sarasti:~/src/otp$ make -j17
alex@sarasti:~/src/otp$ cd lib/kernel
alex@sarasti:~/src/otp/lib/kernel$ make dialyzer
Building kernel plt file
  Creating PLT priv/plt/kernel.plt ... done in 0m19.49s
done (passed successfully)
Running dialyzer on kernel
  Checking whether the PLT priv/plt/kernel.plt is up-to-date... yes
  Proceeding with analysis...
dialyzer: Analysis failed with error:
Could not scan the following file(s):
  Could not get Core Erlang code for: /home/alex/src/otp/lib/kernel/../kernel/ebin/pg.beam
  Could not get Core Erlang code for: /home/alex/src/otp/lib/kernel/../kernel/ebin/global_group.beam
  Could not get Core Erlang code for: /home/alex/src/otp/lib/kernel/../kernel/ebin/erts_debug.beam

Last messages in the log cache:
  Reading files and computing callgraph... 
make: *** [/home/alex/src/otp/make/app_targets.mk:63: dialyzer] Error 1
```

Until this works, I cannot fix anything related to Dialyzer grumblings.

> * In inet_dns, creating a Data 6-tuple on line 384 and dissecting it
>   on line 876 seems unnecessary.  Why not simply pass the #dns_rr_tsig{}
>   record as Data?

Someone may want to construct their own TSIG and encode it for example if they want to be crazy implement GSS application side.

jimdigriz:
I see the Dialyzer complaint you are referring to in https://github.com/jimdigriz/otp/actions/runs/5490334109/jobs/10005986729#step:10:79 but Dialyzer I think it is unable to track that `TSQR == 3` is tied to `mac = {?MODULE,binary()}` so we should neuter Dialyzer here.

RaimoNiskanen:
> Think I picked up a few you missed, but stayed "non-compliant" for inet_dns.hrl to fit in with the existing style used there.

Great!

> ... but as Dialyzer (for me, see below) does not run on lib/kernel this is going to have to be something the OTP team are going to have to fix for me.

I always run `make dialyzer` from the source top level.  But then I also build with `./otp_build setup -a`.  And I often start with running `scripts/otp_build_check` which among other things checks Dialyzer.  (remember to set `MAKEFLAGS=-j10`)

I tried `scripts/otp_build_check` just now and there were no Dialyzer errors.  Was it e85ec8ce121ce419af0932b670194b2ed4b32939 that fixed the issue?  Was it even a bug?

> Someone may want to construct their own TSIG and encode it for example if they want to be crazy and implement GSS application side.

Right. You want an API boundary here.  Fair enough...

> Why the hate on binary:part?
> 
> I thinks it made things easier to follow rather than dumping on the reader scaffolding that exists only for the sake of allowing the binary syntax to drive it.
> 
> Performance makes no sense either as we are about to dispatch into calculating a SHA256 HMAC.

I do not hate `binary:part/2`, but I do like the bit syntax, which might be a personal preference...

Also, in my eyes the code becomes easier to understand when the packet is disassembled into named parts and later the result is composed from the named parts, vs. when the result is composed from unnamed pieces pulled from the incoming packet.

You get a better view of the packet layout when using a bit syntax expression, I think.

In this case:
``` erlang
    <<Hdr:10/binary, ARC:16, Content/binary>> = Pkt,
    TSigRRBin = foo(),
    {ok, <<Hdr/binary, (ARC + 1):16, Content/binary, TSigRRBin/binary>>}.
```
vs.
``` erlang
    TSigRRBin = foo(),
    {ok,
      iolist_to_binary([
         binary:part(Pkt, {0,10}),
         begin <<ARC:16>> = binary:part(Pkt, {10,2}), <<(ARC + 1):16>> end,
         binary:part(Pkt, {12, byte_size(Pkt - 12)}),
         TSigRRBin])}.
```

jimdigriz:
> I tried `scripts/otp_build_check` just now and there were no Dialyzer errors. Was it [e85ec8c](https://github.com/erlang/otp/commit/e85ec8ce121ce419af0932b670194b2ed4b32939) that fixed the issue?

Yes, but also https://github.com/erlang/otp/pull/6985/commits/0c1bb01d8d3b80809fa3b04f7401ac4a6bff045c too.

> Was it even a bug?

It was a bug, though I am surprised the effect was to not break the multi-response TCP verification.

Probably would be happier if I could trigger this edge case. My fault for leaving it for a while, the state machine fell out my head a while back whilst I have been battling with something else far less interesting... :-/

I suspect I need to understand why this works without the tweak, I suspect the code path is triggered for the legacy "up to 99 unsigned packets" but need to double check that. This should not hold up you telling me what else I need to beat into shape though!
 
> > Why the hate on binary:part?
> > I thinks it made things easier to follow rather than dumping on the reader scaffolding that exists only for the sake of allowing the binary syntax to drive it.
> > Performance makes no sense either as we are about to dispatch into calculating a SHA256 HMAC.
> 
> I do not hate `binary:part/2`, but I do like the bit syntax, which might be a personal preference...
> 
> Also, in my eyes the code becomes easier to understand when the packet is disassembled into named parts and later the result is composed from the named parts, vs. when the result is composed from unnamed pieces pulled from the incoming packet.

Swings and roundabouts.

I do prefer bit syntax too but I think here the result is it looks nicer in one area at the cost of making the other function scruffier (forces me to store more state in my head about what is what variable wise).

For me the naming does not really help as we cross the field boundaries of the DNS packet so for me rolling it all up together at the final step is more explicitly stating "right I need the bytes from here, these bytes are incremented, and then slap on the remaining bytes and append some other bytes".

Though, the whole process that needs to be performed here is a bit messy so the scruffiness I think probably is unavoidable :)

RaimoNiskanen:
I pushed two commits, both to be squashed into where `inet_dns_tsig` was added.
* The new module has to be listed in the `kernel.app` file
* If `dig` cannot be found the test case `tsig_server` should be skipped, not fail.
* I added `drill` as an alternative to `dig`.

`inet_res_SUITE:tsig_server/1` failed on all platforms since, apparently none of them had `dig` installed.  I have to talk to our integration team, since it also seems like no platform has got `knotd` installed...  So the test coverage is zero, at the moment :-(

RaimoNiskanen:
I think I found the reason for the timeout failures on FreeBSD.  `drill` there did not like the argument order and thought it was a network error, so the connection to the test case AXFR server never succeeded.  I changed the argument order to something I hope both `drill` and `dig` understands.

jimdigriz:
> I suspect I need to understand why this works without the tweak, I suspect the code path is triggered for the legacy "up to 99 unsigned packets" but need to double check that. This should not hold up you telling me what else I need to beat into shape though!

Got my morning coffee(s) in and checked, yes, is for the legacy support and covers when there is no `#dns_rr_tsig{}` in the AR list (the `case` statement returns `false`) and we have already had a response (`TSQR == 3`), this is why it did not break the existing verify tests as both `nsd` and `knotd` send a TSIG with every response packet.

Feeling comfortable about this now.

Let me know if you need me to do anything or there is something outstanding I have missed.

RaimoNiskanen:
Since the FreeBSD timeout error took a little debugging, I restructured the test to create more explicit crashes if anything goes wrong...

Other than that I would like this to run cleanly in our daily builds, I'll let you know.
Then you can clean up the branch and squash into a few well chosen commits.

RaimoNiskanen:
There seems to be a problem in the knot.conf file (HERE document) in the run-ns script in that the `database:` section is not supported until Knot 2.9 and we have machines in our lab (Ubuntu 20 and earlier) that have Knot 2.6.

So `inet_res_SUITE:basic` fails on those, or should fail but often silently skip.

O.T.O.H. unless we can set a `database:` section the `update` test does not work, so it seems we need Knot 2.9 or later.  The right solution might then be to make this a hard requirement and uninstall all old Knot:s.

jimdigriz:
I would agree to just skip the tests where Knot 2.9 or later is unavailable.

For me the aim of the test here is to make sure we are generating and able to send valid requests and make suitable responses, as long a at least one test machine has this covered I cannot see testing it on other systems provides any value.

I did take a stab and getting it to work without `database` (using a `default` template and `storage`) but for some reason `knotd` is determined to continue to try and use `/var/lib/knot`...I'll stop for now as you seem okay with just skipping this.

Thanks

RaimoNiskanen:
> I did take a stab and getting it to work without `database` (using a `default` template and ...

Thanks for trying!

I wrote a test for Knot version, and skip if no Knot or too old.  That will do.

We have a general problem in our daily tests presentation and that is that no red lights show up if a test is skipped everywhere, and that we only see failed test cases in the overview.  Therefore we cannot see if a test case has run "anywhere" without manually wrestling the viewer into showing the results for one particular test case.  For that to happen you will have to know that you should look at a particular test case...

This is because it is hard to figure out what "everywhere" should be today...
So a test case may silently creep out into oblivion.

In this case, with time, the number of platforms that have a late enough Knot should increase, so ignoring this problem will do.

We just have to remember to install Knot and Drill on our daily build machines.  Now it is in the Dockerfile, today with an Ubuntu version that is too old for a working Knot.  But, as said, that will improve with time.

jimdigriz:
My project actively uses DNS TSIG so I'll come a knocking if it breaks :)

RaimoNiskanen:
The test cases works now, skipped on most of our daily build machines.  It is only Ubuntu 22 and recent BSD:s that have a modern enough `knotd`, and we do not have many such machines in the daily build.

But, @jimdigriz;  go ahead and squash the commits into something decent.

jimdigriz:
> This time it turned out that it was not reliable to use SIGPIPE to stop "tail -f" when detecting name server start, probably due to different output buffering and timing on different OS:es.

Shame `stdbuf` is Linux only.

Though you have found something that works, I wonder if `sed -u` would have helped; it is described also in the BSD [manpage for `sed(1)`](https://man.openbsd.org/sed) too.

Striking that up to the fuzziness of UNIX-esque implementations :)

RaimoNiskanen:
> I wonder if `sed -u` would have helped;

Interesting, but sounds a bit fuzzy.  "load minimal amounts of data" may be «right», but there still are no guarantees about how `tail -f` buffers output, or how the pipe implementation itself buffers.  The symptoms were that `tail -f` had written past the server start line and was waiting for more input.  Then `sed` saw the server start line and exited.  But `tail -f` did not notice the closed output until it tried to write again, which it did when more input arrived (in the test case no more input arrived so server start failed with the timer process killing it).

So maybe even there is no SIGPIPE  signal sent on the BSD:s, but instead the closed pipe is noticed on write.

Anyway, this solution should be «right» since when `sed` exits, `tail -f` gets explicitly killed, provided that the correct PID of `tail -f` has been found, which was a bigger problem than I think it should be.

> Striking that up to the fuzziness of UNIX-esque implementations :)

Writing portable shell scripts is truly a pain :(

RaimoNiskanen:
I removed the 'testing' label, but the branch is still part of our daily builds.  I pull it in and add it internally...

jimdigriz:
I managed to clear the backlog of my other work today, hence the rebasing the slipped out, and hope to have the squashing sorted by today. sorry it took a while to get back to this.

RaimoNiskanen:
No worries :-)

RaimoNiskanen:
For the sake of the procedures I will let it run a few nights in its squashed variant before merging.  I also think it is ready to merge.

RaimoNiskanen:
Thank you for your pull request and hard work!

jimdigriz:
Thanks for the guidance on beating this PR into shape.

See you when I need to implement `SIG(0)` :)

RaimoNiskanen:
I will read up on that when it falls on my table... ;-)

PR #6988, opened 2023-03-07
Document the bit order of the bit syntax

bjorng:


system/doc/reference_manual/expressions.xml
```diff
@@ -1348,6 +1348,27 @@ handle_call(change, From, #{ state := start } = S) ->
   <section>
     <marker id="bit_syntax"></marker>
     <title>Bit Syntax Expressions</title>
+    <p>The bit syntax operates on <em>bit strings</em>.</p>
+    <p>
+      A <em>bit string</em> is a sequence of bits stored in
+      a sequence of (8-bit) bytes in memory address order.
+      The bits are stored in big-endian format in the byte sequence
+      so the first bit is the higest bit of the first byte.
```
sverker, line 1356: I don't think "big-endian" is normally used to refer to bit order.
Suggestion:
_A **bit string** is a sequence of bits. The bits are stored as
a sequence of (8-bit) bytes in memory address order.
Within each byte the bits are stored from highest to lowest bit. That is,
the first bit of a bit string is the highest bit of the first byte._

system/doc/reference_manual/expressions.xml
```diff
@@ -1348,6 +1348,27 @@ handle_call(change, From, #{ state := start } = S) ->
...
+      If the length of the <em>bit string</em> is not a multiple
+      of 8 (bits) then the lowest bits of the last byte are unused.
+      A <em>bit string</em> with a length that is a multiple of 8 (bits)
```
sverker, line 1359: I think it's better with "8 bits" without the parentheses.

system/doc/reference_manual/expressions.xml
```diff
@@ -1348,6 +1348,27 @@ handle_call(change, From, #{ state := start } = S) ->
...
+      is also known as a <em>binary</em>.
+    </p>
+    <p>
+      A <em>bit string</em> can be divided into <em>segments</em>
+      that from left to right are allocated onto the bit sequence
+      in the order above, that is; from high to low bits in each byte,
```
bjorng, line 1365:
```suggestion
      from high to low bits in each byte,
```

system/doc/reference_manual/expressions.xml
```diff
@@ -1348,6 +1348,27 @@ handle_call(change, From, #{ state := start } = S) ->
...
+      with bytes in memory address order.
+      From this follows that a <em>segment</em> that crosses a byte border
+      does that from the low bit in a byte to the high bit
+      in the subsequent byte.
+    </p>
```
bjorng, line 1370: I think there should be one or more examples here.

system/doc/reference_manual/expressions.xml
```diff
@@ -1348,6 +1348,27 @@ handle_call(change, From, #{ state := start } = S) ->
   <section>
     <marker id="bit_syntax"></marker>
     <title>Bit Syntax Expressions</title>
+    <p>The bit syntax operates on <em>bit strings</em>.</p>
+    <p>
+      A <em>bit string</em> is a sequence of bits stored in
+      a sequence of (8-bit) bytes in memory address order.
```
bjorng, line 1354: A young reader may ask what other kind of bytes there are.

```suggestion
      a sequence of bytes (octets) in memory address order.
```

system/doc/reference_manual/expressions.xml
frej: Isn't the usual term "most/least significant" bit?

RaimoNiskanen: Will do

system/doc/reference_manual/expressions.xml
RaimoNiskanen: Will do

system/doc/reference_manual/expressions.xml
RaimoNiskanen: Will fix with a new rewrite

system/doc/reference_manual/expressions.xml
RaimoNiskanen: Will do

system/doc/reference_manual/expressions.xml
RaimoNiskanen: Agreed

RaimoNiskanen:
I have pushed a new attempt that hopefully addressed all comments so far

RaimoNiskanen:
Now the documentation of this PR is ready and updated: [Bit Syntax Expressions](https://erlang.github.io/prs/6988/doc/reference_manual/expressions.html#bit-syntax-expressions)

PR #5292, opened 2021-10-18
Add predefined types to erlang module

garazdawi:
This PR adds all predefined types to the erlang module and links to them from the documentation.

In addition to making the HTML documentation easier to understand, this also allows IDEs that use the EEP-48 style documentation (such as erlang_ls/elixir_ls) to include better IntelliSense documentation for predefined types.

You can find the new documentation [here](https://erlang.org/~lukas/predefined-types//erts-12.1.2/doc/html/erlang.html#Predefined%20datatypes) and links to it [here (see the Packet type)](https://erlang.org/~lukas/predefined-types/lib/kernel-8.1.2/doc/html/gen_tcp.html#send-2) for example.

There are still a couple of rough corners, but I'm opening this PR to get some initial feedback.

PR #6993, opened 2023-03-08
EEP 61: Built-in dynamic type

ilya-klyuchnikov:
A reference implementation for EEP 61 "Built-in dynamic type".

### DIFF
--- a/erts/preloaded/src/erlang.erl
+++ b/erts/preloaded/src/erlang.erl
@@ -86,6 +86,7 @@
 -type boolean() :: true | false.
 -type byte() :: 0..255.
 -type char() :: 0..16#10FFFF.
+-type dynamic() :: dynamic().
 -type float() :: float().
 -type function() :: fun().
 -type identifier() :: pid() | port() | reference().
@@ -123,7 +124,7 @@
 -type timeout() :: 'infinity' | non_neg_integer().
 -type tuple() :: tuple().
 -export_type([any/0, arity/0, atom/0, binary/0, bitstring/0, bool/0, boolean/0, byte/0,
-              char/0, float/0, function/0, identifier/0, integer/0, iodata/0, iolist/0,
+              char/0, dynamic/0, float/0, function/0, identifier/0, integer/0, iodata/0, iolist/0,
               list/0, list/1, map/0, maybe_improper_list/0, maybe_improper_list/2, mfa/0,
               module/0, neg_integer/0, nil/0, no_return/0, node/0, non_neg_integer/0,
               none/0, nonempty_binary/0, nonempty_bitstring/0, nonempty_improper_list/2,
--- a/lib/dialyzer/src/erl_types.erl
+++ b/lib/dialyzer/src/erl_types.erl
@@ -4394,6 +4394,8 @@ from_form({type, _Anno, byte, []}, _S, _D, L, C) ->
   {t_byte(), L, C};
 from_form({type, _Anno, char, []}, _S, _D, L, C) ->
   {t_char(), L, C};
+from_form({type, _Anno, dynamic, []}, _S, _D, L, C) ->
+  {t_any(), L, C};
 from_form({type, _Anno, float, []}, _S, _D, L, C) ->
   {t_float(), L, C};
 from_form({type, _Anno, function, []}, _S, _D, L, C) ->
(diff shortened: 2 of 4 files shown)


jhogberg:
EEP 61: https://github.com/erlang/eep/pull/44

system/doc/reference_manual/typespec.xml
```diff
@@ -83,17 +83,30 @@
...
+    <p>
+      To facilitate <url href="https://en.wikipedia.org/wiki/Gradual_typing">
+      gradual typing</url> of Erlang, the type `dynamic()` is provided.
+      It is similar to
+      <url href="https://docs.python.org/3/library/typing.html#the-any-type">Any</url>
+      in Python,
+      <url href="https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#any">any</url>
+      in TypeScript and
+      <url href="https://docs.hhvm.com/hack/built-in-types/dynamic">dynamic</url>
+      in Hack.
```
josevalim, line 101: Should a note be added that Dialyzer "treats dynamic the same as any/term" (there is probably a better way to phrase this) but other type systems running on the BEAM may rely on its distinctive meaning?

ilya-klyuchnikov: Thanks for pointing this out. - Yes, it is worth to clarify it explicitly.

ilya-klyuchnikov: Clarified in 79697d086ffd6972453cb1c17920280e5a2616b1

lib/dialyzer/src/erl_types.erl
```diff
@@ -4394,6 +4394,8 @@ from_form({type, _Anno, byte, []}, _S, _D, L, C) ->
   {t_byte(), L, C};
 from_form({type, _Anno, char, []}, _S, _D, L, C) ->
   {t_char(), L, C};
+from_form({type, _Anno, dynamic, []}, _S, _D, L, C) ->
+	{t_any(), L, C};
```
kikofernandez, line 4398: there is a tab here

erts/preloaded/src/erlang.erl
```diff
@@ -86,6 +86,7 @@
 -type boolean() :: true | false.
 -type byte() :: 0..255.
 -type char() :: 0..16#10FFFF.
+-type dynamic() :: dynamic().
```
kikofernandez, line 89: Is there any reason why `dynamic()` is not defined as another type alias of `any()`, in the same fashion as `term()`?

**Reasoning**

- Assume that `dynamic()` is not a type alias for `term()` or `any()
- Assume that the standard library replaces `term()` by `dynamic()`, which I think it is not unreasonable (under eqWAlizer and Dialyzer all works)
- Assume there is another static analysis tool. Since `dynamic()` is not a subtype of `any()`, we are breaking backwards compatibility.

Is my reasoning sound?

kikofernandez: This other tool could be (e.g.,) Gradualizer (https://github.com/josefs/Gradualizer), Etylizer (AFAIK https://arxiv.org/abs/2302.12783), or maybe even the bidirectional typing for Erlang (https://dl.acm.org/doi/10.1145/3471871.3472966)

ilya-klyuchnikov: `erlang:dynamic()` is an alias to the builtin type `dynamic()`. Similar to how `erlang:atom()` is an alias to the builtin type `atom()`.

The introduction of the builtin type `dynamic()` is backward-compatible in OTP 26 - thanks to https://github.com/erlang/otp/pull/6335.

ilya-klyuchnikov: Note that line 89 doesn't define `dynamic()`, - it "documents" it - https://github.com/erlang/otp/pull/5292.

In light of https://github.com/erlang/otp/pull/6335 - `erlang:dynamic()` and `dynamic()` should be the same.

jhogberg:
Can you squash these commits? I'd like to merge this before the window for RC2 closes :-)

There's still some activity on the EEP, but there doesn't seem to be a need to hold off on merging this since it's backwards-compatible. The longer people can play with it the better, and we can always fix/revert it in RC3 if something goes wrong.

ilya-klyuchnikov:
squashed and rebased on top of fresh master

jhogberg:
Sweet, thanks, I'll merge it on Friday if all goes well. :-)

jhogberg:
Merged, thanks for the PR!

PR #6998, opened 2023-03-08
Update zlib to 1.2.13

mkuratczyk:
Official changelog:
```
Version 1.2.13 has these key updates from 1.2.12:

Fix a bug when getting a gzip header extra field with inflateGetHeader(). This remedies [CVE-2022-37434](https://nvd.nist.gov/vuln/detail/CVE-2022-37434).
Fix a bug in block type selection when Z_FIXED used. Now the smallest block type is selected, for better compression.
Fix a configure issue that discarded the provided CC definition.
Correct incorrect inputs provided to the CRC functions. This mitigates a bug in Java.
Repair prototypes and exporting of the new CRC functions.
Fix inflateBack to detect invalid input with distances too far.

Due to the first bug fix, any installations of 1.2.12 or earlier should be replaced with 1.2.13.
```

I believe `inflateGetHeader` is not used anywhere in OTP, so this is just a version bump. Formatting changes in the diff are actual differences between 1.2.12 and 1.2.13, not something I introduced.

### DIFF
--- a/HOWTO/INSTALL-RASPBERRYPI3.md
+++ b/HOWTO/INSTALL-RASPBERRYPI3.md
@@ -257,9 +257,9 @@ Uncheck option:
 
   (20)
 
-    $ wget http://zlib.net/zlib-1.2.12.tar.gz
-    $ tar xf zlib-1.2.12.tar.gz
-    $ pushd zlib-1.2.12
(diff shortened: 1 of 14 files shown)


bjorng:

bjorng:
Thanks for your pull request.

PR #7000, opened 2023-03-09
Allow larger column sizes than 8001 in case DB supports it

ignatenkobrain:
In PostgreSQL, the TEXT field can easily have more than 8001 characters and it does return the actual number. The size == 0 would happen in case of some DBs won't return valid value where we would keep the same behavior as of now.

### DIFF
--- a/lib/odbc/c_src/odbcserver.c
+++ b/lib/odbc/c_src/odbcserver.c
@@ -1328,7 +1328,7 @@ static db_result_msg encode_column_name_list(SQLSMALLINT num_of_columns,
 				       &nullable)))
 	    DO_EXIT(EXIT_DESC);
 
-	if(sql_type == SQL_LONGVARCHAR || sql_type == SQL_LONGVARBINARY || sql_type == SQL_WLONGVARCHAR)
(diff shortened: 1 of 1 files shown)


IngelaAndin:
Seems reasonable, I have enabled it in our tests to see that it does not break anything.  I guess you have tested that it works?
We do have Postgres as backend for Linux odbc tests but I see upper limit test of text are skipped due to they being considered to take up to much resources (maybe that is not absolutely true anymore, this code is old).  Do you possible know if there  exist some other easy way to test  what the limit is ?

ignatenkobrain:
We are using this patch in production for like 5 years :) sadly no, I'm not aware of simple way how to test it.

IngelaAndin:
As it does not break anything and is not prioritized by Ericsson, we have decided to see this as an improvement worth  including even if the test are not improved.

gionea:
This change has introduced a problem where the ODBC port crashes (`port_exit, memory_allocation_failed`) when used in conjunction with the Oracle driver.

I have been able to confirm this change as being the culprit by applying a patch that reverts the change from the 26.1 release.

I can open a separate issue for this problem, if you prefer, but it would be helpful if you could give me some pointers to how to collect more information about the crash. For example, it would be useful to know what SQL query it was executing while it crashed, etc.

Issue #6942, opened 2023-02-28
dialyzer regression in 26: nowarn_function being ignored


tsloughter:
**Describe the bug**

I have a use of `-dialyzer({nowarn_function` that on OTP-26 appears to be ignored:

https://github.com/open-telemetry/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_resource_detector.erl#L198-L200

On 25 I get no warning but on 26 I now see:

```
apps/opentelemetry/src/otel_resource_detector.erl
Line 200 Column 9: Unknown function release_handler:which_releases/1
```

**To Reproduce**

Can be reproduced by running `rebar3 dialyzer` on https://github.com/open-telemetry/opentelemetry-erlang

**Expected behavior**

Should be no warning for the function `find_release`

**Affected versions**

26.0-rc1

PR #7001, opened 2023-03-09
dialyzer: do not report warnings for functions marked as `nowarn_function`

kikofernandez:
skip warnings about functions that have inlined nowarn_function flags, e.g.,

```erlang
-dialyzer({nowarn_function, find_release/0}).
find_release() ->
  non_existing_function:foo().
```

closes GH-6942

### DIFF
--- a/lib/dialyzer/src/dialyzer_analysis_callgraph.erl
+++ b/lib/dialyzer/src/dialyzer_analysis_callgraph.erl
@@ -117,25 +117,16 @@ loop(#server_state{parent = Parent} = State,
 analysis_start(Parent, Analysis, LegalWarnings) ->
   CServer = dialyzer_codeserver:new(),
   Plt = Analysis#analysis.plt,
-  State = #analysis_state{codeserver = CServer,
-			  analysis_type = Analysis#analysis.type,
-			  defines = Analysis#analysis.defines,
-			  doc_plt = Analysis#analysis.doc_plt,
-			  include_dirs = Analysis#analysis.include_dirs,
-			  plt = Plt,
-			  parent = Parent,
-                          legal_warnings = LegalWarnings,
-			  start_from = Analysis#analysis.start_from,
-			  use_contracts = Analysis#analysis.use_contracts,
-			  timing_server = Analysis#analysis.timing_server,
-                          solvers = Analysis#analysis.solvers
-			 },
+  Args = {Plt, Analysis, Parent},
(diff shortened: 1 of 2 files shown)


lib/dialyzer/src/dialyzer_cl.erl
```diff
@@ -634,14 +634,15 @@ cl_error(State, Msg) ->
   maybe_close_output_file(State),
   throw({dialyzer_error, lists:flatten(Msg)}).
 
-return_value(State = #cl_state{code_server = CodeServer,
```
kikofernandez, line 637: this is simply the removal of tabs

lib/dialyzer/src/dialyzer_analysis_callgraph.erl
```diff
@@ -117,25 +117,16 @@ loop(#server_state{parent = Parent} = State,
 analysis_start(Parent, Analysis, LegalWarnings) ->
   CServer = dialyzer_codeserver:new(),
   Plt = Analysis#analysis.plt,
-  State = #analysis_state{codeserver = CServer,
```
kikofernandez, line 120: this is the removal of tabs and replace of this record creation into its own function

lib/dialyzer/src/dialyzer_analysis_callgraph.erl
```diff
@@ -160,6 +151,21 @@ analysis_start(Parent, Analysis, LegalWarnings) ->
   Plt4 = dialyzer_plt:delete_list(Plt3, NonExportsList),
   send_analysis_done(Parent, Plt4, DocPlt).
 
+create_analysis_state({Plt, Analysis, Parent}, LegalWarnings, CServer) ->
+  #analysis_state{ codeserver = CServer
+                 , analysis_type = Analysis#analysis.type
```
bjorng, line 156: In my opinion, this comma style looks ugly. Perhaps more to the point, the rest of Dialyzer does not use this comma style, so it should not be used here.

bjorng (commented):
If you have fixed a bug, shouldn't there be a new test case?

When we change code, we have a general rule to try follow the existing code style in the surrounding code, so I don't think the comma style should be changed.

kikofernandez:
I have added tests to:
- test that if we do not use the `nowarn_function`, dialyzer detects unknown functions
- test that if we add the flag `nowarn_function`, dialyzer ignores errors about unknown functions

PR #7003, opened 2023-03-09
sys_core_fold: Inline maps:get/3

jhogberg:
There's little reason not to do this since we're already inlining `maps:get/2`.

### DIFF
--- a/lib/compiler/src/sys_core_fold.erl
+++ b/lib/compiler/src/sys_core_fold.erl
@@ -800,8 +800,6 @@ fold_apply(Apply, _, _) -> Apply.
 call(#c_call{args=As0}=Call0, #c_literal{val=M}=M0, #c_literal{val=N}=N0, Sub) ->
     As1 = expr_list(As0, value, Sub),
     case simplify_call(Call0, M, N, As1) of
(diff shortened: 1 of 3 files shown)


lib/compiler/src/sys_core_fold.erl
```diff
@@ -821,6 +823,31 @@ call(#c_call{args=As0}=Call, M, N, Sub) ->
 %% slightly at the cost of making tracing and stack traces incorrect.
 simplify_call(Call, maps, get, [Key, Map]) ->
     rewrite_call(Call, erlang, map_get, [Key, Map]);
+simplify_call(#c_call{anno=Anno0}, maps, get, [Key, Map, Default]) ->
+    Anno = [compiler_generated] ++ Anno0,
```
bjorng, line 827:
```suggestion
    Anno = [compiler_generated|Anno0],
```

bjorng (approved):

bjorng:
The test case `maps_SUITE:t_get/3` (stdlib) fails.

jhogberg:
Thanks, I've fixed this in a local branch that I'll merge later today once all the test results come in.

PR #7006, opened 2023-03-10
Option 'local' of term_to_binary/2 and term_to_iovec/2

rickard-green:
Introduce a local external term format with an unspecified encoding. This external term format is used for supporting the [`local`](https://erlang.github.io/prs/7006/erts-13.2/doc/html/erlang.html#term_to_binary_local) option of [`term_to_binary/2`](https://erlang.github.io/prs/7006/erts-13.2/doc/html/erlang.html#term_to_binary-2) and [`term_to_iovec/2`](https://erlang.github.io/prs/7006/erts-13.2/doc/html/erlang.html#term_to_iovec-2). Terms encoded on this format is only to be decoded by the same runtime system instance that encoded it.

Also adds a new [`LOCAL_EXT`](https://erlang.github.io/prs/7006/erts-13.2/doc/html/erl_ext_dist.html#local_ext) tag in the external term format.


erts/doc/src/erl_ext_dist.xml
```diff
@@ -1421,6 +1421,31 @@
...
+        decoded by the one that encoded it. The bytes following from here on
+        may contain any unspecified type of encoding of terms. For more
+        information see the
+        <seeerl marker="erts:erlang#term_to_binary_local"><c>local</c></seeerl>
+        option of 
+        <seemfa marker="erts:erlang#term_to_binary/2"><c>term_to_binary/2</c></seemfa>.
+      </p>
+      <p>
+        <c>LOCAL_EXT</c> was introduced in OTP @OTP-18477@.
+      </p>
```
michalmuskala, line 1446: Would it make sense to explicitly clarify that after a restart the term won't be decodable? Or to otherwise specify what it meant by the same system that encoded it

erts/emulator/beam/external.c
```diff
@@ -47,6 +47,12 @@
 #include "erl_proc_sig_queue.h"
 #include "erl_trace.h"
 #include "erl_global_literals.h"
+#include "erl_term_hashing.h"
+
+#if defined(__WIN32__) && !defined(WINDOWS_H_INCLUDES_WINSOCK2_H)
+#include <winsock2.h>
+#endif
```
rickard-green, line 54: For `gethostname()`. Will be moved in to sys, but will live here temporarily

erts/doc/src/erl_ext_dist.xml
rickard-green: It is used that way by `term_to_binary()`, but it is not necessarily only used by the Erlang runtime system or any other OTP functionality which is why I was a bit fuzzy. It can perhaps be better explained, though.

erts/emulator/beam/external.c
```diff
@@ -126,6 +131,70 @@ void erts_init_external(void) {
...
+
+    /*
+     * Use least significant 32 bits of monotonic time as initial
+     * value to hash...
+     */
+    erts_block_hash_init(&hstate, &lnid[0], lnid_ix,
+                         (Uint32) (mtime & 0xffffffff));
+    chunk_size = ERTS_UINT_MAX;
+    res = erts_block_hash(&local_node_hash, &chunk_size, &hstate);
+    ASSERT(res); (void) res;
```
rickard-green, line 194: Hmm, forgot to free the temporary allocated lnid buffer...

okeuday:
@rickard-green The intent of `LOCAL_EXT` looks like ERTS use only (no length is specified and the encoder is not identified in the format).  That would mean `LOCAL_EXT` should not be used in external programming languages (not using BEAM), right?  It might help to make that more explicit in the documentation.

rickard-green:
@okeuday No, it can be used by others as well. Nothing after it is defined since it is up to the user of it to define the format of what follows. Current documentation can be clearer, though. I will modify it some time next week.

mikpe:
> Terms encoded on this format is only to be decoded by the same runtime system instance that encoded it.

I struggle to see how this would be a useful API:
- I can't pass these binaries between nodes and expect to be able to decode them
- I can't store these binaries externally and expect to be able to read and decode them

I can see something like this being used for purely ERTS-internal term_to_binary like conversions (like ETS and sends) but what's the motivation for making it a programmer-visible API?

rickard-green:
@mikpe 😁 I was in a hurry to get this into our builds before the weekend, so I forgot to write the motivation.

If you want to pass identifiers to port drivers and back without them potentially changing identity you need this functionality. In our case we want to pass references to the inet-driver in order to utilize the selective receive optimization. Without this feature, references can change identity which would cause processes communicating with the driver to hang indefinitely.

okeuday:
@rickard-green If a length was provided by `LOCAL_EXT` it could be treated as a binary that doesn't get decoded unless that is necessary.  Otherwise, in other programming languages, `LOCAL_EXT` would require a custom function that decodes every instance of it.  Since the purpose is to keep the opaque data constant normally, it probably doesn't need to get decoded often.  Please consider having a length in the `LOCAL_EXT` encoding.  From a verification view, having a length would help ensure the remaining part of the binary isn't corrupt (e.g., odd situations where a `LOCAL_EXT` encoding is not yet known).

rickard-green:
@okeuday

You seem to want to use this for something it was not intended for. I also don't see a use case for that. If you want data that aren't to be decoded unless under certain circumstances, put it in a binary and introduce a protocol with which you can determine if you should decode the binary or not. The only effect the length field would have is unnecessarily growing the local encoding which I don't want.

Other programming languages *should not* try to interpret `LOCAL_EXT` (just fail) if they have not implemented some protocol of their own for handling something similar to our local encoding.

okeuday:
@rickard-green What you have described appears to indicate that `term_to_binary/2` and `term_to_iovec/2` binaries created with the `local` option should not be used in Erlang ports, cnodes and other uses of the external term format.  If that is the case, it would be good to put that in the documentation (i.e., only for use in NIFs and port drivers).

With `LOCAL_EXT` not having the encoding defined, it is assumed that while the data may contain `NEWER_REFERENCE_EXT`, `NEW_PID_EXT` or `NEW_PORT_EXT` an Erlang port or cnode should not decode it, to instead fail as you described.  However, that would mean any of the extra data encoded would be inaccessible and the data encoded with `LOCAL_EXT` would be unable to be passed back in the future.  I was suggesting a length in `LOCAL_EXT` because I thought Erlang ports, cnodes and other things might want to use the external term format binaries created with the `local` option.

rickard-green:
> @rickard-green What you have described appears to indicate that `term_to_binary/2` and `term_to_iovec/2` binaries created with the `local` option should not be used in Erlang ports, cnodes and other uses of the external term format. If that is the case, it would be good to put that in the documentation (i.e., only for use in NIFs and port drivers).
> 

That is not the case. It can be used  when communicating against anything that communicates with an Erlang runtime system using a serialized format where this issue with node state changes may occur. When you communicate over the distribution you however typically do not want to use it (the issue should also typically not occur then). You can however not just slap on the `local` encoding on everything, you need to make a protocol for what is encoded on `local` term format which will depend on what you want to do.

> With `LOCAL_EXT` not having the encoding defined, it is assumed that while the data may contain `NEWER_REFERENCE_EXT`, `NEW_PID_EXT` or `NEW_PORT_EXT` an Erlang port or cnode should not decode it, to instead fail as you described. However, that would mean any of the extra data encoded would be inaccessible and the data encoded with `LOCAL_EXT` would be unable to be passed back in the future.

Nothing encoded with `local` option should be decoded by anything else than the runtime system instance that encoded it. The user needs to make a protocol that only encodes data on the `local` format that should be encoded like that.

> I was suggesting a length in `LOCAL_EXT` because I thought Erlang ports, cnodes and other things might want to use the external term format binaries created with the `local` option.

The first version of the documentation was quite thin. My hope is that it becomes clearer when reading the new documentation.

okeuday:
@rickard-green Ok, thank you for adding the additional documentation after the first version.  However, the current documentation is a little misleading in the current version because of the `term_to_binary/2` `local` option section:
```
If a term encoded using the local option is stripped of its
leading version number it can be added as part of a larger term
(for example as an element in a tuple) when encoding on the
external term format using, for example, ei. In the ei case, you would
strip it of the version number using ei_decode_version() and
then add the remaining local encoding to what you are encoding
using for example ei_x_append_buf().
```
The problem is that the resulting tuple that contains `LOCAL_EXT` is unable to be decoded by anything except `erlang:binary_to_term/1,2`.  So, it wouldn't be accessible as opaque data in an Erlang port, cnode or something else that attempts to decode the external term format.  To make it accessible as opaque data, it would need to be wrapped as a `BINARY_EXT` or `LOCAL_EXT` would need a length integer (which is why I was concerned about the length of `LOCAL_EXT`).  The documentation may want to change to make the situation clearer by explaining that a binary that contains `LOCAL_EXT` may only be decoded by `erlang:binary_to_term/1,2`.

rickard-green:
> @rickard-green Ok, thank you for adding the additional documentation after the first version. However, the current documentation is a little misleading in the current version because of the `term_to_binary/2` `local` option section:
> 
> ```
> If a term encoded using the local option is stripped of its
> leading version number it can be added as part of a larger term
> (for example as an element in a tuple) when encoding on the
> external term format using, for example, ei. In the ei case, you would
> strip it of the version number using ei_decode_version() and
> then add the remaining local encoding to what you are encoding
> using for example ei_x_append_buf().
> ```
> 
> The problem is that the resulting tuple that contains `LOCAL_EXT` is unable to be decoded by anything except `erlang:binary_to_term/1,2`. So, it wouldn't be accessible as opaque data in an Erlang port, cnode or something else that attempts to decode the external term format. To make it accessible as opaque data, it would need to be wrapped as a `BINARY_EXT` or `LOCAL_EXT` would need a length integer (which is why I was concerned about the length of `LOCAL_EXT`). The documentation may want to change to make the situation clearer by explaining that a binary that contains `LOCAL_EXT` may only be decoded by `erlang:binary_to_term/1,2`.

How the term encoded on the local format is transported to another entity depends on the protocol that the user has defined, but serializing a request like this `term_to_binary({request, term_to_binary(Ref, [local]), Request})` can be one way.

I moved up the paragraph that talks about that it is only the runtime that created the local encoding that can decode it and added the following to the paragraph that you quoted to make it a bit clearer.

> Since it is only the runtime system that encoded a term using the local option that can decode it, the local encoding is typically pieced together with something else to produce a reply to where the local encoding originates from.

Note that it is not only `binary_to_term()` that can decode a term on the local encoding it. All functionality that can decode terms on the external format on a runtime system can also decode terms on the local external format. For example, `erl_drv_send_term()`.

okeuday:
> I moved up the paragraph that talks about that it is only the runtime that created the local encoding that can decode it and added the following to the paragraph that you quoted to make it a bit clearer:
> 
> > Since it is only the runtime system that encoded a term using the local option that can decode it, the local encoding is typically pieced together with something else to produce a reply to where the local encoding originates from.

Ok, that helps to make the usage clearer.  Thanks

PR #7013, opened 2023-03-13
erl_parse: Update types

### LINKED ISSUE #6956: `maybe` not reflected in erl_parse:abstract_expr/0
There's rudimentary support for `maybe` in Gradualizer so that the tool doesn't crash, but rather report it can't type check the expression yet. However, since `maybe` is not part of its `gradualizer_type:abstract_expr/0` originally based on `erl_parse:abstract_expr/0` and updated to match Erlang 24, Gradualizer doesn't self-check correctly. I wanted to borrow `maybe` references from the post Erlang 25 `erl_parse:abstract_expr/0`, but it seems there are none, although the new keyword is now added to the Erlang grammar definition.

bjorng:
* The `binary_op()` type was missing the `!` operator.
* The `abstract_expr()` was missing the `maybe` construct.

Closes #6956

### DIFF
--- a/lib/stdlib/src/erl_parse.yrl
+++ b/lib/stdlib/src/erl_parse.yrl
@@ -661,6 +661,7 @@ Erlang code.
 
 -type abstract_expr() :: af_literal()
                        | af_match(abstract_expr())
+                       | af_maybe_match()
                        | af_variable()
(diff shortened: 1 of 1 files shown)


erszcz:
Thank you!

PR #7025, opened 2023-03-16
[kernel] Reduce memory usage for volatile send in global

rickard-green:


lib/kernel/src/global.erl
```diff
@@ -2760,19 +2760,12 @@ inform_connection_loss(_Node, #state{}) ->
...
-        noconnect ->
-            ok;
-        nosuspend ->
-            _ = spawn(fun () ->
-                              _ = erlang:send(To, Msg, [noconnect])
-                      end),
-            ok
-    end.
+    OldAsyncDist = process_flag(async_dist, true),
+    _ = erlang:send({global_name_server, Node}, Msg, [noconnect]),
```
max-au, line 2766: Is there a chance that send/3 throws an exception (and `async_dist` flag does not get flipped back)?

rickard-green: If it were to throw an exception, `global` would terminate and bring down the whole node with it.

PR #7032, opened 2023-03-21
queue: add examples

kikofernandez:
Improve module `queue` by adding examples to many of its operations.

### DIFF
--- a/lib/stdlib/doc/src/queue.xml
+++ b/lib/stdlib/doc/src/queue.xml
@@ -44,7 +44,6 @@
       not lists. Improper lists cause internal crashes.
       An index out of range for a queue also causes
       a failure with reason <c>badarg</c>.</p>
-
     <p>Some functions, where noted, fail with reason <c>empty</c>
       for an empty queue.</p>
 
@@ -89,7 +88,7 @@
       are reverse operations on the queue.</p>
 
     <p>This module has three sets of interface functions: the
-      "Original API", the "Extended API", and the "Okasaki API".</p>
+      <em>"Original API"</em>, the <em>"Extended API"</em>, and the <em>"Okasaki API"</em>.</p>
 
     <p>The "Original API" and the "Extended API" both use the
       mental picture of a waiting line of items. Both
@@ -133,6 +132,13 @@
         <p>Returns <c>true</c> if <c><anno>Pred</anno>(<anno>Item</anno>)</c>
           returns <c>true</c> for all items <c><anno>Item</anno></c> in
           <c><anno>Q</anno></c>, otherwise <c>false</c>.</p>
+          <p><em>Example:</em></p>
+          <pre>
+1> <input>Queue = queue:from_list([1,2,3,4,5]).</input>
+2> <input>queue:all(fun (E) -> E > 3 end, Queue).</input>
+false
+3> <input>queue:all(fun (E) -> E > 0 end, Queue).</input>
+true</pre>
       </desc>
     </func>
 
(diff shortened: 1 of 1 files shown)


lib/stdlib/doc/src/queue.xml
```diff
@@ -44,9 +44,18 @@
...
+1> <input>queue:len([]).</input>
+** exception error: bad argument</pre>
 
     <p>Some functions, where noted, fail with reason <c>empty</c>
       for an empty queue.</p>
+    <p><em>Example:</em></p>
+      <pre>
+1> <input>queue:get(queue:new()).</input>
+** exception error: empty</pre>
+
```
Maria-12648430, line 58: Are examples of invalid input really needed? :thinking:

lib/stdlib/doc/src/queue.xml
```diff
@@ -44,9 +44,18 @@
       not lists. Improper lists cause internal crashes.
       An index out of range for a queue also causes
       a failure with reason <c>badarg</c>.</p>
+      <p><em>Example:</em></p>
+      <pre>
+1> <input>queue:len([]).</input>
+** exception error: bad argument</pre>
```
Maria-12648430, line 50: This is IMO a bad example of invalid input. `len/1` expects an argument of type `queue`, which is an _opaque_ type. Thus, same as you can't assume anything about what it _is_, you can't make any assumptions about what it is _not_.

Maria-12648430 (changes requested):
Comments inline.

lib/stdlib/doc/src/queue.xml
kikofernandez: Right, I was trying to show that if one passes the wrong argument, you get the described `badarg`.
If this is confusing or does not go well in here, I can move :)

lib/stdlib/doc/src/queue.xml
Maria-12648430: The sentence "fail with reason ..." is used in many places in the documentation with the meaning "throws an exception with reason ...", so I think it is generally understood. Even if it does need to be clarified, I think it would be better just to say "throws an exception" instead of giving an example.

lib/stdlib/doc/src/queue.xml
Maria-12648430: Maybe it would be better to just exemplify with another function, like giving an atom to `queue:from_list/1` or something, where the expected argument type is not opaque.

lib/stdlib/doc/src/queue.xml
```diff
@@ -201,12 +248,30 @@
...
           as returning <c>[]</c> is semantically equivalent to
           returning <c>false</c>. But returning a list builds
           more garbage than returning an atom.</p>
+          <p><em>Example 2:</em></p>
+          <pre>
+1> <input>Queue = queue:from_list([1,2,3,4,5]).</input>
+2> <input>Queue1 = queue:filter(fun (E) -> [E, E+1], Queue).</input>
+3> <input>queue:to_list(Queue1).</input>
+[1,2,2,3,3,4,4,5,5,6]</pre>
+          <p><em>Example 3:</em></p>
```
bjorng, line 269: Here it makes sense having multiple examples. I am not sure that the third example is really necessary.

lib/stdlib/doc/src/queue.xml
```diff
@@ -421,6 +546,16 @@
         <p>Returns <c><anno>Item</anno></c> at the front of queue
           <c><anno>Q</anno></c>.</p>
         <p>Fails with reason <c>empty</c> if <c><anno>Q</anno></c> is empty.</p>
+          <p><em>Example 1:</em></p>
+          <pre>
+1> <input>Queue = queue:from_list([1,2,3,4,5]).</input>
+2> <input>1 == queue:get(Queue).</input>
+</pre>
+        <p><em>Example 2:</em></p>
```
bjorng, line 554: When the examples are very short (with no explaining text between them), I think it makes better sense to combine them. In my opinion, that is easier to read.

bjorng (approved):
Overall looks good to me.

In my opinion, having multiple example blocks ("Example 1" and "Example 2") when they are very short (with no explanatory text between them) makes them harder to read.

I agree with @Maria-12648430 that it might not be necessary to give examples of invalid input.

kikofernandez:
Thanks for the comments, I think I should have addressed all of them with the new commit :)

kikofernandez:
Thanks for your feedback! I thought I had applied it in all places, so I went ahead with the merge :)

PR #7033, opened 2023-03-21
Kuba/ssl/exp ocsp refactor

u3s:


lib/ssl/src/ssl_internal.hrl
```diff
@@ -201,11 +201,17 @@
...
+-define(DEFAULT_OCSP_RESPONDER_CERTS, []).
+-define(GET_OPT_FROM_MAP(MapName, Opt, Default, SslOpts),
+        fun() ->
+            case maps:get(MapName, SslOpts, disabled) of
+                #{Opt := Value} ->
+                    Value;
+                _ ->
+                    Default
+            end
+        end()).
```
IngelaAndin, line 215: Do we need this macro?  We should only need to look inside of the map when it exists.

u3s: You're right. I will remove it.

PR #7034, opened 2023-03-21
orddict: add examples

kikofernandez:
This PR adds usage examples for `orddict` module. In most cases it handles the erroneous case and a successful care.

### DIFF
--- a/lib/stdlib/doc/src/orddict.xml
+++ b/lib/stdlib/doc/src/orddict.xml
@@ -38,7 +38,7 @@
     <p>This module provides a <c>Key</c>-<c>Value</c> dictionary.
       An <c>orddict</c> is a representation of a dictionary, where a
       list of pairs is used to store the keys and values. The list is
-      ordered after the keys in the <em>Erlang term order</em>.</p>
+      ordered after the keys in the <em><seeguide marker="system/reference_manual:expressions#term-comparisons">Erlang term order</seeguide></em>.</p>
 
     <p>This module provides the same interface as the
       <seeerl marker="dict"><c>dict(3)</c></seeerl> module
@@ -69,6 +69,24 @@
           generated if the initial value associated with <c><anno>Key</anno></c>
           is not a list of values.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+          <p><em>Example 1:</em></p>
+          <pre>
+1> <input>OrdDict1 = orddict:from_list([{x, []}]).</input>
+[{x,[]}]
+2> <input>OrdDict2 = orddict:append(x, 1, OrdDict1).</input>
+[{x,[1]}]
+3> <input>OrdDict3 = orddict:append(x, 2, OrdDict2).</input>
+[{x,[1,2]}]
+4> <input>orddict:append(y, 3, OrdDict3).</input>
+[{x,[1,2]},{y,[3]}]</pre>
+          <p><em>Example 2:</em></p>
+          <pre>
+1> <input>OrdDict1 = orddict:from_list([{a, no_list}]).</input>
+[{a,no_list}]
+2> <input>orddict:append(a, 1, OrdDict1).</input>
+** exception error: bad argument
+     in operator  ++/2
+        called as no_list ++ [1]</pre>
       </desc>
     </func>
 
@@ -81,6 +99,14 @@
           An exception is generated if the initial value associated with
           <c><anno>Key</anno></c> is not a list of values.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+          <p><em>Example:</em></p>
+          <pre>
+1> <input>OrdDict1 = orddict:from_list([{x, []}]).</input>
+[{x,[]}]
+2> <input>OrdDict2 = orddict:append_list(x, [1,2], OrdDict1).</input>
+[{x,[1,2]}]
+3> <input>OrdDict3 = orddict:append_list(y, [3,4], OrdDict2).</input>
+[{x,[1,2]},{y,[3,4]}]</pre>
       </desc>
     </func>
 
@@ -89,6 +115,12 @@
       <fsummary>Erase a key from a dictionary.</fsummary>
       <desc>
         <p>Erases all items with a specified key from a dictionary.</p>
+          <p><em>Example:</em></p>
+          <pre>
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
+[{a,1},{b,2}]
+2> <input>orddict:erase(a, OrdDict1).</input>
+[{b,2}]</pre>
       </desc>
     </func>
 
@@ -101,6 +133,14 @@
           the <c><anno>Key</anno></c> is present in the dictionary. An exception
           is generated if <c><anno>Key</anno></c> is not in the dictionary.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+        <p><em>Example:</em></p>
+<pre>
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
+[{a,1},{b,2}]
+2> <input>orddict:fetch(a, OrdDict1).</input>
+1
+3> <input>orddict:fetch(missing, OrdDict1).</input>
+** exception error: no function clause matching orddict:fetch(missing,[])</pre>
       </desc>
     </func>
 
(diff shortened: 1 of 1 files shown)


lib/stdlib/doc/src/orddict.xml
```diff
@@ -38,7 +38,9 @@
     <p>This module provides a <c>Key</c>-<c>Value</c> dictionary.
       An <c>orddict</c> is a representation of a dictionary, where a
       list of pairs is used to store the keys and values. The list is
-      ordered after the keys in the <em>Erlang term order</em>.</p>
+      ordered after the keys in the <em>Erlang term order</em>:</p>
+      <pre>
+number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map &lt; nil &lt; list &lt; bit string</pre>
```
bjorng, line 43: It seems unnecessary to list all types here. I think it is better to turn the text **Erlang term order** into a link to the definition of term order.

lib/stdlib/doc/src/orddict.xml
```diff
@@ -69,6 +71,22 @@
...
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
+[{a,1},{b,2}]
+2> <input>OrdDict2 = orddict:append(c, 3, OrdDict1).</input>
+[{a,1},{b,2},{c,[3]}]
+3> <input>orddict:fetch(c, OrdDict2).</input>
+[3]
+</pre>
+          <p><em>Example 2:</em></p>
+          <pre>
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
```
bjorng, line 85: It seems unnecessary to have two examples, unless you have some explaining text between the examples, which you don't have here. In this case, I think it would be easier to read if you combined the examples. That applies to many of the other examples.

lib/stdlib/doc/src/orddict.xml
```diff
@@ -101,6 +140,17 @@
           the <c><anno>Key</anno></c> is present in the dictionary. An exception
           is generated if <c><anno>Key</anno></c> is not in the dictionary.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+        <p><em>Example 1:</em></p>
+<pre>
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
+2> <input>orddict:fetch(a, OrdDict1).</input>
```
bjorng, line 146: For consistency with other examples, you should show the output of the command. (That applies also to other some other examples.)

lib/stdlib/doc/src/orddict.xml
```diff
@@ -236,6 +337,11 @@ merge(Fun, D1, D2) ->
           dictionary. If the <c><anno>Key</anno></c> already exists in
           <c><anno>Orddict1</anno></c>,
           the associated value is replaced by <c><anno>Value</anno></c>.</p>
+        <p><em>Example:</em></p>
+<pre>
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
+2> <input>orddict:store(a, 99, OrdDict1).</input>
```
bjorng, line 343: I suggest that you also store a key that is not there, for example `c`.

bjorng (commented):
While at it, I suggest that you the `maps` module to the "See Also" section.

lib/stdlib/doc/src/orddict.xml
kikofernandez: Yes, I agree. Will do tomorrow

lib/stdlib/doc/src/orddict.xml
bjorng: In this particular case, having two examples can be clearer, because you can start out with different initial orddicts to make it clearer what is happening.

Here is a suggestion:

**Example 1:**

```
1> OrdDict1 = orddict:from_list([{x, []}]).
[{x,[]}]
2> OrdDict2 = orddict:append(x, 1, OrdDict1).  
[{x,[1]}]
3> OrdDict3 = orddict:append(x, 2, OrdDict2).
[{x,[1,2]}]
4> orddict:append(y, 3, OrdDict3). 
[{x,[1,2]},{y,[3]}]
```

(Your example didn't really show any real appending being done, so I fixed that. By only having one key in the orddict, it is clear enough what is happening without having to use `orddict:fetch/2`.)

**Example 2:**
```
1> OrdDict1 = orddict:from_list([{a, no_list}]).
[{a,no_list}]
2> orddict:append(a, 1, OrdDict1).              
     in operator  ++/2
        called as no_list ++ [1]
```

lib/stdlib/doc/src/orddict.xml
```diff
@@ -81,6 +99,22 @@
...
+          <p><em>Example 1:</em></p>
+          <pre>
+1> <input>OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).</input>
+[{a,1},{b,2}]
+2> <input>OrdDict2 = orddict:append_list(c, [3, 4], OrdDict1).</input>
+[{a,1},{b,2},{c,[3, 4]}]
+3> <input>orddict:fetch(c, OrdDict2).</input>
+[3,4]
+</pre>
+          <p><em>Example 2:</em></p>
```
bjorng, line 111: Here is an instance where (in my opinion) having two examples only makes it harder to read. I suggest:

```
1> OrdDict1 = orddict:from_list([{a, 1}, {b, 2}]).
[{a,1},{b,2}]
2> orddict:fetch(a, OrdDict1).
1
3> orddict:fetch(missing, OrdDict1).
** exception error: no function clause matching orddict:fetch(missing,[])
```

lib/stdlib/doc/src/orddict.xml
```diff
@@ -81,6 +99,15 @@
           An exception is generated if the initial value associated with
           <c><anno>Key</anno></c> is not a list of values.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+          <p><em>Example:</em></p>
```
bjorng, line 102: This example can be improved in the same way as the `orddict:append/3` example. That is, show an actual append operation on an existing entry, and start out with a single existing key so that the example becomes clear without a call to `orddict:fetch/2`.

lib/stdlib/doc/src/orddict.xml
```diff
@@ -101,6 +134,14 @@
           the <c><anno>Key</anno></c> is present in the dictionary. An exception
           is generated if <c><anno>Key</anno></c> is not in the dictionary.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+        <p><em>Example 1:</em></p>
```
bjorng, line 137:
```suggestion
        <p><em>Example:</em></p>
```

lib/stdlib/doc/src/orddict.xml
```diff
@@ -118,6 +165,14 @@
       <desc>
         <p>This function returns value from dictionary and new dictionary without this value.
             Returns <c>error</c> if the key is not present in the dictionary.</p>
+        <p><em>Example 1:</em></p>
```
bjorng, line 168:
```suggestion
        <p><em>Example:</em></p>
```

lib/stdlib/doc/src/orddict.xml
```diff
@@ -141,6 +202,14 @@
           the value associated with <c><anno>Key</anno></c>, or <c>error</c> if
           the key is not present in the dictionary.</p>
         <p>See also section <seeerl marker="#notes">Notes</seeerl>.</p>
+        <p><em>Example 1:</em></p>
```
bjorng, line 205:
```suggestion
        <p><em>Example:</em></p>
```

lib/stdlib/doc/src/orddict.xml
```diff
@@ -254,6 +351,12 @@ merge(Fun, D1, D2) ->
         <p>Updates a value in a dictionary by calling <c><anno>Fun</anno></c>
           on the value to get a new value. An exception is generated if
           <c><anno>Key</anno></c> is not present in the dictionary.</p>
+        <p><em>Example 1:</em></p>
```
bjorng, line 354:
```suggestion
        <p><em>Example:</em></p>
```

bjorng (approved):

lib/stdlib/doc/src/orddict.xml
kikofernandez: Thanks for the feedback. I followed your suggestion

PR #7035, opened 2023-03-21
[xref] drop unused code

max-au:
This code does nothing.

Reported by @mmin on the Erlang Forums,
https://erlangforums.com/t/wondering-why-we-need-the-last-expression-in-this-code-from-xref-reader-erl/2482


bjorng:

PR #7036, opened 2023-03-21
jit: Share more code between architectures

jhogberg:
This PR tries to reduce the duplication of code between architectures, mainly that relating to argument types. A bunch of general code is still kept in the architecture-specific part as sharing it was more trouble than it was worth at the moment, but we'll get to that later on.


bjorng (approved):
Overall looks good to me.

PR #7039, opened 2023-03-22
asn1: fix RELATIVE-OID validation

montag451:
While trying to compile the ASN.1 specification of the latest version of the ETSI 102 232-2 norm ([EmailPDU.txt](https://github.com/erlang/otp/files/11039662/EmailPDU.txt)), I found a bug in the validation of RELATIVE-OID values. The current ASN.1 compiler rejects RELATIVE-OID values containing other RELATIVE-OID values when it's perfectly valid according to the ASN.1 specification (see https://www.itu.int/rec/T-REC-X.680-202102-I/en, section 33.5). This PR fixes this bug and adds more tests to check the behaviour of the compiler regarding RELATIVE-OID validation.


KennethL:
Thanks for the PR, I have put it into testing

PR #7042, opened 2023-03-22
stdlib: Add query optimisation to `ets:fun2ms/1`

TD5:
Unlike writing match specs directly, `ets:fun2ms/1` generates queries by translating an erlang function expression. This is convenient and makes for readable queries, but it necessarily trades-off some expressiveness in favour of simplicity (for example, it's not possible to generate a match spec pattern guard that matches against an in-scope variable: users are forced to use something like an equality guard instead). Here, we resolve that issue by reading the author's _intention_ from the given function expression, generating the match spec as before via `ms_transform`, but then running an optimisation pass over it during compilation in order to generate more efficient queries.

Performance
===========

Amongst other things, we optimise equality guards by moving them into the pattern, which can avoid scanning the whole table, making queries `O(1)` or `O(log(n))` (depending on the table type), rather than `O(n)`, (where `n` is the number of rows in the table). In other words, this is not primarily a micro-optimisation, but rather a very substantial algorithmic complexity improvement for many common queries.

In practice, I have seen no situations where the new `ets:fun2ms/1` queries are slower, but many simple queries can be executed drastically faster when the number of rows in the table is large.

For example, even a simple query over a table of a million rows made up of pairs of keys and values queried with:

```erlang
make_query(Key) ->
  ets:fun2ms(fun({K, V}) when K =:= Key -> {K,V} end). 
```

now executes **>1000x faster** with my local benchmarks. Almost any query which requires that a `=:=` guard always hold will potentially see a substantial performance improvement for queries over big datasets.

Theory
======

From the existing ETS match spec docs:
> Traversals using match and select functions may not need to scan the
> entire table depending on how the key is specified. A match pattern with
> a fully bound key (without any match variables) will optimize the
> operation to a single key lookup without any table traversal at all. For
> ordered_set a partially bound key will limit the traversal to only scan
> a subset of the table based on term order. A partially bound key is
> either a list or a tuple with a prefix that is fully bound.

We can leverage this knowledge to re-write queries to make better use of the key.

For example:

```erlang
make_query(Key) ->
  ets:fun2ms(fun({K, V}) when K =:= Key -> {K,V} end). 
```

was previously compiled to:

```erlang
{
  {'$1', '$2'},
  [
    {'=:=', '$1', Key}
  ],
  [{'$1', '$2'}]
}
```

This was sub-optimal, since the equality guard is less efficient than the functionally-equivalent pattern match because the equality guard did not result in a fast lookup using the table's key.

Now, the same function expression is compiled to this, more efficient, query:

```erlang
{
  {Key, '$2'},
  [],
  [{Key, '$2'}]
}
```

We can also simplify constant parts of queries statically, and perform other rewritings to improve efficiency, but the largest win comes from inlining the values of variables bound by guards such as `(K =:= Key)`.

Implementation
==============

This optimisation is implemented for all relevant literals that I could find. Floats were given extra consideration and testing because of the differences in `==`/`=:=` vs. pattern matching.  In this situation, the handling of floats in `ordered_set` is safe because we only inline `=:=` guards into the the match head and body, but we leave `==` as a guard, since determining statically whether the table type would make this a safe operation or not is not feasible using the the information available in the parse transform.

New unit tests cover the parse transform compiling to the expected match expression, the match expression matching the expected rows, and the equivalence between the naive match expression and the optimised one in terms of data returned.  See the changes to `ms_transform_SUITE.erl` for more information.

This optimisation is specifically applied in `ets:fun2ms/1`, because I think users would expect generated match specs to avoid trivial inefficiencies (and, indeed, utilising the key efficiently when it was given as a parameter was impossible to express before). Moreover, by making use of `ets:fun2ms/1`, users have already ceded some control of the generated match spec to the tooling. Users who construct match specs directly will be unaffected.

Notably, since `ets:fun2ms/1` is transformed at compile time (outside of the shell, at least), we don't pay any unnecessary performance penalty at runtime in order to apply these optimisations, and the cost of doing them at compile time is low relative to other operations.

Later work could explore runtime query-planning for ETS, but avoiding introducing performance regressions for at least some queries will be harder to guarantee, since we then we would have to consider the runtime cost of computing the optimisation itself.

Optimisation can be disabled with the `no_optimise_fun2ms` compiler flag, but by default it is enabled. The flag can be altered via the usual compile flag mechanisms, including the `-compile(no_optimise_fun2ms)` attribute.

### DIFF
--- a/lib/stdlib/src/ms_transform.erl
+++ b/lib/stdlib/src/ms_transform.erl
@@ -19,7 +19,7 @@
 %%
 -module(ms_transform).
 
--export([format_error/1,transform_from_shell/3,
+-export([format_error/1,transform_from_shell/3,transform_from_shell/4,
          parse_transform/2,parse_transform_info/0]).
 
 %% Error codes.
@@ -205,8 +205,22 @@ parse_transform_info() ->
       BoundEnvironment :: erl_eval:binding_struct().
 
 transform_from_shell(Dialect, Clauses, BoundEnvironment) ->
+  transform_from_shell(Dialect, Clauses, BoundEnvironment, true).
+
+-spec transform_from_shell(Dialect, Clauses, BoundEnvironment, ShouldOptimise) -> term() when
+      Dialect :: ets | dbg,
+      Clauses :: [erl_parse:abstract_clause()],
+      BoundEnvironment :: erl_eval:binding_struct(),
+      ShouldOptimise :: boolean().
+
+transform_from_shell(Dialect, Clauses, BoundEnvironment, ShouldOptimise) ->
     SaveFilename = setup_filename(),
-    case catch ms_clause_list(1,Clauses,Dialect,gb_sets:new()) of
+    MaybeOptimise =
+      case ShouldOptimise of
+        true -> fun optimise_ms/1;
+        false -> fun (MS) -> MS end
+      end,
+    case catch MaybeOptimise(ms_clause_list(1,Clauses,Dialect,gb_sets:new())) of
 	{'EXIT',Reason} ->
 	    cleanup_filename(SaveFilename),
 	    exit(Reason);
@@ -224,7 +238,7 @@ transform_from_shell(Dialect, Clauses, BoundEnvironment) ->
 		    Ret
             end
     end.
-    
+
 
 %%
 %% Called when translating during compiling
@@ -237,10 +251,10 @@ transform_from_shell(Dialect, Clauses, BoundEnvironment) ->
       Errors :: {error, ErrInfo :: [tuple()], WarnInfo :: []},
       Warnings :: {warning, Forms2, WarnInfo :: [tuple()]}.
 
-parse_transform(Forms, _Options) ->
+parse_transform(Forms, Options) ->
     SaveFilename = setup_filename(),
-    %io:format("Forms: ~p~n",[Forms]),
-    case catch forms(Forms) of
+    ShouldOptimise = not proplists:get_bool(no_optimise_fun2ms, Options),
+    case catch forms(Forms, ShouldOptimise) of
 	{'EXIT',Reason} ->
 	    cleanup_filename(SaveFilename),
 	    exit(Reason);
@@ -248,7 +262,6 @@ parse_transform(Forms, _Options) ->
 	    {error, [{cleanup_filename(SaveFilename),
 		      [{location(AnnoOrUnknown), ?MODULE, R}]}], []};
 	Else ->
-	    %io:format("Transformed into: ~p~n",[Else]),
 	    case get_warnings() of
 		[] ->
 		    cleanup_filename(SaveFilename),
@@ -345,10 +358,10 @@ record_field({record_field,_,{atom,_,FieldName},Def}, C) ->
 record_field({typed_record_field,Field,_Type}, C) ->
     record_field(Field, C).
 
-forms(Forms0) ->
+forms(Forms0, ShouldOptimise) ->
     put(records_replaced_by_tuples, []),
     try
-        Forms = [form(F) || F <- Forms0],
+        Forms = [form(F, ShouldOptimise) || F <- Forms0],
         %% Add `-compile({nowarn_unused_record, RecordNames}).', where
         %% RecordNames is the names of all records replaced by tuples,
         %% in order to silence the code linter's warnings about unused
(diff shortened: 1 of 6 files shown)


KennethL:
The optimization is good since it creates an efficient query from an inefficient fun2ms usage.
But the fun2ms can also be written:
```erlang
3> ets:fun2ms(fun({42, V}) -> {42,V} end).              
[{{42,'$1'},[],[{{42,'$1'}}]}]
```
which result in the same match spec as the optimized one.
But the optimization let the user write nicer and easier to understand match specs with fun2ms

TD5:
This is true, but perhaps my example was too simplified to show the value of the optimiser in general.

Consider a case where we don't have a literal `42`, but instead a parameter to our function `Key`. In that case, we can't write something like:

```
make_query(Key) ->
  ets:fun2ms(fun({Key, V}) -> {Key,V} end). 
```

because of the existing scoping/shadowing rules of Erlang.

Instead, people tend to write the functionally equivalent, but significantly less efficient:

```
make_query(Key) ->
  ets:fun2ms(fun({K, V}) when K =:= Key -> {K,V} end). 
```

This PR solves the issue by letting users write the second form which makes use of a guard, but optimising it away behind the scenes. In some sense, this is a work around for the inability to write the pattern match in the `fun` due to Erlang's scoping/shadowing rules, but in another way, you can just see it as a general purpose query optimiser.

lib/stdlib/src/ms_transform.erl
```diff
@@ -450,10 +448,689 @@ copy_list([],Bound) ->
...
+  {atom,gen_loc(),false};
+simplify_guard_function('not', {atom, _, false}) ->
+  {atom,gen_loc(),true};
+simplify_guard_function(Op, Operand) ->
+  {tuple, gen_loc(), [{atom, gen_loc(), Op}, Operand]}.
+
+simplify_guard_function('>', {integer, _, X}, {integer, _, Y}) when X > Y ->
+  {atom,gen_loc(),true};
+simplify_guard_function('>', {integer, _, _}, {integer, _, _}) ->
+  {atom,gen_loc(),false};
```
sverker, line 926: I think this is overly verbose. What about
```
simplify_guard_function('>', {integer, _, X}, {integer, _, Y}) ->
   {atom,gen_loc(), X > Y};
```
or even
```
simplify_guard_function('>', {Num, _, X}, {Num, _, Y}) when Num =:= integer; Num =:= float ->
   {atom,gen_loc(), X > Y};
```

and what about if integers and floats are mixed?
```
simplify_guard_function('>', {integer, _, X}, {float, _, Y})  ->
   {atom,gen_loc(), X > Y};
```

sverker (commented):
Thanks a lot. This is a nice optimization.

Some code style comments:
I'm not fond of white space fixes in otherwise untouched lines. I would prefer them in a separate commit if you want to keep them.
Some lines seemed a bit too long for my taste. We try to keep the 80 column limit in general I think.

lib/stdlib/src/ms_transform.erl
TD5: I did none of those things, because I wanted follow the pattern from some of the lines above, and in those I was quite conservative, because the pattern not matching doesn't mean the case is definitely false - only that I can't easily prove it is true. I suppose I could tweak some of these cases.

When it came to integers vs. floats, I just didn't want to open that can of worms. I thought I'd try to stay within the realm of things I feel like I largely understand, and potentially refine the optimiser later for more completeness if it appeared to be useful in practice.

TD5:
> Would this optimisation also impact tracing, or is the functionality too different to matter in the way match specs are run in this case?

I don't know how tracing is implemented at runtime, so I can't say for sure, but I suspect there won't be much of a win there because the re-writes here attempt to leverage ETS table indexing, and it's not clear to me how that would benefit tracing.

That said, the re-write makes the structure of the query more explicit, so even if there's no immediate win, in principle, I suspect it could be leveraged in the future to offer a small performance benefit.

sverker:
There are a bunch of failing test cases in [qlc_SUITE](https://erlang.github.io/prs/7042/ct_logs/ct_run.test_server@04f249767076.2023-04-03_15.44.32/make_test_dir.stdlib_test.logs/run.2023-04-03_15.45.27/suite.log.html)

They seem to be triggered by this PR, but I haven't investigated further.

TD5:
> There are a bunch of failing test cases in qlc_SUITE
> 
> They seem to be triggered by this PR, but I haven't investigated further.

Good find! I was running only a subset of the unit tests locally, and didn't think to include `qlc` 🤦‍♂️ The existing logic there called `ms_transform:parse_transform/2` in a few places which assumed it only output one match function per match spec, which is no longer true after optimisations. Rather than let this change spider out into lots of other code, I just pass disable the optimisations when calling `ms_transform` from `qlc_pt`.

TD5:
I just noticed I left some intentional crashes from debugging in there  ...but I can't remember what I was trying to achieve 🤦‍♂️. I'll get those fixed up after the renewed interest in https://erlangforums.com/t/variable-shadowing-constraint-of-match-spec-generated-using-ms-transform/2656/8

lib/stdlib/test/qlc_SUITE.erl
```diff
@@ -7257,15 +7257,18 @@ manpage(Config) when is_list(Config) ->
     [2,3,4] = qlc:eval(QH),
 
     %% ets(3)
-    MS = ets:fun2ms(fun({X,Y}) when (X > 1) or (X < 5) -> {Y} end),
```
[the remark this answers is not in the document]
TD5: It was both: The old condition seemed odd, but it also broke the test, since the test is asserting the handle is the same via `ets:fun2ms` and `qlc`, but this isn't true if `ets:fun2ms` applies some optimisations (which it did in the original case)

TD5:
> Was the merging of master to fix conflicts?

It seems like CI fails to build due to some sort of caching issue when I base this change on `maint` (it still builds fine for me locally).

TD5:
> It would be nice if you could write something in `ets:fun2ms` docs about this. Maybe use the new `<change>` tag when talking about difference in behavior between OTP versions 

I think that's a really good suggestion. I'll add something.

sverker:
Reading your [post at erlang forums](https://erlangforums.com/t/ets-queries-with-keys-which-look-like-meta-variables/2946) maybe you also realized this cannot be optimized correctly:

```
1> ets:fun2ms(fun({K,V}) when V =:= '$1' -> K end).
[{{'$1',{const,'$1'}},[],['$1']}]
```

I also discovered this bug:
```
2> ets:fun2ms(fun({K,V}) when K =:= V+1 -> K end).
[{{{'+','$2',1},'$2'},[],[{'+','$2',1}]}]
```

PR #7046, opened 2023-03-23
erts: Support is_boolean/1 in match specs

### LINKED ISSUE #7045: Support `erlang:is_boolean/1` in match specifications
# Synopsis

Today, match specifications do not support an equivalent of `erlang:is_boolean/1`. For example:

```erlang
erlang:match_spec_test({true}, [{{'$1'}, [{is_boolean, '$1'}], [{{'$1'}}]}], table).
% {error,[{error,"Function is_boolean/1 does_not_exist."}]}
```

This is surprising behaviour, since most guard-allowed `erlang` functions are available in match specifications, and it is trivial to implement in terms of `or` and `==`, both of which ***are*** available.

This is frustrating when translating existing code into match specs when moving filter/map operations into `ets` for performance reasons.

## Desired Result

Ideally, this would work as such:

```erlang
erlang:match_spec_test({true}, [{{'$1'}, [{is_boolean, '$1'}], [{{'$1'}}]}], table).
% {ok,{true},[],[]}
```

Additionally, `is_boolean/2` would be listed in the match specification documentation grammar guide [here](https://www.erlang.org/doc/apps/erts/match_spec.html#grammar).

## Alternatives

One can always compose the equivalent working match specification:

```erlang
erlang:match_spec_test({true}, [{{'$1'}, [{'or', {'==', '$1', true}, {'==', '$1', false}], [{{'$1'}}]}], table).
% {ok,{true},[],[]}
```

However, this discrepancy furthers the cognitive gap between what is valid Erlang code, and what is valid in match specifications.

jhogberg:
This PR adds a few missing guard BIFs like `is_boolean/1` to the list of functions allowed in match specifications, and adds a test to ensure that all new guard BIFs are added to this list.

Fixes #7045

### DIFF
--- a/erts/emulator/test/match_spec_SUITE.erl
+++ b/erts/emulator/test/match_spec_SUITE.erl
@@ -28,7 +28,8 @@
 	 ms_trace2/1, ms_trace3/1, ms_trace_dead/1, boxed_and_small/1,
 	 destructive_in_test_bif/1, guard_exceptions/1,
 	 empty_list/1,
-	 unary_plus/1, unary_minus/1, moving_labels/1]).
+	 unary_plus/1, unary_minus/1, moving_labels/1,
+         guard_bifs/1]).
 -export([fpe/1]).
 -export([otp_9422/1]).
 -export([faulty_seq_trace/1, do_faulty_seq_trace/0]).
@@ -57,7 +58,8 @@ all() ->
      faulty_seq_trace,
      empty_list,
      otp_9422,
-     maps].
+     maps,
+     guard_bifs].
 
 init_per_testcase(_TestCase, Config) ->
     Config.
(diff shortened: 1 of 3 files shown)


sverker:
Can you also add the BIFs to docs at https://www.erlang.org/doc/apps/erts/match_spec.html,
both for tracing and ets grammar.

erts/emulator/test/match_spec_SUITE.erl
```diff
@@ -1145,7 +1147,32 @@ moving_labels(Config) when is_list(Config) ->
...
+                       3 -> {F, '$1', '$2', '$3'}
+                   end,
+             erlang:match_spec_test({a, b, c},
+                                    [{{'$1','$2', '$3'},
+                                      [BIF],
+                                      [{{'$1', '$2', '$3'}}]}],
+                                    table)
+         end
+         || {F, A} <- erlang:module_info(functions),
+            F =/= is_record,
```
christhekeele, line 1168: I'd point out that `is_record/3` ***does*** work in match specs, it's only `is_record/2` that does not, so we may want to change this filter a little to reflect that.

christhekeele:
Just finished flexing all of these from a src build and they work great!

jhogberg:
Thanks, I've pushed an update with fixed documentation and extended tests. :-)

sverker (approved):

PR #7048, opened 2023-03-23
Fix pattern matching on optional `maps_order`

moogle19:
#6949 added fixes for the optional `maps_order` key in `io_lib:format_spec()`, but forgot to update the `build_limited` function.

This PR updates the `build_limited` function to match the behaviour of `print`, when `maps_order` is missing.


sverker:
@moogle19 Thank you, but this is a duplicate of #7050 which also contains a test case.

PR #7049, opened 2023-03-23
driver call callback fix

rickard-green:
Implementations of the `call()` driver callback that returned a faulty encoded result could cause a memory leak and could cause invalid data on the heap of the processes calling `erlang:port_call/3`.


benonymus:
Hey, 
How could I check if I am impacted by this?

I have an elixir application, that uses OTP_VERSION=25.2.1.

I see the host system memory constantly rising. The beam.smp's size constantly rising.

The beam.smp is above 1GB, but the internal memory metrics of the elixir application or erlang do not match up.
These report in the range fo 300-359mb.
```
:erlang.memory
[
  total: 275998648,
  processes: 81298904,
  processes_used: 81293272,
  system: 194699744,
  atom: 2687473,
  atom_used: 2663790,
  binary: 21397256,
  code: 93399782,
  ets: 21150928
]
```

Thank you

rickard-green:
@benonymus

Drivers using the functionality that this PR fixes are very rare. We got no such drivers in OTP itself.

If `erlang:memory()` doesn't report an increased amount of memory while `top` (or similar OS tools) report an increase, you are not effected by this memory leak. In that case you most likely have a NIF or driver that allocates memory using other APIs than those provided by the Erlang runtime system that leaks memory.

PR #7059, opened 2023-03-24
ssl: Enhance warning alert handling

IngelaAndin:
If a user cancel alert with level warning is received during handshake make it be handled the same regardless of TLS version. If it is received in connection in TLS-1.3 regard it as an error as it is inappropriate.

In TLS-1.3 all error alerts are considered FATAL regardless of legacy alert type. But make sure legacy type is printed in logs to not confuse users that are expecting the same legacy type as sent by peer.

### DIFF
--- a/lib/ssl/src/ssl_gen_statem.erl
+++ b/lib/ssl/src/ssl_gen_statem.erl
@@ -1076,26 +1076,8 @@ handle_normal_shutdown(Alert, StateName, #state{static_env = #static_env{role = 
     Pids = Connection:pids(State),
     alert_user(Pids, Transport, Trackers, Socket, StateName, Opts, Pid, RecvFrom, Alert, Role, StateName, Connection).
 
-handle_alert(#alert{level = ?FATAL} = Alert0, StateName,
-	     #state{static_env = #static_env{role = Role,
-                                             socket = Socket,
-                                             host = Host,
-                                             port = Port,
-                                             trackers = Trackers,
-                                             transport_cb = Transport,
-                                             protocol_cb = Connection},
-                    connection_env  = #connection_env{user_application = {_Mon, Pid}},
-		    ssl_options = #{log_level := LogLevel},
-                    start_or_recv_from = From,
-                    session = Session,
-		    socket_options = Opts} = State) ->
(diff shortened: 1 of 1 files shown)


u3s (approved):

lib/ssl/src/ssl_gen_statem.erl
```diff
@@ -1161,6 +1143,16 @@ handle_alert(#alert{level = ?WARNING, description = ?NO_RENEGOTIATION} = Alert,
     %% Go back to connection!
     State = Connection:reinit(State0#state{handshake_env = HsEnv#handshake_env{renegotiation = undefined}}),
     Connection:next_event(connection, no_record, State);
+handle_alert(#alert{level = ?WARNING, description = ?USER_CANCELED} = Alert, StateName,
+	     #state{static_env = #static_env{role = Role,
+                                             protocol_cb = Connection},
+                    connection_env = #connection_env{negotiated_version = Version},
```
u3s, line 1149: `Version -> _Version` ?

lib/ssl/src/ssl_gen_statem.erl
```diff
@@ -1171,9 +1163,33 @@ handle_alert(#alert{level = ?WARNING} = Alert, StateName,
...
-handle_alert(Alert0, StateName, State) ->
+handle_alert(Alert, StateName, State) ->
     %% In TLS-1.3 all error alerts are fatal not matter of legacy level
-    handle_alert(Alert0#alert{level = ?FATAL}, StateName, State).
+    %% but keep the level for the log so that users looking at what is
+    %% sent and what is logged are not confused! Or if some one sends
+    %% user cancel alert in connection which is inappropriate!
+    handle_fatal_alert(Alert, StateName, State).
+
+handle_fatal_alert(Alert, StateName,
```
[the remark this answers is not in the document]
IngelaAndin: Will fix

lib/ssl/src/ssl_gen_statem.erl
IngelaAndin: Version is not needed, will remove it. Needs merge of other branches before testing properly.

dgud (approved):

PR #7060, opened 2023-03-24
gh: Add a label which forces all tests to be run

TD5:
Adds a check in CI that looks for the `full-build-and-check` label, and if it is present, overrides the checking of which apps have changed, and tests everything.

This is useful for changes which can have consequences beyond the modules which were directly modified, giving confidence without needing to wait for the daily tests (which most contributors don't have access to). In the future, we could expand this to run an even broader suite of checks.


TD5:
Having created this PR, I see that I don't actually have permission to set labels myself. Can/should we allow users to set the `full-build-and-check` label themselves?

TD5:
Closing, since I have a bigger, better version of this coming 😁

PR #7081, opened 2023-03-30
[stdlib] Remove gen_statem call proxy process

rickard-green:
The proxy process was used to prevent late replies from reaching the client at timeout or connection loss. This is no longer needed since process aliases take care of this, are used and supported by all Erlang nodes that an OTP 26 Erlang node can communicate with.


RaimoNiskanen (approved):
LGTM

PR #7083, opened 2023-04-03
JIT: Avoid mixing AVX and SSE instructions

bjorng:
Mixing AVX and SSE instructions results in performance penalities on many CPUs.

Therefore, if AVX is available, the JIT will always emit AVX instructions instead of SSE instructions (for example, `vmovups` instead of `movups`). The JIT will also emit a `vzeroupper` instruction before calling any BIFs or helper routines in the runtime system to avoid a costly state change in case the runtime system uses SSE instructions.

### DIFF
--- a/erts/emulator/beam/jit/x86/beam_asm.hpp
+++ b/erts/emulator/beam/jit/x86/beam_asm.hpp
@@ -640,9 +640,9 @@ protected:
         if (ERTS_LIKELY(erts_frame_layout == ERTS_FRAME_LAYOUT_RA)) {
             if ((Spec & (Update::eHeap | Update::eStack)) ==
                 (Update::eHeap | Update::eStack)) {
-                /* To update both heap and stack we use sse instructions like
-                 * gcc -O3 does. Basically it is this function run through
-                 * gcc -O3:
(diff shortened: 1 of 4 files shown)


erts/emulator/beam/jit/x86/beam_asm.hpp
```diff
@@ -709,6 +723,9 @@ class BeamAssembler : public ErrorHandler {
         a.sub(x86::rsp, imm(15));
         a.and_(x86::rsp, imm(-16));
 #endif
+        if (hasCpuFeature(CpuFeatures::X86::kAVX)) {
+            a.vzeroupper();
+        }
```
jhogberg, line 728:
```suggestion
        /* If the emulator has not been compiled with AVX support (which stops it
         * from using legacy SSE instructions), we'll need to clear the upper bits of
         * all AVX registers to avoid AVX/SSE transition penalties.  */
#if !defined(__AVX__)
        if (hasCpuFeature(CpuFeatures::X86::kAVX)) {
            a.vzeroupper();
        }
#endif
```

jhogberg (approved):

PR #7087, opened 2023-04-03
Remove unused argument

michalmuskala:
This is not used in the `not_string` function


bjorng:

PR #7088, opened 2023-04-03
[erl_lint] Recurse with cheaper function in pattern checks

michalmuskala:
I noticed lists were recursed with the `is_pattern_expr_1/1` function, while tuples were recursed with the more expensive (since it's also doing an eval) `is_pattern_expr/1` function. Since there will be an eval at the end anyway, I don't see a reason for this difference. This makes it so that all internal recursion is using the cheaper function.


bjorng (approved):

bjorng:
Thanks for your pull request.

PR #7095, opened 2023-04-05
asn1: Fix warning about mismatched allocation function

frej:
As enif_alloc is, in applicable cases, declared with the malloc attribute, recent GCCs complain that: "‘free’ called on pointer returned from a mismatched allocation function". Avoid the warning by freeing the new chunk using enif_free().


sverker:
That's a real bug. enif_free and free are not interchangeable in the general case.

bjorng (approved):

PR #7097, opened 2023-04-05
Document the scoping rules for try/catch/of/after expressions

### LINKED ISSUE #6598: Clarifying the scoping rules of try/of/catch/after
Looking at https://www.erlang.org/doc/reference_manual/expressions.html#try, I've not found a description of the rules for scoping of variables in that language construct.

Here is what I've found empirically by giving small testcases to erlc:
- variables bound in the expressions just after `try` are:
  - bound in the `of` cases
  - unsafe in the `catch` and `after` cases, as well as after the whole construct
- variables bound in the `of` section are:
  - unbound in the `catch` section
  - unsafe in the `after` section, as well as after the whole construct
- variables bound in the `catch` section are:
  - unsafe in the `after` section, as well as after the whole construct
- variables bound in the `after` section are:
  - unsafe after the whole construct

Before sending a PR adding this information to the documentation, I would like to make sure that this is all working as intended.
In particular, the following two points are surprising to me:
- variables bound in the`of` section can be re-bound in the  `catch` section, whereas I would have expected them to be unsafe by consistency with all the other cases.
- variables bound in the `after` section cannot be used after it, despite being guaranteed to have been initialized.

RobinMorisset:
Fixes #6598.

### DIFF
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -1938,6 +1938,20 @@ catch
     exit:Reason -> {'EXIT',Reason}
     error:Reason:Stk -> {'EXIT',{Reason,Stk}}
 end</code>
+
+    <p>Variables bound in the various parts of these expressions have different scopes.
+      Variables bound just after the <c>try</c> keyword are:</p>
+    <list type="bulleted">
+      <item>bound in the <c>of</c> section</item>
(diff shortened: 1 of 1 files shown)


bjorng (approved):

bjorng:
Thanks for your pull request.

PR #7102, opened 2023-04-06
erts: Fix warning about set but unused variable

frej:
When assertions are disabled, `offset` is set but not used and Clang warns about it. Add a `(void)object` after the assert to silence the warning.


frej:
The failed `distribution_SUITE:dist_auto_connect_once` works fine locally and appears unrelated to the change in the PR, so unless someone tells me otherwise I will blame something transient in the CI environment.

PR #7103, opened 2023-04-06
Export additional crypto types

mkuratczyk:
New dialyzer default in OTP-26 made it clear that in https://github.com/rabbitmq/credentials-obfuscation/ we are relying on some types from the crypto module that are not exported (specifically: `crypto:cipher_iv` and `crypto:hash_algorithm`), so our pipeline fails with OTP-26.

Would you be ok with exporting them? In this PR I included a few more as they seem at the same "level" as some of the currently exported ones and the ones we would like to be exported. 

There's a comment saying `A minimum exported: only what public_key needs.` but no justification why `public_key` gets special treatment. ;)  These types (and many more) are documented in https://www.erlang.org/doc/man/crypto.html

Thanks,

### DIFF
--- a/lib/crypto/src/crypto.erl
+++ b/lib/crypto/src/crypto.erl
@@ -159,17 +159,27 @@
        ensure_engine_loaded_nif/2
       ]).
 
--export_type([ %% A minimum exported: only what public_key needs.
-               dh_private/0,
-               dh_public/0,
-               dss_digest_type/0,
-               ec_named_curve/0,
-               ecdsa_digest_type/0,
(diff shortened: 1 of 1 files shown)


IngelaAndin:
Alas, this will not make it for OTP-26.0, but I think it can be handled before 27.0 :)  The reason public_key got priority is that the algorithms used there are used in OTP by ssl and ssh applications which in turn means we have team members that know more about them without making any special investigation. We want to rather document to little and have the documentation being correct than it being more complete but wrong.  Also there is always extra work to also try to avoid having names that we later regret, we have too much of that already!

PR #7107, opened 2023-04-06
Kuba/inets/fix httpc remote socket close/otp 18545

u3s:


IngelaAndin (approved):

KennethL (commented):
Seems to me like the 20 first lines in the two terminate clauses are doing the same thing. Maybe they can be combined in a common function?

lib/inets/src/http_client/httpc_handler.erl
```diff
@@ -305,8 +305,9 @@ terminate(normal,
 %% Socket closed remotely
 terminate(normal, 
           #state{session = #session{socket      = {remote_close, Socket},
-                                    socket_type = SocketType, 
-                                    id          = Id}, 
+                                    socket_type = SocketType,
+                                    type        = Type,
+                                    id          = Id},
```
KennethL, line 310: Would be nice if the variable matchings here could be in the same order as on line 334-337, That would make it easier to realize that the two function clauses are almost identical except for the last few lines.

u3s:
@KennethL both comments applied in https://github.com/erlang/otp/pull/7107/commits/8e1dba19eda92e6db7d97775997b7324146f1f24

thank you

PR #7109, opened 2023-04-09
Replace OTPROOT with ERL_ROOTDIR in documentation

sirikid:
I was reading [documentation](https://www.erlang.org/doc/man/code.html#code-path) and found that it mentions `OTPROOT`, but `erl` actually [uses](https://github.com/erlang/otp/blob/91b9369f320ae04de0bcff07cd0d4e5ba9a34414/erts/etc/unix/erl.src.src#L39) `ERL_ROOTDIR`.


rickard-green:
`$OTPROOT` is used as a placeholder for describing where an OTP installation is installed in the documentation. Even though `$ERL_ROOTDIR` is an actual environment variable used, it is not documented and not intended to be set by the user. That is, we do not want to replace `$OTPROOT` in the documentation with `$ERL_ROOTDIR`.

PR #7110, opened 2023-04-10
[RFC] Allow dist input_handler to be added if the dist control is an internal port

zzydxm:
Currently dist input_handler can only be set if dist control is a pid. However, in some case we might want to intercept incoming traffic and do some security check. Allowing dist input_handler to be added in port case would let us achieve this without sacrificing performance while using kTLS in dist. This change won't have impact on any current usage of input_handler.

After this change, if we set socket option to {deliver, term} after handshake, all data will be send to the port owner, and port owner can make dist_ctrl_put_data after validation.

We can also make it work for {deliver, port} socket option, but need a bit more change in driver_output2 in io.c

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -1927,16 +1927,22 @@ end</code>
 	  channel.
 	</p>
 	<note><p>
-	  Only the process registered as distribution
-	  controller for the distribution channel identified by
-	  <c><anno>DHandle</anno></c> is allowed to call this
-	  function.
+          When the distribution controller for the distribution
+          channel identified by <c><anno>DHandle</anno></c> is a
+          process, it is the only process allowed to call this
+          function.
+          This function is also allowed to be called when the
+          distribution controller for the distribution channel
+          identified by <c><anno>DHandle</anno></c> is a port.
+          The data received by the port should in this case be delivered to
+          the process identified by <c><anno>InputHandler</anno></c>
+          which in turn should call
+          <seemfa marker="erlang#dist_ctrl_put_data/2"><c>erlang:dist_ctrl_put_data/2</c></seemfa>.
 	</p></note>
 	<p>
 	  This function is used when implementing an alternative
-	  distribution carrier using processes as distribution
-	  controllers. <c><anno>DHandle</anno></c> is retrieved
-	  via the callback
+	  distribution carrier. <c><anno>DHandle</anno></c> is
+	  retrieved via the callback
 	  <seeguide marker="erts:alt_dist#hs_data_f_handshake_complete"><c>f_handshake_complete</c></seeguide>.
 	  More information can be found in the documentation of
 	  <seeguide marker="erts:alt_dist#distribution_module">ERTS
@@ -1967,9 +1973,8 @@ end</code>
 	</p></note>
 	<p>
 	  This function is used when implementing an alternative
-	  distribution carrier using processes as distribution
-	  controllers. <c><anno>DHandle</anno></c> is retrieved
-	  via the callback
+	  distribution carrier. <c><anno>DHandle</anno></c> is
+	  retrieved via the callback
 	  <seeguide marker="erts:alt_dist#hs_data_f_handshake_complete"><c>f_handshake_complete</c></seeguide>.
 	  More information can be found in the documentation of
 	  <seeguide marker="erts:alt_dist#distribution_module">ERTS
--- a/erts/emulator/beam/dist.c
+++ b/erts/emulator/beam/dist.c
@@ -4386,13 +4386,13 @@ dist_get_stat_1(BIF_ALIST_1)
 BIF_RETTYPE
 dist_ctrl_input_handler_2(BIF_ALIST_2)
 {
-    DistEntry *dep = ERTS_PROC_GET_DIST_ENTRY(BIF_P);
     Uint32 conn_id;
+    DistEntry *dep = erts_dhandle_to_dist_entry(BIF_ARG_1, &conn_id);
 
     if (!dep)
         BIF_ERROR(BIF_P, EXC_NOTSUP);
 
-    if (erts_dhandle_to_dist_entry(BIF_ARG_1, &conn_id) != dep)
+    if ((ERTS_PROC_GET_DIST_ENTRY(BIF_P) != dep) && !is_internal_port(dep->cid))
         BIF_ERROR(BIF_P, BADARG);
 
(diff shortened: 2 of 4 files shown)


zzydxm:
@rickard-green This is the solution we found for doing dist security check with minimum OTP change and without hurting performance a lot.

Previously we tried socket nif with kTLS in dist. However it does not work very well. Both sender and receiver are much more expensive, lots of time spending on erl_process scheduling and it gets bad when there are lots of connections. We tried many optimizations but the performance is still far from the inet_drv case.

rickard-green:
I think this is ok. I don't see a need for implementing this for `{deliver,port}`.  Test case(s) and updated documentation should be enough.

zzydxm:
> I think this is ok. I don't see a need for implementing this for `{deliver,port}`. Test case(s) and updated documentation should be enough.

Thanks! I will work on it.

zzydxm:
Updated a test case. Where would you suggest to add the documentation?

rickard-green:
The documentation of the `erlang:dist_ctrl_input_handler/2` BIF.

zzydxm:
Updated the doc also

erts/doc/src/erlang.xml
```diff
@@ -1927,16 +1927,23 @@ end</code>
...
+          process, it is the only process allowed to call this
+          function.
+          This function is also allowed to be called when the
+          distribution controller for the distribution channel
+          identified by <c><anno>DHandle</anno></c> is a port,
+          for the port owner to be able to call
+          <seemfa marker="erlang#dist_ctrl_put_data/2"><c>erlang:dist_ctrl_put_data/2</c></seemfa>
+          when the port is using
+          <seemfa marker="kernel:inet#setopts/2"><c>{deliver, term}</c></seemfa>
+          option.
```
rickard-green, line 1941:
```suggestion
          This function is also allowed to be called when the
          distribution controller for the distribution channel
          identified by <c><anno>DHandle</anno></c> is a port.
          The data received by the port should in this case be delivered to
          the process identified by <c><anno>InputHandler</anno></c>
          which in turn should call
          <seemfa marker="erlang#dist_ctrl_put_data/2"><c>erlang:dist_ctrl_put_data/2</c></seemfa>.
```

Using the inet-driver is a special case of using a port as distribution controller. You may use any driver you want and the only requirement is that the data received is delivered to the runtime system via the input handler process.

RaimoNiskanen:
I have a number of objections regarding how the input controller is hooked in: `application:get_env(kernel, dist_hs_data_finalize_fun)`.
1)  This is a tweak on the SSL distribution module so it should not be a `kernel` variable.
2)  I do not like that it is a `fun(_)`; that would be impossible to set e.g from a configuration file.
3)  The `fun(_)` has to assume way too much about the internals of the `#hs_data{}` record such as `Driver=inet|inet6`, it should not call `gen_tcp`, it would not work if you combine `ktls=false` with `kernel dist_hs_data_finalize_fun`.
4)  Whatever the hook is it should be documented so others may use this feature.

So is this the right abstraction level for the input handler?  Would it be better to implement the spawn and controlling process in `inet_tls_dist` and then have an SSL application variable `{M,F,A}` that is the entry point for the input handler?  Or would it be even better to register a wrapper for `erlang:dist_ctrl_put_data/2` that `inet_tls_dist` (and also `ssl_gen_statem`) calls.  Such a wrapper could also have a state variable to work on.

zzydxm:
Thanks for the comments!

> This is a tweak on the SSL distribution module so it should not be a kernel variable.

I can also add it to other dist modules as well as test cases, it will work in the same way. I found replacing HSData to be very useful when debugging and developing dist protocols.

> I do not like that it is a fun(_); that would be impossible to set e.g from a configuration file.

This can be done by using `fun some_module:some_function/1`, it works in configuration file as well as init arguments. (e.g. `erl -kernel dist_hs_data_finalize_fun "fun some_module:some_function/1"`). But we can change it to be {Module,Function}.

> 3. The `fun(_)` has to assume way too much about the internals of the `#hs_data{}` record such as `Driver=inet|inet6`, it should not call `gen_tcp`, it would not work if you combine `ktls=false` with `kernel dist_hs_data_finalize_fun`.

I only define the dist_hs_data_finalize_fun in one test case of ssl test SUITE, does the function need to be generalized for other use cases? It is supposed to be a customized function and non-ktls tests won't need a customized HSData.

> 4. Whatever the hook is it should be documented so others may use this feature.

Sure I will update it after we make a decision here.

> So is this the right abstraction level for the input handler? Would it be better to implement the spawn and controlling process in `inet_tls_dist` and then have an SSL application variable `{M,F,A}` that is the entry point for the input handler? Or would it be even better to register a wrapper for `erlang:dist_ctrl_put_data/2` that `inet_tls_dist` (and also `ssl_gen_statem`) calls. Such a wrapper could also have a state variable to work on.

Do you mean move ktls_input_handler from test SUITE to inet_tls_dist and let user customize a MFA to be run in ktls_input_handler_loop? It works for us but seems will make the usage too restricted for others?

RaimoNiskanen:
Let's get on the same page first; what is the purpose of the config variable `dist_hs_data_finalize_fun`?

zzydxm:
By changing dist_hs_data_finalize_fun, people can
1. use input handler in inet_tls_dist/inet_tcp_dist(if we add it also), define customized receiver function and do anything they want for received data.
2. set socket options, dist flags, do garbage collections, or treat different dist connection differently without copying the inet_tls_dist/inet_tcp_dist module.
3. or rather, this is a functionality added to create that input_handler test case in ssl_dist_SUITE, we can copy the dist protocol locally and do the HSData modification locally on our dist protocol. So if you have other suggestion to modify the test we can also remove the dist_hs_data_finalize_fun

RaimoNiskanen:
Do you use it for your implementation, or have you got a dedicated distribution module for that?

Exposing so very internal stuff like this with the purpose of it being a user interface, documented and so on,
would be a bad idea, in my opinion.  It would hinder future rewrites, and not make a good API.

But if this is a test only feature it might be acceptable...

zzydxm:
Yeah I'm taking it as a test only feature. We have our own dist protocol module.

RaimoNiskanen:
I have a bunch of TLS distribution modules in `ssl_dist_bench_SUITE`, using a simplified internal framework.  I will try to see if a test can be added there instead, reasonably easy, without introducing a hook in the regular TLS distribution module.

Stay tuned...

zzydxm:
That looks good, thanks a lot!

rickard-green:
I squashed some commits and force pushed the result.

PR #7112, opened 2023-04-11
erl_interface warning fixes

frej:


erts/etc/common/escript.c
```diff
@@ -286,7 +286,7 @@ find_prog(char *origpath)
                     beg = end + 1;
                     continue;
                 }
-                strncpy(dir, beg, sz);
+                mempcpy(dir, beg, sz);
```
rickard-green, line 289:
```suggestion
                memcpy(dir, beg, sz);
```
After googleing a bit is seems that `mempcpy()` have portability issues and `memcpy()` is identical if the return value is ignored.

frej: Considering that `mempcpy` is a typo on my part, I 100% agree that `memcpy` is what it should be, do you want a refreshed patch @rickard-green ?

rickard-green: You can squash the commit that I added into your commit.

frej: Did that and refreshed the patches to current master. I also fixed some errors in the commit message, I wonder what I was thinking when I wrote it, the fix we're discussing is to escript and not erl_interface as the commit message claimed. :)

rickard-green: 👍

PR #7117, opened 2023-04-13
List additional potential incompatibilities

josevalim:
List the changes discussed in #6683 and #6693
as upcoming incompatibilities.

See https://github.com/erlang/otp/pull/6836#issuecomment-1433215809 for more context.


josevalim:
Note this should be applied to maint/OTP 26. I can rebase but you have my full permission to just copy and paste the contents and place wherever you want with no credits whatsoever. :D

PR #7118, opened 2023-04-13
Optimise accesses to the Table name -> Table id map

RobinMorisset:
Currently these accesses are guarded by (striped) read-write locks.

With this patch, there are still (regular) locks for protecting writing threads from each other, but threads that only need to read that map (so the overwhelming majority) no longer need to use any lock. Instead we now rely on lightweight atomic synchronisation, and on the thread_progress functionality for memory reclaimation.

One ugly hack is that if we add 1 table, then remove it, the corresponding bucket may go
- inline, empty
- inline, with a value
- outlined vector without a value

Once we have a filled the inline vector, we can only transition to an outlined vector, and never go back to the inline state. This saves us from an ABA hazard:
- Reader sees inline vector of size 1
- Reader reads the name NameA that corresponds to table IdA, sees it matches what it looked for
- Writer deletes (or rename the table), moving us to an empty vector
- Writer inserts another table in that bucket, with a different name: NameB -> IdB
- Reader reads IdB, concluding that NameA -> IdB. As long as we stick to outline vectors and always allocate a fresh one when deleting tables, we don't have that risk, as the vectors only get freed (and thus the adress may get reused by another vector) once all readers have made thread progress (and thus are no longer in the middle of an operation).

One consequence of this hack, is that we must lie to the memory counting system, pretending to have an outlined vector at all times, even when we only have the initial inlined vector. Otherwise, the tests detect the memory increase in inline (1 element) -> outlined (0 element) -> outlined (1 element), and report a memory leak.

This patch was motivated by noticing that heavyweight services can spend several % of CPU time in db_get_table_aux. A trial in production confirmed that this patch roughly halves the time spent in that function.


RobinMorisset:
ping @sverker

lpgauth:
@RobinMorisset would be nice to have in OTP 27 :)

erts/emulator/beam/erl_db.c
```diff
@@ -783,57 +807,35 @@ DbTable* db_get_table_aux(Process *p,
...
-            if (name_lck)
-                erts_rwmtx_runlock(name_lck);
             return tb;
         }
 
         DB_HASH_ADAPT_NUMBER_OF_LOCKS(tb);
-	db_lock(tb, kind);
-        if (name_lck)
-            erts_rwmtx_runlock(name_lck);
+        db_lock(tb, kind);
```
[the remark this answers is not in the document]
RobinMorisset: I've tried getting this to work and I'm hitting a small issue.
The lock is not always taken in db_get_table_aux(): if `what == DB_READ_TBL_STRUCT`, then there is an early return before the lock is taken. The double-check might still work, but I'm struggling to figure out how to make it safe, and ensure that the table does not change under us. Apart from that I've got a patch with the relevant changes (removing the hack for size, and doing an atomic write of 0 to the inline_entry size when removing the only table in it (rather than allocating an empty outlined entry).

RobinMorisset:
This new version passes tests, so it should not be completely broken, but I've put the double-checking before the locking because the lock is not always taken (see my comment), so I'm not very confident yet that this last change is correct.

erts/emulator/beam/erl_db.c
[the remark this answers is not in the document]
sverker: You should double check the name **after** locking the table. Otherwise the table op may be seen as happening after the table was renamed.

```
ets:rename(aaa, bbb),
ets:insert(bbb, {key, bbb})
```
A concurrent process doing `ets:lookup(aaa, key)` should never get `{key,bbb}` if it never existed in the tabled when it was named `aaa.`

For `DB_READ_TBL_STRUCT` it's enough to do the check without lock. Just to make sure the table got the correct name _now._ `DB_READ_TBL_STRUCT` is only used by yielding `ets:insert` as a first cheap lookup of the table. It will lock the table later and then also double check the name. See comment in function `ets_insert_2_list_lock_tbl`.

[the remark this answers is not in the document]
RobinMorisset: Thanks for the explanation, I added it to the comment (and moved the double-check to after taking the lock).

erts/emulator/beam/erl_db.c
```diff
@@ -783,57 +807,57 @@ DbTable* db_get_table_aux(Process *p,
...
-	}
+        }
     }
 
     if (tb) {
         if (what == DB_READ_TBL_STRUCT) {
-            if (name_lck)
-                erts_rwmtx_runlock(name_lck);
             return tb;
         }
```
sverker, line 835: I think you must check the name for DB_READ_TBL_STRUCT as well. Otherwise we may start to operate on a completely random wrong table which feels a bit scary. At least I think it can result in incorrect errors if the random table has a different `keypos` for example.

erts/emulator/beam/erl_db.c
```diff
@@ -902,142 +926,164 @@ static DbTable* db_get_table_or_fail_return(Binary* btid,
...
+    ERTS_ETS_MISC_MEM_ADD(-alloc_size);
+
+    erts_free(ERTS_ALC_T_DB_NTAB_ENT, ptr);
+}
+
+static ERTS_INLINE
+void schedule_meta_name_tab_entries_for_deletion(struct meta_name_tab_entries *entries)
+{
+    char* ptr = (char *) entries;
+    ptr -= sizeof(ErtsThrPrgrLaterOp);
```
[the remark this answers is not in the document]
RobinMorisset: Thanks for the suggestion, I did not know about ErtsContainerStruct and it is indeed much cleaner.

RobinMorisset:
I've checked and the new test catches that bug (triggers the segfault, which disappears with the fix). Thanks again for the careful review!

sverker:
Did you get the test case to trigger the race as it is, without `erts_thr_yield` and with only 1000 iterations?

RobinMorisset:
I got it to trigger the race as it is, but I had 1M iterations at the time. I reduced the number of iterations because it takes forever to run otherwise now that the race is fixed.

RobinMorisset:
Testing some more, 1k iterations does not trigger the bug, but 10k iterations is enough to trigger it reliably (2 out of 2 runs).
But once I fix the bug, 10k iterations takes forever (to the point I'm wondering whether it is triggering some kind of deadlock or livelock or something).

RobinMorisset:
I found the issue: I was missing a db_unlock in the double-check branch!
The test now runs fine (and very fast) with the fix, and immediately blows up if I re-introduce the segfault in that branch, confirming that it is being tested.

sverker:
You can also use undocumented `spawn_opt `option `{scheduler,N}` to make sure writer and reader are spawned on different schedulers.

sverker:
You should always also run debug built emulator when developing. That will detect the missing unlock.

RobinMorisset:
> You should always also run debug built emulator when developing. That will detect the missing unlock.

How do you ensure that the debug version is used by the tests? I've just been following https://github.com/erlang/otp/blob/master/HOWTO/TESTING.md and doing
```
make stdlib_test ARGS="-suite ets_SUITE -case racy_rename"
```

sverker:
I don't know if there is a way to select emulator type when running tests like that.

I use the old way by releasing the tests:
https://github.com/erlang/otp/blob/master/HOWTO/TESTING.md#releasing-tests

Then I can start whatever `erl` I want and run tests with `ts:run`.

```
$ERL_TOP/bin/cerl -debug
1> ts:install().
2> ts:run(stdlib,ets_SUITE,racy_rename,[batch]).
```

RobinMorisset:
I've just tried that, and `ts:install()` fails with `** exception error: undefined function ts:install/0`.
Is there an additional command I should run to bring `ts` in scope?

dgud:
You need to stand in the test_server directory.

The more modern variant is using common test:   (though you will still need test_server in the path).
```
cd  TEST_INSTALL_PATH/test/app/
erl -pa /TEST_INSTALL_PATH/test/test_server -sname foo EXTRA_ARGS
> ct:run_test([{suite, ets_SUITE}, {testcase,racy_rename}]).
```

RobinMorisset:
What is TEST_INSTALL_PATH supposed to be?
I tried doing `$ERL_TOP/bin/cerl -debug -pa ../../tests/test_server` in `$ERL_TOP/release/tests/stdlib_test` and then calling `ct:run_test([{suite, ets_SUITE}, {testcase,racy_rename}]).` in the resulting shell, but it gave me an error message "Failed to start CTH, see the CT Log for details" (with no information on how to find that log) and "*** FAILED {ets_SUITE,init_per_suite} ***".
I also tried ts:install(), and it apparently found it, but it gave another error message: {error,no_configure_script}

Sorry for all the questions, I generally find the test and build system rather mystifying and I'd like to figure it out for future PRs.

dgud:
`-pa ../../tests/test_server`  This needs to be an absolute path for a to me unknown reason.

RobinMorisset:
It worked! Thanks for the help.
Unfortunately, the race that reproduces every few thousand iterations in release mode stubbornly refuses to reproduce in debug mode, even with 1M iterations. Still it is good to know how to run tests in debug mode.

sverker:
> It worked! Thanks for the help. Unfortunately, the race that reproduces every few thousand iterations in release mode stubbornly refuses to reproduce in debug mode, even with 1M iterations. Still, it is good to know how to run tests in debug mode.

That is not surprising. The debug VM does a lot of extra checks that can radically change the timing.

garazdawi:
```
make TYPE=debug
make stdlib_test TYPE=debug ARGS="-suite ets_SUITE -case racy_rename"
```

as per https://github.com/erlang/otp/blob/master/HOWTO/DEVELOPMENT.md#types-and-flavors

lpgauth:
@RobinMorisset bumping this is case you have time to complete this PR :) thanks

RobinMorisset:
@sverker I just rebased to fix the trivial conflict in the test suite. I've also re-measured the performance impact of this PR in production, this time on more and larger machines, and it is still reliably a 1.5% win. Is there any blocker left to merging this?

sverker:
@RobinMorisset Sorry for letting this linger. I need to schedule time to do a proper review and assure myself that all the necessary memory barriers are in place without race conditions.
What kind of benchmark was the 1.5% win? An ETS heavy "real" application or a synthetic micro benchmark pounding on name lookup and maybe table renames/create/deletes?

Also, I think John did something similar during a profiling and optimization session. Should maybe look at what he did, if it still exists.

RobinMorisset:
It was a very "real" application: our largest service in production.
I measured the 1.5% win in two ways: comparing 4 machines with the change to 4 machines without at very high throughput (~90% cpu usage, with the 4 machines with the change having between 1 and 2% less cpu usage overall), and looking at profiles (db_get_table falls from 3.3% of total cpu usage to 1.8%)

erts/emulator/beam/erl_db.c
```diff
@@ -904,142 +940,160 @@ static DbTable* db_get_table_or_fail_return(Binary* btid,
...
-    
-	}
+    if (index == size) {
+        goto done;
+    }
+
+    if (entries == &bucket->inline_entry) {
+        // We don't need to do anything to the actual entry: if a reader thread sees size=0 it won't look further.
+        // And if it saw size = 1, then it is as if it had fully run before us.
+        erts_atomic32_set_wb(&bucket->inline_entry.size, (erts_aint_t) 0);
```
sverker: [nitpick] I think this write barrier in `remove_named_tab()` is pointless. As this is the only write we do here, there are no other writes to order against.

erts/emulator/beam/erl_db.c
```diff
@@ -904,142 +940,160 @@ static DbTable* db_get_table_or_fail_return(Binary* btid,
...
+    } else {
+        // Trying to remove an entry from the vector without reallocation
+        // while reader threads concurrently access it would be a nightmare.
+        // So we just move things to a new vector instead
+        if (size - 1 > 1) {
+            new_entries = alloc_meta_name_tab_entries(size - 1);
+        } else {
+            new_entries = &bucket->inline_entry;
```
sverker, line 1098: Also in remove_named_tab():
When shrinking back to the `inline_entry,` I assume we rely on the size still being 1 (not 0) since the last use. If so, an ASSERT would be nice to make that assumption clear.

erts/emulator/beam/erl_db.c
RobinMorisset: I think I added it here (and in insert_named_tab) to order against changes to the table itself. For example:
Thread A:
- remove named table
- write some value into that (now unnamed table)

Thread B:
- reads from the table through its name.

I don't think B should be able to see the value that was written by A, and I'm not sure it is prevented by anything if we remove that barrier.

erts/emulator/beam/erl_db.c
RobinMorisset: I'll add it.

erts/emulator/beam/erl_db.c
sverker: Both when renaming and writing to the table, table locks will be seized. That will imply memory barriers enough, I think.

RobinMorisset:
I've spent a while thinking about the memory barrier in that path in remove_named_tab, and I agree that it does not seem required, so I removed it, along with adding the ASSERT you requested and fixing the formatting of my overly long lines.

erts/emulator/beam/erl_db.c
```diff
@@ -904,142 +946,168 @@ static DbTable* db_get_table_or_fail_return(Binary* btid,
...
-	}
+    if (index == size) {
+        goto done;
+    }
+
+    if (entries == &bucket->inline_entry) {
+        // We don't need to do anything to the actual entry: if a reader thread
+        // sees size=0 it won't look further.
+        // And if it saw size = 1, then it is as if it had fully run before us.
+        ASSERT(erts_atomic32_get_nob(&bucket->inline_entry.size) == 1);
```
sverker: Compile with `debug` target and fix the error on this line.  😀
`get` -> `read`

sverker: Also, the case when I thought an ASSERT would be nice is a few lines down where we shrink from 2 to 1 entry and do:
`new_entries = &bucket->inline_entry;`
There we do not set `inline_entry.size` and just assume it's still 1.

sverker: @RobinMorisset I pushed a fixup to my own comments above.

RobinMorisset: oops, sorry for missing those, and thanks for the fixup.

sverker:
@RobinMorisset I was about to squash all into one commit for merge but noticed your initial commit message is a bit out of date, I think. So, I leave the squashing honors to you if you have some time.

sverker:
@RobinMorisset Squashed and cleaned up commit message. I removed all the talk about the "ugly hack" which seemed irrelevant now.

PR #7124, opened 2023-04-14
crypto: dh.c change len type to ErlNifUInt64

u3s:
- fix for "Can't do fromdata" observed on Windows builds
- and ssh tests

### DIFF
--- a/lib/crypto/c_src/dh.c
+++ b/lib/crypto/c_src/dh.c
@@ -41,7 +41,7 @@ ERL_NIF_TERM dh_compute_key_nif(ErlNifEnv* env, int argc, const ERL_NIF_TERM arg
 /* Has 3_0 */
 ERL_NIF_TERM dh_generate_key_nif(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 {/* (PrivKey|undefined, DHParams=[P,G], 0, Len|0) */
-    unsigned long len = 0;
+    ErlNifUInt64 len = 0;
     int i = 0;
     OSSL_PARAM params[8];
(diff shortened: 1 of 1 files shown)


lib/crypto/c_src/dh.c
```diff
@@ -270,7 +270,7 @@ ERL_NIF_TERM dh_generate_key_nif(ErlNifEnv* env, int argc, const ERL_NIF_TERM ar
     BIGNUM *dh_p_shared;
     BIGNUM *dh_g = NULL;
     BIGNUM *priv_key_in = NULL;
-    unsigned long len = 0;
+    ErlNifUInt64 len = 0;
```
sverker, line 273: This `len` can stay `unsigned long` I think. It's never used as a 64-bit integer.

lib/crypto/c_src/dh.c
```diff
@@ -304,7 +304,7 @@ ERL_NIF_TERM dh_generate_key_nif(ErlNifEnv* env, int argc, const ERL_NIF_TERM ar
     if (mpint != 0 && mpint != 4)
         assign_goto(ret, err, EXCP_BADARG_N(env, 2, "Integer 0 or 4 expected"));
 
-    if (!enif_get_ulong(env, argv[3], &len))
+    if (!enif_get_uint64(env, argv[3], &len))
```
sverker, line 307: ..and then you would have to revert this one also

sverker (approved):

lib/crypto/c_src/dh.c
u3s: You're right.

dgud (approved):

PR #7125, opened 2023-04-16
Remove Vanilla Drivers

### LINKED ISSUE #6725: Ports being initialized through Vanilla Drivers could lead to BEAM failures.
**Describe the bug**
Erlang/OTP still supports [Ports creation by using Vanilla Drivers](https://erlangforums.com/t/ports-communication-through-ipc-safe-to-use-or-likely-to-be-deprecated/2227/2) that are called when you use the first argument as an Atom/String.

This method was deprecated since the early 2000's and got all the documentation removed around 2006, as you can see in the picture below.

Screen Shot 2023-01-23 at 16 46 00

The main problem with this approach is that Vanilla Drivers only do a syscall to open file, doing so can lead to BEAM failures, since it doesn't have any treatment for special files

**To Reproduce**
You can access any file through `open_port/2` by passing a String as first argument.

```
open_port("/dev/urandom", [eof]).
```

**Expected behavior**
Ports usage should only be accepted with tuple arguments like `{spawn, "cat file.txt"}`

Vanilla Drivers shouldn't exist for this case, resulting in a badarg.
```
> open_port("something", [eof]).
badarg error
```

**Affected versions**
It happen on all versions.

This is something I would like to tackle in case you agree that this behavior shouldn't exist any longer. Ideas on that?

WLSF:
This PR removes all code related to Vanilla Drivers port creation, add tests to validate that both Atoms and Strings as 1st argument on `open_port/2` should raise badarg error.

The changes in this PR resolves #6725 so that ports could only be used with tuple args and file manipulation as the way vanilla drivers used to do should be accomplished through [File module](https://www.erlang.org/doc/man/file.html)

### DIFF
--- a/erts/emulator/beam/erl_bif_port.c
+++ b/erts/emulator/beam/erl_bif_port.c
@@ -1,7 +1,7 @@
 /*
  * %CopyrightBegin%
  *
- * Copyright Ericsson AB 2001-2022. All Rights Reserved.
+ * Copyright Ericsson AB 2001-2023. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


WLSF:
Should be good to review by now, let me know if theres anything that could be improved here.

PR #7126, opened 2023-04-17
Update `perf` workflow for OTP26

mkuratczyk:
Scheduler names changed in OTP26, so we need a different `sed` expresion to merge data from all schedulers. The example is now compatible with versions prior to OTP26 and OTP26+


jhogberg:
Merged, thanks for the PR!

Issue #7140, opened 2023-04-21
lcnt crashes when collecting data – no function clause lcnt:creation_bits(120)


stolen:
**Describe the bug**
LCNT crashes on OTP26rc3 when collecting data

**To Reproduce**
  * `$ ./configure --enable-lock-counter`
  * `$ ./bin/erl -emu_type lcnt`
  * `> lcnt:clear(), lcnt:collect().`

```
...
** When Server state == {state,[],0}
** Reason for termination ==
** {function_clause,[{lcnt,creation_bits,"x",[{file,"lcnt.erl"},{line,916}]},
                     {lcnt,clean_id_creation,1,[{file,"lcnt.erl"},{line,895}]},
                     {lcnt,locks2records,1,[{file,"lcnt.erl"},{line,855}]},
                     {lcnt,locks2records,1,[{file,"lcnt.erl"},{line,857}]},
                     {lcnt,data2state,2,[{file,"lcnt.erl"},{line,847}]},
                     {lcnt,handle_call,3,[{file,"lcnt.erl"},{line,549}]},
                     {gen_server,try_handle_call,4,
                                 [{file,"gen_server.erl"},{line,1113}]},
                     {gen_server,handle_msg,6,
                                 [{file,"gen_server.erl"},{line,1142}]}]}
** Client <0.85.0> stacktrace
** [{gen,do_call,4,[{file,"gen.erl"},{line,240}]},
    {gen_server,call,3,[{file,"gen_server.erl"},{line,397}]},
    {erl_eval,do_apply,7,[{file,"erl_eval.erl"},{line,750}]},
    {shell,exprs,7,[{file,"shell.erl"},{line,780}]},
    {shell,eval_exprs,7,[{file,"shell.erl"},{line,736}]},
    {shell,eval_loop,4,[{file,"shell.erl"},{line,721}]}]

** exception exit: {{function_clause,[{lcnt,creation_bits,"x",
                                            [{file,"lcnt.erl"},{line,916}]},
                                      {lcnt,clean_id_creation,1,[{file,"lcnt.erl"},{line,895}]},
                                      {lcnt,locks2records,1,[{file,"lcnt.erl"},{line,855}]},
                                      {lcnt,locks2records,1,[{file,"lcnt.erl"},{line,857}]},
                                      {lcnt,data2state,2,[{file,"lcnt.erl"},{line,847}]},
                                      {lcnt,handle_call,3,[{file,"lcnt.erl"},{line,549}]},
[...]

PR #7141, opened 2023-04-21
LCNT: fix port parser for V4_PORT_EXT

stolen:
See #7140 

After commit e6074fada29f7ec8a8ffd28e6a3aab75f74199f5, term_to_binary always uses V4_PORT_EXT for port encoding. That led to lcnt crashes in clean_id_creation/1 function.

This PR makes lcnt aware of V4_PORT_EXT syntax.


jhogberg:
Merged, thanks for the PR!

PR #7143, opened 2023-04-21
beam_ssa_private_append: Fix crash on oddly structured code

jhogberg:
The private append pass didn't expect to append anything to literals other than `<<>>`. Rare interactions with the type or bool passes could sneak in things like atoms in unexpected places, crashing the pass.

@frej

### DIFF
--- a/lib/compiler/src/beam_ssa_private_append.erl
+++ b/lib/compiler/src/beam_ssa_private_append.erl
@@ -565,6 +565,8 @@ patch_literal_term(<<>>, self, Cnt0) ->
     {V,Cnt} = new_var(Cnt0),
     I = #b_set{op=bs_init_writable,dst=V,args=[#b_literal{val=256}]},
     {V, [I], Cnt};
+patch_literal_term(Lit, self, Cnt) ->
+    {#b_literal{val=Lit}, [], Cnt};
 patch_literal_term([H0|T0], {hd,Element}, Cnt0) ->
     {H,Extra,Cnt1} = patch_literal_term(H0, Element, Cnt0),
     {T,[],Cnt1} = patch_literal_term(T0, [], Cnt1),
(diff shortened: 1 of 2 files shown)


lib/compiler/src/beam_ssa_private_append.erl
```diff
@@ -565,6 +565,8 @@ patch_literal_term(<<>>, self, Cnt0) ->
     {V,Cnt} = new_var(Cnt0),
     I = #b_set{op=bs_init_writable,dst=V,args=[#b_literal{val=256}]},
     {V, [I], Cnt};
+patch_literal_term(Lit, self, Cnt) ->
```
frej, line 568: This fixes the symptom, but it's not the right fix as we should already have discovered that this isn't possible, i.e. `add_literal/3` should never get to add the "patchee" to the set of literals to patch.

For the work I'm currently doing I have refactored `add_literal/3` to do a validity check, so we never add something which later on turns out to be impossible. I could try to backport that work, but as it's a larger patch, this is probably correct for R26.

frej (approved):
LGTM if @jhogberg  agrees that this small change is the right thing this late in the release process.

lib/compiler/src/beam_ssa_private_append.erl
jhogberg: I'd rather keep things small this late in the release process, but then again it might end up biting us later. How large do you think the `add_literal/3` change would be?

frej: It will require adding additional arguments to `add_literal/3` and thus changing all call sites. 

Considering that the only drawback of your fix is that we may keep too many literals around and spend some time trying to patch them before giving up, I'm perfectly happy with your patch.

The no-non-patchable-literals refinement can wait until R27 and the added clause will show up as not visited while running coverage, so there is little risk that it will remain when merging the refined stuff.

jhogberg: Thanks, I'll go ahead with this then :)

PR #7144, opened 2023-04-21
os_mon: Fix failing port_close

sverker:
Use asynchronous close message to not fail on already closed port.

### DIFF
--- a/lib/os_mon/src/cpu_sup.erl
+++ b/lib/os_mon/src/cpu_sup.erl
@@ -613,8 +613,8 @@ port_server_loop(Port, Timeout) ->
 
 	% Close port and this server
 	{Pid, ?quit} ->
-	    port_command(Port, ?quit),
-	    port_close(Port),
+            Port ! {self(), {command, ?quit}},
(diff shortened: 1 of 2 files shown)


lib/os_mon/src/cpu_sup.erl
```diff
@@ -613,8 +613,8 @@ port_server_loop(Port, Timeout) ->
 
 	% Close port and this server
 	{Pid, ?quit} ->
-	    port_command(Port, ?quit),
-	    port_close(Port),
+            Port ! {self(), {command, ?quit}},
```
max-au, line 616: Does this simply ignore exceptions previously fired? And in that case, would it be safer (in terms of not exposing internals) to just add "catch"?

sverker: It will ignore the synchronous badarg error raised by `port_close` if the port has already been closed. The bang operator is asynchronous in that respect and does not fail for an already closed port.
Adding a `catch` would also catch and disguise other errors (like `Port` being of wrong type). I don't get what you mean by exposing internals.

PR #7154, opened 2023-04-24
dialyzer: Expand Dialyzer config file options

TD5:
The Dialyzer config file previously allowed setting fallback modules to analyse and report warnings for in the absence of explicitly given modules in the arguments.

This change extends the config file to include warnings to toggle and modifications to the beginning and end of the path, allowing users to factor these out of their command line arguments.

This change also fixes apps being in default_warning_apps without being in default_apps causing an inconsistency. Now all warning apps are implicitly added to the apps to analyze.

An example Dialyzer config that makes use of the new options:

```erlang
  {incremental,
    {default_apps,[stdlib,kernel,erts]},
    {default_warning_apps,[stdlib]}
  }.
  {warnings, [no_improper_lists]}.
  {add_pathsa,["/users/samwise/potatoes/ebin"]}.
  {add_pathsz,["/users/smeagol/fish/ebin"]}.
```

This config expresses that:
 - If no apps/warnings apps are given explicitly, analyse stdlib, kernel and erts, and report warnings for stdlib only.
 - Set the no_improper_lists option to suppress warnings about improper list construction
 - Add /users/samwise/potatoes/ebin to the start of the path, and /users/smeagol/fish/ebin to the end of it

This change should make incremental mode suitable as the default analysis mode for Dialyzer in OTP 27 and beyond.

### DIFF
--- a/lib/dialyzer/doc/src/dialyzer.xml
+++ b/lib/dialyzer/doc/src/dialyzer.xml
@@ -507,6 +507,32 @@ dialyzer --plts plt_1 ... plt_n -- files_to_analyze</code>
     <p>Currently the only option used is the
     <seeerl marker="#error_location"><c>error_location</c></seeerl> option.
     </p>
+
+    <p><em>Dialyzer configuration file:</em></p>
+
+    <p>Dialyzer's configuration file may also be used to augment the default
+       options and those given directly to the Dialyzer command. It is commonly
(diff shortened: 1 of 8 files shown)


bjorng (approved):
Looks good to me. Added to our daily builds.

bjorng:
Thanks for your pull request.

PR #7169, opened 2023-04-25
kernel, stdlib: multiline editing in the shell

frazze-jobb:
-Support inputting expressions that span multiple lines
  in erl shell.
-Adding new keybindings:
  navigate up (ctrl+up)/(alt+up)
  navigate down (ctrl+down)/(alt+down)
  insert newline in middle of line (alt+enter)
  navigate top (alt+<)/(alt+shift+up)
  navigate bottom (alt+>)/(alt+shift+down)
  clear current expression (alt+c)
  cancel search (alt+c)
-Modifies the prompt for new lines to make it clearer
  that the prompt has entered multiline mode.
-Modifies the search prompt to support multiline statements.
-Including keybinding for opening editor on mac
  (option+o)/(alt+o)
-Redraw the prompt after continuing from jcl menu.

### DIFF
--- a/lib/kernel/src/group.erl
+++ b/lib/kernel/src/group.erl
@@ -470,50 +470,56 @@ get_chars_n(Prompt, M, F, Xa, Drv, Shell, Buf, Encoding) ->
     Pbs = prompt_bytes(Prompt, Encoding),
     case get(echo) of
         true ->
-            get_chars_loop(Pbs, M, F, Xa, Drv, Shell, Buf, start, Encoding);
+            get_chars_loop(Pbs, M, F, Xa, Drv, Shell, Buf, start, [], Encoding);
         false ->
             get_chars_n_loop(Pbs, M, F, Xa, Drv, Shell, Buf, start, Encoding)
     end.
 
 get_chars_line(Prompt, M, F, Xa, Drv, Shell, Buf, Encoding) ->
     Pbs = prompt_bytes(Prompt, Encoding),
-    get_chars_loop(Pbs, M, F, Xa, Drv, Shell, Buf, start, Encoding).
+    get_chars_loop(Pbs, M, F, Xa, Drv, Shell, Buf, start, [], Encoding).
 
-get_chars_loop(Pbs, M, F, Xa, Drv, Shell, Buf0, State, Encoding) ->
+get_chars_loop(Pbs, M, F, Xa, Drv, Shell, Buf0, State, LineCont0, Encoding) ->
     Result = case get(echo) of
-		 true ->
-		     get_line(Buf0, Pbs, Drv, Shell, Encoding);
-		 false ->
-		     % get_line_echo_off only deals with lists
-		     % and does not need encoding...
-		     get_line_echo_off(Buf0, Pbs, Drv, Shell)
-	     end,
+                 true ->
+                     get_line(Buf0, Pbs, LineCont0, Drv, Shell, Encoding);
+                 false ->
+                     %% get_line_echo_off only deals with lists
+                     %% and does not need encoding...
+                     get_line_echo_off(Buf0, Pbs, Drv, Shell)
+             end,
     case Result of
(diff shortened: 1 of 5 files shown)


frazze-jobb:
@josevalim, could you check if this causes problems with elixir?

josevalim:
Awesome. I will play with it later today. Although we will likely need https://github.com/erlang/otp/pull/6921 to be merged (or similar) so we can still customize our prompt during multi line reads.

frazze-jobb:
I am almost done with handling the bug mentioned above.
Currently I output "lists:duplicate($ , size of prompt)++".. " as the multiline prompt, I will review #6921 next thing

josevalim:
@frazze-jobb my initial testing shows that nothing breaks in Elixir but I haven't changed it yet to use the new features. I will try that soon.

Regarding the keybindings, I have some feedback. If you use tools like Slack, you use shift+enter to enter multiline mode, so that could be handy. I also wonder if we could make all of the navigation based on alt. Here is one proposal:

* New line (alt+enter AND shift+enter)
* Line up (alt+up)
* Line down (alt+down)
* Previous word (alt+left - already works before this PR)
* Next word (alt+right - already works before this PR)
* Beginning of text (shift+alt+up)
* End of text (shift+alt+down)
* Beginning of line (shift+alt+left - already works as ctrl+A)
* End of line (shift+alt+right - already works as ctrl+E)

The idea is that "alt+arrow" does a "small movement" (next word / next line) and "shift+alt+arrow" does a big movement (beginning and end of line / text). At the same time, I know the shell often uses emacs bindings, and I don't how those work there. :)

josevalim:
I have an Elixir branch that integrates with this functionality and now I can navigate multiline expressions. It is beautiful. Multiline history also just works. Based on that, I have two additional points:

1. Could we make it so alt+enter works _anywhere_ and not only in the middle of lines? Elixir does not require line terminators, like Erlang does, so I may want to write `1` (which in itself is a complete expression) but force it to go multiline by using alt+enter

2. Based on how history works, I am not sure if #6921 is the right approach. #6921 works per line and I think that, once you are browsing history, we need the prompt for all lines at once? Unsure.

Thank you for the ping!

frazze-jobb:
Regarding keybindings, I could try to add those that you suggested as a redundancy. We will go with this for now, I am planning to look into https://sw.kovidgoyal.net/kitty/keyboard-protocol/ after 26, so that we can offer custom keybindings. shift+enter currently gives the same key as enter and so does ctrl+enter, only alt+enter is distinguishable. But with kitty we might be able to distinguish all of them.

1) I need to think about this one if I can somehow skip the apply step on alt+enter
2) Yes I would assume that you would need that, or have a function that outputs the prompt differently depending on which line in the expression its being printed for.

josevalim:
Cool, thanks. Regarding the keybindings, are the current ones copied from somewhere? I have been playing with different editors and terminals trying to find a common ground but they are all different. :'(

In any case, I don't think we need to have redundancy. If you have decided on the current one, then I am completely fine with supporting it exclusively. My only suggestion would be to move "ctrl+up" and "ctrl+down" to "alt+up" and "alt+down" so I need to move fewer fingers around when navigating multiline!

frazze-jobb:
Not a problem to add those, I also think its nice to not have to move your fingers as much. Later I am hoping we can support custom keybinding configurations for emacs users, and vscode users. I was able to change the behavior of alt+enter. Will push changes later.

bjorng (commented):
The combination Alt/Option + an arrow key does nothing. Alt+O and other Alt combinations do work.

Otherwise it seems to work fine.

bjorng:
> Option + arrow key or Option+shift + arrow key doesn't work on Mac, you have to add it to your terminal settings for them to work.

It is possible to export Terminal profiles that can be imported into the Terminal application, so we will probably supply a profile that adds the missing keys.

PR #7174, opened 2023-04-26
jit: Refactor sigaltstack(2) handling

jhogberg:
Erlang code compiled to x86 native code uses `RSP` as its stack pointer. This improves performance in several ways:

- It permits the use of the x86 `CALL` and `RET` instructions, which reduces code volume and improves branch prediction.
- It avoids stealing a callee-save register to act as a stack pointer.

Unix signal handlers are by default delivered onto the current stack, i.e. `RSP`. This is a problem since our native-code stacks are small and may not have room for the Unix signal handler.

There is a way to redirect signal handlers to an "alternate" signal stack by using the `SA_ONSTACK` flag with the `sigaction(2)` system call. Unfortunately this has to be specified explicitly for each signal, and it is impossible to enforce given the presence of libraries.

We used to attempt to override the C library's signal handler setup procedure with our own that added the `SA_ONSTACK` flag, but it only worked with `GNU libc` which is not always the current libc. As many of our users liked to run docker images with `Alpine` which uses `musl` instead, they got needlessly bad performance without knowing it.

Instead, we now explicitly add `SA_ONSTACK` to our own uses of `sigaction(2)` and ignore the library problem altogether because:

1. We don't care about this problem on non-scheduler threads: if a library wants to fiddle around with signals on its own threads then it doesn't affect us.
2. We don't care about this problem when executing on the runtime stack: if a NIF or driver uses signals in a creative manner locally during a call, then that's fine as long as they restore them before returning to Erlang code.

   A NIF or driver that doesn't do this is misbehaving to begin with and we can't shield ourselves against that.
3. If a library that we're statically linked to messes around with signals in the initialization phase (think C++ constructors of static objects), all of it will happen before `main` runs and we'll set things straight in `sys_init_signal_stack`.

   If a dynamically linked library does the same, the same restrictions as ordinary NIF/driver calls apply to the initialization phase and the library must restore the signals before returning.

   If any threads are created in either of these phases, they're still not scheduler threads so we don't have to care then either.

### DIFF
--- a/erts/emulator/sys/unix/sys.c
+++ b/erts/emulator/sys/unix/sys.c
@@ -256,9 +256,6 @@ erts_sys_pre_init(void)
     /* After creation in parent */
     eid.thread_create_parent_func = thr_create_cleanup,
 
-    /* Must be done really early. */
-    sys_init_signal_stack();
-
 #ifdef ERTS_ENABLE_LOCK_COUNT
     erts_lcnt_pre_thr_init();
 #endif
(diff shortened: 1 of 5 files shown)


jimdigriz:
`signalfd()`?

jhogberg:
That'd work but is Linux-specific, we still need a solution for the *BSDs and Mac.

jimdigriz:
`kqueue()` apparently is what the *BSD peeps use.

jimdigriz:
Easiest solution is make Linux faster, publish a benchmark showing this and explaining it with `signalfd()`. Then detail "`kqueue()` was a pain" then a few minutes later someone will send you a PR resolving it!

Use the *BSD/Linux rivalry to your own benefit! :)

erts/emulator/sys/unix/sys.c
```diff
@@ -341,12 +338,25 @@ erl_sys_init(void)
...
     sigemptyset(&act.sa_mask);
-    act.sa_flags = 0;
+
+#if (defined(BEAMASM) && defined(NATIVE_ERLANG_STACK))
+    /* The JIT assumes that signals don't execute on the current stack (as our
+     * Erlang process stacks may be too small to execute a signal handler).
+     *
+     * Make sure the SA_ONSTACK flag is set when needed so that signals execute
+     * on their own signal-specific stack. */
+    extra_flags |= (func != SIG_DFL && func != SIG_IGN) ? SA_ONSTACK : 0;
```
mikpe, line 351: nit: `if (func != ...) extra_flags |= SA_ONSTACK;`

mikpe (approved):

PR #7176, opened 2023-04-26
Handle feature attribute in epp_dodger

### LINKED ISSUE #7175: `edoc:get_doc/1` crashes on a file using a `feature` attribute
**To Reproduce**
In OTP 26, even with the `maybe_expr` feature enabled:

```
$ asdf global erlang 26.0-rc3   
$ erl -enable-feature maybe_expr
Erlang/OTP 26 [RELEASE CANDIDATE 3] [erts-14.0] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit:ns]

1> erl_features:enabled().
[maybe_expr]
```

Given the file:

```
$ cat /tmp/test.erl
-module(test).
-feature(maybe_expr, enable)
```

The following call fails with an error:

```
2> edoc:get_doc("/tmp/test.erl").
/tmp/test.erl: at line 2: bad attribute
/tmp/test.erl: If the error is caused by too exotic macro
/tmp/test.erl: definitions or uses of macros, adding option
/tmp/test.erl: {preprocess, true} can help. See also edoc(3).
** exception exit: error
     in function  edoc:'-check_forms/3-fun-0-'/3 (edoc.erl, line 771)
     in call from lists:foreach_1/2 (lists.erl, line 1686)
     in call from edoc:read_source/2 (edoc.erl, line 640)
     in call from edoc_extract:source/3 (edoc_extract.erl, line 60)
```

**Expected behavior**
The call to `edoc:get_doc/1` succeeds.

**Affected versions**
* OTP 26.0-rc3

**Additional context**

The issue seems caused by the fact that `edoc` attempts at parsing the form via `erl_parse`, which fails:

```
3> redbug:start("erl_parse:parse_form->return", [{msgs, 99999}, {time, 10000}]).     
{48,1}
4> edoc:get_doc("/tmp/test.erl").                                               
[...]
% 15:45:55 <0.89.0>(dead)
% erl_parse:parse_form([{'-',2},
 {atom,2,feature},
 {'(',2},
 {atom,2,maybe_expr},
 {',',2},
 {atom,2,enable},
 {')',2},
 {dot,2}])
   
% 15:45:55 <0.89.0>(dead)
% erl_parse:parse_form/1 -> {error,{2,erl_parse,"bad attribute"}}
```

In turn, this is probably caused by the `epp_dodger` not handling the `feature` attribute [here](https://github.com/erlang/otp/blob/00c35bdf4514a23b69dbcb4aeb8e9233efad94dc/lib/syntax_tools/src/epp_dodger.erl#L498).

robertoaloi:
* Handle the `feature` attribute in `epp_dodger` (by killing the form)
* Add a testcase (failing before the change, passing after)

Testplan:

```
export ERL_TOP=`pwd`
./otp_build setup -a
make syntax_tools
make edoc_test ARGS="-suite edoc_SUITE -case module_with_feature"
```

Fixes #7175 .

### DIFF
--- a/lib/edoc/test/edoc_SUITE.erl
+++ b/lib/edoc/test/edoc_SUITE.erl
@@ -24,13 +24,14 @@
 
 %% Test cases
 -export([app/1,appup/1,build_std/1,build_map_module/1,otp_12008/1,
-         build_app/1, otp_14285/1, infer_module_app_test/1]).
+         build_app/1, otp_14285/1, infer_module_app_test/1,
+         module_with_feature/1]).
 
 suite() -> [{ct_hooks,[ts_install_cth]}].
 
 all() ->
     [app,appup,build_std,build_map_module,otp_12008, build_app, otp_14285,
-     infer_module_app_test].
+     infer_module_app_test, module_with_feature].
 
 groups() -> 
     [].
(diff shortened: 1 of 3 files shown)


michalmuskala:
I'm not sure this fix will be enough if a module actually uses `maybe` - could you try adding a test case like that?

robertoaloi:
> I'm not sure this fix will be enough if a module actually uses `maybe` - could you try adding a test case like that?

That seems to be indeed the case. It looks like the whole dodger doesn't support the `maybe` syntax yet.

bjorng:
Thanks for your pull request.

PR #7177, opened 2023-04-27
Update erl_nif.xml

ityonemo:
this function appears to have the "incorrect" const-correctness (it *should* be const * but the .h file *appears* to make it not const), but I could be misreading `erl_nif_api_funcs.h`.  This change to documentation reflects the reality of the code (unless the header file is to be changed to give it const correctness).


ityonemo:
withdrawn.  it appears my C parser is getting confused.

PR #7182, opened 2023-04-27
Remove incorrect compatibility section from `gb_sets` docs

Maria-12648430:
The paragraph is misleading, if not outright wrong.

The listed functions are fully compatible with those in the `ordsets` module, that much is true, at least AFAICS.

They are also present and provide the same _functionality_ as those in the `sets` module, but the results may be different since `sets` considers two elements as different iff they do not match (`=:=`), while `gb_sets` and `ordsets` consider two elements as different iff they do not compare equal (`==`).

Examples:
```erlang
1> gb_sets:to_list(gb_sets:from_list([1, 1.0])).
[1]
2> sets:to_list(sets:from_list([1, 1.0])).
[1.0, 1]
3> gb_sets:is_element(1.0, gb_sets:from_list([1])).
true
4> sets:is_element(1.0, sets:from_list([1])).
false
...
```

So, one can _not_ simply swap out the module name `sets` for `gb_sets`/`ordsets` or vice versa.

I see no big value in the paragraph itself, especially with version 2 `sets` becoming the preferred set representation. So instead of correcting the paragraph, making it clumsy and verbose, I think it is best to remove it altogether. Also, none of the other two set modules does have them.

### DIFF
--- a/lib/stdlib/doc/src/gb_sets.xml
+++ b/lib/stdlib/doc/src/gb_sets.xml
@@ -68,48 +68,10 @@
 
   <section>
     <title>Compatibility</title>
-    <p>The following functions in this module also exist and provides
-      the same functionality in the
-      <seeerl marker="sets"><c>sets(3)</c></seeerl> and
-      <seeerl marker="ordsets"><c>ordsets(3)</c></seeerl>
-      modules. That is, by only changing the module name for each call,
-      you can try out different set representations.</p>
-    <list type="bulleted">
-      <item><seemfa marker="#add_element/2"><c>add_element/2</c></seemfa>
-      </item>
-      <item><seemfa marker="#del_element/2"><c>del_element/2</c></seemfa>
-      </item>
-      <item><seemfa marker="#filter/2"><c>filter/2</c></seemfa>
-      </item>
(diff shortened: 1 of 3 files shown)


bjorng:
For many practical uses of sets, the `gb_sets` module is fully compatible with both `sets` and `ordsets`. For example, in the  compiler we often have sets of variable name. Those are guaranteed to never contain floating point values, so we can use any type of set depending on which is fastest for the types of operations we use for that particular set. The difference between `sets` and `gb_sets` only becomes an issue if you can have arbitrary Erlang terms as set members.

I think that this information can be useful, but the documentation for the `gb_sets` modules is perhaps not the best place. Putting it in `sets` could be more logical (because it is a more general name and probably where developers needing a set implementations would look first). Obviously, a note would have to be added about mixing integer and float members. `gb_sets` and `ordsets` could have a link to that part of the `sets` module documentation.

juhlig:
@bjorng moved and augmented the paragraph as you suggested. What do you think?

bjorng:
This looks good to me. I have changed the base to `master` because we will only release emergency patches for OTP 25.

Please squash the commits.

juhlig:
Squashed :+1:

bjorng:
Thanks for your pull request.

PR #7189, opened 2023-04-28
ct: various doc fixes

u3s:
- fix typo group/0 -> groups/1
- in groups example, wrap result term in a list
- add testcase tuple in type specs for all/0 and groups/0

### DIFF
--- a/lib/common_test/doc/src/ct_suite.xml
+++ b/lib/common_test/doc/src/ct_suite.xml
@@ -85,7 +85,9 @@
       <fsummary>Returns the list of all test case groups and test cases
 	in the module.</fsummary>
       <type>
-	<v><seetype marker="#ct_test_def">ct_test_def()</seetype> = TestCase | {group, GroupName} | {group, GroupName, Properties} | {group, GroupName, Properties, SubGroups}</v>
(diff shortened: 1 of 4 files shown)


lib/common_test/doc/src/ct_suite.xml
```diff
@@ -135,11 +138,12 @@
         <v><seetype marker="#ct_group_def">ct_group_def()</seetype> = {GroupName, Properties, GroupsAndTestCases}</v>
         <v>GroupName = <seetype marker="#ct_groupname">ct_groupname()</seetype></v>
         <v>Properties = [parallel | sequence | Shuffle | {RepeatType, N}]</v>
-        <v>GroupsAndTestCases = [Group | {group, GroupName} | TestCase]</v>
+        <v>GroupsAndTestCases = [Group | {group, GroupName} | TestCase | {testcase, TestCase, TestCaseRepeatType}]</v>
```
u3s, line 141: (Ingela) Verify if testcase tuple can also contain 2 elements only and if this should be specced here ...
see 2 element testcase tuple https://www.erlang.org/doc/man/ct.html#run_test-1

u3s: 2 element testcase tuple does not seem to work in test suite code. Therefore, I will not add it.

PR #7193, opened 2023-05-01
Amends enif_release_resource documentation.

ityonemo:
Adds extra clarification on ordering for resource release operations.  If you do it in the wrong order, there is a high likelihood of causing segfault.

### DIFF
--- a/erts/doc/src/erl_nif.xml
+++ b/erts/doc/src/erl_nif.xml
@@ -3040,6 +3040,10 @@ enif_map_iterator_destroy(env, &amp;iter);</code>
           References made by <seecref marker="#enif_make_resource">
           <c>enif_make_resource</c></seecref>
           can only be removed by the garbage collector.</p>
+        <p> You may call this function only after the resource object has been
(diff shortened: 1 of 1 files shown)


erts/doc/src/erl_nif.xml
```diff
@@ -3040,6 +3040,10 @@ enif_map_iterator_destroy(env, &amp;iter);</code>
           References made by <seecref marker="#enif_make_resource">
           <c>enif_make_resource</c></seecref>
           can only be removed by the garbage collector.</p>
+        <p> You may call this function only after the resource object has been
+          bound to a term using <seecref marker="#enif_make_resource><c>
```
savonarola, line 3044:
```suggestion
          bound to a term using <seecref marker="#enif_make_resource"><c>
```

jhogberg:
There is no requirement that every call to `enif_alloc_resource` must land in `enif_make_resource`. It's okay to "back out" of creating a resource by calling `enif_release_resource` without calling `enif_make_resource`, and this change would imply that it's illegal.

PR #7196, opened 2023-05-03
Use map generator in sets:filter/2 for version 2 sets

juhlig:
As briefly discussed in #7183, this PR simplifies the implementation of `sets:filter/2` for version 2 `sets` by using a map generator instead of an iterator. It should also be slightly more efficient.


bjorng:
Thanks for your pull request.

PR #7203, opened 2023-05-03
ssl: Add signature_algs/2

IngelaAndin:
To make it easier to configure signature algorithms with algorithms that are moved from the default add the API function signature_algs/2 that lists possible values.

Also adjust signature algorithms according to TLS-1.3 RFC. That is TLS-1.3 legacy algorithms are the TLS-1.2 algorithms to be supported in TLS-1.2-TLS-1.3 combination. sha224 is also considered legacy.

### DIFF
--- a/lib/ssl/doc/src/ssl.xml
+++ b/lib/ssl/doc/src/ssl.xml
@@ -477,12 +477,10 @@
 {sha384, ecdsa},
 {sha384, rsa},
 {sha256, ecdsa},
-{sha256, rsa},
-{sha224, ecdsa},
-{sha224, rsa}
+{sha256, rsa}
 ]</code>
 
-<p>Support for {md5, rsa} was removed from the the TLS-1.2 default in ssl-8.0 (OTP-22) and support for SHA1 {sha, _} was removed in ssl-11.0 (OTP-26) </p>
+<p>Support for {md5, rsa} was removed from the the TLS-1.2 default in ssl-8.0 (OTP-22) and support for SHA1 {sha, _}  and SHA224 {sha224, _} was removed in ssl-11.0 (OTP-26) </p>
 
  <p><c> rsa_pss_schemes =</c></p>
  <code>
@@ -521,10 +519,10 @@ rsa_pss_schemes()
  <p>EDDSA was made highest priority in ssl-10.8 (OTP-25) </p>
 
 <p>TLS-1.3 default is</p>
-<code>Default_TLS_13_Schemes ++ Legacy_TLS_13_Schemes </code>
+<code>Default_TLS_13_Schemes</code>
 
 <p>If both TLS-1.3 and TLS-1.2 are supported the default will be</p>
-<code>Default_TLS_13_Schemes ++ Default_TLS_12_Alg_Pairs </code>
+<code>Default_TLS_13_Schemes ++ TLS_13_Legacy_Schemes ++ Default_TLS_12_Alg_Pairs (not represented in TLS_13_Legacy_Schemes) </code>
 
 <p>so appropriate algorithms can be chosen for the negotiated
 version.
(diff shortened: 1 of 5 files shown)


lib/ssl/doc/src/ssl.xml
```diff
@@ -477,12 +477,10 @@
...
 {sha384, rsa},
 {sha256, ecdsa},
-{sha256, rsa},
-{sha224, ecdsa},
-{sha224, rsa}
+{sha256, rsa}
 ]</code>
 
-<p>Support for {md5, rsa} was removed from the the TLS-1.2 default in ssl-8.0 (OTP-22) and support for SHA1 {sha, _} was removed in ssl-11.0 (OTP-26) </p>
+<p>Support for {md5, rsa} was removed from the the TLS-1.2 default in ssl-8.0 (OTP-22) and support for SHA1 {sha, _}  and SHA224 {ssh224, _} was removed in ssl-11.0 (OTP-26) </p>
```
u3s, line 483: typo: ssh224 -> sha224

u3s (approved):

lib/ssl/doc/src/ssl.xml
IngelaAndin: Good catch fixed!

lib/ssl/src/ssl.erl
```diff
@@ -1105,6 +1105,42 @@ append_cipher_suites(Filters, Suites) ->
...
 
+%%--------------------------------------------------------------------
+-spec signature_algs(Description, Version) -> [signature_algs()] when
+      Description :: default | all | exclusive,
+      Version :: protocol_version().
+
+%% Description: Returns possible signature algorithms/schemes
+%% for TLS/DTLS version
+%%--------------------------------------------------------------------
+
```
dgud, line 1116: signature_algs(Desc, 'dtlsv1.2') ->
    signature_algs(Desc, 'tlsv1.2').

lib/ssl/doc/src/ssl.xml
```diff
@@ -2175,6 +2173,22 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
+        <c>Description</c> that are available. The
+        <c>exclusive</c> option will exclusively list
+        algorithms/schemes for that protocol version, whereas the
+        default and all options lists the combined list to support the
+        range of protocols from (D)TLS-1.2, the first version to support
+        configuration of the signature algorithms, to <c>Version</c>. </p>
+	<note><p>Some TLS-1-3 scheme names overlap with TLS-1.2 tuple-pair-names
+	and then TLS-1.3 names will be used.</p></note>
+      </desc>
+    </func>
```
dgud, line 2191: Should we have an example? 
I don't think I understand the difference between all and default in this description. 
The default and all options in the description should be in <c> </c> tags.

lib/ssl/src/ssl.erl
IngelaAndin: This is correct! DTLS-1.2 corresponds to TLS-1.2 and hence have the same signature algorithms.  For payload ciphers this is happens to be not true, but it is for signature algs.

dgud: You can simplify the code here with one match for dtls, so you don't have to have it in 3 places.

dgud (approved):

lib/ssl/doc/src/ssl.xml
IngelaAndin: So I tried to address this, pleas read again!

lib/ssl/src/ssl.erl
IngelaAndin: Ah, now I see what you mean!

IngelaAndin: fixed!

lib/ssl/doc/src/ssl.xml
```diff
@@ -2175,6 +2173,26 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
+        <c>Description</c> that are available. The
+        <c>exclusive</c> option will exclusively list
+        algorithms/schemes for that protocol version, whereas the
+        <c>default</c> and <c>all</c> options lists the combined list to support the
+        range of protocols from (D)TLS-1.2, the first version to support
+        configuration of the signature algorithms, to <c>Version</c>. </p>
+	<note><p>Some TLS-1-3 scheme names overlap with TLS-1.2
+	algorithm-tuple-pair-names and then TLS-1.3 names will be
+	used, for example <c>rsa_pkcs1_sha256</c> instead of
+	<c>{sh256, rsa}</c> these are legacy algorithms in TLS-1.3
```
dgud, line 2191: sha256

IngelaAndin: fixed

lib/ssl/src/ssl.erl
```diff
@@ -1105,6 +1105,38 @@ append_cipher_suites(Filters, Suites) ->
...
+        tls_v1:legacy_signature_algs_pre_13();
+signature_algs(all, 'tlsv1.2') ->
+    tls_v1:default_signature_algs([tls_record:protocol_version_name('tlsv1.2')]) ++ 
+        tls_v1:legacy_signature_algs_pre_13();
+signature_algs(exclusive, 'tlsv1.3') ->
+    tls_v1:default_signature_algs([tls_record:protocol_version_name('tlsv1.3')]);
+signature_algs(exclusive, 'tlsv1.2') ->
+    Algs = [Alg || Alg <- tls_v1:default_signature_algs([tls_record:protocol_version_name('tlsv1.2')]),
+                   not lists:member(Alg, tls_v1:legacy_signature_algs_pre_13())],
+    Algs ++ tls_v1:legacy_signature_algs_pre_13();
```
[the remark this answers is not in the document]
IngelaAndin: Order is important. And the thing is that  some algorithms has an atom name in TLS-1.3 context and another tuple representation in TLS-1.2 context.  We will use the TLS-1.3 names initially and if TLS-1.2 is negotiated these will be converted to that format.  On protocol level they will have the same representation.  I will make sure we add a comment. But I might do it after the release as time is running out!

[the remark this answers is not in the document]
IngelaAndin: Looking into this a last time before deciding to merge or not I realized I was a little out of context just looking at the diff, and misunderstood.  A comment will not be needed I think this was actually some residue form a refactor and you where right this was strange. 
```
  Algs = tls_v1:default_signature_algs([tls_record:protocol_version_name('tlsv1.2')]),
    Algs ++ tls_v1:legacy_signature_algs_pre_13();
```
will do fine.  I changed it and will merge tomorrow.

lib/ssl/src/tls_v1.erl
```diff
@@ -910,16 +910,14 @@ default_signature_algs(_) ->
 default_pre_1_3_signature_algs_only() ->
     Default = [%% SHA2
                {sha512, ecdsa},
-               {sha512, rsa},
```
ahovgaard, line 913: @IngelaAndin Sorry to revive this old PR. Is the intention that the `{sha256, rsa}` signature algorithm is still supposed to work with TLS 1.2 with the default list of signature algorithms (when both TLS 1.3 and 1.2 are supported), because the default list contains the `rsa_pkcs1_sha256` TLS 1.3 legacy scheme?

I am trying to understand how it's supposed to work, since I have observed issues with `SERVER ALERT: Fatal - Insufficient Security - :no_suitable_ciphers` after upgrading to OTP 26, since my `ssl` server no longer supports `{sha256, rsa}` by default. It isn't clear to me if this is intended to be changed.

IngelaAndin:
I believe the problem is fixed by #7783

PR #7205, opened 2023-05-03
stdlib: fix exported types in supervisor, gen_server, gen_event

kikofernandez:
Exports types mentioned in the documentation of supervisor, gen_server, and gen_event

Closes #6893

### DIFF
--- a/lib/stdlib/doc/src/gen_event.xml
+++ b/lib/stdlib/doc/src/gen_event.xml
@@ -142,8 +142,12 @@ gen_event:stop             ----->  Module:terminate/2
     </datatype>
 
     <datatype>
-      <name name="emgr_ref"/>
-    </datatype>    
+      <name name="sup_ref"/>
+    </datatype>
+
+    <datatype>
+      <name name="sup_name"/>
+    </datatype>
     
     <datatype>
       <name name="request_id"/>
--- a/lib/stdlib/doc/src/gen_event.xml
+++ b/lib/stdlib/doc/src/gen_event.xml
@@ -142,7 +142,7 @@ gen_event:stop             ----->  Module:terminate/2
     </datatype>
 
     <datatype>
-      <name name="sup_ref"/>
+      <name name="process_ref"/>
     </datatype>
 
     <datatype>
--- a/lib/stdlib/doc/src/gen_event.xml
+++ b/lib/stdlib/doc/src/gen_event.xml
@@ -146,7 +146,7 @@ gen_event:stop             ----->  Module:terminate/2
     </datatype>
 
     <datatype>
-      <name name="sup_name"/>
+      <name name="process_name"/>
     </datatype>
     
     <datatype>
--- a/lib/stdlib/doc/src/gen_server.xml
+++ b/lib/stdlib/doc/src/gen_server.xml
@@ -133,19 +133,19 @@ gen_server:abcast     -----> Module:handle_cast/2
           <seemfa marker="#enter_loop/3"><c>enter_loop/3,4,5</c></seemfa>,
           and the type
 	  <seetype marker="#server_ref"><c>server_ref()</c></seetype>
-	  below.
+	  is an alias to <seetype marker="stdlib:supervisor#sup_name"><c>supervisor:sup_name()</c></seetype>.
         </p>
         <taglist>
-          <tag><c>{local,<anno>LocalName</anno>}</c></tag>
+          <tag><c>{local,LocalName}</c></tag>
           <item>
             <p>
               Register the <c>gen_server</c> locally
-              as <c><anno>LocalName</anno></c> using
+              as <c>LocalName</c> using
               <seemfa marker="erts:erlang#register/2">
                 <c>register/2</c></seemfa>.
             </p>
           </item>
-          <tag><c>{global,<anno>GlobalName</anno>}</c></tag>
+          <tag><c>{global,GlobalName}</c></tag>
           <item>
             <p>
               Register the <c>gen_server</c> process id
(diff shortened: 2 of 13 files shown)


kikofernandez:
Closes #6893

IngelaAndin:
This feels kind of backwards! Should we not have a generic type that used by gen_statem, gen_server and supervisors?  Supervisors actually are implemented as  gen_servers.

KennethL (changes requested):
I think this needs to be taken up at an OTB since it has quite much impact on the public API.
If types are exported they should have well thought names that we are happy with and would like to keep. 
When exporting types they become part of the public API that we have to keep compatible, When the types where only mentioned in the doc but not exported they where only for manual use and we could change them whenever we wanted.

kikofernandez:
> This feels kind of backwards! Should we not have a generic type that used by gen_statem, gen_server and supervisors? Supervisors actually are implemented as gen_servers.

What do you mean with this? that the types from gen_servers are the ones that everyone references? or that the types from gen_server are the ones that everyone references under an alias (similar to what we have now but I chose supervisor to be the "global" type to be shared and every other module creates an alias towards it?

kikofernandez:
@IngelaAndin @KennethL I have made changes as suggested in the OTB thread, where proc_lib is the one that exports those types, which are consumed by `gen_XXX.erl` modules

IngelaAndin (changes requested):
I think proc_lib:sup_name should be proc_lib:process_name, or maybe shorter proc_name. Otherwise I like it now.

IngelaAndin (approved):

kikofernandez:
@IngelaAndin  for consistency, I think that if we have not ever exported `gen_xxx:sup_ref()` and others, then these should be renamed as well as `gen_xxx:process_ref()`?

IngelaAndin:
@kikofernandez I agree with you. I think it makes much more sense to use the names from the main building blocks to represent common types.

kikofernandez:
@IngelaAndin I have updated the `gen_xxx:server_ref()` and `gen_xxx:server_name()` to `gen_xxx:process_ref()` and `gen_xxx:process_name()`. Exactly the same for `gen.erl`. There is a file named `peer.erl` that already exports `server_ref()`, so I cannot touch that one. Let me know if this changes are ok with you :)

lib/stdlib/doc/src/gen_event.xml
```diff
@@ -142,8 +142,12 @@ gen_event:stop             ----->  Module:terminate/2
     </datatype>
 
     <datatype>
-      <name name="emgr_ref"/>
-    </datatype>    
+      <name name="process_ref"/>
```
IngelaAndin, line 145: emgr is not a clear name I think, I guess it stands for event manager. But in this particular place I am not sure if it is obvious if we should use the generalized type or if we should have a type for this abstraction?!  I think the same goes for gen_server and supervisor module. I will make a comment in those place too, so that we can have comments from the @erlang/team-ps

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -186,7 +186,7 @@ gen_server:abcast     -----> Module:handle_cast/2
 	  <seemfa marker="#wait_response/2"><c>wait_response/2</c></seemfa>,
 	  <seemfa marker="#stop/1"><c>stop/2,3</c></seemfa>
           and the type
-	  <seetype marker="#server_name"><c>server_name()</c></seetype>
+	  <seetype marker="#process_name"><c>process_name()</c></seetype>
```
IngelaAndin, line 189: Here is one place

lib/stdlib/doc/src/gen_server.xml
```diff
@@ -122,7 +122,7 @@ gen_server:abcast     -----> Module:handle_cast/2
 
   <datatypes>
     <datatype>
-      <name name="server_name"/>
+      <name name="process_name"/>
       <desc>
```
IngelaAndin, line 126: Here is one place

lib/stdlib/doc/src/supervisor.xml
```diff
@@ -606,7 +609,7 @@ child_spec() = #{id => child_id(),             % mandatory
       <fsummary>Create a supervisor process.</fsummary>
       <type name="startlink_ret"/>
       <type name="startlink_err"/>
-      <type name="sup_name"/>
+      <type name="process_name"/>
       <desc>
```
IngelaAndin, line 613: Here is one place

lib/stdlib/doc/src/gen_statem.xml
```diff
@@ -512,26 +512,26 @@ handle_event(_, _, State, Data) ->
 
   <datatypes>
     <datatype>
-      <name name="server_name"/>
+      <name name="process_name"/>
```
IngelaAndin, line 515: Here is one place

lib/stdlib/doc/src/gen_event.xml
IngelaAndin: Or maybe it is the case that we should not have new types with the same name but let the documentation refer to 
proc_lib:process_name and so on!

kikofernandez: The point on making aliases an exporting the aliases is to provide a type abstraction over it, so that theoretically, in a common type system, `gen_event:process_ref()` is not the same as other type that happens to match its internal representation. That said, the current type checker (Dialyzer) will expand the types to its constituents, so for Dialyzer `gen_event:process_ref()` and `proc_lib:process_ref()` are the same. 
(AFAIK, Dialyzer is a structural type system and will establish the equality of types as mentioned above)

kikofernandez: Regarding the generalisation of the type, I know think you mean that `emgr_ref()` is more concrete than `process_ref()`.
I would like to favour uniformity, so that we can create a "type" understanding without having to know that an `emgr_ref()` is equivalent to what is known in other modules as `process_ref()`

kikofernandez:
@rickard-green Could you take a look at this PR and let us know what you think of the types?

PR #7207, opened 2023-05-04
Fix bug in private-append transform

frej:


bjorng:

PR #7208, opened 2023-05-04
Fix `filelib:safe_relative_path/2`

### LINKED ISSUE #6460: the `filelib:safe_relative_path` may return wrong result in macos
**Describe the bug**
I use macOS Big Sur (Version 11.4) with OTP 24.1.5, and I found that `filelib:safe_relative_path` will return a wrong result if the `Cwd` is one of `/etc` `/tmp` `/var` `/home` 

e.g.
` filelib:safe_relative_path("foo/../foo", "/etc").`
will return
`foo/private/etc/foo`
but the correct result should be `foo`

` filelib:safe_relative_path("foo/../foo", "/home").`
will return
`unsafe`
but the correct result should be `foo`

**To Reproduce**
e.g.
` filelib:safe_relative_path("foo/../foo", "/etc").`

**Expected behavior**
return correct result

**Affected versions**
The version I used is 24.1.5,  but I think other versions also have this issue.

**Additional context**
Add any other context about the problem here. If you wish to attach Erlang code you can either write it directly in the post using [code tags](https://guides.github.com/features/mastering-markdown/), create a [gist](https://gist.github.com), or attach it as a zip file to this post.

juhlig:
Fixes #6460.

The function would return a wrong path if the second argument was a link and the path in the first argument would descend and then ascend again to the top level.

The function would also report a path as unsafe by falsely detecting a link loop if a link would target another link, which in turn would target a file of the same name like `bar` --> `foo/bar` --> (`foo/`)`foo/bar`.

As the link resolving and CWD-breakout and loop detection is pretty tricky, I'm not 100% sure it is correct in every aspect and corner case. Please review carefully.

### DIFF
--- a/lib/stdlib/src/filelib.erl
+++ b/lib/stdlib/src/filelib.erl
@@ -770,65 +770,65 @@ find_regular_file([File|Files]) ->
       Cwd :: filename_all(),
       SafeFilename :: filename_all().
 
+safe_relative_path(Path, "") ->
+    safe_relative_path(Path, ".");
 safe_relative_path(Path, Cwd) ->
-    case filename:pathtype(Path) of
-        relative -> safe_relative_path(filename:split(Path), Cwd, [], "");
-        _ -> unsafe
-    end.
+    srp_path(filename:split(Path),
+             Cwd,
+             sets:new([{version, 2}]),
+             []).
 
-safe_relative_path([], _Cwd, _PrevLinks, Acc) ->
-    Acc;
-
-safe_relative_path([Segment | Segments], Cwd, PrevLinks, Acc) ->
-    AccSegment = join(Acc, Segment),
-    case safe_relative_path(AccSegment) of
-        unsafe ->
-            unsafe;
-        SafeAccSegment ->
-            case file:read_link(join(Cwd, SafeAccSegment)) of
-                {ok, LinkPath} ->
-                    case lists:member(LinkPath, PrevLinks) of
-                        true ->
-                            unsafe;
-                        false ->
-                            case safe_relative_path(filename:split(LinkPath), Cwd, [LinkPath | PrevLinks], Acc) of
-                                unsafe -> unsafe;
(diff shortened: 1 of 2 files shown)


juhlig:
It might be worth noting that a path is always reported as `unsafe` if it contains a link with an _absolute_ target, even if that target is really located under the given `Cwd`. This behavior is the same as in the current implementation.

juhlig:
It may also be worth noting that the given `Cwd` is not checked in any way, ie assumed to be safe. That means that it may actually contain link loops, without `safe_relative_path` reporting the fact. _Only_ the given `Path` is checked.

juhlig:
Finally, it may be worth noting that if a call to `safe_relative_path` reports a given `Path` as safe (or `unsafe` for that matter), it is only so at the very moment when a path segment is checked. The actual directory/link structure may change immediately after or even _while_ the path is being checked.

Maria-12648430:
@rickard-green @bjorng @jhogberg what is the status of this PR, on your side? The issue behind this looks somewhat serious to me, which should be discussed and seen to :sweat_smile:

Maria-12648430:
That said, I would suggest as much as the complete removal of the function because, as @juhlig already said, the result is not unconditionally trustworthy on OSes that support symlinks: A path that is reported as safe (or `unsafe` for that matter) may not be safe (or `unsafe`) when it is used afterwards, as the filesystem may have changed. Even _while_ it is being checked, actually.

Nevertheless, it carries a strong connotation of "if you just use this function, nothing bad can happen". The alternative is to plaster the documentation with appropriate warnings.

Just my 2ct :wink:

bjorng:
@Maria-12648430 We intend to review this pull request but haven't gotten to it yet.

bjorng (commented):
I have taken a first look at the code. So far I haven't found any flaws. I will do some more thinking and probably discuss this PR with one or more colleagues.

juhlig:
@bjorng suggestions applied :+1:

bjorng:

bjorng:
Thanks for your pull request.

Issue #7043, opened 2023-03-22
Defining OLD based on LIBRESSL_VERSION_NUMBER may not be needed anymore


VlkrS:
Starting with [LibreSSL 3.5.0](https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.5.0-relnotes.txt) the reason for this doesn't seem to exist anymore:

https://github.com/erlang/otp/blob/5400ccf243a31d664153a4b9ceb9de3edfce1e0e/lib/crypto/c_src/otp_test_engine.c#LL40C37-L40C37

The [erlang port](https://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/lang/erlang/21/patches/Attic/patch-lib_crypto_c_src_otp_test_engine_c?rev=1.1&content-type=text/x-cvsweb-markup ) in OpenBSD at the time was patched to remove that line.

IngelaAndin:
I believe, this was fixed by @sverker and should be closed. @sverker please reopen if you disagree.

PR #7209, opened 2023-05-04
Libressl Adaptations

VlkrS:
* Define HAVE_CHACHA20_POLY1305 and HAVE_CHACHA20 for LibreSSL >= 3.7.0
* Remove a limitation that is not needed since LibreSSL >= 3.5.0 anymore (#7043)


sverker:
Added a fixup commit to actually test for LibreSSL < 3.5.0, if that is the version where the EVP_MD_meth_* function were introduced, which it seems to be.

lib/crypto/c_src/otp_test_engine.c
```diff
@@ -37,7 +37,7 @@
     PACKED_OPENSSL_VERSION(MAJ,MIN,FIX,('a'-1))
 
 #if OPENSSL_VERSION_NUMBER < PACKED_OPENSSL_VERSION_PLAIN(1,1,0) \
-    || LIBRESSL_VERSION_NUMBER < 0x3050000fL
```
sverker: That's actually not needed due to a quirk/feature in the preprocessor causing undefined macros to be replaced with 0 in #if-conditions.

I thought of maybe (ab)using that to define all-in-one macros like this

```
#ifdef LIBRESSL_VERSION_NUMBER
#  define HAS_LIBRESSL LIBRESSL_VERSION_NUMBER
#else  /* assume OpenSSL */
#  define HAS_OPENSSL OPENSSL_VERSION_NUMBER
#endif
```
in order to write simpler conditions like
```
#if HAS_OPENSSL > PACKED_VERSION_NUMBER(1,1,0) || HAS_LIBRESSL > 0x3050000f
#  define HAS_SOME_FEATURE
#endif
```

lib/crypto/c_src/otp_test_engine.c
```diff
@@ -37,7 +37,7 @@
     PACKED_OPENSSL_VERSION(MAJ,MIN,FIX,('a'-1))
 
 #if OPENSSL_VERSION_NUMBER < PACKED_OPENSSL_VERSION_PLAIN(1,1,0) \
-    || defined(LIBRESSL_VERSION_NUMBER)
+    || (defined(LIBRESSL_VERSION_NUMBER) && LIBRESSL_VERSION_NUMBER < 0x3050000fL)
```
sverker, line 40: That's actually not needed due to a quirk/feature in the preprocessor causing undefined macros to be replaced with 0 in #if-conditions.

I thought of maybe (ab)using that to define all-in-one macros like this

```
#ifdef LIBRESSL_VERSION_NUMBER
#  define HAS_LIBRESSL LIBRESSL_VERSION_NUMBER
#else  /* assume OpenSSL */
#  define HAS_OPENSSL OPENSSL_VERSION_NUMBER
#endif
```
in order to write simpler conditions like
```
#if HAS_OPENSSL > PACKED_VERSION_NUMBER(1,1,0) || HAS_LIBRESSL > 0x3050000f
#  define HAS_SOME_FEATURE
#endif
```

VlkrS: I did see a CI build failure due to LIBRESSL_VERSION_NUMBER being undefined, which prompted me to propose that fix

sverker: I take that back about abusing undefined macros as zero. We use compile option -Werror=undef which forbids such uglyness.

sverker:
I added som more fixes for LibreSSL and rebased on a fresh maint branch.

sverker:
Merged to maint for OTP 26.1.

PR #7211, opened 2023-05-05
kernel: support sending raw byte data over stdout

### LINKED ISSUE #7132: [OTP 26] Unexpected conversion when transferring data over stdout
Under OTP 26-RC3, an unexpected conversion (latin1 to unicode?) seems to be happening behind the scenes when transferring data via stdout. I put together a minimal reproducible example showing the issue:

https://github.com/robertoaloi/otp-26-stdout/

The repo is made of two tiny escripts: a client and a server. Upon request, the server sends EETF encoded data to the client via stdout. This works under OTP 25, but the data seems corrupted on OTP 26-RC3.

Please refer to the README for how to run the sample code, but essentially this is what happens:

OTP 25:

```
asdf global erlang 25.3
./client
<<131,104,4,100,0,2,105,115,100,0,4,116,104,105,115,100,0,1,97,100,0,3,98,117,103>>
{is,this,a,bug}%
```

OTP 26:

```
asdf global erlang 26.0-rc3
./client
<<194,131,104,4,119,2,105,115,119,4,116,104,105,115,119,1,97,119,3,98,117,103>>
escript: exception error: bad argument
  in function  binary_to_term/1
     called as binary_to_term(<<194,131,104,4,119,2,105,115,119,4,116,104,105,
                                115,119,1,97,119,3,98,117,103>>)
     *** argument 1: invalid external representation of a term
```

Things I tried (which didn't help):

* Various emulator flags, trying to enforce latin1
* Playing with various `io:setopts` configuration
* Specify `{minor_version, 1}` and `term_to_binary/2` (due to the UTF8 encoding of atoms by default in OTP 26)

What could explain the inconsistency?

frazze-jobb:
After rewriting the shell, group.erl converts everything to Unicode. However, this is problematic if you want to send raw byte data over stdout which may contain Erlang terms converted to binary that you want to convert back to Erlang term after data transfer. If io:setopts(_, {encoding, latin1}) is set, group.erl will just send the data directly to user_drv.erl which will send the data to prim_tty.erl without converting it to Unicode. Fixes #7132.

### DIFF
--- a/lib/kernel/src/prim_tty.erl
+++ b/lib/kernel/src/prim_tty.erl
@@ -164,6 +164,7 @@
                       sig => boolean()
                     }.
 -type request() ::
+        {putc_raw, binary()} |
         {putc, unicode:unicode_binary()} |
         {expand, unicode:unicode_binary()} |
         {insert, unicode:unicode_binary()} |
@@ -525,6 +526,8 @@ writer_loop(TTY, WriterRef) ->
 -spec handle_request(state(), request()) -> {erlang:iovec(), state()}.
 handle_request(State = #state{ options = #{ tty := false } }, Request) ->
     case Request of
+        {putc_raw, Binary} ->
+            {Binary, State};
         {putc, Binary} ->
             {encode(Binary, State#state.unicode), State};
         beep ->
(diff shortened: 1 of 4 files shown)


lib/kernel/src/prim_tty.erl
```diff
@@ -525,6 +525,8 @@ writer_loop(TTY, WriterRef) ->
 -spec handle_request(state(), request()) -> {erlang:iovec(), state()}.
 handle_request(State = #state{ options = #{ tty := false } }, Request) ->
     case Request of
+        {putc_raw, Binary} ->
```
jhogberg, line 528: Shouldn't this be added to the request type?

https://github.com/erlang/otp/blob/438a98a7e2dc8e39c0eeff91728d16b55581ea83/lib/kernel/src/prim_tty.erl#L166

rickard-green (approved):

lib/kernel/src/prim_tty.erl
rickard-green: @frazze-jobb I added this to the request() type and force pushed the branch so we can get this into the testing over the weekend.

michalmuskala:
Thank you for a quick fix!

PR #7215, opened 2023-05-08
Mnesia consistency

### LINKED ISSUE #7202: Eventual consistency in Mnesia
Mnesia's default partition tolerance model is to leave the programmer during a network partition (https://www.erlang.org/doc/apps/mnesia/mnesia_chap7#recovery-from-communication-failure). Moreover Mnesia only provides two access contexts, dirty and transactions, and lacks intermediate apis that provide consistency. For example, dirty operations might leave the replicas in different states depending on the order of the messages being delivered.

I am proposing a new API for Mnesia, which uses CRDTs for eventual consistency. This provides consistency for asynchronous operations (unlike transactions, which wait for all nodes to complete), but also gives better consistency guarantee than dirty operations. Moreover, this could also be used to add partition tolerance and auto conflict resolution after a network partition, although the exact semantics requires some careful tuning.

It is currently implemented in a [fork](https://github.com/Vincent-lau/otp/tree/converl-25) and I plan to submit PRs shortly. As this prototype is still at an early stage, I would appreciate feedbacks.

Alternative solutions might be using read-write quorums, or maybe use Paxos like consensus algorithms for stronger consistency guarantees. But Mnesia inherently works in a leaderless way, and I feel CRDTs are most suited to this style of communication.

**Additional context**

I also recorded a [demo](https://www.youtube.com/watch?v=qfVU4IQBrQI&t=3s) on the issues of dirty operations and the functionality of this new API.

Feel free to [reach me out](mailto:shuntian.liu@outlook.com) if you want to learn more about the technical details of this implementation.

incipit0:
Mnesia's default partition tolerance model is to leave it to the programmer during a network partition. Moreover Mnesia only provides two access contexts, dirty and transactions, and lacks intermediate apis that provide consistency. For example, dirty operations might leave the replicas in different states depending on the order of the messages being delivered.

This patch implements a new access context `(a)sync_ec` which provides eventual consistency when used in conjunction with a new table type `pawset`. For this patch, the `(a)async_ec` API can handle concurrent addition/deletion in the presence of network delays (while dirty operations might give different results on different replicas). Internally it implements a pure operation-based add-wins set. 

This is an attempt to fix #7202, although this is likely to require a lot more effort, and this patch is only the first step.


Vincent-lau:
Wondering if anyone can review this PR?

lin72h:
> Wondering if anyone can review this PR?

I think the OTP team are busy with releasing OTP 26, and your PR is a good addition to the Mnesia, but also a bit complex, for that reason the team trying to avoid to bring into the stdlib.

Vincent-lau:
> I think the OTP team are busy with releasing OTP 26, and your PR is a good addition to the Mnesia, but also a bit complex, for that reason the team trying to avoid to bring into the stdlib.

Okay, do you recommend keeping waiting until OTP26 has been released, or maybe do something else?

PR #7216, opened 2023-05-08
erlc: Fix compiler server args on windows

dgud:
String could be wrong length if arg was quoted and changed.


frazze-jobb (approved):
Looks good

PR #7220, opened 2023-05-11
kernel/file: Add read_file/2

### LINKED ISSUE #3624: ERL-315: Function file:read_file/2 is missing
Original reporter: `okeuday`
Affected version: `OTP-19.1`
Component: `kernel`
Migrated from: https://bugs.erlang.org/browse/ERL-315

---

```
While file:write_file/3 exists to allow usage of [raw] in the options, file:read_file/2 doesn't exist as a developer would expect.  To bypass this problem requires a developer to use prim_file:read_file/1 (undocumented) directly if they want raw behavior, making the situation more problematic than it needs to be.  Could we please add a file:read_file/2 function with the 2nd parameter as additional options to pass, to mirror the file:write_file/3 functionality?
```

jdamanalo:
Closes #3624. One particular use case is reading many small files.

_As an aside, there seems to be an unintended error in `file:delete(File, [])`. The file server does not handle delete calls with options. This is because there is no `prim_file:delete/2` anyway. Should I open a separate issue or fix it here?_

### DIFF
--- a/lib/kernel/doc/src/file.xml
+++ b/lib/kernel/doc/src/file.xml
@@ -1413,13 +1413,16 @@ f.txt:  {person, "kalle", 25}.
     </func>
     <func>
       <name name="read_file" arity="1" since=""/>
+      <name name="read_file" arity="2" since=""/>
       <fsummary>Read a file.</fsummary>
+      <type name="read_file_option"/>
       <desc>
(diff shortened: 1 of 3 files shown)


jhogberg:
Thanks for the PR!

> Should I open a separate issue or fix it here?

Feel free to fix it in a separate commit in this PR :)

jhogberg:

PR #7222, opened 2023-05-11
beam_call_types: Fix success check for 'bsl'

jhogberg:


lib/compiler/test/beam_type_SUITE.erl
```diff
@@ -1433,10 +1434,22 @@ gh_7197() ->
...
+               band 288230376151711743)) =:= _V0 ->
+    a;
+will_succeed_1(_, _) ->
+    b.
 
 %%%
 %%% Common utilities.
 %%%
-
+%%%
```
bjorng, line 1453: Seems to be an accidental change on this line.

```suggestion

```

bjorng (approved):

lib/compiler/test/beam_type_SUITE.erl
jhogberg: Thanks!

PR #7224, opened 2023-05-11
Fix scrollable cursors in odbc

rupurt:
The current implementation of scrollable cursors in the `odbc` driver does not work on `IBM Db2` & `Microsoft SQLServer`. Both databases require `SQL_ATTR_CURSOR_SENSITIVITY` to be set along with a `SQL_ATTR_CURSOR_TYPE` that is not `SQL_CURSOR_FORWARD_ONLY`.

Setting the default scrollable cursor type to `SQL_CURSOR_DYNAMIC` feels like a reasonable default but for full compatibility it would be great to pass this as an option so that we can use `SQL_CURSOR_STATIC`. This feature is not included in the patch.

- set `SQL_ATTR_CURSOR_SENSITIVITY` in db_select_count
- set `SQL_ATTR_CURSOR_TYPE` in db_select_count
- Db2 & SQLServer don't seem to support scrollable cursors with `SQL_CURSOR_FORWARD_ONLY`
- https://www.ibm.com/docs/en/db2-for-zos/11?topic=odbc-scrollable-cursor-characteristics-in-db2
- https://learn.microsoft.com/en-us/sql/odbc/reference/develop-app/cursor-characteristics-and-cursor-type?view=sql-server-ver16

### DIFF
--- a/lib/odbc/c_src/odbcserver.c
+++ b/lib/odbc/c_src/odbcserver.c
@@ -720,6 +720,20 @@ static db_result_msg db_select_count(byte *sql, db_state *state)
 	DO_EXIT(EXIT_ALLOC);
 
     if(use_srollable_cursors(state)) {
+        /* Db2 requires cursor sensitivity to be set to support scrollable
+         * cursors. By default it is = SQL_UNSPECIFIED  */
+	SQLSetStmtAttr(statement_handle(state),
+		       (SQLINTEGER)SQL_ATTR_CURSOR_SENSITIVITY,
+		       (SQLPOINTER)SQL_INSENSITIVE, (SQLINTEGER)1);
+        /* Db2 z/OS & SQLServer do not support scrollable cursors as
+         * SQL_CURSOR_FORWARD_ONLY. Ideally we would also support passing
+         * `SQL_ATTR_CURSOR_TYPE` as an argument so that we can also support
+         * `SQL_CURSOR_TYPE_STATIC`.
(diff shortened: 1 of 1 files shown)


IngelaAndin:
@rupurt  This patch is not causing any problems for our regression tests, although we only have a working environment for psql at the moment, alas our other environments have gone rotten :( and this is not a very high priority for Ericsson.  We are willing to accept this patch on you recommendation that it will work better for other backends.  When it comes to the suggest feature we would be willing to accept a PR, but it is nothing that we will work on.

rupurt:
Hi @IngelaAndin,

I feel like it's better to hold off. As I've got deeper I've found there are several current issues with the performance of batch workloads. I've documented the issues in further detail in my forked repo https://github.com/rupurt/otp/pull/1

IngelaAndin:
@rupurt So you mean you want to make a bigger PR fixing those thing too? I think the odbc code has not been actively developed in probably more than a decade so it does not surprise me there are thing that could be modernized or in other ways improved. You are welcome to make improvements. It would be nice if you could make one PR per logical change as the make the reviewing easier for us, especially since this is code that we do not work with on a daily bases.

rupurt:
Hi @IngelaAndin,

Yeah, the scope of it has grown a fair bit the deeper I got into the C code. I will make sure that any PR's I submit are broken up into smaller digestible pieces.

IngelaAndin:
I will close this due to inactivity.  The new PRs you where talking about are welcome when you find the time.

PR #7231, opened 2023-05-15
Fix Dialyzer failures in CI

TD5:
Dialyzer failed in CI during testing because we'd run the tests in a Docker image which uses the output of a build of OTP, but that build doesn't include the `scripts/` directory (which contains the script for running Dialyzer) from the main Git repository. This change mirrors that directory into the Docker image used for testing so that it can be run there.

Tested by running the GitHub Actions on my own fork.


bjorng:
Thanks for your pull request.

I changed the base to `maint` before merging.

PR #7243, opened 2023-05-16
Default -code_path_choice to strict

josevalim:
This option requires traversing the loadpaths
several times and doing multiple filesystem
lookups during boot.

This was previously scheduled for Erlang/OTP 26
but it was postponed as there was a need for a longer
deprecation cycle.

Revival of #6683. Follow up of #7117.

PS: Please remember to update the bootscripts on merge. :)


frazze-jobb:
This just merged to master, with updated preloaded modules

PR #7244, opened 2023-05-16
Remove archives fallback

josevalim:
Archives have a feature where both the archive directory and the application directory can co-exist and the contents of the archive will be read as a fallback. This pull request removes this feature as it is implicit behaviour and has a direct impact on each code path added to the code server with two additional file system lookups per code path.

Instead, the correct path must be given to -pa/-pz, and everything will continue to work as before. This pull request does not change anything in relation to escripts.

Revival of #6693. Follow up of #7117.


josevalim:
I will close this one, as it is outdated, but we are likely ready to remove archives (it was scheduled for OTP 28 anyway).

PR #7246, opened 2023-05-16
Kuba/ct/post end per testcase fix/otp 18579

u3s:


lib/common_test/src/test_server.erl
```diff
@@ -561,8 +561,14 @@ run_test_case_msgloop(#st{ref=Ref,pid=Pid,end_conf_pid=EndConfPid0}=St0) ->
...
-	{EndConfPid0,{call_end_conf,Data,_Result}} ->
-	    #st{mf={Mod,Func},config=CurrConf} = St0,
+	{EndConfPid0,{call_end_conf,Data,EndConf,_Result}} ->
+            #st{mf={Mod,Func},config=CurrConfFromState} = St0,
+            CurrConf = case EndConf of
+                           [] ->
+                               CurrConfFromState;
+                           _ ->
+                               EndConf
+                       end,
```
IngelaAndin, line 571: This feels a little strange? Should EndConf override CurrConfFromState only if it adds new config items?  I am  not verify familiar with this code base so only asking if maybe we should instead add EndConf ++ CurrentStateConf?  As I think it is a property list this will make the EndConf take precedence if it configures a value already present!

PR #7249, opened 2023-05-16
erts: Refactor internal term hashing

jhogberg:
Improve performance and hash quality by using MurmurHash3 instead of our old Jenkins96 derivative, and reduce the chances of needing collision nodes in maps by making the hash width follow the word size, allowing an extra 8 levels of collision before that happens on 64-bit platforms.


max-au:
Cool, now that MurmurHash3 is a part of ERTS, would that be possible to actually expose it to work over binaries? E.g. add that support to `crypto:hash(murmur3, Data)`. It would've saved me from having a NIF.

max-au:
Just need to ensure there is no confusion with `erlang:phash2` and `crypto:hash`...

erts/emulator/beam/erl_term_hashing.c
```diff
@@ -1977,38 +2186,50 @@ Uint32 erts_dbg_hashmap_collision_bonanza(Uint32 hash, Eterm key)
         };
         hash = hashvec[hash % (sizeof(hashvec) / sizeof(hashvec[0]))];
     }*/
-    const Uint32 bad_hash = (hash & 0x12482481) * 1442968193;
-    const Uint32 bad_bits = hash % 67;
-    if (bad_bits < 32) {
+    const erts_ihash_t bad_hash =
+        (hash & (erts_ihash_t)0x1248248112482481ull) *
+        (erts_ihash_t)0x5601F2815601F281ull;
```
sverker, line 2191: You have increased number of bad hashes to 65536.
I think 256 bad hashes was a good number. It will exercise collision nodes in a lot more test cases that do not create massive maps.

PR #7255, opened 2023-05-17
mnesia: Integrate implementation.txt to docs

jchristgit:
Mnesia had this file providing great insights into the way it works, but it was so far hard to discover. This commit includes it in the online HTML documentation under the "Internal Documentation" tab, and converts its existing ASCII figures into proper images.

Speaking of the figures, is there some guideline which software to use for making them? I chose xfig to make the first drawing and plantuml for the second, as I did not get good results with it on the first one.


dgud:
Closing this, I'm not sure we want to mix internal implementation docs with other.
And there are some documentation projects going on.

Issue #7165, opened 2023-04-25
gen_udp using {inet_backend, socket} causes crash


voluntas:
**Describe the bug**

I don't know how to reproduce the issue, but a server using gen_udp with the {inet_backend, socket} option enabled occasionally experiences crashes.

Looking at the crash log, it seems to be failing on the pattern match in this section:
https://github.com/erlang/otp/blob/OTP-26.0-rc3/lib/kernel/src/gen_udp_socket.erl#L646-L649

It seems to be expecting ok or {error, closed}, but {error, select_sent} is being received instead.

The conditions for this issue to occur are unknown, but the server experiencing this problem is a low-spec machine with 1 core and 1 GB of memory. It is possible that the low specs are causing the issue.

**To Reproduce**

Currently, I do not know how to reproduce the issue.

**Expected behavior**

I expect that the crash will not occur.

**Affected versions**

OTP 26.0-rc3

**Additional context**

Crash log: 
https://gist.github.com/voluntas/eb2b2f340fab26464be3a149c327d601

The following part of the crash log corresponds to ok = gen_udp:Close(socket):
{app_udp,maybe_socket_close,1, [{file,"/app/src/app_udp.erl"},{line,587}]},

Please let me know if there is any other information needed.

bmk:
Thanks for the report.
We have run into this ourselves.
Should be fixed when 26.0 is released (the fix is not yet merged into the master branch).

voluntas:
I'm sorry, I left out some important information. The environment is Ubuntu 22.04 x86_64.

bmk:
Fix is now merged (to master).

sile:
@bmk Hi,

I had a similiar crash even after updating to Erlang/OTP 26 (`Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:four:4] [ds:four:4:10] [async-threads:1] [jit:ns]`).

```
{{{{case_clause,{error,select_sent}},
      [{gen_udp_socket,cleanup_recv,4,
                       [{file,"gen_udp_socket.erl"},{line,646}]},
       {gen_udp_socket,handle_event,4,
                       [{file,"gen_udp_socket.erl"},{line,1356}]},
       {gen_statem,loop_state_callback,11,
                   [{file,"gen_statem.erl"},{line,1379}]},
       {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,241}]}]},
[...]

PR #7260, opened 2023-05-18
Fix a bug that prim_socket:cancel/3 could return `{error, select_sent|not_found}` while socket module expects `select_sent|not_found`

sile:
The return value of `prim_socket:cancel/3` changed in https://github.com/erlang/otp/commit/640c4a52c806d5a080f7e327d53cb8c6f022b4b0 from `select_sent|not_found` to `{error, select_sent|not_found}`. 
However, `socket` module keeps expecting that `prim_socket:cancel/3` returns `select_sent|not_found` (without `{error, _}`) as following code:

```erlang
%% From: https://github.com/erlang/otp/blob/OTP-26.0/lib/kernel/src/socket.erl#L4595-L4603
cancel(SockRef, Op, Handle) ->                                                                                                                                                                                
    case prim_socket:cancel(SockRef, Op, Handle) of                                                                                                                                                                             
        select_sent ->                                                                                                                                                                                                          
            _ = flush_select_msg(SockRef, Handle),                                                                                                                                                                              
            _ = flush_abort_msg(SockRef, Handle),                                                                                                                                                                               
            ok;                                                                                                                                                                                                                 
        not_found ->                                                                                                                                                                                                            
            _ = flush_completion_msg(SockRef, Handle),                                                                                                                                                                          
            _ = flush_abort_msg(SockRef, Handle),                                                                                                                                                                               
            invalid;   
```

This is the cause of the process crash reported in #7165.
This PR fixes this problem by reverting the changes in priim_socket_nif.c.

### DIFF
--- a/erts/emulator/nifs/common/prim_socket_nif.c
+++ b/erts/emulator/nifs/common/prim_socket_nif.c
@@ -10134,7 +10134,7 @@ ERL_NIF_TERM esock_cancel_mode_select(ErlNifEnv*       env,
             return esock_atom_ok;
         } else {
             /* Has already sent the message */
-            return esock_make_error(env, esock_atom_select_sent);
+            return esock_atom_select_sent;
         }
     } else {
         /* Stopped? */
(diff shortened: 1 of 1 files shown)


sile:
@bmk @IngelaAndin How is the status of the review?
It looks like a simple bug, so I would be happy if it is fixed by the next release (our company maintains a fork of erlang/otp to patch this change).

Issue #7266, opened 2023-05-19
EPP Dodger does not respect features


robertoaloi:
**Describe the bug**
This is a follow up from #7175, which was partially fixed by #7176.

**To Reproduce**
Given the following example module (extracted from the [OTP 26 highlights](https://www.erlang.org/blog/otp-26-highlights/#no-need-to-enable-feature-maybe-in-the-runtime-system) post):

```
-module(t).
-feature(maybe_expr, enable).

-export([listen_port/2]).
listen_port(Port, Options) ->
    maybe
        {ok, ListenSocket} ?= inet_tcp:listen(Port, Options),
        {ok, Address} ?= inet:sockname(ListenSocket),
        {ok, {ListenSocket, Address}}
    end.
```

The following crashes:
```
$ erl -enable-feature maybe_expr
> edoc:get_doc("t.erl").
t.erl: at line 7: syntax error before: '{'
t.erl: If the error is caused by too exotic macro
t.erl: definitions or uses of macros, adding option
t.erl: {preprocess, true} can help. See also edoc(3).
** exception exit: error
     in function  edoc:'-check_forms/3-fun-0-'/3 (edoc.erl, line 771)
     in call from lists:foreach_1/2 (lists.erl, line 1686)
     in call from edoc:read_source/2 (edoc.erl, line 640)
     in call from edoc_extract:source/3 (edoc_extract.erl, line 60)
```

**Expected behavior**
The function call succeeds.

**Affected versions**
* OTP 26.0

**Additional context**
The root cause is in the `epp_dodger` module, where the `io:scan_erl_form/4` function is invoked with an incorrect `reserved_word_fun`  which does not take into account the `feature` macros, which are currently discarded by the dodger.

A possible solution could be to introduce some state and update the features accordingly, similarly to how the `update_feature/4` function works in the `epp` module. This could be tricky to do in a backward-compatible way, since most of the `epp_dodger` parsing functions are exported.

Alternatives could be:

* Treat all potential keywords as keywords (would break the opposite case, where the atoms `maybe` and 
* Use the process dictionary

PR #7267, opened 2023-05-19
Always consider 'maybe' and 'else' as keywords in epp_dodger

robertoaloi:
This PR adds a testcase that highlights the issue in #7266 and provides a temporary fix. A proper solution would be to process the feature attributes in the dodger.


bjorng:
Thanks for your pull request.

PR #7271, opened 2023-05-22
compiler: Fix type inference bug for bit strings

frej:
The beam_ssa_type pass deduces the type of variables from the result of relational operators. In the true-branch following a `A '=:=' B`, the variables `A` and `B` are deduced to be of the type `beam_types:meet(type(A), type(B))` which is not necessarily true for bit strings. Bit strings are special in that nothing about their appendable-status can be deduced from a comparison. The only information gained is the size_unit. The appendable status is unchanged by the comparison.

This patch corrects a type inference bug for bit strings where a comparison would falsely set the appendable flag for non-appendable bit strings in the true-branch. The fix special-cases the type inferring following a `=:=` when one or both arguments are appendable by clearing the appendable status, calculating the resulting type and then restoring the appendable-status of the variables.
[edited to match updated commit message]

### DIFF
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -2512,6 +2512,24 @@ infer_types_br_1(V, Ts, Ds) ->
             {TrueTypes, FalseTypes}
     end.
 
+infer_relop('=:=', [LHS,RHS],
+            [#t_bitstring{appendable=LHSApp}=LType0,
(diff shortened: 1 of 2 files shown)


bjorng (approved):

bjorng:
We want to include this PR in an emergency patch release. Please rebase your branch like this and force-push:

```
git rebase --onto OTP-26.0 HEAD~1
```

frej:
> Please rebase your branch like this and force-push:

Done.

PR #7272, opened 2023-05-22
compiler: bif:binary_part/3 should alias its arguments

frej:
The `bif:binary_part/3` is a guard bif which creates a sub-binary (containing a reference to its argument) from its argument, thus aliasing both. Currently this does not produce any errors as the BEAM will only ever append to binaries, but if we ever attempt to do destructive updates of binaries this will give us grief, so we'd better correct it.

### DIFF
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -740,6 +740,11 @@ aa_bif(Dst, tl, Args, SS, AAS) ->
 %% TODO: Ignored for now, as we don't track what's inside maps.
 %% aa_bif(_Dst, map_get, _Args, SS, _AAS) ->
 %%     SS;
+aa_bif(Dst, binary_part, Args, SS, _AAS) ->
+    %% bif:binary_part/{2,3} is the only guard bif which could lead to
+    %% aliasing, it extracts a sub-binary with a reference to its
+    %% argument.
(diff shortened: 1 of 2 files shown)


lib/compiler/src/beam_ssa_alias.erl
```diff
@@ -740,6 +740,11 @@ aa_bif(Dst, tl, Args, SS, AAS) ->
 %% TODO: Ignored for now, as we don't track what's inside maps.
 %% aa_bif(_Dst, map_get, _Args, SS, _AAS) ->
 %%     SS;
+aa_bif(Dst, binary_part, Args, SS, _AAS) ->
+    %% bif:binary_part/3 is the only guard bif which could lead to
```
bjorng, line 744: Nitpick: The comment does not mention bif:binary_part/2 (which the code handles).

```suggestion
    %% bif:binary_part/2,3 is the only guard bif which could lead to
```

bjorng (approved):

lib/compiler/src/beam_ssa_alias.erl
frej: True, I checked the implementation and saw that both variants of `binary_part` boiled down to the same function, never fed that back to the comment. Fixed (and updated the commit message accordingly).

bjorng:

Issue #7029, opened 2023-03-20
Can't type cjk characters correctly in OTP 26 shell using Microsoft IME


shionryuu:
**Describe the bug**

Can't type CJK correctly using Microsoft IME

**To Reproduce**

- OS: Windows
- Terminal: cmd/powershell

```cmd
Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit:ns]

Eshell V14.0 (press Ctrl+G to abort, type help(). for help)
1> "お疲れ様です。". %% paste from clipboard
[12362,30130,12428,27096,12391,12377,12290]
2> "". %% "お疲れ様です。", type with Microsoft Japanese IME
[]
3> "1717". %% "17时17分", type with Microsoft Pinyin IME, is truncated to "1717"
"1717"
4> "". %% "·", paste from clipboard
[].
5> "·". %% "·", type with Microsoft Pinyin IME
"·"
6> ;2~ %% paste from clipboard, shift+insert
```

**Expected behavior**

Type CJK using Microsoft IME

**Affected versions**

[OTP 26.0-rc1](https://github.com/erlang/otp/releases/tag/OTP-26.0-rc1)

frazze-jobb:
Great, I will close this, I tried on Windows 10 and it seems to work, but I did not have a font in cmd.exe that could display the characters correctly.

PR #7275, opened 2023-05-22
fix ime not work in windows

shionryuu:
Fix https://github.com/erlang/otp/issues/7029

### DIFF
--- a/erts/emulator/nifs/common/prim_tty_nif.c
+++ b/erts/emulator/nifs/common/prim_tty_nif.c
@@ -361,17 +361,74 @@ static ERL_NIF_TERM tty_read_nif(ErlNifEnv* env, int argc, const ERL_NIF_TERM ar
                               &inputs_read)) {
             return make_errno_error(env, "ReadConsoleInput");
         }
+
+        /**
+         * Reading keyevents using ReadConsoleInput is a bit fragile as
+         * different consoles and different input modes cause events to
+         * be triggered in different ways. I've so far identified four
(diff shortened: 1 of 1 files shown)


garazdawi:
We have a few different scenarios to deal with that seem to make it impossible to figure out exactly what keys have been pressed (as far as I can tell).

1. Normal key presses
2. Using IME
3. Pasting text into the shell
4. Using ALT+ modifiers

To make things worse, all of the above except normal key presses work differently depending on the console you use.

### IME

In powershell+cmd IME seems to send only keydown events with the keys pressed, while in WindowsTerminal and alacritty it sends both keydown and keyup.

### Pasting text into shell

All consoles seems to treat pasted ascii text as keydown+keyup events, while unicode text is treated as a keydown with character set to 0 and then a keyup with the unicode value. The keydown and keyup seem to always be available to be read using ReadConsolInput at the same time.

### Using ALT+ modifiers

This does not work in WindowsTerminal, but in cmd.exe if you press ALT+1 (the 1 on the numpad) you get a keydown with character set to 0 and then in a later event you get a keyup with the correct unicode character.

I've pushed a commit to this PR that should make 1,2 and 3 work, but I can't see a way to make ALT+ work and not break IME....

If you could test my patch to make sure IME works as you would expect that would be great. I'll try to dig some more and see if I can figure out a way to make ALT+ to work aswell....

garazdawi:
Thanks for the PR!

PR #7278, opened 2023-05-23
ssl: Improve format_error/1

### LINKED ISSUE #7247: `ssl:format_error/1` crashes when given `{options,incompatible,...}`
**Describe the bug**

My understanding is `format_error(Reason)` should work with any `{error, Reason}` returned from ssl functions. This is not the case for `{options, incompatible, ...}` reason.

**To Reproduce**

```
Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

1> {error, Reason} = ssl:connect("www.erlang.org", 443, []).
{error,{options,incompatible,
                [{verify,verify_peer},{cacerts,undefined}]}}
2> ssl:format_error(Reason).
     in function  atom_to_binary/1
        called as atom_to_binary({options,incompatible,
                                          [{verify,verify_peer},{cacerts,undefined}]})
        *** argument 1: not an atom
     in call from erl_posix_msg:message_1/1 (erl_posix_msg.erl, line 175)
     in call from erl_posix_msg:message/1 (erl_posix_msg.erl, line 29)
     in call from ssl:do_format_error/1 (ssl.erl, line 2769)
```

**Expected behavior**
A formatted error message is returned.

**Affected versions**
26.0

dgud:
Convert more (and new) option tuples to error strings.

Fixes #7247

### DIFF
--- a/lib/ssl/test/ssl_api_SUITE.erl
+++ b/lib/ssl/test/ssl_api_SUITE.erl
@@ -187,7 +187,9 @@
          check_random_nonce/0,
          check_random_nonce/1,
          cipher_listing/0,
-         cipher_listing/1
+         cipher_listing/1,
+         format_error/0,
+         format_error/1
         ]).
 
 %% Apply export
@@ -275,7 +277,8 @@ simple_api_tests() ->
      invalid_cacertfile,
      invalid_options,
      options_not_proplist,
-     options_whitebox
+     options_whitebox,
+     format_error
     ].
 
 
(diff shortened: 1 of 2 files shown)


lib/ssl/test/ssl_api_SUITE.erl
```diff
@@ -3545,6 +3548,33 @@ cipher_listing(Config) when is_list(Config) ->
...
+              {option, server_only, alpn_preferred_protocols},
+              {options, {alpn_advertised_protocols, undefined}},
+              {options, missing_version, {'tlsv1.2',{versions, ['tlsv1.1','tlsv1.3']}}},
+              {options, {no_supported_algorithms, {signature_algs,[]}}},
+              {options, {no_supported_signature_schemes, {signature_algs_cert,[]}}}
+             ],
+    Check = fun(Err) ->
+                    Str = ssl:format_error(Err),
+                    io:format("~p => ~s ~n", [Err, Str]),
+                    %% Verify flat string (why?)
```
u3s, line 3571: if comment is still valid, elaborate more please.
if not needed, remove.

lib/ssl/test/ssl_api_SUITE.erl
```diff
@@ -3545,6 +3548,33 @@ cipher_listing(Config) when is_list(Config) ->
...
+                    Str = [C || C <- Str, is_integer(C)]
```
u3s, line 3572: should it be checked that C is within range of char() (https://www.erlang.org/doc/man/erlang.html#type-char)?

lib/ssl/test/ssl_api_SUITE.erl
```diff
@@ -3545,6 +3548,33 @@ cipher_listing(Config) when is_list(Config) ->
...
+              {options, {certfile, "TestFileName", {error,enoent}}},
+              {options, {cacertfile, "TestFileName",{error,enoent}}},
+              {options, {option_not_a_key_value_tuple, [foo]}},
+              {options, {insufficient_crypto_support,{'tlsv1.3','tlsv1.1'}}},
+              {options, {protocol, foo}},
+              {option, server_only, alpn_preferred_protocols},
+              {options, {alpn_advertised_protocols, undefined}},
+              {options, missing_version, {'tlsv1.2',{versions, ['tlsv1.1','tlsv1.3']}}},
+              {options, {no_supported_algorithms, {signature_algs,[]}}},
+              {options, {no_supported_signature_schemes, {signature_algs_cert,[]}}}
```
u3s, line 3566: include `{options,incompatible...}` from the issue in the test case?

u3s (approved):

lib/ssl/test/ssl_api_SUITE.erl
IngelaAndin: Sound a little overkill to me!

IngelaAndin (approved):

lib/ssl/test/ssl_api_SUITE.erl
dgud: I don't know why we flatten the string in ssl.erl, it just costs performance.

lib/ssl/test/ssl_api_SUITE.erl
dgud: In the unicode-range why should we?

lib/ssl/test/ssl_api_SUITE.erl
dgud: Sigh, how could I miss that

lib/ssl/test/ssl_api_SUITE.erl
u3s: I just checked what string(), char() types are ... 
ssl:format_error/1 is supposed to return string according to docs. Are unicode characters included in string()?

PR #7279, opened 2023-05-23
ssl:negotiated_protocol can return {error,closed}

essen:
The function ssl_gen_statem:call/2 will return `{error,closed}` when the process is not alive.

https://github.com/erlang/otp/blob/master/lib/ssl/src/ssl_gen_statem.erl#L1318-L1329

Related: https://github.com/ninenines/gun/issues/313


essen:
Done.

PR #7280, opened 2023-05-24
Fix shell:start_interactive/1 spec

michallepicki:
See also https://github.com/erlang/otp/pull/7063 and https://github.com/erlang/otp/blob/acedc250f06f02d534ba21b8ab1d84b3b830c50b/lib/kernel/src/user_drv.erl#L110-L113 (there `unicode:charlist()` is used instead of `string()` but I don't know what's more correct)

### DIFF
--- a/lib/stdlib/src/shell.erl
+++ b/lib/stdlib/src/shell.erl
@@ -58,11 +58,11 @@ non_local_allowed(_,_,State) ->
 -spec start_interactive() -> ok | {error, already_started}.
 start_interactive() ->
     user_drv:start_shell().
--spec start_interactive(noshell | mfa()) ->
+-spec start_interactive(noshell | {module(), atom(), [term()]}) ->
           ok | {error, already_started};
                        ({remote, string()}) ->
           ok | {error, already_started | noconnection};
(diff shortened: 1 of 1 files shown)


michallepicki:
As a sidenote, not sure why Dialyzer didn't catch that (and the wrong spec fixed by my previous PR) when analyzing the Erlang/OTP codebase. I found these mistakes when running Dialyzer on the Elixir codebase.

garazdawi:
Thanks for the PR! It is not caught by our dialyzer because none of the Erlang/OTP code uses that function. We would have had to run dialyzer against our testcases to catch it, which we would like but triggers too many false positives right now to be useful...

garazdawi:
Would you mind rebasing the commit to maint so that we can fix this in a patch rather than having to wait a year?

michallepicki:
> Would you mind rebasing the commit to maint so that we can fix this in a patch rather than having to wait a year?

Sorry about that. Cherry-picked onto fresh `maint` and force-pushed.

garazdawi:
Merged, thanks for the PR!

PR #7281, opened 2023-05-24
Fix register allocation in the presence of singled-valued phi nodes

### LINKED ISSUE #7248: [erlc] miscompilation with optimizations turned off
**Describe the bug**
On the following code:
```erlang
f() ->
    try _V6 = (_V5 = bit_size(erlang:iolist_to_binary("a"))) rem 1 of
        _ ->
            _V5;
        _ ->
            _V6
    catch
        _ ->
            ok
    end.

wrapper0() ->
    io:write(f()).
```
Running it with the following commands:
```
erlc -W0 ~/minimized/test62768.erl
erl -pa . -noshell -s test62768 wrapper0 -s init stop
```
results in `8` being printed to the shell. But adding `+no_bool_opt +no_ssa_opt` to erlc makes it output `0` instead.

**Expected behavior**
I believe that 8 is the valid result:
- first _V5 is set to 8
- then _V6 is set to 8 rem 1 which is 0
- then _V6 is matched against the first pattern, which is `_` so it is a match
- and so _V5 is returned.

**Affected versions**
- master
- master + #7240

**Additional context**
The following were tried and make the bug disappear:
- replacing the try/of/catch/end by case/of/end
- replacing the whole computation of _V5 by _V5 = 8
- only passing one of the two flags to erlc

bjorng:
With optimizations disabled, the register allocator in `beam_ssa_pre_codegen` could allocate different variables to the same register if single-valued phi nodes in a try/catch construct were present.

Closes #7248

### DIFF
--- a/lib/compiler/src/beam_ssa_pre_codegen.erl
+++ b/lib/compiler/src/beam_ssa_pre_codegen.erl
@@ -815,9 +815,8 @@ do_sanitize_is(#b_set{op=Op,dst=Dst,args=Args0}=I0,
                Is, Last, InBlocks, Blocks, Count, Values, Changed0, Acc) ->
     Args = sanitize_args(Args0, Values),
     case sanitize_instr(Op, Args, I0, Blocks) of
-        {value,Value0} ->
-            Value = #b_literal{val=Value0},
-            sanitize_is(Is, Last, InBlocks, Blocks, Count, Values#{Dst=>Value},
+        {subst,Subst} ->
+            sanitize_is(Is, Last, InBlocks, Blocks, Count, Values#{Dst => Subst},
                         true, Acc);
         {ok,I} ->
             sanitize_is(Is, Last, InBlocks, Blocks, Count, Values, true, [I|Acc]);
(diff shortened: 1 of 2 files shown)


lib/compiler/src/beam_ssa_pre_codegen.erl
```diff
@@ -862,6 +862,13 @@ sanitize_arg(#b_var{}=Var, Values) ->
 sanitize_arg(Arg, _Values) ->
     Arg.
 
+sanitize_instr(phi, [{#b_var{}=V,_}], I, _Blocks) ->
+    %% GH-7248: With optimizations disabled, the register allocator
+    %% can accidentaly allocate different variables to the same
```
RobinMorisset, line 867: typo: accidentaly -> accidentally

RobinMorisset (approved):
Thanks!

lib/compiler/src/beam_ssa_pre_codegen.erl
bjorng: Thanks! Fixed.

RobinMorisset:
This PR makes the following erlang test cases:
```erlang
f(_V0) ->
    {_V0 and _V0, [ok || <<_:(_V0)/integer-unsigned>> <= port_to_list(_V0 orelse _V0)]}.
```
and
```erlang
f(_V1) when (_V1 xor _V1); {_V1} ->
    erlang:insert_element(_V1 orelse _V1, ok, ok),
    _V1.
```
and
```erlang
f(_V1) ->
    case ok of
        _ ->
            not _V1,
            [ok || _ := <<_:(_V1)/bytes>> <- (_V1 andalso _V1)]
    end.
```
fail to compile with variations of the following error message:
```
  Instruction: {swap,{y,1},{x,0}}
  Error:       {uninitialized_reg,{y,1}}:
```

I can provide more test cases if it would be useful for debugging.

bjorng:
I have now pushed a different solution that I hope has no undesirable side effects.

PR #7287, opened 2023-05-24
Use generator for existing atoms in `lists` property tests

Maria-12648430:
The `atom()` generator of PropEr creates atoms from random strings, and is thereby prone to exhaust the atom limit. By extension, so are the generators containing the `atom()` generator, such as `any()` or `list()`.

The `lists` property test suite exhausted the atom limit at around 80 (of 107) properties, and we had to find a workaround by pre-generating a limited set of random atoms, store them in `persistent_term`, and a generator that subsequently took atoms only from this set instead of generating new ones.

This PR changes the custom atom generator again to not pre-generate anything but pick from the set of already existing atoms. This has the following benefits over the current version:
* no need to generate and store (and remember to delete) anything, as everything is already there
* by using existing atoms, there is a greater likelihood of hitting anything than by simply random ones, as they at least have been written down at least _somewhere_

Unfortunately, there is no official way of getting at the existing atoms, so this has to rely on an obscure "trick" that somebody figured out at some point. For OTP-internal tests like this one, I think it is justifiable, though.

### DIFF
--- a/lib/stdlib/test/lists_property_test_SUITE.erl
+++ b/lib/stdlib/test/lists_property_test_SUITE.erl
@@ -112,7 +112,6 @@ init_per_suite(Config) ->
     ct_property_test:init_per_suite(Config).
 
 end_per_suite(Config) ->
-    persistent_term:erase({lists_prop, random_atoms}),
     Config.
 
 do_proptest(Prop, Config) ->
--- a/lib/stdlib/test/property_test/lists_prop.erl
+++ b/lib/stdlib/test/property_test/lists_prop.erl
@@ -50,8 +50,6 @@
 -endif.
 -endif.
 
--define(RANDOM_ATOMS, 1000).
-
 %%%%%%%%%%%%%%%%%%
 %%% Properties %%%
 %%%%%%%%%%%%%%%%%%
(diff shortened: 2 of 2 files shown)


Maria-12648430:
@RobinMorisset

> ```erlang
> f(X = '原子') ->
>    <<X/utf8>>.
> ```
`'原子'`, nice pun :rofl:

Anyway, this is local to tests for the `lists` module, where the actual "appearance" of the atoms is of no importance whatsoever, as long as matching/comparison of atoms itself works. And if it doesn't, all hell would break loose all over the place, anyway :fire:

That said, one day I will try to do an extension to the property-based testing facility of `common_test` to enable it to work in an atom-limit-safe manner. Or so I keep telling myself each time I do some property testing suite for OTP and get reminded of it :sweat_smile:  (A PR I made in PropEr itself for that issue got somewhat stuck/stalled --> not coming anytime soon it seems :woman_shrugging:). When I get round to it, I will do as you suggest :slightly_smiling_face:

RobinMorisset:
Oh, I had not realized that this is only for the list module.
Thanks for the explanation, your solution seems perfect then.

bjorng:
I don't mind the trick for accessing existing atoms, but I only want to see it in one place in the test suite for STDLIB. That is, if you plan to use the same generator for other tests, please place in its own module that can be shared between property tests for STDLIB.

Maria-12648430:
If #7364 is accepted, this PR is obsolete. In that case, a new PR should be created to make use of the new generators.

Maria-12648430:
Obsoleted by #7364, will make another PR that uses `ct_proper_ext`.

Maria-12648430:
Superseded by #7454.

PR #7291, opened 2023-05-24
Document that NaN, +Inf, -Inf are not valid Erlang terms.

RobinMorisset:
The only documentation I managed to find about those values in Erlang is an 11 years-old mailing list message by Sverker:
http://erlang.org/pipermail/erlang-questions/2012-February/064728.html.

This commit just copies that information into the official documentation.


bjorng:
Please rebase the branch on `maint`.

bjorng:
Thanks for your pull request.

PR #7293, opened 2023-05-24
Document performance reduction with QLC string handles

jchristgit:
`qlc:string_to_handle` seems to have worse performance than `qlc:q` with the same input query, presumably because the parse transform is not optimizing `qlc:string_to_handle`.

Benchmarks that showcase this behaviour can be found here: https://elixirforum.com/t/performance-discrepancies-with-using-qlc-queries-written-in-erlang-and-elixir/

### DIFF
--- a/lib/stdlib/doc/src/qlc.xml
+++ b/lib/stdlib/doc/src/qlc.xml
@@ -1247,6 +1247,11 @@ ets:match_spec_run(
           <anno>Options</anno>, erl_eval:new_bindings())</c>.</p>
         <p>This function is probably mainly useful when called from
           outside of Erlang, for example from a driver written in C.</p>
+        <note>
+          <p>Query handles created this way may have worse
+            performance than when created directly via
+            <seemfa marker="#q/1"><c>q/1,2</c></seemfa>.</p>
+        </note>
       </desc>
     </func>
 


lib/stdlib/doc/src/qlc.xml
```diff
@@ -1247,6 +1247,11 @@ ets:match_spec_run(
           <anno>Options</anno>, erl_eval:new_bindings())</c>.</p>
         <p>This function is probably mainly useful when called from
           outside of Erlang, for example from a driver written in C.</p>
+        <note>
+          <p>Query handles created this way may have worse
+            performance on evaluation than when created directly via
```
bjorng, line 1252: This phrasing seems a little bit strange to me. What about:

```suggestion
            performance than when created directly via
```

bjorng (commented):
Thanks for your pull request.

The documentation does hint that the performance may be worse because the fun is evaluated using `erl_eval`. However, that requires knowledge of `erl_eval` or reading up on it while trying to understand `qlc`, so I approve of this explicit note about performance.

lib/stdlib/doc/src/qlc.xml
jchristgit: Sure, sounds good. I committed your suggestion.

bjorng:
Please squash the commits.

jchristgit:
Squashed.

In general should I squash any changes from review comments, or just for smaller changes like here?

bjorng:
Yes, in general you should squash changes made from review comments. We want a git history that is useful to look at ten years from now. Keeping a history of the development of a pull request is seldom useful.

Here is our guide for [writing good commit messages](https://github.com/erlang/otp/wiki/Writing-good-commit-messages).

bjorng:
Thanks for your pull request.

PR #7294, opened 2023-05-24
Fix docs about apps not supporting upgrade

jchristgit:
HiPE has been removed, erl_interface is a good other example.


bjorng:
Thanks!

PR #7299, opened 2023-05-25
inets: Implement simple command-line serving via httpd

jchristgit:
Often times people want to quickly have a directory index of some folder
to either view directly or share on the local network. For instance, in
Python, `python3 -m http.server` implements a simple server
accomplishing exactly this.

`httpd` has the needed functionality to do this, but did not define a
command-line entrypoint. Users who wanted to quickly serve files with
Erlang needed hacks like shown here:
https://gist.github.com/willurd/5720255#erlang.

This commit adds support for the following forms of hosting a simple
directory listing with support for `index.html` and basic mimetypes,
with the following commands:

    erl -S httpd
    erl -S httpd serve
    erl -S httpd serve path/to/dir

The original suggestion cropped up here:
https://erlangforums.com/t/httpc-httpd-improvements/2622/8

### DIFF
--- a/lib/inets/src/http_server/httpd.erl
+++ b/lib/inets/src/http_server/httpd.erl
@@ -43,6 +43,8 @@
          info/3,
          info/4
         ]).
+%% Command line interface
+-export([start/1, serve/1]).
 
 -deprecated({parse_query, 1,
             "use uri_string:dissect_query/1 instead"}).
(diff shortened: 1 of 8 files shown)


jchristgit:
Thanks for the feedback.

I've pushed a more complete implementation together with docs, tests, and the two command-line options `--bind` and `--port`, parsed via `argparse`.

I also have a rough draft of CGI support stashed away here, but I think that's better for a separate PR.

Argument passing is a bit strange because it requires to parse `-extra` to properly register the additional arguments, perhaps here init can be improved to have a flag like `-S` which internally would work like a combination of `-s` and `-extra` to allow forms such as `erl -S httpd serve --port 4000` directly? I might open a separate issue for this.

As for the test failure, I've noticed that `mod_alias` uses the configured `server_name` when performing redirects instead of the URL of the client, which breaks when the server name is not a hostname under which the server is reachable. Is there a reason this doesn't use the absolute request URL on the rewrite, maybe security? I believe it would also be possible to just redirect to a path and have the browser figure it out. But perhaps we should skip / xfail the test here for now?

jchristgit:
I pushed some small changes in my latest update to the commit:

- instead of serving in wherever the test server runs, serve in the data directory of the suite (added as well)
- when the server instructs us to redirect, use the connect address we figured out on our own instead of the server-provided one: the server-provided one depends on `server_name` which may not be the reachable address
- run the default port tests with `sequence` such that if one fails the others won't run and clutter logs
- actually retrieve random file to test in the `random_file` assertion instead of picking the first one we find
- regard `eaddrnotavail` as a skip (e.g. IPv6 not available), fail on other server startup errors

Hope this fixes CI - if not I'll work at it 🙂

jchristgit:
@u3s the tests are more tolerant of unexpected network configurations now, and pass in CI. Let me know if you have something else I should update :slightly_smiling_face:

jchristgit:
I made a draft PR at https://github.com/erlang/otp/pull/7470 that would ease starting this up, e.g. `erl -S httpd serve --flags` instead of `erl -s httpd serve -extra --flags`. Would be happy for your feedback there.

u3s:
I'm sorry for delay. We will get back to this one when possible.

lib/inets/test/httpd_serve_SUITE.erl
```diff
@@ -0,0 +1,241 @@
...
+
+wait_for_startup_line(["\nStarted HTTP" ++ _Rest = Line | _Lines], _Unexpected, _Tries) ->
+    {ok, Line};
+
+wait_for_startup_line([Line | Lines], Unexpected, Tries) ->
+    wait_for_startup_line(Lines, [Line | Unexpected], Tries).
+
+%% Parse the given line into a tuple.
+%% Example line:
+%%   Started HTTP server on 127.0.0.1:8000 at /path/to/lib/inets/make_test_dir/ct_logs/ct_run.test_server@zulu.2023-06-06_12.07.27\n"
```
u3s, line 234: Output in my env looks like `127.0.0.1 port 8000`. is the comment actual?

lib/inets/test/httpd_serve_SUITE.erl
```diff
@@ -0,0 +1,241 @@
...
+verify_server(FunOrArgs) ->
+    TestFun = fun run_server_assertions/1,
+    with_server(FunOrArgs, TestFun).
+
+with_server(Args, TestFun) when is_list(Args) ->
+    ServerFun = fun () -> httpd:serve(Args) end,
+    run_with_server(TestFun, ServerFun);
+
+with_server(ServerFun, TestFun) when is_function(ServerFun) ->
+    run_with_server(TestFun, ServerFun).
```
u3s, line 183: maybe I'm missing some convention, but I don't see point in change arguments order between `with_server` and `run_with_server`. wouldn't it be simpler is both functions take ServerFun as 1st and TestFun as 2nd argument?

u3s:
for reference, this PR potentially depends on #7470 (which I guess should be merged first)

u3s:
- are some adjustments needed to make `erl -S httpd serve` work properly?
- maybe also update docs to cover `-S` (do you think it would be enough if documentation only covers `-S` ?)
- please squash commits

u3s:
>> - maybe also update docs to cover `-S` (do you think it would be enough if documentation only covers `-S` ?)

> Actually I think it may be cleaner to both only document & only support `-S` because it will prevent any surprises with option parsing. What do you think?

I agree, I think this is inline with `simple` from the PR title and will reduce confusion.

lib/inets/src/http_server/httpd.erl
```diff
@@ -392,6 +394,122 @@ service_info(Pid) ->
...
+          type => {custom, fun parse_ip_address/1},
+          short => $b,
+          long => "-bind",
+          default => {127, 0, 0, 1},
+          help => "IP address to listen on. Use 0.0.0.0 or :: for all interfaces."
+        }
+      ],
+      help => (
+        "Start a HTTP server serving files from DIRECTORY. Any options "
+        "must be passed after the `-extra` flag of the emulator."
```
u3s, line 469: `-extra` is not relevant any longer?

u3s (changes requested):
- please fix error message (--exrta is no longer worth mentioning)
- I might have found some issue with `-S` which affects this PR (see comment in #7470 )

u3s (approved):

PR #7300, opened 2023-05-25
Fix incorrect DTLS retransmission timeout

Shaienn:
Hello. 

In case of UDP some packets may be lost. According to RFC (https://datatracker.ietf.org/doc/html/rfc9147#section-5.8.1-7.1) the DTLS handshake participant should re-send content of last flight after timeout but should retain current timer value if no retransmissions happened (https://datatracker.ietf.org/doc/html/rfc9147#section-5.8.2-2, https://datatracker.ietf.org/doc/html/rfc9147#section-5.8.2-5). 

I have observed that retransmission timer value is doubled each time when DTLS state machine entering next state despite there is no retransmission occurred. 

There is code which shows that if state is changed then `retransmit` value is doubled.

 ```
-module(test).
-include_lib("ssl/src/ssl_connection.hrl").
-compile(export_all).

close(_) -> ok.

start() ->
  application:start(crypto),
  application:start(asn1),
  application:start(public_key),
  application:start(ssl),
  Role = client,
  SslOpts0 = [{protocol, dtls}, {cb_info, {?MODULE,udp,udp_closed,udp_error,udp_passive}},
    {handshake, hello}, {verify_peer, none}],
  {ok, SslConfig} = ssl:handle_options(SslOpts0, Role, dummy),
  #config{ssl = SslOpts, transport_info = CbInfo} = SslConfig,

  DtlsStatem = proc_lib:spawn_link(dtls_connection, init, [[Role, host, port, {{peer, port}, #{role => Role}}, {SslOpts, #socket_options{}, []}, user, CbInfo]]),
  {initial_hello, #state{protocol_specific = #{flight_state := {retransmit, 1000}}} = State1} = sys:get_state(DtlsStatem),
  exit(DtlsStatem, normal),

  {keep_state, State2, _} = dtls_connection:hello(enter, dummy, State1),
  #state{protocol_specific = #{flight_state := {retransmit, 2000}}} = State2,

  {keep_state, State3, _} = dtls_connection:hello(enter, dummy, State2),
  #state{protocol_specific = #{flight_state := {retransmit, 4000}}} = State3,

  {keep_state, State4, _} = dtls_connection:certify(enter, dummy, State3),
  #state{protocol_specific = #{flight_state := {retransmit, 8000}}} = State4,

  {keep_state, State5, _} = dtls_connection:cipher(enter, dummy, State4),
  #state{protocol_specific = #{flight_state := {retransmit, 16000}}} = State5,

  ok.
```

The problem leads to long delay before retransmission at the final flights even if there are no retransmission actually happened at the previous ones. 

I propose the solution which increases timer value only when retransmission timeout have been fired. 

It is not clear for me how it should be tested in OTP therefore i expect your advises here. 

The related issue is here: https://github.com/erlang/otp/issues/7301

Thank you.

### DIFF
--- a/lib/ssl/src/dtls_connection.erl
+++ b/lib/ssl/src/dtls_connection.erl
@@ -734,11 +734,12 @@ handle_client_hello(#client_hello{client_version = ClientVersion} = Hello, State
 
 handle_state_timeout(flight_retransmission_timeout, StateName,
                      #state{protocol_specific = 
-                                #{flight_state := {retransmit, _NextTimeout}}} = State0) ->
+                                #{flight_state := {retransmit, CurrentTimeout}}} = State0) ->
     {State1, Actions0} = dtls_gen_connection:send_handshake_flight(State0, 
                                                                    retransmit_epoch(StateName, State0)),
-    {next_state, StateName, State, Actions} = 
+    {next_state, StateName, #state{protocol_specific = PS} = State2, Actions} =
         dtls_gen_connection:next_event(StateName, no_record, State1, Actions0),
+    State = State2#state{protocol_specific = PS#{flight_state => {retransmit, new_timeout(CurrentTimeout)}}},
     %% This will reset the retransmission timer by repeating the enter state event
     {repeat_state, State, Actions}.
 
@@ -813,7 +814,7 @@ handle_flight_timer(#state{protocol_specific = #{flight_state := reliable}} = St
     {State, []}.
 
 start_retransmision_timer(Timeout, #state{protocol_specific = PS} = State) ->
-    {State#state{protocol_specific = PS#{flight_state => {retransmit, new_timeout(Timeout)}}}, 
+    {State#state{protocol_specific = PS#{flight_state => {retransmit, Timeout}}},
      [{state_timeout, Timeout, flight_retransmission_timeout}]}.
 
 new_timeout(N) when N =< 30000 ->

### CODE REVIEW

**Comment by dgud** [comment]
Please change the author of the commit to your github username/email and force push and then sign the CLA is you have not done it already.

**Comment by Shaienn** [comment]
> Please change the author of the commit to your github username/email and force push and then sign the CLA is you have not done it already.

Done

File: lib/ssl/src/dtls_connection.erl [hunk 1]
```diff
@@ -734,13 +734,14 @@ handle_client_hello(#client_hello{client_version = ClientVersion} = Hello, State
...
+                                #{flight_state := {retransmit, CurrentTimeout}}} = State0) ->
     {State1, Actions0} = dtls_gen_connection:send_handshake_flight(State0, 
                                                                    retransmit_epoch(StateName, State0)),
-    {next_state, StateName, State, Actions} = 
+    {next_state, StateName, #state{protocol_specific = PS} = State, Actions} =
         dtls_gen_connection:next_event(StateName, no_record, State1, Actions0),
+    State2 = State#state{protocol_specific = PS#{flight_state => {retransmit, new_timeout(CurrentTimeout)}}},
     %% This will reset the retransmission timer by repeating the enter state event
-    {repeat_state, State, Actions}.
+    {repeat_state, State2, Actions}.
```
IngelaAndin, line 744: We have a convention to number the states and call the final one only State, so I would switch State2 and State!

Shaienn: Done

IngelaAndin:
Regarding the test this is not very easy to test, we are planning some activates to try to make a good way of testing such things in an automated way during next release cycle, and hopefully it will receive enough priority. So I think we could probably  make an exception accept this after some more code review and maybe some more manual test.

Shaienn:
> 



> Regarding the test this is not very easy to test, we are planning some activates to try to make a good way of testing such things in an automated way during next release cycle, and hopefully it will receive enough priority. So I think we could probably make an exception accept this after some more code review and maybe some more manual test.

Got it, thank you. I use OTP DTLS for WebRTC and this issue sometimes drastically increase connection time in case of network loss. So, this request is quite important for me.

IngelaAndin:
@Shaienn  It is regression testing fine. I guess you would like to rebase it on maint so that it can be part of the OTP-26 release cycle!?

Shaienn:
> @Shaienn It is regression testing fine. I guess you would like to rebase it on maint so that it can be part of the OTP-26 release cycle!?

Hello. I did rebase from `otp/maint`

IngelaAndin:
@Shaienn we have a review meeting on Wednesday and then I think this is ready for inclusion :)

PR #7302, opened 2023-05-25
[public_key] expose error in public_key:cacerts_get/0 if ca bundle is not found

### LINKED ISSUE #7295: [public_key] Dedicated error on missing cacerts file
**Describe the bug**

When calling `public_key:cacerts_get/0` and the operating system does not have a root ca file, the function crashes instead of returning the error tuple.

**To Reproduce**

```bash
docker run -ti hexpm/erlang:26.0-debian-bullseye-20230227-slim erl -eval 'public_key:cacerts_get()'
```

```
** exception error: no match of right hand side value {error,enoent}
     in function  pubkey_os_cacerts:get/0 (pubkey_os_cacerts.erl, line 38)
```

**Expected behavior**

It should return `{error, enoent}` to the caller.

**Affected versions**

`>= 25.0`

**Additional context**

`httpc:ssl_verify_host_options/1` should also be able to handle the error or return an error itself.

PS: Happy to provide a PR for the issue once the solution has been decided.

maennchen:
Fixes #7295

### Intentions

1. Replace a `badmatch` crash with a more descriptive one. `badmatch` feels like a bug to the user. (At least it did to me.)
2. Provide good error detail to the user about the reason of the crash.

### Implementation

I've implemented an extra match of the `load` result to help with 1 and introduced `format_error` which is used in the crash `error_info` for 2.

### DIFF
--- a/lib/public_key/src/pubkey_os_cacerts.erl
+++ b/lib/public_key/src/pubkey_os_cacerts.erl
@@ -23,7 +23,7 @@
 -module(pubkey_os_cacerts).
 
 -include("public_key.hrl").
--export([load/0, load/1, get/0, clear/0]).
+-export([load/0, load/1, get/0, clear/0, format_error/2]).
 
 -on_load(on_load/0).
 -nifs([os_cacerts/0]).
@@ -35,8 +35,16 @@
 get() ->
     case persistent_term:get(?MODULE, not_loaded) of
         not_loaded ->
-            ok = load(),
-            persistent_term:get(?MODULE);
+            case load() of
+                ok ->
+                    persistent_term:get(?MODULE);
+                {error, Reason} ->
+                    erlang:error(
+                        {failed_load_cacerts, conv_error_reason(Reason)},
+                        none,
+                        [{error_info, #{cause => Reason, module => ?MODULE}}]
+                    )
+            end;
         CaCerts ->
             CaCerts
     end.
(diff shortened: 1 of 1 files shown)


maennchen:
Should exit instead of returning error. Setting to draft to fix the PR.

dgud:
I wonder if we should add a describing atom as well to the error reason, 
something like {failed_load_cacerts, enoent | enosupp}

And also think we should add "Failed to load cacerts" to all describing error strings, not just for enoent.

dgud:
I will change this to go towards the master branch, in case someone is matching on the badmatch,
I do think this is more cosmetics and can wait (a long time) to the next major release.

The httpc patch is a real bug and I will try to get it into the next patch.

lib/public_key/src/pubkey_os_cacerts.erl
```diff
@@ -216,3 +224,35 @@ load_nif() ->
...
+%%%
+%%% Error Handling
+%%%
+
+conv_error_reason(enoent) -> enoent;
+conv_error_reason({enotsup, _OS}) -> enotsup;
+conv_error_reason({eopnotsupp, _Reason}) -> eopnotsupp;
+conv_error_reason({eopnotsupp, _Status, _Acc}) -> eopnotsupp.
+
+-spec format_error(Reason, StackTrace) -> ErrorMap when
```
IngelaAndin, line 237: So I guess the intention is that public_key:cacerts_load should call this function?

IngelaAndin: So I think I might have initially misunderstood.  I take this back, it is an other discussion!

maennchen: @IngelaAndin I've updated the PR description to include a description of intention / changes. Does this answer your question?

PR #7305, opened 2023-05-25
Return error if ssl is not started

### LINKED ISSUE #7297: `httpc:request("https://erlang.org")` hangs when ssl is not started
**To Reproduce**

```
$ erl
```

```erlang
> inets:start(), httpc:request("https://www.erlang.org").
%% hangs...
```

Setting the timeout does not help, it hangs too:

```erlang
> inets:start(), httpc:request(get, {"https://www.erlang.org", []}, [{timeout, 1000}], []).
%% hangs...
```

Starting ssl beforehand fixes the issue:

```erlang
> inets:start(), ssl:start(), httpc:request("https://www.erlang.org").
{ok, ...}
```

**Expected behavior**

An error tuple.

**Affected versions**
26.0, 25.3, possibly more

OTP 25.0 works as expected:

```
> inets:start(), httpc:request("https://www.erlang.org").
=WARNING REPORT==== 25-May-2023::10:32:34.000172 ===

{error,{failed_connect,[{to_address,{"www.erlang.org",443}},
                        {inet,[inet],ssl_not_started}]}}
```

dgud:
It hanged since the supervisor re-write, need to catch errors when use spawn on the startup function.

Fixes #7297

### DIFF
--- a/lib/ssl/test/ssl_api_SUITE.erl
+++ b/lib/ssl/test/ssl_api_SUITE.erl
@@ -177,6 +177,8 @@
          server_options_negative_dependency_role/1,
          invalid_options_tls13/0,
          invalid_options_tls13/1,
+         ssl_not_started/0,
+         ssl_not_started/1,
          cookie/0,
          cookie/1,
 	 warn_verify_none/0,
@@ -315,6 +317,7 @@ gen_api_tests() ->
      options_not_proplist,
      invalid_options,
      cb_info,
+     ssl_not_started,
      log_alert,
      getstat,
      warn_verify_none,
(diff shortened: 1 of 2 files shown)


lib/ssl/test/ssl_api_SUITE.erl
```diff
@@ -2638,6 +2641,19 @@ invalid_options_tls13(Config) when is_list(Config) ->
           end,
     [Fun(Option, ErrorMsg, Type) || {Option, ErrorMsg, Type} <- TestOpts].
 
+
+ssl_not_started() ->
+    [{doc, "Test that an error is returned if ssl is not started"}].
+ssl_not_started(Config) when is_list(Config) ->
+    application:stop(ssl),
+    try
+        {error, ssl_not_started} = ssl:connect("www.erlang.org", 443, []),
```
IngelaAndin, line 2650: Should we not have a valid option list in the test, even if might fail before it is checked! So [{verify, none}] should be added!

u3s (approved):

lib/ssl/test/ssl_api_SUITE.erl
dgud: Yes, I have obviously not tested this on a 26 branch, that would fail.

PR #7306, opened 2023-05-25
[httpc] calculate default ssl options on demand

### LINKED ISSUE #7303: [httpc] Do not read OS cacerts if `verify_none`
**Describe the bug**

When suplying `{ssl, [{verify, verify_none}]}` to `http`, it still calls `public_key:cacerts/0` internally. If there are no OS cacerts present on the machine, httpc will crash even though no certs were actually required.

**To Reproduce**

```erlang
inets:start(),
ssl:start(),
Response = httpc:request(get, {\"https://google.com\", []}, [{ssl,[{verify, verify_none}]}], []),
erlang:display(Response).
```

```
** exception error: no match of right hand side value {error,enoent}
     in function  pubkey_os_cacerts:get/0 (pubkey_os_cacerts.erl, line 38)
     in call from httpc:ssl_verify_host_options/1 (httpc.erl, line 476)
     in call from httpc:http_options_default/0 (httpc.erl, line 1012)
     in call from httpc:http_options/1 (httpc.erl, line 927)
     in call from httpc:handle_request/9 (httpc.erl, line 771)
```

**Expected behavior**

No Error

**Affected versions**

`>= 26.0`

**Additional context**

* Follow up of https://github.com/erlang/otp/issues/7295#issuecomment-1562635486

maennchen:
Avoids crash when using verify_none and the OS does not have a CA cert bundle.

I don't know how to properly test this since the only way I know to produce the reason for this PR is to not have any ca bundles on the system. That is relatively hard to reproduce in normal unit tests.

Fixes #7303

### DIFF
--- a/lib/inets/src/http_client/httpc.erl
+++ b/lib/inets/src/http_client/httpc.erl
@@ -955,10 +955,12 @@ http_options([{Tag, Default, Idx, Post} | Defaults], HttpOptions, Acc) ->
 	false ->
 	    DefaultVal = 
 		case Default of
-		    {value, Val} ->
-			Val;
-		    {field, DefaultIdx} ->
-			element(DefaultIdx, Acc)
+            {value, Val} ->
+                Val;
+            {value_lazy, ValFn} ->
+                ValFn();
+            {field, DefaultIdx} ->
+                element(DefaultIdx, Acc)
 		end,
 	    Acc2 = setelement(Idx, Acc, DefaultVal),
 	    http_options(Defaults, HttpOptions, Acc2)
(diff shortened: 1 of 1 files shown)


dgud:
Can you do `git rebase --onto OTP-26.0 cefbdf5c4b6` and force push so I can included it in the next patch.

maennchen:
@dgud Pushed

dgud:
Thanks, noticed that `httpc_SUITE:stream_fun_server_close/1` fails though, both here (on github) and in our nightly runs, so something is strange.

Have not had time to debug yet.

maennchen:
@dgud I believe that test to be failing unrelated to my change. It seems to be flaky (passes like 50% of the time on my local machine).

IngelaAndin:
> @dgud I believe that test to be failing unrelated to my change. It seems to be flaky (passes like 50% of the time on my local machine).

I believe @maennchen  is right about that!

dgud:
Hmm, ok but it failed on 19 runs this night and previously it have failed once or twice a week. 

Something is fishy.

http_transport:connect expects this:
`connect({ssl, SslConfig}, ... ) ->
`
And crashes with function clause:

```
=CRASH REPORT==== 26-May-2023::02:22:32.831322 ===
  crasher:
    initial call: httpc_handler:init/1
    pid: <0.2579.0>
    registered_name: []
    exception error: no function clause matching 
                     http_transport:connect([{verify,verify_peer},
                                             {cacerts,
                                              [{cert,
                                                <<48,130,4,48,48,130,3,24,160,
                                                  3,2,1,2,2,16,80,148,108,236,
                                                  24,234,213,156,77,213,151,
                                                  239,117,143,160,173,48,13,6,
                                                  9,42,134,72,134,247,13,1,1,5,
                                                  5,0,48,129,130,49,11,48,9,6,
                                                  3,85,4,6,19,2,85,83,49,30,48,
                                                  28,6,3,85,4,11,19,21,119,119,`
```

maennchen:
@dgud You are right, I found a bug. I'll push a fix in a few minutes.

maennchen:
@dgud / @IngelaAndin I forgot to wrap the return value of the `SslOptsLazyFn` into a `{ssl, ...}` tuple. That is now fixed.

PR #7310, opened 2023-05-26
Document loading httpd mime_types from file

jchristgit:
Re-reading this, a possible improvement could be trying to load mime types from `/etc/mime.types` as a fallback if it exists, as this seems to be the default location at least on Debian.

https://github.com/erlang/otp/blob/646be911ba326bfffd04042297389ebcb7753fa4/lib/inets/src/http_server/httpd_conf.erl#L347-L371

### DIFF
--- a/lib/inets/doc/src/httpd.xml
+++ b/lib/inets/doc/src/httpd.xml
@@ -280,7 +280,13 @@
           1590. File suffixes are mapped to MIME types before file delivery.
           The mapping between file suffixes and MIME types can be specified
 	  in the property list.</p>
-	<p>Default is [{"html","text/html"},{"htm","text/html"}].</p>
+        <p>Mime types can also be read from a file. The file should contain
+          lines in the form <c>MediaType [Extensions...]</c>, such as
+          <c>text/html html htm</c>. To configure this, specify
+          the path to it, such as <c>{mime_types, "/etc/mime.types"}</c>.</p>
+        <p>If unset, <c>conf/mime.types</c> under <c>server_root</c>
+          will be used if it exists, otherwise, the default is
+          <c>[{"html","text/html"},{"htm","text/html"}]</c>.</p>
       </item>
 
       <tag><marker id="prop_mime_type"></marker>{mime_type, string()}</tag>


IngelaAndin:
Did you try that this works? (I might if it is compatible to how Apache works)  We need a fallback that will work on all platforms that we support, but for a good enough reason we could perhaps have several fallback steps.  I think it is nice if the simplest configuration is not dependent on external configuration files.

jchristgit:
Loading mime types from a file works with `/usr/local/etc/nginx/mime.types` on my Mac:

```sh
$ erl -s inets -eval '{ok, Pid} = inets:start(httpd,[{server_name,"NAME"},{document_root, "."},{server_root,"."},{port, 8000},{mime_types,["/usr/local/etc/nginx/mime.types"]}]).'

3> inets:services().
[{httpc,<0.89.0>},{httpd,<0.92.0>}]
4> httpd:info(list_to_pid("<0.92.0>")). 
[{mime_types,[{"css;","text/css"},
              {"midi","audio/midi"},
              {"dmg;","application/octet-stream"},
              {"wml;","text/vnd.wap.wml"},
              {"img;","application/octet-stream"},
              {"pdb;","application/x-pilot"},
              {"flv;","video/x-flv"},
              {"eps","application/postscript"},
              {"ico;","image/x-icon"},
              {"msm;","application/octet-stream"},
              {"jad;","text/vnd.sun.j2me.app-descriptor"},
              {"prc","application/x-pilot"},
              {"jar","application/java-archive"},
              {"htc;","text/x-component"},
              {"jpg;","image/jpeg"},
              {"m3u8;","application/vnd.apple.mpegurl"},
              {"pem","application/x-x509-ca-cert"},
              {"cco;","application/x-cocoa"},
              {"pdf;","application/pdf"},
              {"rpm;","application/x-redhat-package-manager"},
              {"mov;","video/quicktime"},
              {"ogg;","audio/ogg"},
              {"ear;","application/java-archive"},
              {"tif",[...]},
              {[...],...},
              {...}|...]},
 {server_name,"NAME"},
 {bind_address,any},
 {server_root,"."},
 {port,8000},
 {document_root,"."},
 {ipfamily,inet}]
```

I could reproduce the same with `/etc/mime.types` on Raspbian.

jchristgit:
Hmm, sorry, re-reading your comment I think I misunderstood it a bit.

There is currently no fallback implemented with this PR, that was mainly an idea I had whilst making it to make it easier to configure. In https://github.com/erlang/otp/pull/7299/files#diff-d8239d05391b50fc317f0ce1f9ea048a8011210ebd3361e31f60e379c970fe13R407-R428 I have added some mime type file searching with a fallback (if none were found) which we could maybe integrate here instead.

IngelaAndin:
We can keep it as separate PRs, I think the documentation can probably be accepted fairly soon! I was commenting your comment, rather than the change in this PR.

jchristgit:
Sounds good, thank you 👍

IngelaAndin:
I was thinking should we perhaps mention something about the expected file format?! I think that it is probably a reason why it was not documented, and we might have done that when dropping apache-config files for main configuration but we did not exactly know what to do about the mime-types!

jchristgit:
I updated the commit to include a bit of info on the file format, as taken from `/etc/mime.types` on Debian.

PR #7311, opened 2023-05-26
[dialyzer] Tolerate is_function/2 with invalid arity

### LINKED ISSUE #7181: [dialyzer] crash on is_function/2 with large negative arity
**Describe the bug**
Running dialyzer on the following:
```erlang
f() when is_function(ok, -9223372036854775808) ->
    ok.
```
results in the following error:
```
dialyzer: Analysis failed with error:
{function_clause,[{erl_types,t_fun,
                             [-9223372036854775808,any],
                             [{file,"erl_types.erl"},{line,988}]},
                  {dialyzer_typesig,'-get_bif_constr/4-lc$^9/1-0-',1,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,1371}]},
                  {dialyzer_typesig,'-get_bif_constr/4-fun-10-',3,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,1371}]},
                  {dialyzer_typesig,solve_one_c,2,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,2232}]},
                  {dialyzer_typesig,v2_solve,3,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,1860}]},
                  {dialyzer_typesig,v2_solve_conj,12,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,2067}]},
                  {dialyzer_typesig,v2_solve_disj,10,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,1995}]},
                  {dialyzer_typesig,v2_solve_disjunct,3,
                                    [{file,"dialyzer_typesig.erl"},
                                     {line,1958}]}]}
```

**Affected versions**
master

**Additional context**
Here is an alternate test case that seems to trigger the same bug:
```erlang
f(_V0) ->
    (is_function(ok, 9223372036854775807) andalso _V0) bor _V0.
```

RobinMorisset:
Fixes #7181.

Functions in Erlang have a max arity of 255 (see
https://www.erlang.org/doc/reference_manual/typespec.html), so `is_function(Foo, 999)` will always fail, and obviously no function can have negative arity so `is_function(Foo, -100)` will also always fail.

This patch just teaches dialyzer to deal with such obviously wrong code in a more graceful manner (i.e. without crashing).

I've tried putting the tests in a reasonable place, but there might be a better file for them to live in.

### DIFF
--- a/lib/dialyzer/test/small_SUITE_data/src/guards.erl
+++ b/lib/dialyzer/test/small_SUITE_data/src/guards.erl
@@ -145,6 +145,14 @@ t17(X) when #{x => X} ->
 t18(X) when <<X>> ->
     ok.
 
+%% Should give a warning, see https://github.com/erlang/otp/issues/7181
+t19() when is_function(ok, -9223372036854775808) ->
+    never.
+
+% Could give a warning, see the same link
+t20(X) ->
+    (is_function(ok, 9223372036854775807) andalso X) bor X.
+
 %% Coverage
 
 cover1() when 1 < 2; 1 > 2 ->
--- a/lib/dialyzer/src/dialyzer_dataflow.erl
+++ b/lib/dialyzer/src/dialyzer_dataflow.erl
@@ -1998,7 +1998,7 @@ handle_guard_is_function(Guard, Map, Env, Eval, State) ->
 	case t_number_vals(ArityType, State#state.opaques) of
 	  unknown -> t_fun();
 	  Vals ->
-	    t_sup([t_fun(lists:duplicate(X, t_any()), t_any()) || X <- Vals])
+	    t_sup([t_fun(lists:duplicate(X, t_any()), t_any()) || X <- Vals, X >= 0, X =< 255])
 	end,
       FunType = t_inf(FunType0, FunTypeConstr, Opaques),
       case t_is_none(FunType) of
--- a/lib/dialyzer/src/dialyzer_typesig.erl
+++ b/lib/dialyzer/src/dialyzer_typesig.erl
@@ -1368,7 +1368,7 @@ get_bif_constr({erlang, is_function, 2}, Dst, [Fun, Arity], _State) ->
 		   ArityType = lookup_type(Arity, Map),
 		   case t_number_vals(ArityType) of
 		     unknown -> t_fun();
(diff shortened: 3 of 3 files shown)


lib/dialyzer/test/small_SUITE_data/src/guards.erl
```diff
@@ -145,6 +145,14 @@ t17(X) when #{x => X} ->
 t18(X) when <<X>> ->
     ok.
 
+%% Should give a warning, see https://github.com/erlang/otp/issues/7181
+t19() when is_function(ok, -9223372036854775808) ->
+    never.
```
bjorng, line 150: There will indeed be a warning:

```
guards.erl:149:12: Guard test is_function
         ('ok',
          -9223372036854775808) can never succeed
```

To ensure that the test case will not fail, you will have to add this warning to [file with expected output](https://github.com/erlang/otp/blob/master/lib/dialyzer/test/small_SUITE_data/results/guards).

lib/dialyzer/test/small_SUITE_data/src/guards.erl
```diff
@@ -145,6 +145,14 @@ t17(X) when #{x => X} ->
...
+
+% Could give a warning, see the same link
+t20(X) ->
+    (is_function(ok, 9223372036854775807) andalso X) bor X.
```
bjorng, line 154: The following warnings will be generated:

```
guards.erl:153:1: Function t20/1 has no local return
guards.erl:154:22: The call erlang:is_function
         ('ok',
          9223372036854775807) will never return since it differs in the 2nd argument from the success typing arguments: 
         (any(),
          byte())
```

This is not correct, because this call to `is_function/2` will not fail but return `false`:

```
1> is_function(ok, 9223372036854775807).
false
```

A negative arity will cause the call to fail:

```
2> is_function(ok, -1).
     in function  is_function/2
        called as is_function(ok,-1)
        *** argument 2: out of range
```

bjorng (changes requested):
Thanks for your pull request!

When you've addressed my comments, please rebase the branch to `maint`.

A tip about comments such as `Fixes #7181`: If you put them at the end of the commit message, and if the pull request is for the default branch of the repository (`master` in our case), the pull request will automatically be linked to the issue.

PR #7313, opened 2023-05-27
EEP 64: Triple-Quoted Strings

williamthome:
This PR covers the [EEP 64: Triple-Quoted Strings](https://github.com/erlang/eep/pull/47).

> **Warning**
> The EEP is currently under discussion and probably would be updated/modified.

### DIFF
--- a/lib/compiler/src/core_scan.erl
+++ b/lib/compiler/src/core_scan.erl
@@ -122,6 +122,41 @@ format_error(char) -> "unterminated character";
 format_error(scan) -> "premature end";
 format_error({base,Base}) -> io_lib:fwrite("illegal base '~w'", [Base]);
 format_error(float) -> "bad float";
+format_error({triple_quoted_string,syntax}) ->
+    % Elixir's example of this error:
+    % iex(1)> """foo
+    % ** (SyntaxError) iex:1:1: heredoc allows only zero or more whitespace characters followed by a new line after """
+    %   |
+    % 1 | """foo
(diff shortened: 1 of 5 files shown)


kikofernandez:
Hi @williamthome , before we can proceed, we think it is advisable to create an EEP that details how things will be handled.
I will try to write such EEP (unless you prefer to do so) and the main idea is that we would like to get as close as seems reasonable to Elixir's semantics. In this sense, the triple quotes will create a binary instead of a string.

More details about the EEP coming soon, as soon as I figure out the semantics of Elixir :)

lib/compiler/src/core_scan.erl
```diff
@@ -357,6 +360,82 @@ name_char($_) -> true;
...
+
+% TODO: Improve error handling
+scan_heredocs(Cs) ->
+    scan_heredocs_1(heredocs_trim_leading(Cs), false, []).
+
+scan_heredocs_1([$",$",$"|Cs], {true, _} = I, Acc) ->
+    {Result, _} =
+        lists:foldl(
+            fun
+            % FIXME: We can have spaces between \ and \n
```
williamthome, line 373: This comment makes no sense. The `\` should be the last char before `\n`. I'll remove it.

williamthome:
I'm closing this PR in favor of the fantastic implementation of @RaimoNiskanen on #7451.
I'm happy to see the triple-quoted string in OTP-27.0 milestones :partying_face:

PR #7316, opened 2023-05-27
inets: Make httpd server_name setting optional

jchristgit:
By default, httpd will now use the system DNS name, if available. This makes it easier to start httpd for cases where the server name isn't important. Suggestion came from here:
https://erlangforums.com/t/httpc-httpd-improvements/2622/14


lib/inets/src/http_server/httpd_conf.erl
```diff
@@ -75,12 +75,23 @@ load_mime_types(MimeTypesFile) ->
 	    {error, ?NICE("Can't open " ++ MimeTypesFile)}
     end.
 
+%% Add any optional properties if they are not present.
+optional_properties(Properties) ->
+    case proplists:get_value(server_name, Properties) of
+        undefined ->
+            {ok, Fqdn} = net_adm:dns_hostname(net_adm:localhost()),
```
IngelaAndin, line 82: I believe net_adm:localhost()  would be a better default as it would work for nodes with short names. 

Despription of  net_adm:localhost(): 
Returns the name of the local host. If Erlang was started with command-line flag -name, Name is the fully qualified name.

jchristgit: Thanks for the review.

Revisiting this I actually was not quite sure why I wrapped it into `dns_hostname`. I have amended the commit to only use `net_adm:localhost()` on its own.

lib/inets/doc/src/httpd.xml
```diff
@@ -128,6 +123,13 @@
 	<p>Default is <c>any</c></p>
       </item>
 
+      <tag><marker id="prop_server_name"></marker>{server_name, string()}</tag>
+      <item>
+	<p>The name of your server, normally a fully qualified domain name.</p>
+        <p>If not given, this defaults to the fully qualified hostname as obtained
+          from <c>net_adm:localhost()</c>.</p>
+      </item>
```
IngelaAndin, line 131: We should probably only say defaults to net_adm:local_host(), as it will only be a qualified hostname if you started the node with -name if started with -sname it will not be qualified.  But we could leave that for the net_adm:localhost doc :)

jchristgit: Good point. I updated my commit to just mention it defaults to `net_adm:localhost()`.

IngelaAndin:
I think it looks good, but before we decide to merge this I think you should also take a look at the User Guide doc.

https://www.erlang.org/doc/apps/inets/http_server.html#configuration

u3s (approved):

jchristgit:
> @jchristgit that is should we add/change something here with regard to you update?!

Absolutely :) I updated the "Start with minimal required configuration" to remove the `server_name` param.

PR #7322, opened 2023-05-29
Drop support in the compiler for OTP 22/23

bjorng:
The compiler has undocumented options for generating code for older releases, so that we can test that the runtime system can load code compiled for a previous release.

Remove the support for generating code that can be loaded on OTP 22/23. The support in the runtime system for loading code compiled for OTP 22/23 will be dropped in a separate pull request.

### DIFF
--- a/lib/compiler/src/beam_asm.erl
+++ b/lib/compiler/src/beam_asm.erl
@@ -65,7 +65,7 @@ assemble({Mod,Exp0,Attr0,Asm0,NumLabels}, ExtraChunks, CompileInfo, CompilerOpts
     {1,Dict0} = beam_dict:atom(Mod, beam_dict:new()),
     {0,Dict1} = beam_dict:fname(atom_to_list(Mod) ++ ".erl", Dict0),
     {0,Dict2} = beam_dict:type(any, Dict1),
-    Dict3 = shared_fun_wrappers(CompilerOpts, Dict2),
+    Dict3 = reject_unsupported_versions(Dict2),
     NumFuncs = length(Asm0),
     {Asm,Attr} = on_load(Asm0, Attr0),
     Exp = sets:from_list(Exp0, [{version, 2}]),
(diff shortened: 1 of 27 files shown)


lib/compiler/src/beam_asm.erl
```diff
@@ -65,30 +65,22 @@ assemble({Mod,Exp0,Attr0,Asm0,NumLabels}, ExtraChunks, CompileInfo, CompilerOpts
...
-            Dict
-    end.
+shared_fun_wrappers(Dict) ->
+    %% The compiler in OTP 23 depends on the on the loader using the
+    %% new indices in funs and being able to have multiple make_fun2
+    %% instructions referring to the same fun entry. Artificially set
+    %% the highest opcode for the module to ensure that it cannot be
+    %% loaded in OTP 22 and earlier.
+    Swap = beam_opcodes:opcode(swap, 2),
+    beam_dict:opcode(Swap, Dict).
```
jhogberg, line 83: Since we're removing support for OTP 23 as well, wouldn't it make sense to unconditionally set this to lowest supported version (e.g. `make_fun3` introduced in OTP 24)?

```suggestion
reject_unsupported_versions(Dict) ->
    %% Emit an instruction that was added in our lowest supported
    %% version so that it cannot be loaded by earlier releases.
    true = (?OTP_RELEASE - 3) =:= 24, %Assertion.
    Swap = beam_opcodes:opcode(make_fun3, 2),
    beam_dict:opcode(Swap, Dict).
```

jhogberg (approved):

lib/compiler/src/beam_asm.erl
bjorng: I like idea except for the assertion, which means that when a release is cut the `master` branch will no longer have a working compiler. Releases can be stressful enough without an auto-triggering assertion. Besides, raising the lowest supported release involves some work in our testing environment, so we might not want to raise it in every release.

PR #7326, opened 2023-05-30
Add matchspec BIFs maps_put and maps_remove

### LINKED ISSUE #7309: Add support to in place add/remove key/values of map in ets:select_replace with match spec

ets:select_replace/2 can use match spec to do in place replacement with List, which is efficient. But it doesn’t support updating map in place.

For example, the change to a list works:
`ets:fun2ms(fun({K, L}) when is_list(L) → {K, [marker | L]} end).`
Output: 
`[{{‘$1’,‘$2’},[{is_list,‘$2’}],[{{‘$1’,[marker|‘$2’]}}]}]`

But for something similar with map:
`ets:fun2ms(fun({K, Map}) when is_map(Map) → {K, Map#{marker => true}} end).`
Output:
`Error: the language element map (in body) cannot be translated into match_spec
{error,transform_error}`

Is Erlang/OTP team considering adding support to map as well?

ets:select_replace/2 can use match spec to efficiently in place add/remove key/values from a map.

Right now in order to update a map, it needs to read from ets, make the changes, then copy the new map back to ets, which is very inefficient for bigger map. Since Erlang/OTP already support select by key with match spec, it would be consistent to support modification as well. 

**Additional context**

sverker:
corresponding to maps:put/3 and maps:remove/2.

This is a quick attempt to fix #7309.

The map syntax to update keys cannot simply be used as `Map#{key => value}` would be translated by fun2ms to something like `'$1'#{key => value}` which is not valid Erlang syntax. Instead use the match spec syntax to call BIFs.

```
1> T = ets:new(x,[]).

2> ets:insert(T, {key, #{a=>1, b=>2}}).

3> MS = ets:fun2ms(fun({Key, Map}) when is_map(Map) -> {Key, maps_put(c, 3, Map)} end).
[{{'$1','$2'},
  [{is_map,'$2'}],
  [{{'$1',{maps_put,c,3,'$2'}}}]}]

4> ets:select_replace(T, MS).
1
5> ets:tab2list(T).
[{key,#{c => 3,a => 1,b => 2}}]
```

ToDo: Test and documentation.

### DIFF
--- a/erts/emulator/beam/atom.names
+++ b/erts/emulator/beam/atom.names
@@ -416,6 +416,8 @@ atom Lt='<'
 atom machine
 atom magic_ref
 atom major
+atom maps_put
+atom maps_remove
 atom match
 atom match_limit
 atom match_limit_recursion
--- a/erts/emulator/beam/erl_db_util.c
+++ b/erts/emulator/beam/erl_db_util.c
@@ -697,6 +697,18 @@ static DMCGuardBif guard_tab[] =
         DBIF_ALL
     },
     {
+        am_maps_put,
+        &maps_put_3,
+        3,
+        DBIF_TABLE_BODY
+    },
+    {
+        am_maps_remove,
+        &maps_remove_2,
+        2,
+        DBIF_TABLE_BODY
+    },
+    {
         am_is_map_key,
         &is_map_key_2,
         2,
--- a/lib/stdlib/src/ms_transform.erl
+++ b/lib/stdlib/src/ms_transform.erl
@@ -1020,6 +1020,8 @@ action_function(trace,3) -> true;
 action_function(caller_line,0) -> true;
 action_function(current_stacktrace,0) -> true;
 action_function(current_stacktrace,1) -> true;
+action_function(maps_put,3) -> true;
+action_function(maps_remove,2) -> true;
 action_function(_,_) -> false.
 
 bool_operator('and',2) ->


michalmuskala:
Could it be that fun2ms translates the map update syntax into `map_put` calls?

sverker:
> Could it be that fun2ms translates the map update syntax into `map_put` calls?

It could I suppose. I'm not familiar with the fun2ms code though.
You would still need to call maps_remove directly.

yushli:
Vote for 

> fun2ms translates the map update syntax into map_put calls

christhekeele:
A couple of thoughts:

- Wouldn't it be better to have this be `map_put` and `map_remove`?

    Today's `map_get` and `map_size` equivalents do not pluralize `maps`.

- Should these be added as general auto-imported BIFs?

    Today's `map_get` and `map_size` are named the same as auto-imported BIFs [`map_get/2`](https://www.erlang.org/doc/man/erlang#map_get-2) and [`map_size/2`](https://www.erlang.org/doc/man/erlang#map_size-1).  Do we want this to be internally consistent or just a secret capability of match specs? Or are there equivalent auto-imported BIFS in `master` and I'm just behind the times?

- My own (extreme) ignorance showing here in how this works under the hood, but [the match spec docs](https://www.erlang.org/doc/apps/erts/match_spec) suggest that [action functions](https://github.com/erlang/otp/pull/7326/files#diff-e749e3266505cdf1c43cf6273a06ef2663d85fad8bb25b7b0b927d9c17a32fccR1023) are only allowed in `dbg`-dialect ms's. Would this implementation only allow them in `dbg`-dialects? Or would this be the first set of "action functions" allowed in the `ets` dialect?

sverker:
> * Wouldn't it be better to have this be `map_put` and `map_remove`?

Yes maybe. I called them `maps_put` and `maps_remove` to refer to `maps:put` and `maps:remove`.

> * Should these be added as general auto-imported BIFs?
>   Today's `map_get` and `map_size` are named the same as auto-imported BIFs [`map_get/2`]

Maybe. I did this as a quick proof of concept without too much thought of the naming.

> * My own (extreme) ignorance showing here in how this works under the hood, but [the match spec docs](https://www.erlang.org/doc/apps/erts/match_spec) suggest that [action functions](https://github.com/erlang/otp/pull/7326/files#diff-e749e3266505cdf1c43cf6273a06ef2663d85fad8bb25b7b0b927d9c17a32fccR1023) are only allowed in `dbg`-dialect ms's. Would this implementation only allow them in `dbg`-dialects? Or would this be the first set of "action functions" allowed in the `ets` dialect?

I don't remember why I added them to ms_transform:action_function/1. They are not action functions in the sense of side effects. `maps_put` and `maps_remove` are pure functions returning new maps and will fit in with the existing ets match spec paradigm.

PR #7329, opened 2023-05-31
compiler: Refine too strict assert in beam_ssa_pp:format_tuple_set_1/1

frej:
The assertion in `beam_ssa_pp:format_tuple_set_1/1` is not always true, as the type stored in the `Elems` map doesn't need to be exactly the same type as `Key`, they just have to be compatible.

The assert can be triggered by running `./scripts/diffable --asm --co 'dssaopt' $TMPDIR`.


bjorng (approved):

bjorng:
Thanks for your pull request.

PR #7330, opened 2023-05-31
Avoid optimization that beam_validator might consider unsafe

### LINKED ISSUE #6599: [erlc] internal consistency check failed: "illegal_branch, try_handler"

bjorng:
Eliminating a repeated `=:=` test can result in code that `beam_validator` doesn't consider safe. This pull request makes the optimization keep `=:=` if it seems to be significant, that is, if one or both operands will gain type information from it.
    
That makes it possible to compile code from #6599 such as:
    
        f(X, X) when is_integer(X) ->
            ok;
        f(Y, Y = #{}) ->
            Y#{ok := ok}.
    
(Six other examples from #6599 can now also be compiled.)
    
Running `scripts/diffable` shows that this fix prevented the optimization in 14 modules (out of about 1000).
    
Closes #6599 (the examples that were not fixed by this PR will be transferred to a new issue)

### DIFF
--- a/lib/compiler/src/beam_ssa_dead.erl
+++ b/lib/compiler/src/beam_ssa_dead.erl
@@ -1176,9 +1176,19 @@ opt_redundant_tests_is([#b_set{op=Op,args=Args,dst=Bool}=I0], Tests, Acc) ->
         {Test,MustInvert} ->
             case old_result(Test, Tests) of
                 Result0 when is_boolean(Result0) ->
-                    Result = #b_literal{val=Result0 xor MustInvert},
-                    I = I0#b_set{op={bif,'=:='},args=[Result,#b_literal{val=true}]},
-                    {old_test,reverse(Acc, [I]),Bool,Result};
+                    case gains_type_information(I0) of
+                        false ->
+                            Result = #b_literal{val=Result0 xor MustInvert},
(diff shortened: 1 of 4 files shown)


lib/compiler/src/beam_ssa_dead.erl
```diff
@@ -1187,6 +1197,33 @@ opt_redundant_tests_is([I|Is], Tests, Acc) ->
...
+        0 ->
+            false;
+        1 ->
+            true;
+        2 ->
+            case Types of
+                #{0 := Same,1 := Same} ->
+                    false;
+                #{} ->
+                    true
```
kikofernandez, line 1215: this case was already covered by `map_size/1`, which would fall under the `0` size case.
so it cannot be that `#{}` is reachable, because `map_size(Types)` rules it out. At the same time, when size is 0 we return false, and when size is 0 for #{} we return true. 

I think there is something that may not be correct?

PR #7334, opened 2023-05-31
Add link to appup manpage in release handling doc

jchristgit:
Additionally, document that `UpFromVsn` and `DownToVsn` can also be a regular expression (explained in detail in the `appup` reference): the current document implies, to me, that previous versions must be a fixed string, whilst the appup file can deal with regular expressions as well.


bjorng:
Thanks for the pull request.

PR #7343, opened 2023-06-01
feature: String interpolation

TD5:
Adds four kinds of string interpolation split over two axes (utf-8 binary or unicode codepoint list, and user-facing or developer-facing formatting).

The result are four general classes of syntax with interpolated values:

```
% binary format
<<"A utf-8 binary string: 4"/utf8>> =
  bf"A utf-8 binary string: ~2 + 2~"
```

```
% list format
"A unicode codepoint list string: 4" =
  lf"A unicode codepoint list string: ~2 + 2~"
```

```
% binary debug
<<"A utf-8 binary string: {4, foo, [x, y, z]}"/utf8>> =
  bd"A utf-8 binary string: ~{2 + 2, foo, [x, y, z]}~"
```

```
% list debug
"A unicode codepoint list string: {4, foo, [x, y, z]}" =
  ld"A unicode codepoint list string: ~{2 + 2, foo, [x, y, z]}~"
```

Arbitrary expressions can be nested inside string interpolation substitutions, including variables, function calls, macros and even further string interpolation expressions.

Design
======

Why list- and binary-strings?
-----------------------------

In the `string` module from the stdlib, a string is represented by `unicode:chardata()`, that is, a list of codepoints, binaries with UTF-8-encoded codepoints (UTF-8 binaries), or a mix of the two.

With this in mind, the list- and binary-oriented string interpolation syntaxes accept either type of interpolated value, but the user of the interpolation determines whether they want to generate a `unicode:char_list()` or `unicode:unicode_binary()` based on which kind of interpolation they use (`bf"..."` and `bd"..."` to create binaries, or `lf"..."` and `ld"..."` to create lists).

List-strings are most useful for backwards compatibility and convenience. Binary-strings are most useful for memory-compactness and IO.

Why user- and developer-oriented strings?
-----------------------------------------

There are two similar, but distinct cases where developers typically want to format strings: when logging/debugging, and when displaying data to users.

When logging or debugging, the most important features are typically that any kind of term can be printed, and it should round-trip losslessly and be read by developers unambiguously. Examples of these properties are, for example, retaining runtime type information, e.g. keeping strings quoted when formatting them and printing floats with full range and resolution.

When displaying to users, the most important features are typically that they are always going to be human-readable and cleanly formatted. Examples of these properties are, for example, formatting strings verbatim, without quotation marks, and not retaining any Erlang-isms (e.g. we don't want to be printing Erlang tuples, because they won't make much sense to the average application consumer), so we'd rather get a `badarg` error to push the developer to make an explicit formatting decision.

Why no formatting options?
--------------------------

Let's consider the two use-cases introduced earlier:

- Logging/debugging: Typically you want to fire-and-forget, giving whatever value you care about to the formatter, and just let it print that value unambiguously, meaning there's no need to tweak formatting options: `bd"~Timestamp~: ~Query~ returned ~Result~"`
- Displaying to users: Typically you want to tightly control formatting, and you probably want to do so in a modular and reusable way. In that case, factoring out your formatting decision to a function, and interpolating the result of that function is probably the best way to go: `bf"You account balance is now ~my_app:format_balance(Currency, Balance)~"`.

Notably, nothing in the design and implementation here precludes the future introduction of formatting options such as `bf"float: ~.2f(MyFloat)~"` as one might do with `io_lib:format` etc. But existing stdlib functions can offer similar functionality, e.g. `bf"float: ~float_to_binary(MyFloat, [{decimals, 2}, compact])~"`, and can be factored out into their own reusable functions.

Implementation
==============

To parse interpolated strings, the scanner tracks some additional state regarding whether we are currently in an interpolated string, at which point it enables the recognition of `~` as the delimiter for interpolated expressions, and generates new tokens which represent the various components of an interpolated string.

Early during compilation and shell evaluation, interpolated strings are desugared into calls to functions from the `io_lib` module, and therefore don't impact later stages of compilation or evalution.

The new string interpolation syntax was not previously valid syntax, so should be entirely backwards compatible with existing source code.

### DIFF
--- a/lib/compiler/src/compile.erl
+++ b/lib/compiler/src/compile.erl
@@ -789,6 +789,7 @@ make_ssa_check_pass(PassFlag) ->
 
 standard_passes() ->
     [?pass(transform_module),
+     ?pass(desugar_interpolation),
 
      {iff,makedep_side_effect,?pass(makedep_and_output)},
      {iff,makedep,[
@@ -1248,6 +1249,9 @@ strip_columns(Code) ->
          Form ->
              erl_parse:map_anno(F, Form)
      end || Form <- Code].
+
+desugar_interpolation(Code, #compile{options=Opt}=St) ->
+  {ok, erl_desugar_interpolation:module(Code, Opt), St}.
 
 get_core_transforms(Opts) -> [M || {core_transform,M} <- Opts].
 
--- a/lib/stdlib/src/Makefile
+++ b/lib/stdlib/src/Makefile
@@ -67,6 +67,7 @@ MODULES= \
 	erl_error \
 	erl_eval \
 	erl_expand_records \
+	erl_desugar_interpolation \
 	erl_features \
 	erl_internal \
 	erl_lint \
--- a/lib/stdlib/src/epp.erl
+++ b/lib/stdlib/src/epp.erl
@@ -1930,6 +1930,10 @@ token_src({char,_,C}) ->
 token_src({string, _, X}) ->
     io_lib:write_string(X);
 token_src({_, _, X}) ->
+    io_lib:format("~w", [X]);
+token_src({_, _, _, X}) ->
+    io_lib:format("~w", [X]);
+token_src({_, _, _, _, X}) ->
     io_lib:format("~w", [X]).
 
 stringify1([]) ->
--- a/lib/stdlib/src/erl_desugar_interpolation.erl
+++ b/lib/stdlib/src/erl_desugar_interpolation.erl
@@ -1,39 +1,17 @@
-%% ``Licensed under the Apache License, Version 2.0 (the "License");
-%% you may not use this file except in compliance with the License.
-%% You may obtain a copy of the License at
-%%
-%%     http://www.apache.org/licenses/LICENSE-2.0
-%%
-%% Unless required by applicable law or agreed to in writing, software
-%% distributed under the License is distributed on an "AS IS" BASIS,
-%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-%% See the License for the specific language governing permissions and
-%% limitations under the License.
-%% 
-%% The Initial Developer of the Original Code is Ericsson Utvecklings AB.
-%% Portions created by Ericsson are Copyright 1999, Ericsson Utvecklings
-%% AB. All Rights Reserved.''
-%% 
-%%     $Id$
-%%
--module(erl_id_trans).
+%% Purpose: Expand list-/binary-string interpolation expression syntax into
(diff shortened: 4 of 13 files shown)


TD5:
I see there are a bunch of test failures outside the tests I ran locally. I'll investgate.

kikofernandez:
Thanks for this contribution. 

At the moment, this PR involves many small decisions that need to be consistent and well-thought, e.g., symbol for sigils, complexity of the lexer and parser given that it is defined in a recursive manner, etc.

I do not think we can figure all these details before OTP-27, so we mark it as "stalled" and we will take small steps until all the design decisions are 100% clear.

rvirding:
Which checks failed and why?

mikpe:
1. I see that a lot of changes spill to parts outside of the scanner and parser proper. Why? The feature should be reducible to the base language, and the parser should be able to hide that translation from the rest of the system.
2. It seems the new feature uses a new term formatter not available via `io_lib:format`. That doesn't seem right to me.

TD5:
> The feature should be reducible to the base language, and the parser should be able to hide that translation from the rest of the system.

It's a bit of a trade-off. As I see it, the options are:
 * Confine it to the parser, and therefore have the parser return abstract forms which do not closely match the true source syntax
 * Let the parser handle _source syntax_, and let later stages of the compiler perform re-writes. Notably, preserving the high-level structure can also allow further optimisations to be applicable, since, for example, we are guaranteed to know the number of interpolated sub-expressions to be formatted statically, which isn't true of formatting in general (e.g. with `io_lib:format`, since its even the arity of its arguments can be determined dynamically).
 
Regarding the other spillage, that's for supporting linting, the compiler's partial evaluation, etc. All things which I think should be updated for a change like this.

> It seems the new feature uses a new term formatter not available via `io_lib:format`. That doesn't seem right to me.

This is essentially because the original `io_lib:format` formatter lacked features to meet the design I laid out in the EEP. The implementation here adds new functions to the `io_lib` module which implement these. Notably, these new functions are aware of types with a "natural user-facing format", and the implementation makes use of an explicit accumulator for binaries which leverages recent optimisations for binaries in the runtime. This means we get the existing benefits of binaries (compactness & locality), plus efficient construction (which is optimised by the runtime into mutations rather than naive copying). What's more, the information is there statically to allow even more future optimisations, for example by pre-allocating the binary to minimise re-sizing of the underlying mutable binary.

fnchooft:
Good Morning, are there any plans to advance on this Feature?

TD5:
> Good Morning, are there any plans to advance on this Feature?

My understanding is that this still applies: https://github.com/erlang/otp/pull/7343#issuecomment-1717596887

I am happy to answer any further questions if there are any.

Issue #7257, opened 2023-05-17
Can't use "fun FunctionName/Arity" syntax in escript


shionryuu:
**Describe the bug**

Can't use "fun FunctionName/Arity" syntax in escript

**To Reproduce**

```erl
#!/usr/bin/env escript
%% -*- erlang -*-
-module(script).

-export([main/1]).

main(_) ->
    lists:foreach(fun worker/1, lists:seq(1, 3)).
    
worker(_) ->
    ok.
```

```bash
$ ./script.erl
escript: exception error: undefined function erl_eval:worker/1
  in function  erl_eval:expr/6 (erl_eval.erl, line 338)
  in call from erl_eval:expr_list/7 (erl_eval.erl, line 961)
  in call from erl_eval:expr/6 (erl_eval.erl, line 454)
  in call from erl_eval:local_func/8 (erl_eval.erl, line 644)
  in call from escript:interpret/4 (escript.erl, line 780)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_em/1 
[1]    284351 exit 127   ./script.erl
$ sed -i '1d' script.erl
$ erlc script.erl
$ erl -pa .
Erlang/OTP 25 [erts-13.2] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [jit:ns]

Eshell V13.2  (abort with ^G)
1> script:main("").
ok
2> 

```

**Expected behavior**

use "fun FunctionName/Arity" syntax in escript

**Affected versions**

OTP 25, 26 etc.

stolen:
BTW `-mode(compile).` is a workaround.

bjorng:
It is an old limitation that not all language constructs work in escripts. We don't intend to lift that limitation. The workaround, as @stolen pointed out, is to force  the escript to be compiled.

In Erlang/OTP 27, escripts will probably be compiled by default as suggested in #7348.

PR #7348, opened 2023-06-02
Compile escripts by default

bjorng:
The current default when running an escript is to interpret it using the `erl_eval` module. In the ancient times when escript was invented that made sense. Interpretation would be slower and some language constructs would not work, but execution could start immediately without waiting for the compiler to finish.

Nowadays with faster computers it makes less sense.

Therefore, in Erlang/OTP 27, change the default to compile the escript. That will require the escript to be executed by an Erlang system that includes the Erlang compiler.

### DIFF
--- a/erts/doc/src/escript_cmd.xml
+++ b/erts/doc/src/escript_cmd.xml
@@ -185,15 +185,16 @@ halt(1).</pre>
           <c>127</c>.</p>
         <p>Both the module declaration and the export declaration of
           the <c>main/1</c> function are optional.</p>
-        <p>By default, the script will be interpreted. You can force
-          it to be compiled by including the following line somewhere
-          in the script file:</p>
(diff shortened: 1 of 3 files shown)


bjorng:
I have marked this pull request as a draft because as an potential incompatibility it will have to be approved by the OTP Technical Board.

I also propose that the interpreted mode should be removed in Erlang/OTP 28.

garazdawi:
Is it a lot of work to keep the interpreted mode as a fallback for when the compiler is not available? In small embedded systems it might still be nice to be able to run escripts without including the compiler in those systems. Maybe even we have escripts in applications that don't today need the compiler to be present?

bjorng:
@garazdawi No, there is not a lot of ongoing work to keep the interpretation working, but some language construct don't work (see #7257 for an example). Yes, the embedded systems without an compiler could be a reason that we should not remove the possibility to interpret the code.

PR #7350, opened 2023-06-02
erts: Fix 'init' crash slogan

### LINKED ISSUE #7344: Regression in error reporting with 26.0?
Hi,

While developing a test module of mine, I noticed that with version 26.0 the error reporting was at least in my case a lot less useful than in previous versions.

On the same computer (GNU/Linux, derived from Debian Bullseye) and same codebase, with 25.3 I had:

```
[...]
{"init terminating in do_boot",{{nocatch,{unsupported_random_law_spec,#Fun<random_utils_test.0.14562089>}},[{random_utils,initialise_law,1,[{file,"random_utils.erl"},{line,1187}]},{random_utils_test,run,0,[{file,"random_utils_test.erl"},{line,145}]},{erl_eval,do_apply,7,[{file,"erl_eval.erl"},{line,744}]},{init,start_it,1,[]},{init,start_em,1,[]},{init,do_boot,3,[]}]}}
init terminating in do_boot ({{nocatch,{unsupported_random_law_spec,}},[{random_utils,initialise_law,1,[{_},{_}]},{random_utils_test,run,0,[{_},{_}]},{erl_eval,do_apply,7,[{_},{_}]},{init,start_it,1,[]},{init,start_em,1,[]},{init,do_boot,3,[]}]})

[...]
```

(which is appropriate and useful), whereas now 26.0 returns for the same case:

```
[...]
Error! Failed to eval: random_utils_test:run()

Runtime terminating during boot ({{nocatch,{unsupported_random_law_spec,}},[{random_utils,canonicalise_law_spec,1,[{_},{_}]},{random_utils,initialise_law,1,[{_},{_}]},{random_utils_test,run,0,[{_},{_}]},{erl_eval,do_apply,7,[{_},{_}]},{init,start_it,1,[]},{init,start_em,1,[]},{init,do_boot,3,[]}]})

[...]
```

So the fun was not printed out at all (reproducibly), and files and line numbers are lost.

If it matters, it was run with:

```
erl -eval random_utils_test:run() +W w -pz [...] -smp +K true +A 128 +zdbbl 1048576 +P 400000 -setcookie xxx -kernel inet_dist_listen_min 50000 inet_dist_listen_max 55000 -sname random_utils_run-YYY +pc unicode -noinput -extra
```

Thanks in advance,
Best regards,

Olivier.

jhogberg:
Fixes #7344


garazdawi:
One of the features of the original algorithm was that it tried to limit terms if they were too deep, about the same as `io:format("~P",[Term, 80])` but much more naive. The new implementation only truncates on text length. The reason why trying to limit the term is nice is because if there is a callstack, we can get some very valuable information about the crash.

However, given the size of the printed text that you have chosen, maybe this is not a problem as the majority of error reasons can be printed using 32000 characters...

I do however think there should be some tests for this?

PR #7353, opened 2023-06-02
[dialyzer] Simplify ets usage

michalmuskala:
This simplifies ets:lookup + ets:delete into ets:take calls. Removing one call to ets should be beneficial in terms of performance.

### DIFF
--- a/lib/dialyzer/src/dialyzer_cplt.erl
+++ b/lib/dialyzer/src/dialyzer_cplt.erl
@@ -490,9 +490,8 @@ tab_merge('$end_of_table', T1, T2) ->
       tab_merge(Key, T1, T2)
   end;
 tab_merge(K1, T1, T2) ->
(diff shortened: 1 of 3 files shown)


michallepicki:
Is this safe? Dialyzer seems to be failing. From [`ets:next/2` docs](https://www.erlang.org/doc/man/ets.html#next-2):
> Unless a table of type set, bag, or duplicate_bag is fixated using [safe_fixtable/2](https://www.erlang.org/doc/man/ets.html#safe_fixtable-2), a call to next/2 will fail if Key1 no longer exists in the table

michalmuskala:
Switching the next & take calls around should fix the crashes (at least it seems to locally)

kikofernandez (approved):
I will run our tests tonight and merge tomorrow if everything goes well

Thanks for your contribution 🥳

PR #7362, opened 2023-06-05
dialyzer: Add dry-run mode

TD5:
For big incremental analyses, it can be useful to "dry-run" Dialyzer in order to determine how many modules Dialyzer expects to need to analyse. This allows for determining whether or not a full analysis run is needed, and what sort of resources it might require. In principle, this information could then be used for automatically running Dialyzer on an appropriately sized worker.

### DIFF
--- a/lib/dialyzer/src/dialyzer.hrl
+++ b/lib/dialyzer/src/dialyzer.hrl
@@ -158,6 +158,7 @@
                        | {'warnings', [warn_option()]}
                        | {'get_warnings', boolean()}
                        | {'use_spec', boolean()}
+                       | {'dry_run', boolean()}
                        | {'filename_opt', filename_opt()}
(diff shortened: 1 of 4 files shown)


TD5:
On second thought, we should probably not halt here either, but rather short-circuit the analysis more gracefully, so users of the Dialyzer API don't have their code halted as well.

kikofernandez:
Any updates on those tests? :)

TD5:
> Any updates on those tests? :)

It's only my radar to revisit soon. Watch this space 👍

kikofernandez:
Any news on this? 😃

TD5:
> Any news on this? 😃

Yes! Sort of. I'll replace this with a bigger and better PR with a much more efficient incremental mode.

PR #7376, opened 2023-06-07
common test: Fix typo in event handling chapter

jchristgit:


IngelaAndin:
Could you rebase your branch on maint, so that we can fix the docs for the OTP-26 track and not have to wait for OTP-27?

PR #7380, opened 2023-06-08
common test: Synchronize capture start and stop

jchristgit:
Wait on the reply by the test server group leader to ensure that by the time we continue running the test case all output is captured.

### DIFF
--- a/lib/common_test/src/test_server.erl
+++ b/lib/common_test/src/test_server.erl
@@ -1863,22 +1863,14 @@ log(Msg) ->
     ok.
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-%% capture_start() -> ok
-%% capture_stop() -> ok
-%%
-%% Starts/stops capturing all output from io:format, and similar. Capturing
-%% output doesn't stop output from happening. It just makes it possible
-%% to retrieve the output using capture_get/0.
-%% Starting and stopping capture doesn't affect already captured output.
-%% All output is stored as messages in the message queue until retrieved
+%% @see test_server_gl:capture_start/2
 
 capture_start() ->
-    group_leader() ! {capture,self()},
-    ok.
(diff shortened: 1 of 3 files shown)


lib/common_test/src/test_server.erl
```diff
@@ -1863,22 +1863,14 @@ log(Msg) ->
...
 capture_start() ->
-    group_leader() ! {capture,self()},
-    ok.
+    test_server_gl:capture_start(group_leader(), self()).
 
+%% @see test_server_gl:capture_stop/1
 capture_stop() ->
-    group_leader() ! {capture,false},
-    ok.
+    test_server_gl:capture_stop(group_leader()).
```
jchristgit, line 1873: If wanted, these functions could probably be removed altogether from here - I believe this is a private API - but it's used in a few OTP test cases:
```erl
$ rg test_server:capture_start
lib/common_test/src/ct.erl
579:    test_server:capture_start().

lib/common_test/test/test_server_SUITE_data/test_server_unicode_SUITE.erl
66:    test_server:capture_start(),
75:    test_server:capture_start(),

lib/common_test/test/test_server_SUITE_data/test_server_SUITE.erl
262:    test_server:capture_start(),

lib/observer/test/ttb_SUITE.erl
220:    test_server:capture_start(),

lib/sasl/test/rb_SUITE.erl
526:    test_server:capture_start(),

lib/tools/test/cover_SUITE.erl
835:    test_server:capture_start(),
846:    test_server:capture_start(),
854:    test_server:capture_start(),
865:    test_server:capture_start(),
880:    test_server:capture_start(),
889:    test_server:capture_start(),
901:    test_server:capture_start(),
911:    test_server:capture_start(),

lib/asn1/test/test_compile_options.erl
119:    test_server:capture_start(),
126:    test_server:capture_start(),
```
I can update these if wanted.

u3s: indeed, looks like a private API these days.
historically, this might be a public API of `test_server` behavior.

IMHO, it's not worth doing at this stage.

PR #7383, opened 2023-06-09
erl_parse: Provide friendlier "head mismatch" errors

jchristgit:
Given the following module:
```erl
    %% mismatch.erl
    -module(mismatch).

    test_it(ok) -> ok;

    verify_it(true) -> ok;
    verify_it(false) -> {error, was_false}.
    verify_it(false, Reason) -> {error, Reason}.
```
On master, running `erlc mismatch.erl`  currently reports:
```erl
    mismatch.erl:6:1: head mismatch
    %    6| verify_it(true) -> ok;
    %     | ^
```
The previous clause that the parser is comparing to may not be obvious:
The current compiler warning points to `verify_it/1`, whose two clauses
are terminated properly. When more code is inbetween these two functions
(think documentation, specs, comments) the time needed to find the
mismatched clause can be a lot higher. I also believe that this will
make the error easier to understand for people new to Erlang.

With this commit, the compiler will output the following:
```erl
    mismatch.erl:6:1: head mismatch: previous function test_it/1 is distinct from verify_it/1. Is the semicolon in test_it/1 unwanted?
    %    6| verify_it(true) -> ok;
    %     | ^
```
For functions that are defined with the same name but distinct arities,
a separate error message is used, since the possibility of missed or
superfluous arguments also arises here. The following module illustrates
it:
```erl
    %% mismatch2.erl
    -module(mismatch2).

    verify_it(false) -> {error, was_false};
    verify_it(false, Reason) -> {error, Reason}.
```
The following error message is provided in this case:
```erl
    mismatch2.erl:6:1: head mismatch: function verify_it with arities 1 and 2 is regarded as two distinct functions. Are the number of arguments incorrect or is the semicolon in verify_it/1 unwanted?
    %    6| verify_it(false, Reason) -> {error, Reason}.
    %     | ^
```
For tools matching on the output of the compiler warnings, this is a
breaking change. I have so far only found Erlang LS which matches on
this [1]. Adding support for the extended information should be trivial.

[1]: https://github.com/erlang-ls/erlang_ls/blob/e315f9801d60e1ccbbfd444f32a33d00b0f6795d/apps/els_lsp/src/els_compiler_diagnostics.erl#L614

### DIFF
--- a/lib/compiler/test/error_SUITE.erl
+++ b/lib/compiler/test/error_SUITE.erl
@@ -22,8 +22,8 @@
 
 -export([all/0, suite/0,groups/0,init_per_suite/1, end_per_suite/1, 
 	 init_per_group/2,end_per_group/2,
-	 head_mismatch_line/1,warnings_as_errors/1, bif_clashes/1,
-	 transforms/1,maps_warnings/1,bad_utf8/1,bad_decls/1]).
+	 head_mismatch_line/1, head_mismatch_same_function_name/1, warnings_as_errors/1,
+	 bif_clashes/1, transforms/1,maps_warnings/1,bad_utf8/1,bad_decls/1]).
 
 %% Used by transforms/1 test case.
 -export([parse_transform/2]).
@@ -35,7 +35,8 @@ all() ->
 
 groups() -> 
     [{p,test_lib:parallel(),
-      [head_mismatch_line,warnings_as_errors,bif_clashes,
+      [head_mismatch_line,head_mismatch_same_function_name,
+       warnings_as_errors,bif_clashes,
        transforms,maps_warnings,bad_utf8,bad_decls]}].
 
 init_per_suite(Config) ->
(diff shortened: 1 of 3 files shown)


bjorng:
Thanks for your pull request.

I like the general idea. However, I think you'll need to provide a somewhat different message when the names of the function heads are the same but they have different arities. I often run into this error when I refactor code to remove or add function parameters. For example:

```
foo({A,B}) ->
    A + B;
foo(Other, State) ->
    {Other, State}.
```

The message will be:

```
t.erl:9:1: head mismatch: previous function foo/1 is distinct from foo/2. Is there an unwanted semicolon?
%    9| foo(Other, State) ->
%     | ^
```

Here the problem is not an unwanted semicolon, but that I have forgotten to add the `State` variable to the first clause.

Regarding compatibility, we don't promise compatibility of compiler messages between major releases of OTP.

jchristgit:
Thank you for the feedback!

I have updated my commit to include special handling for functions with
the same name but different arity:

```erl
%% mismatch2.erl
-module(mismatch2).

verify_it(false) -> {error, was_false};
verify_it(false, Reason) -> {error, Reason}.
```

will result in:

```erl
mismatch2.erl:5:1: head mismatch: function verify_it with arities 1 and 2 is regarded as two distinct functions. Are arguments to verify_it/1 missing or is the semicolon unwanted?
%    5| verify_it(false, Reason) -> {error, Reason}.
%     | ^
```

Perhaps a bit verbose, but let me know what you think.

Re: compatibility, good to know. I made a draft PR to Erlang LS to
support the new message should it be merged.

bjorng:
Yes, that's better, but it is still a little bit confusing if the intention in my refactoring was to remove one or more arguments. What about:

```
mismatch2.erl:5:1: head mismatch: function verify_it with arities 1 and 2 is regarded as two distinct functions. Are the number of arguments incorrect or is there an unwanted semicolon?
```

(I found "is the semicolon unwanted?" confusing, because it not clear to which semicolon you are referring. "is the semicolon in verify_it/1 unwanted?" is longer but is clear.)

jchristgit:
I updated the commit to explicitly specify where the unwanted semicolon could be and provide a more generic message for a function with the same name but different arities. The new messages are:
```
    mismatch.erl:6:1: head mismatch: previous function test_it/1 is distinct from verify_it/1. Is the semicolon in test_it/1 unwanted?
```
and
```
    mismatch2.erl:6:1: head mismatch: function verify_it with arities 1 and 2 is regarded as two distinct functions. Are the number of arguments incorrect or is the semicolon in verify_it/1 unwanted?
```

bjorng (approved):

bjorng:
Excellent! 👍 

Added to our daily builds.

PR #7385, opened 2023-06-09
inets: Remove unused files in test suite

jchristgit:
Tests in `httpd_all.erl` and `httpd_mod.erl` have mostly been integrated into `httpd_SUITE.erl`. The deprecated apache-style configuration files that were part of the test data server root have also been removed.


IngelaAndin:
Thanks for helping us clean up the application.

PR #7388, opened 2023-06-11
ft: increase default process limit

hauleth:
Increase the default process limit from quite conservative limit that wasn't
changed since ERTS 5.10.1 to value that is more applicable in 2023. It cause
about 9 MiB more memory to be consumed at startup, but with current machines it
shouldn't be much of an issue.

### DIFF
--- a/erts/emulator/beam/erl_process.h
+++ b/erts/emulator/beam/erl_process.h
@@ -78,7 +78,7 @@ typedef struct process Process;
 #define ERTS_MAX_NO_OF_DIRTY_CPU_SCHEDULERS ERTS_MAX_NO_OF_SCHEDULERS
 #define ERTS_MAX_NO_OF_DIRTY_IO_SCHEDULERS ERTS_MAX_NO_OF_SCHEDULERS
 
-#define ERTS_DEFAULT_MAX_PROCESSES (1 << 18)
+#define ERTS_DEFAULT_MAX_PROCESSES (1 << 20)
 
 #define ERTS_HEAP_ALLOC(Type, Size)					\
      erts_alloc((Type), (Size))
--- a/erts/doc/src/erl_cmd.xml
+++ b/erts/doc/src/erl_cmd.xml
@@ -1131,7 +1131,7 @@ $ <input>erl \
 	however, be changed in the future. The actual value chosen can be
 	checked by calling
 	<seeerl marker="erlang#system_info_process_limit">erlang:system_info(process_limit)</seeerl>.</p>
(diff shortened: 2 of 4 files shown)


jhogberg:
Thanks for the PR! I just have one small nitpick, can you change the commit title to `erts: Increase default process limit` instead? `ft:` is not our convention.

hauleth:
@jhogberg ah, yes. Sorry. The conventional commit approach is so set in my brain that I forgot that OTP uses something different. I need to setup the hook to validate it.

erts/emulator/beam/erl_process.h
```diff
@@ -78,7 +78,7 @@ typedef struct process Process;
 #define ERTS_MAX_NO_OF_DIRTY_CPU_SCHEDULERS ERTS_MAX_NO_OF_SCHEDULERS
 #define ERTS_MAX_NO_OF_DIRTY_IO_SCHEDULERS ERTS_MAX_NO_OF_SCHEDULERS
 
-#define ERTS_DEFAULT_MAX_PROCESSES (1 << 18)
+#define ERTS_DEFAULT_MAX_PROCESSES (1 << 20)
```
rlipscombe, line 81: Is there a particular reason why these are expressed as bit-shifts? If there _is_, then the shift count should be in a #define, if _not_, then the value should just be written out. The compiler's going to constant-fold it anyway, but it would forestall questions (like the one I'm asking right now).

hauleth: I think the reason why it is presented as a shift right now is:

> The actual maximum chosen may be much larger than the `Number` passed. Currently the runtime system often, but not always, chooses a value that is a power of 2. This might, however, be changed in the future.

In the documentation of `+P`. So right now it is always power of 2, but it is prepared for a case when this limitation will be lifted.

rlipscombe: Makes sense. Maybe copy the comment from erl_ptab.c: `/* A size that is a power of 2 is to prefer performance wise */`...?

jhogberg: I think I'd prefer to have the minimal change merged for now, I'd be happy take a general cleanup of things like this in a separate PR. :-)

jhogberg:
The `node_container_SUITE:iter_max_procs/1` test has started failing on 32-bit machines due to running out of memory (it needs to spawn 4x as many processes before it hits the ceiling now). Can you break out that test to run on a peer node that has a lower `+P` setting?

hauleth:
@jhogberg done

jhogberg:

PR #7394, opened 2023-06-13
Fix typo for maybe expression abstract form

J3RN:
The term `Tc_k` was not introduced, and I think `Ec_k` was intended.


bjorng:
Thanks for your pull request.

PR #7398, opened 2023-06-13
syntax_tools: Fix pos during reverting

maxno-kivra:
This fixes/improves the position (really an `erl_anno` annotation) handing in `erl_syntax`.

I discovered today that reverting any `#wrapper` will throw away any modifications to the `pos`, which is fixed in two ways here. First off, make sure `set_pos` also update the inner `tree`. Secondly by always using `revert_root`.
I think either change will fix this, but I prefer having both, as it is nice to keep the `pos` in sync at all times.

While debugging this (by parsing `epp` with `epp` and comparing that with `erl_syntax:revert(erl_syntax_lib:annotate_bindings(...))`), I noticed that the list reverting was also throwing away `pos`/annotations. So I fixed that, mostly anyway. The remaining issues are hard to get around, but the changes reduces the diff.

### DIFF
--- a/lib/syntax_tools/src/erl_syntax.erl
+++ b/lib/syntax_tools/src/erl_syntax.erl
@@ -2456,17 +2456,23 @@ list(Elements, Tail) when Elements =/= [] ->
 
 revert_list(Node) ->
     Pos = get_pos(Node),
-    P = list_prefix(Node),
-    S = case list_suffix(Node) of
+    Prefix = list_prefix(Node),
+    Suffix = case list_suffix(Node) of
 	    none ->
-		revert_nil(set_pos(nil(), Pos));
-	    S1 ->
-		S1
+            LastPos = get_pos(lists:last(Prefix)),
+            LastLocation = case erl_anno:end_location(LastPos) of
+                undefined -> erl_anno:location(LastPos);
+                Location -> Location
+            end,
+            revert_nil(set_pos(nil(), erl_anno:set_location(LastLocation, Pos)));
+	    Suffix1 ->
+            Suffix1
 	end,
-    lists:foldr(fun (X, A) ->
-			{cons, Pos, X, A}
-		end,
-		S, P).
+    lists:foldr(fun (Head, Tail) ->
+        HeadPos = get_pos(Head),
+        HeadLocation = erl_anno:location(HeadPos),
+        {cons, erl_anno:set_location(HeadLocation, Pos), Head, Tail}
+    end, Suffix, Prefix).
 
 %% =====================================================================
 %% @doc Creates an abstract empty list. The result represents
--- a/lib/syntax_tools/src/erl_syntax.erl
+++ b/lib/syntax_tools/src/erl_syntax.erl
@@ -928,14 +928,18 @@ get_pos(Node) ->
 
 set_pos(Node, Pos) ->
     case Node of
-	#tree{attr = Attr} ->
-	    Node#tree{attr = Attr#attr{pos = Pos}};
-	#wrapper{attr = Attr} ->
-	    Node#wrapper{attr = Attr#attr{pos = Pos}};
-	_ ->
-	    %% We then assume we have an `erl_parse' node, and create a
-	    %% wrapper around it to make things more uniform.
-	    set_pos(wrap(Node), Pos)
+        #tree{attr = Attr} ->
+            Node#tree{attr = Attr#attr{pos = Pos}};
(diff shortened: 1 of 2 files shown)


lib/syntax_tools/src/erl_syntax.erl
```diff
@@ -930,8 +930,12 @@ set_pos(Node, Pos) ->
...
 	#tree{attr = Attr} ->
 	    Node#tree{attr = Attr#attr{pos = Pos}};
-	#wrapper{attr = Attr} ->
-	    Node#wrapper{attr = Attr#attr{pos = Pos}};
+	#wrapper{attr = Attr, tree = {error, {_, Module, Reason}}} ->
+	    Node#wrapper{attr = Attr#attr{pos = Pos}, tree = {error, {Pos, Module, Reason}}};
+    #wrapper{attr = Attr, tree = {warning, {_, Module, Reason}}} ->
+        Node#wrapper{attr = Attr#attr{pos = Pos}, tree = {warning, {Pos, Module, Reason}}};
+	#wrapper{attr = Attr, tree = Tree} ->
+	    Node#wrapper{attr = Attr#attr{pos = Pos}, tree = setelement(2, Tree, Pos)};
```
jhogberg, line 938: Nitpick: the indentation is a bit off. While you're at it, you can convert tabs to 8 spaces here.

lib/syntax_tools/src/erl_syntax.erl
```diff
@@ -2828,7 +2838,7 @@ compact_list(Node) ->
 					  copy_attrs(Node,
 						     Node1));
 			_ ->
-			    Node
```
jhogberg, line 2831: Nitpick: unrelated whitespace change. Please remove the others like it as well.

jhogberg (changes requested):
Thanks for the PR, and sorry for the delay in getting around to this. It looks good aside from a few nits, and I'd like it to be rebased to `maint` for inclusion in OTP 26.1. :)

lib/syntax_tools/src/erl_syntax.erl
maxno-kivra: Do you mean all the random trailing spaces? They were removed by my editor, and I suspect that will be the case for most editors. So in order to reduce the next diff, I would like to keep the commit.

lib/syntax_tools/src/erl_syntax.erl
maxno-kivra: I think I got rid of all the tabs now.

maxno-kivra:
Now it's my turn to apologise for the delay 😉. I'm back from vacation and have rebased on latest master and replied to your comments.

lib/syntax_tools/src/erl_syntax.erl
frej: Changes which are not related to what the patch claims to do, mask the real reason for a code line looking like it does. If I'm trying to find a bug, I want git blame to point to the last functional change, not a completely unrelated patch doing a cosmetic change. I guess that's why the [contributor's guide](https://github.com/erlang/otp/blob/master/CONTRIBUTING.md) asks you to not fix preexisting whitespace errors in otherwise untouched source lines.

maxno-kivra: I removed the commit that removes the trailing whitespace, but for posterity I'd like to mention the `--ignore-rev`/`--ignore-revs-file` flags to `git blame` which allows you to ignore/automatically re-blame.

maxno-kivra:
@jhogberg I removed the white space commit a while ago, but I guess it got lost in the ether. 

Anyway, could you please take another look at this? 🙏

maxno-kivra:
Ping @jhogberg please take another look

maxno-kivra:
@jhogberg @frej either close this or re-review it now that I have removed the white space commit. It fixes a bug and adds a test to prove it.

I understand you have limited bandwidth, but please at least communicate that you don't have time. Silence for weeks just makes one wonder if you only have one shot to get you PR approved/merged. If you fail the first review you'll just get ignored. I really hope I'm wrong about that, so please prove me wrong here. Thanks 🙏

jhogberg:
> Silence for weeks just makes one wonder if you only have one shot to get you PR approved/merged.

Sorry, it was my turn to go on vacation just prior to you getting back. I returned to work a bit after you pinged me and missed the notification. Oops. :smile:

I've added it to our nightly runs, and will merge it tomorrow if nothing explodes.

jhogberg:

maxno-kivra:
❤️

PR #7403, opened 2023-06-14
Fix broken windows installer exe link

jchristgit:


garazdawi:
Thanks! I rebased to maint and changed the link to point to https://www.erlang.org/downloads instead.

PR #7405, opened 2023-06-15
ssl: Doc enhancement

IngelaAndin:
So that you find keylog doc easier

### DIFF
--- a/lib/ssl/doc/src/using_ssl.xml
+++ b/lib/ssl/doc/src/using_ssl.xml
@@ -342,6 +342,33 @@ ssl:connect("localhost", 9999,
     
   </section>
 
+ <section>
+   <title>NSS keylog </title>
+   <p>The NSS keylog debug feature can be used by authorized users
+   to for instance enable wireshark to decrypt TLS packets.</p>
+
+   <p><em>Server (with NSS key logging)</em></p>
+    <code type="none">
+    server() ->
+        application:load(ssl),
+        {ok, _} = application:ensure_all_started(ssl),
+        Port = 11029,
+        LOpts = [{certs_keys, [#{certfile => "cert.pem", keyfile => "key.pem"}]},
+        {reuseaddr, true},
+        {versions, ['tlsv1.2','tlsv1.3']},
+        {keep_secrets, true} %% Enable NSS key log (debug option)
+        ],
+        {ok, LSock} = ssl:listen(Port, LOpts),
+        {ok, ASock} = ssl:transport_accept(LSock),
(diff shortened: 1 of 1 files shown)


dgud (approved):

lib/ssl/doc/src/using_ssl.xml
```diff
@@ -342,6 +342,38 @@ ssl:connect("localhost", 9999,
...
+        {versions, ['tlsv1.2','tlsv1.3']},
+        {keep_secrets, true} %% Enable NSS key log (debug option)
+        ],
+        {ok, LSock} = ssl:listen(Port, LOpts),
+        %% Accept first connection
+        {ok, CSock0} = ssl:transport_accept(LSock),
+        {ok, _} = ssl:handshake(CSock0),
+        %% Accept second connection
+        {ok, CSock1} = ssl:transport_accept(LSock),
+        {ok, Sock} = ssl:handshake(CSock1),
```
u3s, line 367: explain somehow why 2 connections are needed?
keylog does not work for 1st connection?

maybe 2 connections are not needed and example can be simplified?

lib/ssl/doc/src/using_ssl.xml
```diff
@@ -689,10 +721,6 @@ ssl:connect("localhost", 9999, [{verify, verify_peer},
     GET, can usually be regarded as safe but even they can be exploited by a large number of
     replays causing resource limit exhaustion and other similar problems.</p>
     <p>An example of sending early data with automatic and manual session ticket handling:</p>
-    <warning>
-    <p>The Early Data feature is experimental in this version of OTP.
-    </p>
-    </warning>
 
     <p><em>Server (with NSS key logging)</em></p>
```
u3s, line 725: Section below no longer contains NSS key logging option, is the title still valid?

u3s (changes requested):

lib/ssl/doc/src/using_ssl.xml
IngelaAndin: Will look into if it can be simplified, I just split a old example into two. Thanks for the comment.

lib/ssl/doc/src/using_ssl.xml
IngelaAndin: Probably forgot to remove it, will fix!

lib/ssl/doc/src/using_ssl.xml
IngelaAndin: Fixed

PR #7418, opened 2023-06-19
logger_olp:check_load/1: only retrieve memory metric if absolutely ne…

### LINKED ISSUE #7417: logger's overload protection gets slower as its message queue length grows
**Describe the bug**
The `logger_olp` overload protection module used by the standard OTP logger handlers has a performance bug where its per-message overheads gets higher as its message queue gets longer. If the logger handler is configured to not drop messages, a spike of messages can put it in a state where it fails to drain its queue and eventually crashes the node with out-of-memory.

The bug is that `logger_olp:check_load/1` unconditionally calls `process_info(self(), memory)`. That call takes time proportional to the length of the message queue, and the result is not used unless specific runtime or compile-time configuration is enabled. In our case the result is never used.

We run our logger in a non-dropping async mode, i.e. with burst and sync modes disabled and the drop mode limit increased to avoid reaching it. Our previous logging framework (a wrapper around `disk_log`) has no problems keeping up with our logging load, including heavy bursts. When we migrated to `logger` we found that no amount of configuration tweaking would prevent it from dropping messages. Worse, as we increased the dropping limit the handler would never leave the long message queue state and eventually crash the node with a memory allocation failure.

**To Reproduce**
I've attached a `logger_olp_test.erl` module (zipped since GH won't let me attach an Erlang file) which illustrates the problem. First plot the time to drain message queues of lengths up to `1 bsl 17` while retrieving `memory` for each message:
```
1>  logger_olp_test:test(17, []).
[{0,0},
 {1,0},
 {2,0},
 {3,0},
 {4,0},
 {5,0},
 {6,0},
 {7,0},
 {8,0},
 {9,0},
 {10,1},
 {11,6},
 {12,28},
 {13,115},
 {14,467},
 {15,1891},
 {16,7501},
 {17,33842}]
```
Next repeat that without retrieving `memory`:
```
2> logger_olp_test:test(17, [{memory, false}]).
[{0,0},
 {1,0},
 {2,0},
 {3,0},
 {4,0},
 {5,0},
 {6,0},
 {7,0},
 {8,0},
 {9,0},
 {10,0},
 {11,0},
 {12,0},
 {13,0},
 {14,1},
 {15,2},
 {16,5},
 {17,10}]
```
You can increase the first parameter to check even longer message queue lengths, but the runtime quickly increases to impractical levels. You can also add `{off_heap, false}` to verify that placing messages on the heap only has a marginal effect (the `memory` metric is still linear).

**Expected behavior**
I expect the per-message processing costs in `logger_olp` to remain constant regardless of message queue length.

**Affected versions**
OTP-24, 25, and 26. (Probably older ones too, but I haven't checked and they're EOL.)

**Additional context**
[logger_olp_test.zip](https://github.com/erlang/otp/files/11786878/logger_olp_test.zip)

mikpe:
…cessary

Fixes #7417

### DIFF
--- a/lib/kernel/src/logger_olp.erl
+++ b/lib/kernel/src/logger_olp.erl
@@ -473,7 +473,7 @@ check_load(State = #{id:=_Name, mode_ref := ModeRef, mode := Mode,
                      sync_mode_qlen := SyncModeQLen,
                      drop_mode_qlen := DropModeQLen,
                      flush_qlen := FlushQLen}) ->
-    {_,Mem} = process_info(self(), memory),
+    Mem = maybe_self_memory(State),
     ?observe(_Name,{max_mem,Mem}),
     {_,QLen} = process_info(self(), message_queue_len),
     ?observe(_Name,{max_qlen,QLen}),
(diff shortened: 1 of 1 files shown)


lib/kernel/src/logger_olp.erl
```diff
@@ -507,6 +507,24 @@ check_load(State = #{id:=_Name, mode_ref := ModeRef, mode := Mode,
      ?update_other(flushes,FLUSHES,_NewFlushes,
                    State4#{last_qlen => QLen})}.
 
+%% Calling process_info(self(), memory) is linear on the size of the message queue,
+%% which is an extremely bad thing to do in high load situations, so only do that
+%% if the value actually is required.
+-ifdef(OBSERVER_MOD). % does ?observe/2 use Mem
+maybe_self_memory(_State) -> do_self_memory().
+-else.
+-ifdef(SAVE_STATS). % does ?update_max_mem/2 use Mem
```
mikpe, line 516: These `ifdef`s are needed because of pre-existing preprocessor usage in `logger_olp`.

garazdawi:
thanks!

PR #7422, opened 2023-06-20
`code`: reorder functions in docs

Maria-12648430:
In the documentation of the `code` module, the functions are apparently ordered randomly. This PR reorders them alphabetically, as in most other modules.

### DIFF
--- a/lib/kernel/doc/src/code.xml
+++ b/lib/kernel/doc/src/code.xml
@@ -327,10 +327,10 @@ zip:create("mnesia-4.4.7.ez",
       <name name="cache"/>
     </datatype>
     <datatype>
-      <name name="load_ret"/>
+      <name name="load_error_rsn"/>
     </datatype>
     <datatype>
-      <name name="load_error_rsn"/>
+      <name name="load_ret"/>
     </datatype>
     <datatype>
       <name name="module_status"/>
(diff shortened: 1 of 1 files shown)


garazdawi:
> the functions are apparently ordered randomly

I think the functions are ordered in functional areas, i.e. code path, loading code, unloading code etc etc.

IMO it helps that `set_path` is next to `get_path` as it allows the discovery of related functions when you browse the documentation. However, if you are looking for a specific function then alphabetic makes more sense. I'll leave it up to @bjorng to decide how we want it :)

Maria-12648430:
@garazdawi @bjorng I won't feel hurt if this PR gets rejected, but let me point out two things ;)

This module would be pretty alone in following this group-by-functionality approach. The only other module that comes to mind is `queue`, and there the functions are grouped by the API they implement (Original, Extended, Okasaki).

That said, it is obvious that grouping is often arbitrary to some degree. There are several ways to group things. Like, by what they do (like `code:set_path` and `code:get_path`) or what they are targeted at (like the `lists:key*` family vs the rest of the `lists` functions).

bjorng:
I will think some more about this pull request, but right now I think that I will approve it. I don't think that the implied groups of the functions for `code` are sufficiently clearly defined to justify breaking the rule about alphabetical ordering.

bjorng (approved):

bjorng:
Thanks for your pull request.

PR #7425, opened 2023-06-20
Handle seconds in `zip:dos_date_time_from_datetime/1`

### LINKED ISSUE #7412: zip:create/2,3 does not accept a valid file:file_info record
**Describe the bug**
`zip:create/2,3` does not accept a valid `file:file_info` record.

**To Reproduce**
In erl:
```
Eshell V13.2.2  (abort with ^G)
1> rr(code:lib_dir(kernel, include) ++ "/file.hrl").
[file_descriptor,file_info]
2> zip:create("archive.zip", [{"file.txt", <<>>, #file_info{type=regular, mtime=0}}], [memory]).
{error,{'EXIT',{function_clause,[{zip,dos_date_time_from_datetime,
                                      [0],
                                      []},
                                 {zip,local_file_header_from_info_method_name,5,
                                      [{file,"zip.erl"},{line,1131}]},
                                 {zip,put_z_files,6,[{file,"zip.erl"},{line,874}]},
                                 {zip,do_zip,3,[{file,"zip.erl"},{line,466}]},
                                 {zip,zip,3,[{file,"zip.erl"},{line,455}]},
                                 {erl_eval,do_apply,7,[{file,"erl_eval.erl"},{line,748}]},
                                 {shell,exprs,7,[{file,"shell.erl"},{line,691}]},
                                 {shell,eval_exprs,7,[{file,"shell.erl"},{line,647}]}]}}}
```

**Expected behavior**
The file modification time in the archive is set to the Unix epoch. (This won't actually happen because I've omitted the rest of the file_info for brevity.)

**Affected versions**
At least 25.3.2

**Additional context**
I want to create a reproducible archive and while it's doable with the current API it's not that concise. It might be worth adding an option for this.

sirikid:
Fixes #7412

### DIFF
--- a/lib/stdlib/src/zip.erl
+++ b/lib/stdlib/src/zip.erl
@@ -1620,6 +1620,9 @@ dos_date_time_to_datetime(DosDate, DosTime) ->
     {{YearFrom1980+1980, Month, Day},
      {Hour, Min, Sec}}.
 
+dos_date_time_from_datetime(Seconds) when is_integer(Seconds) ->
+    DateTime = calendar:now_to_datetime({0, Seconds, 0}),
+    dos_date_time_from_datetime(DateTime);
 dos_date_time_from_datetime({{Year, Month, Day}, {Hour, Min, Sec}}) ->
     YearFrom1980 = Year-1980,
     <<DosTime:16>> = <<Hour:5, Min:6, Sec:5>>,
--- a/lib/stdlib/test/zip_SUITE.erl
+++ b/lib/stdlib/test/zip_SUITE.erl
@@ -27,7 +27,8 @@
          openzip_api/1, zip_api/1, open_leak/1, unzip_jar/1,
 	 unzip_traversal_exploit/1,
          compress_control/1,
-	 foldl/1,fd_leak/1,unicode/1,test_zip_dir/1]).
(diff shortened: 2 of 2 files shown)


lib/stdlib/src/zip.erl
```diff
@@ -1510,7 +1510,7 @@ get_z_file(In0, Z, Input, Output, OpO, FB,
 					   CompSize, Input, Output, OpO, Z),
 			    In5 = skip_z_data_descriptor(GPFlag, Input, In4),
 			    %% TODO This should be fixed some day:
-			    %% In5 = Input({set_file_info, FileName, 
+			    %% In5 = Input({set_file_info, FileName,
```
jhogberg, line 1513: Nitpick: please skip this unrelated whitespace change.

jhogberg (changes requested):
Thanks for the PR! All that's lacking is a test case, can you add one to `zip_SUITE.erl`? :-)

lib/stdlib/src/zip.erl
sirikid: Sure

jhogberg (approved):

jhogberg:

PR #7428, opened 2023-06-21
ct: Add basic dark color scheme support to HTML

jchristgit:
This commits adds basic support for dark theme in Common Test's HTML output by inverting and rotating colors. The separators between individual rows are also changed to grey to make it easier to read.

## Screenshots

image

image

image

image


u3s:
1. I think that in general we would like reports look to be customizable and I speculate this was the intention behind "-stylesheet" option of ct_run command
2. unfortunately this option seems broken ... (maybe I'm missing something) when I try to use it by providing custom CSS, I see it being applied only to test log HTML file and not report files leading to testcase or configuration function log
3. maybe we should have a look on it and fix it if issue is confirmed ? then any CSS could be used by users concerned with default look
4. regarding the PR my thoughts are:
   - in theory "dark mode" seems to be personal preference anyone can turn on or off,
   - but in practice it might be set to default in some browsers (Chrome?) inheriting the settings from OS
   - OS might be managed administrators not caring or willing to make it adjustable
   - dark mode setting might be experimental feature in some browsers - might make user not happy when they are forced to adjust it
   - as a result I can imagine, some users locked with new coloring and not necessarily happy
   - complaining about no option to change display mode from the HTML page directly and missing privileges to change it elsewhere ...
   - or complaining about _colors they don't like_ - and this is a discussion I definitely don't want to participate in :-)

Taking all above into consideration, I think I'm leaning towards checking "stylesheet" option. Fixing it or maybe even improving to make customizations easier for interested users.

What do you think?

jchristgit:
Good evening, hmm, I think I disagree with some of your points:

>    - OS might be managed administrators not caring or willing to make it adjustable

This would be the first time I hear that dark theme is something managed
via group policy or other admin settings, even if it's managed on OS
level, at least in Firefox you can configure it independently. But I
haven't really worked in environments that manage your computer this far
so I can't say for sure...

>    - dark mode setting might be experimental feature in some browsers - might make user not happy when they are forced to adjust it

https://caniuse.com/?search=prefers-color-scheme suggests that 94.93% of
browsers support this - I assume for the remaining 5% the @media query
will just not work and they will stay on light theme (if their settings
are switched on dark theme).


These points both make perfect sense to me:

>    - complaining about no option to change display mode from the HTML page directly and missing privileges to change it elsewhere ...
>    - or complaining about _colors they don't like_ - and this is a discussion I definitely don't want to participate in :-)

We could perhaps do something like here:
https://stackoverflow.com/questions/56300132/how-to-override-css-prefers-color-scheme-setting
to change it with JS but it seems hacky to be honest.


`ct_run` has a `-basic_html` switch which, if I understood it properly,
reverts back to the old output that CT used to produce. Maybe we could
take a similar approach as to my PR calming down the ct_run output:
introducing this and if people do not like it, offer flags to change it?
What do you think?

Regardless of the outcome here I agree fixing stylesheet support may be
a nice idea, but given this wasn't known to be broken before you tried
using it I wonder if this feature is in use at all?

u3s:
> @u3s I looked into the stylesheet option and it seems to work, but you must give it an absolute path. It also does not link to the file but copies its content into the HTML, is that maybe why you saw it not working? If I read the code right (`print_style_error/4` in `ct_logs`) then it will only emit a HTML comment to document failure.

yes I've used absolute path. what I see is that:
- if I open test or config function log (for example init_per_suite), then custom stylesheet is applied
- but if I go up in hierarchy when links to those test logs are displayed in table, then stylesheet is not applied.

u3s:
After discussing, we lean towards accepting your original proposal.

PR #7441, opened 2023-06-27
List optional applications for wx apps

josevalim:
Erlang/OTP 24 added support for optional applications,
this pull request lists them to some apps in order to
exercise the tooling.

If no bugs are found, we could list the optional applications
for more apps in the future.

### DIFF
--- a/lib/debugger/src/debugger.app.src
+++ b/lib/debugger/src/debugger.app.src
@@ -1,8 +1,8 @@
 %%
 %% %CopyrightBegin%
-%% 
+%%
 %% Copyright Ericsson AB 1997-2021. All Rights Reserved.
-%% 
+%%
 %% Licensed under the Apache License, Version 2.0 (the "License");
 %% you may not use this file except in compliance with the License.
 %% You may obtain a copy of the License at
@@ -14,7 +14,7 @@
 %% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 %% See the License for the specific language governing permissions and
 %% limitations under the License.
-%% 
+%%
 %% %CopyrightEnd%
 %%
 {application, debugger,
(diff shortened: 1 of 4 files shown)


josevalim:
There are two separate threads from this issue:

* `ttb` is part of `observer`, which causes `observer` to depend on `et`. I think we should move `ttb` to `runtime_tools` and `ttb_et` to `et`, in order to remove this additional dependency.

* `dialyzer` depends on `ex` because it has a GUI but I am not sure how much this GUI is used and it does not support the new incremental PLT. I will send a PR to remove the Dialyzer GUI and see where that leads.

IngelaAndin:
@dgud  will have a look when he is back from vacation

lib/debugger/src/debugger.app.src
```diff
@@ -47,6 +47,7 @@
 	     int
 	    ]},
   {registered, [dbg_iserver, dbg_wx_mon, dbg_wx_winman]},
-  {applications, [kernel, stdlib]},
+  {applications, [kernel, stdlib, compiler, wx]},
```
IngelaAndin, line 50: wx should only be only in optional applications

lib/et/src/et.app.src
```diff
@@ -31,7 +31,8 @@
     et_wx_viewer
    ]},
   {registered, [et_collector]},
-  {applications, [stdlib, kernel]},
+  {applications, [stdlib, kernel, wx, runtime_tools]},
+  {optional_applications, [wx, runtime_tools]},
```
IngelaAndin, line 35: wx should not  be in applications.

josevalim: @IngelaAndin `optional_applications` should be a subset of `applications`. This was a discussion back to when the feature was implemented, because we need `applications` to still provide a consistent ordering (as in #2675).

IngelaAndin: @josevalim Humm ... I guess I misread the documentation.  @dgud promised he would have a look at this as he has worked more with the applications at hand.

dgud:
I guess @IngelaAndin read this:

> optional_applications
> A list of applications that are optional. Note if you want an optional dependency to be automatically started before the current application whenever it is available, it must be listed on both applications and optional_applications.

And wx doesn't need to be started (it's started automagically by wx:new() ), and in the case of wx I believe we should NOT
try to start it.  

My thinking is, someone starts etop on a a target system then you don't want to load 200 wx modules that is not gonna be used.

josevalim:
Thank you @IngelaAndin and @dgud, I understand it now and I agree. I will have to change some assumptions in Elixir but that's on my end. :)

I have updated the PR so `wx` is only listed under optional applications.

PR #7443, opened 2023-06-27
Remove dialyzer --gui

josevalim:
The GUI does not support the new incremental PLT
and, while support could be added, the --gui provides
only a subset of the CLI functionality, so I thought it
may be simpler to remove the functionality.


bjorng:
The build fails:

https://github.com/erlang/otp/actions/runs/5388259519/jobs/9780664747?pr=7443

There seems to be some remaining references to the removed `gui/1` function.

(Personally, I am all for removing the `--gui` option, but we will have to discuss it in the OTP team.)

KennethL:
I agree that it would be good to remove the GUI both because it is not needed and that we remove a dependency to Wx.

bjorng:
Thanks for your pull request.

PR #7445, opened 2023-06-27
Fixed wrong output for orddict:store code example

tobiazpMA:
Code example `orddict:store` showed wrong output for c value (99 instead of 100).


bjorng:
Thanks for your pull request.

PR #7447, opened 2023-06-27
ssl: Remove compression handling

dgud:
Compression is deprecated and deemed unsecure so we will never implement it, 
remove all handling of it to save memory and performance.


lib/ssl/src/ssl_internal.hrl
```diff
@@ -74,6 +74,10 @@
 -define(TRUE, 0).
 -define(FALSE, 1).
 
+
+-define(NO_COMPRESSION, ?NULL).
+-define(COMPRESSIONS, [?NO_COMPRESSION]).
```
u3s, line 79: I don't know whether COMPRESSIONS macro is worth having. it is used in only 3 places.
IMHO, when reading code [?NO_COMPRESSION] is more informative, albeit a bit harder to write.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -850,28 +847,28 @@ decode_handshake(_, ?NEXT_PROTOCOL, <<?BYTE(SelectedProtocolLength),
     #next_protocol{selected_protocol = SelectedProtocol};
 
 decode_handshake(Version, ?SERVER_HELLO, <<?BYTE(Major), ?BYTE(Minor), Random:32/binary,
-					    ?BYTE(SID_length), Session_ID:SID_length/binary,
-					    Cipher_suite:2/binary, ?BYTE(Comp_method)>>) ->
+                                           ?BYTE(SID_length), Session_ID:SID_length/binary,
+                                           Cipher_suite:2/binary, ?BYTE(CompMethod)>>) ->
+    ?NO_COMPRESSION = CompMethod,
```
u3s, line 852: can't you pattern match in function clause? you want to avoid that?
here and below.

lib/ssl/src/tls_dtls_connection.erl
```diff
@@ -832,21 +829,18 @@ override_server_random(Random, _, _) ->
     Random.
 
 new_server_hello(#server_hello{cipher_suite = CipherSuite,
-			      compression_method = Compression,
-			      session_id = SessionId},
-                 #state{session = Session0,
-                        static_env = #static_env{protocol_cb = Connection}} = State0, Connection) ->
+                               session_id = SessionId},
+                 #state{session = Session0} = State0, Connection) ->
```
u3s, line 833: do you really want to remove pattern matching on protocol_cb. when it's removed different function clause might be selected ... just speculating. I don't know if it was done on purpose ...

here and below ...

u3s (changes requested):
1. ssl.xml - remove Compression argument from server_reuse_session option?
2. ssl_cipher.erl - rename variables so they don't include "Compressed" in names?
3. ssl_alert.hrl - remove DECOMPRESSION_FAILURE macro? also from code comments referring to this macro?

lib/ssl/src/dtls_record.hrl
```diff
@@ -29,7 +29,7 @@
 
 -include("ssl_record.hrl"). %% Common TLS and DTLS records and Constantes
 
-%% Used to handle dtls_plain_text, dtls_compressed and dtls_cipher_text
+%% Used to handle dtls_plain_text and dtls_cipher_text
```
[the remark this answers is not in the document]
dgud: I brought back the comment here

lib/ssl/src/ssl_handshake.erl
dgud: Fixed

u3s (approved):

IngelaAndin (approved):

Issue #7342, opened 2023-06-01
GH-6599 continued


bjorng:
**To Reproduce**
Compile one the following examples:

```
%% Never finishes.
f(X) when true; X ->
    - maybe
        <<>> ?= X,
        f(maybe
            0 ?= ([] =< ok)
        end),
        <<>> ?= ok
    else
        false  ->
            X;
        0  ->
            X
    end.
```
```
%% Never finishes.
f(_, []) ->
    0;
f(X, _) ->
    ?MODULE:foo(
        <<>> =
            try f(ok, X) + 0 of
                X ->
                    X
            catch
                _ ->
                    X
            end
    );
f(_, _) ->
    ok.
```

```
f(_V0) ->
    case
        trunc(
            try
                trunc(<<>> = _V0)
            catch
                _ ->
                    0.0
            end
        )
    of
        _V0 -> ok;
        _ -> 42
    end.
```
Crashes like this:

```
  Internal consistency check failed - please report this bug.
  Instruction: {test,is_eq_exact,
                     {f,4},
                     [{tr,{y,0},{t_integer,any}},{integer,0}]}
  Error:       {bad_typed_register,{t_bitstring,256,true},{t_integer,any}}:
```

bjorng:
Additional examples from #7360 with some optimizations disabled.

Compiling with `erlc +no_ssa_opt`:
```
f(_V0) ->
    [0 || not (_V0 > (0 bsl erlang:adler32(true = (_V7 = (ok < _V0))))), _V7].
```

results in the following message:

```
test21380:1: function f/1+20:
  Internal consistency check failed - please report this bug.
  Instruction: {test,is_eq_exact,{f,10},[{y,0},{atom,true}]}
  Error:       {type_conflict,{t_atom,[ok]},none}:
```

Compile with `erlc +no_copt +no_ssa_opt`:
```
f() ->
    <<_V0/utf32>> = foo,
    [
        0
     || _ <- bar,
        _V4 = (_V2 = _V0) > baz,
        _V4 =
            case _V2 bxor _V2 of
                _ -> ok
            end
    ].
```

Compile with `erlc +no_ssa_opt`:
```
f(_V3) ->
    case foo =< _V3 of
        _V4 when _V4 andalso {trunc(_V3), _V4 orelse bar} ->
            ok
    end.
```

bjorng:
Additional example from #8202:

```erlang
-export([f/1]).
-record(r0, {rf0}).

f(_V1) ->
[...]

PR #7448, opened 2023-06-27
Fix longstanding errors in the compiler's alias analysis pass

frej:
This is a set of patches which fixes a number of errors in the compiler's alias analysis pass which have been there since the code was first merged. [The final commit](https://github.com/erlang/otp/commit/1ed37c62c6ffa6d28dac6b4a471218b18124b359) is unfortunately large, but I have been unable to create a smaller patch, which doesn't disable previously working and correct test cases.

### DIFF
--- a/lib/compiler/src/beam_ssa_alias.erl
+++ b/lib/compiler/src/beam_ssa_alias.erl
@@ -740,14 +740,10 @@ aa_bif(Dst, element, [#b_literal{},Tuple], SS, _AAS) ->
     aa_set_aliased([Dst,Tuple], SS);
 aa_bif(Dst, element, [#b_var{},Tuple], SS, _AAS) ->
     aa_set_aliased([Dst,Tuple], SS);
-aa_bif(Dst, hd, Args, SS, AAS) ->
-    %% If we extract a value and the aggregate dies and wasn't aliased,
-    %% we should not consider this an aliasing operation.
-    aa_alias_if_args_dont_die(Args, Dst, SS, AAS);
-aa_bif(Dst, tl, Args, SS, AAS) ->
-    %% If we extract a value and the aggregate dies and wasn't aliased,
-    %% we should not consider this an aliasing operation.
-    aa_alias_if_args_dont_die(Args, Dst, SS, AAS);
+aa_bif(Dst, hd, [Pair], SS, _AAS) ->
+    aa_pair_extraction(Dst, Pair, hd, SS);
+aa_bif(Dst, tl, [Pair], SS, _AAS) ->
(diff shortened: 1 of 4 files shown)


frej:
Tweak commit subject for 5175353e7c99568fbf21fc79daaf5d74e94e63af.

frej:
The bug found by @RobinMorisset should now be fixed.

frej:
The build failures are due to `write /dev/stdout: no space left on device`, so I'd  say that this PR still qualifies for label:testing.

RobinMorisset:
Thanks for the quick fix!
Erlfuzz found another failing test case overnight:
```erlang
f() ->
    [ 0 || _V1 <- erlang:memory(),
        { maybe
            error ?= _V1,
            ok
        end,
        maybe 
            {<<_>>} ?= _V1,
            ok
        end } ].
```
Without this PR it compiles fine, but with it erlc fails with:
```
Sub pass ssa_opt_alias
/home/rmorisset/minimized/test282060.erl: internal error in pass beam_ssa_opt:
exception error: no match of right hand side value true
  in function  beam_ssa_alias:aa_tuple_extraction/4 (beam_ssa_alias.erl, line 1176)
  in call from beam_ssa_alias:aa_is/4 (beam_ssa_alias.erl, line 456)
  in call from beam_ssa_alias:aa_blocks/3 (beam_ssa_alias.erl, line 392)
  in call from beam_ssa_alias:aa_fun/3 (beam_ssa_alias.erl, line 372)
  in call from beam_ssa_alias:aa_fixpoint/6 (beam_ssa_alias.erl, line 340)
  in call from beam_ssa_opt:run_phases/3 (beam_ssa_opt.erl, line 75)
  in call from beam_ssa_opt:module/2 (beam_ssa_opt.erl, line 71)
```

frej:
Another day, another fix :) 

@RobinMorisset, how much CPU-time was required to find the latest crash? I'm starting to think I should set up my own fuzzing instance. Is it worthwhile to run on single PC overnight, or is a cluster/cloud required to get results?

RobinMorisset:
It took a few hours on a moderately beefy server (on par with a high-end gaming PC I'd say), so it would probably have taken a day or two of running on the average laptop.
I think that the main challenge in running erlfuzz is filtering out the duplicates of #7342 (although it got way better since some recent fixes were merged).

frej:
> I think that the main challenge in running erlfuzz is filtering out the duplicates of https://github.com/erlang/otp/issues/7342

Then I think I'll leave the fuzzing to you :)

RobinMorisset:
The fuzzer did not find any new bug on this PR for a couple of days, so it looks good to me.

PR #7451, opened 2023-06-28
Implementation of EEP 64: Triple-Quoted Strings

RaimoNiskanen:
Triggered by [EEP 64](https://github.com/erlang/eep/blob/master/eeps/eep-0064.md) and @williamthome's PR #7313 I also wrote an implementation of triple-quoted strings, only in the scanner.


RaimoNiskanen:
I write updates of EEP 64 in erlang/eep#49

RaimoNiskanen:
I took a stab at the Erlang mode for Emacs, but have a lingering feeling that I only stabbed myself in the foot...

lib/stdlib/src/erl_scan.erl
```diff
@@ -828,6 +857,218 @@ scan_char([], St, Line, Col, Toks) ->
...
+              Line, Col, Cs);
+        {Tag=white_space, N} ->
+            scan_error(
+              Tag, Line0, incr_column(Col0, Tqs#tqs.qs+N),
+              Line, Col, Cs)
+    end.
+
+%% Reconstruct the scanned triple-quoted string from content lines
+%%
+tqstring_chars_r(ContentR) ->
```
kikofernandez, line 991: Is `tqstring_chars_s` the same as:
```erl
lists:foldl(fun (X, Acc) -> [lists:reverse(X)|Acc] end, [], ContentR).
```
Just checking whether I understood it and wondering if it makes more sense the shorter version or this recursive version

kikofernandez (approved):
Great PR!
I have only one minor comment but it is not mandatory to address, I simply thought some small piece of code could be simplified (but we all have different coding styles)

lib/stdlib/src/erl_scan.erl
RaimoNiskanen: Almost.  Changing to `lists:reverse(X, Acc)` makes it equivalent.  Which leads to this beauty, that I will change to, thanks for the suggestion:
``` erlang
tqstring_chars_r(ContentR, String) ->
    lists:foldl(fun lists:reverse/2, String, ContentR).
```

PR #7457, opened 2023-06-30
Clean up `base64` property test suite

juhlig:
This PR removes the old manual inclusion of the different libraries and replaces it by including `ct_property_test.hrl`.


jhogberg:
Merged, thanks for the PR! :)

PR #7470, opened 2023-07-04
init: Introduce -S flag

jchristgit:
Currently, when wanting to run programs via `-s` or `-run` that take arguments, one has to use the form `-s M F -extra A1 A2 ...`. This is cumbersome to remember and must be documented appropriately. Furthermore, the documentation for `-s` recommends to use `-run` instead, as `-s` passes arguments as atoms.

This commit introduces a new flag, `-S`, which combines `-s`, `-run`, and `-extra`: arguments will be passed as strings, can be passed without the `-extra` flag. Two pieces in Erlang that would benefit from this as is:

- The C sources for dialyzer, escript and typer all use `-extra` to pass additional arguments.
- https://github.com/erlang/otp/pull/7299, if merged, would allow using `erl -S httpd serve --port 1234` instead of `erl -s httpd serve -extra --port 1234`.

**For now, this is just a suggestion.** The `-S` flag was chosen because I find it easier to remember compared to another form of `-run`, and because Elixir uses the same flag.

If there is interest in this, I will refine it with documentation and docs as needed.

### DIFF
--- a/erts/doc/src/init.xml
+++ b/erts/doc/src/init.xml
@@ -310,6 +310,28 @@ BF</pre>
           <seemfa marker="#get_plain_arguments/0">
           <c>get_plain_arguments/0</c></seemfa>.</p>
       </item>
+      <tag><c>-S Mod [Func [Arg1, Arg2, ...]]</c></tag>
+      <item>
+        <p>Evaluates the specified function call during system
+          initialization. <c>Func</c> defaults to <c>start</c>. If no
+          arguments are provided, the function is assumed to be of arity
+          0. Otherwise it is assumed to be of arity 1, taking the list
+          <c>[Arg1,Arg2,...]</c> as argument. All arguments are passed
+          as strings. If an exception is raised, Erlang stops with an
+          error message.</p>
+        <p>Example:</p>
+        <pre>
+          % <input>erl -S httpd serve --port 8080 /var/www/html</input></pre>
+        <p>This starts the Erlang runtime system and evaluates
+          the function <c>httpd:serve(["--port", "8080", "/var/www/html"])</c>.
+          All arguments up to the end of the command line will be passed
+          to the called function.</p>
+        <p>The function is executed sequentially in an initialization
+          process, which then terminates normally and passes control to
+          the user. This means that a <c>-S</c> call that does not
+          return blocks further processing; to avoid this, use
+          some variant of <c>spawn</c> in such cases.</p>
+      </item>
       <tag><c>-run Mod [Func [Arg1, Arg2, ...]]</c></tag>
       <item>
         <p>Evaluates the specified function call during system
@@ -333,6 +355,10 @@ foo:bar(["baz", "1", "2"]).</code>
           the user. This means that a <c>-run</c> call that does not
           return blocks further processing; to avoid this, use
           some variant of <c>spawn</c> in such cases.</p>
+        <note><p>This flag will not forward arguments beginning with
+          a hyphen (-) to the specified function, as these will be
+          interpreted as flags to the runtime. If the function uses
+          flags in this form, it is advised to use <c>-S</c> instead.</p></note>
       </item>
       <tag><c>-s Mod [Func [Arg1, Arg2, ...]]</c></tag>
       <item>
@@ -359,6 +385,12 @@ foo:bar([baz, '1', '2']).</code>
           some variant of <c>spawn</c> in such cases.</p>
         <p>Because of the limited length of atoms, it is recommended to
           use <c>-run</c> instead.</p>
+        <note><p>This flag will not forward arguments beginning with
+          a hyphen (-) to the specified function, as these will be
+          interpreted as flags to the runtime. If the function uses
+          flags in this form, it is advised to use <c>-S</c> instead,
+          with the additional caveat that arguments are passed as strings
+          instead of atoms.</p></note>
       </item>
     </taglist>
   </section>
--- a/erts/doc/src/erl_cmd.xml
+++ b/erts/doc/src/erl_cmd.xml
@@ -627,6 +627,17 @@ $ <input>erl \
           slave node on a remote host; see
           <seeerl marker="stdlib:slave"><c>slave(3)</c></seeerl>.</p>
       </item>
+      <tag><c><![CDATA[-S Mod [Func [Arg1, Arg2, ...]]]]></c> (init
+        flag)</tag>
+      <item>
+        <p>Makes <c><![CDATA[init]]></c> call the specified function.
+          <c><![CDATA[Func]]></c> defaults to <c><![CDATA[start]]></c>.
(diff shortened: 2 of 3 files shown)


bjorng:
A slight correction: you can pass arguments to a function called using `-s` or `-run`, but with the fatal caveat that those arguments must not begin with a hyphen. Thus, if I define:

```
-module(t).
-export([start/1]).
    start(As) ->
    io:format("~p\n", [As]),
    halt().
```

I can start `erl` like this:

```
% erl -s t start port: 80
Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

['port:','80']
```

or like this:

```
% erl -run t start port: 80
Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

["port:","80"]
```

Passing the arguments for simple command-line HTTP server like this:

```
erl -s httpd serve port: 1234
```

would be an improvement over having to write `erl -s httpd serve -extra --port 1234`, but it still not good enough. One should not have to learn a new way to pass arguments on the command line.

Therefore, I am in favour of this pull request, but with one modification. I don't think that the called function should have to call `init:get_plain_arguments/0` to retrieve the arguments. Instead, I think that the function should be passed a list of the arguments as strings. That is:

```
erl -S httpd serve --port 1234
```

should result in the following function call:

```
httpd:serve(["--port","1234"])
```

and:

```
erl -S httpd serve
```

should result in the following call:

```
httpd:serve([])
```

bjorng:
> I assume we would not want to change the behaviour for the existing flags?

Correct, we can't change the current behaviour.

I suggest doing some refactoring. What I would do is to first simplify the second clause of `start_it/1` like so:

```
start_it({apply,M,F,Args}) ->
    case code:ensure_loaded(M) of
        {module, M} ->
            apply(M, F, Args);
        {error, Reason} ->
            Message = [explain_ensure_loaded_error(M, Reason), <<"\r\n\r\n">>],
            erlang:display_string(binary_to_list(iolist_to_binary(Message))),
            erlang:error(undef)
    end.
```

I would then remove `prepare_run_args/1` and integrate its functionality into `parse_boot_args/1`(and some new helper functions) along with the functionality that I removed from `start_it/1`.

erts/doc/src/init.xml
```diff
@@ -310,6 +310,27 @@ BF</pre>
...
+          0. Otherwise it is assumed to be of arity 1, taking the list
+          <c>[Arg1,Arg2,...]</c> as argument. All arguments are passed
+          as strings. If an exception is raised, Erlang stops with an
+          error message.</p>
+        <p>Example:</p>
+        <pre>
+          % <input>erl -S httpd serve --port 8080 /var/www/html</input></pre>
+        <p>This starts the Erlang runtime system and evaluates
+          the function <c>httpd:serve(["--port", "8080", "/var/www/html"])</c>,
+          terminating additional processing of arguments to the runtime.</p>
```
bjorng, line 327: Could you express this in a clearer way? Perhaps saying that all arguments up to the end of the command line will be passed to the called function?

bjorng (approved):

bjorng:
The behaviour when no module name follows `-s` or `-run` has changed. Before this PR:

```
% erl -run              
Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

1> 
```

With this PR:

```
% bin/erl -run
System process <0.0.0> terminated: {function_clause,[{init,run_args_to_mfa,[[]],[]},{init,parse_boot_args,4,[]},{init,boot,1,[]}]}
(no logger present) unexpected logger message: {log,error,"Error in process ~p with exit value:~n~p~n",[<0.0.0>,{function_clause,[{init,run_args_to_mfa,[[]],[]},{init,parse_boot_args,4,[]},{init,boot,1,[]}]}],#{error_logger=>#{emulator=>true,tag=>error},pid=><0.0.0>,time=>1693291497039471,gl=><0.0.0>}}

```

`-s` and `-S` crashes similarly.

There could potentially be scripts that depend on current behaviour, so I think it's best to keep it.

For `-S`, I think it would be better to print an error message explaining how to use the flag.

u3s (approved):

bjorng:
Added to our daily builds.

bjorng:
I am ready to merge this PR now. Before I do that, can you update the commit message so that it correctly describes what is implemented in this PR?

bjorng:
Thanks for your pull request.

u3s:
when testing #7299, I've face potential issue with `-S` flag.

It seems that when `ERL_AFLAGS` is specified (for example containing `-kernel shell_history enabled`), extra `--` is passed to emulator.

Once shell is started, `erlang:system_info(emu_args).` returns a list ending with "--" string.

This becomes a problem for httpd:serve function when it tries to parse the arguments - for example when port is also specified.
Like this:

`ERL_AFLAGS="-kernel shell_history enabled" erl -S httpd serve --port 0`

will crash emulator and display error `error: erl -S httpd serve: unknown argument: --`

@garazdawi helped in troubleshooting this situation. what can/should be done for improving it?

garazdawi:
In my opinion `-S` should ignore all `ERL_*FLAGS`, just like `-extra` does.

```erl
> ERL_ZFLAGS="-A 1" erl -extra a b c
Erlang/OTP 27 [DEVELOPMENT] [erts-14.0.2] [source-231c37794b] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Eshell V14.0.2 (press Ctrl+G to abort, type help(). for help)
1> erlang:system_info(emu_args).
["/home/.../git/otp2/bin/x86_64-pc-linux-gnu/beam.smp",
 "--","-root","/home/.../git/otp2","-bindir",
 "/home/.../git/otp2/bin/x86_64-pc-linux-gnu",
 "-progname","erl","--","-home","/home/...","--",
 "-kernel","shell_history","enabled","--","--","-A","1","--",
 "-extra","a","b","c"]
2> init:get_plain_arguments().
["a","b","c"]
```

I've written some code that does this and fixes some other problems with -S (for instance allowing `+ABC` to be used as an argument). I'll open a PR with the changes later today.

garazdawi:
Opened #7639 with a solution to this issue.

jchristgit:
@bjorng @garazdawi I've been wondering whether it would make sense to have using the `-S` flag automatically set `-noshell`, or at least default to using it (maybe adding a `-shell` option to invert the behavior, if wanted). I believe that for the intended primary usecase (command-line scripts) it would be convenient to set this, and there's still time to change it before OTP 27. What do you think?

garazdawi:
Yes, that seems like a good idea. If the user want a shell they can call `shell:start_interactive()`, no need to add a `-shell` flag.

PR #7473, opened 2023-07-04
Fix deadlock when setting encoding on Windows

### LINKED ISSUE #7459: Erlang/OTP 26 write to stdout hangs in latin1 encoding on Windows
**Describe the bug**
After changing `standard_io` encoding to `latin1` all writes to the output device hang.

**To Reproduce**

Run from windows command prompt
```
erl -noshell -eval "Path = os:find_executable('erl'), Port1 = erlang:open_port({spawn_executable, Path}, [binary, {args, ['-noshell', '-kernel', 'standard_io_encoding', 'latin1', '-eval', \"io:setopts(standard_io, [{encoding, latin1}]), io:format('duπa~n', []), erlang:halt()\"]}, exit_status]), receive {Port1, {data, A}} -> io:fwrite(A, []) after 2000 -> io:fwrite('nothing~n') end, erlang:halt()"
```
Result on windows
```
nothing
```

**Expected behavior**
Should print
```
du\x{3C0}a
```
it works correctly on OTP 25 both windows and *nix, and on OTP 26.0.2 on *nix

**Affected versions**
OTP 26.0.2

**Additional context**

This bug was tricky to track as it hangs only with a spawned process. If the same code is run directly from cmd shell it works correctly

```
erl -noshell -kernel standard_io_encoding latin1 -eval "io:setopts(standard_io, [{encoding, latin1}]), io:format('duπa~n', []), erlang:halt()"
```
prints
```
du\x{3C0}a
```

This bug breaks ElixirLS on windows https://github.com/elixir-lsp/elixir-ls/issues/927

Possibly this is a regression of https://github.com/erlang/otp/issues/7261 once fixed in https://github.com/erlang/otp/commit/2bbeed333b2314f91947c79ce9809bab5a0f6a19. The original script from that bug as well as a modified one hangs as well

```
-module(foo).
-compile(export_all).

run() ->
  io:put_chars("START\n"),

  open_port(
    {spawn_executable, os:find_executable("erl")},
    [exit_status, binary,
     {args, ["-noshell", "-kernel", "standard_io_encoding", "latin1", "-eval", "io:setopts(standard_io, [{encoding, latin1}]), io:write('teπt'), erlang:halt()."]}]
  ),

  loop().

loop() ->
  receive
    {_, {data, Data}} ->
      io:put_chars(Data),
      loop();

    {_, {exit_status, _}} ->
      erlang:halt()
  end.
```

Related to https://github.com/erlang/otp/issues/7230, https://github.com/erlang/otp/pull/7384

garazdawi:
This PR fixes a bug when calling `io:setopts([{encoding,unicode}])` on Windows and also cleans up and documents how to bytewise encoded data should be handled by standard_io.

closes #7459

### DIFF
--- a/erts/doc/src/erlang.xml
+++ b/erts/doc/src/erlang.xml
@@ -2915,8 +2915,23 @@ uncompiled code with the same arity are mapped to the same list by
           groups have a <em>group leader</em>. All I/O from the group
           is channeled to the group leader. When a new process is
           spawned, it gets the same group leader as the spawning
-          process. Initially, at system startup, <c>init</c> is both
-          its own group leader and the group leader of all processes.</p>
+          process.</p>
+        <p>Initially, at system startup, <c>init</c> is both
+          its own group leader and the group leader of all processes.
+          During the boot of a system the group leader for processes
+          will be changed depending on the need of the system. Some examples
+          where this is done are:</p>
+        <list type="bulleted">
+          <item>When an application is started, the top supervisor of that
+            application will have its group leader set to the application
+            master. See <seemfa marker="kernel:application#start/2"><c>
+            application:start/2</c></seemfa> for more details.</item>
+          <item>When running tests, both <seeapp marker="common_test:index"><c>common_test</c></seeapp> and
+            <seeerl marker="eunit:eunit"><c>eunit</c></seeerl> set the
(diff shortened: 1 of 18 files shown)


garazdawi:
Merged to main for release in OTP-26.1.

PR #7475, opened 2023-07-04
Add sign_fun config option (SSL)

ziopio:
# sign_fun 
This option allows the user to define a function tasked to sign an ssl message. This enables to delegate the private key signature to sources like: Secure Elements.

Users are free to program such function the way they think is best, allowing them to support any private key storage
or to delegate signature to any external service or device.
## How to use
The tls settings try to be similar to what is already in place to support crypto engines.
For this reason is necessary to provide a proper `algorithm` atom along side with the `sign_fun`
```erlang
    TLS_Options = [
        {cacertfile, "my.CA.pem"},
        {certfile, "me.pem"},
        {verify, verify_peer},
        {key, #{
            algorithm => rsa, % any of the supported key types for Key-Cert pairs
            sign_fun => {Module, Fun} % or fun()
        }}
    ]
```

## Tests

I implemented a testcase (`client_auth_custom_sign`) in the `ssl_cert_SUITE` which tests this feature with various combinations of key types both on the client and on the server. This test is just to prove that the feature is working, not sure where and what to add to the test suites.
To make all key-cert combos work, I had to adapt few functions that manipulate the configuration to correctly process the new config option. I kept the changes to the minimum, given my limited knowledge about the code-base.

### GRiSP

At Stritzinger, we use it in a slightly old version to enable our GRiSP board to use the OTP SSL app, while storing the private key in a secured element.
https://github.com/grisp/grisp_cryptoauth/blob/5b9ca3cfaf5a1d07baaed2bd4f00bceba9501a2d/src/grisp_cryptoauth.erl#L94

We came up with this idea since we noticed that the PKCS11 crypto engine doesn't cover all use cases for HSMs. We faced few technical difficulties, like loading shared libs.

Currently we have to patch OTP to include such feature in our boards, but we would like to see this part of the lib since we believe it would be a great addition.

### Braid

Recently we also published an open source project that uses this feature in combination with erlang distribution.

Here where it is used to sign ssl for `erlang distribution`:
Where is set: https://github.com/stritzinger/braidnode/blob/main/priv/ssl_dist_opts.rel
Where is implemented: https://github.com/stritzinger/braidnode/blob/main/src/braidnode_crypto.erl

In this last case we do not use a secure element but instead, we detached the key storage on another service that holds the key and signs for us.

### DIFF
--- a/lib/public_key/test/public_key_SUITE.erl
+++ b/lib/public_key/test/public_key_SUITE.erl
@@ -658,9 +658,10 @@ encrypt_decrypt_sign_fun() ->
 encrypt_decrypt_sign_fun(Config) when is_list(Config) ->
     {PrivateKey, _DerKey} = erl_make_certs:gen_rsa(64),
     #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateKey,
-    EncryptFun = fun (PlainText, Options) ->
-        public_key:encrypt_private(PlainText, PrivateKey, Options)
-    end,
+    EncryptFun =
+        fun (PlainText, Options) ->
+                public_key:encrypt_private(PlainText, PrivateKey, Options)
+        end,
     CustomPrivKey = #{sign_fun => EncryptFun},
     PublicKey = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
     Msg = list_to_binary(lists:duplicate(5, "Foo bar 100")),
@@ -753,8 +754,8 @@ custom_sign_fun_verify(Config) when is_list(Config) ->
     #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateRSA,
     PublicRSA = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
     SignFun = fun (Msg, HashAlgo, Options) ->
-        public_key:sign(Msg, HashAlgo, PrivateRSA, Options)
-    end,
+                      public_key:sign(Msg, HashAlgo, PrivateRSA, Options)
+              end,
     CustomKey = #{algorithm => rsa, sign_fun => SignFun},
 
     Msg = list_to_binary(lists:duplicate(5, "Foo bar 100")),
--- a/lib/public_key/test/public_key_SUITE.erl
+++ b/lib/public_key/test/public_key_SUITE.erl
@@ -77,12 +77,16 @@
          cert_pem/1,
          encrypt_decrypt/0,
          encrypt_decrypt/1,
+         encrypt_decrypt_sign_fun/0,
+         encrypt_decrypt_sign_fun/1,
          rsa_sign_verify/0,
          rsa_sign_verify/1,
          rsa_pss_sign_verify/0,
          rsa_pss_sign_verify/1,
          dsa_sign_verify/0,
          dsa_sign_verify/1,
+         custom_sign_fun_verify/0,
+         custom_sign_fun_verify/1,
          pkix/0,
          pkix/1,
          pkix_countryname/0,
@@ -147,6 +151,7 @@ all() ->
      appup,
      {group, pem_decode_encode},
      encrypt_decrypt,
+     encrypt_decrypt_sign_fun,
      {group, sign_verify},
      pkix, 
      pkix_countryname, 
@@ -181,7 +186,7 @@ groups() ->
 			      ec_pem_encode_generated, gen_ec_param_prime_field,
 			      gen_ec_param_char_2_field]},
      {sign_verify, [], [rsa_sign_verify, rsa_pss_sign_verify, dsa_sign_verify,
-                        eddsa_sign_verify_24_compat]}
+                        eddsa_sign_verify_24_compat, custom_sign_fun_verify]}
     ].
 %%-------------------------------------------------------------------
 init_per_suite(Config) ->
@@ -646,6 +651,22 @@ encrypt_decrypt(Config) when is_list(Config) ->
     RsaEncrypted2 = public_key:encrypt_public(Msg, PublicKey),
     Msg = public_key:decrypt_private(RsaEncrypted2, PrivateKey),
     ok.
+
+%%--------------------------------------------------------------------
+encrypt_decrypt_sign_fun() ->
+    [{doc, "Test public_key:encrypt_private with user provided sign_fun"}].
+encrypt_decrypt_sign_fun(Config) when is_list(Config) ->
+    {PrivateKey, _DerKey} = erl_make_certs:gen_rsa(64),
+    #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateKey,
+    EncryptFun = fun (PlainText, Options) ->
+            public_key:encrypt_private(PlainText, PrivateKey, Options)
+        end,
+    CustomPrivKey = #{sign_fun => EncryptFun},
+    PublicKey = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
+    Msg = list_to_binary(lists:duplicate(5, "Foo bar 100")),
+    RsaEncrypted = public_key:encrypt_private(Msg, CustomPrivKey),
+    Msg = public_key:decrypt_public(RsaEncrypted, PublicKey),
+    ok.
        
 %%--------------------------------------------------------------------
 rsa_sign_verify() ->
@@ -722,6 +743,28 @@ dsa_sign_verify(Config) when is_list(Config) ->
 			      {DSAPublicKey, DSAParams}), 
     false = public_key:verify(Digest, none, <<1:8, DigestSign/binary>>, 
 			      {DSAPublicKey, DSAParams}).
+%%--------------------------------------------------------------------
+
+custom_sign_fun_verify() ->
+    [{doc, "Checks that public_key:sign correctly calls the `sign_fun`"}].
+custom_sign_fun_verify(Config) when is_list(Config) ->
+    {_, CaKey} = erl_make_certs:make_cert([{key, rsa}]),
+    PrivateRSA = public_key:pem_entry_decode(CaKey),
+    #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateRSA,
+    PublicRSA = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
+    SignFun = fun (Msg, HashAlgo, Options) ->
+            public_key:sign(Msg, HashAlgo, PrivateRSA, Options)
+        end,
+    CustomKey = #{algorithm => rsa, sign_fun => SignFun},
+
+    Msg = list_to_binary(lists:duplicate(5, "Foo bar 100")),
+    RSASign = public_key:sign(Msg, sha, CustomKey),
+    true = public_key:verify(Msg, sha, RSASign, PublicRSA),
+    false = public_key:verify(<<1:8, Msg/binary>>, sha, RSASign, PublicRSA),
+    false = public_key:verify(Msg, sha, <<1:8, RSASign/binary>>, PublicRSA),
+
+    RSASign1 = public_key:sign(Msg, md5, CustomKey),
+    true = public_key:verify(Msg, md5, RSASign1, PublicRSA).
 
 %%--------------------------------------------------------------------
 pkix() ->
--- a/lib/public_key/test/public_key_SUITE.erl
+++ b/lib/public_key/test/public_key_SUITE.erl
@@ -659,8 +659,8 @@ encrypt_decrypt_sign_fun(Config) when is_list(Config) ->
     {PrivateKey, _DerKey} = erl_make_certs:gen_rsa(64),
     #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateKey,
     EncryptFun = fun (PlainText, Options) ->
-            public_key:encrypt_private(PlainText, PrivateKey, Options)
-        end,
+        public_key:encrypt_private(PlainText, PrivateKey, Options)
+    end,
     CustomPrivKey = #{sign_fun => EncryptFun},
     PublicKey = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
     Msg = list_to_binary(lists:duplicate(5, "Foo bar 100")),
@@ -753,8 +753,8 @@ custom_sign_fun_verify(Config) when is_list(Config) ->
     #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateRSA,
     PublicRSA = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
     SignFun = fun (Msg, HashAlgo, Options) ->
-            public_key:sign(Msg, HashAlgo, PrivateRSA, Options)
-        end,
+        public_key:sign(Msg, HashAlgo, PrivateRSA, Options)
+    end,
     CustomKey = #{algorithm => rsa, sign_fun => SignFun},
 
     Msg = list_to_binary(lists:duplicate(5, "Foo bar 100")),
--- a/lib/ssl/src/ssl_config.erl
+++ b/lib/ssl/src/ssl_config.erl
@@ -61,7 +61,6 @@ group_pairs([#{certs := []}]) ->
     #{eddsa => [],
       ecdsa => [],
       rsa_pss_pss => [],
-      rsa_pss_rsae => [],
       rsa => [],
       dsa => []
      };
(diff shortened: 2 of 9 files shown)


lib/ssl/src/ssl_config.erl
```diff
@@ -87,6 +87,9 @@ group_pairs([#{private_key := #'DSAPrivateKey'{}} = Pair | Rest], #{dsa := DSA}
 group_pairs([#{private_key := #{algorithm := dss, engine := _}} = Pair | Rest], Group) ->
     Pairs = maps:get(dsa, Group),
     group_pairs(Rest, Group#{dsa => [Pair | Pairs]});
+group_pairs([#{private_key := #{algorithm := Alg, sign_fun := _}} = Pair | Rest], Group) ->
```
ziopio, line 90: Here my test would fail with `rsa_pss_rsae`, there is already a conversion in place for `dss` to `dsa` for the engine, I could do something similar and convert here, maybe to `rsa_pss_pss` or just `rsa`.

lib/ssl/test/ssl_cert_tests.erl
```diff
@@ -96,7 +98,42 @@ auth(Config) ->
...
+%%--------------------------------------------------------------------
+client_auth_custom_sign() ->
+    [{doc,"Test that client and server can connect using their own signature function"}].
+
+client_auth_custom_sign(Config) when is_list(Config) ->
+    KeyAlg =  proplists:get_value(cert_key_alg, Config),
+    % 'rsa_pss_rsae' is not an expected config value here, 
+    % using rsa or rsa_pss_pss instad seams to work,
+    % Not sure how to modify SSL to add it in the acceptable key pair groups...
+    CompatibleKeyAlg = case KeyAlg of rsa_pss_rsae -> rsa_pss_pss; _ -> KeyAlg end,
```
ziopio, line 110: I am in doubt on this, should I extend the group pairs, convert in the lib, or require the user to handle it as I do in this test?

ziopio:
SSL lib test run on master:
`Testing tests.ssl_test: TEST COMPLETE, 2702 ok, 84 failed, 690 skipped of 3476 test cases`
SSL lib test run on this PR:
`Testing tests.ssl_test: TEST COMPLETE, 2723 ok, 84 failed, 690 skipped of 3497 test cases`

IngelaAndin:
I think that is desired to include such functionality. I am not sure about how we will prefer the implementation we will need to think about that, and it is still vacation period so please have some patience! I think we can work something out that will work for us you and us in the end.

lib/ssl/test/ssl_cert_tests.erl
IngelaAndin: rsa_pss_pss and rsa_pss_rsae are only vaild for TLS-1.2 and TLS-1.3!

lib/ssl/src/ssl_config.erl
IngelaAndin: This situation is a bit different.  DSA and DSS is just two names for the conceptually same thing. There is a theoretical difference being that one being  a reference  to the whole standard and the other to the algorithm. (I think this is mostly unfortunate and confusing, but it is at it is!) When it comes to rsa_pss_pss and rsa_pss_rsae it now depends on if the rsa key has any parameters or not.  Certificates using a certificate has a key that has no parameters must use the rsa identifier. So we can handle it as rsa or rsa_pss_pss.

IngelaAndin:
In general I think the public_key functions should handle the fun.  The handling of engines could be moved to public_key, or perhaps we could add a check that map value is not a fun and still call crypto directly to avoid overhead (I am not convinced the overhand of  a function call would actually be an issue, but it would be a smaller behavior change).

lib/ssl/src/ssl_config.erl
ziopio: Ok, so I can just add a conversion to `rsa` for example. Since the key will be utilized by the user's function code anyway,  it shouldn't matter much in this case.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2148,6 +2148,12 @@ digitally_signed(Version, Msg, HashAlgo, PrivateKey, SignAlgo) ->
 	    throw(?ALERT_REC(?FATAL, ?HANDSHAKE_FAILURE, bad_key(PrivateKey)))
     end.
 
+do_digitally_signed(Version, Msg, HashAlgo, #{algorithm := _KeyAlg, 
+                                              sign_fun := Sign} = Key, SignAlgo) when is_function(Sign) ->
+    Sign(Version, Msg, HashAlgo, Key, SignAlgo);
+do_digitally_signed(Version, Msg, HashAlgo, #{algorithm := _KeyAlg,
+                                              sign_fun := {Mod, Fun}} = Key, SignAlgo) ->
+    Mod:Fun(Version, Msg, HashAlgo, Key, SignAlgo);
```
ziopio, line 2156: @IngelaAndin Ok I can try moving the handling of `sign_fun` in `public_key`.
But your second options sound like what is already here, or are you saying to move even the calls to crypto into `public_key`?

lib/ssl/test/ssl_cert_tests.erl
ziopio: Good to know, but this line only does a translation of the KeyAlg provided by the test group settings.
So those settings are right I presume. 
So this is not wrong, right? It could be moved to an internal translation as I suggest above.

ziopio:
@IngelaAndin can you provide some more explanation on the desired changes? Then I would try to move things the way you suggest.

IngelaAndin:
@ziopio  Sorry I have been a little preoccupied, but I will try to find some time for this later today, have to attend a meeting at the moment.

lib/ssl/src/ssl_handshake.erl
IngelaAndin: So public key is the more general API. That is why I think it
would be better if it is the place that handles different type of
keys. Although engine keys are very OpenSSL specific and hence
will only be a passthrough to crypto. When Engine was a special
case that did not fit under the general case that was that all
general keys where supplied as DER (raw) or through PEM-format
the reason to go through public_key was not obvious. But now
engine becomes a corner case of the more general case of suppling
a fun. And this may be something that is interesting to other
users, than ssl, of the public_key API.

To support old TLS-versions (pre TLS-1.2) we will also need
support for the encrypt/decrypt_private that only will need to
handle plain RSA keys (with param = NULL).

lib/ssl/test/ssl_cert_tests.erl
IngelaAndin: es we will need to specify that a key can be of type rsa_pss_pss. But rsa_pss_rsae is just a plain
rsa-key used with special options. Here there is a consideration that the person suppling the fun
must supply a fun that can handle rsa_pss_rsae options or they have to disable this signature_algorithm
if TLS-1.2 and/or TLS-1.3 are supported.

lib/ssl/test/ssl_cert_tests.erl
```diff
@@ -516,3 +553,30 @@ group_config(Config, ServerOpts, ClientOpts) ->
                 {[{supported_groups, [x448, x25519]} | ServerOpts],
                  [{groups,"P-256:X25519"} | ClientOpts]}
         end.
+
+choose_sign_fun(#'RSAPrivateKey'{} = Key) ->
+    fun (_, {digest, Digest}, _HashAlgo, _KeyOpts, rsa) ->
+            public_key:encrypt_private(Digest, Key, [{rsa_pad, rsa_pkcs1_padding}]);
+        (_, Msg, HashAlgo, _KeyOpts, SignAlgo) ->
+            Options = signature_options(SignAlgo, HashAlgo),
+            public_key:sign(Msg, HashAlgo, Key, Options)
```
IngelaAndin, line 562: To support old TLS-versions (pre TLS-1.2) we will also need
support for the encrypt/decrypt_private that only will need to
handle plain RSA keys (with param = NULL). This choice needs to be done in ssl.

lib/ssl/test/ssl_cert_tests.erl
IngelaAndin: I think this is something that any way should be added to the engine support (even without this PR)  and was sort of overlooked when introducing the rsa_pss_pss support.  I know both rsa_pss_pss and rsa_pss_rsae use the same option
list. That is related to the the crypto API. The big difference beeing that pss_pss  signature also uses the RSA-keys special params, that do not exist in the pss_rsae case.

IngelaAndin:
@ziopio  I tried to give some explanations please feel free to ask more questions if needed!

ziopio:
@IngelaAndin I applied the refactoring, `ssl_cert_SUITE` and `publick_key_SUITE` run fine. 
What do you think?
Before going on I have a couple of questions:
- `public_key:decrypt_private` looks like is not used by ssl, (?) If I also extend to use sign-fun it how should I test it?
- Can you tell me precisely what/where I need to change regarding `rsa_pss_rsae` handling? It is not clear to me what is the problem right now since I did not touch code that handles that option. I read about it [here](https://crypto.stackexchange.com/questions/58680/whats-the-difference-between-rsa-pss-pss-and-rsa-pss-rsae-schemes) but basically I only stumbled upon it in the test case. Is it maybe out of scope for this PR?

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2152,17 +2152,9 @@ do_digitally_signed(Version, Msg, HashAlgo, {#'RSAPrivateKey'{} = Key,
...
-do_digitally_signed(Version, {digest, Digest}, _,
-                    #{algorithm := rsa} = Engine, rsa) when ?TLS_LTE(Version, ?TLS_1_1) ->
-    crypto:private_encrypt(rsa, Digest, maps:remove(algorithm, Engine),
-                           rsa_pkcs1_padding);
-do_digitally_signed(_, Msg, HashAlgo, #{algorithm := Alg} = Engine, SignAlgo) ->
-    Options = signature_options(SignAlgo, HashAlgo),
-    crypto:sign(Alg, HashAlgo, Msg, maps:remove(algorithm, Engine), Options);
-do_digitally_signed(Version, {digest, _} = Msg , HashAlgo, Key, _) when ?TLS_LTE(Version,?TLS_1_1) ->
+do_digitally_signed(Version, {digest, Digest}, _HashAlgo, Key, rsa) when ?TLS_LTE(Version, ?TLS_1_1) ->
+    public_key:encrypt_private(Digest, Key);
```
IngelaAndin, line 2156: You need to preserve the padding option here!

IngelaAndin:
Line 1214 ssl_hndshake.erl:

```
premaster_secret(EncSecret, #'RSAPrivateKey'{} = RSAPrivateKey) ->
    try public_key:decrypt_private(EncSecret, RSAPrivateKey,
				   [{rsa_pad, rsa_pkcs1_padding}])
    catch
	_:Reason:ST ->
            ?SSL_LOG(debug, decrypt_error, [{reason, Reason}, {stacktrace, ST}]),
	    throw(?ALERT_REC(?FATAL, ?DECRYPT_ERROR))
    end;
```

Only used in RSA-kex exchange ciphers not supported by default, considered broken.

IngelaAndin:
ssl needs to know the type of the key to be able to chose signature algorithm. The function that provides the signature using the key need to provide the signature according to ssl's choice.  So if the key is a plain rsa key two signature algorithms may be chosen rsa or rsa_pss_rsae. These are both variants of  plain RSA-signatures just using different paddings  and hash functions. Today this is reflected by using default options or pss_options to in the end crypto functions.  Now if you provide a external function to handle the signature it need to know what type of signature, and maybe we should provide also an rsa_pss_rsae atom and let public key have the knowledge of crypto pss_options. If the externa function would not support both kinds of  plain rsa-signatures then ssl would have to be configured too. Also in the case of rsa_pss_pss key which is a separate type of key, the pss options needs to be passed to crypto functions.

lib/ssl/src/ssl_handshake.erl
ziopio: If you refer to the old line 2160; I'm not calling `crypto:private_encrypt` directly, but  `public_key:encrypt_private` which applies `default_options([])` that adds  `[{rsa_padding, rsa_pkcs1_padding}];`. Is it too hidden?

IngelaAndin: Ah, that should be ok then. No problem just looking at patch diffs it was not obvious :)

IngelaAndin:
Sorry  if I confused you. But we can not add rsa_pss_rsae as a key type. The key-type needs to still be rsa. However we can add an attribute to "key describing map" with  rsa_pss_rsae atom or rsa to help public-key function know which signing algorithm that was chosen so that it can generate it correctly.

ziopio:
I noticed that `rsa_pss_rsae` was indeed passed to `do_digitally_signed` as `SignAlgo`.
So I though that it was the best place to just pass it to `public_key:sign`.

IngelaAndin:
Well I see how you think, but looking at how the code turned out and the existing API description I  now think
it would be better to just let ssl generate the options using the `SignAlgo.`  As the existing API already will put the requirement
on the  custom_sign function to be able to handle these options. In this way we avoid changing the public_key API in a hard to explain exactly why way. We only extend it with the custom fun, that in the end needs to be documented in public_key and there ought to be a small test in public_key too.

ziopio:
Before touching the type specs and the documentation I wait your approval on the current state of the code.

- As you can see, `public_key` really only needs `sign_fun` attributes but gets it in a map. Do you have any cleaner idea?

- This new type of key would be presented primarily in the `public_key` API docs, while SSL docs will just mention/link to it. right?

- This `sign_fun` can have 2 arg count: 2 arguments if fed to `public_key:encrypt_private` or 3 arguments for `public_key:sign`, in the SSL tests i simulate the user choice by catching if we are going to use the legacy `tls` versions or not. I personally do not like too much that we have a different arity for the same function, since now the user must pay attention to this detail. On the other hand this looks simple... What do you think?

IngelaAndin:
@ziopio  Yes absolutely, we would like to take a little deeper dive into this as a team to straighten out API concerns and not having you jumping back and forth. We will try to plan this for next sprint so please have some patience, we just want it to be good :)

u3s (commented):
AFAIK API discussion did not took place yet. I will clarify that with @IngelaAndin.

Some minor comments dropped inline. When I wrote them I wasn't aware you decided to postpone doc update waiting for final API.

We will try to come back with API suggestion soon.

IngelaAndin:
@ziopio FYI design discussion is initiated. We have a lot of balls in the air the moment but hope to have some feedback for you sometime next week.

IngelaAndin:
@ziopio FYI there has been a really busy week, so final result has to wait a little longer. I have made an internal PR with a prototype commit on top of your branch that needs so further discussions before we can make a recommendation for you.

u3s (commented):
minor indentation comments

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -646,6 +651,22 @@ encrypt_decrypt(Config) when is_list(Config) ->
...
+
+%%--------------------------------------------------------------------
+encrypt_decrypt_sign_fun() ->
+    [{doc, "Test public_key:encrypt_private with user provided sign_fun"}].
+encrypt_decrypt_sign_fun(Config) when is_list(Config) ->
+    {PrivateKey, _DerKey} = erl_make_certs:gen_rsa(64),
+    #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateKey,
+    EncryptFun = fun (PlainText, Options) ->
+        public_key:encrypt_private(PlainText, PrivateKey, Options)
+    end,
```
u3s, line 663: pls format it as Emacs does.
image

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -722,6 +743,28 @@ dsa_sign_verify(Config) when is_list(Config) ->
...
+custom_sign_fun_verify() ->
+    [{doc, "Checks that public_key:sign correctly calls the `sign_fun`"}].
+custom_sign_fun_verify(Config) when is_list(Config) ->
+    {_, CaKey} = erl_make_certs:make_cert([{key, rsa}]),
+    PrivateRSA = public_key:pem_entry_decode(CaKey),
+    #'RSAPrivateKey'{modulus=Mod, publicExponent=Exp} = PrivateRSA,
+    PublicRSA = #'RSAPublicKey'{modulus=Mod, publicExponent=Exp},
+    SignFun = fun (Msg, HashAlgo, Options) ->
+        public_key:sign(Msg, HashAlgo, PrivateRSA, Options)
+    end,
```
u3s, line 757: pls do Emacs formatting
image

IngelaAndin:
@ziopio  #7898  I have made this PR with a commit on top of yours to show what was the outcome of our internal API discussions is. I was not allowed to push it to your repo. If it is ok with you I can use that PR instead to finalize this, I will then squash the commits and keep you as author.

ziopio:
@IngelaAndin The new API looks good to me. I'm fine if you prefer working on the other PR.
If you need any help or further contribution by me just ask.

IngelaAndin:
Closed in favor of #7898

PR #7481, opened 2023-07-06
More precise specs for some functions in the `binary` module

Maria-12648430:
In the `binary` module, the functions `longest_common_prefix/1` and `longest_common_suffix/1` expect the given list of binaries to be non-empty. This is not reflected in the specs of or documentation for those functions.

Also, the functions ` compile_pattern/1`, `match/2,3`, `matches/2,3`, `replace/3,4` and `split/2,3`, if the pattern is given as a list of binaries, expect that list to be non-empty. This is not reflected in the specs of or documentation for those functions.
Also, for those functions, any binary given as pattern must be non-empty. This is mentioned in the documentation, but not reflected in the specs.

### DIFF
--- a/lib/stdlib/doc/src/binary.xml
+++ b/lib/stdlib/doc/src/binary.xml
@@ -149,11 +149,12 @@
       only one element. The order of alternatives in a pattern is
       not significant.</p>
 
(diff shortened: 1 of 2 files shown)


garazdawi:
Thanks!

PR #7484, opened 2023-07-07
(feat) Support CDATA in xml export

leonardb:
`#xmlText` is spec'd with `type = text :: text | cdata` yet the type is ignored when the XML doc is exported

Where `#xmlText.type = cdata` the value for the element will be wrapped as ` ["<![CDATA[", TextValue, "]]>"]`

This allows generating XML which does not convert characters `<`, `>` and `&` to `&lt;`, `&gt;` and `&amp;`


lib/xmerl/src/xmerl_lib.erl
```diff
@@ -45,6 +45,9 @@
 -include("xmerl_xsd.hrl").
 
 
+export_cdata(T) ->
+    ["<![CDATA[", T, "]]>"].
```
Maria-12648430, line 49: Please keep in mind that `T` itself may contain `]]>`, in which case the result will be invalid, or rather, not what you would expect. One way to escape `]]>` in CDATA that I know of is to replace it with `]]]]><[CDATA[>`.

leonardb: Thanks for pointing that out. I suppose the only way to cater for that would be
```diff
diff --git a/lib/xmerl/src/xmerl_lib.erl b/lib/xmerl/src/xmerl_lib.erl
index 2e233637d6..3235b89af7 100644
--- a/lib/xmerl/src/xmerl_lib.erl
+++ b/lib/xmerl/src/xmerl_lib.erl
@@ -46,7 +46,13 @@
 
 
 export_cdata(T) ->
-    ["<![CDATA[", T, "]]>"].
+    ["<![CDATA[",
+     binary:replace(
+       iolist_to_binary(T),
+       <<"]]>">>,
+       <<"]]]]><![CDATA[>">>,
+       [global]),
+     "]]>"].
 
 %% Escape special characters `<' and `&', flattening the text.
 %% Also escapes `>', just for symmetry.
```

Maria-12648430: Hm... I'm pretty sure this doesn't do the right thing.

OTOH, if the argument list `T` contains unicode data, eg `"ħ"` (same as `[295]`), `iolist_to_binary` will fail.

OTOH, if `T` contains something like `"Ä§"` (same as `[196, 167]`), `iolist_to_binary` will accept and convert it to `<<196, 167>>`, but `unicode:characters_to_list` will turn that into `"ħ"` (`[295]`).

leonardb:
Hmm, would you suggest just using lists:flatten/1 and 'walking' the
resulting list rather than binary conversion and replace using Unicode
functions? It should be safer.

lib/xmerl/src/xmerl_lib.erl
leonardb: So looking at `xmerl_lib:export_text/1` it is not unicode safe at all.

The safest way I can see to handle this using existing code is to just convert `T` to graphemes using `string:to_graphemes/1` beforehand. (results using this below)

```erlang
37> io:format("~ts~n", [xmerl_lib:export_cdata(["some ", [<<"Ä§"/utf8>>, " ", "Ä§", " ħaa", <<" ❤️ "/utf8>>], " ]]> ", " other "])]).
some Ä§ Ä§ ħaa ❤️  ]]]]><![CDATA[>  other 
ok
38> io:format("~ts~n", [xmerl_lib:export_text(["some ", [<<"Ä§"/utf8>>, " ", "Ä§", " ħaa", <<" ❤️ "/utf8>>], " ]]> ", " other "])]). 
some ÃÂ§ Ä§ ħaa â¤ï¸  ]]&gt;  other 
```

I'm guess making `xmerl_lib:export_text/1` unicode safe should probably be done in another PR

Maria-12648430:
> Hmm, would you suggest just using lists:flatten/1 and 'walking' the resulting list rather than binary conversion and replace using Unicode functions? It should be safer.

Hm, probably, I'm not sure about performance, though :sweat_smile: But it would be a little complicated if what you could get as input was indeed an iolist... even after flattening, you could have a mixture of bytes and binaries, and the `]]>` could be split up over them, like `[$a, $b, $c, $], <<$]>>, $>, $d, $e, $f]`...

Maria-12648430:
What do you think of this?

```erlang
-define(cdata_end_delim_escape, $], $], $], $], $>, $<, $!, $[, $C, $D, $A, $T, $A, $[, $>).

export_cdata(T) ->
    ["<![CDATA[", escape_cdata(T), "]]>"].

escape_cdata(T) when is_binary(T) ->
    escape_cdata_binary(T);
escape_cdata(T) ->
    escape_cdata_iolist(lists:filter(fun(X) -> X =/= <<>> end,
                        lists:flatten(T))).

escape_cdata_iolist([]) ->
    [];
escape_cdata_iolist([$], $], $> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([$], $], << $> >> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([$], $], << $>, MoreB/binary >> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist([MoreB | More])];
escape_cdata_iolist([$], << $] >>, $> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([$], << $], $> >> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([$], << $], $>, MoreB/binary >> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist([MoreB | More])];
escape_cdata_iolist([<< $] >>, $], $> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([<< $] >>, $], << $> >> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([<< $] >>, $], << $>, MoreB/binary >> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist([MoreB | More])];
escape_cdata_iolist([<< $], $] >>, $> | More]) ->
    [?cdata_end_delim_escape | escape_cdata_iolist(More)];
escape_cdata_iolist([Bin | More]) when is_binary(Bin) ->
    [escape_cdata_binary(Bin) | escape_cdata_iolist(More)];
escape_cdata_iolist([C | More]) ->
    [C | escape_cdata_iolist(More)].

escape_cdata_binary(<<>>) ->
    <<>>;
escape_cdata_binary(<< $], $], $>, More/binary >>) ->
    << ?cdata_end_delim_escape, (escape_cdata_binary(More))/binary >>;
escape_cdata_binary(<< C, More/binary >>) ->
    << C, (escape_cdata_binary(More))/binary >>.
```

It should catch all occurrences of `]]>` in the given iolist, no matter if they are in binary elements, list bytes, or distributed over a mix like `[ $], << $] >>, $> ]`or `[ << $x, $] >>, $], << $>, $y >> ]`, and replace it with `]]]]><![CDATA[>`.

Maria-12648430:
Ah, wait, it has a bug XD

leonardb:
> ~Ah, wait, it has a bug XD~ ~Fixed.~ Fixed again.

It's not unicode safe though.

Compared output for your code `export_cadata/1` vs the code in the PR `export_cdata2/1`

Input:
```erlang
S = ["some ", [<<"Ä§"/utf8>>, " ", "Ä§", " ħaa", <<" ❤️ "/utf8>>], " ]", [ "]", ["> "]], " other "].
```
Your code:
```erlang
io:format("~ts~n",[cdata:export_cdata(S)]).
<![CDATA[some §ÂÃ Ä§ ħaa ¸ï¤â  ]]]]><![CDATA[>  other ]]>

rp(cdata:export_cdata(S)).
["<![CDATA[",
 [115,111,109,101,32,
  <<167,194,132,195>>,
  32,196,167,32,295,97,97,
  <<32,143,184,239,164,157,226,32>>,
  32,93,93,93,93,62,60,33,91,67,68,65,84,65,91,62,32,32,111,
  116,104,101,114,32],
 "]]>"]

```
PR code:
```erlang
io:format("~ts~n",[cdata:export_cdata2(S)]).
<![CDATA[some Ä§ Ä§ ħaa ❤️  ]]]]><![CDATA[>  other ]]>

rp(cdata:export_cdata2(S)).
["<![CDATA[",
 [115,111,109,101,32,196,167,32,196,167,32,295,97,97,32,
  10084,65039,32,32,93,93,93,93,62,60,33,91,67,68,65,84,65,91,
  62,32,32,111,116,104,101,114,32],
 "]]>"]
```

I think we're entering bikeshedding territory.
1. The existing export_text is not unicode safe
2. I don't think it's adding sufficient overhead to even worry about it when using `string:graphemes` to get a unicode safe flattened list of codepoints to parse

Maria-12648430:
> It's not unicode safe though.

It _would_ be if there hadn't been _yet another_ bug in my code in `escape_cdata_binary` :roll_eyes: Should be fixed now, updated the comment yet again, try again if you like.

> I think we're entering bikeshedding territory.

What does that mean? (The phrase is unfamiliar to me :sweat_smile:)

leonardb:
Bike-shedding origins: https://en.wikipedia.org/wiki/Law_of_triviality

Maria-12648430:
> Bike-shedding origins: https://en.wikipedia.org/wiki/Law_of_triviality

Ah, so there is a name for that :smile: But let's just say that I love to exercise my brain :wink:

> I'll also add that it seems that most of xmerl is not unicode safe. We cannot even round-trip the string we're testing with worried

I'm not debating that, xmerl is apparently in a state of neglect and disrepair :cry:

IngelaAndin:
Alas, this is one of the not so prioritized applications that would benefit from some more love.  We will at least try to give this PR some attention after the vacation period.

leonardb:
Thanks Ingela,

Was debating whether to just close it or working on adding Unicode support
to the whole xmerl app.

If the latter I'd probably lean heavily on the string and Unicode apps.
Would that be preferable?

Maria-12648430:
> adding Unicode support to the whole xmerl app. >
> If the latter I'd probably lean heavily on the string and Unicode apps.

IMO, if you are going to tackle unicode for all of xmerl, then I would say yes, by all means, use them 😄 That's what they are there for, and well tested, no need to reinvent the wheel (with bugs 😅).

I was against using them just like that in the context of this PR, ie without the surrounding app being unicode-safe, not because of performance reasons but because it would rely on assumptions about the data flying around that _may_ hold for all actual use cases, but are not _guaranteed_ to.

With xmerl as a whole unicode-safe, I would say `unicode` and `string` galore 😉

PR #7489, opened 2023-07-11
beam_ssa_type: Fix bugs relating to subtraction in type inference

### LINKED ISSUE #7488: Erlang fails to compile valid code on tuple element guards
The following minimized code fails to compile:

```erlang
-module(broken).
-export([bar/1]).

bar(X) ->
    case foo(X) of
        Tuple when element(1, Tuple) =:= ok ->
            ok;
        Tuple when element(1, Tuple) =:= error ->
            element(2, Tuple)
    end.

foo(X) ->
    case X of
        true -> {ok};
        false -> {error, <<"aaaa">>}
    end.
```

with the error:
Sub pass ssa_opt_type_continue

OTP 26-only

jhogberg:
Fixes #7488

### DIFF
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -26,6 +26,7 @@
 %% it goes.
 %%
 
+-feature(maybe_expr, enable).
 -module(beam_ssa_type).
 -export([opt_start/2, opt_continue/4, opt_finish/3, opt_ranges/1]).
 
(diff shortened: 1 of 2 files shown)


RobinMorisset:
erlfuzz found two crashes that only reproduce with this PR:
```erlang
f(_V0) when is_boolean(_V0), is_binary(_V0), _V0 andalso _V0 ->
    ok.
```
causes the following error:
```
Function: f/1
Sub pass ssa_opt_type_start
/home/rmorisset/minimized/test214716.erl: internal error in pass beam_ssa_opt:
exception error: no case clause matching {none,none}
  in function  beam_ssa_type:update_successors/5 (beam_ssa_type.erl, line 1968)
  in call from beam_ssa_type:sig_bs/8 (beam_ssa_type.erl, line 257)
  in call from beam_ssa_type:sig_function_1/4 (beam_ssa_type.erl, line 220)
  in call from beam_ssa_type:do_sig_function/4 (beam_ssa_type.erl, line 168)
  in call from beam_ssa_type:sig_function/4 (beam_ssa_type.erl, line 159)
  in call from beam_ssa_type:signatures_1/3 (beam_ssa_type.erl, line 147)
  in call from beam_ssa_type:signatures/2 (beam_ssa_type.erl, line 141)
  in call from beam_ssa_type:opt_start/2 (beam_ssa_type.erl, line 80)
```

And
```erlang
f(_V0) ->
    try
        _V0 = ok
    catch
        _ ->
            <<
                0
             || _V0 := _ <- ok,
                (try ok of
                    _ when _V0, (_V0 andalso _V0) orelse trunc(ok) ->
                        ok
                catch
                    _ ->
                        ok
                end)
            >>
    end.
```
causes the following errror:
```
Function: '-f/1-lbc$^0/2-0-'/2
Sub pass ssa_opt_type_continue
Function: '-f/1-lbc$^0/2-0-'/2
/home/rmorisset/minimized/test408856.erl: internal error in pass beam_ssa_opt:
exception error: no match of right hand side value none
  in function  beam_ssa_type:concrete_type/2 (beam_ssa_type.erl, line 2449)
  in call from beam_ssa_type:subtract_types/2 (beam_ssa_type.erl, line 2898)
  in call from beam_ssa_type:infer_types_br_1/3 (beam_ssa_type.erl, line 2517)
  in call from beam_ssa_type:infer_types_br/4 (beam_ssa_type.erl, line 2477)
  in call from beam_ssa_type:update_successors/5 (beam_ssa_type.erl, line 1968)
  in call from beam_ssa_type:opt_bs/8 (beam_ssa_type.erl, line 529)
  in call from beam_ssa_type:do_opt_function/6 (beam_ssa_type.erl, line 501)
  in call from beam_ssa_type:opt_function/6 (beam_ssa_type.erl, line 474)
```
I don't know whether these two crashes are symptom of a single bug or separate issues.

jhogberg:
Thanks, that looks like ye olde #6599/#7342 again :(

RobinMorisset:
Ah, I had hope it was something new since it needs the PR to trigger. But I guess it might have just revealed a pre-existing bug :-/.

jhogberg:
I take that back, it was a bug in this PR. Thanks! :D

RobinMorisset:
Thanks for the fix. I checked and it solved all the crashing testcases found by erlfuzz.
Can you add the first test case above to the test suite? So that these lines in the compiler are covered and if anyone breaks them the regression will be caught.

PR #7490, opened 2023-07-11
argparse: Make formatting of argument descriptions less ambiguous

jhogberg:
Currently, the default value is rather ambiguous for choice types, being presented as:

`choice: a, b, c, d, a`

Where it's not entirely clear that the last value, `a`, is the default. This PR attempts to fix that.


max-au (approved):
Thanks!

Fun fact, originally `argparse` was not able to limit help format width, and I tried to make arguments printing as compact as I could. Sometimes too compact :-)

PR #7496, opened 2023-07-13
Kuba/ct/ct hooks order option/otp 18682

u3s:


u3s:
> Not much we can do about that now, but I'm not sure we need to make the same mistake for `ct_hook_order`. IMO it should be enough to allow that option in the `suite` callback and in test specs/command line.

I will gladly reduce supported callbacks to suite. I tried to follow `ct_hooks` assuming it is all well thought and needed. I didn't dare to think it is all _an accident_ ;-)

Also on 2nd thought, I think `ct_hooks_order` is more 'global' thing in framework than `ct_hooks`.

PR #7499, opened 2023-07-15
Fix SSH reverse search with CTRL+R

jjcarstens:
`group.erl` now sends `put_expand_no_trim` when entering a history search. `ssh_cli` does not support that message and so when entering a search when SSH is the underlying driver, the expanded result would not be displayed even though the search functionality would still technically work.

This adds support for that `put_expand_no_trim` message when using SSH

I look through the tests and not entirely sure where I would put this. I'm happy to write that if there is any direction of how others might want it tested. In the meantime, this is my quick manual setup to test it works:

**Start the Daemon**
```erlang
%% Create System host key
Priv = public_key:generate_key({rsa, 2048, 65537}).
PemEntry = public_key:pem_entry_encode(:"RSAPrivateKey", Priv).
Pem = public_key:pem_encode([PemEntry]).
ok = file:mkdir_p("/tmp/ssh"),
ok = file:write("/tmp/ssh/ssh_host_rsa_key", Pem).

%% Start SSH Daemon
Opts = [{system_dir, "/tmp/ssh"}, {pwdfun, fun (_, _, _, _) -> true end}],
application:ensure_all_started(ssh),
ssh:daemon(2222, Opts).
```

**SSH (no password)**
```
ssh localhost -p 2222
# CTRL+r to search history
```

### DIFF
--- a/lib/ssh/src/ssh_cli.erl
+++ b/lib/ssh/src/ssh_cli.erl
@@ -424,6 +424,8 @@ io_request({put_chars, Cs}, Buf, Tty, _Group) ->
     put_chars(bin_to_list(Cs), Buf, Tty);
 io_request({put_chars, unicode, Cs}, Buf, Tty, _Group) ->
     put_chars(unicode:characters_to_list(Cs,unicode), Buf, Tty);
+io_request({put_expand_no_trim, unicode, Expand}, Buf, Tty, _Group) ->
+    insert_chars(unicode:characters_to_list(Expand, unicode), Buf, Tty);
 io_request({insert_chars, Cs}, Buf, Tty, _Group) ->
     insert_chars(bin_to_list(Cs), Buf, Tty);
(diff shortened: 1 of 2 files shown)


IngelaAndin:
Thanks as it is vacation time it may take a little longer, but we will handle it.

u3s:
> I look through the tests and not entirely sure where I would put this. I'm happy to write that if there is any direction of how others might want it tested.

Can you check if some simple test could be developed around ssh_basic_SUITE:do_shell code?

jjcarstens:
Sorry! We seemed to have swapped vacation times and then I let this slip away from me until your recent comment 🤦🏼 Thanks for the help!

u3s:
change of plans. we will merge both tests in this PR (removing the dependency towards #7627). I hope to merge your PR this week.

PR #7500, opened 2023-07-15
erts: check for sys/auxv.h

bkuhls:
Fixes build with uClibc-based toolchains.


garazdawi:
Hello!

Please open a PR to https://github.com/asmjit/asmjit instead and we will get the fix next time we sync with upstream. We do not want theirs and our implementation to diverge.

bkuhls:
> Please open a PR to https://github.com/asmjit/asmjit instead

Hi, this repo has a cmake-based buildsystem and lacks configure.ac.
As I am unfamiliar with the history of both projects and their connection I would like to kindly ask what to do?

garazdawi:
I don't know much about CMAKE myself, so I'm afraid I cannot help you. If we only patch it in Erlang then we are sure to loose the patch the next time we vendor asmjit as we just copy the files from their repo.

Maybe open an issue and ask for help there? The maintainer is usually very helpful.

PR #7511, opened 2023-07-26
Clarify normal shutdown return on gen_server:call/2,3

jhogberg:


smaximov:
@jhogberg I believe this should also apply to 

<blockquote>
<dl><dd><code>_OtherTerm</code></dd><dt>
The server process exited during the call, with reason Reason. Either by returning <code>{stop,Reason,_}</code> from its <a href="https://www.erlang.org/doc/man/gen_server.html#Module:handle_call-3">Module:handle_call/3()</a> callback (without replying), by raising an exception, or due to getting an exit signal it did not trap.</dt></dl>
</blockquote>

PR #7512, opened 2023-07-26
compiler: Don't apply beam_ssa_type to unreachable functions

### LINKED ISSUE #7478: [erlc] internal error in ssa_opt_type_continue
**Describe the bug**
Running erlc on the following testcase:
```erlang
f(_V0) ->
    [ 0 ||
        begin bit_size(maybe 
            [] ?= maybe 
                0 ?= _V0,
                << 0 || _ <- []>>,
                ok
            end,
            _V0
        else
            _V0 -> _V0;
            ok -> _V0
        end),
        _V0 end,
    _ <- ok].
```
results in the following error message:
```
Sub pass ssa_opt_type_continue
Function: '-f/1-lc$^1/1-1-'/1
/home/rmorisset/minimized/test70696.erl: internal error in pass beam_ssa_opt:
exception error: no function clause matching beam_types:join([]) 
  in function  beam_ssa_type:join_arg_types/3 (beam_ssa_type.erl, line 446)
  in call from beam_ssa_type:opt_continue/4 (beam_ssa_type.erl, line 429)
  in call from beam_ssa_opt:ssa_opt_type_continue/1 (beam_ssa_opt.erl, line 437)
  in call from compile:run_sub_passes_1/3 (compile.erl, line 411)
  in call from beam_ssa_opt:phase/4 (beam_ssa_opt.erl, line 117)
  in call from beam_ssa_opt:fixpoint/6 (beam_ssa_opt.erl, line 100)
  in call from beam_ssa_opt:run_phases/3 (beam_ssa_opt.erl, line 86)
```

**Affected versions**
- master + #7474
- maint + #7448

### LINKED ISSUE #7509: Internal compiler error in ssa_opt_type_continue
**Describe the bug**

Erlc can sometimes crash with a back-trace like this:

```
Sub pass ssa_opt_type_continue
Function: '-f3/1-lc$^0/1-3-'/1
test362604.erl: internal error in pass beam_ssa_opt:
exception error: no function clause matching beam_types:join([]) 
  in function  beam_ssa_type:join_arg_types/3 (beam_ssa_type.erl, line 449)
  in call from beam_ssa_type:opt_continue/4 (beam_ssa_type.erl, line 432)
  in call from beam_ssa_opt:ssa_opt_type_continue/1 (beam_ssa_opt.erl, line 449)
  in call from beam_ssa_opt:phase/4 (beam_ssa_opt.erl, line 116)
  in call from beam_ssa_opt:fixpoint/6 (beam_ssa_opt.erl, line 99)
  in call from beam_ssa_opt:run_phases/3 (beam_ssa_opt.erl, line 85)
```

**To Reproduce**
Unless the patch below is applied, the crash appears to be dependent on the phase of the moon :) With the patch applied, the [reproducer](https://gist.github.com/frej/cb3748cea39dd04bace98d4886878206) (found with Erlfuzz, thanks @RobinMorisset ) triggers the error reliably.

```Patch
diff --git a/lib/compiler/src/beam_ssa_type.erl b/lib/compiler/src/beam_ssa_type.erl
index d492ed4b50..9d8b965bd3 100644
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -379,7 +379,7 @@ init_sig_st(StMap, FuncDb) ->
              wl=wl_defer_list(Roots, wl_new()) }.
 
 init_sig_roots(FuncDb) ->
-    [Id || Id := #func_info{exported=true} <- FuncDb].
+    [Id || Id := #func_info{exported=true} <- maps:iterator(FuncDb, ordered)].
 
 init_sig_args([Root | Roots], StMap, Acc) ->
     #opt_st{args=Args0} = map_get(Root, StMap),

```

**Affected versions**
At least `maint` (fe0571ecad3dd01609f2bd237a38816912fbd23f)

**Additional context**

As far as I can tell, `beam_ssa_type:signatures_1/2` can produce a different `#sig_st{}` and `FuncDb` depending on the order of the initial worklist in the `#sig_st{}`. This, in turn, leads to number of list-comprehension functions being eliminated from optimization when the compilation succeeds, but kept around when the crash occurs.

frej:
The `beam_ssa_type:opt_continue/4` pass requires that type information is available for all arguments or else it will crash with a back trace similar to this:

```
Sub pass ssa_opt_type_continue
Function: '-f3/1-lc$^0/1-3-'/1
test362604.erl: internal error in pass beam_ssa_opt: exception error: no function clause matching beam_types:join([])
  in function  beam_ssa_type:join_arg_types/3 (beam_ssa_type.erl, line 449)
  in call from beam_ssa_type:opt_continue/4 (beam_ssa_type.erl, line 432)
  in call from beam_ssa_opt:ssa_opt_type_continue/1 (beam_ssa_opt.erl, line 449)
  in call from beam_ssa_opt:phase/4 (beam_ssa_opt.erl, line 116)
  in call from beam_ssa_opt:fixpoint/6 (beam_ssa_opt.erl, line 99)
  in call from beam_ssa_opt:run_phases/3 (beam_ssa_opt.erl, line 85)
```

If type analysis during `beam_ssa_type:opt_start/2` concludes that a function is never called, type information for the callee will never be recorded, thus leaving no type information for the arguments in the function's entry in the function database, leading to the crash above.

Although naively unreachable functions are pruned when the list of functions is first traversed in `beam_ssa_type:opt_start/2`, it doesn't handle they case when type inference concludes that a function is never called. This patch extends `beam_ssa_type:opt_start/2` to, when all functions have been processed, prune non-exported functions for which `#func_info.arg_types` is just a list of empty dictionaries.  That a function of zero arity is always considered reachable, is harmless as it won't crash opt_continue/4.

Closes #7509
Closes #7478

### DIFF
--- a/lib/compiler/src/beam_ssa_type.erl
+++ b/lib/compiler/src/beam_ssa_type.erl
@@ -109,7 +109,39 @@ opt_start_1([Id | Ids], ArgDb, StMap0, FuncDb0, MetaCache) ->
             opt_start_1(Ids, ArgDb, StMap, FuncDb, MetaCache)
     end;
 opt_start_1([], _CommittedArgs, StMap, FuncDb, _MetaCache) ->
+    %% Although unreachable functions are pruned when the list of
+    %% functions is first traversed above, it doesn't handle the case
+    %% when type inference concludes that a function is never
+    %% called. That is, its #func_info.arg_types is just a list of
+    %% empty dictionaries. Further optimizing such an unreachable
+    %% function is wasteful and will furthermore crash opt_continue/4
+    %% as no type information is available for its
+    %% arguments. Therefore, remove all functions which are
+    %% unreachable.
+    %%
+    %% That a function of zero arity is always considered reachable,
+    %% is harmless as it won't crash opt_continue/4.
+    %%
+    remove_unreachable(maps:keys(StMap), StMap, FuncDb).
+
+remove_unreachable([Id|Ids], StMap, FuncDb) ->
+    #{ Id := #func_info{exported=Exported,arg_types=ArgTypes} } = FuncDb,
+    case Exported orelse is_reachable(ArgTypes) of
+        true ->
+            remove_unreachable(Ids, StMap, FuncDb);
+        false ->
+            remove_unreachable(Ids,
+                               maps:remove(Id, StMap), maps:remove(Id, FuncDb))
+    end;
+remove_unreachable([], StMap, FuncDb) ->
     {StMap, FuncDb}.
+
+is_reachable([X|_]) when map_size(X) == 0 ->
+    false;
+is_reachable([_|Xs]) ->
+    is_reachable(Xs);
+is_reachable([]) ->
+    true.
 
 %%
 %% The initial signature analysis is based on the paper "Practical Type
--- a/lib/compiler/test/beam_type_SUITE.erl
+++ b/lib/compiler/test/beam_type_SUITE.erl
@@ -1490,6 +1490,13 @@ will_succeed_1(_V0, _V1)
 will_succeed_1(_, _) ->
     b.
 
+%% GH-7478 and GH-7509: The beam_ssa_type pass would crash when
+%% insisting on optimizing a function the pass had concluded was
+%% unreachable.
+gh_7478(_V0) ->
+    <<0 || erlang:yield() and ([0 || _ <- _V0] =< 9223372036854775807),
+           <<>> <= ok >>.
+
 %%%
 %%% Common utilities.
 %%%


lib/compiler/src/beam_ssa_type.erl
```diff
@@ -109,8 +109,40 @@ opt_start_1([Id | Ids], ArgDb, StMap0, FuncDb0, MetaCache) ->
             opt_start_1(Ids, ArgDb, StMap, FuncDb, MetaCache)
     end;
 opt_start_1([], _CommittedArgs, StMap, FuncDb, _MetaCache) ->
+    %% Although unreachable functions are pruned when the list of
+    %% functions is first traversed above, it doesn't handle they case
```
RobinMorisset, line 113: nit: they -> the

RobinMorisset (approved):
LGTM, but I don't know that code well, so I recommend waiting for a review by @jhogberg or @bjorng.

lib/compiler/src/beam_ssa_type.erl
frej: Fixed.

frej:
Withdrawn as it masks the [real bug](https://github.com/erlang/otp/issues/7509#issuecomment-1651922199).

PR #7513, opened 2023-07-26
inets: httpd avoid function_clause during startup

u3s:
- avoid function_clause httpd_request:body_data/2
- happening when httpd is stopped during TLS negotiation
- in such cases Manager process is killed and then continue_init fails

### DIFF
--- a/lib/inets/src/http_server/httpd_request_handler.erl
+++ b/lib/inets/src/http_server/httpd_request_handler.erl
@@ -137,34 +137,42 @@ continue_init(Manager, ConfigDB, SocketType, Socket, Peername, Sockname,
                socket_type = SocketType,
                socket = Socket,
                init_data = InitData},
-    
-    MaxHeaderSize = max_header_size(ConfigDB), 
-    MaxURISize    = max_uri_size(ConfigDB), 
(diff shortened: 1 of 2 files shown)


lib/inets/src/http_server/httpd_request_handler.erl
```diff
@@ -137,34 +137,42 @@ continue_init(Manager, ConfigDB, SocketType, Socket, Peername, Sockname,
...
-                   chunk                   = chunk_start(MaxChunk)},
-    setopts(Socket, SocketType, [binary, {packet, 0}, {active, once}]),
-    NewState =  data_receive_counter(activate_request_timeout(State), httpd_util:lookup(ConfigDB, minimum_bytes_per_second, false)),
-     gen_server:enter_loop(?MODULE, [], NewState).
 
+    {Result, Status} = httpd_manager:new_connection(Manager),
+    case Result of
+        error ->
+            httpd_util:error_log(ConfigDB,
+                                 httpd_logger:error_report('TLS', Status,
```
RaimoNiskanen, line 152: It is not known at this point that it is a `'TLS'` socket.  The other atom used in this module, in this argument position would be `internal`.

PR #7515, opened 2023-07-26
Gracefully handle bad validity dates in public_key:pkix_path_validation

andreasanta:
If a certificate contains an invalid date in the `validity` attribute, the function `public_key:pkix_path_validation` raises a `function_clause` error due to the invocation of `calendar:date_to_gregorian_days` with an invalid date.

In the attached unit test, the leaf certificate contains the invalid date string "19000100000000Z", and this would entirely fail the validation of the certificate chain.

In this PR, I am handling this edge case by delegating the decision to `verify_fun` and thus the caller. If `verify_fun` traps the error and decides to continue, validation is not interrupted, otherwise the `function_clause` error is raised.

This is an example error reproduced locally with the attached certificate:

```
{function_clause,
                     [{calendar,date_to_gregorian_days,
                          [1900,1,0],
                          [{file,"calendar.erl"},{line,132}]},
                      {calendar,datetime_to_gregorian_seconds,1,
                          [{file,"calendar.erl"},{line,155}]},
                      {pubkey_cert,validate_time,3,
                          [{file,"pubkey_cert.erl"},{line,147}]},
```

NOTE: I am aware invalid dates should never be passed to this function, but I am dealing with a hardware implementation that cannot be easily changed, thus I have to permit a more lenient parsing and recovery from error. Also, I do not believe a `function_clause` error is the right response, since it's too generic and cannot be handled accordingly or interpreted by the caller.

CC @IngelaAndin @potatosalad for visibility


lib/public_key/src/pubkey_cert.erl
```diff
@@ -144,14 +144,23 @@ validate_time(OtpCert, UserState, VerifyFun) ->
...
-    NotBefore = time_str_2_gregorian_sec(notBefore, NotBeforeStr),
-    NotAfter = time_str_2_gregorian_sec(notAfter, NotAfterStr),
 
-    case ((NotBefore =< Now) and (Now =< NotAfter)) of
-	true ->
-	    UserState;
-	false ->
-	    verify_fun(OtpCert, {bad_cert, cert_expired}, UserState, VerifyFun)
+    % Catch invalid date format errors in the `validity` field of the certificate
+    try
```
potatosalad, line 149: This `try` statement seems wider in scope than it needs to be to me.  I think it would be better to isolate the parsing of `notBefore` and `notAfter` so the `try` statement doesn't also wrap the `verify_fun` call for `{bad_cert, cert_expired}`, which itself could raise a `function_clause` error.

lib/public_key/src/pubkey_cert.erl
```diff
@@ -144,14 +144,23 @@ validate_time(OtpCert, UserState, VerifyFun) ->
...
+            true ->
+                UserState;
+            false ->
+                verify_fun(OtpCert, {bad_cert, cert_expired}, UserState, VerifyFun)
+        end
+
+    % "error:function_clause" is thrown by time_str_2_gregorian_sec if the date format is not valid
+    % verify_fun only throws exceptions
+    catch error:function_clause ->
+        verify_fun(OtpCert, {bad_cert, invalid_date}, UserState, VerifyFun)
```
potatosalad, line 163: Also not sure if `invalid_date` is descriptive enough here?  Is it worth having something as specific as `invalid_not_before` and `invalid_not_after`, or maybe something like `invalid_cert_expiration`?

lib/public_key/src/pubkey_cert.erl
andreasanta: Changed into a clearer message.

andreasanta: Reduced the scope of the try/catch and carried over the state.

lib/public_key/src/pubkey_cert.erl
andreasanta: Changed the error message into a more meaningful one.

IngelaAndin:
I am not very fond of these kind of workarounds, but we have been forced to be pragmatic in other cases too, and it seems harmless enough solution. It is vacation time now so I would like to wait a little for final inclusion decision so that we can have a team review. And we would also need to add the new error to the documentation.

andreasanta:
Thanks @IngelaAndin - To be honest I would have also gladly glossed over this edge case, but we are getting plenty of hardware certificates not abiding to the RFCs and we do not have many alternatives.

Please let me know if you would like
me to add the documentation commit to the PR as well.

IngelaAndin:
Yes we want you to add a documentation commit where the invalid_validity_dates are added to the possible bad_cert_reason()  type.

IngelaAndin:
Also if you base your patch on maint instead of master we can include it for OTP-26.1  instead of waiting for OTP-27. This I think should be ok as it does not  really change any API behavior (only adds an error reason)  and it is unlikely to change any existing verify_fun behavior but rather expands the workaround possibilities for the verify_fun, which is one of its uses although the main purpose is to be able to handle application specific extensions.  But that is up to you.

IngelaAndin:
@andreasanta  and adding the bad_cert_reason() to the documentation is actually  adding it to the spec in public_key.erl

lib/public_key/src/pubkey_cert.erl
```diff
@@ -144,15 +144,31 @@ validate_time(OtpCert, UserState, VerifyFun) ->
...
-    case ((NotBefore =< Now) and (Now =< NotAfter)) of
-	true ->
-	    UserState;
-	false ->
-	    verify_fun(OtpCert, {bad_cert, cert_expired}, UserState, VerifyFun)
+    % Catch invalid date format errors in the `validity` field of the certificate
+    NotExpired = try
+        NotBefore = time_str_2_gregorian_sec(notBefore, NotBeforeStr),
+        NotAfter = time_str_2_gregorian_sec(notAfter, NotAfterStr),
+
```
u3s, line 152: Please remove empty line. I don't think it improves readability of code.

lib/public_key/src/pubkey_cert.erl
```diff
@@ -144,15 +144,31 @@ validate_time(OtpCert, UserState, VerifyFun) ->
...
+    catch error:function_clause ->
+        verify_fun(OtpCert, {bad_cert, invalid_validity_dates}, UserState, VerifyFun)
+    end,
+
+    case NotExpired of
+        true ->
+            UserState;
+        false ->
+            verify_fun(OtpCert, {bad_cert, cert_expired}, UserState, VerifyFun);
+
```
u3s, line 165: Please remove empty line. I don't think it improves readability of code.

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -926,7 +929,47 @@ pkix_path_validation_root_expired(Config) when is_list(Config) ->
...
+    % First test exception throwing without validation function
+    try
+        public_key:pkix_path_validation(Root, CertificateChain, [])
+    catch
+        % We expect this function to throw if the date is wrong
+        error:function_clause ->
+            ok;
+        _ ->
+            throw("Expected function_clause error on bad date")
+    end,
```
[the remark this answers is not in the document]
andreasanta: This is the current behaviour if no verify fun is supplied. Why would you like to change it?

andreasanta:
@IngelaAndin All done. Addressed comments, rebased on maint and added docs. Please confirm if ok.

lib/public_key/test/public_key_SUITE.erl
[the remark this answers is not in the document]
dgud: Since the test will be merged along with your changes there is no reason to test the behavior in the old code.

You shall test the new functionality, i.e. that {error, {bad_cert, invalid_validity_dates}} is returned if no verify_fun is provided.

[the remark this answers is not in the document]
u3s: @andreasanta what about this one?

lib/public_key/src/pubkey_cert.erl
```diff
@@ -144,15 +144,29 @@ validate_time(OtpCert, UserState, VerifyFun) ->
...
+    case NotExpired of
+        true ->
+            UserState;
+        false ->
+            verify_fun(OtpCert, {bad_cert, cert_expired}, UserState, VerifyFun);
+            % This will be called in case an exception is caught when processing dates
+            % and verify_fun alters the state, so we keep the modified state along the way.
+        NewState ->
+            NewState
```
u3s, line 167: I have doubts about this catch everything clause.
Previous code didn't work that way.

Is there a reason for it?

andreasanta: No reason, preventing crash in case of unexpected value. I can remove it if necessary.

lib/public_key/test/public_key_SUITE.erl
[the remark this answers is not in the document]
andreasanta: @dgud This diff does not alter the old functionality if you don't provide `verify_fun`. This is why I am leaving the testcase. In order to preserve backwards compatibility, I am still raising an error if no `verify_fun` is specified.

@u3s Sorry maybe I did not explain my diff correctly. The previous behaviour has not been altered on purpose. You can optionally alter if if you pass a `verify_fun` that intercepts the validation error.  This testcase exists to ensure that the previous behaviour - i.e. throwing an error upon failed validation - still happens if you don't specify a `verify_fun`.

Please let me know if you'd like a different approach.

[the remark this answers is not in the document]
dgud: You have altered the functionality if no verify_fun is used and you fail to test it.

lib/public_key/src/pubkey_cert.erl
IngelaAndin: I actually think that the last case is the case that the catch verify_fun returns a state.  I think the following code could benefit readablility:

``validate_time(OtpCert, UserState, VerifyFun) ->
    TBSCert = OtpCert#'OTPCertificate'.tbsCertificate,
    {'Validity', NotBeforeStr, NotAfterStr} 
	= TBSCert#'OTPTBSCertificate'.validity,
   Now = calendar:datetime_to_gregorian_seconds(calendar:universal_time()),

  case valid_dates(NotBeforeStr, NotAfterStr}) of
   {true, {NotBefore, NotAfter}} ->
         case ((NotBefore =< Now) and (Now =< NotAfter)) of
	    true ->
	      UserState;
	   false ->
	      verify_fun(OtpCert, {bad_cert, cert_expired}, UserState, VerifyFun)
           end;
   {false, Error} ->
        verify_fun(OtpCert, {bad_cert, Error}, UserState, VerifyFun)
  end.

valid_dates(NotBeforeStr, NotAfterStr) ->
  try
    NotBefore = time_str_2_gregorian_sec(notBefore, NotBeforeStr),
    NotAfter = time_str_2_gregorian_sec(notAfter, NotAfterStr),
    {true, {NotBefore, NotAfter}}
catch
    error:function_clause ->
    {false, invalid_validity_dates}
end.

lib/public_key/test/public_key_SUITE.erl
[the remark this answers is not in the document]
IngelaAndin: I do not see a reason to preserve the crash behavior for something that should not really happen.  I think this PR is about  adding a workaround | graceful handling of that it sometimes does due to mistakes that are out of users control.

IngelaAndin:
@andreasanta if you include my suggested code enhancement I think this will be ready for merge.

lib/public_key/test/public_key_SUITE.erl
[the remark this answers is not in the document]
andreasanta: @IngelaAndin Refactored as suggested, plus some other improvements, to improve readability.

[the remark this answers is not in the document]
andreasanta: > You have altered the functionality if no verify_fun is used and you fail to test it.

Sorry I do not understand it. I have written two tests, one that lets the code crash as expected, the other prevents the crash by supplying a custom `verify_fun`. What did I miss?

andreasanta:
> @andreasanta if you include my suggested code enhancement I think this will be ready for merge.

Done, and also refactored a bit further to improve readability.

andreasanta:
Apologies, after the refactor unit tests are failing because the error,function_clause is not bubbled up correctly. I am working on it.

IngelaAndin:
Sure, I am in Berlin for Code Beam. Will look at your changes next week.

andreasanta:
> Sure, I am in Berlin for Code Beam. Will look at your changes next week.

Thanks! I am trying to make "erlang:error" work inside a test case because it fails the entire testcase instead of being bubbled up to the caller. Will keep you posted.

lib/public_key/src/pubkey_cert.erl
```diff
@@ -26,23 +26,23 @@
 
 -export([init_validation_state/3, 
          prepare_for_next_cert/2,
- 	 validate_time/3, 
+         validate_time/3,
```
andreasanta, line 29: I don't know what is going on with whitespaces, but now they are all spaces and not a mix between tabs and spaces.

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -190,9 +194,9 @@ groups() ->
 init_per_suite(Config) ->
     application:stop(crypto),
     try crypto:start() of
-	ok ->
-	    application:start(asn1),
-	    Config
+        ok ->
```
andreasanta, line 197: These are also all spaces now.

lib/public_key/src/pubkey_cert.erl
IngelaAndin: This is probably due to it not being enforced to be one way a long time ago.  So code base slowly becomes more adherent.

lib/public_key/src/pubkey_cert.erl
```diff
@@ -140,19 +140,50 @@ prepare_for_next_cert(OtpCert, ValidationState = #path_validation_state{
 %% current time.
 %%--------------------------------------------------------------------	  
 validate_time(OtpCert, UserState, VerifyFun) ->
+
+    % Parse and check validity of the certificate dates, and if it fails, invoke `verify_fun` to
+    % hand over control to the caller in order to decide what to do.
+    case parse_and_check_validity_dates(OtpCert) of
+
```
IngelaAndin, line 147: @adresanta I do not think we need these empty lines between the case clauses. No other cases in the application are written like that. Otherwise I think this is now looking good.

andreasanta:
> We should adhere to coding style of the application. Otherwise I think it looks good :)

Done away with the spacing within case clauses, thank you.

IngelaAndin:
So last testing round, should be able to merge tomorrow.

IngelaAndin (approved):

IngelaAndin:
@andreasanta some other changes has created a merge conflict, could you please rebase your branch on latest maint and resolve it so that we can use the merge button,

lib/public_key/src/pubkey_cert.erl
```diff
@@ -140,19 +140,47 @@ prepare_for_next_cert(OtpCert, ValidationState = #path_validation_state{
 %% current time.
 %%--------------------------------------------------------------------	  
 validate_time(OtpCert, UserState, VerifyFun) ->
+
```
u3s, line 143: please stick to convention in module and don't add empty line after `->`

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -934,6 +938,39 @@ pkix_path_validation_root_expired(Config) when is_list(Config) ->
     Peer = proplists:get_value(cert, Conf),
     {error, {bad_cert, cert_expired}} = public_key:pkix_path_validation(Root, [ICA, Peer], []).
     
+
+pkix_path_validation_bad_date() ->
+    [{doc, "Ensure bad date formats in `validity` are handled gracefully by verify fun"}].
+pkix_path_validation_bad_date(Config) when is_list(Config) ->
```
u3s, line 944: please stick to convention in module and don't add empty line after `->`

u3s (approved):

IngelaAndin:
@andreasanta  ping

andreasanta:
Rebased and adapted to style reqs. Thank you.

IngelaAndin:
@andreasanta  Thanks, need one more test day (night) for sanity check of the rebase and small adjustments and then we should finally be able to merge.

PR #7518, opened 2023-07-31
Fix crash when compiling is_list/1 call

### LINKED ISSUE #7504: [erlc] internal consistency check: uninitialized_reg
**Describe the bug**
Running erlc on the following:
```erlang
f([]) ->
    ok;
f(_V0) ->
    ((_V1 = erlang:is_list(_V0)) orelse binary_to_integer(list_to_bitstring([]))) / 0,
    _V1.
```
fails with
```
test493867:1: function f/1+16:
  Instruction: {call_ext,1,{extfunc,erlang,list_to_bitstring,1}}
  Error:       {uninitialized_reg,{y,0}}:
```

**Affected versions**
- master
- master + #7489 

**Additional context**
I could not reproduce this with an old version of OTP 25, so I suspect the bug was introduced in the last year.

bjorng:
Fixes #7504

### DIFF
--- a/lib/compiler/src/beam_ssa_codegen.erl
+++ b/lib/compiler/src/beam_ssa_codegen.erl
@@ -1270,9 +1270,22 @@ cg_block([#cg_set{op=is_tagged_tuple,anno=Anno,dst=Bool,args=Args0}], {Bool,Fail
             [Src,{integer,Arity},Tag] = typed_args(Args0, Anno, St),
             {[{test,is_tagged_tuple,ensure_label(Fail, St),[Src,Arity,Tag]}],St}
     end;
-cg_block([#cg_set{op=is_nonempty_list,dst=Bool,args=Args0}], {Bool,Fail}, St) ->
+cg_block([#cg_set{op=is_nonempty_list,dst=Bool0,args=Args0}=Set], {Bool0,Fail0}, St) ->
+    Fail = ensure_label(Fail0, St),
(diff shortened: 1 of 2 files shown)


lib/compiler/src/beam_ssa_codegen.erl
```diff
@@ -1270,9 +1270,22 @@ cg_block([#cg_set{op=is_tagged_tuple,anno=Anno,dst=Bool,args=Args0}], {Bool,Fail
...
-    {[{test,is_nonempty_list,ensure_label(Fail, St),Args}],St};
+    case beam_args([Bool0|Args0], St) of
+        [{z,0}|Args] ->
+            {[{test,is_nonempty_list,Fail,Args}],St};
+        [Dst|Args] ->
+            %% This instruction was a call to is_list/1, which was
+            %% rewritten to an is_nonempty_list test by
+            %% beam_ssa_type. BEAM has no is_nonempty_list instruction
+            %% that will return a boolean, so we must revert it to an
+            %% is_list/1 call.
```
jhogberg, line 1284: Given the relative rarity of needing boolean values from type tests, I wonder if it wouldn't be better to convert them at an early stage from `A = is_xyz(X)` to something like `A = if is_xyz(X) -> true; not is_xyz(X) -> false end` (it would also simplify some of the work I'm doing :smile:).

jhogberg (approved):

PR #7522, opened 2023-08-02
ssl: Handle that inet:getopts may return {ok, []}

IngelaAndin:
If the standard C library call getsockopt(2) returns an error for an option - then no return value is created for that option. There is not check about the socket's state in the call chain; if the inet_drv port is still open, then the getsockopt(2) call is executed.
So it should be possible that inet:getsockopt(Port, [sndbuf]) (or recbuf), for a Port that is not dead yet but the underlying socket is closed, may return {ok,[]}.

Closes 7506


IngelaAndin:
@raimo It also visualizes that  inet:getopts might return an empty list which I think is not completely  obvious from the documentation.

PR #7527, opened 2023-08-02
compiler: Fix crash in beam_ssa_private_append

frej:
The beam_ssa_private_append pass can crash, if it, during initial value tracking, ends up in operations which do not create bit strings. This can happen as the initial value tracking in beam_ssa_private_append doesn't consider types. As the decision to apply the private append transform is using type information, tracking values into not type-compatible execution paths is harmless.

This patch changes the beam_ssa_private_append pass to simply stop tracking a value when it is known that the type is not compatible.

You run Erlfuzz over multiple week-ends without finding any issues, as soon as your branch gets merged, Erlfuzz delivers overnight :)


jhogberg:
Merged, thanks for the PR! :)

Issue #7432, opened 2023-06-22
Undesired(?) slow down in ssa_opt_alias


josevalim:
**Describe the bug**

Take the Erlang file from this gist: https://gist.github.com/josevalim/c59e1d706daf6e863f6f2745a04fb810

Now compile it with `time erlc +time Elixir.HelloTest.erl`. In my machine, it takes 0.430s and `ssa_opt_alias` reports:

    ssa_opt_alias              :      0.111 s  67 %

Now search for `whatever:a(` (on lines 150 and 183) and replace it simply by `a(` (i.e. convert the remote call into a local call). It runs much faster in 0.270s where:

    ssa_opt_alias              :      0.002 s  21 %

This slow down does not happen on Erlang/OTP 25. From an external glance, I wouldn't expect swapping a local call by a remote one to cause such slow down but I can be wrong.

Also note the larger the code contents, the slower it gets. :)

**Affected versions**

Erlang/OTP 26.

**Additional context**

Originally reported here: https://github.com/elixir-lang/elixir/issues/12696

josevalim:
Here is another file which went from a dozen seconds to compile in Erlang/OTP 25 to minutes in Erlang/OTP 26: https://gist.github.com/josevalim/94bfd65e8eaf892ed700df349838796a 

It is an Elixir file because I suspect it is the exact the same issue. I will edit this comment once profiling finishes.

frej:
@jhogberg, I missed this the first time around. I will take a look.

frej:
> From an external glance, I wouldn't expect swapping a local call by a remote one to cause such slow down but I can be wrong.

@josevalim, that's not strange, the alias analysis pass is fairly costly. When you change `whatever:a(` to `a(`, in the first example, the compiler is smart enough to notice that anything after the first `a(` will be unreachable (as `noop` doesn't match `{{ok, _subbaz@1}, [_event1@1]}`) and will therefore just remove it as dead code long before it reaches the alias analysis. If you run `erlc -S` you'll see that the fast variant only produces around a 1/6 of the code.

You can disable the alias analysis pass by running `erlc` with `+no_ssa_opt_alias`. On my machine the compile time goes from 0.8s to 0.5s. I'll investigate further and see what I can do to speed things up.

josevalim:
@frej that makes perfect sense. We were just short-circuiting part of the analysis then, thanks!

josevalim:
[...]

PR #7528, opened 2023-08-02
Speed up alias analysis 

frej:
Combined with improved kill-set calculation (85454714956f54679ae75ef49d10d7992e31b9e2) and an improved data
structure for describing the alias status of variables (c389665cc991fb8a9d123fe34dc95c12bdf1263f), this patch series provides a substantial reduction of the time required for alias analysis. For the set of modules compiled by `scripts/diffable` the time spent in alias analysis is reduced by approximately 55%. For the [example](https://github.com/erlang/otp/issues/7432) in Issue #7432, provided by José Valim, which has a large number of variables, the reduction is even more dramatic. The time spent in the alias analysis pass is reduced by 97%.


frej:
Fixed Dialyzer error and rebase to pick up CI-fixes.

jhogberg:
Merged, thanks for the PR!

Issue #7524, opened 2023-08-02
OTP26: gen_statem:start_link is not linked properly with process during init()


attah:
**Describe the bug**
If init() in my gen_statem returns {stop, ...} or even exit()s, the caller of start_link is not notified.
This works on OTP25 but not OTP26(.0.2).

**To Reproduce**
```
-module(init_error_test).
-behaviour(gen_statem).
-export([test/0, callback_mode/0, init/1]).

callback_mode() ->
    state_functions.

init({}) ->
    io:format("Hello from init/1 @~p ~p~n",
              [self(), erlang:process_info(self(), links)]),
    % exit(schmerror),
    {stop, schmerror}.

test() ->
    io:format("test/0 pid is: ~p~n", [self()]),
    process_flag(trap_exit, true),
    {error, schmerror} =
        gen_statem:start_link({local, my_statem}, ?MODULE, {}, []),
    receive
        {'EXIT', _, Err} ->
            io:format("Got EXIT: ~p~n", [Err])
    after 1234 ->
        io:format("No EXIT for you!~n")
    end.
```

**Expected behavior**
'EXIT' is received.

**Affected versions**
OTP26

**Additional context**
OTP25:
```
test/0 pid is: <0.90.0>
Hello from init/1 @<0.130.0> {links,[<0.90.0>]}
Got EXIT: schmerror
=CRASH REPORT==== 2-Aug-2023::12:43:46.511362 ===
  crasher:
    initial call: init_error_test:init/1
    pid: <0.130.0>
    registered_name: []
    exception exit: schmerror
      in function  gen_statem:init_result/8 (gen_statem.erl, line 1041)
    ancestors: [<0.90.0>]
    message_queue_len: 0
    messages: []
    links: [<0.90.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 376
    stack_size: 28
    reductions: 175
  neighbours:

ok

```

OTP26:
```
test/0 pid is: <0.86.0>
Hello from init/1 @<0.93.0> {links,[<0.86.0>]}
=CRASH REPORT==== 2-Aug-2023::12:43:38.687024 ===
  crasher:
    initial call: init_error_test:init/1
    pid: <0.93.0>
    registered_name: []
    exception exit: schmerror
      in function  gen_statem:init_result/8 (gen_statem.erl, line 1005)
    ancestors: [<0.86.0>,<0.85.0>,<0.71.0>,<0.66.0>,<0.70.0>,<0.65.0>,
[...]

PR #7529, opened 2023-08-02
Clarify `gen_statem:start_link/3,4` semantics

RaimoNiskanen:
Solves #7524

### DIFF
--- a/lib/stdlib/doc/src/gen_statem.xml
+++ b/lib/stdlib/doc/src/gen_statem.xml
@@ -588,8 +588,10 @@ handle_event(_, _, State, Data) ->
       <name name="start_ret"/>
       <desc>
 	<p>
-	  Return value from the <c>start()</c> and <c>start_link()</c> functions,
-	  for example, <seemfa marker="#start_link/3"><c>start_link/3</c></seemfa>.
+	  Return value from the
+          <seemfa marker="#start/3"><c>start/3,4</c></seemfa> and
+          <seemfa marker="#start_link/3"><c>start_link/3,4</c></seemfa>
+          functions.
 	</p>
       </desc>
     </datatype>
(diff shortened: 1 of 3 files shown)


lib/stdlib/doc/src/gen_statem.xml
```diff
@@ -2472,31 +2472,41 @@ handle_event(_, _, State, Data) ->
...
+	  <seetype marker="#init_result"><c>{error,Reason}</c></seetype>
+	  the process is terminated and this function returns
+	  <seetype marker="#start_ret"><c>{error,Reason}</c></seetype>.
+	  If <c>Module:init/1</c> returns
+	  <seetype marker="#init_result"><c>ignore</c></seetype>
+	  the process is terminated and this function returns
+	  <seetype marker="#start_ret"><c>ignore</c></seetype>.
+        </p>
+        <p>
+          The process exits with <c>Reason</c> unless
```
attah, line 2494: Could do with re-asserting that what "the" process is here. At least my reference had expired once i read this far.

lib/stdlib/doc/src/gen_statem.xml
```diff
@@ -2472,31 +2472,41 @@ handle_event(_, _, State, Data) ->
...
+          <seemfa marker="#Module:init/1"><c>Module:init/1</c></seemfa>
+          returns <c>ignore</c> or <c>{error,_}</c>,
+          then the exit reason is <c>normal</c>.
+        </p>
+        <p>
+          If <c>start_link/3,4</c> returns <c>ignore</c> or <c>{error,_}</c>,
```
attah, line 2500: Would it be useful to point to the change notice from here or have a "since OTP-26"?

attah (commented):
Definite improvement. Two minor suggestions though.

lib/stdlib/doc/src/gen_statem.xml
RaimoNiskanen: Absolutely.  I will even add a warning note.

kikofernandez (approved):

PR #7534, opened 2023-08-03
Export documented and exposed type beam_lib:beam/0

paulo-ferraz-oliveira:
From https://www.erlang.org/doc/man/beam_lib.html#type-beam, used e.g. by [`katana-code`](https://github.com/inaka/katana-code/blob/main/src/ktn_code.erl#L37C22-L37C37).


jhogberg:
Merged, thanks for the PR!

PR #7535, opened 2023-08-03
Export documented and exposed type erl_syntax:annotation_or_location/0

paulo-ferraz-oliveira:
Exposed (and documented) at https://www.erlang.org/doc/man/erl_syntax#data-types.


jhogberg:
Merged, thanks for the PR!

PR #7536, opened 2023-08-03
Kuba/inets/adjust error type/otp 18704

u3s:


lib/inets/src/http_server/httpd_request_handler.erl
```diff
@@ -149,7 +149,7 @@ continue_init(Manager, ConfigDB, SocketType, Socket, Peername, Sockname,
     case Result of
         error ->
             httpd_util:error_log(ConfigDB,
-                                 httpd_logger:error_report('TLS', Status,
+                                 httpd_logger:error_report(internal, Status,
```
IngelaAndin, line 152: I do not think we should make it internal, just not hardcode TLS. You should use SocketType to determine if it should be TLS or TCP. See httd_logger.erl

lib/inets/src/http_server/mod_esi.erl
```diff
@@ -220,47 +220,47 @@ erl(#mod{method = "POST", entity_body = Body} = ModData, ESIBody, Modules) ->
...
+                false ->
+                    Error403 =
+                        [{status,
+                          {403, ModData#mod.request_uri,
+                           ?NICE("Client not authorized to evaluate: " ++ ESIBody)}} |
+                         ModData#mod.data],
+                    {proceed, Error403}
+            end;
+        _ ->
+            ConversionResult
```
IngelaAndin, line 263: Sorry but I do not think this makes it more readable,  just unreadable in an other way! I think we should keep two function clauses and make a new function for line 245 to 263.  As we discussed earlier it was probably written recursively after formatting to avoid coming up with a new function name.  We could call the new function do_genraret_webpage to wok around the problem.

u3s: please check new proposal based on discussion

lib/inets/src/http_server/httpd_logger.erl
```diff
@@ -46,11 +46,11 @@ error_report(Protocol, Reason, #mod{init_data = #init_data{peername = PeerName,
...
     case Protocol of
         'HTTP' ->
-            Report1#{transport => transport_type(Type)}; 
+            Report1#{transport => transport_type(Type)};
         _ ->
             Report1
-    end. 
-    
+    end.
+
```
IngelaAndin, line 53: This is only indentation whitespace changes!?

IngelaAndin (approved):

lib/inets/src/http_server/httpd_logger.erl
u3s: yes. I tried to smuggle small whitespace fixes.

PR #7537, opened 2023-08-03
Fix/httpc stand alone cookies

MarkoMin:
I spotted that `stand_alone` `httpc` doesn't track session cookies (at least when started without `data_dir`). I went with PR rather than issue because I included testcase `cookie_stand_alone` (based on `cookie` and `cookie_profile` testcases) that proves the claim. I spent few hours trying to work out why this issue happens, but with little success. Hoping to get some advice from OTP team before going further. I made sure that `cookies` are enabled and checked that `<profile>__handler_db`, `<profile>__session_db` and `<profile>__session_cookie_db` ETS tables exist at runtime. session_db seems to be working, but session_cookie always stays empty. Set-cookie headers are 100% correct because it works when `httpc` is started under `inets` supervision tree.

In the first commit, changes to the `handle_http_msg/3` and `check_cookie/1` must be made because if `ct:fail/1` gets called from there it just blocks the testcase until timetrap triggered so I changed the mechanism to return HTTP 401 when there is no cookie.

Second commit contains a small improvements which I think should be done for tests to work as intended.

### DIFF
--- a/lib/inets/test/httpc_SUITE.erl
+++ b/lib/inets/test/httpc_SUITE.erl
@@ -155,6 +155,7 @@ only_simulated() ->
     [
      cookie,
      cookie_profile,
+     cookie_stand_alone,
      empty_set_cookie,
      invalid_set_cookie,
      trace,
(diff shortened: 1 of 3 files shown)


IngelaAndin:
The stand alone mode was a compromise made a long ago for allowing a more lightweight client for "get one URI wget like" situations.  I think this was mistake, and we should have put more time in improving the client design instead. I still think that! I do not know when this might be prioritized,  but we will try to give this some more attention when everyone is back from summer vacation.  It could be a start!

IngelaAndin:
I am saying that stand_alone mode is a ugly hack that was made for allowing "lightweight" one time request-responses instead of investing time in making more general improvements. And we would rather deprecate the stand_alone mode then try to improve it.  Yes it was a bad decision we made back then, and I am sorry for the confusion it might cause, that said I do not think we should make it worse!

PR #7543, opened 2023-08-08
Kuba/ssl/client ocsp/otp 18606

u3s:


lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -153,27 +156,80 @@ get_nonce_value([#'Extension'{
...
+        %%      Section 4.2.2.2) who holds a specially marked certificate
+        %%      issued directly by the CA, indicating that the responder may
+        %%      issue OCSP responses for that CA (id-kp-OCSPSigning)
+        fun() -> public_key:pkix_is_issuer(Cert, IssuerCert) andalso
+                                 designated_for_ocsp_signing(Cert) end,
+    Case3 =
+        %% a Trusted Responder whose public key is trusted by the requestor
+        fun() ->
+                Pred = fun(#cert{otp = CACert}) -> IssuerCert == CACert end,
+                lists:any(Pred, public_key:cacerts_get())
```
IngelaAndin, line 222: We can not use cacerts_get/0 here, we can not assume who the user wants to trust. cacerts_get/0 is typically who webbrowser would trust.  It needs to be configurable.  The ssl application would default to its cacerts option. In the long run we might want to have a special stapling trusted option, the one we agreed that we could skip for now to reduce complexity.  Also I think verify responder cert is a confusing name as you do not verify the signature in this function, this is done later this is still about finding the potential issuer.  This code path may also be problematic if there exist more than one potential issuer in the set of trusted certs!

IngelaAndin (changes requested):

lib/public_key/doc/src/public_key.xml
```diff
@@ -723,6 +723,16 @@ fun(#'DistributionPoint'{}, #'CertificateList'{},
...
+  <func>
+    <name name="pkix_ocsp_validate" arity="4" since="OTP 26.1"/>
+    <fsummary>Validate OCSP response.</fsummary>
+    <desc>
+      <p>Perform OCSP response validation according to RFC
+      6960. Returns 'valid' when OCSP response is successfully
+      validated and {bad_cert, Reason} otherwise.</p>
+    </desc>
+  </func>
+
```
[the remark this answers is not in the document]
u3s: fixed in https://github.com/erlang/otp/pull/7543/commits/3fdb8d842e21eb1289877ea28940c96e6b2054fc

lib/ssl/src/ssl.erl
```diff
@@ -451,9 +449,7 @@
 -type max_fragment_length()      :: undefined | 512 | 1024 | 2048 | 4096.
 -type fallback()                 :: boolean().
 -type ssl_imp()                  :: new | old.
-%% -type ocsp_stapling()            :: boolean().
-%% -type ocsp_responder_certs()     :: [public_key:der_encoded()].
-%% -type ocsp_nonce()               :: boolean().
+-type stapling()                 :: boolean() | map().
```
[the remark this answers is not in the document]
u3s: fixed in https://github.com/erlang/otp/pull/7543/commits/5715f752973d1c9bd78ac23b556f5d74ddcc3853

lib/public_key/test/public_key_SUITE.erl
```diff
@@ -1338,6 +1342,98 @@ gen_ec_param_char_2_field(Config) when is_list(Config) ->
     Datadir = proplists:get_value(data_dir, Config),
     do_gen_ec_param(filename:join(Datadir, "ec_key_param1.pem")).
 
+%%--------------------------------------------------------------------
+ocsp_extensions() ->
+    [{doc, "Check OCSP extensions"}].
+ocsp_extensions(_Config) ->
+    Nonce = <<4,8,66,243,220,236,16,118,51,215>>,
+    ExpectedExtentions = [{'Extension',
+      {1,3,6,1,5,5,7,48,1,2},
```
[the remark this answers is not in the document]
u3s: fixed in https://github.com/erlang/otp/pull/7543/commits/95f08e0bda06f121bb825ffdf2133d718ef2f27a

lib/ssl/src/ssl.erl
```diff
@@ -2503,10 +2496,10 @@ assert_server_only(client, Bool, Option) ->
...
 assert_server_only(_, _, _) ->
     ok.
-assert_client_only(server, Bool, Option) ->
-    role_error(Bool, client_only, Option);
-assert_client_only(_, _, _) ->
-    ok.
+%% assert_client_only(server, Bool, Option) ->
+%%     role_error(Bool, client_only, Option);
+%% assert_client_only(_, _, _) ->
+%%     ok.
```
dgud, line 2502: Remove

dgud (commented):
Should we handle "-type combined_cert()        :: #cert{}."  in the certificate(s) in the arguments?
To reduce the decode/encode der roundtrips that is done?

lib/ssl/src/ssl.erl
u3s: fixed in [05288e3](https://github.com/erlang/otp/pull/7543/commits/05288e3f32668ad581571e0a1c3b9e0d5943ff09)

lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -153,27 +156,78 @@ get_nonce_value([#'Extension'{
...
+        end,
+
+    case lists:any(fun(E) -> E() end, [Case1, Case2, Case3]) of
         true ->
-            {ok, Cert};
-        false ->
-            find_responder_cert(ResponderID, TCerts)
+            true;
+        _ ->
```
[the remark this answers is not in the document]
u3s: fixed in [05288e3](https://github.com/erlang/otp/pull/7543/commits/05288e3f32668ad581571e0a1c3b9e0d5943ff09)

lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -67,15 +80,15 @@ find_single_response(Cert, IssuerCert, SingleResponseList) ->
...
     match_single_response(IssuerName, IssuerKey, SerialNum, SingleResponseList).
 
--spec ocsp_status({atom(), term()}) -> atom() | {atom(), {atom(), term()}}.
-ocsp_status({good, _}) ->
-    valid;
-ocsp_status({unknown, Reason}) ->
-    {bad_cert, {revocation_status_undetermined, Reason}};
-ocsp_status({revoked, Reason}) ->
-    {bad_cert, {revoked, Reason}}.
+-spec status({atom(), term()}) -> atom() | {atom(), {atom(), term()}}.
```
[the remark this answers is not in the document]
u3s: fixed in [05288e3](https://github.com/erlang/otp/pull/7543/commits/05288e3f32668ad581571e0a1c3b9e0d5943ff09)

lib/public_key/src/pubkey_ocsp.erl
u3s: I've applied this comment apart from passing trusted certs from ssl to public_key, which I would like to discuss first.

u3s:
> Should we handle "-type combined_cert() :: #cert{}." in the certificate(s) in the arguments? To reduce the decode/encode der roundtrips that is done?

I'm starting to understand and like your comment :-).
But are you suggesting this PR context or more general?

lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -153,31 +157,91 @@ get_nonce_value([#'Extension'{
...
+                ResponderCert == IssuerCert
+        end,
+    Case2 =
+        %% a CA Designated Responder (Authorized Responder, defined in
+        %%      Section 4.2.2.2) who holds a specially marked certificate
+        %%      issued directly by the CA, indicating that the responder may
+        %%      issue OCSP responses for that CA (id-kp-OCSPSigning)
+        fun() ->
+                public_key:pkix_is_issuer(ResponderCert, IssuerCert) andalso
+                                 designated_for_ocsp_signing(ResponderCert)
```
IngelaAndin, line 224: I think we want to verify the signature here also, that is public_key:pkix_verify(ResponderCert, IssuerCertsPublicKey) not only see that the issuer name check out!

IngelaAndin:
@u3s  you also need to rebase!

lib/public_key/src/pubkey_ocsp.erl
IngelaAndin: Maybe this is actually done further up!

lib/public_key/src/pubkey_ocsp.erl
```diff
@@ -92,16 +106,23 @@ match_single_response(_IssuerName, _IssuerKey, _SerialNum, []) ->
...
         true ->
-            {ok, Response};
+            {ok, SingleResponse};
         false ->
-            match_single_response(IssuerName, IssuerKey, SerialNum, Responses)
-    end.
+            match_single_response(IssuerName, IssuerKey, SerialNum, Tail)
+    end;
+match_single_response(IssuerName, IssuerKey, SerialNum,
+                      [_BadSingleResponse | Tail]) ->
```
IngelaAndin, line 124: Is it worth doing some common verify_data function in pubkey_cert.

lib/public_key/src/public_key.erl
```diff
@@ -1700,9 +1689,12 @@ otp_cert(Der) when is_binary(Der) ->
 otp_cert(#'OTPCertificate'{} = Cert) ->
     Cert;
 otp_cert(#cert{otp = OtpCert}) ->
-    OtpCert;
-otp_cert(#'Certificate'{} = Cert) ->
-    pkix_decode_cert(der_encode('Certificate', Cert), otp).
```
dgud, line 1705: Why remove this conversion?

lib/ssl/src/ssl_certificate.erl
```diff
@@ -589,18 +588,33 @@ verify_cert_extensions(Cert, #{cert_ext := CertExts} =  UserState) ->
...
+                case ssl_manager:lookup_trusted_cert(
+                       CertDbHandle, CertDbRef, SerialNr, IssuerId) of
+                    {ok, #cert{der = DerResponderCert}} ->
+                        true;
+                    _ ->
+                        false
+                end
+        end,
+    case public_key:pkix_ocsp_validate
+        (Cert, Issuer, OcspResponseDer, Nonce,
```
dgud, line 612: Personally I dislike the newline between function name and left parenthesis.

dgud (approved):

lib/ssl/src/ssl_certificate.erl
IngelaAndin: I would also prefer to brake it differently!  

```
case public_key:pkix_ocsp_validate(Cert, Issuer, 
                                                        OcspResponseDer, Nonce, 
                                                        ....                                        ) of 
```

u3s: I also dislike it ;-)
'(' acts as operator here and it would be better if it does not start the new line.

I will fix it.

lib/public_key/src/public_key.erl
IngelaAndin: I guess since most public_key functions take a der_cert() or otp_cert() and not a plain_cert()!?

IngelaAndin:
Maybe squash some of the commits?

lib/ssl/doc/src/standards_compliance.xml
```diff
@@ -1492,8 +1492,8 @@
       <row>
         <cell align="left" valign="middle"></cell>
         <cell align="left" valign="middle">status_request (RFC6066)</cell>
-	<cell align="left" valign="middle"><em>NC</em></cell>
-	<cell align="left" valign="middle"></cell>
+	<cell align="left" valign="middle"><em>C</em></cell>
```
u3s, line 1495: this is wrong. nothing is implemented on server side within this PR.
Certificate with status request extension should be marked as compliant in client section above and not here.

lib/public_key/src/public_key.erl
```diff
@@ -21,7 +21,7 @@
 %%
 
 -module(public_key).
-
+-feature(maybe_expr,enable).
```
u3s, line 24: check if can be removed (since PR is targeted to OTP-27)

lib/public_key/src/public_key.erl
u3s: because ... I think ... this clause was added for initial OCSP implementation, which I changed in this PR.
removed clause is dropping the DER representation which I need at some later OCSP processing occasion.

In order to get DER later I would need to run similar conversion one more time, which I think is a waste.
For that reason, I created `combined_cert` function which does not loose partial processing result.

removed clause, became not useful.

lib/public_key/src/pubkey_ocsp.erl
u3s: I'm not sure what we could gain here ...
`public_key:verify` is used already in `do_verify_signature`...

Am I missing something? pls explain.

lib/public_key/src/pubkey_ocsp.erl
u3s: Yes. This call realizes this:
`ok ?= do_verify_signature(ResponseDataDer, Signature, SignatureAlgo, ResponderCert)`

OCSP response contains list of certificates, we need to iterate over that list.
Once a potential candidate is found, signature of response is verified by `do_verify_signature` call.

lib/ssl/test/openssl_stapling_SUITE.erl
```diff
@@ -170,24 +213,44 @@ stapling_helper(Config, Opts) ->
...
     [{doc, "Verify OCSP stapling works with certificate with undetermined status."}].
-stapling_undetermined(Config)
+cert_status_undetermined(Config)
   when is_list(Config) ->
     stapling_negative_helper(Config, "undetermined/cacerts.pem",
                                   openssl_ocsp_undetermined, bad_certificate).
 
-stapling_no_staple() ->
+staple_missing() ->
     [{doc, "Verify OCSP stapling works with a missing OCSP response."}].
```
u3s, line 252: is this correct behavior? should we fail connection?
RFC6066 8: "server MAY also choose not to send a CertificateStatus ..."

lib/ssl/test/openssl_stapling_SUITE.erl
```diff
@@ -170,24 +213,44 @@ stapling_helper(Config, Opts) ->
...
+cert_status_undetermined() ->
     [{doc, "Verify OCSP stapling works with certificate with undetermined status."}].
-stapling_undetermined(Config)
+cert_status_undetermined(Config)
   when is_list(Config) ->
     stapling_negative_helper(Config, "undetermined/cacerts.pem",
                                   openssl_ocsp_undetermined, bad_certificate).
 
-stapling_no_staple() ->
+staple_missing() ->
```
u3s, line 251: @u3s - check (at least by code reading) how ssl client behaves if it receives OCSP response which wasn't requested.

lib/ssl/doc/src/ssl.xml
```diff
@@ -1280,28 +1280,19 @@ fun(srp, Username :: binary(), UserState :: term()) ->
...
-  <!--     </datatype> -->
+      <datatype>
+        <name name="stapling"/>
+        <desc><p>If <c>staple</c> or a map, OCSP stapling will be enabled, an
+        extension of type "status_request" will be included in the
+        client hello to indicate the desire to receive certificate
+        status information. If <c>no_staple</c> (the default), OCSP stapling will
+        be disabled.</p>
+
+        <p>When map is used, boolen ocsp_nonce key may indicate if
```
u3s, line 1291: typo: boolen -> boolean

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2209,24 +2203,24 @@ bad_key(#{algorithm := ecdsa}) ->
...
+                  #{stapling_state := #{configured := false}} = SslState,
                   VerifyResult, CertPath, LogLevel) ->
     maybe_check_crl(OtpCert, SslState, VerifyResult, CertPath, LogLevel);
 cert_status_check(_OtpCert,
-                  #{ocsp_state := #{ocsp_stapling := true,
-                                    ocsp_expect := undetermined}},
+                  #{stapling_state := #{configured := true,
+                                        status := undetermined}},
                   _VerifyResult, _CertPath, _LogLevel) ->
     {bad_cert, {revocation_status_undetermined, not_stapled}};
```
u3s, line 2220: check if this clause is dead code. openssl_stapling_SUITE works fine without it.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -804,42 +805,36 @@ update_encryption_state(client, State) ->
...
+                          StaplingInfo).
 
 store_peer_cert(#state{session = Session,
                        handshake_env = HsEnv} = State, PeerCert, PublicKeyInfo) ->
     State#state{session = Session#session{peer_certificate = PeerCert},
                 handshake_env = HsEnv#handshake_env{public_key_info = PublicKeyInfo}}.
 
-split_cert_entries(CertEntries, OcspState) ->
-    split_cert_entries(CertEntries, OcspState, [], #{}).
+stapling_info(CertEntries, StaplingState) ->
```
u3s, line 819: stapling_info/2 wrapper seems not needed. consider removal.

lib/ssl/src/tls_handshake_1_3.erl
```diff
@@ -804,42 +805,36 @@ update_encryption_state(client, State) ->
...
-                        OcspState0;
-                    _ ->
-                        OcspState0#{ocsp_expect => stapled}
-                end,
-    split_cert_entries(CertEntries, OcspState, [DerCert | Chain], Ext#{Id => Extensions}).
+    StaplingState = case maps:get(status_request, Extensions0, undefined) of
+                        undefined ->
+                            StaplingState0;
+                        _ ->
+                            StaplingState0#{status => stapled}
```
u3s, line 834: code here seems to not care much about which certificate had a staple nor about if all certificates got a staple or only one.
this might be problematic, if for example we receive a staple for non peer certificate and treat the whole path as not revoked and OK.

lib/ssl/src/tls_handshake.erl
```diff
@@ -476,9 +476,9 @@ decode_handshake(Version, Tag, Msg) ->
     ssl_handshake:decode_handshake(Version, Tag, Msg).
 
 
-ocsp_expect(true) ->
+default_stapling_status(true) ->
     staple;
```
u3s, line 480: `staple` meaning is not obvious and can be easily confused with `stapled`.
consider renaming atom to something better.

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1957,22 +1951,22 @@ extension_value(#psk_key_exchange_modes{ke_modes = Modes}) ->
...
     Cookie.
 
-handle_ocsp_extension(#{ocsp_stapling := _OcspStapling}, Extensions) ->
+handle_cert_status_extension(#{stapling := _Stapling}, Extensions) ->
     case maps:get(status_request, Extensions, false) of
         undefined -> %% status_request in server hello is empty
-            #{ocsp_stapling => true,
-              ocsp_expect => staple};
+            #{configured => true,
+              status => staple};
```
u3s, line 1958: `negotiated` is more informative than `staple`

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2209,24 +2203,24 @@ bad_key(#{algorithm := ecdsa}) ->
     unacceptable_ecdsa_key.
 
 cert_status_check(_,
-                  #{ocsp_state := #{ocsp_stapling := true,
-                                    ocsp_expect := stapled}},
+                  #{stapling_state := #{configured := true,
+                                        status := stapled}},
```
u3s, line 2207: `stapled` is not precise enough for TLS1.3

lib/ssl/src/ssl_handshake.erl
u3s, line 2220: maybe fallback to CRL here?

lib/ssl/src/ssl_handshake.erl
```diff
@@ -2209,24 +2203,24 @@ bad_key(#{algorithm := ecdsa}) ->
...
+                                        status := undetermined}},
                   _VerifyResult, _CertPath, _LogLevel) ->
     {bad_cert, {revocation_status_undetermined, not_stapled}};
 cert_status_check(_OtpCert,
-                  #{ocsp_state := #{ocsp_stapling := true,
-                                    ocsp_expect := no_staple}},
+                  #{stapling_state := #{configured := true,
+                                        status := no_staple}},
                   _VerifyResult, _CertPath, _LogLevel) ->
     {bad_cert, {revocation_status_undetermined, not_stapled}}.
```
u3s, line 2225: fallback to CRL check if OSCP staple(s) was not received?

lib/ssl/src/ssl_certificate.erl
```diff
@@ -589,18 +588,33 @@ verify_cert_extensions(Cert, #{cert_ext := CertExts} =  UserState) ->
 
 verify_cert_extensions(Cert, UserState, [], _) ->
     {valid, UserState#{issuer => Cert}};
-verify_cert_extensions(Cert, #{ocsp_responder_certs := ResponderCerts,
-                               ocsp_state := OscpState,
-                               issuer := Issuer} = UserState,
-                       [#certificate_status{response = OcspResponsDer} | Exts],
+verify_cert_extensions(Cert, #{stapling_state := StaplingState,
```
u3s, line 591: what if status_request_ext is negotiated but actual OCSP response is not received?
will code fail in such situation?

lib/ssl/src/tls_connection.erl
```diff
@@ -248,27 +248,27 @@ hello(internal, #server_hello{} = Hello,
...
+                          stapling_state = maps:merge(StaplingState0,StaplingState)}});
             %% TLS 1.3
-            {next_state, wait_sh, SelectedVersion, OcspState} ->
+            {next_state, wait_sh, SelectedVersion, StaplingState} ->
                 %% Continue in TLS 1.3 'wait_sh' state
                 {next_state, wait_sh,
                  State#state{handshake_env =
-                                 HsEnv#handshake_env{ocsp_stapling_state =
-                                                         maps:merge(OcspState0, OcspState)},
+                                 HsEnv#handshake_env{stapling_state =
```
u3s, line 270: When running `-suite openssl_stapling_SUITE -group 'tlsv1.3' -case staple_with_nonce` - this section does not get executed. How it works for TLS1.3 ??

u3s: Maybe this is how it should actually work as described in https://datatracker.ietf.org/doc/html/rfc8446#section-4.4.2.1.

u3s: This stapling code seems dead - maybe remove.

lib/ssl/doc/src/standards_compliance.xml
u3s: 609f4ac122

lib/ssl/doc/src/ssl.xml
u3s: 609f4ac122

lib/ssl/src/tls_connection.erl
u3s: for TLS 1.3 status_request extension (server response completing extension negotiation) is checked `tls_handshake_1_3:split_cert_entries` not in ServerHello.
merging StaplingState here is not needed I guess.

@IngelaAndin

lib/ssl/src/tls_handshake_1_3.erl
u3s: 0a6bd7979c

lib/ssl/src/ssl_handshake.erl
```diff
@@ -1957,22 +1951,22 @@ extension_value(#psk_key_exchange_modes{ke_modes = Modes}) ->
...
         undefined -> %% status_request in server hello is empty
-            #{ocsp_stapling => true,
-              ocsp_expect => staple};
+            #{configured => true,
+              status => staple};
         false -> %% status_request is missing (not negotiated)
-            #{ocsp_stapling => true,
-              ocsp_expect => no_staple};
+            #{configured => true,
+              status => no_staple};
```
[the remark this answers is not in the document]
u3s: 5b711cb4d1

lib/ssl/src/tls_handshake.erl
u3s: changed to `received_peer` which is supposed to mean, _staple(OCSP response) received for peer certificate_
9e5f3ce3ae

lib/public_key/src/pubkey_ocsp.erl
IngelaAndin: This was a long time ago. I think I was think that the logic of verify_past_timstamp and verify_next_update also exists in pubkey_cert module.

lib/ssl/doc/src/standards_compliance.xml
IngelaAndin: You can mark it as PC (partially complient)

u3s: 4.4.2.1. OCSP Status and SCT Extensions -> mark as PC

lib/ssl/src/ssl_certificate.erl
IngelaAndin: It is handled by this code in  ssl_handshake.  (ocsp_expect := stapled means that the stapel information was recived)

cert_status_check(_,
                  #{ocsp_state := #{ocsp_stapling := true,
                                    ocsp_expect := stapled}},
                  _VerifyResult, _, _) ->
    %% OCSP staple will now be checked by
    %% ssl_certificate:verify_cert_extensions/2 in ssl_certificate:validate
    valid;

lib/ssl/src/ssl_handshake.erl
```diff
@@ -3805,23 +3799,28 @@ path_validate([], _, _, _, _, _, _, _, _, _, {error, {bad_cert, root_cert_expire
 path_validate([], _, _, _, _, _, _, _, _, _, Error) ->
     Error;
 path_validate([{TrustedCert, Path} | Rest], ServerName, Role, CertDbHandle, CertDbRef, CRLDbHandle,
-              Version, SslOptions, CertExt, InvalidatedList, Error) ->
+              Version, SslOptions, StaplingInfo, InvalidatedList, Error) ->
```
IngelaAndin, line 3802: There can be more certificate extensions that the staplinginfo I do not understand why you are changing this variable name.

lib/ssl/src/ssl_handshake.erl
IngelaAndin: Yes TLS-1.3 is handled here. Please elaborate!

u3s: I mean in TLS1.3 maybe we should be more specific when describing for which cert we received OCSP response:
1. received_peer
2. received_chain
3. received some other OCSP response for certificate from chain
4. received OCSP response not related to certificate chain

- maybe this is indirectly handled later when cert chain gets verified, but I would like to check it
- also crl_check option by default (set to `true`) tries to verify the whole chain, then it can be set to less strict with `peer` value ... for TLS1.3 stapling we could find some similarity and maybe we should try to have it?

lib/ssl/src/ssl_handshake.erl
u3s: 66e77332e7 and b1c1d3d770

lib/ssl/src/ssl_handshake.erl
u3s: maybe `received some OCSP response` is good enough at this stage ... I need to analyze it

lib/ssl/src/ssl_certificate.erl
u3s: yes. thank you.

lib/ssl/doc/src/standards_compliance.xml
u3s: c7424afdb4

lib/ssl/src/tls_handshake_1_3.erl
IngelaAndin: I think this code determines that there is at least on staple  and all present staples will be verified in the path validation. They are saved in the state for this purpose.

lib/ssl/src/ssl_handshake.erl
u3s: probably need, but current tests don't verify it.
- not_negotiated(aka no staple) - is tested in `staple_missing` testcase
- not_received(aka undetermined) - can't be tested easily and currently it is not covered
I'm leaving that clause

lib/ssl/src/tls_handshake_1_3.erl
IngelaAndin: Do we need to be able to verify that the whole chain has been stapled? If so should be done during path validation 
in verify callback code in ssl_certificate.

lib/ssl/src/ssl_handshake.erl
IngelaAndin: I do not see how nr 4 could happen!?

IngelaAndin: The other alternatives I read as peer was stapled all certs in chain where stapled only some certs in the chain where stapled. As I said the code now would check all staples that are sent but it will not verify which certs that are stapled.
I think nr 1 and 2 are the most likly senarios.

lib/ssl/src/ssl_handshake.erl
IngelaAndin: Or maybe there is a missing test case!  I think this looks like we want to fail if no staple was received .

lib/ssl/src/ssl_handshake.erl
IngelaAndin: To be able to fallback to CRL check, CRL options must have been specified so it can not be done unconditionally.

lib/ssl/src/ssl_handshake.erl
IngelaAndin: There can be if we implement it!

lib/ssl/src/tls_connection.erl
IngelaAndin: I guess you are right, try removing it and run the tests. No it will not be covered by refactor other than the code being placed in an other file called client something.

lib/ssl/src/ssl_handshake.erl
u3s: I think you're right. I was misleaded because StaplingInfo was in a lot of places.
For TLS1.3 in tls_handshake_1_3 all extensions will be collected delivered here.

I will change that.

lib/ssl/src/ssl_handshake.erl
u3s: > I do not see how nr 4 could happen!?

bad actor knows server private key, tries to setup fake server using stolen certificate. can't provide real OCSP response and attaches a dummy one hoping it will not be checked ... a bit theoretical but still one possible flow

lib/ssl/src/tls_connection.erl
u3s: > I guess you are right, try removing it and run the tests. No it will not be covered by refactor other than the code being placed in an other file called client something.

Do you mean `maps:merge(StaplingState0, StaplingState)` part or whole TLS1.3 clause (I think coverage report marks it as used, but I don't understand what scenario that could be)

lib/ssl/src/ssl_handshake.erl
u3s: changed StaplingInfo into `ExtInfo` (picked it because map contains not only extensions, but also stapling state which is something additional)
18649553ea

lib/ssl/src/ssl_handshake.erl
IngelaAndin: Was thinking should we call it received_staple as the response in this case is the staple which I think is less obvious if we only call it response.

lib/ssl/src/tls_connection.erl
IngelaAndin: Clarification it will then change to the TLS-1.3 state machine. 

```
 [{change_callback_module, tls_client_connection_1_3},
                  {next_event, internal, Hello}]}
```

lib/ssl/src/ssl_certificate.erl
u3s: fixed in 509784a4e6

lib/ssl/src/tls_connection.erl
u3s: I think I will leave this code as is - leave the merge.
For TLS1.3, in this clause we will only initialized StaplingState to `#{configured => true, status => not_negotiated}`

lib/ssl/src/ssl_handshake.erl
u3s: let's leave for now and decide once I push updated branch. I have no strong opinion,
- `received_response` status key matches with `response` key containing CertStatus ...

lib/ssl/test/openssl_stapling_SUITE.erl
u3s: handshake will fail with bad cert, when
- stapling is configured but no status_request was received for peer certificate
- stapling is NOT configured AND any status_request was received
fixed in f983d0b909

lib/ssl/src/tls_handshake_1_3.erl
u3s: f983d0b909 adjusts that. handshake will fail if stapling is configured but peer(server) cert does not arrive with a valid status extension.

Modifying it to verifying whole chain should be easy to achieve, but currently we are strict only about peer cert.

lib/public_key/src/pubkey_ocsp.erl
u3s: Well potentially, pubkey_cert could be refactored and make use of atomic time comparison functions introduced in pubkey_ocsp.

But the overlap is relatively small and time verifications for OCSP are a bit specific not matching 1:1 with pubkey_cert, I don't think it makes sense now. Will resolve that one without any change.

u3s:
I've pushed updated branch. fixup commits contain updates since preview review. I will squash them when finish discussions.

lib/public_key/src/pubkey_ocsp.erl
u3s: this is not realized in `ssl_certificate:verify_cert_extensions/4` which provides public key with IsTrustedResponderFun

resolving

u3s:
> Should we handle "-type combined_cert() :: #cert{}." in the certificate(s) in the arguments? To reduce the decode/encode der roundtrips that is done?

I created backlog item, to handle this in separate PR. Plan is to do it soon.

lib/ssl/src/ssl_handshake.erl
u3s: Your suggestion make sense. Response is very generic, let's use staple in this context. I've renamed both keys:
- received_response -> received_staple
- response -> staple

f06f636277

lib/ssl/src/ssl_handshake.erl
u3s: I'm not sure how this CRL fallback should look like. Maybe we can take that discussion after merging this PR ...

IngelaAndin: I think it will look like when OCSP check finds itself in result "undetermined" it should call mayb_check_crl if crl_check is set to relevant value  (not false, and not peer if CA cert in TLS-1.3, only peer relevant in TLS-1.2) to see if it can get a valid result.  It would have to be handled in different places for TLS-1.3 (certificate callback) and it can be handled in cert_status_check for TLS-1.2.  We can fix it in follow up PR for master, as we really like to merge this soon to be able to also continue work with other TLS refactor jobs.

lib/ssl/src/tls_dtls_connection.erl
```diff
@@ -341,10 +341,12 @@ certify(internal, #certificate{asn1_certificates = [Peer|_]} = Cert,
...
                                crl_db = CRLDbInfo},
                handshake_env = #handshake_env{
-                                  stapling_state = #{status := Status} = StaplingState},
+                                  stapling_state = #{status := StaplingStatus} =
+                                      StaplingState},
                connection_env = #connection_env{
                                    negotiated_version = Version},
-               ssl_options = Opts} = State0) when Status =/= staple ->
+               ssl_options = Opts} = State0)
+  when StaplingStatus == not_negotiated; StaplingStatus == received_peer ->
```
IngelaAndin: I do not understand why StaplingStatus could be equal to not_negotiated here? This code is not executed in TLS-1.3 context, to my understanding it should only received_peer.

IngelaAndin: Or probably received_response when all changes are applied!

IngelaAndin: Ah, I think new commit clears it up :)  Old behavior was kind of double negation.

IngelaAndin (approved):

lib/ssl/src/ssl_handshake.erl
u3s: I created backlog item for finalizing this thread.
Resolving it here.

lib/ssl/src/ssl_handshake.erl
u3s: This might be tricky to test. You would need TLS server negotiated status_request but not sending the response ...

PR #7552, opened 2023-08-11
otp: Explain the term equivalence operators

jhogberg:
This PR attempts to document the term equivalence operators, whose behavior is [set to change slightly in OTP 27](https://erlangforums.com/t/in-erlang-otp-27-0-0-will-no-longer-be-exactly-equal-to-0-0/2586). It's a pretty subtle (and quite pedantic) change so I'd really appreciate all the feedback I can get, especially from the community at large.

### DIFF
--- a/system/doc/reference_manual/data_types.xml
+++ b/system/doc/reference_manual/data_types.xml
@@ -87,6 +87,37 @@
 11> <input>1_234.333_333</input>
 1234.333333
 </pre>
+<marker id="numeric_comparisons" />
+    <section>
+      <title>Comparisons</title>
+      <p>Both integers and floats share the same linear order. That is,
+        <c>1</c> compares less than <c>2.4</c>, <c>3</c> compares greater than
+        <c>2.99999</c>, and <c>5</c> is equal to <c>5.0</c>.</p>
+      <p>When wanting to compare an integer with another integer or a float
+        with another float, it may be tempting to use the term equivalence
+        operators (<c>=:=</c>, <c>=/=</c>) or pattern matching. This works for
+        integers which has a distinct representation for every number, but
+        there's a surprising edge case for floating-point as the latter has two
+        representations for zero which are considered different by the term
+        equivalence operators and pattern matching.</p>
+      <p>If you wish to compare floating-point numbers <em>numerically</em>,
+        use the regular comparison operators (such as <c>==</c>) and add guards
+        that require both the arguments to be floating-point.</p>
+      <note>
+        <p>Prior to OTP 27, the term equivalence operators had a bug where they
+          considered <c>0.0</c> and <c>-0.0</c> to be the same term. Legacy
+          code that makes equality comparisons on floating-point zero should
+          migrate to using the equal-to (<c>==</c>) operator with
+          <c>is_float/1</c> guards, and compiler warnings have been added to
+          that effect. These can be silenced by writing <c>+0.0</c> instead,
+          which is the same as <c>0.0</c> but makes the compiler interpret the
+          comparison as being purposely made against <c>0.0</c>.</p>
+        <p>Note that this does <em>not</em> break compatibility with IEEE 754
+          which mandates that <c>0.0</c> and <c>-0.0</c> should compare
+          equal: they are equal when interpreted as numbers (<c>==</c>), and
+          unequal when interpreted as opaque terms (<c>=:=</c>).</p>
+      </note>
+    </section>
 <marker id="float_representation_problem" />
     <section>
       <title>Representation of Floating Point Numbers</title>
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -840,7 +840,7 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map 
       whether two terms are indistinguishable. While the other operators
       consider the same <em>numbers</em> equal even when their types differ
       (<c>1 == 1.0</c> is true), the term equivalence operators return whether
-      there exists any function that can tell their arguments apart.</p>
+      or not there exists a way to tell the arguments apart.</p>
 
     <p>For example, while the terms <c>0</c> and <c>0.0</c> represent the same
       <em>number</em>, we can tell them apart by using the <c>is_integer/1</c>
@@ -855,8 +855,9 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map 
 
     <p>The term equivalence operators are useful when reasoning about terms as 
       opaque values, for example in associative containers or memoized
-      functions where using the equal-to operator (<c>==</c>) can result in
-      subtly incorrect results.</p>
+      functions where using the equal-to operator (<c>==</c>) risks producing
+      incorrect results as a consequence of mixing up numbers of different
+      types.</p>
 
     <p>Term comparison operators return the Boolean value of the
       expression, <c>true</c> or <c>false</c>.</p>
--- a/system/doc/reference_manual/expressions.xml
+++ b/system/doc/reference_manual/expressions.xml
@@ -801,11 +801,11 @@ Expr1 <input>op</input> Expr2</pre>
       </row>
       <row>
         <cell align="left" valign="middle">=:=</cell>
-        <cell align="left" valign="middle">Exactly equal to</cell>
+        <cell align="left" valign="middle">Term equivalence</cell>
       </row>
       <row>
         <cell align="left" valign="middle">=/=</cell>
-        <cell align="left" valign="middle">Exactly not equal to</cell>
+        <cell align="left" valign="middle">Term non-equivalence</cell>
       </row>
       <tcaption>Term Comparison Operators.</tcaption>
     </table>
@@ -836,6 +836,28 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map 
       depending on the size of the float because otherwise comparison of large
       floats and integers would lose their transitivity.</p>
 
+    <p>The term equivalence operators, <c>=:=</c> and <c>=/=</c>, return
+      whether two terms are indistinguishable. While the other operators
+      consider the same <em>numbers</em> equal even when their types differ
+      (<c>1 == 1.0</c> is true), the term equivalence operators return whether
+      there exists any function that can tell their arguments apart.</p>
+
+    <p>For example, while the terms <c>0</c> and <c>0.0</c> represent the same
+      <em>number</em>, we can tell them apart by using the <c>is_integer/1</c>
+      function. Hence, <c>=:=</c> and <c>=/=</c> consider them different.</p>
+
+    <p>Furthermore, the terms <c>0.0</c> and <c>-0.0</c> also represent the
+      same <em>number</em>, but they yield different results when converted to
+      string form through <c>float_to_list/1</c>: when given the former it
+      returns a string without a sign, and when given the latter it returns a
+      string with a sign. Therefore, <c>=:=</c> and <c>=/=</c> consider
+      them different.</p>
+
+    <p>The term equivalence operators are useful when reasoning about terms as 
+      opaque values, for example in associative containers or memoized
+      functions where using the equal-to operator (<c>==</c>) can result in
+      subtly incorrect results.</p>
+
     <p>Term comparison operators return the Boolean value of the
       expression, <c>true</c> or <c>false</c>.</p>
 
@@ -845,17 +867,35 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map 
 true
 2> <input>1=:=1.0.</input>
 false
-3> <input>1 > a.</input>
+3> <input>0=:=0.0.</input>
 false
-4> <input>#{c => 3} > #{a => 1, b => 2}.</input>
+4> <input>0.0=:=-0.0.</input>
 false
-5> <input>#{a => 1, b => 2} == #{a => 1.0, b => 2.0}.</input>
+5> <input>0.0=:=+0.0.</input>
 true
-6> <input>&lt;&lt;2:2>> &lt; &lt;&lt;128>>.</input>
+6> <input>1 > a.</input>
+false
+7> <input>#{c => 3} > #{a => 1, b => 2}.</input>
+false
+8> <input>#{a => 1, b => 2} == #{a => 1.0, b => 2.0}.</input>
 true
-7> <input>&lt;&lt;3:2>> &lt; &lt;&lt;128>>.</input>
+9> <input>&lt;&lt;2:2>> &lt; &lt;&lt;128>>.</input>
+true
+10> <input>&lt;&lt;3:2>> &lt; &lt;&lt;128>>.</input>
 false
 </pre>
+    <note>
+      <p>Prior to OTP 27, the term equivalence operators had a bug where they
+        considered <c>0.0</c> and <c>-0.0</c> to be the same term.</p>
+      <p>This was fixed in OTP 27 but legacy code may have expected them to
+        be considered the same. To help users catch errors that may arise from
+        an upgrade, the compiler raises a warning when <c>0.0</c> is
+        pattern-matched or used in a term equivalence test.</p>
+      <p>If you need to match <c>0.0</c> specifically, then the warning can be
+        silenced by writing <c>+0.0</c> instead, which produces the same term
+        but makes the compiler interpret the match as being done on
+        purpose.</p>
+    </note>
   </section>
 
   <section>


system/doc/reference_manual/expressions.xml
```diff
@@ -845,6 +859,8 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
 true
 2> <input>1=:=1.0.</input>
 false
+2> <input>+0.0=:=-0.0.</input>
+false
```
RaimoNiskanen, line 863: Should we have `0.0 =:= -0.0` and `+0.0 =:= 0.0`?

system/doc/reference_manual/expressions.xml
```diff
@@ -836,6 +836,20 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
...
+      floats or integers (<c>1 == 1.0</c> is true), the term equivalence
+      operators return whether there exists any function that can tell their
+      arguments apart.</p>
+    <p>For example, while the terms <c>0.0</c> and <c>-0.0</c> represent the
+      exact same <em>number</em>, they yield different results when converted
+      to string form through <c>float_to_list/1</c>, so <c>=:=</c> and
+      <c>=/=</c> consider them different. This is useful when reasoning about
+      terms as opaque values, for example in associative containers or memoized
+      functions where using the equal-to (<c>==</c>) operator can result in
+      subtly incorrect results.</p>
```
RaimoNiskanen, line 851: To me this explanation explains why `==` and `=:=` behaves differently on `0.0` and `0`, but not why `=:=` regards `0.0` as different from `-0.0`.

RaimoNiskanen: Where should we explain that `0.0` and `+0.0` are exactly equal in runtime but have slightly different meaning in compile time?

jhogberg: >  To me this explanation explains why == and =:= behaves differently on 0.0 and 0, but not why =:= regards 0.0 as different from -0.0.

I've pushed up a commit that I hope will be a bit clearer.

> Where should we explain that 0.0 and +0.0 are exactly equal in runtime but have slightly different meaning in compile time?

I'm not sure that's necessary: they are the same even in compile-time, the compiler just happens to give the user a helpful warning when using the former in a pattern or term equality test.

system/doc/reference_manual/expressions.xml
```diff
@@ -836,6 +836,21 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
       depending on the size of the float because otherwise comparison of large
       floats and integers would lose their transitivity.</p>
 
+    <p>The term equivalence operators, <c>=:=</c> and <c>=/=</c>, return
+      whether two terms are indistinguishable. While the other operators
+      consider the same <em>numbers</em> equal regardless of whether they are
```
bjorng, line 841: I think that "regardless of" can be omitted. At least, that sounds good to me.

bjorng (approved):

system/doc/reference_manual/expressions.xml
jhogberg: Thanks, I've tweaked the wording a bit. :)

system/doc/reference_manual/expressions.xml
```diff
@@ -836,6 +836,21 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
       depending on the size of the float because otherwise comparison of large
       floats and integers would lose their transitivity.</p>
 
+    <p>The term equivalence operators, <c>=:=</c> and <c>=/=</c>, return
+      whether two terms are indistinguishable. While the other operators
+      consider the same <em>numbers</em> equal even when their types differ
+      (<c>1 == 1.0</c> is true), the term equivalence operators return whether
+      there exists any function that can tell their arguments apart.</p>
+    <p>For example, while the terms <c>0.0</c> and <c>-0.0</c> represent the
+      exact same <em>number</em>, they yield different results when converted
```
RaimoNiskanen, line 845: I think we need a *since* note of some kind here since that distinction is new.

jhogberg: Good point! :)

RaimoNiskanen: We should also explain that `0.0` and `-0.0` are distinct terms here:
https://www.erlang.org/doc/reference_manual/data_types#representation-of-floating-point-numbers
with a note when that changed...

system/doc/reference_manual/expressions.xml
RaimoNiskanen: But we need to state that they are the same

system/doc/reference_manual/expressions.xml
RaimoNiskanen: Perhaps a good place to state that `0.0` and `+0.0` are equivalent

system/doc/reference_manual/expressions.xml
jhogberg: Sure, I've pushed up another commit to that effect. :)

system/doc/reference_manual/expressions.xml
```diff
@@ -836,6 +836,21 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
...
+      to string form through <c>float_to_list/1</c>: when given the former it
+      returns a string without a sign, and when given the latter it returns a
+      string with a sign. Therefore <c>=:=</c> and <c>=/=</c> consider these
+      terms different. This is useful when reasoning about terms as opaque
+      values, for example in associative containers or memoized functions where
```
RaimoNiskanen, line 850: I still think it becomes confusing here, maybe partly because I am familiar with the history.

Above the text talks about `0.0` and `-0.0` and below it talks about `1` vs. `1.0`.
They should not be in the same paragraph.

I suggest to paragraph break here, and add to the paragraph above that also `0 =/= 0.0` so `0 =/= 0.0 andalso 0.0 =/= -0.0 andaso -0.0 =/= 0` is `true`, that is: all these 3 zeros are different.
Then the paragraph below can start with "The equivalence operators are useful when reasoning ...".

I think we should not mix up the new distinction between `0.0` and `-0.0` with the old distinction between `0` and `0.0`.

The Note further down about the "bug" in the equivalence between `0.0` and `-=0.0` makes things clear.  I like that.

system/doc/reference_manual/data_types.xml
```diff
@@ -87,6 +87,19 @@
...
+      <title>Comparisons</title>
+      <p>Both integers and floats share the same linear order. That is,
+        <c>1</c> compares less than <c>2.4</c>, <c>3</c> compares greater than
+        <c>2.99999</c>, and <c>5</c> is equal to <c>5.0</c>.</p>
+      <p>When wanting to match a specific number without also matching its
+        integer- or floating-point twin, the term equivalence operators
+        (<c>=:=</c>, <c>=/=</c>) should be used instead.</p>
+      <p>Note that since OTP 27, <c>0.0</c> and <c>-0.0</c> are <em>not</em>
+        considered to be the same term since they can be distinguished through
+        the use of string or bitstring conversion functions.</p>
```
RaimoNiskanen, line 101: At the end of the next section it is stated that we don not follow IEEE completely, and that Nan, +Inf and -Inf are not implemented in Erlang.  I suggest writing there that 0.0 vs. -0.0 *is* since it is also an IEEE feature, but one that we *do* support.

system/doc/reference_manual/data_types.xml
```diff
@@ -87,6 +87,37 @@
...
 1234.333333
 </pre>
+<marker id="numeric_comparisons" />
+    <section>
+      <title>Comparisons</title>
+      <p>Both integers and floats share the same linear order. That is,
+        <c>1</c> compares less than <c>2.4</c>, <c>3</c> compares greater than
+        <c>2.99999</c>, and <c>5</c> is equal to <c>5.0</c>.</p>
+      <p>When wanting to match a specific number without also matching its
+        integer- or floating-point twin, it may be tempting to use the term
```
josevalim, line 97: Nitpick: the negation is hard to process here, a suggestion to special out the specific comparison pairs:

```suggestion
      <p>When wanting to match an integer with another integer or a float
        with another float, it may be tempting to use the term
```

system/doc/reference_manual/data_types.xml
```diff
@@ -87,6 +87,37 @@
...
+        equivalence operators (<c>=:=</c>, <c>=/=</c>) or pattern matching.
+        This works for integers which has a distinct representation for every
+        number, but does not work for floating-point as the latter has two
+        representations for zero which are considered different by the term
+        equivalence operator (<c>=:=</c>).</p>
+      <p>If you wish to specifically compare floating-point numbers, use the
+        regular comparison operators (such as <c>==</c>) and add guards that
+        require both the arguments to be floating-point.</p>
```
josevalim, line 105: Nitpick: I am not sure if we should label it as "does not work". As you said later, you are comparing them as terms, not as numbers, which is fine by IEEE754. To me this is 

```suggestion
        However, beware floating-point numbers have two representations for
        zero (<c>0.0</c> and <c>-0.0</c>) and the term equivalent operator
        (<c>=:=</c>) will consider them to be distinct, which may be undesired.
        This is not a concern for integers, which have a distinct representation
        for every number.</p>
      <p>If you wish to specifically compare floating-point numbers while
        keeping the zero equivalence, use the regular comparison operators
        (such as <c>==</c>) and add guards that require both the arguments
        to be floating-point.</p>
```

jhogberg: Thanks, I've reworded it a bit :)

system/doc/reference_manual/data_types.xml
```diff
@@ -87,6 +87,37 @@
 11> <input>1_234.333_333</input>
 1234.333333
 </pre>
+<marker id="numeric_comparisons" />
+    <section>
+      <title>Comparisons</title>
+      <p>Both integers and floats share the same linear order. That is,
+        <c>1</c> compares less than <c>2.4</c>, <c>3</c> compares greater than
+        <c>2.99999</c>, and <c>5</c> is equal to <c>5.0</c>.</p>
+      <p>When wanting to an integer with another integer or a float with
```
Maria-12648430, line 96:
```suggestion
      <p>When wanting to compare an integer with another integer or a float with
```

jhogberg: Thanks :)

system/doc/reference_manual/data_types.xml
```diff
@@ -87,6 +87,37 @@
 11> <input>1_234.333_333</input>
 1234.333333
 </pre>
+<marker id="numeric_comparisons" />
+    <section>
+      <title>Comparisons</title>
+      <p>Both integers and floats share the same linear order. That is,
+        <c>1</c> compares less than <c>2.4</c>, <c>3</c> compares greater than
+        <c>2.99999</c>, and <c>5</c> is equal to <c>5.0</c>.</p>
+      <p>When wanting compare an integer with another integer or a float with
```
Maria-12648430, line 96:
```suggestion
      <p>When wanting to compare an integer with another integer or a float with
```
:sweat_smile::sweat_smile::sweat_smile:

Maria-12648430: On another note, in my previous comment I suggested "match", but you used "compare", which is certainly the better choice here.
But since you're subsequently talking only about testing for (in-)equality, it may be good to end the sentence with something like "... to check for equality or inequality", or write it like "... to use pattern matching, or the term equivalence operators <c>=:=</c> to test for equality and <c>=/=</c> to test for inequality"?

jhogberg: I fear that'll just make things too wordy, it's pretty bad as it is. :thinking:

Maria-12648430: Yeah, it is pretty hard to be concise _and_ precise :sweat: I guess the message gets across even without the equality/inequality part, so it's fine as far as I'm concerned.

system/doc/reference_manual/expressions.xml
```diff
@@ -836,6 +836,28 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
...
+      same <em>number</em>, but they yield different results when converted to
+      string form through <c>float_to_list/1</c>: when given the former it
+      returns a string without a sign, and when given the latter it returns a
+      string with a sign. Therefore, <c>=:=</c> and <c>=/=</c> consider
+      them different.</p>
+
+    <p>The term equivalence operators are useful when reasoning about terms as 
+      opaque values, for example in associative containers or memoized
+      functions where using the equal-to operator (<c>==</c>) can result in
+      subtly incorrect results.</p>
```
Maria-12648430, line 859: This is also somewhat confusing. It brings up two new concepts, without explanation or examples (which of course would blow this document out of all proportion). I mean, I know about the subtly incorrect results you mention here, because they did bite me in the past. But I'd say that I wouldn't have been aware of them if I had read this paragraph before :woman_shrugging:

Maybe you should just show briefly how different implementations of the same `$THING` behave differently by using the equality vs term equivalence operators/pattern matching, like `ordsets` (working by equality) and `sets` (working by term equivalence), or the like?
As of late, there is a NOTE box at the top of the [`sets` documentation](https://www.erlang.org/doc/man/sets.html) outlining the differences in behavior between `sets`, `ordsets` and `gb_sets` that you could link to.

Also, I think it is wrong to unconditionally say that using the equality operators can result in incorrect results, it kinda leaves an aftertaste of "equality operators ***bad***!!!". Which they are not: If equality is what you _want_, I'd say it is not wrong.
IMO the real issue is that there is the possibility of a mismatch between how you _expect_ something to behave and how it _really_ behaves if you don't pay attention to equality vs term equivalence, a mismatch that will probably manifest itself every so often, and is hard to find.

Maria-12648430: Reading back on my comment, I realize it sounds a bit harsh :sweat: Sorry for that :pray:

But I realize the wording and maybe-needed verbosity depends to some degree on what assumptions we can make about prior knowledge of a reader.

mikpe (approved):

system/doc/reference_manual/expressions.xml
```diff
@@ -836,6 +836,28 @@ number &lt; atom &lt; reference &lt; fun &lt; port &lt; pid &lt; tuple &lt; map
       depending on the size of the float because otherwise comparison of large
       floats and integers would lose their transitivity.</p>
 
+    <p>The term equivalence operators, <c>=:=</c> and <c>=/=</c>, return
+      whether two terms are indistinguishable. While the other operators
+      consider the same <em>numbers</em> equal even when their types differ
+      (<c>1 == 1.0</c> is true), the term equivalence operators return whether
+      there exists any function that can tell their arguments apart.</p>
```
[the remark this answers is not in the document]
jhogberg: `∀A∀B ¬(A =/= B) ≡ (A =:= B) ≡ ∀f (f(A) =:= f(B)))`, what's the problem? :grin:

> "... whether there exists **no** function...", 

The operators are mirrors of each other, so they both say the same thing, and I'm not sure the extra negation helps. :thinking: 

> Maybe just stick with the conventional notion of "... are exactly the same...", 

Hacker News got all up in arms because they misunderstood what we meant by "exactly equal to," so I'm not sure that'll help much either. :sweat_smile:

system/doc/reference_manual/expressions.xml
jhogberg: > But I realize the wording and maybe-needed verbosity depends to some degree on what assumptions we can make about prior knowledge of a reader.

If only we could assume that the reader knew everything to begin with... :smile: 

> it kinda leaves an aftertaste of "equality operators bad!!!". Which they are not: If equality is what you want, I'd say it is not wrong.

It's wrong when reasoning about terms as opaque values, which is what that paragraph is about.

Can you try playing around with the wording to see if there's any particular direction you'd prefer (in general, both files if you wish)?

system/doc/reference_manual/expressions.xml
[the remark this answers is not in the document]
josevalim: I definitely prefer term-equivalence operators as a name. I took a stab at writing this section. Feel free use any part of it or throw it all out. The idea is that we would describe the general rules, as they are currently written, provide examples, and then discuss comparing numbers in their own section.

---

The rules above apply to all comparison operators. The distinction from `==` and `/=` to `=:=` and `=/=` arises when comparing integers and floats. Let's explore this case in more detail.

## Comparing numbers

Similar to `+` and other arithmetic operators, Erlang provides a series of equality operators to compare numbers, regardless if they are integers or floats:

```erlang
1> 1 == 1.0.
true
2> 1 /= 1.0.
false
3> 10 < 5.0.
false
4> 10 >= 5.
true
```

When comparing an integer to a float, the number with the lesser precision is converted into the type of the other. A float is more precise than an integer until all significant figures of the float are to the left of the decimal point. This happens when the float is larger/smaller than `+/-9007199254740992.0`. The conversion strategy is changed depending on the size of the float because otherwise comparison of large floats and integers would lose their transitivity.

Additionally, floating-point numbers have an interesting property where zero is either negative (-0.0) or positive (+0.0), although in practice they compare equal:

```erlang
1> +0.0 == -0.0
true
```

However, in some cases, you may want to distinguish between integers and floats, or between positive and negative zero, as they have different in-memory representations (and in some mathematical domains they may also exhibit different properties). Erlang provides two _term equivalence operators_ (`=:=` and `=/=`) for this exact purpose:

```erlang
1> 1 =:= 1.0.
false
2> +0.0 =:= -0.0
false
```

Finally, note pattern matching relies on term equivalence. Therefore, the following expressions won't match either:

```erlang
1> 1 = 1.0.
** exception error: no match of right hand side value 1.0
2> -0.0 = +0.0.
```

To avoid ambiguity, Erlang will warn if 0.0 is used inside a pattern. You must instead match on `+0.0` or `-0.0`. If you want to match on both positive and negative zero floating points, you can write:

```erlang
is_floating_zero(Num) when is_float(Num), Num == 0.0 -> true;
is_floating_zero(_Num) -> false.
```

In other words, the comparison operators will behave in accordance to the other arithemetic operators in the language, and adhere to the IEEE 754 specification. The term equivalence operators allow developers to explicitly compare their underlying representations.

> Note: Prior to OTP 27, the term equivalence operators had a bug where they considered <c>0.0</c> and <c>-0.0</c> to be the same term.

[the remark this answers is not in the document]
josevalim: FWIW, Elixir calls those operators "strictly equal". It may make more sense to think some comparisons are more strict than others (rather than more exact).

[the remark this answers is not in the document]
jhogberg: Thanks, I like that take on things. I'll try to incorporate it when I'm back from vacation mid-to-late September.

system/doc/reference_manual/expressions.xml
Maria-12648430: > Can you try playing around with the wording to see if there's any particular direction you'd prefer (in general, both files if you wish)?

I guess I had that coming :rofl: Ok, I'll think and play :wink:

jhogberg:
I've changed the wording a slight bit, keeping most of the PR the same. I'd like to continue in the direction @josevalim took but I think that's better done in a general overhaul of this subject as it's currently split between "expressions" and "data types." I think a separate PR will be better for that.

If that sounds good, I'll merge this PR next week and start work on implementing the comparison change itself.

PR #7554, opened 2023-08-11
public_key: Fix case insensitive match for countryname

### LINKED ISSUE #7546: Case Sensitive X509 PKI Name Constraint Checking
**Describe the bug**
During the TLS handshake with `verify` set to `verify_peer`, when the peer certificate is evaluated against the directory name constraints of the intermediate CA that issued it, they appear to be compared in a mostly case sensitive manner in the `is_dir_name2` functions in pubkey_cert.erl https://github.com/erlang/otp/blob/87d24fee5b75653899e83c0c3333b4f83c22afd7/lib/public_key/src/pubkey_cert.erl#L721 

I believe this should be a case insensitive check for all attribute types, not just PrintableString as [RFC 5280 Section 7.1](https://datatracker.ietf.org/doc/html/rfc5280#section-7.1) specifies: 
> Conforming implementations MUST support name comparisons using caseIgnoreMatch.

The error that comes back is `{bad_cert, name_not_permitted}`

The peer certificates we are having issues with passes validation using other tools such as `openssl verify` and a NodeJS test program. If I edit the peer certificate in the Erlang debugger to have all lowercase relative distinguished names the connection succeeds.

**To Reproduce**
Issue a client auth certificate that has a subject:
CN = test.program
OU = APPLICATIONS
O = PROJECT
C = GB

From an intermediate CA that includes a directory name constraint:
OU = applications
O = project
C = gb

Attempt a connection to a TLS endpoint with `verify` set to `verify_peer`.

**Expected behavior**
The check succeeds.

**Affected versions**
23.3.2.5

IngelaAndin:
Closes #7546


lib/public_key/src/pubkey_cert.erl
```diff
@@ -676,7 +676,15 @@ is_dir_name2({utf8String, Value1}, String) ->
     is_dir_name2({printableString, unicode:characters_to_list(Value1)}, String);
 is_dir_name2(String, {utf8String, Value1}) ->
     is_dir_name2(String, {printableString, unicode:characters_to_list(Value1)});
-is_dir_name2(_, _) ->
+is_dir_name2(Value1, Value2) when is_list(Value1), is_list(Value2) ->
+    Comp1 = string:to_lower(strip_spaces(Value1)),
```
dgud, line 680: Why not use string:equal/3  with Ignorecase = true, instead of an obsolete to_lower function.
Using equal/3 you don't have to convert the binary to list either.

dgud: If strip space is needed you should use string:casefold/1 instead of to_lower.

IngelaAndin: Sure this code is pretty old so I will look into improving it according to your suggestion :) We probably want to release it on OTP-24 but that should be fine?!  public_key code is older than that.

IngelaAndin: Should be good now with our common efforts :)

kikofernandez (approved):

